Issue #3563, opened 2021-08-18
Project includes a dependancy that has a license that forbids its use


notbasetwo:
Dapr depends on [bouk/monkey](https://github.com/bouk/monkey), which has a [license that forbids anybody to use it](https://github.com/bouk/monkey/blob/master/LICENSE.md).

The license is as follows:
```
Copyright Bouke van der Bijl

I do not give anyone permissions to use this tool for any purpose. Don't use it.

I’m not interested in changing this license. Please don’t ask.
```

The package seems to be used in pkg/components/standalone_loader_test.go.

CodeMonkeyLeet:
Thanks for the heads up!

@yaron2 @artursouza this seems like it was introduced recently by PR https://github.com/dapr/dapr/pull/3158. Should this change be reverted?

daixiang0:
Still need to consider adding a license checker in case this issue occurs again.

yaron2:
> Thanks for the heads up!
> 
> @yaron2 @artursouza this seems like it was introduced recently by PR #3158. Should this change be reverted?

This is only used in the test (and thus not.compiled and distributed in our binaries).

We should not revert the entire PR, just rewrite the test.

yaron2:
Also thanks for reporting @daixiang0

89z:
I mean, you could just use it before the license was added:

https://github.com/bouk/monkey/tree/697673d58359af7ebdcad66f99b81d011359d8ef

seems pretty close

mpldr:
> I mean, you could just use it before the license was added:
> — @89z

Nope, before the License, it's All rights reserved.

eligrey:
If you dynamically link to this repository as a submodule and entirely switch over CI, deployment, and development to Codespaces then this is technically still legal. IANAL & TINLA

Oppen:
The T&C of GH imply you can use it due to the fact it's in a public repository. You would probably need to add it as a submodule just in case, because the terms say the grant applies when accessed via the GH service.

See [clause 5](https://docs.github.com/en/github/site-policy/github-terms-of-service#5-license-grant-to-other-users):
> If you set your pages and repositories to be viewed publicly, **you grant each User of GitHub a nonexclusive, worldwide license to use, display, and perform Your Content through the GitHub Service** and to reproduce Your Content solely on GitHub as permitted through GitHub's functionality (for example, through forking). You may **grant further rights** if you adopt a license.

Emphasis mine.

StoneCypher:
[...]

PR #3570, opened 2021-08-20
remove bou.ke/monkey depencences

daixiang0:
Signed-off-by: Long <long0dai@foxmail.com>

# Description

`loadComponentsFromFile` is equal to `io.ReadFile` + `decodeYaml`, no need to test ` io.ReadFile` as it is standard.

## Issue reference

Please reference the issue this PR will close: #3563

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing

### DIFF
--- a/pkg/components/standalone_loader_test.go
+++ b/pkg/components/standalone_loader_test.go
@@ -1,71 +1,12 @@
 package components
 
 import (
-	"fmt"
-	"io/ioutil"
 	"testing"
 
-	"bou.ke/monkey"
 	"github.com/stretchr/testify/assert"
 
 	config "github.com/dapr/dapr/pkg/config/modes"
 )
-
-func TestLoadComponentsFromFile(t *testing.T) {
-	request := &StandaloneComponents{
-		config: config.StandaloneConfig{
-			ComponentsPath: "test_component_path",
-		},
-	}
-	t.Run("valid yaml content", func(t *testing.T) {
-		filename := "test-component.yaml"
-		yaml := `
-apiVersion: dapr.io/v1alpha1
-kind: Component
-metadata:
-   name: statestore
-spec:
-   type: state.couchbase
-   metadata:
-   - name: prop1
-     value: value1
-   - name: prop2
-     value: value2
-`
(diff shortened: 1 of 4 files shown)


yaron2 (changes requested):
Can we still make sure this functionality is tested? If possible Id rather change the test and not remove it completely.

daixiang0:
After check code, `loadComponentsFromFile` is equal to `io.ReadFile` + `decodeYaml`, no need to test ` io.ReadFile` as it is standard.

pkg/components/standalone_loader_test.go
```diff
@@ -1,72 +1,13 @@
...
-	"io/ioutil"
 	"testing"
 
-	"bou.ke/monkey"
 	"github.com/stretchr/testify/assert"
 
 	config "github.com/dapr/dapr/pkg/config/modes"
 )
 
-func TestLoadComponentsFromFile(t *testing.T) {
```
jjcollinge, line 14: Just a thought, rather than remove the test - would it not be fairly simple to inject an `io.Reader` and change `loadComponentsFromFile` to use that or even just write to a temporary file to keep the test but remove the `bou.ke/monkey` dependency?

jjcollinge: P.S. Didn't realize @yaron2 had already comments something to this effect.

jjcollinge:
> After check code, `loadComponentsFromFile` is equal to `io.ReadFile` + `decodeYaml`, no need to test ` io.ReadFile` as it is standard.

Not to be pedantic but this test is there to ensure that the assumption you have made (that this function only uses `io.ReadFile` + `decodeYaml`) remains the case going forward - if a human PR review misses a change to this file in the future then a bug can be introduced because that assumption has been invalidated.

yaron2 (approved):
I'm merging this to remove the licensing issue.

@jjcollinge if you can submit a PR that rewrites the test that'd be great :)

jjcollinge:
Looks like @daixiang0 beat me to it #3575 🚀

Issue #1090, opened 2020-02-22
Enable automatic encryption of state stores


jeffhollan:
## In what area(s)?

/area runtime

## Describe the feature
Today dapr enables you to store and fetch state and potentially encrypt yourself.  While feasible to encrypt and decrypt before `/state` calls, it's be great if dapr had a way to configure the `state` component with encryption and have it automatically be encrypted and decrypted for me.

yaron2:
/cc @msfussell 

This is a low hanging fruit, can be implemented in the runtime itself (covering all current and future state stores) and encrypt data-at-rest for state stores as well as data-in-transit (mTLS).

amanbha:
Use user provided key from secret store to encrypt the date before saving. Then use they same key to decrypt it before returning it to user on a get cakk.

yaron2:
Triaged into 1.4. /cc @artursouza

rovangju:
I'd be concerned about performance implications of an 'all or nothing' state store; or, I guess having to run two state store configurations to get around that, but ... Would it be worthwhile to entertain the ability to opt-in to an 'encryption enabled' state store via a parameter passed with the SDK state calls?

yaron2:
> I'd be concerned about performance implications of an 'all or nothing' state store; or, I guess having to run two state store configurations to get around that, but ... Would it be worthwhile to entertain the ability to opt-in to an 'encryption enabled' state store via a parameter passed with the SDK state calls?

Encryption will be opt in, not enabled by default, and will be enabled either by configuring the state store component globally or on a per call basis using metadata.

yaron2:
## Design proposal

### Overview

This design proposal aims to capture the following assumptions:

1. The configuration of automatic encryption should be *simple* for both developers and operators
2. Have "best practices" in mind regarding security of the encryption key
3. Automatic encryption should be a non-breaking change and an opt-in feature

#### Operator/Developer Workflow

1. Define a secret containing an AES256 encryption key. The secret can be stored in any Dapr supported secret store
2. Add a metadata key/value pair to a secret component, to explicitly specify the intention of binding the encryption key to the state store
3. Save state using the current, unchanged APIs

### Component schema

To enable automatic data encryption for a state store, the following metadata can be added to any Dapr state store:

```yaml
metadata:
- name: encryptionSecretRef
  value: encryptionKey
```

In order to promote best practices, users will not be able to put an encryption key in plain-text in the `value` field.

### Secret schema

[...]

PR #3589, opened 2021-08-25
add automatic state store encryption

yaron2:
This PR brings automatic state encryption with key rotation support to all Dapr state stores.

Closes https://github.com/dapr/dapr/issues/1090.


pkg/encryption/encryption.go
```diff
@@ -0,0 +1,186 @@
...
+	AES256Algorithm        = "AES256"
+)
+
+// ComponentEncryptionKeys holds the encryption keys set for a component.
+type ComponentEncryptionKeys struct {
+	Primary   Key
+	Secondary Key
+}
+
+// EncryptionKey holds the key to encrypt an arbitrary object.
```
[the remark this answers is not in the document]
yaron2: fixed

pkg/encryption/state.go
```diff
@@ -0,0 +1,78 @@
...
+		enc, err := encrypt(value, keys.Primary, AES256Algorithm)
+		if err != nil {
+			return value, err
+		}
+
+		sEnc := b64.StdEncoding.EncodeToString(enc) + separator + keys.Primary.Name
+		return []byte(sEnc), nil
+	}
+
+	return value, nil
```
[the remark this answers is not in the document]
yaron2: there shouldn't be an error, but after checking the codepaths I realized that map will never return false so removed the check.

pkg/encryption/encryption.go
```diff
@@ -0,0 +1,187 @@
...
+)
+
+type Algorithm string
+
+const (
+	primaryEncryptionKey   = "primaryEncryptionKey"
+	secondaryEncryptionKey = "secondaryEncryptionKey"
+	errPrefix              = "failed to extract encryption key"
+	AES256Algorithm        = "AES256"
+	FeatureName            = "encryption"
```
artursouza, line 29: nit: please, use this enum instead: https://github.com/dapr/dapr/blob/a30488d55fa493f95ed3430539d7a348a96e6e22/pkg/config/configuration.go#L40

artursouza (changes requested):

pkg/encryption/encryption.go
artursouza: Also, name it State.Encryption.

yaron2: That's the first time I see the "centralized" list.. it's mostly an anti-pattern in Go as packages are expected to be self-governed and all encompassing, also test dependencies or 3rd parties wanting to use the feature names now need to pull the dependencies of configuration. I made that change for now but will probably open an issue to disperse these among their respective packages.

artursouza (approved):

Issue #2918, opened 2021-03-12
Dapr lifecycle management improve - start


skyao:
## In what area(s)?

/area runtime

## Background

We, Allbaba Cloud, are trying to use dapr in our production. We found many errors like "dapr unavailable" in our logs so that we want to improve the dapr lifeciecle management between App and daprd.

Because there are so many detais, I split the proposal to three parts: start / stop and running.  

In this issue we will focus on the first part : start.

## user cases when app and daprd start

It's very complicated for start scenario because there are different user case for outbound requests and inbound requests. I will talk about them one by one.

### For outbound Requests

**case1: App sends outbound requests after starts**

Description: When App is started, it perhaps sends some outbound requests imediately. So at this time, the daprd should be able to serve these outbound requests from App. 

Requirement: Daprd should be ready for outbound before App is ready

**case2: App loads configurations during starting**

Description: We are planning to add configuration building block in Dapr. When App is starting, it will load configuration through daprd . So the daprd should be able to serve these outbound requests (loading configurations) from App. 

Requirement: Daprd should be ready for outbound when App is starting.

### For inbound Requests

**case3: Daprd forwards inbound requests to App**

Description: When Daprd is started and subscribed some topic, it will receive inbound requests imediately and forward them to App. So at this time, the App should be able to serve these inbound requests. 

Requirement: App should be ready for inbound before Daprd starts to subscribe and forward inbound requests.

**case4: Daprd gets configuration from App during starting**

Description: During starting, Daprd need to get some dapr related configuration from App like topics to subscribe/input binding to read. So App should be ready to serve dapr's request before Daprd start to do subscribe .

Requirement: App should be ready before Daprd start to get configuration for  inbound.

### Proposal Overview

From above user cases we know that daprd and app need to wait for each other:

- for outbound: Daprd should be ready before app
- for inbound: App should be ready before daprd

So we can't simply let Daprd start first or let App start first. We need to split daprd start process into two part and there are two milestones:  1. Daprd is ready for serve outbound traffic (short for outbound ready) 2. Daprd is ready to subscribe and forward inbound requests (short for inbound ready) . And the process from App start to App ready should between these two part of Daprd.

As shown in following timeline picture, there are three phases for the start of Daprd and App:

![image](https://user-images.githubusercontent.com/1582369/110908217-74c74600-8349-11eb-8832-0a3cb2b41ac5.png)

[...]

PR #3605, opened 2021-09-01
daprd wait app ready at starting phase

theonefx:
# Description

This PR provides an optional capability for daprd to check and wait for the app to complete startup, this ensure that the app is ready to handle incoming traffic before daprd starts receiving it.

there are two optional check point that developer can use:
- url
- k8s containers name

## Issue reference

this is a part of issue : https://github.com/dapr/dapr/issues/2918

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests
* [X] Unit tests passing


pkg/runtime/runtime.go
```diff
@@ -1910,6 +1920,60 @@ func (a *DaprRuntime) blockUntilAppIsReady() {
...
+
+		if err == nil && resp.Status().Code == nethttp.StatusOK {
+			log.Infof("application discovered on readiness address %s", a.runtimeConfig.ReadinessAddress)
+			break
+		}
+
+		time.Sleep(time.Millisecond * 200)
+	}
+
+	log.Infof("wait applicatoin ready, continue init")
```
jjcollinge, line 1974: nit: typo

tests/config/dapr_kafka_bindings_custom_route.yaml
```diff
@@ -12,19 +12,19 @@ spec:
...
-    value: group2
-  # publisher configuration: topic
-  - name: publishTopic
-    value: dapr-test
-  - name: authRequired
-    value: "false"
-    # set specify route
-  - name: route
-    value: /custom-path
+    # Kafka broker connection setting
```
daixiang0, line 15: seems no truly changes here.

theonefx: you are rignt, i will revert it.

theonefx: revert done

pkg/runtime/runtime.go
theonefx: it's a spelling mistake. fixed ^_^

pkedy (commented):
This is an interesting approach! I see a new resource permission is required in the ClusterRoleBinding. Hypothetically, how does the code handle not having this permission?

theonefx:
> This is an interesting approach! I see a new resource permission is required in the ClusterRoleBinding. Hypothetically, how does the code handle not having this permission?

I've tried, and I got an error message with no permissions. So, I have to add the permission definition.

pkg/runtime/wait.go
```diff
@@ -69,3 +75,108 @@ func checkIfOutboundReady(client *http.Client, outboundReadyHealthURL string) er
...
+					}
+				}
+			}
+
+			log.Infof("all containers are ready")
+			break
+		}
+	}
+
+	if a.runtimeConfig.WaitingProbe != "" {
```
artursouza, line 148: Doesn't K8s already have readiness check? I see that this is helpful for Standalone mode but this is too specific - it is on HTTP only (does not support gRPC) and content-type is forced to JSON. Can the waitingProbe (maybe readinessCheck is a better name here) be configured in the Configuration CRD?

Again, this check only adds value to standalone mode.

artursouza (changes requested):
/cc @yaron2

pkg/injector/pod_patch.go
```diff
@@ -593,6 +593,14 @@ func getSidecarContainer(annotations map[string]string, id, daprSidecarImage, im
 				Name:  "NAMESPACE",
 				Value: namespace,
 			},
+			{
+				Name: "POD_NAME",
```
yaron2, line 597: We need to find another way to either pass the pod name to the container or identify the pod in the operator. This uses the FieldPath ref (Downward API) which is not supported on platforms like Virtual Kubelet.

yaron2 (changes requested):

yaron2:
Added comment, looks good otherwise.

daixiang0:
@theonefx please resolve conflicts.

yaron2:
Ping @theonefx

daixiang0:
ping @theonefx

PR #3659, opened 2021-09-13
Is binding subscribed status code tweak

### LINKED ISSUE #3472: Dapr runtime sends InputBinding events to undesignated locations
On startup Dapr sends OPTIONS request to check if the app has already subscribed to an input binding or not, but while sending request if the app is not available for any reason, the sidecar starts to read from all input binding and sends unrelated binding events to the app. 

## In what area(s)?
Input Binding

## Expected Behavior
App should only receive InputBinding event only if it is intended to handle it (corresponding endpoint/subscriber to handle it). 
## Actual Behavior
If the app is not available while being checked from sidecar for the input binding subscriber, the runtime takes the connection error as a positive response and starts reading from binding. 

1- use an InputBinding dapr configuration and an app without having subscriber for the binding
2- Set a delay to answer the option request in the app
3- Start sidecar
4- If option request arrived, shut down the app 
5- Dapr runtime starts to read binding
6- Start the app again
7- Binding event will be sent to the app (wrong location, cause there is no subscriber for the binding)

I did some investigations and found the http_channel converts connection refusal error to 500 error, instead of throwing an exception. 
image

And because not error received here and response is other than StatusNotFound (404), runtime takes it as App is subscribed to the binding.

[runtime.go](https://github.com/dapr/dapr/blob/58ebe431029706bbe9663296b36ec53db385bf01/pkg/runtime/runtime.go#L860)
image

This issue is somehow related to #3376.

pkedy:
# Description

This PR tweaks the HTTP status code check for testing if an application is subscribed to a binding.

In a future release the logic may be expanded to something like:

* Return `true` on `2XX || 405`,
* Return `false` on `4XX`,
* Retry ~5 times otherwise (`5XX` & network error).

## Issue reference

Closes #3472

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests
* [x] Unit tests passing

### DIFF
--- a/pkg/runtime/runtime.go
+++ b/pkg/runtime/runtime.go
@@ -976,7 +976,12 @@ func (a *DaprRuntime) isAppSubscribedToBinding(binding string) bool {
 		// TODO: Propagate Context
 		ctx := context.Background()
 		resp, err := a.appChannel.InvokeMethod(ctx, req)
-		return err == nil && resp.Status().Code != nethttp.StatusNotFound
+		if err != nil {
+			return false
+		}
+		code := resp.Status().Code
+
+		return code/100 == 2 || code == nethttp.StatusMethodNotAllowed
 	}
 	return false
 }
--- a/pkg/runtime/runtime.go
+++ b/pkg/runtime/runtime.go
@@ -977,7 +977,7 @@ func (a *DaprRuntime) isAppSubscribedToBinding(binding string) bool {
 		ctx := context.Background()
 		resp, err := a.appChannel.InvokeMethod(ctx, req)
 		if err != nil {
-			return false
+			log.Fatalf("could not invoke OPTIONS method on input binding subscription endpoint %q: %w", path, err)
 		}
 		code := resp.Status().Code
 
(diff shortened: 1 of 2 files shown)


pkg/runtime/runtime.go
```diff
@@ -976,7 +976,12 @@ func (a *DaprRuntime) isAppSubscribedToBinding(binding string) bool {
 		// TODO: Propagate Context
 		ctx := context.Background()
 		resp, err := a.appChannel.InvokeMethod(ctx, req)
-		return err == nil && resp.Status().Code != nethttp.StatusNotFound
+		if err != nil {
+			return false
```
artursouza, line 980: Where is the fatal log?

artursouza (changes requested):

pkg/runtime/runtime.go
pkedy: I'll add it here, though I'm afraid this will bite us somehow.

pkedy: It's also worth noting that this unlikely return `err != nil` because `parseChannelResponse` in `http_channel.go` returns them as `500` instead. Would you also like it for `500` to `log.Fatal`?

FYI: I created issue #3660 to capture what users on Discord have reported.

artursouza (approved):

yaron2 (approved):
lgtm

Issue #2940, opened 2021-03-16
Dapr sidecar should only check the health of components used by the service it is injected into when starting.


kamesh1677:
> /area runtime
Kubernetes env.
> /area operator
> /area placement
> /area docs
> /area test-and-release

## What version of Dapr?

Dapr 1.0 ,Dapr 0.11 

## Expected Behavior

Dapr sidecar  should run based on respective/assigned  component. 

## Actual Behavior

Scenario :  Deployed an Dapr Microservice with State-management using Redis . Able to store the data no issues here.  

Deployed another Microservice with pub-sub using Kafka able to  publish and subscribe the messages .

 After some time VM got down where Kafka installed ,By that time User restarted the service which is using  Stage-management  then it is observed that dapr side car is not coming up. 

When user check the logs found that due to Kafka component problem .

So The service which is not using Kafka component should be up and running without depending on independent component Health status.

## Steps to Reproduce the Problem

Deploy any service which uses State management with Redis..

Deploy another service which uses Pub-sub with Kafka.

Make Kafka down. 

Restart the State management service . Then issue occurred. 

## Release Note

RELEASE NOTE:

artursouza:
I see this as a feature request for App to announce to sidecar which components are required. Right now, Dapr does not know which components the app will actually use or not.

In the meantime, you might use the ignoreErrors flag in the component that can be ignored but that is a global setting: https://docs.dapr.io/operations/components/component-schema/#fields

theonefx:
My thoughts on this issue are as follows:
1. Each Microservice application use it's own module config. In your case, the daprd of state-manager application will not load the pub/sub module at all. @kamesh1677 
2. Provides a mechanism for sending messages from applications to daprd to decide which modules to start.

---
In the suggestion 2, we can  enhance the “runtime.loadAppConfiguration()” method, to let daprd got which component the app want to use.

theonefx:
As my suggestion 2, i submit a PR https://github.com/dapr/dapr/pull/3688

olitomlinson:
This can be solved by using `scopes` on the PubSub and State-Management components, so that they are only loaded by the Apps that care about them.

In OPs example, if `scopes` were assigned accordingly, Kafka being down wouldn't affect Apps that haven't loaded the Kafka component.

@artursouza this can probably be closed, right?

PR #3688, opened 2021-09-17
Sidecar judge which component can be process by app configuration

theonefx:
# Description

This PR provide an optional way for daprd to judge which module should process. If app provide a module name list, daprd will only init module which in the list.
By default, daprd will process all component which can be loaded.

## Issue reference

Please reference the issue this PR will close: https://github.com/dapr/dapr/issues/2940

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests
* [X] Unit tests passing


pkg/runtime/runtime.go
```diff
@@ -400,7 +401,7 @@ func (a *DaprRuntime) initRuntime(opts *runtimeOpts) error {
 	a.daprHTTPAPI.SetActorRuntime(a.actor)
 	grpcAPI.SetActorRuntime(a.actor)
 
-	// TODO: Remove feature flag once feature is ratified
+	// TODO: Remove feature flag once feature is l
```
daixiang0, line 404: What is it for?

pkg/runtime/runtime.go
```diff
@@ -387,6 +386,8 @@ func (a *DaprRuntime) initRuntime(opts *runtimeOpts) error {
 
 	a.loadAppConfiguration()
 
+	go a.processComponents()
```
daixiang0, line 389: What is it for?

theonefx: after get app configuration, then start process components. so runtime can judge which module can be process.

pkg/runtime/runtime.go
theonefx: my mistake, no need change this

yaron2:
@theonefx tests are failing.

theonefx:
> @theonefx tests are failing.

I'll fix it right away

theonefx:
@yaron2 @daixiang0 the e2e test has passed

artursouza:
Please, validate this with E2E tests first. Overall, I am concerned about this feature changing the order of sidecar initialization. 

How is this different than https://docs.dapr.io/operations/components/component-scopes/ ?

artursouza (changes requested):
Please, compare advantages of this vs https://docs.dapr.io/operations/components/component-scopes/ ?

theonefx:
> Please, validate this with E2E tests first. Overall, I am concerned about this feature changing the order of sidecar initialization.
> 
> How is this different than https://docs.dapr.io/operations/components/component-scopes/ ?


There is a key difference: "Who decides which components can be loaded" :
1. Namespace-based solutions defer this decision until the deployment phase;
2. AppConfig-based solution, this is decided at the development phase;

The difference is that developers or operations maintain component data;

If AppConfig-based solution is supported, the following scenario can be implemented:
Operations can maintain a large component pool, and application developers can determine which components are loaded just by looking at component names.

Of course, AppConfig-based solution is an optional feature, and Namespace-based is valid whether or not developers configure it;

artursouza:
> > Please, validate this with E2E tests first. Overall, I am concerned about this feature changing the order of sidecar initialization.
> > How is this different than https://docs.dapr.io/operations/components/component-scopes/ ?
> 
> There is a key difference: "Who decides which components can be loaded" :
> 
> 1. Namespace-based solutions defer this decision until the deployment phase;
> 2. AppConfig-based solution, this is decided at the development phase;
> 
> The difference is that developers or operations maintain component data;
> 
> If AppConfig-based solution is supported, the following scenario can be implemented: Operations can maintain a large component pool, and application developers can determine which components are loaded just by looking at component names.
> 
> Of course, AppConfig-based solution is an optional feature, and Namespace-based is valid whether or not developers configure it;


My main question is about the component scope section. What is the added value of this changed compared to the existing filter we have already implemented here: https://github.com/dapr/dapr/blob/0ba79334e6fe81e2a0539b70abfe2b696beef05f/pkg/runtime/runtime.go#L1688

In summary, this feature is already present in runtime. Unless I am missing the context here.

berndverst:
Let me summarize my understanding:

The current approach lets the operator / devops person define which Dapr components exist in a certain namespace and which applications have access to the component.

But additionally an application developer may want to improve the performance of their application (e.g. limit the memory footprint of the sidecar, remove components that are slow to initialize and not needed etc) from the sidecar that connects to their application.

The request here is to **additionally instruct the sidecar only to initialize certain components, even if the operator allowed the application to have access to more components**.

I can see this being desirable in organizations where developers cannot manage / modify component definitions themselves.

PR #3752, opened 2021-10-11
bugfix/operator: fix resource leak for grpc server connection

1046102779:
Currently operator server grpc connection's updateChan nerver be closed and cause resource leak.

and I provide a method to close updateChan gracefully that avoid conncurrency panic. 

codes.Unavailable  =  "rpc error: code = Unavailable desc = transport is closing"

image

### DIFF
--- a/pkg/runtime/options.go
+++ b/pkg/runtime/options.go
@@ -19,6 +19,8 @@ type (
 		inputBindings   []bindings.InputBinding
 		outputBindings  []bindings.OutputBinding
 		httpMiddleware  []http.Middleware
+
+		componentsCallback ComponentsCallback
 	}
 
 	// Option is a function that customizes the runtime.
@@ -71,5 +73,12 @@ func WithOutputBindings(outputBindings ...bindings.OutputBinding) Option {
 func WithHTTPMiddleware(httpMiddleware ...http.Middleware) Option {
 	return func(o *runtimeOpts) {
 		o.httpMiddleware = append(o.httpMiddleware, httpMiddleware...)
+	}
+}
+
+// WithComponentsCallback sets the components callback for applications that embed Dapr.
+func WithComponentsCallback(componentsCallback ComponentsCallback) Option {
+	return func(o *runtimeOpts) {
+		o.componentsCallback = componentsCallback
 	}
 }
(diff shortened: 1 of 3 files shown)


1046102779:
@artursouza  chWrapper.Close() will close updateChan, only the location actively exit grpc server connection.

darren-wang:
I met this resource leak scenerio in test, which version of operator has fixed this bug?

yaron2:
> I met this resource leak scenerio in test too, which version of operator has fixed this bug?

1.5

darren-wang:
@1046102779 @yaron2 @artursouza @daixiang0 
I read codes to find that with this fix, the channels are closed when new components added or existing components updated.

This is better than the version before since there was no cleaning before and in out tests, the memory used by operator grows all the time. 

In out tests, we restart pods every 30 minutes in 48 hours to test the stablity of sidecar-injector and operator, with old sidecars stopped and new sidecars started, I can see that operator didn't clean the resources used by sidecars no long existing.

This fix mitigates the resources leakage by closing unavailabe channels when components added or updated. 

However, what if there is no component creation or update, will operator clean unavailable channels?

In production scenarios, components are rather stable when configured, while pods come and go. When componnets change little while pods creation frequently, operators still has leakage.  

I think operators should scan channels periodically to remove unavailable ones, that's the way to solve the resource leakage finally.

1046102779:
> @1046102779 @yaron2 @artursouza @daixiang0 I read codes to find that with this fix, the channels are closed when new components added or existing components updated.
> 
> This is better than the version before since there was no cleaning before and in out tests using dapr-1.4.2, the memory used by operator grows all the time.
> 
> In out tests, we restart pods every 30 minutes in 48 hours to test the stablity of sidecar-injector and operator, with old sidecars stopped and new sidecars started, I can see that operator didn't clean the resources used by sidecars no long existing.
> 
> This fix mitigates the resources leakage by closing unavailabe channels when components added or updated.
> 
> However, what if there is no component creation or update, will operator clean unavailable channels?
> 
> In production scenarios, components are rather stable when configured, while pods come and go. When componnets change little while pods start frequently, operators still has leakage.
> 
> I think operators should scan channels periodically to remove unavailable ones, that's the way to solve the resource leakage finally.

OK, I will look at it immediately.

1046102779:
@darren-wang Wait for MR to CR and pass, the next version should be able to solve this problem.

PR #3821, opened 2021-10-28
injector: add openssl

daixiang0:
Signed-off-by: Long <long0dai@foxmail.com>

# Description

It is only a POC for Dapr to use OpenSSL via CGO, the sidecar injector is a single component and is used frequently in the cases that lots of services start and stop.

Perf tests increase by about 10% at local, @artursouza @yaron2 please call perf tests with GH to reproduce it as I do not have permission.

## Issue reference

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,7 +14,7 @@ export GOSUMDB ?= sum.golang.org
 GIT_COMMIT  = $(shell git rev-list -1 HEAD)
 GIT_VERSION = $(shell git describe --always --abbrev=7 --dirty)
 # By default, disable CGO_ENABLED. See the details on https://golang.org/cmd/cgo
-CGO         ?= 0
+CGO         ?= 1
 BINARIES    ?= daprd placement operator injector sentry
 HA_MODE     ?= false
 # Force in-memory log for placement
--- a/go.mod
+++ b/go.mod
@@ -28,6 +28,7 @@ require (
 	github.com/hashicorp/raft-boltdb v0.0.0-20171010151810-6e5ba93211ea
 	github.com/json-iterator/go v1.1.11
 	github.com/kelseyhightower/envconfig v1.4.0
+	github.com/libp2p/go-openssl v0.0.7
 	github.com/minio/blake2b-simd v0.0.0-20160723061019-3f5f724cb5b1
 	github.com/mitchellh/mapstructure v1.4.1
 	github.com/openzipkin/zipkin-go v0.2.2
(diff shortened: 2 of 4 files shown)


artursouza:
Does not build on ARM.

daixiang0:
@artursouza @yaron2 still need all CI pass then we can run perf test? I just want to show how we can get benefit from CGO with OpenSSL, so it is not a complete PR, could we run perf tests directly?

If not, I will make it complete.

wcs1only:
Can you help us understand why injector performance is worth introducing a CGO dependency? Even on the busiest clusters, pod launch rate is very likely to be <1/second. Our own perf test suite only launches 11 containers over roughly 30 minutes of the test run. I'm not sure how a 10% performance improvement on those 11 requests could even be visible?

Also, the KinD tests are failing, which means perf tests aren't likely to be able to run.

daixiang0:
@wcs1only I would fix KinD issues. It is to show we can introduce CGO to optimize SSL related, a try indeed. It does not very obvious like calling something like SHA256 but works in some cases. In Spike/Big promotion case, some clusters need to inject 100+/s pods in responding to big traffic.

wcs1only:
Even at a burst of 100/s of pods/s, the injector TLS performance is not at all likely to be a scaling factor. You are free to fix this PR enough that we can run the perf tests against it, but we'll need to prove measurable benefit to actual workloads before accepting a CGO dependency like this.

yaron2:
> @wcs1only I would fix KinD issues. It is to show we can introduce CGO to optimize SSL related, a try indeed, not for merge. It is not very obvious like calling something like SHA256 but works in some cases. In Spike/Big promotion case, some clusters need to inject 100+/s pods in responding to big traffic.

@daixiang0 the sidecar injector can be scaled out horizontally, as the k8s API server uses a round-robin load balancing approach for the Injector's k8s service. This is unlike the placement service, for example, that can only have one *active* instance at all times. So according to this, I'm not sure the CGO drawbacks are worth the single instance performance benefits. What do you think?

daixiang0:
Yes, it is not a significant factor I think, maybe I need to choose a better one for POC.

daixiang0:
@yaron2 Accelerating the connection to the k8s API server will make the sidecar creation requests get faster response, which will be particularly obvious in large-scale clusters. Although k8s API server uses a round-robin load balancing approach, accelerating each connection response still makes sense.

@beiwei30 WDYT?

PR #3855, opened 2021-11-05
Fix when PartitionCount is configured as 0, the index of the array is illegal

CrazyHZM:
# Description

Fix when `PartitionCount` is configured as 0, the index of the array is illegal
Handling some exceptions
Fix typo

## Issue reference

Fix https://github.com/dapr/dapr/issues/3899

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing


pkg/actors/actors.go
```diff
@@ -904,7 +918,7 @@ func (m *ActorMetadata) removeReminderFromPartition(reminderRefs []actorReminder
 }
 
 func (m *ActorMetadata) insertReminderInPartition(reminderRefs []actorReminderReference, reminder *Reminder) ([]Reminder, actorReminderReference, string, *string) {
-	newReminderRef := m.createReminderReference(reminder)
+	newReminderRef, _ := m.createReminderReference(reminder)
```
daixiang0, line 921: Could we skip this error safely?

CrazyHZM: > Could we skip this error safely?

I think this cannot be ignored, it has been modified.

yaron2:
Thanks @CrazyHZM. Can you please create an issue for this and reference it here?

/cc @artursouza

CrazyHZM:
> Thanks @CrazyHZM. Can you please create an issue for this and reference it here?
> 
> /cc @artursouza

Done.

daixiang0:
Could we have a ut for this case?

daixiang0:
@CrazyHZM please fix DCO issue.

CrazyHZM:
@daixiang0 
Because there are several commits that take too long, the rebase operation to solve the singoff will cause the PR to be not clean enough, I re-submitted the PR:
https://github.com/dapr/dapr/pull/4100

PR #3856, opened 2021-11-05
optimize/tests: add minikube ip

1046102779:
We don’t need to pass the `DAPR_TEST_MINIKUBE_IP` environment variable。

```shell
minikube ip
```


daixiang0:
Please use `make describe-minikube-env` as the doc says. Also we support both KinD and minikube, so it is not good enough.

1046102779:
> Please use `make describe-minikube-env` as the doc says. Also we support both KinD and minikube, so it is not good enough.

ok, i close it.

PR #3858, opened 2021-11-05
bugfix/e2e: allowlist failed  when mtls is in closed state

1046102779:
If DAPR_MTLS_ENABLED=false, the allowlist test fails. Because allowlist server depends on SPIFFE, and it is the encrypted data in the mtls. So when mtls is closed, the allowlist server will report the following error:

image

Because SPIFFE is nil,  and  server's appid, namespace and trustdomain  are empty
image


daixiang0 (approved):
@artursouza  Looks that we do not test this case in e2e :(

1046102779:
> 

@artursouza +1

1046102779:
@yaron2 @daixiang0 +1

PR #3863, opened 2021-11-08
feature/e2e: add env DAPR_E2E_TEST to pick many tests

1046102779:
Add env `DAPR_E2E_TESTS` to pick many e2e tests, and use `make test-e2e-selected` to run.

**make test-e2e-selected**

for example:

`export DAPR_E2E_TESTS=allowlist hellodapr`


tests/dapr_tests.mk
```diff
@@ -209,6 +209,12 @@ test-e2e-all: check-e2e-env test-deps
 	# Note2: use env variable DAPR_E2E_TEST to pick one e2e test to run.
```
daixiang0, line 209: We can use this one.

1046102779: Yes, but the env param `DAPR_E2E_TEST ` only picks all e2e apps or one e2e app, it can't pick multiple e2e apps at once.

daixiang0: A target for one or all is common, the rest often write in their own script.

tests/dapr_tests.mk
```diff
@@ -221,6 +221,12 @@ test-e2e-all: check-e2e-env test-deps
...
 	DAPR_CONTAINER_LOG_PATH=$(DAPR_CONTAINER_LOG_PATH) GOOS=$(TARGET_OS_LOCAL) DAPR_TEST_NAMESPACE=$(DAPR_TEST_NAMESPACE) DAPR_TEST_TAG=$(DAPR_TEST_TAG) DAPR_TEST_REGISTRY=$(DAPR_TEST_REGISTRY) DAPR_TEST_MINIKUBE_IP=$(MINIKUBE_NODE_IP) gotestsum --jsonfile $(TEST_OUTPUT_FILE_PREFIX)_e2e.json --junitfile $(TEST_OUTPUT_FILE_PREFIX)_e2e.xml --format standard-quiet -- -p 2 -count=1 -v -tags=e2e ./tests/e2e/$(DAPR_E2E_TEST)/...
 
+# start selected e2e tests.
+test-e2e-selected: check-e2e-env test-deps
```
artursouza, line 225: This can be super useful used in the existing target. This allows us to change the tests to run and still use the same workflow file in GitHub. Also, it would be great to keep it compatible with the existing DAPR_E2E_TEST. I think this would mean adding "if" statements in the existing test e2e target.

artursouza (changes requested):

tests/dapr_tests.mk
1046102779: fixed.

1046102779:
@artursouza +1

1046102779:
@yaron2 @daixiang0 +1

artursouza (approved):

artursouza:
@1046102779 Please, commit with `-s` for DCO to pass. You might want to squash it all into a single commit, so you don't need to `-s` each one.

Issue #3788, opened 2021-10-20
Setting global.ha.enabled: true produces deployment yaml with 2 affinity elements


andrey-dubnik:
We are using the FluxCD and it does not like the fact helm produces duplicates of the "affinity:" for the following templates where HA is enabled:

dapr/charts/dapr_sentry/templates/dapr_sentry_deployment.yaml
dapr/charts/dapr_operator/templates/dapr_operator_deployment.yaml
dapr/charts/dapr_sidecar_injector/templates/dapr_sidecar_injector_deployment.yaml

It may sound like FluxCD being picky and it seem k8s is forgiving when it comes to duplicates where it may be something to fix in the helm so the definition is consistent and there is only one affinity block produced? 

## In what area(s)?

> /area helm

## What version of Dapr?

> 1.4.3

## Expected Behavior

Helm to produce templates with a single affinity block

## Actual Behavior

Helm fails to apply via FluxCD cause duplicate affinity causes validation to fail

artursouza:
Reopening because test to upgrade in HA mode failed in 3-node cluster during release activities for 1.5. This issue needs to be revisited.

andrey-dubnik:
What was the error?

artursouza:
It gets stuck in pending due to the affinity rules.

andrey-dubnik:
I guess the solution can be to switch to soft rules, which won't guarantee HA as with soft rules pods can still consolidate in non-HA way...

Does test system runs with a single node?

artursouza:
It is a 3 node cluster.

andrey-dubnik:
Changed to use the soft anti-affinity, which should not block the upgrade

ewassef:
We have had to turn off HA in all of our clusters until this is resolved

artursouza:
PR is merged. Closing.

PR #3868, opened 2021-11-08
use soft anti-affinity rules

andrey-dubnik:
#3788


andrey-dubnik:
There seem to be a glitch in the testing pipeline....

artursouza (approved):

Taction:
@andrey-dubnik Thanks for this PR! Would you please fix DCO issue?

andrey-dubnik:
> @andrey-dubnik Thanks for this PR! Would you please fix DCO issue? You can refer to [here](https://github.com/src-d/guide/blob/master/developer-community/fix-DCO.md) about how to fix it.

@Taction I can see the commit DCO complains about is signed... do you see any possible cause why DCO complains about it?

yaron2:
> 

See here: https://docs.dapr.io/contributing/contributing-overview/#i-didnt-sign-my-commit-now-what

Taction:
@andrey-dubnik You do have pushed a signed commit with hash `cda2ae0`, but your first commit with hash `2e30320` is still there with unsigned status. The expected behaviour should be to replace the first commit rather than append a new commit.  
Besides, if it is too confusing for you to solve this, feel free to open another pr and copy your changes to that.

andrey-dubnik:
By the looks of it the PR does not actually have all the necessary changes, I will create a new one

PR #3872, opened 2021-11-08
Add darwin arm64 binaries

berndverst:
# Description

Publishes darwin arm64 binaries of the dapr runtime

Merge this after #3871

### DIFF
--- a/.github/workflows/dapr.yml
+++ b/.github/workflows/dapr.yml
@@ -46,8 +46,6 @@ jobs:
             target_arch: arm64
           - os: macOS-latest
             target_arch: arm
-          - os: macOS-latest
-            target_arch: arm64
     steps:
       - name: Set up Go ${{ env.GOVER }}
         uses: actions/setup-go@v2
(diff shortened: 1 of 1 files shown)


.github/workflows/dapr.yml
```diff
@@ -204,6 +202,11 @@ jobs:
...
           name: dapr_darwin_amd64
           path: ${{ env.ARTIFACT_DIR }}
+      - name: download artifacts - dapr_darwin_arm64
+        uses: actions/download-artifact@master
+        with:
+          name: dapr_darwin_arm64
+          path: ${{ env.ARTIFACT_DIR }}
       - name: download artifacts - dapr_grafana_dashboards
         uses: actions/download-artifact@master
         with:
```
artursouza, line 212: You might need to add to the "tar release binaries" step too.

artursouza (changes requested):

.github/workflows/dapr.yml
berndverst: Linux only

artursouza (approved):

Issue #3906, opened 2021-11-16
Unsubscribe of configuration API.


LaurenceLiZhixin:
I'm working on configuration API of multi-language sdks https://github.com/dapr/go-sdk/pull/200, and found subscription problems of configuration api https://github.com/dapr/dapr/pull/3903 . Now the implement of subscribe API is server-side-grpc-stream, and once the subscription is called, the return statement of server side causes the stream ends, and user's client can't get update event from config state store.
```proto
 rpc SubscribeConfigurationAlpha1(SubscribeConfigurationRequest) returns (stream SubscribeConfigurationResponse) {}
```
I would fix this bug, and encountered a new problem. After subscription function blocking, it would never be restart. How can user unsubscribe (stop) the blocking go routine if his app is shutdown, or the grpc connection broken? I think there is two way to solve, one is to add a new api called unsubscribe, the second is change the server-side-grpc-stream to a full duplex stream, and dapr runtime can know if the subscription should be shutdown, or the streaming is broken, and return the broking function.
@yaron2 @artursouza @daixiang0

yaron2:
I would always encourage the explicit (and in this case, also non-breaking) API design. So, I support adding an Unsubscribe method.

LaurenceLiZhixin:
OK, I'll follow up.

greenie-msft:
Doc PR for this: https://github.com/dapr/docs/pull/2010

artursouza:
Moving to 1.7 since PR is still in progress.

LaurenceLiZhixin:
My colleague Lan Luo raised a problem with the existing design. My current design is based on the config's key granularity model. For example, if there are multiple subscribe operations of the configuration of key "myKey", one unsubscribe operation for "myKey" will impact all previous subscribe operations.
But with a granularity of the whole subscribe operation, each call to subscribe returns a specific ID that the user needs to pass when unsubscription, thus terminating only this subscribe without any other Subscribe Goroutine being affected.

My current scheme requires the user to guarantee that the key can be unsubscribed only if the user really does not need to subscribe to it. And the new idea that unsubscribe by ID is more flexible I think, and the user needs to record each subscription ID. 

Yaron, what do you think about? I will follow up . @yaron2

LaurenceLiZhixin:
[...]

PR #3927, opened 2021-11-19
Add configuration unsubscribe api.

LaurenceLiZhixin:
# Description

Add configuration unsubscribe api of dapr runtime. As subscribe api is a gRPC stream api, and never return until client call unsubscribe of the target 'store name + keys' subscription.

## Issue reference
#3906 

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests
* [X] Unit tests passing
* [X] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_
* [X] Specification has been updated / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_
* [X] Provided sample for the feature / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_


pkg/grpc/api.go
```diff
@@ -1296,12 +1290,42 @@ func (a *api) SubscribeConfigurationAlpha1(request *runtimev1pb.SubscribeConfigu
...
+	return nil
+}
+
+func (a *api) UnSubscribeConfigurationAlpha1(ctx context.Context, request *runtimev1pb.UnSubscribeConfigurationRequest) (*runtimev1pb.UnSubscribeConfigurationResponse, error) {
+	sort.Slice(request.Keys, func(i, j int) bool {
+		return request.Keys[i] < request.Keys[j]
+	})
+
+	a.configurationSubscribeLock.Lock()
+	a.configurationSubscribeLock.Unlock()
```
daixiang0, line 1312: ?

LaurenceLiZhixin: Fixed

LaurenceLiZhixin:
I fixed ut, please take a look. @daixiang0  @yaron2

yaron2:
Thanks @LaurenceLiZhixin! can you please update the docs section also with this method?

LaurenceLiZhixin:
> method

Fine!

LaurenceLiZhixin:
I updated the docs of this method: https://github.com/dapr/docs/pull/2010

pkg/grpc/api.go
```diff
@@ -1309,12 +1302,46 @@ func (a *api) SubscribeConfigurationAlpha1(request *runtimev1pb.SubscribeConfigu
...
+	// TODO(@laurence) deal with failed subscription and retires
+	err = store.Subscribe(ctx, req, handler.updateEventHandler)
+	if err != nil {
+		err = status.Errorf(codes.InvalidArgument, messages.ErrConfigurationSubscribe, req.Keys, request.StoreName, err.Error())
+		apiServerLogger.Debug(err)
+		return err
 	}
+	stop := make(chan struct{})
+	a.configurationSubscribeLock.Lock()
+	a.configurationSubscribe[getConfigSubscribeUniqueKey(request.GetStoreName(), request.GetKeys())] = stop
```
mukundansundar, line 1317: isn't this changing the unique map key pattern? 
If my understanding is correct this is changing the map key to "storename||key1 key2 key3" etc. since request.GetKeys() is a slice. 
If that is the scenario, there will be no option to subscribe and unsubscribe to individual keys right ?

Previously it was "storeName||key1", "storeName||key2" etc ...

LaurenceLiZhixin: This is worth considering I think. Considering that we start subscribing by a GRPC stream request, in which multiple keys may be subscribed at the same time, and subscribing and unsubscribing are the entire GRPC Stream dimension, So here will be the key design for "storename | | key1 key2 key3" pattern
@mukundansundar

yaron2: @LaurenceLiZhixin it might be confusing for users to know they need to include all the keys they want to subscribe/unsubscribe to. Since we don't expect big slices (ie. thousands) here, can we change the key to only contain the store name and then keep and have a data structure that encapsulates the keys as the value?

LaurenceLiZhixin: If the key is storeName so, I think the unsubscribe API can only unsubscribe specific storeName's subscription. And `UnSubscribeConfigurationRequest` contains only storeName fields. After user calling Unsubscribe API, all blocked grpc-stream of the storeName's subscription are released.

In this way, we need to maintain two maps, one to hold the map of storeName to encapsulated keys data structure and one to hold the stop channel of this storeName that grpc-stream subscribing to.
@yaron2

yaron2: > In this way, we need to maintain two maps, one to hold the map of storeName to encapsulated keys data structure and one to hold the stop channel of this storeName that grpc-stream subscribing to.

Currently, the proto API in this PR allows users to unsubscribe from individual keys, because you have a store name and a list of keys. We need to ensure this is doable. This can be either done with two maps or by keeping the key as `storename||keyX`, where `keyX` is a single key.

dapr/proto/runtime/v1/dapr.proto
```diff
@@ -89,6 +89,9 @@ service Dapr {
   // SubscribeConfiguration gets configuration from configuration store and subscribe the updates event by grpc stream
   rpc SubscribeConfigurationAlpha1(SubscribeConfigurationRequest) returns (stream SubscribeConfigurationResponse) {}
 
+  // UnSubscribeConfiguration unsubscribe the subscription of configuration
+  rpc UnSubscribeConfigurationAlpha1(UnSubscribeConfigurationRequest) returns (UnSubscribeConfigurationResponse) {}
```
jjcollinge, line 93: nit: UnSubscribe should be Unsubscribe

LaurenceLiZhixin:
The continue work of this pr is moved to https://github.com/dapr/dapr/pull/4130#partial-pull-merging , as this pr is containing many conflicts when i fix dco.

PR #3935, opened 2021-11-23
Ignore go.work for Go 1.18 Workspaces

berndverst:
Go 1.18 introduces Go workspaces. These can be used to pin specific go versions or maybe locally built library versions / dependencies.
See `https://github.com/golang/go/issues/45713`

This avoids the need to constantly update go.mod for local development. For example I commonly use replace to refer to components-contrib and dapr on my local file system instead of the remote.


daixiang0 (approved):

daixiang0:
Flaky test still exists :(

berndverst:
@artursouza can you merge? This won't have any negative side effect. Just adding a file to gitignore which does not exist today, but will exist in the future and should never be checked in.

mukundansundar (approved):

artursouza (approved):

PR #3952, opened 2021-11-25
Add namespace check for component updates

yaron2:
This PR adds a check when a new component is updated. The check makes sure the component updated matches the namespace of the sidecar that registers for updates before attempting to process the component, which will lead to unnecessary goroutines and warning messages on the operator log.


pkg/operator/api/api.go
```diff
@@ -221,8 +221,12 @@ func (a *apiServer) ComponentUpdate(in *operatorv1pb.ComponentUpdateRequest, srv
 	chWrapper := initChanGracefully(updateChan)
 
 	for c := range updateChan {
+		if c.Namespace != in.Namespace {
+			continue
```
daixiang0, line 225: Could add test cases for those?

yaron2: Done.

daixiang0:
@yaron2 please resolve conflicts.

yaron2:
> @yaron2 please resolve conflicts.

Resolved.

pkg/operator/api/api.go
```diff
@@ -220,6 +220,10 @@ func (a *apiServer) ComponentUpdate(in *operatorv1pb.ComponentUpdateRequest, srv
 	}()
 	chWrapper := initChanGracefully(updateChan)
 	updateComponentFunc := func(c *componentsapi.Component) {
+		if c.Namespace != in.Namespace {
+			return
```
daixiang0, line 224: How about adding a warn here then admin can find invalid calls?

yaron2: No, because this is a valid operation. It's an internal decision of the operator to not update sidecars from a different namespace, but it's not an invalid operation.

daixiang0 (approved):
lgtm

PR #4013, opened 2021-12-09
Add breaking change for PostgreSQL State Store

berndverst:
# Description

Adds breaking change information for PostgreSQL State Store (Alpha) component.

### DIFF
--- a/docs/release_notes/v1.5.0.md
+++ b/docs/release_notes/v1.5.0.md
@@ -278,4 +278,12 @@ kubectl rollout restart deploy/<deployment-name>
 
 ## Breaking Changes
 
-None.
+### PostgreSQL State Store component (Alpha)
+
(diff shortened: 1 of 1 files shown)


docs/release_notes/v1.5.0.md
```diff
@@ -278,4 +278,12 @@ kubectl rollout restart deploy/<deployment-name>
...
 
-None.
+### PostgreSQL State Store component (Alpha)
+
+This release of the PostgreSQL State Store component (Alpha) cannot fetch data created with previous versions due to a schema change.
+
+To fetch data created with previous versions please manually update the state store table (default name: `state`) schema.
+
+```sql
+ALTER TABLE state ADD COLUMN isbinary boolean NOT NULL
```
pkedy, line 288: Wondering if this statement might fail because it is `NOT NULL` without a `DEFAULT false`.

PR #4021, opened 2021-12-13
Fix actor reminder e2e tests + fixes to reminder partition

artursouza:
# Description

Fixes the reminder E2E tests.
Fixes to the reminder storage solution with partitioning.

## Issue reference

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing

### DIFF
--- a/pkg/actors/actors.go
+++ b/pkg/actors/actors.go
@@ -454,6 +454,9 @@ func (a *actorsRuntime) GetState(ctx context.Context, req *GetStateRequest) (*St
 	resp, err := a.store.Get(&state.GetRequest{
 		Key:      key,
 		Metadata: metadata,
+		Options: state.GetStateOption{
+			Consistency: state.Strong,
+		},
 	})
 	if err != nil {
 		return nil, err
@@ -663,6 +666,9 @@ func (a *actorsRuntime) getReminderTrack(actorKey, name string) (*ReminderTrack,
 
 	resp, err := a.store.Get(&state.GetRequest{
 		Key: constructCompositeKey(actorKey, name),
+		Options: state.GetStateOption{
+			Consistency: state.Strong,
+		},
 	})
 	if err != nil {
 		return nil, err
@@ -688,6 +694,9 @@ func (a *actorsRuntime) updateReminderTrack(actorKey, name string, repetition in
 	err := a.store.Set(&state.SetRequest{
 		Key:   constructCompositeKey(actorKey, name),
 		Value: track,
+		Options: state.SetStateOption{
+			Consistency: state.Strong,
+		},
 	})
 	return err
 }
@@ -1272,6 +1281,9 @@ func (a *actorsRuntime) getActorTypeMetadata(actorType string, migrate bool) (*A
 		metadataKey := constructCompositeKey("actors", actorType, "metadata")
 		resp, err := a.store.Get(&state.GetRequest{
 			Key: metadataKey,
+			Options: state.GetStateOption{
+				Consistency: state.Strong,
+			},
 		})
 		if err != nil {
 			return err
@@ -1416,6 +1428,9 @@ func (a *actorsRuntime) getRemindersForActorType(actorType string, migrate bool)
 			getRequests = append(getRequests, state.GetRequest{
 				Key:      key,
 				Metadata: metadata,
+				Options: state.GetStateOption{
+					Consistency: state.Strong,
+				},
 			})
 		}
 
@@ -1489,6 +1504,9 @@ func (a *actorsRuntime) getRemindersForActorType(actorType string, migrate bool)
 	key := constructCompositeKey("actors", actorType)
 	resp, err := a.store.Get(&state.GetRequest{
 		Key: key,
+		Options: state.GetStateOption{
+			Consistency: state.Strong,
+		},
 	})
 	if err != nil {
 		return nil, nil, err
@@ -1530,6 +1548,10 @@ func (a *actorsRuntime) saveRemindersInPartition(ctx context.Context, stateKey s
 		Value:    reminders,
 		ETag:     etag,
 		Metadata: map[string]string{metadataPartitionKey: databasePartitionKey},
+		Options: state.SetStateOption{
+			Concurrency: state.FirstWrite,
+			Consistency: state.Strong,
+		},
 	})
 }
 
--- a/pkg/actors/actors.go
+++ b/pkg/actors/actors.go
@@ -127,7 +127,7 @@ type ActorRemindersMetadata struct {
 type actorReminderReference struct {
 	actorMetadataID           string
 	actorRemindersPartitionID uint32
-	reminder                  *Reminder
+	reminder                  Reminder
 }
 
 const (
@@ -631,7 +631,7 @@ func (a *actorsRuntime) evaluateReminders() {
 						if !exists {
 							stop := make(chan bool)
 							a.activeReminders.Store(reminderKey, stop)
-							err := a.startReminder(r.reminder, stop)
+							err := a.startReminder(&r.reminder, stop)
 							if err != nil {
 								log.Errorf("error starting reminder: %s", err)
 							} else {
@@ -851,7 +851,7 @@ func (a *actorsRuntime) getReminder(req *CreateReminderRequest) (*Reminder, bool
 
 	for _, r := range reminders {
 		if r.reminder.ActorID == req.ActorID && r.reminder.ActorType == req.ActorType && r.reminder.Name == req.Name {
-			return r.reminder, true
+			return &r.reminder, true
 		}
 	}
 
@@ -870,7 +870,7 @@ func (m *ActorMetadata) calculateReminderPartition(actorID, reminderName string)
 	return (h.Sum32() % uint32(m.RemindersMetadata.PartitionCount)) + 1
 }
 
-func (m *ActorMetadata) createReminderReference(reminder *Reminder) actorReminderReference {
+func (m *ActorMetadata) createReminderReference(reminder Reminder) actorReminderReference {
 	if m.RemindersMetadata.PartitionCount > 0 {
 		return actorReminderReference{
 			actorMetadataID:           m.ID,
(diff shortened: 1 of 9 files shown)


pkg/actors/actors.go
```diff
@@ -1226,7 +1233,7 @@ func (a *actorsRuntime) getActorTypeMetadata(actorType string, migrate bool) (*A
 
 	if !a.actorTypeMetadataEnabled {
 		return &ActorMetadata{
-			ID: uuid.NewString(),
+			ID: metadataZeroID,
```
artursouza, line 1236: This is the default metadata ID when porting the "zero partition" configuration. We should keep the ID the same for idempotency.

pkg/actors/actors.go
```diff
@@ -1235,89 +1242,89 @@ func (a *actorsRuntime) getActorTypeMetadata(actorType string, migrate bool) (*A
...
-		// We fallback to the default "zero" partition behavior.
+	result := ActorMetadata{
+		ID: metadataZeroID,
+		RemindersMetadata: ActorRemindersMetadata{
+			partitionsEtag: nil,
+			PartitionCount: 0,
+		},
+		Etag: nil,
+	}
+	retryErr := backoff.Retry(func() error {
```
artursouza, line 1253: Add a retry logic to read+migrate actor metadata.

pkg/actors/actors.go
```diff
@@ -1235,89 +1242,89 @@ func (a *actorsRuntime) getActorTypeMetadata(actorType string, migrate bool) (*A
...
 	}
 	if refreshedActorMetadata.ID != actorMetadata.ID {
-		return nil, errors.Errorf("could not migrate reminders for actor type %s due to race condition in actor metadata", actorType)
+		return errors.Errorf("could not migrate reminders for actor type %s due to race condition in actor metadata", actorType)
 	}
 
 	// Recreate as a new metadata identifier.
 	actorMetadata.ID = uuid.NewString()
 	actorMetadata.RemindersMetadata.PartitionCount = a.config.RemindersStoragePartitions
+	actorMetadata.Etag = refreshedActorMetadata.Etag
```
artursouza, line 1327: This is needed to avoid race condition in the database.

pkg/actors/actors.go
```diff
@@ -1235,89 +1242,89 @@ func (a *actorsRuntime) getActorTypeMetadata(actorType string, migrate bool) (*A
...
+		return nil
 	}
 
 	if actorMetadata.RemindersMetadata.PartitionCount > a.config.RemindersStoragePartitions {
 		log.Warnf("cannot decrease number of partitions for reminders of actor type %s", actorType)
-		return actorMetadata, nil
+		return nil
 	}
 
+	// Nice to have: avoid conflicting migration within the same process.
```
artursouza, line 1310: Just to highlight the comment in the code.

tests/apps/actorfeatures/app.go
```diff
@@ -308,13 +308,15 @@ func testCallActorHandler(w http.ResponseWriter, r *http.Request) {
...
 
+	expectedHTTPCode := 200
 	var req timerReminderRequest
 	switch callType {
 	case "method":
 		// NO OP
 	case "timers":
 		fallthrough
 	case "reminders":
+		expectedHTTPCode = 204
```
artursouza, line 319: This will silent some unnecessary error in the logs.

tests/apps/actorfeatures/app.go
```diff
@@ -362,6 +364,16 @@ func testCallMetadataHandler(w http.ResponseWriter, r *http.Request) {
 	w.Write(body)
 }
 
+func testEnvHandler(w http.ResponseWriter, r *http.Request) {
```
artursouza, line 367: This can be handy to validate that the app is running with the expected configuration.

pkg/actors/actors.go
```diff
@@ -1216,6 +1219,10 @@ func (a *actorsRuntime) saveActorTypeMetadata(actorType string, actorMetadata *A
 		Key:   metadataKey,
 		Value: actorMetadata,
 		ETag:  actorMetadata.Etag,
+		Options: state.SetStateOption{
```
artursouza, line 1222: This is a must to make sure first-write behavior is respected when etag is nil (write works only if record does not exist).

tests/apps/actorfeatures/app.go
```diff
@@ -172,7 +173,12 @@ func appendLog(actorType string, actorID string, action string, start int) {
 }
 
 func getLogs() []actorLogEntry {
-	return actorLogs
+	actorLogsMutex.Lock()
```
artursouza, line 176: This was another source of instability in our tests. It was hard to find but it made a difference.

pkg/actors/actors.go
```diff
@@ -1330,37 +1369,25 @@ func (a *actorsRuntime) migrateRemindersForActorType(actorType string, actorMeta
...
-				Key:      stateKey,
-				Value:    stateValue,
-				Metadata: metadata,
-			},
-		})
-	}
-	err = a.transactionalStore.Multi(&transaction)
-	if err != nil {
-		return nil, err
+		err = a.saveRemindersInPartition(context.TODO(), stateKey, stateValue, nil, actorMetadata.ID)
```
artursouza, line 1376: No need for transactions here. It simplifies debugging.

pkg/actors/actors.go
```diff
@@ -1409,11 +1442,22 @@ func (a *actorsRuntime) getRemindersForActorType(actorType string, migrate bool)
...
+						return
+					}
+
+					if resp == nil {
+						r.Error = "response not found for partition"
+						return
+					}
+
+					if len(resp.Data) == 0 {
+						r.Error = "data not found for reminder partition"
```
artursouza, line 1454: Every  partition should have data even if empty list.

tests/config/dapr_cosmosdb_query_state.yaml
```diff
@@ -34,5 +34,3 @@ spec:
     secretKeyRef:
       name: cosmosdb-secret
       key: queryCollection
-  - name: actorStateStore
```
artursouza, line 37: This was the root cause of increased failure rate of the reminder E2E tests.

pkg/runtime/runtime.go
```diff
@@ -1184,17 +1185,18 @@ func (a *DaprRuntime) initState(s components_v1alpha1.Component) error {
...
 				a.actorStateStoreName = s.ObjectMeta.Name
+			} else {
+				log.Warnf("ignoring duplicate actor state store: %s", s.ObjectMeta.Name)
 			}
+			a.actorStateStoreLock.Unlock()
 		}
 		diag.DefaultMonitoring.ComponentInitialized(s.Spec.Type)
 	}
 
-	if a.hostingActors() && (a.actorStateStoreName == "" || a.actorStateStoreCount != 1) {
```
artursouza, line 1194: This code is executed before sidecar invokes /dapr/config, which means that a.hostingActors() would return false. This was not always like that but we had many changes in the order of loading things in Dapr sidecar and this was a side effect of that.

yaron2 (approved):
lgtm

tests/config/dapr_cosmosdb_query_state.yaml
wcs1only: So, the root cause was that we were using cosmosdb with actorStateStore = true? In that case, why were the failures intermittent? Should we not have had a 100% failure rate with this kind of misconfiguration?

wcs1only: I think I understand, based on an offline conversation, the actual issue is that we had two actoreStateStores configured and we were racing between which one to use?

tests/e2e/actor_features/actor_features_test.go
```diff
@@ -368,23 +370,30 @@ func TestActorFeatures(t *testing.T) {
...
-		err = tr.Platform.Restart(appName)
-		assert.NoError(t, err)
-		externalURL = tr.Platform.AcquireAppExternalURL(appName)
-		require.NotEmpty(t, externalURL, "Could not get external URL after app restart.")
+		t.Logf("Restarting %s ...", appName)
+		err := tr.Platform.Restart(appName)
+		require.NoError(t, err)
 
-		time.Sleep(sleepTime)
+		time.Sleep(120 * time.Second)
```
wcs1only, line 377: Is this really necessary? Or was this something you were trying while debugging?

tests/runner/kube_testplatform.go
```diff
@@ -296,9 +297,16 @@ func (c *KubeTestPlatform) SetAppEnv(name, key, value string) error {
 		return err
 	}
 
-	_, err := appManager.WaitUntilDeploymentState(appManager.IsDeploymentDone)
+	// 120s is the observed time to spawn a new test app POD and terminate the old one.
+	// This sleep time is important for the stream log to work.
+	time.Sleep(120 * time.Second)
```
wcs1only, line 302: This seems extremely high, and we're already cutting close the 10 minutes limit for some of our test suites. The line below this one should gate this appropriately. If deployment state is running, your logs should stream properly. If you are seeing errors, it might be because the StreamContainerLogs() attempts to run against all pods in a deployment, and might try against a terminated pod, I think? The new pod should still work, even in the presence of this error.

tests/runner/kube_testplatform.go
```diff
@@ -316,6 +324,10 @@ func (c *KubeTestPlatform) Restart(name string) error {
 		return err
 	}
 
+	if _, err := m.WaitUntilDeploymentState(m.IsDeploymentDone); err != nil {
```
wcs1only, line 327: Isn't this a dupe of line 286 from the Scale call above?

tests/e2e/actor_features/actor_features_test.go
artursouza: I don't think this is necessary anymore.

tests/config/dapr_cosmosdb_query_state.yaml
artursouza: Correct.

tests/runner/kube_testplatform.go
artursouza: This one is needed because it is the time I observed for the new POD to be created and the old one to be terminated. Otherwise, tests can fail because the requests can go to any of the PODs and add flakiness.

tests/runner/kube_testplatform.go
artursouza: Yes. let me remove this.

tests/runner/kube_testplatform.go
wcs1only: So WaitUntilDeploymentState doesn't sufficiently gate on the deployment being completed? Could we just update WaitUntilDeploymentState to also gate on the number of replicas being what we expect?

artursouza: I checked the code. It does that already. Let me remove this. Nice catch.

wcs1only (approved):
Thanks for removing the sleeps. LGTM

Issue #3995, opened 2021-12-05
Reminders and Timer ISO-8601 off-by-one error


withinboredom:
## In what area(s)?

/area runtime

> /area operator

> /area placement

> /area docs

> /area test-and-release

## What version of Dapr?

1.2+

## Expected Behavior

Not every day has 24 hours, every minute 60 seconds, nor every year has 365 days. When delaying a reminder from 2021-12-02 for 4 years, I'd expect it to be 2025-12-02.

## Actual Behavior

The reminder gets scheduled for 2025-12-01.

## Steps to Reproduce the Problem

Create a reminder in the future that passes a leap year or a leap second, or running on an OS with the timezone set to a local timezone.

It looks like the number of days in a year is [hard coded](https://github.com/dapr/dapr/blob/master/pkg/actors/actors.go#L1651).

## Release Note

RELEASE NOTE:

PR #4027, opened 2021-12-14
use addDate to support leap year & leap month

theonefx:
# Description

In the original implementation logic, the number of years was directly converted to days. But without knowing the base time, it's impossible to know if it's a leap year. So the calculation is not accurate. As code below:
```go
		switch name {
		case "year":
			duration += time.Hour * 24 * 365 * time.Duration(val)
		case "month":
			duration += time.Hour * 24 * 30 * time.Duration(val)
		case "week":
			duration += time.Hour * 24 * 7 * time.Duration(val)
		case "day":
			duration += time.Hour * 24 * time.Duration(val)
		case "hour":
			duration += time.Hour * time.Duration(val)
		case "minute":
			duration += time.Minute * time.Duration(val)
		case "second":
			duration += time.Second * time.Duration(val)
		case "repetition":
			repetition = val
		default:
			return 0, 0, fmt.Errorf("unsupported ISO8601 duration field %s", name)
		}
```

So, return the years and duration at the same time, and invoke "AddDate(years, months, days)" & "Add(period)" to get the correct time.

## Issue reference
https://github.com/dapr/dapr/issues/3995

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Unit tests passing

### DIFF
--- a/pkg/actors/actors.go
+++ b/pkg/actors/actors.go
@@ -1065,7 +1065,7 @@ func (a *actorsRuntime) CreateTimer(ctx context.Context, req *CreateTimerRequest
 		repeats             int
 		dueTime, ttl        time.Time
 		period              time.Duration
-		yeads, months, days int
+		years, months, days int
 	)
 	a.activeTimersLock.Lock()
 	defer a.activeTimersLock.Unlock()
@@ -1092,7 +1092,7 @@ func (a *actorsRuntime) CreateTimer(ctx context.Context, req *CreateTimerRequest
 
 	repeats = -1 // set to default
 	if len(req.Period) != 0 {
-		if yeads, months, days, period, repeats, err = parseDuration(req.Period); err != nil {
+		if years, months, days, period, repeats, err = parseDuration(req.Period); err != nil {
 			return errors.Wrap(err, "error parsing timer period")
 		}
 		// error on timers with zero repetitions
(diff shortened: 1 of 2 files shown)


daixiang0:
Please add tests for leap year/month.

theonefx:
> Please add tests for leap year/month.

done

pkg/actors/actors_test.go
```diff
@@ -1736,6 +1736,22 @@ func TestParseDuration(t *testing.T) {
 		assert.Equal(t, time.Hour*2+time.Minute*10+time.Second*3, duration)
 		assert.Equal(t, -1, repetition)
 	})
+	t.Run("parse ISO 8610 with leap yean", func(t *testing.T) {
```
daixiang0, line 1739: typo `yean` here.

theonefx: done

theonefx:
@artursouza @yaron2

pkg/actors/actors.go
```diff
@@ -1060,10 +1061,11 @@ func (a *actorsRuntime) CreateReminder(ctx context.Context, req *CreateReminderR
...
 	var (
-		err          error
-		repeats      int
-		dueTime, ttl time.Time
-		period       time.Duration
+		err                 error
+		repeats             int
+		dueTime, ttl        time.Time
+		period              time.Duration
+		yeads, months, days int
```
yaron2, line 1068: typo - change `yeads` to `years`.

yaron2 (changes requested):

pkg/actors/actors.go
theonefx: done

yaron2 (approved):
lgtm

Issue #3586, opened 2021-08-25
[Proposal] Resiliency policies across all building blocks


pkedy:
/area runtime
/area operator

The goal of this proposal is to create a mechanism where [resiliency patterns](https://docs.microsoft.com/en-us/dotnet/architecture/cloud-native/application-resiliency-patterns) can be applied consistently configured across all Dapr building blocks.

Design goals are:

* Support high-value resiliency policies: Timeouts, [Retries/Back off](https://docs.microsoft.com/en-us/azure/architecture/patterns/retry) and [Circuit Breakers](https://docs.microsoft.com/en-us/azure/architecture/patterns/circuit-breaker)
* Resiliency policies are configured at the application scope
* Compatibility across all component types and service invocation
* Developers must opt-in, default behavior is unchanged
* Policies can be configured with sane defaults
* Rules can be used to assign specific policies to Dapr API operation

Non-goals are:

* Distributed circuit breaking (additive feature)
* [Bulkheads](https://docs.microsoft.com/en-us/azure/architecture/patterns/bulkhead) and [Throttling](https://docs.microsoft.com/en-us/azure/architecture/patterns/throttling): These can be addressed by future proposals
* Fallback operations: These would be handled in code, not declaratively

Because this proposal targets many resiliency policy types and spans across all building blocks, it is related to several existing issues:

* #1820
* #3000
* #777
* #2716 (Required for timeouts & retry)

## Design

Even the simplest resiliency configuration can be 100+ lines of YAML. To prevent polluting the main configuration, a new `Resiliency` CRD is proposed in order to promote clear separation of concerns from other features like RBAC. It also creates a distinct boundary for SREs and architects to work in without worrying about impacting other aspects of the system. The new CRD has three main sections: **policies**, **buildingBlocks** and **apis**.

### Policy definition

The `policies` section is where timeout, retry and circuit breaker polices are defined. Each is given a name so they can be referred to from the `buildingBlocks` and `apis` sections.

#### Timeouts

Timeouts only require a duration expressed in [Go's time.Duration format](https://pkg.go.dev/time#ParseDuration).

#### Retries

Retries follow either a `constant` or `exponential` configuration with unique settings for each. These will be based on the existing [dapr/kit retry package](https://github.com/dapr/kit/tree/main/retry) which uses the popular `github.com/cenkalti/backoff` package.

Constant configuration:

| Field                     | Description                                 | Default   |
|---------------------------|---------------------------------------------|-----------|
[...]

PR #4032, opened 2021-12-14
Add hooks to load resiliency into runtime

halspang:
# Description

This commit adds the hooks to load a resiliency configuration into
the runtime when an app starts for both standalone and k8s.

The configuration is not used anywhere, it is simply loaded.

## Issue reference

Partial: #3586 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Unit tests passing

### DIFF
--- a/charts/dapr/charts/dapr_rbac/templates/ClusterRoleBinding.yaml
+++ b/charts/dapr/charts/dapr_rbac/templates/ClusterRoleBinding.yaml
@@ -30,13 +30,13 @@ metadata:
   name: dapr-operator-admin
 rules:
 - apiGroups: ["*"]
-  resources: ["customresourcedefinitions", "serviceaccounts", "deployments", "services", "configmaps", "secrets", "components", "configurations", "subscriptions", "leases"]
+  resources: ["customresourcedefinitions", "serviceaccounts", "deployments", "services", "configmaps", "secrets", "components", "configurations", "subscriptions", "leases", "resiliencies"]
   verbs: ["get"]
 - apiGroups: ["*"]
-  resources: ["deployments", "services", "components", "configurations", "subscriptions", "leases", "secrets"]
+  resources: ["deployments", "services", "components", "configurations", "subscriptions", "leases", "secrets", "resiliencies"]
   verbs: ["list"]
 - apiGroups: ["*"]
-  resources: ["deployments", "services", "components", "configurations", "subscriptions", "leases", "secrets"]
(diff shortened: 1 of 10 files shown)


daixiang0:
Need to update docs?

halspang:
> Need to update docs?

We will yes, but not immediately. This is going into a feature branch instead of master as there's a lot to it and @pkedy and I are trying to get to a point where there's something actually worth while before we merge it into master.

Given how much things will potentially change before we're ready to merge, I'd prefer to keep documentation out until we have a stronger foundation. Thoughts?

pkedy (approved):
LGTM

Just noting this deviates from the proposal in that it uses a switch/arg to point to the resiliency config instead of using `scopes` inside the CRD itself. This is fine. I actually kind of like it. We went this `scopes` to be consistent with everything else. I say lets merge this and keep working and we can revisit this at a later time.

Please review/merge @artursouza

artursouza (approved):

yaron2:
Actually @pkedy there's a real good reason why it's preferable to go with the proposal and not use a flag as done in this PR.

In effect, the current implementation limits Dapr to only having a single resiliency feature per app. While this is inconsistent with our Subscriptions and Components CRDs, and by extension doesn't conform to users expectations about how Kubernetes specifically works, the greatest issue with this approach is in how it limits areas of ownership between development and operations teams.

Resiliency policies can be applied to multiple Dapr APIs, and to multiple components within these APIs.
Requiring policies to live in a single resource forces prohibits multiple teams with different permissions to different components to apply policies in regards to their RBAC rules.

So an operations team that has ownership of the app database and it's associated resiliency policy needs to give access to the development team which owns the resiliency policies for service invocation calls, for example.

This creates an awkward situation where multiple teams needs to have shared access to a single resource, and this problem becomes even bigger in CI pipelines where all changes (including additions and removals of resiliency policies) cause mutation of the existing resiliency policy.

Another unwelcome side effect of a single resiliency policy is a long and unreadable YAML. The developer experience should not be optimized around one or two policies in a resource, but rather the opposite - what does the user experience look like when the feature is highly used and is prevalent in the system?

cc @artursouza @pkedy @halspang

pkedy:
> So an operations team that has ownership of the app database and it's associated resiliency policy needs to give access to the development team which owns the resiliency policies for service invocation calls, for example.

The ownership aspect is what resonates with me the most. Thanks for chiming in on this, @yaron2.

Issue #3839, opened 2021-11-03
Retry /dapr/subscribe in case of failure or wait for app to get into Ready state


mcrio:
Hi,

Posting a pub/sub related problem for discussion:

Problem:
In case `/dapr/subscribe` is not available when the sidecar starts, the app will not subscribe to the event queue.

Scenario (reproducible using Tye):

- App has readiness and liveness health-checks defined
- Dapr sidecar starts, App is not yet in `Ready` state
- Dapr sends `/dapr/subscribe` request but the connection is refused because the app is not in `Ready` state yet
- Dapr will not retry getting the subscriptions. App finally gets into `Ready` state. No events will be processed.

Possible solutions:
Retry logic for the `/dapr/subscribe` endpoint or wait until app gets into `Ready` state by monitoring health-check endpoints.

This problem was partly discussed in:
https://github.com/dapr/dapr/issues/2918
https://github.com/dapr/dapr/pull/3605

rtrindvg:
I am seeing the same behavior on my environment, with the added problem of the super verbose stderr logging.

We are using a redis as a pub/sub service. We had to upgrade the node pool to a bigger instance and since all the pods were relocated to the new nodes, it seems the redis was late to be available to the pub/sub services. This is the sample line which logs nonstop (millions of lines per day) on the affecting services:
`time="2021-11-16T14:30:28.353396986Z" level=error msg="redis streams: error reading from stream svcname: NOGROUP No such key 'svcname' or consumer group 'svcname' in XREADGROUP with GROUP option" app_id=svcname instance=svcname scope=dapr.contrib type=log ver=1.4.3`

Redis pod was stable, 0 restarts since the move. After deleting/recreating the affecting consumer pods, each daprd sidecar was able to talk with the pub/sub service and the issue was gone. So beyond the service not working, it generated a couple thousand dollars of cost because of the log verbosity ingest on Azure.

mcrio:
Hi @rtrindvg It seems like the problem you described deserves a separate issue. This one is just about the subscriptions discovery endpoint that the application provides, while the scenario you described is a sidecar <> pub/sub connection problem.

@Team-Dapr do you think it makes sense to have this retry logic. The subscriptions discovery procedure seems a bit unreliable at the moment.

yaron2:
[...]

PR #4036, opened 2021-12-16
Add exponential backoff to topic subscriptions calls

yaron2:
Closes https://github.com/dapr/dapr/issues/3839.

This PR adds exponential retries to the methods that get topic subscriptions from the app using HTTP and gRPC and adds test cases.

### DIFF
--- a/pkg/runtime/pubsub/subscriptions.go
+++ b/pkg/runtime/pubsub/subscriptions.go
@@ -10,13 +10,14 @@ import (
 	"time"
 
 	"github.com/cenkalti/backoff/v4"
-	"github.com/dapr/kit/retry"
 	"github.com/ghodss/yaml"
 	"github.com/pkg/errors"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 	"google.golang.org/protobuf/types/known/emptypb"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/dapr/kit/retry"
 
 	subscriptionsapi_v1alpha1 "github.com/dapr/dapr/pkg/apis/subscriptions/v1alpha1"
 	subscriptionsapi_v2alpha1 "github.com/dapr/dapr/pkg/apis/subscriptions/v2alpha1"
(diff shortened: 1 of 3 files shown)


pkg/runtime/pubsub/subscriptions.go
```diff
@@ -8,8 +8,11 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/cenkalti/backoff"
```
halspang, line 11: Should this be using https://github.com/dapr/kit/tree/main/retry?

halspang (commented):
Looks good to me overall, had a quick question about consistency in regards to `dapr/kit`.

Also, do we think this would fall under the umbrella of resiliency? Or do we think that stuff in the init path is kind of outside of resiliency for now as it's focusing a lot on invocations/component interactions?

yaron2:
> Looks good to me overall, had a quick question about consistency in regards to `dapr/kit`.
> 
> Also, do we think this would fall under the umbrella of resiliency? Or do we think that stuff in the init path is kind of outside of resiliency for now as it's focusing a lot on invocations/component interactions?

This is not related to the resiliency feature, which focuses on optional, opt-in resiliency policies for the Dapr APIs and are very tight with underlying components behavior and set up. Getting topic configurations, like operator and placement retries, is a runtime init `daprd` concern.

pkg/runtime/pubsub/subscriptions.go
yaron2: Thanks for the pointer, switched to `kit/retry`.

tests/apps/service_invocation_grpc_proxy_server/service.yaml
```diff
@@ -20,7 +20,6 @@ spec:
         dapr.io/enabled: "true"
         dapr.io/app-id: "grpcproxyserver"
         dapr.io/app-port: "50051"
-        dapr.io/app-protocol: "grpc"
```
artursouza, line 23: Why reverting this? Is it a left over from debugging?

yaron2: Redundant, was never needed.

artursouza (approved):

halspang (approved):

PR #4049, opened 2021-12-22
Charts - Supports to customize livenessProbe and readinessProbe

haoxins:
# Description

When I use Argo CD to deploy the Dapr, some components (Operator and Placement) need more time to start.
Sometimes the Argo CD may kill/restart the pods because of the health check.

So I think it's a easy and safe way to increase the initial delay and period seconds

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing

### DIFF
--- a/charts/dapr/charts/dapr_operator/templates/dapr_operator_deployment.yaml
+++ b/charts/dapr/charts/dapr_operator/templates/dapr_operator_deployment.yaml
@@ -86,22 +86,22 @@ spec:
             path: /healthz
             port: 8080
 {{- if eq .Values.debug.enabled false }}
-          initialDelaySeconds: 3
+          initialDelaySeconds: {{ .Values.livenessProbe.initialDelaySeconds }}
 {{- else }}
           initialDelaySeconds: {{ .Values.debug.initialDelaySeconds }}
 {{- end }}
-          periodSeconds: 3
+          periodSeconds: {{ .Values.livenessProbe.periodSeconds }}
           failureThreshold: 5
         readinessProbe:
           httpGet:
             path: /healthz
             port: 8080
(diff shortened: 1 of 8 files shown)


yaron2:
Instead of changing the defaults globally in Dapr, isn't it better to use the already configurable settings to increase the delay for your Argo deployment?

haoxins:
```
{{- if eq .Values.debug.enabled false }}
          initialDelaySeconds: 5
{{- else }}
          initialDelaySeconds: {{ .Values.debug.initialDelaySeconds }}
{{- end }}
```

Maybe we should accept the custom params when `Values.debug.enabled = false`?

yaron2:
> ```
> {{- if eq .Values.debug.enabled false }}
>           initialDelaySeconds: 5
> {{- else }}
>           initialDelaySeconds: {{ .Values.debug.initialDelaySeconds }}
> {{- end }}
> ```
> 
> Maybe we should accept the custom params when `Values.debug.enabled = false`?

That's a much better idea.

haoxins:
> > ```
> > {{- if eq .Values.debug.enabled false }}
> >           initialDelaySeconds: 5
> > {{- else }}
> >           initialDelaySeconds: {{ .Values.debug.initialDelaySeconds }}
> > {{- end }}
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Maybe we should accept the custom params when `Values.debug.enabled = false`?
> 
> That's a much better idea.

PR updated~

mukundansundar (approved):
lgtm

daixiang0 (approved):

PR #4072, opened 2021-12-31
Fixed timer call actor judgment

ls-2018:
# Description
The timer calling actor should determine if the application is local

  
## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests
* [x] Unit tests passing

### DIFF
--- a/pkg/actors/actors.go
+++ b/pkg/actors/actors.go
@@ -1199,7 +1199,7 @@ func (a *actorsRuntime) executeTimer(actorType, actorID, name, dueTime, period, 
 	req := invokev1.NewInvokeMethodRequest(fmt.Sprintf("timer/%s", name))
 	req.WithActor(actorType, actorID)
 	req.WithRawData(b, invokev1.JSONContentType)
-	_, err = a.callLocalActor(context.Background(), req)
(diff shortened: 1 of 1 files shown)


pkg/actors/actors.go
```diff
@@ -1199,7 +1199,7 @@ func (a *actorsRuntime) executeTimer(actorType, actorID, name, dueTime, period,
 	req := invokev1.NewInvokeMethodRequest(fmt.Sprintf("timer/%s", name))
 	req.WithActor(actorType, actorID)
 	req.WithRawData(b, invokev1.JSONContentType)
-	_, err = a.callLocalActor(context.Background(), req)
+	_, err = a.Call(context.Background(), req)
```
yaron2, line 1202: Please revert back to `callLocalActor` since timers can only be created for actors that are activated in the local Dapr sidecar.

yaron2 (changes requested):

pkg/actors/actors.go
yaron2: Ok, I see this is the only change. I think we should close this PR, as mentioned above, actor timers can only be activated locally.

Issue #3982, opened 2021-12-02
daprd crashes if gRPC request arrives before runtime init completes


berndku:
## In what area(s)?

/area runtime

## What version of Dapr?

1.5.0

## Expected Behavior
Using below two building blocks together in a single app is working: 
- call gRPC server over [dapr gRPC proxying with asyncio channel](https://github.com/dapr/python-sdk/blob/master/examples/grpc_proxying/invoke-caller.py)
- [Programmatically subscribing](https://docs.dapr.io/developing-applications/building-blocks/pubsub/howto-publish-subscribe/#programmatic-subscriptions) to MQTT topics with dapr Pub/Sub

## Actual Behavior

Sidecar is crashing:

panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x13e4deb]

goroutine 204 [running]:
github.com/dapr/dapr/pkg/messaging.(*proxy).intercept(0xc0005f0550, {0x3f0d238, 0xc000da8420}, {0xc000bb0880, 0x1d})
        /home/runner/work/dapr/dapr/pkg/messaging/grpc_proxy.go:89 +0x40b
github.com/dapr/dapr/pkg/grpc/proxy.(*handler).handler(0xc000596cb0, {0x0, 0x0}, {0x3f34900, 0xc000bba940})
        /home/runner/work/dapr/dapr/pkg/grpc/proxy/handler.go:68 +0x9d
github.com/dapr/dapr/pkg/diagnostics.GRPCTraceStreamServerInterceptor.func1({0x0, 0x0}, {0x3f34ea0, 0xc000db6000}, 0xc000bb28e8, 0xc000c28280)
        /home/runner/work/dapr/dapr/pkg/diagnostics/grpc_tracing.go:121 +0x3fa
github.com/grpc-ecosystem/go-grpc-middleware.ChainStreamServer.func1.1.1({0x0, 0x0}, {0x3f34ea0, 0xc000db6000})
        /home/runner/go/pkg/mod/github.com/grpc-ecosystem/go-grpc-middleware@v1.2.2/chain.go:49 +0x3a
github.com/grpc-ecosystem/go-grpc-middleware.ChainStreamServer.func1({0x0, 0x0}, {0x3f34ea0, 0xc000db6000}, 0x37206a0, 0xc000bd4d80)
[...]

PR #4079, opened 2022-01-04
Return error instead of crashing when receiving grpc_proxy request on uninitialized daprd

wcs1only:
# Description

If a grpc proxy request comes in before daprd has completed initialization (for instance if daprd is blocked waiting on the app to start listening), daprd will crash. This change returns an error instead.

## Issue reference

Please reference the issue this PR will close: #3982
## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing

### DIFF
--- a/pkg/messaging/grpc_proxy.go
+++ b/pkg/messaging/grpc_proxy.go
@@ -80,6 +80,10 @@ func (p *proxy) intercept(ctx context.Context, fullName string) (context.Context
 	outCtx := metadata.NewOutgoingContext(ctx, md.Copy())
 	appID := v[0]
 
+	if p.remoteAppFn == nil {
+		return ctx, nil, errors.Errorf("failed to proxy request: proxy not initialized. daprd startup may be incomplete.")
+	}
+
 	target, err := p.remoteAppFn(appID)
 	if err != nil {
 		return ctx, nil, err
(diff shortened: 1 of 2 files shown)


pkg/messaging/grpc_proxy.go
```diff
@@ -98,6 +98,10 @@ func (p *proxy) intercept(ctx context.Context, fullName string) (context.Context
 		return outCtx, conn, cErr
 	}
 
+	if p.remoteAppFn == nil {
```
yaron2, line 101: Please move the if block before line 83: https://github.com/dapr/dapr/blob/a782b85439c7d3ed87a76b2eb086e2f9eedb89b4/pkg/messaging/grpc_proxy.go#L83.

yaron2 (changes requested):

yaron2:
Left one comment, but would also be good to add a unit test for this when `remoteAppFn` is nil.

pkg/messaging/grpc_proxy.go
wcs1only: Oh, bad rebase from release-1.5 to master. Good catch.

wcs1only:
> Left one comment, but would also be good to add a unit test for this when `remoteAppFn` is nil.

Indeed. Will add.

yaron2 (approved):
lgtm

PR #4080, opened 2022-01-04
Fix flakiness on get metadata e2e.

artursouza:
Fix flakiness on get metadata e2e when there are additional tests running in parallel that can influence the actor count.

## Issue reference

Please reference the issue this PR will close: #N/A

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests
* [X] Unit tests passing

### DIFF
--- a/tests/e2e/actor_features/actor_features_test.go
+++ b/tests/e2e/actor_features/actor_features_test.go
@@ -607,13 +607,10 @@ func TestActorFeatures(t *testing.T) {
 		res, err := utils.HTTPGet(fmt.Sprintf(actorMetadataURLFormat, externalURL))
 		require.NoError(t, err)
 
-		var prevMetadata metadata
-		err = json.Unmarshal(res, &prevMetadata)
+		var previousMetadata metadata
+		err = json.Unmarshal(res, &previousMetadata)
 		require.NoError(t, err)
-		var prevActors int
-		if len(prevMetadata.Actors) > 0 {
-			prevActors = prevMetadata.Actors[0].Count
-		}
+		require.NotNil(t, previousMetadata)
 
 		// Each test needs to have a different actorID
 		actorIDBase := "1008Instance"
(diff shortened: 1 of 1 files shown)


tests/e2e/actor_features/actor_features_test.go
```diff
@@ -629,16 +626,17 @@ func TestActorFeatures(t *testing.T) {
...
-		require.Equal(t, expected, actual)
+		require.NotNil(t, currentMetadata)
+
+		require.Equal(t, appName, currentMetadata.ID)
+		require.Equal(t, appName, previousMetadata.ID)
+		require.Greater(t, len(previousMetadata.Actors), 0)
+		require.Greater(t, len(currentMetadata.Actors), 0)
+		require.Equal(t, "testactorfeatures", currentMetadata.Actors[0].Type)
+		require.Equal(t, "testactorfeatures", previousMetadata.Actors[0].Type)
+		require.Greater(t, currentMetadata.Actors[0].Count, previousMetadata.Actors[0].Count)
```
artursouza, line 640: We just assert it is greater and not necessarily the exact count.

yaron2 (approved):

daixiang0:
Developers may not consider parallel issues, do we really need run tests in parallel? If so, we should better highlight those tests.

artursouza:
> Developers may not consider parallel issues, do we really need run tests in parallel? If so, we should better highlight those tests.

The time to run tests increase a lot without parallelism. We need to keep this in mind when reviewing PRs. This is the only way to reduce time to run tests. IMO.

daixiang0:
Look at https://github.com/dapr/dapr/runs/4698628375?check_suite_focus=true, build and push containers spend ~50% time, if we do it only once before all e2e tests, we can reduce much time.

Refer to https://github.com/docker/build-push-action/blob/master/docs/advanced/share-image-jobs.md.

artursouza:
> Look at https://github.com/dapr/dapr/runs/4698628375?check_suite_focus=true, build and push containers spend ~50% time, if we do it only once before all e2e tests, we can reduce much time.
> 
> Refer to https://github.com/docker/build-push-action/blob/master/docs/advanced/share-image-jobs.md.

Yes, that is right. We have kept the build of E2E apps in the pipeline because it is very common to have PRs that require changes to E2E test app to be merged together with the test code. I agree that we can expedite the build time too.

PR #4016, opened 2021-12-10
dapr support rocketmq

x-shadow-man:
We have combined dapr and rocketmq at work, so the rocketmq component is introduced in the dapr project, so that dapr can support rocketmq

PR #4081, opened 2022-01-04
dapr support rocketmq #4016

x-shadow-man:
We have combined dapr and rocketmq at work, so the rocketmq component is introduced in the dapr project, so that dapr can support rocketmq.


x-shadow-man:
> We have combined dapr and rocketmq at work, so the rocketmq component is introduced in the dapr project, so that dapr can support rocketmq.

@yaron2 

this RP refer to closed PR #4081 

pls feel free to check and merge. thx for your help.

PR #4091, opened 2022-01-05
CI: speed up e2e tests

daixiang0:
Signed-off-by: Loong <loong.dai@intel.com>

# Description

Build one time to reduce.

## Issue reference

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing


daixiang0:
Compared with https://github.com/dapr/dapr/pull/4088, the time of the e2e test reduces.

docker/docker.mk
```diff
@@ -121,6 +121,14 @@ else
 	$(DOCKER) buildx build --build-arg PKG_FILES=sentry --platform $(DOCKER_IMAGE_PLATFORM) -f $(DOCKERFILE_DIR)/$(DOCKERFILE) $(BIN_PATH) -t $(DAPR_SENTRY_DOCKER_IMAGE_TAG)-$(TARGET_OS)-$(TARGET_ARCH) --push
 endif
 
+# push docker image for e2e tests
+docker-push-e2e:
```
wcs1only, line 125: Why do we need a separate target for this? Isn't docker-push above sufficient?

.github/workflows/kind-e2e.yaml
```diff
@@ -37,8 +37,39 @@ jobs:
...
+    - name: Set up Go ${{ env.GOVER }}
+      uses: actions/setup-go@v2
+      with:
+        go-version: ${{ env.GOVER }}
+    - name: Check out code onto GOPATH
+      uses: actions/checkout@v2
+      with:
+        path: ./src/github.com/dapr/dapr
+    # setup image cache
+    - uses: satackey/action-docker-layer-caching@v0.0.11
```
wcs1only, line 60: We have attempted this in the past. Trouble was the Windows images being extremely large ended up paving the cache and breaking this optimization. But, let's give it a try and see what happens.

docker/docker.mk
daixiang0: `docker-push` depends on `docker-build` :(

.github/workflows/kind-e2e.yaml
daixiang0: It is for kind e2e test which only run at Linux I think.

daixiang0:
@wcs1only could we make it land?

docker/docker.mk
artursouza: I would rather create a special env variable that will skip the build and keep the docker-push target above only. There is more logic to support multi arch, for example, that this one does not replicate.

artursouza (changes requested):

Issue #3761, opened 2021-10-13
Add kubernetes annotation to enable/disable placement component in daprd


auhlik:
## In what area(s)?

 /area runtime

> /area operator

> /area placement

> /area docs

> /area test-and-release

## Describe the feature
**Placement** component is marked as [optional](https://docs.dapr.io/operations/hosting/kubernetes/kubernetes-production/#optional-components) in dapr docs. **Dapr** and **daprd** indeed work without it, however, in debug mode  **daprd** log is full of unsuccessful attempts to connect to **placement**:
```
time="2021-10-12T15:10:43.546021623Z" level=debug msg="try to connect to placement service: dns:///dapr-placement-server.default.svc.cluster.local:50005" app_id=test-app instance=test-app-62dd487dfd-m4dre scope=dapr.runtime.actor.internal.placement type=log ver=edge
```
This might create an additional load on the network, cpu and make user think that **placement** component is mandatory.

It would be nice to have a kubernetes annotation in, e.g. `dapr.io/enable-placement` which can be used to enable/disable **placement** component in **daprd**. Some code changes will be needed to implement the feature.
## Release Note

RELEASE NOTE:
**ADD** kubernetes annotation to enable/disable placement component in daprd.

yaron2:
Thanks for reporting this @auhlik, thanks to @CrazyHZM this will be fixed in the upcoming 1.5 release.

artursouza:
Moving this to 1.6

CrazyHZM:
@artursouza 
Please reopen this issues.

msfussell:
@CrazyHZM - Please ensure that you also write the docs for this, before closing this issue. I have created this issue dapr/docs#2074

artursouza:
Can we achieve the same by setting an annotation that sets the placement addresses as empty? This will simplify the code a lot and make this an extensible solution.

CrazyHZM:
> Can we achieve the same by setting an annotation that sets the placement addresses as empty? This will simplify the code a lot and make this an extensible solution.

I'm not sure if there is an error message missing: 
when the user wants to turn on placement, but forgets to configure the address.
If you use address as the basis for judging whether to disable placement, you will lose this error message
@artursouza

msfussell:
Reopen

auhlik:
Reopen

PR #3823, opened 2021-10-28
Support enable/disable placement component

CrazyHZM:
# Description

Add kubernetes annotation to enable/disable placement component in daprd

## Issue reference

Please reference the issue this PR will close: #3761

## Checklist

Please make sure you've  completed the relevant tasks for this PR, out of the following list:

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Unit tests passing
* [ ] End-to-end tests passing
* [ ] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_
* [ ] Specification has been updated / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_
* [ ] Provided sample for the feature / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_

PR #4094, opened 2022-01-05
Support enable/disable placement component

CrazyHZM:
Signed-off-by: crazyhzm <crazyhzm@gmail.com>

# Description

Add kubernetes annotation to enable/disable placement component in daprd
Clean up https://github.com/dapr/dapr/pull/3823
## Issue reference

Please reference the issue this PR will close: #3761

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing


tests/platforms/kubernetes/kubeobj.go
```diff
@@ -77,6 +77,7 @@ func buildDaprAnnotations(appDesc AppDescription) map[string]string {
 			"dapr.io/sidecar-readiness-probe-threshold": "15",
 			"dapr.io/sidecar-liveness-probe-threshold":  "15",
 			"dapr.io/enable-metrics":                    strconv.FormatBool(appDesc.MetricsEnabled),
+			"dapr.io/enable-placement":                  strconv.FormatBool(appDesc.EnablePlacement),
```
artursouza, line 80: instead, set "dapr.io/placement-addresses" and set it as empty for disabling it. The code would be much simpler and would be a more extensible solution for future scenarios. Also, we already have a way to set the placement address via daprd, so this would also remove the need for extra param in the daprd binary.

artursouza (changes requested):

tests/platforms/kubernetes/kubeobj.go
CrazyHZM: I'm not sure if there is an error message missing:
when the user wants to turn on placement, but forgets to configure the address.
If you use address as the basis for judging whether to disable placement, you will lose this error message

yaron2: Actually, we might also want to allow users to skip the installation of the placement service all together.

CrazyHZM: If  also want to use this configuration to skip the installation of the placement service all together, is `dapr.io/enable-placement` more appropriate?

yaron2: Skipping the installation should be done in the Helm Chart, and add a setting like `global.placementService=true/false`.

CrazyHZM: I can understand the idea of ​​replacing `dapr.io/enable-placement` with `dapr.io/placement-addresses` to simplify configuration, but I'm still not sure if users will lose the error message that they forgot to configure addresses. I think users should expect to receive error message feedback if no address is configured when placement is turned on.
Are we going to consider this situation? I'll use `dapr.io/placement-addresses` later if we don't need to consider this situation.
@yaron2

yaron2: @CrazyHZM skipping the installation of the placement service needs to happen at the Dapr installation level. the `dapr.io` annotations targets configuration for a specific deployment. This is where the Helm setting comes into play.

As for the annotation, that's how I see it for `dapr.io/placement-address`:

If no annotation exists, the default value is used (set by the Sidecar Injector).
If an annotation exists (dapr.io/placement-address) with the value `""` for a given deployment, the sidecar will skip connecting to placement.
If an annotation exists (dapr.io/placement-address) with a non empty string value, the sidecar will connect to the configured address.

Thoughts?

CrazyHZM:
cc @yaron2 @artursouza 
I re-adjusted the logic, please review when you have time.

artursouza (approved):

CrazyHZM:
@artursouza Please review this.

CrazyHZM:
cc @artursouza @yaron2

CrazyHZM:
I found that `--placement-address` has been replaced by `--placement-host-address`
Are we going to align the names: replace `dapr.io/placement-addresses` with `dapr.io/placement-host-address`
@artursouza @yaron2 
cc @msfussell

msfussell:
> I found that `--placement-address` has been replaced by `--placement-host-address` Are we going to align the names: replace `dapr.io/placement-addresses` with `dapr.io/placement-host-address` @artursouza @yaron2 cc @msfussell

Yes we are going to align names to make this consistent between the CLI and K8s annotations

CrazyHZM:
> @CrazyHZM is this change backward compatible with older versions? Meaning, if I'm using actors today (without the new annotation), will that still work after this change?

Yes, it is guaranteed to be backward compatible.
@yaron2

pkg/injector/pod_patch.go
```diff
@@ -629,6 +643,10 @@ func getSidecarContainer(annotations map[string]string, id, daprSidecarImage, im
 
 	HTTPStreamRequestBodyEnabled := HTTPStreamRequestBodyEnabled(annotations)
 
+	if existPlacementAddressesAnnotation(annotations) {
+		placementServiceAddress = getPlacementAddresses(annotations)
+	}
+
```
CrazyHZM, line 649: @yaron2 
It will only take effect when this annotation is added.

CrazyHZM:
ping @artursouza @yaron2

pkg/injector/pod_patch.go
artursouza: LGTM

artursouza (approved):

PR #4114, opened 2022-01-10
Resiliency policy runner

pkedy:
# Description

This PR adds logic to decode resiliency configurations into a `resiliency.Provider`. The provider is used to wrap operations with resiliency policies similar to Polly for .NET. This has been been applied to a few "low lying" places in `runtime.go`:

* Invoking output bindings (component)
* Input binding events (route)
* Publish event (component)
* Receiving PubSub events (route)
* State store bulk set in `onAppResponse` (component)

The resiliency provider is also injected into the HTTP and gRPC APIs but not yet leverage it.

**Next steps:**

To simplify the integration with components, one approach could be to create "resiliency wrappers" for each component type so that resiliency policies are applied transparently and universally.

We need tests that exercise timeout, retry, and circuit breaking. My thought was this would come in a later PR. For timeouts, we will also need all component interfaces to accept `ctx context.Context`.

Multiple resiliency configs can apply to a singe app. Currently, ordering of the configs returned by the operator API and filesystem are non deterministic. We might consider adding a `priority` field to the CRD so that the configs can be sorted by something like `spec.priority` and `metadata.name`.

## Issue reference

Part of #3586

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Unit tests passing


pkg/resiliency/breaker/circuitbreaker.go
```diff
@@ -0,0 +1,79 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package breaker
+
+import (
+	"time"
+
+	"github.com/sony/gobreaker"
```
artursouza, line 19: License is MIT. We are good.

pkg/resiliency/resiliency.go
```diff
@@ -1,132 +1,504 @@
...
-func LoadDefaultResiliency() *Resiliency {
-	return &Resiliency{resiliency_v1alpha.Resiliency{}}
-}
+const (
+	// ActorCircuitBreakerScopeType indicates the type scope (less granular).
+	ActorCircuitBreakerScopeType ActorCircuitBreakerScope = iota
+	// ActorCircuitBreakerScopeID indicates the type+id scope (more granular).
+	ActorCircuitBreakerScopeID
+	// ActorCircuitBreakerScopeBoth indicates both type and type+id are used for scope.
+	ActorCircuitBreakerScopeBoth // Usage is TODO
```
artursouza, line 54: Why even keep it here then?

pkg/resiliency/resiliency.go
```diff
@@ -1,132 +1,504 @@
...
+	for name, t := range buildingBlocks.Components {
+		r.components[name] = PolicyNames{
+			Timeout:        t.Timeout,
+			Retry:          t.Retry,
+			CircuitBreaker: t.CircuitBreaker,
+		}
+	}
+
+	for name, t := range buildingBlocks.Routes {
+		name = strings.TrimPrefix(name, "/") // Ignore slash prefix.
```
artursouza, line 277: Does this handle multiple "/"?

pkg/resiliency/resiliency.go
```diff
@@ -1,132 +1,504 @@
...
+
+	return nil
+}
+
+// RoutePolicy returns the policy for a route.
+func (r *Resiliency) RoutePolicy(ctx context.Context, name string) Runner {
+	var t time.Duration
+	var rc *retry.Config
+	var cb *breaker.CircuitBreaker
+	name = strings.TrimPrefix(name, "/") // Ignore slash prefix.
```
artursouza, line 293: same

pkg/resiliency/policy.go
```diff
@@ -0,0 +1,78 @@
...
+		if r == nil {
+			return operation(ctx)
+		}
+
+		// Use retry/back off
+		b := r.NewBackOffWithContext(ctx)
+		return retry.NotifyRecover(func() error {
+			return operation(ctx)
+		}, b, func(_ error, _ time.Duration) {
+			log.Printf("Error processing operation %s. Retrying...", operationName)
```
artursouza, line 73: Infof

pkg/resiliency/policy.go
```diff
@@ -0,0 +1,78 @@
...
+		}, func() {
+			log.Printf("Recovered processing operation %s.", operationName)
```
artursouza, line 75: Infof

pkg/resiliency/breaker/circuitbreaker.go
```diff
@@ -0,0 +1,79 @@
...
+	Name        string
+	MaxRequests uint32        `mapstructure:"maxRequests"`
+	Interval    time.Duration `mapstructure:"interval"`
+	Timeout     time.Duration `mapstructure:"timeout"`
+	Trip        *expr.Expr    `mapstructure:"trip"`
+	breaker     *gobreaker.CircuitBreaker
+}
+
+func IsErrorPermanent(err error) bool {
+	return err == gobreaker.ErrOpenState || err == gobreaker.ErrTooManyRequests
```
artursouza, line 34: How can components emit permanent errors too?

artursouza (changes requested):

pkg/resiliency/resiliency.go
pkedy: We may end up nixing "both" by the end just due to complexity.

pkg/resiliency/resiliency.go
pkedy: It doesn't need to. It's based on the path which throughout Dapr can have 1 or 0 prefix forward slashes.

pkg/resiliency/breaker/circuitbreaker.go
pkedy: Wrap the error with `backoff.Permanent(err)`

pkg/resiliency/resiliency.go
halspang: If it's just an index into the map, is there a reason to ignore it? Seems like it'll work as a key regardless of the slash.

pkg/resiliency/resiliency.go
```diff
@@ -1,132 +1,508 @@
...
+				cb = ep.Get(endpoint, template)
+			}
+		}
+	}
+
+	return Policy(ctx, operationName, t, rc, cb)
+}
+
+// RemoveEndpoint removes an endpoint from a service.
+// This would be used when a service is taken out of service.
```
halspang, line 348: For my own edification, this is just saying that a pod hosting something has gone down so we don't need to track it anymore, right?

pkg/resiliency/resiliency.go
```diff
@@ -1,132 +1,508 @@
...
+func (r *Resiliency) ActorPolicy(ctx context.Context, actorType string, id string) Runner {
+	var t time.Duration
+	var rc *retry.Config
+	var cb *breaker.CircuitBreaker
+	operationName := fmt.Sprintf("actor[%s, %s]", actorType, id)
+	if r == nil {
+		return Policy(ctx, operationName, t, rc, cb)
 	}
+	policyNames, ok := r.actors[actorType]
+	if ok {
```
halspang, line 366: Does this not trip a linter? I feel like there's a nesting limit around 5.

pkg/runtime/cli.go
```diff
@@ -248,22 +248,26 @@ func FromFlags() (*DaprRuntime, error) {
...
 		namespace = os.Getenv("NAMESPACE")
-		resiliencyConfigs = resiliency_config.LoadKubernetesResiliency(*appID, namespace, operatorClient, log)
+		resiliencyConfigs = resiliency_config.LoadKubernetesResiliency(*appID, namespace, operatorClient)
 	case modes.StandaloneMode:
-		resiliencyConfigs = resiliency_config.LoadStandaloneResiliency(*componentsPath, *appID, log)
+		resiliencyConfigs = resiliency_config.LoadStandaloneResiliency(*componentsPath, *appID)
+	}
+	if resiliencyConfigs == nil {
+		resiliencyConfigs = []*resiliency_v1alpha.Resiliency{}
 	}
```
halspang, line 262: Given that you're still checking this here and replacing with an empty array, why'd you change the method to return a nil instead of the empty array?

pkg/resiliency/resiliency.go
pkedy: Good question. We have examples that both use prefix slashes and not. I actually had to add this line in order to normalize the values for comparison.

pkg/resiliency/resiliency.go
pkedy: See above

pkg/resiliency/breaker/circuitbreaker.go
pkedy: This can be wrapped in `kit/retry` to hide the implementations but ultimately it needs to be `backoff.Permanent` and not an interface of some sort.

https://github.com/cenkalti/backoff/blob/6b0e4ad0cd65431b217393bff47d1dff727b264b/retry.go#L59-L62

pkg/resiliency/breaker/circuitbreaker_test.go
```diff
@@ -0,0 +1,45 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/dapr/dapr/pkg/expr"
+	"github.com/dapr/dapr/pkg/resiliency/breaker"
+)
+
+func TestCircuitBreaker(t *testing.T) {
```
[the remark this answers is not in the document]
pkedy: Yes, we do need more tests that exercise timeout, retry, and circuit breaking. My thought was this would come in a later PR. For timeouts, we will also need all component interfaces to accept `ctx context.Context`. (added this to the next steps in the description)

pkg/resiliency/resiliency.go
```diff
@@ -1,132 +1,508 @@
...
+
+	// Resiliency encapsulates configuration for timeouts, retries, and circuit breakers.
+	// It maps services, actors, components, and routes to each of these configurations.
+	// Lastly, it maintains circuit breaker state across invocations.
+	Resiliency struct {
+		timeouts        map[string]time.Duration
+		retries         map[string]*retry.Config
+		circuitBreakers map[string]*breaker.CircuitBreaker
+
+		actorCBCaches map[string]*lru.Cache
```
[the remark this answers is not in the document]
pkedy: Yes. Since there is an unbounded amount of actors, the cache size is a mechanism to evict older circuit breaker state. We might also apply this approach to endpoints instead of explicitly removing them from service via `RemoveEndpoint`.

pkg/resiliency/resiliency.go
```diff
@@ -1,132 +1,508 @@
...
+		EndpointPolicy(ctx context.Context, service string, endpoint string) Runner
+		RemoveEndpoint(name string, endpoint string)
+		ActorPolicy(ctx context.Context, actorType string, id string) Runner
+		ComponentPolicy(ctx context.Context, name string) Runner
+	}
+
+	// Resiliency encapsulates configuration for timeouts, retries, and circuit breakers.
+	// It maps services, actors, components, and routes to each of these configurations.
+	// Lastly, it maintains circuit breaker state across invocations.
+	Resiliency struct {
```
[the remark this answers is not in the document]
pkedy: All of those things are a concern :)  We should run longhaul tests before merging this feature branch.

pkg/resiliency/resiliency.go
```diff
@@ -1,132 +1,508 @@
...
+	// ActorPolicyNames add the circuit breaker scope to PolicyNames.
+	ActorPolicyNames struct {
+		PolicyNames
+		CircuitBreakerScope ActorCircuitBreakerScope
+	}
+)
 
-func LoadStandaloneResiliency(path, runtimeId string, log logger.Logger) []Resiliency {
-	var resiliencies []Resiliency
+var _ = (Provider)((*Resiliency)(nil))
```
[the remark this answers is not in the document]
pkedy: Its a compiler check to ensure that `*Resiliency` satisfies the `Provider` interface.

pkg/resiliency/resiliency.go
pkedy: It probably does. I asked @artursouza to add `feature/*` to the github workflows to run this and the e2e tests. Those probably won't work until we merge master into this branch.

pkedy: This did not come up when running `make lint` but I fixed all the other issues that did :)

pkg/resiliency/resiliency.go
pkedy: Correct

pkg/resiliency/resiliency.go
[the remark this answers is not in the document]
halspang: I think it'd be good to have both under an LRU just to reduce the number of concepts that we're tracking here. Thoughts?

pkg/resiliency/resiliency.go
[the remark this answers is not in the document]
halspang: I am highly supportive of this idea :)

pkg/resiliency/resiliency.go
```diff
@@ -1,133 +1,535 @@
...
+	// Provider is the interface for returning a `Runner` for the various
+	// resiliency scenarios in the runtime.
+	Provider interface {
+		// RoutePolicy returns the policy for a route.
+		RoutePolicy(ctx context.Context, name string) Runner
+		// EndpointPolicy returns the policy for a service endpoint.
+		EndpointPolicy(ctx context.Context, service string, endpoint string) Runner
+		// RemoveEndpoint removes an endpoint from a service.
+		// This would be used when a service is taken out of service.
+		RemoveEndpoint(name string, endpoint string)
```
halspang, line 67: I think it'd be good to move this to an LRU like Actors and not have to worry about the additional method here. Just trying to keep concepts in the config lower where possible given how much it's already covering.

halspang (commented):
I think this is coming along well, just one comment about whether we should keep the `RemoveEndpoint` or leverage more LRU stuff.

pkg/resiliency/resiliency.go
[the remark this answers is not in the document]
pkedy: Both as in also have Endpoints under an LRU? I can get behind that.

pkg/resiliency/resiliency.go
pkedy: You confirmed my question above. I agree and will make the change.

halspang (approved):

pkg/resiliency/breaker/circuitbreaker.go
artursouza: This will need to be incorporated for the pluggable component later on since the response will come from another process.

artursouza (approved):

PR #4130, opened 2022-01-14
unsubscribe api

LaurenceLiZhixin:
# Description

The continue work of https://github.com/dapr/dapr/pull/3927, as the old pr containing many conflicts when i fix dco.
Add configuration unsubscribe api of dapr runtime. As subscribe api is a gRPC stream api, and never return until client call unsubscribe of the target 'store name + keys' subscription.

## Issue reference
#3906 

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests
* [X] Unit tests passing
* [X] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_
* [X] Specification has been updated / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_
* [X] Provided sample for the feature / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_


LaurenceLiZhixin:
This pr can't be merged until unsubscribe of  config component is supported: https://github.com/dapr/components-contrib/pull/1440

charts/dapr/charts/dapr_operator/templates/dapr_operator_deployment.yaml
```diff
@@ -67,6 +67,18 @@ spec:
         prometheus.io/path: "/"
 {{- end }}
     spec:
+{{- if eq .Values.global.ha.enabled true }}
```
beiwei30: Not sure if it's included by accident, but it looks irrelevant to this change.

pkg/grpc/api.go
```diff
@@ -1325,20 +1320,66 @@ func (a *api) SubscribeConfigurationAlpha1(request *runtimev1pb.SubscribeConfigu
...
 	}
+	stop := make(chan struct{})
+	a.configurationSubscribeLock.Lock()
+	a.configurationSubscribe[getConfigSubscribeUniqueKey(request.GetStoreName(), request.GetKeys())] = stop
+	a.configurationSubscribeLock.Unlock()
+	<-stop
+	return nil
+}
+
+func (a *api) UnSubscribeConfigurationAlpha1(ctx context.Context, request *runtimev1pb.UnSubscribeConfigurationRequest) (*runtimev1pb.UnSubscribeConfigurationResponse, error) {
```
beiwei30, line 1341: If the same client subscribe twice with the same store and the same key set, but the lifecycle of the subscriptions are different deliberately, for example, one subscription will keep alive to the end of the application, while the other subscription (with the same interested set to the same store) probably starts and ends under certain circumstance, then the current UnsubscribeConfigurationAlpha1 is problematic, since it will end both subscriptions which is a unwanted side effect.

If we introduce an id in the return values of SubscribeConfigurationAlpha1, and then use the id to unsubscribe, I believe it would solve the problem mentioned above.

beiwei30 (changes requested):

dapr/proto/runtime/v1/dapr.proto
```diff
@@ -92,6 +92,9 @@ service Dapr {
   // SubscribeConfiguration gets configuration from configuration store and subscribe the updates event by grpc stream
   rpc SubscribeConfigurationAlpha1(SubscribeConfigurationRequest) returns (stream SubscribeConfigurationResponse) {}
 
+  // UnSubscribeConfiguration unsubscribe the subscription of configuration
+  rpc UnSubscribeConfigurationAlpha1(UnSubscribeConfigurationRequest) returns (UnSubscribeConfigurationResponse) {}
```
jjcollinge, line 96: nit: `UnSubscribeConfigurationAlpha1` should be `UnsubscribeConfigurationAlpha1`. Same for other references of UnSubscribe.

LaurenceLiZhixin: Thanks, I would fix it.

pkg/grpc/api.go
LaurenceLiZhixin: Thanks, I mentioned your great idea in [issue](https://github.com/dapr/dapr/issues/3906).
 As gRPC's streaming invocation, the response can't be returned directly, but be returned in async streaming response  , so I think putting the subscribe id in each response is ideal. 
In dapr runtime api level, the id is returned as I explained, as gRPC limitation, but in component level, I put the subscrption id into return value of configuration API interface, which is here: https://github.com/dapr/components-contrib/pull/1483/files
@beiwei30

LaurenceLiZhixin:
This pr should pass ci after https://github.com/dapr/components-contrib/runs/5106499336?check_suite_focus=true is merged

charts/dapr/charts/dapr_operator/templates/dapr_operator_deployment.yaml
LaurenceLiZhixin: I fixed it, it's included by accident.

yaron2:
@LaurenceLiZhixin this is merged now, please update this PR so we can get it merged.

berndverst:
```
=== FAIL: pkg/grpc TestSubscribeConfiguration (unknown)
panic: assert: arguments: Cannot call Get(1) because there are 1 argument(s).

=== FAIL: pkg/grpc TestSubscribeConfiguration/get_store_with_single_key (unknown)
panic: assert: arguments: Cannot call Get(1) because there are 1 argument(s).
```

artursouza:
> ```
> === FAIL: pkg/grpc TestSubscribeConfiguration (unknown)
> panic: assert: arguments: Cannot call Get(1) because there are 1 argument(s).
> 
> === FAIL: pkg/grpc TestSubscribeConfiguration/get_store_with_single_key (unknown)
> panic: assert: arguments: Cannot call Get(1) because there are 1 argument(s).
> ```

I know. Fixing that too.

dapr/proto/runtime/v1/dapr.proto
```diff
@@ -511,8 +514,25 @@ message SubscribeConfigurationRequest {
...
+  string id = 2;
+}
+
 message SubscribeConfigurationResponse {
+  // Subscribe id, used to stop subscription.
+  string id = 1;
+
   // The list of items containing configuration values
-  repeated common.v1.ConfigurationItem items = 1;
+  repeated common.v1.ConfigurationItem items = 2;
```
mukundansundar, line 531: @artursouza @yaron2 Please not this change will break existing implementations of the configuration API particularly the subscribe call.

yaron2: It's an Alpha API

Issue #3801, opened 2021-10-22
PubSub subscription `dataAsPayload` flag that sends only the `data/data_base64` value of the data attribute to the application


pkedy:
/area runtime

This feature applies only to the HTTP application protocol. This is not a usability issue with the gRPC API spec.

In some cases, the application may not need the full CloudEvent payload sent to the HTTP application. This is particularly the case with raw events, where receiving the CloudEvent wrapper is unexpected/confusing. This also avoids some performance hit in Base64 encoding/decoding when sending binary data (e.g. Avro/Protobuf). It is important to propagate the telemetry trace IDs through the call flow, but those are added to the header of the request to the app.

Here is what the configuration for raw events could look like:

```yaml
apiVersion: dapr.io/v1alpha1
kind: Subscription
metadata:
  name: products
spec:
  pubsubname: pubsub
  topic: inventory
  route: products
  dataAsPayload: true
```

Note: dataAsPayload is not defined as a metadata attribute because it could apply per route in the case of PubSub Routing:

```yaml
apiVersion: dapr.io/v2alpha1
kind: Subscription
metadata:
  name: mysubscriptions
spec:
  pubsubname: pubsub
  topic: inventory
  dataAsPayload: false   # (maybe) if specified, acts as the default for the routes
  routes:
    rules:
      - match: "event.type == 'widget'"
        path: widgets
        dataAsPayload: true    # overrides the default
      - match: "event.type == 'gadget'"
        path: gadgets
    default: products
    defaultDataAsPayload: true
```

Site note: In hindsight, `default` should have been a nested structure:

```yaml
    default:
      path: products
      dataAsPayload: true
```

I don't think the CRD can be changed now. 🤷 

From an SDK perspective (C# shown) the handler's method signature would look like this for `dataAsPayload: false`

```csharp
public IActionResult Widget([FromBody] CloudEvent<Widget> widgetEvent)
```

Or like this with `dataAsPayload: true`

```csharp
public IActionResult Widget([FromBody] Widget widget)
```

## Release Note

RELEASE NOTE: Subscriptions can optionally send only the data of a CloudEvent to the application

PR #4134, opened 2022-01-14
Add PubSub subscription dataAsPayload flag

Taction:
Signed-off-by: Taction <zchao9100@gmail.com>

# Description

Add PubSub subscription dataAsPayload flag that sends only the data/data_base64 value of the data attribute to the application.

## Issue reference

Please reference the issue this PR will close: #3801 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Unit tests passing


daixiang0:
@artursouza please help trigger /ok-to-test

pkedy:
@Taction Thanks for creating this PR! Please check the e2e test logs.

Taction:
@pkedy I have checked the latest e2e error, it seems related to actor and I don't think it is related to this pr.

pkg/runtime/runtime.go
```diff
@@ -585,6 +588,14 @@ func (a *DaprRuntime) beginPubSub(name string, ps pubsub.PubSub) error {
 				log.Debugf("no matching route for event %v in pubsub %s and topic %s; skipping", cloudEvent[pubsub.IDField], name, msg.Topic)
 				return nil
 			}
+			if dataAsPayload && !rawPayload {
+				cloudData := cloudEvent[pubsub.DataField]
```
famarting, line 592: I was just thinking of this feature and noticed this PR , data can be in two different fields in the CloudEvent so this logic would suffice. I suggest taking a look at the method `publishMessageGRPC` and maybe even to extract to a new method the log get the `data` from the cloudEvent

pkg/runtime/runtime.go
```diff
@@ -554,7 +555,9 @@ func (a *DaprRuntime) beginPubSub(name string, ps pubsub.PubSub) error {
 
 			var cloudEvent map[string]interface{}
 			data := msg.Data
-			if rawPayload {
+			if rawPayload && defaultDataAsPayload {
+				cloudEvent = pubsub.FromRawPayload(nil, msg.Topic, name)
```
famarting, line 559: what is the idea with this?

Taction: I want to avoid base64 encoding when `rawPayload` and `DataAsPayload` are both true. But haven't found a proper way, I'll reconsider this logic later.(`DataAsPayload` is identified by applying matcher to cloudevent)

pkg/runtime/runtime.go
Taction: I don't think `GRPC` needs to do so, it already has a data field. Just as the related issue, this is only for HTTP.

pkg/runtime/runtime.go
Taction: Deleted it for a better review.

daixiang0:
Please fix CI build.

Taction:
CI build fixed.

daixiang0:
ping @Taction

PR #4140, opened 2022-01-17
Reduce placement blocking client

ls-2018:
Signed-off-by: ls-2018 <acejilam@gamil.com>

# Description
Existing logic: an actor Type change ,eg A; will replace all the actor runtime [client] saved by the actor information
then.

1. Send a lock to all hosts, if a client calls an actor A; wait for `WaitUntilPlacementTableIsReady` to finish;
2. Then this will cause all the client calls in the cluster to block;
3. If the process is as it is, then only one client calling actor A will block,other client calls do not block
 

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests
* [x] Unit tests passing

### DIFF
--- a/pkg/placement/membership.go
+++ b/pkg/placement/membership.go
@@ -299,9 +299,11 @@ func (p *Service) performTableDissemination() {
 func (p *Service) performTablesUpdate(hosts []placementGRPCStream, newTable *v1pb.PlacementTables) {
 	// TODO: error from disseminationOperation needs to be handle properly.
 	// Otherwise, each Dapr runtime will have inconsistent hashing table.
-	p.disseminateOperation(hosts, "lock", nil)
-	p.disseminateOperation(hosts, "update", newTable)
-	p.disseminateOperation(hosts, "unlock", nil)
+	for _, host := range hosts {
+		p.disseminateOperation([]placementGRPCStream{host}, "lock", nil)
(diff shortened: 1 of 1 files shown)


daixiang0:
@aronchick I think we need perf tests to see results.

ls-2018:
If possible, I would like someone to help me with this test

pkg/placement/membership.go
```diff
@@ -307,9 +307,11 @@ func (p *Service) performTableDissemination() {
 func (p *Service) performTablesUpdate(hosts []placementGRPCStream, newTable *v1pb.PlacementTables) {
 	// TODO: error from disseminationOperation needs to be handle properly.
 	// Otherwise, each Dapr runtime will have inconsistent hashing table.
-	p.disseminateOperation(hosts, "lock", nil)
-	p.disseminateOperation(hosts, "update", newTable)
```
yaron2, line 311: by removing this you're allowing the sidecars to update with the latest table entries.

ls-2018: My idea is to change from locking all hosts first to locking each host in turn for efficiency.

ls-2018: It is more like a CP model as currently implemented, and more like an AP model after the change. Because we don't need every DAPR client actor to have the same data at the same time, as long as he has data that the client can call

yaron2:
@ls-2018 tests are failing.

Issue #3934, opened 2021-11-22
Configuration per Actor Type


artursouza:
Actor configuration is global instead of per Actor Type. This is an issue when there are multiple actor types in an app and each one has different requirements for garbage collection.

This issue also includes SDK changes.

```csharp
// Today
public void ConfigureServices(IServiceCollection services)
{
    // Register actor runtime with DI
    services.AddActors(options =>
    {
        // Register actor types
        options.Actors.RegisterActor<MyActor>();
        
        // Configure default settings
        options.ActorIdleTimeout = TimeSpan.FromMinutes(10);
        options.ActorScanInterval = TimeSpan.FromSeconds(35);
        options.DrainOngoingCallTimeout = TimeSpan.FromSeconds(35);
        options.DrainRebalancedActors = true;
    });

    // Register additional services for use with actors
    services.AddSingleton<BankService>();
}
```

```csharp
// Proposed
public void ConfigureServices(IServiceCollection services)
{
    // Register actor runtime with DI
    services.AddActors(globalOptions =>
    {
        // Set override settings per actor type
        ActorOptions options = new ActorOptions();
        options.ActorIdleTimeout = TimeSpan.FromMinutes(1);
        options.ActorScanInterval = TimeSpan.FromSeconds(10);
        options.DrainOngoingCallTimeout = TimeSpan.FromSeconds(35);
        options.DrainRebalancedActors = true;

        // Register actor types and configure actor settings
        globalOptions.Actors.RegisterActor<MyActor>(options);
        
        // Configure default settings
        globalOptions.ActorIdleTimeout = TimeSpan.FromMinutes(10);
        globalOptions.ActorScanInterval = TimeSpan.FromSeconds(35);
        globalOptions.DrainOngoingCallTimeout = TimeSpan.FromSeconds(35);
        globalOptions.DrainRebalancedActors = true;
    });

    // Register additional services for use with actors
    services.AddSingleton<BankService>();
}
```

Optionally, the actor type configuration can come from class annotations. That is a decision per language SDK.

RELEASE NOTE: **ENABLES** actor options to be configured per actor type

halspang:
## Overview

As this is a non-versioned API, backwards compatibility will be important. As such, we have a couple ways of doing things:

1. Add a version field in the payload and create a new version of the object (mapping a v1 -> v2 internally)
2. Make this an additive change where we just introduce new fields
3. A combination of the two above

[...]

PR #4201, opened 2022-02-04
Allow per actor type configuration

halspang:
# Description

This commit allows actors to specify a type specific configuration.
This allows for actors in the same host to have different timeouts
and features enabled. The original config is still allowed and is
used a fallback if a type specific config is not specified.

## Issue reference

Please reference the issue this PR will close: #3934 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Unit tests passing
* [ ] Extended the documentation / https://github.com/dapr/docs/issues/2176


pkg/config/app_configuration.go
```diff
@@ -16,18 +16,32 @@ package config
...
 	ActorScanInterval string `json:"actorScanInterval"`
-	// Duration. example: "30s"
+	// Duration. example: "30s".
 	DrainOngoingCallTimeout    string           `json:"drainOngoingCallTimeout"`
 	DrainRebalancedActors      bool             `json:"drainRebalancedActors"`
 	Reentrancy                 ReentrancyConfig `json:"reentrancy,omitempty"`
 	RemindersStoragePartitions int              `json:"remindersStoragePartitions"`
+
+	// Duplicate of the above config so we can assign it to individual entities.
+	EntityConfigs []EntityConfig `json:"entitiesConfig,omitempty"`
```
artursouza, line 30: Please, validate if any of the entities in EntityConfig.entities is not present in entities. If so, IMO sidecar should give a warning for that type.

artursouza (changes requested):
Just a minor change.

@yaron2 any concerns over this? Please, pay attention to the new struct expected by dapr/config since that is not versioned and we don't want to break it now or in the future.

pkg/config/app_configuration.go
yaron2: Looks good.

artursouza (approved):

daixiang0 (approved):

Issue #4124, opened 2022-01-13
Nightly builds for Dapr


artursouza:
## In what area(s)?

> /area runtime

> /area operator

> /area placement

> /area docs

/area test-and-release
## Describe the proposal

As we add new features or APIs in master branch, SDKs need to have a local override solution to rebuild the master branch to run integration tests on changes that are not yet released in runtime. Instead, we should have a unified solution where Dapr is released nightly and SDK workflows can make use of that pre-built release instead. It can also enable users to try unreleased features way before the RC is cut.

We already have the workflow to release Dapr and Dapr RCs, so it would mean a relative small modification to have nightly builds too.

The current solution to install RCs should also work for nightly builds directly from the install script or the CLI.

Proposed tag: nightly-YYYYMMDD

Also, add a CLI feature where `nightly` version would automatically scan the nightly build releases in GH and use the latest.

berndverst:
This would be fantastic.

`dapr init --runtime-version=nightly` should install the latest nightly release. This would be pretty close to running against dapr/dapr@master, but without having to build daprd from scratch which is very slow.

Cowlord711:
I can try to work on this

yaron2:
> I can try to work on this

That's much appreciated. We're here to assist!

artursouza:
@Cowlord711 It is working! Thanks a lot for this contribution.

![image](https://user-images.githubusercontent.com/130954/158114854-7d0a551d-5325-440e-be59-1383c9424833.png)

lloydfischer:
I'm told, on Discord, that this no longer works

PR #4208, opened 2022-02-06
Add workflow for Nightly build

Cowlord711:
# Description

Adding new workflow file to create a nightly tag that should trigger dapr build workflow. Modified get-release-version.py and dapr.yaml accordingly to account for changes.

## Issue reference

Please reference the issue this PR will close: #_[4124](https://github.com/dapr/dapr/issues/4124)_

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing


yaron2:
@Cowlord711 are there any other changes you're expecting in this PR? (it's labeled as WIP)

Cowlord711:
> @Cowlord711 are there any other changes you're expecting in this PR? (it's labeled as WIP)

Was just looking for some initial feedback because I am new to open source. Was hoping to do some more testing and then see if there is a way to clean up older nightly releases.

Cowlord711:
Ready for review

yaron2:
ping for review @artursouza

.github/scripts/get_release_version.py
```diff
@@ -20,14 +20,25 @@
...
-    if gitRef is None or not gitRef.startswith(tagRefPrefix):
+    if gitRef is None or not gitRef.startswith((tagRefPrefix, nightlyTagPrefix)):
         githubEnv.write("REL_VERSION=edge\n")
+        githubEnv.write("RELEASE_TO_GH=False\n")
         print ("This is daily build from {}...".format(gitRef))
         sys.exit(0)
 
+    githubEnv.write("RELEASE_TO_GH=True\n")
+
+    if gitRef.find("nightly") > 0:
```
artursouza, line 35: We cannot have a full release for nightly build because of 2 reasons:
1. Helm chart needs to follow semantic versioning, nightly-X-Y-Z format is not a valid format for that.
2. Even if there is a workaround for (1), it would "explore" the number of helm chart entries in the index file: https://github.com/dapr/helm-charts/blob/master/index.yaml


My proposal is to only release the dapr images to Docker Registry and not release Helm charts as part of the nightly release. People that need to use any of the nightly images can do so by using the latest release Helm + changing the image of specific containers to use the nightly build.

artursouza (changes requested):

.github/scripts/get_release_version.py
Cowlord711: @artursouza I see that we already push an edge image to Docker after new changes are merged into master. I was wondering how would this nightly build docker image differ from that? If people want the most recent changes can they not use the edge image that is published?

artursouza: > @artursouza I see that we already push an edge image to Docker after new changes are merged into master. I was wondering how would this nightly build docker image differ from that? If people want the most recent changes can they not use the edge image that is published?

Yes, there is a big difference. The nightly build is a fixed point in time while the `edge` tag is a moving target. You can depend on a nightly build with guarantees it will never change. The `edge` tag is always changing. We, for example, plan to use nightly build to allow SDK builds to use a runtime version that is not yet released but not a moving target to keep the build reproducible. I hope that clarifies it.

Cowlord711: Yes that makes sense, thank you. I will make the changes

artursouza (approved):

Issue #2892, opened 2021-03-05
Add content type to state operations


rynowak:
## In what area(s)?

/area runtime

## Describe the feature

We should add content type to state operations to allow callers to specify the representation of the data they are providing. There are some state stores where it's valuable to provide guarantees about the data representation. Since the state store does not know what kind of content its getting in the common case (gRPC) it has to make the most general assumption about how to store it.

For example in a database like CosmosDB it would be useful to be specify that the data is JSON so that it can be stored without additional encoding for querying. We changed this to base64 encode late in 1.0 and I heard about it from a number of users in discord.

Issues:

- https://github.com/dapr/components-contrib/issues/665
- https://github.com/dapr/dotnet-sdk/issues/614

----

Currently we don't provide any information about the type of content to a state store component. It's not possible for a user to influence this other than through their choice of protocol.

The state store gets a value of type `interface{}` which depends on the protocol used:

-  HTTP: a Go value (an `int`, `string`, or `map[string]interface{}` that was unmarshalled from JSON
- gRPC: a `[]byte` value

**Note:** in the case of HTTP the caller MUST provide JSON as part of the request.

---- 

How this would work in practice:

- `.proto` messages get a content type field added for each state item
- state component interfaces get a content type field added as well - this will be `application/json` in the case of an HTTP request, or whatever the caller provided for gRPC
- SDKs that perform serialization will provide a content type
- the state component implementations can now choose the best way to represent data based on the content type
- state component implementations can now reject data that they are unable to store
- the state component implementations now return a content type as part of GET operations - SDKs can use this for diagnostics and validation

## Release Note

RELEASE NOTE: Add the ability to specify a content type for state operations

jigargandhi:
@rynowak @artursouza can metadata be used to specify content type of the payload, or does it have to be an additional field?

artursouza:
@jigargandhi I like the use of metadata in runtime API. Although the SDKs might decide to handle it as a first class attribute.

artursouza:
@jigargandhi I will assign it to you. Beware this is not an easy issue, and it has impacts across the codebase and must be done in a backwards compatible way. Also, consider the impact to the SDKs.

jigargandhi:
I got a feedback on the [discord](https://discord.com/channels/778680217417809931/901141713089863710/914206213254823946) that we should add content type to more components, however not all components require a content type. On looking at a list of components here are the component types that require a content-type

So *apart from state management*, here are the places I think require content type
[...]

PR #4210, opened 2022-02-07
state store: include URL query parameters in the metadata of HTTP API requests

dmitsh:
# Description

state store: include URL query parameters in the metadata of "save state" HTTP requests

## Issue reference

Please reference the issue this PR will close: #2892

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing
* [x] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#2186

### DIFF
--- a/pkg/http/api.go
+++ b/pkg/http/api.go
@@ -460,7 +460,15 @@ func (a *api) onBulkGetState(reqCtx *fasthttp.RequestCtx) {
 		return
 	}
 
+	// merge metadata from URL query parameters
 	metadata := getMetadataFromRequest(reqCtx)
+	if req.Metadata == nil {
+		req.Metadata = metadata
+	} else {
+		for k, v := range metadata {
+			req.Metadata[k] = v
+		}
+	}
 
 	bulkResp := make([]BulkGetResponse, len(req.Keys))
 	if len(req.Keys) == 0 {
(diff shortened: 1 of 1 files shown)


artursouza (commented):
This is an interesting idea. The body already have the metadata attribute per key, not? So, how would this work in a scenario where metadata from query params conflict? Later, can you create unit tests and E2E tests?

dmitsh:
Usage example:

Save data as JSON document:
```
curl -X POST http://localhost:3500/v1.0/state/starwars?metadata.contentType=application/json  \
  -H "Content-Type: application/json" \
  -d '[
        {
          "key": "weapon",
          "value": "DeathStar",
          "etag": "1234"
        },
        {
          "key": "planet",
          "value": {
            "name": "Tatooine"
          }
        }
      ]'
```

Retrieve data as JSON document:
```
curl -X GET http://localhost:3500/v1.0/state/starwars/planet?metadata.contentType=application/json
```

dmitsh:
> The body already have the metadata attribute per key, not? So, how would this work in a scenario where metadata from query params conflict?

We need to decide on that. I would suggest that URL query parameters take precedence, since they are set explicitly.

PR #4213, opened 2022-02-07
fix component exists check when updating

yaron2:
Fixes https://github.com/dapr/dapr/issues/4212.

Also, removed redundant function from `runtime.go` that wasn't used.

PR #4215, opened 2022-02-08
cherry pick fix component exists check when updating #4213

daixiang0:
hotfix


yaron2 (approved):

pkg/runtime/runtime.go
```diff
@@ -1632,10 +1639,6 @@ func (a *DaprRuntime) initActors() error {
 	return err
 }
 
-func (a *DaprRuntime) hostingActors() bool {
```
yaron2, line 1635: We can't remove this function for 1.5.1 since it's being used by `runtime.go` in that branch...

yaron2 (changes requested):

yaron2 (approved):

Issue #3920, opened 2021-11-18
Issue matching wildcard topics for pubsub


lullen:
## In what area(s)?

/area runtime

> /area operator

> /area placement

> /area docs

> /area test-and-release

## What version of Dapr?

1.3.1

## Expected Behavior

Dapr should call my application when an event is received on all matching topics.

## Actual Behavior

Dapr skipps processing event and gives the following message

`no matching route for event <nil> in pubsub mqtt-pubsub and topic app/device-simple/event; skipping`

## Steps to Reproduce the Problem

Publish an mqtt message to `app/device-simple/event` and then subscribe to `app/+/event`

Subscribing on the exact topic works:
`app/device-simple/event`

## Release Note

RELEASE NOTE: **FIX** Bug in runtime matching wildcard routes for pubsub.

yaron2:
@thielepaul can you comment here with `/assign` please? :)

artursouza:
Reopening since the E2E test does not pass on Windows and should not be specific to mqtt.

thielepaul:
> Reopening since the E2E test does not pass on Windows and should not be specific to mqtt.

Ok I see why you needed to revert the e2e test. Can we merge the fix to the problem anyway? https://github.com/dapr/dapr/pull/4259

I do not think this can be tested with redis streams as we use `XGroupCreateMkStream` for subscriptions which does not support wildcards.

artursouza:
Thanks for your contribution! When you have some time, it would be great to have E2E test for this.

PR #4220, opened 2022-02-08
add mqtt to e2e-pubsub-test and fix wildcard subscription

thielepaul:
Signed-off-by: Paul Thiele <thielepaul@gmail.com>

# Description

Added mosquitto as a mqtt message broker to the e2e tests and added a mqtt subscription to e2e pubsub test.

## Issue reference

Please reference the issue this PR will close: #3920

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Unit tests passing

### DIFF
--- a/tests/apps/pubsub-publisher/app.go
+++ b/tests/apps/pubsub-publisher/app.go
@@ -44,6 +44,7 @@ type publishCommand struct {
 	Data        interface{}       `json:"data"`
 	Protocol    string            `json:"protocol"`
 	Metadata    map[string]string `json:"metadata"`
+	PubSubName  string            `json:"pubsubname"`
 }
 
 type appResponse struct {
@@ -85,12 +86,16 @@ func performPublish(w http.ResponseWriter, r *http.Request) {
 		})
 		return
 	}
+	if len(commandBody.PubSubName) == 0 {
+		commandBody.PubSubName = pubsubName
+	}
 
 	log.Printf("    commandBody.ContentType=%s", commandBody.ContentType)
 	log.Printf("    commandBody.Topic=%s", commandBody.Topic)
 	log.Printf("    commandBody.Data=%v", commandBody.Data)
 	log.Printf("    commandBody.Protocol=%s", commandBody.Protocol)
 	log.Printf("    commandBody.Metadata=%s", commandBody.Metadata)
+	log.Printf("    commandBody.PubSubName=%s", commandBody.PubSubName)
 
 	// based on commandBody.Topic, send to the appropriate topic
 	resp := appResponse{Message: fmt.Sprintf("%s is not supported", commandBody.Topic)}
@@ -115,9 +120,9 @@ func performPublish(w http.ResponseWriter, r *http.Request) {
 	// publish to dapr
 	var status int
 	if commandBody.Protocol == "grpc" {
-		status, err = performPublishGRPC(commandBody.Topic, jsonValue, contentType, commandBody.Metadata)
+		status, err = performPublishGRPC(commandBody.Topic, jsonValue, contentType, commandBody)
 	} else {
-		status, err = performPublishHTTP(commandBody.Topic, jsonValue, contentType, commandBody.Metadata)
+		status, err = performPublishHTTP(commandBody.Topic, jsonValue, contentType, commandBody)
 	}
 
 	if err != nil {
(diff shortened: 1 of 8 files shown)


thielepaul:
@artursouza can you trigger the e2e tests? The pubsub test should fail, and then pass after adding this: https://github.com/thielepaul/dapr/pull/1

thielepaul:
@maintainers-dapr please approve the test workflow :)

yaron2 (approved):
lgtm

artursouza:
I am reverting this change because it broke the test on Windows and the test should not be specific to a pubsub component.

yaron2:
> I am reverting this change because it broke the test on Windows and the test should not be specific to a pubsub component.

Can we fix instead of revert?

tests/apps/pubsub-publisher/app.go
```diff
@@ -85,12 +86,16 @@ func performPublish(w http.ResponseWriter, r *http.Request) {
...
+	if len(commandBody.PubSubName) == 0 {
+		commandBody.PubSubName = pubsubName
+	}
 
 	log.Printf("    commandBody.ContentType=%s", commandBody.ContentType)
 	log.Printf("    commandBody.Topic=%s", commandBody.Topic)
 	log.Printf("    commandBody.Data=%v", commandBody.Data)
 	log.Printf("    commandBody.Protocol=%s", commandBody.Protocol)
 	log.Printf("    commandBody.Metadata=%s", commandBody.Metadata)
+	log.Printf("    commandBody.PubSubName=%s", commandBody.PubSubName)
```
artursouza, line 98: Please, do not make this test aware of pubsub. We keep the test generic.

tests/config/dapr_mqtt_pubsub.yaml
```diff
@@ -0,0 +1,18 @@
+apiVersion: dapr.io/v1alpha1
+kind: Component
+metadata:
+  name: mqtt-pubsub
+spec:
+  type: pubsub.mqtt
+  version: v1
+  metadata:
+  - name: url
+    value: "tcp://mosquitto.dapr-tests.svc.cluster.local:1883"
```
artursouza, line 10: Why bringing MQTT to the test?

tests/dapr_tests.mk
```diff
@@ -313,6 +313,8 @@ setup-test-components: setup-app-configurations
 	$(KUBECTL) apply -f ./tests/config/app_topic_subscription_routing.yaml --namespace $(DAPR_TEST_NAMESPACE)
 	$(KUBECTL) apply -f ./tests/config/app_topic_subscription_routing_grpc.yaml --namespace $(DAPR_TEST_NAMESPACE)
 	$(KUBECTL) apply -f ./tests/config/app_pubsub_routing.yaml --namespace $(DAPR_TEST_NAMESPACE)
+	$(KUBECTL) apply -f ./tests/config/mosquitto.yaml --namespace $(DAPR_TEST_NAMESPACE)
```
artursouza, line 316: Please, keep the test independent of mqtt, see how we have DAPR_TEST_PUBSUB as an env var to configure which pubsub to use for E2E tests. The e2e tests should pass independent of a component. For component specific test, we have certification tests in components-contrib.

artursouza:
> > I am reverting this change because it broke the test on Windows and the test should not be specific to a pubsub component.
> 
> Can we fix instead of revert?

No, it is a rewrite.

artursouza:
I know we don't do this for every PR, but in case of E2E test related PRs, we must run ok-to-test to make sure it passes on Windows too.

tests/config/dapr_mqtt_pubsub.yaml
thielepaul: while redis does support wildcards when using `psubscribe`, dapr uses `XGroupCreateMkStream` (https://github.com/dapr/components-contrib/blob/6e4139c7b1da21ea4a5d14d2e323a7ff37ee2d9f/pubsub/redis/redis.go#L172), which does not support wildcards.

tests/dapr_tests.mk
thielepaul: I think it is not really mqtt specific, as there are other brokers supported by dapr, which also support wildcards in the topics, but redis does not, so I am not sure what would be the best solution here.
If we move the test to components-contrib, this has the disadvantage, that we won't notice it if a PR to the main dapr repo breaks this again.

tests/apps/pubsub-publisher/app.go
thielepaul: yes, I agree why this is preferable. Maybe we can run the test for all supported brokers and make it somehow configurable, as not all brokers support the same features.

Issue #4249, opened 2022-02-15
redis state store: implement state query API and add e2e tests


dmitsh:
## In what area(s)?

/area runtime
/area docs
/area test-and-release

## Describe the feature

redis state store: implement state query API and add e2e tests

## Release Note

RELEASE NOTE:
**ADD** state store query API for Redis

PR #4251, opened 2022-02-15
redis state store: implement state query API and add e2e tests

dmitsh:
# Description

redis state store: implement state query API and add e2e tests

## Issue reference

Please reference the issue this PR will close: #4249

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing


tests/dapr_tests.mk
```diff
@@ -309,6 +309,7 @@ setup-test-components: setup-app-configurations
 	$(KUBECTL) apply -f ./tests/config/app_topic_subscription_pubsub_grpc.yaml --namespace $(DAPR_TEST_NAMESPACE)
 	$(KUBECTL) apply -f ./tests/config/kubernetes_allowlists_config.yaml --namespace $(DAPR_TEST_NAMESPACE)
 	$(KUBECTL) apply -f ./tests/config/kubernetes_allowlists_grpc_config.yaml --namespace $(DAPR_TEST_NAMESPACE)
+	$(KUBECTL) apply -f ./tests/config/dapr_redis_state_query.yaml --namespace $(DAPR_TEST_NAMESPACE)
```
artursouza, line 312: Can we use the existing DAPR_TEST_QUERY_STATE_STORE attribute to use redis instead of creating a new one?

tests/e2e/stateapp/stateapp_test.go
```diff
@@ -608,38 +608,58 @@ func TestQueryStateStore(t *testing.T) {
...
-		keys int
+	for _, store := range []struct {
+		name    string
+		isRedis bool
 	}{
 		{
-			path: "./query-data/query.json",
-			keys: 3,
+			name:    "querystatestore",
+			isRedis: false,
```
artursouza, line 617: Same comment as in the previous PR, please, just add the content type all the time and we can remove any redis specific reference here.

tests/e2e/stateapp/stateapp_test.go
```diff
@@ -608,38 +608,58 @@ func TestQueryStateStore(t *testing.T) {
...
+		}
+		for _, test := range tests {
+			body, err := os.ReadFile(test.path)
 			require.NoError(t, err)
-			require.Equal(t, test.keys, len(states.States))
+
+			for _, protocol := range []string{"http", "grpc"} {
+				url := fmt.Sprintf("%s/test/%s/query/%s", externalURL, protocol, store.name)
+				if store.isRedis {
+					url += "?contentType=application/json&queryIndexName=orgIndx"
```
artursouza, line 652: I think it is OK to add set the queryIndexName here. Can these two be moved to the component metadata, so we don't need this to be in the request? The fact that our E2E test now must know about being redis or not is a sign that we are breaking portability.

tests/dapr_tests.mk
dmitsh: DAPR_TEST_QUERY_STATE_STORE is used to alternate between MongoDB and CosmosDB.
The Redis query e2e test is an additional test.

artursouza: E2E tests are not component specific.

dmitsh: No problem, I will replace mongodb with redis

tests/e2e/stateapp/stateapp_test.go
```diff
@@ -608,38 +608,40 @@ func TestQueryStateStore(t *testing.T) {
...
 
-	// Populate store.
-	url := fmt.Sprintf("%s/test/%s/%s/querystatestore", externalURL, "http", "load")
-	body, err := os.ReadFile("query-data/dataset.json")
-	require.NoError(t, err)
-
-	_, status, err := utils.HTTPPostWithStatus(url, body)
-	require.NoError(t, err)
-	require.Equal(t, 204, status)
+	for _, storename := range []string{"querystatestore", "redisquerystore"} {
```
artursouza, line 611: Please, keep the test store independent.

artursouza (changes requested):

tests/dapr_tests.mk
artursouza: We can keep a mix of both and use both in our kind tests.

dmitsh: > We can keep a mix of both and use both in our kind tests.

I don't think so. We only can have one component with the name `querystatestore`, and we already alternate it between AKS (CosmosDB) and KIND (MongoDB) e2e tests. Beside, if e2e tests are component-agnostic, it should not matter which component is used.

artursouza: Running E2E tests with the same code but against different components can help uncover inconsistencies. I will look into that later. For now, let's just keep one configured component for query state. I will look into doing a combination of those later for Kind. Thanks.

dmitsh: > Running E2E tests with the same code but against different components can help uncover inconsistencies. I will look into that later. For now, let's just keep one configured component for query state. I will look into doing a combination of those later for Kind. Thanks.

I actually agree that testing more components is beneficial. This is why I added Redis component to the existing ones ...

artursouza: My point is more in the how. The idea is that we keep the code component independent and configure the different components outside the test code. Similar to how we switch the state store between redis and cosmosdb today.

dmitsh: I made a cosmetic change in an attempt to keep both Redis and MongoDB query tests. LMK.

tests/e2e/stateapp/stateapp_test.go
```diff
@@ -608,38 +608,40 @@ func TestQueryStateStore(t *testing.T) {
...
 
-	// Populate store.
-	url := fmt.Sprintf("%s/test/%s/%s/querystatestore", externalURL, "http", "load")
-	body, err := os.ReadFile("query-data/dataset.json")
-	require.NoError(t, err)
-
-	_, status, err := utils.HTTPPostWithStatus(url, body)
-	require.NoError(t, err)
-	require.Equal(t, 204, status)
+	for _, storename := range []string{"querystatestore", "querystatestore2"} {
```
artursouza, line 611: I think we can do it with 1 state store. But we can merge this one for now.

artursouza (approved):

PR #4254, opened 2022-02-16
Add resiliency to state APIs

halspang:
# Description

Add resiliency to state APIs.

## Issue reference

## Checklist

* [x] Code compiles correctly
* [] Created/updated tests
* [x] Unit tests passing

### DIFF
--- a/pkg/grpc/api.go
+++ b/pkg/grpc/api.go
@@ -405,7 +405,14 @@ func (a *api) GetBulkState(ctx context.Context, in *runtimev1pb.GetBulkStateRequ
 		}
 		reqs[i] = r
 	}
-	bulkGet, responses, err := store.BulkGet(reqs)
+
+	policy := a.resiliency.ComponentPolicy(ctx, in.StoreName)
+	var bulkGet bool
+	var responses []state.BulkGetResponse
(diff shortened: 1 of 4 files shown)


mukundansundar (commented):
@halspang How do we make sure that the operations are idempotent for set, delete and receptive bulk operations?

pkedy (approved):

halspang:
> @halspang How do we make sure that the operations are idempotent for set, delete and receptive bulk operations?

Idempotency tends to come from either the user app, or the underlying statestore. This is a common concern in resiliency and I think is honestly something solved with a big warning flag on the documentation.

We have the option of introducing that as a feature flag to the statestore components, but I don't think it's something we'd capture here. 

What do you think?

halspang:
> There is a failing test, but I approve assuming that is resolved.

I think I just made the tolerance for the test too low for non-local runs. 5ms is pretty short in a github action. I'll increase the window for success.

daixiang0:
> > There is a failing test, but I approve assuming that is resolved.
> 
> I think I just made the tolerance for the test too low for non-local runs. 5ms is pretty short in a github action. I'll increase the window for success.

Is it better to set different value for different env, like local uses 5ms, CI use 100ms?

Issue #4065, opened 2021-12-29
Proposal: support multi nameresolution


1046102779:
At present, the nameresolution of the Dapr community is only used for service discovery between Dapr Runtimes, and the configuration initialization is completed through `a.globalConfig.Spec.NameResolutionSpec.Component` in the global config configuration;

However, when the business team uses binding external resources or accesses other external resources, it also needs to use nameresolution, and it should also support multi nameresolution instances. How to use it depends on the nameresolution instance configuration of the business team, and the addressing TargetID of each resource , And proceed with processing.

So we should remove the a.globalConfig.Spec.NameResolutionSpec.Component global configuration, and then directly use the nameresolution instance list in the component to initialize, and store it in map[componentName]nameResolution for processing separately

At the same time, one more thing to note:

If we use map to store multi nameresolution, there is currently only one communication between Dapr Runtimes, and there is no choice; for multiple nameresolutions, we need to support parameter input to specify who should be selected for communication between Dapr Runtimes?

1046102779:
https://github.com/dapr/components-contrib/issues/1398

ls-2018:
Domain name resolution is currently only available in Direct Message. In K8S, based on the current messaging mechanism, is there any other way to resolve domain names?

1046102779:
https://github.com/polarismesh/polaris
https://github.com/hashicorp/consul
and etc.

ls-2018:
Okay, I see what you're saying.

I see that the components library has an implementation of consul, but ```initNameResolution```  has some problems

daixiang0:
Something like https://github.com/dapr/components-contrib/issues/1358

/cc @artursouza @yaron2

artursouza:
Can someone provide an example of how this would work? I am interested to see how components, like Kafka or Azure CosmosDB would benefit from a custom DNS resolution (if possible).

daixiang0:
ping @1046102779

1046102779:
@daixiang0 @artursouza @yaron2 

Example after we transform dapr, as follow:

# component
## bindings.http
specify nameresolution:  polaris
<img width="495" alt="wecom-temp-f5bd6c9d79adf30247c05015e7059f63" src="https://user-images.githubusercontent.com/20457624/150056993-3406910f-b75f-4907-b637-8f3ec174e68d.png">

## nameresolution.polaris
[...]

PR #4264, opened 2022-02-17
Dns policy

ls-2018:
# Description
If DNSPolicy is not ClusterFirst, the dapr sidecar may not able to connect to other dapr sidecar.
 
## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests
* [x] Unit tests passing


ls-2018:
It's not the same question. I'm going to give you an example to prove that the change is correct.
I have two programs `foo`, `bar`, where `foo` calls `bar`, <s>but it seems that the logic is not used.</s>
As you can see, when we set the pod's dns policy to Default, the daprd container will never be ready, let alone the routing behind it.There is nothing wrong with the logic of the program itself.In this case, it should not be injected.

### bar.py
```
import time

from flask import Flask

app = Flask("___")


@app.route('/')
@app.route('/index')
def index():
    return "bar:" + str(time.time())


if __name__ == '__main__':
    app.run(port=8000)
```
### foo.py
```
import requests

from flask import Flask

app = Flask("___")


@app.route('/')
def index():
    try:
        return requests.get(f"http://localhost:3500/v1.0/invoke/bar/method/index").text
    except Exception as e:
        return str(e)


if __name__ == '__main__':
    app.run(port=8000)

```

### pod.yaml
```
---
kind: Deployment
apiVersion: apps/v1
metadata:
  name: foo
spec:
  template:
    metadata:
      annotations:
        dapr.io/enabled: 'true'
        dapr.io/app-id: foo
        dapr.io/app-port: '8000'
      labels:
        app: foo
    spec:
      dnsPolicy: "Default"
      containers:
        - name: foo
          image: acejilam/example:latest
          args:
            - "/app/foo.py"
  selector:
    matchLabels:
      app: foo
---
kind: Deployment
apiVersion: apps/v1
metadata:
  name: bar
spec:
  template:
    metadata:
      annotations:
        dapr.io/enabled: 'true'
        dapr.io/app-id: bar
        dapr.io/app-port: '8000'
      labels:
        app: bar
    spec:
      containers:
        - name: bar
          image: acejilam/example:latest
          args:
            - "/app/bar.py"
  selector:
    matchLabels:
      app: bar

```

### Dockerfile
```
#FROM python:3.9.6-slim as base
FROM centos:8 as base
RUN #sed -i 's#http://deb.debian.org#https://mirrors.aliyun.com#g' /etc/apt/sources.list
RUN #sed -i 's#http://security.debian.org#https://mirrors.aliyun.com#g' /etc/apt/sources.list
RUN rm -f /etc/localtime && ln -sv /usr/share/zoneinfo/Asia/Shanghai /etc/localtime && echo "Asia/Shanghai" > /etc/timezone
RUN yum install python3.9 epel-release  lsof -y && yum install htop -y
RUN #apt-get update && apt-get install vim -y
#mirrors.aliyun.com|mirror.lzu.edu.cn|mirrors.ustc.edu.cn|mirrors.cloud.tencent.com
# ---- compile image -----------------------------------------------
FROM base AS compile-image
RUN python3 -m pip install --upgrade pip
RUN python3 -m venv /venv
RUN #apt-get update -y && apt-get install -y build-essential
# pip install is fast here (while slow without the venv) :
RUN pip3 config set global.index-url https://pypi.tuna.tsinghua.edu.cn/simple
RUN /venv/bin/python3 -m pip install --upgrade pip
RUN /venv/bin/pip3 install flask requests
# ---- build image -----------------------------------------------
FROM base AS build-image
COPY --from=compile-image /venv /venv
## Make sure we use the virtualenv:
COPY . /app
WORKDIR /app
ENV PATH="/venv/bin:$PATH"
#RUN . /venv/bin/activate
ENV LC_ALL zh_CN.UTF-8
ENV LANG zh_CN.UTF-8
EXPOSE 8000
ENTRYPOINT ["/venv/bin/python3"]

```

### test.sh
```
docker build -t acejilam/example:latest .
docker push acejilam/example:latest
kind delete cluster
echo -e 'kind: Cluster
apiVersion: kind.x-k8s.io/v1alpha4
nodes:
- role: control-plane
  extraPortMappings:
  - containerPort: 80
    hostPort: 80
    listenAddress: "0.0.0.0" # Optional, defaults to "0.0.0.0"
    protocol: udp # Optional, defaults to tcp
- role: worker
' > ./kind.yaml

kind create cluster --config ./kind.yaml
kubectl cluster-info --context kind-kind
dapr init -k
kubectl apply -f pod.yaml
sleep 300
# shellcheck disable=SC2046
kubectl logs $(kubectl get pod|grep foo|awk '{print $1}') -c daprd
#kubectl exec -it $(kubectl get pod|grep Running|grep foo|awk '{print $1}') -c foo -- curl http://127.0.0.1:8000/
```
image
image

pkg/injector/injector.go
```diff
@@ -224,6 +234,7 @@ func (i *injector) handleRequest(w http.ResponseWriter, r *http.Request) {
 		}
 	}
 
+SkipPatch:
```
daixiang0, line 237: I really do not like any `goto` codes, could we use other way to do it?

daixiang0:
If we add the DNS component mentioned in #4065, the policy-related checks should be in that component.

If not, we need this check to ensure Dapr sidecar works.

/cc @artursouza

ls-2018:
I then add some data to illustrate the difference between the #4264  #4065

#4264：

image
image
image
 
We can see that foo Container based on 'dnsPolicy: "Default" configuration cannot resolve the k8S internal service name. Because he hasn't been resolved by CoreDNS.
In the case of 'DirectMessage', there is communication through the K8S service name, but in this case there is no communication at all. Other components based on K8S domain name communication will be affected.
image


#4065:
This proposal is intended to support communication between different components using different name resolution components.
For example, Foo and Bar communicate with each other using Consul.
DNS communication is used between foo2 and bar2.
However, no matter which parsing component is used, a service is required to implement this functionality.
If DAPRD and the service communicate via IP :port, then this case is useless.
However, if it is resolved by domain name, how can it be implemented in K8S cluster without coreDNS?【The external DNS of the cluster is excluded】

pkg/injector/injector.go
ls-2018: I'll change it.

pkg/injector/injector.go
```diff
@@ -209,9 +209,21 @@ func (i *injector) handleRequest(w http.ResponseWriter, r *http.Request) {
...
 	ar := v1.AdmissionReview{}
 	_, gvk, err := i.deserializer.Decode(body, nil, &ar)
+	isDNSClusterFirst := true
 	if err != nil {
 		log.Errorf("Can't decode body: %v", err)
 	} else {
+		var pod corev1.Pod
+		if err = json.Unmarshal(ar.Request.Object.Raw, &pod); err != nil {
+			log.Warnf("could not unmarshal raw object: %v", err)
+		} else {
```
daixiang0, line 219: Codes look good, just a nit: use `elsif` here.

ls-2018: 👌🏻

yaron2:
@ls-2018 please fix the failing checks.

ls-2018:
golangci-lint run -c .golangci.yml --fix

PR #4275, opened 2022-02-19
Add resiliency to service invocation

halspang:
# Description

Add resiliency to service invocation.

## Issue reference

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Unit tests passing

### DIFF
--- a/pkg/grpc/proxy/handler.go
+++ b/pkg/grpc/proxy/handler.go
@@ -4,14 +4,18 @@
 package proxy
 
 import (
+	"fmt"
 	"io"
 
 	"golang.org/x/net/context"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/metadata"
 	"google.golang.org/grpc/status"
 
+	"github.com/dapr/dapr/pkg/diagnostics"
 	"github.com/dapr/dapr/pkg/grpc/proxy/codec"
+	"github.com/dapr/dapr/pkg/resiliency"
 )
 
 var clientStreamDescForProxying = &grpc.StreamDesc{
@@ -23,8 +27,11 @@ var clientStreamDescForProxying = &grpc.StreamDesc{
 // The behaviour is the same as if you were registering a handler method, e.g. from a codegenerated pb.go file.
 //
 // This can *only* be used if the `server` also uses grpcproxy.CodecForServer() ServerOption.
-func RegisterService(server *grpc.Server, director StreamDirector, serviceName string, methodNames ...string) {
-	streamer := &handler{director}
+func RegisterService(server *grpc.Server, director StreamDirector, resiliency resiliency.Provider, serviceName string, methodNames ...string) {
+	streamer := &handler{
+		director:   director,
+		resiliency: resiliency,
+	}
 	fakeDesc := &grpc.ServiceDesc{
 		ServiceName: serviceName,
 		HandlerType: (*interface{})(nil),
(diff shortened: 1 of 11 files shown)


yaron2:
This needs to cover the gRPC proxy path too.

yaron2:
> > This needs to cover the gRPC proxy path too.
> 
> Yup, I was going to include it in a separate PR but I can also add it here since it's just another version of service invocation.

Better to add here

halspang:
@yaron2 - Can you recommend where to put the retries for proxying? I tried in `handler.go`, which seemed to be the natural case. However, when I retried within the `forward_to_` methods, the retry never seemed to trigger a request. When I retried the around the entire handler functionality, I kept getting errors like the following for each retry:

```
== APP == fail: Grpc.AspNetCore.Server.ServerCallHandler[14]                                                                                                                                                                                                
== APP ==       Error reading message.                                                                                                                                                                                                                      
== APP == Grpc.Core.RpcException: Status(StatusCode=Internal, Detail="Incomplete message.")  
```

I thought this may be because I was closing the stream, but I have the stream construction within the retry. Any direction here would be appreciated!

yaron2:
> @yaron2 - Can you recommend where to put the retries for proxying? I tried in `handler.go`, which seemed to be the natural case. However, when I retried within the `forward_to_` methods, the retry never seemed to trigger a request. When I retried the around the entire handler functionality, I kept getting errors like the following for each retry:
> 
> ```
> == APP == fail: Grpc.AspNetCore.Server.ServerCallHandler[14]                                                                                                                                                                                                
> == APP ==       Error reading message.                                                                                                                                                                                                                      
> == APP == Grpc.Core.RpcException: Status(StatusCode=Internal, Detail="Incomplete message.")  
> ```
> 
> I thought this may be because I was closing the stream, but I have the stream construction within the retry. Any direction here would be appreciated!

Can you try redialing the connection?

yaron2:
> @yaron2 - Can you recommend where to put the retries for proxying? I tried in `handler.go`, which seemed to be the natural case. However, when I retried within the `forward_to_` methods, the retry never seemed to trigger a request. When I retried the around the entire handler functionality, I kept getting errors like the following for each retry:
> 
> ```
> == APP == fail: Grpc.AspNetCore.Server.ServerCallHandler[14]                                                                                                                                                                                                
> == APP ==       Error reading message.                                                                                                                                                                                                                      
> == APP == Grpc.Core.RpcException: Status(StatusCode=Internal, Detail="Incomplete message.")  
> ```
> 
> I thought this may be because I was closing the stream, but I have the stream construction within the retry. Any direction here would be appreciated!

I believe this is where you should be retrying: https://github.com/dapr/dapr/blob/c503710ab2d12de04aae0c86f3031cb187cc453e/pkg/grpc/proxy/handler.go#L107.

Instead of returning the error, you should (most likely) literally `goto` here: https://github.com/dapr/dapr/blob/c503710ab2d12de04aae0c86f3031cb187cc453e/pkg/grpc/proxy/handler.go#L63.

halspang:
@yaron2 - I tried your suggestion and confirmed that we're going through the reconnection path. Still getting the same error though. So we can browse through it, I've put the code up here.

pkg/grpc/proxy/handler.go
```diff
@@ -148,13 +169,22 @@ func (s *handler) forwardServerToClient(src grpc.ServerStream, dst grpc.ClientSt
...
+					break
+				}
+			}
+		}
 		for i := 0; ; i++ {
 			if err := src.RecvMsg(f); err != nil {
 				ret <- err // this can be io.EOF which is happy case
 				break
 			}
+			s.bufferedCalls = append(s.bufferedCalls, f)
```
yaron2, line 186: 1. Isn't this a memory leak? we keep adding frames but never cleaning them up.
2. Shouldn't a call be added to `bufferedCalls` only if there's an error?

halspang: We do reset `s.bufferedCalls` at the start of the call, but we could also clean it after we're done.

Regardless, this is actually how the grpc middleware does this (albeit with locking): https://github.com/grpc-ecosystem/go-grpc-middleware/blob/560829fc74fcf9a69b7ab01d484f8b8961dc734b/retry/retry.go#L158

yaron2: Lets clean it as soon as we're done, please :)

halspang: Done!

artursouza:
I will approve after test pass.

artursouza (approved):

pkg/grpc/proxy/handler.go
```diff
@@ -46,70 +53,98 @@ func RegisterService(server *grpc.Server, director StreamDirector, serviceName s
...
-	}
 
-	clientCtx, clientCancel := context.WithCancel(outgoingCtx)
-	// TODO(mwitkow): Add a `forwarded` header to metadata, https://en.wikipedia.org/wiki/X-Forwarded-For.
-	clientStream, err := grpc.NewClientStream(clientCtx, clientStreamDescForProxying, backendConn, fullMethodName, grpc.CallContentSubtype((&codec.Proxy{}).Name()))
-	if err != nil {
-		return err
+	// Fetch the AppId so we can reference it for resiliency.
+	md, _ := metadata.FromIncomingContext(serverStream.Context())
+	v := md.Get(diagnostics.GRPCProxyAppIDKey)
```
yaron2, line 85: Can we move that to the section where a resiliency policy is used?

yaron2 (changes requested):

pkg/grpc/proxy/handler.go
halspang: We need it to get the resiliency policy, so it has to be before the policy. Or am I misunderstanding what you're asking?

yaron2: Yeah I realize it's needed for the policy.

Issue #4119, opened 2022-01-11
Log Dapr API calls


artursouza:
## In what area(s)?

On every call into the Sidecar or into the App, a log message should be printed for both gRPC and HTTP. I propose this to be have DEBUG verbosity.

> /area runtime

> /area operator

> /area placement

> /area docs

> /area test-and-release

## Describe the feature

Display API calls to/from Dapr sidecar.

## Release Note

RELEASE NOTE: **ADDED** logging for Dapr API calls

yaron2:
This was discussed in the past and decided against. It's way too verbose and doesn't provide any value. We currently log API calls on Debug level.

Info would create an IO bottleneck,.degrade performance and create noisy output for user's log sinks.

yaron2:
Removed from 1.7 until further discussed.

yaron2:
@artursouza please update the proposal to `Debug` verbosity, then we can put in 1.7 as we'll have consensus.

yaron2:
As for the implementation, ideally this can be accomplished as a global HTTP handler for the HTTP API and as a gRPC interceptor for gRPC, instead of per endpoint/method.

artursouza:
@yaron2 @msfussell would it make sense to have this log level configurable with DEBUG as default, so some developers might want to have this at INFO and not DEBUG where the performance hit is not critical to their application. I am saying this because I have many past experiences where logging API calls were set as INFO and critical to debug production issues. Changing to DEBUG to force the API log output after the fact of the incident would not add value.

yaron2:
> @yaron2 @msfussell would it make sense to have this log level configurable with DEBUG as default, so some developers might want to have this at INFO and not DEBUG where the performance hit is not critical to their application. I am saying this because I have many past experiences where logging API calls were set as INFO and critical to debug production issues. Changing to DEBUG to force the API log output after the fact of the incident would not add value.

Fine by me.

msfussell:
@amulyavarote - can you create a docs issue for this to track. The docs issues should contain examples on how to use this.

PR #4288, opened 2022-02-23
Add feature to log Dapr API calls

amulyavarote:
Signed-off-by: Amulya Varote <amulyavarote@microsoft.com>

# Description

Added feature to log Dapr API calls in both http and gRPC.

## Issue reference

Please reference the issue this PR will close: #4119 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Unit tests passing

### DIFF
--- a/pkg/grpc/server.go
+++ b/pkg/grpc/server.go
@@ -307,7 +307,7 @@ func shouldRenewCert(certExpiryDate time.Time, certDuration time.Duration) bool 
 
 func (s *server) getGRPCAPILogging() grpc_go.UnaryServerInterceptor {
 	return func(ctx context.Context, req interface{}, info *grpc_go.UnaryServerInfo, handler grpc_go.UnaryHandler) (interface{}, error) {
-		s.logger.Infof("Dapr gRPC API logging: , %s", info)
+		s.logger.Debugf("Dapr gRPC API logging: , %s", info)
 		return handler(ctx, req)
 	}
 }
--- a/pkg/grpc/server.go
+++ b/pkg/grpc/server.go
@@ -307,7 +307,7 @@ func shouldRenewCert(certExpiryDate time.Time, certDuration time.Duration) bool 
 
 func (s *server) getGRPCAPILogging() grpc_go.UnaryServerInterceptor {
 	return func(ctx context.Context, req interface{}, info *grpc_go.UnaryServerInfo, handler grpc_go.UnaryHandler) (interface{}, error) {
-		s.logger.Debugf("Dapr gRPC API logging: , %s", info)
+		s.logger.Infof("Dapr gRPC API logging: , %s", info)
 		return handler(ctx, req)
 	}
 }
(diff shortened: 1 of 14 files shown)


yaron2:
Thanks, we need to see how this affects service invocation performance (if at all).

pkg/http/server.go
```diff
@@ -208,6 +209,13 @@ func (s *server) useMetrics(next fasthttp.RequestHandler) fasthttp.RequestHandle
 	return next
 }
 
+func (s *server) apiLogging(next fasthttp.RequestHandler) fasthttp.RequestHandler {
+	return func(ctx *fasthttp.RequestCtx) {
+		log.Debugf("Dapr http API logging: , %s", string(ctx.Path()))
```
artursouza, line 214: For HTTP, in addition to path, also include the HTTP Method (GET, POST, ETC).

Please, use a simpler format: "HTTP API Called: %s %s"

/cc @yaron2  @msfussell

pkg/grpc/server.go
```diff
@@ -301,3 +304,10 @@ func shouldRenewCert(certExpiryDate time.Time, certDuration time.Duration) bool
 	percentagePassed := 100 - ((expiresInSeconds / certDurationSeconds) * 100)
 	return percentagePassed >= renewWhenPercentagePassed
 }
+
+func (s *server) getGRPCAPILogging() grpc_go.UnaryServerInterceptor {
+	return func(ctx context.Context, req interface{}, info *grpc_go.UnaryServerInfo, handler grpc_go.UnaryHandler) (interface{}, error) {
+		s.logger.Debugf("Dapr gRPC API logging: , %s", info)
```
artursouza, line 310: Please, use "gRPC API Called: %s"

artursouza (changes requested):
To make the log level configurable, please use a command line argument and annotation, similar to `log-level`. For this one, we might want to call `api-log-level`. See how the command line flag and K8s annotations are all defined here: https://docs.dapr.io/reference/arguments-annotations-overview/

artursouza:
> Thanks, we need to see how this affects service invocation performance (if at all).

We can also have a log level "off" by default and it would not have any performance impact. This way, there is no code executed apart from the if check.

pkg/grpc/server.go
```diff
@@ -218,6 +219,8 @@ func (s *server) getMiddlewareOptions() []grpc_go.ServerOption {
 		intr = append(intr, diag.DefaultGRPCMonitoring.UnaryServerInterceptor())
 	}
 
+	intr = append(intr, s.getGRPCAPILogging())
```
artursouza, line 222: If `api-log-level` is "off" or "", we don't even need to execute this line. Same for HTTP.

artursouza (changes requested):

msfussell:
@amulyavarote - can you create a docs issue for this to track. The docs issues should contain examples on how to use this.

artursouza (approved):
Please, cut a PR in docs too.

Issue #4292, opened 2022-02-24
Helm install dapr fails when running e2e test locally


CrazyHZM:
## In what area(s)?

> /area test-and-release

## What version of Dapr?

> 1.1.x
> 1.0.x

## Actual Behavior

Error: validation: chart.metadata.version "DAPR_VERSION_HELM" is invalid
make: *** [docker-deploy-k8s] Error 1

## Steps to Reproduce the Problem

make docker-deploy-k8s
## Release Note

RELEASE NOTE:
FIX helm install dapr fails when running e2e test locally

CrazyHZM:
Because `DAPR_VERSION_HELM` and `DAPR_VERSION_TAG` will not be replaced when e2e is executed locally.

CrazyHZM:
Resolved at https://github.com/dapr/dapr/pull/4334

PR #4293, opened 2022-02-24
FIX helm install dapr fails when running e2e test locally

CrazyHZM:
Signed-off-by: crazyhzm <crazyhzm@gmail.com>

# Description

## Issue reference

Fix https://github.com/dapr/dapr/issues/4292

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing


tests/docs/running-e2e-test.md
```diff
@@ -90,8 +95,14 @@ make docker-build
...
 make docker-push
 
+# Replace version placeholders in helm configuration files
+sh .github/scripts/set_helm_dapr_version.sh
+
 # Deploy Dapr runtime to your cluster
 make docker-deploy-k8s
+
+# Revert version placeholders in helm config files
+sh .github/scripts/revert_helm_dapr_version.sh
```
daixiang0, line 105: `git checkout` would revert changes, still need this?

CrazyHZM: When debugging locally, because you need to manually replace `DAPR_VERSION_HELM` and `DAPR_VERSION_TAG`, after debugging, you need to revert the change. In these operations, git checkout will not be used

daixiang0: Change version string -> debug -> revert, changes to those files can be git checkout to the original.

CrazyHZM: If you change some content under the `charts` pkg, git checkout is not necessarily a very good behavior, because it will revert the records you really want to change.It is not the reverse of `set_helm_dapr_version.sh`
@daixiang0 What do you think?

daixiang0: Let's use one script `set_helm_dapr_version.sh`, add a param to revert/set.

CrazyHZM: > Let's use one script `set_helm_dapr_version.sh`, add a param to revert/set.

@daixiang0 Done

yaron2:
@daixiang0 please re-review.

ItalyPaleAle:
@CrazyHZM Thanks for sending this PR, I noticed it only now!

I think this issue was fixed in master a few weeks back with https://github.com/dapr/dapr/pull/4334 . Can you please check if it's working for you now?

CrazyHZM:
> @CrazyHZM Thanks for sending this PR, I noticed it only now!
> 
> I think this issue was fixed in master a few weeks back with #4334 . Can you please check if it's working for you now?

I checked, this problem has been fixed, whether this PR can be modified into a functional enhancement type to replace the version number information in full, if you think this function is not necessary, you can easily close the PR
@ItalyPaleAle

ItalyPaleAle:
@CrazyHZM  What would the advantage be in this case?

We noticed that in a dev environment the version sent to Helm doesn't need to be anything specific; it just needs to be _a_ version number, and 0.0.0 seems to be ok.

CrazyHZM:
@ItalyPaleAle 
I will close the pr first.

PR #4299, opened 2022-02-25
Add resiliency feature flag

berndverst:

# Description

Adds resiliency feature flag


pkg/runtime/cli.go
```diff
@@ -265,17 +265,26 @@ func FromFlags() (*DaprRuntime, error) {
...
+			namespace = os.Getenv("NAMESPACE")
+			resiliencyConfigs = resiliency_config.LoadKubernetesResiliency(log, *appID, namespace, operatorClient)
+		case modes.StandaloneMode:
+			resiliencyConfigs = resiliency_config.LoadStandaloneResiliency(log, *appID, *componentsPath)
+		}
+		log.Debugf("Found %d resiliency configurations.", len(resiliencyConfigs))
+		r = resiliency_config.FromConfigurations(log, resiliencyConfigs...)
+	} else {
+		log.Debug("Resiliency is not enabled.")
+		r = resiliency_config.New(log)
```
pkedy, line 286: Should we create a NoOp implementation to avoid the codepath of looking in maps, etc (even if they are empty)?

/cc: @artursouza

artursouza: I not fully understand this comment.

berndverst: @pkedy if that is a question for me I don't know the answer.

pkedy: If you add a `noop.go` with the following code, you can completely bypass the resiliency logic we are adding. Without it, the code path would still access empty maps which results in an empty resiliency policy instead of a NoOp. Just a thought.

```golang
package resiliency

import (
	"context"
)

type NoOp struct{}

// RoutePolicy returns a NoOp policy runner for a route.
func (*NoOp) RoutePolicy(ctx context.Context, name string) Runner {
	return func(oper Operation) error {
		return oper(ctx)
	}
}

// EndpointPolicy returns a NoOp policy runner for a service endpoint.
func (*NoOp) EndpointPolicy(ctx context.Context, service string, endpoint string) Runner {
	return func(oper Operation) error {
		return oper(ctx)
	}
}

// ActorPolicy returns a NoOp policy runner for an actor instance.
func (*NoOp) ActorPolicy(ctx context.Context, actorType string, id string) Runner {
	return func(oper Operation) error {
		return oper(ctx)
	}
}

// ComponentPolicy returns a NoOp policy runner for a component.
func (*NoOp) ComponentPolicy(ctx context.Context, name string) Runner {
	return func(oper Operation) error {
		return oper(ctx)
	}
}
```

Usage from `pkg/runtime/cli.go`:

```golang
r = &resiliency_config.NoOp{}
```

pkedy: Created PR #4308 to add this

berndverst: Updated.

artursouza (approved):

PR #4308, opened 2022-02-28
NoOp implementation of Resiliency

pkedy:
# Description

Adding a NoOp implementation of Resiliency for true bypass behavior.

/cc: @artursouza @halspang @berndverst

### DIFF
--- a/pkg/resiliency/noop.go
+++ b/pkg/resiliency/noop.go
@@ -4,7 +4,7 @@ import (
 	"context"
 )
 
-// NoOp is a true bypass implementation of `Provider``.
+// NoOp is a true bypass implementation of `Provider`.
 type NoOp struct{}
 
 // Ensure `*NoOp` satisfies the `Provider` interface.
--- a/pkg/resiliency/noop.go
+++ b/pkg/resiliency/noop.go
@@ -1,0 +1,39 @@
+package resiliency
+
+import (
+	"context"
+)
+
+// NoOp is a true bypass implementation of `Provider``.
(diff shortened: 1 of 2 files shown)


pkg/resiliency/noop.go
```diff
@@ -0,0 +1,39 @@
+package resiliency
```
artursouza, line 1: license headers.

artursouza (changes requested):

pkg/resiliency/noop.go
pkedy: Thanks!

pkg/resiliency/noop_test.go
```diff
@@ -0,0 +1,55 @@
+package resiliency
+
+import (
+	"context"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestNoOp(t *testing.T) {
```
halspang, line 23: Should the no-op policy have tests that show it is truly a no-op (return failure and not retry).

pkedy: Yeah I'll tweak the tests to return an error for each call type.

halspang (approved):

yaron2:
Thanks for this PR, true bypass is extremely important. We learned it the hard way early on with tracing.

yaron2 (approved):

berndverst (approved):

artursouza (approved):

PR #4315, opened 2022-03-01
Fix devcontainer image not building

ItalyPaleAle:
The devcontainer image was failing to build on a MacOS host when docker-in-docker was used, with a permission error.


berndverst (approved):
Thanks for fixing!

artursouza (approved):
Please, cherry pick into release-1.6 branch.

PR #4322, opened 2022-03-03
Add missing component metrics

### LINKED ISSUE #4068: important metrics are missing
## In what area(s)?

/area runtime

## Describe the feature
while dapr exposes a while truckload of metrics, it seems to be missing what I would call the most relevant ones, such as pubsub messages in flight, service invocation requests in flight etc.

## Release Note

RELEASE NOTE:

yaron2:
Fixes #4068.

### DIFF
--- a/pkg/runtime/runtime.go
+++ b/pkg/runtime/runtime.go
@@ -212,6 +212,7 @@ type pubsubSubscribedMessage struct {
 	topic      string
 	metadata   map[string]string
 	path       string
+	pubsub     string
 }
 
 // NewDaprRuntime returns a new runtime with the given runtime config and global config.
@@ -320,9 +321,11 @@ func (a *DaprRuntime) setupTracing(hostAddress string, exporters traceExporterSt
 }
 
 func (a *DaprRuntime) initRuntime(opts *runtimeOpts) error {
+	a.namespace = a.getNamespace()
+
 	// Initialize metrics only if MetricSpec is enabled.
 	if a.globalConfig.Spec.MetricSpec.Enabled {
-		if err := diag.InitMetrics(a.runtimeConfig.ID); err != nil {
+		if err := diag.InitMetrics(a.runtimeConfig.ID, a.namespace); err != nil {
 			log.Errorf("failed to initialize metrics: %v", err)
 		}
 	}
@@ -331,7 +334,6 @@ func (a *DaprRuntime) initRuntime(opts *runtimeOpts) error {
 	if err != nil {
 		return err
 	}
-	a.namespace = a.getNamespace()
 	a.podName = a.getPodName()
 	a.operatorClient, err = a.getOperatorClient()
 	if err != nil {
@@ -598,6 +600,7 @@ func (a *DaprRuntime) beginPubSub(name string, ps pubsub.PubSub) error {
 				topic:      msg.Topic,
 				metadata:   msg.Metadata,
 				path:       routePath,
+				pubsub:     name,
 			})
 		}); err != nil {
 			log.Errorf("failed to subscribe to topic %s: %s", topic, err)
@@ -953,11 +956,17 @@ func (a *DaprRuntime) sendBindingEventToApp(bindingName string, data []byte, met
 func (a *DaprRuntime) readFromBinding(name string, binding bindings.InputBinding) error {
 	err := binding.Read(func(resp *bindings.ReadResponse) ([]byte, error) {
 		if resp != nil {
+			start := time.Now()
 			b, err := a.sendBindingEventToApp(name, resp.Data, resp.Metadata)
+			elapsed := diag.ElapsedSince(start)
+
+			diag.DefaultComponentMonitoring.InputBindingEvent(context.TODO(), name, err == nil, elapsed)
+
 			if err != nil {
 				log.Debugf("error from app consumer for binding [%s]: %s", name, err)
 				return nil, err
 			}
+
 			return b, err
 		}
 		return nil, nil
@@ -1466,8 +1475,12 @@ func (a *DaprRuntime) publishMessageHTTP(ctx context.Context, msg *pubsubSubscri
 		ctx, span = diag.StartInternalCallbackSpan(ctx, spanName, sc, a.globalConfig.Spec.TracingSpec)
 	}
 
+	start := time.Now()
 	resp, err := a.appChannel.InvokeMethod(ctx, req)
+	elapsed := diag.ElapsedSince(start)
+
 	if err != nil {
+		diag.DefaultComponentMonitoring.PubsubIngressEvent(ctx, msg.pubsub, string(pubsub.Retry), msg.topic, elapsed)
 		return errors.Wrap(err, "error from app channel while sending pub/sub event to app")
 	}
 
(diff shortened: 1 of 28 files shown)


yaron2:
@artursouza I used Go Benchmarks to test performance for per call component metrics.

The result it an additional 1423 op/ns which translates to ~0.001423ms.

pkg/runtime/runtime.go
```diff
@@ -1540,12 +1560,14 @@ func (a *DaprRuntime) publishMessageGRPC(ctx context.Context, msg *pubsubSubscri
 			decoded, decodeErr := base64.StdEncoding.DecodeString(dataAsString)
 			if decodeErr != nil {
 				log.Debugf("unable to base64 decode cloudEvent field data_base64: %s", decodeErr)
+				diag.DefaultComponentMonitoring.PubsubIngressEvent(ctx, msg.pubsub, string(pubsub.Retry), msg.topic, 0)
```
mukundansundar, line 1563: this ideally should not come under `retry` since it seems to be a data error.

pkg/runtime/runtime.go
```diff
@@ -1513,11 +1531,13 @@ func (a *DaprRuntime) publishMessageHTTP(ctx context.Context, msg *pubsubSubscri
 		// When adding/removing an error here, check if that is also applicable to GRPC since there is a mapping between HTTP and GRPC errors:
 		// https://cloud.google.com/apis/design/errors#handling_errors
 		log.Errorf("non-retriable error returned from app while processing pub/sub event %v: %s. status code returned: %v", cloudEvent[pubsub.IDField], body, statusCode)
+		diag.DefaultComponentMonitoring.PubsubIngressEvent(ctx, msg.pubsub, string(pubsub.Drop), msg.topic, elapsed)
 		return nil
 	}
 
 	// Every error from now on is a retriable error.
 	log.Warnf("retriable error returned from app while processing pub/sub event %v, topic: %v, body: %s. status code returned: %v", cloudEvent[pubsub.IDField], cloudEvent[pubsub.TopicField], body, statusCode)
+	diag.DefaultComponentMonitoring.PubsubIngressEvent(ctx, msg.pubsub, string(pubsub.Retry), msg.topic, elapsed)
```
mukundansundar, line 1540: convert `pubsub.Retry`, `pubsub.Drop` etc to be lowercase to be consistent across all diag paths.

pkg/runtime/runtime.go
```diff
@@ -1540,12 +1560,14 @@ func (a *DaprRuntime) publishMessageGRPC(ctx context.Context, msg *pubsubSubscri
...
 			if decodeErr != nil {
 				log.Debugf("unable to base64 decode cloudEvent field data_base64: %s", decodeErr)
+				diag.DefaultComponentMonitoring.PubsubIngressEvent(ctx, msg.pubsub, string(pubsub.Retry), msg.topic, 0)
 
 				return decodeErr
 			}
 
 			envelope.Data = decoded
 		} else {
+			diag.DefaultComponentMonitoring.PubsubIngressEvent(ctx, msg.pubsub, string(pubsub.Retry), msg.topic, 0)
```
mukundansundar, line 1570: I believe this should not come under `Retry` since this seems to be a data error.

pkg/runtime/runtime.go
```diff
@@ -1558,6 +1580,7 @@ func (a *DaprRuntime) publishMessageGRPC(ctx context.Context, msg *pubsubSubscri
 			case []byte:
 				envelope.Data = v
 			default:
+				diag.DefaultComponentMonitoring.PubsubIngressEvent(ctx, msg.pubsub, string(pubsub.Retry), msg.topic, 0)
```
mukundansundar, line 1583: same asa above.

pkg/runtime/runtime.go
```diff
@@ -1612,16 +1639,20 @@ func (a *DaprRuntime) publishMessageGRPC(ctx context.Context, msg *pubsubSubscri
 	case runtimev1pb.TopicEventResponse_SUCCESS:
 		// on uninitialized status, this is the case it defaults to as an uninitialized status defaults to 0 which is
 		// success from protobuf definition
+		diag.DefaultComponentMonitoring.PubsubIngressEvent(ctx, msg.pubsub, string(pubsub.Success), msg.topic, elapsed)
```
mukundansundar, line 1642: lowercase for `success`

pkg/runtime/runtime.go
```diff
@@ -1612,16 +1639,20 @@ func (a *DaprRuntime) publishMessageGRPC(ctx context.Context, msg *pubsubSubscri
...
 		return nil
 	case runtimev1pb.TopicEventResponse_RETRY:
+		diag.DefaultComponentMonitoring.PubsubIngressEvent(ctx, msg.pubsub, string(pubsub.Retry), msg.topic, elapsed)
```
mukundansundar, line 1645: same as above.

mukundansundar (commented):
Have left some comments on this PR.

pkg/runtime/runtime.go
yaron2: Dapr treats this as a Retry, and the monitoring simply follows the logic. This PR doesn't deal at all with how Dapr handles pub/sub errors. Will not change.

pkg/runtime/runtime.go
yaron2: Repeated from above: Dapr treats this as a Retry, and the monitoring simply follows the logic. This PR doesn't deal at all with how Dapr handles pub/sub errors. Will not change.

pkg/runtime/runtime.go
yaron2: Repeated from above: Dapr treats this as a Retry, and the monitoring simply follows the logic. This PR doesn't deal at all with how Dapr handles pub/sub errors. Will not change.

pkg/runtime/runtime.go
yaron2: 1. These values are taken from Components Contrib
2. This is not how Golang works. Lowercasing them would mean these fields are unexported, and thus unavailable to `runtime.go`.

pkg/runtime/runtime.go
yaron2: 1. These values are taken from Components Contrib
2. This is not how Golang works. Lowercasing them would mean these fields are unexported, and thus unavailable to `runtime.go`.

pkg/runtime/runtime.go
yaron2: 1. These values are taken from Components Contrib
2. This is not how Golang works. Lowercasing them would mean these fields are unexported, and thus unavailable to `runtime.go`.

pkg/runtime/runtime.go
mukundansundar: When I meant lowercase, I meant the values not the variable names.

yaron2: Ah well that makes sense! should have said so :)

Changed

pkg/runtime/runtime.go
```diff
@@ -1579,8 +1602,10 @@ func (a *DaprRuntime) publishMessageGRPC(ctx context.Context, msg *pubsubSubscri
 
 	ctx = invokev1.WithCustomGRPCMetadata(ctx, msg.metadata)
 
-	// call appcallback
+	start := time.Now()
 	clientV1 := runtimev1pb.NewAppCallbackClient(a.grpc.AppClient)
+	elapsed := diag.ElapsedSince(start)
+
 	res, err := clientV1.OnTopicEvent(ctx, envelope)
```
Taction, line 1609: nits: calculate `elapsed` should behind `OnTopicEvent`.
```
res, err := clientV1.OnTopicEvent(ctx, envelope)
elapsed := diag.ElapsedSince(start)
```

Taction (commented):
LGTM overall, just a small change commented above.
Besides, for pubsub there may be some error/drop cases in `beginPubSub` func not counted, such as https://github.com/dapr/dapr/blob/master/pkg/runtime/runtime.go#L559

pkg/runtime/runtime.go
yaron2: Done

Taction (approved):

PR #4328, opened 2022-03-04
Add Sentry monitoring of certificate expiry

berndverst:

# Description

Production users of Dapr need a way to set up alerting for expiring certificates. This PR checks the certificate expiration on an hourly basis and beginning 30 days from expiry on an hourly basis emits a log entry to the Sentry warning logs indicating that the certificate will expire in X hours. Additionally an expired certificate is also logged.

Production users can setup alerting for the specific warning strings from the Sentry service logs.


berndverst:
This is the same code for determining the certificate expiration as is present in the CLI:

See https://github.com/dapr/cli/blob/master/pkg/kubernetes/mtls.go#L168
```go
	secret, err := getTrustChainSecret()
	if err != nil {
		return nil, err
	}

	caCrt := secret.Data["ca.crt"]
	block, _ := pem.Decode(caCrt)
	cert, err := x509.ParseCertificate(block.Bytes)
	if err != nil {
		return nil, err
	}
	return &cert.NotAfter, nil
```

pkg/sentry/sentry.go
```diff
@@ -76,6 +79,27 @@ func (s *sentry) Run(ctx context.Context, conf config.SentryConfig, readyCh chan
...
+			if err != nil {
+				log.Warn("could not determine Dapr root certificate expiration time")
+				continue
+			}
+			if cert.NotAfter.Before(time.Now()) {
+				log.Warn("Dapr root certificate expiration warning: certificate has expired.")
+				continue
+			}
+			if (cert.NotAfter.Add(-30 * 24 * time.Hour)).Before(time.Now()) {
+				log.Warnf("Dapr root certificate expiration warning: certificate expires in %s hours", time.Until(cert.NotAfter).Hours())
```
artursouza, line 99: Can this message be printed as "x day(s) and y hour(s)". It will improve readability IMO.

artursouza (changes requested):

berndverst:
@yaron2 any thoughts on this? Should be pretty straightforward.

It's intended to help people set up alerting in prod for cert rotation.

yaron2 (approved):

PR #4329, opened 2022-03-04
Replace Route Policy with Component Policy

berndverst:

# Description

Removes Route Policy in favor of Component Policy for PubSub and Bindings. This simplifies the user experience a lot at this time.


yaron2:
Can you give an example here of the difference in user experience?

berndverst:
Copying from an internal chat:

yaron2:
Thanks,  it makes sense overall, but:

> If you really want to have different policies for publish/subscribe you can always make more than one component.

That might be a problem since Dapr will essentially be initializing the same component twice with 2 different connections That might lead to some odd behaviors.

berndverst:
The Python SDK for example abstracts the concept of routes because it magically registers default routes for PubSub topics. A user would not be explicitly aware of the existence of routes. It makes more sense for the simple PubSub topic subscription resiliency to be covered by the overall component policy.

I do not think we should expect folks to know about the existence of routes (and their implementation) and having to set route policies at this time.

berndverst:
@artursouza @yaron2 @halspang are we going to proceed with this change? It will certainly make documenting how resiliency works a lot easier.

halspang:
> That might be a problem since Dapr will essentially be initializing the same component twice with 2 different connections. That might lead to some odd behaviors.

Yeah, that's not ideal. That point was more to illustrate that we aren't really losing anything here if a customer needs it immediately. I think this kind of scenario is better covered in Phase 2/with user feedback about _how_ they want it covered.

yaron2:
@halspang @berndverst I agree we can proceed, having looked at the user experience with route policies, I think this makes more sense for this stage of the APIs.

However, regardless of that, I strongly recommend we change this:

```go
type ResiliencySpec struct {
	Policies       Policies       `json:"policies"`
	BuildingBlocks BuildingBlocks `json:"buildingBlocks" yaml:"targets"`
}

type BuildingBlocks struct {
	Services   map[string]EndpointPolicyNames `json:"services,omitempty" yaml:"services,omitempty"`
	Actors     map[string]ActorPolicyNames    `json:"actors,omitempty" yaml:"actors,omitempty"`
	Components map[string]PolicyNames         `json:"components,omitempty" yaml:"components,omitempty"`
}
```

To this:

```go
type ResiliencySpec struct {
	Policies       Policies       `json:"policies"`
	Targets Targets `json:"buildingBlocks" yaml:"targets"`
}

type Targets struct {
	Apps   map[string]EndpointPolicyNames `json:"apps,omitempty" yaml:"services,omitempty"`
	Actors     map[string]ActorPolicyNames    `json:"actors,omitempty" yaml:"actors,omitempty"`
	Components map[string]PolicyNames         `json:"components,omitempty" yaml:"components,omitempty"`
}
```

OR

```go
type ResiliencySpec struct {
	Policies       Policies       `json:"policies"`
	Targets Targets `json:"appliesTo" yaml:"targets"`
}

type Targets struct {
	Apps   map[string]EndpointPolicyNames `json:"apps,omitempty" yaml:"services,omitempty"`
	Actors     map[string]ActorPolicyNames    `json:"actors,omitempty" yaml:"actors,omitempty"`
	Components map[string]PolicyNames         `json:"components,omitempty" yaml:"components,omitempty"`
}
```

### Why?

The term building blocks for what it currently stands is confusing, if not wrong. Under building blocks we find `Components` - Components are not building blocks.

We also find `Services`, which is an ambiguous name, and also not a building block per-se. To Kubernetes users, this might mean a selector of a Kubernetes service, for example. In Dapr jargon we often try to refer to higher level constructs, in this case, it's an app, and so I recommend to change `services` to `apps` and `buildingBlocks` to `targets` or `appliesTo`, for clearer understanding and readability.

berndverst:
@yaron2 can you take another look now? I made the changes you suggested.

```
targets:
  components:
      pubsub1:
        // specifying values only at the top level is equivalent to setting an output policy
        retry: pubsubRetry
        circuitBreaker: pubsubCB

      pubsub2:
        output:
          retry: pubsubRetry
          circuitBreaker: pubsubCB
        input:
          timeout: general
          retry: pubsubRetry
          circuitBreaker: pubsubCB
```

yaron2:
@berndverst great changes, thanks!

My only remaining concerns is with the ambiguity of having a top level policy that defaults to an output policy.
I think it would be simpler and much more to the "element of least surprise" if we cancel the top policy feature.

berndverst:
@yaron2 yes I agree - I am totally fine removing the top level way to specify a policy.

yaron2 (approved):

artursouza (approved):

PR #4330, opened 2022-03-04
Devcontainer should install protoc 3.14.0

berndverst:

# Description

Protoc 3.14.0 is specifically required for generating Dapr protos, but today the devcontainer installs the latest version (3.19.4 etc). This fixes that.


docker/Dockerfile-dev
```diff
@@ -53,13 +53,12 @@ RUN apt-get update \
...
     #
-    # Install protoc compiler.
-    && LATEST_PROTOC_URL=$(curl -s https://api.github.com/repos/protocolbuffers/protobuf/releases/latest | grep -oP '"browser_download_url": "\K(.*)(?=")' | grep "protoc-.*-linux-x86_64") \
-    && wget -q -O /tmp/staging/protoc-latest.zip ${LATEST_PROTOC_URL} \
-    && unzip /tmp/staging/protoc-latest.zip -d /tmp/staging/protoc \
-    && chmod -R 755 /tmp/staging/protoc \
-    && mv /tmp/staging/protoc/include/* /usr/local/include \
-    && mv /tmp/staging/protoc/bin/* /usr/local/bin \
+    # Install protoc compiler 3.14.0 currently required by 'make gen-proto'.
+    && PROTOC_ZIP=protoc-3.14.0-linux-x86_64.zip \
```
ItalyPaleAle, line 57: Nit: suggest keeping the version in its own variable.

ItalyPaleAle (approved):
Approved but with a suggested change

docker/Dockerfile-dev
berndverst: Since you are planning to update it anyway for Darwin ARM64 native feel free to make the change then. We never update the protoc version so far, so this isn't necessary :)

shubham1172 (approved):

berndverst:
@artursouza please review and merge. We'll include this change when we build the 0.7 dev container for release 1.7.

daixiang0:
We need use an ENV or config to unify protoc version.

berndverst:
> We need use an ENV or config to unify protoc version.

Sure. But this PR only updates the dev container given that today this isn't exposed via ENV var.

Also updating protoc to a newer version produces different Protos (I tried this)

It seems we likely will need to keep protoc version locked for now.

Can we merge this change so that the release 1.7.0 dev container will have this fix? Right now you cannot actually run "make gen-proto" from the dev container.

I recommend the idea of changing protoc version and controlling that with a unified ENV var should be discussed in a new issue.

Thoughts @daixiang0 ?

daixiang0:
Sure, please create an issue to track it.

Also fix DCO.

berndverst:
@daixiang0 done. I opened https://github.com/dapr/dapr/issues/4350

artursouza (approved):

PR #4342, opened 2022-03-08
gRPC stream tracing: Ignore dapr api calls

yaron2:
The Configuration API's subscribe method introduced a gRPC streaming method to Dapr for the first time.

Consequently, the tracing middleware for gRPC proxying (which relies on gRPC streaming) needed an update to only handle requests that are meant for proxying, ie. not Dapr API calls.

### DIFF
--- a/pkg/diagnostics/grpc_tracing.go
+++ b/pkg/diagnostics/grpc_tracing.go
@@ -102,7 +102,7 @@ func GRPCTraceUnaryServerInterceptor(appID string, spec config.TracingSpec) grpc
 func GRPCTraceStreamServerInterceptor(appID string, spec config.TracingSpec) grpc.StreamServerInterceptor {
 	return func(srv interface{}, ss grpc.ServerStream, info *grpc.StreamServerInfo, handler grpc.StreamHandler) error {
 		if strings.Index(info.FullMethod, daprPackagePrefix) == 0 {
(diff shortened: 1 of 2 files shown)


halspang:
I tried using this change while testing the .NET implementation of config subscribe and found that I still needed to remove proxying in order to get the stream to read.

yaron2:
> I tried using this change while testing the .NET implementation of config subscribe and found that I still needed to remove proxying in order to get the stream to read.

@halspang please confirm this is working now.

halspang (approved):
Worked for my local implementation of the subscribe API for .NET. Also worked for a gRPC proxying call, also through a .NET app.

@mukundansundar - Can you verify this with your implementation of the subscribe API in the Java SDK?

mukundansundar (dismissed):

Taction (approved):
LGTM

Taction:
I think most users may not interested in this information, also if any user asking for that we can add a configuration param to define whether add this into tracing.

mukundansundar:
> Worked for my local implementation of the subscribe API for .NET. Also worked for a gRPC proxying call, also through a .NET app.
> 
> @mukundansundar - Can you verify this with your implementation of the subscribe API in the Java SDK?

Have verified that this works for the subscribe API with Java SDK

Issue #4357, opened 2022-03-12
placement raft cluster hangs when large number of pods restart


psychomelet:
Dapr version v1.5.1

Dapr placement using ha deployment: 3 replicas

We're monitoring the cases of dapr-placement raft cluster stop in our prod k8s cluster, the placements stopped with no nodes working as leader in cluster, causing the entire actor service down, and the cluster can not recover automatically.

When the incident happens, we have to follow exactly 3 steps to recover the service. Neither restart dapr-placement nor restart actor deployment can recover the service when placement down.
1. kill all `dapr-placement` pods to restart placement.
2. scale all actor deployments to 0 and wait for all pods terminated.
3. scale actor deployment back to normal.

It can be easily reproduced with large number for actor replicas in deployment, steps:
1. start a clean k8s
2. start a deployment with ~500 replicas as actor service, we're using the `dapr-ext-fastapi`
3. update the deployment.
4. repeats [step2&3] multiple times, our tests show in about restarting for 5~10 times the hang will happen. (monitoring large number of hosts terminating and creating at the same time)
5. the placement hangs, all actor down.

P.S:
when following the reproduce steps, the #4353 will occur first and crash the placement pods. We fixed the concurrent map write and the craft hang issues still exist. So I opened this as a separate issue.

PR #4359, opened 2022-03-12
Speedup placement hosts update

psychomelet:
Main changes:
1. add dead grpc stream detection, skip the update for terminated pods
2. update and wait all hosts parallel instead of sequential
3. table updates will wait for at most 10s and continue to next update, preventing the cluster being blocked by single update.

Discussion 1: about the 10s timeout
The `disseminateOperation` may block for a very long time, possible blocked in the `err = s.Send(o)`.

When this happened, a single host stream connection will block the whole cluster table updates, causing large cluster downtime.

The code shows the runtime client have a 5s auto unlock, so it seems it's not making things worse if the server ignore and continue the next table updates after 10s.

Discussion 2: about the table updates design choice
1. all hosts locks sequential -> all hosts updates sequential -> all hosts unlock sequential
2. all locks have a 5s auto unlock

This implementation actually makes me very confusing.

1. Doing all things sequential make the performance very bad under hundreds or more of pods
2. then that combined with client 5s auto unlock, if anything blocks the procedure in the middle, 5s later all locks auto unlocked, and data inconsistent may occur. In the same time the table update goroutine in the placement service also blocks for a very long time, about ~15s, sometimes even longer (retry 3 times with backoff 5s).

It seems the current implementation lost both consistency and performance.

An alternative approach will be using 2 WaitGroup, first wg parallel all `lock + update`, wait for complete, then start 2nd wg to parallel `unlock`, this will make the cluster state more consistent, if that's important enough.

### DIFF
--- a/pkg/placement/membership.go
+++ b/pkg/placement/membership.go
@@ -216,6 +216,8 @@ func (p *Service) processRaftStateCommand(stopCh chan struct{}) {
 	// of raft apply command.
 	logApplyConcurrency := make(chan struct{}, raftApplyCommandMaxConcurrency)
 
+	processingTicker := time.NewTicker(10 * time.Second)
+
 	for {
 		select {
 		case <-stopCh:
@@ -223,6 +225,10 @@ func (p *Service) processRaftStateCommand(stopCh chan struct{}) {
 
 		case <-p.shutdownCh:
 			return
+
+		case <-processingTicker.C:
+			log.Infof("process raft state command: nothing happened...")
+			continue
 
 		case op := <-p.membershipCh:
 			switch op.cmdType {
@@ -307,11 +313,55 @@ func (p *Service) performTableDissemination() {
 func (p *Service) performTablesUpdate(hosts []placementGRPCStream, newTable *v1pb.PlacementTables) {
 	// TODO: error from disseminationOperation needs to be handle properly.
 	// Otherwise, each Dapr runtime will have inconsistent hashing table.
+	startedAt := time.Now()
+
+	var wg sync.WaitGroup
 	for _, host := range hosts {
-		p.disseminateOperation([]placementGRPCStream{host}, "lock", nil)
-		p.disseminateOperation([]placementGRPCStream{host}, "update", newTable)
-		p.disseminateOperation([]placementGRPCStream{host}, "unlock", nil)
+		wg.Add(1)
+
+		go func(h placementGRPCStream) {
+			defer wg.Done()
+
+			timeoutC := make(chan bool)
+			stopC := make(chan bool)
+			step := ""
+			isTimeout := false
+
+			go func() {
+				defer close(stopC)
+
+				timeoutC <- true
+				step = "lock"
+				p.disseminateOperation([]placementGRPCStream{h}, "lock", nil)
+				if isTimeout {
+					return
+				}
+				step = "update"
+				p.disseminateOperation([]placementGRPCStream{h}, "update", newTable)
+				if isTimeout {
+					return
+				}
+				step = "unlock"
+				p.disseminateOperation([]placementGRPCStream{h}, "unlock", nil)
+			}()
+
+			<-timeoutC
+
+			select {
+			case <-time.After(10 * time.Second):
+				// timeout
+				isTimeout = true
+				log.Errorf("performTablesUpdate(%v) timeout 10s at step %v!!!", h, step)
+				return
+			case <-stopC:
+				// NOTE normal
+				return
+			}
+		}(host)
 	}
+	wg.Wait()
+
+	log.Debugf("performTablesUpdate succeed %v", time.Since(startedAt))
 }
 
 func (p *Service) disseminateOperation(targets []placementGRPCStream, operation string, tables *v1pb.PlacementTables) error {
@@ -321,7 +371,6 @@ func (p *Service) disseminateOperation(targets []placementGRPCStream, operation 
 	}
 	var err error
 	for _, s := range targets {
-
 		config := retry.DefaultConfig()
 		config.MaxRetries = 3
 		backoff := config.NewBackOff()
@@ -330,8 +379,8 @@ func (p *Service) disseminateOperation(targets []placementGRPCStream, operation 
 				// Check stream in stream pool, if stream is not available, skip to next.
 				if !p.hasStreamConn(s) {
 					remoteAddr := "n/a"
-					if peer, ok := peer.FromContext(s.Context()); ok {
-						remoteAddr = peer.Addr.String()
+					if _peer, ok := peer.FromContext(s.Context()); ok {
+						remoteAddr = _peer.Addr.String()
 					}
 					log.Debugf("runtime host (%q) is disconnected with server. go with next dissemination (operation: %s).", remoteAddr, operation)
 					return nil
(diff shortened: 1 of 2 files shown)


yaron2:
Thanks for these improvements @psychomelet. Please fix the failing build checks.

> Doing all things sequential make the performance very bad under hundreds or more of pods

In the "happy path", meaning connections aren't blocked for long periods of time by the host pod for any reason, a single pod update operation takes an observed ~2ms end to end. So in the case of 500 pods, for example, this would cause the total update operation to last 1000ms or 1s. Still, parallelizing this would be a welcome addition as long as it doesn't break existing guarantees.

> 5s later all locks auto unlocked, and data inconsistent may occur

There should be two levels of protection here from data inconsistency in the event of a host pod update timeout.
The first, is that once a pod timed out, it means it did not finish it's update successfully, and it should assume that some actors might now be hosted on a different pod. All pending and incoming requests from the app to this sidecar instance should be rejected until the next tables update finishes successfully.

The second, is how state transactions are managed. In the unlikely event that an actor is handling an event, and during that time the pod is now blocked for new actor calls, and the same actor instance is now receiving calls in a new pod, then the mechanism of saving state by actors is using Transactions, which, depending on the state store component used, offer different levels of isolation. Serializable and Read Committed transactions will likely guarantee data consistency from the database perspective.

psychomelet:
Thanks for the explanation.

Based on my observation and load tests, the happy path easily failed. It's highly possible something blocked in the `err = s.Send(o)` which I don't know why. But the result is, after the patch to parallel all hosts updates in goroutine, some host `lock -> update -> unlock` triggered 60s timeout, which will be then be let free and stuck in it's goroutine, but the reason why the host update blocked should be dig further.

> All pending and incoming requests from the app to this sidecar instance should be rejected until the next tables update finishes successfully.

This can cause the actor service experience a long downtime for whole cluster if the placement service somehow stuck in the middle. Maybe we can add another feature in the runtime client side to solve this:
1. (same as now) if update timeout aborted in the middle, reject further requests.
2. in the same time, the client make a gRPC request to placement to get the latest hash table. (init by client)
3. as long as the client init rpc request success, continue the service.

> Serializable and Read Committed transactions will likely guarantee data consistency from the database perspective.

Under load tests we do produced data inconsistent situation, when dapr-placement finally reached complete state and all actor pods are running success, but there're still "can't find actor" error message, which can only be recovered by scale all actor pods to 0 (clean all local state) and re-scale actor pods back. (This can be reproduced with same steps in #4357 , to make the inconsistent happen more quickly, add a `sleep 6s` right after the `p.disseminateOperation 'lock'`, to manually trigger client lock timeout.)

I'm wondering if there's any data consistent check & full sync between runtime & placement, maybe we can make the client periodically check the md5sum of server hash table and if it's not matched, process a client inited full sync.

yaron2:
@psychomelet see linter errors.

yaron2:
> (same as now) if update timeout aborted in the middle, reject further requests.
in the same time, the client make a gRPC request to placement to get the latest hash table. (init by client)
as long as the client init rpc request success, continue the service.

This is an acceptable solution.

> Under load tests we do produced data inconsistent situation

You and I meant different things by data inconsistency; I meant it in the scope of the actor's business data, while you meant it at the level of different host tables across pods.

> OR when client find it's unable to invoke method of some actor

This is too much non-deterministic, there could be many other reasons why an actor won't be found. This will generate a big number of false positives and put unnecessary load on the placement service.

> maybe we can make the client periodically check the md5sum of server hash table and if it's not matched

This is a better solution I believe.

pkg/placement/membership.go
```diff
@@ -224,6 +226,10 @@ func (p *Service) processRaftStateCommand(stopCh chan struct{}) {
 		case <-p.shutdownCh:
 			return
 
+		case <-processingTicker.C:
+			log.Infof("process raft state command: nothing happened...")
+			continue
```
psychomelet, line 231: The logging print is like a heartbeat that prints to console to indicate the node is alive.

pkg/placement/membership.go
```diff
@@ -307,9 +313,56 @@ func (p *Service) performTableDissemination() {
...
+					return
+				}
+				step = "unlock"
+				p.disseminateOperation([]placementGRPCStream{h}, "unlock", nil)
+			}()
+
+			<-timeoutC
+
+			select {
+			case <-time.After(10 * time.Second):
```
psychomelet, line 352: quit host level 'lock -> update -> unlock' after 10s timeout, the goroutine will be left there, and the isTimeout flag will quit whatever next action asap.

psychomelet:
hmm the uncovered tests looks not very intuitive to add, I'v add comments about the function. And the TestActorReentrancy tests have failed, but seems non related to the changes of the pr.

psychomelet:
For what's next, let me summary the discussion for possible further improvements:

Firstly, Replace the client-side silent auto unlock with client initi full sync rpc.
1. (same as now) if update timeout aborted in the middle, reject further requests.
2. in the same time, the client make a gRPC request to placement to get the latest hash table. (init by client)
3. as long as the client init rpc request success, continue the service.

Secondly, client <> server heartbeat md5sum check & full sync when needed:
1. runtime <> placement service doing heartbeat with current md5sum hash of current hash table in placement service.
2. If the client finds out the md5sum hash not matched, and there's no current updating process ongoing, client will lock & make a gRPC request to get latest hash table.
3. as long as the client init rpc request success, continue the service.

Theses additional 2 checks depends on a full-sync gRPC API on the placement service side, which beyond my current scope of the code, shall we open another issue for that?

pkg/placement/membership.go
yaron2: This should probably be a debug log?

yaron2:
@psychomelet please update the branch with master and see my comment above.

yaron2:
@psychomelet please resolve conflicts.

psychomelet:
@yaron2 comments resolved, but it seems the CI is blocked.

daixiang0:
Please rebase main branch.

psychomelet:
@daixiang0 rebased.

psychomelet:
@daixiang0 it seems the ci broken in non-related code

daixiang0:
Please check and fix lint issues.

yaron2:
@daixiang0 did you close by mistake?

daixiang0:
Sorry about that :(

psychomelet:
@daixiang0 rebased

daixiang0:
Please rebase latest main branch.

psychomelet:
@daixiang0 rebased

psychomelet:
@daixiang0 what's the status of this PR, as I've rebased many times, anything more shall I do?

daixiang0 (dismissed):
/cc @yaron2

psychomelet:
@yaron2 @daixiang0 rebased to latest.

yaron2:
@artursouza can you please review as well?

pkg/placement/membership.go
```diff
@@ -307,11 +313,55 @@ func (p *Service) performTableDissemination() {
...
 	// TODO: error from disseminationOperation needs to be handle properly.
 	// Otherwise, each Dapr runtime will have inconsistent hashing table.
+	startedAt := time.Now()
+
+	var wg sync.WaitGroup
 	for _, host := range hosts {
-		p.disseminateOperation([]placementGRPCStream{host}, "lock", nil)
-		p.disseminateOperation([]placementGRPCStream{host}, "update", newTable)
-		p.disseminateOperation([]placementGRPCStream{host}, "unlock", nil)
+		wg.Add(1)
```
artursouza, line 320: I wonder if this can cause issues for high scale clusters?  I think we should add a `parallelism` setting to make sure we don't spawn too many go routines at once. We can add it as a separate PR since this one makes things better already.

yaron2: I'm not overly concerned here, hundreds of goroutines (representing hundreds of nodes, a very unlikely and uncommon deployment topology) would be fine IMO. a goroutine takes 2kb of memory and so memory usage would go up minimally, while CPU usage should increase but not to unsustainable levels as the IO operation in dissemination uses already established streams and not creating/destroying new connections.

artursouza:
Merged as part of the PR above since contributor's fork does not allow updates from maintainers.

PR #4361, opened 2022-03-12
Reduce placement blocking client

ls-2018:
Signed-off-by: ls-2018 <acejilam@gmail.com>

The PR was submitted a long time ago, and the test case is now complete

If n nodes node to send message in sequence, first send a lock message to each node, then send an update message, and finally send an unlock message .from the receipt of the lock message to lock itself (stop call), to the receipt of the unlock message will stop during min (5s,2n times network communication),.
The following changes can greatly reduce the client pod blocking time when the pod changes frequently
Ref : 
- https://github.com/dapr/dapr/pull/4140 
- https://github.com/dapr/dapr/pull/4356
# Description
### test method 
100 clients
Simulate the delay in each communication of the server, get the time required for each node lock->update->unlock, and output the maximum stopping time of the node
image

You can see that changing the call method changes the blocking time from 20000+ms to 200+ms

### The test results
Don't use the for loop
image

After the change
image

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Unit tests passing


yaron2:
@ls-2018 can you explain why performing the loop in the outer method changes the behavior? the end result looks the same - there's still a loop over all the hosts?

ls-2018:
```
node :       1                2                3               4               5              n-1            n
            lock(1) ------> lock(2) ------> lock(3) ------> lock(4) ------> lock(5) ....... lock(n-1) ----> lock(n)  -->
process     update(n+1) --> update(n+2) --> update(n+3) --> update(n+4) --> update(n+5) ... update(2n-1)--> update(2n) -->
            unlock(2n+1)--> unlock(2n+2)--> unlock(2n+3)--> unlock(2n+4)--> unlock(2n+5)... unlock(3n-1)--> unlock(3n)

```
If there are n nodes, each node will go through lock, update, and unlock states when the actor services node information changes and synchronizes messages to all nodes. This blocks each node first until it times out or receives an unlock message. This means that each node will block 2n times the network communication time ` This time depends on the number of nodes, network latency`.
After my change, each node only blocks 2 times. This also reduces the blocking time of the client, which in turn improves performance
image
image

image

yaron2 (approved):
lgtm

Issue #4218, opened 2022-02-08
Updating components such as pubsub, the old components is never closed when running in a Kubernetes mode.


63hogan:
## In what area(s)?
pkg/runtime/runtime.go

## What version of Dapr?
newest

## Expected Behavior
close old component before updating component with same pubsubname.

## Actual Behavior
when updating some pubsub component config,  the old should be closed.
```go
// pkg/runtime/runtime.go
func (a *DaprRuntime) onComponentUpdated(component components_v1alpha1.Component) {
}

func (a *DaprRuntime) initPubSub(c components_v1alpha1.Component) error {
a.pubSubs[pubsubName] = pubSub
}
```
when received update request, if the pubsubName is same as old component(maybe for  updating config?) it just replaced with the new component pusSub, but the old is never closed.

yaron2:
Thanks, will be fixed in 1.7.

mstrYoda:
> 

@yaron2 There is also another issue which I opened a time ago, it is automatically closed, is that issue will be fixed in 1.7 too? https://github.com/dapr/dapr/issues/3015

shubham1172:
FYI, to reproduce the issue:
- Deploy the tutorial hello-world: https://github.com/dapr/quickstarts/tree/master/tutorials/hello-kubernetes
- Make changes to the state store configuration and run `kubectl apply -f redis-state.yaml`
- With each run, there is an extra Redis client added (can be seen in the screenshot) 

![image](https://user-images.githubusercontent.com/22132944/159647057-948c3e20-b3e6-46e3-8e57-4e1c2007c366.png)

shubham1172:
After the PR: 
(1 is CLI, 1 is GUI, 1 is Dapr)
![image](https://user-images.githubusercontent.com/22132944/159654703-6d2ea00d-2982-4bd5-bd70-38dea1275120.png)

yaron2:
@63hogan hot-reloading of components is actually not an intended feature, and more of a bug at the moment.

The expected behavior is to perform a rolling update of your deployments after a change to a component. As such, we will be preventing the in-place updates in this release and introduce hot-reload (properly, with a feature flag) in the next release (1.8).

msfussell:
#4433 PR closes this

shubham1172:
@msfussell, #4433 closed #4432, which turns off component updates. However this bug still remains open. It should be fixed before we turn on components updates again. Can you please reopen the issue?

PR #4376, opened 2022-03-15
[WIP] Close old components during component updates

shubham1172:

# Description

This PR adds a step to shutdown components that are no more in use. There are two basic changes:
1. Shutdown logic for each component is moved to its own function, and referred in `shutdownComponents` and `shutdownComponentIfExists` (this is similar to init)
2. In `processComponents`, old component is shut down after the new component is processed.

## Issue reference

Please reference the issue this PR will close: #4218

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests
* [X] Unit tests passing
* [X] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#2296


yaron2:
We must add tests (e2e or unit) that validate zero downtime for building blocks like pub/sub, bindings etc.

It would be logical to load the new components first and shutdown the old components after.

pkg/runtime/runtime.go
```diff
@@ -1270,6 +1306,18 @@ func (a *DaprRuntime) initState(s components_v1alpha1.Component) error {
...
 
+func (a *DaprRuntime) shutdownState(name string, state state.Store) error {
+	if closer, ok := state.(io.Closer); ok {
+		if err := closer.Close(); err != nil {
+			err = fmt.Errorf("error closing state store %s: %w", name, err)
+			log.Warn(err)
+			return err
+		}
+	}
+
```
surenderssm, line 1317: NIT : new line can be avoided

pkg/runtime/runtime.go
```diff
@@ -1208,6 +1232,18 @@ func (a *DaprRuntime) initConfiguration(s components_v1alpha1.Component) error {
 	return nil
 }
 
+func (a *DaprRuntime) shutdownConfiguration(name string, store configuration.Store) error {
```
surenderssm, line 1235: Is there any opportunity to move shutdown* (new functions) to a new file within the same `runtime` package?
With current state of 2K+ sloc is single file is overwhelming and it gets little tricky to review/write code.

pkg/runtime/runtime.go
```diff
@@ -1942,53 +2039,68 @@ func (a *DaprRuntime) stopActor() {
...
+func (a *DaprRuntime) shutdownComponentInstance(name string, instance interface{}, category ComponentCategory) error {
+	switch category {
+	case bindingsComponent:
+		if outputBinding, ok := instance.(bindings.OutputBinding); ok {
+			if err := a.shutdownOutputBinding(name, outputBinding); err != nil {
+				return err
+			}
+		}
+		if inputBinding, ok := instance.(bindings.InputBinding); ok {
+			if err := a.shutdownInputBinding(name, inputBinding); err != nil {
```
surenderssm, line 2051:
```suggestion
			return  a.shutdownInputBinding(name, inputBinding)
```

pkg/runtime/runtime.go
shubham1172: Agreed, and AFAIK this is already being tracked. I **can** move the shutdown functions simply to a new file, but I would propose (1) firstly understanding if it's the best way to split code - one way is to group similar functions, another will be extracting out modules. (2) separate that work in another PR

pkg/runtime/runtime.go
shubham1172: We don't want to return when it's `nil`.

pkg/runtime/runtime.go
DeepanshuA: I think we would need to weigh-in pros and cons of what approach we take to modularize code - simply taking out parts to different corresponding files or using Generics or some other way. Also, we would need to ensure that everything is good regression-wise after these changes. Given the time we have now for 1.7 release, separating out work in another PR for modularizing sounds like a better option.

mukundansundar:
since this is a behaviour change ... shouldn't there be E2E for this ?

pkg/runtime/runtime.go
shubham1172: Sorry, I did not get it earlier, fixed!

tests/apps/runtime/app.go
```diff
@@ -107,6 +113,27 @@ func invokeDaprGRPCAPI() error {
...
+		return err
+	}
+	defer r.Body.Close()
+
+	bytes, err := io.ReadAll(r.Body)
+	if err != nil {
+		return err
+	}
+
+	if r.StatusCode < 200 || r.StatusCode >= 400 {
```
surenderssm, line 130: Do we have scenarios with status code < 200?

shubham1172: https://docs.dapr.io/reference/api/state_api/#http-response-1
I expect 200 or 204, both of them are okay.

pkg/runtime/runtime.go
```diff
@@ -1845,10 +1908,45 @@ func (a *DaprRuntime) processComponents() {
 				log.Fatalf(e)
 			}
 			log.Errorf(e)
+		} else if exists {
+			log.Infof("process component %s is replaced with a newer version, the older version will be shutdown", comp.Name)
```
artursouza, line 1912: As we agreed, let's just not do the update of the component in this fix. Reloading components dynamically requires further planning and testing.

artursouza (changes requested):

PR #4378, opened 2022-03-15
Move sentry certificate expiry logging before blocking call

berndverst:

# Description

PR #4328 didn't work completely because the logging goroutine was started after a blocking call. (Thanks to @ItalyPaleAle for root causing this issue.)


berndverst:
@yaron2 @artursouza my sentry logging PR was never able to actually write / flush the logs because the goroutine never ran. This PR changes that (and reduces the number of goroutines too).

artursouza (approved):

PR #4405, opened 2022-03-22
fix/typo: change from traceid to traceparent

1046102779:
Signed-off-by: 1046102779 <seachen@tencent.com>

https://github.com/dapr/components-contrib/pull/1604

in MQ message, `Publish` method  does not need to keep compatibility, `Subscribe` method keeps compatibility.


pkg/runtime/pubsub/cloudevents_test.go
```diff
@@ -44,14 +44,14 @@ func TestNewCloudEvent(t *testing.T) {
...
-			TraceID:         "d",
+			TraceParent:     "d",
 		})
 		assert.NoError(t, err)
 		assert.Equal(t, "a", ce["source"].(string))
 		assert.Equal(t, "b", ce["topic"].(string))
 		assert.Empty(t, ce["data"])
 		assert.Equal(t, "text/plain", ce["datacontenttype"].(string))
-		assert.Equal(t, "d", ce["traceid"].(string))
+		assert.Equal(t, "d", ce["traceparent"].(string))
```
artursouza, line 54: This is a breaking change.

artursouza (changes requested):

pkg/runtime/pubsub/cloudevents_test.go
yaron2: @artursouza it seems like that from this PR's perspective, but the corresponding components-contrib PR retains backward compatibility: https://github.com/dapr/components-contrib/pull/1604.

yaron2: There is a place in this PR that does probably needs to get adjusted for backward compat actually. I'll add a comment.

pkg/runtime/runtime.go
```diff
@@ -1521,12 +1521,15 @@ func (a *DaprRuntime) publishMessageHTTP(ctx context.Context, msg *pubsubSubscri
 	req.WithRawData(msg.data, contenttype.CloudEventContentType)
 	req.WithCustomHTTPMetadata(msg.metadata)
 
-	if cloudEvent[pubsub.TraceIDField] != nil {
-		traceID := cloudEvent[pubsub.TraceIDField].(string)
-		sc, _ := diag.SpanContextFromW3CString(traceID)
-		spanName := fmt.Sprintf("pubsub/%s", msg.topic)
-		ctx, span = diag.StartInternalCallbackSpan(ctx, spanName, sc, a.globalConfig.Spec.TracingSpec)
+	traceParent, _ := cloudEvent[pubsub.TraceParentField].(string)
```
yaron2, line 1524: In order to preserve backwards compat, we need to check for either trace parent or trace ID.

pkg/runtime/runtime.go
```diff
@@ -1642,18 +1645,18 @@ func (a *DaprRuntime) publishMessageGRPC(ctx context.Context, msg *pubsubSubscri
...
-			spanName := fmt.Sprintf("pubsub/%s", msg.topic)
-
-			// no ops if trace is off
-			ctx, span = diag.StartInternalCallbackSpan(ctx, spanName, sc, a.globalConfig.Spec.TracingSpec)
-			ctx = diag.SpanContextToGRPCMetadata(ctx, span.SpanContext())
-		} else {
-			log.Warnf("ignored non-string traceid value: %v", iTraceID)
-		}
-	}
+	traceParent, _ := cloudEvent[pubsub.TraceParentField].(string)
```
yaron2, line 1648: In order to preserve backwards compat, we need to check for either trace parent or trace ID.

yaron2 (changes requested):

pkg/runtime/runtime.go
yaron2: This method name is a bit confusing - this is Dapr sending the message to the consumer.

pkg/runtime/runtime.go
yaron2: This method name is a bit confusing - this is Dapr sending the message to the consumer.

1046102779:
If multiple versions coexist( by `dapr.io/sidecar-image` specified) in a dapr cluster, the publish msg may also need to be compatible with traceid. How do you think it ? @yaron2 @artursouza

yaron2:
> If multiple versions coexist( by `dapr.io/sidecar-image` specified) in a dapr cluster, the publish msg may also need to be compatible with traceid and traceparent. How do you think it ? @yaron2 @artursouza

We're not accounting today for this scenario in our tests, so I would say this is an unsupported and unrecommended scenario. However, if it's an easy fix to maintain backward compat all the way through, please do it.

yaron2:
ping @1046102779

1046102779:
> ping @1046102779

Now or later, there will be some compatibility issues, it is really difficult to do, or the code looks difficult to understand.

For example, making some adjustments now may cause the linkage of Component-contribs and Dapr to cause version incompatibility, which is the most troublesome. It will also make us move slower. For example: Component API Context, Multi namresolution, Opentelemetry, etc.

PR #4407, opened 2022-03-22
Add flag to disable built-in retries

halspang:
# Description

This commit allows a customer to switch over to resiliency
completely. The purpose here is to allow users to enable resiliency
and experiment with it without losing all the retries dapr already
has.

## Issue reference

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Unit tests passing

### DIFF
--- a/pkg/actors/actors.go
+++ b/pkg/actors/actors.go
@@ -107,6 +107,7 @@ type actorsRuntime struct {
 	actorTypeMetadataEnabled bool
 	resiliency               resiliency.Provider
 	storeName                string
+	disableBuiltInRetries    bool
 }
 
 // ActiveActorsCount contain actorType and count of actors each type has.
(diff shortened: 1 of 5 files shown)


mukundansundar:
@halspang Is this feature not for pubsub? Because pubsub does have built in retries.

yaron2:
I don't think we should add this flag, because some aspects of Dapr like service to service retries are there to guarantee stability and functionality specifically for certain cases. If a user disables these retires and fails to provide a policy, or a policy which is at least as aggressive as the internal retires (to which they have zero visibility), this will result in failures and breaking their apps.

yaron2:
@halspang are we ok closing this?

halspang:
> @halspang are we ok closing this?

Yeah, I'll close this and make a follow-up item for 1.8.

PR #4425, opened 2022-03-24
New infrastructure for E2E and perf tests on Azure

ItalyPaleAle:
This PR changes how E2E and perf tests are executed on Azure, using ephemeral infrastructure that is deployed on-demand.

It should address a number of concerns with our existing test infra. Tests should now be:

- More reliable: less flakiness
- Faster: the E2E test run covering Linux + Windows should complete in less than 75 minutes
- More scalable: because we're not limited to 2 pairs of E2E test clusters (2 for Linux and 2 for Windows) and 1 perf test cluster, we can now run as many E2E and perf tests as we want
- Using resources in a more streamlined way, by not maintaining idle clusters
- Not reliant on Docker Hub anymore: all tests use a private Azure Container Registry instance

## Main changes and design decisions

- Test clusters are now created on-demand at the start of each pipeline that needs them. They are deployed using a set of Bicep templates that can be found in `tests/test-infra`.
  - `azure-aks.bicep` deploys an AKS cluster and an associated Azure Container Registry
  - `azure-cosmosdb.bicep` deploys a Cosmos DB cluster
    - `azure-cosmosdb-rbac.bicep` is used to assign a service principal access to Cosmos DB via RBAC. It should allow using Cosmos DB via MSI inside AKS. However, this is currently unused/disabled until https://github.com/dapr/components-contrib/pull/1613 is merged (once that's merged, a little bit more work will be needed to make MSI work).
  - `azure-servicebus.bicep` and `azure-servicebus-topic.bicep` deploy Azure Service Bus and the topics we need
  - `azure.bicep` deploys all the resources above (AKS, ACR, Cosmos DB, Service Bus) in an Azure region, creating all the resources that are needed to test on Linux **or** Windows
  - `azure-all.bicep` deploys **both** Linux and Windows clusters in a subscription. This needs to be used in a subscription-level deployment, and it creates 2 resource groups.
  - The same Bicep templates can also be used to create test clusters manually and on-demand. Instructions have been provided in the README file.
- Each pipeline deploys new resource groups with a "random" name (technically, it's generated from a hash that identifies the pipeline run). Once tests are over, the resource groups and all resources contained in them are automatically destroyed.
- A big change is that all test controllers are now running on Linux, even when testing on a Windows AKS cluster. I believe that this will greatly help with reducing flakiness–and makes tests faster too. The OS that the test controller runs on should be irrelevant anyways.
- The pipelines are now parallelized, so the infrastructure is deployed at the same time as the apps are being built.
- Test clusters are deployed in random Azure regions selected from a list that contains 3 regions currently (WestUS3, EastUS2, CanadaCentral). This is to spread usage given that we've been seeing capacity issues on certain Azure regions. Any region that has availability zones can be used; these 3 were selected as they should not have significant capacity issues at the moment.
- I've also tried implementing some caching, first by caching Go modules (including compiled ones). This should help reducing execution time, especially on Windows.

## Notes

- This PR requires creating a new service principal that has full access to the test Azure subscription (level "Owner" is required because it needs to access Azure RBAC). Someone with admin permissions on this GitHub repo needs to create a new secret with the credentials.
- This PR has been in work for a while, but given that we have to migrate our test infrastructure into a new Azure subscription, I accelerated the work on this as it makes sense to just use the new infrastructure with the new subscription.
- https://github.com/dapr/dapr/pull/4424 is a prerequisite for this PR
- You can see the pipelines … in action (ha ha) in my own test environment (which was using my own Azure test sub):
  - [E2E](https://github.com/ItalyPaleAle/dapr-ci-test/actions/workflows/dapr-test.yml)
  - [Perf](https://github.com/ItalyPaleAle/dapr-ci-test/actions/workflows/dapr-perf.yml)


artursouza (approved):
Impressive work here. Our test workflow is much better.

ItalyPaleAle:
Our new subscription was not enabled for creating DSv2 cores in East US 2, one of the regions. I've opened a support ticket to get that enabled.
If this takes too long, we can simply remove "eastus2" from the list of regions in the meanwhile.

Issue #4432, opened 2022-03-28
Prevent component updates for k8s


shubham1172:
Hot reloading for components is currently unsupported and is planned to be implemented and shipped by v1.8. As discussed in #4218, the k8s components updates were buggy and should be disabled till the hot-reloading feature is completed (for both k8s and standalone).

Till then, in-place component updates should be prevented.

PR #4433, opened 2022-03-28
Prevent component updates for k8s

shubham1172:
Hot reloading is currently unsupported. This PR moves the partial support for component updates behind an always-disabled feature flag. Once hot reloading is completely implemented, it will be moved behind an actual feature flag.

# Description

## Issue reference

Please reference the issue this PR will close: #4432

## Checklist

* [X] Code compiles correctly
* [ ] Created/updated tests
* [X] Unit tests passing

### DIFF
--- a/pkg/runtime/runtime.go
+++ b/pkg/runtime/runtime.go
@@ -190,8 +190,8 @@ type DaprRuntime struct {
 
 	resiliency resiliency.Provider
 
-	// TODO: Remove feature flag once feature is ratified
-	featureRoutingEnabled bool
+	featureRoutingEnabled      bool // TODO: Remove feature flag once feature is ratified
+	featureHotReloadingEnabled bool
 }
 
 type ComponentsCallback func(components ComponentRegistry) error
@@ -259,6 +259,8 @@ func NewDaprRuntime(runtimeConfig *Config, globalConfig *config.Configuration, a
 		shutdownC:                  make(chan error, 1),
 
 		resiliency: resiliencyProvider,
+
+		featureHotReloadingEnabled: false, // TODO: Use an actual feature flag once hot reloading feature is supported.
 	}
 }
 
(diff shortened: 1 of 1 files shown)


pkg/runtime/runtime.go
```diff
@@ -371,10 +373,15 @@ func (a *DaprRuntime) initRuntime(opts *runtimeOpts) error {
 	a.httpMiddlewareRegistry.Register(opts.httpMiddleware...)
 
 	go a.processComponents()
-	err = a.beginComponentsUpdates()
-	if err != nil {
-		log.Warnf("failed to watch component updates: %s", err)
+
+	if a.featureHotReloadingEnabled {
```
artursouza, line 377: Can we have an environment variable instead of an official feature flag? We know it does not work at all today, so why offer as a preview feature?

artursouza (changes requested):

pkg/runtime/runtime.go
shubham1172: Actually it's not even a preview feature (just a bool), but env var makes sense. Let me add it, thanks!

shubham1172: @artursouza since it's unofficial, we don't need any documentation, right?

artursouza: I would not document this environment variable. We are just being extra cautious and not removing the code to being with.

artursouza (dismissed):

yaron2:
@daixiang0 for FYI

pkg/runtime/runtime.go
```diff
@@ -94,6 +94,11 @@ const (
 	bindingsConcurrencyParallel   = "parallel"
 	bindingsConcurrencySequential = "sequential"
 	pubsubName                    = "pubsubName"
+
+	// hot reloading is currently unsupported, but
+	// setting this environment variable restores the
+	// partial hot reloading support for k8s.
+	hotReloadingEnvVar = "DAPR_ALLOW_HOT_RELOADING"
```
yaron2, line 101: Change `ALLOW` to `ENABLE` please.

yaron2 (changes requested):

pkg/runtime/runtime.go
shubham1172: @yaron2 updated.

yaron2 (approved):

PR #4439, opened 2022-03-28
E2E and perf test improvements

ItalyPaleAle:
This PR makes some improvements that are needed since deployment the new E2E test framework late last week. Since then, I've noticed most failures were due to 2 things:

1. Tests running in Canada Central _always_ fail, while tests running in West US 3 succeed. I am really not sure what's going on there, but it looks like it's due to a networking issue between GitHub Actions and Canada Central. Adding a bunch of debugging lines, it looks like the Actions runner fails to connect to apps running on AKS in that region in certain cases (getting timeouts with "context deadline exceeded"). I repro'd the issue, which is somehow intermitted and seems to depend to what IP your services are getting, when running the tests manually from a Codespace, which is in the GitHub network too. I am not sure what's going on, but seems that there's an infra-level issue that it's not worth dealing with.
  - Fix: this PR changes the list of regions removing Canada Central and replacing that with East Asia and East US 2 (which was previously, temporarily removed on Friday). East Asia is interesting because it adds some latency too.
  - Fix part 2: change AKS cluster node to use Dsv3 VMs, which are available for our subscription in East Asia and East US 2 too, without having to wait for Azure support to enable our subscription.
3. In a few cases Azure resources deployments failed due to transient failures with Azure (for example temporary capacity issues). This is not unusual but it was not addressed before.
  - Fix: I'm addressing the issue by re-trying (after a short deploy) the deployment in case of error.

In addition to that, I've added a new pipeline that runs every 6 hours and performs a "garbage collection" of Azure resources that were not automatically deleted. The pipelines should delete all Azure resources (and so far they've done it well), but this adds an extra layer of safety.

The PR also includes a couple more small changes, including some docs updates and adding a flag to disable CosmosDB or Service Bus if needed for manual tests (by default they're enabled).


tests/dapr_tests.mk
```diff
@@ -60,8 +60,6 @@ KUBECTL=kubectl
 
 DAPR_CONTAINER_LOG_PATH?=./dist/container_logs
 
-DAPR_TEST_SECONDARY_NAMESPACE=dapr-tests-2
```
ItalyPaleAle, line 63: The secondary test namespace was not used anywhere, yet it contributed to challenges while running tests manually.

artursouza (approved):

Issue #4451, opened 2022-03-30
When actor is executing long running operation single instance of the actor is executing two actor methods in parallel in two different pods


techgeek03:
## In what area(s)?

/area runtime

/area operator

/area placement

> /area docs

> /area test-and-release

## What version of Dapr?

1.6.0
edge

## Expected Behavior

1. When actor is executing long running operation the runtime should block subsequent call on that actor. 
2. The Dapr sidecar should respect the dapr.io/graceful-shutdown-seconds defined time and still accept requests from the main application

## Actual Behavior

When actor is executing long running operation (5+ seconds), under certain circumstances; single instance of the actor can be executing two methods in parallel in two different pods. "under certain circumstances" in this case means the following:

1. Dapr runtime deployed in Kubernetes with HA enabled.
3. Actor with Actor ID has been created and is placed on a single pod.
4. Actor method has been invoked and is taking long time for this operation to complete. (RPC operation to external service, database call, etc.)
5. While the method is still running; the pod where this actor is placed is evicted/killed by kubernetes or user or just by redeploying the application.
6. The same or other actor method is called while the first method is still running.

There are **two** problems (bugs) now:

1. The actor with the same instance (Actor ID) now is been placed on other pod and has started invoking the requested method while the first method is still running.
2. Since **daprd** sidecar received the **SIGTERM** signal from kubernetes it has shout down the http endpoints and is not receiving any requests. However since the first method is still running (imagine external http call and slow network) the next step in that method can be to use the actor state store or maybe publish event; which are using the dapr SDK; and this will fail since **daprd** is no longer receiving requests.

I have also prepared an repo https://github.com/techgeek03/dapr-actor-reminders/ that has implementation, detailed explanation of the problem and how to reproduce it. The actual explanation of the bug is here https://github.com/techgeek03/dapr-actor-reminders/blob/main/bugs/02/BUG.md. Logs can also be found in the repo.

### Project description

From the high level component architecture displayed [here](https://github.com/techgeek03/dapr-actor-reminders/blob/main/README.md#high-level-components-architecture) we can see that we are having two deployable units:

- Rest API that is exposed thru an ingress and receives requests. The API uses the Dapr SDK and the Dapr Actor Proxy to invoke actors. (Dapr.Testing.WebApi.csproj)
- The dapr actor runtime that hosts the actors. (Dapr.Testing.Actors.Runtime.csproj)

[...]

Issue #4067, opened 2021-12-29
clarify intended behavior during shutdown


trondhindenes:
## In what area(s)?
/area runtime

## Ask your question here
We find dapr pretty amazing, but we find ourselves struggling with how to deal with pod shutdowns.
I'd like to ask the maintainers to document the current expected functionality of dapr, no matter if it's less than perfect, so that we have a reference to validate against. Right now I have no idea how dapr is supposed to behave during pod shutdown so I don't know if the behavior we're seeing is expected or not.

**a bit of background:**
we're using dapr to process messages that may take a relatively long time to complete, often several minutes. We're using rabbitmq as the pubsub broker. Here's my test:
1. send a message to dapr topic which a test app subscribes to
2. the test app will simply sleep for 30 seconds before returning `{'status': 'SUCCESS'}`
3. a few seconds after the test app has received its message, we issue `kubectl rollout restart deployment awesome` which causes the test app pod to go into a `Terminating` state before it's ultimately deleted

**here's what we're seeing (dapr logs):**
```
time="2021-12-29T15:53:25.618876625Z" level=debug msg="rabbitmq pub/sub: publishing message to orders" app_id=awesome instance=awesome-7fb9fdbbc8-9ncrx scope=dapr.contrib type=log ver=1.5.1
time="2021-12-29T15:53:39.441189001Z" level=info msg="Shutting down actor" app_id=awesome instance=awesome-7fb9fdbbc8-9ncrx scope=dapr.runtime type=log ver=1.5.1
time="2021-12-29T15:53:40.437639992Z" level=info msg="dapr shutting down." app_id=awesome instance=awesome-7fb9fdbbc8-9ncrx scope=dapr.runtime type=log ver=1.5.1
time="2021-12-29T15:53:40.437743435Z" level=info msg="Stopping Dapr APIs" app_id=awesome instance=awesome-7fb9fdbbc8-9ncrx scope=dapr.runtime type=log ver=1.5.1
time="2021-12-29T15:53:40.437974282Z" level=info msg="Waiting 5s to finish outstanding operations" app_id=awesome instance=awesome-7fb9fdbbc8-9ncrx scope=dapr.runtime type=log ver=1.5.1
time="2021-12-29T15:53:45.438952469Z" level=info msg="Shutting down all components" app_id=awesome instance=awesome-7fb9fdbbc8-9ncrx scope=dapr.runtime type=log ver=1.5.1
```
[...]

PR #4456, opened 2022-03-30
Change shutdown order of operations

yaron2:
This PR has `daprd` wait for the duration of the specified grace period before cleaning actors and closing the API servers.

Closes https://github.com/dapr/dapr/issues/4451.

*Needs further testing, do not merge*.


artursouza:
We need to change the approach:
+ Stops new incoming messages right away for PubSub, Binding and Service Invocation.
+ Continue allowing outgoing API calls during grace period for all APIs.
+ Allow this to be configurable via K8s annotation.

yaron2:
> We need to change the approach:
> 
> * Stops new incoming messages right away for PubSub, Binding and Service Invocation.
> * Continue allowing outgoing API calls during grace period for all APIs.
> * Allow this to be configurable via K8s annotation.

This is legit, however we need to find a way to shutdown Service Invocation without killing requests that are in-flight: https://github.com/dapr/dapr/issues/4067#issuecomment-1021507479.

sujitdmello:
I realize this is closed but wonder if the proposed changes were ever merged? We are seeing this in our testing while recycling Actor Pods. A few actors don't get the chance to update state before the Pod is terminated because the sidecar has disabled the gRPC interface and is refusing Dapr API calls during the shutdown sequence.

Issue #4457, opened 2022-03-30
Actor samples are not working in dotnet-sdk


amulyavarote:
The way actors report their errors in the runtime has changed.

## In what area(s)?

> /area runtime

> /area operator

> /area placement

> /area docs

> /area test-and-release

## What version of Dapr?

> 1.1.x
> 1.0.x
> edge: output of `git describe --dirty`

## Release Note

RELEASE NOTE:

amulyavarote:
@halspang The issue is created for the actor samples in dotnet-sdk. Can you please add the description here?

halspang:
The reason this is happening is .NET actors communicate errors with a header instead of sending failing requests. The SDK handles the error internally by inspecting for that header. With resiliency, we made that error a real error (so we could retry on it). It seems like .NET didn't like that so I'll fix the issue to no longer return an error after the retry loop.

Thanks for finding this @amulyavarote! I was worried we'd run into stuff like this when we started fiddling with the actor call path.

yaron2:
@halspang Discord seems to be down, can you join a short call with me?

halspang:
@yaron2 - Sure can.

yaron2:
@halspang https://meet.google.com/qbe-eehz-yfc

PR #4458, opened 2022-03-30
Send response back to .NET Actor on error

halspang:
# Description 
Actors implemented in the .NET SDK communicate failures with a
header. Resiliency changed this to be an actual error which is a
change in behavior for .NET actors. This change reverts that as it
was not intentional and restores the intended behavior.

https://github.com/dapr/dapr/issues/4457

## Issue reference

Please reference the issue this PR will close: #4457 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Unit tests passing

### DIFF
--- a/pkg/actors/actors.go
+++ b/pkg/actors/actors.go
@@ -138,6 +138,8 @@ const (
 	incompatibleStateStore = "state store does not support transactions which actors require to save state - please see https://docs.dapr.io/operations/components/setup-state-store/supported-state-stores/"
 )
 
+var ErrDaprResponseHeader = errors.New("error indicated via actor header response")
+
 // NewActors create a new actors runtime with given config.
 func NewActors(
 	stateStore state.Store,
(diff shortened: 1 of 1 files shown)


pkg/actors/actors.go
```diff
@@ -138,6 +138,8 @@ const (
 	incompatibleStateStore = "state store does not support transactions which actors require to save state - please see https://docs.dapr.io/operations/components/setup-state-store/supported-state-stores/"
 )
 
+var ErrInvokeDotnetActor = errors.New("error invoking .NET actor")
```
yaron2, line 141: Can your remove all occurrences specific to .NET? Just because the header itself might be used either now or in the future by other SDKs (perhaps ones that are not in the `dapr` org, for example).

yaron2 (changes requested):

pkg/actors/actors.go
```diff
@@ -434,7 +436,7 @@ func (a *actorsRuntime) callLocalActor(ctx context.Context, req *invokev1.Invoke
 
 	// The .NET SDK signifies Actor failure via a header instead of a bad response.
 	if _, ok := resp.Headers()["X-Daprerrorresponseheader"]; ok {
-		return nil, errors.Errorf("error from a .NET actor")
+		return resp, ErrDaprResponseHeader
```
pkedy, line 439: Just curious, why return `resp` _and_ the error?

halspang: In the case of .NET actors, they communicate the error via the header in the response. If we don't return the response, the .NET actor runtime gets very upset and is a breaking behavior change.

halspang: For the record, before I inspected for this header, this request would've returned normally and we wouldn't have seen any error at all. So this is a return to earlier behavior.

artursouza (approved):

PR #4467, opened 2022-04-01
Updated dev container

ItalyPaleAle:
This PR updates the Dapr dev container with full support for **running on arm64 natively**!

The Action that builds the dev container has been updated to create multi-arch images too

Additionally, it includes other changes:

- Updated base OS from Debian Buster to Bullseye (required for arm64 support)
- Merged changes in dev container from upstream (microsoft/vscode-dev-containers)
- Re-introduced the Microsoft copyright on files that were copied from microsoft/vscode-dev-containers: when the license was changed throughout the repo, files that were copies from the upstream repo had their license changed. That should not have happened, as those files were licensed by Microsoft under a MIT license. I've re-introduced that license header (below the Dapr license)
- Enabling ZSH (which is also on by default in Codespaces)
- Installing additional tools: the GitHub CLI (very useful when working with PRs) and Azure CLI (helpful when working with our E2E tests)
- Docker now uses BuildKit (buildx) by default. This also allows using QEMU and build cross-platform images; QEMU is not included in the container image, but it can be enabled by running the script `/usr/local/share/setup-docker-multiarch.sh` within the container.
- Code is always cloned into `/workspace` now. Before it was sometimes cloned in `/workspace`, and sometimes in the GOPATH (which is probably a bad idea now anyways since we've been using Go modules for a while)

Note: the new Action is VERY SLOW because it uses emulation to build arm64 images. This is not unexpected and can't really be worked around.

Tested scenarios:

- [X] Run on AMD64
- [X] Run on ARM64
- [X] Use Docker-in-Docker
- [X] Use Docker-from-Docker
- [X] Run on Codespaces
  - [X] Use Docker-in-Docker in Codespaces
  - [X] Use Docker-from-Docker in Codespaces


berndverst (dismissed):
Great. Thanks for this.

I verified by building locally on M1 and running directly in privileged mode.

Then I also pushed to my registry and updated the devcontainer json to refer to my image to use this as a DevContainer in VSCode.

This works and correctly returns an architecture of aarm64. CLI and other binaries are also correctly installed as `aarm64`.

ItalyPaleAle:
Converted back to draft because:

1. Need to update the GitHub Action too to build cross-platform images, or it will only build for amd64 (and only those who build from the Dockerfile will be able to use arm64).
1. Need to rebase against release-1.7

Will make the changes then move back to ready

ItalyPaleAle:
Pipeline's been updated. This is ready for review

berndverst (approved):

berndverst:
Can you please open this against `release-1.7` @ItalyPaleAle

berndverst (approved):

Issue #4469, opened 2022-04-04
Fix handling of content-type header in service invocation


artursouza:
Issue created from https://github.com/dapr/dapr/issues/4301#issuecomment-1082278188

From @timoknapp

=======================================================================

Hi @artursouza , @daixiang0 ,

I recently noticed a weird behavior with Dapr. Why is it, that Dapr sets a default Content-Type header of "text/plain;charset=UTF-8" whenever the Content-Type header has not been set by the outgoing request? I am getting weird errors of my upstream services when performing outbound GET requests, which do not consist of a Content-Type header.
Is there any way of telling Dapr not to set such headers by default? 🤔 I would expect Dapr to be transparent for these common headers.

Request when NOT manually setting Content-Type:

Dapr setting "text/plain ;charset=UTF-8" as Content-Type
GET http://127.0.0.1:8080/******/*****/****
{User-Agent=[PostmanRuntime/7.29.0], Host=[127.0.0.1:8080], Content-Type=[text/plain;charset=UTF-8],...
Request when manually setting Content-Type to be 'application/json':

Dapr will not override the Content-Type header but forwards the initially set header to the upstream service.
GET http://127.0.0.1:8080/******/*****/****
{User-Agent=[PostmanRuntime/7.29.0], Host=[127.0.0.1:8080], Content-Type=[application/json],...
FYI: I am using Dapr sidecars with my Azure Container Apps environment.

Regards,
Timo

yaron2:
@timoknapp please notice we enabled a preview feature to disable the default content-type to not break existing clients: https://v1-7.docs.dapr.io/operations/support/support-preview-features/#current-preview-features.

timoknapp:
> @timoknapp please notice we enabled a preview feature to disable the default content-type to not break existing clients: https://v1-7.docs.dapr.io/operations/support/support-preview-features/#current-preview-features.

Hey @yaron2 , can you maybe share a sample snippet on how to enable this feature? I am missing a sample in your docs.

Thanks!

yaron2:
> > @timoknapp please notice we enabled a preview feature to disable the default content-type to not break existing clients: https://v1-7.docs.dapr.io/operations/support/support-preview-features/#current-preview-features.
> 
> Hey @yaron2 , can you maybe share a sample snippet on how to enable this feature? I am missing a sample in your docs.
> 
> Thanks!

You'd need to create this configuration file:

```yaml
apiVersion: dapr.io/v1alpha1
kind: Configuration
[...]

PR #4472, opened 2022-04-04
Fix default content type

yaron2:
Fixes https://github.com/dapr/dapr/issues/4469.

This PR does the following:

1. Removes original and wrong intention of using `application/json` as a default content-type if non was supplied by the user
2. Fixes the default setting of `text/plain` content type by FastHTTP which prevented the original intention from taking place
3. Modifies tests to account for the correct behavior

As requested by @artursouza and agreed upon by myself, we'll be merging this to the `1.7` branch.

### DIFF
--- a/pkg/channel/http/http_channel.go
+++ b/pkg/channel/http/http_channel.go
@@ -179,6 +179,7 @@ func (h *Channel) invokeMethodV1(ctx context.Context, req *invokev1.InvokeMethod
 
 	// Send request to user application
 	resp := fasthttp.AcquireResponse()
+
 	err := h.client.Do(channelReq, resp)
 	defer func() {
 		fasthttp.ReleaseRequest(channelReq)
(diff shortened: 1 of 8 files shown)


timoknapp:
> As requested by @artursouza and agreed upon by myself, we'll be merging this to the `1.7` branch.

Thanks for prioritizing this -  really appreciated!

artursouza (approved):

duglin:
This PR seems to have broken some existing usecases, such as the quickstart service-inv/python one.
While I agree that not defaulting to app/json is the right behavior, it might be a bit disruptive for users if we make this change without some kind of warning period.
Should we consider:
- adding a flag so people can "opt-in" to the new/correct behavior
- but keep the default value of the new flag so the v1.6's behavior is what people will see in v1.7
- during the v1.8 dev cycle decide whether we'll break backwards compatibility in v1.8 by changing the default value of this flag. Then people have a warning period during which they should add code to set the content-type header before v1.8 is released.

By doing something like this we're not rushed into a breaking change but still provide a work-around for any customer who is impacted by the current behavior.

Issue #4426, opened 2022-03-25
[m1 support]Improve go build in dockerfile to support build docker images on m1 macbook


skyao:
## In what area(s)?

 /area test-and-release

## What version of Dapr?

master

## Expected Behavior

Develop on m1 macbook (mac + arm64), update code and build/push docker images to docker hub. Then run e2e tests and performance in k8s (linux + amd64) for test before commit the code to github.

The images should work on k8s (linux + amd64).

## Actual Behavior

The images for performance test don't work and the pod is CrashLoopBackOff with error "exec user process caused: exec format error":

```bash
$ k get pods -A
NAMESPACE     NAME                                     READY   STATUS             RESTARTS   AGE
dapr-tests    testapp-85d8d9db89-p6mms                 0/2     CrashLoopBackOff   13         11m

$ k logs testapp-85d8d9db89-p6mms -n dapr-tests testapp
standard_init_linux.go:228: exec user process caused: exec format error
```

## Steps to Reproduce the Problem

1. Build images on m1 macbook:

    ```bash
   $ make build-perf-app-all

   $ make push-perf-app-all
   ```

2. Run performance test case with remote k8s cluster

## Solution

The cause is that some test images is build by the dockerfile with `go build` command, but without GOOS and GOARCH settings. So the binary built on m1 book can't work on k8s ( linux + amd64).

To resolve this issue, we need to do two updates:

1. pass GOARCH as docker build argement in dapr makefile, such as dapr_tests.mk

    ```makefile
    define genPerfTestAppImageBuild
   .PHONY: build-perf-app-$(1)
   build-perf-app-$(1): check-e2e-env
	$(DOCKER) build -f $(PERF_TESTAPP_DIR)/$(1)/$(DOCKERFILE) $(PERF_TESTAPP_DIR)/$(1)/.
    -t    $(DAPR_TEST_REGISTRY)/perf-$(1):$(DAPR_TEST_TAG) 
   --build-arg GOARCH_ARG=$(GOARCH)
   endef
    ```
2. in the dockerfile with "go build", add GOOS and GOARCH settings

    ```makefile
    ARG GOARCH_ARG=amd64

    WORKDIR /app
    COPY app.go go.mod ./
    RUN go get -d -v && GOOS=linux GOARCH=$GOARCH_ARG go build -o app .
    ```

I verified for performance test cases and it works well.  The test cases of e2e test should do similar updates.

## Release Note

RELEASE NOTE:

[...]

PR #4482, opened 2022-04-06
Improve to support building amd64 images on m1 macbook

skyao:
Signed-off-by: Sky Ao <aoxiaojian@gmail.com>

# Description

Support to build amd64 images in m1 macbook and run k8s in somewhere (like aks).

## Issue reference

https://github.com/dapr/dapr/issues/4426

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing

### DIFF
--- a/tests/apps/perf/tester/Dockerfile
+++ b/tests/apps/perf/tester/Dockerfile
@@ -1,15 +1,19 @@
 FROM golang:1.17 as build_env
 
+ARG GOARCH_ARG=amd64
+
 WORKDIR /app
 COPY app.go go.mod ./
-RUN go get -d -v && go build -o tester .
+RUN go get -d -v && GOOS=linux GOARCH=$GOARCH_ARG go build -o tester .
 
 FROM golang:1.17 as fortio_build_env
+
+ARG GOARCH_ARG=amd64
 
 WORKDIR /fortio
 ADD "https://api.github.com/repos/fortio/fortio/branches/master" skipcache
 RUN git clone https://github.com/fortio/fortio.git
-RUN cd fortio && git checkout v1.16.1 && go build
+RUN cd fortio && git checkout v1.16.1 && GOOS=linux GOARCH=$GOARCH_ARG go build
 
 FROM debian:buster-slim
 #RUN apt update
--- a/tests/apps/perf/actorfeatures/Dockerfile
+++ b/tests/apps/perf/actorfeatures/Dockerfile
@@ -13,12 +13,14 @@
 
 FROM golang:1.17 as build_env
 
+ARG GOARCH_ARG=amd64
+
 WORKDIR /app
 COPY app.go .
 COPY go.mod .
 COPY go.sum .
 RUN go get -d -v
-RUN go build -o app .
+RUN GOOS=linux GOARCH=$GOARCH_ARG go build -o app .
 
 FROM debian:buster-slim
 WORKDIR /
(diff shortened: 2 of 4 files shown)


daixiang0 (approved):

tests/apps/perf/tester/Dockerfile
```diff
@@ -1,15 +1,19 @@
 FROM golang:1.17 as build_env
 
+ARG GOARCH_ARG=amd64
+
 WORKDIR /app
 COPY app.go go.mod ./
-RUN go get -d -v && go build -o tester .
+RUN go get -d -v && GOOS=linux GOARCH=$GOARCH_ARG go build -o tester .
```
mukundansundar, line 7: why hardcode as linux ?

ItalyPaleAle:
I still don't think this is the approach we should take, as this is not going to scale to ALL our Docker images.

I see that you've updated the Dockerfiles for the perf test apps only, because those are the only ones that build the apps within the container itself. However those are only a small subset of the test (e2e or perf) apps we build. The e2e test apps that are written in Go are first built inside the machine "natively", and then copied into the container. The other e2e test apps (like those written in Java or Python) are instead built within the container.

Sadly this approach of setting Go build tags in the Doc kerfiles is specific to some kinds of apps, it can't scale to all apps, and it could become harder to maintain in the future.

The approach I think we should take is based on Docker buildx instead, which does scale to support building anything from anywhere. Good news is that the new devcontainers (which are in the `release-1.7` branch and will be published together with the 1.7 release) offer support for Docker buildx too (just needs to be enabled by running `/usr/local/share/setup-docker-multiarch.sh` within the container).

Looks like our Makefile already has some support for Buildx, but it only enables that if you're targeting a different platform than amd64 (I think those Makefiles were written before ARM laptops became a thing):
https://github.com/dapr/dapr/blob/master/docker/docker.mk#L91-L105

What would need to change is to default to using Buildx for all Docker builds. And then we could pass the target arch with `$TARGET_ARCH`...

skyao:
> I still don't think this is the approach we should take, as this is not going to scale to ALL our Docker images.
> 
> I see that you've updated the Dockerfiles for the perf test apps only, because those are the only ones that build the apps within the container itself. However those are only a small subset of the test (e2e or perf) apps we build. The e2e test apps that are written in Go are first built inside the machine "natively", and then copied into the container. The other e2e test apps (like those written in Java or Python) are instead built within the container.
> 
> Sadly this approach of setting Go build tags in the Doc kerfiles is specific to some kinds of apps, it can't scale to all apps, and it could become harder to maintain in the future.
> 
> The approach I think we should take is based on Docker buildx instead, which does scale to support building anything from anywhere. Good news is that the new devcontainers (which are in the `release-1.7` branch and will be published together with the 1.7 release) offer support for Docker buildx too (just needs to be enabled by running `/usr/local/share/setup-docker-multiarch.sh` within the container).
> 
> Looks like our Makefile already has some support for Buildx, but it only enables that if you're targeting a different platform than amd64 (I think those Makefiles were written before ARM laptops became a thing): https://github.com/dapr/dapr/blob/master/docker/docker.mk#L91-L105
> 
> What would need to change is to default to using Buildx for all Docker builds. And then we could pass the target arch with `$TARGET_ARCH`...

Agree!

I will close this issue and do testing on your solution with buildx.

PR #4499, opened 2022-04-08
Updates to dev container

ItalyPaleAle:
Fixes some minor teething issues with the new image:

1. Setting permissions to the `/workspace` folder to the dapr user
2. Changed the workspace folder to the correct one
3. Removed the setting that makes bash the default shell in the dev container. VS Code / Codespaces will then default to whatever is the user's preferred shell.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -57,7 +57,6 @@
 		"--init"
 	],
 	"settings": {
-		"terminal.integrated.defaultProfile.linux": "bash",
 		"go.toolsManagement.checkForUpdates": "local",
 		"go.useLanguageServer": true,
 		"go.gopath": "/go"
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -61,6 +61,6 @@
 		"go.useLanguageServer": true,
 		"go.gopath": "/go"
 	},
-	"workspaceFolder": "/go/src/github.com/dapr/dapr",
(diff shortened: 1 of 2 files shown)


.devcontainer/devcontainer.json
```diff
@@ -57,11 +57,10 @@
...
 	],
 	"settings": {
-		"terminal.integrated.defaultProfile.linux": "bash",
 		"go.toolsManagement.checkForUpdates": "local",
 		"go.useLanguageServer": true,
 		"go.gopath": "/go"
 	},
-	"workspaceFolder": "/go/src/github.com/dapr/dapr",
-	"workspaceMount": "type=bind,source=${localWorkspaceFolder},target=/go/src/github.com/dapr/dapr",
+	"workspaceFolder": "/workspaces/dapr",
```
artursouza, line 64: We have used github.com/dapr/<repo> as a standard here. Why the change?

artursouza (changes requested):

.devcontainer/devcontainer.json
ItalyPaleAle: The devcontainer in 0.1.7 was already cloning into /workspaces - this part was left to be updated. Previously, the workspace was mounted inside the /go/src folder, but Codespaces was using /workspaces, so we were doing some hacky things to keep everything in sync.

I assume that the requirement for cloning into /go/src was due to a time when we were not using Go modules. Now that we use modules, cloning inside the GOPATH is not recommended anymore AFAIK.

artursouza (approved):

Issue #4509, opened 2022-04-11
Actor client requires an actor state store even if not using actors.


withinboredom:
## In what area(s)?

/area runtime

> /area operator

> /area placement

> /area docs

> /area test-and-release

## What version of Dapr?

- 1.5.2
- 1.6.1
- 1.7.0

## Expected Behavior

An actor client shouldn't need any access to state stores unless it is utilizing a state store.

## Actual Behavior

An actor client needs access to a state store configured for actors, not necessarily the same state store.

## Steps to Reproduce the Problem

The easiest way is to check out this [commit](https://github.com/dapr/php-sdk/pull/130/commits/6daf3ad736273305beed9872abb9ef42fbcc4c02) in the PHP repo where this issue was discovered. You can revert the changes in that example and run `make && make start` (see the example README). After `curl -s http://localhost:8080/method/hello_world/get_count | jq .`, you'll receive the error:

```
{\"errorCode\":\"ERR_ACTOR_RUNTIME_NOT_FOUND\",\"message\":\"actor runtime is not configured\"}"
```

Potentially related issues/PRs: #4273

## Release Note

RELEASE NOTE:

yaron2:
@Taction can you look into this please?

Taction:
@yaron2 Sure, I'll look into this.

Taction:
This is because the `initActors()` func has checked if the actor state store is defined: https://github.com/dapr/dapr/blob/master/pkg/runtime/runtime.go#L1771

To fix this issue, we can add the check of `actorConfig.EntityConfigs.Entities` to tell if it is server-side or client-side.
But consider #4273, we can turn this error into a warning log. If the actor server uses the state store and the actor state store is not defined, the error is `state store does not exist` (and we can add more info to that error).

yaron2:
> This is because the `initActors()` func has checked if the actor state store is defined: https://github.com/dapr/dapr/blob/master/pkg/runtime/runtime.go#L1771
> 
> To fix this issue, we can add the check of `actorConfig.EntityConfigs.Entities` to tell if it is server-side or client-side. But consider #4273, we can turn this error into a warning log. If the actor server uses the state store and the actor state store is not defined, the error is `state store does not exist` (and we can add more info to that error).

That's a great find, and agree about changing the error to a warning log.

Issue #4273, opened 2022-02-18
Improve the error message for when a state store has not been set for actor usage and actors are used


msfussell:
## In what area(s)?
> /area actor runtime
## What version of Dapr?
v1.6
## Expected Behavior
See this doc issues that was raised https://github.com/dapr/docs/issues/2184

This is to add more detail to the error message when you try to save and actor and the statestore has not been configured for actors. It would be good to provide and indication of what you can do to fix the issue. Suggested error can be improve similar to below

{ "errorCode": "ERR_ACTOR_RUNTIME_NOT_FOUND", "message": "The state store is not configured to use the actor runtime. Have you set the - name: actorStateStore value: "true" in your state store component file?" }

## Actual Behavior
{ "errorCode": "ERR_ACTOR_RUNTIME_NOT_FOUND", "message": "actor runtime is not configured" }

See https://github.com/dapr/docs/issues/2184 as an example

PR #4516, opened 2022-04-13
Fix actor client not requires state store.

Taction:
# Description
Fix actor client do not require state store.
Add `Have you set the - name: actorStateStore value: "true" in your state store component file?` to actor state store absent error.

## Issue reference

Please reference the issue this PR will close: #4509 #4273

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing

### DIFF
--- a/pkg/actors/actors.go
+++ b/pkg/actors/actors.go
@@ -497,7 +497,7 @@ func (a *actorsRuntime) GetState(ctx context.Context, req *GetStateRequest) (*St
 
 func (a *actorsRuntime) TransactionalStateOperation(ctx context.Context, req *TransactionalRequest) error {
 	if a.store == nil || a.transactionalStore == nil {
-		return errors.New("actors: state store does not exist or incorrectly configured")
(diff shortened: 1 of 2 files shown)


pkg/actors/actors.go
```diff
@@ -497,7 +497,7 @@ func (a *actorsRuntime) GetState(ctx context.Context, req *GetStateRequest) (*St
 
 func (a *actorsRuntime) TransactionalStateOperation(ctx context.Context, req *TransactionalRequest) error {
 	if a.store == nil || a.transactionalStore == nil {
-		return errors.New("actors: state store does not exist or incorrectly configured")
+		return errors.New("actors: state store does not exist or incorrectly configured. Have you set the - name: actorStateStore value: \"true\" in your state store component file?")
```
yaron2, line 500: Shouldn't this return an error if Entities > 0?

Taction: Actors are using state store but actor state store does not exist, regardless of the Entities count, the error should be always returned. WDYT?

yaron2:
Thanks @Taction

Issue #4237, opened 2022-02-11
increase mTLS service invocation retry interval


berndverst:
In order to aid with the replacing of TLS certificates, this is a proposal to try mTLS service invocation requests for a period of 60 seconds (?) in case a certificate mismatch is encountered.

This would allow for more time to restart applications without errors in case of Dapr control plane TLS certificate upgrades

artursouza:
This should be done in the context of resiliency policies.

PR #4542, opened 2022-04-21
[WIP] Increase retry time interval for service invocation

pravinpushkar:

# Description

Increasing the retry time interval for service invocation to 60 sec.

Output logs - (**Some of the below logs statements are added just for debugging purpose and are not actually present in the code**)

**Logs from sidecar when Caller and Callee both apps are present** - 
 in Invoke direct_messaging, appid= nodeapp, err=%!s(<nil>), namespace=default<nil>
app.id=nodeapp, app.namespace=default
 in invokeWithRetry, numRetries :  3
in invokeWithRetry, backoffInterval:  1m0s
Got response :  200

**Logs from sidecar when Callee app became unavailable** - 
app.id=nodeapp, app.namespace=default
in invokeWithRetry, numRetries :  3
in invokeWithRetry, backoffInterval:  1m0s
retry count: 1, grpc call failed, ns: default, addr: nodeapp-dapr.default.svc.cluster.local:50002, appid: nodeapp, **sleepTime: 1m0.0005653s**, err: rpc error: code = Unavailable desc = last connection error: connection error: desc = "transport: Error while dialing dial tcp 172.17.0.12:50002: connect: connection refused"

## Issue reference

Please reference the issue this PR will close: #4237 

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing


pkg/retry/retry.go
```diff
@@ -16,6 +16,6 @@ package retry
 import "time"
 
 const (
-	DefaultLinearBackoffInterval = time.Second
+	DefaultLinearBackoffInterval = time.Minute
```
pravinpushkar, line 19: Same retry is used for remote actor invocation also here - https://github.com/dapr/dapr/blob/cc4df6b440b41f3b51e8253cc08dffb286ac1082/pkg/actors/actors.go#L332

Please let me know if retry interval for both needs to be different.

CrazyHZM, line 19: I am more worried that the retry interval is so long, resulting in high CPU usage.

beiwei30: agree, I guess we should use time.Minute for TLS setup only instead of changing the default value.

pravinpushkar: @beiwei30 Are you suggesting to have a separate backoffInterval for mtls? Currently this retry is being used in this file [direct_messaging.go](https://github.com/dapr/dapr/blob/master/pkg/messaging/direct_messaging.go), which seems to be independent of mTLS setting. Which files do you suggest I should take a look at?

berndverst: You should use a separate instance of `dapr/kit` retries just for service invocation retries.

Please do not change a default that is used elsewhere - that is not the correct approach.

https://github.com/dapr/kit/blob/main/retry/retry.go

berndverst (changes requested):

mukundansundar:
@artursouza @berndverst @yaron2 I believe a 1 minute retry interval is too high ... Can it be somewhat like 5 seconds or 10 seconds ?

halspang:
@mukundansundar - This is specifically for handling something like cert rotation which we expect to have a longer downtime. Due to that though, it should definitely be a much more specific change than this so we don't affect all retries.

pkg/retry/retry.go
beiwei30: @pravinpushkar, correct, we shouldn't change this default value. Bernd made a very good point, pls. follow his suggestion.

pravinpushkar:
There are issue/proposal opened for better root cert rotation -  https://github.com/dapr/dapr/issues/5958, https://github.com/dapr/dapr/issues/5756

Once these are implemented, increasing time interval for service invocation will not be of any use as this issues is specifically for taking care of service invocation when the cert is being rotated.

pravinpushkar:
Closing this for now. If anyone feels otherwise , feel free to re-open this.

Issue #4544, opened 2022-04-22
API logging enabled crashes sidecar for Service Invocation


yaron2:
How to reproduce:

1. Install Dapr 1.7.1
2. Run the Dapr [hello-kubernetes tutorial](https://github.com/dapr/quickstarts/tree/master/tutorials/hello-kubernetes), and add `dapr.io/enable-api-logging: "true"` to the node and python app YAMLs.
3. Deploy the apps and see node app crashing

Stack trace:

```
github.com/dapr/dapr/pkg/grpc.(*server).getGRPCAPILoggingInfo.func1({0x4510638, 0xc000bbb050}, {0x3a60ba0, 0xc000bf0190}, 0xc000be24a0, 0xc0006d44c8)
	/home/runner/work/dapr/dapr/pkg/grpc/server.go:318 +0xc7
github.com/grpc-ecosystem/go-grpc-middleware.ChainUnaryServer.func1.1.1({0x4510638, 0xc000bbb050}, {0x3a60ba0, 0xc000bf0190})
	/home/runner/go/pkg/mod/github.com/grpc-ecosystem/go-grpc-middleware@v1.3.0/chain.go:25 +0x3a
github.com/dapr/dapr/pkg/diagnostics.(*grpcMetrics).UnaryServerInterceptor.func1({0x4510638, 0xc000bbb050}, {0x3a60ba0, 0xc000bf0190}, 0xc000be24a0, 0xc000be24c0)
	/home/runner/work/dapr/dapr/pkg/diagnostics/grpc_monitoring.go:181 +0xf8
github.com/grpc-ecosystem/go-grpc-middleware.ChainUnaryServer.func1.1.1({0x4510638, 0xc000bbb050}, {0x3a60ba0, 0xc000bf0190})
	/home/runner/go/pkg/mod/github.com/grpc-ecosystem/go-grpc-middleware@v1.3.0/chain.go:25 +0x3a
github.com/dapr/dapr/pkg/diagnostics.GRPCTraceUnaryServerInterceptor.func1({0x4510638, 0xc000bba210}, {0x3a60ba0, 0xc000bf0190}, 0xc000be24a0, 0xc000be24e0)
	/home/runner/work/dapr/dapr/pkg/diagnostics/grpc_tracing.go:73 +0x407
github.com/grpc-ecosystem/go-grpc-middleware.ChainUnaryServer.func1.1.1({0x4510638, 0xc000bba210}, {0x3a60ba0, 0xc000bf0190})
[...]

PR #4545, opened 2022-04-22
fix nil pointer error with api-logging enabled

yaron2:
Fixes https://github.com/dapr/dapr/issues/4544.

This PR also enabled API logging for all e2e tests.

### Root Cause Analysis

Service Invocation calls go through the internal gRPC server on the callee side. The `infoLogger` was not set on the internal server, which caused a nil pointer panic error for service invocation calls.

### DIFF
--- a/pkg/grpc/server.go
+++ b/pkg/grpc/server.go
@@ -107,6 +107,7 @@ func NewInternalServer(api API, config ServerConfig, tracingSpec config.TracingS
 		renewMutex:       &sync.Mutex{},
 		kind:             internalServer,
 		logger:           internalServerLogger,
+		infoLogger:       apiServerInfoLogger,
 		maxConnectionAge: getDefaultMaxAgeDuration(),
 		proxy:            proxy,
 	}
--- a/pkg/grpc/server.go
+++ b/pkg/grpc/server.go
@@ -315,7 +315,7 @@ func shouldRenewCert(certExpiryDate time.Time, certDuration time.Duration) bool 
 func (s *server) getGRPCAPILoggingInfo() grpc_go.UnaryServerInterceptor {
 	return func(ctx context.Context, req interface{}, info *grpc_go.UnaryServerInfo, handler grpc_go.UnaryHandler) (interface{}, error) {
 		if s.infoLogger != nil && info != nil {
-			s.infoLogger.Info("gRPC API Called: ", *info)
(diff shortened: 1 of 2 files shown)


tests/platforms/kubernetes/kubeobj.go
```diff
@@ -77,6 +77,7 @@ func buildDaprAnnotations(appDesc AppDescription) map[string]string {
 			"dapr.io/sidecar-readiness-probe-threshold": "15",
 			"dapr.io/sidecar-liveness-probe-threshold":  "15",
 			"dapr.io/enable-metrics":                    strconv.FormatBool(appDesc.MetricsEnabled),
+			"dapr.io/enable-api-logging":                "true",
```
mukundansundar, line 80: +1

artursouza (approved):

mukundansundar (approved):
lgtm. contingent on tests passing ...

yaron2:
> lgtm. contingent on tests passing ...

See my comment on the release channel please. I think the logic here should change to not log internal API requests as they carry no meaning for the user, ie:

```
"gRPC API Called: {0xc000b3c000 /dapr.proto.internals.v1.ServiceInvocation/CallLocal}"
```

mukundansundar:
@yaron2 Why is the memory reference being logged ? Is it being logged even for normal gRPC API calls?

yaron2:
> @yaron2 Why is the memory reference being logged ? Is it being logged even for normal gRPC API calls?

Needs input from original author here.

amulyavarote:
Yes. We should not print the memory reference. I agree with @mukundansundar

pkg/grpc/server.go
```diff
@@ -314,8 +314,8 @@ func shouldRenewCert(certExpiryDate time.Time, certDuration time.Duration) bool
 
 func (s *server) getGRPCAPILoggingInfo() grpc_go.UnaryServerInterceptor {
 	return func(ctx context.Context, req interface{}, info *grpc_go.UnaryServerInfo, handler grpc_go.UnaryHandler) (interface{}, error) {
-		if info != nil {
-			s.infoLogger.Info("gRPC API Called: ", *info)
+		if s.infoLogger != nil && info != nil {
```
artursouza, line 317: Much better. Thanks!

artursouza (approved):

mukundansundar (approved):
lgtm. contingent on tests passing ...

PR #4555, opened 2022-04-26
Use revive instead of golint

pigletfly:
Signed-off-by: pigletfly <wangbing.adam@gmail.com>

# Description

## Issue reference

part of https://github.com/dapr/components-contrib/issues/1687

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Unit tests passing


yaron2:
cc @artursouza

artursouza:
I saw this PR before. What is the advantage of revive over golint?

yaron2:
> dapr/components-contrib#1687

It appears that golint is deprecated. @shubham1172 seems to have replaced it in the CLI: https://github.com/dapr/cli/issues/960.

yaron2:
@pigletfly please see there's one test that's failing: https://github.com/dapr/dapr/runs/6236197796?check_suite_focus=true.

ItalyPaleAle:
Took a review since I have a vested interest in this PR now :)

Looks like the issue is caused by this change here: https://github.com/dapr/dapr/pull/4555/files#diff-30164ead47f86ee4b8e0c078aa25c30a60a6f241ee7d0f19ac778eb360854879R23

Before, in case there was an error importing the TLS certificate, it would just return `nil` as error and essentially not use TLS. This is what the tests are asserting too: https://github.com/dapr/dapr/blob/master/pkg/credentials/grpc_test.go#L21

Now, probably because of the linter's own complaining, an error is returned instead.

Was the previous behavior on purpose or accidental? If TLS certs are invalid, do we want to return an error or silently ignore the failure and proceed without TLS?

CC: @pigletfly @yaron2

yaron2:
> https://github.com/dapr/dapr/pull/4555/files#diff-30164ead47f86ee4b8e0c078aa25c30a60a6f241ee7d0f19ac778eb360854879R23

Accidental.

ItalyPaleAle:
Thanks for confirming.

I've opened a PR against @pigletfly 's PR: https://github.com/pigletfly/dapr/pull/1 I think you _may_ have permissions to merge it yourself since you're a maintainer here, otherwise we'll just wait for @pigletfly to be back :)

pigletfly:
@ItalyPaleAle thank you for your review, I'm back from vacation.

ItalyPaleAle:
@pigletfly Welcome back :) Not sure what happened but the Git history got messed up and the error is back.

I've updated https://github.com/pigletfly/dapr/pull/1/ and this should fix the test failures

artursouza:
I have merged @ItalyPaleAle's PR already.

artursouza (approved):

artursouza:
No need to assign PRs. The author is given credit as per the release notes script.

PR #4570, opened 2022-04-29
Update to Go 1.18

ItalyPaleAle:
This PR updates Go to 1.18 in all places, including setting the minimum version in `go.mod` files and in all the CI's and containers.

A number of dependencies, most notably the newest Azure SDKs, require Go 1.18 (because of generics) and we will need to update the version of Go used by Dapr to be able to rely on them.

Companion components-contrib PR: https://github.com/dapr/components-contrib/pull/1697


ItalyPaleAle:
This PR is currently failing because the version of the linter we're using in the Action is outdated and doesn't support Go 1.18.

As per discussion on Discord, we'll wait until https://github.com/dapr/dapr/pull/4555 is merged, which changes the linter to revive.

ItalyPaleAle:
The linter has been updated and this PR is ready for review.

tests/e2e/utils/helpers.go
```diff
@@ -128,10 +128,11 @@ func HTTPGetNTimes(url string, n int) ([]byte, error) {
 
 // httpGet is a helper to make GET request call to url.
 func httpGet(url string, timeout time.Duration) ([]byte, error) {
-	resp, err := httpGetRaw(url, timeout) //nolint
+	resp, err := httpGetRaw(url, timeout)
 	if err != nil {
 		return nil, err
 	}
+	defer resp.Body.Close()
```
artursouza, line 135: scary shit. Nice catch.

artursouza (approved):

tests/e2e/utils/helpers.go
ItalyPaleAle: It was actually ok. Before, this was inside `extractBody` so it was being executed.

But......

1. Being inside `extractBody` meant that the linter would not see it so it was marked as error. This came from the `bodyclose` linter.
2. However, `bodyclose` is [not working with Go 1.18 yet](https://github.com/golangci/golangci-lint/issues/2649).
3. This means that the `//nolint` rule that was on that line was actually not doing anything, because no linter was finding error there (because `bodyclose` is not enabled until it's fixed for Go 1.18)
4. Because `//nolint` was not doing anything, another linter, `nolintlint` was throwing an error saying that there was a `//nolint` command that was necessary

Now, I could have just disabled `nolintlint`, but I figured that solving the problem at the root (and making it so we won't need `//nolint:bodyclose`) would just be easier.

PS: Did I say how much I ❤️ linters

PR #4575, opened 2022-05-02
Placement chart: do not allow arbitrary replica count

ItalyPaleAle:
This changes the behavior of the replica count of the placement service Helm chart. Now, it's one of two options:

- If `global.ha.enabled` is true, use 3 instances of the placement service
- Otherwise, use 1 instance only

There's no point in allowing any other option:

1. Even numbers should not be allowed because with Raft you always need an odd number of nodes (a replica count of 2 would not tolerate the failure of a single node anyways, giving the same reliability of a replica count of 1; likewise, replica count of 4 has same reliability as 3). It's assumed that > 3 replicas aren't necessary (and they'd be incredibly chatty too).
2. When in HA mode, we need to set a value for the `initial-cluster` option. This is currently done by a template that has hardcoded values for 3 replicas. As such, running with 2 replicas (besides not being useful) would also require changing that value.


yaron2 (approved):

yaron2:
MSTM (makes sense to me, new term).

LGTM on PR.

daixiang0 (dismissed):

ItalyPaleAle:
Made one extra change, as I realized that the `replicaCount` option was still listed in the `values.yaml` file. Should not have changed the substance.

artursouza (dismissed):
This can break deployments to production for existing users.

yaron2:
> This can break deployments to production for existing users.

It might change the number of replicas from, say, `4` to `3` if HA mode is specified. What do you mean specifically by breaking a deployment? Do you mean this causes downtime? placement not working?

ItalyPaleAle:
@artursouza  and I chatted about this in-person. There may be some small risk of things breaking, but as it is today, the current experience is broken.

**if `global.ha.enabled` is false**

In this case, the placement service is created with [as many replicas as `dapr_placement.replicaCount`](https://github.com/dapr/dapr/blob/master/charts/dapr/charts/dapr_placement/templates/dapr_placement_deployment.yaml#L11). However, [they all work as standalone instances and are not clustered together](https://github.com/dapr/dapr/blob/master/charts/dapr/charts/dapr_placement/templates/dapr_placement_deployment.yaml#L110-L123). Depending on how things are load-balanced, this may cause issues where a placement server here doesn't know what the placement server there knows.

**if `global.ha.enabled` is true**

In this case, the placement service is created with [as many replicas as `global.ha.replicaCount`](https://github.com/dapr/dapr/blob/master/charts/dapr/charts/dapr_placement/templates/dapr_placement_deployment.yaml#L9). This means that it's scaled as much as every other service (injector, sentry, operator...).

This is not good as, like I mentioned earlier, for HA mode you should have exactly 3 nodes. Having 2 doesn't offer any more resiliency to failures than 1 (actually, it doubles the likelihood that there's a failure, because now you have 2 services that can each fail independently and even if just goes down, the cluster stops). Having 4 has the same resiliency as having 3, so it's a waste of an extra instance. Having 5 would increase the resiliency, but would also increase the chattiness between the services by a lot (they all ping each other every .5 seconds, and although the pings are small packets, they add up).

There's one other aspect. Although with `global.ha.replicaCount` you could have deployed 5 instances of placement before, the list of nodes in the cluster [is hardcoded](https://github.com/dapr/dapr/blob/master/charts/dapr/charts/dapr_placement/templates/dapr_placement_deployment.yaml#L113-L114) and contains [exactly 3 nodes](https://github.com/dapr/dapr/blob/master/charts/dapr/charts/dapr_placement/templates/_helpers.tpl#L37-L39). This means that one could have deployed 5 instances of placement, but only 3 were talking to each other. The other 2 would have an undefined behavior (they may crash actually, but I haven't tried that).

yaron2:
> @artursouza and I chatted about this in-person. There may be some small risk of things breaking, but as it is today, the current experience is broken.
> 
> **if `global.ha.enabled` is false**
> 
> In this case, the placement service is created with [as many replicas as `dapr_placement.replicaCount`](https://github.com/dapr/dapr/blob/master/charts/dapr/charts/dapr_placement/templates/dapr_placement_deployment.yaml#L11). However, [they all work as standalone instances and are not clustered together](https://github.com/dapr/dapr/blob/master/charts/dapr/charts/dapr_placement/templates/dapr_placement_deployment.yaml#L110-L123). Depending on how things are load-balanced, this may cause issues where a placement server here doesn't know what the placement server there knows.
> 
> **if `global.ha.enabled` is true**
> 
> In this case, the placement service is created with [as many replicas as `global.ha.replicaCount`](https://github.com/dapr/dapr/blob/master/charts/dapr/charts/dapr_placement/templates/dapr_placement_deployment.yaml#L9). This means that it's scaled as much as every other service (injector, sentry, operator...).
> 
> This is not good as, like I mentioned earlier, for HA mode you should have exactly 3 nodes. Having 2 doesn't offer any more resiliency to failures than 1 (actually, it doubles the likelihood that there's a failure, because now you have 2 services that can each fail independently and even if just goes down, the cluster stops). Having 4 has the same resiliency as having 3, so it's a waste of an extra instance. Having 5 would increase the resiliency, but would also increase the chattiness between the services by a lot (they all ping each other every .5 seconds, and although the pings are small packets, they add up).
> 
> There's one other aspect. Although with `global.ha.replicaCount` you could have deployed 5 instances of placement before, the list of nodes in the cluster [is hardcoded](https://github.com/dapr/dapr/blob/master/charts/dapr/charts/dapr_placement/templates/dapr_placement_deployment.yaml#L113-L114) and contains [exactly 3 nodes](https://github.com/dapr/dapr/blob/master/charts/dapr/charts/dapr_placement/templates/_helpers.tpl#L37-L39). This means that one could have deployed 5 instances of placement, but only 3 were talking to each other. The other 2 would have an undefined behavior (they may crash actually, but I haven't tried that).

This is a good summary of things that were all known for a very long time :)

My question was about the scenarios that could break (meaning, cause temporary or permanent downtime) based on the changes made in this PR.

ItalyPaleAle:
I am not sure about what would break. What I was trying to say and didn't do a good job at putting forward, is that I do not think this PR would break things that are working today. In one way or another, customers that were relying on behaviors different from what this PR was setting would have noticed failures/errors.

yaron2:
> I am not sure about what would break. What I was trying to say and didn't do a good job at putting forward, is that I do not think this PR would break things that are working today. In one way or another, customers that were relying on behaviors different from what this PR was setting would have noticed failures/errors.

Good, this aligns with what I thought originally (that it doesn't actually break users). That's why I asked @artursouza for his opinion in case there's something I missed.

yaron2 (approved):

PR #4577, opened 2022-05-03
Test for placement service in HA

ItalyPaleAle:
Adds tests to ensure that the placement service behaves correctly in HA mode, being able to withstand the failure of nodes and them coming back online.

### DIFF
--- a/pkg/placement/ha_test.go
+++ b/pkg/placement/ha_test.go
@@ -1,0 +1,193 @@
+package placement
+
+import (
+	"fmt"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/dapr/dapr/pkg/placement/raft"
+)
+
+func TestPlacementHA(t *testing.T) {
+	testMembers := []*raft.DaprHostMember{
+		{
+			Name:     "127.0.0.1:3031",
+			AppID:    "testmember1",
+			Entities: []string{"red"},
+		},
+		{
+			Name:     "127.0.0.1:3032",
+			AppID:    "testmember2",
+			Entities: []string{"blue"},
+		},
+		{
+			Name:     "127.0.0.1:3033",
+			AppID:    "testmember3",
+			Entities: []string{"red", "blue"},
+		},
+	}
+
+	// Create 3 raft servers
+	raftServers := make([]*raft.Server, 3)
+	peers := make([]raft.PeerInfo, 3)
+	for i := 0; i < 3; i++ {
+		peers[i] = raft.PeerInfo{
+			ID:      fmt.Sprintf("mynode-%d", i),
+			Address: fmt.Sprintf("127.0.0.1:%d", 6100+i),
+		}
+	}
+	for i := 0; i < 3; i++ {
+		raftServers[i] = createRaftServer(t, i, peers)
+	}
+
+	// Run tests
+	var leader = -1
+	t.Run("elects leader with 3 nodes", func(t *testing.T) {
+		leader = findLeader(t, raftServers)
+	})
+	if leader == -1 {
+		// If there's no leader, no point in running the other tests
+		t.FailNow()
+	}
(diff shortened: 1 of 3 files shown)


pkg/placement/ha_test.go
```diff
@@ -0,0 +1,193 @@
...
+		retrieveValidState(t, raftServers[follower], testMembers[0])
+	})
+	t.Run("new leader after leader fails", func(t *testing.T) {
+		raftServers[leader].Shutdown()
+		raftServers[leader] = nil
+		time.Sleep(2500 * time.Millisecond)
+		leader = findLeader(t, raftServers)
+		raftServers[leader].FSM().State()
+	})
+	t.Run("set and retrieve state in leader", func(t *testing.T) {
```
halspang, line 72: Can this have a unique name to distinguish it from the previous test case? "set and retrieve state in leader after reelection"?

pkg/placement/ha_test.go
```diff
@@ -0,0 +1,193 @@
...
+		time.Sleep(2500 * time.Millisecond)
+		follower := (leader + 1) % 3
+		retrieveValidState(t, raftServers[follower], testMembers[0])
+	})
+	t.Run("new leader after leader fails", func(t *testing.T) {
+		raftServers[leader].Shutdown()
+		raftServers[leader] = nil
+		time.Sleep(2500 * time.Millisecond)
+		leader = findLeader(t, raftServers)
+		raftServers[leader].FSM().State()
```
halspang, line 70: What's the purpose of this call? I know `findLeader` will fail if no leader is found, but do we need to validate whatever the output of this is? Or are we expecting it to fail the test if it's nil?

halspang (commented):
Looks good overall, just 2 small things.

pkg/placement/ha_test.go
ItalyPaleAle: Fixed

pkg/placement/ha_test.go
ItalyPaleAle: Oh, good catch. That was a leftover thing that I should have deleted.

ItalyPaleAle:
(Can you re-trigger the CI, please.... something failed there)

artursouza (approved):

artursouza (changes requested):

pkg/placement/ha_test.go
```diff
@@ -0,0 +1,192 @@
...
+		},
+	}
+
+	// Create 3 raft servers
+	raftServers := make([]*raft.Server, 3)
+	peers := make([]raft.PeerInfo, 3)
+	for i := 0; i < 3; i++ {
+		peers[i] = raft.PeerInfo{
+			ID:      fmt.Sprintf("mynode-%d", i),
+			Address: fmt.Sprintf("127.0.0.1:%d", 6100+i),
```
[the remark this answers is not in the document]
ItalyPaleAle: Fixed in both the new tests and the old one (the non-HA test that currently exists was using a hardcoded port too)

pkg/placement/ha_test.go
```diff
@@ -0,0 +1,201 @@
...
+	ports, err := daprtesting.GetFreePorts(3)
+	if err != nil {
+		log.Fatalf("failed to get 3 ports: %v", err)
+		return
+	}
+
+	// Note that ports below are unused (i.e. no service is started on those ports), they are just used as identifiers with the IP address
+	testMembers := []*raft.DaprHostMember{
+		{
+			Name:     "127.0.0.1:3031",
```
artursouza, line 25: These are also hardcoded.

artursouza (changes requested):

pkg/placement/ha_test.go
ItalyPaleAle: As per the comment, those are unused. We just need an IP:PORT to use as identifier for objects stored in Raft.

artursouza (approved):

PR #4582, opened 2022-05-05
Refactorings to sentry and other services 

ItalyPaleAle:
Background: I was working on the sentry service and all of a sudden I went down a rabbit hole of small refactorings and fixes, primarily on sentry but also on other services except daprd 🙃

This is a collection of small fixes and refactorings:

- Sentry can now import Ed25519 keys and certificates (which must be generated externally)
    - Why: read on [ECDSA vs EdDSA](https://crypto.stackexchange.com/questions/60383/what-is-the-difference-between-ecdsa-and-eddsa) and note that some people prefer to use non-NIST curves (for reasons explained there too). Note however that sentry continues to sign issue certificates with P-256 keys only (for now?)
- Fixed: loading of TLS certs in operator and placement is more reliable
    - Fixed a race condition that may make certs not be loaded in some cases
    - Continuously try to load certs and not give up after the first signal from the watcher
    - In placement svc, fixed a goroutine leak
- Fixed: shutting down sentry, operator, placement, and injector should not panic when the healthz server is shut down gracefully (by ignoring `ErrServerClosed`)
- Fixed: rare crash in sentry when reading invalid certificates
- Refactored some code in sentry to improve code readability and remove some extra background goroutines.


Makefile
```diff
@@ -229,6 +229,7 @@ docker-deploy-k8s: check-docker-env check-arch
 		--set dapr_placement.logLevel=debug --set dapr_sidecar_injector.sidecarImagePullPolicy=$(PULL_POLICY) \
 		--set global.imagePullPolicy=$(PULL_POLICY) --set global.imagePullSecrets=${DAPR_TEST_REGISTRY_SECRET} \
 		--set global.mtls.enabled=${DAPR_MTLS_ENABLED} \
+		--set dapr_sentry.logLevel=debug \
```
ItalyPaleAle, line 232: Remove this

artursouza (commented):
My questions are: is this backwards compatible with existing clusters with an existing cert? Does it allow rollback to the previous version?

ItalyPaleAle:
What do you mean, @artursouza? This did not introduce new functionality so it should be fully backwards-compatible.

Just fixed some bugs and cleaned up goroutines :)

artursouza (approved):

pravinpushkar:
> What do you mean, @artursouza? This did not introduce new functionality so it should be fully backwards-compatible.
> 
> Just fixed some bugs and cleaned up goroutines :)
> 
> Also, @pravinpushkar requested to make one method exported so the CLI can use it.

Yes, Exporting a method for generating new certificate would hide the internal implementation for it to be used in CLI during certificate rotation.

PR #4597, opened 2022-05-09
Update Debug Image installation of dlv

berndverst:

# Description

Since Go 1.18 `go install` must be used to install go libraries outside of modules. This change is required to install `dlv` correctly. Additionally installing outside of modules requiring specifying a specific version. This change picks the tag associated with the latest release to install.

### DIFF
--- a/docker/Dockerfile-debug
+++ b/docker/Dockerfile-debug
@@ -2,7 +2,7 @@
 
 FROM golang:latest AS golang
 ENV GOPROXY=https://goproxy.io,direct
-RUN CGO_ENABLED=0 go get -ldflags '-s -w -extldflags -static' github.com/go-delve/delve/cmd/dlv
+RUN CGO_ENABLED=0 go install -ldflags '-s -w -extldflags -static' github.com/go-delve/delve/cmd/dlv@latest
 
 FROM gcr.io/distroless/static:nonroot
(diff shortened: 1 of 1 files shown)


docker/Dockerfile-debug
```diff
@@ -2,10 +2,10 @@
 
 FROM golang:latest AS golang
 ENV GOPROXY=https://goproxy.io,direct
-RUN CGO_ENABLED=0 go get -ldflags '-s -w -extldflags -static' github.com/go-delve/delve/cmd/dlv
+RUN export DELVETAG=$(curl -s https://api.github.com/repos/go-delve/delve/releases/latest | grep "tag_name" | cut -f2 -d':' | tr -d ' ",') && CGO_ENABLED=0 go install -ldflags '-s -w -extldflags -static' github.com/go-delve/delve/cmd/dlv@$DELVETAG
```
ItalyPaleAle, line 5: Any reason why this couldn't have been `go install github.com/go-delve/delve/cmd/dlv@latest`?

artursouza (approved):

docker/Dockerfile-debug
berndverst: The reason is that I didn't know this is a thing... hahaha

yaron2 (approved):

Issue #4227, opened 2022-02-09
Switch from Service Account Tokens to Bound Tokens


artursouza:
## In what area(s)?

 /area runtime

## Describe the feature
In some setups, the application's POD must run without access to K8s APIs. On the other hand, the Dapr sidecar crashes as in #3667. The sidecar should still work without access to K8s API, such as when `automountServiceAccountToken: false` is set at the pod.

## Release Note

RELEASE NOTE: Allow sidecar to run in pods with `automountServiceAccountToken: false`

yaron2:
Without a service account token, there is zero validation on the pod identity in Kubernetes and this presents a risk for clusters that users control (99.9% of Dapr use cases I imagine).

What needs to be done is the same thing Linkerd did and switch to Bound Tokens: https://linkerd.io/2021/12/28/using-kubernetess-new-bound-service-account-tokens-for-secure-workload-identity/.

But we should allow no token verification at all.

artursouza:
@yaron2 I think the work here might be as simple as disabling (via configuration or automatically) the K8s secret store, which is the root cause of the crash. This way, users run on K8s while the app POD has `automountServiceAccountToken: false`. Do you see another use case for the sidecar to depend on the K8s API?

yaron2:
> @yaron2 I think the work here might be as simple as disabling (via configuration or automatically) the K8s secret store, which is the root cause of the crash. This way, users run on K8s while the app POD has `automountServiceAccountToken: false`. Do you see another use case for the sidecar to depend on the K8s API?

Dapr service invocation (with mTLS) cannot operate when `automountServiceAccountToken` is set to `false`.

yaron2:
@artursouza the cause of the crash in https://github.com/dapr/dapr/issues/3667 is no longer relevant. The Dapr sidecar today makes no k8s API calls. The secrets now are fetched by the Operator and delivered to the sidecar.

yaron2:
> @artursouza the cause of the crash in #3667 is no longer relevant. The Dapr sidecar today makes no k8s API calls. The secrets now are fetched by the Operator and delivered to the sidecar.

Well, there's an exception there. If the app uses the Secrets API, then the sidecar will call the k8s API as part of the component logic. We need to make the k8s secret component explicit, and only if applicable to the app we should load it.

artursouza:
[...]

PR #4605, opened 2022-05-11
Enable flag to enable/disable kubernetes builtin secret store

DeepanshuA:
# Description

Currently, builtin kubernetes secret store is being called by default every time dapr runtime starts. 
Adding enable-builtin-k8s-secret-store flag so that a user can disable calling it.
By default, behavior will remain same and if not specified, its value will be true by default.

## Issue reference

Please reference the issue this PR will close: #4227 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Unit tests passing


pkg/runtime/cli.go
```diff
@@ -72,6 +72,7 @@ func FromFlags() (*DaprRuntime, error) {
 	daprHTTPStreamRequestBody := flag.Bool("dapr-http-stream-request-body", false, "Enables request body streaming on http server")
 	daprGracefulShutdownSeconds := flag.Int("dapr-graceful-shutdown-seconds", -1, "Graceful shutdown time in seconds.")
 	enableAPILogging := flag.Bool("enable-api-logging", false, "Enable API logging for API calls")
+	enableBuiltinK8sSecretStore := flag.Bool("enable-builtin-k8s-secret-store", true, "Enable Builtin Kubernetes Secret Store")
```
artursouza, line 75: Default being `true` is great. It makes this change backwards compatible.

pkg/runtime/runtime.go
```diff
@@ -386,8 +386,9 @@ func (a *DaprRuntime) initRuntime(opts *runtimeOpts) error {
 			log.Warnf("failed to watch component updates: %s", err)
 		}
 	}
-
-	a.appendBuiltinSecretStore()
+	if a.runtimeConfig.EnableBuiltinSecretStore {
+		a.appendBuiltinSecretStore()
```
artursouza, line 390: Please, create a new E2E test to validate that the k8s secret store can be enabled/disabled correctly.

artursouza (changes requested):

pkg/runtime/runtime.go
yaron2: Generally I wouldn't require an E2E test for something so small in scope, but in this case there's also the Injector component involved, and we need to ensure backwards compatibility all the way from injection, so E2E makes sense here.

yaron2: Well ACTUALLY... backwards compatibility is tested by the virtue of having the existing e2e pass.

pkg/runtime/cli.go
ItalyPaleAle: If the default is `true`, shouldn't the flag be called `no-builtin-k8s-secret-store` (or `disable-builtin-k8s-secret-store`) with a default of `false`? Most UNIX CLIs do not use `true` for missing flags.
(Same for the annotation)

DeepanshuA: @ItalyPaleAle @artursouza If disable-builtin-k8s-secret-store, with a default value of `false` makes better sense here to users, then I can change the flag/annotation accordingly. Is that the final decision here? Just want to confirm before I make the actual change.

ItalyPaleAle: I would recommend the `--no-` option to turn the feature off. If the default is to have the feature on, that seems the most common pattern among CLIs.

artursouza: OK, let's change to `disable-builtin-k8s-secret-store`

pkg/runtime/runtime.go
artursouza: That is true. But we should have an E2E test showing that this feature works. The E2E test should have a scenario targeting reading from K8s secrets directly. One for on and another one for off.

pkg/runtime/cli.go
DeepanshuA: Please review the changes now.

pkg/runtime/runtime.go
DeepanshuA: We already have `secretapp` taking care of default case where-in built-in secret store named `kubernetes` is used. I have added further case in the same test to take care of disabled k8s secret store.

ItalyPaleAle (approved):
LGTM

artursouza (approved):

Issue #4512, opened 2022-04-12
Provide a Default Policy Experience for Resiliency


halspang:
## In what area(s)?

/area runtime

> /area operator

> /area placement

> /area docs

> /area test-and-release

## Describe the proposal
With the initial launch of Resiliency in 1.7, we should provide an easier way for people to adopt it. Specifically, there will be a time when we want people to move to Resiliency and _remove_ the internal, built-in Dapr retries. This proposal is to get users to a place where they can more confidently adopt Resiliency without worrying that they'll lose all the retries that Dapr provides.

### Existing Retries
Dapr performs it's own retries in a few places already (this list may not be complete):

- Service invocation (specifically calling remote hosts)
- Actor invocation (specifically calling remote hosts)
- Actor state management (retries around storing reminders)
- Internal initialization (listing subscriptions, listing components)

Aside from the internal calls, the above calls can be covered by a Resiliency policy. As of 1.7, they exist in tandem with the Resiliency policy, which can cause more retries than intended.

The goal here will be to allow users to easily cover the existing retries and give them an avenue for using only Resiliency for all of their retry/backoff needs.

### Providing a Default Configuration
The easiest way to do this will be to provide a default configuration for Dapr's own retries. This configuration will use special names that we can recognize where retries exist and utilize them in our code base. This is necessary because Resiliency targets specific entities. The built-in retries are global. So, we'll need a way to make the policies recognize the global aspect of it.

This configuration can either be provided by default or setup by the customer. For an initial revision, it would be good to bundle it with Dapr so users are guaranteed their retries. Resiliency allows the setting of multiple policies (in k8s) and thus anything the customer wishes to set can be added without having to touch this configuration.

The policy itself will be provided in our documentation, but in Dapr, it will exist as part of the base level structure and be loaded into memory as a constant. The reason for this is that there has been a concern with users deleting/removing built-in retries entirely without intending to. By having it as an in-memory constant as part of the binary, the customer won't be able to delete it without explicitly overriding it. That can be accomplished by taking the documented configuration and changing the values which will override it, but not delete it. If a user goes through this much effort to override the built-ins we can assume they are a power user and responsible for their own actions.

### Utilizing the Configuration/Opting-Out of Built-ins
 Built-in retries don't always take place in the same area of code that the Resiliency policy retries do. So, in order to allow customers to utilize their own policy/opt-out of built-in retries we need to reference the policies in multiple places.

[...]

PR #4609, opened 2022-05-12
Add resiliency policy for built-in retries

halspang:
# Description

This commit brings Dapr's existing built-in retries into the
resiliency framework. The built-in retries still exist as resiliency
is still a preview feature. However, if resiliency is enabled, an
in-memory policy is loaded that will be used where the built-in
retries currently exist.

The built-in retries can be overwritten if the customer specifies
the special policy name in their policy. They can also be ignored
if the target application has a policy defined for it.

https://github.com/dapr/dapr/issues/4512

Note: This only covers service invocation to show it as an example. Actors will follow or be added to this PR if people prefer.

## Issue reference

Please reference the issue this PR will close: #4512 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Unit tests passing

### DIFF
--- a/pkg/messaging/direct_messaging.go
+++ b/pkg/messaging/direct_messaging.go
@@ -19,6 +19,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/cenkalti/backoff"
 	"github.com/pkg/errors"
 	"github.com/valyala/fasthttp"
 	"google.golang.org/grpc"
@@ -33,6 +34,7 @@ import (
 	diag "github.com/dapr/dapr/pkg/diagnostics"
 	diag_utils "github.com/dapr/dapr/pkg/diagnostics/utils"
 	"github.com/dapr/dapr/pkg/modes"
+	"github.com/dapr/dapr/pkg/resiliency"
 	"github.com/dapr/dapr/pkg/retry"
 	"github.com/dapr/dapr/utils"
 
@@ -65,6 +67,8 @@ type directMessaging struct {
 	maxRequestBodySize  int
 	proxy               Proxy
 	readBufferSize      int
+	resiliency          resiliency.Provider
+	isResiliencyEnabled bool
 }
 
 type remoteApp struct {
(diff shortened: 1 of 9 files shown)


pkg/messaging/direct_messaging.go
```diff
@@ -145,6 +157,46 @@ func (d *directMessaging) invokeWithRetry(
...
+			err := policy(func(ctx context.Context) (rErr error) {
+				retriesExhaustedPath = false
+				resp, rErr = fn(ctx, app.id, app.namespace, app.address, req)
+				if rErr == nil {
+					return nil
+				}
+
+				code := status.Code(rErr)
+				if code == codes.Unavailable || code == codes.Unauthenticated {
+					_, connerr := d.connectionCreatorFn(context.TODO(), app.address, app.id, app.namespace, false, true, false)
```
artursouza, line 176: Can't you use the `ctx` context here?

pkg/messaging/direct_messaging.go
```diff
@@ -145,6 +157,46 @@ func (d *directMessaging) invokeWithRetry(
...
+					if connerr != nil {
+						nullifyResponsePath = true
+						return backoff.Permanent(connerr)
+					}
+					retriesExhaustedPath = true
+					return rErr
+				}
+				return backoff.Permanent(rErr)
```
artursouza, line 184: Is it right to `return backoff.Permanent()` even when `retriesExhaustedPath` is `false`?

artursouza (commented):
Overall, LGTM. Just a minor comment.

pkg/messaging/direct_messaging.go
halspang: Yes, the reason we do that is because we actually don't retry everything. Here's the original retry method:

```go
for i := 0; i < numRetries; i++ {
	resp, err := fn(ctx, app.id, app.namespace, app.address, req)
	if err == nil {
		return resp, nil
	}
	log.Debugf("retry count: %d, grpc call failed, ns: %s, addr: %s, appid: %s, err: %s",
		i+1, app.namespace, app.address, app.id, err.Error())
	time.Sleep(backoffInterval)

	code := status.Code(err)
		_, connerr := d.connectionCreatorFn(context.TODO(), app.address, app.id, app.namespace, false, true, false)
		if connerr != nil {
			return nil, connerr
		}
		continue
	}
	return resp, err
}
```

You can see we only call `continue` when we've gotten `Unavailable` or `Unauthenticated` and were able to reconnect. If it wasn't one of those two errors, we exit from the loop immediately. That's what the permanent error does.

pkg/messaging/direct_messaging.go
halspang: Good call.

artursouza (approved):

PR #4613, opened 2022-05-14
Updated components-contrib pinned version

ItalyPaleAle:
Updated components-contrib pinned version to bring in the latest changes (including new Azure SDKs)


berndverst:
Currently breaks E2E tests and more investigation is needed.

This contrib release introduces changes to the service bus components (now using the Track 2 SDK).

ItalyPaleAle:
I have re-run the tests manually (targeting AKS and Service Bus) and I cannot repro the failures from the Action. Will re-trigger them again.

PS: Note that Event Hubs is not used in the E2e tests, and it was not really impacted (the tests in components-contrib for event hubs are broken and were failing 100% of times before and after the PR).

ItalyPaleAle:
I have confirmed the root cause of the test failures, and it's a combination of:

- Issues with the tests themselves, which set back-off times too low when a message had to be re-delivered (among other issues) - PR for improving the tests is coming soon
- An issue where Dapr's runtime keeps sending messages to the apps even after shutdown has been initiated, during the 5 seconds "grace period". In this case, messages fail to be processed because the app has already been terminated, so the sidecar gets "connection refused" errors while connecting to the app to try and process the messages. I am working on a PR for this as well.

This is to say, it should be safe to merge this PR. Additional PRs fixing the issues above are coming.

ItalyPaleAle:
Superseded by #4619

PR #4618, opened 2022-05-15
E2E tests: cache app images for significant speed-ups

ItalyPaleAle:
This PR aims to improve the speed of E2E tests (both on Azure and KIND) and perf tests by caching the app images.

Currently, one of the longest steps in E2E tests is building the images for the E2E test apps. In fact, it takes A LOT of time, especially on Windows. Here's a screenshot of a [recent run](https://github.com/dapr/dapr/actions/runs/2325025877):

Screen Shot 2022-05-14 at 16 23 22

In particular, expanding the "Build for Windows" step, you can see that "Build e2e test apps" took a whole 19 minutes. (On Linux it took 6m7s, which is still a lot of time although not as terrible).

Because tests cannot start until all images are built, that adds to a significant delay.

Because the E2E test images don't change often, in fact they see a commit every few months only, they can be cached! When images are cached, there are significant speed-ups.

Here's an example of a run where images were cached (performed in a fork to test the pipeline):

image

In this example, it took 16 minutes less! 🥳

How this PR works:

1. I created an Azure Container Registry in our "Dapr Tests" subscription (the one in the dapr.io tenant where we run our E2E tests).
2. If the environmental variable `DAPR_CACHE_REGISTRY` is set with the address of the ACR, the `make build-e2e-app-*` step now invokes a new script to build the Docker images (same for perf tests)
  - The script first computes the checksum of the test app. It does that by calculating the SHA-256 hash of all files (and their content) within the test app, then it computes the SHA-256 hash of these hashes and the file names.
  - The checksum is used as tag for the Docker images.
  - The script first tries to pull the image from the cache registry, using the tag generated above.
  - If the image doesn't exist, it builds it from scratch and pushes it to the cache.
  - Whether the image was cached or freshly built, it then tags the image with the right tag so our tests can proceed (ie. the test runner will then push it to whatever container registry we use)

Because the cache is based on the hash of the files in the test apps, it works fine with pull requests that may change the test apps too.

PS: As part of this work, the purpose of `dapr-test-apps` seemed redundant and I deleted the pipeline. I can re-add it if needed.


tests/build_cache_images.sh
```diff
@@ -0,0 +1,67 @@
...
+HASH=$(hashdir "$APP_DIR")
+echo "HASH: ${HASH:0:10} (${HASH})"
+HASH="${HASH:0:10}"
+CACHE_TAG="${CACHE_REGISTRY}/${APP_DIR}:${DOCKERFILE}-${HASH}"
+
+# Check if the image already exists
+set +e
+docker pull "${CACHE_TAG}"
+EXITCODE=$?
+set -e
```
daixiang0, line 54: Only check the failure of pull image from cache?

tests/build_cache_images.sh
```diff
@@ -0,0 +1,67 @@
...
+    done
+    cd "$starting_dir"
+    echo -e "$result" | sha256sum | awk '{ print $1 }'
+}
+
+# Compute the hash of the app's files
+HASH=$(hashdir "$APP_DIR")
+echo "HASH: ${HASH:0:10} (${HASH})"
+HASH="${HASH:0:10}"
+CACHE_TAG="${CACHE_REGISTRY}/${APP_DIR}:${DOCKERFILE}-${HASH}"
```
daixiang0, line 48: CACHE_NAME is better, it is not only about tag.

tests/build_cache_images.sh
```diff
@@ -0,0 +1,67 @@
...
+    echo "Found cached image"
+else
+    echo "Cached image not found; building it"
+    docker build -f "${APP_DIR}/${DOCKERFILE}" "${APP_DIR}/." -t "${CACHE_TAG}"
+    docker push "${CACHE_TAG}"
+fi
+
+# Tag the image with the desired tag
+echo "Tagging image as ${TARGET_NAME}"
+docker tag "${CACHE_TAG}" "${TARGET_NAME}"
```
daixiang0, line 67: Do we need to push TARGET_NAME since tests will pull this?

tests/build_cache_images.sh
ItalyPaleAle: We cannot determine the cause of the failure of this command. So we assume that whatever the reason is, the image doesn’t exist in the cache.
If the issue is related to authentication, the push command later would fail too.

tests/build_cache_images.sh
ItalyPaleAle: Yes, because the cache container registry is different from the one used by the tests. The cache CR is fixed and pre-deployed. E2E tests use a dynamically-created CR, while KIND tests use the CR that is part of KIND

tests/build_cache_images.sh
ItalyPaleAle: Done

tests/build_cache_images.sh
daixiang0: `set -e` helps here since we want all commands to succeed.

ItalyPaleAle: No, this specific command may fail   If the image is not in the cache, it should fail and exit with non-0 status code. We capture the status code to check it the image exists.

If it doesn't exist, we build the image.

daixiang0: `func pull_from_cache || func build_and_push` then we can use `set -e`.

ItalyPaleAle: That sounds reasonable. I made the change as you requested

.github/workflows/kind-e2e.yaml
```diff
@@ -10,9 +10,16 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
+
+# Required secrets:
+# - AZURE_CREDENTIALS: JSON object containing the Azure service principal credentials. This is necessary if you want to cache the E2E test images. Docs: https://github.com/Azure/login#configure-a-service-principal-with-a-secret
```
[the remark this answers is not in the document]
ItalyPaleAle: I have updated the ACR to allow un-authenticated pulls. The script will still try to push even if there are no credentials, but will ignore failures.

This way, KIND pipelines will never be able to push images to the cache, but can consume them if they're available. That seems an acceptable trade-off given that apps are updated very infrequently.

artursouza (approved):

PR #4622, opened 2022-05-17
Hotfix for actor subsystem initialization in 1.7 (and possibly others)

johnewart:
# Description

Changes the actor initialization subsystem to fail when there is no state store only when there are hosted actors. When there are hosted actors and no state store, don't initialize the actor subsystem. Moved actor state storage into a separate (identically configured) component for Cosmos and Redis and scoped it to only tests needing actor state storage. 

Added a test to the actor_reminder to verify that it gets a 500 when trying to call a service that has hosted actors but no state store available, and also ensure that actor clients can work without a state store. 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Unit tests passing
* [x] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#2441

/cc @artursouza

### DIFF
--- a/docs/release_notes/v1.7.3.md
+++ b/docs/release_notes/v1.7.3.md
@@ -4,7 +4,7 @@
 
 #### Problem
 
-The Actor API fails to initialize without an actor state store component being provided. This is the correct behavior for services that register actors, as they require state storage, but clients that invoke actors should operate with or without an actor state store. 
+All Actor APIs return errors without an actor state store component being provided. This is the correct behavior for services that register actors, as they require state storage, but clients that invoke actors should operate with or without an actor state store. 
 
 #### Root cause
 
--- a/docs/release_notes/v1.7.3.md
+++ b/docs/release_notes/v1.7.3.md
@@ -1,0 +1,15 @@
+# Dapr 1.7.3
+
+### Fixes incorrect actor state-store configuration behavior
+
+#### Problem
+
+Omitting a state store configuration will cause the actor subsystem to fail to initialize. This is the correct behavior for services providing actors but this not for services acting as consumers / clients of actors that other services provide. A service acting purely as a consumer does not need state storage and should initiazlize correctly in order to invoke actors elsewhere. 
+
+#### Root cause
+
+Absence of a state store causes the actor subsystem to throw an error, regardless of whether or not the service is providing hosted actors.
+
+#### Solution
+
+Thix fix changes the actor runtime initialization logic such that, when there is no state store available, the actor subsystem will successfully initialize for consumers but will fail for services with hosted actors. 
--- a/docs/release_notes/v1.7.3.md
+++ b/docs/release_notes/v1.7.3.md
@@ -4,12 +4,13 @@
 
 #### Problem
 
-Omitting a state store configuration will cause the actor subsystem to fail to initialize. This is the correct behavior for services providing actors but this not for services acting as consumers / clients of actors that other services provide. A service acting purely as a consumer does not need state storage and should initiazlize correctly in order to invoke actors elsewhere. 
+The Actor API fails to initialize without an actor state store component being provided. This is the correct behavior for services that register actors, as they require state storage, but clients that invoke actors should operate with or without an actor state store. 
 
 #### Root cause
 
-Absence of a state store causes the actor subsystem to throw an error, regardless of whether or not the service is providing hosted actors.
+The code that initializes the Actor API raises an error when there is no actor state storage component available, regardless of whether or not the service
+registers actors. 
 
 #### Solution
 
-Thix fix changes the actor runtime initialization logic such that, when there is no state store available, the actor subsystem will successfully initialize for consumers but will fail for services with hosted actors. 
+Thix fix changes the actor runtime initialization logic such that, when there is no actor state store available, the API will initialize correctly as long as the service does not register actors. As a result, the Actor API will be available to services acting only as clients (i.e invoking actors) with or without an actor state store component; the Actor API will continue to be unavailable in those services that register actors without providing an actor state store component.
--- a/pkg/actors/actors.go
+++ b/pkg/actors/actors.go
@@ -192,7 +192,8 @@ func (a *actorsRuntime) Init() error {
 
 	if len(a.config.HostedActorTypes) > 0 {
 		if a.store == nil {
-			log.Warn("actors: state store must be present to initialize the actor runtime")
+			// If we have hosted actors and no store, we can't initialize the actor runtime
+			return fmt.Errorf("hosted actors: state store must be present to initialize the actor runtime")
 		} else {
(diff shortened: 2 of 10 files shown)


johnewart:
Draft because I need @artursouza to flag as ok to test to ensure cosmosdb works as expected

artursouza:
Please, fix linter.

johnewart:
Added release notes for 1.7.3 and opened an issue for the docs to update references to the latest point release

docs/release_notes/v1.7.3.md
```diff
@@ -0,0 +1,15 @@
+# Dapr 1.7.3
+
+### Fixes incorrect actor state-store configuration behavior
+
+#### Problem
+
+Omitting a state store configuration will cause the actor subsystem to fail to initialize. This is the correct behavior for services providing actors but this not for services acting as consumers / clients of actors that other services provide. A service acting purely as a consumer does not need state storage and should initiazlize correctly in order to invoke actors elsewhere.
```
berndverst, line 7: "state storage" -> "actor state storage" (both occurrences)

"but this not for" -> "but not for"

"A service acting purely as a consumer" -> client instead of consumer

docs/release_notes/v1.7.3.md
```diff
@@ -0,0 +1,15 @@
...
+
+### Fixes incorrect actor state-store configuration behavior
+
+#### Problem
+
+Omitting a state store configuration will cause the actor subsystem to fail to initialize. This is the correct behavior for services providing actors but this not for services acting as consumers / clients of actors that other services provide. A service acting purely as a consumer does not need state storage and should initiazlize correctly in order to invoke actors elsewhere. 
+
+#### Root cause
+
+Absence of a state store causes the actor subsystem to throw an error, regardless of whether or not the service is providing hosted actors.
```
berndverst, line 11: Root Cause: All Actor APIs return an error if no actor state store is defined, regardless of where actors are registered.

FYI SDKs use the term "register" for telling a sidecar to host particular actors. So this may be less confusing to use in language.

docs/release_notes/v1.7.3.md
```diff
@@ -0,0 +1,15 @@
...
+
+#### Solution
+
+Thix fix changes the actor runtime initialization logic such that, when there is no state store available, the actor subsystem will successfully initialize for consumers but will fail for services with hosted actors.
```
berndverst, line 15: state store -> actor state store
consumer -> client

berndverst (changes requested):

docs/release_notes/v1.7.3.md
```diff
@@ -0,0 +1,16 @@
+# Dapr 1.7.3
+
+### Fixes incorrect actor state-store configuration behavior
+
+#### Problem
+
+The Actor API fails to initialize without an actor state store component being provided. This is the correct behavior for services that register actors, as they require state storage, but clients that invoke actors should operate with or without an actor state store.
```
berndverst, line 7:
```suggestion
All Actor APIs return errors without an actor state store component being provided. This is the correct behavior for services that register actors, as they require state storage, but clients that invoke actors should operate with or without an actor state store. 
```

artursouza (approved):

pkg/runtime/runtime.go
```diff
@@ -1768,12 +1768,14 @@ func (a *DaprRuntime) initActors() error {
 	a.actorStateStoreLock.Lock()
 	defer a.actorStateStoreLock.Unlock()
 	if a.actorStateStoreName == "" {
-		return errors.New("no actor state store defined")
+		log.Warn("actors: state store is not configured - this is okay for clients but services with hosted actors will fail to initialize!")
```
halspang, line 1771: Will this introduce a warning to all instances that don't use actors? If it's OK in some scenarios we shouldn't say anything here.

pkg/actors/actors.go
```diff
@@ -192,12 +192,13 @@ func (a *actorsRuntime) Init() error {
...
-			if !state.FeatureETag.IsPresent(features) || !state.FeatureTransactional.IsPresent(features) {
-				return errors.New(incompatibleStateStore)
-			}
+			// If we have hosted actors and no store, we can't initialize the actor runtime
+			return fmt.Errorf("hosted actors: state store must be present to initialize the actor runtime")
+		}
+
+		features := a.store.Features()
+		if !state.FeatureETag.IsPresent(features) || !state.FeatureTransactional.IsPresent(features) {
+			return errors.New(incompatibleStateStore)
```
halspang, line 201: These seem pretty easy to unit test, are there tests for this?

halspang (changes requested):
Quick question about a test but if I'm understanding this correctly we're adding a warning to every startup without actors which I don't think is the right behavior.

pkg/actors/actors.go
johnewart: The E2E tests cover this for sure -- there is an exclusion of the `actortestclient` app-id in the actor state storage component (I just added a commit with comments in each actor state store YAML file to make a note of that) to ensure that the client-only app used by `test_actor_features` is able to invoke actors in the the other app being deployed for that test without a state store. I don't know if we have the plumbing to fake having hosted actors in the unit test but I can look.

pkg/runtime/runtime.go
johnewart: I could change it to info level -- we warn if the app has hosted actors but no actor state store. I think it's maybe worth letting people know informationally, but we can also probably get rid of it if it's overly verbose. I personally prefer more logs to sift through than less context but will go with whatever our usual policy / behavior is.

pkg/actors/actors.go
halspang: Yeah, I know the e2e tests cover them, but the more places we can enforce the appropriate behavior the better. And the `NewActors` method is quite literally just this if statement, so it shouldn't be too bad to cover.

pkg/runtime/runtime.go
halspang: I'm also OK with slightly more verbose logging, especially at startup. I just don't think we should warn someone if their behavior is expected and rational. I'm fine with moving it to an info, it's a log message that happens once during init and it's a piece of information they may want. Warning, to me, means it's something they should go fix but not critically.

pkg/actors/actors.go
johnewart: Added unit tests in 9bb68e0ef5feb

halspang (approved):

pkg/runtime/runtime.go
johnewart: Swapped to Info in 32dbce326945caa3f2b7a4ee6c4d9a6704cc51a0

artursouza (approved):

halspang (approved):

berndverst (approved):

PR #4624, opened 2022-05-17
Shutting down input components before the grace period

ItalyPaleAle:
The current shutdown sequence is:

1. Stop receiving new messages via service invocation
2. Wait 5 seconds
3. Shutdown all components

However, some components (pubsub and input bindings) are bringing in new work as well and their shutdown should not be delayed.

While debugging E2E tests, this behavior was causing the runtime to continue receiving messages and attempt re-delivering them during the shutdown grace period, even though the app had already been shut down. This was causing messages to be retried constantly and they were exhausting their max retry attempts.

### DIFF
--- a/pkg/runtime/runtime.go
+++ b/pkg/runtime/runtime.go
@@ -569,7 +569,7 @@ func (a *DaprRuntime) sendToDeadLetterIfConfigured(name string, msg *pubsub.NewM
 	return true, err
 }
 
-func (a *DaprRuntime) beginPubSub(name string, ps pubsub.PubSub) error {
+func (a *DaprRuntime) beginPubSub(subscribeCtx context.Context, name string, ps pubsub.PubSub) error {
 	var publishFunc func(ctx context.Context, msg *pubsubSubscribedMessage) error
 	switch a.runtimeConfig.ApplicationProtocol {
 	case HTTPProtocol:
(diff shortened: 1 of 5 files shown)


ItalyPaleAle:
Given how this is not going to make it into a hotfix, moving it back to draft stage so we can review it more thoroughly.

ItalyPaleAle:
@berndverst As I looked more into this, we should do something coordinated between the runtime and components-contrib.

Right now, the shutdown calls `Close()` on them. However, that can close the output channels too (e.g. the ability to perform `Invoke` on a pubsub component and the output part of a binding). For example, i have noticed that to be the case at least in the ASB queues binding and in Postgres.

Proposal is to add an extra method to close the inbound part:

- For pubsub, that could be `Unsubscribe` (to mirror `Subscribe`). `Close` can remain to close the outbound channel
- For bindings, we should probably split the `Close` in `CloseInput` and `CloseOutput`

@skyao had some ideas here too.

I wouldn't say that these changes are blockers to merge this PR. The current situation is arguably worse than closing output bindings earlier, but...

pkg/runtime/runtime.go
```diff
@@ -2088,50 +2088,55 @@ func (a *DaprRuntime) stopActor() {
...
 		if err := pubSub.Close(); err != nil {
 			err = fmt.Errorf("error closing pub sub %s: %w", name, err)
 			merr = multierror.Append(merr, err)
 			log.Warn(err)
 		}
 	}
-	for name, stateStore := range a.stateStores {
-		if closer, ok := stateStore.(io.Closer); ok {
+	// Close bindings if they implement `io.Closer`
+	// TODO: Separate the input part of bindings and close output here, then close the input via cancelation of a.ctx
```
halspang, line 2135: Aren't the output bindings closed above this? I'm a little confused by this todo, can you clarify it?

halspang (commented):
Seems fine overall, but I'm confused over the TODO you have listed. Can you clarify it?

Also linter :)

artursouza (approved):

Issue #4447, opened 2022-03-29
E2E tests: move the "test controller" to a container within the Kubernetes cluster itself


ItalyPaleAle:
## In what area(s)?

/area test-and-release

## Describe the proposal

Currently, E2E tests are designed to be executed by a "test controller" app that runs "locally" (in the case of the GitHub Action, that means within the Actions runner itself). This "test controller" communicates with the K8s cluster to create new apps, then connects to the apps to verify their behavior.

When the tests run on a K8s cluster that's not in the runner itself, the communication goes over some network.

During my recent work on improving E2E tests, I noticed that the network between the Actions runner and AKS is not always reliable. Tests are not-infrequently failing due to network issues (connections dropped, timeouts, etc).

In some cases these are not transient: even when a retry logic is in place, the connection kept failing for at least 10 minutes. It's unclear exactly what causes this, but tests have been showing this behavior pretty much consistently when targeting AKS in Canada Central or East Asia.

Even when the failures are transient and a retry would fix them, it would still require a lot of work to go through our test infrastructure and add some resiliency support (in some cases operations aren't idempotent either, so automatically re-trying them in case of network failures may not be ideal).

A possible solution to this problem could be to containerize the "test controller" and run it inside the K8s cluster itself. So all communication would be "local" inside the K8s cluster.

Inside Actions, the Runner would create a container of the test app and deploy that like any other workload on K8s. The test controller would do what it needs to do and then return the response to the Actions Runner. This way I hope we would eliminate flakiness due to issues with the network.

(Note that this does **not** require the creation of self-hosted Actions runner - we'd still be using the public Actions Runners, but their job would more limited to just kickstarting the tests and waiting for the results).

artursouza:
I am supportive of running the test code from within the AKS cluster itself. My only concern would be if that can be achieved without moving the GitHub Actions Runner entirely too.

ItalyPaleAle:
Yes, not moving the Actions Runner (that is, not needing to create self-hosted ones) is a requirement.

ItalyPaleAle:
My hope is that given that this would make less network calls and less frequently, it should be more stable.

If the runner couldn't even make a single call to deploy a pod, and then another one to retrieve the results, that would be really bad

ItalyPaleAle:
[...]

PR #4626, opened 2022-05-17
Pubsub E2E tests fixes

ItalyPaleAle:
These ([and the work done in components-contrib](https://github.com/dapr/components-contrib/pull/1729)) should fix the E2E tests that were failing after updating the Azure Service Bus SDK.
In general, this commit improves the reliability of the E2E PubSub tests and the logging.


yaron2 (approved):

ItalyPaleAle:
Added: looks like tests are green even with a smaller service bus (4 units instead of 16!). This may trigger some retries, which is actually a good thing so we test resiliency to that.

artursouza (approved):

artursouza (approved):

ItalyPaleAle:
@artursouza  as discussed, this PR is as good as it will ever be at this point.

KIND tests are passing, although we are seeing some weird issues that trigger built-in retries. In particular, the publisher service sometimes has issues invoking subscriber services: we are seeing calls that get completed in 22 seconds (and according to the logs, the request does not hit the subscriber service until that time, so there's an issue with connecting in the first place), and GRPC calls that simply fail to connect.
See for example the logs for [this action run](https://github.com/dapr/dapr/actions/runs/2353177691) (which overall is green but contains failures that get retried).
I can't explain this in any other way besides thinking there's _something_ wrong either in the KIND network stack, or in the name resolution (the latter of which I can investigate).

AKS tests sometimes fail because requests to `/publish` fail with timeout errors and that causes the entire tests to be red. Those requests cannot be retried because they're not idempotent. These requests are going over the network and seems that they never hit the apps running on AKS (see request with trace ID `c-b3edf4f5-2f7b-4fce-93a6-edb5dcfc8e06` in [this run](https://github.com/dapr/dapr/actions/runs/2353188189)). This is likely indicating an issue with either the network, or the AKS ingress controller.

Weirdly, all tests seem to be passing in my private fork, which targets the same Azure subscription as this one.

Actions:

- I will open a support ticket with Azure and see if they can help us troubleshooting this by looking at the network traces.
- I will also prioritize #4447 that hopefully improves our resiliency in this space.

PR #4644, opened 2022-05-20
Fix for actor subsystem initialization (#4622)

johnewart:
* Adds a new E2E test to ensure that a service, which registers an actor but does not have an actor state store component will correctly generate a 500 when trying to invoke Actor APIs
* Adding unit tests for the actor initialization with no state store for the case where we have 0 hosted actors (client) and one hosted actor registered

Uses existing E2E tests with some changes to components to validate the correct behavior:
  * Separates the actor state store into new components for cosmos and redis, removing the actor state store flag from the original ones
  * Scopes the new actor state store components to all E2E test apps requiring actors (except one test that acts as a client only) so that we can re-use the existing E2E tests to prove correctness
  * Adds comments to the actor store component YAML files regarding which app id is excluded to ensure client works without an actor state store component

Signed-off-by: John Ewart <johnewart@microsoft.com>
Co-authored-by: Bernd Verst <4535280+berndverst@users.noreply.github.com>

# Description

In 1.7.x (and earlier), not having actor state storage would incorrectly cause the Actor API to be unavailable regardless of
whether or not the service registered actors. This meant that services that interacted with actors (i.e acting as a client) would not work without an actor state store component - which is not required to be an actor client. This is a forward port of the fix to 1.7.x, along with tests to ensure we don't accidentally regress.

Changes the actor API initialization behavior so that:
* Services that register actors, but no actor state storage will not initialize the Actor API 
* Services that do not register actors will initialize the Actor API regardless of whether or not state storage is present

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Unit tests passing
* [ ] ~~Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_~~
* [ ] ~~Specification has been updated / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_~~
* [ ] ~~Provided sample for the feature / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_~~

### DIFF
--- a/docker/Dockerfile-debug
+++ b/docker/Dockerfile-debug
@@ -1,6 +1,6 @@
 # current directory must be ./dist
 
-FROM golang:latest AS golang
+FROM golang:1.18 AS golang
 ENV GOPROXY=https://goproxy.io,direct
 RUN CGO_ENABLED=0 go install -ldflags '-s -w -extldflags -static' github.com/go-delve/delve/cmd/dlv@latest
 
--- a/docs/release_notes/v1.5.3.md
+++ b/docs/release_notes/v1.5.3.md
@@ -1,0 +1,16 @@
+# Dapr 1.5.3
+
+### Fixes incorrect actor state-store configuration behavior
+
+#### Problem
+
+All Actor APIs return errors without an actor state store component being provided. This is the correct behavior for services that register actors, as they require state storage, but clients that invoke actors should operate with or without an actor state store. 
+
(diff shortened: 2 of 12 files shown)


yaron2 (approved):

yaron2:
Great stuff.

halspang (approved):

PR #4647, opened 2022-05-21
[WIP] Mitigations for potential SNAT port exhaustion

ItalyPaleAle:
There's a possibility (to be confirmed) that our flakiness with E2E tests may be caused by SNAT port exhaustion in the Actions runners. Runners are Azure VMs behind a load balancer, so [SNAT port exhaustion may be a possibility](https://docs.microsoft.com/en-us/azure/load-balancer/troubleshoot-outbound-connection) if we're making a lot of outbound requests.

This commit includes improvements with recommendations that could mitigate that:

1. Ensure that HTTP clients are re-used and not recreated every time
2. Ensure that responses' `Body` streams are always closed properly, so that the client is not idle

Without the ability to control the Actions infrastructure, these application-level changes are our best option, assuming our issues are indeed caused by SNAT.


ItalyPaleAle:
Closing as the experiment was successful. Will create a PR with a squash of all commits

Issue #4662, opened 2022-05-25
VolumeMounts support for Side Car Daprd


jollyaakash:
## In what area(s)?
 /area runtime

## Background
Today Dapr doesn't support  providing a capability via which the Volumes that are part of the Pod can be mounted on the side car - daprd container as well. The Volume might be needed by the Side car to communicate with host or other elements living on a K8s cluster. One real case example of this is how SPIRES implementation of SPIFFE [concepts here](https://spiffe.io/docs/latest/spire-about/spire-concepts/) work via a _workload api socket_ created on Host Path. This socket present on Host Path mounted as a volume allows workloads to call into workload api to get a SPIFFE identity and likewise provide agent/server to attest the workloads. 

The current [MQTT side car component](https://github.com/dapr/components-contrib/blob/master/pubsub/mqtt/mqtt.go) can be extended to talk to workload api and use the SPIFFE identity while talking to any MQTT broker. This can only be possible when the Volumes of the Pod can be mounted on the Daprd side car as well. The main purpose of the proposal is to allow such capabilities to happen, such that other Side Car components can use it as well and help drive development of others.

## Describe the proposal
Add two new annotations to the set of annotations present right now in Dapr Kubernetes [annotations list](https://docs.dapr.io/reference/arguments-annotations-overview/) 

```
"dapr.io/side-volume-mounts"
"dapr.io/side-volume-mounts-read-write"
```

The first annotation **"dapr.io/side-volume-mounts"** will be used to mention a list of all the Volumes present in the Pod that should be mounted in the Side car Daprd as well separated by a comma.  This will support mounting multiple volumes using one annotation only. The value should be of form _VolumeName=MountPath, VolumeName2=MountPath2_. For example - 

```
....
  template:
    metadata:
      labels:
        app: publisher
      annotations:
        dapr.io/enabled: "true"
        dapr.io/app-id: "checkout"
        dapr.io/side-volume-mounts: "spires-agent-socket=/home/nonroot/sockets/, sample-volume=/tmp/sample"
        dapr.io/side-volume-mounts-read-write: "true"
....
      volumes:
      - name: spires-agent-socket
        hostPath:
          path: /run/spires/sockets/
          type: DirectoryOrCreate
      - name: sample-volume
        emptyDir: {}
....
```

[...]

PR #4663, opened 2022-05-25
Add volume mount support for sidecar

shubham1172:
# Description

This PR adds a way to mound Pod volumes to the container using two new Dapr annotations:
1. **dapr.io/volume-mounts**: for read-only volume mounts
1. **dapr.io/volume-mounts-rw**: for read-write volume mounts

The format of both the annotations are comma separated pairs of "pod-volume-name:/path/to/mount/in/container".
If a volume mount is annotated but not present in the pod, we ignore it and log a warning.

There are unit and e2e tests added. The e2e test follows a scenario similar to [custom-secrets-storage-using-local-file-secret-store](https://github.com/dapr/docs/blob/840486d9f8c6836b8125b755abe231c40c8250fb/daprdocs/content/en/operations/hosting/kubernetes/kubernetes-volume-mounts.md#custom-secrets-storage-using-local-file-secret-store), wherein an init container is used to populate data to the pod volume, which is then used as local file-based secret.

## Issue reference

Please reference the issue this PR will close: #4662

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests
* [X] Unit tests passing
* [x] Extended the documentation / Created issue in the https://github.com/dapr/docs/issues/2454


yaron2:
Nice! I like this feature.

shubham1172:
The e2e test has been verified manually and via the GH pipeline.
image

image

shubham1172:
Apparently this is causing all other e2e tests to fail. Will investigate further.

halspang (commented):
Change looks good overall, looks like you have a linter error though. Address that and we can see how the actions go.

halspang (dismissed):

shubham1172:
Failure seems unrelated: `tests/e2e/pubsub_grpc TestPubSubGRPC/publish_with_subscriber_error_test_redelivery_of_messages_grpc`

shubham1172:
I see why e2e tests for Windows are failing. `Message:Back-off pulling image "docker.io/busybox"` - since there is no Windows version of it. One option is to write an custom init app with OS agnostic implementation.

artursouza (approved):

shubham1172:
> > I see why e2e tests for Windows are failing. `Message:Back-off pulling image "docker.io/busybox"` - since there is no Windows version of it. One option is to write an custom init container binary with OS agnostic implementation.
> 
> Which test is requiring this image? This is new.

I was trying to introduce one, as the initcontainer image for injectorapp. Its job is to create a new file in the volume mount. I have not created an injectorapp-init golang binary that does that (and handles both Linux and Windows).

shubham1172:
Both Linux and Windows error seem to be transient. Linux error fails for injectorapp test and Windows error fails for allowlists test with the same error type: `cannot find dapr sidecar in pod <app-name>-5cd8ccc77-g7pkl`. Shall we retrigger this once again?

artursouza (approved):

PR #4665, opened 2022-05-25
AWS SNS/SQS pubsub should be named pubsub.aws.snssqs

ItalyPaleAle:
# Description

Following our conventions, the name for the AWS SNS/SQS pubsub component should have been `pubsub.aws.snssqs` rather than `pubsub.snssqs` (in some parts of our codebase we did use `pubsub.aws.snssqs`, although docs always included `pubsub.snssqs`)

This PR adds supports for aliases in component names so it's possible to support both the new and old names (for backwards compatibility).

I will open a docs PR to update the recommended name to use.

## Issue reference

Fixes dapr/components-contrib#1753

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests
* [X] Unit tests passing
* [ ] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_ @ItalyPaleAle  TBD

### DIFF
--- a/pkg/components/bindings/registry.go
+++ b/pkg/components/bindings/registry.go
@@ -26,13 +26,13 @@ import (
 type (
 	// InputBinding is an input binding component definition.
 	InputBinding struct {
-		Name          string
+		Names         []string
 		FactoryMethod func() bindings.InputBinding
 	}
 
 	// OutputBinding is an output binding component definition.
 	OutputBinding struct {
-		Name          string
+		Names         []string
 		FactoryMethod func() bindings.OutputBinding
 	}
 
@@ -46,6 +46,10 @@ type (
 		CreateOutputBinding(name, version string) (bindings.OutputBinding, error)
 	}
 
+	stringOrSliceOfStrings interface {
+		string | []string
+	}
+
 	bindingsRegistry struct {
 		inputBindings  map[string]func() bindings.InputBinding
 		outputBindings map[string]func() bindings.OutputBinding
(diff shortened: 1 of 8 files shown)


artursouza:
LGTM.

artursouza (dismissed):

pkg/components/bindings/registry.go
```diff
@@ -46,24 +46,42 @@ type (
...
+
 	bindingsRegistry struct {
 		inputBindings  map[string]func() bindings.InputBinding
 		outputBindings map[string]func() bindings.OutputBinding
 	}
 )
 
 // NewInput creates a InputBinding.
-func NewInput(name string, factoryMethod func() bindings.InputBinding) InputBinding {
+func NewInput[T stringOrSliceOfStrings](name T, factoryMethod func() bindings.InputBinding) InputBinding {
```
yaron2, line 60: Disclaimer: I dislike generics in general, not just in Go.

While I wouldn't put a blank ban for generics in Dapr, this usage in particular is a bit of a smell in my opinion, where the developer reasons about the valid input to the method by looking at the name, in this case `stringOrSliceOfStrings`.

What's your take @artursouza? You hail from a Java background so your input is valuable here.

ItalyPaleAle: Are you suggesting that the naming could be improved (such as `nameOrNames` for the argument, or changing the type `stringOrSliceOfStrings`)... Or that I shouldn't have used generics at all?

Option 2 I was considering was changing the method signature to `New(name string, factoryMethod func(), aliases ...string)`, where `aliases` was optional and could have been left empty. But I disliked that `name` and `aliases` were separated by `factoryMethod` (of a different type).

yaron2: `New(name string, factoryMethod func(), aliases ...string)` looks like a much better alternative to me.

It also removes the boilerplate code of the switch case statement and allows the method to focus on business logic instead.

ItalyPaleAle: I can make that change. The run with generics was fun while it lasted :)

yaron2: :)

ItalyPaleAle: Made the change.

artursouza: This allows backwards compatibility to allow this method to receive both an array and a string. Generics fits well here because Go does not allow method overloading.

PR #4666, opened 2022-05-25
Fix performance tests

johnewart:
# Description

Converted app name magic strings to constants in perf tests and added their identifiers to actor state store configurations so that the performance tests utilizing actors can run. 

Signed-off-by: John Ewart <johnewart@microsoft.com>

/cc @skyao

### DIFF
--- a/tests/config/dapr_cosmosdb_state_actorstore.yaml
+++ b/tests/config/dapr_cosmosdb_state_actorstore.yaml
@@ -50,3 +50,9 @@ scopes:
 - actorphp
 - resiliencyapp
 - resiliencyappgrpc
+- perf-actor-activation-service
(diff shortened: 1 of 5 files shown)


yaron2 (approved):

skyao:
@yaron2 why the statue of this PR is still open, not merged.

yaron2:
> @yaron2 why the statue of this PR is still open, not merged.

It shows as merged.

skyao:
> 

OK, it seems a bug of github. I refreshed the pages for times and it was open all the time. 

And now it is merged.

skyao:
@johnewart  @yaron2  the perf test is not recovered after this PR merged.  We need to follow it: the perf test is NOT work for more than one week now.

PR #4667, opened 2022-05-26
fix concurrent gRPC proxy requests

Syuparn:
# Description

fix gRPC proxy to handle requests concurrently

## Issue reference

Please reference the issue this PR will close: #4537

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Unit tests passing

### DIFF
--- a/pkg/actors/actors.go
+++ b/pkg/actors/actors.go
@@ -87,7 +87,7 @@ type actorsRuntime struct {
 	store                    state.Store
 	transactionalStore       state.TransactionalStore
 	placement                *internal.ActorPlacement
-	grpcConnectionFn         func(ctx context.Context, address, id string, namespace string, skipTLS, recreateIfExists, enableSSL bool, customOpts ...grpc.DialOption) (*grpc.ClientConn, error)
+	grpcConnectionFn         func(ctx context.Context, address, id string, namespace string, skipTLS, recreateIfExists, enableSSL bool, customOpts ...grpc.DialOption) (*grpc.ClientConn, func(), error)
 	config                   Config
 	actorsTable              *sync.Map
 	activeTimers             *sync.Map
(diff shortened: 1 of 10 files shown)


pkg/runtime/runtime_test.go
```diff
@@ -4117,3 +4122,132 @@ func TestComponentsCallback(t *testing.T) {
...
+		require.NotNil(t, resp, "resp must not be nil")
+
+		require.NoError(t, stream2.Send(req), "sending to PingStream must not fail")
+		resp, err = stream2.Recv()
+		require.NoError(t, err)
+		require.NotNil(t, resp, "resp must not be nil")
+
+		require.NoError(t, stream1.CloseSend(), "no error on close send")
+		require.NoError(t, stream2.CloseSend(), "no error on close send")
+	})
```
Syuparn, line 4196: I made a concurrent request unittest here because the issue is caused by the combination of `GetGRPCConnection` and `TransparentHandler`.

yaron2:
Thanks for this fix. Notice there are linter failures

yaron2:
The linter logs are a mess here but I think the message `##[error]shadow: declaration of "err" shadows declaration at line 351 (govet)` refers to this https://github.com/dapr/dapr/pull/4667/files#diff-aeb183ea3b81464cb579aa3e309be6bbfa88d6d900cfdbddb8c74d89ee7d4dc6R359.

Syuparn:
I fixed error shadowing.

pkg/actors/actors.go
```diff
@@ -356,9 +356,10 @@ func (a *actorsRuntime) callRemoteActorWithRetry(
 
 		code := status.Code(err)
 		if code == codes.Unavailable || code == codes.Unauthenticated {
-			_, err = a.grpcConnectionFn(context.TODO(), targetAddress, targetID, a.config.Namespace, false, true, false)
-			if err != nil {
-				return nil, err
+			_, teardown, cerr := a.grpcConnectionFn(context.TODO(), targetAddress, targetID, a.config.Namespace, false, true, false)
+			teardown()
```
yaron2, line 360: Should this be deferred?

Syuparn: `defer` is not used intentionally.

This line just creates a new connection and does not receive it, so `teardown()` can be called immediately.

Also, since this line is in a `for` loop, `defer` is evaluated after the whole loop ends. All old connections created by each iteration remain until the end of the function.

I referred to this post `#2 — Defer inside a loop`.
https://blog.learngoprogramming.com/gotchas-of-defer-in-go-1-8d070894cb01

yaron2 (approved):

skyao:
@Syuparn  @yaron2  Daprd is crashed now in perf test because of race condition. 

Please have a look at issue https://github.com/dapr/dapr/issues/4688

Issue #4641, opened 2022-05-20
Declarative pub/sub subscriptions are picked from stand alone components even if they are not in a .yaml file


francotiveron:
> /area runtime
## What version of Dapr?
> 1.7.3
## Expected Behavior
If I rename subscription.yaml to subscription.txt, the subscription should not be picked. This is useful for test purpose, otherwise we must delete the file and then recreate it.
## Actual Behavior
Subscription is activated as if the file still has .yaml subscription

PR #4679, opened 2022-05-31
add file extension check when loading subscription declaratively in selfhosted mode

wwbweibo:
Signed-off-by: wwbweibo <wwbweibo@outlook.com>

# Description

add file extension check when loading subscription declaratively in selfhosted mode

## Issue reference

Please reference the issue this PR will close: #4641

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Unit tests passing


ItalyPaleAle (approved):
LGTM, thanks for this contribution :)

halspang (approved):
This looks like the correct behavior to me, but I worry people may have been using improperly named files (with correct yaml) to load their subscriptions. Which, I think, may make this a breaking change.

@yaron2 / @artursouza - Thoughts?

beiwei30:
If we check where `os.ReadDir()` gets called in runtime project, we'll notice there are three usages which all intend to load yaml config files:

1. standalone_loader.go:49 verify file extension:

```go
// LoadComponents loads dapr components from a given directory.
func (s *StandaloneComponents) LoadComponents() ([]components_v1alpha1.Component, error) {
	files, err := os.ReadDir(s.config.ComponentsPath)
	if err != nil {
		return nil, err
	}

	list := []components_v1alpha1.Component{}

	for _, file := range files {
		if !file.IsDir() && s.isYaml(file.Name()) {
			components := s.loadComponentsFromFile(file.Name())
			if len(components) > 0 {
				list = append(list, components...)
			}
		}
	}

	return list, nil
}
```

2. resiliency.go:152

```go
// LoadStandaloneResiliency loads resiliency configurations from a file path.
func LoadStandaloneResiliency(log logger.Logger, runtimeID, path string) []*resiliency_v1alpha.Resiliency {
	if _, err := os.Stat(path); os.IsNotExist(err) {
		return nil
	}

	files, err := os.ReadDir(path)
	if err != nil {
		log.Errorf("failed to read resiliences from path %s: %s", err)
		return nil
	}
```

3. subscriptions.go:225

```go
// DeclarativeSelfHosted loads subscriptions from the given components path.
func DeclarativeSelfHosted(componentsPath string, log logger.Logger) []Subscription {
	var subs []Subscription

	if _, err := os.Stat(componentsPath); os.IsNotExist(err) {
		return subs
	}

	files, err := os.ReadDir(componentsPath)
	if err != nil {
		log.Errorf("failed to read subscriptions from path %s: %s", err)
		return subs
	}
```

We should make sure all these three places behave same, either more strict or more relax. I personally support adding file ext check.

ItalyPaleAle:
That’s a good point, this could be a breaking change.

perhaps this could be handled by showing a warning first?

1. In Dapr 1.8 when we find a file without extension we continue loading it (in all places), but show a warning with a link to our preview features page ( https://docs.dapr.io/operations/support/support-preview-features/ )
2. In a future version (probably Dapr 1.10?) we stop loading those files entirely.

wwbweibo:
> I agree with @beiwei30 that we should modify all methods that read files to have a file extension check.
> 
> @wwbweibo If you can please make the following changes:
> 
> 1. Add file extension checks to the methods listed above
> 2. Add a warning log message whenever we load a file with the wrong extension, prompting that it will not be loaded
> 
> Re: breaking changes, I think we're ok here considering the backward compatibility guarantees are applied to our APIs and CRD schemas plus the warning message.

I have added the warning message for the three file listed above, and moved the IsYaml file to utils package.

beiwei30 (approved):
LGTM

pkg/components/standalone_loader.go
```diff
@@ -54,7 +53,11 @@ func (s *StandaloneComponents) LoadComponents() ([]components_v1alpha1.Component
 	list := []components_v1alpha1.Component{}
 
 	for _, file := range files {
-		if !file.IsDir() && s.isYaml(file.Name()) {
+		if !file.IsDir() {
+			if !s.isYaml(file.Name()) {
```
beiwei30, line 57: would it better to directly call `utils.IsYaml()` and remove local `isYaml`?

pkg/components/standalone_loader.go
```diff
@@ -54,7 +53,11 @@ func (s *StandaloneComponents) LoadComponents() ([]components_v1alpha1.Component
...
+				log.Warnf("A non-YAML file %s was detected, it will not be loaded", file.Name())
```
ItalyPaleAle, line 58: I would recommend making these logs more explicit

1. What kind of YAML file? Should specify if this is a component definition, etc
2. If possible, add a link to something. Could be this PR too, if there's a comment that's relevant.

pkg/components/standalone_loader.go
wwbweibo: Ok, I will change it later.

pkg/components/standalone_loader.go
wwbweibo: 1. To add the kind of YAML file require read and parse the file first, I think that may cause unnecessary performance loss, I'm not sure that's necessary.
2. I will add a URL refer to this PR.

ItalyPaleAle: For 1, is it not possible to guess the kind of file here? In this place, for example, you're reading components, so it may be possible to say that you found "A non-YAML component file"?

wwbweibo: Sorry for misunderstanding, it's ok to do so.

pkg/components/standalone_loader.go
```diff
@@ -54,7 +53,11 @@ func (s *StandaloneComponents) LoadComponents() ([]components_v1alpha1.Component
 	list := []components_v1alpha1.Component{}
 
 	for _, file := range files {
-		if !file.IsDir() && s.isYaml(file.Name()) {
+		if !file.IsDir() {
+			if !utils.IsYaml(file.Name()) {
+				log.Warnf("A non-YAML component file %s was detected, it will not be loaded, for more detail, please refer to https://github.com/dapr/dapr/pull/4679", file.Name())
```
artursouza, line 58: Please, do not refer to GitHub. You can keep a simpler message now and then update once you have added documentation in the docs website.

utils/resolvconf.go
```diff
@@ -70,3 +71,12 @@ func getResolvSearchDomains(resolvConf []byte) []string {
 
 	return domains
 }
+
+// IsYaml checks whether the file is yaml or not.
+func IsYaml(fileName string) bool {
```
artursouza, line 76: This seems to be "too generic" for this file. Please, move to utils.go.

pkg/runtime/pubsub/subscriptions.go
```diff
@@ -230,6 +231,10 @@ func DeclarativeSelfHosted(componentsPath string, log logger.Logger) []Subscript
 
 	for _, f := range files {
 		if !f.IsDir() {
+			if !utils.IsYaml(f.Name()) {
+				log.Warnf("A non-YAML pubsub file %s was detected, it will not be loaded, for more detail, please refer to https://github.com/dapr/dapr/pull/4679", f.Name())
```
artursouza, line 235: No reference to GitHub, please. Same comment.

pkg/resiliency/resiliency.go
```diff
@@ -158,6 +159,10 @@ func LoadStandaloneResiliency(log logger.Logger, runtimeID, path string) []*resi
 	configs := make([]*resiliency_v1alpha.Resiliency, 0, len(files))
 
 	for _, file := range files {
+		if !utils.IsYaml(file.Name()) {
+			log.Warnf("A non-YAML resiliency file %s was detected, it will not be loaded, for more detail, please refer to https://github.com/dapr/dapr/pull/4679", file.Name())
```
artursouza, line 163: No reference to GitHub, please. Same comment.

artursouza (changes requested):

yaron2:
@wwbweibo please address the comments so we can merge this.

wwbweibo:
I force pushed some code, so I can not address the requested change. I'm so sorry.

yaron2 (approved):
I fixed the linter errors. LGTM

yaron2:
Thank you for this contribution @wwbweibo.

Issue #4688, opened 2022-06-01
Daprd crashed for race condition of grpc connection pool


skyao:
## In what area(s)?

/area runtime

## What version of Dapr?

> master

## Expected Behavior

Perf test should pass.

## Actual Behavior

In perf test case (service invoke http) , Daprd crashed for race condition of grpc connection pool.

See page: 
https://github.com/dapr/dapr/actions/runs/2418445782

Download the file [perf_container_logs](https://github.com/dapr/dapr/suites/6737249116/artifacts/257110634) in this page and open the log file tester-7fdf9754fc-vzgq6.daprd.log:

```go
fatal error: concurrent map writes
fatal error: concurrent map writes

goroutine 563 [running]:
runtime.throw({0x3eabf72?, 0x3a57320?})
	/opt/hostedtoolcache/go/1.18.2/x64/src/runtime/panic.go:992 +0x71 fp=0xc0011f6e78 sp=0xc0011f6e48 pc=0x438471
runtime.mapassign_faststr(0x37117c0, 0xc000ce1c80, {0xc0005bc8a0, 0x2f})
	/opt/hostedtoolcache/go/1.18.2/x64/src/runtime/map_faststr.go:295 +0x38b fp=0xc0011f6ee0 sp=0xc0011f6e78 pc=0x41364b
github.com/dapr/dapr/pkg/grpc.(*connectionPool).Register(0xc000ccf1c0, {0xc0005bc8a0, 0x2f}, 0xc000c52000)
	/home/runner/work/dapr/dapr/pkg/grpc/grpc.go:191 +0x7c fp=0xc0011f6f10 sp=0xc0011f6ee0 pc=0x16f759c
github.com/dapr/dapr/pkg/grpc.(*Manager).GetGRPCConnection(0xc000c99600, {0x45f5be8, 0xc000132008}, {0xc0005bc8a0, 0x2f}, {0xc0007cb248, 0x7}, {0xc00007c00a, 0xa}, 0x0, ...)
	/home/runner/work/dapr/dapr/pkg/grpc/grpc.go:168 +0xfba fp=0xc0011f7238 sp=0xc0011f6f10 pc=0x16f715a
[...]

PR #4691, opened 2022-06-01
Add referenceCount lock to grpc connection

yaron2:
Closes https://github.com/dapr/dapr/issues/4688.


yaron2:
@skyao please review.

pkg/grpc/grpc.go
```diff
@@ -202,15 +207,23 @@ func (p *connectionPool) Share(address string) (*grpc.ClientConn, bool) {
...
+	p.referenceLock.Lock()
+	defer p.referenceLock.Unlock()
+
 	p.referenceCount[conn]++
 	return conn, true
 }
 
 func (p *connectionPool) Release(conn *grpc.ClientConn) {
+	p.referenceLock.RLock()
 	if _, ok := p.referenceCount[conn]; !ok {
```
skyao, line 219: Hi, yaron, I would like to know what are the odds of this if happening? 

- if very high: then we should keep this optimization to use a lightweight read lock and expect most of the call be return in this if check.
-  if not very high: then I suggest to remove this optimization to use a BIG lock for the whole method scope. And if we do this, we can replace this RW lock with a simple (and also be quick) lock since we only use READ lock here.

yaron2: Looks like perf is unaffected. cc @skyao

skyao: @yaron let's quickly competed this PR to make perf test pass.

skyao (approved):
OK

Issue #4428, opened 2022-03-25
Issuer filenames configurable via options


safanaj:
/area runtime
/area sentry

## Describe the feature
Currently the path for the dapr-trust-bundle secret, used by sentry to issuing certs are set as constants:
https://github.com/dapr/dapr/blob/4cf499448ef6ee87c83db6a11b84e83237e92665/pkg/credentials/certchain.go#L7

This don't play well for integration with other CA/certificates managers like cert-manager, and the k8s standard way to layout TLS certs via secrets of type kubernetes.io/tls

I would like to have a command line option on all the binaries to change those key/path values.

The goal is to have option similar to --issuer-credentials (https://github.com/dapr/dapr/blob/master/cmd/sentry/main.go#L49) but for the constants in credentials package related to the IssuerCertFilename and IssuerKeyFilename, to be able to change the values from issuer.crt to tls.crt and issuer.key to tls.key.

Doing so the integration with auto-renewal external certs manager (see: #3968) will be really easy and smooth way.

safanaj:
Hi @daixiang0 , sorry for the delay, yes of course I am interested into have this implemented, are you asking for a patch proposal?

safanaj:
Hi @daixiang0 , I started to work on a proposal for this feature, the draft PR is #4699 

I would appreciate some guidance about make that PR compliant/ready for the approval.

Thanks

Issue #3968, opened 2021-11-30
Support use third-party CA to replace sentry


daixiang0:
/area sentry

## Describe the proposal

From what I know, now `sentry` component works as a CA server and control mTLS between Dapr sidecars. 

Consider this case -- a Kubernetes cluster has API server, it works as a CA, setup service mesh like Istio, it also setup a CA, setup Dapr, another CA. Now we have 3 CA servers, maybe more due to business needs. Each CA has its own expire date but provides similar functions, maintaining them is very complex.

/cc @artursouza @yaron2 @wcs1only WDYT?

yaron2:
Yes, I think having cert-manager support will be of great help, especially towards auto-rotating root certs.

daixiang0:
Yes, now we need to rotate it by hand refer to https://docs.dapr.io/operations/security/mtls/#updating-root-or-issuer-certs

ItalyPaleAle:
I really like this idea.

In the theme of supporting third-party CA's, I would like to suggest that we support Azure Key Vault as well, as it would allow better security (keys can't be extracted), automatic cert rotation, etc.

CC: @paulyuk

ItalyPaleAle:
👋 bot

daixiang0:
@safanaj I see you create a transformer as https://github.com/safanaj/dapr-cert-transformer, are you interested in introducing it into dapr, not a single project but an internal pkg?

safanaj:
@daixiang0 I wrote that "transformer" as workaround because the filenames/keys for the TLS secret are constants, I would suggest to fix that instead to add a component, but also the latter approach could be ok, just I don't think I have time to work on a proper integration code now

ItalyPaleAle:
@safanaj what does your code do? It wasn't clear from the README and I am not allowed to read your code because it's GPL'd 😔

safanaj:
Hi @ItalyPaleAle , basically it just keep watching a secret (especially dapr-trust-bundle) for the keys "tls.key" and "tls.crt" and copy the content respectively to "issuer.key" and "issuer.crt" , this is pretty useful if you want the dapt-trust-bundle to be be managed by cert-manager, and provide it to dapr-sentry and all other components.

> I am not allowed to read your code because it's GPL'd 😔
what do you mean?

ItalyPaleAle:
@safanaj thanks for explaining, makes sense now and I see its value - although I agree that we should probably be looking into an integrated solution rather than (yet another) background service.

[...]

PR #4699, opened 2022-06-01
allow to set credentials related filenames/secret keys by flags

safanaj:
# Description

This add 3 flags to allow to change the values for TLS related filenames (usually used as keys in Secret object) to each binary. 

## Issue reference

#4428 

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing


pkg/credentials/certchain.go
```diff
@@ -4,13 +4,13 @@ import (
 	"os"
 )
 
-const (
+var (
 	// RootCertFilename is the filename that holds the root certificate.
-	RootCertFilename = "ca.crt"
+	RootCertFilename string = "ca.crt"
```
daixiang0, line 9: Where is the default path for certs?

pkg/credentials/certchain.go
```diff
@@ -4,13 +4,13 @@ import (
 	"os"
 )
 
-const (
+var (
```
daixiang0, line 7: Why change to var?

pkg/credentials/certchain.go
safanaj: each binary is defining it, a bit of duplication there,
https://github.com/dapr/dapr/blob/master/cmd/sentry/main.go#L42
https://github.com/dapr/dapr/blob/master/cmd/placement/config.go#L27
https://github.com/dapr/dapr/blob/master/cmd/operator/main.go#L39

injector is getting them from the Secret directly, no need to mount it as volume,
https://github.com/dapr/dapr/blob/master/pkg/injector/pod_patch.go#L320

pkg/credentials/certchain.go
safanaj: to allow each binary to change them during the early init phase, like parsing flags. This could be dirty, but is the easiest and more consistent way avoiding to pass additional parameters where ever those credentials are used

safanaj: https://github.com/dapr/dapr/search?q=credentials.IssuerCertFilename

maybe not many places, but with the var is easier and no other change in the code.

pkg/credentials/certchain.go
daixiang0: I would like to put path in `pkg/credentials` so that we only one config entry.

pkg/credentials/certchain.go
daixiang0: It is fine for now since they are all independent services.

pkg/credentials/certchain.go
safanaj: totally agree, I just changed as few code as possible just because I don't know much about dapr, this proposal, for me, is just to get rid of this workaround: https://github.com/safanaj/dapr-cert-transformer/blob/master/main.go#L63

I am running this workaround because I used to manage all the tls certs with cert-manager, as also wished in here #3968

safanaj:
Hi @yaron2 @daixiang0 , I didn't get if this is gonna to be merged/included or what? Is it something needed to have this merged and released?

daixiang0 (dismissed):
LGTM

yaron2:
@safanaj please resolve the last linter warnings and we can merge this.

https://github.com/dapr/dapr/pull/4699/files

safanaj:
@yaron2 if I understood what were the linter warnings I removed the type in the var declarations.

yaron2:
> @yaron2 if I understood what were the linter warnings I removed the type in the var declarations.

Thanks, there's just a simple conflict to resolve now. Will merge once that's done.

yaron2 (approved):

yaron2:
Thanks for your contribution @safanaj!

PR #4719, opened 2022-06-07
Added option to capture diagnostic logs from AKS

ItalyPaleAle:
This PR adds the ability to collect diagnostic logs from E2E and Perf tests on AKS.

These logs include:

- All container logs, sent to Azure Log Analytics. This is better than our current log collection as it captures logs for crashed containers and even after the test has stopped running.
- Diagnostic logs (`kube-apiserver` and `kube-controller-manager`), sent to Azure Log Analytics.
- Audit logs (`kube-audit`), sent to Azure Storage.

These logs are particularly useful for debugging issues with E2E and Perf tests on AKS.

Note that the infrastructure where logs are collected (the Log Analytics workspace and Storage account) is static and pre-deployed, and it's not deployed as part of each test (otherwise the logs would get deleted automatically too!). Data is persisted for a short number of days only (30 for Log Analytics, 14 for Azure Storage). Because those logs could contain sensitive information (audit logs include a lot of secrets!), they are not exported from the places where they are stored.

**Note:** I've already deployed the infrastructure in our test Azure subscription. To make the diagnostic logs work in our Actions runs, a STC member will need to add two secrets to the dapr/dapr repo: `AZURE_DIAG_LOG_ANALYTICS_WORKSPACE_ID` and `AZURE_DIAG_STORAGE_ID` (I can provide the values privately). The pipelines will work even without those secrets, but diagnostic logs won't be collected

### DIFF
--- a/.github/workflows/dapr-perf.yml
+++ b/.github/workflows/dapr-perf.yml
@@ -16,7 +16,7 @@
 # - PERF_AZURE_STORAGE_ACCOUNT and PERF_AZURE_STORAGE_KEY: Credentials for the Storage Account where to store the result of perf tests
 # - DAPR_BOT_TOKEN: Token for the Dapr bot
 #
-# Optional secets:
+# Optional secrets:
 # - AZURE_DIAG_LOG_ANALYTICS_WORKSPACE_ID: Resource ID of the Log Analytics Workspace where to store certain diagnostic logs (e.g. `/subscriptions/<subscription>/resourcegroups/<resource group>/providers/Microsoft.OperationalInsights/workspaces/<workspace name>`)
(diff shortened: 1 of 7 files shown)


skyao (approved):
Great PR. Let's try it!

yaron2:
> a STC member will need to add two secrets to the dapr/dapr repo: AZURE_DIAG_LOG_ANALYTICS_WORKSPACE_ID and AZURE_DIAG_STORAGE_ID (I can provide the values privately)

Please provide these to @artursouza.

artursouza (changes requested):

artursouza (approved):

Issue #4171, opened 2022-01-27
Make Sidecar injection resilient in case of Kubernetes API interruptions / admissions webhook failures etc


berndverst:
## In what area(s)?

/area injector
/area sidecar-injector

## Describe the feature
Dapr Sidecar Injector is responsible for handling the creation of Dapr sidecars and injecting the Dapr Trust Bundle secret from the Dapr-System namespace. Internally this approach uses the Kubernetes `MutatingAdmissionWebhook` feature.

On occasion the MutatingAdmissionRequest to the webhook which is registered to the Dapr Sidecar injector (see `kubectl get MutatingWebhookConfiguration/dapr-sidecar-injector -oyaml`) may not occur or may fail. At this time Dapr Helm Charts are configured to resume with the deployment in that case. This means new application deployments with Dapr annotations that should have sidecars injected will have missing sidecars.

**The Dapr control plane should periodically (or upon Kubernetes API failure) determine whether any existing deployments need sidecars injected and whether new components must be loaded.** The Dapr control plane also needs to gracefully resume watching of resource deployments when the Kubernetes API is available again.

## Release Note

RELEASE NOTE:

berndverst:
This issue may also come in handy in case an existing deployment is modified with dapr annotations when it previously didn't have dapr annotations. I'm unclear on the behavior today (haven't further looked into this).

yaron2:
> This issue may also come in handy in case an existing deployment is modified with dapr annotations when it previously didn't have dapr annotations. I'm unclear on the behavior today (haven't further looked into this).

An existing deployment with added annotations will get the sidecars injected.

artursouza:
> > This issue may also come in handy in case an existing deployment is modified with dapr annotations when it previously didn't have dapr annotations. I'm unclear on the behavior today (haven't further looked into this).
> 
> An existing deployment with added annotations will get the sidecars injected.

We have identified other scenarios where the sidecar is not injected. We need to create some sort of background process to identify those instances and proactively fix it.

MattCosturos:
Will this issue be fixed in 1.8? 
We had an outage, and our system did not recover due to this issue.

yaron2:
> Will this issue be fixed in 1.8? We had an outage, and our system did not recover due to this issue.

@artursouza to provide info here as triager.

berndverst:
@yaron2 I just updated all my incorrect understanding in the issue. @ItalyPaleAle is currently investigating.

[...]

PR #4721, opened 2022-06-07
Changed default `webhookFailurePolicy` to `Fail`

ItalyPaleAle:
This is an **experimental** PR that changes our Helm chart's default to `Fail`. See #4171

We will run some tests to see how this performs, and decide if we want to keep it.

Because this may make apps fail to deploy, changed the E2E test framework too to add a retry policy.


ItalyPaleAle:
Moving to draft because of the concerns expressed here: https://github.com/dapr/dapr/issues/4171#issuecomment-1149079172

PR #4729, opened 2022-06-08
Build tools CLI (plus speed up E2E tests by copying test apps without pulling them)

ItalyPaleAle:
One of the longest steps in our E2E tests right now is building E2E test apps and pushing them to the container registry used for tests, especially on Windows.

#4618 made this much faster by caching some of the Docker images (for the apps that had a Dockerfile - not for those that were built with `go build`). But we can do better :)

The biggest problem is Windows, as the GitHub Actions agents on Windows seem to have issues with disk I/O performance. This makes even using the cache for the Docker images slow: images are first pulled into the cache, and then pushed to the other container registry.

All of the above takes time. For example, in this run (which was faster than the average), it took 9 minutes to build the E2E test images on Windows (including pulling from the cache those that can be cached). Pushing all images (including E2E test apps and Dapr images themselves) took an additional 11m44s:

image
image

We can do much better 💪

image
image

The jobs here are grouped separately, but in this test run it took not even 3.5 minutes to build all E2E test app images, and push all images for Dapr and E2E test apps to the container registry, on Windows.

**The end result is that the "Build for Windows" step took ~18 minutes less to complete, making our entire E2E tests start (and complete) much faster! 🏎** (Of course, when cached images for E2E test apps are available)

This PR implements two strategies to speed up building and pushing E2E test app images to the test container registry:

1. When an image is found in the cache, it is copied directly between the two container registries without being pulled locally first. This happens using [`crane`](https://github.com/google/go-containerregistry/tree/main/cmd/crane) and makes it possible for us to completely avoid the local disk I/O (which is what is slow on Windows)
   - `make build-e2e-app-all` and `make push-e2e-app-all` continue to build images locally and pushing images that are stored locally. To use the new registry-to-registry workflow, one needs to use `make build-push-e2e-app-all`. If the image isn't available in the cache, this automatically goes back to building and pushing.
2. We are now caching all Docker images, even those that are built by just building Go apps and copying them into the container. Although building those containers is quick, on Windows it causes quite a bit of (slow) disk I/O when then pushing the images.

PS: As an additional thing, all the operations above are now performed by a new tool written in Go called `build-tools`. The makefiles continue to work, but now invoke `build-tools` behind the scenes. The reason for this is that I found myself having to deal with things that could not be conveniently built in Bash (such as having to compute the hash of all files in a folder while taking into account files excluded by `.gitignore`), and there were too many possible states and options, resulting in the solution not being easy to maintain when it was split between the makefile and a (set of) Bash scripts.


build-tools/cmd/zz-e2e-perf.go
```diff
@@ -0,0 +1,460 @@
...
+// Builds a Docker image for the app
+// It also pushes it to the cache registry if that's enabled
+func (c *cmdE2EPerf) buildDockerImage(cachedImage string) error {
+	destImage := c.getDestImage()
+	appDir := c.getAppDir()
+
+	// First, check if the image has its own Dockerfile
+	dockerfile := filepath.Join(appDir, c.flags.Name, c.flags.Dockerfile)
+	_, err := os.Stat(dockerfile)
+	if err != nil {
```
halspang, line 270: Other than pointing this at the 'shared' Dockerfile, is there a point for building here? Wont' the docker build do that right after this?

.github/workflows/dapr-perf.yml
```diff
@@ -231,11 +226,15 @@ jobs:
         run: |
           az acr login --name ${{ env.TEST_PREFIX }}acr
         shell: bash
-      - name: Push all container images to the registry
+      - name: Push container images
```
halspang, line 229: Can we call this `Push Dapr container images` or something more specific than just `container`? I was confused at first as this push is happening before we build the apps below.

build-tools/cmd/zz-e2e-perf.go
```diff
@@ -0,0 +1,460 @@
...
+	return cmd
+}
+
+// Handler for the "build" sub-command
+func (c *cmdE2EPerf) buildCmd(cmd *cobra.Command, args []string) error {
+	// Target image and tag
+	destImage := c.getDestImage()
+
+	// If cache is enable, try pulling from cache first
+	cachedImage, err := c.getCachedImage()
```
halspang, line 147: It took a bit for me to realize that the hash validation happens because the filepath is no longer there. Think you can add a comment here (or at the method) stating that it handles hash validation there?

halspang (commented):
Overall, this change seems to be good to me. Though I'm no expert in Makefiles :)

I think the main thing I'd like to see is an additional README in the `build-tools` dir explaining why they're there and how we may extend them if we need to. This change does add complexity and I think it's important to keep note of why we're doing that in there instead of making people look for the PR.

ItalyPaleAle:
Thanks for the review @halspang 

> Though I'm no expert in Makefiles :)

Yes, that's precisely the point... Makefiles are particularly complex and messy. The other day, when I was researching on why my Makefile wasn't working and turns out that `=` is for recursive assignment while I had to do `:=` for a simple assignment, I read someone writing this on StackOverflow:

> GNU make is hardly known for elegant solutions. Unless you find trapdoors and minefields to be elegant.

What I was trying to do here was, IMHO, far beyond what could sensibly be done with Make, and beyond even what I could have accomplished with a Bash script (the hashing part).

I am also not a huge fan of how our Makefiles depend heavily on environmental variables (because they didn't have many alternatives). Env vars are essentially part of the global state of the shell. They're tied to a specific window/tab in the terminal, and it's hard to know at a given time what the state is (many times I've deleted the wrong namespace on the K8s test cluster - it was just test stuff, but still). It's also unclear what env vars are available for each command, what the default value is, etc.

I am hoping that a solution based on a CLI like this is more approachable, and friendlier to new contributors. Each command can be self-documenting with `--help` which explains the options (which are always CLI flags), what they do, and the default values.

As for the other points of your review, I'll address them shortly :)

build-tools/cmd/root.go
```diff
@@ -0,0 +1,51 @@
...
+// Execute adds all child commands to the root command and sets flags appropriately.
+// This is called by main.main(). It only needs to happen once to the rootCmd.
+func Execute() {
+	err := rootCmd.Execute()
+	if err != nil {
+		os.Exit(1)
+	}
+}
+
+func init() {
```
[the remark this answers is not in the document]
ItalyPaleAle: All these comments were placeholders created by the cobra generator. I'll remove them (I am adding a README however).

build-tools/cmd/zz-e2e-perf.go
ItalyPaleAle: Done

build-tools/cmd/zz-e2e-perf.go
ItalyPaleAle: No, apps that don't have a Dockerfile are built outside of Docker, then the Dockerfile just copies the binary in the image. This has been like this for a while, and I like that it helps reducing the number of go.mod files (and makes creating tests much simpler)

.github/workflows/dapr-perf.yml
ItalyPaleAle: Done

ItalyPaleAle:
Made the changes you've requested @halspang

halspang (dismissed):

Makefile
```diff
@@ -385,6 +395,14 @@ check-proto-diff:
 	git diff --exit-code ./pkg/proto/sentry/v1/sentry.pb.go # check no changes
 
 
+################################################################################
+# Target: compile-build-tools                                                              #
+################################################################################
+compile-build-tools:
+ifeq (,$(wildcard $(BUILD_TOOLS)))
+	cd build-tools; CGO_ENABLED=$(CGO) GOOS=$(TARGET_OS_LOCAL) GOARCH=$(TARGET_ARCH_LOCAL) go build -o $(BUILD_TOOLS_BIN) .
```
artursouza, line 403: Please, rename this folder to `.build-tools` to keep it separate from the main code.

ItalyPaleAle: Done

artursouza (approved):

PR #4733, opened 2022-06-09
[DO NOT MERGE] Update AKS version

ItalyPaleAle:
Testing if a newer K8s version improves our tests. Do not merge


ItalyPaleAle:
Upgrading the K8s version to 1.23 doesn't seem to fix the issue. Experiment failed.

PR #4741, opened 2022-06-10
Allow setting custom certificates in Windows containers

shubham1172:

# Description

With the feature to add volume mounts using #4662, it is possible for users to mount a volume with custom certificates. On Linux containers, the `SSL_CERT_DIR` environment variable can be used to make the daprd sidecar trust these certificates. On Windows, we need to add a special entry-point script that adds these certificates to the root store.

With this PR, we make the following changes
- In the Windows base Docker image, add the `certoc` utility and the certificate setup script to be able to add certs to the root store.
- Add an entrypoint to run the setup script and accept args to `cmd` for running dapr.
- This script looks for the env variable `SSL_CERT_DIR` and if present, it adds the certificates present in the directory point by the environment variable to the root store.
- In `pod_patch.go`, we no longer use both `Command` and `Args`. `Command` in container spec translated to the Docker ENTRYPOINT, which we no longer override. The ENTRYPOINT is either explicitly specified (as for Windows), or the default is used (as for Linux). `pod_patch.go` passes `Args` to the ENTRYPOINT to invoke Dapr (or Delve).
- Also, if `SSL_CERT_DIR` is set, `pod_patch.go` configures the Windows container to use the `ContainerAdministrator` user. This is required by `certoc.exe` to install certificates to the root store.

## Issue reference

Please reference the issue this PR will close: #4587 

## Checklist

* [X] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing
* [ ] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: https://github.com/dapr/docs/issues/2570 


docker/Dockerfile-windows
```diff
@@ -1,6 +1,11 @@
 ARG WINDOWS_VERSION=1809
 #This version confirmed to work for both github and AKS
 FROM ghcr.io/dapr/windows-base:$WINDOWS_VERSION
+
+# Administrator access is required to install certificates.
+USER ContainerAdministrator
+
 ARG PKG_FILES
 WORKDIR /
+COPY /setup-certificates.cmd /
```
ItalyPaleAle, line 10: Should this be added to the base dockerfile (`Dockerfile-windows-base`) instead? It would allow skipping the "awkward" copying here: https://github.com/dapr/dapr/pull/4741/files#diff-c5390b6b2b8f0bc8fb6702bb8aede7ad4d8344393456674aa8678480d56c4c76R93-R95

pkg/injector/pod_patch.go
```diff
@@ -847,6 +848,17 @@ func getSidecarContainer(annotations map[string]string, id, daprSidecarImage, im
 	if resources != nil {
 		c.Resources = *resources
 	}
+
+	if runtime.GOOS == "windows" {
```
ItalyPaleAle, line 852: Rather than checking the OS of the control plane, can you check the OS of the container that is being deployed?

Imagine the case where Dapr's control plane is running on Linux, but it's deploying a Windows container.

shubham1172: Actually I was exploring that. So the pod spec has a field called `PodOS`, [documentation](https://pkg.go.dev/k8s.io/api/core/v1#PodSpec), but it requires a special feature flag to be set - `IdentifyPodOS`.

From [Feature Gates | Kubernetes](https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/)
> `IdentifyPodOS`: Allows the Pod OS field to be specified. This helps in identifying the OS of the pod authoritatively during the API server admission time. In Kubernetes 1.24, the allowed values for the `pod.spec.os.name` are `windows` and `linux`.


|Feature| Default|Stage|Since|Until|
-- | -- | -- | -- | --
|IdentifyPodOS | false | Alpha | 1.23 | 1.23
IdentifyPodOS | true | Beta | 1.24 |  

Since Dapr supports versions of k8s < 1.24, we will need to explicitly enable this feature to work. So we can conditionally support this feature based on the feature flag for Windows. What do you think?

ItalyPaleAle: Uhm that doesn't sound good either.

How about just running the `.cmd` file on every Dapr container? That is, we modify the launch command for *all* Windows containers to make that run before any script. I see that you've already added a line that checks if there's an env var, otherwise it just exits.

I believe this could be done by making the `.cmd` file the ENTRYPOINT of the container. You can still run any command you want which will be passed as shell arg to the `.cmd` file. The `.cmd` file will do what it needs to do, then read the shell arg and invoke that. See example here: https://stackoverflow.com/a/56587335/192024

shubham1172: This hypothesis is not turning out to be true. If I have an ENTRYPOINT and I invoke `docker run --rm -it <image name>` it seems to be working, but in the context of k8s, command seems to be overriding the ENTRYPOINT.

From [Define a Command and Arguments for a Container | Kubernetes](https://kubernetes.io/docs/tasks/inject-data-application/define-command-argument-container/),
> The command and arguments that you define in the configuration file override the default command and arguments provided by the container image. If you define args, but do not define a command, the default command is used with your new arguments.
Note: The command field corresponds to entrypoint in some container runtimes. 

Exploring more on this.

ItalyPaleAle: Uhm I see. What about just setting the command as "args"? See: https://stackoverflow.com/a/67113557/192024

Downsides is that this _may_ require updating the Linux image too (not 100% sure about that)

daixiang0: I think for windows case, should better add a doc to guide users on what should do, as considering and handling all situations is hard.

daixiang0: Yes, it is easy to implement.

shubham1172: > Uhm I see. What about just setting the command as "args"? See: https://stackoverflow.com/a/67113557/192024
> 
> Downsides is that this _may_ require updating the Linux image too (not 100% sure about that)

That does indeed work for Windows. We will need to add an `ENTRYPOINT` for Linux with this approach. Will confirm and update this thread.

shubham1172: Let me know what you think of the current approach @ItalyPaleAle @daixiang0 

Going the feature flag route we can skip the entrypoint modification and changing the pod_patch, which might be better since our new entry point will be a shell that handles dapr (as compared to running dapr directly). If we use feature flag, we can set our script to run via the k8s PostStart hook.

ItalyPaleAle: @shubham1172 excellent work testing this, thanks!! If you confirm that everything works (even on Linux, where our containers do not have `/bin/sh`), then it’s good with me!

shubham1172: Thanks @ItalyPaleAle. I confirm that this works on Linux distroless images (without a shell). Also from the docs:
image

ItalyPaleAle (dismissed):

docker/Dockerfile-windows
```diff
@@ -1,6 +1,12 @@
 ARG WINDOWS_VERSION=1809
 #This version confirmed to work for both github and AKS
 FROM ghcr.io/dapr/windows-base:$WINDOWS_VERSION
+
+# Administrator access is required to install certificates.
+USER ContainerAdministrator
```
artursouza, line 6: Is there a way for this not to be the default user? I assume most users will not need to install certificates. Maybe we need to publish a new image.

docker/Dockerfile-windows-base
```diff
@@ -1,9 +1,16 @@
 ARG WINDOWS_VERSION=1809
 #This version confirmed to work for both github and AKS
 FROM mcr.microsoft.com/windows/servercore:$WINDOWS_VERSION as servercore
-FROM mcr.microsoft.com/windows/nanoserver:$WINDOWS_VERSION
+
 # Nanoserver is missing exactly one dll that go needs. The servercore image has this dll, but
 # is nearly 2GB. This has the potential to be somewhat fragile, but it is the only way I can
 # find to keep the container image size small. To keep from doing this rather expensive
 # multistage build every time, we build a small base image and push it to our repo
+FROM mcr.microsoft.com/windows/nanoserver:$WINDOWS_VERSION
```
shubham1172, line 9: I was not able to find a reference to this Dockerfile in our GitHub workflows. Is there a process in place to update this image?

pkg/injector/pod_patch.go
```diff
@@ -723,8 +723,8 @@ func getSidecarContainer(annotations map[string]string, id, daprSidecarImage, im
 		SecurityContext: &corev1.SecurityContext{
 			AllowPrivilegeEscalation: &allowPrivilegeEscalation,
 		},
-		Ports:   ports,
-		Command: cmd,
```
artursouza, line 727: Why cmd moved to args?

artursouza (changes requested):

docker/Dockerfile-windows-base
ItalyPaleAle: https://github.com/dapr/dapr/actions/workflows/dapr-base-containers.yaml

pkg/injector/pod_patch.go
ItalyPaleAle: It's discussed in the comments above.

Essentially in K8s, Cmd is the ENTRYPOINT in Dockerfiles, and we need not to override that.

docker/Dockerfile-windows-base
shubham1172: Ah, thanks!

docker/Dockerfile-windows
shubham1172: This is a build-time setting, right? We can publish two images and select one in pod_patch based on if this environment variable is set, but that won't be very clean. 

Another option is to add another entrypoint logic in a `cmd` file, and based on if admin permissions are needed or not (i.e. if this environment variable is not set), it can runas `ContainerUser` instead of `ContainerAdministrator`. I will explore more on this.

artursouza: Please. I think not forcing Admin user is key.

shubham1172: Yes @ItalyPaleAle, I will test these out! Will reply on this thread.

shubham1172: With the latest update, we have a logic in-place to use Admin user if `SSL_CERT_DIR` was set using k8s YAML (as suggested by @mukundansundar) in the sidecar injector. This has been verified on both Linux and Windows.

Note, for Linux containers, the container will still contain the `windowsOptions` YAML, but it won't have any effect.
image

pkg/injector/pod_patch.go
```diff
@@ -776,6 +775,19 @@ func getSidecarContainer(annotations map[string]string, id, daprSidecarImage, im
 
 	c.Env = append(c.Env, utils.ParseEnvString(annotations[daprEnvKey])...)
 
+	// This is a special case that requires administrator privileges in Windows containers
+	// to install the certificates to the root store. If this environment variable is set,
+	// the container security context should be set to run as administrator.
+	for _, env := range c.Env {
+		if env.Name == "SSL_CERT_DIR" {
+			userName := "ContainerAdministrator"
+			c.SecurityContext.WindowsOptions = &corev1.WindowsSecurityContextOptions{
```
mukundansundar, line 784: why is the windows security option also added for linux containers if SSL CERT DIR is given?
Also here only an implicit consent is used to elevate container user to Admin ie SSL CERT DIR is given ... shouldn't there be an explicit consent given by the user say set an annotation so that user explicitly turns on this feature?

ItalyPaleAle: > why is the windows security option also added for linux containers if SSL CERT DIR is given?

We have no way to know if a container is Linux- or Windows- based

> shouldn't there be an explicit consent given by the user say set an annotation so that user explicitly turns on this feature?

IMHO, using ContainerAdministrator is necessary and sufficient for, and only for, using custom TLS certs. Perhaps this should be explained in the docs: if you enable that feature (and only in that case), the user within the container will be administrator.

docker/Dockerfile-windows
ItalyPaleAle: This works for me, thanks @shubham1172 for the investigation

pkg/injector/pod_patch.go
artursouza: I agree that we need docs for this.

shubham1172: https://github.com/dapr/docs/issues/2570 docs issue.

ItalyPaleAle:
@shubham1172 there's something happening on Windows and all apps fail to deploy :(

ItalyPaleAle:
@shubham1172 i think I know why windows tests fail. It's because you modified the base containers and our E2E tests do not rebuild the base containers in every run (for perf reasons).

To test this PR, what you need to do is:
1.  Build docker/Dockerfile-windows-base and push it to your own container registry (for example a public Docker hub)
2. TEMPORARILY change docker/Dockerfile-windows to point to your base image
3. Run the tests
4. When tests pass, revert change at point 2

ItalyPaleAle:
By the way, do you think we should create an E2E test specifically for root cert injection?

shubham1172:
From the logs:
`{Waiting:nil Running:nil Terminated:&ContainerStateTerminated{ExitCode:127,Signal:0,Reason:ContainerCannotRun,Message:container 2f63d3bb85ddfd03718dccd5a9b50e5afdaf1efcd978c4496964fe542b296ec0 encountered an error during hcsshim::System::CreateProcess: failure in a Windows system call: The system cannot find the file specified. (0x2),StartedAt:2022-06-25 12:45:11 +0000 UTC,FinishedAt:2022-06-25 12:45:11 +0000 UTC,ContainerID:docker://2f63d3bb85ddfd03718dccd5a9b50e5afdaf1efcd978c4496964fe542b296ec0,}`

Makes sense, let me push the suggested temporary changes to this PR and try again.

For E2E test, do you suggest using the HTTP binding scenario? Also, can we track it separately from this PR?

ItalyPaleAle:
> Makes sense, let me push the suggested temporary changes to this PR and try again.

Thanks! Now that we have build-tools I can also look into automating this for future PRs. I'll do that on Monday.

> For E2E test, do you suggest using the HTTP binding scenario?

I think that's fine. Perhaps:

1. Deploy an app that offers an HTTP server with a TLS cert signed by a custom CA.
2. This test would invoke the app directly (to get a baseline and make sure the root cert is injected)
3. Then it would try to invoke via HTTP binding to test another scenario.

> Also, can we track it separately from this PR?

Of course.

shubham1172:
Following @ItalyPaleAle's suggestion, the e2e test on Windows are now passing. I will revert the last commit to use the original Windows base image if this looks good to everyone. Note, we will probably have to manually trigger a windows-base image build once the PR is merged.

E2E test issue: #4835

shubham1172:
Update: this PR has been targeted to release-1.8 now.

ItalyPaleAle (approved):
Good work #ShipIt

pkg/injector/pod_patch.go
```diff
@@ -741,8 +741,8 @@ func getSidecarContainer(annotations map[string]string, id, daprSidecarImage, im
 		SecurityContext: &corev1.SecurityContext{
 			AllowPrivilegeEscalation: &allowPrivilegeEscalation,
 		},
-		Ports:   ports,
-		Command: cmd,
+		Ports: ports,
+		Args:  append(cmd, args...),
```
yaron2, line 745: Why this change?

ItalyPaleAle: See here: https://github.com/dapr/dapr/pull/4741#discussion_r901261917

yaron2: 10x

shubham1172: Command overrides the entrypoint in the docker image. This change ensures that we can run our custom entrypoint scripts. If Command is passed, the entrypoint gets overridden. Dapr and its args, are now passed as Arguments.

On non windows container, the default entrypoint is used since we don't specify an explicit one.

yaron2 (approved):

pkg/injector/pod_patch.go
ItalyPaleAle: 10x?

yaron2: > 10x?

ten + x = thanks

itac-admin:
Why did you change command to args?
"AKS with virtual kubelet/ACI" doesn't work with this change.

https://github.com/dapr/dapr/commit/9822921a92ecfce8f271feeb3e9825bfda95e295#diff-877ab4b4bd43d88fc79f83e7a1cbe540f963a07cea4b571b16ef6303dc0bf2f1R744-R745

ItalyPaleAle:
@itac-admin can you please provide more details? What's the error?

itac-admin:
@ItalyPaleAle 
The following error occurs when deploying with "aks with aci".

----------------------------
Events:
  Type     Reason                Age                From                                   Message
  ----     ------                ----               ----                                   -------
  Normal   Scheduled             10s                default-scheduler                      Successfully assigned backend-test/nginx-deployment-67f6b7c46-qvbnp to virtual-node-aci-linux
  Warning  ProviderCreateFailed  4s (x10 over 10s)  virtual-node-aci-linux/pod-controller  ACI does not support providing args without specifying the command. Please supply both command and args to the pod spec.
----------------------------

-----------------------
  daprd:
    Image:       mcr.microsoft.com/daprio/daprd:1.8.4
    Ports:       3500/TCP, 50001/TCP, 50002/TCP, 9090/TCP
    Host Ports:  0/TCP, 0/TCP, 0/TCP, 0/TCP
    Args:
      /daprd
      --mode
      kubernetes
-----------------------

ItalyPaleAle:
Thanks. I'm trying to connect with the ACI team and get their take on this too.

itac-admin:
thank you.
I want to use the latest Dapr.

ItalyPaleAle:
@itac-admin  can you confirm if you're using Windows or Linux containers?

ItalyPaleAle:
@itac-admin to close the loop on this, @shubham1172 has identified a possible solution and is working on it. PR is here: https://github.com/dapr/dapr/pull/5024

Feel free to provide some feedback there!

PR #4757, opened 2022-06-11
[DO NOT MERGE] Avoid running tests in parallel

ItalyPaleAle:
Test PR to see if running tests not in parallel makes a difference


ItalyPaleAle:
This doesn't seem to be making a significant difference. Closing.

PR #4764, opened 2022-06-14
Sidecar watchdog in operator service

### LINKED ISSUE #4171: Make Sidecar injection resilient in case of Kubernetes API interruptions / admissions webhook failures etc

ItalyPaleAle:
# Description

This PR adds another job to the operator service, adding a "watchdog" that periodically checks all pods to make sure that the sidecar is present.

This should help fixing issues such as #4171 (the sidecar may not be present in pods if the entire cluster is restarted and apps' pods are restarted before the dapr-sidecar-injector service), as well as mitigate issues such as those we are experiencing in our E2E tests, where the injector service does not receive the admission webhook.

## Issue reference

Fixes #4171

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Unit tests passing


ItalyPaleAle:
Looks like this is working and it is catching pods where the sidecar wasn't injected. From logs:

```json
{"instance":"dapr-operator-5445c95fbd-bgdtw","level":"warning","msg":"Pod hellobluedapr-98869bdf7-lr5ff does not have the Dapr sidecar and will be deleted","scope":"dapr.operator","time":"2022-06-14T02:22:04.821961274Z","type":"log","ver":"edge"}
{"instance":"dapr-operator-5445c95fbd-bgdtw","level":"info","msg":"Deleted pod hellobluedapr-98869bdf7-lr5ff","scope":"dapr.operator","time":"2022-06-14T02:22:04.832526661Z","type":"log","ver":"edge"}
```

Just need to complete the PR now :)

charts/dapr/README.md
```diff
@@ -114,6 +114,8 @@ The Helm chart has the follow configuration options that can be supplied:
 |-------------------------------------------|-------------------------------------------------------------------------|-------------------------|
 | `dapr_operator.replicaCount`              | Number of replicas                                                      | `1`                     |
 | `dapr_operator.logLevel`                  | Log level                                                               | `info`                  |
+| `dapr_operator.watchInterval`             | Interval for polling pods' state, in seconds (e.g. `120`). Set to `0` to disable, or `once` to only run once when the operator starts | `0` |
```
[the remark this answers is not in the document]
ItalyPaleAle: Done

cmd/operator/main.go
```diff
@@ -28,25 +30,60 @@ import (
...
+				log.Fatalf("invalid watch-interval value: if not '0' or 'once', must be at least 1s")
+			}
+			operatorOpts.WatchdogInterval = dur
+		}
+	}
+
 	ctx := signals.Context()
 
-	go operator.NewOperator(config, certChainPath, !disableLeaderElection).Run(ctx)
+	go operator.NewOperator(operatorOpts).Run(ctx)
```
artursouza, line 86: ❤️

pkg/operator/handlers/dapr_handler.go
```diff
@@ -273,12 +273,7 @@ func (h *DaprHandler) isAnnotatedForDapr(wrapper ObjectWrapper) bool {
 	if !ok {
 		return false
 	}
-	switch strings.ToLower(enabled) {
-	case "y", "yes", "true", "on", "1":
-		return true
-	default:
-		return false
-	}
+	return utils.IsTruthy(enabled)
```
artursouza, line 276: ❤️

pkg/operator/watchdog.go
```diff
@@ -0,0 +1,172 @@
...
+	maxRestartsPerMin int
+
+	client         client.Client
+	restartLimiter ratelimit.Limiter
+}
+
+// NeedLeaderElection makes it so the controller runs on the leader node only.
+// Implements sigs.k8s.io/controller-runtime/pkg/manager.LeaderElectionRunnable .
+func (dw *DaprWatchdog) NeedLeaderElection() bool {
+	return true
```
artursouza, line 34: Why even do this?

ItalyPaleAle:
Because  ok-to-test can't show the watchdog running in this PR, I've triggered it in a fork: https://github.com/ItalyPaleAle/dapr-ci-test/actions/runs/2505352950

pkg/operator/watchdog.go
```diff
@@ -0,0 +1,172 @@
...
+
+// Start the controller. This method blocks until the context is canceled.
+// Implements sigs.k8s.io/controller-runtime/pkg/manager.Runnable .
+func (dw *DaprWatchdog) Start(parentCtx context.Context) error {
+	if !dw.enabled {
+		log.Infof("DaprWatchdog is not enabled")
+		return nil
+	}
+
+	log.Debugf("DaprWatchdog worker started")
```
artursouza, line 45: Log info.

Also, it has not started yet. It is starting.

pkg/operator/watchdog.go
```diff
@@ -0,0 +1,172 @@
...
+			case <-workCh:
+				dw.listPods(ctx)
+				if firstCompleteCh != nil {
+					close(firstCompleteCh)
+					firstCompleteCh = nil
+				}
+			}
+		}
+	}()
+
```
artursouza, line 79: Now you can log that it has started. It can be log info too.

pkg/operator/watchdog.go
```diff
@@ -0,0 +1,172 @@
...
+				// Successfully queued another iteration
+				log.Debugf("Added listPods iteration to the queue")
+			default:
+				// Do nothing - there's already an iteration in the queue
+				log.Debugf("There's already an iteration in the listPods queue-not adding another one")
+			}
+		}
+	}
+
+	log.Debugf("DaprWatchdog worker stopping")
```
artursouza, line 111: Info.

Also, log when it stops.

artursouza (changes requested):

pkg/operator/watchdog.go
ItalyPaleAle: Operator service can run in HA mode where there are 3 instances. We need to make sure that only one instance of operator is killing pods, otherwise there are race conditions and pods can be killed too many times (and in general there can be too many invocations of the K8s APIs).

Luckily for us, the operator service already implemented a Controller Manager (see https://github.com/dapr/dapr/pull/4764/files#diff-99c0290799daf9abc6240df64063e20bfaf67b371577b67ac7eec6f4725622ffR91 ) , so I plugged this watchdog as a controller. 

All controllers need to implement the `sigs.k8s.io/controller-runtime/pkg/manager.Runnable` interface, which includes the `Start` method below.

This `NeedLeaderElection` method is an optional one to implement `sigs.k8s.io/controller-runtime/pkg/manager.LeaderElectionRunnable` which makes it so the controller only runs on the instance of the service that K8s elects as leader.

So if there are 3 instances of Operator, the watchdog runs only on one. If the leader dies, another instance will re-start the operator. I have manually verified this behavior.

PS: if the watchdog was configured to run only `once` and the leader of the operator service dies, when a new leader is elected the watchdog be triggered again - I don't think this should be an issue however.

pkg/operator/watchdog.go
ItalyPaleAle: Done. Stopped needs to be logged here: https://github.com/dapr/dapr/pull/4764/files#diff-76c380a58f117638bd5ab67cbcf2704d7e623b3d3651204a8fcccda177317ebaR66

pkg/operator/watchdog.go
artursouza: OK. I understand now. Thanks.

artursouza (approved):

pkg/operator/watchdog.go
```diff
@@ -0,0 +1,175 @@
...
+
+func (dw *DaprWatchdog) listPods(ctx context.Context) {
+	log.Infof("DaprWatchdog started checking pods")
+
+	// Request the list of pods
+	// We are not using pagination because we may be deleting pods during the iterations
+	// The client implements some level of caching anyways
+	pod := &corev1.PodList{}
+	err := dw.client.List(ctx, pod,
+		client.InNamespace(GetNamespace()),
```
gunniwho, line 126: @ItalyPaleAle why only check in the dapr namespace? 🤔

ItalyPaleAle: Hey good question! I didn't think checking all namespaces was needed but perhaps it is?

Problem would be if you have multiple namespaces with Dapr installed, in which case there may be a race condition.

Perhaps this needs to have another configuration option to pass the namespace(s) to check?

yaron2: There can only be a single Dapr installation in a cluster, but regardless this list operation should list pods in all namespaces.

ItalyPaleAle: Yes  :) https://github.com/dapr/dapr/pull/4837

PR #4773, opened 2022-06-14
Increase resources given to E2E apps

ItalyPaleAle:
Although this isn't **the** fix to all of our issues, it seems safe to merge this and can fix issues such as apps crashing (common with certain E2E test apps).


tests/e2e/actor_features/actor_features_test.go
```diff
@@ -142,10 +142,6 @@ func TestMain(m *testing.M) {
 			Replicas:       1,
 			IngressEnabled: true,
 			MetricsEnabled: true,
-			DaprCPULimit:   "2.0",
```
daixiang0, line 145: We need to track why old resources do not work and why it needs more resources.

PR #4776, opened 2022-06-14
Add distributed lock http1.1 endpoints

yaron2:
This PR adds a RESTFul HTTP1.1 interface for the Distributed Lock API.

Also, fixes typo in state error response.


pkg/http/api.go
```diff
@@ -1906,7 +2033,7 @@ func getMetadataFromRequest(reqCtx *fasthttp.RequestCtx) map[string]string {
 
 func (a *api) onPostStateTransaction(reqCtx *fasthttp.RequestCtx) {
 	if a.stateStores == nil || len(a.stateStores) == 0 {
-		msg := NewErrorResponse("ERR_STATE_STORES_NOT_CONFIGURED", messages.ErrStateStoresNotConfigured)
+		msg := NewErrorResponse("ERR_STATE_STORE_NOT_CONFIGURED", messages.ErrStateStoresNotConfigured)
```
artursouza, line 2036: Breaking change, not?

artursouza (changes requested):

pkg/http/api.go
artursouza: We might need to document this. Otherwise, LGTM.

yaron2: It's a breaking change in the response text, yes. So far we've considered an API change to be a breaking change if it included a modification for: 1. URL path 2. request/response object 3 .status code change.

If we consider the response text for errors a breaking change then we need to revert this change. Personally I recommend to change and document.

johnewart:
It looks like lock stores are not required to be transactional -- is this by design? I would think that we need some sort of atomic / transactional store to avoid race conditions between two concurrent lock requests, how do we expect to handle that? (I looked for the proposal but can't seem to find it, either my GitHub search-fu is bad or my memory banks are 😄  -- if it's in there or I am missing something then feel free to point me to TFM)

yaron2:
> It looks like lock stores are not required to be transactional -- is this by design? I would think that we need some sort of atomic / transactional store to avoid race conditions between two concurrent lock requests, how do we expect to handle that? (I looked for the proposal but can't seem to find it, either my GitHub search-fu is bad or my memory banks are 😄 -- if it's in there or I am missing something then feel free to point me to TFM)

No, they don't need to be transactional. It's an implementation detail for each component. The Redis lock store actually uses the Redis lock feature, for example.

pkg/http/api.go
artursouza: I know. Error codes don't have the same treatment as the rest of the API spec. We should rethink this at some point when we start designing the v2 API spec.

artursouza (approved):

PR #4783, opened 2022-06-15
Fix dead letter topic keeping message in original topic

yaron2:
This PR adds a missing return statement that caused messages dropped into the dead letter topic to be kept and retried on the original topic.

Also adds a minor change for code readability and fixes wrong metric from `Retry` to `Success` when the app returns `2xx`.

### DIFF
--- a/pkg/runtime/runtime.go
+++ b/pkg/runtime/runtime.go
@@ -690,6 +690,8 @@ func (a *DaprRuntime) beginPubSub(subscribeCtx context.Context, name string, ps 
 				if configured, _ := a.sendToDeadLetterIfConfigured(name, msg); !configured {
 					return err
 				}
+
+				return nil
 			}
 			return err
 		}); err != nil {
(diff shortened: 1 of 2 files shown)


halspang (commented):
Change makes sense but I see a lot of pubsub errors in the e2e.

yaron2:
> Change makes sense but I see a lot of pubsub errors in the e2e.

Yeah the e2e test needs to be rechecked.

yaron2:
@Taction I want to get your take on the e2e test case you wrote.

This PR fixes a bug where messages would still be kept on the original topic after an app returned an error and the topic was sent to dead letter queue. This means topics would get redelivered to the app from the original topic when they shouldn't have been.

I can confirm the correct behavior by:

1. Publishing 10 messages to topic X
2. Subscribing to topic X and returning an error, with having a dead letter topic Y configured
3. Confirm messages from topic X are not getting delievered
4. Consume 10 messages from dead letter topic Y
5. Confirm messages from dead letter topic Y are not getting redelivered

Looking at the e2e test logs above, it looks like the test was written to expect redelivery of messages sent to the dead letter topic. Can you confirm this needs to be changed?

Taction:
@yaron2 Thanks for your fix, I'll look into the e2e tests.

Taction:
@yaron2 The e2e topic check is changed, fixed now.

halspang (approved):

yaron2:
> @yaron2 The e2e topic check is changed, fixed now.

Thanks!

ItalyPaleAle:
@yaron2  @Taction  looks like after this PR was merged, E2E tests on AKS are now failing almost every time with errors due to PubSub.

ItalyPaleAle:
My attempt (hoping I understood the intent correctly): https://github.com/dapr/dapr/pull/4792

ItalyPaleAle:
I think I found the issue, it was in the tests. The messages were re-delivered in the DLQ a bit too slow and they came in too late. I increased the sleep times here: https://github.com/dapr/dapr/pull/4793

PR #4786, opened 2022-06-16
Reduce size of Azure Service Bus instance

ItalyPaleAle:
If this works, we can reduce our COGS quite a bit for E2E tests


ItalyPaleAle:
Looks like initial E2E test failures were due to the fact that pubsub tests that day were broken regardless. After the fix, looks like this works!

skyao (approved):

PR #4792, opened 2022-06-17
Should fix failures due to #4783

ItalyPaleAle:
See https://github.com/dapr/dapr/pull/4783#issuecomment-1158426778


yaron2:
@ItalyPaleAle this introduced a wrong behavior that's a breaking change. If there's no DLT and there's an error from the app, an error should be returned to retry the message, not drop it.

ItalyPaleAle:
Thanks for the review. Right, I misunderstood the logic. Closing this.

PR #4793, opened 2022-06-17
Should fix test failures due to #4783

ItalyPaleAle:
After further investigation, test failures were due to messages arriving a second or two too late. This increases the the sleep time so we give messages more time to be delivered.

### DIFF
--- a/pkg/runtime/runtime.go
+++ b/pkg/runtime/runtime.go
@@ -679,12 +679,9 @@ func (a *DaprRuntime) beginPubSub(subscribeCtx context.Context, name string, ps 
 					pubsub:     name,
 				})
 			})
-			log.Infof("Error processing message %v: %v", cloudEvent, err)
 			if err != nil && err != context.Canceled {
 				// Sending msg to dead letter queue, if no DLQ is configured, return error for backwards compatibility(component level retry).
-				configured, _ := a.sendToDeadLetterIfConfigured(name, msg)
-				log.Infof("Error processing message %v: %v (DLQ configured? %v)", cloudEvent, err, configured)
-				if !configured {
+				if configured, _ := a.sendToDeadLetterIfConfigured(name, msg); !configured {
 					return err
 				}
 
--- a/pkg/runtime/runtime.go
+++ b/pkg/runtime/runtime.go
@@ -679,9 +679,12 @@ func (a *DaprRuntime) beginPubSub(subscribeCtx context.Context, name string, ps 
 					pubsub:     name,
 				})
 			})
+			log.Infof("Error processing message %v: %v", cloudEvent, err)
 			if err != nil && err != context.Canceled {
 				// Sending msg to dead letter queue, if no DLQ is configured, return error for backwards compatibility(component level retry).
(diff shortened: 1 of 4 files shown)


yaron2:
@ItalyPaleAle looks like more messages are coming in, but not all of them make it in time.

ItalyPaleAle:
Uhmmmm the increase in the wait time was supposed to address that. But i see they are still failing. Taking a look at the new logs.

yaron2:
@ItalyPaleAle look at the KiND test failures. For the failing tests, it looks like most but not all messages arrived. It might be a combination of resiliency + longer wait time

ItalyPaleAle:
Uhm my last commit was supposed to fix both resiliency and increase timeouts :D but it's not working. I added a test commit with more logging (because I can't repro the issue when running the tests manually). Needs more investigation

ItalyPaleAle:
Ok I think I figured out the biggest issue (perhaps in addition to the prior ones).

The test was sending all messages and then invoking `setDesiredResponse(...."success"...)` right after. From the moment that invocation is made, all incoming messages are treated as successful. https://github.com/dapr/dapr/pull/4793/files#diff-bf3056b76a9a9696fdfd37479a4d118b75f2d6a7cbb52070ee99508a4b06bc27R274

The problem is that the last messages sent are those in the "dead topic", which are going to be put in the DLQ when failing. The messages are delivered to the receiver fairly quickly, but the last messages are delivered after the desired response is changed to `success`.

I mitigated that by:

1. Changing the order messages are published: the ones that go in the "dead topic" are now sent first, so we have more time to make sure they are processed before the desired response is changed (messages in the other topics never go in the DLQ and are just retried "indefinitely" - until TTL - so eventually end up in the "successful" queue).
2. Making the app sleep for 5s before changing the desired response to "success", just to give the queue a bit more time to deliver messages that need to be sent to the DLQ.

pkg/runtime/runtime.go
```diff
@@ -679,9 +679,12 @@ func (a *DaprRuntime) beginPubSub(subscribeCtx context.Context, name string, ps
 					pubsub:     name,
 				})
 			})
+			log.Infof("Error processing message %v: %v", cloudEvent, err)
```
artursouza, line 682: Remove log.

pkg/runtime/runtime.go
```diff
@@ -679,9 +679,12 @@ func (a *DaprRuntime) beginPubSub(subscribeCtx context.Context, name string, ps
...
 			if err != nil && err != context.Canceled {
 				// Sending msg to dead letter queue, if no DLQ is configured, return error for backwards compatibility(component level retry).
-				if configured, _ := a.sendToDeadLetterIfConfigured(name, msg); !configured {
+				configured, _ := a.sendToDeadLetterIfConfigured(name, msg)
+				log.Infof("Error processing message %v: %v (DLQ configured? %v)", cloudEvent, err, configured)
```
artursouza, line 686: remove log

artursouza (approved):

PR #4782, opened 2022-06-15
increase the checking value to let perf test continue to run

skyao:
Signed-off-by: Sky Ao <aoxiaojian@gmail.com>

# Description

Perf test failed sometimes because of the checking is not passed.

Increase the value and let perf test continue to run and let us collect more test data. 

In the issue I will record the problem and follow them.

## Issue reference

https://github.com/dapr/dapr/issues/4743

## Checklist

Please make sure you've  completed the relevant tasks for this PR, out of the following list:

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing
* [ ] End-to-end tests passing
* [ ] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_
* [ ] Specification has been updated / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_
* [ ] Provided sample for the feature / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_

PR #4802, opened 2022-06-20
Require perf test-er and test-ed pods to be on the same node

ItalyPaleAle:
This removes the need to get the traffic go over the network, which could be highly variable and add noise. Fixes #4782

Also reverts the previous performance targets added in #4744 to the prior values.

PS: I also added the ability to run a single perf test using the env var `DAPR_PERF_TEST`, just like there was the ability to run a single E2E test using `DAPR_E2E_TEST`


skyao (approved):
Let's try it on perf test workflow.

PR #4805, opened 2022-06-21
Correctly reporting perf test results

ItalyPaleAle:
Previously, was misreporting results so that the 99th percentile was repoted as 90th.

Also restored the original perf targets


skyao (approved):
Good fix!

artursouza (approved):

Issue #2836, opened 2021-02-22
Support trace exporting to Open Telemetry protocol


tcnghia:
## In what area(s)?

/area runtime
> /area operator
> /area placement
> /area docs
> /area test-and-release

## Describe the feature
Now that OpenTelemetry [reached 1.0](https://medium.com/opentelemetry/opentelemetry-specification-v1-0-0-tracing-edition-72dd08936978), we need to support OpenTelemetry protocol. (Currently OpenTelemetry is supported by translating Zipkin protocol with OpenTelemetry Collector).

## Release Note

RELEASE NOTE: **ADD** trace export using OpenTelemetry protocol

Note: Submit a Docs PR on this issue https://github.com/dapr/docs/issues/2405 at the time of code checkin

daixiang0:
Now Go SDK has reached 1.0 RC2, see https://github.com/open-telemetry/opentelemetry-go/releases/tag/v1.0.0-RC2

yaron2:
@juazasan can you work on this?

juazasan:
Hi Yaron,

I can work on it right after Kubecon. Does that work?

Thanks.

yaron2:
@juazasan sure thing, thanks!

juazasan:
@yaron2 I am looking at this but have some doubts about the approach for this issue. Wondering if you could please confirm the intend is to completely replace OpenCensus with OpenTelemetry and avoid introducing a breaking change by keep leveraging the Zipkin exporter.

Also, do we need to allow for other otel exporters to be used as part of the issue?

Thanks.

yaron2:
> @yaron2 I am looking at this but have some doubts about the approach for this issue. Wondering if you could please confirm the intend is to completely replace OpenCensus with OpenTelemetry and avoid introducing a breaking change by keep leveraging the Zipkin exporter.
> 
> Also, do we need to allow for other otel exporters to be used as part of the issue?
> 
> Thanks.

I confirm that's the intent. We want to introduce support for the OTLP exporter in addition to Zipkin.

juazasan:
[...]

PR #4781, opened 2022-06-15
Issue/2836

juazasan:
# Description

Changes to tracing libraries to leverage OpenTelemetry instead of Opencensus.
Added support for otel protocol in addition to zipkin to allow exporting traces to an opentelemetry agent using native protocol over both http and grpc.

## Issue reference

Please reference the issue this PR will close: #2836 

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests
* [X] Unit tests passing
* [X] End-to-end tests passing
* [ ] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/doc#2405

PR #4808, opened 2022-06-21
merged changes including opentelemetry tracing on behalf of opencensus

juazasan:
Signed-off-by: Juan Carlos Zamora <juazasan@msn.com>

# Description

Changes to tracing libraries to leverage OpenTelemetry instead of Opencensus.
Added support for otel protocol in addition to zipkin to allow exporting traces to an opentelemetry agent using native protocol over both http and grpc.

## Issue reference

Please reference the issue this PR will close: #2836 

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests
* [X] Unit tests passing
* [ ] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#2405


juazasan:
This PR replaces #4781

yaron2:
@juazasan this is awesome work, thank you. telemetry is one of Dapr's weaknesses in terms of test coverage, and we have to review and test this very carefully on multiple platforms to make sure there are no regressions here. considering that, we will push this to 1.9 and make sure this is merged into `master` branch as soon as possible after `1.8` is released.

juazasan:
> @juazasan this is awesome work, thank you. telemetry is one of Dapr's weaknesses in terms of test coverage, and we have to review and test this very carefully on multiple platforms to make sure there are no regressions here. considering that, we will push this to 1.9 and make sure this is merged into `master` branch as soon as possible after `1.8` is released.

@yaron2 I spent quite a lot of time running manual end to end tests for different scenarios and comparing the resulting traces with the same got with opencensus for the same scenarios. I also added some test cases to identify the breaks that I was seeing in those tests which were mostly due to not handling spans propagation properly.

That said. I understand your concern and decision of postponing this merge to a later release.

yaron2:
> 

@artursouza would like to get your take here as well.

yaron2:
@juazasan please resolve the conflict.

@artursouza ping.for review

pkg/channel/http/http_channel.go
```diff
@@ -222,8 +221,8 @@ func (h *Channel) constructRequest(ctx context.Context, req *invokev1.InvokeMeth
 
 	// HTTP client needs to inject traceparent header for proper tracing stack.
 	span := diag_utils.SpanFromContext(ctx)
-	httpFormat := &tracecontext.HTTPFormat{}
-	tp, ts := httpFormat.SpanContextToHeaders(span.SpanContext())
+	tp := diag.SpanContextToW3CString((*span).SpanContext())
```
ItalyPaleAle, line 224: Why are these `(*span)` necessary? They look... awkward. Could `diag_utils.SpanFromContext` return a pointer directly?

juazasan: Sure. Just for my knowledge: Is that the idiomatic way rather than create a struct to wrap the interface? Thanks

ItalyPaleAle: Yes, structs implement interfaces rather than wrapping them, usually. Take a look at this which seems to be well-written: https://medium.com/rungo/interfaces-in-go-ab1601159b3a 

An example is a readable stream.

```go
buf := bytes.Buffer{}
```

`buf`, whose type is `bytes.Buffer` implements some interfaces, including for example `io.Reader` (both `bytes.Buffer` and `*bytes.Buffer` would implement that interface).

So if you have a variable of type `io.Reader` or a method that accepts an `io.Reader`, you can pass that directly:

```go
// Assign buf to a variable of type io.Reader
var myreader io.Reader
myreader = buf

// Pass buf to a method that accepts an io.Reader
io.ReadAll(buf)
```

As for handling nils, variables that are of an interface type can always be nil, even when the type you're assigning to it is not a pointer.

```go
if myreader != nil { /* ... */ }
```

Hope this (and the article above) help :)

Although what you wrote is probably ok, the fact that it uses an uncommon syntax can be hard to follow and maintain in the future IMHO

yaron2: Ping @juazasan

yaron2:
@juazasan can you please provide a brief explanation on what a user would need to do to enable the OpenTelemetry native protocol? (in terms of user experience). I'd like to take this feature for a spin and test it.

pkg/channel/http/http_channel.go
juazasan: Thank @ItalyPaleAle for the example. However not sure they are representative of the scenario here. In your example bytes.Buffer is a struct that implements an interface and that is what you pass around, while the challenge here is that the Span creator from Opentelemetry returns objects of type interface not a struct. See: https://pkg.go.dev/go.opentelemetry.io/otel/trace#Tracer

juazasan:
> 

You only need to add a new trace exporter of type otel in the dapr configuration as in the example below. Please note that both grpc and http are supported as transport protocols.
```yaml
apiVersion: dapr.io/v1alpha1
kind: Configuration
metadata:
  name: daprConfig
spec:
  tracing:
    samplingRate: "1"
    stdout: true
    otel:
      endpointAddress: "localhost:4317"
      isSecure: false
      protocol: "grpc"
```

pkg/channel/http/http_channel.go
juazasan: Anyways, I will look at changing the functions signatures and pass the variables of type interface by value instead of by reference as you are recommending instead of creating my own struct that wraps the object returned by the creator functions.

ItalyPaleAle: @juazasan I have to admit that I'm a bit confused as to where the issue is in this case. Where in the code are you defining the struct that you need to pass around?

juazasan: I have not defined it for now. I would have defined it for instance in the diagnostic_utils package if we had agreed on that being the best option to get rid of the (*.span).

yaron2:
> > 
> 
> @yaron2
> 
> You only need to add a new trace exporter of type otel in the dapr configuration as in the example below. Please note that both grpc and http are supported as transport protocols.
> 
> ```yaml
> apiVersion: dapr.io/v1alpha1
> kind: Configuration
> metadata:
>   name: daprConfig
> spec:
>   tracing:
>     samplingRate: "1"
>     stdout: true
>     otel:
>       endpointAddress: "localhost:4317"
>       isSecure: false
>       protocol: "grpc"
> ```
> 
> If you want to keep using zipkin then no changes are required as definition is backwards compatible:
> 
> ```yaml
> apiVersion: dapr.io/v1alpha1
> kind: Configuration
> metadata:
>   name: daprConfig
> spec:
>   tracing:
>     samplingRate: "1"
>     stdout: true
>     zipkin:
>       endpointAddress: http://localhost:9411/api/v2/spans
>     otel:
>       endpointAddress: "localhost:4318"
>       isSecure: false
>       protocol: "http"
> ```

This is great @juazasan.

yaron2:
@juazasan see this: https://github.com/dapr/dapr/runs/7582389237?check_suite_focus=true.

ItalyPaleAle:
@yaron2 it's because of an issue with the CI that was fixed in release-1.8. Please see this: https://github.com/dapr/dapr/pull/4944

pkg/diagnostics/http_tracing.go
```diff
@@ -123,43 +119,25 @@ func isHealthzRequest(name string) bool {
...
-	if code == codes.Unknown {
-		if httpCode >= 100 && httpCode < 300 {
-			code = codes.OK
-		} else if httpCode >= 300 && httpCode < 400 {
-			code = codes.DeadlineExceeded
-		} else if httpCode >= 400 && httpCode < 500 {
-			code = codes.InvalidArgument
-		} else if httpCode >= 500 {
-			code = codes.Internal
+func traceStatusFromHTTPCode(httpCode int) (otelcodes.Code, string) {
```
artursouza, line 128: Is this a breaking change?

pkg/diagnostics/utils/trace_utils.go
```diff
@@ -54,29 +69,74 @@ func GetTraceSamplingRate(rate string) float64 {
...
+func BinaryFromSpanContext(sc trace.SpanContext) []byte {
+	traceID := sc.TraceID()
+	spanID := sc.SpanID()
+	traceFlags := sc.TraceFlags()
+	if sc.Equal(trace.SpanContext{}) {
+		return nil
+	}
+	var b [29]byte
+	copy(b[2:18], traceID[:])
+	b[18] = 1
```
artursouza, line 111: Where are these magic numbers coming from?

pkg/diagnostics/utils/trace_utils.go
```diff
@@ -54,29 +69,74 @@ func GetTraceSamplingRate(rate string) float64 {
...
+	b[27] = 2
+	b[28] = uint8(traceFlags)
+	return b[:]
+}
+
+// SpanContextFromBinary returns the SpanContext represented by b.
+//
+// If b has an unsupported version ID or contains no TraceID, SpanContextFromBinary returns with ok==false.
+func SpanContextFromBinary(b []byte) (sc trace.SpanContext, ok bool) {
+	var scConfig trace.SpanContextConfig
```
artursouza, line 122: Magic numbers? Is there a spec this is following?

artursouza (commented):
Can we have a E2E test that validates this new exporting protocol?

yaron2:
> Can we have a E2E test that validates this new exporting protocol?

@juazasan, as discussed with @artursouza we will accept this PR without additional E2E tests for now, and will follow up with them in a separate PR.

pkg/diagnostics/http_tracing.go
juazasan: It could be indeed a breaking change. I just tried to apply the logic referenced in the function documentation. See: https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/http.md#status
Not sure previous implementation was actually aligned with the opentelemetry convention.

pkg/channel/http/http_channel.go
juazasan: @ItalyPaleAle This issue should have been addressed with my last commit.

pkg/diagnostics/utils/trace_utils.go
juazasan: That comes from here: https://github.com/census-instrumentation/opencensus-go/blob/49838f207d61097fc0ebb8aeef306913388376ca/trace/propagation/propagation.go#L57

Previous tracing implementation used that function from opencensus. I had to create my own function to be able to remove all references to opencensus.

pkg/diagnostics/utils/trace_utils.go
juazasan: Same than previous comment. See: https://github.com/census-instrumentation/opencensus-go/blob/49838f207d61097fc0ebb8aeef306913388376ca/trace/propagation/propagation.go#L74

pkg/diagnostics/utils/trace_utils.go
ItalyPaleAle: It would be great if you could add a comment saying where that code comes from, and report its copyright. Something like:

```go
// Adapted from https://github.com/census-instrumentation/opencensus-go/blob/49838f207d61097fc0ebb8aeef306913388376ca/trace/propagation/propagation.go
// Copyright 2017 OpenCensus Authors
// License: https://github.com/census-instrumentation/opencensus-go/blob/49838f207d61097fc0ebb8aeef306913388376ca/LICENSE (Apache-2.0)
```

pkg/http/api.go
```diff
@@ -489,10 +489,10 @@ func (a *api) onOutputBindingMessage(reqCtx *fasthttp.RequestCtx) {
 			req.Metadata = map[string]string{}
 		}
 		// if sc is not empty context, set traceparent Header.
-		if sc != (trace.SpanContext{}) {
+		if !sc.Equal(trace.SpanContext{}) {
 			req.Metadata[traceparentHeader] = diag.SpanContextToW3CString(sc)
 		}
-		if sc.Tracestate != nil {
+		if sc.TraceState().Len() == 0 {
```
ItalyPaleAle, line 495: Should we add a nil-check for `sc.TraceState()`? Or is that guaranteed to be non-nil?

pkg/runtime/runtime.go
```diff
@@ -322,22 +330,64 @@ func (a *DaprRuntime) getOperatorClient() (operatorv1pb.OperatorClient, error) {
...
+			}
+			client = otlptracehttp.NewClient(clientOptions...)
+		} else {
+			clientOptions := []otlptracegrpc.Option{otlptracegrpc.WithEndpoint(endpoint)}
+			if !isSecure {
+				clientOptions = append(clientOptions, otlptracegrpc.WithInsecure())
+			}
+			client = otlptracegrpc.NewClient(clientOptions...)
+		}
+		otelExporter, err := otlptrace.New(context.Background(), client)
```
ItalyPaleAle, line 371: Should probably use `a.ctx` as context.

pkg/runtime/runtime.go
```diff
@@ -2191,6 +2241,9 @@ func (a *DaprRuntime) Shutdown(duration time.Duration) {
 			log.Warnf("error closing API: %v", err)
 		}
 	}
+	if a.tracerProvider != nil {
```
ItalyPaleAle, line 2244: This needs to go into a goroutine or it can delay the shutdown.


```go
shutdownCtx, shutdownCancel := context.WithCancel(context.Background())
go func() {
	if a.tracerProvider != nil {
		a.tracerProvider.Shutdown(shutdownCtx)
	}
}()

log.Infof("Waiting %s to finish outstanding operations", duration)
<-time.After(duration)
shutdownCancel()
```

pkg/runtime/trace.go
```diff
@@ -14,36 +14,106 @@ limitations under the License.
...
+	s.exporters = append(s.exporters, exporter)
+}
+
+// RegisterResource adds a Resource for registration with open telemetry global trace provider
+func (s *opentelemetryTracerProviderStore) RegisterResource(res *resource.Resource) {
+	s.res = res
+}
+
+// RegisterSampler adds a custom sampler for registration with open telemetry global trace provider
+
```
ItalyPaleAle, line 59: Please remove the newline here

pkg/runtime/trace.go
```diff
@@ -14,36 +14,106 @@ limitations under the License.
...
+	s.exporters = append(s.exporters, exporter)
+}
+
+// RegisterResource adds a Resource for registration with open telemetry global trace provider
+func (s *fakeTracerProviderStore) RegisterResource(res *resource.Resource) {
+	s.res = res
+}
+
+// RegisterSampler adds a custom sampler for registration with open telemetry global trace provider
+
```
ItalyPaleAle, line 113: Here too

yaron2:
There's an additional missing piece for this PR, and that's to update the [Configuration API type](https://github.com/dapr/dapr/blob/master/pkg/apis/configuration/v1alpha1/types.go) and regenerate the [Configuration CRD schema](https://github.com/dapr/dapr/blob/master/charts/dapr/crds/configuration.yaml). This is to enable the new `otel` section on Kubernetes.

Due to the size of this PR and aided by the fact we have a backlog of currently blocked PRs, I'll merge this now and assign myself and @juazasan for the follow ups.

@juazasan, please see remaining comments above and submit new PR(s) if further code changes are required.

yaron2 (approved):

yaron2:
Verified linux e2e test failure failed due to known and repeated failure.

juazasan:
@msfussell Sure. However, I cannot commit to having that completed before the 12th of September. Also, please be aware I already had this other issue assigned: https://github.com/dapr/docs/issues/2405

pkg/http/api.go
juazasan: @ItalyPaleAle sc.TraceState() returns a struct which as far as I know it cannot be nul. Lenght is the way I figure out to check whether tracestate is empty. Hope that makes sense.

PR #4820, opened 2022-06-22
Replace github.com/golang/protobuf with google.golang.org/protobuf

### LINKED ISSUE #4813: go get warning: module github.com/golang/protobuf is deprecated
When running `go get -v ./...`, the following appears in the console:

```
go: module github.com/golang/protobuf is deprecated: Use the "google.golang.org/protobuf" module instead.
```

Need to update the protobuf package.

ItalyPaleAle:
Fixes #4813

Module `github.com/golang/protobuf` is deprecated and needs to be replaced with `google.golang.org/protobuf`.

Note that while `github.com/golang/protobuf` is now fully-removed from dapr/dapr, it's still an indirect dependency for now.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -16,7 +16,6 @@ require (
 	github.com/ghodss/yaml v1.0.0
 	github.com/go-logr/logr v1.2.3
 	github.com/gogo/protobuf v1.3.2
-	github.com/golang/protobuf v1.5.2
 	github.com/google/cel-go v0.9.0
 	github.com/google/go-cmp v0.5.8
 	github.com/google/uuid v1.3.0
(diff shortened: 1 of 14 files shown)


go.mod
```diff
@@ -368,6 +368,7 @@ require (
 )
 
 replace (
+	github.com/trusch/grpc-proxy => github.com/ItalyPaleAle/grpc-proxy v0.0.0-20220622194522-bb83906fa0cb
```
ItalyPaleAle, line 371: Repo is un-maintained and had to fork this

daixiang0: I recommend put the repo under dapr org.

yaron2: That's a good call out @daixiang0.

Please remove this line, due to:

1. Code in `trusch/grpc-proxy` used in Dapr is used for tests only and isn't compiled in `daprd`
2. Using unmaintained repositories is allowed and this ain't the only one

ItalyPaleAle: Just to clarify, the reason for the fork wasn't that the repo was un-maintained.

The code (which was used for tests only, but was used) did not build anymore after updating protobuf. I had to make [changes](https://github.com/ItalyPaleAle/grpc-proxy/commit/bb83906fa0cb7db6d39a2f36daa1cee3c819ea16), and couldn't open a PR upstream because the repo is unmaintained.

Happy to move the repo to the Dapr org or even just copy/paste the code we need in this repo, it that's better.

yaron2: We have most of the code from that repository copied directly into `dapr/dapr` under `pkg/grpc/proxy`, so the correct thing to do is move the test code also to `dapr/dapr`.

ItalyPaleAle: I have copied the remaining files into this repo.

daixiang0 (approved):
lgtm

PR #4821, opened 2022-06-22
Disable TCP keepalives and HTTP/2 in injector's server

ItalyPaleAle:
This was an experimental fix to see if it could help reducing the rate of issues with the injector not receiving the webhooks.

Sadly, it doesn't look like it (completely) fixed the issue (although it _may_ have made it a bit better, as in 5 couples of tests–5 windows and 5 linux–there was only 1 injector failure). Runs: [1](https://github.com/ItalyPaleAle/dapr-ci-test/actions/runs/2543651192) [2](https://github.com/ItalyPaleAle/dapr-ci-test/actions/runs/2543681952) [3](https://github.com/ItalyPaleAle/dapr-ci-test/actions/runs/2543682488) [4](https://github.com/ItalyPaleAle/dapr-ci-test/actions/runs/2543939476) [5](https://github.com/ItalyPaleAle/dapr-ci-test/actions/runs/2543751067) (has one injector failure on Linux)

Nevertheless, this is probably a good change to merge anyways. The injector is responding to webhook requests and there's no reason why it needs to use HTTP/2 or why clients need to maintain a persistent TCP connection to it.

### DIFF
--- a/pkg/injector/injector.go
+++ b/pkg/injector/injector.go
@@ -15,6 +15,7 @@ package injector
 
 import (
 	"context"
+	"crypto/tls"
 	"encoding/json"
 	"fmt"
 	"io"
@@ -102,15 +103,20 @@ func getAppIDFromRequest(req *v1.AdmissionRequest) string {
 func NewInjector(authUIDs []string, config Config, daprClient scheme.Interface, kubeClient kubernetes.Interface) Injector {
 	mux := http.NewServeMux()
 
+	server := &http.Server{
+		Addr:    fmt.Sprintf(":%d", port),
+		Handler: mux,
+		// Disable HTTP/2
+		TLSNextProto: map[string]func(*http.Server, *tls.Conn, http.Handler){},
+	}
+	server.SetKeepAlivesEnabled(false)
+
 	i := &injector{
 		config: config,
 		deserializer: serializer.NewCodecFactory(
 			runtime.NewScheme(),
 		).UniversalDeserializer(),
-		server: &http.Server{
-			Addr:    fmt.Sprintf(":%d", port),
-			Handler: mux,
-		},
+		server:     server,
 		kubeClient: kubeClient,
 		daprClient: daprClient,
 		authUIDs:   authUIDs,


pkg/injector/injector.go
```diff
@@ -102,15 +103,20 @@ func getAppIDFromRequest(req *v1.AdmissionRequest) string {
 func NewInjector(authUIDs []string, config Config, daprClient scheme.Interface, kubeClient kubernetes.Interface) Injector {
 	mux := http.NewServeMux()
 
+	server := &http.Server{
+		Addr:    fmt.Sprintf(":%d", port),
+		Handler: mux,
+		// Disable HTTP/2
+		TLSNextProto: map[string]func(*http.Server, *tls.Conn, http.Handler){},
+	}
+	server.SetKeepAlivesEnabled(false)
```
yaron2, line 112: This will reduce throughput by a very wide margin, which might ben an issue for large deployments.

ItalyPaleAle: We are seeing issues with TCP connections being closed hence the proposed change.

I am not too concerned with reducing throughput here, as this is only responding to admission webhooks.  If this does help, I think the improvement in reliability should overcome the very small performance impact.

**First**
On large deployments, there will be multiple instances of sidecar-injector and multiple instances of the K8s control plane so TCP connection reuse shouldn't be common anyways.

Let's assume for simplicity that there is only 1 K8s control plane instance, so at least the caller is single. If there are 3 instances of sidecar-injector, injection webhooks will be sent to all 3 of them in a round-robin fashion. So you have 3 TCP connections being established there.

In Go, default TCP keepalive is 15 seconds, so to be able to have the same TCP connection re-used in this case you need to deploy > 3 pods per minute (see point 2 below).

If there is more than 1 K8s control plane instance, then there's even more round-robin happening.

**Second**
When you think about everything that's happening to deploy a pod in K8s, the time spent on our webhook is likely a relatively less important part of the entire process. A slight delay in this step should not cause any noticeable impact on the performance of the entire cluster IMHO.

yaron2: > In Go, default TCP keepalive is 15 seconds, so to be able to have the same TCP connection re-used in this case you need to deploy > 3 pods per minute (see point 2 below).

Common use cases that we know of today deploy 100s of pods < 1 second, whether it's for rolling updates of the apps, rolling updates after a Dapr upgrade, Kubernetes node upgrades, downtimes etc. In these situations this would incur a significant performance penalty.

> When you think about everything that's happening to deploy a pod in K8s, the time spent on our webhook is likely a relatively less important part of the entire process. A slight delay in this step should not cause any noticeable impact on the performance of the entire cluster IMHO.

Bottlenecks on an admission webhook == slower rollouts == potential downtime / business impact. So while there would also be impact on the performance of the cluster (the API server is waiting on I/O requests), that's not as critical as delaying the rollout time of pods, which is a side effect of disabling keep-alive at scale.

ItalyPaleAle: I have benchmarked it (within K8s).

See details of tests with `ab` with 10,000 requests and 200 concurrent.

<detail>
<summary>With keepalive (`-k`)</summary>
```
# ab -k -n 10000 -c 200 https://dapr-sidecar-injector.dapr-tests.svc.cluster.local/mutate
This is ApacheBench, Version 2.3 <$Revision: 1879490 $>
Copyright 1996 Adam Twiss, Zeus Technology Ltd, http://www.zeustech.net/
Licensed to The Apache Software Foundation, http://www.apache.org/

Benchmarking dapr-sidecar-injector.dapr-tests.svc.cluster.local (be patient)
Completed 1000 requests
Completed 2000 requests
Completed 3000 requests
Completed 4000 requests
Completed 5000 requests
Completed 6000 requests
Completed 7000 requests
Completed 8000 requests
Completed 9000 requests
Completed 10000 requests
Finished 10000 requests


Server Software:        
Server Hostname:        dapr-sidecar-injector.dapr-tests.svc.cluster.local
Server Port:            443
SSL/TLS Protocol:       TLSv1.2,ECDHE-RSA-AES128-GCM-SHA256,2048,128
Server Temp Key:        X25519 253 bits
TLS Server Name:        dapr-sidecar-injector.dapr-tests.svc.cluster.local

Document Path:          /mutate
Document Length:        11 bytes

Concurrency Level:      200
Time taken for tests:   0.582 seconds
Complete requests:      10000
Failed requests:        0
Non-2xx responses:      10000
Keep-Alive requests:    10000
Total transferred:      1940000 bytes
HTML transferred:       110000 bytes
Requests per second:    17169.21 [#/sec] (mean)
Time per request:       11.649 [ms] (mean)
Time per request:       0.058 [ms] (mean, across all concurrent requests)
Transfer rate:          3252.76 [Kbytes/sec] received

Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        0    3  22.5      0     269
Processing:     0    9  13.5      5     224
Waiting:        0    9  13.5      5     224
Total:          0   11  26.5      6     277

Percentage of the requests served within a certain time (ms)
  50%      6
  66%      8
  75%     10
  80%     12
  90%     20
  95%     30
  98%     90
  99%    168
 100%    277 (longest request)
```
</details>

<details>
<summary>Without keepalive</summary>
```
# ab -n 10000 -c 200 https://dapr-sidecar-injector.dapr-tests.svc.cluster.local/mutate
This is ApacheBench, Version 2.3 <$Revision: 1879490 $>
Copyright 1996 Adam Twiss, Zeus Technology Ltd, http://www.zeustech.net/
Licensed to The Apache Software Foundation, http://www.apache.org/

Benchmarking dapr-sidecar-injector.dapr-tests.svc.cluster.local (be patient)
Completed 1000 requests
Completed 2000 requests
Completed 3000 requests
Completed 4000 requests
Completed 5000 requests
Completed 6000 requests
Completed 7000 requests
Completed 8000 requests
Completed 9000 requests
Completed 10000 requests
Finished 10000 requests


Server Software:        
Server Hostname:        dapr-sidecar-injector.dapr-tests.svc.cluster.local
Server Port:            443
SSL/TLS Protocol:       TLSv1.2,ECDHE-RSA-AES128-GCM-SHA256,2048,128
Server Temp Key:        X25519 253 bits
TLS Server Name:        dapr-sidecar-injector.dapr-tests.svc.cluster.local

Document Path:          /mutate
Document Length:        11 bytes

Concurrency Level:      200
Time taken for tests:   7.916 seconds
Complete requests:      10000
Failed requests:        0
Non-2xx responses:      10000
Total transferred:      1700000 bytes
HTML transferred:       110000 bytes
Requests per second:    1263.31 [#/sec] (mean)
Time per request:       158.314 [ms] (mean)
Time per request:       0.792 [ms] (mean, across all concurrent requests)
Transfer rate:          209.73 [Kbytes/sec] received

Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        3  120  70.3    108     552
Processing:     0   35  41.2     21     411
Waiting:        0   25  30.6     14     338
Total:          4  155  81.5    144     658

Percentage of the requests served within a certain time (ms)
  50%    144
  66%    175
  75%    201
  80%    217
  90%    255
  95%    298
  98%    374
  99%    420
 100%    658 (longest request)
```
</details>

You're right that it's much slower. In particular, with keepalives the time spent connecting is essentially zero, so the mean time is 0.058ms per request. Without keepalives, it's 0.792ms per request, so 13.6x slower.

Although the slowdown is significant in relative terms, we're still sub-ms, and this is at 200 concurrent requests.

If the assumption I have is correct, which is that keepalives are indeed causing reliability issues, I think the downtime/business impact caused by pods starting without sidecar is bigger than this iMHO.

yaron2: > I think the downtime/business impact caused by pods starting without sidecar is bigger than this iMHO.

That may be true, but before making this trade-off I want us to consider alternatives to disabling keep-alive.

PR #4824, opened 2022-06-22
Changing AKS control plane SKU

ItalyPaleAle:
This seems to actually fix all issues with sidecar injectors randomly failing on AKS

Since this was implemented in my fork (where the watchdog is disabled), no test has failed with sidecar injection issues: https://github.com/ItalyPaleAle/dapr-ci-test/actions/workflows/dapr-test.yml (start from run 112) - the two failures are unrelated.

### DIFF
--- a/tests/test-infra/azure-aks.bicep
+++ b/tests/test-infra/azure-aks.bicep
@@ -156,7 +156,7 @@ resource aks 'Microsoft.ContainerService/managedClusters@2021-07-01' = {
   tags: {}
   sku: {
     name: 'Basic'
-    tier: 'Free'
+    tier: 'Paid'
   }
   identity: {
     type: 'SystemAssigned'


tests/test-infra/azure-aks.bicep
```diff
@@ -156,7 +156,7 @@ resource aks 'Microsoft.ContainerService/managedClusters@2021-07-01' = {
   tags: {}
   sku: {
     name: 'Basic'
-    tier: 'Free'
+    tier: 'Paid'
```
daixiang0, line 159: I am curious about why it caused failures.

mukundansundar (approved):

tests/test-infra/azure-aks.bicep
ItalyPaleAle: I am curious too!!

First, this is NOT a Dapr issue. The issue is with the K8s control plane.

I spent about 2 weeks debugging this with the AKS team and this was the result.

At a high level, the "paid" control plane has better HA (control plane resources are replicated). More relevant to us, in the "free" tier the nodes connect to the control plane via a  sort of P2P VPN which we've seen sometimes can have glitches. In the "paid" tier, the VPN is different.

daixiang0: That sounds reasonable to me, the network issue caused failures.

PR #4829, opened 2022-06-23
Fix resiliency logic in gRPC proxying

halspang:
# Description

In the case of gRPC proxying, we need to record all the calls that
have occurred and replay them if we need to perform retries. The
array of calls was being shared between goroutines which was causing
different calls to share frames. This causes requests to exhibit
incorrect behavior/fail. The fix for this is to isolate the call
array.

Additionally, this commit also fixes an issue with headers being
sent during stream calls. When a stream fails mid-stream, it
attempted to send the initial headers again. This hung forever as
the listener was no longer expecting headers.

https://github.com/dapr/dapr/issues/4537

## Issue reference

Please reference the issue this PR will close: #4537 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Unit tests passing

### DIFF
--- a/pkg/grpc/proxy/handler.go
+++ b/pkg/grpc/proxy/handler.go
@@ -6,12 +6,15 @@ package proxy
 import (
 	"fmt"
 	"io"
+	"sync"
 
 	"golang.org/x/net/context"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/metadata"
 	"google.golang.org/grpc/status"
+
+	"github.com/google/uuid"
 
 	"github.com/dapr/dapr/pkg/diagnostics"
 	"github.com/dapr/dapr/pkg/grpc/proxy/codec"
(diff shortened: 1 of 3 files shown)


pkg/grpc/proxy/handler.go
```diff
@@ -186,26 +199,27 @@ func (s *handler) forwardClientToServer(src grpc.ClientStream, dst grpc.ServerSt
...
 			}
 		}
 		for i := 0; ; i++ {
 			if err := src.RecvMsg(f); err != nil {
+				s.bufferedCalls.Store(requestID, bufferedFrames)
 				ret <- err // this can be io.EOF which is happy case
 				break
 			}
-			s.bufferedCalls = append(s.bufferedCalls, f)
+			bufferedFrames = append(bufferedFrames, f)
```
pkedy, line 220: Is there a concurrency issue here if >1 goroutine is accessing the underlying buffer?

halspang: It's a local variable in the goroutine which is referenced out of the sync map by a uuid. For two goroutines to access the same underlying buffer, we'd need to kick off two streams from the same request at the same time, which I don't believe is possible. 

https://github.com/dapr/dapr/pull/4829/files#diff-b4595741c39d4944ac576f1228c698c0871002bfb2d2bbf6eab939aa66dac7feR207

yaron2: There should likely be a lock here per request ID

yaron2: > It's a local variable in the goroutine which is referenced out of the sync map by a uuid. For two goroutines to access the same underlying buffer, we'd need to kick off two streams from the same request at the same time, which I don't believe is possible.
> 
> https://github.com/dapr/dapr/pull/4829/files#diff-b4595741c39d4944ac576f1228c698c0871002bfb2d2bbf6eab939aa66dac7feR207

That assumption should hold true.

halspang: @yaron2 - If we have multiple goroutines from the same request in here, wouldn't we also be sending/receiving incorrectly because we'd have multiple listeners on the same stream? Or is this a "just for safety's sake" thing?

yaron2 (dismissed):

yaron2:
@halspang see failures

yaron2 (approved):

PR #4836, opened 2022-06-27
Watchdog: list pods from all namespaces

yaron2:
Currently the watch dog only lists pods from the namespace of the Dapr control plane, while it should list pods from all namespaces.

This PR changes the pod list operation to use all namespaces.

cc @ItalyPaleAle


yaron2:
Closing this as @ItalyPaleAle is working on the same. Can re-open if needed.

Issue #501, opened 2019-10-01
Support pubsub for Actors


amanbha:
## Describe the proposal
Support pubsub between actors, currently the publisher can be a service or an actor. But when it comes to subscribing, the subscribe is supported at service level. We need to extend this to support subscribe at an Actor level. An actor should be able to subscribe to a message of type "Message_One" with its actor id and when that message is published a specific method is invoked on that particular actor id. If multiple actor ids have subscribed for that message then method will be invoked for all the subscribed actor ids.

mchmarny:
Given the unique need for some kind of mapping between inbound event and actor (based on the event type and and some kind of logic to derive ID to create/activate actor), I wonder if this possibly better handled in new kind of binding that can be configured with ref to pubsub binding name as a "source" in metadata. Trying to handle this in PubSub construct could potentially add complexity to the simple pubsub construct we support now.

amanbha:
@mchmarny  The implementation for ACtors pubsub will built on top of Dapr's pubsub so the description is incorrect. I will modify it keep and this issue for bindings only.
@LMWF Could you open another issue for Actors pubsub?

amanbha:
@mchmarny @LMWF With the proposal for Actors pubsub, it will not require logic to determine id from the event. The Actors layer in dapr runtime will save that information and will build on top of Dapr's pubsub building block.

yaron2:
> @mchmarny @LMWF With the proposal for Actors pubsub, it will not require logic to determine id from the event. The Actors layer in dapr runtime will save that information and will build on top of Dapr's pubsub building block.

Yeah, but this proposal assumes the actor can first subscribe to events only after it gets activated via service invocation.

@mchmarny is talking about the scenario where event come in (either through bindings or Pub/Sub) that need to trigger an actor irrespective of it being created earlier.

amanbha:
Yes, these are 2 scenarios.

LMWF:
Note #1601 above was forked for bindings.

artursouza:
I think there is only one scenario: subscription for ActorType+ActorId. The other scenarios are a specific scenario of this one. The reason why is that Dapr does not track which actors exist since they are virtual actors. The fact that an actor is active does not mean it only exists in that period of time. Virtual actors "always" exist.

[...]

PR #4844, opened 2022-06-29
Add Pubsub for Actors

DiegoAJR:
# Description

Support Pubsub for Actors with new API endpoint and modified runtime to invoke actors' method.
Added API enpoint to publish for an specific actor type and id.
Added into Actors runtime configuration field to declare subscriptions
Modified runtime to handle subscriptions to topics.
Created e2e test: Publishing for an actor and invoking the right method. 

## Issue reference

WIP #501

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Unit tests passing
* [X] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_2735_
* [X] Specification has been updated / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_2735_
* [X] Provided sample for the feature / Created issue in the https://github.com/dapr/python-sdk/ repo: dapr/python-sdk#_438_


pkg/grpc/api.go
```diff
@@ -365,6 +366,112 @@ func (a *api) CallActor(ctx context.Context, in *internalv1pb.InternalInvokeRequ
 	return resp.Proto(), nil
 }
 
+func (a *api) PublishActorEvent(ctx context.Context, in *runtimev1pb.PublishActorEventRequest) (*emptypb.Empty, error) {
+	if a.pubsubAdapter == nil {
```
berndverst, line 370: @halspang can you think of a way to avoid code duplication here?

```
publishEventInternal(ctx context.Context, in interface{})
{

    switch in.(type) {
        case *runtimev1pb.PublishActorEventRequest: {
          // Actor publish path
        }
        case *runtimev1pb.PublishEventRequest: {
          // Regular publish path
        }
    }
}
```

or some generic way?

Essentially we should have the same code somehow for everything other than the addition of cloudevent fields.

halspang: Given that everything is the same except for the two new fields, I'd think you could just use a single common method. In that method either choose to add the fields if they're defined or, just add them anyway as they'll be null in the message. 

That last approach does add two attributes to all requests though, which I'm not sure everyone would be happy about so I'd go with the first approach.

berndverst:
@halspang can you run the tests again?

pkg/http/api.go
```diff
@@ -369,6 +369,12 @@ func (a *api) constructActorEndpoints() []Endpoint {
 			Version: apiVersionV1,
 			Handler: a.onRenameActorReminder,
 		},
+		{
+			Methods: []string{fasthttp.MethodPost, fasthttp.MethodPut},
+			Route:   "actors/{actorType}/{actorId}/publish/{pubsubname}/{topic:*}",
+			Version: apiVersionV1,
```
berndverst, line 375: this needs to be Alpha1

yaron2:
@berndverst please re-review.

dapr/proto/runtime/v1/dapr.proto
```diff
@@ -53,6 +53,9 @@ service Dapr {
   // Publishes events to the specific topic.
   rpc PublishEvent(PublishEventRequest) returns (google.protobuf.Empty) {}
 
+  // PublishActorEventAlpha1 Publishes an event from an actor with a specific topic.
+  rpc PublishActorEventAlpha1(PublishActorEventRequest) returns (google.protobuf.Empty) {}
```
artursouza, line 57: Please, move this close to the other Actor methods. On the client side, this belongs to the ActorProxy IMO.

pkg/grpc/api.go
```diff
@@ -73,6 +73,7 @@ type API interface {
 
 	// Dapr Service methods
 	PublishEvent(ctx context.Context, in *runtimev1pb.PublishEventRequest) (*emptypb.Empty, error)
+	PublishActorEventAlpha1(ctx context.Context, in *runtimev1pb.PublishActorEventRequest) (*emptypb.Empty, error)
```
artursouza, line 76: nit: move close to the other Actor methods.

pkg/http/api.go
```diff
@@ -1879,11 +1885,20 @@ func (a *api) onShutdown(reqCtx *fasthttp.RequestCtx) {
 }
 
 func (a *api) onPublish(reqCtx *fasthttp.RequestCtx) {
+	a.onPubSubPublish(reqCtx, "", "")
+}
+
+func (a *api) onActorPublishAlpha1(reqCtx *fasthttp.RequestCtx) {
```
artursouza, line 1891: nit: just onActorPublish is enough since internal method name can change freely.

pkg/http/api.go
```diff
@@ -1879,11 +1885,20 @@ func (a *api) onShutdown(reqCtx *fasthttp.RequestCtx) {
...
+	actortype := reqCtx.UserValue(actorTypeParam).(string)
```
artursouza, line 1892: nit: Maybe we can keep the original method and check for the presence of actorTypeParam and actorIdParam.

pkg/runtime/pubsub/cloudevents.go
```diff
@@ -29,13 +29,15 @@ type CloudEvent struct {
 	DataContentType string
 	TraceID         string
 	TraceState      string
+	ActorType       string
```
artursouza, line 32: Do we want to rename it to "DaprActorType" and "DaprActorId" to avoid conflict with existing names applications might be using?

pkg/runtime/runtime.go
```diff
@@ -573,14 +573,9 @@ func (a *DaprRuntime) beginPubSub(subscribeCtx context.Context, name string, ps
 	case GRPCProtocol:
 		publishFunc = a.publishMessageGRPC
 	}
-	topicRoutes, err := a.getTopicRoutes()
```
artursouza, line 576: Why did you remove the error handling here?

pkg/runtime/runtime.go
```diff
@@ -695,6 +690,110 @@ func (a *DaprRuntime) beginPubSub(subscribeCtx context.Context, name string, ps
 		}
 	}
 
+	for _, actorConfigSubscriptions := range a.appConfig.Pubsub {
+		if actorConfigSubscriptions.PubSubName == name {
+			allowed := a.isPubSubOperationAllowed(name, actorConfigSubscriptions.Topic, a.scopedSubscriptions[name])
```
artursouza, line 695: I wonder if we can reuse the subscription code above. It is odd that we need to implement DLQ and TTL again here.

pkg/runtime/runtime.go
```diff
@@ -695,6 +690,110 @@ func (a *DaprRuntime) beginPubSub(subscribeCtx context.Context, name string, ps
...
+	return nil
+}
+
+// actorCallPubsub calls the actor that was declared in the cloudevent or in the app.Config
+func (a *DaprRuntime) callActorPubsub(ctx context.Context, cloudEvent map[string]interface{}, actorPubSubSubscription config.PubSubConfig) error {
+	sendData, _ := json.Marshal(cloudEvent[pubsub.DataField])
+	if cloudEvent[pubsub.ActorIdField] == "" {
+		// Actor id is not in the cloudevent
+		if actorPubSubSubscription.ActorIDDataAttribute == "" {
+			log.Debug("Actor Pubsub: ActorID and ActorIDDataAttribute not declared in cloud event id - %s", cloudEvent[pubsub.IDField])
```
artursouza, line 755: Log Warning here. Users probably want to know the event was ignored.

Should these go to the DLQ instead?

pkg/runtime/runtime.go
```diff
@@ -695,6 +690,110 @@ func (a *DaprRuntime) beginPubSub(subscribeCtx context.Context, name string, ps
...
+			if attribute, ok := dataMap[actorPubSubSubscription.ActorIDDataAttribute]; ok {
+				req := invokev1.NewInvokeMethodRequest(actorPubSubSubscription.Method)
+				req.WithActor(actorPubSubSubscription.ActorType, attribute)
+				req.WithHTTPExtension(nethttp.MethodPost, "")
+				req.WithRawData(sendData, contenttype.CloudEventContentType)
+				log.Infof("Invoking Actor %+v: with method %s", req.Actor(), req.Message().GetMethod())
+
+				res, err := a.actor.Call(ctx, req)
+				if err != nil {
+					log.Debugf("Error invoking actor: %s", err.Error())
```
artursouza, line 773: We probably want an log error here?

pkg/runtime/runtime.go
```diff
@@ -695,6 +690,110 @@ func (a *DaprRuntime) beginPubSub(subscribeCtx context.Context, name string, ps
...
+					return err
+				}
+				log.Debugf("Actor response: %v", res.Message().GetData())
+			} else {
+				log.Debug("Actor Pubsub: Not a valid ActorIDAttribute for this actor")
```
artursouza, line 778: Log warning? DLQ?

pkg/runtime/runtime.go
```diff
@@ -695,6 +690,110 @@ func (a *DaprRuntime) beginPubSub(subscribeCtx context.Context, name string, ps
...
+			}
+		}
+	} else {
+		// actor id is in the cloud event
+		actorid := fmt.Sprint(cloudEvent[pubsub.ActorIdField])
+		req := invokev1.NewInvokeMethodRequest(actorPubSubSubscription.Method)
```
artursouza, line 784: This code block seems to be very similar to the invocation above. Can this be refactored so we reuse the actor invocation logic?

tests/e2e/actor_pubsub/actor_pubsub_test.go
```diff
@@ -0,0 +1,357 @@
...
+	"github.com/stretchr/testify/require"
+	"go.uber.org/ratelimit"
+)
+
+var tr *runner.TestRunner
+
+const (
+	numberOfMessagesToPublish = 60
+	publishRateLimitRPS       = 25
+	defaultActorType          = "myActorType"
```
artursouza, line 43: Please, use an actor type that is specific to this test to avoid conflicting with other test apps.

tests/perf/actor_pubsub/actor_pubsub_test.go
```diff
@@ -0,0 +1,185 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+const (
+	numHealthChecks        = 60 // Number of times to check for endpoint health per app.
+	serviceApplicationName = "perf-actor-subscriber"
+	clientApplicationName  = "perf-actor-publisher"
+	actorType              = "testactorpubsub"
+	defaultpubsubName      = "messagebus"
+	defaultTopicNameHTTP   = "mytopic"
```
artursouza, line 39: pick something more unique.

artursouza (changes requested):

pkg/runtime/pubsub/cloudevents.go
DiegoAJR: We could rename them. Will require update in the components-contrib/pubsub/envelope.go file

pkg/grpc/api.go
```diff
@@ -412,21 +443,23 @@ func (a *api) PublishEvent(ctx context.Context, in *runtimev1pb.PublishEventRequ
...
-			DataContentType: in.DataContentType,
+			Topic:           in.topic,
+			DataContentType: in.datacontectType,
 			Data:            body,
 			TraceID:         corID,
 			TraceState:      traceState,
-			Pubsub:          in.PubsubName,
+			Pubsub:          in.pubsubName,
+			ActorType:       actorType,
+			ActorID:         actorID,
```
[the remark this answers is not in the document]
DiegoAJR: We could declare a different cloud event structure. Update in components-contrib

andrewhathaway:
@DiegoAJR @halspang @artursouza Hi all, appreciate this PR has not changed in the last 67 days however this is still a feature we would still greatly benefit from and are having to work around it in the meantime. Is this functionality still something we can expect to be added in the future? :)

rotcafer:
@andrewhathaway can I ask how you are working around this now?

paule96:
@DiegoAJR, @andrewhathaway, @rotcafer, @halspang, @artursouza 

would it make sense to take these changes and continue working on it, on an updatodate branch?
I could maybe invest my next week to work on this. But I'm not a Go programmer so probably my progress wouldn't be that fast.

But having pub/sub in actors would be so nice. As far as I understand this PR was almost done, just some minor adjustments were missing.

What are your opinions on it?

artursouza:
This PR should not be merged. First, we should have a proposal to review some design decisions. If you want to take action in creating the proposal, see the Dapr/proposals repo.

PR #4846, opened 2022-06-29
Add context support for input bindings + shutdown output bindings before input ones

ItalyPaleAle:
Updated input bindings APIs to add support for contexts. This allows shutting down input components via context cancellation, before output bindings are shut down at the end of the graceful shutdown sequence. See also work done for PubSub: https://github.com/dapr/dapr/pull/4624


go.mod
```diff
@@ -374,6 +374,8 @@ replace (
 
 replace github.com/eclipse/paho.mqtt.golang => github.com/shivamkm07/paho.mqtt.golang v1.3.6-0.20220106130409-e28a1db639f8
 
+replace github.com/dapr/components-contrib => github.com/italypaleale/dapr-components-contrib v1.4.0-rc1.0.20220629203416-07ba3de80dc8
```
ItalyPaleAle, line 377: This is temporary until the PR is merged in components-contrib

artursouza (approved):

Issue #4830, opened 2022-06-24
Expose runtime version in Metadata API


greenie-msft:
The feature ask - expose Dapr runtime version in the Metadata API: https://docs.dapr.io/reference/api/metadata_api/

There are certain scenarios where apps would like to check that they are running within a tested Dapr version.

PR #4862, opened 2022-07-05
Add dapr runtime version to metadata

addjuarez:
Signed-off-by: Addison Juarez <adjuarez@microsoft.com>

# Description

Added the dapr runtime version to the extended metadata in the metadata api. Also fixed a small bug in http metadata api where extended metadata values were being set to keys instead.

## Issue reference
#4830

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Unit tests passing


pkg/grpc/api.go
```diff
@@ -289,6 +291,9 @@ func NewAPI(
 		}
 	}
 
+	var extendedMetadata sync.Map
```
ItalyPaleAle, line 294: Why sync.Map here? Would a regular map (or better, a key in the struct below) not be enough?

pkg/http/api.go
```diff
@@ -158,6 +160,10 @@ func NewAPI(
 			transactionalStateStores[key] = store.(state.TransactionalStore)
 		}
 	}
+
+	var extendedMetadata sync.Map
```
ItalyPaleAle, line 164: Same here. Since this is used only once, is there a reason for using sync.Map?

pkg/grpc/api.go
addjuarez: The extendedMetadata sync map is a field already present in the api struct and is used to fill the extended_metadata map in the GetMetadataResponse proto. We could add a new field for the version in the api struct but we would still need to add it to extendedMetadata.

pkg/http/api.go
```diff
@@ -1807,7 +1814,7 @@ func (a *api) onGetMetadata(reqCtx *fasthttp.RequestCtx) {
 
 	// Copy synchronously so it can be serialized to JSON.
 	a.extendedMetadata.Range(func(key, value interface{}) bool {
-		temp[key.(string)] = key.(string)
+		temp[key.(string)] = value.(string)
```
ItalyPaleAle, line 1817: Great catch

pkg/grpc/api.go
ItalyPaleAle: I think it makes sense to move it to a separate value, since this is a constant. A sync map is most useful for things that change. But because this is a constant value, it perhaps makes more sense to move it to its own space.

But these are my .02 :)

berndverst: I do not think we should change this as part of this PR.

The PUT / GET `/v1.0/metadata` API stores values in a `sync.map` today. Values may indeed change because we cannot foresee how this API is being used @ItalyPaleAle.

I strongly recommend not changing this.

pkg/http/api.go
berndverst: This is already fixed upstream. I ran into this very problem today with @diegoAJR :)

berndverst (approved):
LGTM

daixiang0:
@addjuarez please fix conflicts.

artursouza (approved):

PR #4868, opened 2022-07-06
optimize: modify context package path

1046102779:


daixiang0:
Thanks for your contribution, could you fix `govet` issue to make CI happy?

pkg/messaging/direct_messaging.go
```diff
@@ -262,13 +262,20 @@ func (d *directMessaging) invokeRemote(ctx context.Context, appID, namespace, ap
 }
 
 func (d *directMessaging) addDestinationAppIDHeaderToMetadata(appID string, req *invokev1.InvokeMethodRequest) {
-	req.Metadata()[invokev1.DestinationIDHeader] = &internalv1pb.ListStringValue{
+	metadata := req.Metadata()
```
daixiang0, line 265: It is another issue, isn't it? I would create another PR to fix it. Also, do not forget add tests.

1046102779: ok, I check and repair it.

1046102779:
image

it be fixed.

daixiang0:
@1046102779 please make CI happy then we can make it land.

pkg/messaging/direct_messaging.go
```diff
@@ -237,7 +237,7 @@ func (d *directMessaging) setContextSpan(ctx context.Context) context.Context {
 }
 
 func (d *directMessaging) invokeRemote(ctx context.Context, appID, namespace, appAddress string, req *invokev1.InvokeMethodRequest) (*invokev1.InvokeMethodResponse, error) {
-	conn, teardown, err := d.connectionCreatorFn(context.TODO(), appAddress, appID, namespace, false, false, false)
```
ItalyPaleAle, line 240: I don't think we should change it here, it was meant as a TODO on purpse IMHO

pkg/grpc/proxy/handler_test.go
```diff
@@ -177,7 +185,10 @@ func (s *ProxyHappySuite) TestDirectorErrorIsPropagated() {
 }
 
 func (s *ProxyHappySuite) TestPingStream_FullDuplexWorks() {
-	stream, err := s.testClient.PingStream(s.ctx())
```
ItalyPaleAle, line 180: Why this change? This reduces the timeout

ItalyPaleAle (changes requested):

pkg/messaging/direct_messaging.go
1046102779: Ok, i change it back.

1046102779: ok . i change it back.

pkg/grpc/proxy/handler_test.go
1046102779: I removed the following method, and then used the `context. WithTimeout` method directly. It is not appropriate to put ctx in the `ProxyHappySuite` object, it should be placed in a entire request.

```golang
func (s *ProxyHappySuite) ctx() context.Context {
```

daixiang0:
@1046102779 could you rebase main branch?

ItalyPaleAle (approved):

daixiang0 (approved):

PR #4873, opened 2022-07-07
Adds cloud event extension attributes & custom properties

### LINKED ISSUE #3991: Add optional & extension CloudEvents attributes
Per the [1.0.1 CloudEvents spec](https://github.com/cloudevents/spec/blob/v1.0.1/spec.md), the following optional fields are missing from [TopicEvent](https://github.com/dapr/dapr/blob/master/dapr/proto/runtime/v1/appcallback.proto#L49-L85).

* dataschema
* subject
* time

Extension attributes are also unsupported.

It might make sense to borrow for CloudEvents does this:

```proto
  // Optional & Extension Attributes
  map<string, CloudEventAttributeValue> attributes = 5;

  ...

  message CloudEventAttributeValue {
    oneof attr {
      bool ce_boolean = 1;
      int32 ce_integer = 2;
      string ce_string = 3;
      bytes ce_bytes = 4;
      string ce_uri = 5;
      string ce_uri_ref = 6;
      google.protobuf.Timestamp ce_timestamp = 7;
    }
  }
```

At some point (v2) it might make sense to adopt the CE proto directly.

https://github.com/cloudevents/spec/blob/v1.0.1/spec.proto

RELEASE NOTE: **ADD** Optional attributes from the CloudEvents 1.0.1 spec.

berndverst:
# Description

Allows cloud event extensions and custom attributes to be sent over gRPC via the TopicEventRequest.

The full Cloud Event types cannot be retained - namely the two URL formats which are specialized string types. Here we simply return strings.

Work in Progress

## Issue reference

Fixes #3991 

## Checklist

* [X] Code compiles correctly
* [ ] Created/updated tests
* [X] Unit tests passing


berndverst:
The Python SDK can be updated to use this like so:

```python

def OnTopicEvent(
            self, request: appcallback_pb2.TopicEventRequest,
            context: grpc.ServicerContext) -> appcallback_v1.TopicEventResponse:
        """Subscribes events from Pubsub."""
        pubsub_topic = request.pubsub_name + DELIMITER + \
            request.topic + DELIMITER + request.path
        no_validation_key = request.pubsub_name + DELIMITER + request.path

        if pubsub_topic not in self._topic_map:
            if no_validation_key in self._topic_map:
                pubsub_topic = no_validation_key
            else:
                context.set_code(grpc.StatusCode.UNIMPLEMENTED)  # type: ignore
                raise NotImplementedError(
                    f'topic {request.topic} is not implemented!')

        event = v1.Event()
        event.SetEventType(request.type)
        event.SetEventID(request.id)
        event.SetSource(request.source)
        event.SetData(request.data)
        event.SetContentType(request.data_content_type)

        # NEW CODE FOR EXTENSIONS STARTS HERE

        attributes = dict()
        for k, v in request.attributes.items():
            field = v.WhichOneof(oneof_name='attr')
            if field:
                attributes[k] = v.__getattribute__(field)
            else:
                attributes[k] = None
        event.SetExtensions(attributes)

        # NEW CODE FOR EXTENSIONS END HERE

        # TODO: add metadata from context to CE envelope

        response = self._topic_map[pubsub_topic](event)
        if isinstance(response, TopicEventResponse):
            return appcallback_v1.TopicEventResponse(status=response.status.value)
        return empty_pb2.Empty()

```

dapr/proto/runtime/v1/appcallback.proto
```diff
@@ -82,6 +82,20 @@ message TopicEventRequest {
...
   // This value is used by OnTopicEvent to "switch" inside the handler.
   string path = 9;
+
+  // The cloud event extension attributes (custom attributes).
+  // Compatible with the cloud events v1.0.2 spec.
+  map<string, CloudEventAttributeValue> attributes = 10;
+
+}
+
+message CloudEventAttributeValue {
```
berndverst, line 92: This is a subset of the options from the official CloudEvent proto spec. They define additional string fields for URLs, but we cannot distinguish between these types because we are not receiving data via the cloudevents proto. Similarly for the Time field defined in that spec.

pkg/runtime/pubsub/cloudevents.go
```diff
@@ -31,6 +31,27 @@ type CloudEvent struct {
...
+		contrib_pubsub.DataContentTypeField,
+		contrib_pubsub.SourceField,
+		contrib_pubsub.TypeField,
+		contrib_pubsub.TopicField,
+		contrib_pubsub.PubsubField,
+		contrib_pubsub.SubjectField,
+		contrib_pubsub.DataBase64Field,
+		contrib_pubsub.DataField,
+		contrib_pubsub.ExpirationField,
+		contrib_pubsub.TraceIDField,
```
mukundansundar, line 47: This is not as per spec

PR #4887, opened 2022-07-08
Fix release noted to include 1845 from contrib

pravinpushkar:

Fix release notes to include missed issue from contrib repository


artursouza (changes requested):
Let's merge into release branch first, please. Then we can merge release branch into master, as usual.

PR #4889, opened 2022-07-08
Some fixes for dev container and running Go code

ItalyPaleAle:
Fixed a few issues in the 1.8 dev container:

1. Fixed: there was a permission issue in the GOROOT so Go could not fetch modules.
2. Installing protoc-gen-go-grpc too, which was missing and it's required to rebuild protos
3. Switched Dapr CLI to use GHCR by default
4. Installing golangci-lint from pre-compiled binaries as that's what is recommended by the project's authors


artursouza (approved):

ItalyPaleAle:
After merging this, please trigger https://github.com/dapr/dapr/actions/workflows/dapr-dev-container.yml

PR #4890, opened 2022-07-09
Fixed maxConcurrency issue in HTTP channel

ItalyPaleAle:
In case of error, the item was not removed from the channel


pkg/channel/http/http_channel.go
```diff
@@ -167,8 +167,13 @@ func (h *Channel) invokeMethodV1(ctx context.Context, req *invokev1.InvokeMethod
...
 
 	if h.ch != nil {
-		h.ch <- 1
+		h.ch <- struct{}{}
 	}
+	defer func() {
+		if h.ch != nil {
+			<-h.ch
+		}
+	}()
```
halspang, line 176: Seems like we should get a test for this. Is that possible?

halspang (commented):
Change seems fine, just want a test for it if possible.

pkg/channel/http/http_channel.go
ItalyPaleAle: Added test

ItalyPaleAle:
@halspang I added the tests, can this be approved and merged now?

daixiang0 (approved):

Issue #4866, opened 2022-07-06
Why is grpc via dapr slower than http?


hspaay:
**Note:** If you have a general support question and are looking for a quicker response, please checkout our discord channel for answers from the community:
https://aka.ms/dapr-discord

## In what area(s)?

/area runtime

> /area operator

> /area placement

> /area docs

> /area test-and-release

## Ask your question here
A simplistic performance comparison of invoking a method via gRPC and http with and without dapr shows that without dapr the grpc calls are more efficient by factor 2 but via dapr less efficient by 10%. Also interesting is that grpc performance is nearly 5 times slower via dapr compared to direct calls. Is this expected?

The tests are in the makefile in this repo: https://github.com/wostzone/echo-dapr and the readme shows the numbers. Repeated here for convenience:
| Makefile | test                         | duration  |
|----------|------------------------------|-----------|
| run1     | 1000 plain grpc calls        | 164 msec  |
| run2     | 1000 plain http calls        | 320 msec  |
| run3     | 1000 dapr wrapped grpc calls | 986 msec  |
| run4     | 1000 dapr wrapped http calls | 824 msec  |
| run5     | 1000 grpc sdk calls          | 772 msec  |
| run6     | 1000 http sdk calls          | 700 msec  |

Please note that this is a simplistic performance test without additional middleware, just simple invoke.

berndverst:
What is the payload size here? Could you repeat the test for small payloads and large payloads?

There will be overhead in deserializing and serializing protobufs, but this should be negligible for larger requests. So I wonder if you see gRPC being more performant for larger requests.

hspaay:
Running the test with a small, 1KB, 10KB and 100KB random text payload yields the following timing:

| Makefile | test                         | Hello world | 1K text | 10K text | 100K text  |
|----------|------------------------------|-------------|---------|----------|------------|
| run1     | 1000 plain grpc calls        | 150 msec    | 150 msec    | 200 msec  | 550 msec   |
| run2     | 1000 plain http calls        | 320 msec    | 355 msec    | 520 msec  | 2050 msec  |
| run3     | 1000 dapr wrapped grpc calls | 970 msec    | 980 msec    | 1110 msec | fails (*1) |
| run4     | 1000 dapr wrapped http calls | 810 msec    | 820 msec    | 1110 msec | 3160 msec  |
| run5     | 1000 grpc sdk calls          | 760 msec    | 760 msec    | 980 msec  | 3230 msec  |
[...]

Issue #4804, opened 2022-06-20
Blob storage building block


artursouza:
Today, we use binding for multiple components, without a consistent API spec but there are a subset of those components that behave like a blob storage. There is an opportunity to consolidate those into a defined API spec (new building block). The API would be simple to POST and GET to/from blob storage, offering portability.

yaron2:
> This is different from state store where the API surface is expected to be small with no deviation from the generic behavior of getting and posting one object.

If the API for blobs is a subset of the API for state, I'm concerned that multiple implementations (3 - bindings, state stores, blobs) might be confusing, and developers will naturally go to the one that has the most API surface, ie. state.

In the likely event that developers will also want to delete and post multiple objects, we'll see the Blob API become an almost mirror reflection of the state API.

Thoughts?

ItalyPaleAle:
I think the missing part in current state stores, to make them effective for managing blobs, is support for streaming. Our current APIs load everything in memory in multiple steps (application and sidecar both load the entire blob in memory), so this is not ideal for large blobs.

A suggestion I brought up to @berndverst recently was to add support for streams to state stores, when the underlying service and SDK support that. We'd have two version of the APIs to work with byte slices and streams separately, for example:

```go
// Uses byte slices -> data is in the Data property in the response object
Get(req *state.GetRequest) (*state.GetResponse, error)
// Uses streams -> in the response object, Data is nil
GetStream(req *state.GetRequest, out io.Writer) (*state.GetResponse, error)
// Uses byte slices -> data is in the Value property of the request object
Set(req *state.SetRequest) error
// Uses streams -> in the request object, the Value property is nil
SetStream(req *state.StateRequest, in io.Reader) error
// Alternatively, we could set a stream in the req object and not add a new signature
```

For components that support streams, methods would only use streams. If users pass a byte slice, then a reader is created from that.

For components that do not support streams, you can still use the `*Stream` methods. Internally, the method reads the data from a stream into a buffer.

olitomlinson:
> If the API for blobs is a subset of the API for state, I'm concerned that multiple implementations (3 - bindings, state stores, blobs) might be confusing, and developers will naturally go to the one that has the most API surface, ie. state.

Completely agree with this.

I still struggle with there being a first-class State Management building block, but then also a myriad of bindings that do state operations as well... it just feel somewhat fragmented and IMO doesn't paint a solid story for Dapr and state management concerns.

[...]

PR #4903, opened 2022-07-14
Streaming for communication between sidecars and for (HTTP) service invocation

### LINKED ISSUE #3103: Stream requests and not buffer on invoke service method.
## In what area(s)?

/area runtime
> /area operator
> /area placement
> /area docs
> /area test-and-release
## Describe the proposal
Dapr should stream requests (and responses) and not buffer when invoking remote service. Use cases: sending large file to another service (request), receiving large file or result set from service (response).

I know that there's MaxRequestBodySize (here's related issue https://github.com/dapr/cli/issues/699) but large request bodies should be handled by streaming them directly to the destination service and not buffering requests in memory. I'm not sure how responses are handled by Dapr but I suppose that the same way.

ItalyPaleAle:
# Description

This PR is the first step in enabling support for streaming data that transits through Dapr runtimes, with the initial focus on sidecar-to-sidecar communication, as well as service invocation(currently limited to HTTP only).

To explain the issue, here's a demo of a service "client" calling into "server" through Dapr, making an HTTP request. The payload is "big" (about 70KB) and for demonstration purposes, both "client" and "server" are slow at producing or consuming the data.

https://user-images.githubusercontent.com/43508/178880142-4f170a0e-6ab2-487a-9f3a-0f29f519ea6f.mp4

As you can see, the flow is:

1. "client" makes a request to its sidecar
2. The sidecar waits for the entire request to be complete and loads the entire data in memory, then serializes it (to protobuf)
3. "client" app's sidecar then sends the data to "server" app's sidecar (I did not add any "slowdown" in this step, but this does take some time especially with large payloads)
4. "server" app's sidecar waits for the entire payload to be received (from "client" app's sidecar) and loaded in memory, then un-serializes it
5. "server" app's sidecar sends the data to the "server" app

The flow is then repeated exactly the same way (but with roles inverted) for the response.

Because in every step the sidecars read the entire message in memory, this can make it challenging to use Dapr service-to-service invocation with large payloads (for example, images) because it adds significant latency. Additionally, because all the data is kept in memory in both serialized and unserialized forms, there's an increase in memory usage.

This PR changes the behavior, and now service-to-service invocation (same example) behaves this way:

https://user-images.githubusercontent.com/43508/178880049-4b377ea8-b78e-4dce-95ca-5b89e3bb71d7.mp4

By leveraging streaming, this is much faster, especially when transferring large files or when the producer is slow. Especially the TTFB (Time To First Byte) is much shorter.

What this PR includes, at a higher level:

- Changed the way sidecars invoke one another to use streams. The new method `ServiceInvocation.CallLocalStream` is a variant of `ServiceInvocation.CallLocal` that uses streams rather than making unary calls. Internally this uses a buffer of up to 4KB, so messages that are smaller than that (and that are received in a single chunk by the runtime) are transferred in a single block. This should allow having no performance impact for small payloads (compared to unary RPCs).
- In the HTTP API server, service invocation calls are now read as streams, so the runtime can begin invoking the other app as soon as the first bytes come in.
- Likewise, when the sidecar invokes the apps using HTTP, the data is now sent to the app as a stream. Apps can begin receiving data as soon as the runtime gets it.

It also includes a lot of internal changes to support the two things above.

For clients that invoke Dapr via regular HTTP calls and receive responses through a regular HTTP server, no change will be needed. They will be able to enjoy the benefits of streaming for free and without doing anything!

For clients that use the Dapr SDKs, we may need to update the SDKs to leverage streaming when possible and appropriate. That is a separate work item.

Lastly, this PR is not "complete" as it doesn't add support for streams _everywhere_ where we could want them. For example, we will want to support clients that use gRPC too. We will also want to leverage streaming for reading/writing into state stores and bindings.

Talking to @artursouza, he recommended focusing on HTTP service invocation first, and take this as an opportunity to lay the groundwork that will ultimately be useful in supporting all other scenarios. In this regard, this PR does provide some solid foundations for the remaining work.

## Implementation details

This PR creates a new gRPC method that is used for sidecar-to-sidecar communication.

Currently, sidecar-to-sidecar communication uses the [`CallLocal` RPC](https://github.com/dapr/dapr/blob/2addb5754398b84bde7aa3891a5bb9a11395b413/dapr/proto/internals/v1/service_invocation.proto#L39):

```proto3
// Invokes a method of the specific service.
rpc CallLocal (InternalInvokeRequest) returns (InternalInvokeResponse) {}
```

This RPC is not removed and is retained for backwards compatibility (e.g. when Dapr 1.8 invokes Dapr vNext), however it is not used otherwise.

A new RPC is added:

```proto3
// Invokes a method of the specific service using a stream of data.
rpc CallLocalStream (stream InternalInvokeRequestStream) returns (stream InternalInvokeResponseStream) {}
```

The request and response objects are defined as:

```proto3
// InternalInvokeRequestStream is a variant of InternalInvokeRequest used in streaming RPCs.
message InternalInvokeRequestStream {
  // Request details.
  // This does not contain any data in message.data.
  InternalInvokeRequest request = 1;

  // Chunk of data.
  StreamPayload payload = 2;
}

// InternalInvokeResponseStream is a variant of InternalInvokeResponse used in streaming RPCs.
message InternalInvokeResponseStream {
  // Response details.
  // This does not contain any data in message.data.
  InternalInvokeResponse response = 1;

  // Chunk of data.
  StreamPayload payload = 2;
}

// Chunk of data sent in a streaming request or response.
message StreamPayload {
  // Data sent in the chunk.
  google.protobuf.Any data = 1;

  // Set to true if this is the last chunk.
  bool complete = 2;
}
```

In the protos above, `InternalInvokeRequest` and `InternalInvokeResponse` are the existing protos, that are used by the unary `CallLocal` RPC. In the streamed version, they are wrapped in a proto together with a `StreamPayload` object.

> Note that, although the RPC uses bi-directional streaming, it still **behaves like a unary call**, in the sense that once the connection has been established, first the sender sends the entire request, and then it reads the entire response from the receiver. We use streams to be able to send data in chunks, but the sidecars do not maintain a persistent stream that is used to exchange multiple messages or where communication is constantly bi-directional.

How it works:

1. The caller sidecar invokes `CallLocalStream` on the target sidecar and establishes the gRPC connection and the streaming RPC.
2. The caller sidecar sends the data in chunks (see below for details on how chunking works).
3. The target sidecar can begin processing the message (i.e. send it to the target app) as soon as the first chunk is received.
4. Once the full payload has been sent, the target sidecar waits for the response from the app (which is read as a stream).
5. The target sidecar sends the response in chunks to the caller sidecar, with the first chunk sent as soon as it's available.

Each chunk is a message of type `InternalInvokeRequestStream` or `InternalInvokeResponseStream` sent over the RPC, where:

- The first message in the stream MUST contain an `InternalInvokeRequest` or `InternalInvokeResponse` with the required keys present.
- The first message in the stream MAY contain a `payload`, but that is not required.
- Subsequent messages (any message except the first in the stream) MUST contain a `payload` and MUST NOT contain any other property.
- The last message in the stream MUST contain a `payload` with `complete` set to `true`. That message is assumed to be the last one from the sender and no more messages are to be sent in the stream after that.

The amount of data contained in `payload.data` is variable and it's up to the discretion of the sender. In this PR, chunks are at most 4KB in size, although senders may send smaller chunks if they wish. Receivers must not assume that messages will contain any specific number of bytes in the payload.

Note that it's possible for senders to send a single message in the stream. If the data is small and could fit in a single chunk, senders MAY choose to include a `payload` with `complete=true` in the first message. In this case, the entire message contains both an `InternalInvokeRequest`/`InternalInvokeResponse` with the required keys present AND a `payload` with data (if any) and `complete=true`. Receivers should assume that single message to be the entire communication from the sender.

## To do

- [ ] Figure out a way to ensure backwards compatibility when a runtime that supports streaming (e.g. daprd "vNext") tries to invoke a runtime that does not (e.g. daprd 1.8)
- [ ] Update docs
- [ ] Review what updates are needed in the SDKs

## Issue reference

Fixes #3103
Also impacts other issues such as #4866

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests
* [X] Unit tests passing
* [ ] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: @ItalyPaleAle TODO


yaron2:
Please make sure that this supports load balancing, otherwise we cannot accept this PR.

Meaning, if you call between app 1 and app 2 through Dapr and another instance pops up for app 2, it will start receiving requests.

yaron2:
Also, for the scenario mentioned of transferring a file between two sidecars, why can't users do that with gRPC streaming with Dapr? As Dapr supports streaming naturally all the way through from app to sidecar, sidecar to remote sidecar and then to the target app.

ItalyPaleAle:
> Please make sure that this supports load balancing, otherwise we cannot accept this PR.
> Meaning, if you call between app 1 and app 2 through Dapr and another instance pops up for app 2, it will start receiving requests.

Yes, this should have no issue with load balancing. It creates a new stream for each invocation, so it can load balance. Internally, at the wire level, it behaves the same as an unary invocation.

> Also, for the scenario mentioned of transferring a file between two sidecars, why can't users do that with gRPC streaming with Dapr?

This is based on HTTP which supports streaming natively as a protocol, we just didn’t support it in Dapr. HTTP is simpler and is easier to integrate with existing services.

In fact, I started thinking about this problem about 2 months ago when I made an attempt at developing a demo using [Imaginary](https://github.com/h2non/imaginary) (which exposes a HTTP endpoint) through Dapr, and noticed these issues.

One other thing to note is that this only supports a scenario where there’s a single request followed by a single response (is not meant for long-running streams). The goal is to provide a different, stream-based transport for service invocation: for end-users, they should continue to think of this as regular service invocation, and not think of this as a stream. gRPC proxying remains the best solution when users need long-running streams and the ability to have two-way, alternating communication.

Lastly, as I wrote, this starts with HTTP but it also lays the foundation for supporting streams throughout the runtime. My goal is to support streams for other things including state stores (see #4804) and bindings.

yaron2:
> Please make sure that this supports load balancing, otherwise we cannot accept this PR.
> 
> Meaning, if you call between app 1 and app 2 through Dapr and another instance pops up for app 2, it will start receiving requests.

Alternatively, you can keep existing mechanisms untouched while making streaming explicit and respecting HTTP semantics by introducing the `Transfer-Encoding: chunked` header.

ItalyPaleAle:
I don’t think we need `Transfer-Encoding: Chunked`: that is meant for different purposes such as receiving a sequence of messages (like server-side events). This is meant to make service invocation faster and more efficient, by processing HTTP messages as a byte stream. It doesn’t require any change at the application level, and will work out-of-the-box even with regular curl requests.

yaron2:
> Yes, this should have no issue with load balancing. It creates a new stream for each invocation, so it can load balance. Internally, at the wire level, it behaves the same as an unary invocation.

Ok, just please make sure to validate that either manually or with a test (better).

See my comment above about leaving current mechanism untouched and enabling streaming explicitly using a header.

yaron2:
> I don’t think we need `Transfer-Encoding: Chunked`: that is meant for different purposes such as receiving a sequence of messages (like server-side events). This is meant to make service invocation faster and more efficient, by processing HTTP messages as a byte stream. It doesn’t require any change at the application level, and will work out-of-the-box even with regular curl requests.

Sounds good if there's no risk of behavioral breaking change like load balancing. Once this PR compiles I'll help verify.

ItalyPaleAle:
Note that I've not abandoned this PR. I just realized this is very big, and a lot of things won't work well until #4979 is merged.

I'm the process of splitting it into multiple PRs, also to make it easier to review.

yaron2:
> Note that I've not abandoned this PR. I just realized this is very big, and a lot of things won't work well until #4979 is merged.
> 
> I'm the process of splitting it into multiple PRs, also to make it easier to review.

Since #4979 will only be merge in 1.10 the soonest, doesn't it make sense to close this PR for now as this is at least 4/5 months away?

ItalyPaleAle:
The “core” PR I’m working on does not involve the API server (app-to-dapr).

It only implements partial streaming support by changing the messaging APIs (dapr-to-dapr) and the HTTP channel (dapr-to-app). That’s not full support for streaming but at least implements the core functionality and the required APIs to unblock things like blob stores and cryptography, who depend on core streaming support.

Updating the HTTP API server will indeed depend on #4979.

I will close this PR once the “core” one is ready, hopefully by end of next week.

ItalyPaleAle:
Closing this in favor of #5170

Issue #4852, opened 2022-06-30
Generalized Resiliency Policies


halspang:
## In what area(s)?

/area runtime

> /area operator

> /area placement

> /area docs

> /area test-and-release
## Describe the proposal
Resiliency as it exists in 1.8/1.7 requires that policies be applied to specific targets (apps, actors, or components). In sufficiently complex systems, this could lead to a very large resiliency configuration. Importantly, this could also lead to applications/actors/components not having a policy by mistake. 

To address this, I propose the addition of generic/general/default policies in Resiliency. These are policies that could be defined once and, if there is no other policy defined, could then be applied to any target that does not already have an existing policy.

### Proposal
I propose that we use follow the pattern set by the Built-in retries. That is, we will utilize keywords to signify policies that can be used as a default if no other policies are referenced. This would look something like this in the yaml:

```yaml
apiVersion: dapr.io/v1alpha1
kind: Resiliency
metadata:
  name: resiliency
spec:

  policies:
    # Retries are named templates for and are instantiated for life of the operation.
    retries:
      DaprBuiltInServiceRetries:
        policy: constant
        duration: 2s
        maxRetries: 5 

      DefaultServiceInvocationRetries:
        policy: constant
        duration: 1s
        maxRetries: 3

      5Retries:
        policy: constant
        duration: 10s
        maxRetries: 5

    timeouts:
      DefaultServiceInvocationTimeouts: 5s
  
  targets:
    apps:
      DemoApp:
        retry: 5Retries

    actors:
      DemoActor:
        retry: 5Retries
```

With this policy, which is user defined, any application that is _not_ "DemoApp" will use the retry and timeout combo from `DefaultServiceInvocationRetries` and `DefaultServiceInvocationTimeouts` (note there could also be a default circuit breaker). These do not start with the `Dapr` prefix because they are defined by users and Dapr does not include them automatically. 

Actors and components would also be able to define their own defaults with similar names like `DefaultActorRetries` or `DefaultComponentTimeouts`. 

An open consideration is that components might want to striate their defaults. This is because the act of publishing to a broker will most likely require a vastly different timeout than saving state.

It's also worth noting that default policies won't mix and match with defined policies. For example:

```yaml
apiVersion: dapr.io/v1alpha1
kind: Resiliency
metadata:
  name: resiliency
spec:

  policies:
    # Retries are named templates for and are instantiated for life of the operation.
    retries:
      DefaultServiceInvocationRetries:
        policy: constant
        duration: 1s
[...]

PR #4906, opened 2022-07-14
Add default policy support for apps and actors

halspang:
# Description

This commit adds support for specifying default resiliency policies
for both apps and actors. These policies, if defined, will be used
if there are no specific policies set for the target being called.

https://github.com/dapr/dapr/issues/4852

## Issue reference

Please reference the issue this PR will close: #4852 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Unit tests passing

### DIFF
--- a/pkg/resiliency/resiliency.go
+++ b/pkg/resiliency/resiliency.go
@@ -47,14 +47,17 @@ const (
 	defaultEndpointCacheSize = 100
 	defaultActorCacheSize    = 5000
 
-	BuiltInServiceRetries        BuiltInPolicyName = "DaprBuiltInServiceRetries"
-	BuiltInActorRetries          BuiltInPolicyName = "DaprBuiltInActorRetries"
-	BuiltInActorReminderRetries  BuiltInPolicyName = "DaprBuiltInActorReminderRetries"
-	BuiltInActorNotFoundRetries  BuiltInPolicyName = "DaprBuiltInActorNotFoundRetries"
-	BuiltInInitializationRetries BuiltInPolicyName = "DaprBuiltInInitializationRetries"
-	Endpoint                     PolicyType        = "endpoint"
-	Component                    PolicyType        = "component"
-	Actor                        PolicyType        = "actor"
+	BuiltInServiceRetries         BuiltInPolicyName     = "DaprBuiltInServiceRetries"
+	BuiltInActorRetries           BuiltInPolicyName     = "DaprBuiltInActorRetries"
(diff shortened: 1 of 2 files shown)


daixiang0:
Do we have a clear document about the default policy to ensure users know what will happen?

halspang:
@daixiang0 - I made a docs issue for it: https://github.com/dapr/docs/issues/2664

pkg/resiliency/resiliency.go
```diff
@@ -532,6 +536,39 @@ func (r *Resiliency) EndpointPolicy(ctx context.Context, app string, endpoint st
 				}
 			}
 		}
+	} else {
+		if defaultNames, ok := r.getDefaultPolicy(Endpoint); ok {
+			r.log.Debugf("Found Default Policy for Endpoint %s: %+v", app, defaultNames)
```
artursouza, line 541: Please, log info since this can help a lot with troubleshooting when users/customers report an issue.

pkg/resiliency/resiliency.go
```diff
@@ -581,6 +618,43 @@ func (r *Resiliency) ActorPreLockPolicy(ctx context.Context, actorType string, i
 				}
 			}
 		}
+	} else {
+		if defaultNames, ok := r.getDefaultPolicy(Actor); ok {
+			r.log.Debugf("Found Default Policy for Actor type %s: %+v", actorType, defaultNames)
```
artursouza, line 623: same.

pkg/resiliency/resiliency.go
```diff
@@ -601,6 +675,13 @@ func (r *Resiliency) ActorPostLockPolicy(ctx context.Context, actorType string,
 		if policyNames.Timeout != "" {
 			t = r.timeouts[policyNames.Timeout]
 		}
+	} else {
+		if defaultPolicies, ok := r.getDefaultPolicy(Actor); ok {
+			r.log.Debugf("Found Default Policy for Actor type %s: %+v", actorType, defaultPolicies)
```
artursouza, line 680: Same.

artursouza (commented):
Just some minor comment on log level. LGTM.

pkg/resiliency/resiliency.go
halspang: I think that'd be a bit verbose. It would be displayed at every call since policies are loaded at call time, not stored in memory permanently. Here's that same log as it exists today:

https://github.com/dapr/dapr/blob/master/pkg/resiliency/resiliency.go#L506

artursouza (approved):
LGTM. Log level should be debug because it is executed per API call and not init.

PR #4909, opened 2022-07-14
Enable tcpdump in E2E runners

ItalyPaleAle:
As requested by the Actions team to help debug our network connectivity issues

https://github.com/actions/virtual-environments/issues/5850


ItalyPaleAle:
@yaron2 I'm afraid the tests won't show the TCP dumps being collected here because the pipeline comes from master. You can see an example here: https://github.com/ItalyPaleAle/dapr-ci-test/actions/runs/2673418016

PR #4926, opened 2022-07-21
[Hotfix] Use the correct version of backoff

ItalyPaleAle:
In two files, including (critically) `pkg/messaging/direct_messaging.go`, the `github.com/cenkalti/backoff` package was imported referencing the wrong module, missing `/v4` at the end. This caused an older version of the package (2.2.1) to be used instead.

Because of that, every time we returned `backoff.Permanent` (such as within `invokeWithRetry`), the permanent error was ignored. This caused a variety of issues including retrying when  up to a panic.

For example, when invoking `CallLocal` over gRPC and receiving an UNAUTHENTICATED or UNAVAILABLE error, then a connection error, the policy would retry even though that error was meant to be permanent.

Worse, when the response was for example UNIMPLEMENTED or another gRPC error code, which should have been a permanent error, the policy would have retried regardless and eventually (due to a chain of events), daprd would have panicked.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -61,7 +61,6 @@ require (
 )
 
 require (
-	github.com/cenkalti/backoff v2.2.1+incompatible
 	github.com/golang/mock v1.6.0
 	github.com/golang/protobuf v1.5.2
 )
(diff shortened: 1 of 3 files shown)


halspang (approved):

yaron2:
Is this a regression? If so, can you point out when/where it occured?

ItalyPaleAle:
@yaron2 this is the PR that added the regression: https://github.com/dapr/dapr/pull/4609 It was added in 1.8.

User who reported this issue said it was working fine in 1.7.

yaron2:
> @yaron2 this is the PR that added the regression: #4609 It was added in 1.8.
> 
> User who reported this issue said it was working fine in 1.7.

Alright, I agree this requires 1.8.2.

yaron2 (approved):

artursouza (approved):

Issue #4692, opened 2022-06-01
Proposal: Option to disable certain components at runtime


ItalyPaleAle:
## In what area(s)?

/area runtime

> /area operator

> /area placement

> /area docs

> /area test-and-release

## Describe the proposal

At present, developers are able to use every supported component in Dapr, by creating a component CRD with the desired component type.

In some situations, that may not be desired. For example, cluster administrators may want to restrict the type of available components (e.g. to disable certain state stores, secret stores, etc, either via an allowlist or a denylist).

It would be helpful for that to offer a new configuration option (set through the [configuration CRD](https://docs.dapr.io/operations/configuration/configuration-overview)) to include either an allowlist or a denylist of components.

For example:

- `dapr.io/components-allow` -> allowlist of components that can be used, as a list of component types (e.g. `pubsub.in-memory`), comma-separated.
- `dapr.io/components-deny` -> disallowlist of components that cannot be used, just like the previous.

Administrators can use either one of the two options, but not both, to control what components are allowed (only those in the allowlist can be used) or what components are disallowed (all components can be used except those in the denylist).

yaron2:
This is possible via scopes today.

https://docs.dapr.io/operations/components/component-scopes/

ItalyPaleAle:
I think that’s something different though, isn’t it?

scoping is about making a configured component visible to certain apps only. For example, a component of type pubsub.in-memory instantiated with name “mypubsub” could be used by certain apps only.

This proposal is about restricting the types of components that can be created. Let’s say that the admin configures Dapr to disallow in-memory pubsub components: then the entire component is not instantiated in the runtime (the code is still present, but it’s not loaded). A developer who tried to instantiate a component of that type would then get an error saying that component type does not exist.

yaron2:
> I think that’s something different though, isn’t it?
> 
> scoping is about making a configured component visible to certain apps only. For example, a component of type pubsub.in-memory instantiated with name “mypubsub” could be used by certain apps only.
> 
> This proposal is about restricting the types of components that can be created. Let’s say that the admin configures Dapr to disallow in-memory pubsub components: then the entire component is not instantiated in the runtime (the code is still present, but it’s not loaded). A developer who tried to instantiate a component of that type would then get an error saying that component type does not exist.

This is exactly what scopes does. If a component is scoped to app one,  app two cannot instantiate the component or use it in any way. The component never goes through init and cannot be used.

ItalyPaleAle:
[...]

PR #4945, opened 2022-07-29
Component denylist

ItalyPaleAle:
This PR adds the ability to create a denylist that forbids loading certain types of components. This is passed as an environmental variable (for the daprd container) called `DENY_COMPONENTS`.

The denylist is a comma-spearated list of component identifiers. For example:

```
DENY_COMPONENTS="bindings.smtp,bindings.kubernetes"
```

Disables the "bindings.smtp" and "bindings.kubernetes" components. It's possible to specify a version too, to disable that version only (if the version is not specified, all versions are disabled):

```
DENY_COMPONENTS="bindings.kubernetes/v1"
```

The goal is being able to disable certain components from being instantiated. This issue fixes https://github.com/dapr/dapr/issues/4692 although it is done in a much softer way, including a denylist only and using env vars instead of annotations. Additionally, it prevents components from being loaded, but does not prevent them from being registered.

### DIFF
--- a/pkg/components/secretstores/registry.go
+++ b/pkg/components/secretstores/registry.go
@@ -19,9 +19,11 @@ import (
 	"github.com/pkg/errors"
 
 	"github.com/dapr/components-contrib/secretstores"
-
 	"github.com/dapr/dapr/pkg/components"
 )
+
+// Name of the built-in Kubernetes secret store component.
+const BuiltinKubernetesSecretStore = "kubernetes"
 
 type (
 	// SecretStore is a secret store component definition.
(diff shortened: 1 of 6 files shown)


yaron2:
Please change the PR to use Configuration CRD instead of an env var. It's more explicit and will allow extensibility.

yaron2:
Also, let's wait with merging until https://github.com/dapr/dapr/pull/4808 is in as there will likely be a merge conflict.

ItalyPaleAle:
Closing this in favor of #4946 which uses the configuration CRD.

PR #4946, opened 2022-07-30
Component denylist via Configuration CRD

ItalyPaleAle:
This PR adds the ability to create a denylist that forbids loading certain types of components. This is done by adding a new property to the configuration CRD:

```yaml
apiVersion: dapr.io/v1alpha1
kind: Configuration
metadata:
  name: secretappconfig
spec: 
  components:
    deny:
      - bindings.smtp
      - bindings.kubernetes
```

Disables the "bindings.smtp" and "bindings.kubernetes" components. It's possible to specify a version too with the syntax `component/version` (e.g. `bindings.kubernetes/v1`), to disable that version only (if the version is not specified, all versions are disabled).

The goal is being able to disable certain components from being instantiated. This issue fixes https://github.com/dapr/dapr/issues/4692 although it is done in a much softer way, including a denylist only and using env vars instead of annotations. Additionally, it prevents components from being loaded, but does not prevent them from being registered.

One special note for the `secretstores.kubernetes` component: adding it to the denylist will prevent developers from configuring new components of that type. However, when on Kubernetes, the built-in secret store called `kubernetes` is configured regardless of whether `secretstores.kubernetes` is disallowed and users can still use that. That is by design, as that component is usually necessary for example to configure secrets for other components' metadata. Users who want to disable the built-in secret store can use the `dapr.io/disable-builtin-k8s-secret-store` annotation that's been available since 1.8.

This behavior is validated by the updated E2E test too, which checks that components can be disabled and yet that the built-in secret store can still be used.


artursouza (approved):

yaron2 (approved):
LGTM

yaron2:
@ItalyPaleAle please follow up with a docs PR. I'll merge this once tests pass.

ItalyPaleAle:
Thanks, @yaron2 ! I opened an issue in Docs here because I don't have time to do it _today_, but I've assigned it to me: https://github.com/dapr/docs/issues/2704

yaron2:
> Thanks, @yaron2 ! I opened an issue in Docs because I don't have time to make a PR _today_, but I've assigned it to me: [dapr/docs#2704](https://github.com/dapr/docs/issues/2704)

Needs to be done by end of milestone, certainly not today :)

PR #4975, opened 2022-08-04
Improve component init error logs and metrics

addjuarez:
# Description

Added a componentName tag to the **dapr_runtime_component_init_fail_total** metric. The metric should now be produced when init failures are logged. 

Also have standardized the component failure error logging for creation/init stages. All component init/create errors should now be logged. 

Also removed some redundant logging of errors. 

A new error type InitErrors was created and was modeled after custom etag errors. We have for initError codes and have a standardized format for error messages. 
The new error message will look like : 
` (InitComponentFailure): initialization error occurred for statestore2 (state.redis/v1): init timeout for component statestore2 exceeded after 5s"`
The new error codes.
- InitComponetFailure : Any component failures to initialize or complete their init() method.
- InitFailure : Any non component failures like actor init failures
- CreateComponentFailure : component failed to be added to the registry

## Issue reference

Part of the work involved with this.
https://github.com/dapr/components-contrib/issues/1752

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Unit tests passing


berndverst:
For standardizing the error - what if instead we created a new error type DaprComponentInitError or something like that and change the Init interface to return that. And then we update the runtime to just log that error if Init returns an error of this type.

Creating a new error type and expecting this in the return is a much more guaranteed way to keep things consistent going forward. Nothing prevents a new component from failing to return the right error type with your proposal here. This puts unnecessary burden on approvers and maintainers to ensure Init errors are consistent.

pkg/runtime/runtime.go
```diff
@@ -1356,18 +1356,18 @@ func (a *DaprRuntime) isAppSubscribedToBinding(binding string) bool {
 func (a *DaprRuntime) initInputBinding(c components_v1alpha1.Component) error {
 	binding, err := a.bindingsRegistry.CreateInputBinding(c.Spec.Type, c.Spec.Version)
 	if err != nil {
-		log.Warnf("failed to create input binding %s (%s/%s): %s", c.ObjectMeta.Name, c.Spec.Type, c.Spec.Version, err)
```
artursouza, line 1359: Any reason you removed this logging?

pkg/runtime/runtime.go
```diff
@@ -1356,18 +1356,18 @@ func (a *DaprRuntime) isAppSubscribedToBinding(binding string) bool {
...
 	}
 	err = binding.Init(bindings.Metadata{
 		Properties: a.convertMetadataItemsToProperties(c.Spec.Metadata),
 		Name:       c.ObjectMeta.Name,
 	})
 	if err != nil {
-		log.Errorf("failed to init input binding %s (%s/%s): %s", c.ObjectMeta.Name, c.Spec.Type, c.Spec.Version, err)
-		diag.DefaultMonitoring.ComponentInitFailed(c.Spec.Type, "init")
-		return err
+		diag.DefaultMonitoring.ComponentInitFailed(c.Spec.Type, "init", c.ObjectMeta.Name)
```
artursouza, line 1368: Same.

pkg/runtime/runtime.go
```diff
@@ -1385,9 +1385,9 @@ func (a *DaprRuntime) initInputBinding(c components_v1alpha1.Component) error {
 func (a *DaprRuntime) initOutputBinding(c components_v1alpha1.Component) error {
 	binding, err := a.bindingsRegistry.CreateOutputBinding(c.Spec.Type, c.Spec.Version)
 	if err != nil {
-		log.Warnf("failed to create output binding %s (%s/%s): %s", c.ObjectMeta.Name, c.Spec.Type, c.Spec.Version, err)
-		diag.DefaultMonitoring.ComponentInitFailed(c.Spec.Type, "creation")
-		return err
+		diag.DefaultMonitoring.ComponentInitFailed(c.Spec.Type, "creation", c.ObjectMeta.Name)
```
artursouza, line 1388: Same.

pkg/runtime/runtime.go
```diff
@@ -2536,18 +2545,18 @@ func (a *DaprRuntime) appendBuiltinSecretStore() {
...
 	}
 
 	err = secretStore.Init(secretstores.Metadata{
 		Properties: a.convertMetadataItemsToProperties(c.Spec.Metadata),
 	})
 	if err != nil {
-		log.Warnf("failed to init secret store %s/%s named %s: %s", c.Spec.Type, c.Spec.Version, c.ObjectMeta.Name, err)
-		diag.DefaultMonitoring.ComponentInitFailed(c.Spec.Type, "init")
-		return err
+		diag.DefaultMonitoring.ComponentInitFailed(c.Spec.Type, "init", c.ObjectMeta.Name)
```
artursouza, line 2557: Are these breaking changes? The metrics are part of the contracts and should be compatible, it means that someone with alerts on the existing metrics should not have to change their alerts when they upgrade to this version.

artursouza (changes requested):

pkg/runtime/initError.go
```diff
@@ -0,0 +1,49 @@
...
+package runtime
+
+import (
+	"fmt"
+)
+
+type InitErrorKind string
+
+const (
+	InitFailure            InitErrorKind = "InitFailure"
```
artursouza, line 23: If we want these to be printed out, we should agree if these should be app caps, camel case or pascal case. Last but not least, we should agree on these error types since these become contractual.

I personally think these should be ALL_CAPS and we can keep these failure types here.

/cc @yaron2

yaron2: I agree on ALL_CAPS. cc @daixiang0

pkg/runtime/runtime.go
addjuarez: The error is log multiple times and will eventually get captured here. https://github.com/dapr/dapr/blob/master/pkg/runtime/runtime.go#L2158

pkg/runtime/runtime.go
addjuarez: The error is log multiple times and will eventually get captured here. https://github.com/dapr/dapr/blob/master/pkg/runtime/runtime.go#L2158
Can add it back but it seemed redundant to print the same message multiple times

pkg/runtime/runtime.go
addjuarez: Answered above

pkg/runtime/runtime.go
addjuarez: These should not be breaking changes, I have validated that our dapr grafana dashboards are able to still recieve this metric and trigger alerts unchanged.

pkg/runtime/initError.go
```diff
@@ -0,0 +1,49 @@
...
+
+type InitError struct {
+	err    error
+	kind   InitErrorKind
+	entity string
+}
+
+func (e *InitError) Error() string {
+	if e.entity == "" {
+		return fmt.Sprintf("(%s): %s", e.kind, e.err)
```
ItalyPaleAle, line 36: Use square brackets for `e.kind` please

artursouza, line 36: "[ ]" is fine.

pkg/runtime/initError.go
```diff
@@ -0,0 +1,49 @@
...
+	}
+
+	return fmt.Sprintf("(%s): initialization error occurred for %s: %s", e.kind, e.entity, e.err)
```
artursouza, line 39: "[ ]" is better.

pkg/runtime/initError.go
daixiang0: I agree on ALL_CAPS as well.

artursouza (approved):

halspang (approved):

Issue #4576, opened 2022-05-02
[Proposal] Workflow building block and engine


johnewart:
## In what area(s)?
/area runtime

## What? 

This document proposes that the Dapr runtime be extended to include a new workflow building block. This building block, in combination with a lightweight, portable, workflow engine will enable developers to express workflows as code that can be executed, interacted with, monitored, and debugged using the Dapr runtime.

## Why?

Many complex business processes are well modeled as a *workflow* - a set of steps needing to be orchestrated that require resiliency and guarantee completion (success or failure are both completions). To build such workflows, developers are often faced with needing to solve a host of complex problems, including (but not limited to):

* Scheduling
* Lifecycle management
* State storage
* Monitoring and debugging
* Resiliency
* Failure handling mechanisms

Based on available data, it is clear that workflows are quite popular; at the time of this writing, the number of daily executions for hosted workflows and their tasks is in the billions per day across tens of thousands of Azure subscriptions. 

## What is a workflow?

A *workflow*, for the purpose of this proposal, is defined as application logic that defines a business process or data flow that:

* Has a specific, pre-defined, deterministic lifecycle (e.g Pending -> Running -> [Completed | Failed | Terminated])
* Is guaranteed to complete 
* Is durable (i.e completion in the face of transient errors)
* Can be scheduled to start or execute steps at or after some future time
* Can be paused and resumed (explicitly or implicitly)
* Can execute portions of the workflow in serial or parallel
* Can be directly addressed by external agents (i.e an instance of the workflow can be interacted with directly - paused, resumed, queried, etc.)
* May be versioned
* May be stateful
* May create new sub-workflows and optionally wait for those to complete before progressing
* May rely on external components to perform its job (i.e HTTPS API calls, pub/sub message queues, etc.)

## Why Dapr?

Dapr already contains many of the building blocks required to provide reliability, scalability, and durability to the execution of workflows. Building such an engine inside Dapr, and providing the necessary building blocks will help to increase developer productivity through re-usability of existing features and independence from the underlying execution mechanism thereby increasing portability.

In addition to the built-in execution engine, Dapr can provide a consistent programming interface for interacting with third-party workflow execution systems (i.e AWS SWF, Apache Camel, Drools) for those who are already using these tools. Thereby providing a standardized interface for working with both external workflows as well as those running inside Dapr. 

# Proposal

## High-level overview of changes 

We propose that the following features / capabilities be added to the Dapr runtime:

* A new "workflow" building block 
[...]

PR #4986, opened 2022-08-09
Initial push for workflow api

RyanLettieri:
# Description

## Issue reference

Related Proposal: https://github.com/dapr/dapr/issues/4576

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing


cgillum:
@RyanLettieri should we have this PR target the `feature/workflows` branch instead of `main`?

pkg/runtime/options.go
```diff
@@ -19,6 +20,7 @@ type (
 		configurations  []configuration.Configuration
 		locks           []lock.Lock
 		pubsubs         []pubsub.PubSub
+		workFlows       []workflows.Workflow
```
cgillum, line 23: nit: "workflow" is one word, so we should use all lowercase letters in our "workflow" variable and function names.

pkg/http/api.go
```diff
@@ -224,6 +233,29 @@ func (a *api) MarkStatusAsOutboundReady() {
...
+			Route:   "workflows/{workflowName}",
+			Version: apiVersionV1alpha1,
+			Handler: a.onStartWorkflow,
+		},
+		{
+			Methods: []string{fasthttp.MethodDelete},
+			Route:   "workflows/{workflowId}/{workflowRunId}",
+			Version: apiVersionV1alpha1,
+			Handler: a.onTerminateWorkflow,
+		},
```
cgillum, line 255: We'll need to update these routes to match the latest iteration of the spec. Some of the variable names are also different in the spec (I assume you're using the Temporal terminology here). Ping me for details.

pkg/http/api.go
```diff
@@ -698,6 +730,70 @@ func (a *api) getLockStoreWithRequestValidation(reqCtx *fasthttp.RequestCtx) (lo
...
+		log.Debug("No workflow, or empty workflow was provided.")
+		return
+	}
+
+	err := json.Unmarshal(reqCtx.PostBody(), &startReq)
+	if err != nil {
+		log.Debug(err)
+	}
+	req := workflows.StartRequest{
+		WorkflowName: startReq.WorkflowName,
```
cgillum, line 747: The workflow name be coming from the URL instead of the request body. I see we grabbed it and saved it in the `wf` variable but aren't actually using it.

RyanLettieri:
> @RyanLettieri should we have this PR target the `feature/workflows` branch instead of `main`?

@cgillum We could, however, there is no associated feature/workflows branch in components-contrib so we may just use 'main' for both

pkg/http/api.go
```diff
@@ -233,23 +233,26 @@ func (a *api) MarkStatusAsOutboundReady() {
...
 			Methods: []string{fasthttp.MethodPost},
-			Route:   "workflows/{workflowName}",
+			Route:   "workflows/{workflowComponent}{workflowType}/{instanceId}",
 			Version: apiVersionV1alpha1,
 			Handler: a.onStartWorkflow,
 		},
 		{
 			Methods: []string{fasthttp.MethodDelete},
-			Route:   "workflows/{workflowId}/{instanceId}",
+			Route:   "workflows/{workflowComponent}{workflowType}/{instanceId}",
```
cgillum, line 255: Does there need to be a slash in between `{workflowComponent}` and `{workflowType}` here, and in the above routes?
```suggestion
			Route:   "workflows/{workflowComponent}/{workflowType}/{instanceId}",
```

pkg/grpc/api.go
```diff
@@ -299,6 +302,7 @@ func NewAPI(
 		appChannel:                 appChannel,
 		pubsubAdapter:              pubsubAdapter,
 		stateStores:                stateStores,
+		workFlows:                  workfFlows,
```
johnewart, line 305: Typo: `workfFlows`, also should it be `workflows`? (or maybe `workflowEngines`?)

pkg/grpc/proxy/codec/codec.go
```diff
@@ -22,7 +22,7 @@ func init() {
 
 // Register manually registers the codec.
 func Register() {
-	encoding.RegisterCodec(codec())
+	// encoding.RegisterCodec(codec())
```
johnewart, line 25: Is this needed or not? If it's needed, why? (I recall mention of a gRPC issue but wasn't sure what the resolution was)

pkg/http/api.go
```diff
@@ -224,6 +233,32 @@ func (a *api) MarkStatusAsOutboundReady() {
...
 }
 
+// Workflow Component: Component specified in yaml (temporal, etc..)
+// Workflow Type: Name of the workflow to run (function name)
+// Instance ID: Identifier of the specific run
+func (a *api) constructWorkflowEndpoints() []Endpoint {
+	return []Endpoint{
+		{
+			Methods: []string{fasthttp.MethodGet},
+			Route:   "workflows/{workflowComponent}/{workflowType}/{instanceId}",
```
johnewart, line 243: Is the `workflowComponent` the actual component (i.e `temporal`, `built-in`, etc.) or is it the _name_ of the workflow engine component? (i.e a "temporal workflow component named `foobar`")

pkg/http/api.go
```diff
@@ -698,6 +733,128 @@ func (a *api) getLockStoreWithRequestValidation(reqCtx *fasthttp.RequestCtx) (lo
...
+func (a *api) onStartWorkflow(reqCtx *fasthttp.RequestCtx) {
+	startReq := wfs.StartRequest{}
+
+	wfType := reqCtx.UserValue(workflowType).(string)
+	if wfType == "" {
+		log.Debug("No workflow, or empty workflow was provided.")
+		return
+	}
+
+	component := reqCtx.UserValue(workflowComponent).(string)
```
johnewart, line 749: Is this guaranteed to be non-nil? (`reqCtx.UserValue(....)`)

pkg/http/api.go
```diff
@@ -698,6 +733,128 @@ func (a *api) getLockStoreWithRequestValidation(reqCtx *fasthttp.RequestCtx) (lo
...
+// Workflow Component: Component specified in yaml (temporal, etc..)
+// Workflow Type: Name of the workflow to run (function name)
+// Instance ID: Identifier of the specific run
+func (a *api) onStartWorkflow(reqCtx *fasthttp.RequestCtx) {
+	startReq := wfs.StartRequest{}
+
+	wfType := reqCtx.UserValue(workflowType).(string)
+	if wfType == "" {
+		log.Debug("No workflow, or empty workflow was provided.")
+		return
```
johnewart, line 746: HTTP 400 / 404 here? (or will it automatically do that? - also same below)

pkg/http/api.go
```diff
@@ -698,6 +733,128 @@ func (a *api) getLockStoreWithRequestValidation(reqCtx *fasthttp.RequestCtx) (lo
...
+	req := workflows.StartRequest{
+		Options:      startReq.Options,
+		WorkflowName: wfType,
+		Parameters:   startReq.Parameters,
+	}
+	req.WorkflowInfo.InstanceId = instance
+
+	resp, err := a.workFlows[component].Start(reqCtx, &req)
+	if err != nil {
+		log.Debug(err)
```
johnewart, line 779: If there's an error, we should `log.Error` it - also we should probably handle it here and not continue, right?

pkg/http/api.go
```diff
@@ -698,6 +733,128 @@ func (a *api) getLockStoreWithRequestValidation(reqCtx *fasthttp.RequestCtx) (lo
...
+	}
+	log.Debug(resp)
+	respond(reqCtx, withJSON(202, response))
+}
+
+func (a *api) onGetWorkflow(reqCtx *fasthttp.RequestCtx) {
+
+	wfName := reqCtx.UserValue(workflowType).(string)
```
johnewart, line 792: Same as above, if it's guaranteed to be non-nil then you can ignore

pkg/http/api.go
```diff
@@ -698,6 +733,128 @@ func (a *api) getLockStoreWithRequestValidation(reqCtx *fasthttp.RequestCtx) (lo
...
+		return
+	}
+
+	req := workflows.WorkflowStruct{
+		InstanceId: instance,
+	}
+
+	err := a.workFlows[component].Terminate(reqCtx, &req)
+	if err != nil {
+		log.Debug(err)
```
johnewart, line 853: Same as above, if it's an error we should log it as such and also handle it gracefully

pkg/runtime/options.go
```diff
@@ -99,3 +101,10 @@ func WithComponentsCallback(componentsCallback ComponentsCallback) Option {
 		o.componentsCallback = componentsCallback
 	}
 }
+
+// WithWorkflows adds workflow components to the runtime.
+func WithWorkflows(workFlows ...workflows.Workflow) Option {
+	return func(o *runtimeOpts) {
+		o.workFlows = append(o.workFlows, workFlows...)
```
johnewart, line 108: `workflow`? (or `workflowEngine`?)

pkg/runtime/runtime.go
```diff
@@ -544,6 +546,7 @@ func (a *DaprRuntime) initRuntime(opts *runtimeOpts) error {
 			OutputBindings:  a.outputBindings,
 			SecretStores:    a.secretStores,
 			PubSubs:         a.pubSubs,
+			WorkFlows:       a.workFlows,
```
johnewart, line 549: `Workflows`?

go.mod
```diff
@@ -386,10 +393,15 @@ replace (
...
 
+replace go.temporal.io/api => go.temporal.io/api v1.8.0
+
+replace go.temporal.io/sdk => go.temporal.io/sdk v1.15.0
+
 // Uncomment for local development for testing with changes in the components-contrib repository.
 // Don't commit with this uncommented!
 //
-// replace github.com/dapr/components-contrib => ../components-contrib
+replace github.com/dapr/components-contrib => ../components-contrib
```
johnewart, line 403: You probably don't want to merge this

go.mod
```diff
@@ -386,10 +393,15 @@ replace (
 
 replace github.com/eclipse/paho.mqtt.golang => github.com/shivamkm07/paho.mqtt.golang v1.3.6-0.20220106130409-e28a1db639f8
 
+replace go.temporal.io/api => go.temporal.io/api v1.8.0
```
johnewart, line 396: Can we say why we need to pin these so it's not a mystery? 😄

johnewart (changes requested):
Looking good, some small nitpicks / error handling comments below and a few things (like the replace local path) that need changing. I only have one question re: naming -- right now we really are registering a _workflow engine_ as opposed to individual workflows, should we consider naming it as such?

pkg/http/api.go
RyanLettieri: This is from the proposal document: The {workflowComponent} segment is the name of a workflow component – i.e., something that would be described in a workflow.yaml file (similar to how state store and pub/sub components have names)

pkg/runtime/runtime.go
RyanLettieri: "workFlows" follows the same naming convention as the items above it

pkg/http/api.go
RyanLettieri: A 404 error is returned

pkg/grpc/api.go
RyanLettieri: Updated name to be workflowEngines

pkg/grpc/proxy/codec/codec.go
RyanLettieri: It does not appear to be needed any longer. It was most likely an error on my part

go.mod
RyanLettieri: This also seemed to be an error on my end. I've removed the replace statement and it appears to work as expected

PR #4990, opened 2022-08-10
Block calls from sidecar to app when app is not healthy - fix 4883

### LINKED ISSUE #4883: Block calls from sidecar to app when app is not healthy
## In what area(s)?

/area runtime

> /area operator

> /area placement

> /area docs

> /area test-and-release

## Describe the feature

Dapr sidecar should proactively ping a health endpoint on the app and block forever any calls into the app channel until app is healthy again. This should be configurable via K8s anotations and sidecar command line args:

`dapr.io/app-readiness-path` - path for health check via HTTP endpoint or method name for gRPC. Example: `/health` or `health`. No default - feature is off if this is not provided.
`dapr.io/app-readiness-probe-period-seconds` - how often (in seconds) to perform the app liveness probe. Defaults to 5.
`dapr.io/app-readiness-probe-period-threshold` - number of consecutive failures for app to be considered not ready. Defaults to 3.

New E2E test scenario:
Validate that bindings, pubsub subscription and service invocation block and do not call into the app's endpoint while it is down.

## Release Note

RELEASE NOTE:  **ADD** support for app's channel readiness probe from sidecar.

ItalyPaleAle:
# Description

This PR adds a new apphealth module which is used to probe for the health of the app. The feature is disabled by default.

As described in #4883, when the feature is enabled, Dapr periodically polls the app at a given endpoint (HTTP or gRPC, depending on app-protocol). When probes fail for more than a certain threshold in a row, Dapr stops all subscriptions (pubsub topics and input bindings) and short-circuits service invocation calls (that are rejected before attempting to invoke the app).

This PR adds a few new CLI options for daprd:

- `enable-app-health-check` (annotation: `dapr.io/enable-app-health-check`): Boolean that enables the health-checks (it's `false` by default, so no health checks are performed)
- `app-health-check-path` (annotation: `dapr.io/app-health-check-path`): String, used only when app-channel is HTTP. This is the path that Dapr invokes for health probes (default: `/health`).
- `app-health-probe-interval` (annotation: `dapr.io/app-health-probe-interval`): Integer with number of seconds between each health probe (default: 5)
- `app-health-probe-timeout` (annotation: `dapr.io/app-health-probe-timeout`): Integer with number of milliseconds for timing out health probe requests (default: 500)
- `app-health-threshold` (annotation: `dapr.io/app-health-threshold"`): Integer with max number of consecutive probe failures before the app is considered unhealthy

_Almost_ everything in #4883 was included. The missing thing was being able to use successful requests from Dapr to the app as passed health probes (e.g. a network failure there would be considered the same as a failed health probe). Support for that is included in the apphealth module, but I have not (yet) implemented it because this PR was already large enough, and after initial investigation, that feature requires a lot more work and care (because signals would have to be sent in multiple packages, and with resiliency still optional there are lots of code paths to consider). If that is a feature that's strongly desired, I can revise that in the future.

This PR also includes E2E tests to validate the behavior.

## Issue reference

Fixes #4883

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests
* [X] Unit tests passing
* [X] Extended the documentation https://github.com/dapr/docs/issues/2734


pkg/apphealth/health_test.go
```diff
@@ -0,0 +1,179 @@
...
+			}
+		}
+	}
+
+	// Trigger threshold+10 failures and detect the invocation after the threshold
+	simulateFailures(threshold + 10)
+	assert.Empty(t, unexpectedStatusChanges.Load())
+	assert.Equal(t, threshold+10, h.failureCount.Load())
+
+	// First success should bring the app back to healthy
```
mukundansundar, line 71: @ItalyPaleAle Since the health probe waits for a threshold to be passed before considering the app is unhealthy, should there be a threshold also set to consider the app to be health after an unhealthy state? Can't there be cases where the app become healthy (health endpoint works) and immediately becomes unhealthy (probably restart of container, lazy load init issue etc)?

In these scenarios should we consider a threshold of successful health checks before we change status to healthy? 
cc @yaron2 @artursouza WDYT?

ItalyPaleAle: A threshold ≥ 1 is common to detect failures, as it can happen that apps are just busy. Failure thresholds are particularly useful when there's a network involved as the network is unstable; that's not a big concern in this case as sidecar and app are in the same pod,  so perhaps a lower default threshold (2 or even 1) may even be an option.

A success threshold is a thing, but I don't believe it's a very common pattern. In the situations you describe, it sounds like the health endpoint would be implemented incorrectly, and the app should not report itself as healthy until all init work is performed.

go.mod
```diff
@@ -8,7 +8,7 @@ require (
 	github.com/PuerkitoBio/purell v1.1.1
 	github.com/agrea/ptr v0.0.0-20180711073057-77a518d99b7b
 	github.com/cenkalti/backoff/v4 v4.1.3
-	github.com/dapr/components-contrib v1.8.0-rc.1.0.20220808165304-1148776d370d
+	github.com/dapr/components-contrib v1.8.0-rc.1.0.20220810215246-f2ac0cda69ec
```
ItalyPaleAle, line 11: On reason for updating components-contrib: needed a fix in the cron component for the E2E tests

dapr/proto/runtime/v1/appcallback.proto
```diff
@@ -46,6 +46,13 @@ service AppCallback {
   rpc OnBindingEvent(BindingEventRequest) returns (BindingEventResponse) {}
 }
 
+// AppCallbackHealthCheck V1 is an optional extension to AppCallback V1 to implement
+// the HealthCheck method.
+service AppCallbackHealthCheck {
```
ItalyPaleAle, line 51: This is implemented as a separate service so it's optional for clients and preserves backwards-compatibility

pkg/apphealth/health.go
```diff
@@ -0,0 +1,225 @@
...
+	select {
+	case h.queue <- struct{}{}:
+		// Do nothing
+	default:
+		// Do nothing
+	}
+}
+
+// ReportHealth is used by the runtime to report a health signal from the app.
+func (h *AppHealth) ReportHealth(status uint8) {
```
ItalyPaleAle, line 121: This method is currently unused. It was implemented to support using other signals to report health, but the downstream implementation is not (yet) done. Nevertheless, I don't think there's harm in leaving this here, as it also comes with unit tests

pkg/channel/grpc/grpc_channel.go
```diff
@@ -52,7 +56,7 @@ func CreateLocalChannel(port, maxConcurrency int, conn *grpc.ClientConn, spec co
 		readBufferSize:     readBufferSize,
 	}
 	if maxConcurrency > 0 {
-		c.ch = make(chan int, maxConcurrency)
+		c.ch = make(chan struct{}, maxConcurrency)
```
ItalyPaleAle, line 59: See: https://hackthedeveloper.com/golang-empty-struct-channel/

pkg/runtime/config.go
```diff
@@ -77,46 +82,91 @@ type Config struct {
...
 }
 
 // NewRuntimeConfig returns a new runtime config.
-func NewRuntimeConfig(
-	id string, placementAddresses []string, controlPlaneAddress, allowedOrigins, globalConfig, componentsPath,
-	appProtocol, mode string, httpPort, internalGRPCPort, apiGRPCPort int, apiListenAddresses []string, publicPort *int,
-	appPort, profilePort int, enableProfiling bool, maxConcurrency int, mtlsEnabled bool, sentryAddress string, appSSL bool,
-	maxRequestBodySize int, unixDomainSocket string, readBufferSize int, streamRequestBody bool, gracefulShutdownDuration time.Duration, enableAPILogging bool, disableBuiltinK8sSecretStore bool,
-) *Config {
+func NewRuntimeConfig(opts NewRuntimeConfigOpts) *Config {
```
ItalyPaleAle, line 126: Why this change: the function already had almost 30 parameters, which were hard to maintain ("what is this 'false' for?". And I had to add 5 more...

pkg/apphealth/health_test.go
```diff
@@ -0,0 +1,182 @@
+/*
+Copyright 2021 The Dapr Authors
```
artursouza, line 2: nit: 2022 everywhere

pkg/channel/grpc/grpc_channel.go
```diff
@@ -69,6 +73,10 @@ func (g *Channel) GetAppConfig() (*config.ApplicationConfig, error) {
 
 // InvokeMethod invokes user code via gRPC.
 func (g *Channel) InvokeMethod(ctx context.Context, req *invokev1.InvokeMethodRequest) (*invokev1.InvokeMethodResponse, error) {
+	if g.appHealth != nil && g.appHealth.GetStatus() != apphealth.AppStatusHealthy {
```
artursouza, line 76: Can we default the app channel to always be healthy instead? This way we don't need to check if appHealth is not nil.

tests/platforms/kubernetes/kubeobj.go
```diff
@@ -107,7 +107,21 @@ func buildDaprAnnotations(appDesc AppDescription) map[string]string {
 	if appDesc.DaprEnv != "" {
 		annotationObject["dapr.io/env"] = appDesc.DaprEnv
 	}
-
+	if appDesc.EnableAppHealthCheck {
+		annotationObject["dapr.io/enable-app-health-check"] = "true"
```
artursouza, line 111: I think we need a preview feature flag for user's awareness that we can change this behavior still.

pkg/channel/grpc/grpc_channel.go
ItalyPaleAle: The nil check is to avoid instantiating the `appHealth` object if not needed.

ItalyPaleAle:
@artursouza added a feature flag `AppHealthCheck` which needs to be enabled for this to work (will need to be put in docs).

If the feature flag is not set and the user adds `--enable-app-health-check`, health checks are disabled and this is shown in the logs: 

> App health checks are a preview feature and require the AppHealthCheck feature flag to be enabled. See https://docs.dapr.io/operations/configuration/preview-features/ on how to enable preview features.

artursouza (approved):

Issue #4962, opened 2022-08-02
E2E tests: connect to AKS using a Wireguard tunnel


ItalyPaleAle:
## In what area(s)?

/area test-and-release

## Describe the proposal

One of the leading causes of E2E test failures (on AKS) continues to be network timeouts. These are commonly reported in the logs with errors such as "context deadline exceeded".

The issue, as [confirmed by the Actions team](https://github.com/actions/virtual-environments/issues/5850), is that we are making too many outbound connections from the Actions runner to the Internet (especially, to the same few target IPs), in a short amount of time, and we are frequently exhausting our SNAT ports in the Actions runners (because they are behind a load balancer themselves).

Over the last few months, we have made some improvements so that our E2E test apps try to re-use TCP sockets as much as possible, thus limiting the number of outbound connections. Nevertheless, per the TCP dumps the Actions team analyzed, we are still making thousands of outbound connections, and there's not much more we can do to optimize on our end.

A few months ago (before we root-caused the issue to SNAT port exhaustion), I wrote [a proposal](https://github.com/dapr/dapr/issues/4447) for addressing the problem by moving the test controller (i.e. the code that is executed when running `make e2e-test-all`) into the Kubernetes cluster itself. I do believe this would solve the issue; however, it does require some significant re-architecting of the E2E test infrastructure. While running tests from within the AKS cluster is not impossible, coordinating the work with the Actions runner so that logs and errors are correctly reported requires careful consideration.

An alternative approach which I also floated with the Actions team in the issue linked above could be to use a tunnel, for example a Wireguard one, to maintain a connection between the Actions runner and the AKS cluster. In fact, if the problem is that we are making too many outbound connections, using a tunnel would make it so we'd have a grand total of.... 1 connection only!

Wireguard seems to be a solid protocol for implementing a tunnel because it can run in the userspace and can thus be executed directly into the AKS cluster, so no new infrastructure is needed, and it allows accessing all services in the cluster by connecting with their local IPs (bonus: tests would be faster because we don't need to wait for a service to get a LoadBalancer endpoint!). An example is here: https://github.com/ivanmorenoj/k8s-wireguard 

The idea would be:

- Change the Actions runner to create a Wireguard tunnel with the AKS cluster
  - This requires deploying a Wireguard server as K8s workload before starting the tests
[...]

PR #4991, opened 2022-08-10
[WIP] Run e2e tests using a VPN Tunnel

mcandeia:
# Description

Now, the connection between the actions runner and the AKS clusters is over a VPN tunnel using [tailscale](https://tailscale.com/). To achieve that was required basically two main changes:

- Our AKS cluster is now configured using a tailscale basic subnet router configuration (allowing action runner to call internal IPs directly)
- The actions runner is connecting to tailscale using an Ephemeral Key (so that, when the actions runner finish its work, the node is basically discarded)
- The app manager now exposes the app private IP rather than external loadbalancer one.
## Issue reference

Please reference the issue this PR will close: #4962 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Unit tests passing

### DIFF
--- a/tests/platforms/kubernetes/app_description.go
+++ b/tests/platforms/kubernetes/app_description.go
@@ -15,8 +15,15 @@ package kubernetes
 
 import (
 	"encoding/json"
+	"os"
+	"strings"
 
 	apiv1 "k8s.io/api/core/v1"
+)
+
+const (
+	// useServiceInternalIP is used to identify wether the connection between the kubernetes platform could be made using its internal ips.
+	useServiceInternalIP = "TEST_E2E_USE_INTERNAL_IP"
 )
 
 // AppDescription holds the deployment information of test app.
@@ -60,6 +67,11 @@ func (a AppDescription) String() string {
 	// This method overrides the default stringifier to use the custom JSON stringifier which hides ImageSecret
 	j, _ := json.Marshal(a)
 	return string(j)
+}
+
+// ShouldBeExposed returns if the app should be exposed as a loadbalancer/nodeport service.
+func (a AppDescription) ShouldBeExposed() bool {
+	return a.IngressEnabled && !strings.EqualFold(os.Getenv(useServiceInternalIP), "true")
 }
 
 func (a AppDescription) MarshalJSON() ([]byte, error) {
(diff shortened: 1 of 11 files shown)


ItalyPaleAle:
I love that as a side effect to this, the tests are also quite a bit faster! Hopefully you can sort out the last hurdles :)

mcandeia:
We have faced some limitations using tailscale related to the solution maintainability. Unfortunately, tailscale does not support an automatic way to rotate client keys, and the client keys has only 1-month expiry.

Things that we should consider when back to this in the future:

- Auth Keys rotation
- Only 1 subnet router is allowed for free plan
- Cluster CIDR isolation (we should avoid using the same CIDR for parallel tests running)
- ACL should be configured for the generate keys. Use autoApprovers feature (https://tailscale.com/kb/1018/acls/#auto-approvers-for-routes-and-exit-nodes)

**I'm putting this PR on hold until we have a way to rotate auth keys.**

[issue](https://github.com/dapr/dapr/issues/4962)

tests/platforms/kubernetes/app_description.go
```diff
@@ -62,6 +69,11 @@ func (a AppDescription) String() string {
 	return string(j)
 }
 
+// ShouldBeExposed returns if the app should be exposed as a loadbalancer/nodeport service.
+func (a AppDescription) ShouldBeExposed() bool {
+	return a.IngressEnabled && !strings.EqualFold(os.Getenv(useServiceInternalIP), "true")
```
ItalyPaleAle, line 74: Nit: rather than checking for value `true`, please use `utils.IsTruthy` which checks for other allowed values too (like `1`, `y`, `yes`...)

ItalyPaleAle:
While this can't be merged yet as-is today because of a missing feature that we need in our CI, do you think you could split out the change to the workflow file and then we could merge this PR regardless?

Although that won't make our CI better yet, I for one would LOVE to have this as it makes running E2E tests manually (and developing them) much, much faster (just not having to wait for the load balancer would be awesome).

(PS: don't forget to add some docs here too please: https://github.com/dapr/dapr/blob/master/tests/docs/running-e2e-test.md )

tests/platforms/kubernetes/app_description.go
mcandeia: fixed!

ItalyPaleAle:
@mcandeia why did you close this?

PR #5002, opened 2022-08-11
Fixed: send messages to DLQ when topic name has wildcard

### LINKED ISSUE #5000: Dead letter queue does not work when topic contains a wildcard
## In what area(s)?

/area runtime

## What version of Dapr?

> 1.8.3
> master

## Issue

When an app subscribes to a topic using a wildcard (something that can be done for example with `pubsub.mqtt` by subscribing to `foo/#`), failed messages are not sent to the DLQ.

The problem is here:

https://github.com/dapr/dapr/blob/e8080af869f56d78ca44e8fe4603d2e41c4d9ee7/pkg/runtime/runtime.go#L610

This checks for the DLQ topic by using `msg.Topic`. However the topic used to subscribe contains the wildcard (e.g. `foo/#`), while the topic in `msg.Topic` is the "actual" topic (e.g. `foo/3`).

## Release Note

RELEASE NOTE: **FIX** Failed pubsub messages are not sent to DLQ if the topic subscription has a wildcard

ItalyPaleAle:
# Description

Fixes #5000 by ensuring that messages received on a subscription which uses a wildcard can be sent to the DLQ if configured.

## Issue reference

Fixes #5000


pkg/runtime/runtime.go
```diff
@@ -630,11 +628,12 @@ func (a *DaprRuntime) sendToDeadLetterIfConfigured(name string, msg *pubsub.NewM
 	err = a.Publish(req)
 	if err != nil {
 		log.Errorf("error sending message to dead letter, origin topic: %s dead letter topic %s err: %w", msg.Topic, deadLetterTopic, err)
+		return err
 	}
-	return true, err
+	return nil
```
mcandeia, line 633: nitpicking:

How about removing the `return err` from line 631? instead `return err` on line 633 directly, the behavior is pretty much the same

pkg/runtime/runtime.go
```diff
@@ -140,8 +140,11 @@ type Route struct {
 	rules    []*runtime_pubsub.Rule
 }
 
+type TopicRoutes map[string]TopicRoute
```
mcandeia, line 143: nitpicking:

Generally speaking, we read the code from top lines to bottom lines, given that this line is using the TopicRoute type (that's defined below) I think it should be better placed right after the `TopicRoute` type,

just nitpicking

pkg/runtime/runtime.go
```diff
@@ -140,8 +140,11 @@ type Route struct {
...
+
 type TopicRoute struct {
-	routes map[string]Route
+	routes          Route
```
mcandeia, line 146: just a comment:

this annoys me a bit. TopicRoute contains route**s** that are not actually a collection,

but it could be my lack of knowledge

pkg/runtime/runtime.go
```diff
@@ -655,21 +654,23 @@ func (a *DaprRuntime) subscribeTopic(parentCtx context.Context, name string, top
...
-			if configured, dlqErr := a.sendToDeadLetterIfConfigured(name, msg); configured && dlqErr == nil {
-				// dlq has been configured and message is successfully sent to dlq.
-				diag.DefaultComponentMonitoring.PubsubIngressEvent(ctx, pubsubName, strings.ToLower(string(pubsub.Drop)), msg.Topic, 0)
-				return nil
+			if route.deadLetterTopic != "" {
+				if dlqErr := a.sendToDeadLetter(name, msg, route.deadLetterTopic); dlqErr == nil {
+					// dlq has been configured and message is successfully sent to dlq.
+					diag.DefaultComponentMonitoring.PubsubIngressEvent(ctx, pubsubName, strings.ToLower(string(pubsub.Drop)), msg.Topic, 0)
+					return nil
+				}
```
mcandeia, line 673: If I'm not wrong, this code is pretty much the same as lines 686, 700, and 725,

do you think it worth extracting it to an internal func and reusing it instead? it could be a long-named function like

sendDeadLetter[andPublishIngressEvent]

pkg/runtime/runtime.go
```diff
@@ -1615,27 +1626,33 @@ func (a *DaprRuntime) getTopicRoutes() (map[string]TopicRoute, error) {
...
-			for topic := range v.routes {
-				topics = append(topics, topic)
+			var topics string
+			for topic := range v {
+				if topics == "" {
+					topics += topic
+				} else {
+					topics += " " + topic
+				}
 			}
```
mcandeia, line 1651: [strings.Join](https://pkg.go.dev/strings#Join) might help here

pkg/runtime/runtime.go
ItalyPaleAle: Yeah on this and the next comment: I agree naming can be improved. I'll change

pkg/runtime/runtime.go
ItalyPaleAle: Sadly we cannot do that. We could if Go supported macros...

The reason is that in line 672, we want to return from the entire function, but ONLY if sending to the DLQ did not error. Otherwise, we want to continue execution and hit the return on line 676

mcandeia: how about having a flag for it? you think it's a good idea?

like

if ok := a.trySendToDeadLetter(...); ok {
    return nil
}

pkg/runtime/runtime.go
ItalyPaleAle: You're totally rightYou're right that it would be the same. The reason why I don't do that, in general, is because, especially on a project like Dapr with lots of contributors, it can help prevent future regressions :)

Example, following your suggestion:

```go
err = a.Publish(req)
if err != nil {
	log.Errorf("error sending message to dead letter, origin topic: %s dead letter topic %s err: %w", msg.Topic, deadLetterTopic, err)
}
return err
```

Now, if someone wants to add more code before returning, they will need to re-add a `return err` within the block.

```go
err = a.Publish(req)
if err != nil {
	log.Errorf("error sending message to dead letter, origin topic: %s dead letter topic %s err: %w", msg.Topic, deadLetterTopic, err)
	return err
}
// Do something else
return nil
```

I've seen it happen before many times where people forget the `return err`. So as a good practice I try to prevent these and save others (and myself!) from potentially making a mistake in the future.

pkg/runtime/runtime.go
ItalyPaleAle: Good callout. I actually think that this can be flattened without having `route` as a struct.

artursouza (commented):
We had issues with wildcard topic names before. Do we have E2E tests for this?

ItalyPaleAle:
We do not. Only certain pubsub components support wildcards (MQTT and another one which I can’t remember right now), and those are not present in the test cluster. Both are non-stable components too.

I can see if perhaps this can be done with some unit test in the runtime that mocks the components. (If that’s ok, I’d do a separate PR for that)

ItalyPaleAle:
Thinking about it, one way it could be done in E2E would be leveraging the in-memory pubsub component. It would need to be modified to support wildcards, but it would be easier than spinning up more infrastructure. 

https://github.com/dapr/components-contrib/issues/1964

ItalyPaleAle:
Update: I will have an E2E test ready in a separate PR shortly.

pkg/runtime/runtime.go
ItalyPaleAle: Note that I've refactored this again in a new PR that's almost ready

yaron2 (approved):

PR #5004, opened 2022-08-12
copy redis v6 image from dockerhub to ghcr

mukundansundar:
Signed-off-by: Mukundan Sundararajan <65565396+mukundansundar@users.noreply.github.com>

# Description

WRT to the change made in CLI to restrict Redis version to compatible version 6, this PR adds ability to copy version 6 of redid as a separate tag into GHCR. 
https://github.com/dapr/cli/pull/1046

Till this PR is merged and workflow run, `dapr init` with GHCR will fail. 

## Issue reference

Please reference the issue this PR will close: CLI hotfix
## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing

### DIFF
--- a/.github/workflows/dapr-3rdparty-images.yaml
+++ b/.github/workflows/dapr-3rdparty-images.yaml
@@ -46,6 +46,13 @@ jobs:
           src: redis:latest
           dst: |
             ghcr.io/${{ github.repository_owner }}/3rdparty/redis:latest
+      - # copy 3rd party redis v6 image from dockerhub to ghcr
+        name: Push Image (redis) to target registries
+        uses: akhilerm/tag-push-action@v2.0.0
+        with:
+          src: redis:6
+          dst: |
+            ghcr.io/${{ github.repository_owner }}/3rdparty/redis:6
       - # copy 3rd party rejson image from dockerhub to ghcr
         name: Push Image (rejson) to target registries
         uses: akhilerm/tag-push-action@v2.0.0


yaron2:
This PR is wrong. Currently, the GHCR image remains unchanged as Redis, not redis:6. What this flow needs to do is push Redis 6 to the : latest tag of GHCR

mukundansundar:
> This PR is wrong. Currently, the GHCR image remains unchanged as Redis, not redis:6. What this flow needs to do is push Redis 6 to the : latest tag of GHCR

I do not think that `redid:6` should be tagged as `redis:latest`. 
If that's the case, my suggestion would be to change GHCR also in CLI and have it point to the `redis:6` image instead of latest similar to dockerHub. (This would require 1.8.2 in CLI)
@yaron2 WDYT?

mukundansundar:
> > > This PR is wrong. Currently, the GHCR image remains unchanged as Redis, not redis:6. What this flow needs to do is push Redis 6 to the : latest tag of GHCR
> > 
> > 
> > I do not think that `redid:6` should be tagged as `redis:latest`. If that's the case, my suggestion would be to change GHCR also in CLI and have it point to the `redis:6` image instead of latest similar to dockerHub. (This would require 1.8.2 in CLI) @yaron2 WDYT?
> 
> That's fine, but doesn't neccesitate 1.8.2 as GHCR is optional and we have data that shows an incredibly low usage for these images.

I suggest we do merge this PR as it is since it will at least be used for next version of CLI (GHCR, if not for 1.8.2), till the compatibility with v7 Redis is fixed.

yaron2 (approved):

PR #5008, opened 2022-08-12
Updated list of gRPC endpoints

### LINKED ISSUE #5007: `pkg/grpc/endpoints.go` needs to be updated to the last APIs
## In what area(s)?

/area runtime

## What version of Dapr?

> 1.8.3
> master branch

## Expected Behavior

API allowlists work as expected.

## Actual Behavior

API allowlists are currently not working as expected with gRPC because the file [`pkg/grpc/endpoints.go`](https://github.com/dapr/dapr/blob/master/pkg/grpc/endpoints.go) is not up-to-date.

A user reported an issue with being unable to use the State Query API via gRPC, which seems to be due to the fact that the endpoint includes `"/dapr.proto.runtime.v1.Dapr/QueryState"` and that's an old name (should have been `"/dapr.proto.runtime.v1.Dapr/QueryStateAlpha1"`.

These other APIs are missing:

- `configuration.v1`: (all APIs)
   - `/dapr.proto.runtime.v1.Dapr/GetConfigurationAlpha1`
   - `/dapr.proto.runtime.v1.Dapr/SubscribeConfigurationAlpha1`
   - `/dapr.proto.runtime.v1.Dapr/UnsubscribeConfigurationAlpha1`
- `lock.v1` (all APIs):
   - `/dapr.proto.runtime.v1.Dapr/TryLockAlpha1`
- `unlock.v1` (all APIs):
   - `/dapr.proto.runtime.v1.Dapr/UnlockAlpha1`

To catch issues like these in the future, we should add some tests to ensure that the endpoint list matches the gRPC endpoints (that's not an issue for HTTP).

## Release Note

RELEASE NOTE: **FIX** Added missing endpoints for API allowlist

ItalyPaleAle:
# Description

Making sure the list is up-to-date with the methods that Dapr actually offers.

Also added tests to ensure there won't be a drift again in the future

PS: with the occasion, also changed the unary interceptor to avoid re-building the allowlist every time (saving many allocations on every invocation)

## Issue reference

Fixes #5007

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests
* [X] Unit tests passing

### DIFF
--- a/pkg/grpc/endpoints_test.go
+++ b/pkg/grpc/endpoints_test.go
@@ -15,20 +15,44 @@ package grpc
 
 import (
 	"context"
+	"sort"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
 	"google.golang.org/grpc"
 
 	"github.com/dapr/dapr/pkg/config"
+	runtimev1pb "github.com/dapr/dapr/pkg/proto/runtime/v1"
 )
+
+func TestEndpointCompleteness(t *testing.T) {
+	// Get the list of endpoints in the runtime
+	runtimeEndpoints := []string{}
+	prefix := "/" + runtimev1pb.Dapr_ServiceDesc.ServiceName + "/"
+	for _, m := range runtimev1pb.Dapr_ServiceDesc.Methods {
+		runtimeEndpoints = append(runtimeEndpoints, prefix+m.MethodName)
+	}
+	for _, m := range runtimev1pb.Dapr_ServiceDesc.Streams {
+		runtimeEndpoints = append(runtimeEndpoints, prefix+m.StreamName)
+	}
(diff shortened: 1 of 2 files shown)


pkg/grpc/endpoints_test.go
```diff
@@ -15,20 +15,44 @@ package grpc
...
 	"testing"
 
 	"github.com/stretchr/testify/assert"
 	"google.golang.org/grpc"
 
 	"github.com/dapr/dapr/pkg/config"
+	runtimev1pb "github.com/dapr/dapr/pkg/proto/runtime/v1"
 )
 
+func TestEndpointCompleteness(t *testing.T) {
```
yaron2, line 28: That's a very good test!

artursouza (approved):

PR #5009, opened 2022-08-12
Use latest k8s for local kind

berndverst:

# Description

The local kind configurations were using a super outdated Kubernetes. This updated everything to the latest Kubernetes (assuming kind 0.14.0)

This only impacts running `make setup-kind`

### DIFF
--- a/tests/test-infra/azure-aks.bicep
+++ b/tests/test-infra/azure-aks.bicep
@@ -37,7 +37,7 @@ param diagStorageResourceId string = ''
 var osDiskSizeGB = 0
 
 // Version of Kubernetes
-var kubernetesVersion = '1.24.0'
+var kubernetesVersion = '1.22.6'
 
 resource containerRegistry 'Microsoft.ContainerRegistry/registries@2019-05-01' = {
   name: '${namePrefix}acr'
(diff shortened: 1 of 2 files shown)


tests/test-infra/azure-aks.bicep
```diff
@@ -37,7 +37,7 @@ param diagStorageResourceId string = ''
 var osDiskSizeGB = 0
 
 // Version of Kubernetes
-var kubernetesVersion = '1.22.6'
```
ItalyPaleAle, line 40: Can we keep it on 1.22.6 for now? That's what our partner teams are currently using in prod.

yaron2 (dismissed):

tests/test-infra/azure-aks.bicep
berndverst: These partner teams really should upgrade their Kubernetes version :)

I suppose we can keep this, but we constantly tend to fall behind in versions, so it's good to update this to the latest AKS version every now and then. We probably should have a separate approach for the partner teams going forward.

tests/test-infra/azure-aks.bicep
```diff
@@ -37,7 +37,7 @@ param diagStorageResourceId string = ''
...
+var kubernetesVersion = '1.24.0'
```
berndverst, line 40:
```suggestion
var kubernetesVersion = '1.22.6'
```

artursouza (approved):

PR #5018, opened 2022-08-16
Return Active Actors count in gRPC metadata API

berndverst:

# Description

Even though the actors API is http only, the Metadata API can provide information about hosted actors via gRPC. This change includes the missing information about active actors.

Note that some SDKs such as Python generally query APIs via gRPC and hence this implementation is necessary.

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Unit tests passing

### DIFF
--- a/pkg/grpc/api_test.go
+++ b/pkg/grpc/api_test.go
@@ -48,6 +48,7 @@ import (
 	"github.com/dapr/components-contrib/pubsub"
 	"github.com/dapr/components-contrib/secretstores"
 	"github.com/dapr/components-contrib/state"
+	"github.com/dapr/dapr/pkg/actors"
 	components_v1alpha "github.com/dapr/dapr/pkg/apis/components/v1alpha1"
 	"github.com/dapr/dapr/pkg/apis/resiliency/v1alpha1"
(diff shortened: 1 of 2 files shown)


berndverst:
cc @halspang @artursouza @tmacam

pkg/grpc/api_test.go
```diff
@@ -1,3 +1,6 @@
+//go:build unit
+// +build unit
```
halspang, line 2: I think I may have been mistaken when I said you needed this here. I think you just needed the `-tags=unit` in the command to run the test.

I don't think there's a problem with having it, but we should see if we need it.

halspang (commented):
Code looks fine to me, just a question about whether we need the thing I told you to add offline :)

pkg/grpc/api_test.go
berndverst: confirmed not needed

halspang (approved):

tmacam:
LGTM

berndverst:
This always fails on a different Kind test. The tests are flaky but that has nothing to do with this PR :)
And Fossa is flakey as everyone knows. Their API seems to be having trouble again.

PR #5022, opened 2022-08-17
Remove goroutine in ComponentUpdate to avoid data race

### LINKED ISSUE #5016: gRPC `SendMsg` should not be called in different goroutines when using streams
## In what area(s)?

/area operator

## What version of Dapr?

1.8.3

## Expected Behavior

No data race when using grpc streams.

## Actual Behavior
Currently the component is sent to sidecar by multiple goroutines when it is updated:
https://github.com/dapr/dapr/blob/327900f9d95747a1421bec44d899fae7e484916b/pkg/operator/api/api.go?_pjax=%23js-repo-pjax-container%2C%20div%5Bitemtype%3D%22http%3A%2F%2Fschema.org%2FSoftwareSourceCode%22%5D%20main%2C%20%5Bdata-pjax-container%5D#L345

However, the grpc `SendMsg` or `RecvMsg` should not be used in different goroutines on the same stream, which is described as follows in [grpc doc](https://github.com/grpc/grpc-go/blob/master/Documentation/concurrency.md#streams): 
> When using streams, one must take care to avoid calling either SendMsg or RecvMsg multiple times against the same [Stream](https://godoc.org/google.golang.org/grpc#Stream) from different goroutines. In other words, it's safe to have a goroutine calling SendMsg and another goroutine calling RecvMsg on the same stream at the same time. But it is not safe to call SendMsg on the same stream in different goroutines, or to call RecvMsg on the same stream in different goroutines.

## Release Note

RELEASE NOTE: FIX data race in component update.

sunzhaochang:
Signed-off-by: sunzhaochang <zhchsun1992@gmail.com>

# Description

## Issue reference

Please reference the issue this PR will close: #5016 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Unit tests passing


yaron2:
@daixiang0 can you please review?

pkg/operator/api/api.go
```diff
@@ -342,33 +341,7 @@ func (a *apiServer) ComponentUpdate(in *operatorv1pb.ComponentUpdateRequest, srv
 			if !ok {
 				return nil
 			}
-			go updateComponentFunc(c)
+			updateComponentFunc(c)
```
daixiang0, line 344: Could we add a test to cover this case?

sunzhaochang: added a test case

pkg/operator/api/api_test.go
```diff
@@ -260,6 +251,50 @@ func TestComponentUpdate(t *testing.T) {
...
+		mockSidecar := &mockComponentUpdateServer{}
+		api := NewAPIServer(client).(*apiServer)
+		updateIterations := 100
+		go func() {
+			// Send a component update, give sidecar time to register
+			time.Sleep(time.Millisecond * 500)
+
+			for _, connUpdateChan := range api.allConnUpdateChan {
+				for i := 0; i < updateIterations; i++ {
+					connUpdateChan <- &c
```
daixiang0, line 282: Seems that it updates sidecar many times with the same data in a loop, which is not a race case I think.

sunzhaochang: The data race occurs in `updateComponentFunc` function which receives component from connUpdateChan and sends it to sidecar. So if we send component to connUpdateChan many times in a loop, `updateComponentFunc` will run in different goroutines and then lead to race condition. It does not matter if the component data sent to connUpdateChan is same or not, what do you think?

daixiang0 (dismissed):

tthophan (approved):

Issue #5023, opened 2022-08-17
Support Dapr deployments on ACI virtual kubelets


shubham1172:
## In what area(s)?

/area runtime

> /area operator

> /area placement

> /area docs

> /area test-and-release

## Describe the feature

**Background**
With https://github.com/dapr/dapr/pull/4741 we modified a behavior for our k8s deployment. 
- Originally, both Command and Args were used, where Command would be `daprd` or `dlv` and Args would be the appropriate arguments. 
- After the above PR, Command is no longer used. The Args now contain `daprd` or `dlv` followed by other arguments. This allows using custom ENTRYPOINT in Docker images. For example, the Windows image now uses a custom ENTRYPOINT `setup-certificates.cmd` that installs certificates in the Windows container if necessary. For more, see [docs](https://kubernetes.io/docs/tasks/inject-data-application/define-command-argument-container/#running-a-command-in-a-shell).

**Problem**
This broke the ACI virtual-kubelet deployments. Root cause of the issue can be found here https://github.com/virtual-kubelet/azure-aci/issues/10. It is because of the way ACI API is structured, it expects Command to be always set. 

Error message:
> Type Reason Age From Message
Normal Scheduled 10s default-scheduler Successfully assigned backend-test/nginx-deployment-67f6b7c46-qvbnp to virtual-node-aci-linux
Warning ProviderCreateFailed 4s (x10 over 10s) virtual-node-aci-linux/pod-controller ACI does not support providing args without specifying the command. Please supply both command and args to the pod spec.

For more details on the error, see the comments in #4741.

**Solution**
We can add a special handling for ACI virtual-kubelet, and have Dapr specify both Command and Args. Note, this will cause ENTRYPOINTs in Docker images to no longer work and will be a limitation. The immediate side-effect will be that [certificate installation](https://docs.dapr.io/operations/configuration/install-certificates/) won't be allowed on Windows nodes in ACI virtual-kubelet.

## Release Note

RELEASE NOTE: **ADD** Support Dapr deployment in ACI virtual-kubelet.

ItalyPaleAle:
> The immediate side-effect will be that [certificate installation](https://docs.dapr.io/operations/configuration/install-certificates/) won't be allowed on Windows nodes in ACI virtual-kubelet.

[...]

PR #5024, opened 2022-08-17
[WIP] Support Dapr deployments on ACI virtual kubelets

shubham1172:
# Description

TODO

## Issue reference

Please reference the issue this PR will close: #5023

## Checklist

* [X] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing


pkg/injector/pod_patch.go
```diff
@@ -613,7 +613,7 @@ func getPullPolicy(pullPolicy string) corev1.PullPolicy {
 	}
 }
 
-func getSidecarContainer(annotations map[string]string, id, daprSidecarImage, imagePullPolicy, namespace, controlPlaneAddress, placementServiceAddress string, socketVolumeMount, tokenVolumeMount *corev1.VolumeMount, volumeMounts []corev1.VolumeMount, trustAnchors, certChain, certKey, sentryAddress string, mtlsEnabled bool, identity string) (*corev1.Container, error) {
+func getSidecarContainer(annotations map[string]string, id, daprSidecarImage, imagePullPolicy, namespace, controlPlaneAddress, placementServiceAddress string, socketVolumeMount, tokenVolumeMount *corev1.VolumeMount, volumeMounts []corev1.VolumeMount, trustAnchors, certChain, certKey, sentryAddress string, mtlsEnabled bool, identity string, tolerations []corev1.Toleration) (*corev1.Container, error) {
```
ItalyPaleAle, line 616: Not strictly related to this PR, but this method now has 17 parameters (if I didn't count wrong). Perhaps should be changed to accept a single struct rather than an ever-growing, confusing list of parameters?

pkg/injector/pod_patch.go
```diff
@@ -773,6 +772,20 @@ func getSidecarContainer(annotations map[string]string, id, daprSidecarImage, im
 		},
 	}
 
+	// This is a special case for virtual kubelets on ACI.
```
ItalyPaleAle, line 775: Nit: Please expand acronym on first instance: Azure Container Instances.

pkg/injector/pod_patch.go
```diff
@@ -920,3 +933,15 @@ func getVolumeMounts(pod corev1.Pod) []corev1.VolumeMount {
...
 	return volumeMounts
 }
+
+// isACIVirtualKubelet returns true if the pod is scheduled in ACI.
+// If it is an ACI deployment, the pods have a toleration "azure.com/aci".
+// See https://github.com/virtual-kubelet/azure-aci#schedule-a-pod-in-aci
+func isACIVirtualKubelet(ts []corev1.Toleration) bool {
+	for _, t := range ts {
+		if t.Key == "azure.com/aci" {
+			return true
```
ItalyPaleAle, line 943: I think we need to check for the value too (`&& t.Effect == "NoSchedule"`)

shubham1172: Do we need this check? In case they expand the scope of `Effect` later.

ItalyPaleAle: I think we need it. The Kubernetes docs state:

image

https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/

I actually think we need to check that:

1. `t.Key` is `azure.com/aci`
2. One of:
   1. `t.Operator` is either `Equal` or missing (empty string) **and** `t.Value` is `NoSchedule`
   2. `t.Operator` is `Exist`

ItalyPaleAle: However this is a bit beyond my understanding of tolerations on K8s. I think the real expert is @yaron2

pkg/injector/pod_patch.go
shubham1172: I have refactored this in this PR.

pkg/injector/pod_patch.go
```diff
@@ -759,21 +793,35 @@ func getSidecarContainer(annotations map[string]string, id, daprSidecarImage, im
...
 	}
 
-	c.Env = append(c.Env, utils.ParseEnvString(annotations[daprEnvKey])...)
+	// This is a special case for virtual kubelets on Azure Container Instances (ACI).
+	// ACI does not support separate Command and Args,
+	// Ref. https://github.com/virtual-kubelet/azure-aci/issues/10.
+	// For other deployments, the Command is passed as a part of Args.
+	// This allows the Docker images to specify an ENTRYPOINT
+	// which is otherwise overridden by Command.
+	if isACIVirtualKubelet(cfg.tolerations) {
```
yaron2, line 816: Please remove all and every reference to ACI preferably to Virtual Kubelet also from the code. Please make these settings configurable on the Helm Chart and have the code act based on the configured settings.

yaron2 (changes requested):

pkg/injector/pod_patch.go
artursouza, line 816: I like that we found a solution for these cases. Thanks, @shubham1172! I think we can avoid having "if cloud provider" logic in the code here. We can change it to receive the list of tolerations from an environment variable and have the list to come from the Helm chart. The list would still be hardcoded, but in the Helm chart instead. This way, other cloud providers living the same situation can add to the list of tolerations in the Helm chart. I don't think that the list should be configurable by the end user since it gives a better experience out of the box without requiring user to have a problem and then discover the documentation about which toleration to add.

pkg/injector/pod_patch.go
shubham1172: Hello @ItalyPaleAle, in the new Helm mode, the configuration now contains an array of tolerations that must be matched with logical AND. The default configuration is now `[{\"effect\":\"NoSchedule\",\"key\":\"azure.com/aci\"}]`.

pkg/injector/pod_patch.go
shubham1172: Updated.

shubham1172: Thanks @artursouza, updated! The default list contains just one toleration for ACI, and can be updated via Helm charts to include more, they are matched with pod tolerations using logical AND.

pkg/injector/pod_patch.go
```diff
@@ -162,16 +183,32 @@ func (i *injector) getPodPatchOperations(ar *v1.AdmissionReview,
...
 	trustAnchors, certChain, certKey = getTrustAnchorsAndCertChain(kubeClient, namespace)
-	identity = fmt.Sprintf("%s:%s", req.Namespace, pod.Spec.ServiceAccountName)
 
-	socketMount := appendUnixDomainSocketVolume(&pod)
-	tokenMount := getTokenVolumeMount(pod)
-	volumeMounts := getVolumeMounts(pod)
-	sidecarContainer, err := getSidecarContainer(pod.Annotations, id, image, imagePullPolicy, req.Namespace, apiSvcAddress, placementAddress, socketMount, tokenMount, volumeMounts, trustAnchors, certChain, certKey, sentryAddress, mtlsEnabled, identity)
+	socketVolumeMount := appendUnixDomainSocketVolume(&pod)
+
+	cfg := sidecarContainerConfig{
```
ItalyPaleAle, line 191: ❤️ this

pkg/injector/pod_patch.go
```diff
@@ -760,21 +796,33 @@ func getSidecarContainer(annotations map[string]string, id, daprSidecarImage, im
...
+			FailureThreshold:    getInt32AnnotationOrDefault(cfg.annotations, daprLivenessProbeThresholdKey, defaultHealthzProbeThreshold),
 		},
 	}
 
-	c.Env = append(c.Env, utils.ParseEnvString(annotations[daprEnvKey])...)
+	// If the pod contains all the tolerations specified by the configuration,
+	// the Command and Args are passed as is. Otherwise, the Command is passed as a part of Args.
+	// This is to allow the Docker images to specify an ENTRYPOINT
+	// which is otherwise overridden by Command.
+	if podContainsTolerations(cfg.ignoreEntrypointTolerations, cfg.tolerations) {
```
ItalyPaleAle, line 817: Nit: perhaps a name like "podRequiresCommand" or something like that may be more descriptive

ItalyPaleAle:
Thanks for making the changes @shubham1172 

@yaron2 does this work for you? (Once the tests pass)

pkg/injector/pod_patch.go
shubham1172: Strictly functionally, it does check if the pod contains a list of tolerations from the configuration. Its consumer calls it to check where command is required or not. Should we rename this function to reflect the same? Isn't it a consumer detail? WDYT?

shubham1172:
Continued in #5067 to troubleshoot E2E failure and fixing weird DCO issue.

PR #5025, opened 2022-08-17
Feat/test parallel suites

mcandeia:
# Description

This is an attempt to make our e2e test stability better. The general idea is to divide our tests in N suites that could be configurable by just setting up an environment variable "E2E_TEST_SUITE_COUNT". The test chunks are arbitrarily chosen, following the lexicographical order of the folders' name. If we intentionally want some tests to execute in same suite one option is just name the folder as 1-testA and 1-testB.

We expect the following improvements:
- The total test duration will decrease
- The SNAT ports exaustion should not happen very often

## Issue reference

Please reference the issue this PR will close: #4962 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Unit tests passing


tests/list_test_suites.sh
```diff
@@ -0,0 +1,32 @@
...
+
+# divided suites
+suites=
+
+for((i=0; i < $total_tests; i+=tests_n))
+do
+  remaining=$(expr $i + $double)
+
+  # if it can't be divided exactly so the last test will include the remaining tests.
+  if [[ $remaining -lt $total_tests ]]; then
```
ItalyPaleAle, line 23: Note that some tests should always be in the same suite (i.e. executed in parallel). For example, all pubsub tests should be executed in parallel to avoid overloading Service Bus

mcandeia: In this case all pubsub tests should be executed sequentially rather than parallel, is that you meant?

If yes, indeed, as you can see the test suite division is not explicit, I was wondering if it really needs to be. Thoughts?

An option might be prefixing the folder name with the desired test suite, something like:

1-pubsub_test_b
1-pubsub_test_a

Folders that not contains the prefix wouldn't have any restriction, how about that?

Another option is having subfolders

/pubsub/test_a
/pubsub/test_b

ItalyPaleAle: Yes, sorry, I meant sequentially.

Any implementation would be fine.

Issue #5030, opened 2022-08-18
Support Lock component aliasing


mcandeia:
## In what area(s)?

/area runtime

> /area operator

> /area placement

> /area docs

> /area test-and-release

## Describe the feature

The lock component should support aliasing.

## Release Note

RELEASE NOTE:

PR #5033, opened 2022-08-18
Allow lock components to define aliases

mcandeia:

# Description

Lock component is the only component that doesn't support aliasing. This PR adds the support to it.

## Issue reference

Please reference the issue this PR will close: #5030 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Unit tests passing
* [] End-to-end tests passing
* [] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_

### DIFF
--- a/pkg/components/lock/registry.go
+++ b/pkg/components/lock/registry.go
@@ -15,12 +15,8 @@ type Lock struct {
 }
 
 func New(name string, f func() lock.Store, aliases ...string) Lock {
-	names := []string{name}
-	if len(aliases) > 0 {
-		names = append(names, aliases...)
-	}
 	return Lock{
-		Names:         names,
+		Names:         append(aliases, name),
 		FactoryMethod: f,
 	}
 }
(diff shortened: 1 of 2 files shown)


ItalyPaleAle (dismissed):

artursouza (approved):
Please, make the changes in SDKs too. .Net, Java, Python, JS, Cpp and Go.

artursouza:
We also need to update the docs.

yaron2:
> We also need to update the docs.

I don't think this change needs SDK and docs updates.. @mcandeia can you confirm?

mcandeia:
I shouldn't, I guess.

This feature has been used for renaming/supporting multiple names for the same component on the component manifest.

You can see the motivation [here](https://github.com/dapr/dapr/pull/4665)

@ItalyPaleAle can you help to clarify here?

ItalyPaleAle:
Correct. This was added in all the other components because of https://github.com/dapr/components-contrib/issues/1753

Aliases are generally used to maintain backwards-compatibility when we need to rename a component, for example because it was incorrectly named to begin with.

yaron2 (approved):

Issue #4979, opened 2022-08-05
Migrate HTTP server from fasthttp to net/http


ItalyPaleAle:
Since at least 2020, Dapr has been using the [fasthttp](https://github.com/valyala/fasthttp) library for the internal HTTP API server. An [Architectural Decision Record](https://github.com/dapr/dapr/blob/1cbf67f8a5da0bb5be4e3a123632c17ebb173f4e/docs/decision_records/architecture/ARC-004-http-server.md) from June 2020 (that is still in "proposed" status) contains information on this decision, which was based on the fact that, at least on synthetic benchmarks, fasthttp is optimized for performance and raw speed, primarily thanks to avoiding almost any memory allocation.

However, with more experience building Dapr and with the desire to push it forward, we are now uncovering a series of issues that could make it advisable to convert the internal HTTP server to be based on the Go standard library (net/http). It's likely that our use cases are not what fasthttp is meant for, and as we invest in things like [streaming supports](https://github.com/dapr/dapr/pull/4903), the issues will only be growing. This is advice that was provided by the maintainer (it appears to be a single person now) of fasthttp too.

Although switching from net/http to fasthttp will have a negative performance impact, arguably this should be negligible in the context of Dapr. Each one of our handlers performs a lot of operations (and are in large part I/O bound) and a lot of allocations and spanning of goroutines, which mask the effect of any micro-optimization performed by fasthttp.

> Note that fasthttp contains both a server and client, and Dapr uses both, although the client part plays a very small role.

I am categorizing the issues in 3 groups.

### Lack of context support 

With fasthttp, each request does not get a context that is tied to the lifetime of the request.

Although each handler is passed a `ctx` which does implement the `context.Context` interface (and we use it a lot in Dapr, [example](https://github.com/dapr/dapr/blob/master/pkg/http/api.go#L2181)), the lifecycle of that context is tied to the server's context and not the request's context. This seems to be an [explicit design decision](https://github.com/valyala/fasthttp/issues/1350) (although not clearly documented).

This is an issue for Dapr because if the user (app) cancels a request, we do not have a way to be notified of that and abort in-progress operations.

According to the [fasthttp maintainer](https://github.com/valyala/fasthttp/issues/1219#issuecomment-1041548933):

> If people want to use contexts they really shouldn't be using fasthttp. When you are using contexts it shows that you don't need the performance increase fasthttp gives. If you really needed that performance you wouldn't be using contexts to begin with.

### No support for HTTP/2

[...]

PR #5048, opened 2022-08-19
Remove `http-stream-request-body`

ItalyPaleAle:
This is another spin-off from #4903 and it removes `http-stream-request-body`. Two reasons:

1. As we move towards support for streaming throughout the runtime, that flag will not be needed anymore
2. The flag doesn't work today anyways. Per my research on the internals of fasthttp (from #4979):
   1. Setting that flag enables consumers to read data as stream (with the caveats from #4979) but currently Dapr doesn't use the methods to read data as streams: Dapr only uses methods like `reqCtx.PostBody()` which return the entire request body as a byte slice, regardless of streaming support.
   2. That flag causes some unintended behavior, such as `http-max-request-size` to not work as expected

Docs issue: https://github.com/dapr/docs/issues/2737


pkg/messaging/direct_messaging.go
```diff
@@ -61,7 +60,6 @@ type directMessaging struct {
 	grpcPort            int
 	namespace           string
 	resolver            nr.Resolver
-	tracingSpec         config.TracingSpec
```
ItalyPaleAle, line 64: Removed as it was not used

pkg/messaging/direct_messaging.go
```diff
@@ -77,45 +75,47 @@ type remoteApp struct {
...
-	resolver nr.Resolver,
-	tracingSpec config.TracingSpec,
-	maxRequestBodySize int,
-	proxy Proxy,
-	readBufferSize int,
-	streamRequestBody bool,
-	resiliency resiliency.Provider,
-	isResiliencyEnabled bool,
-) DirectMessaging {
+func NewDirectMessaging(opts NewDirectMessagingOpts) DirectMessaging {
```
ItalyPaleAle, line 95: Why this change: the old method's signature had 15 arguments and it was very hard to maintain when we needed to remove or add parameters (like in this case), especially in tests

ItalyPaleAle:
Now that the other PR is merged, this is ready for review

pkg/messaging/direct_messaging.go
berndverst: There are some slight performance difference in using the options pattern vs the named params. But I don't recall which one is better haha

berndverst (approved):

berndverst:
LGTM

pkg/messaging/direct_messaging.go
ItalyPaleAle: This may be slightly worse, but this code path is only executed once during initialization.

yaron2 (approved):

PR #5050, opened 2022-08-19
`resiliency.PolicyDefined` returns the policy object

ItalyPaleAle:
This is another spin-off from #4903

It changes the `resiliency.PolicyDefined` method to return the actual policy object and not just a boolean indicating whether a policy exists.

This was necessary for #4903, but I co-authored this change with @halspang who said he thinks we need this for other things too, in the future.


pkg/resiliency/resiliency.go
```diff
@@ -840,6 +897,11 @@ func (e *circuitBreakerInstances) Remove(name string) {
 	e.Unlock()
 }
 
+// HasRetries returns true if the policy is configured to have more than 1 retry.
+func (p PolicyDescription) HasRetries() bool {
```
mukundansundar, line 901: any specific reason this has to be an exported function in this PR?

mukundansundar (approved):

pkg/resiliency/resiliency.go
ItalyPaleAle: That is needed in #4903

artursouza (approved):
Can you follow up with a PR renaming PolicyDefined() to GetPolicy()?

Issue #5053, opened 2022-08-20
Support running E2E tests over tailscale network with a Personal Token


mcandeia:
## In what area(s)?

> /area runtime

> /area operator

> /area placement

> /area docs

/area test-and-release

The original issue is the #4962. However it stills on hold because of [this](https://github.com/dapr/dapr/pull/4991#issuecomment-1218304428).

I'm creating a new issue to address the possibility of executing the tests using the tailscale subnet router but using their personal keys.

PR #5054, opened 2022-08-20
E2E Tailscale local run support

mcandeia:
# Description

Documentation and make-like commands to setup tailscale subnet router with a personal access key.

## Issue reference

Please reference the issue this PR will close: #5053 

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing


tests/config/tailscale_key.yaml
```diff
@@ -0,0 +1,9 @@
+# Copyright (c) 2021 Tailscale Inc & AUTHORS All rights reserved.
+# Use of this source code is governed by a BSD-style
+# license that can be found in the LICENSE file.
```
ItalyPaleAle, line 3: I think we need to replace the copyright with the Dapr one and then add theirs below. For theirs, include:
- A link to where the code comes from
- Their copyright (“Copyright (C) 2021 Tailscale Inc & AUTHORS. All rights reserved”)
- A mention of the license and a link to the LICENSE file that’s mentioned there

tests/docs/running-e2e-test.md
```diff
@@ -197,6 +197,67 @@ If you want to run the tests in a similar environment, you can deploy the test i
...
+        "tag:dapr-tests": ["your_email_comes_here@your_domain.com"],
+      }
+    }
+    ```
+   2. Assign permissions to the created tag. Since we are going to use the [tailscale subnet router](https://tailscale.com/kb/1185/kubernetes/), it is much convenient that the subnet router should auto approve the registered routes, for that, use the following acl.
+   ```json
+     {...
+       "autoApprovers": {
+         "routes": {
+           "0.0.0.0/0": ["tag:dapr-tests"],
```
ItalyPaleAle, line 224: Since all Kubernetes routes are on 10.0.0.0/8 can you restrict it to that? For security reasons (so people can’t change the route to, say, google.com) and also to reduce the risk of mistakes.

tests/docs/running-e2e-test.md
```diff
@@ -197,6 +197,67 @@ If you want to run the tests in a similar environment, you can deploy the test i
...
+
+The next step will be install the tailscale subnet router in your kubernetes cluster, for that, run
+
+```sh
+TAILSCALE_AUTH_KEY=your_key_goes_here make setup-tailscale
+```
+
+If you prefer you can also export the key
+```sh
+export TAILSCALE_AUTH_KEY=your_key_goes_here
```
ItalyPaleAle, line 243: I’d remove this. It shouldn’t be set in the shell env in general, for security reasons, so not a practice we should recommend

ItalyPaleAle (commented):
Great work on this. It will make it much quicker to develop tests and run them against a remote Kubernetes cluster, as we won’t need to wait for the load balancer anymore (it can take up to 1 min per each test run, per app). Also, it helps us MS employees to test against AKS thanks to the NAT traversal :)

Please do fix the linter issues however.

tests/docs/running-e2e-test.md
```diff
@@ -197,6 +197,67 @@ If you want to run the tests in a similar environment, you can deploy the test i
...
+One more config is necessary, `E2E_USE_INTERNAL_IP=true`, you can use it as a variable when running tests as the following:
+
+```sh
+E2E_USE_INTERNAL_IP=true make test-e2e-all
+```
+
+Or exported as the previous ones,
+
+```sh
+export E2E_USE_INTERNAL_IP=true
```
mcandeia, line 258: Rename to TEST_E2E_USE_INTERNAL_IP

tests/docs/running-e2e-test.md
mcandeia: Yes, it makes sense, thanks

tests/docs/running-e2e-test.md
mcandeia: You're right, some malicious script could steal our environment variables. I was wondering if the same applies to ~/.bash_history ~/.zsh_history, so I'm going to add a section to advice using `unset HISTFILE` before running the command

mcandeia:
It looks like some tests failed. I'll take a look.

tests/config/tailscale_key.yaml
mcandeia: could you check now?

tests/docs/running-e2e-test.md
ItalyPaleAle: I was more concerned about the fact that with “export”, any process you launch from that shell gets the env var. The risk of saving in the history is a different one. Some shells (like ZSH) can be configured to not save in the history if the command starts with a space.

ItalyPaleAle (approved):

mcandeia:
Just fixed the lint problem!

ItalyPaleAle (approved):

ItalyPaleAle (approved):

tests/config/tailscale_key.yaml
artursouza: This does not make sense. This is Dapr specific.

ItalyPaleAle: @artursouza is right on this file. The other ones seem copied from https://github.com/tailscale/tailscale/tree/main/docs/k8s but this one is original from us, right? So no need to keep their copyright on this specific file.

artursouza (approved):

PR #4860, opened 2022-07-05
Add web3.ipfs binding

hypernova3:
# Description

for https://github.com/dapr/components-contrib/pull/1849

## Checklist

Please make sure you've  completed the relevant tasks for this PR, out of the following list:

* [x] Code compiles correctly
* [ ] Created/updated tests
* [x] Unit tests passing
* [x] End-to-end tests passing
* [ ] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_
* [ ] Specification has been updated / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_
* [ ] Provided sample for the feature / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_

PR #5062, opened 2022-08-23
Components auto-registration

ItalyPaleAle:
This PR changes the way that the component registry works internally.

Before, each component had to have an import in the `cmd/daprd/main.go` file and then be added to the relevant registry in that file. With over 100 components, that started to become complex to track too.

Now, each component is automatically registered when a file is placed in `cmd/daprd/components`, with each component getting its own file. For example, this is the `state_in-memory.go` file:

```go
package components

import (
	inmemory "github.com/dapr/components-contrib/state/in-memory"
	stateLoader "github.com/dapr/dapr/pkg/components/state"
)

func init() {
	stateLoader.DefaultRegistry.RegisterComponent(inmemory.NewInMemoryStateStore, "in-memory")
}
```

This PR is needed to simplify (greatly) the life of users who are building Dapr with a custom set of components, either adding new ones or stripping down components. The biggest advantage is a significant reduction in merge conflicts when managing a custom set of components.

PS: A companion PR will be needed in components-contrib. Once that's merged, I will remove the temporary replacement in the go.mod https://github.com/dapr/components-contrib/pull/1997

PPS: Also fixes #4860


yaron2:
Pinging @pkedy for review and thoughts as the main author of the current registration mechanism

daixiang0:
I have a thought that could we use a method like a template to generate files in `cmd/daprd/components` based on config (maybe build config), make whole build like a menu?

ItalyPaleAle:
It’s an option for the future. It does raise questions about what the supportability story would be in that case.

Note that some components have slightly different files, especially bindings (they can have both input and output) and middlewares.

for now, this PR is optimized to allow the people who absolutely need to build their own Dapr with a custom selection of components to do that more easily. Especially if you want to remove a lot of components, doing it without this PR makes it very likely to have a lot of merge conflicts over time.

daixiang0:
In this way the dependences still exist in `go.mod`, how do you fix the "merge conflicts" issue?

ItalyPaleAle:
The merge conflicts would be in the `cmd/runtime/main.go`. If you remove a lot of dependencies (like a user wants to do right now, removing almost all components), then any chance to the `main.go` file is likely to cause merge conflicts.

For the `go.mod` file, users could place their own additional dependencies at the very top or very bottom in a separate section. If they just want to remove components, then there's no problem there.

artursouza:
LGTM. Pending validation passing.

ItalyPaleAle:
Something's up with the change. I'll investigate and the E2E test failures tomorrow morning.

ItalyPaleAle:
@artursouza  fixed the issue that was causing tests to fail.

ItalyPaleAle:
@yaron2 @artursouza @pkedy any update on this?

I heard from @mcandeia that this will help his work on pluggable components too

artursouza (approved):

PR #5067, opened 2022-08-23
Add Helm option for using separate Commands and Args in sidecar container

shubham1172:
Signed-off-by: Shubham Sharma <shubhash@microsoft.com>

# Description

This PR introduces a new Helm chart configuration in the dapr sidecar injector - `ignoreEntrypointTolerations`. It is a JSON array of tolerations.

If the application pod contains any of the tolerations specified by the Helm config above, the sidecar image has an explicit Command set (`daprd` or `dlv`) , along with the correct Args. Otherwise, Command is kept as empty and `daprd` or `dlv` is a part of Args. This allows the ENTRYPOINTS in Dapr Docker image to be executed first.

The motivation for this PR is to support cloud providers (e.g. ACI virtual-kubelet) that mandate both Command and Args to be set. In these cases, the ENTRYPOINT of Docker images are overridden. See the attached issue for more details on this.   

## Note to reviewers

Please refer to existing comments in #5024

## Issue reference

Please reference the issue this PR will close #5023

## Checklist

Please make sure you've  completed the relevant tasks for this PR, out of the following list:

* [X] Code compiles correctly
* [X] Created/updated tests
* [X] Unit tests passing
* [X] End-to-end tests passing
* [ ] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_
* [ ] Specification has been updated / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_
* [ ] Provided sample for the feature / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_

PR #5081, opened 2022-08-24
Add additional Virtual Kubelet providers

yaron2:
Follow up to https://github.com/dapr/dapr/pull/5067 to add more providers.


daixiang0 (approved):
LGTM

shubham1172:
@yaron2 what was the intent of this PR? Do all the virtual kubelets added by you require Command and Args to be separate. For the original PR, ACI had this problem specifically because of the way its API was structured, see https://github.com/virtual-kubelet/azure-aci/issues/10

yaron2:
> @yaron2 what was the intent of this PR? Do all the virtual kubelets added by you require Command and Args to be separate. For the original PR, ACI had this problem specifically because of the way its API was structured, see [virtual-kubelet/azure-aci#10](https://github.com/virtual-kubelet/azure-aci/issues/10)

Yes. Most of the VKubelet providers actually append the args to command and not use Entrypoint. Another reason is to provide consistency in case this changes in the underlying platform for any reason.

Issue #4925, opened 2022-07-21
[Proposal] (Updated) gRPC Components (aka "Pluggable components")


johnewart:
To reset the conversation a bit in preparation for 1.9, this is an updated proposal for gRPC-based components. This proposal and its underlying implementation does not deviate much from the previous proposal and work from @patrickhuber but adds a bit more clarity around the design and proposed packaging of components. 

# Dapr Pluggable Components 

## Proposal 

To build a general-purpose mechanism for implementing Dapr components outside of the daprd binary itself. The goal of this is to both improve the overall health of the Dapr sidecar as well as increase flexibility when building new component implementations.  

### Terminology 

_Building block_ -- an abstract, reusable, description of a set of operations including their expected behavior, inputs and outputs. For example, a "state store" building block can store state, fetch state, delete state or update state.  

_Component_ an implementation of a Dapr building block that provides the desired functionality. For example, a Redis state store would implement the "state store" building block operations by using Redis as a storage engine.  

### Design goals 

* Create a mechanism for building Dapr components in a way that does not require modifying the dapr source code and recompiling the binaries. 
* Minimize the performance penalty incurred by such a system. 
* Allow components to be developed in languages other than Go (specifically, those supported by gRPC) 
* Increase flexibility of the dapr sidecar 
* Improve long-term health of the dapr project 

### Non-goals 

* Support components running on machines separate from where the dapr sidecar or daemon is being run (i.e., via remote TCP/IP sockets). 
* Supporting programming languages that do not have existing protobuf compilers. 

## Background 
 
Today, building a new component requires that the code for the component be built into the Dapr binary itself. This mechanism presents a few hurdles for Dapr (though this also applies to any other generally extensible system) that make development of new features and long-term maintenance of the platform itself challenging. In addition to project health-related challenges, it increases the friction for adoption of Dapr when customers need to integrate with systems that are not otherwise open-source. As a result, Dapr will benefit from a framework that enables the development of components that are not tightly coupled to the Dapr source code in both maintainability and adoption.  

### Challenges related to project health 

As Dapr grows, and along with it the number of components that exist, there arFe several technical challenges that the project will face including development, deployment and security-related problems. These include (but are not necessarily limited to): 

* New components increase the size of the dependency-graph, leading to a higher likelihood of conflicts or forced-upgrades. 

* Larger application footprint as all modules will be built into the binary distribution whether they are used or not.  

* Increased surface area for security vulnerabilities due to the inclusion of unused code and dependencies.  

[...]

PR #5102, opened 2022-08-30
Support State Store Pluggable Components

mcandeia:
# Description

## Preface

This PR adds support for gRPC components (a.k.a pluggable components). For the context of this PR only State Store pluggable component is covered along with the required barebones for the upcoming pluggable components.

* [x] State Store
* [ ] Input Bindings
* [ ] Output Bindings
* [ ] Lock Stores
* [ ] Configuration Stores
* [ ] Name Resolution
* [ ] Middlewares
* [ ] Secret Stores
* [ ] PubSub

## Hacking with It

1. [.NET](https://github.com/mcandeia/DaprPluggableComponentSDK.NET/tree/fix/service-lifecycle) thanks to @johnewart
2. [Java](https://github.com/tmacam/DaprPluggableComponent-Java) thanks to @tmacam
3. [Go](https://github.com/mcandeia/dapr-components-go-sdk)

All you need is an application that listen to a **default component-specific** unix domain socket that responds the gRPC calls defined in our of a [State Store Service](https://github.com/mcandeia/dapr/blob/feat/statestore-pluggable/dapr/proto/components/v1/state.proto).

From the runtime perspective the only thing that link a pluggable component to its implementation is the same socket as prior mentioned. It will be unique for each **Component Instance** to avoid complexity related to multiplexing sockets when implementing multiple components.

As a Pluggable Component implementer you don't need to worry about understand gRPC, unix sockets or any communication mechanism that we've chosen. We're going to provide SDKs for a set of languages (starting from .NET, Java and Go, probably) and the implementation of a pluggable component will be pretty much the same as doing it inside components-contrib with few nuances.

## Pluggable Concept

Pluggable Components are implementation of building blocks running as a separate process - generally containers - communicating via a Unix Domain Sockets. For the this initial release our expectation is that people will **self-manage their pluggable components**, and again, from the runtime perspective a pluggable component is nothing more than a socket responding gRPC requests.

## Pluggable Components CRD

The way we declare pluggable components is via CRD. A new CRD of Kind `PluggableComponent` was added and for now, the idea is to having this CRD living together with others' components CRD inside components folders - you will probably notice that I have to make a change in the component-loaders to support loading the pluggable components from disk.

It will look like this:

```yaml
apiVersion: dapr.io/v1alpha1
kind: PluggableComponent
metadata:
  name: imdotnet
spec:
  type: state
  version: v1
```

## Pluggable Components Proto

The pluggable components proto will sit in the same root folder as the dapr API yet in different subfolders.

## Pluggable Components Registry

Pluggable components are not concrete instances of components, in fact, if we were to make an analogy they would be the definitions of the classes that implement the component. So indeed, it doesn't have its own registry, it uses the other registry as a proxy for itself.

### Kubernetes Mode

**Kubernetes mode is not supported yet**, another PR will be needed to not add more complexity on this. It will  probably require new operators APIs.

[see here a non-production ready draft-just-for-poc-purposes](https://github.com/mcandeia/dapr/pull/5)

### Conformance/Certification tests

The same conformance and certification testes that has been applied for the others components should be re-used for pluggable components as well.

My proposal for conformance test is provide a way to run it as a github action along with the pluggable component implementation. [Here](https://github.com/mcandeia/action-dapr-conformance-tests)

For conformance test I was able to wrap the actual Redis component into a pluggable component and got conformance tests passed.

You can see it here: https://github.com/mcandeia/dapr-components-go-sdk/blob/main/.github/workflows/ci.yaml#L33 the pipeline that covers it.

You may ask: Why is the conformance test not written inside the current components-contrib repository?

Well, the conformance test for pluggable components is not exactly what it means to, because pluggable components serve as an intermediate the communication between dapr sidecar and the component implementation itself. So ideally all current conformance tests should apply for a wrapped pluggable component. So having it as a github action is a good option to allow pluggable component developers to get their components tested.

Above you can see the current state of Redis Pluggable component.
image

[take a look on the runner logs](https://github.com/mcandeia/dapr-components-go-sdk/runs/8142932577?check_suite_focus=true)

> Disclaimer: The github action and the sdks by no means should be used in production, they are only proof of concepts that will come along with a proper proposal and discussion. The idea of having they running is just to provide some level of confidence that we're on the right path.

### Monitoring/Observabilty

The metrics that has been used for the others components should work out-of-the-box for pluggable components. New metrics related to the gRPC connection must be added soon - maybe an option would be reuse our grpc_monitoring.go, should require remove app-specific metrics/monitoring.

The same applies for tracing, we should manage to have access to the parent context

### Resiliency

Since we're using unix sockets, for now only one connection is being created and reused, this connection should be monitored.

## Issue reference

Please reference the issue this PR will close: #4925 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Unit tests passing


dapr/proto/components/v1/state.proto
```diff
@@ -0,0 +1,212 @@
...
+  // Operations that should be performed.
+  repeated TransactionalStateOperation operations = 1;
+  // Request associated metadata.
+  map<string, string> metadata = 2;
+}
+
+// StateStore service provides a gRPC interface for state store components.
+service StateStore {
+  // Initializes the state store component with the given metadata.
+  rpc Init(MetadataRequest) returns (google.protobuf.Empty) {}
```
ItalyPaleAle, line 117: For future-proofing, I'd strongly recommend defining a type for each request or response, even if it's empty. For example, here you'd define messages `InitRequest` (`MetadataRequest` would be part of that) and `InitResponse` (which would be empty)

This way, we will be able to extend the proto in the future without breaking anything.

You can read more here: https://stackoverflow.com/a/50996412

pkg/apis/components/v1alpha1/types.go
```diff
@@ -37,6 +37,36 @@ type Component struct {
 	Scopes []string `json:"scopes,omitempty"`
 }
 
+// Kind returns the component kind.
+func (Component) Kind() string {
+	return "Component"
+}
+
+// +genclient
```
ItalyPaleAle, line 45: Comments meant to be read by a machine should not have a space after `//`

mcandeia: Thanks for catching this!

As you can [see](https://github.com/dapr/dapr/pull/5102/files/653dac8cb4caa5328d3b027ae9538a59258eb362#diff-b61d06975d6d1f5da12fdb8a6368b859979ae451e15d9c02825e91ea305f5b64R23) I'm actually using the same format as for Component CRD.

dapr/proto/components/v1/state.proto
mcandeia: While I agree with you that this is cool, I think it ends up not having as much effect if you look at the whole context.

For this to change to occur, theoretically we would have to change the interface on code, which will lead us to change the entire chain of implementers/sdks, see that having an empty object in this case means not returning anything in the code, which will surely become a breaking changes anyway.

Another point is that from [what I noticed](https://github.com/dapr/dapr/blob/master/dapr/proto/runtime/v1/dapr.proto#L19), we already have this pattern in other APIs, is it interesting to deviate only here?

I'm not against it, but seems that in the other hand it could create empty objects without any strong reason,

unless we already have a strong intention to change an API soon, maybe it's a good one.

does it makes sense to you @ItalyPaleAle ?

pkg/apis/components/v1alpha1/types.go
ItalyPaleAle: Yes the convention throughout Go is that comments have a space after // and directives (meant to instruct a compiler or something else) have no space after //

dapr/proto/components/v1/state.proto
ItalyPaleAle: Yeah I see what you're saying. I guess it depends on how we think the API will evolve. Being able to add extra fields may allow us to make incremental evolutions without requiring an entire v2 version. Whether that's needed or not, perhaps should be a discussion.

Indeed we use that pattern everywhere in the code already. Still doesn't mean that we cannot use better patterns (if appropriate) in newer code.

mcandeia: >Indeed we use that pattern everywhere in the code already. Still doesn't mean that we cannot use better patterns (if appropriate) in newer code.

Yes, I agree with you. The main question is, worth to pay API consistency vs non-predictable extensibility in this case?

Notice that changing the API is different from code. The API is what we expose as the interface for all SDKs, a change here will require a change in all SDKs, I generally care less about changes that cause inconsistency in the code if it is well scoped - new code for instance.

In the case of an API inconsistency - whether objects are empty or not - I think this should be taken as a decision at the API guidelines level and not punctually,

What do you think?

pkg/apis/components/v1alpha1/types.go
mcandeia: Fixed!

dapr/proto/components/v1/state.proto
ItalyPaleAle: Perhaps we should discuss this in a separate issue or during a call?

mcandeia: I agree, where we should start it?

mcandeia: As a followup discussion with @artursouza he convinced me that it is worth deviating here especially because the pluggable components proto is kinda new API that could change a lot soon. 

And even dough at the end of the day changing proto means changing the entire sdk chain, having empty messages will allow to break semantically but no structurally.

So, let's say we have a new field on `SetResponse` and this will require a new SDK version and a new component interface on code. For changes backwards compatible for instance, adding a new field on SetResponse, will also be semantically backwards compatible with components made on the previous version because the type didn't change. As opposite if it as a protobuf.empty, which will enforce components to update even if they don't want to use the new SetResponse field.

Thanks for suggesting this @ItalyPaleAle

dapr/proto/components/v1/state.proto
```diff
@@ -0,0 +1,241 @@
...
+  // The response token.
+  string token = 2;
+  // Response associated metadata.
+  map<string, string> metadata = 3;
+}
+
+// TransactionalStateStore service provides a gRPC interface for transactional state store components.
+service TransactionalStateStore {
+    // Multi executes multiples operation in a transactional environment.
+    rpc Multi(TransactionalStateRequest) returns (TransactionalStateResponse) {}
```
tmacam, line 92: Why are we calling this operation/method `Multi`?

dapr/proto/components/v1/state.proto
```diff
@@ -0,0 +1,241 @@
...
+import "dapr/proto/components/v1/common.proto";
+
+option go_package = "github.com/dapr/dapr/pkg/proto/components/v1;components";
+
+
+// QueriableStateStore service provides a gRPC interface for querier state store components.
+service QueriableStateStore {
+  // Query performs a query request on the statestore.
+  rpc Query(QueryRequest) returns (QueryResponse) {}
+}
```
tmacam, line 28: I noticed we are missing `Ping` and `Init` both here and in the Transactional one. I guess the assumption is that one **must** implement a `service StateStore`  before exposing a `QueriableStateStore` or a `TransactionalStateStore`. Correct? Might be worth stating this in the comments.

mcandeia: That's a good point thanks for asking this,

I got a bit confusing about this as too, but at the end of the day you need a `StateStore` implementation for either, queriable and transacional, you can see [here](https://github.com/dapr/components-contrib/blob/master/state/transactional_store.go#L17) and [here](https://github.com/dapr/components-contrib/blob/master/state/store.go#L98) that we do require init method for transacional state store but not for querier.

However I noticed that it does not make difference because we don't have a separate registry for the transactionalstatestore or for the querier. Looks like that statestore is the base interface for the others and in runtime - based on state store features, a [cast](https://github.com/dapr/dapr/blob/a6153bb26c5ae21dbb88569d7faf42d5011003f8/pkg/grpc/api.go#L293) to transactional state store is made. So it's not necessary to repeat ping/init methods for those.

dapr/proto/components/v1/state.proto
mcandeia: Good point, I just copy/paste the same as used in code [here](https://github.com/dapr/components-contrib/blob/master/state/transactional_store.go#L19). 

I have no objection to changing it to something more suitable ideas?, "rpc ExecTransaction" "rpc Transaction" "rpc Transact"

artursouza: I agree. Can we change to Transact?

pkg/apis/components/v1alpha1/types.go
```diff
@@ -20,40 +20,70 @@ import (
...
+
+// Kind returns the component kind.
+func (PluggableComponent) Kind() string {
+	return "PluggableComponent"
+}
+
+// PluggableComponentSpec is the spec for a pluggable component.
+type PluggableComponentSpec struct {
+	Type    string `json:"type"`
+	Version string `json:"version"`
```
artursouza, line 67: Version is not applicable because this only applies to the implementation of the component. In this case, versioning of a pluggable component does not belong to Dapr (redis v1, redis v2, etc). So, we can just remove this attribute.

pkg/components/disk_manifest_loader_test.go
```diff
@@ -0,0 +1,201 @@
+/*
+Copyright 2021 The Dapr Authors
```
artursouza, line 2: nit: 2022 on every new file.

pkg/components/manifest_loader.go
```diff
@@ -0,0 +1,24 @@
...
+*/
+
+package components
+
+type kubernetesManifest interface {
+	Kind() string
+}
+
+// ManifestLoader loads manifest-like files.
+type ManifestLoader[T kubernetesManifest] interface {
```
artursouza, line 21: @yaron2 are you OK with the use of generics?

pkg/components/state/pluggable.go
```diff
@@ -0,0 +1,464 @@
...
+	"github.com/dapr/kit/logger"
+
+	"google.golang.org/grpc"
+	"google.golang.org/protobuf/types/known/anypb"
+)
+
+var (
+	ErrNilSetValue                = errors.New("an attempt to set a nil value was received, try to use Delete instead")
+	ErrRespNil                    = errors.New("the response for GetRequest is nil")
+	ErrMultiOperationNotSupported = errors.New("multi operation not supported")
```
artursouza, line 36: Transact

pkg/components/pluggable/registry.go
```diff
@@ -0,0 +1,53 @@
...
+	}
+}
+
+// Register register all plugable components and returns the total number of registered components.
+func Register(pcs ...components.Pluggable) int {
+	registeredComponents := 0
+	for _, pc := range pcs {
+		register, ok := registries[pc.Type]
+		if !ok {
+			log.Warnf("%s is not registered as a pluggable component", pc.Type)
```
artursouza, line 42: Can you log.Info() when the registration is successful? It can help debugging and is not too noisy since is not a per request type of message.

artursouza (commented):
Some minor comments. LGTM overall.

pkg/components/pluggable/loader.go
```diff
@@ -0,0 +1,58 @@
...
+// LoadFromDisk loads PluggableComponents from the given path.
+func LoadFromDisk(pluggableComponentsPath string) ([]components.Pluggable, error) {
+	comp, err := newFromPath(pluggableComponentsPath).Load()
+	if os.IsNotExist(err) {
+		return make([]components.Pluggable, 0), nil
+	}
+	return mapComponents(comp), err
+}
+
+// TODO Load from K8S
```
tmacam, line 54: Should we have an issue to track this TODO?

mcandeia: Dont know if we need an issue for that, I think the overall progress of pluggable components must include this, hence the current issue should be sufficient, what do you think?

pkg/components/state/pluggable.go
```diff
@@ -0,0 +1,464 @@
...
+		results[idx] = itemIdx
+	}
+	return &state.QueryResponse{
+		Results:  results,
+		Token:    resp.Token,
+		Metadata: resp.Metadata,
+	}
+}
+
+func toMultiOperation(req state.TransactionalStateOperation) (*proto.TransactionalStateOperation, error) {
```
tmacam, line 276: Transact?

utils/utils.go
```diff
@@ -93,6 +93,14 @@ func SetEnvVariables(variables map[string]string) error {
...
 }
 
+// GetEnvOrElse get the value from the OS environment or use the else value if variable is not present.
+func GetEnvOrElse(name, orElse string) string {
+	if value, ok := os.LookupEnv(name); ok {
+		return value
+	}
+	return orElse
+}
+
```
tmacam, line 103: Not sure if this is required but... shouldn't we check whether the envVar value is also is non-empty and/or non-blank?

tmacam (commented):
Some nitpicking with "Multi". LGTM otherwise.

dapr/proto/components/v1/state.proto
tmacam: Might be worth adding as a comment this to the .proto, that one must define a base StateStore over which Queries and Transactions can be run.

pkg/components/pluggable/loader.go
tmacam: I don't know what is the practice for this project. If this is being addressed in a follow up PR, I don't see the reason for it. If this is just a "good to have" feature then, again, an issue would not really be required. For all other cases I would say it is worth having an issue open to keep track of this -- no one is keeping track of the number of open TODOs, right? :-P

dapr/proto/components/v1/state.proto
```diff
@@ -0,0 +1,241 @@
...
+  rpc Delete(DeleteRequest) returns (DeleteResponse) {}
+
+  // Get data from the given key.
+  rpc Get(GetRequest) returns (GetResponse) {}
+
+  // Sets the value of the specified key.
+  rpc Set(SetRequest) returns (SetResponse) {}
+
+  // Ping the state store. Used for liveness porpuses.
+  rpc Ping(PingRequest) returns (PingResponse) {}
```
tmacam, line 134: Something I have been wondering: what exactly is the semantics of this method? Is it a "component is up and responding" or does it work more like a deep ping, reaching out to the underlying data store it is connecting Dapr to? Should it return a boolen or error status?

Suppose I am a component that provides connectivity to a ElasticSearch cluster. The cluster is unreachable. What should `ping` implementation do in this case?

1.  Raise an exception?
2. Nothing -- the component is up, that the cluster is unreachable is another concern. 
3. Ping should be a blocking call down to the downstream service, so... unreachability means it times out?

dapr/proto/components/v1/state.proto
mcandeia: Fixed

dapr/proto/components/v1/state.proto
mcandeia: added!

pkg/apis/components/v1alpha1/types.go
mcandeia: I don't know if I caught your concern,

The version attr is necessary when you have multiples pluggable components versions being used by the same app (i.e making a transition between two majors), it must be used to disambiguate between them. And, currently is being used for the socket name creation - which contains the verison number in it.

So let's say we have the following component

```yaml
apiVersion: dapr.io/v1alpha1
kind: Component
metadata:
  name: state
spec:
  type: state.my-pluggable-component
  version: v1
```

and another one

```yaml
apiVersion: dapr.io/v1alpha1
kind: Component
metadata:
  name: another-state
spec:
  type: state.my-pluggable-component
  version: v2
```

I can't see this as a problem, it's totally fine. So, going back to the problem: We do need two pluggable components definitions to allow such configuration, right? 

They're different applications, different sockets and probably different behaviors,

Please, correct me if I'm wrong

pkg/components/pluggable/registry.go
mcandeia: I do log how many components were successfully registered [here](https://github.com/dapr/dapr/pull/5102/files#diff-e56f9bb947923c6f8bb4c995bf4b9469a0ed117276223d16b177dcb04399a4bcR638).

Seems good enough to you?

pkg/components/pluggable/loader.go
mcandeia: tbh I agree with you that we should have "small" issues instead of a big one that should be closed when everything is ready - especially because being ready is too relative.

One thing that would become confuse is when the big issue should be closed then, perhaps it shouldn't be treat as an issue but a milestone/project/something bigger instead? 

anyways I'll create these issues shortly after having this PR closed when we have more clear vision of what we need in the near future, how about that?

dapr/proto/components/v1/state.proto
mcandeia: Liveness in this case means: I can reach the component and he's responding to a Ping request. It is exactly like `ping` for testing connectivity over ICMP.

However, I agree with you that we should define the exactly associated semantics that we expect from the component, for instance:

If I can reach the component but some dependency is down, should it return an OK PingResponse? 

I think we can have different probes for it in the next releases like readiness/liveness, or even better - kubernetes might solve this for us.

Thanks for addressing this.

tmacam: > Liveness in this case means: I can reach the component and he's responding to a Ping request. It is exactly like ping for testing connectivity over ICMP.

True. As long as that is the semantic we are aiming at, which is different than "TCP port is open" which is also different than and "I got a 200 OK HTTP response".  In my head I thought ping was a "deep ping" but.. my bad. :)

pkg/apis/components/v1alpha1/types.go
artursouza: OK. I see where you are going. I thought we would not own versioning the pluggable components and let them just be a different type: `state.my-pluggable-component-v2`. But I am OK to keep it as you proposed so we keep consistency with the rest of Dapr.

pkg/components/pluggable/registry.go
artursouza: The goal is to log each one, not just the count. It is more helpful for debugging in production.

dapr/proto/components/v1/state.proto
artursouza: Some components do offer a deep check, like `select 1 from dual;`.

artursouza (approved):

PR #5109, opened 2022-08-31
optimize(runtime): add --namespace param to set namespace in standalone mode.

1046102779:
Use of `NAMESPACE` name is prohibited in standalone mode, We need to add cmdline params to fix it.

the purpose is to service discovery in standalone mode.

in standalone mode, I hope to be able to discover target services like cluster mode. such as: 

curl -i "http://127.0.0.1:3500/v1.0/invoke/micrologic.default/method/activity/info?id=123"

image

/cc @daixiang0 @artursouza @yaron2 @berndverst @ItalyPaleAle


pkg/runtime/cli.go
```diff
@@ -41,11 +39,14 @@ import (
...
 	"github.com/dapr/dapr/pkg/version"
 	"github.com/dapr/dapr/utils"
+	"github.com/dapr/kit/logger"
+	"github.com/dapr/kit/ptr"
 )
 
 // FromFlags parses command flags and returns DaprRuntime instance.
 func FromFlags() (*DaprRuntime, error) {
 	mode := flag.String("mode", string(modes.StandaloneMode), "Runtime mode for Dapr")
+	namespace := flag.String("namespace", "", "micrologic namespace, default empty. and it also be set by env `NAMESPACE`")
```
ItalyPaleAle, line 49: I don't think we refer to "micrologic" in our codebase anywhere

ItalyPaleAle (changes requested):

1046102779:
|mode|namespace|
|---|---|
|kubernetes|Yes|
|standalone|No|

Our service architecture: `BFF(Apisix)--->Dapr(MicroLogic)`

BFF router: https://domain.com/v1.0/invoke/${appid}.${namespace}/method/${method}

such as: 

curl -i "http://domain.com/v1.0/invoke/micrologic.default/method/activity/info?id=123"

For some reasons, we want to switch the service from cluster(k8s) mode to standalone mode, and at the same time, we want the BFF route to remain unchanged, because this will affect the front-end, so we want to import a fake (virtual) namespace to ensure that the front-end access does not change, only change backend service vip in BFF router.


|mode|namespace source|
|---|---|
|kubernetes|`NAMESPACE`|
|standanlone| if not exist `NAMESPACE`, select cmdline `namespace` param valid.|

1046102779:
A solution I just came up with to see how feasible it is:

We may introduce an instance in the middleware of the `components-contrib` repo: `middleware.http.routeralias`

```yaml
apiVersion: dapr.io/v1alpha1
kind: Component
metadata:
  name: routeralias 
spec:
  type: middleware.http.routeralias
  version: v1
  metadata:
  - name: "/v1.0/mall/activity/info"  # gray router.
    value: "/v1.0/invoke/micrologic.default/method/activity/info" # alias router.
```

```golang
// code: dapr/dapr/pkg/http/server.go: line 89.
// StartNonBlocking starts a new server in a goroutine.
func (s *server) StartNonBlocking() error {
        handler := useAPIAuthentication(
                s.useCors(
                        s.useComponents(  // middleware.http.routeralias
                                s.useRouter())))
```

/cc @ItalyPaleAle

pkg/runtime/runtime_test.go
```diff
@@ -3664,6 +3664,7 @@ func NewTestDaprRuntimeWithProtocol(mode modes.DaprMode, protocol string, appPor
 func NewTestDaprRuntimeConfig(mode modes.DaprMode, protocol string, appPort int) *Config {
 	return NewRuntimeConfig(NewRuntimeConfigOpts{
 		ID:                           TestRuntimeConfigID,
+		Namespace:                    os.Getenv("NAMESPACE"),
```
Taction, line 3667: Since this is used for tests, better to define a const string.

1046102779: Because uts need to test whether the writing and reading of namespace environment variables are as expected. If this is changed to the default value or fake value, there will be dependencies elsewhere, resulting in failure.

image

PR #5111, opened 2022-08-31
Run Arm64 nodepool to e2e test runs

addjuarez:
# Description
AKS has recently added support for arm64 node pools. So we are able to now add an arm64 aks cluster to run our e2e tests on. 
These changes allow for the deployment of an arm64 aks cluster and allows for the creation of the docker/app docker images for arm64 on github's  action runners.

## Issue reference

Adding an Arm64 nodepool to e2e workflows

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing


.github/workflows/dapr-test.yml
```diff
@@ -30,7 +30,11 @@ on:
   # Dispatch on external events
   repository_dispatch:
     types: [e2e-test]
-
+  pull_request:
```
ItalyPaleAle, line 33: Please don't add this. For PRs, we will need to use `/ok-to-test`.

During development on your own, you will probably need to create your own fork. Feel free to ping me offline and I'll explain how!

.github/workflows/dapr-test.yml
```diff
@@ -306,8 +315,17 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        target_os: [linux, windows]
-        target_arch: [amd64]
+        os: [ubuntu-latest, windows-2019]
```
ItalyPaleAle, line 318: Unlike in the "build" step, this is used in a different way and all the Actions runners are **always** on Linux/amd64. In this case, you want to keep `os` as `linux, windows` because that's not the OS of GitHub Actions, but just the OS used in AKS

tests/test-infra/azure-aks.bicep
```diff
@@ -17,6 +17,9 @@ param namePrefix string
 @description('The location of the resources')
 param location string = resourceGroup().location
 
+@description('If enabled, add a Arm64 pool')
```
ItalyPaleAle, line 20: Nit: `an ARM64 pool`

tests/test-infra/azure-aks.bicep
```diff
@@ -26,6 +29,9 @@ param linuxVMSize string = 'Standard_DS2_v2'
 @description('VM size to use for Windows nodes, if enabled')
 param windowsVMSize string = 'Standard_DS3_v2'
 
+@description('VM size to use for Arm64 nodes if enabled')
```
ItalyPaleAle, line 32: Nit here too, ARM64 is spelled uppercase :)

tests/test-infra/azure-aks.bicep
```diff
@@ -26,6 +29,9 @@ param linuxVMSize string = 'Standard_DS2_v2'
...
+param armVMSize string = 'Standard_D4pds_v5'
```
ItalyPaleAle, line 33: This seems a pretty beefy VM. On Linux we use a DS2 which has 2 cores only. This one has 4

tests/test-infra/azure-all.bicep
```diff
@@ -78,9 +82,33 @@ module windowsCluster 'azure.bicep' = {
...
+    enableArm64 : false
     diagLogAnalyticsWorkspaceResourceId: diagLogAnalyticsWorkspaceResourceId
     diagStorageResourceId: diagStorageResourceId
     enableCosmosDB: enableCosmosDB
     enableServiceBus: enableServiceBus
   }
 }
+
+// Deploy the Arm cluster in the third location
+resource linuxResources 'Microsoft.Resources/resourceGroups@2020-10-01' = {
```
ItalyPaleAle, line 94: There's already a "linuxResources", so give this a different name

tests/test-infra/azure-all.bicep
```diff
@@ -78,9 +82,33 @@ module windowsCluster 'azure.bicep' = {
...
+  name: 'Dapr-E2E-${namePrefix}l'
```
ItalyPaleAle, line 95: Here too, needs a different suffix, not `l` which is used by Linux. For example `la`?

tests/test-infra/azure-all.bicep
```diff
@@ -78,9 +82,33 @@ module windowsCluster 'azure.bicep' = {
...
+  location: location3
+  tags: dateTag != '' ? {
+    date: dateTag
+  } : {}
+}
+module armCluster 'azure.bicep' = {
+  name: 'armCluster'
+  scope: ArmResources
+  params: {
+    namePrefix: '${namePrefix}l'
```
ItalyPaleAle, line 105: Here too, `la` as suffix?

tests/test-infra/azure-all.bicep
```diff
@@ -78,9 +82,33 @@ module windowsCluster 'azure.bicep' = {
...
+    location: location1
```
ItalyPaleAle, line 106: location 3

tests/test-infra/azure-all.bicep
```diff
@@ -78,9 +82,33 @@ module windowsCluster 'azure.bicep' = {
...
+    enableWindows: false
+    enableArm64 : false
```
ItalyPaleAle, line 108: True

ItalyPaleAle (changes requested):
Good work so far! Please also make sure to create a difference resource group for the ARM test, not just the one ending in `l`, but perhaps ending in `la`?

.github/workflows/dapr-test.yml
addjuarez: Yeah I only added this to test it.

tests/test-infra/azure-aks.bicep
addjuarez: I tried using a smaller one, but I received an VMCannotFitEphemeralOSDisk when trying to deploy.

.github/workflows/dapr-test.yml
```diff
@@ -101,15 +101,14 @@ jobs:
       - name: Deploy the test cluster
         if: env.TEST_PREFIX != ''
         run: |
-          # Select two random Azure regions
+          # Select three random Azure regions
```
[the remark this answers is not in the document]
ItalyPaleAle: The reason for the 2 regions for x86 was that we could potentially use more regions in the future and spread capacity if needed. Then, turned out that westus3 was the only one that was stable enough for our tests :)

.build-tools/cmd/zz-e2e-perf.go
```diff
@@ -311,6 +311,19 @@ func (c *cmdE2EPerf) buildDockerImage(cachedImage string) error {
 		"-t", destImage,
 		filepath.Join(appDir, c.flags.Name, "."),
 	)
+
+	if c.flags.TargetArch == "arm64" {
```
ItalyPaleAle, line 315: Should this be a switch instead?

```
args := []string{
		"build",
		"-f", dockerfile,
		"-t", destImage,
		filepath.Join(appDir, c.flags.Name, "."),
	}
	switch c.flags.TargetArch {
	case "arm64":
		args = append(args, "--platform", "linux/arm64/v8")
		// Add --load if needed>?
	case "amd64":
		args = append(args, "--platform", "linux/amd64")
	default:
		args = append(args, "--platform", "linux/amd64")
	}
	e := exec.Command("docker", args...)
```

This way should also be possible to build amd64 images on an arm64 machine!

.github/workflows/dapr-test.yml
```diff
@@ -246,7 +251,7 @@ jobs:
         if: env.TEST_PREFIX != ''
         run: |
           make build
-          if [ "${{ env.TARGET_OS }}" = "linux" ] && [ "${{ env.TARGET_ARCH }}" != "arm" ]; then
+          if [ "${{ env.TARGET_OS }}" = "linux" ] && [ "${{ env.TARGET_ARCH }}" != "arm64" ]; then
```
ItalyPaleAle, line 254: This needs to keep "arm". Mariner images are available for "arm64" but are not available for "arm"

.github/workflows/dapr-test.yml
```diff
@@ -278,7 +283,7 @@ jobs:
       - name: Push Dapr container images
         if: env.TEST_PREFIX != ''
         run: |
-          if [ "${{ env.TARGET_OS }}" = "linux" ] && [ "${{ env.TARGET_ARCH }}" != "arm" ]; then
+          if [ "${{ env.TARGET_OS }}" = "linux" ] && [ "${{ env.TARGET_ARCH }}" != "arm64" ]; then
```
ItalyPaleAle, line 286: Same - needs to keep "arm"

ItalyPaleAle (changes requested):
Just a couple small changes and, assuming this works and you tested it, should be good to go for me.

Please also do update https://github.com/dapr/dapr/blob/master/tests/docs/running-e2e-test.md to explain the addition and how to enable ARM node pools

ItalyPaleAle (dismissed):
Approved, as long as it’s working!

.github/workflows/dapr-test.yml
```diff
@@ -120,17 +120,17 @@ jobs:
...
           REGIONS_IDX=$(($RANDOM % $REGIONS_SIZE))
           REGION1=${REGIONS[$REGIONS_IDX]}
           echo "REGION1=${REGION1}" >> $GITHUB_ENV
-          REGIONS_IDX=$(($RANDOM % $REGIONS_SIZE))
           REGION2=${REGIONS[$REGIONS_IDX]}
           echo "REGION2=${REGION2}" >> $GITHUB_ENV
-          echo "Deploying to Azure regions: Linux=${REGION1} Windows=${REGION2}"
-
+          REGION3=${{ env.AZURE_ARM_REGIONS }}
+          echo "REGION3=${REGION3}" >> $GITHUB_ENV
```
ItalyPaleAle, line 132: Please also add below in the table in the message that there's this region too (see lines 173 and 188)

artursouza (approved):

Issue #5056, opened 2022-08-21
E2E tests consistently failing on Windows


ItalyPaleAle:
Since about 2 days ago (Friday), our E2E tests are consistently failing on Windows (and only on Windows), with actor tests failing almost every time. Current pass rate is about 10%.

Requests to actor apps fail with "connection refused".

Examples:
- https://github.com/dapr/dapr/actions/runs/2899097648
- https://github.com/dapr/dapr/actions/runs/2898743651
- https://github.com/dapr/dapr/actions/runs/2898057237
- https://github.com/dapr/dapr/actions/runs/2897418321
- https://github.com/dapr/dapr/actions/runs/2896933383
- https://github.com/dapr/dapr/actions/runs/2896604126

skyao:
There are many different errors in different action run. 

I need to find the reason for them case by case.

skyao:
### Start / End time checking fail

https://github.com/dapr/dapr/runs/8027778812?check_suite_focus=true

```bash
=== FAIL: tests/e2e/actor_features TestActorFeatures/Actor_concurrency_different_actor_ids. (10.20s)
2022/08/26 00:39:48 POST http://20.14.110.190:3000/test/testactorfeatures/1004b/method/concurrency?reqid=aad6a24b-3d62-49a8-9319-07a85f9e329a completed in 5.07398099s
2022/08/26 00:39:54 POST http://20.14.110.190:3000/test/testactorfeatures/1004a/method/concurrency?reqid=b76ecc76-2b66-47cc-9ff1-cdea08f93f01 completed in 10.152406245s
2022/08/26 00:39:54 GET http://20.14.110.190:3000/test/logs?reqid=f9bb9465-001c-4398-8e91-674c7a8abbc1 completed in 51.060052ms
    actor_features_test.go:740: 
        	Error Trace:	/home/runner/work/dapr/dapr/tests/e2e/actor_features/actor_features_test.go:740
        	Error:      	Should be true
        	Test:       	TestActorFeatures/Actor_concurrency_different_actor_ids.
    --- FAIL: TestActorFeatures/Actor_concurrency_different_actor_ids. (10.20s)
```

For the code of test case:

```go
require.True(t, (logOne.StartTimestamp >= logTwo.EndTimestamp) || (logTwo.StartTimestamp >= logOne.EndTimestamp))
```

From the time information in the logs

```
[...]

PR #5113, opened 2022-09-01
update external URL after app restarted

skyao:
Signed-off-by: Sky Ao <aoxiaojian@gmail.com>

# Description

Try to confirm if the test case failed on windows is due to the external URL changed after app restarted.

## Issue reference

https://github.com/dapr/dapr/issues/5056

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing


ItalyPaleAle:
This shouldn't be the case however, right? The external URL is determined by the load balancer which is determined by the K8s Service. Whether the app is restarted or not, that shouldn't matter?

skyao:
> This shouldn't be the case however, right? The external URL is determined by the load balancer which is determined by the K8s Service. Whether the app is restarted or not, that shouldn't matter?

You are right! Also confirmed by test case, this should not happen.

See more details in  https://github.com/dapr/dapr/issues/5056#issuecomment-1234975790

Issue #4475, opened 2022-04-05
Proposal: HTTP Pipeline isn't executed in service-to-service invocation


ewassef:
## In what area(s)?
 

  /area runtime
 
## What version of Dapr?
 
> 1.6
 

## Expected Behavior

If I am setting an HTTP pipeline in a configuration and the service is an HTTP service. I expect the svc-to-svc invocation to go through this pipeline. This seems to be omitted due to the grpc channel used for the execution. This should invoke the local appchannel AFTER the grpc pipeline to keep the functionality

## Actual Behavior

Direct invocation to the *-dapr service or from within the container will execute the pipeline, any svc to svc invocation will not

## Steps to Reproduce the Problem

I have asked this question in Discord (https://discord.com/channels/778680217417809931/846943333824987156/959559729359753216)
I have asked this question in StackOverflow (https://stackoverflow.com/questions/71711579/opa-middleware-with-dapr-not-invoking)
I have a repo with repo scripts (https://github.com/ewassef/opa-dapr) 

## Release Note
 

RELEASE NOTE: **FIX** Bug in runtime. HTTP Pipeline invocation executes as designed

@yaron2 FYI

berndverst:
Sidecar to Sidecar communication is always done via gRPC.

You have to choice whether to call your sidecar with gRPC or HTTP, but for sidecar to sidecar communication it is gRPC only.

ewassef:
@berndverst pls take a look at the repo above. This is using the sample from the dapr.io website and a service calling another service that has an HTTP Pipeline configured.

ewassef:
@yaron2 @berndverst any luck?

360build:
I am facing the same problem.

Configured 2 services: Service A and Service B where service A is talking to service B
Service B is configured with a DAPR middle ware. This middle ware is an OPA configuration witch is checking an AD access token. 

When making direct calls to the service B from an other pod or outside of the cluster the middle-ware gets called. :
| Service A |  --> | dapr ---> service B| 
With these I have to give a valid JWT token to be able to access Service B.

When I make the call via dapr:     
           | service A -> dapr | ---> | dapr ---> service B |   
the middle ware is not called. It only relies on the dapr configuration to allow or deny access. I does not mather if my JWT token is valid or not.

artursouza:
For gRPC, the HTTP middleware should not apply. We tried making a mapping between the HTTP and gRPC protocols and it falls short. Each middleware component should be protocol specific IMO.

[...]

PR #5132, opened 2022-09-05
App http channel middlewares

mcandeia:
# Description

This PR adds support to app http channel middlewares.

Currently, our http middlewares don't execute in a service-to-service invocation context, and some of our middlewares are intented to be executed regardless of used dapr protocol.

App Http Channel middlewares solve this by always executing in a context of the app channel protocol, meaning they are also applied when making service-to-service invocations.

The middleware component itself still the same as used for dapr HTTP API, it could be reused with the caveat of having a mapping from fasthttp.Request to fasthttp.RequestCtx. although they are very similar they don't provide same set of features.

> in draft until having an e2e test using service-to-service invocation + middleware execution

## Issue reference

Please reference the issue this PR will close: #4475 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Unit tests passing
* [x] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#2778
* [x] Specification has been updated / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#2778
* [x] Provided sample for the feature / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#2778


artursouza (approved):

artursouza:
@mcandeia Please, resolve conflicts.

pkg/channel/http/http_channel.go
```diff
@@ -235,24 +240,29 @@ func (h *Channel) invokeMethodV1(ctx context.Context, req *invokev1.InvokeMethod
 	diag.DefaultHTTPMonitoring.ClientRequestStarted(ctx, verb, req.Message().Method, int64(len(req.Message().Data.GetValue())))
 	startRequest := time.Now()
 
-	// Send request to user application
-	resp := fasthttp.AcquireResponse()
```
yaron2, line 239: Why is this removed?

mcandeia: Sadly, this was a side-effect of converting from fasthttp.Request to fasthttp.RequestCtx,

One solution was simply getting one fasthttp.Response from the pool and set on the fasthttp.RequestCtx, but it's not possible because responses doesn't support copying, and fasthttp.RequestCtx requires a value rather than pointer,

I can use ctx.Response.CopyTo(&resp) but as per [documentation](https://pkg.go.dev/github.com/valyala/fasthttp#Response.CopyTo) it doesn't copy the body stream

image

Also, if the idea is to provide response recycling and less GC pressure I think that copying won't work anyways,

Any suggestion on this @yaron2 ?

mcandeia: I'm completely aligned if you think this is a blocker for this PR. Especially if there have been previous experiences that resulted in more memory usage and increased latency (related to GC pauses)

yaron2: Currently I'd put this as a blocker but it doesn't mean it'll remain that way unless fixed. We need to be able to measure the performance hit here. An immediate solution would be to revert to the current (and more performant) request construction if no middleware is present.

mcandeia: Great idea @yaron2. Applied your suggestions

> We need to be able to measure the performance hit here

Yes, I'll probably ask `@tanvigour` for help to look at the graphs we already have that the memory consumption/p99 latency increases

mcandeia:
@yaron2 @artursouza 

I’m afraid I don’t have bandwidth to get this ready until 1.9.

Do you guys agree on adding a feature flag for this - and launch as an alpha feature, and let the performance impact measurement as a prerequisite to become stable?

From the user perspective there’s no breaking change involved for future releases

artursouza:
> @yaron2 @artursouza
> 
> I’m afraid I don’t have bandwidth to get this ready until 1.9.
> 
> Do you guys agree on adding a feature flag for this - and launch as an alpha feature, and let the performance impact measurement as a prerequisite to become stable?
> 
> From the user perspective there’s no breaking change involved for future releases

This feature should go in 1.9 - putting behind a feature flag makes sense. @yaron2 would this unblock this to be in 1.9?

yaron2:
> > @yaron2 @artursouza
> > I’m afraid I don’t have bandwidth to get this ready until 1.9.
> > Do you guys agree on adding a feature flag for this - and launch as an alpha feature, and let the performance impact measurement as a prerequisite to become stable?
> > From the user perspective there’s no breaking change involved for future releases
> 
> This feature should go in 1.9 - putting behind a feature flag makes sense. @yaron2 would this unblock this to be in 1.9?

If the feature flag guarantees no performance hit then I am ok with this for 1.9

mcandeia:
Perfect, I think it’s already covered by the changed code, if no middleware are specified the actual behavior is kept.

I’ll add a feature for extra caution, so both should be satisfied, pipeline specified and flag enabled, thanks

mcandeia:
Done @artursouza

yaron2:
> > @yaron2 @artursouza
> > I’m afraid I don’t have bandwidth to get this ready until 1.9.
> > Do you guys agree on adding a feature flag for this - and launch as an alpha feature, and let the performance impact measurement as a prerequisite to become stable?
> > From the user perspective there’s no breaking change involved for future releases
> 
> This feature should go in 1.9 - putting behind a feature flag makes sense. @yaron2 would this unblock this to be in 1.9?

Ok, to be clear: what is missing in your eyes from making this ready?

Also, if we retain the pooled responses if no middleware is specified, then I don't think we need a feature flag. But that depends on the answer to my first question :)

mcandeia:
From my perspective this feature is done - when considering it for alpha/preview release,

What's missing: Measure the performance hit when using it - if no middleware is specified nothing changes from the current behavior as @yaron2 suggested

> Also, if we retain the pooled responses if no middleware is specified, then I don't think we need a feature flag. But that depends on the answer to my first question :)

Well, if so, I can revert my previous commit.

artursouza (approved):

ItalyPaleAle:
I just discovered this PR while working on a merge conflict for #5170. In there, fasthttp is being removed from the HTTP channel (and it will have to be removed from Dapr entirely in #4979), so we may need to figure out what to do with this if we want to merge streaming in 1.10.

msfussell:
Why do we add features like this without any explanation to the "why" of this, even if this may seem important?

yaron2:
> @artursouza @mcandeia @yaron2 Why do we add features like this without any explanation to the "why" of this, even if this may seem important? And then there is no communication on this until a doc issue arrives. Looks like this is not just a surprise to me either.

The rational and why are all in the referenced issue: https://github.com/dapr/dapr/issues/4475

PR #5143, opened 2022-09-07
Fix data race in state config

sunzhaochang:
Signed-off-by: sunzhaochang <zhchsun1992@gmail.com>

# Description

Fix data race in state config

## Issue reference

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing

### DIFF
--- a/pkg/components/state/state_config.go
+++ b/pkg/components/state/state_config.go
@@ -17,6 +17,7 @@ import (
 	"fmt"
 	"os"
 	"strings"
+	"sync"
 
 	"github.com/pkg/errors"
 )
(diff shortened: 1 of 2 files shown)


pkg/components/state/state_config.go
```diff
@@ -96,6 +100,8 @@ func GetOriginalStateKey(modifiedStateKey string) string {
 }
 
 func getStateConfiguration(storeName string) *StoreConfiguration {
+	statesConfigurationLock.Lock()
```
shubham1172, line 103: Is it better to use `RWMutex` with `RLock` here?

sunzhaochang:
> Just curious, is frequent configuration modification a common scenario?

@daixiang0 It is not common in fact, there is just little chance of race condition.

pkg/components/state/state_config.go
sunzhaochang: The statesConfiguration is also updated in this function. see: https://github.com/dapr/dapr/blob/master/pkg/components/state/state_config.go#L102

shubham1172: Just saw that, thanks!

artursouza (approved):

Issue #5149, opened 2022-09-09
feature: Kubernetes Mode Support for Pluggable Components


mcandeia:
## In what area(s)?

/area runtime

/area operator

> /area placement

> /area docs

> /area test-and-release

## Describe the feature

[Original issue](https://github.com/dapr/dapr/issues/4925)

In order to list all available pluggable components a new operator API should be added,

```proto
service Operator {
    // ...
    rpc ListPluggableComponents(ListPluggableComponentsRequest) returns (ListPluggableComponentsResponse) {}
}
```

**This API must not be namespaced.** The rationale is that the pluggable component makes no sense without a component pointing to it. So if there's a component depending on a pluggable component (i.e its type is pointing to a pluggable component) so it should be available to configure and use it without namespace restriction, as it used to be for the built-in ones, which means that pluggable component CRD should be cluster-wide as the usage of  (let's say)`state.redis` is today.

```proto
// ListPluggableComponentsRequest is the request to get pluggable components for a sidecar [edited ~in a namespace~].
message ListPluggableComponentsRequest {
  string namespace = 1;
  string pod_name = 2;
}

// ListPluggableComponentsResponse includes the list of available pluggable components.
message ListPluggableComponentsResponse {
  repeated bytes pluggable_components = 1;
}
```

This will return all available cluster-wide pluggable components.

## Release Note

RELEASE NOTE:

yaron2:
I've read this several times and cannot understand why this needs to be cluster-wide.

> so it should be available to configure and use as it used to be for the built-in ones

What is `it` referring to in this sentence? a pluggable component? 

> It means that pluggable component CRD should be cluster-wide.

I don't understand the rationale behind this. Say we have two namespaces, A and B. There are two state store components, with the same name, each deployed to the same namespace. Now developers want to add corresponding pluggable components for components, also with the same name:

1. Why wouldn't they namespace them just like regular components and most CRDs?
2. If cluster-wide, how would they keep the same name?

yaron2:
> This API must not be namespaced

> This will return all available cluster-wide pluggable components.

But then in the request you require a namespace, which will return all pluggable components for the requested namespace:

```go
string namespace = 1;
```

Am I missing something?

mcandeia:
You're right, it got a little confused reading it now again,

[...]

PR #5151, opened 2022-09-09
Kubernetes support for Pluggable Components

mcandeia:

# Description

This PR adds kubernetes mode support for pluggable components. New operators API were created.

Also I've made a few changes in the Pluggable Components CRD, the most important one was moving from namespaced to cluster-wide, check the issue for further information.

To see it working, deploy the current version in a kubernetes cluster

1) Inside dapr/charts/dapr
change the `values.yaml` inside chart folder to:

```yaml
global:
  registry: ghcr.io/mcandeia
  tag: 1.9.0-pluggable-components-alpha.2-linux-amd64
  dnsSuffix: ".cluster.local"
  logAsJson: false
  imagePullPolicy: IfNotPresent
  imagePullSecrets: ""
  nodeSelector: {}
  tolerations: []
  ha:
    enabled: false
    replicaCount: 3
    disruption:
      minimumAvailable: ""
      maximumUnavailable: "25%"
  prometheus:
    enabled: true
    port: 9090
  mtls:
    enabled: true
    workloadCertTTL: 24h
    allowedClockSkew: 15m
  daprControlPlaneOs: linux
  labels: {}
```

2) Install dapr release
```sh
helm install dapr .
```
> Notice that I've changed the versions to point to a personal image registry

3)
Add component and pluggable component CRD using `kubectl apply -f ...`
```yaml
apiVersion: dapr.io/v1alpha1
kind: PluggableComponent
metadata:
  name: store
spec:
  type: state
  version: v1
---
apiVersion: dapr.io/v1alpha1
kind: Component
metadata:
  name: default
spec:
  type: state.store
  version: v1
  initTimeout: 10s
  metadata:
    - name: timeout
      value: 100s
```

Deploy the dapr with the pluggable component container using `kubectl apply -f ...`

4)
```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: sample
  labels:
    app: sample
spec:
  replicas: 1
  selector:
    matchLabels:
      app: sample
  template:
    metadata:
      labels:
        app: sample
      annotations:
        dapr.io/unix-domain-socket-path: "/var/run"
        dapr.io/app-id: "backend"
        dapr.io/enabled: "true"
        dapr.io/sidecar-listen-addresses: "0.0.0.0"
    spec:
      volumes:
      - name: dapr-unix-domain-socket
        emptyDir: {}
      containers:
        - name: in-memory-pluggable
          volumeMounts:
              - name: dapr-unix-domain-socket
                mountPath: /dapr-unix-domain-sockets
          image: tmacam/dapr-memstore-java:latest
          resources:
            requests:
              memory: "64Mi"
              cpu: "250m"
            limits:
              memory: "128Mi"
              cpu: "500m"
          env:
            - name: DAPR_COMPONENT_SOCKET_PATH
              value: /dapr-unix-domain-sockets/dapr-state.store-v1-default.sock

```

> Notice that I reused the current unix socket directory, we might need to change that to have a specific folder for components - perhaps one file per component in a different volume mount?  It brings more security given that one context cannot access the application data neither others components data.

Forward daprd http port, then you can try out using http calls.

```sh
curl --location --request POST 'http://localhost:$PORT/v1.0/state/default' \
--header 'Accept: text' \
--header 'Content-Type: application/json' \
--data-raw '[{ "key": "name", "value": "Bruce s", "metadata": { "ttlInSeconds": "60"}}]'
```

```sh
curl --location --request GET 'http://localhost:$PORT/v1.0/state/default/name' \
--header 'Accept: text'
```

> Replace $PORT for the forwarded port

## Issue reference

Please reference the issue this PR will close: #5149 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Unit tests passing


yaron2 (changes requested):
Overall looks ok, please perform the changes requested here https://github.com/dapr/dapr/issues/5149#issuecomment-1242822038.

mcandeia:
applied your suggestions @yaron2, I'll keep it in draft until have at least one e2e test working

mcandeia:
@artursouza @yaron2 I managed to have an e2e test working with pluggable components, but for that I had to change our test framework a bit to support the additional containers,

Two questions:

1) Do you guys prefer a separate PR from this one or I can keep the changes?
2) I'm using a [pluggable component image](https://github.com/dapr/dapr/pull/5151/files#diff-50213871251d4d0b426f8a7e43739423f65380c0425ff8e0396d51de79c85813R409) from my personal fork (the wrapped redis pluggable component [here](https://github.com/dapr-sandbox/components-go-sdk/blob/main/examples/redis/main.go#L25), any suggestions on this? Maybe publishing it in the dapr-sandbox or perhaps deploy as an ordinary e2etest app?

yaron2:
> @artursouza @yaron2 I managed to have an e2e test working with pluggable components, but for that I had to change our test framework a bit to support the additional containers,
> 
> Two questions:
> 
> 1. Do you guys prefer a separate PR from this one or I can keep the changes?
> 2. I'm using a [pluggable component image](https://github.com/dapr/dapr/pull/5151/files#diff-50213871251d4d0b426f8a7e43739423f65380c0425ff8e0396d51de79c85813R409) from my personal fork (the wrapped redis pluggable component [here](https://github.com/dapr-sandbox/components-go-sdk/blob/main/examples/redis/main.go#L25)), any suggestions on this? Maybe publishing it in the dapr-sandbox registry instead or perhaps deploy as an ordinary e2etest app?

I'm ok with keeping the changes.

Re: image, we can't merge as long as it's using an image that's not owned by the Dapr project.

mcandeia:
can you see if it looks better now @yaron2 ?

tests/config/pluggable_state.yaml
```diff
@@ -0,0 +1,48 @@
...
+    - name: redisHost
+      secretKeyRef:
+        name: redissecret
+        key: host
+    - name: redisPassword
+      value: ""
+scopes:
+  - stateapp-pluggable
+---
+# TODO: Remove once PluggableComponents feature is finalized
```
artursouza, line 40: You can turn on this feature by test app.

tests/dapr_tests.mk
```diff
@@ -445,6 +446,7 @@ setup-test-components: setup-app-configurations
 	$(KUBECTL) apply -f ./tests/config/dapr_vault_secretstore.yaml --namespace $(DAPR_TEST_NAMESPACE)
 	$(KUBECTL) apply -f ./tests/config/uppercase.yaml --namespace $(DAPR_TEST_NAMESPACE)
 	$(KUBECTL) apply -f ./tests/config/pipeline.yaml --namespace $(DAPR_TEST_NAMESPACE)
+	$(KUBECTL) apply -f ./tests/config/pluggable_state.yaml --namespace $(DAPR_TEST_NAMESPACE)
```
artursouza, line 449: Call it dapr_redis_pluggable_state.yaml

tests/config/pluggable_state.yaml
mcandeia: can you send an example of this?

I see others doing the same like: [resiliency](https://github.com/dapr/dapr/blob/master/tests/config/app_resiliency.yaml#L1) and [app healthcheck](https://github.com/dapr/dapr/blob/master/tests/config/app_healthcheck.yaml#L15)

tests/dapr_tests.mk
mcandeia: I guess your idea is to use the DAPR_TEST_STATE_STORE variable to switch between cosmos/redis/redis_pluggable is that correct?

If yes, wouldn't better to have both running at same time and switch between cosmos and redis as it is currently?

if I got it wrong, just ignore this comment.

mcandeia:
@yaron2 @artursouza can you guys take a look again? I believe the test that is failing is flaky.
btw: /ok-to-test doesn't seem to work

artursouza:
> @yaron2 @artursouza can you guys take a look again? I believe the test that is failing is flaky. btw: /ok-to-test doesn't seem to work

It is working but the workflow is not ideal. It does not comment on the issue when it starts (@ItalyPaleAle - can you fix that?) but it also does not comment if the build step fails - which was the case for this.

artursouza:
It is failing to build on Windows: https://github.com/dapr/dapr/actions/runs/3048187369/jobs/4912984388

mcandeia:
Fixed

tests/config/pluggable_state.yaml
artursouza: It is just in the same file here then. It is OK. I will not the PR for this.

artursouza (approved):

Issue #2218, opened 2020-10-07
Proposal: Support batching of send and receive pub/sub messages for performance


artursouza:
The current PubSub model notifies consumers by making a roundtrip per message. A way to improve message throughput would be to allow consumers to declare the intention of consuming messages in batches. Processing messages in batches lead to the following optimizations:
- roundtrip reduction between dapr and consuming component by sending messages batches
- consumer computation by observing a message batch instead of a single item

Expected behavior: Subscribers can (optionally) choose to receive messages in batches. 

Example of a subscriber's response where batching is requested:

```
{
  "subscriptions": [
    {
       "topic": "A",
       "maxBatchSize": "10"
    }
  ]
}
```

The optional parameter `maxBatchSize` defines the maximum number of messages to be delivered in a single receive loop. By default, maxBatchSize should be 1.

Message checkpointing in batch mode are for all messages (all succeeded or failed).

Distributed tracing should track each message inside the batch in such a way that users can still find them in the target telemetry backend. In other words, if messages A, B and C were processed in the same batch, searching for message B telemetry should return the batch execution.

In order to enable receiving multiple messages the subscriber callback contract could be changed from:

```go
// NewMessage is an event arriving from a message bus instance
type NewMessage struct {
	Data     []byte            `json:"data"`
	Topic    string            `json:"topic"`
	Metadata map[string]string `json:"metadata"`
}
```

to

```go
// Message is one event arriving from a message bus instance
type Message struct {
	Data []byte `json:"data"`
	Metadata map[string]string `json:"metadata"`
}

// NewMessages contains one or more events arriving from a message bus instance for a given topic
type NewMessages struct {
	Messages []Message `json:"messages"`
	Topic    string    `json:"topic"`
}
```

## Impact
* Components Contrib: each implementation needs to handle the new property (impact on certification), defaults to "1" otherwise. Still need to change code so it can return an array of 1 and avoid compilation error.
* For GRPC, this is a breaking change given the change in proto change. Any example for GRPC will need to be fixed in every SDK as well.
* For HTTP, we can make use of the lack of structure and send the payload to subscriber as-is (CloudEvent) when batch size is configured to 1. When batch size is configured to greater value, the payload to subscriber is an array of CloudEvent.

Parent: #843 
Note: At time of code checkin, submit a docs PR on this issue https://github.com/dapr/docs/issues/2407

artursouza:
[...]

PR #5154, opened 2022-09-12
Add Initial http bulk subscribe Implementation

DeepanshuA:

# Description

It implements http bulk subscribe implementation in runtime.

## Issue reference

Please reference the issue this PR will close: #2218 

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing


go.mod
```diff
@@ -393,7 +393,8 @@ replace github.com/eclipse/paho.mqtt.golang => github.com/shivamkm07/paho.mqtt.g
 // Uncomment for local development for testing with changes in the components-contrib repository.
 // Don't commit with this uncommented!
 //
-// replace github.com/dapr/components-contrib => ../components-contrib
+replace github.com/dapr/components-contrib => github.com/DeepanshuA/components-contrib v1.6.0-rc.2.0.20220912174448-d1d3384d6575
```
DeepanshuA, line 396: Need to revert this change after PR introducing bulk pubsub interfaces is merged

ItalyPaleAle: Ok for using a temp replace, but please keep it at the bottom and don't replace this line, which is a hint to contributors for using a local clone.

pkg/runtime/pubsub/bulksubscribe_events.go
```diff
@@ -0,0 +1,73 @@
...
+	Topic     string
+	Pubsub    string
+	EventType string
+}
+
+func NewBulkSubscribeEnvelope(req *BulkSubscribeEnvelope) map[string]interface{} {
+	// defaults
+	id := req.ID
+	if id == "" {
+		id = uuid.New().String()
```
ItalyPaleAle, line 51: If there isn't enough entropy in the kernel, this will panic. Do we use this emthod in other hot paths? Otherwise, would be better to use the method that returns an error rather than panics in case of lack of entropy.

pkg/runtime/config.go
```diff
@@ -50,6 +50,8 @@ const (
 	DefaultGracefulShutdownDuration = time.Second * 5
 	// DefaultAppHealthCheckPath is the default path for HTTP health checks.
 	DefaultAppHealthCheckPath = "/health"
+	// Metadata needed to call bulk subscribe
+	BulkSubscribe = "bulkSubscribe"
```
ItalyPaleAle, line 54: Because this isn't a config option, can you put it in the runtime.go file?

pkg/runtime/runtime.go
```diff
@@ -1747,6 +1765,23 @@ func (a *DaprRuntime) initPubSub(c componentsV1alpha1.Component) error {
 	return nil
 }
 
+func (a *DaprRuntime) BulkPublish(req *pubsub.BulkPublishRequest) error {
+	ps, ok := a.pubSubs[req.PubsubName]
+	if !ok {
+		return runtimePubsub.NotFoundError{PubsubName: req.PubsubName}
+	}
+
+	if allowed := a.isPubSubOperationAllowed(req.PubsubName, req.Topic, ps.scopedPublishings); !allowed {
```
ItalyPaleAle, line 1774: Nit:

```
if !a.isPubSubOperationAllowed(req.PubsubName, req.Topic, ps.scopedPublishings) {
```

pkg/runtime/runtime.go
```diff
@@ -1878,6 +1913,272 @@ func (a *DaprRuntime) initNameResolution() error {
 	return nil
 }
 
+func (a *DaprRuntime) sendBulkToDeadLetter(
+	name string, msg *pubsub.BulkMessage, deadLetterTopic string,
+	bulkResponses []pubsub.BulkSubscribeResponseEntry,
+) error {
+	data := make([]pubsub.BulkMessageEntry, 0)
```
ItalyPaleAle, line 1920: You can pre-allocate this since you know the size for avoiding lots of allocations in a hot path.

```
data := make([]pubsub.BulkMessageEntry, len(msg.Entries))
```

then:

```
n := 0
for _, message := range msg.Entries {
    ...
    if bulkResponses == nil || bulkResponses[entryID].Error != nil {
		data[n] = message
    }
}
data = data[:n]
```

pkg/runtime/runtime.go
```diff
@@ -1878,6 +1913,272 @@ func (a *DaprRuntime) initNameResolution() error {
...
+					// dlq has been configured and whole bulk of messages is successfully sent to dlq.
+					diag.DefaultComponentMonitoring.PubsubIngressEvent(ctx, pubsubName, strings.ToLower(string(pubsub.Drop)), msg.Topic, 0)
+					return nil, nil
+				}
+			}
+			diag.DefaultComponentMonitoring.PubsubIngressEvent(ctx, pubsubName, strings.ToLower(string(pubsub.Retry)), msg.Topic, 0)
+			return nil, err
+		}
+
+		cloudEvents := make([]map[string]interface{}, 0)
```
ItalyPaleAle, line 1972: Here too you should be able to pre-allocate this with a given size.

pkg/runtime/runtime.go
```diff
@@ -1878,6 +1913,272 @@ func (a *DaprRuntime) initNameResolution() error {
...
+	return err
+}
+
+func (a *DaprRuntime) bulkSubscribeTopic(ctx context.Context, policy resiliency.Runner,
+	name string, topic string, route TopicRouteElem,
+) error {
+	err := a.pubSubs[name].component.BulkSubscribe(ctx, pubsub.SubscribeRequest{
+		Topic:    topic,
+		Metadata: route.metadata,
+	}, func(ctx context.Context, msg *pubsub.BulkMessage) ([]pubsub.BulkSubscribeResponseEntry, error) {
```
ItalyPaleAle, line 1951: Is there a way this method can be "merged" into the handler above for the "regular" subscribe? Looks lilke a lot of code is duplicated, and then there's another method to maintain...

ItalyPaleAle (changes requested):

pkg/runtime/runtime.go
DeepanshuA: Actually, this handler's input parameters, return types and working is totally different than normal subscribe handler. 
It definitely takes inspiration from the normal subscribe handler but treatment of incoming item(s) and thus the code is as well different.

ItalyPaleAle: Ok

pkg/runtime/config.go
DeepanshuA: Moved to runtime.go

pkg/runtime/runtime.go
DeepanshuA: Done

pkg/runtime/runtime.go
DeepanshuA: Done

pkg/runtime/pubsub/bulksubscribe_events.go
DeepanshuA: Agreed, good suggestion. I have used uuid.NewRandom() instead now.

go.mod
DeepanshuA: Plan to remove this reference - as first https://github.com/dapr/components-contrib/pull/2024 would be merged.

pkg/runtime/bulk_subscriber.go
```diff
@@ -0,0 +1,503 @@
...
+//  2. Iterate through each message and validate entryID is NOT blank
+//     2A. If it is a raw payload:
+//     2Aa. Get route path, if processable
+//     2Ab. If route path is non-blank, generate base64 encoding of event data
+//     and set contentType, if provided, else set to "application/octet-stream"
+//     2Ac. Finally, form a child message to be sent to app and add it to the list of messages,
+//     to be sent to app (this list of messages is registered against correct path in an internal map)
+//     2B. If it is NOT a raw payload (it is considered a cloud event):
+//     2Ba. Unmarshal it into a map[string]interface{}
+//     2Bb. If any error while unmatrshalling, send to DLQ if configured, else register error for this message
```
mukundansundar, line 52: Nitpick typo on unmarshalling

pkg/runtime/bulk_subscriber.go
```diff
@@ -0,0 +1,503 @@
...
+					bulkResponses[i].Error = nil
+					continue
+				}
+				rPath, routeErr := a.getRouteIfProcessable(ctx, route, &(msg.Entries[i]), i, &bulkResponses, cloudEvent, psName, topic)
+				if routeErr != nil {
+					hasAnyError = true
+					continue
+				}
+				if rPath == "" {
+					continue
```
mukundansundar, line 148: What does it mean when route is nil, the message cannot be sent to app so why not set the bulk response to component as error there it self?

pkg/runtime/bulk_subscriber.go
```diff
@@ -0,0 +1,503 @@
...
+			}
+		}
+		for path, psm := range routePathBulkMessageMap {
+			invokeErr := a.createEnvelopeAndInvokeSubscriber(ctx, psm, topic, psName, msg, route, &bulkResponses, &entryIDIndexMap, path, policy)
+			if invokeErr != nil {
+				hasAnyError = true
+				err = invokeErr
+			}
+		}
+		if hasAnyError && err != context.Canceled {
```
mukundansundar, line 160: Qq what happens when hasAnyError is true and error is overwritten by any of the following lines since the error occurred and it is context cancelled...then we are not considering that scenario here... Would that lead to messages not being sent to dlq?

pkg/runtime/pubsub/bulksub.go
```diff
@@ -0,0 +1,161 @@
...
+	"github.com/google/uuid"
+
+	"github.com/dapr/dapr/utils"
+
+	contribPubsub "github.com/dapr/components-contrib/pubsub"
+)
+
+const (
+	maxBulkCountKey     string = "maxBulkCount"
+	maxBulkLatencyMsKey string = "maxBulkLatencyMilliSeconds"
```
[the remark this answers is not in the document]
DeepanshuA: Yeah, updated with correct property name.

pkg/runtime/bulk_subscriber.go
DeepanshuA: Actually, this route is calculated against each message/entry inside a bulk message; so we just need to catch appropriate error for this message only (if any), and then continue for other messages/entries in this bulk message.

ItalyPaleAle:
FYI made changes in a PR against this PR: https://github.com/DeepanshuA/dapr/pull/29

ItalyPaleAle (approved):

DeepanshuA:
> FYI made changes in a PR against this PR: [DeepanshuA#29](https://github.com/DeepanshuA/dapr/pull/29)

thanks @ItalyPaleAle , for taking care of these Misc fixes.

pkg/runtime/pubsub/bulksub.go
```diff
@@ -0,0 +1,161 @@
...
+}
+
+// flushMessages writes messages to a BulkHandler and clears the messages slice.
+func flushMessages(ctx context.Context, topic string, messages []contribPubsub.BulkMessageEntry, msgCbMap map[string]func(error), handler contribPubsub.BulkHandler) (
+	[]contribPubsub.BulkMessageEntry, map[string]func(error),
+) {
+	if len(messages) > 0 {
+		responses, err := handler(ctx, &contribPubsub.BulkMessage{
+			Topic:    topic,
+			Metadata: map[string]string{},
```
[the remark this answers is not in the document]
DeepanshuA: Have kept Metadata in BulkMessage with view that if in some use case, we may need to provide a metadata on bulk level, across all messages. Right now, it doesn't seem to require any data in default impl.

pkg/runtime/bulk_subscriber.go
```diff
@@ -0,0 +1,507 @@
...
+	req.WithCustomHTTPMetadata(msg.metadata)
+
+	n := 0
+	for _, cloudEvent := range msg.cloudEvents {
+		if cloudEvent[pubsub.TraceIDField] != nil {
+			traceID := cloudEvent[pubsub.TraceIDField].(string)
+			sc, _ := diag.SpanContextFromW3CString(traceID)
+			spanName := fmt.Sprintf("pubsub/%s", msg.topic)
+			var span trace.Span
+			ctx, span = diag.StartInternalCallbackSpan(ctx, spanName, sc, a.globalConfig.Spec.TracingSpec)
```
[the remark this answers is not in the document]
DeepanshuA: I found this one working well in case of normal subscription and for tracing, my use case is almost similar, where-in I trace against the published traceID only, if any. I found this working well in bulk subscription case as well.

pkg/runtime/bulk_subscriber.go
```diff
@@ -0,0 +1,507 @@
...
+		// When adding/removing an error here, check if that is also applicable to GRPC since there is a mapping between HTTP and GRPC errors:
+		// https://cloud.google.com/apis/design/errors#handling_errors
+		log.Errorf("non-retriable error returned from app while processing bulk pub/sub event: %s. status code returned: %v", body, statusCode)
+		diag.DefaultComponentMonitoring.BulkPubsubIngressEvent(ctx, msg.pubsub, strings.ToLower(string(pubsub.Drop)), msg.topic, elapsed)
+		populateBulkSubscribeResponsesWithError(msg.entries, bulkResponses, &entryIDIndexMap, nil)
+		return nil
+	}
+
+	// Every error from now on is a retriable error.
+	log.Warnf("retriable error returned from app while processing bulk pub/sub event, topic: %v, body: %s. status code returned: %v", msg.topic, body, statusCode)
```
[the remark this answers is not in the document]
DeepanshuA: Checked this one - this is done as it is done for all the left status codes, considering them as retriable. This is totally in sync with normal subscribe as well. And, the body here is of response, so that user can check back what message got passed back in response:
```	_, body := resp.RawData()
```

I don't think this should be considered as PII then, but if the view is against this, then I can remove it.

pkg/runtime/bulk_subscriber.go
DeepanshuA: It is a good question. I thought about it again that in this case, when do I actually need to check for context cancellation. I think it would be good to check for context cancellation, specifically when hearing back errror after invoke calls. So, changing logic a bit here. Please provide your thoughts. I am not individual invokeErrs but overall as want to return back complete bulkResponses, in any case.

pkg/runtime/bulk_subscriber.go
```diff
@@ -0,0 +1,507 @@
...
+// Major steps inside a bulk handler:
+//  1. Deserialize pubsub metadata and determine if rawPayload or not
+//     1A. If any error occurs, send to DLQ if configured, else send back error for all messages
+//  2. Iterate through each message and validate entryID is NOT blank
+//     2A. If it is a raw payload:
+//     2Aa. Get route path, if processable
+//     2Ab. If route path is non-blank, generate base64 encoding of event data
+//     and set contentType, if provided, else set to "application/octet-stream"
+//     2Ac. Finally, form a child message to be sent to app and add it to the list of messages,
+//     to be sent to app (this list of messages is registered against correct path in an internal map)
```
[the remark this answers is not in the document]
DeepanshuA: I converted to 2.A.i. type but one indentation is failing in lint with:
``` pkg/runtime/bulk_subscriber.go:50  gofmt  File is not `gofmt`-ed with `-s` ```

File: pkg/runtime/runtime.go [hunk 16]
```diff
@@ -721,72 +753,49 @@ func (a *DaprRuntime) subscribeTopic(parentCtx context.Context, name string, top
...
-
-		if pubsub.HasExpired(cloudEvent) {
-			log.Warnf("dropping expired pub/sub event %v as of %v", cloudEvent[pubsub.IDField], cloudEvent[pubsub.ExpirationField])
-			diag.DefaultComponentMonitoring.PubsubIngressEvent(ctx, pubsubName, strings.ToLower(string(pubsub.Drop)), msg.Topic, 0)
-
-			if route.deadLetterTopic != "" {
-				_ = a.sendToDeadLetter(name, msg, route.deadLetterTopic)
+			if !shouldProcess {
+				// The event does not match any route specified so ignore it.
+				log.Debugf("no matching route for event %v in pubsub %s and topic %s; skipping", cloudEvent[pubsub.IDField], name, msg.Topic)
```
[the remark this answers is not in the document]
DeepanshuA: Agreed, changed to Warnf

pkg/runtime/bulk_subscriber.go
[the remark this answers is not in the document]
halspang: I'd say that we should probably sanitize it out of all of them. I feel like this is the generally accept practice but maybe I'm just being over cautious. Thoughts @artursouza?

Regardless, we can do that as a separate issue once we decide on it, especially if it's the existing path.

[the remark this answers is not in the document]
artursouza: We should not log body since we don't know the content. It can also impact perf. At best, it can be a setting where it is OFF by default.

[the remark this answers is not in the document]
DeepanshuA: Removed setting of response `body` from all the places for now.

go.mod
```diff
@@ -396,8 +396,8 @@ replace github.com/eclipse/paho.mqtt.golang => github.com/shivamkm07/paho.mqtt.g
 // Uncomment for local development for testing with changes in the components-contrib repository.
 // Don't commit with this uncommented!
 //
-//replace github.com/dapr/components-contrib => ../components-contrib
-replace github.com/dapr/components-contrib => github.com/dapr/components-contrib v1.8.0-rc.1.0.20220915201717-3f9c42c4b8d9
+// replace github.com/dapr/components-contrib => ../components-contrib
+replace github.com/dapr/components-contrib => github.com/dapr/components-contrib v1.8.0-rc.1.0.20220923041529-e1d9a3bf9816
```
ItalyPaleAle, line 400: Please remove this before it's merged.

ItalyPaleAle (changes requested):

go.mod
DeepanshuA: Yeah, removed.

ItalyPaleAle (approved):

artursouza (approved):

Issue #5156, opened 2022-09-12
feature: Add Pluggable Components Feature Flag


mcandeia:
## In what area(s)?

/area runtime

> /area operator

> /area placement

> /area docs

> /area test-and-release

## Describe the feature

Add a new feature flag to enable/disable pluggable components since it is a preview feature.

## Release Note

RELEASE NOTE:

PR #5157, opened 2022-09-12
feature: Pluggable component feature flag

mcandeia:

# Description

This PR adds feature toggle for Pluggable Components.

## Issue reference

Please reference the issue this PR will close: #5156 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Unit tests passing

### DIFF
--- a/pkg/runtime/runtime.go
+++ b/pkg/runtime/runtime.go
@@ -630,6 +630,9 @@ func (a *DaprRuntime) buildHTTPPipeline() (httpMiddleware.Pipeline, error) {
 
 // registerPluggableComponents loads and register the loaded pluggable components.
 func (a *DaprRuntime) registerPluggableComponents() error {
+	if !config.IsFeatureEnabled(a.globalConfig.Spec.Features, config.PluggableComponents) {
+		return nil
+	}
 	pluggables, err := a.loadPluggableComponents()
 	if err != nil {
 		return err
--- a/pkg/runtime/runtime.go
+++ b/pkg/runtime/runtime.go
@@ -447,9 +447,7 @@ func (a *DaprRuntime) initRuntime(opts *runtimeOpts) error {
 	a.httpMiddlewareRegistry = opts.httpMiddlewareRegistry
 	a.lockStoreRegistry = opts.lockRegistry
 
-	if err = a.registerPluggableComponents(); err != nil {
-		log.Warnf("failed to register pluggable components: %s", err)
-	}
+	a.initPluggableComponents()
 
 	go a.processComponents()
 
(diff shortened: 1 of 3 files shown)


pkg/runtime/runtime.go
```diff
@@ -630,6 +630,9 @@ func (a *DaprRuntime) buildHTTPPipeline() (httpMiddleware.Pipeline, error) {
 
 // registerPluggableComponents loads and register the loaded pluggable components.
 func (a *DaprRuntime) registerPluggableComponents() error {
```
berndverst, line 632: IMO it should never call this function if the feature is not enabled. I would not put that within this function.

berndverst (changes requested):

pkg/runtime/runtime.go
mcandeia: Okay, I can do that, but I chose that function because we don't have any `initRuntime` tests and I think it's a bit difficult to do because it has a lot of dependencies - and I don't believe that this is the context of this PR. Having that said, I'd rather have the test than put it inside the initRuntime, otherwise I can create an initPluggableComponents() that does this check before calling the register, at least that way it can be "testable".

wdyt @berndverst ?

berndverst: I will leave this up to dapr/dapr maintainers.

We have already proven that our feature flags work, so we don't need to specifically test for that again.

IMO your approach is less readable. If the method were called `registerPluggableComponentsIfFeatureEnabled` it makes sense - but simply `registerPluggableComponents` is weird if I am debugging a scenario where I have the feature disabled.

Thoughts?

mukundansundar: IMO registerPluggableComponents should only register any pluggable component that is there. It should not check if the feature is enabled or not. And having an initPluggableComponents make sense in a way to modularize the code as well as making it testable. And that initialization code checking if it has to start registering pluggable components if feature is enabled makes its more clear Imo.

mcandeia: >IMO your approach is less readable. If the method were called registerPluggableComponentsIfFeatureEnabled it makes sense - but simply registerPluggableComponents is weird if I am debugging a scenario where I have the feature disabled. I would not expect that method to be entered in such a case.

I would agree with you if we had a single place where featureflags are initialized/checked, which is not the case. In that sense I prefer InitPluggableComponents because it turns the entrypoint for anyone to understand if the components will be loaded or not, even because I think the initRuntime code is getting too many lines and bad to maintain, so I think discussing name when the we have untested code ends up becoming irrelevant

mcandeia: > IMO registerPluggableComponents should only register any pluggable component that is there. It should not check if the feature is enabled or not. And having an initPluggableComponents make sense in a way to modularize the code as well as making it testable. And that initialization code checking if it has to start registering pluggable components if feature is enabled makes its more clear Imo.

I agree moving to there

artursouza (approved):
@tmacam We can start documenting this as well.

tmacam:
> @tmacam We can start documenting this as well.

Sure. Should this be added to `dapr-docs` directly or as part of the sandbox-projects?

PR #5153, opened 2022-09-12
[Work-In-Progress] gRPC implementation for workflow building block

akhilac1:
# Description
This is the initial gRPC implementation for workflow building block. 
Subsequent PRs will address doc and test

- Added proto ffor workflow start,terminate,get
- Implementation of gRPC API for workflow start/terminate/get
- 
## Issue reference
#4576 - Implementation of workflow building block

Please reference the issue this PR will close: #4576 partly

## Checklist

Please make sure you've  completed the relevant tasks for this PR, out of the following list:

* [X] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing
* [ ] End-to-end tests passing
* [ ] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_
* [ ] Specification has been updated / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_
* [ ] Provided sample for the feature / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_

PR #5158, opened 2022-09-12
Initial push for workflow api

RyanLettieri:
# Description

Note that this is essentially a duplicate of https://github.com/dapr/dapr/pull/4986 but with additional changes since the branch was very far behind and needed to be updated with recent changes to master

Note that this PR will fail until https://github.com/dapr/components-contrib/pull/2269 is merged into components contrib
## Issue reference

Related Proposal: https://github.com/dapr/dapr/issues/4576

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing


nyemade-uversky:
@akhilac1 and @RyanLettieri  would you both be able to sign off on the following:

1. Code from Akhila's PR for the gRPC code has been properly moved over to this PR (https://github.com/dapr/dapr/pull/5153)
2. The other PR on gRPC is OK to close

pkg/http/api.go
```diff
@@ -714,6 +748,133 @@ func (a *api) getLockStoreWithRequestValidation(reqCtx *fasthttp.RequestCtx) (lo
...
+		log.Error("No workflow, or empty workflow was provided.")
+		return
+	}
+
+	component := reqCtx.UserValue(workflowComponent).(string)
+	if component == "" {
+		log.Error("No component, or empty component was provided.")
+		return
+	}
+	if a.workflows[component] == nil {
```
cgillum, line 769: Cache the result of this map lookup into a local variable so that you can use the local variable instead of doing another map lookup on line 793.

pkg/http/api.go
```diff
@@ -714,6 +748,133 @@ func (a *api) getLockStoreWithRequestValidation(reqCtx *fasthttp.RequestCtx) (lo
...
+		return
+	}
+
+	component := reqCtx.UserValue(workflowComponent).(string)
+	if component == "" {
+		log.Error("No component, or empty component was provided.")
+		return
+	}
+
+	if a.workflows[component] == nil {
```
cgillum, line 827: Same comment about caching the map lookup result.

pkg/http/api.go
```diff
@@ -714,6 +748,133 @@ func (a *api) getLockStoreWithRequestValidation(reqCtx *fasthttp.RequestCtx) (lo
...
+		log.Error("No instance, or empty instance was provided.")
+		return
+	}
+
+	component := reqCtx.UserValue(workflowComponent).(string)
+	if component == "" {
+		log.Error("No component, or empty component was provided.")
+		return
+	}
+	if a.workflows[component] == nil {
```
cgillum, line 862: Same comment about caching the map lookup result.

pkg/runtime/options.go
```diff
@@ -22,6 +23,7 @@ type (
 		nameResolutionRegistry *nrLoader.Registry
 		bindingRegistry        *bindingsLoader.Registry
 		httpMiddlewareRegistry *httpMiddlewareLoader.Registry
+		workflowRegistry       *workflowsLoader.Registry
```
cgillum, line 26: `workflowRegistry` --> `workflowComponentRegistry`

pkg/runtime/options.go
```diff
@@ -85,6 +87,13 @@ func WithHTTPMiddlewares(registry *httpMiddlewareLoader.Registry) Option {
 	}
 }
 
+// WithWorkflows adds workflow components to the runtime.
+func WithWorkflows(registry *workflowsLoader.Registry) Option {
```
cgillum, line 91: `WithWorkflows` --> `WithWorkflowComponents`

pkg/runtime/runtime.go
```diff
@@ -474,7 +480,6 @@ func (a *DaprRuntime) initRuntime(opts *runtimeOpts) error {
 
 	// Start proxy
 	a.initProxy()
-
```
cgillum, line 477: I suggest reverting these newline changes since I don't think they're contributing to this PR or to readability generally.

pkg/runtime/runtime.go
```diff
@@ -1593,6 +1598,34 @@ func (a *DaprRuntime) initLock(s componentsV1alpha1.Component) error {
 	return nil
 }
 
+func (a *DaprRuntime) initWorkflow(s componentsV1alpha1.Component) error {
```
cgillum, line 1601: `initWorkflow` --> `initWorkflowComponent`

pkg/components/workflows/registry.go
```diff
@@ -0,0 +1,80 @@
...
+
+	wfs "github.com/dapr/components-contrib/workflows"
+	"github.com/dapr/dapr/pkg/components"
+	"github.com/dapr/kit/logger"
+)
+
+// Registry is an interface for a component that returns registered state store implementations.
+type Registry struct {
+	Logger    logger.Logger
+	workflows map[string]func(logger.Logger) wfs.Workflow
```
cgillum, line 29: I suggest renaming to either `workflowComponents` or just `components`.

pkg/components/workflows/registry.go
```diff
@@ -0,0 +1,80 @@
...
+}
+
+func (s *Registry) Create(name, version string) (wfs.Workflow, error) {
+	if method, ok := s.getWorkflow(name, version); ok {
+		return method(), nil
+	}
+	return nil, errors.Errorf("couldn't find wokflow %s/%s", name, version)
+}
+
+func (s *Registry) getWorkflow(name, version string) (func() wfs.Workflow, bool) {
```
cgillum, line 56: `getWorkflow` --> `getWorkflowComponent`

dapr/proto/runtime/v1/dapr.proto
```diff
@@ -654,4 +664,34 @@ message UnlockResponse {
...
+}
+
+message WorkflowReference {
+  string InstanceID = 1;
+}
+
+message GetWorkflowRequest {
+  WorkflowReference WorkflowReference  = 1;
+  string WorkflowType = 2;
+  string InstanceID = 3;
```
cgillum, line 676: What's the difference between this `InstanceID` and `WorkflowReference.InstanceID`?

dapr/proto/runtime/v1/dapr.proto
```diff
@@ -654,4 +664,34 @@ message UnlockResponse {
...
+  google.protobuf.Timestamp StartTime = 2;
+  map<string, string> Metadata = 3;
+}
+
+message StartWorkflowRequest {
+  WorkflowReference WorkflowReference = 1;
+  string WorkflowComponent = 2;
+  string WorkflowName = 3;
+  map<string, string> Options = 4;
+  google.protobuf.Any Parameters = 5;
```
cgillum, line 691: `Parameters` as a name for this field is a little confusing because using a plural often implies some kind of collection type, which isn't the case here. My recommendation would be to call this `StartParameter` (single) or `Input`, which is the terminology we use for the built-in workflow engine.

tests/apps/workflowsapp/app.go
```diff
@@ -0,0 +1,210 @@
...
+	w.RegisterActivity(HelloDaprAct)
+
+	// Start listening to the Task Queue
+	log.Println("e2e worker created")
+	w.Start()
+
+	utils.StartServer(appPort, appRouter, true, false)
+}
+
+func HelloDaprWF(ctx workflow.Context) (string, error) {
```
cgillum, line 184: `HelloDaprWF` --> `HelloTemporalWF`

tests/apps/workflowsapp/app.go
```diff
@@ -0,0 +1,210 @@
...
+	err := workflow.ExecuteActivity(ctx, HelloDaprAct).Get(ctx, &resp)
+	if err != nil {
+		fmt.Println("ERROR During activity call: ", err.Error())
+	}
+
+	fmt.Println("WF Activity Finished")
+	return resp, nil
+}
+
+func HelloDaprAct(ctx context.Context) (result string, err error) {
```
cgillum, line 208: `HelloDaprAct` --> `HelloTemporalAct`

tests/apps/workflowsapp/app.go
```diff
@@ -0,0 +1,210 @@
...
+		HeartbeatTimeout:       time.Second * 2,
+		WaitForCancellation:    true,
+	}
+
+	ctx = workflow.WithActivityOptions(ctx, options)
+
+	var resp string
+	err := workflow.ExecuteActivity(ctx, HelloDaprAct).Get(ctx, &resp)
+	if err != nil {
+		fmt.Println("ERROR During activity call: ", err.Error())
```
cgillum, line 201: Shouldn't we be returning `err` here?

tests/apps/workflowsapp/service.yaml
```diff
@@ -0,0 +1,46 @@
+# In e2e test, this will not be used to deploy the app to test cluster.
+# This is created for testing purpose in order to deploy this app using kubectl
+# before writing e2e test.
+
+kind: Service
+apiVersion: v1
+metadata:
+  name: workflowsapp
```
cgillum, line 8: Should this be named `temporalapp`?

cgillum (commented):
A few suggestions around naming and error handling. As far as naming is concerned, my main feedback is that we be explicit in referring to workflow components as "components". Otherwise, there will be a ton of confusion and terminology overlap between the API infrastructure and the built-in workflow engine. It's a small break from existing naming conventions, but I think it's justified in this case to remove ambiguity.

pkg/http/api.go
```diff
@@ -714,6 +748,161 @@ func (a *api) getLockStoreWithRequestValidation(reqCtx *fasthttp.RequestCtx) (lo
...
+		WorkflowReference: startReq.WorkflowReference,
+		WorkflowName:      wfType,
+		Input:             startReq.Input,
+	}
+	req.WorkflowReference.InstanceID = instance
+
+	resp, err := workflowRun.Start(reqCtx, &req)
+	if err != nil {
+		msg := NewErrorResponse("ERR_START_WORKFLOW", fmt.Sprintf(messages.ErrStartWorkflow, err))
+		respond(reqCtx, withError(fasthttp.StatusBadRequest, msg))
```
cgillum, line 805: This looks like it should be `fasthttp.StatusInternalServerError`, since we can't know if it was bad input or just some internal issue with the workflow component's `Start` implementation.

pkg/http/api.go
```diff
@@ -714,6 +748,161 @@ func (a *api) getLockStoreWithRequestValidation(reqCtx *fasthttp.RequestCtx) (lo
...
+	}
+
+	req := wfs.WorkflowReference{
+		InstanceID: instance,
+	}
+
+	resp, err := workflowRun.Get(reqCtx, &req)
+	if err != nil {
+		msg := NewErrorResponse("ERR_GET_WORKFLOW", fmt.Sprintf(messages.ErrWorkflowGetResponse, err))
+		respond(reqCtx, withError(fasthttp.StatusBadRequest, msg))
```
cgillum, line 858: Same comment about using `fasthttp.StatusInternalServerError` instead of `fasthttp.StatusBadRequest`.

pkg/http/api.go
```diff
@@ -714,6 +748,161 @@ func (a *api) getLockStoreWithRequestValidation(reqCtx *fasthttp.RequestCtx) (lo
...
+		log.Debug(msg)
+		return
+	}
+	response, err := json.Marshal(resp)
+	if err != nil {
+		log.Errorf("Error when matshalling workflow information: %v\n", err)
+		return
```
cgillum, line 865: We need to return an HTTP error response in this case too, right?

pkg/http/api.go
```diff
@@ -714,6 +748,161 @@ func (a *api) getLockStoreWithRequestValidation(reqCtx *fasthttp.RequestCtx) (lo
...
+	}
+
+	req := wfs.WorkflowReference{
+		InstanceID: instance,
+	}
+
+	err := workflowRun.Terminate(reqCtx, &req)
+	if err != nil {
+		msg := NewErrorResponse("ERR_TERMINATE_WORKFLOW", fmt.Sprintf(messages.ErrTerminateWorkflow, err))
+		respond(reqCtx, withError(fasthttp.StatusBadRequest, msg))
```
cgillum, line 900: Same comment about using `fasthttp.StatusInternalServerError` instead of `fasthttp.StatusBadRequest`.

pkg/http/api.go
```diff
@@ -714,6 +748,161 @@ func (a *api) getLockStoreWithRequestValidation(reqCtx *fasthttp.RequestCtx) (lo
...
+		msg := NewErrorResponse("ERR_NO_WORKFLOW_COMPONENT_PROVIDED", fmt.Sprintf(messages.ErrNoOrMissingComponent))
+		respond(reqCtx, withError(fasthttp.StatusBadRequest, msg))
+		log.Debug(msg)
+		return
+	}
+
+	workflowRun := a.workflowComponents[component]
+	if workflowRun == nil {
+		log.Error("The provided component does not exist.")
+		return
```
cgillum, line 890: Same comment about sending HTTP error responses.

cgillum (commented):
A few small things I noticed in the most recent iteration. I haven't quite gone through all the changes yet, though.

cgillum:
Besides my comments above, the rest of the changes look good to me. However, it looks like there are some build failures that still need to be resolved.

I'm a bit confused about the recent commit history for this PR (there appears to be several duplicates, unrelated things, etc.) but I'll assume that Dapr maintainers will comment on that if it's at all problematic.

RyanLettieri:
> Besides my comments above, the rest of the changes look good to me. However, it looks like there are some build failures that still need to be resolved.
> 
> I'm a bit confused about the recent commit history for this PR (there appears to be several duplicates, unrelated things, etc.) but I'll assume that Dapr maintainers will comment on that if it's at all problematic.

@cgillum the build failures are due to the fact that the building block in components contrib is also getting updated based off some of the comments/changes in this PR. Here's the link to the component-contrib PR: https://github.com/dapr/components-contrib/pull/2269
The large number of commits is because I messed up when merging in changes from master, but they are all resolved now

cgillum (dismissed):
Thanks @RyanLettieri. I've reviewed the latest commit and I'm good with these changes!

dapr/proto/runtime/v1/dapr.proto
```diff
@@ -672,4 +682,33 @@ message UnlockResponse {
...
+}
+
+message GetWorkflowResponse {
+  WorkflowReference WorkflowReference  = 1;
+  google.protobuf.Timestamp StartTime = 2;
+  map<string, string> Metadata = 3;
+}
+
+message StartWorkflowRequest {
+  WorkflowReference WorkflowReference = 1;
```
ItalyPaleAle, line 704: Why using a message here? Do you think you'll need to add more properties in the future? Or can you just pass the workflow ID here?

dapr/proto/runtime/v1/dapr.proto
```diff
@@ -672,4 +682,33 @@ message UnlockResponse {
...
+
+message GetWorkflowRequest {
+  WorkflowReference WorkflowReference  = 1;
+  string WorkflowType = 2;
+  string WorkflowComponent = 3; 
+}
+
+message GetWorkflowResponse {
+  WorkflowReference WorkflowReference  = 1;
+  google.protobuf.Timestamp StartTime = 2;
```
ItalyPaleAle, line 699: I don't believe we use `google.protobuf.Timestamp` anywhere, but it's a struct... Can we use int64 and pass a UNIX timestamp? Unless nanoseconds matter?

pkg/components/workflows/registry.go
```diff
@@ -0,0 +1,80 @@
...
+	for _, name := range names {
+		s.workflowComponents[createFullName(name)] = componentFactory
+	}
+}
+
+func (s *Registry) Create(name, version string) (wfs.Workflow, error) {
+	if method, ok := s.getWorkflowComponent(name, version); ok {
+		return method(), nil
+	}
+	return nil, errors.Errorf("couldn't find wokflow %s/%s", name, version)
```
ItalyPaleAle, line 53: Let's avoid using `github.com/pkg/errors` in new code please. Here, use `fmt.Errorf`

pkg/grpc/api.go
```diff
@@ -1715,6 +1723,135 @@ func (a *api) SetMetadata(ctx context.Context, in *runtimev1pb.SetMetadataReques
 	return &emptypb.Empty{}, nil
 }
 
+func (a *api) GetWorkflowAlpha1(ctx context.Context, in *runtimev1pb.GetWorkflowRequest) (*runtimev1pb.GetWorkflowResponse, error) {
+	if in.WorkflowReference.InstanceID == "" {
+		err := status.Errorf(codes.InvalidArgument, fmt.Sprintf(messages.ErrMissingOrEmptyInstance))
```
ItalyPaleAle, line 1728: Why fmt.Sprintf here if there's no argument passed? Can you just pass `messages.ErrMissingOrEmptyInstance`?

pkg/grpc/api.go
```diff
@@ -1715,6 +1723,135 @@ func (a *api) SetMetadata(ctx context.Context, in *runtimev1pb.SetMetadataReques
...
+		apiServerLogger.Debug(err)
+		return &runtimev1pb.GetWorkflowResponse{}, err
+	}
+	if in.WorkflowComponent == "" {
+		err := status.Errorf(codes.InvalidArgument, fmt.Sprintf(messages.ErrNoOrMissingComponent))
```
ItalyPaleAle, line 1733: Same

pkg/grpc/api.go
```diff
@@ -1715,6 +1723,135 @@ func (a *api) SetMetadata(ctx context.Context, in *runtimev1pb.SetMetadataReques
...
+	}
+	response, err := workflowRun.Get(ctx, &req)
+	if err != nil {
+		err = status.Errorf(codes.Internal, fmt.Sprintf(messages.ErrWorkflowGetResponse, err))
+		apiServerLogger.Debug(err)
+		return &runtimev1pb.GetWorkflowResponse{}, err
+	}
+
+	id := &runtimev1pb.WorkflowReference{
+		InstanceID: response.WFInfo.InstanceID,
```
ItalyPaleAle, line 1754: Can `response` ever be nil? do we need a nil-check here?

pkg/grpc/api.go
```diff
@@ -1715,6 +1723,135 @@ func (a *api) SetMetadata(ctx context.Context, in *runtimev1pb.SetMetadataReques
...
+		WorkflowReference: id,
+		StartTime:         startTime,
+		Metadata:          response.Metadata,
+	}
+	return res, nil
+}
+
+func (a *api) StartWorkflowAlpha1(ctx context.Context, in *runtimev1pb.StartWorkflowRequest) (*runtimev1pb.WorkflowReference, error) {
+	if in.WorkflowName == "" {
+		err := status.Errorf(codes.InvalidArgument, fmt.Sprintf(messages.ErrWorkflowNameMissing))
```
ItalyPaleAle, line 1776: Ditto

pkg/grpc/api.go
```diff
@@ -1715,6 +1723,135 @@ func (a *api) SetMetadata(ctx context.Context, in *runtimev1pb.SetMetadataReques
...
+		apiServerLogger.Debug(err)
+		return &runtimev1pb.WorkflowReference{}, err
+	}
+
+	if in.WorkflowComponent == "" {
+		err := status.Errorf(codes.InvalidArgument, fmt.Sprintf(messages.ErrNoOrMissingComponent))
```
ItalyPaleAle, line 1782: Ditto

pkg/grpc/api.go
```diff
@@ -1715,6 +1723,135 @@ func (a *api) SetMetadata(ctx context.Context, in *runtimev1pb.SetMetadataReques
...
+		apiServerLogger.Debug(err)
+		return &runtimev1pb.WorkflowReference{}, err
+	}
+
+	if in.WorkflowReference.InstanceID == "" {
+		err := status.Errorf(codes.InvalidArgument, fmt.Sprintf(messages.ErrMissingOrEmptyInstance))
```
ItalyPaleAle, line 1788: Ditto

pkg/grpc/api.go
```diff
@@ -1715,6 +1723,135 @@ func (a *api) SetMetadata(ctx context.Context, in *runtimev1pb.SetMetadataReques
...
+	}
+
+	resp, err := workflowRun.Start(ctx, &req)
+	if err != nil {
+		err = status.Errorf(codes.Internal, fmt.Sprintf(messages.ErrStartWorkflow, err))
+		apiServerLogger.Debug(err)
+		return &runtimev1pb.WorkflowReference{}, err
+	}
+	ret := &runtimev1pb.WorkflowReference{
+		InstanceID: resp.InstanceID,
```
ItalyPaleAle, line 1817: Ditto: do we need a nil-check for `resp`?

pkg/grpc/api.go
```diff
@@ -1715,6 +1723,135 @@ func (a *api) SetMetadata(ctx context.Context, in *runtimev1pb.SetMetadataReques
...
+	}
+	return ret, nil
+}
+
+func (a *api) TerminateWorkflowAlpha1(ctx context.Context, in *runtimev1pb.TerminateWorkflowRequest) (*emptypb.Empty, error) {
+	if in.WorkflowReference.InstanceID == "" {
+		err := status.Errorf(codes.InvalidArgument, fmt.Sprintf(messages.ErrMissingOrEmptyInstance))
```
ItalyPaleAle, line 1824: Ditto

pkg/grpc/api.go
```diff
@@ -1715,6 +1723,135 @@ func (a *api) SetMetadata(ctx context.Context, in *runtimev1pb.SetMetadataReques
...
+		apiServerLogger.Debug(err)
+		return &emptypb.Empty{}, err
+	}
+
+	if in.WorkflowComponent == "" {
+		err := status.Errorf(codes.InvalidArgument, fmt.Sprintf(messages.ErrNoOrMissingComponent))
```
ItalyPaleAle, line 1830: Ditto

pkg/grpc/api.go
```diff
@@ -1715,6 +1723,135 @@ func (a *api) SetMetadata(ctx context.Context, in *runtimev1pb.SetMetadataReques
...
+		apiServerLogger.Debug(err)
+		return &emptypb.Empty{}, err
+	}
+
+	workflowRun := a.workflowComponents[in.WorkflowComponent]
+	if workflowRun == nil {
+		err := status.Errorf(codes.InvalidArgument, fmt.Sprintf(messages.ErrComponentDoesNotExist, in.WorkflowComponent))
```
ItalyPaleAle, line 1837: Ditto

pkg/http/api.go
```diff
@@ -731,6 +765,167 @@ func (a *api) getLockStoreWithRequestValidation(reqCtx *fasthttp.RequestCtx) (lo
...
+// Route:   "workflows/{workflowComponent}/{workflowType}/{instanceId}",
+// Workflow Component: Component specified in yaml (temporal, etc..)
+// Workflow Type: Name of the workflow to run (function name)
+// Instance ID: Identifier of the specific run
+func (a *api) onStartWorkflow(reqCtx *fasthttp.RequestCtx) {
+	startReq := wfs.StartRequest{}
+
+	wfType := reqCtx.UserValue(workflowType).(string)
+	if wfType == "" {
+		msg := NewErrorResponse("ERR_NO_WORKFLOW_TYPE_PROVIDED", fmt.Sprintf(messages.ErrWorkflowNameMissing))
```
ItalyPaleAle, line 777: Ditto

pkg/http/api.go
```diff
@@ -731,6 +765,167 @@ func (a *api) getLockStoreWithRequestValidation(reqCtx *fasthttp.RequestCtx) (lo
...
+		respond(reqCtx, withError(fasthttp.StatusBadRequest, msg))
+		log.Debug(msg)
+		return
+	}
+
+	component := reqCtx.UserValue(workflowComponent).(string)
+	if component == "" {
+		msg := NewErrorResponse("ERR_NO_WORKFLOW_COMPONENT_PROVIDED", fmt.Sprintf(messages.ErrNoOrMissingComponent))
```
ItalyPaleAle, line 785: Ditto

pkg/http/api.go
```diff
@@ -731,6 +765,167 @@ func (a *api) getLockStoreWithRequestValidation(reqCtx *fasthttp.RequestCtx) (lo
...
+	if workflowRun == nil {
+		msg := NewErrorResponse("ERR_NON_EXISTENT_WORKFLOW_COMPONENT_PROVIDED", fmt.Sprintf(messages.ErrComponentDoesNotExist, component))
+		respond(reqCtx, withError(fasthttp.StatusBadRequest, msg))
+		log.Debug(msg)
+		return
+	}
+
+	err := json.Unmarshal(reqCtx.PostBody(), &startReq)
+	if err != nil {
+		log.Errorf("Error while unmarshalling workflow info: %v\n", err)
```
ItalyPaleAle, line 800: Just logging the error? No returning?

pkg/http/api.go
```diff
@@ -731,6 +765,167 @@ func (a *api) getLockStoreWithRequestValidation(reqCtx *fasthttp.RequestCtx) (lo
...
+	}
+	response, err := json.Marshal(resp)
+	if err != nil {
+		msg := NewErrorResponse("ERR_METADATA_GET", fmt.Sprintf(messages.ErrMetadataGet, err))
+		respond(reqCtx, withError(fasthttp.StatusInternalServerError, msg))
+		log.Debug(msg)
+		return
+	}
+	log.Debug(resp)
+	respond(reqCtx, withJSON(202, response))
```
ItalyPaleAle, line 834: Use a constant like `fasthttp.StatusAccepted` rather than 202

pkg/http/api.go
```diff
@@ -731,6 +765,167 @@ func (a *api) getLockStoreWithRequestValidation(reqCtx *fasthttp.RequestCtx) (lo
...
+}
+
+func (a *api) onGetWorkflow(reqCtx *fasthttp.RequestCtx) {
```
ItalyPaleAle, line 837: For this method, my feedback is the same as in the previous method

pkg/http/api.go
```diff
@@ -731,6 +765,167 @@ func (a *api) getLockStoreWithRequestValidation(reqCtx *fasthttp.RequestCtx) (lo
...
+	if err != nil {
+		msg := NewErrorResponse("ERR_METADATA_GET", fmt.Sprintf(messages.ErrMetadataGet, err))
+		respond(reqCtx, withError(fasthttp.StatusInternalServerError, msg))
+		log.Debug(msg)
+		return
+	}
+	respond(reqCtx, withJSON(200, response))
+}
+
+func (a *api) onTerminateWorkflow(reqCtx *fasthttp.RequestCtx) {
```
ItalyPaleAle, line 891: Same in this method

pkg/messages/api_errors.go
```diff
@@ -88,4 +88,15 @@ const (
 	ErrLockOwnerEmpty             = "LockOwner is empty in lock store %s"
 	ErrExpiryInSecondsNotPositive = "ExpiryInSeconds is not positive in lock store %s"
 	ErrLockStoreNotFound          = "lock store %s not found"
+
+	// Workflow
+	ErrWorkflowNameMissing    = "workflow name is not configured"
+	ErrMissingOrEmptyInstance = "no instance or empty instance was provided"
```
ItalyPaleAle, line 94: For constants in this block, they should all have `Workflow` in the name. For example, this one should be `ErrMissingOrEmptyWorkflowInstance` or something like that

tests/apps/workflowsapp/app.go
```diff
@@ -0,0 +1,205 @@
...
+	// Parse the data from the Get call, specifically status and return it. It should be complete
+	body, _ = io.ReadAll(res.Body)
+	var stateData workflows.StateResponse
+	json.Unmarshal(body, &stateData)
+	return http.StatusOK, appResponse{Message: stateData.Metadata["status"]}
+}
+
+// epoch returns the current unix epoch timestamp
+func epoch() int {
+	return int(time.Now().UnixMilli())
```
ItalyPaleAle, line 129: Use int64 or you'll have big trouble on 32-bit machines here

tests/apps/workflowsapp/app.go
```diff
@@ -0,0 +1,205 @@
...
+	log.Printf("StartTest - message: %s", commandRequest.Message)
+
+	// Create json payload to send over HTTP to start workflow by providing task_queue
+	jsonData := []byte(`{
+		"workflow_options" :
+		{
+			"task_queue" : "e2e_test_queue"
+		}
+	}`)
+	res, err := httpClient.Post("http://localhost:3500/v1.0-alpha1/workflows/temporal/HelloTemporalWF/WorkflowID/start", "application/json", bytes.NewBuffer(jsonData))
```
ItalyPaleAle, line 101: Note that dapr's port can change. Please use the DAPR_HTTP_PORT env var

tests/apps/workflowsapp/app.go
```diff
@@ -0,0 +1,205 @@
...
+	defer res.Body.Close()
+
+	// Get the data response from the start workflow call
+	body, _ := io.ReadAll(res.Body)
+	var resultData workflows.WorkflowReference
+	json.Unmarshal(body, &resultData)
+	time.Sleep(2 * time.Second)
+
+	// USe the data that was retrieved back from the start workflow call (InstanceID) to get info on the workflow
+	res, err = httpClient.Get("http://localhost:3500/v1.0-alpha1/workflows/temporal/HelloTemporalWF/" + resultData.InstanceID + "")
```
ItalyPaleAle, line 114: Ditto

tests/apps/workflowsapp/app.go
```diff
@@ -0,0 +1,205 @@
...
+	if err != nil {
+		return http.StatusInternalServerError, appResponse{Message: err.Error()}
+	}
+	defer res.Body.Close()
+
+	// Get the data response from the start workflow call
+	body, _ := io.ReadAll(res.Body)
+	var resultData workflows.WorkflowReference
+	json.Unmarshal(body, &resultData)
+	time.Sleep(2 * time.Second)
```
ItalyPaleAle, line 111: Why the sleep call?

tests/apps/workflowsapp/app.go
```diff
@@ -0,0 +1,205 @@
...
+
+	router.Use(mux.CORSMethodMiddleware(router))
+
+	return router
+}
+
+func main() {
+	log.Printf("Workflow Test - listening on http://localhost:%d", appPort)
+
+	client2, _ := client.NewNamespaceClient(client.Options{HostPort: "dapr-temporal-frontend.dapr-tests.svc.cluster.local:7233"})
```
ItalyPaleAle, line 150: I think the address should be configurable via env var. The test will pass it as env vars, such as "SERVICE_NAME" and "NAMESPACE" and "PORT"... Don't just hardocde it!

tests/apps/workflowsapp/app.go
```diff
@@ -0,0 +1,205 @@
+/*
```
ItalyPaleAle, line 1: Note that this app needs a .cache_include file (see other apps as examples)

tests/apps/workflowsapp/app.go
```diff
@@ -0,0 +1,205 @@
...
+	startTime := epoch()
+	statusCode, res := startTest(commandBody)
+	res.StartTime = startTime
+	res.EndTime = epoch()
+
+	w.WriteHeader(statusCode)
+	json.NewEncoder(w).Encode(res)
+}
+
+func startTest(commandRequest testCommandRequest) (int, appResponse) {
```
ItalyPaleAle, line 91: Two things on this:

1. Should also test the "workflow terminate" method
2. Tests should work with both HTTP and gRPC (see other apps how they do it)

tests/dapr_tests.mk
```diff
@@ -369,6 +370,8 @@ setup-helm-init:
 	$(HELM) repo add bitnami https://charts.bitnami.com/bitnami
 	$(HELM) repo add stable https://charts.helm.sh/stable
 	$(HELM) repo add incubator https://charts.helm.sh/incubator
+	$(HELM) repo add armory https://armory.jfrog.io/artifactory/charts/
```
ItalyPaleAle, line 373: Is this one still used? I see werner referenced below, but not this one

ItalyPaleAle (changes requested):

pkg/grpc/api.go
RyanLettieri: No, the only time the response will be nil is when the error is not nil, and that is covered by the check above.

pkg/grpc/api.go
RyanLettieri: No, the only time that the resp is nil is when the error is not nil.

pkg/grpc/api.go
RyanLettieri: SprintF is needed here since we are passing in the in.WorkflowComponent as a parameter to the message error

tests/apps/workflowsapp/app.go
RyanLettieri: The sleep call is to ensure that the activity has been finished prior to executing the "Get" call

dapr/proto/runtime/v1/dapr.proto
RyanLettieri: I do anticipate that we might add more properties in the future

ItalyPaleAle: More properties under "WorkflowReference"? Or just in the "StartWorkflowRequest"?

ItalyPaleAle (approved):

ItalyPaleAle:
@artursouza 

> @ItalyPaleAle would you be OK if @RyanLettieri follows up with gRPC tests in a separate PR? I would like to unblock SDK work with this being merged.

I’m ok with that. @RyanLettieri  can you please create an issue however so we don’t forget about it?

However, there are still 2 comments I hope @RyanLettieri  can address in the proto files:

1. Using the Timestamp, which is a message, rather than just passing an int64 with a UNIX timestamp (the Timestamp message has the difference that it allows passing nanoseconds, and don’t think that’s needed?)
2. IDs as a struct rather than just a string

Less structs, easier it is to invoke them :)

RyanLettieri:
> @artursouza
> 
> > @ItalyPaleAle would you be OK if @RyanLettieri follows up with gRPC tests in a separate PR? I would like to unblock SDK work with this being merged.
> 
> I’m ok with that. @RyanLettieri can you please create an issue however so we don’t forget about it?
> 
> However, there are still 2 comments I hope @RyanLettieri can address in the proto files:
> 
> 1. Using the Timestamp, which is a message, rather than just passing an int64 with a UNIX timestamp (the Timestamp message has the difference that it allows passing nanoseconds, and don’t think that’s needed?)
> 2. IDs as a struct rather than just a string
> 
> Less structs, easier it is to invoke them :)

Here's the gRPC issue: https://github.com/dapr/dapr/issues/5581

ItalyPaleAle (dismissed):

yaron2:
@RyanLettieri notice tests are failing.

RyanLettieri:
> @RyanLettieri notice tests are failing.

@yaron2 Third time's a charm! It turns to be a flakey test

artursouza (approved):

Issue #5136, opened 2022-09-06
Sidecar logs error on RETRY pubsub result from App


artursouza:
## In what area(s)?

> /area runtime

> /area operator

> /area placement

> /area docs

> /area test-and-release

## What version of Dapr?

1.8.x

## Expected Behavior

Should log WARN instead.

## Actual Behavior

Logs ERROR because it returns an error to the upstream call stack:

https://github.com/dapr/dapr/blob/490ef74dbd06e13a4557a249afd58d1abecb01af/pkg/runtime/runtime.go#L1937
https://github.com/dapr/dapr/blob/490ef74dbd06e13a4557a249afd58d1abecb01af/pkg/runtime/runtime.go#L1945
https://github.com/dapr/dapr/blob/490ef74dbd06e13a4557a249afd58d1abecb01af/pkg/runtime/runtime.go#L1960

Code should handle retriable vs non-retriable error as different return objects

## Release Note

RELEASE NOTE:

yaron2:
Maybe this should be `Debug` logs?

anoymouscoder:
@artursouza I want to make sure I understand correctly that for retriable errors we don't need to log Error, but just log Warn or log Debug, while for non-retriable errors we need to log Error. 
The log behavior is done by upstream call, so we need different types of error to distinguish retriable from non-retriable error, is that right?

artursouza:
> Maybe this should be `Debug` logs?

I think it should log warning since users have asked for this behavior before.

artursouza:
> @artursouza I want to make sure I understand correctly that for retriable errors we don't need to log Error, but just log Warn or log Debug, while for non-retriable errors we need to log Error. 
> 
> The log behavior is done by upstream call, so we need different types of error to distinguish retriable from non-retriable error, is that right?

Correct. It probably means that the function should return two error objects, one for retriable error and one for non-retriable.

anoymouscoder:
@artursouza The act of catching these errors and logging them happens in the pubsub component(e.g. rabbitmq) of the components-contrib repo, right? 
So if I want to change the log level I also need to change the components-contrib repository?

artursouza:
> @artursouza The act of catching these errors and logging them happens in the pubsub component(e.g. rabbitmq) of the components-contrib repo, right? So if I want to change the log level I also need to change the components-contrib repository?

[...]

PR #5159, opened 2022-09-13
Distinguish between retriable and non-retriable error When publish a …

SpikeWong:
# Description

Distinguish between retrialable and non-retriable errors when publishing messages with pubsub

## Issue reference

Please reference the issue this PR will close: #5136 

## Checklist

* [ ] Code compiles correctly
* [x] Created/updated tests
* [x] Unit tests passing


pkg/runtime/runtime.go
```diff
@@ -1976,7 +1987,7 @@ func (a *DaprRuntime) publishMessageHTTP(ctx context.Context, msg *pubsubSubscri
 	// Every error from now on is a retriable error.
 	log.Warnf("retriable error returned from app while processing pub/sub event %v, topic: %v, body: %s. status code returned: %v", cloudEvent[pubsub.IDField], cloudEvent[pubsub.TopicField], body, statusCode)
 	diag.DefaultComponentMonitoring.PubsubIngressEvent(ctx, msg.pubsub, strings.ToLower(string(pubsub.Retry)), msg.topic, elapsed)
-	return errors.Errorf("retriable error returned from app while processing pub/sub event %v, topic: %v, body: %s. status code returned: %v", cloudEvent[pubsub.IDField], cloudEvent[pubsub.TopicField], body, statusCode)
+	return fmt.Errorf("retriable error returned from app while processing pub/sub event %v, topic: %v, body: %s. status code returned: %v, err: %w", cloudEvent[pubsub.IDField], cloudEvent[pubsub.TopicField], body, statusCode, ErrRetriable)
 }
 
 func (a *DaprRuntime) publishMessageGRPC(ctx context.Context, msg *pubsubSubscribedMessage) error {
```
anoymouscoder, line 1993: Not sure if this method needs to be modified at the same time

pkg/runtime/runtime.go
```diff
@@ -807,6 +812,12 @@ func (a *DaprRuntime) subscribeTopic(parentCtx context.Context, name string, top
 			}
 		})
 		if err != nil && err != context.Canceled {
+			if errors.Is(err, ErrRetriable) {
```
artursouza, line 815: This code seems to be executed after the retries were exhausted. The change should take place inside the policy callback function (lines 805 to 812 in this version.

artursouza (changes requested):

pkg/runtime/runtime.go
anoymouscoder: @artursouza  Do we also need to distinguish between retriable and non-retriable errors in the `publishMessageGRPC` func?

anoymouscoder:
Do I need to change this error, it also looks like a retriable error
https://github.com/dapr/dapr/blob/a85b29fbd745e4e963594133b578f1ce685110cb/pkg/runtime/runtime.go#L1934

artursouza:
> Do I need to change this error, it also looks like a retriable error
> 
> https://github.com/dapr/dapr/blob/a85b29fbd745e4e963594133b578f1ce685110cb/pkg/runtime/runtime.go#L1934

Yes, it seems to be retriable.

pkg/runtime/runtime.go
```diff
@@ -1928,7 +1939,7 @@ func (a *DaprRuntime) publishMessageHTTP(ctx context.Context, msg *pubsubSubscri
 
 	if err != nil {
 		diag.DefaultComponentMonitoring.PubsubIngressEvent(ctx, msg.pubsub, strings.ToLower(string(pubsub.Retry)), msg.topic, elapsed)
-		return errors.Wrap(err, "error from app channel while sending pub/sub event to app")
+		return fmt.Errorf("%w from app channel while sending pub/sub event to app: %v", ErrRetriable, err)
```
spikecdc, line 1942: Not sure if I should handle it this way, do I need to use multierror to wrap multiple errors at the same time?

anoymouscoder, line 1942: Not sure if I should handle it this way, do I need to use multierror to wrap multiple errors at the same time?

artursouza: I am not aware of multierror being used anywhere else in the code. What would be the pros and cons?

anoymouscoder: I am referring to this [library](https://github.com/uber-go/multierr), which is currently available in the [component-contrib](https://github.com/dapr/components-contrib/search?q=multierr).
The reason I mention this is that when I use `fmt.Errorf` to wrap the lower level error as `ErrRetriable`, I can no longer do the `errors.Is` and `errors.As` check of the original lower level error for the new error. I don't know if we have such a need, but if we use multierr, we can satisfy it.

SpikeWong:
> @SpikeWong please resolve the conflicts.

Finished

SpikeWong:
A little bit curious why there are some e2e tests failed,  because they are unstable?

pkg/runtime/runtime.go
ItalyPaleAle: Perhaps consider creating a custom Error class instead: https://www.digitalocean.com/community/tutorials/creating-custom-errors-in-go

SpikeWong: Do you mean using a custom error struct to wrap original error?

ItalyPaleAle: Yes :)

SpikeWong: Ok, I can implement a custom error struct which has `error` field, so it can wrap original error. This custom error struct also provide `Unwrap` func, which allow user to assert the original error's type, do you think is it ok?

ItalyPaleAle: Perfect :)

SpikeWong: Finished, please take a look :)

pkg/runtime/retriableError.go
```diff
@@ -0,0 +1,41 @@
+/*
```
SpikeWong, line 1: I am not sure about the file naming format,  there is file named `initError.go` under the same pkg , but some file located in other pkg use `_` to separate words.

pkg/runtime/retriableError.go
```diff
@@ -0,0 +1,41 @@
...
+limitations under the License.
+*/
+
+package runtime
+
+import (
+	"fmt"
+)
+
+type RetriableError struct {
```
SpikeWong, line 20: I'm not sure if `retriable` is typo, that's how it was written before and I've followed the previous way of writing it

SpikeWong: Maybe it should be retryable?

yaron2:
@SpikeWong please fix conflicts, we'll try to get this merged right after 1.10 is out the door.

pkg/runtime/retriableError.go
ItalyPaleAle: should be retriable_error. Feel free to rename initError to init_error too

pkg/runtime/retriableError.go
ItalyPaleAle: https://english.stackexchange.com/questions/305273/retriable-or-retryable

Retriable works :)

pkg/runtime/runtime.go
```diff
@@ -862,6 +862,12 @@ func (a *DaprRuntime) subscribeTopic(parentCtx context.Context, name string, top
 			default:
 				pErr = backoff.Permanent(errors.New("invalid application protocol"))
 			}
+			var rErr *RetriableError
+			if errors.As(pErr, &rErr) {
+				log.Warnf("encountered a retriable error while publishing a subscribed message to topic %s, err: %s", msgTopic, rErr.Unwrap())
```
ItalyPaleAle, line 867:
```suggestion
				log.Warnf("encountered a retriable error while publishing a subscribed message to topic %s, err: %v", msgTopic, rErr.Unwrap())
```

pkg/runtime/runtime.go
```diff
@@ -862,6 +862,12 @@ func (a *DaprRuntime) subscribeTopic(parentCtx context.Context, name string, top
...
+			} else {
+				log.Errorf("encountered a non-retriable error while publishing a subscribed message to topic %s, err: %s", msgTopic, pErr)
```
ItalyPaleAle, line 869:
```suggestion
				log.Errorf("encountered a non-retriable error while publishing a subscribed message to topic %s, err: %v", msgTopic, pErr)
```

pkg/runtime/runtime.go
```diff
@@ -862,6 +862,12 @@ func (a *DaprRuntime) subscribeTopic(parentCtx context.Context, name string, top
 			default:
 				pErr = backoff.Permanent(errors.New("invalid application protocol"))
 			}
+			var rErr *RetriableError
+			if errors.As(pErr, &rErr) {
+				log.Warnf("encountered a retriable error while publishing a subscribed message to topic %s, err: %s", msgTopic, rErr.Unwrap())
+			} else {
```
ItalyPaleAle, line 868: Shouldn't this check `else if pErr != nil`?

artursouza (approved):

ItalyPaleAle (dismissed):

yaron2 (approved):

pkg/runtime/runtime.go
```diff
@@ -900,6 +900,12 @@ func (a *DaprRuntime) subscribeTopic(parentCtx context.Context, name string, top
 			default:
 				pErr = backoff.Permanent(errors.New("invalid application protocol"))
 			}
+			var rErr *RetriableError
+			if errors.As(pErr, &rErr) {
+				log.Warnf("encountered a retriable error while publishing a subscribed message to topic %s, err: %v", msgTopic, rErr.Unwrap())
+			} else if pErr != nil {
+				log.Errorf("encountered a non-retriable error while publishing a subscribed message to topic %s, err: %v", msgTopic, pErr)
```
mukundansundar, line 907: isn't this publshing the message to app and not topic?

mukundansundar:
similar changes if possible needs to be added to the bulk subscribe API as well

mukundansundar:
milestone tracked by issue closed by this PR

PR #5170, opened 2022-09-14
Streaming support for service invocation (Core)

ItalyPaleAle:
This PR is the "core" part #4903 which implements support for streaming sidecar-to-sidecar communication.

Unlike what #4903 attempted to do, this has a slightly narrower scope. In particular, it implements streaming support in:

- Communications between Dapr sidecars (via the gRPC local channel)
- HTTP app channel (from Dapr to apps)
- Responses from direct messages and actor invocations over HTTP APIs

It does **not** add full streaming support to service invocation, as a few things are missing:

- Nothing is changing on gRPC APIs for now, which lack streaming support
- HTTP API server does not read request bodies as streams, due to buggy support in fasthttp. This will require removing fasthttp per https://github.com/dapr/dapr/issues/4979

At a high level, this PR includes:

- Changed the way sidecars invoke one another to use streams. The new method ServiceInvocation.CallLocalStream is a variant of ServiceInvocation.CallLocal that uses streams rather than making unary calls. Internally, this uses a buffer of up to 4KB, so messages that are smaller than that (and that are received in a single chunk by the runtime) are transferred in a single block. This should allow having no performance impact for small payloads (compared to unary RPCs).  
(If the callee app doesn't support CallLocalStream (e.g. it's on an older version of Dapr), this gracefully falls back to using CallLocal. Because we don't officially support clusters running mixed Dapr version, however, this is only meant to ensure smooth upgrades, and a warning is shown.)
- When the sidecar invokes the apps using the HTTP channel, the data is now sent to the app as a stream. Apps can begin receiving data as soon as the runtime gets it.
- The HTTP API server responds to requests for service and actor invocation by sending the data as a stream. However, due to bugs in fasthttp (see #4979), the HTTP API server still reads all incoming requests' bodies without using streams.

The most important contribution from this PR is the addition of a new gRPC data structure `common.v1.StreamPayload` which underpins streaming capabilities for sidecar-to-sidecar communication and for other, future uses of streaming in Dapr. See the "Implementation details" section below for more info.

## Related issues

https://github.com/dapr/dapr/issues/3103 (partially fixed by this PR)
https://github.com/dapr/dapr/issues/4866

## Implementation details

This PR creates a new gRPC method that is used for sidecar-to-sidecar communication.

Currently, sidecar-to-sidecar communication uses the [`CallLocal` RPC](https://github.com/dapr/dapr/blob/2addb5754398b84bde7aa3891a5bb9a11395b413/dapr/proto/internals/v1/service_invocation.proto#L39):

```proto3
// Invokes a method of the specific service.
rpc CallLocal (InternalInvokeRequest) returns (InternalInvokeResponse) {}
```

This RPC is not removed and is retained for backwards compatibility (e.g. when Dapr 1.8 invokes Dapr vNext), however it is not used otherwise.

A new RPC is added:

```proto3
// Invokes a method of the specific service using a stream of data.
rpc CallLocalStream (stream InternalInvokeRequestStream) returns (stream InternalInvokeResponseStream) {}
```

The request and response objects are defined as:

```proto3
// InternalInvokeRequestStream is a variant of InternalInvokeRequest used in streaming RPCs.
message InternalInvokeRequestStream {
  // Request details.
  // This does not contain any data in message.data.
  InternalInvokeRequest request = 1;

  // Chunk of data.
  StreamPayload payload = 2;
}

// InternalInvokeResponseStream is a variant of InternalInvokeResponse used in streaming RPCs.
message InternalInvokeResponseStream {
  // Response details.
  // This does not contain any data in message.data.
  InternalInvokeResponse response = 1;

  // Chunk of data.
  StreamPayload payload = 2;
}

// Chunk of data sent in a streaming request or response.
message StreamPayload {
  // Data sent in the chunk.
  google.protobuf.Any data = 1;

  // Set to true if this is the last chunk.
  bool complete = 2;
}
```

In the protos above, `InternalInvokeRequest` and `InternalInvokeResponse` are the existing protos, that are used by the unary `CallLocal` RPC. In the streamed version, they are wrapped in a proto together with a `StreamPayload` object.

> Note that, although the RPC uses bi-directional streaming, it still **behaves like a unary call**, in the sense that once the connection has been established, first the sender sends the entire request, and then it reads the entire response from the receiver. We use streams to be able to send data in chunks, but the sidecars do not maintain a persistent stream that is used to exchange multiple messages or where communication is constantly bi-directional.

How it works:

1. The caller sidecar invokes `CallLocalStream` on the target sidecar and establishes the gRPC connection and the streaming RPC.
2. The caller sidecar sends the data in chunks (see below for details on how chunking works).
3. The target sidecar can begin processing the message (i.e. send it to the target app) as soon as the first chunk is received.
4. Once the full payload has been sent, the target sidecar waits for the response from the app (which is read as a stream).
5. The target sidecar sends the response in chunks to the caller sidecar, with the first chunk sent as soon as it's available.

Each chunk is a message of type `InternalInvokeRequestStream` or `InternalInvokeResponseStream` sent over the RPC, where:

- The first message in the stream MUST contain an `InternalInvokeRequest` or `InternalInvokeResponse` with the required keys present.
- The first message in the stream MAY contain a `payload`, but that is not required.
- Subsequent messages (any message except the first in the stream) MUST contain a `payload` and MUST NOT contain any other property.
- The last message in the stream MUST contain a `payload` with `complete` set to `true`. That message is assumed to be the last one from the sender and no more messages are to be sent in the stream after that.

The amount of data contained in `payload.data` is variable and it's up to the discretion of the sender. In this PR, chunks are at most 4KB in size, although senders may send smaller chunks if they wish. Receivers must not assume that messages will contain any specific number of bytes in the payload.

Note that it's possible for senders to send a single message in the stream. If the data is small and could fit in a single chunk, senders MAY choose to include a `payload` with `complete=true` in the first message. In this case, the entire message contains both an `InternalInvokeRequest`/`InternalInvokeResponse` with the required keys present AND a `payload` with data (if any) and `complete=true`. Receivers should assume that single message to be the entire communication from the sender.


artursouza:
Let's merge this right AFTER code freeze for 1.9, please.

ItalyPaleAle:
Tests are passing but there's an issue with the pipelines (#5174) that causes them to be reported as failing.

tony-clarke-amdocs:
@ItalyPaleAle Would you expect that this change might reduce latency for large payloads? If so, any benchmarks? Thanks.

ItalyPaleAle:
@tony-clarke-amdocs 

> Would you expect that this change might reduce latency for large payloads?

Yes, this will reduce latency for large payloads. The larger the payload, the better the latency reduction. You can see a couple of GIFs of before/after in the original PR: https://github.com/dapr/dapr/pull/4903#issue-1304148620 (the GIFs were purposely slowed down to show the impact better).

The impact will be bigger on the TTFB (Time To First Byte), but it should also reduce the TTLB (Time To Last Byte) because of the less buffering.

> If so, any benchmarks? 

Not yet :)

pkg/grpc/api_local.go
```diff
@@ -0,0 +1,232 @@
...
+				pw.CloseWithError(readErr)
+				return
+			}
+
+			if done {
+				break
+			}
+
+			readErr = stream.RecvMsg(chunk)
+			if err != nil {
```
tony-clarke-amdocs, line 109: Should this not reference readErr?

ItalyPaleAle: Yes it should be. Thanks for catching it

ItalyPaleAle:
After talking to @artursouza  today, we're going to change strategy with this PR, which has been stale for too long.

At a high level, adding support for streaming with HTTP requests/responses requires changing 4 things in Dapr:

1. Change the API server to process requests in a streaming way when possible
2. Change all internals of the runtime to work with streams. Biggest part is the pkg/messaging/v1 package which in the current runtime is dealing with byte slices only (classes are InvokeMethodRequest and InvokeMethodResponse). Changes in those classes of course have the consequence of requiring widespread changes throughout the runtime.
3. Change the sidecar-to-sidecar communication to send data in chunks using a stream-based gRPC: from CallLocal to CallLocalStream
4. Change the HTTP channel to send data to the client in a streaming way.

Current status is:

- Work on part 1 this has not started as it's pending on #4979
- Parts 2 and 3 are included in this PR
- Part 4 has been merged in master with #5330

This PR, trying to include both 2 and 3 (and  part 4, before it was split out in #5330), is pretty massive and hard to review. Also, every so often changes in the runtime cause major merge conflicts, and after those, lots of things break. In fact, things are broken right now after 5330 was merged, and I've already spent 7 hours trying to get the tests to pass, unsuccessfully. Part of the issues are caused by fasthttp, which is fairly buggy (e.g. `make test-race` in this PR right now fails due to a race condition in fasthttp itself).

The new plan is:

1. Split out part 3 (change the sidecar-to-sidecar communication to use CallLocalStream) into a separate PR for 1.10, ideally behind a feature flag. I will do it in the near future.
2. Work on 1, fixing #4979 (I could use help with this :) )
3. When fasthttp is out and we hopefully have a more stable implementation with net/http, complete part 2, probably taking from this PR.

Although the full streaming support as shown in the GIF in #4903 will require all 4 parts to be done, every step should help improving the experience for end-users gradually!

ItalyPaleAle:
👋🤖

ItalyPaleAle:
Closing as this is now superseeded.

- #5648 implemented the changes in `pkg/messaging/v1`, with code that was spun off this PR (part 2 from the list above)
- The only other thing left in this PR that hasn't been merged yet is the implementation of CallLocalStream (part 3 from the list above). That will happen in a separate PR.

Still waiting on #4979 to complete the feature E2E.

PR #5171, opened 2022-09-14
Dapr bot provides more updates during e2e tests

ItalyPaleAle:
# Description

As requested by @artursouza, this updates the "/ok-to-test" flow to report more updates.

Fixes these errors:

1. Bot did not post anything if the build failed or the Azure resources failed to deploy
2. Bot did not put a link to the running Action to better track progress

It changes the behavior of "/ok-to-test" by posting a message as soon as the workflow starts, with a link to the workflow run.
image

The message is then updated (not reposted, to avoid spam!) with the progress once the Azure deployment completes, the build succeeds, and tests complete. It also provides useful information for debugging, such as the name of the Resource Group which correlates with the logs in Azure Monitor.

image
image
image

(test failures here are due to the environment inside my fork)

### DIFF
--- a/.github/workflows/dapr-test.yml
+++ b/.github/workflows/dapr-test.yml
@@ -352,7 +352,7 @@ jobs:
           append: true
           GITHUB_TOKEN: ${{ secrets.DAPR_BOT_TOKEN }}
           message: |
-            ## ✅ Build succeeded for ${{ runner.os }}
+            ## ✅ Build succeeded for ${{ matrix.target_os }}/${{ matrix.target_arch }}
 
             - Image tag: `${{ env.DAPR_TAG }}`
             - Test image tag: `${{ env.DAPR_TEST_TAG }}`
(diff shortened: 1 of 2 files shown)


.github/workflows/dapr-test.yml
```diff
@@ -76,8 +78,23 @@ jobs:
...
+        if: env.PR_NUMBER != ''
+        uses: marocchino/sticky-pull-request-comment@v2.2.0
+        with:
+          header: ${{ github.run_id }}
+          number: ${{ env.PR_NUMBER }}
+          hide: true
+          hide_classify: OUTDATED
+          GITHUB_TOKEN: ${{ secrets.DAPR_BOT_TOKEN }}
+          message: |
+            # Dapr E2E test
```
artursouza, line 95: Please, add the HEAD commit hash so we know which version of the code is running.

Issue #5073, opened 2022-08-23
gRPC Service Invocation does not emit metrics


yaron2:
When using gRPC for service invocation, call metrics are not being emitted. This is because Dapr uses gRPC streaming for `daprd` to `daprd` communications and our interceptor is of a Unary interceptor type.

PR #5176, opened 2022-09-15
Add missing metrics for gRPC proxying

yaron2:
This PR fixes https://github.com/dapr/dapr/issues/5073 and adds industry standard golden metrics for proxying calls for gRPC.


pkg/grpc/server.go
```diff
@@ -213,15 +213,18 @@ func (s *server) getMiddlewareOptions() []grpcGo.ServerOption {
 	if diagUtils.IsTracingEnabled(s.tracingSpec.SamplingRate) {
 		s.logger.Info("enabled gRPC tracing middleware")
 		intr = append(intr, diag.GRPCTraceUnaryServerInterceptor(s.config.AppID, s.tracingSpec))
-
-		if s.proxy != nil {
```
mukundansundar, line 217: Quick question why is this check removed? @yaron2

yaron2: This is from the time when the proxy was a preview feature and could have been nil. There is no way proxy can be nil now.

mukundansundar (approved):
Lgtm

Issue #5194, opened 2022-09-17
proposal: Protos for Bindings Pluggable Components


mcandeia:
## In what area(s)?

/area runtime

> /area operator

> /area placement

> /area docs

> /area test-and-release
## Describe the proposal

> If you don't have context about pluggable components, start from here https://github.com/dapr/dapr/issues/4925

As a preparation for 1.9 arrival, we should agree on proto definitions for pluggable components. This proposal is focused on bindings protos.

The experience with the [pubsub](https://github.com/dapr/dapr/issues/5173) proto made this one easier to implement. So the protos follows the same interface for components-contrib receive and send bunch of data with its associated payload and metadata. With bi-di stream used for Inputbindings.

```proto
syntax = "proto3";

package dapr.proto.components.v1;

import "dapr/proto/components/v1/common.proto";

option go_package = "github.com/dapr/dapr/pkg/proto/components/v1;components";

// Interface for input bindings
service InputBinding {
  // Initializes the inputbinding component component with the given metadata.
  rpc Init(InputBindingInitRequest) returns (InputBindingInitResponse) {}

  // Establishes a stream with the server, which sends messages down to the
  // client. The client streams acknowledgements back to the server. The server
  // will close the stream and return the status on any error. In case of closed
  // connection, the client should re-establish the stream.
  rpc Read(stream ReadRequest) returns (stream ReadResponse) {}

  // Ping the pubsub. Used for liveness porpuses.
  rpc Ping(PingRequest) returns (PingResponse) {}
}

service OutputBinding {
  // Initializes the outputbinding component component with the given metadata.
  rpc Init(OutputBindingInitRequest) returns (OutputBindingInitResponse) {}

  // Invoke remote systems with optional payloads.
  rpc Invoke(InvokeRequest) returns (InvokeResponse) {}

  // ListOperations list system supported operations.
  rpc ListOperations(ListOperationsRequest) returns (ListOperationsResponse) {}

  // Ping the pubsub. Used for liveness porpuses.
  rpc Ping(PingRequest) returns (PingResponse) {}
}
// reserved for future-proof extensibility
message ListOperationsRequest {}

message ListOperationsResponse {
  // the list of all supported component operations.
  repeated string operations = 1;
}

// InputBindingInitRequest is the request for initializing the input binding
// component.
message InputBindingInitRequest {
  // The metadata request.
  MetadataRequest metadata = 1;
}

// reserved for future-proof extensibility
message InputBindingInitResponse {}

[...]

PR #5195, opened 2022-09-18
feature: Support Bindings Pluggable Components

mcandeia:
# Description

This PR adds support for bindings pluggable components.

I had to change a bit the pluggable components CRD to allow use name as a spec property instead of using the `metadata.name`.

Please note that this change is a workaround to unblock the bindings e2e tests and the final design will be what we agree in [this](https://github.com/dapr/dapr/issues/5210) issue.

## Issue reference

Please reference the issue this PR will close: #5194 

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing


charts/dapr/crds/pluggable-components.yaml
```diff
@@ -15,36 +16,38 @@ spec:
...
+              object represents. Servers may infer this from the endpoint the client
+              submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds'
+            type: string
+          metadata:
+            type: object
+          spec:
+            description: PluggableComponentSpec is the spec for a pluggable component.
+            properties:
+              name:
+                description: Name is the component name if not specified metadata.Name
```
artursouza, line 40: This is odd. Is there a way for us to avoid this double name thing?

pkg/components/bindings/registry.go
```diff
@@ -31,11 +31,7 @@ type Registry struct {
 }
 
 // DefaultRegistry is the singleton with the registry.
-var DefaultRegistry *Registry
-
-func init() {
-	DefaultRegistry = NewRegistry()
```
artursouza, line 37: @ItalyPaleAle  Is this OK to move out of init()

pkg/components/pluggable.go
```diff
@@ -50,3 +62,8 @@ type Pluggable struct {
 	// Version is the pluggable component version.
 	Version string
 }
+
+// Category returns the component category based on its pluggable type.
+func (p Pluggable) Category() Category {
```
artursouza, line 67: What is the difference between Type and Category? I saw it being used interchangeably in the code. We should probably pick one if they mean the same thing.

/cc @ItalyPaleAle  @yaron2

artursouza (changes requested):

pkg/components/bindings/registry.go
mcandeia: It was previously aligned offline with @ItalyPaleAle

pkg/components/pluggable.go
mcandeia: Basically, it differs only for bindings.

the way we reference bindings components regardless of being input or output is bindings.$component

so the bindings registry is the only one that has a function to check if the component implements input or/and outputbinding

however they have different registries, so when implementing a pluggable component it should be clear wether the component is input or output,

category maps from the given component “interface” to that component category, note that the category concept already exists it was basically moved to components package

pkg/components/bindings/output_pluggable.go
```diff
@@ -0,0 +1,113 @@
...
+// grpcOutputBinding is a implementation of a outputbinding over a gRPC Protocol.
+type grpcOutputBinding struct {
+	*pluggable.GRPCConnector[proto.OutputBindingClient]
+	bindings.OutputBinding
+	operations []bindings.OperationKind
+}
+
+// Init initializes the grpc outputbinding passing out the metadata to the grpc component.
+func (b *grpcOutputBinding) Init(metadata bindings.Metadata) error {
+	if err := b.Dial(metadata.Name); err != nil {
```
ItalyPaleAle, line 35: If a component, the same component, has both input and output binding support, this will create two connections to the same component. Do you think there'd be a way to use a single connection perhaps?

mcandeia: Yes, that's totally possible and I've been considering this - it's on track. Actually if you look closely you will see that this would happen not only for bindings but the same component can implement bindings, lock, statestore, pubsub and all building blocks at once, this is a matter of implementing all proto services that we have exposed,

However we decided to measure (and let this "improvement" for later) if having multiples connections vs single one is a problem considering the fact that we're using unix sockets which by the way is a very lightweight IPC protocol, it's always a tradeoff between throughput and compute resources,

\- connections / - resources usages / - throughput
\+ connections / + resource usages / + throughput

For bindings specifically, I like the idea of having a connection just for input binding since it's a continuous read stream.

wdyt @ItalyPaleAle ?

ItalyPaleAle: I don't think it's likely that the same component will implement more than 1 component type (e.g. a binding + a state store), so I am not concerned with that.

If you are using UDS perhaps this is not as much of a concern. It would be more if you were using HTTP/2 as transport IMHO, as each connection uses another port and multiple connections don't allow the use of HTTP/2 multiplexing. But for UDS perhaps this is ok?

ItalyPaleAle: > do you still think it's OK to keep both connections for now?

Given that you are using UDS rather than HTTP/2, perhaps it's ok ad interim. But if you think this is a small change and want to make it right away.... :)

mcandeia: thanks!

ItalyPaleAle (dismissed):
LGTM but please either implement the change to use a single connection to input+output bindings, or open an issue to track that :)

charts/dapr/crds/pluggable-components.yaml
artursouza, line 40: I was considering use array of types instead to have a normalized experience. On the other hand, it can be harder to extend later since it would need to be an array of tuples (not just `type`).

The best experience IMO would be to rename `name` to `componentName` - this way, it is clear that is not the k8s resource name but the componentName referenced in the component CRD. So, users could have a pluggable component that implements multiple component types with the following PluggableComponent resources in K8s:

(resource) name | metadata.type | metadata.version | metadata.componentName
-|-|-|-
myredis-state | state | 1 | myredis
myredis-inputbinding | inputbinding | 1 | myredis
myredis-outputbinding | outputbinding | 1 | myredis

Then in the component CRD, user would just need to reference `myredis` as a state, inputbinding or outputbinding.

artursouza (changes requested):
See my suggestion to remove the `name` and `metadata.name` ambiguity.

charts/dapr/crds/pluggable-components.yaml
mcandeia: Yes, I agree with you. Applied your suggestions thanks @artursouza

artursouza (dismissed):

artursouza (approved):

Issue #4760, opened 2022-06-09
[ Proposal ] Support for dynamic loading of component configuration in Dapr Self-hosted mode


akhilac1:
/area runtime
/area docs

## Describe the proposal 

Prototype :  [ https://github.com/akhilac1/dapr_delayedComponentLoading](https://github.com/akhilac1/dapr/tree/dapr_delayedComponentLoading)

## What?
This proposal outlines the approach to enable dynamic loading of components using HTTP and gRPC API, when Dapr is running in  *self-hosted* mode. Dapr should be able to persist the configuration in a statestore to gurantee reloading of the components in case of crash of the Dapr process. This proposal focuses on bindings, but can be extended to all supported Dapr components like pub-sub / statestore etc. as a follow-up

## Why?
Cloud-based serverless offerings like Azure Logic Apps, abstract the underlying complexity and integration concerns, enabling seamless interaction with external systems. These solutions build and manage connectors to enable such abstraction. Dapr bindings also addresses the same problem space.  However, components such as bindings cannot be added to a running Dapr process [ in self-hosted mode] and hence user should decide upfront on the various components they wish to connect to. As a result the integration experience in not efficient.

## How?
In self-hosted mode Dapr should supports dynamic loading of configuration, pushed as JSON/YAML payload over HTTP /  gRPC API. Component creation follows the existing flow in Dapr runtime. An additional caveat is to fail the request to dynamically enable component loading if state store is not configured. This is mandatory as Dapr will not be restored to the desired state if there is a failure

Proposed flow for loading the configuration of components is illustrated as follows : 
![63236931-39fb-4915-9dd3-bdcd24f2f09a](https://user-images.githubusercontent.com/3939554/176890210-9a1f0d5c-8f52-45ca-8e5f-d9557656bfa7.jpg)

This can be achieved by enabling the following features in Dapr  for stand-alone mode:

> 1. GET API to get schema of connection, request and response metadata for supported components 
> 2. POST API to request dynamically load component[s] - *Since component loading is an asynchronous process, the API will respond successful loading request creation. User should query back using metadata API to validate if component is loaded.*
> 3. PATCH API to dynamically update loaded components
> 4. DELETE API to dynamically delete the loaded components
> 5. Persist component configuration in state-store to enable configuration recovery in case of failures
> 6. Improve logging to enable 

[Draft API Specification](https://github.com/akhilac1/wip_proposals/blob/main/dapr-metadata-api-spec.json)

 **Related Proposals :** 
[...]

PR #5197, opened 2022-09-19
Dynamic loading of components through file system events

sarsharma:
# Description

Dynamic loading of components can be achieved for standalone mode by adding a feature flag for config CRD
`DynamicLoading`. If this flag it set, we will start a [fsnotify](https://github.com/fsnotify/fsnotify) watcher for the `components-path` directory for file system events, when dapr is running in standalone mode.
Dynamic loading will enable users to add new components for already running daprd. 

* During dapr runtime init, components from the YAML files already present in the `components-path` directory are loaded using the usual component loading flow.
* After these components are loaded, we start a fsnotify watcher goroutine for `components-path`
* Whenever a new file is found this directory (create event), we read the component(s) defined in the file, initialise them,  and maintain a mapping of the file name and components contained in the file.  
* Whenever a file is deleted from the directory (delete event), we check the mapping and give a warning to the user : `file: xyz deleted, component: abc will not be loaded on sidecar restart.` Unloading is not supported, user will have to restart the sidecar to unload the components. Reasons : https://github.com/dapr/dapr/pull/5197#issuecomment-1253540403

## Issue reference

https://github.com/dapr/dapr/issues/4760

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing


pkg/runtime/cli.go
```diff
@@ -55,6 +55,7 @@ func FromFlags() (*DaprRuntime, error) {
 	profilePort := flag.String("profile-port", strconv.Itoa(DefaultProfilePort), "The port for the profile server")
 	appProtocol := flag.String("app-protocol", string(HTTPProtocol), "Protocol for the application: grpc or http")
 	componentsPath := flag.String("components-path", "", "Path for components directory. If empty, components will not be loaded. Self-hosted mode only")
+	dynamicComponentsPath := flag.String("dynamic-components-path", "", "Path of components directory from where components are loaded dynamically. If empty, dynamic loading of components will be disabled. Self-hosted mode")
```
mcandeia, line 58: Do we have a reason to not reuse the `components-path` flag? Is there a usecase when dynamic is not desired?

pkg/runtime/runtime.go
```diff
@@ -2154,6 +2169,25 @@ func (a *DaprRuntime) isComponentAuthorized(component componentsV1alpha1.Compone
 	return true
 }
 
+func (a *DaprRuntime) IsComponentLoaded(component componentsV1alpha1.Component) bool {
+	for _, loadedComp := range a.components {
```
mcandeia, line 2173: I guess we need a readlock RLock/RUnlock here, correct?

pkg/runtime/cli.go
akhilac1: @mcandeia - The intention is to preserve the baseline configuration of dapr sidecar and disallow the app from changing this at runtime. App will not be able to unload components that were loaded as part of initial configuration.

pkg/runtime/runtime.go
sarsharma: added, thanks!

sarsharma:
@ItalyPaleAle 
Right now when the yaml file is removed from the directory, the components that were loaded from that file are removed from `DaprRuntime.components` list. It would not show up in the /metadata API response. This is not actual unloading since that component would still be running and functional. With this PR I was focusing on loading a component dynamically after daprd has started. Unloading might probably require deleting the component from `a.pubsubs`, `a.inputbindings`, `a.configurationstores` etc. safely.

pkg/runtime/cli.go
mcandeia: Got it,
In that case, wouldn't be better to use something policy-based per component like we do for [denying](https://github.com/dapr/dapr/blob/master/pkg/config/configuration.go#L227) components?

sarsharma: added a CLI flag to enable dynamic loading for the same directory `components-path`, instead of having a separate path arg.

pkg/runtime/runtime.go
```diff
@@ -573,6 +578,17 @@ func (a *DaprRuntime) initRuntime(opts *runtimeOpts) error {
...
 
+	// Enable dynamic loading in standalone mode if dynamic components directory is provided
+	if a.runtimeConfig.Mode == modes.StandaloneMode && a.runtimeConfig.Standalone.EnableDynamicLoading == true {
+		dir, err := os.Stat(a.runtimeConfig.Standalone.ComponentsPath)
+		if err != nil {
+			log.Fatalf("failed to get components directory: %s", err)
+		} else if !dir.IsDir() {
+			log.Fatalf("components path is not a directory: %s", a.runtimeConfig.Standalone.ComponentsPath)
+		} else {
+			a.watchPathForDynamicLoading()
```
akhilac1, line 589: Suggestion - after components are loaded successfully we can start watching for dynamic loading - after a.loadComponents

pkg/runtime/runtime.go
```diff
@@ -2162,6 +2178,27 @@ func (a *DaprRuntime) isComponentAuthorized(component componentsV1alpha1.Compone
...
 }
 
+func (a *DaprRuntime) IsComponentLoaded(component componentsV1alpha1.Component) bool {
+	a.componentsLock.RLock()
+	defer a.componentsLock.RUnlock()
+	for _, loadedComp := range a.components {
+		if strings.Compare(component.ObjectMeta.Name, loadedComp.ObjectMeta.Name) == 0 {
+			if strings.Compare(component.Spec.Type, loadedComp.Spec.Type) == 0 {
+				if strings.Compare(component.Spec.Version, loadedComp.Spec.Version) == 0 {
+					log.Warnf("Cannot load multiple instances of same component with dynamic loading. Component name already in use - name: %s type: %s/%s. Skipping dynamic loading.", component.ObjectMeta.Name, component.Spec.Type, component.Spec.Version)
```
akhilac1, line 2188: Actually we can load multiple instances of a component - say kafka. The challenge is using the same identifier. Also, this method only validates if a component matches an already loaded component. So there should not be any warnings logged in this method. IMHO it should be in the calling method.

akhilac1 (changes requested):

akhilac1 (changes requested):

pkg/runtime/dynamicload.go
```diff
@@ -11,23 +10,6 @@ import (
...
-		}
-		for _, file := range files {
-			err := a.loadDynamicComponents(filepath.Join(a.runtimeConfig.Standalone.DynamicComponentsPath, file.Name()))
-			if err != nil {
-				log.Errorf("failed to load components from file : %s", file.Name())
-			}
-		}
-	}
 
 	watcher, err := fsnotify.NewWatcher()
```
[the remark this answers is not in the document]
sarsharma: updated to allow dapr to continue without dynamic loading support if unable to create new watcher

pkg/runtime/runtime.go
sarsharma: Changed to using reflect.DeepEqual() for exact component match. Warning now logged from calling method.

pkg/runtime/dynamicload.go
```diff
@@ -0,0 +1,102 @@
...
+					err := a.loadDynamicComponents(event.Name)
+					if err != nil {
+						log.Errorf("failed to load components from file: %s", event.Name)
+					}
+				} else if event.Op == fsnotify.Remove || event.Op == fsnotify.Rename {
+					log.Warnf("manifest file: %s removed from components directory.", filepath.Base(event.Name))
+					for _, comp := range a.dynamicComponents[DynamicComponentsManifest(event.Name)] {
+						log.Warnf("manifest for dynamic loaded component: %s of type: %s removed.", comp.Name, comp.Spec.Type)
+					}
+					log.Warnf("to unload components from dapr, please restart dapr.")
```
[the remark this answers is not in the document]
sarsharma: Rename|Remove events are emitted for file deletion. For actual file rename, it would give Chmod event. This is also OS specific. We will have to watch for any file edits also if we consider rename.

pkg/runtime/dynamicload.go
```diff
@@ -0,0 +1,102 @@
...
+				}
+			case err, ok := <-watcher.Errors:
+				if !ok {
```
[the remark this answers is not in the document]
sarsharma: ok could be false if it is a zero value generated because the channel is closed and empty. In that case err might not be populated.  https://go.dev/ref/spec#Receive_operator
These are the default max watch numbers given in fsnotify readme https://github.com/fsnotify/fsnotify#platform-specific-notes

pkg/runtime/dynamicload.go
```diff
@@ -0,0 +1,102 @@
...
+		err = a.initDynamicComponent(comp)
+		if err != nil {
+			log.Errorf("error initializing component %s, error: %s", comp.Name, err.Error())
+		}
+	}
+	return nil
+}
+
+func (a *DaprRuntime) initDynamicComponent(comp componentsV1alpha1.Component) error {
+	cat := a.extractComponentCategory(comp)
```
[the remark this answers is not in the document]
sarsharma: We need extra initialisation steps for subscribing to pubsub/ start reading from bindings. This is not required for other component types. Added a comment for that.

pkg/runtime/config.go
```diff
@@ -149,7 +149,7 @@ func NewRuntimeConfig(opts NewRuntimeConfigOpts) *Config {
 		AllowedOrigins:      opts.AllowedOrigins,
 		Standalone: config.StandaloneConfig{
 			ComponentsPath: opts.ComponentsPath,
-			DynamicComponentsPath: opts.DynamicComponentsPath,
+			EnableDynamicLoading: opts.EnableDynamicLoading,
```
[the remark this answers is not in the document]
sarsharma: added validation in cli.go

pkg/runtime/dynamicload.go
```diff
@@ -0,0 +1,97 @@
...
+	if cat == pubsubComponent {
+		a.beginPubSub(comp.Name)
+	} else if cat == bindingsComponent {
+		if a.bindingsRegistry.HasInputBinding(comp.Spec.Type, comp.Spec.Version) {
+			a.readFromBinding(a.inputBindingsCtx, comp.Name, a.inputBindings[comp.Name])
+		}
+	}
+	// No separate initialization is required for other component categories.
+	return nil
```
mukundansundar, line 96: error out if it is a different category?

pkg/runtime/dynamicload.go
```diff
@@ -0,0 +1,97 @@
...
+	authorizedComps := a.getAuthorizedComponents(componentsToLoad)
+	a.componentsLock.Lock()
+	a.components = append(a.components, authorizedComps...)
+	a.componentsLock.Unlock()
+
+	for _, comp := range authorizedComps {
+		err := a.processComponentAndDependents(comp)
+		if err != nil {
+			if !comp.Spec.IgnoreErrors {
+				log.Warnf("error processing component %s, error: %s", comp.Name, err.Error())
```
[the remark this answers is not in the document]
sarsharma: The intention was that daprd should not stop if user is trying to load component dynamically and there is some error in the manifest, giving a chance to rectify the issue without a sidecar restart.

pkg/runtime/dynamicload.go
```diff
@@ -0,0 +1,116 @@
...
+func (a *DaprRuntime) watchPathForDynamicLoading() {
+	watcher, err := fsnotify.NewWatcher()
+	if err != nil {
+		log.Errorf("unable to create watcher for dynamic components, dynamic loading will not be supported: %s", err)
+		return
+	}
+	defer watcher.Close()
+	err = watcher.Add(a.runtimeConfig.Standalone.ComponentsPath)
+	if err != nil {
+		log.Errorf("unable to watch components directory: %s , err: %s , dynamic loading will not be supported", a.runtimeConfig.Standalone.ComponentsPath, err)
```
pravinpushkar, line 22: nit: .... dynamic loading of **components** ....

pkg/runtime/dynamicload.go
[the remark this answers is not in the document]
mukundansundar: The intent behind `ignoreErrors` field is lost in the case of dynamic components then. We are saying that even if there is some error in loading a dynamic component, then `daprd` will continue on as such. 

@dapr/maintainers-dapr Is this the intended flow for dynamic components?

pravinpushkar:
@sarsharma  Do we want to keep filename as `dynamicload.go` or put something more meaningful as in what is loaded dynamically here..like `dynamic_component_loader.go` etc.

pkg/runtime/runtime.go
```diff
@@ -1424,7 +1424,7 @@ func (a *DaprRuntime) getGRPCAPI() grpc.API {
 }
 
 func (a *DaprRuntime) getPublishAdapter() runtimePubsub.Adapter {
-	if len(a.pubSubs) == 0 {
+	if len(a.pubSubs) == 0 && !a.runtimeConfig.Standalone.EnableDynamicLoading {
```
mukundansundar, line 1427: Why is this condition added here?

mukundansundar (commented):
should dynamic loading of components be behind the preview feature flag ?

pkg/runtime/dynamicload.go
sarsharma: updated!

pkg/runtime/runtime.go
sarsharma: It is to pass the pubsubAdapter to the api even if currently there are no pubsub components in the runtime. Since the API servers get initialised during dapr runtime init, incase we want to add pubsub components later through dynamic loading and use it using API, we need the adapter. Tweaked the condition and added a comment for more clarity.

sarsharma:
> should dynamic loading of components be behind the preview feature flag ?

We could add it to CLI preview features at https://docs.dapr.io/operations/support/support-preview-features/
The doc says for CLI there is no explicit opt-in.

sarsharma:
@yaron2 @artursouza @halspang

pkg/runtime/dynamic_components_loader.go
```diff
@@ -0,0 +1,129 @@
...
+			}
+			log.Debug("observed a file event in components directory:", event)
+			if event.Op == fsnotify.Create || event.Op == fsnotify.Write {
+				err := a.loadDynamicComponents(event.Name)
+				if err != nil {
+					log.Errorf("failed to load components from file: %s err: %w", event.Name, err)
+				}
+			} else if event.Op == fsnotify.Remove || event.Op == fsnotify.Rename {
+				for _, comp := range a.dynamicComponents[DynamicComponentsManifest(event.Name)] {
+					log.Warnf("file: %s deleted, component: %s will not be loaded on sidecar restart.", filepath.Base(event.Name), comp.Name)
```
ItalyPaleAle, line 52: Nit: `will not be loaded when daprd is restarted` (because technically "sidecar" is a K8s-only concept)

pkg/runtime/runtime.go
```diff
@@ -2150,6 +2170,17 @@ func (a *DaprRuntime) isComponentAuthorized(component componentsV1alpha1.Compone
 	return true
 }
 
+func (a *DaprRuntime) IsComponentLoaded(component componentsV1alpha1.Component) bool {
+	a.componentsLock.RLock()
+	defer a.componentsLock.RUnlock()
+	for _, loadedComp := range a.components {
+		if reflect.DeepEqual(loadedComp, component) {
```
ItalyPaleAle, line 2177: Is there a better way to check if a component is already loaded besides using `reflect.DeepEqual`? Since the properties of `component` can change throughout its lifecycle, I presume?

pkg/runtime/cli.go
```diff
@@ -55,6 +55,7 @@ func FromFlags() (*DaprRuntime, error) {
 	profilePort := flag.String("profile-port", strconv.Itoa(DefaultProfilePort), "The port for the profile server")
 	appProtocol := flag.String("app-protocol", string(HTTPProtocol), "Protocol for the application: grpc or http")
 	componentsPath := flag.String("components-path", "", "Path for components directory. If empty, components will not be loaded. Self-hosted mode only")
+	enableDynamicLoading := flag.Bool("enable-dynamic-loading", false, "Allow dynamic loading of components. This option is supported in self-hosted mode only")
```
artursouza, line 58: Can this be moved to the config CRD instead?

artursouza (changes requested):

pkg/runtime/runtime.go
sarsharma: Updated to compare using fields that are read only throughout component lifecycle (ObjectMetadata.Name, Spec.Type, Version, TypeMeta)
https://pkg.go.dev/k8s.io/apimachinery@v0.23.0/pkg/apis/meta/v1?utm_source=gopls#ObjectMeta

pkg/runtime/cli.go
sarsharma: switched to config feature instead of cli flag

ItalyPaleAle (dismissed):

mukundansundar (approved):
Lgtm

sarsharma:
@yaron2

akhilac1 (approved):
LGTM

pkg/runtime/dynamicload.go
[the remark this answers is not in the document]
yaron2: Then again, not restarting the sidecar would make the error in the new dynamically loaded component go unnoticed in most cases, which is what `ignoreErrors` is trying to solve for mainly. I recommend to honor `ignoreErrors` for dynamically loaded components.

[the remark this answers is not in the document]
akhilac1: @yaron2 / @mukundansundar - for a daprd startup ignore errors indicates that all the configuration SHOULD be loaded else, dapr should not continue to run. Hence the implementation to flag loading failures as critical and trigger a restart, which is the most we can do to initiate recovery.
For dynamic loading however, the intent is to minimise disruption to the running setup. So user is able to correct the error and retry the loading. Since the loading is halted at the initial phase, there is no impact on the rest of the daprd. Restart would not add any value since we would technically be bringing Daprd back to the state that it was in before restart. Also, some configuration needs to be redone, for example subscription to configuration via the config building block. Hence the approach to log a proper error and allow daprd to function un-interrupted. 
So ignore errors would be true always for dynamic loading.

go.mod
```diff
@@ -186,7 +186,7 @@ require (
 	github.com/facebookgo/clock v0.0.0-20150410010913-600d898af40a // indirect
 	github.com/fasthttp-contrib/sessions v0.0.0-20160905201309-74f6ac73d5d5 // indirect
 	github.com/fatih/color v1.13.0 // indirect
-	github.com/fsnotify/fsnotify v1.6.0 // indirect
+	github.com/fsnotify/fsnotify v1.6.0
```
ItalyPaleAle, line 189: Don't forget `make modtidy-all` please

pkg/runtime/cli.go
```diff
@@ -369,6 +369,13 @@ func FromFlags() (*DaprRuntime, error) {
 		runtimeConfig.AppHealthCheck = nil
 	}
 
+	if daprGlobalConfig.IsFeatureEnabled(globalConfig.Spec.Features, daprGlobalConfig.DynamicLoading) {
```
ItalyPaleAle, line 372: Not a fan of this approach because config files, especially in self-hosted, are often put in `~/.dapr/config.yaml` and auto-loaded by all apps.

1. Do we absolutely need a feature flag? Can't we use a CLI flag instead, that is adding a new flag to `daprd` similar to `--watch-path <path>` (maybe as an alternative to `--components-path`, which then doesn't use hot-reload) or `--enable-component-reloading` (boolean) or whatever else?
2. If we absolutely need a feature flag, we shouldn't crash the runtime if components-path is unset. Just just a warning and don't enable hot-reloading. Because of what I said above, taht config files are often reused across many apps.

pkg/runtime/dynamic_components_loader.go
```diff
@@ -0,0 +1,134 @@
...
+package runtime
+
+import (
+	"path/filepath"
+
+	"github.com/fsnotify/fsnotify"
+	"github.com/pkg/errors"
+
+	"github.com/dapr/dapr/pkg/components"
+
```
ItalyPaleAle, line 23: Nit: remove empty line here

pkg/runtime/dynamic_components_loader.go
```diff
@@ -0,0 +1,134 @@
...
+	componentsV1alpha1 "github.com/dapr/dapr/pkg/apis/components/v1alpha1"
+)
+
+func (a *DaprRuntime) watchPathForDynamicLoading() {
```
ItalyPaleAle, line 27: Should probably accept a `context.Context` that can be used to stop the watcher. Right now, there's no way to stop it.

pkg/runtime/runtime.go
```diff
@@ -2391,6 +2414,21 @@ func (a *DaprRuntime) isComponentAuthorized(component componentsV1alpha1.Compone
 	return true
 }
 
+func (a *DaprRuntime) IsComponentLoaded(component componentsV1alpha1.Component) bool {
+	a.componentsLock.RLock()
+	defer a.componentsLock.RUnlock()
+	for _, loadedComp := range a.components {
+		if loadedComp.Spec.Type == component.Spec.Type && loadedComp.ObjectMeta.Name == component.ObjectMeta.Name {
+			if loadedComp.Spec.Version == component.Spec.Version && loadedComp.TypeMeta == component.TypeMeta {
+				if reflect.DeepEqual(loadedComp.Spec.Metadata, component.Spec.Metadata) {
```
ItalyPaleAle, line 2423: Can you squash these `if`'s in a single one?

pkg/runtime/runtime.go
```diff
@@ -478,6 +484,19 @@ func (a *DaprRuntime) initRuntime(opts *runtimeOpts) error {
 		log.Warnf("failed to load components: %s", err)
 	}
 
+	// Enable dynamic loading in standalone mode if dynamic loading preview feature is enabled and components path is provided
+	if a.runtimeConfig.Mode == modes.StandaloneMode && a.runtimeConfig.Standalone.EnableDynamicLoading {
```
ItalyPaleAle, line 488: I would lean towards saying that this block should be moved at the very end of the initialization of the runtime. So we don't risk interfering with the loading of "static" components. WDYT?

pkg/runtime/runtime_test.go
```diff
@@ -5339,6 +5340,65 @@ func matchContextInterface(v any) bool {
 	return ok
 }
 
+func TestDynamicLoading(t *testing.T) {
```
ItalyPaleAle, line 5343: Nit: consider moving this to `dynamic_components_loader_test.go`

ItalyPaleAle (changes requested):

pkg/runtime/dynamicload.go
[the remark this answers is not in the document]
akhilac1: @yaron2 -  Exploring options to make the loading error visible for action - One thought is to be able to surface errors to the app instead of restarting the sidecar, we could send a notification on the app channel with a relevant error code and the detailed failure reason. This would bring attention on the failure. So, if ignoreErrors is set to false, a notification can be sent to the app channel when a dynamically loaded component fails during load. Any thoughts on this?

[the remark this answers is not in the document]
ItalyPaleAle: @akhilac1 was the intent for this feature to be used primarily for development (so was my understanding) or for production? In the first case, asking the developer to add a route to the app, seems especially an overkill, as they'd also be responsible for removing the route in production.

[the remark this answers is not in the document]
mukundansundar: There are two options that I can see here. 

One we continue to honor `ignoreErrors` and that will disrupt sidecar runs for any dynamic component that fails to initialize. This way `ignoreErrors` as a property will be consistent in definition and use across both normal components and dynamically loaded components. 

The other option is that we ignore the `ignoreErrors` field to reduce disruptions to sidecar. But in this case we need a way to surface the component init errors to the app/user. **Error logs are not sufficient for that.** Even if we have a way to notify the app that the dynamic load of a component failed, we also need to clearly document the behavioral differences between `ignoreErrors` field when used with a normal component vs when used with a pluggable component. 

IMO I believe that both options offers different benefits. The first option seems to have a lesser learning curve and lesser implementation change wrt to the app. Also it follows the expected behavior when the field is configured. 
While in the second option, it offers uninterrupted run except for the particular features that the dynamically loaded component alone targets. But it also has different behaviors for a field in different modes, and might potentially lead to code change in the app to consume any component init errors. 

I think we should continue to honor `ignoreErrors` field so that there are no two different behaviors in two different modes. We should maintain the consistency.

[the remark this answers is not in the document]
yaron2: I agree. Hot reloading is meant to allow users to load components without restarting the sidecar. It says nothing about changing the behavior when encountering errors. If a hot reload event occured, at that point this feature ends and the regular lifecycle of component validation begins.

1046102779:
I remember that when the configuration hot loading and update is implemented in dapr's control panel, we don’t recommend dynamically updating the configuration. Of course, the hot updating of the newly added configuration is recommended; but if the configuration of the same resource is changed repeatedly, this behavior is not recommended. Because it may lead to connection leaks, because all APIs of the current Building blocks do not provide a `Close` method to close the connection. I remember like this.

yaron2:
@sarsharma do you intend to complete this PR?

sarsharma:
> @sarsharma do you intend to complete this PR?

Yes, will try to update it soon!

mukundansundar:
@sarsharma any ETA on this PR?

PR #5241, opened 2022-09-23
Pin latest Contrib for CosmosDB fix

berndverst:

# Description

Pin latest Contrib for CosmosDB fix


artursouza:
This is still not enough to fix the regression.

berndverst:
> This is still not enough to fix the regression.

That makes no sense actually. Very weird. I'll write a conformance test instead to see what's going on.

berndverst:
@artursouza as @ItalyPaleAle the issue is that the test app images are cached, so our code changes were never actually run in this test! Alessandro will purge the cache and rerun.

artursouza:
The test app cache should not matter because the fix is in the sidecar.

artursouza (dismissed):

berndverst:
Didn't pin the right contrib commit earlier. Whoops

artursouza (approved):

PR #5248, opened 2022-09-26
Allows to skip E2E tests based on bash script condition.

artursouza:
E2E tests on ARM64 are still flaky. This allows work to stabilize it to happen via PRs and only enable it again once proved stable for the first time.

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing

### DIFF
--- a/.github/workflows/dapr-test.yml
+++ b/.github/workflows/dapr-test.yml
@@ -533,8 +533,12 @@ jobs:
       - name: Show dapr configurations
         if: env.TEST_PREFIX != ''
         run: kubectl get configurations daprsystem -n ${{ env.DAPR_NAMESPACE }} -o yaml
+      - name: Determine if E2E tests should run
+        if: env.TEST_PREFIX != '' && env.TEST_CLOUD_ENV != ''
+        run: ./tests/test-infra/skip_${{ env.TEST_CLOUD_ENV }}.sh
+        shell: bash
       - name: Run E2E tests
-        if: env.TEST_PREFIX != ''
+        if: env.TEST_PREFIX != '' && env.SKIP_E2E != 'true'
         run: make test-e2e-all
       - name: Save control plane K8s resources
         if: always() && env.TEST_PREFIX != ''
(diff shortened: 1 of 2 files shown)


.github/workflows/dapr-test.yml
```diff
@@ -533,8 +533,12 @@ jobs:
       - name: Show dapr configurations
         if: env.TEST_PREFIX != ''
         run: kubectl get configurations daprsystem -n ${{ env.DAPR_NAMESPACE }} -o yaml
+      - name: Determine if E2E tests should run
+        if: env.TEST_PREFIX != '' && env.TEST_CLOUD_ENV != ''
+        run: ./tests/test-infra/skip_${{ env.TEST_CLOUD_ENV }}.sh
+        shell: bash
       - name: Run E2E tests
-        if: env.TEST_PREFIX != ''
+        if: env.TEST_PREFIX != '' && env.SKIP_E2E != 'true'
```
halspang, line 541: Why not remove it here: https://github.com/dapr/dapr/blob/master/.github/workflows/dapr-test.yml#L409

Or just use the already defined variables instead of making a script? Seems like this way will spend a lot of time doing all the setup work only to not run tests.

halspang (changes requested):
This seems like a more complicated way of doing this. What's the backing behind it?

yaron2 (approved):

.github/workflows/dapr-test.yml
artursouza: I have explained why as comments in the script. It will allow us to enable in a PR and run /ok-to-test multiple times to validate the stability before merging to master. Changes to the workflow YAML do not affect /ok-to-test unless merged into master.

halspang: > # This script allows the stabilization to take place prior to merging into master.

This wasn't super clear to me that it was unblocking /ok-to-test. Mentioning it there would be good, but I won't block on it since we have the record in the PR.

halspang (approved):

Issue #5236, opened 2022-09-23
Runtime panic in Invokemethod for Subscribe configuration notification


akhilac1:
## In what area(s)?

 /area runtime

> /area operator

> /area placement

> /area docs

> /area test-and-release

## What version of Dapr?
1.8.x

> 1.1.x
> 1.0.x
> edge: output of `git describe --dirty`

## Expected Behavior

https://github.com/dapr/dapr/blob/9da16f4a2ee9658dc7b26cf154d75c2e7a3128b7/pkg/http/api.go#L806-L811

This code causes panic in runtime when appchannel is nil or when handler is not defined for the specific key
## Actual Behavior

If appchannel is nil or invoke fails, runtime should not panic

## Steps to Reproduce the Problem

- Subscribe to a key in config store
- close the app / donot list a route related to the key
- add/modify the key in configuration. runtime panics
- 
## Release Note

RELEASE NOTE: **UPDATE** - Configuration Building Block- Http Subscribe AlphaV1 API would need the app to be in ready status. 

RELEASE NOTE:

akhilac1:
@artursouza - This Issue is related to runtime panic and needs to be tagged as a blocker for release
@ItalyPaleAle , @berndverst - FYI

mukundansundar:
closed by #5250

PR #5250, opened 2022-09-26
Breaking change in Subscribe configuration HTTP Alpha API - mitigate runtime panic due to empty appchannel

akhilac1:
Signed-off-by: akhilac1 <chetlapalle.akhila@gmail.com>

# Description
When HTTP request to subscribe to a configuration store is made, the request is accepted based on validations on store.
When the keys are updated, the notifications are raised to the app via the app channel. So if appchannel is empty, the user would never receive the notifications and it incurs a compute overhead of consuming database connections and receiving and discarding notifications from the configuration store. So disallow the user from subscribing to configuration store if appchannel is nil

https://github.com/akhilac1/dapr/blob/8130e7ffca658c6ff4984722ad307ae80bd7ed82/pkg/http/api.go#L920-L925

Check if app channel is nil and return error in updateHandler to prevent panic when h.appchannel is referenced

https://github.com/akhilac1/dapr/blob/8130e7ffca658c6ff4984722ad307ae80bd7ed82/pkg/http/api.go#L799-L804

Update unit tests - TestV1Alpha1ConfigurationUnsubscribe- to accept 500 response code when subscription request is created without appchannel being ready and fixes to support testing for unsubscribe scenarios

Update grpc implementation of subscribe method for configuration building block to return an error when unable to successfully post the notification to the client.

## Issue reference

Please reference the issue this PR will close: #https://github.com/dapr/dapr/issues/5236

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests
* [X] Unit tests passing


pkg/http/api_test.go
```diff
@@ -2284,12 +2284,14 @@ func TestV1Alpha1ConfigurationUnsubscribe(t *testing.T) {
 	t.Run("subscribe and unsubscribe configurations", func(t *testing.T) {
 		apiPath1 := fmt.Sprintf("v1.0-alpha1/configuration/%s/subscribe", storeName)
 		resp1 := fakeServer.DoRequest("GET", apiPath1, nil, nil)
-		assert.Equal(t, 200, resp1.StatusCode, "subscribe configuration store, should return 200")
+		assert.Equal(t, 500, resp1.StatusCode, "subscribe configuration store, should return 500 when app channel is empty")
```
artursouza, line 2287: Do we have test scenarios for the happy path still?

akhilac1: We are mocking a fake http server  with the endpoints of interest - 

https://github.com/akhilac1/dapr/blob/8130e7ffca658c6ff4984722ad307ae80bd7ed82/pkg/http/api_test.go#L2174

Happy path would need initialization of the api instance with NewAPI func. So we cannot test happy path subscribe in unit tests.

artursouza: Why did we have a test checking for http 200?

akhilac1: Earlier subscribe call was always passing as long as the name of the configuration store was accurate. There was no check to see if the notifications can be actually consumed. In scenarios where no application was connected and appchannel was nil - code panicked and this was handled in the defer building block. We could do 2 things - 
1. check if app channel is nil on receiving notification from configuration store and silently drop the notification
2. block subscription if the notifications cannot be consumed due to app not being ready or app not available to consume notifications. 
This implementation ensures that there is a consumer for notifications and disallows subscription otherwise using the appchannel check for http subscribe API call.

artursouza (approved):

mukundansundar (approved):
IMO it's a bug fix for panic and returning error on error scenario and it's not a breaking change.

akhilac1:
> IMO it's a bug fix for panic and returning error on error scenario and it's not a breaking change.

@mukundansundar - earlier users could subscribe to configuration updates over HTTP even when there was no app connected to the sidecar. With this change subscription to configuration updates without an appchannel setup [aka app that can consume the notifications] is not supported. So it is good to call out as part of release note.

PR #5254, opened 2022-09-27
Refactor script to check for registry in E2E test

artursouza:
Refactor script to check for E2E registry:

```txt
$ tests/test-infra/wait_azure_registry.sh  daprtests2
The resource with name 'daprtests2' and type 'Microsoft.ContainerRegistry/registries' could not be found in subscription 'Dapr Tests (38875a89-0178-4f27-a141-dc6fc01f183d)'.
Azure Container Registry not ready yet: sleeping for 20 seconds
^C
$ tests/test-infra/wait_azure_registry.sh  daprtests
"/subscriptions/38875a89-0178-4f27-a141-dc6fc01f183d/resourceGroups/dapr-test/providers/Microsoft.ContainerRegistry/registries/daprtests"
$ echo $?
0
```

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing


.github/workflows/dapr-test.yml
```diff
@@ -328,18 +328,7 @@ jobs:
...
-          do
-              echo "Azure Container Registry not ready yet: sleeping for 20 seconds"
-              sleep 20
-              elapsed=$(expr $elapsed + 20)
-          done
-
-          if [ $elapsed -ge 1200 ]; then
-            echo "Azure Container Registry not ready on time"
-          fi
+        run: ./tests/test-infra/wait_${{ env.TEST_CLOUD_ENV }}_registry.sh ${{ env.TEST_PREFIX }}acr
```
artursouza, line 331: Please, read this carefully and check with the actual file path since this code is not verified until merged into master.

.github/workflows/dapr-test.yml
```diff
@@ -594,7 +583,7 @@ jobs:
           name: ${{ matrix.target_os }}_${{ matrix.target_arch }}_e2e.json
           path: ${{ env.TEST_OUTPUT_FILE_PREFIX }}_e2e.*
       - name: Update PR comment for success
-        if: success() && env.PR_NUMBER != ''
+        if: success() && env.PR_NUMBER != '' && env.SKIP_E2E != 'true'
```
artursouza, line 586: same

.github/workflows/dapr-test.yml
```diff
@@ -604,6 +593,19 @@ jobs:
           message: |
             ## ✅ Tests succeeded on ${{ matrix.target_os }}/${{ matrix.target_arch }}
 
+            - Image tag: `${{ env.DAPR_TAG }}`
+            - Test image tag: `${{ env.DAPR_TEST_TAG }}`
+      - name: Update PR comment for skipped test run
```
artursouza, line 598: same.

tests/test-infra/wait_azure_registry.sh
```diff
@@ -0,0 +1,30 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+# Allows custom logic to validate registry readiness.
+# This is useful when we are still working to stabilize some targets.
+# This script allows the stabilization to take place prior to merging into master.
+registry_name=$1
+elapsed=0
+until [ $elapsed -ge 1200 ] || az acr show --name $registry_name --query "id"
```
artursouza, line 21: I validated this locally on localhost, see validation output in PR description.

tests/test-infra/wait_azure_registry.sh
```diff
@@ -0,0 +1,30 @@
...
+do
+  echo "Azure Container Registry not ready yet: sleeping for 20 seconds"
+  sleep 20
+  elapsed=$(expr $elapsed + 20)
+done
+
+if [ $elapsed -ge 1200 ]; then
```
artursouza, line 28: My goal is to reduce this to 300s (5 minutes) in a follow up PR. This refactor should not have any behavior change.

ItalyPaleAle:
I totally get the value of using scripts. The downside however is that when we trigger the pipelines, they always run the pipeline from the master branch, but the scripts are loaded from the repo at the version that is cloned.

For example, say that this gets merged and then you want to run the E2E pipeline on the release-1.8 branch. The pipeline will always come from master, but then the pipeline does a  “git checkout”  for the “release-1.8” branch, which does not have the scripts.

If we go down this path of relying on external scripts, we should have another pipeline to keep other branches in sync.

artursouza:
> I totally get the value of using scripts. The downside however is that when we trigger the pipelines, they always run the pipeline from the master branch, but the scripts are loaded from the repo at the version that is cloned.
> 
> For example, say that this gets merged and then you want to run the E2E pipeline on the release-1.8 branch. The pipeline will always come from master, but then the pipeline does a “git checkout” for the “release-1.8” branch, which does not have the scripts.
> 
> If we go down this path of relying on external scripts, we should have another pipeline to keep other branches in sync.

I think we should add the scripts on-demand once we need to run this workflow on another branch.

.github/workflows/dapr-test.yml
```diff
@@ -604,6 +593,19 @@ jobs:
...
+        if: success() && env.PR_NUMBER != '' && env.SKIP_E2E = 'true'
+        uses: marocchino/sticky-pull-request-comment@v2.2.0
+        with:
+          header: ${{ github.run_id }}
+          number: ${{ env.PR_NUMBER }}
+          append: true
+          GITHUB_TOKEN: ${{ secrets.DAPR_BOT_TOKEN }}
+          message: |
+            ## 😕 Tests skipped on ${{ matrix.target_os }}/${{ matrix.target_arch }}
```
ItalyPaleAle, line 607: Should the emoji here be a yello ⚠️ (to be in line with the ❌ and ✅) ?

ItalyPaleAle:
Should the perf test workflow be updated too? https://github.com/dapr/dapr/blob/master/.github/workflows/dapr-perf.yml#L219

.github/workflows/dapr-test.yml
artursouza: Sure.

.github/workflows/dapr-test.yml
```diff
@@ -604,6 +593,19 @@ jobs:
           message: |
             ## ✅ Tests succeeded on ${{ matrix.target_os }}/${{ matrix.target_arch }}
 
+            - Image tag: `${{ env.DAPR_TAG }}`
+            - Test image tag: `${{ env.DAPR_TEST_TAG }}`
+      - name: Update PR comment for skipped test run
+        if: success() && env.PR_NUMBER != '' && env.SKIP_E2E = 'true'
+        uses: marocchino/sticky-pull-request-comment@v2.2.0
```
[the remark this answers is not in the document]
artursouza: Let me fork that too.

artursouza:
> Should the perf test workflow be updated too? https://github.com/dapr/dapr/blob/master/.github/workflows/dapr-perf.yml#L219

There is less of a demand for that because we don't have the instability with ARM64 there.

ItalyPaleAle (approved):

yaron2 (approved):

PR #5270, opened 2022-09-28
Fix failing test apps build

mcandeia:
Signed-off-by: Marcos Candeia <marrcooos@gmail.com>

# Description

Tests are failing in newly created branches when trying to build apps due to empty cache. Tests are working in master because the apps didn't change so the cache is used instead.

This PR updates the apps go module to make it work again with or without cache.

## Issue reference

n/a

## Checklist

Please make sure you've  completed the relevant tasks for this PR, out of the following list:

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing
* [ ] End-to-end tests passing
* [ ] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_
* [ ] Specification has been updated / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_
* [ ] Provided sample for the feature / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_

PR #5278, opened 2022-09-28
Build tools: add support for cache-include files 

ItalyPaleAle:
These optional files placed in E2E apps include a list of files that should be included when computing the hash.

Should allow better support for testing in feature branches, as well as should reduce the need to manually purge the E2E app cache in certain situations.

See https://github.com/dapr/dapr/pull/5270


.build-tools/cmd/zz-e2e-perf.go
```diff
@@ -417,25 +423,98 @@ func (c *cmdE2EPerf) getHashDir() (string, error) {
...
+			for _, match := range matches {
+				if match == "" {
+					continue
+				}
+				info, err := os.Stat(match)
+				if err != nil {
+					continue
+				}
+				if info.IsDir() {
+					addFiles, err := hashFilesInDir(match, nil)
```
mcandeia, line 498: should we check for ignored files here?

I think ignores should have precedence over includes, WDYT?

.build-tools/cmd/zz-e2e-perf.go
```diff
@@ -417,25 +423,98 @@ func (c *cmdE2EPerf) getHashDir() (string, error) {
...
+	}
+
+	// Load the files to exclude
+	ignores := c.getIgnores()
+
+	// Load additional paths to include
+	includes := c.getIncludes()
+
+	// Compute the hash of the app's files
+	files, err := hashFilesInDir(basePath, ignores)
```
mcandeia, line 471: nit: here is another way of seeing this:

how about making the `basePath/**` as part of "includes" patterns?

.build-tools/cmd/zz-e2e-perf.go
```diff
@@ -417,25 +423,98 @@ func (c *cmdE2EPerf) getHashDir() (string, error) {
...
 	if err != nil {
 		return "", err
 	}
 	if len(files) == 0 {
 		return "", fmt.Errorf("no file found in the folder")
 	}
 
+	// Include other files or folders as needed
+	if len(includes) > 0 {
```
mcandeia, line 480: I guess you don't need to check for length as if includes are empty so the loop will not be executed, right?

tests/apps/actorapp/.cache-include
```diff
@@ -0,0 +1 @@
+../utils/*.go
```
mcandeia, line 1: at least for me, sounds more natural if the path is relative from root rather than relative from app directory,

but I don't have a strong opinion though

mcandeia (changes requested):

.build-tools/cmd/zz-e2e-perf.go
ItalyPaleAle: I did spend some thoughts on these.

My thought was that .gitignore files are usually specific for the folder they live in, and the includes are in other folders, so it felt "wrong" to apply the same .gitignores. Also, if the included paths are specific enough (you can see that I've tried to be specific, such as also adding `*.proto` or `*.go` at the end), then ignores shouldn't matter.

WDYT?

.build-tools/cmd/zz-e2e-perf.go
ItalyPaleAle: Only if we decide that `ignores` applies to those too, per conversation above.

.build-tools/cmd/zz-e2e-perf.go
ItalyPaleAle: `includes` could be `nil` so this is actually a nil-check

tests/apps/actorapp/.cache-include
ItalyPaleAle: I don't have a strong opinion either.

ItalyPaleAle: Although thinking about it, implementation-wise, "root" is a bit harder to compute, because the build-tools app doesn't know what is the project's root. Doing that is possible but would require some other refactorings.

.build-tools/cmd/zz-e2e-perf.go
mcandeia: Looks good, unless we have a usecase for including an entire folder and ignore a single file there, so the way to go in this scenario is a bit verbose than including dir/** and ignoring dir/onlythisfile

if we do, we can change it later

.build-tools/cmd/zz-e2e-perf.go
mcandeia: I was not sure about that so I tested it out https://go.dev/play/p/ar7iqfsZtpR

and you can actually iterate over nil arrays which won’t result in a nil dereference panic

tests/apps/actorapp/.cache-include
mcandeia: I can live with that

.build-tools/cmd/zz-e2e-perf.go
mcandeia: can you add a comment on it explaining the rational to not use ignores?

ItalyPaleAle: Done

.build-tools/cmd/zz-e2e-perf.go
ItalyPaleAle: Done

mcandeia (approved):

artursouza (approved):

PR #5286, opened 2022-09-29
Convert Dapr middlewares from fasthttp to net/http

ItalyPaleAle:
# Description

This PR (and its [companion PR in components-contrib](https://github.com/dapr/components-contrib/pull/2143)) converts all Dapr middlewares from fasthttp to net/http (standard library).

Why:

1. fasthttp as a server is going to be removed from Dapr, so the middlewares need to be converted (#4979)
2. When streaming is merged (#5170), the HTTP app channel switches from using fasthttp as client to net/http, so fasthttp middlewares can't be used (as implemented in #5132)
3. From the perspective of end users, writing middlewares for net/http is arguably simpler than writing them for fasthttp, given that it's part of the standard library and has lots of examples and existing libraries

## Issue reference

Part of #4979 

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests
* [X] Unit tests passing


ItalyPaleAle:
Closed in favor of #5316 which also changes the HTTP channel to use net/http

PR #5287, opened 2022-09-29
Sidecar watchdog fixes

ItalyPaleAle:
1. Do not run an iteration of the sidecar watchdog if the dapr-sidecar-injector deployment is not running or not ready, otherwise we'd kill pods that would come back up without the sidecar again
2. When running the watchdog only "once", make sure that there's at least one successful attempt completed (including one attempt with the sidecar injector deployment ready)
3. Fixed an issue that caused the watchdog to run an iteration when the method was returning. It could have caused the watchdog to run twice when it should have only run "once"


pkg/operator/watchdog.go
```diff
@@ -115,17 +129,36 @@ forloop:
...
 }
 
-func (dw *DaprWatchdog) listPods(ctx context.Context) {
+func (dw *DaprWatchdog) listPods(ctx context.Context) bool {
 	log.Infof("DaprWatchdog started checking pods")
 
+	// Look for the dapr-sidecar-injector deployment first and ensure it's running
+	// Otherwise, the pods would come back up without the sidecar again
+	deployment := &appsv1.DeploymentList{}
+	err := dw.client.List(ctx, deployment, &client.ListOptions{
```
yaron2, line 138: If we were to list only in the Dapr system namespace, would that result in removing the list deployments in all namespaces permission or do we have that anyway l?

ItalyPaleAle: We still need to list the workload pods, correct. 

For the sidecar injector, I'm listing the deployments rather than the pods, to be able to check if the deployment has at least 1 pod in "ready" state.

yaron2 (approved):

PR #5296, opened 2022-09-30
Fixed all race conditions in tests in multiple packages

### LINKED ISSUE #5289: Race conditions causing flaky unit test for pubsub unsubscribe on Windows
## In what area(s)?

 /area runtime

> /area operator

> /area placement

> /area docs

> /area test-and-release

## What version of Dapr?

1.9

## Expected Behavior

This unit test seems to fail with some frequency on Windows:

image

## Actual Behavior

Unit test runs at 100% success rate (5 times in a row should be enough).

Run the unit test on Windows.

## Release Note

RELEASE NOTE:

ItalyPaleAle:
Fixes #5289

Fixes race conditions in multiple packages (most notably pkg/runtime and pkg/actors, among others), so running `go test -race .` reports no errors.

### DIFF
--- a/pkg/runtime/runtime.go
+++ b/pkg/runtime/runtime.go
@@ -1749,9 +1749,9 @@ func (a *DaprRuntime) initPubSub(c componentsV1alpha1.Component) error {
 // This method is used by the HTTP and gRPC APIs.
 func (a *DaprRuntime) Publish(req *pubsub.PublishRequest) error {
 	a.topicsLock.RLock()
-	defer a.topicsLock.RUnlock()
-
 	ps, ok := a.pubSubs[req.PubsubName]
+	a.topicsLock.RUnlock()
+
 	if !ok {
 		return runtimePubsub.NotFoundError{PubsubName: req.PubsubName}
 	}
--- a/pkg/runtime/runtime.go
+++ b/pkg/runtime/runtime.go
@@ -182,7 +182,7 @@ type DaprRuntime struct {
 	operatorClient         operatorv1pb.OperatorClient
 	pubsubCtx              context.Context
 	pubsubCancel           context.CancelFunc
-	topicsLock             *sync.Mutex
+	topicsLock             *sync.RWMutex
 	topicRoutes            map[string]TopicRoutes        // Key is "componentName"
 	topicCtxCancels        map[string]context.CancelFunc // Key is "componentName||topicName"
 	inputBindingRoutes     map[string]string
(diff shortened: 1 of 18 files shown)


artursouza:
> Fixes #5289
> 
> Now, running `go test . -race` in pkg/runtime reports no data races

Can you create a PR adding that to the validation steps for our build going forward? Going into the master branch for that is fine.

artursouza (dismissed):

pkg/runtime/runtime.go
```diff
@@ -1749,6 +1748,9 @@ func (a *DaprRuntime) initPubSub(c componentsV1alpha1.Component) error {
 // And then forward them to the Pub/Sub component.
 // This method is used by the HTTP and gRPC APIs.
 func (a *DaprRuntime) Publish(req *pubsub.PublishRequest) error {
+	a.topicsLock.RLock()
```
yaron2, line 1751: This is likely to have adverse performance implications. If races here are rare (for example, occuring only when someone is trying to publish a message while processing subscriptions on init), I'd remove this.

yaron2 (changes requested):

pkg/runtime/runtime.go
ItalyPaleAle: It can happen every time the subscriptions are changed, for example if apphealth is turning pubsub off.

Although this will negatively impact performance a bit due to the overhead of acquiring a lock, I am using a RWMutex with a read-lock here, so it doesn't block if other goroutines are trying to publish at the same time. It will only block if there's an operation that's modifying the topics/subscriptions concurrently.

What do you think?

artursouza: Is there a way to have this map to be immutable?

artursouza: I prefer the correctness over the performance here. BTW, we don't have performance tests for pubsub (yet).

yaron2: At the very least then lock when accessing the map and not the entire function?

ItalyPaleAle: That sounds very reasonable. Will do

yaron2: > I prefer the correctness over the performance here. BTW, we don't have performance tests for pubsub (yet).

We actually do have pub/sub perf tests, courtesy of @skyao

ItalyPaleAle:
@yaron2 changed the lock in the runtime

Also: 

1. `go test -race` passes on pkg/actors too now
1. added a `make test-race` with 3 packages that are known to pass. I'll use that for a pipeline in another PR targeting master

yaron2 (approved):

mukundansundar:
Qq on this PR should we add a CI check only for files that currently do not have race and fail if they are introduced later on?
@dapr/maintainers-dapr @ItalyPaleAle

ItalyPaleAle:
> Qq on this PR should we add a CI check only for files that currently do not have race and fail if they are introduced later on?

Yes, @artursouza requested that already:

> Can you create a PR adding that to the validation steps for our build going forward? Going into the master branch for that is fine.

I will do that on Monday, I was just waiting for this PR to be merged.

PR #5301, opened 2022-10-03
Dapr WF engine implementation starter

cgillum:
# Description

This is the initial implementation of the Dapr WF engine. This PR contains the following changes to dapr/dapr:

- A new dependency on https://github.com/microsoft/durabletask-go (currently targeting the `dapr` branch)
- A new `InternalActor` abstraction that the workflow engine will rely on, including changes to the existing actor runtime to invoke them
- A new `backend` implementation for the durable task engine based on internal actors, including the first internal actor implementation (of type `dapr.internal.workflow`)
- Changes to the runtime initialization code to initialize the workflow engine, including registering a new gRPC endpoint on the API server

I've implemented just enough of the engine to support executing a "no-op" orchestration using the existing, unmodified .NET Durable Task SDK, which tells us that the core parts are working and that the embedded engine is compatible with the gRPC protocol used by the existing Durable Task SDKs. However, I have not yet implemented any state storage, so the workflows are not yet scalable or reliable. There are also many more features that need to be implemented. However, I wanted to get this PR started to checkpoint my progress and get any initial feedback.

The most important files to review are the existing files, like `go.mod`, `actors.go`, `runtime.go` and `server.go`.  This is a PR to the `feature/workflows` feature branch, so I don't think we don't need a full-blown review for this.

I've also added a new `README.md` where I add more information, including some basic instructions for how to build and test the new feature.

FYI @johnewart @artursouza @yaron2


pkg/actors/actors.go
```diff
@@ -538,11 +569,71 @@ func (a *actorsRuntime) callRemoteActor(
...
+	factory, ok := a.internalActors[act.actorType]
+	if !ok {
+		return nil, fmt.Errorf("internal actor type '%s' is not registered", act.actorType)
+	}
+
+	// The factory is expected to manage its own internal caching of active actor instances
+	impl := factory(a)
+
+	// Internal actors don't support any form of reentrancy (for now)
+	reentrancyID := uuid.New().String()
```
artursouza, line 584: Reentrancy is a stable feature for Actors. We should track that as a work item in the project board.

pkg/actors/actors.go
```diff
@@ -992,11 +1083,17 @@ func (a *actorsRuntime) executeReminder(reminder *Reminder) error {
 	log.Debugf("executing reminder %s for actor type %s with id %s", reminder.Name, reminder.ActorType, reminder.ActorID)
 	req := invokev1.NewInvokeMethodRequest(fmt.Sprintf("remind/%s", reminder.Name))
 	req.WithActor(reminder.ActorType, reminder.ActorID)
-	req.WithRawData(b, invokev1.JSONContentType)
 
 	policy := a.resiliency.ActorPreLockPolicy(context.Background(), reminder.ActorType, reminder.ActorID)
 	return policy(func(ctx context.Context) error {
-		_, err := a.callLocalActor(ctx, req)
+		var err error
+		if a.isInternalActor(reminder.ActorType) {
```
artursouza, line 1090: I think it is better to make each ActorType to declare its contentType, so this logic is generic and not specific for InternalActors.

pkg/actors/actors.go
```diff
@@ -344,7 +371,11 @@ func (a *actorsRuntime) Call(ctx context.Context, req *invokev1.InvokeMethodRequ
 	var err error
 
 	if a.isActorLocal(targetActorAddress, a.config.HostAddress, a.config.Port) {
-		resp, err = a.callLocalActor(ctx, req)
+		if a.isInternalActor(actor.GetActorType()) {
```
artursouza, line 374: What if it is not a local actor? Will internal actors also work as remote?

pkg/actors/actors.go
```diff
@@ -182,6 +194,7 @@ func NewActors(
 		resiliency:             resiliency,
 		storeName:              stateStoreName,
 		isResiliencyEnabled:    configuration.IsFeatureEnabled(features, configuration.Resiliency),
+		internalActors:         internalActors,
```
artursouza, line 197: Is there a way to avoid separating internal vs non-internal actors?

cgillum: I think the alternative would be to use regular actors. However, this seemed like a bad idea from a performance perspective since it introduces an unnecesary gRPC network hop into the same process for local actors. It also complicates the app channel abstraction because we need to support two types of application endpoints, one for internal endpoints (inside the sidecar) and another for endpoints defined in the app. The current design to distinguish internal vs. external actors felt more natural to me and less obtrusive to the overall Dapr sidecar architecture.

cgillum (commented):
Thanks for your feedback. Please see my comments.

pkg/actors/actors.go
```diff
@@ -538,11 +569,71 @@ func (a *actorsRuntime) callRemoteActor(
...
+	// Call the appropriate method based on the method name and verb.
+	var result interface{} = nil
+	var err error = nil
+	if strings.HasPrefix(methodName, "remind/") {
+		reminderName := strings.TrimPrefix(methodName, "remind/")
+		err = impl.InvokeReminder(ctx, act.actorID, reminderName, requestData)
+	} else if strings.HasPrefix(methodName, "timer/") {
+		timerName := strings.TrimPrefix(methodName, "timer/")
+		err = impl.InvokeTimer(ctx, act.actorID, timerName, requestData)
+	} else if verb == commonv1pb.HTTPExtension_DELETE {
```
cgillum, line 603: I'm not sure what to do with the linter warning I'm getting here. It says the following:

> HTTPExtension_DELETE contains underscore. You should use mixedCap or MixedCap. (nosnakecase).

However, this is generated code and I'm not actually declaring it here, just referencing it. I can't change the declaration and I don't see a way to suppress this.

I could maybe do something like this:

```go
} else if int(verb) == 5 /* DELETE */ {
```

This seems like a bad idea for obvious reasons. Any suggestions? I assume this will block the PR from being merged.

pkg/actors/actors.go
cgillum: I think the right way to introduce something like this would be as a new property in `EntityConfig`. But again, I worry that this is out of scope for this PR. Rather, I think this needs to be a separate feature that's created and documented independently, and that we should just continue hardcoding the content types as we have already been doing.

pkg/actors/actors.go
artursouza: So, does that mean that the internal actors can also be invoked remotely?

pkg/actors/actors.go
artursouza: My point was not to avoid the separation as concepts but to wrap them under the same interface, so we can abstract the fact that one is over gRPC and another one is over regular method invocation.

pkg/actors/actors.go
cgillum: Correct. Internal actors can be invoked locally or remotely.

pkg/actors/actors.go
cgillum: I think the best we can do is create a new `AppChannel` implementation for internal actors. Currently there's a single "global" app channel for invoking the app over gRPC. We could create a second one for making local calls and no-op things like health checks. Would that be better?

cgillum:
Per feedback from @artursouza, I'm going to change the design a bit. Rather than have the actor runtime worry about local vs. internal actors, we'll instead have the actor runtime support two types of app channels, the regular one and also an internal one. Internal actors will be implemented inside the internal app channel. This should reduce some of the churn in the actor runtime code.

yaron2:
> Per feedback from @artursouza, I'm going to change the design a bit. Rather than have the actor runtime worry about local vs. internal actors, we'll instead have the actor runtime support two types of app channels, the regular one and also an internal one. Internal actors will be implemented inside the internal app channel. This should reduce some of the churn in the actor runtime code.

Can we break that up into a separate PR?

cgillum:
> > Per feedback from @artursouza, I'm going to change the design a bit. Rather than have the actor runtime worry about local vs. internal actors, we'll instead have the actor runtime support two types of app channels, the regular one and also an internal one. Internal actors will be implemented inside the internal app channel. This should reduce some of the churn in the actor runtime code.
> 
> Can we break that up into a separate PR?

Sorry @yaron2, I only just now noticed your comment and already made this change. It reduces the complexity of the actor runtime changes slightly, so hopefully that makes things a little easier to review.

pkg/actors/actors.go
cgillum: Note that in the latest iteration based on `AppChannel`, we're reusing all the existing reentrancy machinery, which means internal actors can also leverage this feature (if they want).

pkg/actors/actors.go
artursouza: Maybe @johnewart  can pick this up and implement this feature in dapr/dapr.

artursouza (commented):
I like how the runtime changes go now. I still need to take a look at the wfengine/* code.

pkg/runtime/wfengine/backend.go
```diff
@@ -0,0 +1,208 @@
...
+
+// DeleteTaskHub implements backend.Backend
+func (*actorBackend) DeleteTaskHub(context.Context) error {
+	panic("unimplemented")
+}
+
+// GetActivityWorkItem implements backend.Backend
+func (be *actorBackend) GetActivityWorkItem(ctx context.Context) (*backend.ActivityWorkItem, error) {
+	// Wait for the workflow actor to signal us with some work to do
+	select {
```
johnewart, line 164: Should we have a timeout in this select?

pkg/runtime/wfengine/backend.go
```diff
@@ -0,0 +1,208 @@
...
+
+type actorBackend struct {
+	actors                    actors.Actors
+	orchestrationWorkItemChan chan *backend.OrchestrationWorkItem
+	activityWorkItemChan      chan *backend.ActivityWorkItem
+}
+
+func NewActorBackend() *actorBackend {
+	return &actorBackend{
+		orchestrationWorkItemChan: make(chan *backend.OrchestrationWorkItem),
```
johnewart, line 45: Do you deliberately want these channels un-buffered?

pkg/runtime/wfengine/backend.go
```diff
@@ -0,0 +1,208 @@
...
+//
+// Internally, creating a workflow instance also creates a new actor with the same ID. The create
+// request is saved into the actor's "inbox" and then executed via a reminder thread. If the app is
+// scaled out across multiple replicas, the actor might get assigned to a replicas other than this one.
+func (be *actorBackend) CreateOrchestrationInstance(ctx context.Context, e *backend.HistoryEvent) error {
+	if err := be.validateConfiguration(); err != nil {
+		return err
+	}
+
+	workflowInstanceID := e.GetExecutionStarted().GetOrchestrationInstance().GetInstanceId()
```
johnewart, line 74: In looking at the durable task library, these are not guaranteed to be non-nil return values, which could cause this to catch on fire.

pkg/runtime/wfengine/backend.go
```diff
@@ -0,0 +1,208 @@
...
+	if res, err := be.actors.Call(ctx, req); err != nil {
+		return nil, err
+	} else {
+		_, data := res.RawData()
+		if len(data) == 0 {
+			return nil, api.ErrInstanceNotFound
+		}
+
+		// Decode the data using encoding/gob (https://go.dev/blog/gob)
+		buffer := bytes.NewBuffer(data)
```
johnewart, line 112: Using gob is faster than JSON, particularly during decoding, but makes it much harder to debug things where you may need to look at the data; is this a trade-off we want to make? It also means that the only tools that can decode the data have to be written in Go (I think) -- if we care about speed and portability then msgpack may be better if we don't want to use JSON

pkg/runtime/wfengine/workflow.go
```diff
@@ -0,0 +1,245 @@
...
+		wfLogger.Warnf("%s: ignoring run request for workflow with empty inbox", actorID)
+	}
+
+	runtimeState := getAndCacheRuntimeState(actorID, state)
+	wi := &backend.OrchestrationWorkItem{
+		InstanceID: runtimeState.InstanceID(),
+		NewEvents:  state.inbox,
+		RetryCount: -1, // TODO
+		State:      runtimeState,
+		Properties: make(map[string]interface{}),
```
johnewart, line 190: Can we avoid interfaces? They tend to be dangerous and hard to read code; it would be clearer (and safer) if we use explicit members like `CallbackChannel`. Thoughts?

pkg/runtime/wfengine/workflow.go
```diff
@@ -0,0 +1,245 @@
...
+	}
+
+	// Executing workflow code is a one-way operation. We must wait for the app code to report its completion, which
+	// will trigger this callback channel.
+	callback := make(chan bool)
+	wi.Properties[CallbackChannelProperty] = callback
+	wf.scheduler.ScheduleWorkflow(wi)
+	select {
+	case <-ctx.Done():
```
johnewart, line 199: Should this `select` intentionally block indefinitely?

johnewart (changes requested):

johnewart:
I think this is looking pretty good overall; personally I think that we should move some of the actor-relate bits over into the actor package, and there are a few places where code will indefinitely block without timeouts (which may be okay if intentional) or we could bump into NPEs (which would probably be bad 😄)

pkg/actors/actors.go
```diff
@@ -150,6 +151,7 @@ func NewActors(
 	features []configuration.FeatureSpec,
 	resiliency resiliency.Provider,
 	stateStoreName string,
+	workflowEngineChannel channel.AppChannel,
```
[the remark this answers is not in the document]
cgillum: Is your concern just about constructor arguments? If so, yes, I think I can refactor the initialization path. However, making it more generic will bloat the changes to the actor runtime. My previous iteration was more generic such that actors didn't know anything about workflows, but I got pushback on this because of the new concepts that were introduced. I'm a little worried that I won't be able to make everyone happy with this PR.

pkg/runtime/runtime.go
```diff
@@ -1338,7 +1348,7 @@ func (a *DaprRuntime) startGRPCInternalServer(api grpc.API, port int) error {
 
 func (a *DaprRuntime) startGRPCAPIServer(api grpc.API, port int) error {
 	serverConf := a.getNewServerConfig(a.runtimeConfig.APIListenAddresses, port)
-	server := grpc.NewAPIServer(api, serverConf, a.globalConfig.Spec.TracingSpec, a.globalConfig.Spec.MetricSpec, a.globalConfig.Spec.APISpec, a.proxy)
+	server := grpc.NewAPIServer(api, serverConf, a.globalConfig.Spec.TracingSpec, a.globalConfig.Spec.MetricSpec, a.globalConfig.Spec.APISpec, a.proxy, a.workflowEngine)
```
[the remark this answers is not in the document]
cgillum: For non-public APIs, I don't have a strong opinion one way or the other. I don't want to take on this kind of refactoring as part of this PR, however.

pkg/runtime/wfengine/actors.go
```diff
@@ -0,0 +1,133 @@
...
+	commonv1pb "github.com/dapr/dapr/pkg/proto/common/v1"
+)
+
+const (
+	WorkflowActorType = "dapr.wfengine.workflow"
+	ActivityActorType = "dapr.wfengine.activity"
+)
+
+// InternalActor represents the interface for invoking an "internal" actor (one which is built into daprd directly).
+type internalActor interface {
```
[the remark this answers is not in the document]
cgillum: I mentioned this in response to one of @artursouza's comments, but the interface for external actors is defined in the Go SDK. We don't want daprd taking a dependency on a Dapr SDK so I'm instead defining my own interface for internal actors here.

pkg/runtime/wfengine/workflow.go
cgillum: This `OrchestrationWorkItem` struct is defined in durabletask-go. The use of channels, however, is specific to the Dapr Actors-based backend implementation. It therefore doesn't make sense to make channels part of the base abstraction - for example, the sqlite implementation wouldn't use it. It's best to think of `Properties` like you think of the arbitrary key/value pairs in `context.Context` objects where the values are defined by the specific implementation.

pkg/runtime/wfengine/workflow.go
cgillum: No, there needs to be a timeout. I added a TODO comment in the code that calls `runWorkflow()` but I can add a similar comment here too to make the intentions more clear.

pkg/actors/actors.go
```diff
@@ -235,7 +247,7 @@ func (a *actorsRuntime) Init() error {
 }
 
 func (a *actorsRuntime) startAppHealthCheck(opts ...health.Option) {
-	if len(a.config.HostedActorTypes) == 0 {
+	if len(a.config.HostedActorTypes) == 0 || a.appChannel == nil {
```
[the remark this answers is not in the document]
cgillum: There's only ever one app attaching itself to daprd AFAICT, but I'm fine with changing this to support an arbitrary number of `appChannel` objects for internal use if you think that would be a more maintainable design.

pkg/runtime/wfengine/actors.go
```diff
@@ -0,0 +1,133 @@
...
+	invokev1 "github.com/dapr/dapr/pkg/messaging/v1"
+	commonv1pb "github.com/dapr/dapr/pkg/proto/common/v1"
+)
+
+const (
+	WorkflowActorType = "dapr.wfengine.workflow"
+	ActivityActorType = "dapr.wfengine.activity"
+)
+
+// InternalActor represents the interface for invoking an "internal" actor (one which is built into daprd directly).
```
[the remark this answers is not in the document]
cgillum: Fair enough - I'll move it back. :)

yaron2:
Having the actor package know about the workflow package feels wrong; As Workflows are a consumer of actors it makes sense to have workflow import actors only and not the other way around.

cgillum:
> Having the actor package know about the workflow package feels wrong; As Workflows are a consumer of actors it makes sense to have workflow import actors only and not the other way around.

Sounds good. This will be fixed in the next iteration.

pkg/actors/actors.go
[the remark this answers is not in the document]
cgillum: So I can make it so the actor system doesn't need to know about the workflow engine. However, I don't think I can remove the need for a constructor argument. This constructor returns an `actors.Actors` interface. However, the `AppChannel` is an implementation detail that wouldn't make sense to add to the `actors.Actors` interface. On top of that, all the actor initialization happens inside the `Init()` method, meaning all internal actors need to be registered before the runtime is initialized. I looked at this from a few different angles, and there really isn't any cleaner way to inject internal actors via the `AppChannel` abstraction except via this constructor.

pkg/actors/actors.go
```diff
@@ -182,6 +184,7 @@ func NewActors(
 		resiliency:             resiliency,
 		storeName:              stateStoreName,
 		isResiliencyEnabled:    configuration.IsFeatureEnabled(features, configuration.Resiliency),
+		workflowEngineChannel:  workflowEngineChannel,
```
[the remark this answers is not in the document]
cgillum: Done - we instead allow other parts of the system to add internal app channels. The actor subsystem doesn't know or care whether they are workflow actors.

pkg/runtime/runtime.go
```diff
@@ -2103,7 +2113,7 @@ func (a *DaprRuntime) initActors() error {
 		a.namespace, a.appConfig)
 	act := actors.NewActors(a.stateStores[a.actorStateStoreName], a.appChannel, a.grpc.GetGRPCConnection, actorConfig,
 		a.runtimeConfig.CertChain, a.globalConfig.Spec.TracingSpec, a.globalConfig.Spec.Features,
-		a.resiliency, a.actorStateStoreName)
+		a.resiliency, a.actorStateStoreName, a.workflowEngine.AppChannel)
```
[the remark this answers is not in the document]
cgillum: This was already technically O(1), even with the string prefix comparison. However, I've since updated this to use a map. Map lookups are probably slower in terms of CPU cycles, but it feels a bit more robust.

pkg/runtime/wfengine/actors.go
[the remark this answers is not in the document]
cgillum: This is done. Now anyone should be able to implement internal actors pretty easily.

pkg/runtime/wfengine/actors.go
```diff
@@ -0,0 +1,133 @@
...
+type internalActor interface {
+	InvokeMethod(ctx context.Context, actorID string, methodName string, request []byte) (interface{}, error)
+	DeactivateActor(ctx context.Context, actorID string) error
+	InvokeReminder(ctx context.Context, actorID string, reminderName string, params []byte) error
+	InvokeTimer(ctx context.Context, actorID string, timerName string, params []byte) error
+}
+
+type internalActorChannel struct {
+	actors map[string]internalActor
```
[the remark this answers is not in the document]
cgillum: This has been moved into the `actors` package.

pkg/runtime/wfengine/actors.go
```diff
@@ -0,0 +1,133 @@
...
+}
+
+func newInternalActorChannel() *internalActorChannel {
+	return &internalActorChannel{}
```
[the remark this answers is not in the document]
cgillum: Agreed. Fixed.

pkg/runtime/wfengine/workflow.go
cgillum: I went ahead and added a 30-second timeout via `context.WithTimeout` (see further up where this method gets called).

cgillum (commented):
Thanks for the feedback so far. Here's a quick summary of the changes in this latest iteration:

- The actor runtime now knows nothing about workflows. The only new concept it understands is internal actors.
- I made the `InternalActor` a first-class member of the `actors` package that anyone can use. See `internal_actor.go`.
- We have two `AppChannel` variables: `externalAppChannel` (just a rename of the existing field) and `internalAppChannel`, which is responsible for invoking *all* internal actors, regardless of who creates them.

I've also responded to all comments. Let me know if anything else is required for this PR.

pkg/runtime/wfengine/backend.go
johnewart: You can't method chain and let a `try` block catch failures in Go -- trying to call a method on a nil reference will segfault so you will have to explicitly check (or better, return an error along with nil, which is arguably better if those things are error-ish)

pkg/runtime/wfengine/backend.go
johnewart: That seems fine, as long as we aren't planning to store it or humans need to interact with it :+1:

pkg/runtime/wfengine/backend.go
johnewart: That's fine as long as anything calling this expects that behavior :+1: - might it be worth documenting?

pkg/runtime/wfengine/backend.go
cgillum: I need you to trust me that if any of these methods return `nil`, the result of the entire expression chain will be `""`. Every one of these methods test to see whether the receiver is `nil` and avoid segfaulting. When reading this generated code, it's very clear that the designers of protoc did this intentionally to allow this kind of chaining. I take advantage of this everywhere in the durabletask-go code and it works exactly as I expect, allowing the code to stay very concise and elegant.

pkg/actors/actors.go
johnewart: Verb is already an int32 I think -- if so you should be able to define a constant up top like `const HTTPDelete = 5` and then just say `if (verb == HTTPDelete)` (or whatever you want to call the constant)

johnewart (approved):
I think we are in a pretty good place for this round -- thanks for shuffling around the actors / workflow engine relationship bits!

pkg/actors/actors.go
cgillum: I figured out that I can use a `//nolint:nosnakecase` comment to suppress this linter error (found it used elsewhere in the dapr/dapr code for this same type), so I was able to keep the original code and make the error go away.

pkg/runtime/wfengine/workflow.go
```diff
@@ -0,0 +1,251 @@
...
+	return result, err
+}
+
+// InvokeReminder implements actors.InternalActor
+func (wf *workflowActor) InvokeReminder(ctx context.Context, actorID string, reminderName string, params []byte) error {
+	wfLogger.Debugf("invoking reminder '%s' on workflow actor '%s'", reminderName, actorID)
+
+	var err error
+	switch reminderName {
+	case createWorkflowReminder:
```
RyanLettieri, line 88: Are we going to eventually have multiple types of reminders? If not, what is the purpose of using the switch case statement here?

pkg/runtime/wfengine/workflow.go
```diff
@@ -0,0 +1,251 @@
...
+// DeactivateActor implements actors.InternalActor
+func (wf *workflowActor) DeactivateActor(ctx context.Context, actorID string) error {
+	wfLogger.Debugf("deactivating workflow actor '%s'", actorID)
+	delete(wf.states, actorID)
+	return nil
+}
+
+func (wf *workflowActor) createWorkflowInstanceMethod(ctx context.Context, actorID string, startEventBytes []byte) error {
+	// create a new state entry if one doesn't already exist
+	state, exists, err := wf.loadInternalState(actorID)
```
RyanLettieri, line 116: loadInternalState only ever returns nil for the err. Is the check on "if err != nil" below to account for the updating of the loadInternalState function in the future? I saw that there's a todo line on the return in that function:
`return nil, false, nil // TODO: Implement fetching from storage after validating saving to storage`
Is that "TODO" line going to cover the possibility of a non-nil err return?

pkg/runtime/wfengine/workflow.go
```diff
@@ -0,0 +1,251 @@
...
+	)
+	return metadata, nil
+}
+
+func (wf *workflowActor) runWorkflow(ctx context.Context, actorID string) error {
+	state, exists, err := wf.loadInternalState(actorID)
+	if err != nil {
+		return err
+	} else if !exists {
+		return fmt.Errorf("no workflow state found for actor '%s'", actorID)
```
RyanLettieri, line 184: Should we use this return statement on all scenarios where the state does not exist rather than just returning "err"?

pkg/runtime/wfengine/workflow.go
cgillum: Originally, I thought there would be separate reminders, but as I continued working on this in a separate branch, I realized that I only need it for this one thing. This `switch` will be removed in a separate PR.

pkg/runtime/wfengine/workflow.go
cgillum: Yes, exactly. I haven't yet implemented the fetching and decoding of fetched state, and those are opportunities for us to return non-nil error values.

pkg/runtime/wfengine/workflow.go
cgillum: There is one case where we don't return an error if the state doesn't exist, which is in `createWorkflowInstance` (it's expected to not exist in that case). For that reason, we can't have a single behavior for anytime the state isn't found.

cgillum:
Thanks @johnewart and @RyanLettieri for the reviews! I made a few more small changes in the latest commits:
* Merged the latest changes from `feature/workflows`, which included several conflicting changes (mostly around constructor parameters)
* I updated `GetActiveActorsCount` to exclucde internal actors. This was caught by a failing e2e test. Before this change, `GetActiveActorsCount` was returning the internal workflow actor, which I don't think we want.
* Made some of the nil checks more explicit to avoid reader confusion.

Let me know if anyone thinks there's anything else we need to merge this initial PR. Fingers crossed that the CI will be happy this time. 🤞🏽

artursouza (approved):

PR #5313, opened 2022-10-05
Refactor: avoid constructors with dozens of arguments

ItalyPaleAle:
Doesn't add any new feature, just helps with our sanity when we need to add (or remove) options from constructors. Uses a struct instead now.

Example: grpc.NewAPI had 18 parameters. And this was how it was used in tests:

```go
api := NewAPI("", nil, nil, nil, nil, nil, nil, map[string]lock.Store{"mock": mockLockStore}, nil, nil, nil, nil, config.TracingSpec{}, nil, "", nil, nil, nil)
```

image

PS: Spin-off from #5170


johnewart:
For what it's worth I, for one, wholeheartedly appreciate this <3

pkg/grpc/server_test.go
```diff
@@ -105,7 +115,17 @@ func TestClose(t *testing.T) {
...
+			Port:               port,
+			APIListenAddresses: []string{"127.0.0.1"},
+			NameSpace:          "test",
+			TrustDomain:        "test",
+			MaxRequestBodySize: 4,
+			ReadBufferSize:     4,
+			EnableAPILogging:   false,
+		}
 		a := &api{}
 		server := NewAPIServer(a, serverConfig, config.TracingSpec{}, config.MetricSpec{}, config.APISpec{}, nil)
```
johnewart, line 130: Out of curiosity, why not include the tracing, metrics and api specs along with the proxy, in the `ServerConfig` struct?

johnewart (commented):
Just a question on the API server constructor opts

pkg/grpc/server_test.go
```diff
@@ -94,7 +94,17 @@ func TestClose(t *testing.T) {
...
-		serverConfig := NewServerConfig("test", "127.0.0.1", port, []string{"127.0.0.1"}, "test", "test", 4, "", 4, true)
+		serverConfig := ServerConfig{
+			AppID:              "test",
+			HostAddress:        "127.0.0.1",
+			Port:               port,
+			APIListenAddresses: []string{"127.0.0.1"},
+			NameSpace:          "test",
+			TrustDomain:        "test",
+			MaxRequestBodySize: 4,
+			ReadBufferSize:     4,
```
johnewart, line 105: While you're at it, should we include the metric in the option name? `ReadBufferSizeKB/GB/MB/TB....`? 😄

johnewart: Clearly my sort algorithm for SI units is horribly broken...

pkg/grpc/server_test.go
ItalyPaleAle: I have ported this 1:1 so this is what it was doing before. Can change it however if needed.

pkg/grpc/server_test.go
ItalyPaleAle: Sure I can make that change

pkg/actors/config.go
```diff
@@ -55,52 +55,62 @@ const (
 	defaultReentrancyStackLimit = 32
 )
 
+// NewConfigOpts contains options for NewConfig.
+type NewConfigOpts struct {
```
mukundansundar, line 59: Should this just be ConfigOpts or if stuttering is ok , ActorConfigOpts to qualify it s bit more? Similarly for other types ... 

What benefit does New provide ?

Methods are named such since they create a new instance to work with ... @ItalyPaleAle ?

pkg/grpc/api.go
```diff
@@ -267,53 +267,57 @@ func UnlockResponseToGrpcResponse(compResp *lock.UnlockResponse) *runtimev1pb.Un
 	return result
 }
 
+// NewAPIOpts contains options for NewAPI.
+type NewAPIOpts struct {
```
mukundansundar, line 271: Same as above

DeepanshuA:
This is really great work, trying to make code more readable and better usable.
But while we are at it, just thought of discussing, if we should simply create structs and use them, like in current PR; Or it may be worth-while to discuss if should also consider using either a builder pattern (https://golangbyexample.com/builder-pattern-golang/) OR something like function currying (https://medium.com/@meeusdylan/go-reduce-function-parameters-19b785a87a59).
I think it may also depend on case-by-case basis.

mukundansundar:
> This is really great work, trying to make code more readable and better usable.
> 
> But while we are at it, just thought of discussing, if we should simply create structs and use them, like in current PR; Or it may be worth-while to discuss if should also consider using either a builder pattern (https://golangbyexample.com/builder-pattern-golang/) OR something like function currying (https://medium.com/@meeusdylan/go-reduce-function-parameters-19b785a87a59).
> 
> I think it may also depend on case-by-case basis.

I think both have their own scenarios for use. 
For example we use currying in the case of http response building with the respond() function.

Though in this case I believe it is simpler to refactor based on type.

Do we see a large number of use cases (except in tests) where some parameters are omitted and some are not ?
If not I believe this is a simpler way to refactor these set of parameters imo.

ItalyPaleAle:
Tests aside, they are used once or twice, and all properties are always set. Agree that a builder pattern would add more complexity (and more code to maintain).

pkg/grpc/server_test.go
ItalyPaleAle: Changed

pkg/actors/config.go
ItalyPaleAle: Done

pkg/grpc/grpc.go
```diff
@@ -160,15 +157,18 @@ func (g *Manager) GetGRPCConnection(ctx context.Context, address, id string, nam
...
+	cancel()
 	if err != nil {
 		return nil, func() {}, err
 	}
 
 	teardown := releaseFactory(conn)
 	g.lock.Lock()
-	defer g.lock.Unlock()
 	g.connectionPool.Register(address, conn)
+	g.lock.Unlock()
```
mukundansundar, line 171: Is there a possibility of flow never reaching this line due to some issue in Register function ?

ItalyPaleAle: No (aside from perhaps a panic? But that would be a different - and bigger - problem)

DeepanshuA:
> > This is really great work, trying to make code more readable and better usable.
> > But while we are at it, just thought of discussing, if we should simply create structs and use them, like in current PR; Or it may be worth-while to discuss if should also consider using either a builder pattern (https://golangbyexample.com/builder-pattern-golang/) OR something like function currying (https://medium.com/@meeusdylan/go-reduce-function-parameters-19b785a87a59).
> > I think it may also depend on case-by-case basis.
> 
> I think both have their own scenarios for use. For example we use currying in the case of http response building with the respond() function.
> 
> Though in this case I believe it is simpler to refactor based on type.
> 
> Do we see a large number of use cases (except in tests) where some parameters are omitted and some are not ? If not I believe this is a simpler way to refactor these set of parameters imo.

Yeah, if that is the case that parameters are not omitted, except some tests, I too tend to agree that work done here would be good enough. I will take a look at http response building that how we are using currying over there. Thanks for pointing that out.

PR #5315, opened 2022-10-06
Add resiliency metrics to dapr-metrics doc

yaron2:
Adds the list of resiliency specific metrics emitted by `daprd` to docs.

### DIFF
--- a/docs/development/dapr-metrics.md
+++ b/docs/development/dapr-metrics.md
@@ -83,6 +83,11 @@ Dapr uses prometheus process and go collectors by default.
 * dapr_runtime_actor_deactivated_total: The number of the successful actor deactivation.
 * dapr_runtime_actor_deactivated_failed_total: The number of the failed actor deactivation.
 
+#### Resiliency
+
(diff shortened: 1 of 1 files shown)


docs/development/dapr-metrics.md
```diff
@@ -83,6 +83,11 @@ Dapr uses prometheus process and go collectors by default.
 * dapr_runtime_actor_deactivated_total: The number of the successful actor deactivation.
 * dapr_runtime_actor_deactivated_failed_total: The number of the failed actor deactivation.
 
+#### Resiliency
+
+* dapr_runtime_resiliency_loaded: The number of resiliency policies loaded.
+* dapr_runtime_resiliency_count: The number of times a resiliency policy has been executed.
```
mukundansundar, line 89: Would prefer 
dapr_runtime_resiliency_exec_count instead of just count

mukundansundar (approved):

docs/development/dapr-metrics.md
yaron2: We can look into changing this for 1.10

Issue #5183, opened 2022-09-16
Configuration Store: Subscribe to all keys will not notify changes for any new keys


sarsharma:
The dapr docs specify that we can subscribe for all configuration changes if we do not pass any keys to the subscribe endpoint.
* https://docs.dapr.io/reference/api/configuration_api/#query-parameters-1
`GET http://localhost:<daprPort>/v1.0-alpha1/configuration/<storename>/subscribe`

If the user does not pass any params to the endpoint, the config store is queried for all existing keys and then these keys are subscribed to. The current api code will not support subscribing to new keys that are added to the config store once we call this endpoint. 
https://github.com/dapr/dapr/blob/4550ffdd0bee3b460f9e090ba6023c242a6dfca1/pkg/http/api.go#L925

### Expected behaviour 
We should be able to subscribe to all configuration changes in the store, even for keys added after we have called the subscribe endpoint.

akhilac1:
When no keys are specified as part of the subscribe request, the full set of keys are retrieved using Get operation and these keys are added to the subscribe request sent to the configuration component. 

https://github.com/dapr/dapr/blob/4550ffdd0bee3b460f9e090ba6023c242a6dfca1/pkg/http/api.go#L926-L953

the component sees this as a subscribe request for a specific set of keys and subscribes to those keys only. 
Component expects and empty slice when user wants to subscribe to all keys in a configuration store - Refer to Redis using keyspace as "__keyspace@0__:*" when the intent is to subscribe to all keys.

Proposal is to correct the above code block to pass an empty list of keys to the configuration component.

pravinpushkar:
> When no keys are specified as part of the subscribe request, the full set of keys are retrieved using Get operation and these keys are added to the subscribe request sent to the configuration component.
> 
> https://github.com/dapr/dapr/blob/4550ffdd0bee3b460f9e090ba6023c242a6dfca1/pkg/http/api.go#L926-L953
> 
> the component sees this as a subscribe request for a specific set of keys and subscribes to those keys only. Component expects and empty slice when user wants to subscribe to all keys in a configuration store - Refer to Redis using keyspace as "**keyspace@0**:*" when the intent is to subscribe to all keys.
> 
> Proposal is to correct the above code block to pass an empty list of keys to the configuration component.

I agree. Basically we should pass the keys as it is to components. If the keys list is empty then pass the empty list and leave it to the components on how they want to implement subscribe all. Some components have OOB apis for events and notifications like redis while some do not have.

[...]

PR #5317, opened 2022-10-06
fix for config subscribe all

pravinpushkar:

# Description

Removed fetch all key logic when `subscribe all` is needed.

## Issue reference

Please reference the issue this PR will close: #5183 

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing


artursouza (changes requested):
We have cherry-picked this into the release-1.9 - in the meantime, can this have an e2e test to reproduce the issue?

pravinpushkar:
> We have cherry-picked this into the release-1.9 - in the meantime, can this have an e2e test to reproduce the issue?

Thanks for cherry-picking this to 1.9 rel. 

Regarding E2E, ATM we do not have E2E for any of the configuration apis. 
IMO we should create a different issue to add E2E for all config apis and we can cover this scenario as well there
Let me know your thoughts, I would create another issue and PR.

artursouza:
> > We have cherry-picked this into the release-1.9 - in the meantime, can this have an e2e test to reproduce the issue?
> 
> Thanks for cherry-picking this to 1.9 rel.
> 
> Regarding E2E, ATM we do not have E2E for any of the configuration apis. IMO we should create a different issue to add E2E for all config apis and we can cover this scenario as well there. Let me know your thoughts, I can create another issue and PR and we can merge this as is to avoid conflict.

Please, track that in a separate issue. it would be great to have it on 1.10

artursouza:
Zero delta now. Closing PR.

PR #5303, opened 2022-10-03
Cherry pick #5302 into release-1.9

ItalyPaleAle:
Cherry-picks #5302 into the release branch so E2E tests can be run there too.

**PLEASE DO NOT SQUASH** as this is a merge PR.

PR #5319, opened 2022-10-06
Adds fix for config API all keys subscription, pins contrib rc2

berndverst:
# Description

Adds fix for config API all keys subscription, pins contrib rc2


artursouza (approved):

berndverst:
`/ok-to-test` fails because test infrastructure cannot be deployed because the Kubernetes version is no longer available in AKS. Therefore the following PR needs to be merged first.

https://github.com/dapr/dapr/pull/5303/files

PR #5316, opened 2022-10-06
Convert Dapr middlewares and HTTP channel from fasthttp to net/http

ItalyPaleAle:
# Description

This PR (and its [companion PR in components-contrib](https://github.com/dapr/components-contrib/pull/2143)) converts all Dapr middlewares from fasthttp to net/http (standard library).

In order for that to work, I also had to backport a change from #5170, changing the HTTP channel from using fasthttp to net/http. 

Why:

1. fasthttp as a server is going to be removed from Dapr, so the middlewares need to be converted (#4979)
2. When streaming is merged (#5170), the HTTP app channel switches from using fasthttp as client to net/http, so fasthttp middlewares can't be used (as implemented in #5132)
3. From the perspective of end users, writing middlewares for net/http is arguably simpler than writing them for fasthttp, given that it's part of the standard library and has lots of examples and existing libraries

## Issue reference

Part of #4979 

## Checklist

Please make sure you've  completed the relevant tasks for this PR, out of the following list:

* [X] Code compiles correctly
* [X] Created/updated tests
* [X] Unit tests passing
* [x] End-to-end tests passing

PR #5329, opened 2022-10-10
Revert "Convert Dapr middlewares and HTTP channel from fasthttp to net/http"

yaron2:
Reverts dapr/dapr#5316


artursouza (approved):

ItalyPaleAle:
I will work on exploring how to optimize this. We cannot continue using fasthttp so sooner or later this change needs to happen...

yaron2:
> I will work on exploring how to optimize this. We cannot continue using fasthttp so sooner or later this change needs to happen...

We can, and will continue using fasthttp for as long as we cannot guarantee at the very least an acceptable level of performance degredation with net/http. The current regression is extremely high, to be honest I was surprised by how severe it was and am hoping there's a low hanging fruit there that might explain this.

yaron2:
Verified regression in gone in all latest perf runs for the affected test.

PR #5344, opened 2022-10-12
Core WF engine features: activities, timers, sub-workflows, etc.

cgillum:

# Description

In https://github.com/dapr/dapr/pull/5301, we did the initial introduction of the workflow engine into the Dapr sidecar. The engine didn't implement any features - it could only run "no-op" workflows.

This PR introduces the following core features into the workflow engine:

- Activities
- Durable timers
- Sub-workflows
- Custom status payloads
- External event processing
- Continue-as-new
- App-level retry policies

This PR also adds more testing and fixes some issues that were discovered with internal actors. For example:

- The deactivation path for internal actors was broken, but now is fixed
- Reminders were unable to preserve data payload fidelity due to JSON roundtripping. For example, `[]byte` values were changed to `string` and `int32` values were changed to `float32`. To fix this, reminder serialization is changed to use `encoding/gob` instead of `encoding/json`. This change is only for internal actors. There's no change for external actors.

Lastly, this PR adds some starter code for wfengine black-box testing (see wfengine-test.go). It only tests startup right now, but more tests will be added over time.

External integration testing was done by taking the [Durable Task .NET SDK integration tests](https://github.com/microsoft/durabletask-dotnet/tree/7fb1ba464f42f675f10ddf46d43f1573abde240d/test/DurableTask.Sdk.Tests) and configuring them to talk to the Dapr sidecar. All runtime tests are passing.

## Issue reference

This is one step of many in the workflow engine proposal: https://github.com/dapr/dapr/issues/4576

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Unit tests passing

### DIFF
--- a/pkg/runtime/wfengine/activity.go
+++ b/pkg/runtime/wfengine/activity.go
@@ -16,9 +16,15 @@ package wfengine
 
 import (
 	"context"
-	"errors"
+	"fmt"
+	"strings"
+	"time"
+
+	"github.com/microsoft/durabletask-go/api"
+	"github.com/microsoft/durabletask-go/backend"
 
 	"github.com/dapr/dapr/pkg/actors"
+	invokev1 "github.com/dapr/dapr/pkg/messaging/v1"
 )
 
 type activityActor struct {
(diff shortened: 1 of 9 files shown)


cgillum:
Anyone know why the codecov tool thinks my PR decreases test coverable in `pkg/components/pubsub/pluggable.go`? I haven't changed any code related to that.

yaron2:
> Anyone know why the codecov tool thinks my PR decreases test coverable in `pkg/components/pubsub/pluggable.go`? I haven't changed any code related to that.

It's known to be a hit or miss with that tool

cgillum:
I pushed two additional commits to this PR which add support for continue-as-new and workflow re-creation. With these added features, the runtime compatibility tests with the existing Durable Task SDK are 100% passing.

pkg/runtime/wfengine/activity.go
```diff
@@ -38,14 +44,72 @@
...
+	wi := &backend.ActivityWorkItem{
+		SequenceNumber: int64(taskEvent.EventId),
+		InstanceID:     api.InstanceID(workflowID),
+		NewEvent:       taskEvent,
+		Properties:     make(map[string]interface{}),
+	}
+
+	// Executing activity code is a one-way operation. We must wait for the app code to report its completion, which
+	// will trigger this callback channel.
+	// TODO: Need to come up with a design for timeouts. Some activities may need to run for hours but we also need
```
johnewart, line 81: Iin addition to the potential for blocking indefinitely if the app crashes (I believe, right? Since the channel will never receive anything and the context won't get cancelled), will this also possibly cause double-activations if daprd crashes and this gets re-invoked? (Or do we just assume the app will handle it even if that does happen?)

johnewart (commented):
Looks good to me -- I have a quick question regarding the behavior of InvokeReminder and timeouts or error cases in the workflow activities. Do we need to (or should we) impose a maximum activity lifetime to ensure that we don't get stuck indefinitely? It seems like for really long-running steps, some asynchronous feedback / events might help alleviate some of your concern around timeouts. Thoughts?

pkg/runtime/wfengine/activity.go
cgillum: Yeah, if daprd crashes before the completion callback is received, the activity may be invoked a second time. Since the intent is to provide at-least-once guarantees for activities, we put the burden on the app to deal with this. We could improve it a little by introducing some kind of heartbeat from the app to tell the recovering daprd sidecar that it's still working on that activity, but it's probably not worth the added design complexity, IMO. I don't expect users will run into daprd crashes very often.

cgillum:
Thanks @johnewart for the review.

> Do we need to (or should we) impose a maximum activity lifetime to ensure that we don't get stuck indefinitely?

Yes, I think we need to do this, but I haven't figured out the design yet. Heartbeats from the app might be a good way to detect crashes, etc. For example, if we get no heartbeats in 5 minutes, fail the operation as timed-out. There's also a question if we should impose maximum timeouts to account for the case where the activity is in some kind of deadlock state.

artursouza (approved):

PR #5355, opened 2022-10-13
Refactor and update gRPC connection manager

ItalyPaleAle:
# Description

This PR refactors the gRPC connection manager with two goals:

- Increase re-usage of gRPC connections, so they are not closed as soon as they're not actively being used, and keep them alive and idle for a certain amount of time (currently 3 minutes). This, while keeping into account that each gRPC connection has a limit of 100 active concurrent streams, so if usage of a connection goes over 100, it creates a new channel.
- Allow more flexibility in how the local channel (gRPC connection from Dapr to app) is created. This is to allow for future expansion such as passing through firewalls.
## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests
* [X] Unit tests passing


artursouza:
@ItalyPaleAle Please, resolve conflicts.

pkg/grpc/pool.go
```diff
@@ -0,0 +1,318 @@
...
+
+	"github.com/dapr/kit/ptr"
+)
+
+// Real-time clock (wrapper around time.Time) to allow mocking
+var clock = clocklib.New()
+
+// Maximum number of concurrent streams in a single gRPC connection
+// This is the default value used by gRPC servers and clients
+const grpcMaxConcurrentStreams = 100
```
artursouza, line 32: Is this configurable?

pkg/grpc/pool.go
```diff
@@ -0,0 +1,318 @@
...
+
+// ConnectionPool holds a pool of connections to the same address.
+type ConnectionPool struct {
```
artursouza, line 35: Is there a way for us to use a library to do pooling instead of writing one ourselves?

artursouza (commented):
I need @yaron2 to look at this too. Touches on some critical parts.

pkg/grpc/pool.go
ItalyPaleAle: Sort of, but on the servers (so each sidecar can control the max concurrent streams incoming only). It's something configured at the HTTP/2 level actually: https://httpwg.org/specs/rfc7540.html#rfc.section.5.1.2 In Go, the default is "at least 100" (https://pkg.go.dev/golang.org/x/net/http2#Server)

pkg/grpc/pool.go
ItalyPaleAle: Not sure if there's any optimized for our specific use case.

pkg/actors/actors.go
```diff
@@ -461,7 +464,11 @@ func (a *actorsRuntime) callLocalActor(ctx context.Context, req *invokev1.Invoke
 			reentrancyID = &headerValue.GetValues()[0]
 		} else {
 			reentrancyHeader := fasthttp.RequestHeader{}
-			uuid := uuid.New().String()
```
ItalyPaleAle, line 464: Note that methods like `uuid.New()` can panic in certain cases (e.g. if the kernel doesn't have enough entropy). Hence why the change.

yaron2:
Before this PR can be merged we need to validate the following:

1. No memory leaks (this area of the code is historically prone to these)
2. Continuous service to service invocation calls between app A to app B are not disrupted when a workload certificate renewals occurs in the cluster (mTLS enabled)

In general for refactoring PRs I really like to see tangible benefits and I don't necessarily see that in the current description, so would like to get more feedback here before we can merge this. 

For example:

> This is to allow for future expansion such as passing through firewalls.

We have no open issues about passing through firewalls. So changing critical aspects of the code to cater for non-issues is a red flag for me.

ItalyPaleAle:
Thanks for sharing your thoughts.

> We have no open issues about passing through firewalls. So changing critical aspects of the code to cater for non-issues is a red flag for me.

Yes, you're right that we hadn't opened an issue yet - partly because we hadn't completed the initial investigation until yesterday. You can read the rationale and proposal here: #5392

Although I wrote "passing through firewalls", that's a simplification of the problem that we're trying to address, of having a "server-less" app.

> No memory leaks (this area of the code is historically prone to these)

Could you please share more details on where the leaks happened in the past? I will check to make sure that this implementation isn't doing the same things.

> Continuous service to service invocation calls between app A to app B are not disrupted when a workload certificate renewals occurs in the cluster (mTLS enabled)

This needs to be tested, although I am not too concerned. mTLS certificates are only presented during the handshake of the connection, but a mTLS certificate rotation should have no impact on active connections, since they've already performed the handshake. OTOH, if a server is restarted, all existing connections would be closed, and the sidecar would automatically reconnect (and re-do the TLS handshake). _(There's a lot of work to be done on mTLS renewal at this stage however :) )_

ItalyPaleAle:
> This broke in the past too when changes were done to the code being touched on here.. so this must be tested. Luckily it's easy - Run two gRPC proxy apps (client and server), configure the mTLS cert rotation for 1m and make sure there are no interruptions in traffic after the workload is reloaded. (Run with debug logs and check for the cert renewal message).

I have validated this with the Docker images you shared.

I can see the cert renewals in the logs for daprd, and the client and server communicate without issue. Seems to be fine

pkg/grpc/metadata/metadata.go
```diff
@@ -0,0 +1,57 @@
...
+// FromIncomingContext returns the incoming metadata in ctx if it exists.
+func FromIncomingContext(ctx context.Context) (MD, bool) {
+	md, ok := ctx.Value(ctxKey{}).(MD)
+	if !ok {
+		return nil, false
+	}
+	return md, true
+}
+
+// SetMetadataInContextUnary sets the metadata in the context for an unary gRPC invocation.
```
ItalyPaleAle, line 39: The default metadata package creates a copy of the metadata dictionary every time `FromIncomingContext` is called, additionally processing the dictionary by lowercasing all keys. This changes the code to reuse the same object, saving allocations.

ItalyPaleAle:
@yaron2 I have re-run the perf checks and looks like the perf is pretty much unchanged from the baseline, looking at service invocation tests over HTTP and gRPC

yaron2:
> Link to Action run

Perf looks good, and you've also validated the certificate renewal use case so I think we're good to go here.

ItalyPaleAle:
> Perf looks good, and you've also validated the certificate renewal use case so I think we're good to go here.

Excellent. If you're good, please approve and remove the "do not merge" label :)

yaron2 (approved):

PR #5371, opened 2022-10-17
Full support for `automountServiceAccountToken: false` - use bound service account tokens

ItalyPaleAle:
The injector is changed to create a new projected volume with a service account token with an audience specific for Dapr's Sentry service. Also changes the default value for `--token-audience` in Sentry to be the one the injector uses. Completes the work started in #5359 (which changed Sentry to be able to accept tokens with a different audience)

This effectively means that daprd does not require a service account token with access to the Kubernetes APIs anymore, increasing the security of the overall solution.

> Note: although this PR enables support for setting `automountServiceAccountToken: false`, users will need to add that to their Deployment specs explicitly. The injector will not add it for users, as it's entirely possible the user's own app may require a service account token with access to the K8s APIs. In any case, whether `automountServiceAccountToken` is set to true or false, with this PR daprd will ignore that token regardless.
>
> Note also that setting `automountServiceAccountToken: false` is required for using the Kubernetes secret store component through the Secret Store APIs. (It's not needed for storing component secrets, but it's required for programmatically accessing the secret store)

**BREAKING:** This depends on a Kubernetes feature (projected token volumes) that was not available before 1.21 and became stable in 1.22. This PR will effectively increase the required Kubernetes version to 1.22 (or 1.21 if we accept relying on non-stable APIs). It shouldn't be a big problem given that 1.21 is already EOL and 1.22 will reach EOL next week.

-- EDIT

After discussing with @yaron2, the PR was slightly changed to allow for a smoother upgrade of the control plane. Now:

- if Sentry is started with `--token-audience`, then only that token audience is accepted (no change)
- By default, now Sentry will first try validating tokens with the new token audience (`dapr.io/sentry`). If that fails, it will fall back to passing `nil` as audience, which tells Kubernetes to use the built-in API token audience. A warning message is shown on the logs to tell people that this is only meant as a stopgap measure to ease upgrades
- The behavior above is meant to be temporary and should be removed in Dapr 1.11.

I have validated the behavior above by forcing a Dapr 1.9 sidecar to be deployed with the control plane part of this PR. I am not going to write E2E tests for this, however, as our E2E tests are not equipped for specifying a different sidecar version AND this is meant as just as a stopgap measure to ease upgrades from 1.9 to 1.10.


artursouza:
@ItalyPaleAle Please, resolve conflicts

ItalyPaleAle:
> @ItalyPaleAle Please, resolve conflicts

Done

pkg/injector/sidecar/consts.go
```diff
@@ -15,21 +15,23 @@ limitations under the License.
...
-	SidecarInternalGRPCPortName   = "dapr-internal"
-	SidecarMetricsPortName        = "dapr-metrics"
-	SidecarDebugPortName          = "dapr-debug"
-	SidecarHealthzPath            = "healthz"
-	APIVersionV1                  = "v1.0"
-	UnixDomainSocketVolume        = "dapr-unix-domain-socket" // Name of the Unix domain socket volume.
-	UserContainerDaprHTTPPortName = "DAPR_HTTP_PORT"          // Name of the variable exposed to the app containing the Dapr HTTP port.
-	UserContainerDaprGRPCPortName = "DAPR_GRPC_PORT"          // Name of the variable exposed to the app containing the Dapr gRPC port.
-	ContainersPath                = "/spec/containers"
-	KubernetesMountPath           = "/var/run/secrets/kubernetes.io/serviceaccount" // Mount path for the Kubernetes service account volume.
```
artursouza, line 34: Is this a breaking change?

ItalyPaleAle: No, unless people are doing some really weird stuff.

Here's how it works:

- A new bound token is mounted to `/var/run/secrets/dapr.io/sentry` regardless. That has no permissions to the K8s APIs and it's only used for the runtime to authenticate itself with Sentry. Applications should not have attempted to communicate with Sentry regardless.
- *If and only if* `automountServiceAccountToken: true` (the default value), then a service account token with permissions to talk to the K8s APIs is mounted at the same path as before `/var/run/secrets/kubernetes.io/serviceaccount`. This token is also used by Dapr components that talk to K8s (such as the K8s secret store - but only for accessing secrets at runtime, and not for component secrets - or K8s binding). The Dapr runtime doesn't care about this token at all, whether it's there or not. Some Dapr components may require it. It's up to the developers to decide to add `automountServiceAccountToken: false` to the deployment if they don't need it (by default Dapr does not alter that value in either direction).

ItalyPaleAle: So the constant has been removed because the runtime doesn't need it anymore. But components still know where to find it (https://github.com/dapr/components-contrib/blob/master/internal/authentication/kubernetes/client.go)

artursouza (approved):

Issue #4712, opened 2022-06-06
Use the correct Docker images in Helm charts


shubham1172:
## In what area(s)?

> /area runtime

> /area operator

> /area placement

> /area docs

/area test-and-release

## Describe the feature

For sentry, operator, placement, and injector, we always use the Dapr image.

https://github.com/dapr/dapr/blob/ae49999f48858116830d18e3075f25fe7bf8cac3/charts/dapr/charts/dapr_sentry/templates/dapr_sentry_deployment.yaml#L78-L82

https://github.com/dapr/dapr/blob/ae49999f48858116830d18e3075f25fe7bf8cac3/charts/dapr/charts/dapr_sentry/values.yaml#L4-L7

Although we publish individual docker images for these services, we end up using only the main Dapr image.
https://github.com/dapr/dapr/blob/ae49999f48858116830d18e3075f25fe7bf8cac3/docker/docker.mk#L145-L154

This issue is to update the charts to use the correct images for each service.

## Release Note

RELEASE NOTE: **UPDATE** Helm charts to use specific images for different services.

artursouza:
I like the idea. But it would be a breaking change. In the mean time, can we remove daprd from dapr image?  /cc @yaron2

yaron2:
> This is a great catch. Using smaller images makes startup times smaller too.

Startup times for the control plane are almost irrelevant here. We shoukd do this for other reasons namely security.

yaron2:
> I like the idea. But it would be a breaking change. In the mean time, can we remove daprd from dapr image? /cc @yaron2

I think we can do this safely. Why are you concerned about this being a breaking change if this doesn't cause a change in behavior?

ItalyPaleAle:
I was the one concerned about it being a breaking change. We don't know if someone is using the "dapr" image also for the runtime, perhaps in some self-hosted scenario.

Perhaps a way we can fix this is.... we continue publishing "dapr" with everything inside. We already have separate images for sentry, placement, and daprd. So we just need to create one more image for the operator.

PR #5395, opened 2022-10-21
Update helm charts to use specific control plane images

shubham1172:
# Description

This PR updates
1. docker.mk to start building and publishing the operator and injector images
2. Helm charts for sentry, placement, injector, and operator to use their respective images instead of `dapr`
3. Adds a new option `USE_DEV_IMAGE` to Makefile that can be used to only build, push, and use the "kitchen sink" `dapr` image. This is for facilitating faster dev experiences.

## Issue reference

Please reference the issue this PR will close: #4712

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing


ItalyPaleAle:
Good that this doesn't remove the existing "kitchen sink" image ("dapr") with all binaries.

2 things:

1. I see there's no extra image for the injector, is that on purpose? (I would suggest to keep the injector with the operator, I don't think it needs its own image, since the two can only be used on K8s)
3. Do you think there's any way this could be backwards-incompatible? I can't think of one, but just want to make sure

shubham1172:
> I see there's no extra image for the injector, is that on purpose? (I would suggest to keep the injector with the operator, I don't think it needs its own image, since the two can only be used on K8s) 

Oh no, it was using `dapr` too, now I have moved it to `injector`. If you think we can combine `injector` and `operator` into one image, do you think we call them something else? Or maybe we can continue to keep them separate so it's easy to add/remove images independently as needed.

berndverst (approved):
LGTM

ItalyPaleAle (commented):
Although LGTM, thinking about it, there's one downside of this PR.... Because it builds and pushes more Docker images, it will slow down our development experience a bit. Wonder if maybe there's a way this can be set with an option, to use the "kitchen sink" image?

pravinpushkar (approved):

berndverst (changes requested):
> Although LGTM, thinking about it, there's one downside of this PR.... Because it builds and pushes more Docker images, it will slow down our development experience a bit. Wonder if maybe there's a way this can be set with an option, to use the "kitchen sink" image?

I think a template variable `useSpecificImage` or whatever you want to call this which then render the template to use the specific images or the existing image seems like the way to go.

I support making such a change.

shubham1172:
> I think a template variable useSpecificImage or whatever you want to call this which then render the template to use the specific images or the existing image seems like the way to go.

So all deployments (operator, sentry, placement, injector), have parameters already that can be overridden. I have added support for a new setting, `USE_DEV_IMAGE`, that when set, only builds and pushes the `dapr` image. It also updates the helm charts to use override our deployment images to use it. WDYT @ItalyPaleAle, @berndverst ?

ItalyPaleAle (approved):
LGTM thanks!

mukundansundar (dismissed):
lgtm

pravinpushkar (dismissed):

docker/docker.mk
```diff
@@ -16,10 +16,18 @@
 DOCKER:=docker
 DOCKERFILE_DIR?=./docker
 
+# If set to true, only the `dapr` image will be built and pushed to the registry.
+# This is a "kitchen sink" image that contains all the components.
+# The Helm charts will also be configured to use this image.
+# This is useful for faster development and testing experience.
+USE_DEV_IMAGE?=false
```
artursouza, line 23: Sorry but can this have a name that better relates to the intent?

artursouza (changes requested):

docker/docker.mk
shubham1172: Any suggestions? `USE_DEV_IMAGE` sounded ok because the intent is to reduce development time by building/creating a single image.

ItalyPaleAle: How about `NO_SPLIT_IMAGES` or `SINGLE_IMAGES` (or `KITCHEN_SINK_IMAGE`)?

shubham1172: Good ideas, shall we go ahead with `USE_SINGLE_IMAGE`? @artursouza, thoughts?

daixiang0: If it is true, only build `RELEASE_NAME` image, isn't it?

shubham1172: Updated it!

daixiang0 (approved):

docker/docker.mk
daixiang0: Need @artursouza to merge since it is a requested change.

ItalyPaleAle (approved):

PR #5409, opened 2022-10-24
Changed API logging for HTTP to report abstract method name

### LINKED ISSUE #5408: API logging could log PII with HTTP APIs
## In what area(s)?

/area runtime

## What version of Dapr?

All

## Description

Currently, when API logging is enabled:

- In gRPC, we log the gRPC method name, which is something like `/dapr.proto.runtime.v1.Dapr/GetState`
- In HTTP, we log the entire request path, which is something like `GET /v1.0/state/my-state-store/state-key`

All good with gRPC, as we just log a method name.

But with HTTP, that includes logging something that could be PII. For example, if the user is storing data indexed by a SSN, then the state key would be a SSN, and that would end up in logs.

We should change the format of the HTTP API logs to log just the method name and not the actual path. For example: `GET /v1.0/state/{storeName}/{key}`

## Release Note

RELEASE NOTE: **FIX** Do not log method params in API logs for HTTP APIs, which could contain PII

ItalyPaleAle:
# Description

Changes the API logging so it always logs the "abstract" method name. This was already the case for gRPC, but now applies that to HTTP too, to avoid the possibility of logging PII.

Also, changes the logs to be more structured.

Before:

```json
{"app_id":"orderprocessing","instance":"AleTestVM","level":"info","msg":"HTTP API Called: POST /v1.0/bindings/checkout","scope":"dapr.runtime.http-info","time":"2022-10-24T19:43:36.958061389Z","type":"log","useragent":"node-fetch/1.0 (+https://github.com/bitinn/node-fetch)","ver":"1.9.0"}
```

Note that the full request URL (`/v1.0/bindings/checkout`) is logged, and that can expose PII (e.g. for state store keys).

After:

```json
{"app_id":"orderprocessing","instance":"AleTestVM","level":"info","method":"POST /v1.0/bindings/{name}","msg":"HTTP API Called","scope":"dapr.runtime.http-info","time":"2022-10-24T19:44:23.088517894Z","type":"log","useragent":"node-fetch/1.0 (+https://github.com/bitinn/node-fetch)","ver":"dev"}
```

Note that the path was moved to a "method" key

## Issue reference

Fixes #5408

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests
* [X] Unit tests passing
* [ ] Extended the documentation - TODO: Needs to update https://docs.dapr.io/operations/troubleshooting/api-logs-troubleshooting/


artursouza (commented):
Is there a way to unit test this?

ItalyPaleAle:
> @artursouza  Is there a way to unit test this?

Added unit tests, but please note now the PR depends on https://github.com/dapr/kit/pull/25

artursouza (approved):

Issue #5334, opened 2022-10-11
Pub Sub Polling interval for Kafka


rahulpoddar-fyndna:
The subscriber polls to get the messages from subscription topic in kafka. The interval at which request is made to get the new message is not configurable as of now. Making this parameter configurable will help in tuning the  application for performance needs.

@pravinpushkar

PR #5134, opened 2022-09-06
bugfix(apphealth): app health status no change and return directly.

1046102779:
Signed-off-by: 1046102779 <seachen@tencent.com>

When app health status does not change, initialization or destruction operations are repeated, resulting in resource leakage

PR #5412, opened 2022-10-25
bugfix(apphealthy): app health status no change and return directly.

1046102779:
Signed-off-by: 1046102779 <seachen@tencent.com>

# Description

[dapr/dapr#5334](https://github.com/dapr/dapr/pull/5134) 

My fork `dapr/dapr` repo was dirty in master branch.

## Issue reference

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing

### DIFF
--- a/pkg/runtime/runtime.go
+++ b/pkg/runtime/runtime.go
@@ -190,6 +190,7 @@ type DaprRuntime struct {
 	apiClosers             []io.Closer
 	componentAuthorizers   []ComponentAuthorizer
 	appHealth              *apphealth.AppHealth
+	appHealthStatus        apphealth.AppHealthStatus
 
 	secretsConfiguration map[string]config.SecretsScope
 
@@ -571,7 +572,13 @@ func (a *DaprRuntime) initPluggableComponents() {
 
 // Sets the status of the app to healthy or un-healthy
 // Callback for apphealth when the detected status changed
-func (a *DaprRuntime) appHealthChanged(status uint8) {
+func (a *DaprRuntime) appHealthChanged(status apphealth.AppHealthStatus) {
+	if a.appHealthStatus == status {
+		// Status has not changed and return directly.
+		return
+	}
+	a.appHealthStatus = status
+
 	switch status {
 	case apphealth.AppStatusHealthy:
 		// Start subscribing to topics and reading from input bindings
--- a/pkg/apphealth/health.go
+++ b/pkg/apphealth/health.go
@@ -23,9 +23,11 @@ import (
 	"github.com/dapr/kit/logger"
 )
 
+type AppHealthStatus uint8
+
 const (
-	AppStatusUnhealthy uint8 = 0
-	AppStatusHealthy   uint8 = 1
+	AppStatusUnhealthy AppHealthStatus = 0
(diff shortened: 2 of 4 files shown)


1046102779:
cc @daixiang0 @yaron2 @artursouza

pkg/runtime/runtime.go
```diff
@@ -571,7 +572,13 @@ func (a *DaprRuntime) initPluggableComponents() {
 
 // Sets the status of the app to healthy or un-healthy
 // Callback for apphealth when the detected status changed
-func (a *DaprRuntime) appHealthChanged(status uint8) {
+func (a *DaprRuntime) appHealthChanged(status apphealth.AppHealthStatus) {
+	if a.appHealthStatus == status {
```
ItalyPaleAle, line 576: Can you use an atomic flag here for a.appHealthStatus? Otherwise this could cause race conditions

ItalyPaleAle (changes requested):
LGTM except for a possible race condition

pkg/runtime/runtime.go
1046102779: https://github.com/dapr/dapr/blob/master/pkg/apphealth/health.go#L195

It will not appear here, the current detection interval is the shortest 1s, and it is executed serially, so the callback API  changes the status, and there will be cause race conditions.

ItalyPaleAle: Although the interval is 1s, they are invoked as background goroutines (https://github.com/dapr/dapr/blob/master/pkg/apphealth/health.go#L205) so depending on how busy the application is, they may be delayed in their execution

1046102779: Does it mean that the goroutine is heavily blocked? If the goroutine blocks for more than the interval period, there will indeed be a race condition.

But there is another point here: if the later request arrives first, the first request arrives later, and then the Callback  is executed, is the state itself wrong? Shouldn't that goroutine be removed? become blocking execution?

ItalyPaleAle: > But there is another point here: if the later request (goroutine) arrives first, the first request (goroutine) arrives later, and then the Callback is executed, is the state itself wrong (maybe)? Shouldn't that goroutine be removed? become blocking execution?

That's a good point.

Actually, I do first wonder why is the callback invoked twice. Rather than adding a check to the runtime, shouldn't there be a check in the apphealth package to make sure that the change is never sent to the runtime if it's not actually changed in the first place?

1046102779:
This PR is wrong and has no bugs. The runtime object does not need to set the health status. The health status is stored in apphealth package. Only when the health status changes, the runtime will be notified to subscribe or unsubscribe through the Callback API.  @ItalyPaleAle 

And this PR should be closed.

ItalyPaleAle:
Thanks! Please open another PR, I'll close this for now

1046102779:
> Thanks! Please open another PR, I'll close this for now

Is it to open another PR to keep partial codes of this PR?   https://github.com/dapr/dapr/pull/5412/files#diff-6b98a646df2db6212a6e133614d6ed25804de975bf8e5217d47aec99fd7be5b3R254

PR #5413, opened 2022-10-25
Wait for app to be healthy before invoking /configuration

ItalyPaleAle:
Fixes a bug reported by @artursouza . When AppHealth is enabled, waits for the app to be in a healthy state before invoking the `/configuration` callback


pkg/runtime/runtime.go
```diff
@@ -502,22 +507,49 @@ func (a *DaprRuntime) initRuntime(opts *runtimeOpts) error {
...
-	a.loadAppConfiguration()
+	a.daprGRPCAPI.SetAppChannel(a.appChannel)
 
 	a.initDirectMessaging(a.nameResolver)
 
 	a.daprHTTPAPI.SetDirectMessaging(a.directMessaging)
-	grpcAPI.SetDirectMessaging(a.directMessaging)
+	a.daprGRPCAPI.SetDirectMessaging(a.directMessaging)
+
+	a.appHealthReady = func() {
```
artursouza, line 517: Why do you need to override the existing implementation here? It makes this a little bit hard to read. Wouldn't this need to simply be a local variable?

ItalyPaleAle: When AppHealth is enabled, appHealthReady needs to be invoked once, and only once, as a callback by appHealthChanged: https://github.com/dapr/dapr/pull/5413/files/baf8748c1f6c4ba508a25da0778173868fa56eb6#diff-e56f9bb947923c6f8bb4c995bf4b9469a0ed117276223d16b177dcb04399a4bcR588

PR #5438, opened 2022-10-31
Updated depguard linter config

ItalyPaleAle:
1. Add older (pre-v4) versions of `github.com/cenkalti/backoff` to the denylist
   - Often, gopls (VS Code) automatically adds the "v1" import when typing `backoff`. Because of that, we had a panic in the runtime that forced us to release a hotfix in the 1.8.x branch.
3. Add `github.com/agrea/ptr` to the denylist (use `github.com/dapr/kit/ptr` instead)
4. Fixed error messages for denied deps


mukundansundar:
Can you add https://github.com/pkg/errors to the deny list as well?

artursouza (approved):

Issue #5443, opened 2022-11-02
OpenTelemetry tracing does not work on 1.9.0


zhenqxuMSFT:
## In what area(s)?

/area runtime

## What version of Dapr?

1.9.0

## Expected Behavior

Tracing data is collected by otel collector

## Actual Behavior

No tracing data is available. No error log in both otel collector and daprd sidecar

## Steps to Reproduce the Problem
1. Prepare two pods with dapr enabled
2. Invoke service to service call between pods
3. Check the data in destination configured in otel collector

Configuration:
```
  tracing:
    samplingRate: "0"
    zipkin:
      endpointAddress: http://otel-collector.ns.svc.cluster.local:9411/api/v2/spans
```

I switched the daprd version to 1.8.3 and everything works.

berndverst:
@zhenqxuMSFT samplingRate "0" means tracing disabled. Do you still experience this issue if you set samplingRate to the default value "1" ?

zhenqxuMSFT:
Hi @berndverst, oops I pasted a wrong configuration. Here's the configuration on the problematic cluster, the samplingRate is "1" and it's not working. And we just changed the image version of daprd to 1.8.3 and tracing started to work. The configuration itself is unchanged.
```
  tracing:
    samplingRate: "1"
    zipkin:
      endpointAddress: http://otel-collector.ns.svc.cluster.local:9411/api/v2/spans
```

amolenk:
Running into the same problem with the following config and 1.9.2:
```yaml
apiVersion: dapr.io/v1alpha1
kind: Configuration
metadata:
  name: daprConfig
spec:
  tracing:
    samplingRate: "1"
    zipkin:
      endpointAddress: "http://localhost:9411/api/v2/spans"
```

Service invocations using HTTP get logged and are visible in Zipkin. Service invocations using gRPC do not show up. When I downgrade to 1.8.6, both HTTP and gRPC calls show up.

ItalyPaleAle:
Fixed in release-1.9

PR #5454, opened 2022-11-03
Fix tracing sampler

johnewart:
Addresses #5443 in that when a traceparent header is passed into Dapr and the sampling flag is zero, we will use the sampling rate instead of not sampling at all.

# Description

Added a DaprTraceSampler that has the logic we use for handling sampling decisions - this deviates somewhat from the previous implementation in that if the parent header exists and is supposed to be sampled then we will follow suit, otherwise we will make a determination based on sampling rate.

## Issue reference

Please reference the issue this PR will close: #5443

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Unit tests passing

### DIFF
--- a/pkg/diagnostics/tracing_sampler.go
+++ b/pkg/diagnostics/tracing_sampler.go
@@ -1,0 +1,44 @@
+package diagnostics
+
+import (
+	"fmt"
+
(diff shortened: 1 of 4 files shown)


ItalyPaleAle:
Closing this as it's been merged into release-1.9 and that PR had some improvements

PR #5455, opened 2022-11-03
Hotfix: Fix tracing sampler behavior

johnewart:
# Dapr 1.9.3
 
### Problem
 
With tracing enabled, Dapr operators can choose the sampling rate using the `spec.tracing.sampling` option in the Dapr configuration. In Dapr 1.9.0-1.9.2, when a request coming to the Dapr runtime contains a `traceparent` header, the decision on whether to sample the request or not is solely based on the "sampling bit" set in the header's value.
 
This behavior was introduced with Dapr 1.9.0 during the transition to the new tracing framework based on the OpenTelemetry (OTEL) standard. Although this behavior is compliant with the W3C specs for distributed tracing, it caused Dapr to omit sending traces to the telemetry collector (e.g. Zipkin, Azure Monitor, etc) in many instances.
 
### Impact
 
The issue impacts users on Dapr 1.9.0-1.9.2 who have enabled collection of traces.
 
Based on reports from Dapr users, developers building apps with ASP.NET Core seem to be particularly impacted, as the .NET Core framework can automatically include a `traceparent` header in every request made to the Dapr sidecar which has the "sampling bit" set to `0` (disabled).
 
### Root cause
 
Older versions of Dapr (before 1.9.0) ignored the "sampling bit" in the `traceparent` header when making decisions on whether to sample a request. That behavior changed in Dapr 1.9.0, where the decision made by the caller with the `traceparent` header determines the Dapr runtime's sampling choice too.
 
### Solution
 
We have patched the way Dapr makes decisions on whether to sample a request and submit the trace to the telemetry collector (e.g. Zipkin, Azure Monitor, etc). The new behavior consists of:
 
- If the `traceparent` header has the "sampling bit" set to `1` (enabled), the request is always sampled.
- If the "sampling bit" is `0` (disabled), Dapr decides on whether to sample the request based on its own internal policies. For example, with `spec.tracing.sampling` set to `1`, all requests are traced; instead, a value of `0` generates no trace. Numbers in-between 0 and 1 cause Dapr to sample only a fraction of requests.
 
This behavior, while more lax, remains compliant with the W3C specs for distributed tracing.

### DIFF
--- a/pkg/diagnostics/tracing_test.go
+++ b/pkg/diagnostics/tracing_test.go
@@ -15,8 +15,6 @@ package diagnostics
 
 import (
 	"context"
-	crand "crypto/rand"
-	"encoding/binary"
 	"fmt"
 	"math/rand"
 	"sync"
@@ -163,7 +161,8 @@ func TestStartInternalCallbackSpan(t *testing.T) {
 	t.Run("traceparent is provided with sampling flag = 0 and sampling is enabled (but not P=1.00)", func(t *testing.T) {
 		numTraces := 100000
 		sampledCount := runTraces(t, "test_trace", numTraces, "0.01", 0)
-		require.Greater(t, sampledCount, 0, "Expected to sample at least one span, but did not sample any!")
+		// We use a fixed seed for the RNG so we can use an exact number here
+		require.Equal(t, sampledCount, 1012, "Expected to sample 1012 traces but only sampled %d", sampledCount)
 		require.Less(t, sampledCount, numTraces, "Expected to sample fewer than the total number of traces, but sampled all of them!")
 	})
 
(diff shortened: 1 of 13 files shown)


pkg/diagnostics/tracing_test.go
```diff
@@ -221,3 +285,58 @@ func (o *otelFakeSpanProcessor) Shutdown(ctx context.Context) error {
 func (o *otelFakeSpanProcessor) ForceFlush(ctx context.Context) error {
 	return nil
 }
+
+// This was taken from the otel testing to generate IDs
+// origin: go.opentelemetry.io/otel/sdk@v1.11.1/trace/id_generator.go
```
ItalyPaleAle, line 290: Please add a copyright notice here for that file

pkg/diagnostics/tracing_test.go
```diff
@@ -221,3 +285,58 @@ func (o *otelFakeSpanProcessor) Shutdown(ctx context.Context) error {
...
+	sid := trace.SpanID{}
+	_, _ = gen.randSource.Read(sid[:])
+	return tid, sid
+}
+
+func defaultIDGenerator() IDGenerator {
+	gen := &randomIDGenerator{}
+	var rngSeed int64
+	_ = binary.Read(crand.Reader, binary.LittleEndian, &rngSeed)
+	gen.randSource = rand.New(rand.NewSource(rngSeed))
```
ItalyPaleAle, line 340: If you use a fixed seed here (like `1`) then it's deterministic

ItalyPaleAle (changes requested):

ItalyPaleAle (approved):

artursouza (approved):

yaron2 (approved):

PR #5462, opened 2022-11-04
Added apiLogging options to config CRD

### LINKED ISSUE #5457: Add option to hide health checks from API logging
## In what area(s)?

/area runtime

## Describe the feature

With API logging enabled, any call to the HTTP or gRPC APIs gets logged, and that includes health checks (the `/v1.0/healthz` endpoint). In certain environments, that can add a significant amount of noise, which may not be desirable: for example, on K8s I've seen apps logging 10-20 health checks per minute from the control plane. Aside from taking up unnecessary space, those logs can even be dangerous if they cause too much noise that hides important information.

Propose to add a new boolean CLI flag `--no-health-check-api-logging` (and equivalent annotation `dapr.io/no-health-check-api-logging`) that when used in conjunction with `--enable-api-logging` prevents calls to the health check APIs to be printed. The CLI flag would have no effect if API logging is not enabled.

## Release Note

RELEASE NOTE: **ADD** CLI option `--no-health-check-api-logging` to omit health checks from API logging

ItalyPaleAle:
New options:

- `spec.logging.apiLogging.enabled`: Sets the default value for `--enable-api-logging` (and corresponding annotation). This is the value that's used if the flag is not explicitly set with either a value or true (implied) or false.
- `spec.logging.apiLogging.omitHealthChecks`: if true, health checks are not included in API logging

Fixes #5457

TODO:

- [x] Docs issue: https://github.com/dapr/docs/issues/2938
- [ ] Update Dapr CLI to not automatically add `--enable-api-logging` if unspecified


pkg/apis/configuration/v1alpha1/types.go
```diff
@@ -209,6 +211,25 @@ type ConfigurationList struct {
...
+	// +optional
+	APILogging APILoggingSpec `json:"apiLogging" yaml:"apiLogging"`
+}
+
+// APILoggingSpec defines the configuration for API logging.
+type APILoggingSpec struct {
+	// Default value for enabling API logging. Sidecars can always override this by setting `--enable-api-logging` to true or false explicitly.
+	// The default value is false.
+	// +optional
+	DefaultEnabled bool `json:"defaultEnabled" yaml:"defaultEnabled"`
```
artursouza, line 226: Just Enabled is fine.

ItalyPaleAle: Done

artursouza (approved):

PR #5468, opened 2022-11-07
Update build-tools compilation and minor changes

shubham1172:

# Description

While [running e2e tests](https://github.com/dapr/dapr/blob/master/tests/docs/running-e2e-test.md), we set up `GOOS` and `GOARCH` locally to match linux/amd64. This interferes with build tools compilation that just runs a `go run .`. I am on darwin/arm64 and I see the following:

```
fork/exec /var/folders/g0/3c007k8s5577w7csd7j2trcw0000gn/T/go-build3790960422/b001/exe/build-tools: exec format error
```

This change prefixes the `go run` command with the local `GOOS` and `GOARCH`, so that build-tools always builds in context of the local system.

Also a few minor changes 💄

## Issue reference

None

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing


Makefile
```diff
@@ -83,17 +83,7 @@ ifeq ($(TARGET_OS_LOCAL),windows)
...
-
-ifeq ($(TARGET_OS_LOCAL),windows)
-	BUILD_TOOLS_BIN ?= build-tools.exe
-	BUILD_TOOLS ?= ./.build-tools/$(BUILD_TOOLS_BIN)
-	RUN_BUILD_TOOLS ?= cd .build-tools; go.exe run .
-else
-	BUILD_TOOLS_BIN ?= build-tools
-	BUILD_TOOLS ?= ./.build-tools/$(BUILD_TOOLS_BIN)
-	RUN_BUILD_TOOLS ?= cd .build-tools; go run .
+	RUN_BUILD_TOOLS ?= cd .build-tools; GOOS=$(TARGET_OS_LOCAL) GOARCH=$(TARGET_ARCH_LOCAL) go run .
```
ItalyPaleAle, line 86: Is there a way to still allow using the pre-compiled ones? It's faster than recompiling with go run every time.

ItalyPaleAle (changes requested):

ItalyPaleAle:
Thanks for catching this. My work environment was primarily Linux/amd64 and didn't think about this. There's a lot of work we need to do really allow users on ARM64 to run our E2E tests, but this is a good start.

Makefile
shubham1172: Weren't the two blocks exactly identical? I thought that was unintentional, am I missing something? 🤔

ItalyPaleAle: Uh, looks like the pre-compiled binaries weren’t actually used. oops :)

ItalyPaleAle (dismissed):

Makefile
shubham1172: Let's try to add it separately. Maybe using an environment variable that points to the binary? If missing, always use `go run .`. What do you think?

ItalyPaleAle: That could be nice! But if we haven0t used it so far, maybe not necessary either.

shubham1172: Yup, we will add it when it's needed then, thanks!

.build-tools/cmd/zz-e2e-perf.go
```diff
@@ -411,7 +411,7 @@ func (c *cmdE2EPerf) getIncludes() []string {
 func hashFilesInDir(basePath string, ignores *gitignore.GitIgnore) ([]string, error) {
 	files := []string{}
 
-	err := filepath.WalkDir(basePath, func(path string, d fs.DirEntry, err error) error {
+	err := filepath.WalkDir(basePath, func(path string, d fs.DirEntry, _err error) error {
```
daixiang0, line 414: Just remove the value name?

daixiang0, line 414: Just use `_`.

.build-tools/README.md
```diff
@@ -8,7 +8,7 @@ The CLI is written in Go and based on the [cobra](https://github.com/spf13/cobra
 
 You have two ways to run the CLI:
 
-1. Within the `build-tools` directly, you can run the CLI with `go run .` directly. For example `go run . help` shows the help page.
+1. Within the `build-tools` directly, you can run the CLI with `go run .` directly. For example `go run . help` shows the help page. Make sure that `GOOS` and `GOARCH` are set to the correct values for your system.
```
daixiang0, line 11: In fact, those values are set autosomally, so a better fix is that remove GOOS set in Makefile. Is it possible?

shubham1172: This is for manually running only. Makefile sets `GOOS` for e2e tests today, which is required. If you are relying on Makefile, it should handle GOOS as well as build-tools compilations automatically.

daixiang0: But the GOOS in Makefile only works during Make, unless users set it by hand, or it is always same as system.

daixiang0 (approved):

.build-tools/README.md
shubham1172: I miswrote, Makefile does not set `GOOS`, it looks for it, and uses `TARGET_OS_LOCAL` (`shell uname`) if it's not present.

artursouza (approved):

PR #5476, opened 2022-11-07
Added state storage for workflow execution

cgillum:
# Description

This PR enables state storage for the embedded workflow engine. Previously, state was only held in memory, which meant that workflows could only be used in one-box setups.

Details of the state-storage design can be found in the README.md updates. TL/DR version is this:

- State is stored using state storage actor APIs, so any state store supporting actors can be used
- Each workflow instance has its state spread out across multiple state store keys
- The design allows for arbitrarily large workflow histories

This PR also improves the existing workflow tests by having them run with and without caching. The default behavior is to run with caching, which effectively means we never read state from the state store. However, with caching _disabled_, the tests heavily exercise the state storage read path.

The version of `durabletask-go` was also updated to get a faster version of workflow status polling, using exponential backoff instead of hard-coded 1 second intervals. This significantly speeds up the workflow execution unit tests.

## Issue reference

This is a continuation of the implementation for the workflow engine proposed here: https://github.com/dapr/dapr/issues/4576

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Unit tests passing
* [ ] ~~Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]~~
* [ ] ~~Specification has been updated / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]~~
* [ ] ~~Provided sample for the feature / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]~~

### DIFF
--- a/pkg/runtime/wfengine/README.md
+++ b/pkg/runtime/wfengine/README.md
@@ -24,7 +24,7 @@ DEBUG=1 make build
 After building, the following command can be run from the project root to test the code:
 
 ```bash
-./dist/linux_amd64/debug/daprd --app-id foo --dapr-grpc-port 4001 --placement-host-address :50005 ~/.dapr/components/
+./dist/linux_amd64/debug/daprd --app-id foo --dapr-grpc-port 4001 --placement-host-address :50005 --components-path ~/.dapr/components/
 ```
 * The gRPC port is set to `4001` since that's what the Durable Task test clients default to.
 * This assumes a placement service running locally on port `50005` (the default).
@@ -64,6 +64,36 @@ In the future, instructions will be added for how to tests in a more automated w
 
 The Dapr Workflow engine introduced a new concept of *internal actors*. These are actors that are registered and implemented directly in Daprd with no host application dependency. Just like regular actors, they have turn-based concurrency, support reminders, and are scaled out using the placement service. Internal actors also leverage the configured state store for actors. The workflow engine uses these actors as the core runtime primitives for workflows.
 
-Each workflow instance corresponds to a single `dapr.wfengine.workflow` actor instance. The ID of the workflow instance is the same as the internal actor ID. The internal actor is responsible for triggering workflow execution and for storing workflow state. The actual workflow logic lives outside the Dapr sidecar in a host application. The host application uses a new gRPC endpoint on the daprd gRPC API server to send and receive workflow-specific commands to/from the actor-based workflow engine. The workflow app doesn't need to take on any actor dependencies, nor is it aware that actors are involved in the execution of the workflows. Actors are purely an implementation detail.
+Each workflow instance corresponds to a single `dapr.internal.wfengine.workflow` actor instance. The ID of the workflow instance is the same as the internal actor ID. The internal actor is responsible for triggering workflow execution and for storing workflow state. The actual workflow logic lives outside the Dapr sidecar in a host application. The host application uses a new gRPC endpoint on the daprd gRPC API server to send and receive workflow-specific commands to/from the actor-based workflow engine. The workflow app doesn't need to take on any actor dependencies, nor is it aware that actors are involved in the execution of the workflows. Actors are purely an implementation detail.
 
-TODO: More information on activity actors, the details of state storage, and information about how reminders are used to ensure reliability.
+### State storage
+
+Each workflow actor saves its state using the following keys:
+
+* `metadata`: Contains meta information about the workflow as a JSON blob. Includes information such as the length of the inbox, the length of the history, and a UUID representing the workflow generation (for cases where the instance ID gets reused). The length information is used to determine which keys need to be read or written to when loading or saving workflow state updates.
+* `inbox-NNNNNN`: Multiple keys containing an ordered list of workflow inbox events. Each key holds the data for a single event. The inbox is effectively a FIFO queue of events that the workflow needs to process, with items removed from the earlier indeces and added to the end indices.
+* `history-NNNNNN`: Multiple keys containing an ordered list of history events. Each key holds the data for a single event. History events are only added and never removed, except in the case of "continue as new", where all history events are purged.
+* `customStatus`: Contains a user-defined workflow status value.
+
+The `inbox-NNNNNN` and `history-NNNNNN` key schemes are used to enable arbitrarily large amounts of data. These schemes are also designed for efficient updates. An alternate design would be to store the workflow history as a blob in a single key. However, this would limit the maximum size of the history and would make updates more expensive, since the full history would need to be serialized instead of just inserting incremental additions (the history is an append-only log of events).
+
+The tradeoff with this key scheme design is that loading workflow state becomes more expensive since it's spread out across multiple keys. This is mitigated by the fact that actor state can be cached in memory, removing the need for any reads while the actors are active. However, it could be a problem if workflow histories get large and if actors get moved around or activated frequently.
+
+Below is an example of what keys would be used to store the state of a simple workflow execution with ID '797f67f0c10846f592d0ac82dea1f248', as shown using `redis-cli`.
+
+```
+127.0.0.1:6379> keys *797f67f0c10846f592d0ac82dea1f248*
+1) "myapp||dapr.internal.wfengine.workflow||797f67f0c10846f592d0ac82dea1f248||history-000002"
+2) "myapp||dapr.internal.wfengine.workflow||797f67f0c10846f592d0ac82dea1f248||customStatus"
+3) "myapp||dapr.internal.wfengine.workflow||797f67f0c10846f592d0ac82dea1f248||metadata"
+4) "myapp||dapr.internal.wfengine.workflow||797f67f0c10846f592d0ac82dea1f248||history-000003"
+5) "myapp||dapr.internal.wfengine.workflow||797f67f0c10846f592d0ac82dea1f248||history-000005"
+6) "myapp||dapr.internal.wfengine.workflow||797f67f0c10846f592d0ac82dea1f248||history-000001"
+7) "myapp||dapr.internal.wfengine.workflow||797f67f0c10846f592d0ac82dea1f248||history-000000"
+8) "myapp||dapr.internal.wfengine.workflow||797f67f0c10846f592d0ac82dea1f248||history-000004"
+9) "myapp||dapr.internal.wfengine.workflow||797f67f0c10846f592d0ac82dea1f248||inbox-000000"
+```
+
+### Resiliency
+
+TODO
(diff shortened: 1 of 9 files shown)


pkg/runtime/wfengine/workflowstate.go
```diff
@@ -0,0 +1,229 @@
...
+	s.historyRemovedCount += len(s.History)
+	s.History = nil
+	s.CustomStatus = ""
+	s.Generation = uuid.New()
+}
+
+func (s *workflowState) ApplyRuntimeStateChanges(runtimeState *backend.OrchestrationRuntimeState) {
+	if runtimeState.ContinuedAsNew() {
+		s.historyRemovedCount += len(s.History)
+		s.History = nil
```
daixiang0, line 73: Need to cover this case.

pkg/runtime/wfengine/workflow.go
```diff
@@ -133,7 +142,7 @@ func (wf *workflowActor) createWorkflowInstance(ctx context.Context, actorID str
 	if exists {
 		runtimeState := getRuntimeState(actorID, state)
 		if runtimeState.IsCompleted() {
-			state = &workflowState{}
+			state.Reset()
```
daixiang0, line 145: Please cover reset case.

pkg/runtime/wfengine/workflow.go
```diff
@@ -52,7 +39,29 @@ const (
...
+}
+
+type durableTimer struct {
+	Bytes      []byte
+	Generation uuid.UUID
+}
+
+func init() {
+	// gob is used to serialize internal actor reminder data
+	gob.Register(durableTimer{})
```
daixiang0, line 57: Could you share more details about it? I do not get the benefit of it.

cgillum: Please see my PR description [here](https://github.com/dapr/dapr/pull/5344), where I talk about the reasons for introducing the gob dependency. If you're specifically asking about why I need to call `gob.Register(...)`, it's because reminder payloads are defined as `interface{}` (see [create_reminder_request.go](https://github.com/dapr/dapr/blob/a738f7f200036c200236fa83856ce5b8d8c79ed5/pkg/actors/create_reminder_request.go#L21)), so any structs that get serialized by `gob` have to be declared explicitly. See [here](https://stackoverflow.com/questions/32676898/whats-the-purpose-of-gob-register-method) for more info on this.

pkg/runtime/wfengine/workflowstate.go
cgillum: Done. I also fixed a bug that was exposed by the newly added test, so thanks for calling this out.

pkg/runtime/wfengine/workflow.go
cgillum: Done. I added the `TestRecreateCompletedWorkflow` to cover this, and also fixed a bug that this new test exposed.

pkg/runtime/wfengine/README.md
```diff
@@ -64,6 +64,36 @@ In the future, instructions will be added for how to tests in a more automated w
...
 
-TODO: More information on activity actors, the details of state storage, and information about how reminders are used to ensure reliability.
\ No newline at end of file
+### State storage
+
+Each workflow actor saves its state using the following keys:
+
+* `metadata`: Contains meta information about the workflow as a JSON blob. Includes information such as the length of the inbox, the length of the history, and a UUID representing the workflow generation (for cases where the instance ID gets reused). The length information is used to determine which keys need to be read or written to when loading or saving workflow state updates.
+* `inbox-NNNNNN`: Multiple keys containing an ordered list of workflow inbox events. Each key holds the data for a single event. The inbox is effectively a FIFO queue of events that the workflow needs to process, with items removed from the earlier indeces and added to the end indices.
```
johnewart, line 74: `indeces` -> `indices`

pkg/runtime/wfengine/README.md
```diff
@@ -64,6 +64,36 @@ In the future, instructions will be added for how to tests in a more automated w
...
+
+The tradeoff with this key scheme design is that loading workflow state becomes more expensive since it's spread out across multiple keys. This is mitigated by the fact that actor state can be cached in memory, removing the need for any reads while the actors are active. However, it could be a problem if workflow histories get large and if actors get moved around or activated frequently.
+
+Below is an example of what keys would be used to store the state of a simple workflow execution with ID '797f67f0c10846f592d0ac82dea1f248', as shown using `redis-cli`.
+
+```
+127.0.0.1:6379> keys *797f67f0c10846f592d0ac82dea1f248*
+1) "myapp||dapr.internal.wfengine.workflow||797f67f0c10846f592d0ac82dea1f248||history-000002"
```
johnewart, line 86: My only concern with this is that there is a limitation on the number of keys that a single redis instance can deal with - this limit may far exceed the practical ceiling we would see but if we don't purge the data we could end up hitting it. 

Thoughts?

pkg/runtime/wfengine/wfengine_test.go
```diff
@@ -368,22 +397,96 @@ func TestConcurrentActivityExecution(t *testing.T) {
...
+		t.Run(opt(engine), func(t *testing.T) {
+			id, err := client.ScheduleNewOrchestration(ctx, "ActivityFanOut")
+			if assert.NoError(t, err) {
+				metadata, err := client.WaitForOrchestrationCompletion(ctx, id)
+				if assert.NoError(t, err) {
+					assert.True(t, metadata.IsComplete())
+					assert.Equal(t, `["9","8","7","6","5","4","3","2","1","0"]`, metadata.SerializedOutput)
+
+					// Because all the activities run in parallel, they should complete very quickly
+					assert.Less(t, metadata.LastUpdatedAt.Sub(metadata.CreatedAt), 3*time.Second)
```
johnewart, line 411: Is this critical? I think checks like this are best left to performance tests to avoid flaky tests that are validating correctness. Thoughts?

pkg/runtime/wfengine/wfengine_test.go
```diff
@@ -368,22 +397,96 @@ func TestConcurrentActivityExecution(t *testing.T) {
...
+			var metadata *api.OrchestrationMetadata
+			id, err := client.ScheduleNewOrchestration(ctx, "EchoWorkflow", api.WithInput("echo!"))
+			if assert.NoError(t, err) {
+				if metadata, err = client.WaitForOrchestrationCompletion(ctx, id); assert.NoError(t, err) {
+					assert.True(t, metadata.IsComplete())
+					assert.Equal(t, `"echo!"`, metadata.SerializedOutput)
+				}
+			}
+
+			// Second workflow, using the same ID as the first but a different input
```
johnewart, line 477: I don't think we can test this here, but should we have tests (like in e2e scenarios) that check for behavior if the first instance is still running and you try to do this?

pkg/runtime/wfengine/workflow.go
```diff
@@ -75,7 +84,7 @@ func (wf *workflowActor) InvokeMethod(ctx context.Context, actorID string, metho
 	case CreateWorkflowInstanceMethod:
 		err = wf.createWorkflowInstance(ctx, actorID, request)
 	case GetWorkflowMetadataMethod:
-		result, err = wf.getWorkflowMetadata(actorID)
+		result, err = wf.getWorkflowMetadata(ctx, actorID)
```
johnewart, line 87: Yay, context! :D

pkg/runtime/wfengine/workflowstate.go
```diff
@@ -0,0 +1,232 @@
...
+	res, err = actorRuntime.GetState(ctx, &req)
+	loadedRecords++
+	if err != nil {
+		return nil, fmt.Errorf("failed to load workflow custom status key '%s': %w", req.Key, err)
+	}
+	if err = json.Unmarshal(res.Data, &state.CustomStatus); err != nil {
+		return nil, fmt.Errorf("failed to unmarshal JSON from inbox state key entry: %w", err)
+	}
+
+	wfLogger.Infof("%s: loaded %d state records in %v", actorID, loadedRecords, time.Since(loadStartTime))
```
johnewart, line 226: I like that we are tracking these and would love if we had metrics emitted for workflow operations; I don't think this PR is the place because it's somewhat tangential to this (I would like to make metric instruments easier to create and work with in general beforehand anyway) and can be done separately, but I think it's good to continue at least collecting and logging the important data so we can emit them. :+1:

pkg/runtime/wfengine/wfengine_test.go
```diff
@@ -368,22 +397,96 @@ func TestConcurrentActivityExecution(t *testing.T) {
...
+	ctx := context.Background()
+	client, engine := startEngine(ctx, r)
+	for _, opt := range GetTestOptions() {
+		t.Run(opt(engine), func(t *testing.T) {
+			id, err := client.ScheduleNewOrchestration(ctx, "ContinueAsNewTest", api.WithInput(0))
+			if assert.NoError(t, err) {
+				metadata, err := client.WaitForOrchestrationCompletion(ctx, id)
+				if assert.NoError(t, err) {
+					assert.True(t, metadata.IsComplete())
+					assert.Equal(t, `10`, metadata.SerializedOutput)
```
johnewart, line 445: Do we want to check the history size? I think there are other tests that validate the history gets reset on continuation so I don't know if we need to do it here as well, only if you think it's a useful signal.

johnewart (approved):
Nothing major, looks like the `Reset()` condition has test coverage so my feedback is mostly just few open-ended questions, minor typos and that we should probably remove any hard timing assertions to avoid flaky tests. I think performance tests will be better for ensuring a consistent baseline for overhead, etc.

pkg/runtime/wfengine/README.md
cgillum: I'm thinking we cap the maximum history size to 10K and the maximum inbox size to 1K. This is something we probably want to do anyways to limit the impact of runaway workflows. Also, good point about data purging. I'm going to add a work item to revisit both of these in a separate data governance PR.

pkg/runtime/wfengine/wfengine_test.go
cgillum: I agree we want to avoid tests that involve timing (since they run the risk of becoming flaky) and I debated with myself on whether to include this specific validation. In my personal deliberation, it met the "correctness" validation bar because the ability to execute activities in parallel is foundational, IMO. I would like to keep it if there's no major objection.

pkg/runtime/wfengine/wfengine_test.go
cgillum: I think we can test this here. I can create a workflow that sleeps forever, try to start it twice, and then verify that the second start fails.

pkg/runtime/wfengine/wfengine_test.go
cgillum: I'm a bit torn on this because it's a good thing to do but it's an implementation detail we don't have full control over. The actual history size is controlled by the `durabletask-go` dependency. The Dapr code is mostly trying to ensure that we faithfully represent this history in actor state. It's not straightforward to know what exactly to assert in terms of history size.

PR #5486, opened 2022-11-09
Renames wasm middleware file as it clashes with GOARCH=wasm

codefromthecrypt:
While testing, I noticed the wasm middleware wasn't loading. This was due to the file naming convention, as it is the same as a valid GOARCH.

cc @daixiang0 @ItalyPaleAle @yaron2


artursouza:
What about replacing "wasm" by "webassembly"

codefromthecrypt:
good idea @artursouza. It is now "webassembly"

codefromthecrypt:
here's a sample that works once this is merged, cc @anuraaga https://github.com/dapr/samples/pull/122

ItalyPaleAle (approved):

PR #5493, opened 2022-11-10
Update to latest contrib commit

yaron2:
Incorporate fix from https://github.com/dapr/components-contrib/pull/2268.


yaron2:
The interface between `daprd` and contrib has been broken:

```
pkg/components/state/pluggable.go:457:10: cannot use fromConnector(l, pluggable.NewGRPCConnectorWithDialer(dialer, newStateStoreClient)) (value of type *grpcStateStore) as type "github.com/dapr/components-contrib/state".Store in return statement:
	*grpcStateStore does not implement "github.com/dapr/components-contrib/state".Store (missing GetComponentMetadata method)
```

@mcandeia can you advise?

ItalyPaleAle:
That’s due to @berndverst’s change on metadata

berndverst:
Just add this please: https://github.com/berndverst/dapr/commit/159383e4fac0e505f9d092ffa88e04aead8c13b5

In `components/state/pluggable.go`

```go
func (ss *grpcStateStore) GetComponentMetadata() map[string]string {
	return map[string]string{}
}
```

I was planning to make the PR to dapr/dapr, but I have a couple more such breaking changes for other component types.

berndverst:
@yaron2 I'm adding `GetComponentMetadata() map[string]string` to every component type, which isn't a problem generally.. but it turns out the way the pluggable components code works, this must be updated in `dapr/dapr` for each pluggable component type.

berndverst:
I'll open a separate PR to add this everywhere! @yaron2

berndverst:
Please merge this PR first @yaron2  @artursouza 
https://github.com/dapr/dapr/pull/5494

berndverst:
Merge master into this branch now and it should work!

yaron2:
> Merge master into this branch now and it should work!

Thanks :)

ItalyPaleAle (approved):

yaron2:
Still failing @berndverst

berndverst:
Oh there are a lot more places that need this method added apparently. But only in tests - what I added lets you build Dapr. Will take a look.

berndverst:
@yaron2 please close this PR.

I kept your commits in my new PR which also fixes the error experienced here and also pins an even newer contrib.

https://github.com/dapr/dapr/pull/5496

PR #5498, opened 2022-11-11
optimize bulkpub resp processing

### LINKED ISSUE #5598: sending failed entries and err does not work as expected in gRPC Bulk Publish
## In what area(s)?

/area runtime

## What version of Dapr?

edge

## Expected Behavior

When there are partial failures on a gRPC BulkPublishEventAlpha1() call, the response field must be populated with the failed entries as discussed in dapr/components-contrib#2260.

## Actual Behavior

When err field is returned, the response if by default set to nil on the client side. And only the error status is used. 

For testing this, hard coded in bulk publish API that `entry_id: "1"` will always fail e.g.: Hardcoded in line https://github.com/dapr/dapr/blob/master/pkg/runtime/pubsub/default_bulkpub.go#L82
```go
// TODO REMOVE HARDCODE
	if entry.EntryId == "1" {
		return &contribPubsub.BulkPublishResponseEntry{
			EntryId: entry.EntryId,
			Error:   ErrBulkPublishFailure,
		}
	}
```

Then using post man posted a gRPC request such as : 
```
{
    "entries": [
        {
            "content_type": "application/text",
            "event": "e30=",
            "entry_id": "1"
        },
        {
            "content_type": "application/text",
            "event": "e31=",
            "entry_id": "2"
        }
    ],
    "metadata": {},
    "pubsub_name": "pubsub",
    "topic": "testingtopic"
}
```
The response got back if err is there is 

image

## The issue 

Reading the issue linked here `https://github.com/grpc/grpc/issues/12824`, it seems that this is the expected behavior. That is the 
```
Server-implementations of gRPC MUST drop the Message if it is returned with a non-OK response.
Client side SHOULD drop any response messages received with a non-OK response.
```

The reason stated here is that if there is an error returned on a gRPC call, there is no guarantee that the response object is complete so it must be discarded and only the error status not-OK be considered. 

How this is handled when there are partial failures in GetBulkState is that, even if there is error for individual items, then only the itemized error is populated in the response and the overall `err` for the gRPC call is still nil. See https://github.com/dapr/dapr/blob/master/pkg/grpc/api.go#L802-L817. 

```go
          if bgr.bulkGet {
// There is overall error, so send empty response with the err
		if err != nil {
			return bulkResp, err
		}
		for i := 0; i < len(bgr.responses); i++ {
			item := &runtimev1pb.BulkStateItem{
				Key:      stateLoader.GetOriginalStateKey(bgr.responses[i].Key),
				Data:     bgr.responses[i].Data,
				Etag:     stringValueOrEmpty(bgr.responses[i].ETag),
				Metadata: bgr.responses[i].Metadata,
				Error:    bgr.responses[i].Error,
			}
			bulkResp.Items = append(bulkResp.Items, item)
		}
// If there is partial error as seen above, still only send nil overall error and a valid response with the details populated correctly
		return bulkResp, nil
	}
```
Similar processing is done when state store does not implement bulkGet API [also](https://github.com/dapr/dapr/blob/master/pkg/grpc/api.go#L823-L864).

Should this be the one that is followed for BulkPublishEventAlpha1 request as well? 

**That is if there are partial failures, the err is still nil but the BulkPublishResponse is populated with the failed entries alone.** 
Only on the cases of overall error being such as the following will an empty response with overall error be returned. 

```go

nerr := status.Errorf(codes.Internal, messages.ErrPubsubPublishMessage, topic, pubsubName, err.Error())
if errors.As(err, &runtimePubsub.NotAllowedError{}) {
	nerr = status.Errorf(codes.PermissionDenied, err.Error())
}

if errors.As(err, &runtimePubsub.NotFoundError{}) {
	nerr = status.Errorf(codes.NotFound, err.Error())
}
// if pubsub not allowed or not found return empty response with the corresponding error status
return &BulkPublishResponse{}, nerr
```
@yaron2 @artursouza @ItalyPaleAle 

cc @dapr/maintainers-dapr thoughts? 

## Release Note

RELEASE NOTE:

mukundansundar:
# Description

With change https://github.com/dapr/components-contrib/pull/2270, the bulk publish response from contrib only contains event entries that have failed. Refactor code to take that into account. 

~~This needs to be merged only after https://github.com/dapr/components-contrib/pull/2270 is merged.~~ ready to merge

One quick point. With the conversation in #5598 , now this PR changes the conditions when err is returned from gRPC API of BulkPublish. 
Only on scenarios where request is malformed, or there is total failure due to permissions or not found etc, does the API return an err. 
For other partial or total failures due to network issues or from component, no error is returned and the response contains the IDs that have failed. 

In HTTP though, the status code is 500 even on partial failures. 

Please review @dapr/maintainers-dapr @dapr/approvers-dapr 

## Issue reference

related to dapr/components-contrib#2260

closes #5598 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Unit tests passing


dapr/proto/runtime/v1/dapr.proto
```diff
@@ -332,24 +332,16 @@ message BulkPublishRequestEntry {
 
 // BulkPublishResponse is the message returned from a BulkPublishEvent call
 message BulkPublishResponse {
-  // The publish statuses for different events in the BulkPublishEvent call
-  repeated BulkPublishResponseEntry statuses = 1;
+  // The entries for different events that failed publish in the BulkPublishEvent call
+  repeated BulkPublishResponseFailedEntry failedEntries = 1;
```
ItalyPaleAle, line 336: Just confirming this has not been released yet, right? So we're not modifying a released proto?

ItalyPaleAle (approved):

ItalyPaleAle (approved):

dapr/proto/runtime/v1/dapr.proto
mukundansundar: only nightly builds will be brone ... and BulkPubsub as a feature was not released in 1.9 ... 

See https://github.com/dapr/dapr/blob/release-1.9/dapr/proto/runtime/v1/dapr.proto

ItalyPaleAle: Perfect.

DeepanshuA (approved):
LGTM

ItalyPaleAle (approved):

go.mod
```diff
@@ -415,6 +418,7 @@ replace (
 // Don't commit with this uncommented!
 //
 // replace github.com/dapr/components-contrib => ../components-contrib
+
```
ItalyPaleAle, line 421: Important: please don't add  a newline here

go.mod
```diff
@@ -72,6 +72,12 @@ require (
 	sigs.k8s.io/controller-runtime v0.14.1
 )
 
+require (
+	github.com/didip/tollbooth/v7 v7.0.1 // indirect
+	github.com/go-pkgz/expirable-cache v0.1.0 // indirect
+	github.com/go-redis/redis/v9 v9.0.0-rc.2 // indirect
+)
+
```
ItalyPaleAle, line 80: Can you merge this block into the next?

ItalyPaleAle (changes requested):

ItalyPaleAle (approved):

DeepanshuA (approved):

pkg/grpc/api.go
```diff
@@ -636,7 +626,17 @@ func (a *api) BulkPublishEventAlpha1(ctx context.Context, in *runtimev1pb.BulkPu
 		return &bulkRes, nerr
 	}
 
+	bulkRes.FailedEntries = make([]*runtimev1pb.BulkPublishResponseFailedEntry, 0, len(res.FailedEntries))
+	for _, r := range res.FailedEntries {
+		resEntry := runtimev1pb.BulkPublishResponseFailedEntry{}
```
yaron2, line 631: nit: suggest 
```runtimev1pb.BulkPublishResponseFailedEntry{ EntryId: r.EntryId }```

File: pkg/http/api.go [hunk 5]
```diff
@@ -2426,26 +2426,23 @@ func (a *api) onBulkPublish(reqCtx *fasthttp.RequestCtx) {
...
+	if len(res.FailedEntries) != 0 {
+		eventsPublished -= int64(len(res.FailedEntries))
 	}
+
 	diag.DefaultComponentMonitoring.BulkPubsubEgressEvent(context.Background(), pubsubName, topic, err == nil, eventsPublished, elapsed)
+
 	if err != nil {
+		bulkRes.FailedEntries = make([]BulkPublishResponseFailedEntry, 0, len(res.FailedEntries))
+		for _, r := range res.FailedEntries {
+			resEntry := BulkPublishResponseFailedEntry{}
```
yaron2, line 2439: Same as above

mukundansundar: will followup in #5646

mukundansundar: done

pkg/grpc/api.go
mukundansundar: done

PR #5505, opened 2022-11-12
Workflow engine resiliency

cgillum:
# Description

This PR contains the code necessary to ensure that workflows and workflow activities can run to completion in the face of infrastructure failures. Reminders are the basic tool to ensure that the workflows are always making progress.

A few additional things included:
- Configurable operation execution timeouts (this allowed me to remove some TODOs from the code)
- De-duplication logic for workflow activity actors
- Added a small feature to internal actors to immediately delete reminders after they execute
- New tests
- Updated technical documentation in the wfengine README.md file

A quick note on WF engine tests: on my local machine, all tests under `wfengine` (technically the `wfengine_test` package) are completing in about 16 seconds with 75.3% code coverage.

## Issue reference

This is ongoing work that covers the workflow engine component of the Dapr Workflow proposal in https://github.com/dapr/dapr/issues/4576.

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [ ] Unit tests passing


pkg/runtime/wfengine/activity.go
```diff
@@ -124,6 +210,68 @@ func (*activityActor) DeactivateActor(ctx context.Context, actorID string) error
 	return nil // nothing to do - this actor maintains no state
 }
 
+func (a *activityActor) loadActivityState(ctx context.Context, actorID string, generation uuid.UUID) (*activityState, error) {
+	// See if the state for this actor is already cached in memory.
+	result, ok := a.statesCache.Load(actorID)
+	if ok {
+		cachedState := result.(*activityState)
```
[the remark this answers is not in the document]
johnewart: Also, per a separate conversation, should we include the generation in the cache key?

pkg/runtime/wfengine/activity_test.go
```diff
@@ -0,0 +1,57 @@
...
+	"github.com/microsoft/durabletask-go/task"
+	"github.com/stretchr/testify/assert"
+
+	"github.com/dapr/dapr/pkg/actors"
+	"github.com/dapr/dapr/pkg/runtime/wfengine"
+)
+
+// TestDedupeActivityInvocation executes the activity actor invocation path for the same (simulated) workflow
+// twice to verify that the second call fails with a duplication invocation error.
+func TestDedupeActivityInvocation(t *testing.T) {
```
johnewart, line 31: Is there any way we can test multiple generations of a workflow happening concurrently? The caching mechanism above (and possibly the storage) seems like it could have race conditions if this were to happen. (Or, if we intend for this to never happen, can we guarantee it doesn't happen?)

pkg/runtime/wfengine/activity.go
```diff
@@ -16,26 +16,55 @@ package wfengine
...
+	statesCache      sync.Map
+	cachingDisabled  bool
+	defaultTimeout   time.Duration
+	reminderInterval time.Duration
+}
+
+// ActivityRequest represents a request by a worklow to invoke an activity.
+type ActivityRequest struct {
+	HistoryEvent []byte
+	Generation   uuid.UUID
```
johnewart, line 49: Should these be somehow related to one another like monotonically increasing numbers or timestamps? Would I ever need to know that one is older than another -- maybe for the purposes of autocancelling an older generation if I have seen a newer generation, or viewing them in time-order as a human?

pkg/actors/actors.go
```diff
@@ -978,8 +980,13 @@ func (a *actorsRuntime) startReminder(reminder *Reminder, stopChannel chan bool)
 				break L
 			}
 			if err = a.executeReminder(reminder); err != nil {
-				log.Errorf("error execution of reminder %q for actor type %s with id %s: %v",
-					reminder.Name, reminder.ActorType, reminder.ActorID, err)
+				if err == ErrReminderCanceled {
+					// The handler is explicitly canceling the timer
+					break L
```
johnewart, line 985: Should we log this event?

johnewart (commented):
Just some questions about the concurrency / generation / cache model; talking offline you mentioned adding some additional tests. It would also be good if we can prove to ourselves that this won't break if there are multiple generations concurrently running (or prevent that case from happening if we don't want it to occur) given that reminders are inherently asynchronous

pkg/runtime/wfengine/activity.go
```diff
@@ -93,7 +175,11 @@ loop:
 		case <-ctx.Done():
 			return ctx.Err()
 		case <-time.After(10 * time.Minute):
```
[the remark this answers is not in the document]
cgillum: I don't see any benefit in making this configurable, TBH. It's just logging. We'll almost certainly remove this if/when we get around to doing proper heartbeating for long-running activities (captured by the TODO a few lines above here).

pkg/actors/actors.go
cgillum: Yeah, that's probably a good idea. I'm guessing `Info` is the right level? I feel like reminders are way too chatty right now, but `Info` makes sense for consistency, I guess.

pkg/runtime/wfengine/activity.go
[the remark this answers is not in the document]
cgillum: Good point about caching a copy vs. a reference. Let me look into that.

I don't want to include the generation in the cache key, though. My intent is that these cache keys be strictly actor IDs. TBD on whether to include generations in cache keys, but I think I want to avoid this as well.

pkg/runtime/wfengine/activity.go
cgillum: Yeah, that might actually be useful to simplify debugging. I chose UUID because the Durable Task Framework uses UUIDs for this same concept, but I don't think it's strictly necessary.

pkg/runtime/wfengine/activity_test.go
cgillum: It's impossible to have two generations of the same workflow or activity to run concurrently since all generations share the same actor ID (and actors already guarantee one-at-a-time execution).

pkg/runtime/wfengine/activity.go
```diff
@@ -50,15 +72,68 @@ func (a *activityActor) SetActorRuntime(actorsRuntime actors.Actors) {
...
+	// TODO: Purge actor state based on some data retention policy
+
+	// We delete the reminder on success and on non-recoverable errors.
+	return actors.ErrReminderCanceled
+}
+
+func (a *activityActor) executeActivity(ctx context.Context, actorID string, name string, eventPayload []byte) error {
+	taskEvent, err := backend.UnmarshalHistoryEvent(eventPayload)
 	if err != nil {
 		return err
```
RyanLettieri, line 138: Do we want the fmt.Errorf here as well?

cgillum: I'm intentionally not doing `fmt.Errorf` here because I think the error message generated by the `UnmarshalHistoryEvent` is sufficiently descriptive. It's also an error message that we theoretically should never run into, so I don't think it's important to add additional context.

pkg/runtime/wfengine/activity.go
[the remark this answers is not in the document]
cgillum: I made the change to store/fetch a copy of the cache data instead of a reference. However, I want to leave the generation out of the cache key. There should technically never be two generations of the same activity running twice (the orchestration ensures this) but even if there was, the guarantees of the actor subsystem ensures that we don't have to worry about them running in parallel.

cgillum:
@artursouza looks like I was impacted by another flaky test: https://github.com/dapr/dapr/actions/runs/3494248271/jobs/5849862441
```
=== FAIL: pkg/actors TestPerActorTimeout (3.00s)
    actors_test.go:531: 
        	Error Trace:	/Users/runner/work/dapr/dapr/pkg/actors/actors_test.go:531
        	Error:      	Should be false
        	Test:       	TestPerActorTimeout

DONE 1972 tests, 1 failure in 237.272s
```

It passes locally and in the other jobs.

ItalyPaleAle:
Yes our tests are flaky. Don’t worry too much about them :)

pkg/actors/actors.go
```diff
@@ -978,8 +980,14 @@ func (a *actorsRuntime) startReminder(reminder *Reminder, stopChannel chan bool)
 				break L
 			}
 			if err = a.executeReminder(reminder); err != nil {
-				log.Errorf("error execution of reminder %q for actor type %s with id %s: %v",
-					reminder.Name, reminder.ActorType, reminder.ActorID, err)
+				if err == ErrReminderCanceled {
```
ItalyPaleAle, line 983: Please use `errors.Is(err, ErrReminderCanceled)` which allows wrapping too

pkg/runtime/wfengine/activity.go
```diff
@@ -50,15 +72,68 @@ func (a *activityActor) SetActorRuntime(actorsRuntime actors.Actors) {
...
+	// The actual execution is triggered by a reminder
+	err := a.createReliableReminder(ctx, actorID, "run-activity", ar.Generation)
 	return nil, err
 }
 
 // InvokeReminder implements actors.InternalActor and executes the activity logic.
-// TODO: Need to review how errors for reminders are handled.
 func (a *activityActor) InvokeReminder(ctx context.Context, actorID string, reminderName string, data any, dueTime string, period string) error {
-	taskEvent, err := backend.UnmarshalHistoryEvent(data.([]byte))
+	wfLogger.Debugf("invoking reminder '%s' on activity actor '%s'", reminderName, actorID)
```
ItalyPaleAle, line 103: Is there any chance this may include PII? If so, we shouldn't log it

pkg/runtime/wfengine/activity.go
```diff
@@ -50,15 +72,68 @@ func (a *activityActor) SetActorRuntime(actorsRuntime actors.Actors) {
...
+
+	generation := data.(uint64)
+	state, _ := a.loadActivityState(ctx, actorID, generation)
+	// TODO: On error, reply with a failure - this requires support from durabletask-go to produce TaskFailure results
+
+	timeoutCtx, cancelTimeout := context.WithTimeout(ctx, a.defaultTimeout)
+	defer cancelTimeout()
+
+	if err := a.executeActivity(timeoutCtx, actorID, reminderName, state.EventPayload); err != nil {
+		if err == timeoutCtx.Err() {
```
ItalyPaleAle, line 113: Because `timeoutCtx` is a children of `ctx`, its error could also be `context.Canceled`.

I suggest changing this to be explicitly looking for `errors.Is(err, context.DeadlineExceeded)`

ItalyPaleAle: (and maybe have a separate log for `context.Canceled`?)

pkg/runtime/wfengine/wfengine_test.go
```diff
@@ -517,6 +517,104 @@ func TestRecreateRunningWorkflowFails(t *testing.T) {
...
+		}
+		return actualCallCount, nil
+	})
+
+	ctx := context.Background()
+	client, engine := startEngine(ctx, r)
+
+	// Set a really short timeout to override the default activity timeout (1 hour at the time of writing)
+	// so that we can exercise the timeout handling codepath in a short period of time.
+	engine.SetActivityTimeout(1 * time.Second)
```
ItalyPaleAle, line 593: Is there a way we can use a mock clock in unit tests? We've seen many times that using the real wall clock leads to flakiness. If you can add support for mocking the time object, you can use the `clock` library then (it's already in use by some packages)

pkg/runtime/wfengine/workflow.go
```diff
@@ -98,13 +112,26 @@ func (wf *workflowActor) InvokeReminder(ctx context.Context, actorID string, rem
 	wfLogger.Debugf("invoking reminder '%s' on workflow actor '%s'", reminderName, actorID)
 
 	// Workflow executions should never take longer than a few seconds at the most
-	timeoutCtx, cancelTimeout := context.WithTimeout(ctx, 30*time.Second) // TODO: Configurable
+	timeoutCtx, cancelTimeout := context.WithTimeout(ctx, wf.defaultTimeout)
 	defer cancelTimeout()
 	if err := wf.runWorkflow(timeoutCtx, actorID, reminderName, data); err != nil {
-		return err
+		if err == timeoutCtx.Err() {
```
ItalyPaleAle, line 118: Here too note the error should be checked against `errors.Is(err, context.Canceled)` and `errors.Is(err, context.DeadlineExceeded)`

pkg/runtime/wfengine/activity.go
cgillum: No, we control the name of the reminder. However, we only have a single reminder name, so there's probably no benefit in logging it. I can remove that part of this statement.

pkg/runtime/wfengine/activity.go
cgillum: I changed this to use `errors.Is(err, context.DeadlineExceeded)` as you suggested. However, I didn't add a separate log for `context.Canceled` since we never attempt to explicitly cancel the context.

pkg/runtime/wfengine/workflow.go
cgillum: Updated to use `errors.Is(err, context.DeadlineExceeded)` but not `errors.Is(err, context.Canceled)` since we never explicitly cancel.

pkg/runtime/wfengine/wfengine_test.go
cgillum: I'm not familiar with the `clock` library or with mocking time more generally. I'll need to spend some time investigating how to do this.

pkg/runtime/wfengine/activity.go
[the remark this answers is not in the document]
johnewart: Cool, thanks! :+1:

pkg/runtime/wfengine/activity.go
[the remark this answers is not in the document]
johnewart: Seems reasonable

johnewart (approved):
LGTM, the only suggestion would be to look at mocking time like @ItalyPaleAle mentioned just to reduce potential flakiness in tests but I think that can be done in a follow-on PR if he's okay with it

ItalyPaleAle:
I'm ok with that but now wondering if you could wait for #5538 to be merged?

cgillum:
> I'm ok with that but now wondering if you could wait for #5538 to be merged?

I'm okay with waiting on #5538. Looks like it's signed off on already, which is encouraging. :)

ItalyPaleAle:
It's merged! Thanks for your patience. You may need to adjust the code in this PR now

cgillum:
@artursouza it looks like the `feature/workflows` branch got out of sync with `master` again due to merge conflicts: https://github.com/dapr/dapr/actions/runs/3441220741/jobs/5740529341. Can you help us with getting this feature branch back into good shape?

artursouza:
Yes. Fixing it now.

cgillum:
Thanks @artursouza! I've rebased my PR on the latest changes in `feature/workflows`. Assuming the tests pass, this PR should be good to merge.

pkg/actors/internal_actor.go
```diff
@@ -174,3 +183,11 @@ func DecodeInternalActorData(data []byte, e any) error {
 	}
 	return nil
 }
+
+// DecodeInternalActorReminderData decodes internal actor reminder data payloads and stores the result in e.
+func DecodeInternalActorReminderData(data []byte, e any) error {
```
johnewart, line 188: nit: It's not clear why this method exists - do we intend to add extra data or functionality to it? If not why not just `json.Unmarshal` where this gets called?

johnewart (approved):
I don't have any concerns with the recent changes - always good to handle canceled contexts 😄 

LGTM 👍

Issue #5471, opened 2022-11-07
Proposal: Support k6 load testing


mcandeia:
## In what area(s)?

> /area runtime

> /area operator

> /area placement

> /area docs

/area test-and-release
### What is this proposing?
Currently, Dapr supports load testing (for tracking performance issues) by using a tool named [fortio](https://github.com/fortio/fortio) - a forked version of it, which is the Istio's load testing tool graduated to be its own project. Fortio runs at a specified query per second (qps) and records an histogram of execution time and calculates percentiles (e.g. p99 i.e. the response time such as 99% of the requests take less than that number (in seconds, SI unit)). It can run for a set duration, for a fixed number of calls, or until interrupted (at a constant target QPS, or max speed/load per connection/thread). This proposal suggests expanding the possibilities of load testing tools, initially, with k6.

### Why do we need it?

Supporting more types of load testing tools allows us to create from simpler to complex scenarios that reproduce different load shapes for different purposes. Fortio is a perfect fit when your scenario requires only one endpoint and you have to measure performance without complex scenarios, i.e literally for QPS.
Currently, to be able to run a load test against an app, a fortio container should be deployed together with your apps, and, later on, can be used to run the specified query. When using our test framework, fortio is encapsulated under our perf-tester application. This application is a simple http application that receives the fortio payload and executes the load testing by running the fortio command line programmatically. To exemplify that, let's look at some of our performance tests, 

At the test main, we have to deploy the perf-tester app alongside the target test app

```go
var tr *runner.TestRunner
 
func TestMain(m *testing.M) {
   utils.SetupLogs("actor_activation")
 
   testApps := []kube.AppDescription{
       {
           AppName:           serviceApplicationName,
           DaprEnabled:       true,
           ImageName:         "perf-actorjava",
           Replicas:          1,
           IngressEnabled:    true,
           MetricsEnabled:    true,
           AppPort:           3000,
           DaprCPULimit:      "4.0",
           DaprCPURequest:    "0.1",
           DaprMemoryLimit:   "512Mi",
           DaprMemoryRequest: "250Mi",
           AppCPULimit:       "4.0",
           AppCPURequest:     "0.1",
           AppMemoryLimit:    "800Mi",
           AppMemoryRequest:  "2500Mi",
           Labels: map[string]string{
               "daprtest": serviceApplicationName,
           },
       },
       {
           AppName:           clientApplicationName,
           DaprEnabled:       true,
           ImageName:         "perf-tester",
[...]

PR #5514, opened 2022-11-14
Support k6 loadtesting + Actor Activation Stress test for Double Activation

mcandeia:
# Description

This PR adds support to k6 loadtesting by including a new `LoadTest` method to our test platform. For k6 specifically it is implemented only for kubernetes.

Summary is not experted yet but k6 supports its own internal checks and those are being used to verify if test was successfully executed.

> In draft until a new https://github.com/grafana/k6-operator is released including my fix, meanwhile I'm using my private k6 operator image.

## The double activation test

The double activation test is basically a stress test under a single actor ID that uses a single a redis instance for locking purposes.

So the idea behind is that for each single call to the actor's `Lock` [method](https://github.com/dapr/dapr/pull/5514/files#diff-271e6fe59970199442db695883c2dc697825cc66909dea31657d6e7a95fd2980R52) will 

1) Acquire lock for 1s
2) Release lock and, if error, wait for 1s to make sure the lock was released before returning

Notice that if any disruption occurs on redis or on the app side between these two steps so a false-positive double activation could be caught.

The idea is that if we allow for double activation it means that the `Lock` could be executed before the lock is released.

## Issue reference

Please reference the issue this PR will close: #5471 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Unit tests passing


mcandeia:
@sarsharma @akhilac1 would love to receive feedback from you folks

See [this](https://github.com/dapr/dapr/pull/5514/files#diff-f444ea727e20de932ac66bfd89ee17f18f622cdaf8cc10b3e2a5ab1804e1954aR74) as an example

Considerations:

1. The test summary is being printed on console and the test result is based on k6 job status - I'm having conversations regard exporting these metrics to somewhere (perhaps grafana cloud would work?)
2. There's no way to use any JS bundle yet (which means that you can't have node_modules or something like this, so you're limited to xk6 extensions)

ItalyPaleAle (approved):

artursouza (approved):

Issue #5440, opened 2022-11-01
Make resiliency policies stable


msfussell:
Resiliency Policies.

## Describe the feature
Get resiliency features to stable status. Resiliency was released in the v1.4 milestone. 
Link any other issues that need to be resolved in 1.10
This is the docs issue https://github.com/dapr/docs/issues/2873

ItalyPaleAle:
We'll need to remove the feature flag and all code paths where we bifurcate depending on whether resiliency is enabled or not.

@halspang  can you think of anything else we need to do to make this stable?

ItalyPaleAle:
Perf-wise, enabling resiliency doesn't seem to have a significant negative impact:

|              | service_invocation_http | service_invocation_grpc | state_get_grpc | state_get_http | pubsub_publish_grpc | actor_timer_java | actor_activatio_java | actor_reminder_registration |                                                                      |                      |
|--------------|-------------------------|-------------------------|----------------|----------------|---------------------|------------------|----------------------|-----------------------------|----------------------------------------------------------------------|----------------------|
| Baseline     | 1.05                    | 1.6                     | 0.85           | 0.91           | 0.89                | 8.92             | 2.96                 | 228.88                      | [Link](https://github.com/dapr/dapr/actions/runs/3396739440/jobs/5648225701) |                      |
| Resiliency 1 | 1.07                    | 1.68                    | 0.83           | 0.47           | 0.87                | 8.79             | 3.23                 | 227.74                      | [Link](https://github.com/dapr/dapr/actions/runs/3397608725/jobs/5649985451) |                      |
| Resiliency 1 | 1.81                    | 1.98                    | 0.08           | 1.45           | 0.87                | 9.43             | 3.65                 | 228.92                      | [Link](https://github.com/dapr/dapr/actions/runs/3397608629/jobs/5650025543) | (0.08 is NOT a typo) |
| Resiliency 3 | 1.07                    | 1.76                    | 0.87           | 0.92           | 0.05                | 9                | 3.31                 | 231                         | [Link](https://github.com/dapr/dapr/actions/runs/3397609004/jobs/5649963954) | (0.05 is NOT a typo) |

https://github.com/dapr/dapr/pull/5451

yaron2:
The following issues need to be resolved before making resiliency stable:

* https://github.com/dapr/dapr/issues/5511
* https://github.com/dapr/dapr/issues/5321

msfussell:
The following CLI issue need to be completed for resiliency 
[...]

PR #5523, opened 2022-11-16
Force resiliency to be enabled

ItalyPaleAle:
This PR sets the `Resiliency` to always be enabled in Dapr, regardless of the value of the feature flag in the config file.

We are aiming to have the resiliency feature stable in Dapr 1.10 (per #5440), and this change essentially forces resiliency to be enabled in the nightly builds (and all future CI runs), allowing us to do even more validation that we've done so far.

This approach (forcing the feature flag to be always on) is a lower-risk alternative to actually removing all code paths that don't use resiliency, so if we discover some critical issue, we can easily revert this change.


ItalyPaleAle:
@artursouza @yaron2 

Ready for review now. Please consider merging this sooner than later so we can run more tests with resiliency forcibly on in the nightly builds.

If we're comfortable with how things work, I'll go ahead and actually remove all now-unused code paths.

halspang (approved):

daixiang0:
Also need to update doc.

ItalyPaleAle:
> Also need to update doc.

Will do that when the change is definitive. This PR is meant to be a quick "patch" that enables resiliency with the smallest number of changes, so we can test it in the nightly builds before we actually remove all non-resiliency code.

PR #5531, opened 2022-11-17
Resiliency: fix `maxRetries: 0` ignored

### LINKED ISSUE #5530: Resiliency: setting `maxRetries: 0` is ignored, retries indefinitely instead
## In what area(s)?

/area runtime

## What version of Dapr?

> 1.9.3
> master

## Description

With Resiliency enabled, create a policy with `maxRetries: 0`, for example:

```yaml
apiVersion: dapr.io/v1alpha1
kind: Resiliency
metadata:
  name: pubsub-resiliency
spec:
  policies:
    retries:
      noRetry:
        policy: constant
        maxRetries: 0
  targets:
    components:
      mypubsub:
        inbound:
          retry: noRetry
```

(This is used by a pubsub, but the error happens regardless of what the target is)

Per documentation, `maxRetries: 0` should make it so Dapr never retries in case of errors.

However, the value of `0` is internally converted to `-1` and it causes Dapr to retry indefinitely.

The reason for the error is due to this line:

https://github.com/ItalyPaleAle/dapr/blob/0aab2ca532688eab79443e8501dcbddfa4d112c4/pkg/resiliency/resiliency.go#L393

Before that line, the policy is in an object of type `resiliencyV1alpha.Retry`. That line converts it to a `map[string]any` and does that by first converting it to JSON and back. However, in the `resiliencyV1alpha.Retry` struct, all fields have `omitempty`, so when they're converted to JSON, they're removed entirely.

ItalyPaleAle:
Fixes #5530

Includes unit test to catch regressions


pkg/apis/resiliency/v1alpha1/types.go
```diff
@@ -46,7 +46,7 @@ type Retry struct {
 	Policy      string `json:"policy,omitempty" yaml:"policy,omitempty"`
 	Duration    string `json:"duration,omitempty" yaml:"duration,omitempty"`
 	MaxInterval string `json:"maxInterval,omitempty" yaml:"maxInterval,omitempty"`
-	MaxRetries  int    `json:"maxRetries,omitempty" yaml:"maxRetries,omitempty"`
+	MaxRetries  *int   `json:"maxRetries,omitempty" yaml:"maxRetries,omitempty"`
```
ItalyPaleAle, line 49: By making this a pointer, when it's serialized to JSON, a value of `0` is preserved.

See #5530 for an explanation of why this was causing errors.

mukundansundar (approved):
lgtm

artursouza (approved):

WebscienceXYZ:
YOUR A FUCKING SHIT BAG FAGGOT

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

PR #5538, opened 2022-11-17
Fix multiple race conditions when resiliency is enabled

ItalyPaleAle:
# Description

This PR fixes a bunch of race conditions that appear when resiliency is enabled

## Issue reference

Part of #5440


pkg/actors/actors.go
```diff
@@ -394,25 +397,30 @@ func (a *actorsRuntime) callRemoteActorWithRetry(
...
-			if retriesExhaustedPath {
-				return nil, errors.Errorf("failed to invoke target %s after %v retries", targetAddress, 3)
-			}
+			if err != nil {
+				var permanent *backoff.PermanentError
+				if !errors.As(err, &permanent) {
+					// If the error wasn't a permanent one, it means we just exhausted the retries
+					var retries int64
+					if pd.RetryPolicy != nil {
+						retries = pd.RetryPolicy.MaxRetries
```
johnewart, line 414: Isn't `pd` nil here? (line 381)

pkg/actors/actors.go
```diff
@@ -620,38 +645,37 @@ func (a *actorsRuntime) TransactionalStateOperation(ctx context.Context, req *Tr
...
 		policy = noOp.BuiltInPolicy(ctx, resiliency.BuiltInActorNotFoundRetries)
 	}
-	policy(func(ctx context.Context) error {
-		_, exists = a.actorsTable.Load(key)
-
+	err := policy(func(ctx context.Context) error {
+		_, exists := a.actorsTable.Load(key)
 		if !exists {
-			return fmt.Errorf("Actor")
+			return errors.New("Actor")
```
johnewart, line 674: Should this error message provide more information?

ItalyPaleAle: I agree but that was there before (I just changed the method name) and I have no idea what  the original author meant this to be

pkg/actors/actors.go
ItalyPaleAle: Yes it is. Thanks, I'll fix it

pkg/actors/actors.go
johnewart: Sure, but since it's being touched something like `fmt("actor %s %s not found in actor table", req.ActorType, req.ActorID)` might be handy :-)

ItalyPaleAle: TBH I don't think that matters. If you look below, we just check to see if `err != nil`. The actual error message is ignored. 
So I just changed it to a message that's much better ;)

ItalyPaleAle:
This PR now fixes every single invocation of `policy` making sure that there are no race conditions with the handler trying to modify a variable in the parent scope.

However, after chatting with @johnewart , we wonder if it may make sense to simply refactor the `policy` method so it returns values, rather than having it rely on side effects. This could be implemented with [promises](https://github.com/chebyrash/promise).

mukundansundar (commented):
@ItalyPaleAle should policy call in line 1862 runtime.go in function 
```func (a *DaprRuntime) Publish(req *pubsub.PublishRequest) error {``` be also changed?

**Comment by mukundansundar** [comment]
Thanks for the explanation!

File: pkg/resiliency/resiliency.go [hunk 3]
```diff
@@ -536,7 +536,7 @@ func (r *Resiliency) EndpointPolicy(ctx context.Context, app string, endpoint st
 	var t time.Duration
 	var rc *retry.Config
 	var cb *breaker.CircuitBreaker
-	operationName := fmt.Sprintf("endpoint[%s, %s]", app, endpoint)
+	operationName := "endpoint[" + app + ", " + endpoint + "]"
```
johnewart, line 539: While faster, this is arguably less readable; we don't hit this path a lot, do we?

pkg/resiliency/resiliency.go
```diff
@@ -618,7 +618,7 @@ func (r *Resiliency) ActorPreLockPolicy(ctx context.Context, actorType string, i
 	var t time.Duration
 	var rc *retry.Config
 	var cb *breaker.CircuitBreaker
-	operationName := fmt.Sprintf("actor[%s, %s]", actorType, id)
+	operationName := "actor[" + actorType + ", " + id + "]"
```
johnewart, line 621: Same, not a huge deal but we don't do concatentation elsewhere really that I have noticed

pkg/resiliency/policy_test.go
```diff
@@ -54,15 +54,14 @@ func TestPolicy(t *testing.T) {
 	ctx := context.Background()
 	for name, tt := range tests {
 		t.Run(name, func(t *testing.T) {
-			called := false
-			fn := func(ctx context.Context) error {
-				called = true
-
-				return nil
+			called := atomic.Bool{}
```
johnewart, line 57: Not really the best place for this comment but it's contextually relevant -- Can we simulate double-activation (or is that tested elsewhere?) to ensure that we handle race conditions now?

pkg/resiliency/policy.go
```diff
@@ -26,63 +26,68 @@ import (
...
 				go func() {
-					done <- operCopy(ctx)
+					rRes, rErr := operCopy(ctx)
+					done <- doneCh{rRes, rErr}
 				}()
 
 				select {
-				case err := <-done:
-					return err
+				case v := <-done:
```
johnewart, line 59: Do we want to timeout here? (Or is that what the comment on 45 is alluding to?)

pkg/resiliency/policy.go
```diff
@@ -26,63 +26,68 @@ import (
...
-// resiliency policies in a simple execution wrapper.
+type doneCh struct {
+	res any
+	err error
+}
+
+// Policy returns a policy runner that encapsulates the configured resiliency policies in a simple execution wrapper.
 func Policy(ctx context.Context, log logger.Logger, operationName string, t time.Duration, r *retry.Config, cb *breaker.CircuitBreaker) Runner {
-	return func(oper Operation) error {
+	return func(oper Operation) (any, error) {
```
johnewart, line 42: What do you think about wrapping this up into a struct with a type instead of unwrapping `any`? Something (admittedly a little hand-wavey for the sake of a PR comment) like: 

```
type PolicyExecutor[T any] struct {
	wrappedFunc func(context.Context) (T, error)
}

func NewPolicy[T any](wrapped func(context.Context) (T, error)) PolicyExecutor[T] {
	return PolicyExecutor[T]{
		wrappedFunc: wrapped,
	}
}

func (p PolicyExecutor[T]) Execute(ctx context.Context) (T, error) {
	return p.wrappedFunc(ctx)
}

type Result struct {
	StatusCode int
	Message    string
}


// Then we can use it like ...
func main() {
	p := NewPolicy[Result](func(ctx context.Context) (Result, error) {
		return Result{1, "ohai"}, nil
	})
	ctx := context.TODO()
	if res, err := p.Execute(ctx); err != nil {
        fmt.Printf("Oops! %v", err)
	} else {
		fmt.Printf("Result: %v\n", res)
	}

}

```

johnewart (commented):
I like this, it's much easier to read when using than having to manage atomic bits yourself! I just have a few small questions (and maybe one slightly larger one about wrapping it up into a struct that's a little more type safe than unwrapping `any` types)

pkg/resiliency/resiliency.go
ItalyPaleAle: We do, it's a very hot path actually

pkg/resiliency/resiliency.go
ItalyPaleAle: Same, this is a very hot path

pkg/resiliency/policy.go
ItalyPaleAle: The timeout happens with the context. Line 61 below is what executes the timeout

pkg/resiliency/policy.go
ItalyPaleAle: As discussed with John separately, John will work on a refactoring that uses generics in a separate PR.

pkg/messaging/direct_messaging.go
```diff
@@ -160,39 +162,31 @@ func (d *directMessaging) invokeWithRetry(
...
+					return rResp, fmt.Errorf("failed to invoke target %s after %d retries. Error: %w", app.id, attempt, rErr)
 				}
 				teardown(false)
-				return backoff.Permanent(rErr)
+				return rResp, backoff.Permanent(rErr)
 			})
-			// To maintain consistency with the existing built-in retries, we do some transformations/error handling.
-			if retriesExhaustedPath {
-				return nil, errors.Errorf("failed to invoke target %s after %v retries. Error: %s", app.id, numRetries, err.Error())
+			if err != nil {
```
ItalyPaleAle, line 184: Reviewers: note we don't need to do any unwrapping here because the backoff library already unwraps returned errors. Actually, unwrapping an error would return nil.

mukundansundar (commented):
had some questions on the new nil checks added ...

pkg/grpc/api.go
```diff
@@ -756,29 +780,35 @@ func (a *api) GetBulkState(ctx context.Context, in *runtimev1pb.GetBulkStateRequ
...
+		}, rErr
 	})
 	elapsed := diag.ElapsedSince(start)
 
 	diag.DefaultComponentMonitoring.StateInvoked(ctx, in.StoreName, diag.BulkGet, err == nil, elapsed)
 
+	bgr, _ := bgrAny.(*bulkGetRes)
+	if bgr == nil {
+		bgr = &bulkGetRes{}
+	}
```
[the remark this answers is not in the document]
ItalyPaleAle: If every single retry times out, this could happen, as in that case the policy would return an error and `nil` data.

I can look into adding a unit test

pkg/messaging/v1/invoke_method_response.go
```diff
@@ -116,22 +122,31 @@ func (imr *InvokeMethodResponse) Proto() *internalv1pb.InternalInvokeResponse {
...
 func (imr *InvokeMethodResponse) Trailers() DaprInternalMetadata {
+	if imr.r == nil {
+		return nil
+	}
 	return imr.r.Trailers
 }
 
 // Message returns message field in InvokeMethodResponse.
 func (imr *InvokeMethodResponse) Message() *commonv1pb.InvokeResponse {
+	if imr.r == nil {
```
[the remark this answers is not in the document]
ItalyPaleAle: It seems a pretty trivial thing to write a unit test for? I added this for safety, as during development it panicked once. Although it's likely the bug was somewhere else, and it got fixed then, it's probably not a bad idea to ensure nil-checks are in place regardless IMHO.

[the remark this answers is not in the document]
mukundansundar: Agreed ... these specific nil checks might not need the UTs, I was looking at any new code path that was added and commented on this too.

pkg/grpc/api.go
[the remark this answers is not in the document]
ItalyPaleAle: Added unit test

pkg/grpc/api.go
```diff
@@ -792,28 +822,25 @@ func (a *api) GetBulkState(ctx context.Context, in *runtimev1pb.GetBulkStateRequ
...
-				item.Etag = stringValueOrEmpty(r.ETag)
-				item.Metadata = r.Metadata
+			} else if resAny != nil {
+				res, ok := resAny.(*state.GetResponse)
+				if ok && res != nil {
+					item.Data = res.Data
+					item.Etag = stringValueOrEmpty(res.ETag)
+					item.Metadata = res.Metadata
+				} else {
+					item.Error = "res is empty"
```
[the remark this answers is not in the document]
ItalyPaleAle: Done

mukundansundar (approved):
lgtm

ItalyPaleAle:
Tests are green 🎉 Please let me know when this can be merged so resiliency can be unlocked

mukundansundar (approved):

mukundansundar:
@ItalyPaleAle please fix lint errors

ItalyPaleAle:
Fixed. I hope

pkg/resiliency/policy.go
ItalyPaleAle: Changed the code so this uses generics now

ItalyPaleAle:
@johnewart completed the refactoring (yet another!) to use generics rather than doing runtime type assertions

yaron2:
> @johnewart completed the refactoring (yet another!) to use generics rather than doing runtime type assertions

Please revert the latest commit, otherwise we won't be able to accept this otherwise import PR. This is because:

1. The generics refactoring isn't related to the original issue this PR set out to solve, and adds additional and uncalled for changes

2. This amount of refactoring with a new feature in the Go language that still has unknowns in terms of [performance](https://planetscale.com/blog/generics-can-make-your-go-code-slower) and general stability should be discussed before submitting a PR because of the potential downsides. This would have been better off communicated before hand. I'll add that an argument of Dapr already using generics is devoid as this is not about using generics or not, but about the scope of usage and prevailance in critical aspects.

ItalyPaleAle:
It was discussed above... https://github.com/dapr/dapr/pull/5538/#discussion_r1026994823

@johnewart raised concerns with the fact that using runtime assertions was not ideal, and I agree with him. 

We are using generics a lot in Dapr, in a lot of packages. If there was a concern with using generics, it would have been great to know that before I spent 4 hours on a Saturday refactoring this...

yaron2:
> It was discussed above... #5538
> 
> @johnewart raised concerns with the fact that using runtime assertions was not ideal, and I agree with him.
> 
> We are using generics a lot in Dapr, in a lot of packages. If there was a concern with using generics, it would have been great to know that before I spent 4 hours on a Saturday refactoring this...

The link above leads to the PR itself.

Again, the issue with this commit is twofold:

1. It doesn't belong in this PR and creates an uncalled for burden on reviewers (maintainers and approvers) that need to discern between the functional and new, entirely non-functional commits in this PR, ie. code that fixes the issue at hand and refactoring code
2. I understand that you and @johnewart were having this discussion (which I still cannot find), but the correct thing to do in this case when a really big, non-functional change is introduced is to at least mention a single maintainer (preferable) or reviewer. Maintainers can't follow all conversations in all PRs, and the review stage is where these things get called out unless attention is called to the change by the author prior.

yaron2:
> We are using generics a lot in Dapr, in a lot of packages. If there was a concern with using generics, it would have been great to know that before I spent 4 hours on a Saturday refactoring this...
> 
> Some examples of generics that have been accepted:
> 
> https://github.com/dapr/dapr/blob/fe1be20cd51e063fa81766ab51bfc598356b3993/pkg/components/disk_manifest_loader.go#L31-L32
> 
> 
> https://github.com/dapr/dapr/blob/fe1be20cd51e063fa81766ab51bfc598356b3993/pkg/components/pluggable/grpc.go#L41-L42
> 
> 
> https://github.com/dapr/dapr/blob/fe1be20cd51e063fa81766ab51bfc598356b3993/utils/utils.go#L78-L79
> 
> 
> https://github.com/dapr/components-contrib/blob/d8c2ce47fdb49db8e9ba3ad825d0f9492c0e6f27/internal/utils/utils.go#L34
> https://github.com/dapr/kit/blob/76dbc2afea63836fc9d8bacfea2945db85790136/ptr/of.go#L21-L22
> Not to mention the usage in the `backoff` package we depend on and this PR uses...

Please read the last part of my comment [here](https://github.com/dapr/dapr/pull/5538#issuecomment-1328174908) again - this is not about the usage of Generics as an absolute manner, this is about discussing the scope and criticality of the change. The changes introduced here are far wider than any existing usage of Generics we currently have.

ItalyPaleAle:
> It doesn't belong in this PR and creates an uncalled for burden on reviewers (maintainers and approvers) that need to discern between the functional and new, entirely non-functional commits in this PR, ie. code that fixes the issue at hand and refactoring code

The refactoring was on the code I added on this PR, not unrelated code.

First version of this PR used the atomic package to handle side effects in a sync-safe way. That was very complex to write, and read.

Version 2 refactored the `policy` method to return data, but always returned `(any, error)`. That was much easier to write and maintain, but used runtime assertions.

Version 3 changed the `policy` method again to support generics. It was a refactoring of the refactoring...

yaron2:
Looking at these comments, it was actually communicated 8 days ago this would be a separate PR done by a different author, but then this refactoring work arrives today as part of this PR, opposite of the earlier comment communicated:

Screen Shot 2022-11-26 at 9 47 40 PM

Even for reviewers who have been following the discussion, this creates difficulty in setting up expectations. No worries though, we can discuss those changes in a separate PR (as the code is already written, and would make it easier to discuss affected areas) as originally communicated.

yaron2:
> > It doesn't belong in this PR and creates an uncalled for burden on reviewers (maintainers and approvers) that need to discern between the functional and new, entirely non-functional commits in this PR, ie. code that fixes the issue at hand and refactoring code
> 
> The refactoring was on the code I added on this PR, not unrelated code.
> 
> First version of this PR used the atomic package to handle side effects in a sync-safe way. That was very complex to write, and read.
> 
> Version 2 refactored the `policy` method to return data, but always returned `(any, error)`. That was much easier to write and maintain, but used runtime assertions.
> 
> Version 3 changed the `policy` method again to support generics. It was a refactoring of the refactoring...

I understand that

> > It doesn't belong in this PR and creates an uncalled for burden on reviewers (maintainers and approvers) that need to discern between the functional and new, entirely non-functional commits in this PR, ie. code that fixes the issue at hand and refactoring code
> 
> The refactoring was on the code I added on this PR, not unrelated code.
> 
> First version of this PR used the atomic package to handle side effects in a sync-safe way. That was very complex to write, and read.
> 
> Version 2 refactored the `policy` method to return data, but always returned `(any, error)`. That was much easier to write and maintain, but used runtime assertions.
> 
> Version 3 changed the `policy` method again to support generics. It was a refactoring of the refactoring...

Thanks for the explanation! However I understood that, and those changes touch critical sections of the code base and affect existing, hot path code paths, and in the end are not needed or required to achieve what this PR sets out to do - which is why I ask to review them in a separate PR (as you communicated would be the case) where we can discuss and benchmark them.

ItalyPaleAle:
Hi @yaron2 thanks for your feedback.

To start, please let me explain the situation on why the original message said that John would do it in a separate PR. John and I were chatting offline and I reported here the final part of the conversation only. John convinced me that a solution based on generics was the best option, first of all due to ergonomics (I agree); however, I had just completed "v1" and "v2" in the same day, my eyes were "bleeding", and I did not feel like doing yet another refactoring (it's mostly a boring, repetitive tasks). John offered to do that _eventually_, but I didn't want this refactoring to hold up this PR, as this PR is a blocker for making Resiliency GA in 1.10: without this, even our E2E tests don't pass. However, as of yesterday the PR had not been merged yet, and I had some time to kill, so I just did it.

I apologize for not @'ing you in the conversation above. I (incorrectly) assumed that the choice of generics vs runtime assertion was an implementation detail that wasn't reason for concern.

That said, I've been thinking about your feedback, and I don't think this PR should be rolled back to "v2". There are multiple reasons why.

**Correctness**

I consider "v3" the most correct implementation at this point. Aside from the fact that it contains a few bug fixes (in one case, I had forgotten to do a type casting in the "panic-safe" way myself, i.e. like `res, _ = resAny.(foo)` and omitted the `, _`), it's the most ergonomic one to use and gives the best experience to developers working on Dapr. It's also an implementation without risks of panics, which can happen easily if you make a mistake like I did.

**Use of generics**

I understand your concerns with generics (although at this point I'm not as worried as you are about them anymore, as we've been them extensively in the runtime since 1.8), but rolling back to "v2" won't actually remove generics.

Although the API users would use would not use generics, they're still used internally in the same code paths. This is because the underlying method we use from the `backoff` library uses generics: [`backoff.RetryWithData`](https://pkg.go.dev/github.com/cenkalti/backoff/v4#RetryWithData). If we really had to avoid using generics, we'd need to roll back all the way to "v1" of this PR, which relies on the use of atomics to synchronize changes in side effects. I would not recommend that, as the code becomes much, much more complex to maintain and even read.

**Review process**

Two things on this front.

I don't think that having 2 PRs makes reviews easier. This is currently a pretty large PR, with ~1300 LOCs added, but it can be reviewed as-is. The ["v3" refactoring](https://github.com/ItalyPaleAle/dapr/commit/fdc3f99d9b2d2c5a428c19a8cea395bddd8cfd80) itself is another ~500 LOC PR that changes pretty much the same lines, so if it were to be a separate PR, it wouldn't be a lightweight to review either.

Which brings me to my concern about having such large PRs lingering. These changes touch a lot of parts of the codebase and they're very prone to merge conflicts. This last week was a holiday in the US so less changes came in–and yet, I still had to fix merge conflicts in this PR twice. I would much prefer, for my own mental health, if there wasn't yet another PR I need to keep up-to-date for potentially weeks, fixing merge conflicts in this PR and in next ones.

**Performance**

I hear your concerns about performance, so I created a branch which added back the old `resiliency. Policy` method (no generics) and benchmarks that against the new `resiliency.NewRunner[T]` method. You can find it here: https://github.com/ItalyPaleAle/dapr/tree/fix-race-conditions-resiliency-bench:

There are 3 benchmarks:

- `resiliency.NewRunner[int]`
- `resiliency. Policy` leaving the result as an `any` variable
- `resiliency. Policy` with casting the result from `any` to `int` (because runtime type assertions aren't free either)

Each benchmark is repeated twice, once without any retry configuration, and one with a retry config. When no retry config is used, `backoff.RetryWithData` is not invoked so there's actually no use of generics on the "backend".

These are the results on my dev box:

```
go test -benchmem -run=^$ -tags e2e,perf,conftests,unit,integration_test,certtests -bench ^Benchmark github.com/dapr/dapr/pkg/resiliency -v
goos: linux
goarch: amd64
pkg: github.com/dapr/dapr/pkg/resiliency
cpu: Intel(R) Xeon(R) Platinum 8272CL CPU @ 2.60GHz
BenchmarkNewRunner
BenchmarkNewRunner-4                    19138453                60.05 ns/op           48 B/op          1 allocs/op
BenchmarkPolicy
BenchmarkPolicy-4                       15378357                76.80 ns/op           80 B/op          1 allocs/op
BenchmarkPolicyCasting
BenchmarkPolicyCasting-4                15300786                76.65 ns/op           80 B/op          1 allocs/op
BenchmarkNewRunnerWithRetry
BenchmarkNewRunnerWithRetry-4            4357815               277.0 ns/op           136 B/op          6 allocs/op
BenchmarkPolicyWithRetry
BenchmarkPolicyWithRetry-4               4082368               295.0 ns/op           168 B/op          6 allocs/op
BenchmarkPolicyCastingWithRetry
BenchmarkPolicyCastingWithRetry-4        4017154               299.7 ns/op           168 B/op          6 allocs/op
PASS
ok      github.com/dapr/dapr/pkg/resiliency     8.287s
```

Looks like the method based on generics is about 16ns faster actually: without retries, that's 20% improvement (and the absolute improvement in ns is about the same when retries are added). However, we're still talking about tiny differences. (As an aside, looks like the impact of adding retries, which calls the code through the backoff library, does have a pretty sizable performance impact and adds 5 allocations - that's something that could be worth looking into, separately from this PR)

> Speaking of performance, one thing this PR changed (and which John commented on) was changing from usages like `fmt.Sprintf("endpoint[%s, %s]", app, endpoint)` to `"endpoint[" + app + ", " + endpoint + "]"`. We do this pattern (using `fmt.Sprintf` for what's essentially just string concatenation) a lot in the runtime, and every time I come across one in a PR, especially in a hot path like in this case, I try to change it. The perf difference in this case is actually significant, both by being 80% faster and by using 0 vs 3 allocations:
>
> ```
> go test -benchmem -run=^$ -tags e2e,perf,conftests,unit,integration_test,certtests -bench ^Benchmark github.> com/dapr/dapr/utils -v
> goos: linux
> goarch: amd64
> pkg: github.com/dapr/dapr/utils
> cpu: Intel(R) Xeon(R) Platinum 8272CL CPU @ 2.60GHz
> BenchmarkStringConcatFmt
> BenchmarkStringConcatFmt-4       5483677               212.7 ns/op            56 B/op          3 allocs/op
> BenchmarkStringConcat
> BenchmarkStringConcat-4         29105240                41.22 ns/op            0 B/op          0 allocs/op
> PASS
> ok      github.com/dapr/dapr/utils      2.662s
> ```

**TL;DR**

If you absolutely want me to revert back to "v2", I will (although it's not just about doing a "git revert" because I'd need to backport a few fixes made in "v3" that are unrelated from generics). However, I personally don't think we should.

yaron2:
> I apologize for not @'ing you in the conversation above. I (incorrectly) assumed that the choice of generics vs runtime assertion was an implementation detail that wasn't reason for concern.

As mentioned above, it's best to mention maintainers (not just me specifically :)) when in doubt. I understand that the decision to flag something for maintainers isn't always clear, which is why PR authors should be prepared to receive this feedback during the review stage, as happened here. All good.

> I consider "v3" the most correct implementation at this point. Aside from the fact that it contains a few bug fixes (in one case, I had forgotten to do a type casting in the "panic-safe" way myself, i.e. like res, _ = resAny.(foo) and omitted the , _), it's the most ergonomic one to use and gives the best experience to developers working on Dapr. It's also an implementation without risks of panics, which can happen easily if you make a mistake like I did.

I understand, yet it is still possible to have a panic safe implementation including those bug fixes without generics, is that correct?

> but rolling back to "v2" won't actually remove generics.

As I've mentioned [here](https://github.com/dapr/dapr/pull/5538#issuecomment-1328174908) and [here](https://github.com/dapr/dapr/pull/5538#issuecomment-1328176248), this is not about removing generics, merely about introducing them in a controlled way because of the scope of change and usage in the code paths affected by this PR

> I don't think that having 2 PRs makes reviews easier. This is currently a pretty large PR, with ~1300 LOCs added, but it can be reviewed as-is. The ["v3" refactoring](https://github.com/ItalyPaleAle/dapr/commit/fdc3f99d9b2d2c5a428c19a8cea395bddd8cfd80) itself is another ~500 LOC PR that changes pretty much the same lines, so if it were to be a separate PR, it wouldn't be a lightweight to review either.

I beg to differ - we have been consistently asking authors to break up PRs down for easier reviews, not just in dapr/dapr but in most repositories. The purpose of this in most cases is not to make it lightweight, but to make it *more* lightweight than what came before the break up.

> Which brings me to my concern about having such large PRs lingering. These changes touch a lot of parts of the codebase and they're very prone to merge conflicts. This last week was a holiday in the US so less changes came in–and yet, I still had to fix merge conflicts in this PR twice. I would much prefer, for my own mental health, if there wasn't yet another PR I need to keep up-to-date for potentially weeks, fixing merge conflicts in this PR and in next ones.

I agree 100%, which to me makes adding a commit that increases the LOCs by ~40% counter intuitive, because it now requires another review and necessarily keeps the PR from getting merged sooner. Breaking up the Generics refactoring into its own PR would have unblocked this large PR and allowed the Generics PR to have a smaller surface.

> Looks like the method based on generics is about 16ns faster actually: without retries, that's 20% improvement (and the absolute improvement in ns is about the same when retries are added).

Great that we have these benchmarks; one question here - do the generics methods contain the string formatting improvements you mentioned? because if they do, it skews the results greatly.

-----------------------

I leave the decision here up to you - we can either keep the "v3" changes in this PR and delay merging (as it must be re-reviewed now with the latest changes, and more carefully) or you can modify this PR to be panic safe and then commit the generics refactoring as a separate PR. That would allow us to merge the existing PR much faster.

ItalyPaleAle:
Thanks. I'd prefer to leave this in at this point as it's in a "done" state now, if reviewers don't mind it. I understand it may delay the merging a bit, but at least there's only 1 PR I need to worry about.

No, the benchmark ran only against the policy method, which was the smallest unit I could apply the benchmark to. The string concatenation change was in a "higher level" method and not included in the benchmarks.

yaron2:
> Thanks. I'd prefer to leave this in at this point as it's in a "done" state now, if reviewers don't mind it. I understand it may delay the merging a bit, but at least there's only 1 PR I need to worry about.
> 
> No, the benchmark ran only against the policy method, which was the smallest unit I could apply the benchmark to. The string concatenation change was in a "higher level" method and not included in the benchmarks.

Fine with me. @daixiang0 please help review.

pkg/http/api.go
```diff
@@ -1412,63 +1446,67 @@ func (a *api) onDirectMessage(reqCtx *fasthttp.RequestCtx) {
...
 		}
 
-		errorOccurred = false
-		invokev1.InternalMetadataToHTTPHeader(reqCtx, resp.Headers(), reqCtx.Response.Header.Set)
-		var contentType string
-		contentType, body = resp.RawData()
-		reqCtx.Response.Header.SetContentType(contentType)
+		dmpr := &directMessagingPolicyRes{}
+
+		invokev1.InternalMetadataToHTTPHeader(reqCtx, rResp.Headers(), reqCtx.Response.Header.Set)
```
mukundansundar, line 1470: qq Is it ok to share `reqCtx` across go routine calls? 
https://pkg.go.dev/github.com/valyala/fasthttp#RequestCtx
`It is unsafe modifying/reading RequestCtx instance from concurrently running goroutines. The only exception is TimeoutError*, which may be called while other goroutines accessing RequestCtx.`

pkg/runtime/runtime.go
```diff
@@ -1252,7 +1257,7 @@ func (a *DaprRuntime) sendBindingEventToApp(bindingName string, data []byte, met
 			if resp != nil {
 				body = resp.Data
 			}
-			return nil, errors.Wrap(err, fmt.Sprintf("error invoking app, body: %s", string(body)))
+			return nil, fmt.Errorf("error invoking app, body: '%s', error: %w", string(body), err)
```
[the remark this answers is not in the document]
ItalyPaleAle: I think you're right, the body should not be sent back as it could contain PII. I'll removeit

pkg/runtime/runtime.go
```diff
@@ -1284,22 +1289,26 @@ func (a *DaprRuntime) sendBindingEventToApp(bindingName string, data []byte, met
...
-			if err != nil {
-				return err
+		policy := resiliency.NewRunner[*invokev1.InvokeMethodResponse](ctx,
+			a.resiliency.ComponentInboundPolicy(bindingName, resiliency.Binding),
+		)
+		resp, err := policy(func(ctx context.Context) (*invokev1.InvokeMethodResponse, error) {
+			rResp, rErr := a.appChannel.InvokeMethod(ctx, req)
+			if rErr != nil {
+				return rResp, rErr
 			}
```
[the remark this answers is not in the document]
ItalyPaleAle: It is not a behavior change because previously `resp` was set as a side effect

pkg/http/api.go
ItalyPaleAle: Good catch. I'll change how this works so `reqCtx` is not modified within a callback, as the callback may continue even when the response was sent to the user (e.g. in case of timeouts)

mukundansundar (dismissed):
lgtm.

johnewart (approved):
I personally find this much easier to read where we use resiliency policies :+1:  

I don't have any blocking feedback - just a few little questions (and one about maybe adding some metadata to policy runners in the future, like attempt counters).

pkg/actors/actors.go
```diff
@@ -323,38 +323,46 @@ func (a *actorsRuntime) startDeactivationTicker(configuration Config) {
...
+			appID:              rAppID,
+		}, nil
 	})
-
-	if rErr != nil {
-		return nil, rErr
+	if err != nil {
+		return nil, err
+	}
+	if lar == nil {
```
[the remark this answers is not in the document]
ItalyPaleAle: it doesn't really deviate from the previous behavior because before `targetActorAddress` and `appID` were set as a side effect. The policyRunner allows returning one value only, so I had to wrap it in a struct.

Although policyRunner always returns an `*lookupActorRes` object, if the policy method times out then we only get an error and no data value. So `lar` could be nil. In this specific case, if `lar == nil` then automatically `err != nil` so L358 should never be hit....... but an extra nil-check never hurt nobody IMHO

pkg/actors/actors.go
```diff
@@ -378,41 +386,28 @@ func (a *actorsRuntime) callRemoteActorWithRetry(
...
-			policy := a.resiliency.BuiltInPolicy(ctx, resiliency.BuiltInActorRetries)
-			var resp *invokev1.InvokeMethodResponse
-			err := policy(func(ctx context.Context) (rErr error) {
-				var teardown func(destroy bool)
-				retriesExhaustedPath = false
-				resp, teardown, rErr = fn(ctx, targetAddress, targetID, req)
+			policyRunner := resiliency.NewRunner[*invokev1.InvokeMethodResponse](ctx,
+				a.resiliency.BuiltInPolicy(resiliency.BuiltInActorRetries),
+			)
+			attempts := atomic.Int32{}
```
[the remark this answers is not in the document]
ItalyPaleAle: Right now there are only 2 instances when an attempt counter is needed (and only for logs). If more methods need that in the future we can always add it later.

pkg/actors/actors.go
```diff
@@ -1008,11 +1021,13 @@ func (a *actorsRuntime) executeReminder(reminder *Reminder) error {
...
 	req.WithRawData(b, invokev1.JSONContentType)
 
-	policy := a.resiliency.ActorPreLockPolicy(context.Background(), reminder.ActorType, reminder.ActorID)
-	return policy(func(ctx context.Context) error {
-		_, err := a.callLocalActor(ctx, req)
-		return err
+	policyRunner := resiliency.NewRunner[any](context.TODO(),
+		a.resiliency.ActorPreLockPolicy(reminder.ActorType, reminder.ActorID),
+	)
+	_, err = policyRunner(func(ctx context.Context) (any, error) {
```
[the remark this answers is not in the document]
ItalyPaleAle: I can make it on a separate PR, so not to remove the approval on this :)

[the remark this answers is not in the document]
yaron2: I solved the problem by dismissing the approval :)

Please change to the actual response type

pkg/actors/actors.go
```diff
@@ -1365,10 +1387,11 @@ func (a *actorsRuntime) executeTimer(actorType, actorID, name, dueTime, period,
...
 
-	policy := a.resiliency.ActorPreLockPolicy(context.Background(), actorType, actorID)
-	err = policy(func(ctx context.Context) error {
-		_, err = a.callLocalActor(ctx, req)
-		return err
+	policyRunner := resiliency.NewRunner[any](context.TODO(),
+		a.resiliency.ActorPreLockPolicy(actorType, actorID),
+	)
+	_, err = policyRunner(func(ctx context.Context) (any, error) {
+		return a.callLocalActor(ctx, req)
```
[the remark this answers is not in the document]
yaron2: Please change to the concrete response type

pkg/actors/actors.go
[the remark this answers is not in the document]
ItalyPaleAle: Done

pkg/actors/actors.go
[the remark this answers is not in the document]
ItalyPaleAle: Done

yaron2 (approved):

yaron2:
My minority opinion here is that the addition of generics to the resiliency runner was not needed nor improved much by the way of code readability at least.

Other than that, great PR.

daixiang0:
Agree that generics may be better to put in a single PR.

PR #5544, opened 2022-11-21
Fixed issue with tracing not working in HTTP service invocation

ItalyPaleAle:
This PR fixes a regression reported by @artursouza that was caught thanks the E2E tests in the Java SDK repos.

The regression was introduced by #5316 . When we converted the HTTP middlewares from fasthttp to net/http, we used two "adapters" to convert between the frameworks. As part of that, user values set in the HTTP context were not passed correctly from the router to the tracing middleware, and they were lost in the conversion. These user values are used by the router to store the values from the URLs (e.g. the method name and target ID app).

This PR fixes the issue by forking the [fasthttpadaptor](https://github.com/valyala/fasthttp/tree/master/fasthttpadaptor) and adding an extension to preserve changes to user values across the adaptors. I opted for a hard fork given that the upstream package seems to be mostly stable and this change is pretty much Dapr-specific. It's also meant to be temporary, until #4979 is addressed.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -10,7 +10,7 @@ require (
 	github.com/cenkalti/backoff/v4 v4.1.3
 	github.com/dapr/components-contrib v1.9.1-0.20221111215803-c92827c3defc
 	github.com/dapr/kit v0.0.3-0.20221102045011-c213121f0b4f
-	github.com/fasthttp/router v1.4.12
+	github.com/fasthttp/router v1.4.13
 	github.com/ghodss/yaml v1.0.0
 	github.com/go-logr/logr v1.2.3
 	github.com/gogo/protobuf v1.3.2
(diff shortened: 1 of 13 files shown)


yaron2:
I just verified that HTTP tracing is working in Dapr 1.9.4, I can see traces with method names and app_id. Can you explain in more detail which scenarios are broken?

yaron2:
Does this PR contain tests to capture this regression?

ItalyPaleAle:
No. But @artursouza told me he wanted to create an E2E test. I'll let him chime in if he's still planning to do that, otherwise I can do (even on a separate PR).

It should be an E2E and not unit test because it'd be very hard to write this as unit test AND this situation is hopefully temporary (see #4979)

artursouza:
> 

I am writing E2E test to catch this bug. It is taking some time because we will need to install zipkin and use zipkin's API directly for the first time in our tests. This PR can wait for the E2E tests to be completed from my side.

artursouza (approved):
I will follow up with the E2E test myself. Merging for the sake of having higher quality nightly builds.

PR #5556, opened 2022-11-23
Disallow app-id containing dots

### LINKED ISSUE #5552: Dapr service can not be called when the app-id contains dots
## In what area(s)?

 /area runtime

## What version of Dapr?

1.8.x

## Expected Behavior

The services are resolved by app-id or I receive an error that dapr does not support app-ids containing dots.

## Actual Behavior

Dapr tries to resolve the service id as an url

Try to do a service call to an app with id "services.my.great.service."

## Release Note

Depends on the solution chosen.

ItalyPaleAle:
The daprd binary now shows a fatal error if `--app-id` contains a dot, which is not allowed.

Fixes #5552

### DIFF
--- a/pkg/runtime/cli.go
+++ b/pkg/runtime/cli.go
@@ -127,6 +127,9 @@ func FromFlags() (*DaprRuntime, error) {
 	if *appID == "" {
 		return nil, errors.New("app-id parameter cannot be empty")
 	}
+	if strings.IndexRune(*appID, '.') != -1 {
+		return nil, errors.New("app-id parameter cannot contain dots")
+	}
 
 	// Apply options to all loggers
 	loggerOptions.SetAppID(*appID)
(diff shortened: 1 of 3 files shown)


mukundansundar (commented):
Should the `ValidateKubernetesAppID` method in pkg/validation/validation.go be modified and used here as well to be completely consistent with k8s mode `AppID` validation?

ItalyPaleAle:
> Should the ValidateKubernetesAppID method in pkg/validation/validation.go be modified and used here as well to be completely consistent with k8s mode AppID validation?

Uhm seems that method is used in the operator and injector. I don't think that's necessary, this validation was added to the `daprd` binary, so putting a dot in the `--app-id` flag there would just cause it to fail to start.

ItalyPaleAle:
@mukundansundar oh yeah, good idea. I changed it to use a method from the validation package.

mukundansundar:
@ItalyPaleAle if docs are not there please add docs for this.

artursouza (approved):

Issue #5321, opened 2022-10-07
Trace issue while using Resiliency


rahulpoddar-fyndna:
We are using resiliency with 1.8.4 version of Dapr and having following issue
Parent trace id getting generated as 00-00000000000000000000000000000000-0000000000000000-00 when timeout resiliency parameter is configured for the first entry point service. 

@akhilac1 @mukundansundar

rabollin:
Tagging @halspang to review this issue.

rahulpoddar-fyndna:
Few more details
Observations/Concerns :
1. If timeout is defined for the entry point service i.e ARef service, then the traceId gets generated as 00000000000000000000000000000000. While proper trace id gets generated if the timeout is defined as 0s or the timeout property is not at all defined.
2. If there is a dapr interaction happening from ARef -> CRef, traceId for ARef and CRef is not the same and it gets changed i.e trace is broken when timeout policy for CRef is defined. While same trace id gets generated if the timeout is defined as 0s or the timeout property is not at all defined for CRef and hence the trace is not broken.
3. If in case the timeout value is defined something greater than 60s, still dapr does not consider the same and gives timeout error assuming the 60s as upper limit.

Questions :
1. What is the default timeout value, if no timeout policy is defined in resiliency?

pooja-fyndna:
[RefService.zip](https://github.com/dapr/dapr/files/9772142/RefService.zip)

Attached zip contains following artifacts to simulate issues : 
1. ARef and CRef - Sample applications
2. /RefService/local/dapr-components - Folder containing the resiliency policy details
3. /RefService/local/dapr-config - Folder containing the dapr configuration details
4. /RefService/README.md - Text containing dapr run command details and steps to reproduce different issues
5. /RefService/postman/Reference.postman_collection.json - Postman collection to run the APIs

Observations/Concerns : 
1. If timeout is defined for the entry point service i.e ARef service, then the traceId gets generated as 00000000000000000000000000000000. While proper trace id gets generated if the timeout is defined as 0s or the timeout property is not at all defined.

2. If there is a dapr interaction happening from ARef -> CRef, traceId for ARef and CRef is not the same and it gets changed i.e trace is broken when timeout policy for CRef is defined. While same trace id gets generated if the timeout is defined as 0s or the timeout property is not at all defined for CRef and hence the trace is not broken.

3. If in case the timeout value is defined something greater than 60s, still dapr does not consider the same and gives timeout error assuming the 60s as upper limit.

Questions : 
[...]

PR #5560, opened 2022-11-24
Issue 5321 - Broken trace when timeout resiliency policy is enabled

akhilac1:

# Description
When timeout resiliency policy is enabled, the traceID does not propagate with the request and a broken trace is shown with a Span of 1, separately for client and server. To replicate the issue run the service invocation Quickstart and check the traces in zipkin. connected traces for order processor and checkout will be shown. 
Now enable resiliency [ since it is a preview feature ] and apply timeout policy to both the services. You will see individual traces for order processor and checkout app 

With this fix, you will be able to see a connected trace in both the scenarios.

Retrieving the span using the utility function fails as the function checks for the context object to be of type fast http context. When resiliency policy for timeout is enabled, the context object passed is of type timeout Context. So the span is retrieved as <nil>.

Fixed the utility function to directly look for value in the context if the context is not of type fasthttp. Retained default behaviour if the span value is not found based on the key "daprSpanContextKey"

## Issue reference

Please reference the issue this PR will close: #5321 

## Checklist

* [X] Code compiles correctly
* [ ] Created/updated tests
* [X] Unit tests passing

### DIFF
--- a/pkg/diagnostics/utils/trace_utils.go
+++ b/pkg/diagnostics/utils/trace_utils.go
@@ -80,6 +80,11 @@ func SpanFromContext(ctx context.Context) trace.Span {
 		if val != nil {
 			return val.(trace.Span)
 		}
+	} else {
+		val := ctx.Value(daprFastHTTPContextKey)
+		if val != nil {
+			return val.(trace.Span)
+		}
 	}
 
 	span := trace.SpanFromContext(ctx)


yaron2 (approved):
LGTM

yaron2:
Thanks @akhilac1, this helps the efforts of making Resiliency stable for the 1.10 release.

Issue #5481, opened 2022-11-08
Scale down in environment with Dapr Side Car does not work as expected - Getting 503 Response


rahulpoddar-fyndna:
We are using Dapr Side car in our application and it is deployed in GKE. Our application is based on spring boot. We use KEDA for HPA. While scaling down the pods in GKE environment, we observed 503 http response code for many request.  We have configured the dapr.io/graceful-shutdown-seconds for Dapr container and gracefultimeout for our application as well (spring boot).

**Analysis of issue**

**A. Kubernetes Graceful Shutdown Process**
1. POD is marked in TERMINATING State
2. Containers in the POD receives preStop request first via preStop configured. preStop will be received only if preStop is configured.
3. After preStop, SIGTERM is sent to Containers. At this stage the containers should not receive any request and it should shut its http gates.

**B. What does GKE do when preStop is being executed by containers.**
POD is marked in TERMINATING state, the GKE sends preStop requests to containers. While preStop is being handled by Container, the GKE utilizes this time to update its IP table.  GKE removes the IP of POD marked in terminating state. While container is handling preStop , GKE can still send new request to this container till it finally removes POD IP.  Container can receive new requests as its http gates are still open.
In a way, preStop provides a way to delay the sending of SIGTERM to containers in POD so that IP table can be updated by GKE. Refer below links for more details.

**C. What happens when SIGTEMR is received by container**
The container shuts down gracefully respecting the time configured. Inbound ports are closed gracefully.

**D. Scale down behaviour without Dapr side car**
Application with spring has preStop hook. We configure this hook and provide some sleep. During this time GKE updates the IP tables to remove the POD IP.
While this is happening spring container still receives new request as it has not yet closed its http ports and GKE is sending the new requests.
After preStop , Spring Container receives SIGTERM and close its http ports. So execution of preStop hook by spring container provides sufficient time to GKE to update IP tables.

**Please refer to**
https://cloud.google.com/blog/products/containers-kubernetes/kubernetes-best-practices-terminating-with-grace
Refer point no 2 in above URL
For Spring, please refer
https://docs.spring.io/spring-boot/docs/current/reference/html/deployment.html#deployment.cloud.kubernetes.container-lifecycle

**E. Scale down behaviour with Dapr side car**
[...]

PR #5562, opened 2022-11-24
Initiate Dapr shutdown after expiry of grace period - Issue 5481

akhilac1:

# Description

Issue #5481 limits usage of Dapr for a subset of users, especially during rolling upgrade. The load balancer continues to send requests even after SIGTERM is sent to the Pod. The app being written in spring can support a pre-stop hook and delay shutdown to ensure the requests coming un while load balancer identifies the pod as Terminating and removes it from the list of active pods.

Dapr, however, closes inbound channels and hence will not receive any requests once SIGTERM is received.  So many failed requests are noticed in this environment.

This PR moves the closing of inbound channels to after the graceperiod expiry to address the immediate closure and hence dropping of requests.

## Issue reference
#5481 

Please reference the issue this PR will close: #5481 

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests
* [X] Unit tests passing


ItalyPaleAle (changes requested):
I do not support this change, sorry.

A few months ago we went through an extensive amount of work to be able to shut down PubSub and input binding before the grace period, as this is the most correct behavior for the majority of users.

Imagine a pod that gets shut down. Both daprd and the app get the SIGKILL at the same time and should immediately begin the shutdown sequence. Usually apps terminate instantly. If dapr were to still take messages from PubSub and input bindings, we'd try to deliver them to apps that are already shut down, and that's not polite (and could cause messages to go to the DLQ incorrectly).

We have the grace period on output things just so the app can complete the work and store the output. But we shouldn't try to bring more work to the app during the grace period.

mukundansundar:
> I do not support this change, sorry.
> 
> A few months ago we went through an extensive amount of work to be able to shut down PubSub and input binding before the grace period, as this is the most correct behavior for the majority of users.
> 
> Imagine a pod that gets shut down. Both daprd and the app get the SIGKILL at the same time and should immediately begin the shutdown sequence. Usually apps terminate instantly. If dapr were to still take messages from PubSub and input bindings, we'd try to deliver them to apps that are already shut down, and that's not polite (and could cause messages to go to the DLQ incorrectly).
> 
> We have the grace period on output things just so the app can complete the work and store the output. But we shouldn't try to bring more work to the app during the grace period.

I agree with this comment. I think all input bindings and subscriptions and subscriptions need to be closed immediately.

But we can delay closing the Dapr APIs and Actors as calls to these are generally initiated by the application.

cc @ItalyPaleAle WDYT?

mukundansundar:
> I do not support this change, sorry.
> 
> A few months ago we went through an extensive amount of work to be able to shut down PubSub and input binding before the grace period, as this is the most correct behavior for the majority of users.
> 
> Imagine a pod that gets shut down. Both daprd and the app get the SIGKILL at the same time and should immediately begin the shutdown sequence. Usually apps terminate instantly. If dapr were to still take messages from PubSub and input bindings, we'd try to deliver them to apps that are already shut down, and that's not polite (and could cause messages to go to the DLQ incorrectly).
> 
> We have the grace period on output things just so the app can complete the work and store the output. But we shouldn't try to bring more work to the app during the grace period.

This makes sense. We are closing all Dapr APIs, and Dapr to App communication. 
Only the output components doing any work will be closed after graceful time period. 

Though I have once question, most of the output component APIs are synchronous in Dapr, so what happens to in-flight requests if we suddenly close down the servers? 
Can we configure the servers to not accept new requests during the time period but keep the server still running till the in-flight requests are complete? 

@ItalyPaleAle WDYT?

ItalyPaleAle:
> Though I have once question, most of the output component APIs are synchronous in Dapr, so what happens to in-flight requests if we suddenly close down the servers?
> Can we configure the servers to not accept new requests during the time period but keep the server still running till the in-flight requests are complete?

Right now, the servers are shut down at the end of the grace period.

Interesting idea about not accepting new requests, but I am not sure how I feel about it. The app should be able to invoke Dapr APIs during the graceful shutdown period if it needs to store its output somewhere IMHO.

Here's how I see the shutdown sequence working:

1. When the pod starts getting terminated, the app should immediately stop accepting new work. This is outside of the control of Dapr, except for the part where we should not be sending more work to the app during the graceful shutdown period: that's what we do with input bindings, PubSub, and input service invocation.
2. The app may still need some time to process in-progress work. For example, imagine that a half-second before the SIGKILL, the app received a message from PubSub and that takes 3 seconds to process. That's why there's the graceful shutdown period in Dapr, after all.
3. After the message is done processing, the app may want to store its output somewhere, using Dapr.

That's why I don't think Dapr should stop accepting new work (from the app), as it may still be needed. It's the app's responsibility to complete all work within the grace period, however.

mukundansundar:
> That's why I don't think Dapr should stop accepting new work (from the app), as it may still be needed. It's the app's responsibility to complete all work within the grace period, however.

But with the current sequence in the code, the Dapr APIs are closed immediately. We are closing the APIs even before the graceful time period. (closer.Close()) method.

```go
log.Info("dapr shutting down.")

log.Info("Stopping PubSub subscribers and input bindings")
a.stopSubscriptions()
a.stopReadingFromBindings()
a.cancel()
a.stopActor()
log.Info("Stopping Dapr APIs")
for _, closer := range a.apiClosers {
  if err := closer.Close(); err != nil {
	  log.Warnf("error closing API: %v", err)
  }
}
shutdownCtx, shutdownCancel := context.WithCancel(context.Background())
go func() {
  if a.tracerProvider != nil {
	  a.tracerProvider.Shutdown(shutdownCtx)
   }
}()

log.Infof("Waiting %s to finish outstanding operations", duration)
<-time.After(duration)
```

Shouldn't the `<-time.After(duration)` wait be called before closing the Dapr APIs? 

Additionally, if DAPR APIs are still available during graceful shutdown period, shouldn't actors (`a.stopActor()`) also be available? 

We are also immediately calling the `cancel()` function of the overall context `a.ctx` that is being shared in runtime. Shouldn't that also be after the timeout?

ItalyPaleAle:
Yes I agree with you.

> Shouldn't the <-time.After(duration) wait be called before closing the Dapr APIs?

I think that would probably be correct

> Additionally, if DAPR APIs are still available during graceful shutdown period, shouldn't actors (a.stopActor()) also be available?

Not 100% sure on this, but I think what you're saying makes sense

> We are also immediately calling the cancel() function of the overall context a.ctx that is being shared in runtime. Shouldn't that also be after the timeout?

Possibly, but I'm not exactly sure what the context is used for right now. But you are probably correct.

yaron2:
> Shouldn't the <-time.After(duration) wait be called before closing the Dapr APIs?

Yes, that is the correct behavior.

mukundansundar:
> > Shouldn't the <-time.After(duration) wait be called before closing the Dapr APIs?
> 
> Yes, that is the correct behavior.

Also @yaron2  ... what is your thought on the following two lines ? 
>Additionally, if DAPR APIs are still available during graceful shutdown period, shouldn't actors (a.stopActor()) also be available?

> We are also immediately calling the cancel() function of the overall context a.ctx that is being shared in runtime. Shouldn't that also be after the timeout?

yaron2:
> Additionally, if DAPR APIs are still available during graceful shutdown period, shouldn't actors (a.stopActor()) also be available?

Actors are different, they should when the signal is received to give the runtime the chance to disconnect from placement as soon as possible and finish the rehashing in a controlled manner. Since Dapr is the actual compute orchestrator here, I think that makes sense.

> We are also immediately calling the cancel() function of the overall context a.ctx that is being shared in runtime. Shouldn't that also be after the timeout?

It should probably be called after the timeout.

ItalyPaleAle (changes requested):
@akhilac1 please make the changes as discussed above

akhilac1:
> @akhilac1 please make the changes as discussed above

@ItalyPaleAle @yaron2 @mukundansundar - tagging for review

pkg/runtime/runtime_test.go
```diff
@@ -5392,3 +5371,201 @@ func TestNamespacedPublisher(t *testing.T) {
...
+
+	var activeActCount int
+	activeActors := rt.actor.GetActiveActorsCount(rt.ctx)
+	for _, v := range activeActors {
+		activeActCount += v.Count
+	}
+	assert.Equal(t, activeActCount, 0)
+}
+
+func initMockStateStoreForRuntime(rt *DaprRuntime, primaryKey string, e error) *daprt.MockStateStore {
```
mukundansundar, line 5550:
```suggestion
func initMockStateStoreForRuntime(rt *DaprRuntime, encKey string, e error) *daprt.MockStateStore {
```
to not get confused with `primaryKey` in a state store ... Please change all occurrences while calling also if possible.

pkg/runtime/runtime_test.go
```diff
@@ -5392,3 +5371,201 @@ func TestNamespacedPublisher(t *testing.T) {
...
+	go func() {
+		<-sigs
+		rt.Shutdown(rt.runtimeConfig.GracefulShutdownDuration)
+	}()
+
+	if p, err := os.FindProcess(os.Getpid()); err != nil {
+		rt.Shutdown(rt.runtimeConfig.GracefulShutdownDuration)
+	} else {
+		p.Signal(syscall.SIGTERM)
+	}
```
mukundansundar, line 5428: can this just be `sigs <- syscall.SIGINT` or `sigs <- syscall.SIGTERM`? Also what piece of code is being tested here since rt.Shutdown() is being called manually anyway? Can we completely remove the signal processing logic in this test and call rt.Shutdown(3s) directly?

Similarly comment for other tests also below.

pkg/runtime/runtime_test.go
```diff
@@ -5392,3 +5371,201 @@ func TestNamespacedPublisher(t *testing.T) {
 
 	assert.Equal(t, "ns1topic0", rt.pubSubs[TestPubsubName].component.(*mockPublishPubSub).PublishedRequest.Topic)
 }
+
+func TestGracefulShutdownPubSub(t *testing.T) {
+	rt := NewTestDaprRuntime(modes.StandaloneMode)
+	// setup pubsub
+	testGracefulRestartPubSub := "shutdownPubsub"
```
mukundansundar, line 5378: nitpick 
```suggestion
	testGracefulShutdownPubSub := "shutdownPubsub"
```

pkg/runtime/runtime_test.go
```diff
@@ -5392,3 +5371,201 @@ func TestNamespacedPublisher(t *testing.T) {
...
+		rt.Shutdown(rt.runtimeConfig.GracefulShutdownDuration)
+	}()
+
+	if p, err := os.FindProcess(os.Getpid()); err != nil {
+		rt.Shutdown(rt.runtimeConfig.GracefulShutdownDuration)
+	} else {
+		p.Signal(syscall.SIGTERM)
+	}
+	time.Sleep(1 * time.Second)
+	assert.Nil(t, rt.pubsubCtx)
```
mukundansundar, line 5430: instead of sleeping for 1s, can this be blocked on something like this :
```golang
select {
  case <-rt.pubsubCtx.Done():
     // assert context closed, topic context closed, topic routes nil 
    // assert elapsed time is less than graceful shutdown time
  case <-time.After(gracefulShutdownTime):
    // fail the test as this case should not be triggered    t.Fail()
}

```

mukundansundar (changes requested):
Thanks for adding the UTs. They seem to be flaky in windows. please take a look. 
Have added comments.

pkg/runtime/runtime_test.go
akhilac1: The idea of this test is to init a SIGTERM and test. If we are unable to send sigterm we bypass the signal sending piece and call shutdown anyway. This logic is needed to ensure the test does not break in Windows. For Mac and Linux the test runs with a SIGTERM being sent to runtime.

pkg/runtime/runtime_test.go
akhilac1: We would need to check the cleanup after graceperiod expires. Updated as suggested

pkg/runtime/runtime_test.go
akhilac1: Updated. Please check @mukundansundar

pkg/runtime/runtime_test.go
akhilac1: Updated. Please check @mukundansundar

akhilac1:
> Thanks for adding the UTs. They seem to be flaky in windows. please take a look.
> Have added comments.

Yes. Sending SIGTERM fails and hence the assert operation fails. Fixed this to call rt.Shutdown in case we are unable to send SIGTERM

pkg/runtime/runtime_test.go
```diff
@@ -5392,3 +5371,183 @@ func TestNamespacedPublisher(t *testing.T) {
...
+	sigs := make(chan os.Signal, 1)
+	signal.Notify(sigs, syscall.SIGINT, syscall.SIGTERM)
+	go func() {
+		<-sigs
+		rt.Shutdown(rt.runtimeConfig.GracefulShutdownDuration)
+	}()
+	if p, err := os.FindProcess(os.Getpid()); err != nil {
+		rt.Shutdown(rt.runtimeConfig.GracefulShutdownDuration)
+	} else {
+		if err := p.Signal(syscall.SIGTERM); err != nil { // SIGTERM cannot be sent on WINDOWS
```
ItalyPaleAle, line 5549: Rather than sending an actual SIGTERM to our tests, which could cause unintended consequences, can we do that differently? For example, invoke the shutdown method directly, or use some mocking.

pkg/runtime/runtime_test.go
```diff
@@ -5392,3 +5371,183 @@ func TestNamespacedPublisher(t *testing.T) {
...
+
+	// assert
+	assert.NoError(t, err, "expected no error")
+
+	rt.namespace = "test"
+	rt.runtimeConfig.mtlsEnabled = true
+	assert.Nil(t, rt.initActors())
+
+	sendSigterm(rt)
+	<-time.After(rt.runtimeConfig.GracefulShutdownDuration)
```
ItalyPaleAle, line 5505: Same as in the previous test, please use something that adds some buffer to reduce flakiness

pkg/runtime/runtime_test.go
```diff
@@ -5392,3 +5371,183 @@ func TestNamespacedPublisher(t *testing.T) {
...
+	err := rt.initState(mockStateComponent)
+
+	// assert
+	assert.NoError(t, err, "expected no error")
+
+	rt.namespace = "test"
+	rt.runtimeConfig.mtlsEnabled = true
+	assert.Nil(t, rt.initActors())
+
+	sendSigterm(rt)
```
ItalyPaleAle, line 5504: Because actors are stopped before the graceful shutdown duration, should the test for actors being shut down be moved to before that?

ItalyPaleAle (changes requested):

pkg/runtime/runtime_test.go
akhilac1: Actor shutdown is initiated before graceperiod so that we provide enough time for the shutdown. The test checks after grace period for actor shutdown completion

pkg/runtime/runtime_test.go
akhilac1: Added +2 seconds additional wait before assertion.

pkg/runtime/runtime_test.go
```diff
@@ -5392,3 +5371,183 @@ func TestNamespacedPublisher(t *testing.T) {
...
+	mockAppChannel.On("InvokeMethod", mock.MatchedBy(matchContextInterface), req).Return(fakeResp, nil)
+
+	require.NoError(t, rt.initPubSub(pubsubComponent))
+	rt.startSubscriptions()
+	assert.NotNil(t, rt.pubsubCtx)
+	assert.NotNil(t, rt.topicCtxCancels)
+	assert.NotNil(t, rt.topicRoutes)
+
+	sendSigterm(rt)
+	<-time.After(rt.runtimeConfig.GracefulShutdownDuration)
```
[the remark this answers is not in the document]
akhilac1: time.After with the additional  buffer would cover for rest of the scenarios. So added the buffer of 2 seconds.

pkg/runtime/runtime_test.go
```diff
@@ -5392,3 +5371,183 @@ func TestNamespacedPublisher(t *testing.T) {
...
+	cout := componentsV1alpha1.Component{}
+	cout.ObjectMeta.Name = "testOutputBinding"
+	cout.Spec.Type = "bindings.testOutputBinding"
+	require.NoError(t, rt.initInputBinding(cin))
+	require.NoError(t, rt.initOutputBinding(cout))
+
+	assert.Equal(t, len(rt.inputBindings), 1)
+	assert.Equal(t, len(rt.outputBindings), 1)
+
+	sendSigterm(rt)
```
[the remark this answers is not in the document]
akhilac1: updated

pkg/runtime/runtime_test.go
```diff
@@ -5392,3 +5371,183 @@ func TestNamespacedPublisher(t *testing.T) {
...
+	require.NoError(t, rt.initPubSub(pubsubComponent))
+	rt.startSubscriptions()
+	assert.NotNil(t, rt.pubsubCtx)
+	assert.NotNil(t, rt.topicCtxCancels)
+	assert.NotNil(t, rt.topicRoutes)
+
+	sendSigterm(rt)
+	<-time.After(rt.runtimeConfig.GracefulShutdownDuration)
+	assert.Nil(t, rt.pubsubCtx)
+	assert.Nil(t, rt.topicCtxCancels)
```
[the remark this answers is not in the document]
akhilac1: Close method in the mockPubSub is not invoked as expected and this needs additional troubleshooting. Since this method in mockpubsub is not working as expected [fails invocation with no arguments], I did not include it here.

pkg/runtime/runtime_test.go
akhilac1: Agree. To reduce flakiness in Windows switched from SIGTERM to invoking Shutdown

pkg/runtime/runtime_test.go
akhilac1: Removed SIGTERM and invoking Shutdown directly to mitigate flakiness of the tests in Windows environment

pkg/runtime/runtime.go
```diff
@@ -2620,10 +2623,10 @@ func (a *DaprRuntime) Shutdown(duration time.Duration) {
 		}
 	}()
 
-	log.Infof("Waiting %s to finish outstanding operations", duration)
-	<-time.After(duration)
+	log.Info("Shutting down tracing provider")
```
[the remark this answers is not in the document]
akhilac1: The intent is to indicate the components that are being shutdown. Agree that this is only initiation. Updated the log message to indicate the same

pkg/runtime/runtime.go
```diff
@@ -2620,10 +2623,10 @@ func (a *DaprRuntime) Shutdown(duration time.Duration) {
 		}
```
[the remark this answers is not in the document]
akhilac1: Shutdown for trace is called with a background context and I have tested it to close neatly even when no output components were setup.

pkg/runtime/runtime_test.go
```diff
@@ -5392,3 +5371,183 @@ func TestNamespacedPublisher(t *testing.T) {
...
+	fakeResp := invokev1.NewInvokeMethodResponse(200, "OK", nil)
+	fakeResp.WithRawData(sub, "application/json")
+
+	mockAppChannel := new(channelt.MockAppChannel)
+	rt.appChannel = mockAppChannel
+	mockAppChannel.On("InvokeMethod", mock.MatchedBy(matchContextInterface), req).Return(fakeResp, nil)
+
+	require.NoError(t, rt.initPubSub(pubsubComponent))
+	rt.startSubscriptions()
+	assert.NotNil(t, rt.pubsubCtx)
```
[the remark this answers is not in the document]
akhilac1: Done

yaron2 (approved):

pkg/runtime/runtime_test.go
mukundansundar:
```suggestion
	<-time.After(rt.runtimeConfig.GracefulShutdownDuration + 2*time.Second)
	assert.Nil(t, rt.pubsubCtx)
	assert.Nil(t, rt.topicCtxCancels)
	assert.Nil(t, rt.topicRoutes)
```

IMO the main logic that needs to be tested here is that Pubsub subscriptions and input bindings  are closed before the `GracefulShutdownDuration`. But the assertions only happen 2 seconds after graceful shutdown period. 

The logic that has been changed is the shut down sequence and the order of operations matter.
Similarly before the graceful shut down, we can verify that the api closers are not called.

pkg/runtime/runtime.go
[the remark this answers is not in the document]
ItalyPaleAle: The problem isn't the background context (whcih is just a "cancellation token") but that there's not much time for this to work in background now, since there's no "time.After" after this anymore. So it's like a "race against time".

Maybe move it out of a background goroutine?

pkg/runtime/runtime.go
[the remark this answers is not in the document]
akhilac1: Updated to indicate the initiation of shutdown to maintain continuity of logs during shutdown

pkg/runtime/runtime.go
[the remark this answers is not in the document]
ItalyPaleAle: It's at least a possibility, yes. If you keep it in background, I'd add a wait group with a delay. Or if you want to move it to foreground, maybe add a timeout to the context.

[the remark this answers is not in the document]
akhilac1: The current implementation in runtime does not capture tracerProvider post registry. All validations related to trace shutdown require this handler.Fixed the trace interface method, trace initiation, close and added tests to validate closure of trace.

https://github.com/dapr/dapr/blob/4d350a72191b113ea43741f3a9f639fc2f4b20d7/pkg/runtime/runtime.go#L409-L411


 ForceFlush() helps push out all the completed spans before closure of the trace. Earlier this shutdown was dead code since a.tracerProvider was always nil. The shutdown method would always fail with context.Cancelled error. Triggering ForceFlush and Shutdown with a.ctx helps give more time for the trace to gracefully exit. Shutdown method is deferred and not called until exit to allow forceFlush and shutdown to complete.

https://github.com/dapr/dapr/blob/4d350a72191b113ea43741f3a9f639fc2f4b20d7/pkg/runtime/runtime.go#L2615-L2630

@yaron2 @ItalyPaleAle @mukundansundar

akhilac1:
@ItalyPaleAle @mukundansundar - Pinging for attention

pkg/runtime/runtime.go
```diff
@@ -2649,32 +2648,42 @@ func (a *DaprRuntime) cleanSocket() {
...
-			log.Warnf("error closing API: %v", err)
-		}
-	}
 
-	shutdownCtx, shutdownCancel := context.WithCancel(context.Background())
+	// Flush and shutdown the tracing provider.
+	shutdownCtx, shutdownCancel := context.WithCancel(a.ctx)
+	defer shutdownCancel()
 	go func() {
 		if a.tracerProvider != nil {
```
ItalyPaleAle, line 2663: I don't think we can stop the trace provider *before* the APIs, otherwise we'll be losing traces from the API calls that come in after.

ItalyPaleAle: Still think the trace provider should be stopped synchronously (not in a goroutine), with a short timeout, afterthe APIs

pkg/runtime/runtime_test.go
```diff
@@ -5396,3 +5373,191 @@ func TestNamespacedPublisher(t *testing.T) {
...
+
+	// assert
+	assert.NoError(t, err, "expected no error")
+
+	rt.namespace = "test"
+	rt.runtimeConfig.mtlsEnabled = true
+	assert.Nil(t, rt.initActors())
+
+	go sendSigterm(rt)
+	<-time.After(rt.runtimeConfig.GracefulShutdownDuration + 300*time.Second)
```
ItalyPaleAle, line 5508: If we're stopping actors right away, why waiting 300 seconds (5 minutes)?

pkg/runtime/runtime_test.go
```diff
@@ -5396,3 +5373,191 @@ func TestNamespacedPublisher(t *testing.T) {
...
+	cout.ObjectMeta.Name = "testOutputBinding"
+	cout.Spec.Type = "bindings.testOutputBinding"
+	require.NoError(t, rt.initInputBinding(cin))
+	require.NoError(t, rt.initOutputBinding(cout))
+
+	assert.Equal(t, len(rt.inputBindings), 1)
+	assert.Equal(t, len(rt.outputBindings), 1)
+
+	go sendSigterm(rt)
+	<-time.After(rt.runtimeConfig.GracefulShutdownDuration)
```
ItalyPaleAle, line 5455: May want to add 2-3 seconds as buffer

ItalyPaleAle (changes requested):

pkg/runtime/runtime_test.go
akhilac1: @ItalyPaleAle - We are initiating binding shutdown before graceperiod and expect this to be closed within graceperiod. I have increased the grace period to 5sec from 3 sec to cover for the additional wait. This should address the additional buffer for bindings shutdown.

ItalyPaleAle: That’s fine, thanks

pkg/runtime/runtime.go
akhilac1: https://github.com/dapr/dapr/blob/1f75b16a21344188d41dd9d7ec39b87a8df50cd4/pkg/runtime/runtime.go#L2648-L2671

akhilac1: @ItalyPaleAle - this comment is resolved. Can you please check and mark this as resolved? Thanks.

pkg/runtime/runtime_test.go
```diff
@@ -5400,3 +5377,191 @@ func TestNamespacedPublisher(t *testing.T) {
...
+	require.NoError(t, rt.setupTracing(rt.hostAddress, tpStore))
+	assert.NotNil(t, rt.tracerProvider)
+
+	go sendSigterm(rt)
+	<-rt.ctx.Done()
+	assert.Nil(t, rt.tracerProvider)
+}
+
+func sendSigterm(rt *DaprRuntime) {
+	rt.runtimeConfig.GracefulShutdownDuration = 5 * time.Second
```
ItalyPaleAle, line 5565: Can you change this to 2? Just so tests end quicker.

ItalyPaleAle (approved):

mukundansundar (approved):
lgtm

Issue #5472, opened 2022-11-07
Add namespaced topics to pub/sub namespaced consumer groups


yaron2:
The namespaced consumer groups feature today allows a group of applications in a namespace to compete for messages sent in a given namespace by prefixing the consumer groups with the deployed namespace.

While this works and does the right thing (messages for a given topic will hit multiple consumer groups per namespace instead of a shared consumer group), the topics that are subscribed and published to need to also be namespaced. This is in order to prevent applications in one namespace potentially receiving messages sent to apps in different namespaces.

PR #5564, opened 2022-11-25
Add missing topics to namespaced consumer groups

yaron2:
Fixes https://github.com/dapr/dapr/issues/5472.

### DIFF
--- a/pkg/runtime/runtime.go
+++ b/pkg/runtime/runtime.go
@@ -722,20 +722,23 @@ func (a *DaprRuntime) subscribeTopic(parentCtx context.Context, name string, top
 	policy := a.resiliency.ComponentInboundPolicy(ctx, name, resiliency.Pubsub)
 	routeMetadata := route.metadata
 
-	subscribeTopic := topic
-	if a.pubSubs[name].namespaceScoped {
-		subscribeTopic = a.namespace + topic
-	}
+	namespaced := a.pubSubs[name].namespaceScoped
 
 	if utils.IsTruthy(routeMetadata[BulkSubscribe]) {
-		err := a.bulkSubscribeTopic(ctx, policy, name, subscribeTopic, route)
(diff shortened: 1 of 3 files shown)


pkg/runtime/runtime.go
```diff
@@ -1873,6 +1886,11 @@ func (a *DaprRuntime) Publish(req *pubsub.PublishRequest) error {
 	}
 
 	policy := a.resiliency.ComponentOutboundPolicy(a.ctx, req.PubsubName, resiliency.Pubsub)
+
+	if a.pubSubs[req.PubsubName].namespaceScoped {
```
Taction, line 1890: Please use `if ps.namespaceScoped {`, this map's read is under RLock.

yaron2: Done

Taction:
Generally looks good to me, please check if `bulkSubscribeTopic` needs to remove the namespace prefix when sending messages to app. Also, it is best to have e2e tests to ensure these namespaced subscribe behaviors.

yaron2:
> Generally looks good to me, please check if `bulkSubscribeTopic` needs to remove the namespace prefix when sending messages to app. Also, it is best to have e2e tests to ensure these namespaced subscribe behaviors.

Thanks, and you're right, this feature was enabled before bulk pub/sub so `bulkSubscribeTopic` needed an update, which I've now added. an e2 test for both namespaced publishers and subscribers is being worked on separately.

Taction:
Another thing needs to be considered: does the dead-letter topic need to add a namespace prefix?

yaron2:
> Another thing needs to be considered: does the dead-letter topic need to add a namespace prefix?

I see namespaced DLTs as almost a separate sub feature. It's optional, and I'm leaning towards adding support for it, but shouldn't be part of this PR.

Taction (approved):
Thanks for this pr and the explanation, LGTM.

PR #5566, opened 2022-11-25
Revert self-hosted AppID validation and only disallow dots

shubham1172:

# Description

This PR reverts the logic introduced in #5556 to validate self-hosted app-ids with RFC-1123. It only throws an error if the app-id contains a period (`.`).

A major reason is that upper-case in app-ids was no longer allowed after #5556, which broke the case-sensitive implementations in state store (app-id is prefixed) and pub/sub (consumer groups use app-id).

~~From #5556, we introduced stricter rules for appID validation in self-hosted mode. This PR relaxes the case restrictions by transforming the appID into a lowercase string. This will prevent breaking existing Dapr integrations where users are using appIDs with uppercase characters.~~

## Issue reference

Please reference the issue this PR will close: #5552

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing

### DIFF
--- a/pkg/runtime/cli.go
+++ b/pkg/runtime/cli.go
@@ -125,8 +125,16 @@ func FromFlags() (*DaprRuntime, error) {
 		os.Exit(0)
 	}
 
-	if err := validation.ValidateSelfHostedAppID(*appID); err != nil {
-		return nil, err
+	if *mode == string(modes.KubernetesMode) {
+		if err := validation.ValidateKubernetesAppID(*appID); err != nil {
+			return nil, err
+		}
+	} else {
+		// Convert appID to lowercase to be compatible with the validation.
+		*appID = strings.ToLower(*appID)
+		if err := validation.ValidateSelfHostedAppID(*appID); err != nil {
+			return nil, err
+		}
 	}
 
 	// Apply options to all loggers
--- a/pkg/runtime/cli.go
+++ b/pkg/runtime/cli.go
@@ -126,11 +126,6 @@ func FromFlags() (*DaprRuntime, error) {
 	}
 
 	if *mode == string(modes.StandaloneMode) {
-		appIDLower := strings.ToLower(*appID)
-		if *appID != appIDLower {
-			log.Warnf("app-id should be lowercase, using %s instead", appIDLower)
-			*appID = appIDLower
-		}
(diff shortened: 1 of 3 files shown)


shubham1172:
FYI @ItalyPaleAle @mukundansundar @artursouza

ItalyPaleAle (approved):

pkg/runtime/cli.go
```diff
@@ -125,8 +125,16 @@ func FromFlags() (*DaprRuntime, error) {
 		os.Exit(0)
 	}
 
-	if err := validation.ValidateSelfHostedAppID(*appID); err != nil {
-		return nil, err
+	if *mode == string(modes.KubernetesMode) {
+		if err := validation.ValidateKubernetesAppID(*appID); err != nil {
+			return nil, err
```
mukundansundar, line 130: this call is not needed here ... it is done during pod injection.

yaron2:
I have concerns this might break subscriptions due to case sensitive pub/sub implementations. Need to look into this further.

pkg/runtime/cli.go
```diff
@@ -125,8 +125,16 @@ func FromFlags() (*DaprRuntime, error) {
...
+		}
+	} else {
+		// Convert appID to lowercase to be compatible with the validation.
+		*appID = strings.ToLower(*appID)
+		if err := validation.ValidateSelfHostedAppID(*appID); err != nil {
+			return nil, err
```
mukundansundar, line 136: should we warn as this might lead to some confusion for users using Self Hosted mode if the AppID changes without info.

shubham1172: Yes, we can do that. Let me add it.

ItalyPaleAle (approved):

yaron2 (approved):

ItalyPaleAle:
> We should add new E2E test scenarios to make sure things break on breaking changes.

We do not have E2E scenarios for self-hosted currently :(

yaron2:
> > We should add new E2E test scenarios to make sure things break on breaking changes.
> 
> We do not have E2E scenarios for self-hosted currently :(

We don't, but that's a point in time situation. It actually would be easier to orchestrate self hosted tests than our k8s e2e tests. I'm just wondering if the kind of test to catch something like this should actually be done on the components side or on the daprd side.

Issue #5558, opened 2022-11-23
Add actor perf/stress test: Scale actor ids of a single type


mcandeia:
## In what area(s)?

> /area runtime

> /area operator

> /area placement

> /area docs

/area test-and-release

### Introduction

Measure performance metrics* for a single app with many pods hosting a single actor, receiving concurrent requests for the different actors using a huge number of IDs

*Metrics:

- CPU/Memory of clients
- CPU/Memory of clients' sidecars
- CPU/Memory of servers
- CPU/Memory of servers' sidecars
- Request latency from client perspective
- Number of errors per type of error

### Environment

- Up to five clients
- Up to ten pods of the same app each containing the single actor
- Up to 1k QPS with randomized ID

### Test scenario

<img width="1195" alt="image" src="https://user-images.githubusercontent.com/5839364/203607971-dc2c89a1-3f72-47be-9a3a-4c8fe503660c.png">

- The app will be a `golang` implemented without the actors SDK - to not get biased by the SDK overhead/performance.
- The actor will be an "echo" actor, receive and return the same as received

### Expected Numbers

There's no previous execution for such scenario, so the numbers should be an estimation based on the current numbers with a 5% variation allowed.

Issue #5550, opened 2022-11-22
Actor perf/stress tests: Scale actor types - single app 


mcandeia:
## In what area(s)?

/area runtime

> /area operator

> /area placement

> /area docs

> /area test-and-release

### Introduction

Measure performance metrics* for a single app with many pods hosting a huge number of actors (500+), receiving concurrent requests for the different actors using different IDs

*Metrics:

- CPU/Memory of clients
- CPU/Memory of clients' sidecars
- CPU/Memory of servers
- CPU/Memory of servers' sidecars
- Request latency from client perspective
- Number of errors per type of error

### Environment

- Up to five clients
- Up to ten service applications each containing the 500+ actors
- Up to 1k QPS

### Test scenario

<img width="1113" alt="image" src="https://user-images.githubusercontent.com/5839364/203383781-acff18de-e23b-41ea-ba02-b43c3cfb1288.png">

- The app will be a `golang` implemented without the actors SDK - to not get biased by the SDK overhead/performance.
- The actor will be an "echo" actor, receive and return the same as received

### Expected Numbers

There's no previous execution for such scenario, so the numbers should be a estimation based on the current numbers with a 5% variation allowed.

PR #5569, opened 2022-11-25
Actor ID and Type stress tests

mcandeia:
# Description

This pr adds stress tests for actors based on actor ID distribution. I've found a latency increase when used a high amount of distinct ids, further investigation would be necessary

## Issue reference

Please reference the issue this PR will close: #5558  and #5550

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing


.devcontainer/devcontainer.json
```diff
@@ -61,6 +61,11 @@
 		"go.useLanguageServer": true,
 		"go.gopath": "/go"
 	},
+    "hostRequirements": {
```
ItalyPaleAle, line 64: Nit: this uses spaces but the rest of the file seems to use tabs?

tests/perf/actor_id_scale/test.js
```diff
@@ -0,0 +1,55 @@
+import http from "k6/http";
```
ItalyPaleAle, line 1: Should this file be included with `go:embed`? Unless you add a custom Dockerfile, I believe we just compile the Go app and don't have a way to include other files in the container. Unless I'm wrong

.devcontainer/devcontainer.json
ItalyPaleAle, line 64: Is this strictly necessary?

The problem is that this will make Codespaces only use the largest VMs, which are expensive for our users unless they're really needed

tests/perf/actor_id_scale/test.js
ItalyPaleAle, line 1: I don't see a package.json. Or k6 doesn't need it?

tests/runner/loadtest/k6.go
```diff
@@ -152,10 +152,14 @@ type K6 struct {
 	testMemoryRequest string
 	daprMemoryLimit   string
 	daprMemoryRequest string
+	logEnabled        bool
 }
 
 // collectResult read the pod logs and transform into json output.
 func collectResult[T any](k6 *K6, podName string) (*T, error) {
+	if k6.logEnabled {
```
ItalyPaleAle, line 160: Should this have been `!k6.logEnabled` maybe?

ItalyPaleAle (changes requested):

tests/perf/actor_id_scale/test.js
mcandeia: The k6 javascript runtime is backed by https://github.com/dop251/goja,

Since this script is not aimed to be backed by a nodejs runtime or a headless browser, you can't use functions like from `window` `os`, `fs`, etc. When you need to include libraries/dependencies you have to bundle it all together using a bundling tool like webpack or something similar.

The imports that I'm using are all [provided by k6 in runtime](https://k6.io/docs/javascript-api/), that's why I don't need any node_modules nor package.json, because I don't depend on external libraries. btw you can also use remote module import https://k6.io/docs/using-k6/modules/#the-jslib-repository

tests/runner/loadtest/k6.go
mcandeia: Not really, the log is disabled by default to not mess up with the test summary output which is sent straight to the stdout.

So when the log is enabled I'm assuming that the summary might not be well-formatted

tests/perf/actor_id_scale/test.js
mcandeia: Thanks for introducing me the `go:embed`, very interesting.

But I don't think we do need to embed the test file because it's being used by the test files only, so no dockerfile is needed,

Even considering that we don't need a dockerfile/compile go code, I have to change the k6 runner from receiving a script name to (probably) receive a bytes array - because the k6.go code is the one that read the file content which is dynamic. I feel that this "optimization" is unnecessary for the usecase of running tests, what do you think?

ItalyPaleAle: Oh you’re right. I thought this was a test app but instead it’s a test case, so you’re right there’s no need for embeds.

Don’t worry about optimizations in test cases :)

ItalyPaleAle: Perfect, thanks for explaining!

ItalyPaleAle (approved):

tests/apps/perf/actorfeatures/app.go
```diff
@@ -46,23 +47,28 @@ type daprConfig struct {
...
+		registeredActorType[defaultActorType] = true
+		return []string{defaultActorType}
 	}
 
-	return actorType
+	actorTypes := strings.Split(actorType, ",")
+	for _, tp := range actorTypes {
+		registeredActorType[tp] = true
+	}
+	return actorTypes
```
tmacam, line 71: I am assuming there is no concurrency concern here because the worst that could happen is a race condition to set the same key to true, correct?

tests/perf/actor_id_scale/actor_id_scale_test.go
```diff
@@ -0,0 +1,107 @@
...
+	}
+
+	tr = runner.NewTestRunner("actoridstress", testApps, nil, nil)
+	os.Exit(tr.Start(m))
+}
+
+func TestActorIdStress(t *testing.T) {
+	// Get the ingress external url of test app
+	testServiceAppURL := tr.Platform.AcquireAppExternalURL(serviceApplicationName)
+	require.NotEmpty(t, testServiceAppURL, "test app external URL must not be empty")
```
tmacam, line 75: TIL about require! I was baffled that `assert` would just proceed with test execution and that was "normal". Thanks.

tests/perf/actor_id_scale/test.js
```diff
@@ -0,0 +1,55 @@
...
+import { check } from "k6";
+import exec from "k6/execution";
```
tmacam, line 16: nit: why the inconsistency?

tests/perf/actor_id_scale/test.js
```diff
@@ -0,0 +1,55 @@
...
+
+const defaultMethod = "default"
+const actorType = __ENV.ACTOR_TYPE
```
tmacam, line 19: ALL_CAPS?

tests/perf/actor_id_scale/test.js
```diff
@@ -0,0 +1,55 @@
...
+
+const DAPR_ADDRESS = `http://127.0.0.1:${__ENV.DAPR_HTTP_PORT}/v1.0`;
+
+function callActorMethod(id, method) {
+  return http.put(
+    `${DAPR_ADDRESS}/actors/${actorType}/${id}/method/${method}`,
+    JSON.stringify({})
+  );
+}
+export default function () {
```
tmacam, line 49: TIL about `export default`

tests/perf/actor_id_scale/test.js
ItalyPaleAle: That's how ESM imports work.

`import { check } from 'k6'` -> dereferences the object exported by k6 (so k6 probably has many things like `export function check()...` and other exports)

`import exec from k6/execution` -> imports the default export from the file.

ItalyPaleAle (dismissed):

tests/perf/actor_id_scale/test.js
tmacam: I had to check [this on MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import). This PR has been a series of TIL.

tests/apps/perf/actorfeatures/app.go
```diff
@@ -46,23 +47,28 @@ type daprConfig struct {
...
 }
```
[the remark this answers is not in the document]
mcandeia: Yes. this method is called only once at the startup time

tests/perf/actor_id_scale/actor_id_scale_test.go
```diff
@@ -0,0 +1,107 @@
+//go:build perf
+// +build perf
```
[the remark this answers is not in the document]
mcandeia: they need to, as the others because it is a flag for the go compiler

tests/perf/actor_id_scale/test.js
mcandeia: it is common to have env vars as caps. ps: the __ENV is from the k6 framework

Issue #5571, opened 2022-11-28
[Discussion] Why Dapr runtine subscribe to topics 1 by 1?


DeepanshuA:
In Dapr runtime, we see that for a particular PubSub component, if there are n topics to be subscribed, then these calls are made 1 by 1, rather does it make sense, if the component is made aware of topics to be subscribed along with respective metadata in 1 go.

https://github.com/dapr/dapr/blob/5a17813a159ec68965c58406a0915d74a3ada2c5/pkg/runtime/runtime.go#L904

Right now, due to multiple calls, the resources vested in subscribing topics are cancelled/closed and are re-opened with an additional topic to subscribe, which seems like an overhead.
Rather, runtime should take necessary steps to take care of at its end and inform component about topics to be subscribed at once, along with respective metadata.

Along with overhead, following scenarios can definitely keep occuring in components:

Start subscribers S1, S2 and S3:::::: S1 -> S2 -> S3

Topics are formed

Stop subscribers

Publish Messages on these topics S1, S2 and S3

Start back subscribers. 

- Subscribe for S1 is called.
	- Added to topics list.
	- Create connection and S1 receives pending message.
- Subscribe for S2 is called.
	- Add to topic List. Now, there are S1 and S2 topics to be subscribed. Disconnection for S1 is triggered before any ACK could be sent for pending message.
	- Create back connection and subscribe for S1 and S2. Pending message(s) received for S1 and S2.
- Subscribe for S3 is called.
	- Add to topic list. Now, there are S1, S2 and S3 topics to be subscribed. Disconnection for S1 and S2 is triggered before any ACK could be sent for pending message.
	- Create connection again and subscribe for S1, S2 and S3. Pending message(s) received for S1, S2 and S3.

- In all, 3 times pending message received for S1, 2 times for S2 and 1 time for S3.

Thoughts?

P.S.: I will try to take out some more time tomorrow and show with some visual diagrams, in case description above doesn't create clarity.

ItalyPaleAle:
> > Right now, due to multiple calls, the resources vested in subscribing topics are cancelled/closed and are re-opened with an additional topic to subscribe, which seems like an overhead.
>
> That's an unfortunate implementation detail for the components that do this. There are components that do the "right thing".

I rewrote how this works in each component and can say that it depends on the component and what the SDKs support.

[...]

PR #5031, opened 2022-08-18
Support for dynamic subscriptions

ItalyPaleAle:
# Description

As the last part of the journey that started with https://github.com/dapr/components-contrib/pull/1756 , added the ability to subscribe to and unsubscribe from topics on-demand.

This PR adds 3 new APIs:

- HTTP:
   - `GET  /v1.0/subscriptions` -> returns the list of topics (and routes, etc) the application is subscribed to
   - `POST /v1.0/subscriptions` -> subscribes to a new topic (on an existing pubsub component)
   - `DELETE /v1.0/subscriptions` -> unsubscribes from an existing topic subscription
- gRPC equivalents:
   - `ListActiveTopicSubscriptions`
   - `SubscribeTopic`
   - `UnsubscribeTopic`

Altogether, they allow developers to subscribe to new topics on-demand, and unsubscribe when not needed anymore.

The biggest concern here is that the list of active subscriptions is not persisted anywhere, so in case `daprd` crashed (and the app stayed active), that would be lost. This is how we are addressing that problem:

- Applications are recommended to continue implementing the callback to return the initial list of subscriptions (in the case of HTTP, that would be `/dapr/subscribe`), returning an empty list if appropriate
- Methods to subscribe to and unsubscribe from a topic return in the response the full list of active subscriptions. Thanks to the use of locks, no more than one concurrent call to subscribe/unsubscribe could happen at the same time, so it's guaranteed that the response will contain the exact list of topics the runtime is subscribed to at the time it's sent.
- Apps are recommended to maintain the list of subscriptions (ie. the response from the subscribe/unsubscribe methods) and serve that on subsequent invocations to `/dapr/subscribe`.
   - This is something that the developers need to do, but we could potentially add that to the Dapr SDKs
- If the runtime crashed, but not the app, when the runtime comes back online it will query the `/dapr/subscribe` endpoint (or equivalent gRPC) and the app should respond with the last list of subscriptions

## Issue reference

Fixes #814

## Docs

TODO @ItalyPaleAle 

## Checklist

Please make sure you've  completed the relevant tasks for this PR, out of the following list:

* [X] Code compiles correctly
* [X] Created/updated tests
* [X] Unit tests passing
* [X] End-to-end tests passing
* [ ] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_
* [ ] Specification has been updated / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_
[...]

PR #5572, opened 2022-11-28
Send flag for last subscription topic

DeepanshuA:

# Description

While analysing https://github.com/dapr/components-contrib/issues/2306, realized that there may be cases when a component wants to subscribe topics from broker at one go. For this, runtime should enable component with knowledge whether the current topic for which `Subscribe` is called is last topic in current list or not.

Ref: https://github.com/dapr/components-contrib/issues/2306#issuecomment-1328605232

## Issue reference

Please reference the issue this PR is related to: #5571  and https://github.com/dapr/components-contrib/issues/2306

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing


DeepanshuA:
> Overall LGTM but I wonder how this will interact with dynamic subscriptions and if it may be a problem See #5031

Sorry, coming back to it after couple of weeks, as I was trying to close some other items meanwhile.

I just went through the Dynamic loading issue and solution that you have proposed for it.

As far as I understand, that there are two types of component SDKs out there, which allow us two kind of subscription mechanisms:
1. which can Subscribe individually for each new topic 
2. which needs to keep a list of subscription, revoke old connection and invoke new connection with new list of topics.

For, 1st case, we need not bother and whatever flag this change is talking about, doesn't bother them.

For 2nd:
I feel with current approach, we would need to tell with each Dynamic Subscription that this is the topic which needs to be added and with each new topic addition/deletion, we need to re-connect. So, with each Dynamic Subscription we would need to send this `lastSubscriptionTopic`. But, this seems overhead.
Instead, it would be better if runtime just informs the component that it is a non-terminal topic subscription via metadata, then component can take a decision whether to use this metadata or not. And, for any other usecase like Dynamic, we need to not send any flag.

Hence, I am changing the flag from `lastSubscriptionTopic` and being sent with LAST topic of topics list
TO
`nonTerminalSubscriptionTopic` and being sent with each topic BUT last.
And, we need not send anything for Dynamic case.

cc @ItalyPaleAle

ItalyPaleAle:
Note that with the last changes to MQTT, the component maintain a single connection

mukundansundar:
@DeepanshuA is this PR still needed?

DeepanshuA:
> @DeepanshuA is this PR still needed?

I will check and get back on this.

PR #5582, opened 2022-11-29
Fix placement nil dereference when reading stream

mcandeia:

# Description

I've removed the nil assignment for the clientStream when closing the stream that was causing a nil dereference. This should not be a problem in terms of garbage collection, given the `closeStream` calling frequency. Also, calling CloseSend and Close twice does not panic, so it will not be a problema as well.

## Issue reference

Please reference the issue this PR will close: #5575

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Unit tests passing
* [] End-to-end tests passing

### DIFF
--- a/pkg/actors/internal/placement.go
+++ b/pkg/actors/internal/placement.go
@@ -181,12 +181,11 @@ func (p *ActorPlacement) Start() {
 			p.clientLock.RLock()
 			clientStream := p.clientStream
 			p.clientLock.RUnlock()
-
+			resp, err := clientStream.Recv()
 			if p.shutdown.Load() {
 				break
 			}
 
-			resp, err := clientStream.Recv()
 			// TODO: we may need to handle specific errors later.
 			if err != nil {
 				p.closeStream()
--- a/pkg/actors/internal/placement.go
+++ b/pkg/actors/internal/placement.go
@@ -181,11 +181,12 @@ func (p *ActorPlacement) Start() {
 			p.clientLock.RLock()
 			clientStream := p.clientStream
 			p.clientLock.RUnlock()
-			resp, err := clientStream.Recv()
+
 			if p.shutdown.Load() {
 				break
 			}
 
+			resp, err := clientStream.Recv()
 			// TODO: we may need to handle specific errors later.
 			if err != nil {
 				p.closeStream()
@@ -304,15 +305,8 @@ func (p *ActorPlacement) closeStream() {
 		p.clientLock.Lock()
 		defer p.clientLock.Unlock()
 
-		if p.clientStream != nil {
-			p.clientStream.CloseSend()
-			p.clientStream = nil
-		}
-
-		if p.clientConn != nil {
-			p.clientConn.Close()
-			p.clientConn = nil
-		}
+		p.clientStream.CloseSend()
+		p.clientConn.Close()
 	}()
 
 	func() {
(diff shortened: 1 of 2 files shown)


mcandeia:
please review @yaron2 @artursouza ,

Another interesting fact is that I think we are not handling the stream close very well. According to the [documentation](https://github.com/grpc/grpc-go/blob/be1fb4f27549f736b9b4ec26104c7c6b29845ad0/stream.go#L106) we should cancel context and recv until receiving an EOF or any error from header.

1) We are not calling context cancel, actually we are using a [background context](https://github.com/dapr/dapr/blob/master/pkg/actors/internal/placement.go#L372) 
2) We are not reading the stream until it ends

It might lead in memory leaks. I'll create an issue for further investigation soon, but as we are going to deploy this change by the EOD or tomorrow morning, I'd rather prefer to do it as a separate issue/task.

pkg/actors/internal/placement.go
```diff
@@ -181,11 +181,12 @@ func (p *ActorPlacement) Start() {
 			p.clientLock.RLock()
 			clientStream := p.clientStream
 			p.clientLock.RUnlock()
-			resp, err := clientStream.Recv()
+
 			if p.shutdown.Load() {
 				break
 			}
 
+			resp, err := clientStream.Recv()
```
ItalyPaleAle, line 189: why was this moved after the `break`?

pkg/actors/internal/placement.go
```diff
@@ -304,15 +305,8 @@ func (p *ActorPlacement) closeStream() {
 		p.clientLock.Lock()
```
ItalyPaleAle, line 305: Another (unrealted) thing is that we should switch from "uber/atomic" to the built-in atomic package in the stdlib

ItalyPaleAle (approved):

ItalyPaleAle (approved):

pkg/actors/internal/placement.go
mcandeia: before the current change it was necessary to avoid race conditions between Stop() and this, but now there are no reasons to keep this, so I'm going to put it back there, thanks

pkg/actors/internal/placement.go
```diff
@@ -181,11 +181,12 @@ func (p *ActorPlacement) Start() {
 			p.clientLock.RLock()
```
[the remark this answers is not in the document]
mcandeia: I agree, I think it is just a convenience way of using "defer" without needing to duplicate the `.Unlock` you know? but there are alternatives for sure.

pkg/actors/internal/placement.go
```diff
@@ -306,12 +306,10 @@ func (p *ActorPlacement) closeStream() {
 
 		if p.clientStream != nil {
 			p.clientStream.CloseSend()
-			p.clientStream = nil
 		}
 
 		if p.clientConn != nil {
 			p.clientConn.Close()
-			p.clientConn = nil
```
ItalyPaleAle, line 314: Now that the nil-check is back.... what happens if you call Close twice on the same clientConn?

mcandeia: Great question, I've tested it out because I had the same doubt, but there are no panics thanks to the [clientconn check](https://github.com/grpc/grpc-go/blob/master/clientconn.go#L1019-L1024)

yaron2 (approved):

PR #5584, opened 2022-11-30
Fixed: components that reference secrets cannot be initialized when `--disable-builtin-k8s-secret-store=true`

ItalyPaleAle:
When the flag `--disable-builtin-k8s-secret-store` is set to true, an issue is preventing components that reference secrets from being initialized. Those components fail with a message saying that the secret store `"kubernetes"` doesn't exist - even though the secrets should be fetched from the Dapr Operator and not the secret store itself.

### DIFF
--- a/pkg/runtime/runtime.go
+++ b/pkg/runtime/runtime.go
@@ -2459,11 +2459,6 @@ func (a *DaprRuntime) processComponentSecrets(component componentsV1alpha1.Compo
 		}
 
 		secretStoreName := a.authSecretStoreOrDefault(component)
-		secretStore := a.getSecretStore(secretStoreName)
-		if secretStore == nil {
-			log.Warnf("component %s references a secret store that isn't loaded: %s", component.Name, secretStoreName)
-			return component, secretStoreName
-		}
 
 		// If running in Kubernetes, do not fetch secrets from the Kubernetes secret store as they will be populated by the operator.
 		// Instead, base64 decode the secret values into their real self.
@@ -2489,6 +2484,12 @@ func (a *DaprRuntime) processComponentSecrets(component componentsV1alpha1.Compo
 
 			component.Spec.Metadata[i] = m
 			continue
+		}
+
+		secretStore := a.getSecretStore(secretStoreName)
+		if secretStore == nil {
+			log.Warnf("component %s references a secret store that isn't loaded: %s", component.Name, secretStoreName)
+			return component, secretStoreName
 		}
 
 		resp, ok := cache[m.SecretKeyRef.Name]


ItalyPaleAle:
Rebased onto release-1.9 per @artursouza

shubham1172 (commented):
Can we add a test case to `TestProcessComponentSecrets`?

shubham1172 (approved):

ItalyPaleAle:
I can look at adding one test case in master later on - it’s a method with lots of calls so it may take a while and would rather not delay this PR (for a hotfix) due to that.

I have validated that it fixes the error reported by the user, in the meanwhile.

shubham1172:
> I can look at adding one test case in master later on - it’s a method with lots of calls so it may take a while and would rather not delay this PR (for a hotfix) due to that.
> 
> I have validated that it fixes the error reported by the user, in the meanwhile.

Sure, sounds good, thanks @ItalyPaleAle!

shivamkm07 (approved):

pkg/runtime/runtime.go
```diff
@@ -2491,6 +2486,12 @@ func (a *DaprRuntime) processComponentSecrets(component componentsV1alpha1.Compo
 			continue
 		}
 
+		secretStore := a.getSecretStore(secretStoreName)
+		if secretStore == nil {
+			log.Warnf("component %s references a secret store that isn't loaded: %s", component.Name, secretStoreName)
+			return component, secretStoreName
+		}
+
```
mukundansundar, line 2494: with this change, the secret store is still considered unready, but the secret is resolved because it is loaded from operator? 
Does the init step complete with a secret store being not ready? 
Also for the case of builtin secret store, if the secret is being resolved by the operator, why are we warning here that the secret store is not loaded?

mukundansundar (commented):
Please add a comment for the function that it should set the return value string only for components that are not ready.

pkg/runtime/runtime.go
ItalyPaleAle: I don't understand, sorry.

The problem I was trying to solve is that if the component references a secret from the built-in K8s secret store, it should be read from the operator and not from a secret store component. This is explained by the existing comment:

```
// If running in Kubernetes, do not fetch secrets from the Kubernetes secret store as they will be populated by the operator.
```

The problem is when the built-in K8s secret store is not initialized. In this case, there's no component loaded caled `"kubernetes"` so `secretStore := a.getSecretStore(secretStoreName)` returns nil. But that's totally fine because that secret store isn't used anyways. So I had to move the check to after the case for using the built-in secret store.

mukundansundar: this function is being called in `preprocessOneComponent` function lines https://github.com/dapr/dapr/blob/master/pkg/runtime/runtime.go#LL2522-L2526, and when `kubernetes` secret store is not found, it will return that as an unready dependency. 
Does the initialization continue successfully after that, or does dapr runtime wait for this dependency to be ready? 

Also do we want to print the log warning `"component %s references a secret store that isn't loaded: %s" for the `kubernetes` secret store when we know that `--disable-builtin-k9s-secret-store` is given?

ItalyPaleAle: > Does the initialization continue successfully after that, or does dapr runtime wait for this dependency to be ready?

As long as the component only references secrets from the built-in secret store, initialization will continue successfully and Dapr won't need to wait.

> Also do we want to print the log warning "component %s references a secret store that isn't loaded: %s" for the kubernetessecret store when we know that--disable-builtin-k9s-secret-store` is given?

I would say that's not needed because component secrets are not loaded from the component but through the Dapr Operator. There's an expectation those will "just work" IMHO.

mukundansundar (approved):
lgtm

artursouza (approved):

Issue #5501, opened 2022-11-11
Add Resiliency Support for Bulk Subscribe


DeepanshuA:
## In what area(s)?

/area runtime
/area docs

## Describe the feature
In current bulk subscribe implementation, resiliency policy directly calls all messages for retry etc., whereas only failed messages should be subscribed again.

## Release Note

RELEASE NOTE:

PR #5603, opened 2022-12-03
Resiliency Support for Bulk Subscribe

DeepanshuA:

# Description

Till now, for bulk subscribe, all messages would be replayed under resiliency policy, irrespective of whether they passed earlier or not. Whereas, a Filter should be applied so that successful entries are not replayed.

## Issue reference

Please reference the issue this PR will close: #5501 

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing


utils/utils.go
```diff
@@ -160,3 +160,16 @@ func PopulateMetadataForBulkPublishEntry(reqMeta, entryMeta map[string]string) m
...
+	if isFilterNotApplicable {
+		return items
+	}
+	filteredItems := []T{}
+	for _, value := range items {
+		if fn(value) {
+			filteredItems = append(filteredItems, value)
+		}
+	}
+	return filteredItems
```
mukundansundar, line 174:
```suggestion
	filteredItems := make([]T, 0, len(items))
	for _, value := range items {
		if fn(value) {
			filteredItems = append(filteredItems, value)
		}
	}
	return filteredItems[:len(filteredItems):len(filteredItems)] // return and trunc capacity of slice
```

utils/utils.go
```diff
@@ -160,3 +160,16 @@ func PopulateMetadataForBulkPublishEntry(reqMeta, entryMeta map[string]string) m
 
 	return resMeta
 }
+
+func Filter[T any](items []T, fn func(item T) bool, isFilterNotApplicable bool) []T {
+	if isFilterNotApplicable {
```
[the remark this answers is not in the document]
DeepanshuA: Sure, that actually makes sense. What I had started to think was someone using `Filter` and forgetting about it, with some specific flags - but then instead of setting/re-setting the flag, they can chose to call/not-call Filter itself.

pkg/runtime/bulk_subscriber.go
```diff
@@ -9,6 +9,8 @@ import (
 	"strings"
 	"time"
 
+	"go.uber.org/atomic"
```
ItalyPaleAle, line 12: Please use sync/atomic instead

utils/utils.go
shubham1172: `return filteredItems[:len(filteredItems)] // return the truncated slice`

utils/utils_test.go
```diff
@@ -238,3 +238,27 @@ func TestPopulateMetadataForBulkPublishEntry(t *testing.T) {
 		assert.Equal(t, "22s", resMeta["ttl"], "expected val to be equal")
 	})
 }
+
+func TestFilter(t *testing.T) {
+	t.Run("should filter out empty values", func(t *testing.T) {
```
shubham1172, line 243: Please also add a test for capacity `cap`.

pkg/runtime/bulk_subscriber.go
```diff
@@ -259,80 +266,98 @@ func (a *DaprRuntime) createEnvelopeAndInvokeSubscriber(ctx context.Context, psm
...
+	_, err = policy(func(ctx context.Context) (any, error) {
+		if !firstAttempt.Load() {
+			newPubSubMsg := utils.Filter(psm.pubSubMessages, func(ps pubSubMessage) bool {
+				index := (*entryIdIndexMap)[ps.entry.EntryId]
+				return (*bulkResponses)[index].Error != nil
+			}, firstAttempt.Load())
+			psm.pubSubMessages = newPubSubMsg
+			psm.length = len(newPubSubMsg)
+		} else {
+			firstAttempt.CompareAndSwap(true, false)
```
shubham1172, line 288: This can be simplified to a `Store` instead of `CompareAndSwap`, right?

pkg/runtime/bulk_subscriber.go
```diff
@@ -259,80 +266,98 @@ func (a *DaprRuntime) createEnvelopeAndInvokeSubscriber(ctx context.Context, psm
...
 		Metadata: msg.Metadata,
 	})
+	var firstAttempt atomic.Bool
+	firstAttempt.Store(true)
+	_, err = policy(func(ctx context.Context) (any, error) {
+		if !firstAttempt.Load() {
+			newPubSubMsg := utils.Filter(psm.pubSubMessages, func(ps pubSubMessage) bool {
+				index := (*entryIdIndexMap)[ps.entry.EntryId]
+				return (*bulkResponses)[index].Error != nil
+			}, firstAttempt.Load())
```
shubham1172, line 284:
```suggestion
			}, false)
```

Or, rather we will remove this since the `Filter` method is changing.

pkg/runtime/bulk_subscriber.go
ItalyPaleAle: CAS is correct to use in this case to ensure the operation is actually atomic, but it shouldn't be done here because this can lead to race conditions: while operations like `Load` are performed atomically, there's no lock around them. CAS returns a boolean indicating if the swap happened and you can use that instead:

```go
// If the swap happened, returns true
if firstAttempt.CompareAndSwap(true, false) {
	newPubSubMsg := utils.Filter(psm.pubSubMessages, func(ps pubSubMessage) bool {
		index := (*entryIdIndexMap)[ps.entry.EntryId]
		return (*bulkResponses)[index].Error != nil
	}, false) // We can put false directly because we know it's false in this block
	psm.pubSubMessages = newPubSubMsg
	psm.length = len(newPubSubMsg)
}
```

pkg/runtime/bulk_subscriber.go
```diff
@@ -259,80 +266,98 @@ func (a *DaprRuntime) createEnvelopeAndInvokeSubscriber(ctx context.Context, psm
...
 
 	n := 0
-	for _, cloudEvent := range msg.cloudEvents {
+	for _, pubsubMsg := range psm.pubSubMessages {
+		cloudEvent := pubsubMsg.cloudEvent
 		if cloudEvent[pubsub.TraceIDField] != nil {
 			traceID := cloudEvent[pubsub.TraceIDField].(string)
 			sc, _ := diag.SpanContextFromW3CString(traceID)
-			spanName := fmt.Sprintf("pubsub/%s", msg.topic)
+			spanName := fmt.Sprintf("pubsub/%s", psm.topic)
```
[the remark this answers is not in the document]
shubham1172: Haha, I saw you refactoring other instances of `fmt.Sprintf` to `+` as well. Is it purely due to perf. reasons?

image

[[source](https://forum.golangbridge.org/t/fmt-sprintf-vs-string-concatenation/23006)]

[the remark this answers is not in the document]
ItalyPaleAle: Yes. It's consistent with benchmarks I ran too in another PR. When it's just about string concatenation in a hot path, the perf difference is significant. Not just the ns/op, but the fact that the second has 0 heap allocations, while the first one has a number (in my test, I got 3) - and that's more work for the allocator and the GC.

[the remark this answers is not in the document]
shubham1172: Good stuff! I wonder if it's worth adding it to our linter because it can get tedious to manually review it each time.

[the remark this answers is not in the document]
ItalyPaleAle: I just looked but couldn't see an option in our linters to catch these kinds of things. I don't think we should forbid `fmt.Sprintf` entirely because there are many situations where it's fine: in tests and in other non-hot paths, and/or when there are very complex concatenations (such as where the benefits in terms of code readability outweight the perf impact)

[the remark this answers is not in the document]
shubham1172: Yes, if we ever decide on configuring our linter, it should only be confined to simple concatenations.

pkg/runtime/bulk_subscriber.go
DeepanshuA: Actually, I had tried to look in that direction when your PR https://github.com/dapr/dapr/pull/5538 was raised, but issue is that in case of these Bulk APIs, we want to keep trying (say retries) decrementally i.e. omitting out all Successful entries and just want to keep re-try for Failed entries. And, that all becomes a part of Policy Only, But if we return from policy, we get outside Policy. 
How can we refactor so that we use return of `policy` but also remain inside it?

ItalyPaleAle: Ok, I see what you're trying to do now. I don't think this can be done without side effects, indeed. You should be able to make this work safely, even with side effects, by relying on atomic's and CAS. Ideally you will want to use [`atomic.Pointer`](https://pkg.go.dev/sync/atomic#Pointer) to keep a pointer to the last list of objects that need to be requested It's going to be a bit tricky though :)

utils/utils.go
mukundansundar: @ItalyPaleAle when i tried with this, I was getting some failures here ... when running multiple unit tests in series, even though the type is only slice and not pointer to slice, changing the values in place was affecting the source !

utils/utils.go
```diff
@@ -160,3 +160,16 @@ func PopulateMetadataForBulkPublishEntry(reqMeta, entryMeta map[string]string) m
 
 	return resMeta
 }
+
+func Filter[T any](items []T, fn func(item T) bool, isFilterNotApplicable bool) []T {
```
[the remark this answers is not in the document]
DeepanshuA: Done.

utils/utils.go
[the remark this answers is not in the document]
DeepanshuA: Removed the flag.

utils/utils.go
DeepanshuA: Used array with a capacity pre-defined. But, as discussed, not doing in-place.

pkg/runtime/bulk_subscriber.go
DeepanshuA: Done.

utils/utils.go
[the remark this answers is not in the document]
DeepanshuA: Done

pkg/runtime/bulk_subscriber.go
DeepanshuA: As discussed with @ItalyPaleAle and @mukundansundar, changed the implementation. Using Accumulator now from NewRunnerWithOptions.

pkg/runtime/bulk_subscriber.go
DeepanshuA: Yup, removed.

pkg/runtime/bulk_subscriber.go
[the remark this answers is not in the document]
DeepanshuA: Changed.

pkg/runtime/bulk_subscriber.go
```diff
@@ -459,20 +485,21 @@ func (a *DaprRuntime) publishBulkMessageGRPC(ctx context.Context, msg *pubsubBul
...
+	spans := make([]trace.Span, len(psm.pubSubMessages))
 	n := 0
-	for _, cloudEvent := range msg.cloudEvents {
+	for _, pubSubMsg := range psm.pubSubMessages {
+		cloudEvent := pubSubMsg.cloudEvent
 		if iTraceID, ok := cloudEvent[pubsub.TraceIDField]; ok {
 			if traceID, ok := iTraceID.(string); ok {
 				sc, _ := diag.SpanContextFromW3CString(traceID)
-				spanName := fmt.Sprintf("pubsub/%s", msg.topic)
+				spanName := fmt.Sprintf("pubsub/%s", psm.topic)
```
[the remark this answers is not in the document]
DeepanshuA: Done

pkg/runtime/bulksub_resiliency.go
```diff
@@ -0,0 +1,63 @@
...
+					index := (*bscData.entryIdIndexMap)[v.EntryId]
+					(*bscData.bulkResponses)[index].EntryId = v.EntryId
+					(*bscData.bulkResponses)[index].Error = v.Error
+				}
+				filteredPubSubMsgs := utils.Filter(*subscribedEntries.Load(), func(ps pubSubMessage) bool {
+					index := (*bscData.entryIdIndexMap)[ps.entry.EntryId]
+					return (*bscData.bulkResponses)[index].Error != nil
+				})
+				psm.pubSubMessages = filteredPubSubMsgs
+				psm.length = len(filteredPubSubMsgs)
```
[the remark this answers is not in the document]
DeepanshuA: 1. Since subscribedEntries load psm.pubSubMessages (`subscribedEntries.Store(&psm.pubSubMessages)`), which have already changed as in line above, so modified psm is loaded, instead of original psm. And, it is inside Accumulator, so it seems safe.
2. I am just thinking, if atomic.Pointer is still needed, since these modifications are getting done only inside Accumulator now.

DeepanshuA:
@ItalyPaleAle @shubham1172 @mukundansundar @dapr/approvers-dapr @dapr/maintainers-dapr Re-requesting review here. Whenever I re-request review from someone, some other re-review request gets removed, so requesting via comment.

utils/utils_test.go
shubham1172: @DeepanshuA

pkg/runtime/runtime.go
```diff
@@ -852,6 +850,9 @@ func (a *DaprRuntime) subscribeTopic(parentCtx context.Context, name string, top
 			path:       routePath,
 			pubsub:     name,
 		}
+		policyRunner := resiliency.NewRunner[any](ctx,
+			a.resiliency.ComponentInboundPolicy(name, resiliency.Pubsub),
```
ItalyPaleAle, line 854: Isn't this the same as policyDef (from L730)?

pkg/runtime/bulksub_resiliency.go
```diff
@@ -0,0 +1,63 @@
...
+	_, err := policyRunner(func(ctx context.Context) (*[]pubsub.BulkSubscribeResponseEntry, error) {
+		var pErr error
+		bsre := []pubsub.BulkSubscribeResponseEntry{}
+		switch a.runtimeConfig.ApplicationProtocol {
+		case HTTPProtocol:
+			pErr = a.publishBulkMessageHTTP(&bscData, &psm, &bsre, envelope, deadLetterTopic)
+		case GRPCProtocol:
+			pErr = a.publishBulkMessageGRPC(&bscData, &psm, &bsre, rawPayload)
+		default:
+			pErr = backoff.Permanent(errors.New("invalid application protocol"))
```
ItalyPaleAle, line 51: Nit: you may want to `panic` here as this should never happen, and if it happens, it should be caught during development

pkg/runtime/bulksub_resiliency.go
```diff
@@ -0,0 +1,63 @@
...
+	bscData := *bulkSubCallData
+	ctx := bscData.ctx
+	var subscribedEntries atomic.Pointer[[]pubSubMessage]
+	subscribedEntries.Store(&psm.pubSubMessages)
+	policyRunner := resiliency.NewRunnerWithOptions(
+		ctx, policyDef, resiliency.RunnerOpts[*[]pubsub.BulkSubscribeResponseEntry]{
+			Accumulator: func(bsre *[]pubsub.BulkSubscribeResponseEntry) {
+				for _, v := range *bsre {
+					// add to main bulkResponses
+					index := (*bscData.entryIdIndexMap)[v.EntryId]
```
ItalyPaleAle, line 30: Is this guaranteed to be found, or is there a chance that index will be the zero value and so `(*bscData.bulkResponses)[index]` would be nil? should there be a nil-check?

pkg/runtime/bulksub_resiliency.go
```diff
@@ -0,0 +1,63 @@
...
+// ApplyBulkSubscribeResiliency applies resiliency support to bulk subscribe. It tries to filter
+// out the messages that have been successfully processed and only retries the ones that have failed
+func (a *DaprRuntime) ApplyBulkSubscribeResiliency(bulkSubCallData *bulkSubscribeCallData,
+	psm pubsubBulkSubscribedMessage, deadLetterTopic string, path string, policyDef *resiliency.PolicyDefinition,
+	rawPayload bool, envelope map[string]interface{},
+) (*[]pubsub.BulkSubscribeResponseEntry, error) {
+	bscData := *bulkSubCallData
+	ctx := bscData.ctx
+	var subscribedEntries atomic.Pointer[[]pubSubMessage]
+	subscribedEntries.Store(&psm.pubSubMessages)
```
ItalyPaleAle, line 24: Why is this using atomic? I see it's only used in the accumulator, so it shouldn't need to be atomic anymore, unless I misunderstand

ItalyPaleAle:
@DeepanshuA  FYI I have merged changes from master into this PR. Because of #5648, the `invokev1` package changed a lot and didn't want to put the burden on you to fix the merge conflicts :) (and it also helped me find a bug!)

pkg/runtime/runtime.go
DeepanshuA: Yeah, correct, I will use the already defined policyDef here.

pkg/runtime/bulksub_resiliency.go
DeepanshuA: Sure, changing that in bulk Subscribe, though not touching same logic in runtime `subscribe` functionality and letting that remain same.

pkg/runtime/bulksub_resiliency.go
DeepanshuA: Yeah, this is guaranteed to be found. But, still it doesn't harm to put an extra check here, so putting an explicit check here as well.

pkg/runtime/bulksub_resiliency.go
DeepanshuA: Yeah, correct - atomic shouldn't be required - changing that.

utils/utils_test.go
DeepanshuA: I have added cap assertions in the test cases, please see if that helps.

shubham1172: Looks good, thanks @DeepanshuA!

DeepanshuA:
go test -race results:

/usr/local/go/bin/go test -race -run ^TestBulkSubscribeResiliency$ github.com/dapr/dapr/pkg/runtime
ok      github.com/dapr/dapr/pkg/runtime        28.265s

/usr/local/go/bin/go test -race -run ^TestBulkSubscribeResiliencyWithLongRetries$ github.com/dapr/dapr/pkg/runtime
ok      github.com/dapr/dapr/pkg/runtime        12.202s

/usr/local/go/bin/go test -race -run ^TestBulkSubscribeResiliencyStateConversionsFromHalfOpen$ github.com/dapr/dapr/pkg/runtime
ok      github.com/dapr/dapr/pkg/runtime        13.232s

pkg/runtime/bulk_subscriber.go
```diff
@@ -429,21 +458,19 @@ func (a *DaprRuntime) publishBulkMessageHTTP(ctx context.Context, msg *pubsubBul
...
-	bulkSubDiag.statusWiseDiag[string(pubsub.Retry)] += int64(len(msg.entries))
-	bulkSubDiag.elapsed = elapsed
-	populateBulkSubscribeResponsesWithError(
-		msg.entries, bulkResponses, &entryIdIndexMap,
-		fmt.Errorf("retriable error returned from app while processing bulk pub/sub event, topic: %v. status code returned: %v", msg.topic, statusCode))
-	return fmt.Errorf("retriable error returned from app while processing bulk pub/sub event, topic: %v. status code returned: %v", msg.topic, statusCode)
+	log.Warnf("Retriable error returned from app while processing bulk pub/sub event, topic: %v. status code returned: %v", psm.topic, statusCode)
+	bscData.bulkSubDiag.statusWiseDiag[string(pubsub.Retry)] += int64(len(rawMsgEntries))
+	bscData.bulkSubDiag.elapsed = elapsed
+	populateBulkSubscribeResponsesWithError(psm, bscData.bulkResponses, fmt.Errorf("retriable error returned from app while processing bulk pub/sub event, topic: %v. status code returned: %v", psm.topic, statusCode))
```
mukundansundar, line 472: extract as error var and reuse in line below.

pkg/runtime/bulk_subscriber.go
```diff
@@ -646,32 +688,32 @@ func endSpans(spans []trace.Span) {
...
 
-	if bulkResponses == nil {
+	if sendAllEntries {
 		data = msg.Entries
 	} else {
 		n := 0
 		for _, message := range msg.Entries {
-			entryId := (*entryIdIndexMap)[message.EntryId] //nolint:stylecheck
-			if (*bulkResponses)[entryId].Error != nil {
+			entryId := (*bscData.entryIdIndexMap)[message.EntryId] //nolint:stylecheck
```
mukundansundar, line 702: can there be scenario that the entryId is not present in the map?

DeepanshuA: No, the way it can reach this code-piece, entryIdIndexMap will have those entries.
That is taken care of separately when entries are sent by App, and if they are outside the known entries.

pkg/runtime/bulk_subscriber.go
DeepanshuA: Done.

pkg/runtime/bulk_subscriber.go
```diff
@@ -93,19 +104,27 @@ func (a *DaprRuntime) bulkSubscribeTopic(ctx context.Context, policy resiliency.
...
+			bulkSubDiag:     &bulkSubDiag,
+			entryIdIndexMap: &entryIdIndexMap,
+			psName:          psName,
+			topic:           topic,
+		}
 		rawPayload, err := contribMetadata.IsRawPayload(route.metadata)
 		if err != nil {
 			log.Errorf("error deserializing pubsub metadata: %s", err)
-			if dlqErr := a.sendBulkToDLQIfConfigured(ctx, psName, msg, route, nil, nil, &bulkSubDiag); dlqErr != nil {
+			if dlqErr := a.sendBulkToDLQIfConfigured(&bulkSubCallData, msg, true, route); dlqErr != nil {
```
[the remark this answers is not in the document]
DeepanshuA: Thanks, useful info. Changed accordingly.

pkg/runtime/bulk_subscriber.go
```diff
@@ -67,7 +78,7 @@ type bulkSubIngressDiagnostics struct {
 //     3.B. Send the envelope to app by invoking http endpoint
 //  4. Check if any error has occurred so far in processing for any of the message and invoke DLQ, if configured.
 //  5. Send back responses array to broker interface.
-func (a *DaprRuntime) bulkSubscribeTopic(ctx context.Context, policy resiliency.Runner[any],
+func (a *DaprRuntime) bulkSubscribeTopic(ctx context.Context, policyDef *resiliency.PolicyDefinition,
```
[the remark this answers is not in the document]
DeepanshuA: Changed comments as required.

pkg/runtime/bulk_subscriber.go
```diff
@@ -26,16 +25,19 @@ import (
 	runtimePubsub "github.com/dapr/dapr/pkg/runtime/pubsub"
 )
 
+type pubSubMessage struct {
+	cloudEvent map[string]interface{}
+	rawData    *runtimePubsub.BulkSubscribeMessageItem
+	entry      *pubsub.BulkMessageEntry
+}
```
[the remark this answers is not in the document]
DeepanshuA: Provided.

artursouza (approved):

artursouza:
Will merge after @ItalyPaleAle approves.

ItalyPaleAle (approved):
LGTM

PR #5609, opened 2022-12-06
Misc refactorings extracted from #5170

ItalyPaleAle:
This PR contains misc refactorings that are extracted from #5170. No new feature or capability was added. Refactorings include:

- Splitting the APIs part of DaprInternal (CallLocal, CallLocalActor) from pkg/grpc/api.go into a separate file
- Adding a few more tests and removing dead code from test files
- Improve code readability in some areas
- Minor improvements to allocation efficiency


pkg/actors/internal/placement.go
```diff
@@ -224,9 +225,15 @@ func (p *ActorPlacement) Stop() {
 }
 
 // WaitUntilPlacementTableIsReady waits until placement table is until table lock is unlocked.
-func (p *ActorPlacement) WaitUntilPlacementTableIsReady() {
-	if p.tableIsBlocked.Load() {
-		<-p.unblockSignal
+func (p *ActorPlacement) WaitUntilPlacementTableIsReady(ctx context.Context) error {
+	if !p.tableIsBlocked.Load() {
+		return nil
+	}
```
mukundansundar, line 231: please add a UT for this scenario.
Also will this code be executed in normal scenario?

pkg/actors/internal/placement_test.go
```diff
@@ -203,22 +204,66 @@ func TestWaitUntilPlacementTableIsReady(t *testing.T) {
...
+	tableUpdateFunc := func() {}
+	testPlacement := NewActorPlacement(
+		[]string{}, nil,
+		"testAppID", "127.0.0.1:1000",
+		[]string{"actorOne", "actorTwo"},
+		appHealthFunc, tableUpdateFunc)
+
+	testPlacement.onPlacementOrder(&placementv1pb.PlacementOrder{Operation: "lock"})
+
+	readyCh := make(chan struct{})
```
mukundansundar, line 243: nitpick
```suggestion
	testSuccessCh := make(chan struct{})
```

pkg/grpc/api_daprinternal.go
```diff
@@ -0,0 +1,118 @@
...
+		}
+
+		err = status.Errorf(codes.Internal, messages.ErrActorInvoke, err)
+		return nil, err
+	}
+	return resp.Proto(), nil
+}
+
+// Used by CallLocal and CallLocalStream to check the request against the access control list
+func (a *api) callLocalACL(ctx context.Context, req *invokev1.InvokeMethodRequest) error {
```
mukundansundar, line 98:
```suggestion
func (a *api) validateAgainstACL(ctx context.Context, req *invokev1.InvokeMethodRequest) error {
```

pkg/http/api_test.go
```diff
@@ -2477,20 +2402,17 @@ func TestAPIToken(t *testing.T) {
 		apiPath := "v1.0/invoke/fakeDaprID/method/fakeMethod"
 		fakeData := []byte("fakeData")
 
-		fakeReq := invokev1.NewInvokeMethodRequest("fakeMethod")
```
mukundansundar, line 2480: do we need to add separate tests for the scenarios that are there in these fakeReq assignments? There are multiple being removed. Are we covering all the cases that are needed?

pkg/messaging/v1/invoke_method_response.go
```diff
@@ -115,7 +115,7 @@ func (imr *InvokeMethodResponse) IsHTTPResponse() bool {
 	return imr.r.Status.Code >= 100
 }
 
-// Proto clones the internal InvokeMethodResponse pb object.
```
mukundansundar, line 118: the comment here says it returns a clone. 
Is it necessary to return a clone and not the direct object to make sure no changes are made?

pkg/resiliency/resiliency.go
```diff
@@ -827,18 +827,21 @@ func (r *Resiliency) GetPolicy(target string, policyType PolicyType) *PolicyDesc
 		componentPolicy, exists = r.components[target]
 		if exists {
 			policy, _ := policyType.(*ComponentPolicy)
-			if policy.componentDirection == "Inbound" {
+			switch policy.componentDirection {
+			case "Inbound":
```
mukundansundar, line 831: can we have this and outbound as constants ?

pkg/resiliency/resiliency.go
```diff
@@ -827,18 +827,21 @@ func (r *Resiliency) GetPolicy(target string, policyType PolicyType) *PolicyDesc
...
 				}
-			} else {
+			case "Outbound":
 				policyName = PolicyNames{
 					Retry:          componentPolicy.Outbound.Retry,
 					CircuitBreaker: componentPolicy.Outbound.CircuitBreaker,
 					Timeout:        componentPolicy.Outbound.Timeout,
 				}
+			default:
+				panic(fmt.Errorf("invalid component policy direction: '%s'", policy.componentDirection))
```
mukundansundar, line 844: I think we should not panic here. We should return an error and handle that error correctly.

pkg/testing/directmessaging_mock.go
```diff
@@ -1,4 +1,15 @@
-// Code generated by mockery v1.0.0. DO NOT EDIT.
```
mukundansundar, line 1: is this file autogenerated ? Why is it being modified then?

tests/apps/resiliencyapp/app.go
```diff
@@ -56,12 +64,20 @@ type PubsubResponse struct {
...
-var (
-	daprClient   runtimev1pb.DaprClient
-	callTracking map[string][]CallRecord
-)
-
-var httpClient = utils.NewHTTPClient()
+func init() {
+	p := os.Getenv("DAPR_HTTP_PORT")
+	if p != "" && p != "0" {
+		daprHttpPort, _ = strconv.Atoi(p)
```
mukundansundar, line 70: even in test should we just default on error instead of ignoring it?

tests/apps/resiliencyapp/app.go
```diff
@@ -125,10 +141,11 @@ func resiliencyBindingHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	body, _ := io.ReadAll(r.Body)
 	var message FailureMessage
-	json.NewDecoder(r.Body).Decode(&message)
+	json.Unmarshal(body, &message)
```
mukundansundar, line 146: qq is this faster than decoder? why is this change needed?

pkg/http/api_test.go
ItalyPaleAle: All these objects that were removed where completely unused. My guess is that whoever wrote the test cases copy/pasted from cases where the reqs were needed, and didn't then realize they were unused.

pkg/messaging/v1/invoke_method_response.go
ItalyPaleAle: We've been going all this time without cloning and it's been working fine :) Cloning would also degrade performance.

pkg/testing/directmessaging_mock.go
ItalyPaleAle: It was originally auto-generated, but over the course of time people had been adding patches manually.

pkg/resiliency/resiliency.go
ItalyPaleAle: IMHO a panic here is appropriate because this should never happen during runtime. It's only here to help catch development-time errors.

tests/apps/resiliencyapp/app.go
ItalyPaleAle: This env var was added to allow running the tests on a different port and it's only used during development.

tests/apps/resiliencyapp/app.go
ItalyPaleAle: No, actually `json.NewDecoder(r.Body).Decode(&message)` is usually faster because it works on the stream.

The reason is that in the line below, I am now printing out the raw message (before it gets decoded, in case it's invalid JSON). Doing that requires loading the message in memory first (see `body, _ := io.ReadAll(r.Body)`). If the message isn't loaded in memory first, `NewDecoder(r.Body)` will consume the stream and cannot be read again.

pkg/actors/internal/placement.go
ItalyPaleAle: I am not sure when this would be executed, probably just during init phases. The code was there before, I have just added support for contexts.

pkg/messaging/v1/invoke_method_response.go
mukundansundar: agreed cloning would degrade performance. as long as in the code path where this Proto is used, we are ok with the direct access, then it's fine.

pkg/resiliency/resiliency.go
mukundansundar: my question is should the application crash because of an invalid resiliency policy? Should it not gracefully degrade and use default polices where applicable and let user know that there is an error?

In this case, my opinion is that a faulty resiliency policy config should not cause the application to crash.

ItalyPaleAle: Oh an invalid resiliency policy would not cause this crash. Only using the wrong constant in the source code would.

pkg/actors/internal/placement.go
ItalyPaleAle: Added test

pkg/actors/internal/placement_test.go
ItalyPaleAle: Changed

pkg/grpc/api_daprinternal.go
ItalyPaleAle: Changed

pkg/messaging/direct_messaging.go
```diff
@@ -152,9 +152,8 @@ func (d *directMessaging) requestAppIDAndNamespace(targetAppID string) (string,
 	}
 }
 
-// invokeWithRetry will call a remote endpoint for the specified number of retries and will only retry in the case of transient failures
-// TODO: check why https://github.com/grpc-ecosystem/go-grpc-middleware/blob/master/retry/examples_test.go doesn't recover the connection when target
-// Server shuts down.
+// invokeWithRetry will call a remote endpoint for the specified number of retries and will only retry in the case of transient failures.
+// TODO: check why https://github.com/grpc-ecosystem/go-grpc-middleware/blob/master/retry/examples_test.go doesn't recover the connection when target server shuts down.
```
mukundansundar, line 156: not related to this PR. Can an issue be created to track this?

pkg/resiliency/resiliency.go
mukundansundar: went back and saw the full code flow for reading in resiliency policy. Agree with your comment.

pkg/grpc/api.go
```diff
@@ -2030,8 +1952,6 @@ func (a *api) SubscribeConfigurationAlpha1(request *runtimev1pb.SubscribeConfigu
 	subscribeKeys := make([]string, 0)
 
 	// TODO(@halspang) provide a switch to use just resiliency or this.
```
mukundansundar, line 1954: is this comment still valid?

pkg/grpc/api.go
```diff
@@ -2050,7 +1970,7 @@ func (a *api) SubscribeConfigurationAlpha1(request *runtimev1pb.SubscribeConfigu
 
 	// TODO(@laurence) deal with failed subscription and retires
```
mukundansundar, line 1971: is this comment still valid considering resiliency is in place for this call?

pkg/grpc/api.go
ItalyPaleAle: No clue what this even meant. @halspang ?

mukundansundar: ping @halspang

PR #5611, opened 2022-12-06
Add support for tracing to streamed gRPC calls

ItalyPaleAle:
This PR modifies the gRPC interceptor for stream-based calls to ensure that tracing is added correctly to methods that use streams.

The current method is designed to instrument only calls that are used by gRPC proxying, however this is not enough.

We already have one (alpha) gRPC method that is stream-based: `SubscribeConfigurationAlpha1`. We are also going to add new stream-based methods in the next weeks, including in the crypto building block (part of dapr.runtime) and in stream-based HTTP service invocation (part of dapr.internal).

This PR ensures that traces are included in all those kinds of calls: in addition to gRPC proxying, also gRPC invocations from clients (dapr.runtime) and other runtimes (dapr.internal)


pkg/diagnostics/grpc_tracing.go
```diff
@@ -98,73 +101,89 @@ func GRPCTraceUnaryServerInterceptor(appID string, spec config.TracingSpec) grpc
...
-	if span.SpanContext().IsSampled() {
-		// users can add dapr- prefix if they want to see the header values in span attributes.
-		prefixedMetadata = userDefinedMetadata(ctx)
-	}
+		err := handler(srv, wrapped)
 
-	if span.SpanContext().IsSampled() {
-		// Populates dapr- prefixed header first
-		AddAttributesToSpan(span, prefixedMetadata)
+		if isSampled {
```
johnewart, line 164: nit: Can this be merged with the above sample check for simplicity? It doesn't seem like the span object gets used in `handler` and if we care about avoiding extra work if the handler fails then the above block could be moved down here anyway, right?

ItalyPaleAle: Uhm I'm not sure if there's a possibility the handler modifies the span. The context is added to `wrapped` so the handler could access that.

The code for unary RPCs does something similar too. I am not comfortable enough touching this sorry

pkg/diagnostics/grpc_tracing.go
```diff
@@ -248,14 +267,10 @@ func SpanContextToGRPCMetadata(ctx context.Context, spanContext trace.SpanContex
...
-func isInternalCalls(method string) bool {
-	return strings.HasPrefix(method, "/dapr.proto.internals.")
-}
-
 // spanAttributesMapFromGRPC builds the span trace attributes map for gRPC calls based on given parameters as per open-telemetry specs.
-func spanAttributesMapFromGRPC(appID string, req interface{}, rpcMethod string) map[string]string {
+func spanAttributesMapFromGRPC(appID string, req any, rpcMethod string) map[string]string {
 	// RPC Span Attribute reference https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/rpc.md
-	m := map[string]string{}
+	m := make(map[string]string, 8)
```
johnewart, line 273: I think you picked 8 because it looks like there are 8-ish unique keys below; can we add a note on why the map is deliberately sized? 😄

pkg/diagnostics/grpc_tracing.go
```diff
@@ -98,73 +101,89 @@ func GRPCTraceUnaryServerInterceptor(appID string, spec config.TracingSpec) grpc
...
-		wrapped := grpcMiddleware.WrapServerStream(ss)
-		sc, _ := SpanContextFromIncomingGRPCMetadata(ctx)
 
-		var spanKind trace.SpanStartOption
-
-		if appID == targetID {
+		// This middleware is shared by multiple services and proxied requests, which need to be handled separately
+		switch {
+		// For gRPC service invocation, this generates ServerSpan
+		case strings.HasPrefix(info.FullMethod, daprInternalPrefix):
```
johnewart, line 120: Looks like the tests might not cover this condition -- is it possible to write one?

ItalyPaleAle: I'll see to add one.

pkg/diagnostics/grpc_tracing.go
johnewart: No worries, it just seems a little odd that we check it twice.

pkg/diagnostics/grpc_tracing.go
ItalyPaleAle: Done

johnewart (approved):
Post-change question / thought - I assume we need to make changes to E2E tests for components that support streaming to be sure tracing data is collected, right?

ItalyPaleAle:
We do.... except we don't have a test for that yet :) but @asouza is working on one

ItalyPaleAle:
We do - when we get a test because right now we don't have any! But @artursouza is working on it on #5618

pkg/diagnostics/utils/trace_utils.go
```diff
@@ -31,6 +31,8 @@ const (
 	daprFastHTTPContextKey = "daprSpanContextKey"
 )
 
+var nilSpanContext trace.SpanContext
```
mukundansundar, line 34: nitpick
```suggestion
var emptySpanContext trace.SpanContext
```

pkg/diagnostics/grpc_tracing.go
```diff
@@ -98,73 +101,89 @@ func GRPCTraceUnaryServerInterceptor(appID string, spec config.TracingSpec) grpc
...
+			// users can add dapr- prefix if they want to see the header values in span attributes.
+			prefixedMetadata = userDefinedMetadata(ctx)
+			if isProxied {
+				reqSpanAttr = map[string]string{
+					daprAPISpanNameInternal: info.FullMethod,
+				}
+			} else {
+				reqSpanAttr = spanAttributesMapFromGRPC(appID, nil, info.FullMethod)
+			}
+		}
```
mukundansundar, line 160: Can this come after the `handler` is called? Combing this block and the `if` clause in line 164 together.

`addSpanMetadataAndUpdateStatus` function was called after `handler` was called in the previous code.

pkg/diagnostics/grpc_tracing.go
```diff
@@ -98,73 +101,89 @@ func GRPCTraceUnaryServerInterceptor(appID string, spec config.TracingSpec) grpc
...
+
+		// For proxied requests, this generates a span depending on whether this is the server (target) or client
+		default:
+			isProxied = true
+			md, _ := metadata.FromIncomingContext(ctx)
+			vals := md.Get(GRPCProxyAppIDKey)
+			if len(vals) == 0 {
+				return fmt.Errorf("cannot proxy request: missing %s metadata", GRPCProxyAppIDKey)
+			}
+			if appID == vals[0] {
```
mukundansundar, line 135: do set vals[0] to targetID, adding more context to the code ...

pkg/diagnostics/grpc_tracing.go
```diff
@@ -98,73 +101,89 @@ func GRPCTraceUnaryServerInterceptor(appID string, spec config.TracingSpec) grpc
...
+				spanKind = trace.WithSpanKind(trace.SpanKindServer)
+			} else {
+				spanKind = trace.WithSpanKind(trace.SpanKindClient)
+			}
 		}
 
+		// Overwrite context
+		sc, _ := SpanContextFromIncomingGRPCMetadata(ctx)
```
mukundansundar, line 143: qq why is the context being overwritten here?

pkg/diagnostics/grpc_tracing.go
```diff
@@ -33,36 +33,39 @@ import (
 )
 
 const (
-	GRPCTraceContextKey = "grpc-trace-bin"
-	GRPCProxyAppIDKey   = "dapr-app-id"
-	daprPackagePrefix   = "/dapr.proto"
+	GRPCTraceContextKey     = "grpc-trace-bin"
+	GRPCProxyAppIDKey       = "dapr-app-id"
+	daprInternalPrefix      = "/dapr.proto.internals."
+	daprRuntimePrefix       = "/dapr.proto.runtime."
```
mukundansundar, line 39: and why is the `daprPackagePrefix` being split into two? 
can there be any other cases other than these two(leaving aside service invocation)?

ItalyPaleAle: We need to be able to distinguish calls from other Dapr sidecars (internals - primarily, but not only, service invocation) from calls from apps to Dapr (runtime).

ItalyPaleAle: This may change a bit once the streaming version of CallLocal is implemented by the way.

pkg/diagnostics/grpc_tracing.go
ItalyPaleAle: @johnewart  commented the same above. In the unary interceptor there was a similar logic where `handler` is called sandwiched between two `if isSampled`. I think there's a chance that in the handler the context is changed and I do not feel comfortable changing that.

mukundansundar (approved):

pkg/diagnostics/grpc_tracing.go
```diff
@@ -98,73 +101,90 @@ func GRPCTraceUnaryServerInterceptor(appID string, spec config.TracingSpec) grpc
...
 			spanKind = trace.WithSpanKind(trace.SpanKindServer)
-		} else {
+
+		// For gRPC API, this generates ClientSpan
+		case strings.HasPrefix(info.FullMethod, daprRuntimePrefix):
 			spanKind = trace.WithSpanKind(trace.SpanKindClient)
+
+		// For proxied requests, this generates a span depending on whether this is the server (target) or client
+		default:
+			isProxied = true
```
DeepanshuA, line 129: Just wanted to understand one thing here: any calls from "/dapr.proto.operator." etc. will fall in this category, so based on GRPCProxyAppIDKey, it would be decided if Server span to be allocated or client span to be allocated. Ideally, GRPCProxyAppIDKey would be present here, as it is dapr-app-id. Just wondering, would calls like these from control plane service etc. be falling under server or client category? 
Or there is a gap in my understanding here.

ItalyPaleAle: I may be wrong but I don't believe we have instrumented calls from the runtime to the operator service - I don't think those calls would ever go through this.

DeepanshuA (approved):

artursouza (approved):

PR #5614, opened 2022-12-07
Updated dev container CI

ItalyPaleAle:
1. Setting tag to a non-versioned "latest" given that the container is always built manually
  - Dev Containers from previous versions of Dapr are [failing to build](https://github.com/dapr/dapr/actions/workflows/dapr-dev-container.yml) right now (until 1.8, the Action had an automatic trigger and we had to turn that off as it was causing issues like overriding the dev container on every hotfix)
  - We often make changes in the "master" branch such as upgrading the version of Go or the linter being used, so we should be able to re-publish dev containers whenever we feel like we should
2. Updated the Action to upgrade dependencies that will be deprecated soon ([see warnings here](https://github.com/dapr/dapr/actions/runs/3606497150))


.devcontainer/devcontainer.json
```diff
@@ -1,7 +1,7 @@
 {
 	"name": "Dapr Dev Environment",
 	// Update the container version when you publish dev-container
-	"image": "ghcr.io/dapr/dapr-dev:0.1.9",
+	"image": "ghcr.io/dapr/dapr-dev:latest",
```
daixiang0, line 4: Now does it auto-build and push?

shubham1172 (dismissed):

.devcontainer/devcontainer.json
ItalyPaleAle: https://github.com/dapr/dapr/actions/runs/3606497150/workflow#L16

daixiang0: Although we have updated “build dev container” workflow, but for developers, they can not get those changes, right?

Could we re-enable this job?

ItalyPaleAle: What do you mean? The job is enabled, just needs to be triggered manually

ItalyPaleAle: Done. It now runs weekly too

shubham1172 (dismissed):

.github/workflows/dapr-dev-container.yml
```diff
@@ -15,11 +15,14 @@ name: dapr-dev-container
 
 on:
   workflow_dispatch:
+  schedule:
+    # Run weekly on Tuesdays
+    - cron: "22 03 * * 2"
```
shubham1172, line 20: What's with 3:22 though? 😄

ItalyPaleAle: A random time that is not on top of the hour (when a lot of concurrent tasks are happening), and that’s in the middle of the night PST, when we see less activity on the repo.
I actually put a bit of thought on this :)

.devcontainer/devcontainer.json
mukundansundar: If its being triggered automatically, can this create an issue if we are always going to use the latest dev container?

DeepanshuA: I think mostly developers would be using it, so may be that is better if we get latest.

ItalyPaleAle: What kind of issues are you concerned with?

DeepanshuA (approved):

.devcontainer/devcontainer.json
artursouza: Can we avoid using latest?

.github/workflows/dapr-dev-container.yml
artursouza: Genesis 3:22

And the Lord God said, “The man has now become like one of us, knowing good and evil. He must not be allowed to reach out his hand and take also from the tree of life and eat, and live forever.”

docker/docker.mk
```diff
@@ -242,7 +242,7 @@ docker-windows-base-push: check-windows-version
 ################################################################################
 
 # Update whenever you upgrade dev container image
-DEV_CONTAINER_VERSION_TAG?=0.1.9
```
artursouza, line 245: Please, avoid using `latest` - we got hit by this in multiple places before.

artursouza (changes requested):
Avoid `latest` unless it is needed for sure.

ItalyPaleAle:
> Please, avoid using latest - we got hit by this in multiple places before.

What were the issues before?

The reason for `latest` is actually because "master" (which is where we should assume most contributors are working in) moves fast and we should push images quickly.

Note that for the last few months, dev containers have been built manually, so hotfixes wouldn't override older containers. In the past, they were built on every tag. This meant that when 1.8.0 was released, dev container 1.8 was built. However, "master" branch evolved, and we (for example) upgraded Go. We re-triggered the dev container so 1.8 was re-built. Everything was fine until 1.8.1 came out, and then the dev container was rebuilt with the old version of Go. This is why we've moved months ago to only triggering the dev container manually.

berndverst:
While you are touching things related to the dev container, can you please enable the SSH feature in the dev container json?

```
"features": {
    "sshd": "latest"
}
```

ItalyPaleAle:
@berndverst done

artursouza (approved):

PR #5618, opened 2022-12-08
E2E test for tracing with HTTP service invocation with Zipkin

artursouza:
# Description

E2e test for tracing with HTTP service invocation

## Issue reference

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing


tests/config/dapr_tracing_config.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: dapr.io/v1alpha1
+kind: Configuration
+metadata:
+  name: tracingconfig
+spec:
+  tracing:
+    samplingRate: "1"
+    zipkin:
```
yaron2, line 8: Shouldn't we use OTEL instead of Zipkin? That would be the recommended path now that we support OTEL. We can use Zipkin as the backend, just configure it through the OTel collector.

ItalyPaleAle: Or, higher level question... Do we need to deploy Zipkin? Or can we create an endpoint (even inside the test app itself) that simulates zipkin and that accepts the requests? This way we don't need to deploy another service.

artursouza: This was the easiest way to reproduce the bug. I think we will have more scenarios but having an endpoint to handle zipkin would be too much IMO - the code is much simpler if we can reuse it instead of reimplementing zipkin API.

When it comes to OTEL instead of Zipkin, I can look into it - but these are not exclusive. We need to expand the tracing test scenarios beyond this one. This is a first step.

tests/apps/tracingapp/app.go
```diff
@@ -0,0 +1,275 @@
...
+
+	url := fmt.Sprintf("http://127.0.0.1:%d/v1.0/invoke/%s/method/invoke/something", daprHTTPPort, appId)
+	/* #nosec */
+	req, _ := http.NewRequest(http.MethodPost, url, nil)
+	req = req.WithContext(newCtx)
+	hc := propagation.HeaderCarrier(req.Header)
+	propagators.Inject(newCtx, hc)
+	req.Header.Add("Content-Type", jsonContentType)
+
+	fmt.Printf("span's name is %s and invoke url is %s\n", uuid, url)
```
ItalyPaleAle, line 116: Consider using `log` rather than `fmt`

tests/apps/tracingapp/app.go
```diff
@@ -0,0 +1,275 @@
...
+
+// invoke is the handler for end-to-end to invoke
+func invoke(w http.ResponseWriter, r *http.Request) {
+	log.Printf("Processing %s %s", r.Method, r.URL.RequestURI())
+	ctx := propagators.Extract(r.Context(), propagation.HeaderCarrier(r.Header))
+	_, span := tracer.Start(ctx, "invokedMethod")
+	defer span.End()
+
+	// Pretend to do work
+	time.Sleep(time.Second)
```
ItalyPaleAle, line 151: Missed a chance to actually mine some bitcoins here 🙃

ItalyPaleAle (approved):
Once tests passes, LGTM

tests/apps/tracingapp/app.go
artursouza: Will do

artursouza: Done.

tests/apps/tracingapp/app.go
artursouza: It would be interesting to see the cost-benefit analysis of doing so in the cloud when you pay for reserved capacity and not for utilized capacity. If you have reserved and unused capacity going to waste, then mining might make sense. Good idea for our LH tests.

artursouza:
I have made changes to make it clear this one is for zipkin with HTTP and allow us to expand into using OTEL too.

ItalyPaleAle (approved):

PR #5623, opened 2022-12-08
Allow daprd to load the legacy service account token

ItalyPaleAle:
With #5371 , Dapr added support for using bound tokens. In <= 1.9, daprd authenticates itself against sentry using a K8s service account token, which has the ability to access the K8s APIs. In the upcoming 1.10, this was changed so the injector issues a new token for daprd with no K8s API access at all.

While the implementation in #5371 was backwards-compatible, meaning that a 1.9 daprd could be used with a 1.10 control plane (specifically, sentry and injector), it was not forwards-compatible. This means that daprd built from master right now cannot be used with a 1.9 control plane (sentry and injector).

This PR addresses the issue of forward compatibility by allowing daprd to fall back to loading a service account token (w/ access to K8s APIs) when the bound token is not available. I also made sure to print a warning in the console when that happens, as it should be considered as an exception.

### DIFF
--- a/pkg/runtime/security/auth.go
+++ b/pkg/runtime/security/auth.go
@@ -4,13 +4,13 @@ import (
 	"context"
 	"crypto/x509"
 	"encoding/pem"
+	"fmt"
 	"os"
 	"sync"
 	"time"
 
 	grpcMiddleware "github.com/grpc-ecosystem/go-grpc-middleware"
 	grpcRetry "github.com/grpc-ecosystem/go-grpc-middleware/retry"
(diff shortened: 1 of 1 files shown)


johnewart (approved):
Do we want to include a link to where someone should go to fix this (i.e docs) or find more info?

ItalyPaleAle:
The only "fix" is upgrading the Dapr control plane.

mukundansundar (approved):
LGTM

artursouza (approved):

PR #5633, opened 2022-12-13
Update Mariner Docker images

ItalyPaleAle:
The Mariner team recently made some changes to the base images, so the `distroless/minimal:2.0` images now include the tzdata package and the root CA certs (which were not included before). They are also publishing a `-nonroot` variant. Those changes mean that we can now simplify greatly our Dockerfile for Mariner.


mukundansundar (approved):
I am not seeing license comment in other Dockerfiles also. Do we need one?

mukundansundar:
> > I am not seeing license comment in other Dockerfiles also. Do we need one?
> 
> 
> 
> The license was needed in the old Dockerfile because it contained code copied from a Dockerfile licensed under MIT. The new Dockerfile is "original" and doesn't contain that code anymore.
> 
> 
> 
> As for tests, doesn't matter if Windows fails; only Linux matters here :)

The license that we add to our files.
Dapr authors license is what I meant.

artursouza (approved):

Issue #5636, opened 2022-12-13
Remove sandbox dependencies


yaron2:
Our E2E tests depend on Go packages from the sandbox org. All dependencies must be removed; Sandbox packages should not be a dependency for any Dapr project outside of sandbox.

shubham1172:
Should this issue be closed since https://github.com/dapr/dapr/pull/5639 was closed? /cc @artursouza

shubham1172:
Should this issue be closed since https://github.com/dapr/dapr/pull/5639 was closed? /cc @artursouza

yaron2:
> Should this issue be closed since #5639 was closed? /cc @artursouza

No, we won't close this yet, but I moved this to 1.11.

PR #5639, opened 2022-12-13
Remove pluggable components from E2E tests.

tmacam:
# Description

These tests currently depend on the Pluggable Components Go SDK, which
is a dapr-sandbox project (`dapr-sandbox/components-go-sdk`).

While valuable to tests the pluggable component interface, Dapr core
repositories (such as `dapr/dapr`) should not have dependencies on
projects in dapr-sandbox.

Ideally, we should have these E2E tests coded without depending on
pluggable components SDKs. Or at the very least, depend on pinned
version of them _once they graduate_. But that's for the medium/long
term.

I am removing these tests for the time being to address this
dapr-sandbox dependency in the short term. We can revert this if
required.

Fixes https://github.com/dapr/dapr/issues/5636.

## Issue reference

Please reference the issue this PR will close: #5636.

## Checklist

* [ ] Code compiles correctly
* [x] Created/updated tests
* [ ] Unit tests passing


ItalyPaleAle:
CC: @yaron2

ItalyPaleAle (changes requested):
Rather than completely removing the code, can we just disable the tests with a flag?

1. Delete the apps from `tests/dapr_tests.mk` so they aren't built (this is the step that was causing the tests failures in the PR this morning)
2. Leave the `tests/apps/pluggable_*` folders for when we can re-enable them (maybe add a README saying they're currently unused and say why)
3. In the tests (`tests/e2e/bindigns/bindings_test.go` et similiar) rather than deleting code, just do a `t.Skip()` to skip the pluggable tests

tmacam:
> Rather than completely removing the code, can we just disable the tests with a flag?

We could do that but I wonder if this would mean that the _buildability_ of dapr/dapr would be subject to the _buildability_ of components-go-sdk.

yaron2:
> > Rather than completely removing the code, can we just disable the tests with a flag?
> 
> We could do that but I wonder if this would mean that the _buildability_ of dapr/dapr would be subjected to the _buildability_ of components-go-sdk.

It would be, albeit not in a deliverable path. Please remove the dependencies all together.

tmacam:
> It would be, albeit not in a deliverable path. Please remove the dependencies all together.

Done.

ItalyPaleAle:
Currently the only dependency is in the E2E test apps, not even in the test files however. If that makes a difference. So removing the pluggable apps from `tests/dapr_tests.mk` would make "everything" build

artursouza:
I have a slightly different opinion on this matter. Pluggable components is a preview feature, not stable yet. Preview features are expected to also have E2E tests and those tests are OK to depend on a sandbox project.

These tests are super important for us to keep making progress on pluggable components towards stability. Removing or disabling them will remove this value.

Today, Dapr already depends on many 3rd party dependencies with version 0.x in the core runtime binary - depending on dapr-sandbox for test apps is minor compared to that.

yaron2:
Due to conflicts and the fact we do want to test pluggable components, we can close this PR but track the work to either replace the sandbox dependencies or provide a stable, versioned artifact into dapr/dapr.

berndverst:
When Dapr component interfaces are modified Dapr/Dapr tests fail because the pluggable components SDK doesn't match the interface signature.

That's completely unacceptable. Dapr/dapr should be merged first and the pluggable SDK corrected afterwards.

Yet we need to modify the pluggable SDKs living in the sandbox first. This is not reasonable for contributors.

I respectfully ask maintainers to merge this PR. If maintainers do not wish to merge this PR I maintainers will personally commit to coordinating the updates to the pluggable SDKs to get PRs to pass the test suite.

@

artursouza:
This is an important discussion now: changes to component interfaces should not be breaking now that we have pluggable components. There will be no zero-downtime path to upgrade Dapr in production when using pluggable components.

It seems that these tests have uncovered another issue: how do we guarantee backwards compatibility of components interfaces. I know we did not have that guarantee until now, but we cannot simply delete the tests to feel good because users will be hurt in production and make pluggable components unusable.

berndverst:
Just like with regular SDKs as part of release activities the pluggable components SDKs should be updated before a new release.

As far as backwards compatibility is concerned - yes we should make sure that works.

Maybe we need to define some additional interfaces to make sure we don't break things when we introduce methods that for example pluggable components don't need to implement.

As an example the GetComponentMetadata function only needs to be implement by the server of pluggable components - the actual component itself does not need to implement this as this function would not be used for pluggable components.

The problem is that the pluggable Go SDK should not directly be importing from Dapr/Dapr. It should define its own copy of interfaces. That way it doesn't run into these problems we are encountering.

artursouza:
We decided not to merge this and pinned the sandbox SDK versions instead.

PR #5645, opened 2022-12-15
Changes to the resiliency.NewRunner

ItalyPaleAle:
Three main changes:

## Fixes a goroutine leak when a timeout happens

Previously, when a timeout happened, there was likely going to be a goroutine leak because the runner was storing the return value in an unbuffered channel which was never read. The background goroutine would then leak "forever".

## Add accumulator functions

(Co-authored by @mukundansundar )

Runners can now accept an optional "Accumulator" which is a function that receives all non-zero values returned by the operation functions, even when there's an error, except when there's a timeout (because in case of timeouts, the policy should move on to the next retry and not wait for results).

The use case is for PRs such as #5603 where we need to be able to do operations in bulk and not retry things in the batch that have been successful already.

## Add disposer functions

Runners can also now accept an optional "Disposer" which is a function that receives all non-zero values returned by the operation functions when there's an error (including timeouts).

Use case is being able to dispose objects that are otherwise going to "leak" (until the GC picks them up). Example is if the operation function returns a `*http.Response` object, in which case we need to be able to call `res.Body.Close()` in the returned value _even when_ there's an error such as a timeout (otherwise, those returned values would be lost).

This will be needed in another PR I'm working on.

### DIFF
--- a/pkg/resiliency/policy.go
+++ b/pkg/resiliency/policy.go
@@ -16,6 +16,8 @@ package resiliency
 import (
 	"context"
 	"errors"
+	"reflect"
+	"sync/atomic"
 	"time"
 
 	"github.com/cenkalti/backoff/v4"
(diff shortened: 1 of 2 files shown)


mukundansundar (dismissed):
lgtm

mukundansundar (approved):

mukundansundar (approved):

PR #5651, opened 2022-12-16
Deploy (& dispose) E2E test apps in parallel

ItalyPaleAle:
Should help making E2E tests complete a few mins quicker by deploying (and disposing) E2E test apps in parallel. This should be especially useful in AKS (because of having to pull from a container registry through the network and because of the need to wait for a load balancer), and especially for tests like "service_invocation" that requires deploying half a dozen apps.


ItalyPaleAle:
Looks like some things do need to be deployed in order. Closing for now - will maybe look into this again in the future

PR #5652, opened 2022-12-17
Fix call to fetch subscriptions over gRPC but not implemented.

artursouza:
Fix call to fetch subscriptions over gRPC but not implemented.

This issue is detected in the Java SDK and `git bisect` accused commit eeca309f0ca1630d85415cd4a6c4d52cbaaf4092 as the root cause. This change will handle `not implemented` case correctly in gRPC.

## Issue reference

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing


pkg/runtime/pubsub/subscriptions.go
```diff
@@ -199,7 +199,8 @@ func GetSubscriptionsGRPC(channel runtimev1pb.AppCallbackClient, log logger.Logg
 				s, ok := status.FromError(rErr)
 				if ok && s != nil {
 					if s.Code() == codes.Unimplemented {
-						return rResp, backoff.Permanent(rErr)
+						log.Infof("pubsub subscriptions: gRPC app does not implement ListTopicSubscriptions")
```
mukundansundar, line 202:
```suggestion
						log.Warnf("pubsub subscriptions: gRPC app does not implement ListTopicSubscriptions")
```

pkg/runtime/pubsub/subscriptions.go
```diff
@@ -214,7 +215,8 @@ func GetSubscriptionsGRPC(channel runtimev1pb.AppCallbackClient, log logger.Logg
 			if rErr != nil {
 				s, ok := status.FromError(rErr)
 				if ok && s != nil && s.Code() == codes.Unimplemented {
-					return rResp, backoff.Permanent(rErr)
+					log.Infof("pubsub subscriptions: gRPC app does not implement ListTopicSubscriptions")
```
mukundansundar, line 218:
```suggestion
					log.Warnf("pubsub subscriptions: gRPC app does not implement ListTopicSubscriptions")
```

mukundansundar (commented):
Can a unit test be added for this?

artursouza:
> Can a unit test be added for this?

Well. There is one as I see but the expectations are wrong on that one. So I will fix it.

ItalyPaleAle (approved):
LGTM but agree with Mukundan’s feedback that log message may be a warning in this case. If people use the Dapr SDKs, it should be implemented for them anyways.

pkg/runtime/pubsub/subscriptions.go
```diff
@@ -199,7 +199,8 @@ func GetSubscriptionsGRPC(channel runtimev1pb.AppCallbackClient, log logger.Logg
 				s, ok := status.FromError(rErr)
 				if ok && s != nil {
 					if s.Code() == codes.Unimplemented {
-						return rResp, backoff.Permanent(rErr)
+						log.Infof("pubsub subscriptions: gRPC app does not implement ListTopicSubscriptions")
+						return rResp, nil
```
ItalyPaleAle, line 203: What is rResp in this case? `nil`? Should we just return `nil` (or another zero value) here?

ItalyPaleAle (dismissed):

pkg/runtime/pubsub/subscriptions.go
artursouza: Why warning? Apps are not required or expected to implement this. It means a hello world app would get this warning - which is why I am using Info.

pkg/runtime/pubsub/subscriptions.go
artursouza: It is not nil, but I am changing this and the existing code to guarantee that and not depend on this behavior from the downstream function.

pkg/runtime/pubsub/subscriptions.go
artursouza: I believe this should be Info since it is really not expected for apps to implement this.

ItalyPaleAle (approved):

PR #5663, opened 2022-12-20
Update protoc version

ItalyPaleAle:
Extracted from #5648


berndverst:
Please be careful with these changes -- for example using newer versions of protoc generates protos which in Python SDK may require newer versions of `protobuf` (and consequently `grpcio`) dependencies -- but we cannot bump those libraries because it would break compatibility with many ML libraries.

Protobuf has significant changes even between minor versions.

ItalyPaleAle:
I was getting some incredibly weird errors from the CI with `check-proto-diff` failing for apparently no reason. This was an attempt at fixing it.

The generated protos however are not different in the Dapr case, so SDKs should not need updating

berndverst:
We generate the protos separately from the proto files in the Python SDK anyway, so I can always use a different version of protoc there -- so no problem. Proceed as you wish :)

berndverst (approved):

PR #5664, opened 2022-12-20
Misc test fixes

ItalyPaleAle:
1. Fixes some (not all) race conditions in tests for pkg/runtime
2. Improvements to test platform and the actorfeatures test to make testing locally (outside of K8s) easier
3. Some more logging in E2E test apps
4. Increase resource allocation to Tailscale (if used), which was crashing with OOM a few times for me

Extracted from #5648


berndverst (dismissed):

pkg/runtime/runtime_test.go
```diff
@@ -5422,7 +5423,7 @@ func TestGracefulShutdownPubSub(t *testing.T) {
 	assert.NotNil(t, rt.topicRoutes)
 	go sendSigterm(rt)
 	select {
-	case <-rt.pubsubCtx.Done():
+	case <-rt.shutdownC:
```
mukundansundar, line 5426: this was specifically the pubsub context as it is stopped before dapr starts waiting the grace time period ... 
This was a UT specifically to test that. This change does not test that. This change just verifies that pubsub is closed after all the shutdown is down ...

mukundansundar (changes requested):

pkg/runtime/runtime_test.go
ItalyPaleAle: I see.

I can revert this specific change but I will need @akhilac1 to look into this then. `make test-race` was failing on `pkg/runtime` on this line due to using `pubsubCtx.Done()` which then gets set to `nil` in the test.

mukundansundar (approved):

PR #5677, opened 2022-12-28
Allow enabling preview features at build-time

ItalyPaleAle:
Added the `ENABLED_FEATURES` env var to the Makefile to define a (comma-separated) list of features that are always enabled, regardless of what's in the Configuration spec.

For example, building with `ENABLED_FEATURES="Resiliency,Foo" make build` produces builds that always have the features `Resiliency` and `Foo` enabled, and they cannot be disabled with the Configuration document.

Use cases include:

- Simpler to test with certain preview features enabled
- Enable certain features in testing / CI environments only
- Being able to produce custom builds with certain features enabled, for validation or other reasons

Note: The `Resiliency` feature was added to the list of always-enabled features for now (replacing the previous "hack" to have it always enabled with #5523).


artursouza (approved):

pkg/runtime/runtime_test.go
```diff
@@ -1374,11 +1374,6 @@ func TestInitPubSub(t *testing.T) {
 		rt.appChannel = mockAppChannel
 
 		// User App subscribes 2 topics via http app channel
-		fakeReq := invokev1.NewInvokeMethodRequest("dapr/subscribe").
```
ItalyPaleAle, line 1377: Reviewers: no idea why changes to these tests were needed. This doesn't really change the test, just makes the matcher a bit more lax:

- Before, it required to receive the exact `fakeReq` object
- Now it requires an object which is a request object for `dapr/subscribe`

The change to the matcher is similar to what's used in this test file in many other places.

Without this change, the matcher doesn't recognize the object because a property inside is `nil` rather than another zero value.

I have no idea why this test worked as-is in master and fails here, since nothing should have changed. But this fixes the test - and it's arguably more robust anyways.

ItalyPaleAle: From the failing test, where the matcher was failing:

image

🤷🤷🤷

ItalyPaleAle: Actually I think I know now why this is happening... In master, resiliency is enabled in unit tests, but with this PR it is not. (it's still enabled in E2E tests)

ItalyPaleAle: Changed to add a way to enable features in unit tests

artursouza (approved):

PR #5680, opened 2022-12-28
Add option to annotate application pods

shubham1172:
# Description

This PR adds an option to the sidecar injector helm charts to annotate the application pods. Users can specify `dapr_sidecar_injector.appPodAnnotations` via helm and configure application pods running Dapr to contain custom annotations.

## Issue reference

NA

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing

### DIFF
--- a/pkg/injector/pod_patch.go
+++ b/pkg/injector/pod_patch.go
@@ -101,15 +101,6 @@ func (i *injector) getPodPatchOperations(ar *v1.AdmissionReview,
 	// Projected volume with the token
 	tokenVolume := sidecar.GetTokenVolume()
 
-	// Pod annotations
-	podPatchOps := []common.PatchOperation{
-		{
-			Op:    "add",
-			Path:  patchPathAnnotations,
-			Value: i.config.GetAppPodAnnotations(),
-		},
-	}
-
 	// Get the sidecar container
 	sidecarContainer, err := sidecar.GetSidecarContainer(sidecar.ContainerConfig{
 		AppID:                        appID,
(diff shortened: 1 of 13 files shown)


pkg/injector/pod_patch.go
```diff
@@ -98,6 +101,15 @@ func (i *injector) getPodPatchOperations(ar *v1.AdmissionReview,
 	// Projected volume with the token
 	tokenVolume := sidecar.GetTokenVolume()
 
+	// Pod annotations
+	podPatchOps := []common.PatchOperation{
```
ItalyPaleAle, line 105: Should this be empty if `GetAppPodAnnotations` returns an object with len == 0?

shubham1172: Yes, changed!

yaron2:
Why is this PR needed? Users can annotate pods directly. Also, what is an application pod? Every pod is an application pod that contains two containers - the app container and a Dapr sidecar container. I'm unsure what this PR achieves or what problem it solves.

ItalyPaleAle (dismissed):

yaron2:
@shubham1172 after reviewing this PR I believe it should be closed. Dapr's control plane should not be extended to handle non-Dapr operations. Users control their pod specifications and metadata annotations via the same mechanisms that annotates Dapr. Any CI/CD system, GitOps enabled or otherwise, would be the place that users manage their Kubernetes objects. Having a setting on Dapr which touches non-Dapr related changes is unnecessary at best, and can cause issues with CI/CD system's desired state configuration at worst.

shubham1172:
> there's no need to let a Dapr specific mutating webhook to do that

I spent some more time thinking about this change and I have come to agree with this. I had specific scenarios in mind but I realize that they should be handled differently, instead of letting the Dapr admissions controller do the same. Thank you, I am closing this PR!

Issue #5682, opened 2022-12-28
[Investigation] Pubsub latency increase when publishing large messages proportionally to the delay between publishing them


mcandeia:
## In what area(s)?

/area runtime

> /area operator

> /area placement

> /area docs

/area test-and-release

## What version of Dapr?

1.8.4
> 1.1.x
> 1.0.x
> edge: output of `git describe --dirty`

## Expected Behavior

Publishing large messages should not have any latency increase aside from data transfer.

## Actual Behavior

A Dapr user observed the following issue with PubSub (confirmed with Redis and Kafka, so appears to be a runtime issue)
This problem only occurs if the Payload is 31KB (or larger presumably), but does not occur when the payload is 2KB.
What we observe is that the more sleep time there are between dapr API pubsub calls there are, the slower the response time. 

## Steps to Reproduce the Problem

Test app to reproduce here: https://github.com/Itellissi/performance/tree/main/dapr-tests
[Here](https://discord.com/channels/778680217417809931/901141467840524289/1055516295287803904) is where the thread starts.

## Release Note

RELEASE NOTE:

mcandeia:
CPU profiling for in-memory pubsub 500 iteraction with rate 10 req/s showed no significant issues (aside from the CPU time used for the fasthttp adaptor)

<img width="919" alt="image" src="https://user-images.githubusercontent.com/5839364/209971094-3061f2d4-2284-4f29-8129-e527a59553c2.png">

mcandeia:
<img width="1299" alt="image" src="https://user-images.githubusercontent.com/5839364/209971362-df0f2770-922b-465d-9d54-793d8f1d203f.png">

In-memory benchmarks did not show any significative problem too.

mcandeia:
<img width="970" alt="image" src="https://user-images.githubusercontent.com/5839364/209971795-c6565c83-7575-4969-a2e5-4afb87b76a6b.png">

Heap usage seems ok too. Aside from the memory usage from my-sql lib that deserves investigation.

mcandeia:
![pprof001](https://user-images.githubusercontent.com/5839364/209972391-f50660a7-1940-412a-b5d3-8acae6449385.svg)

Heap web view

berndverst:
Any chance you can run pprof again with a rate of 2000 samples or so? And could you use the flame graph output? :)

mcandeia:
Sure! Let's go through the results;

[...]

PR #5683, opened 2022-12-28
Performance tests for pubsub http

mcandeia:
# Description

I've added perf-tests for the mentioned scenarios and I wasn't able to reproduce the issue with our perf tests tools.

Check the results [here](https://pastebin.com/9Sq01G9z).

## Issue reference

Please reference the issue this PR will close: #5682

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing

### DIFF
--- a/tests/perf/pubsub_publish_http/pubsub_publish_http_test.go
+++ b/tests/perf/pubsub_publish_http/pubsub_publish_http_test.go
@@ -41,36 +41,11 @@ const (
 
 var brokers = []kube.ComponentDescription{
 	{
-		Name:      "redis-broker",
+		Name:      "memory-broker",
 		Namespace: &kube.DaprTestNamespace,
-		TypeName:  "pubsub.redis",
-		MetaData: map[string]kube.MetadataValue{
-			"redisHost": {
-				FromSecretRef: &kube.SecretRef{
-					Name: "redissecret",
-					Key:  "host",
-				},
-			},
-			"redisPassword": {Raw: `""`},
-			"consumerID":    {Raw: `"myGroup"`},
-			"enableTLS":     {Raw: `"false"`},
-		},
-		Scopes: []string{k6AppName},
-	},
-	{
-		Name:      "kafka-broker",
-		Namespace: &kube.DaprTestNamespace,
-		TypeName:  "pubsub.kafka",
-		MetaData: map[string]kube.MetadataValue{
-			"brokers":              {Raw: `"dapr-kafka:9092"`},
-			"authType":             {Raw: `"none"`},
-			"maxMessageBytes":      {Raw: `"1000000"`},
-			"consumeRetryInterval": {Raw: `"200ms"`},
-			"disableTls":           {Raw: `"true"`},
-			"skipVerify":           {Raw: `"true"`},
-			"initialOffset":        {Raw: `"oldest"`},
-		},
-		Scopes: []string{k6AppName},
+		TypeName:  "pubsub.in-memory",
+		MetaData:  map[string]kube.MetadataValue{},
+		Scopes:    []string{k6AppName},
 	},
 }
 
--- a/tests/perf/pubsub_publish_http/pubsub_publish_http_test.go
+++ b/tests/perf/pubsub_publish_http/pubsub_publish_http_test.go
@@ -1,0 +1,106 @@
+//go:build perf
+// +build perf
+
+/*
+Copyright 2022 The Dapr Authors
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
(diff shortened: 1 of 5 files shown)


yaron2:
Great addition to perf tests. Since this is only testing publish, why not use the in-memory pubsub and remove all possible invariance associated with network and performance issues for Redis?

mcandeia:
> Great addition to perf tests. Since this is only testing publish, why not use the in-memory pubsub and remove all possible invariance associated with network and performance issues for Redis?

That makes sense! Initially I was trying to reproduce the same scenario pointed by the user, but I think the only thing that matters is to not have latency increases when using large messages regardless of used broker. Will change it, thanks for the tip.

mcandeia:
Done @yaron2

tests/perf/pubsub_publish_http/pubsub_publish_http_test.go
```diff
@@ -0,0 +1,81 @@
...
+
+func TestMain(m *testing.M) {
+	utils.SetupLogs("pubsub_publish_http_test")
+
+	tr = runner.NewTestRunner("pubsub_publish_http", []kube.AppDescription{}, brokers, nil)
+	os.Exit(tr.Start(m))
+}
+
+func TestPubsubPublishHttpPerformance(t *testing.T) {
+	k6Test := loadtest.NewK6("./test.js", loadtest.WithAppID(k6AppName), loadtest.WithName(k6AppName), loadtest.WithParallelism(1), loadtest.WithRunnerEnvVar(brokersEnvVar, brokersNames))
```
ItalyPaleAle, line 70: 100% a nit, but these could have been split into multiple lines

tests/perf/pubsub_publish_http/test.js
```diff
@@ -0,0 +1,107 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import http from "k6/http";
+import { check } from "k6";
+import crypto from "k6/crypto";
+
+const KB = 1000
```
ItalyPaleAle, line 18: 1000 or 1024?

tests/perf/pubsub_publish_http/test.js
```diff
@@ -0,0 +1,107 @@
...
+const MAX_MS_ALLOWED = 1
+// padd with leading 0 if <16
+function i2hex(i) {
+    return ('0' + i.toString(16)).slice(-2);
+}
+
+function randomStringOfSize(size) {
+    const bytes = crypto.randomBytes(size);
+    const view = new Uint8Array(bytes);
+    return view.reduce(function (memo, i) { return memo + i2hex(i) }, '')
```
ItalyPaleAle, line 28: https://github.com/ItalyPaleAle/arraybuffer for next time :)

tests/perf/pubsub_publish_http/test.js
```diff
@@ -0,0 +1,107 @@
...
+const samples = [200]
+
+let scenarios = {}
+let thresholds = {
+    checks: ['rate==1'],
+}
+
+for (const messageSizeKbIdx in messageSizeKb) {
+    const msgSize = messageSizeKb[messageSizeKbIdx]
+    const msgStr = randomStringOfSize(msgSize * KB)
```
ItalyPaleAle, line 49: This is technically doubling the length however, as it's hex-encoding the message of msgSize. If I'm not mistaken.

tests/perf/pubsub_publish_http/test.js
mcandeia: well it depends on the [base system used](https://en.wikipedia.org/wiki/Kilobyte), but I think we can assume base 2 and use 1024

tests/perf/pubsub_publish_http/test.js
mcandeia: It would be an option if k6 allows node packages, things gets a bit complicated when using external dependencies. Unless you want to convert your library into a k6 module, otherwise I think I'll wait for a next time yet

tests/perf/pubsub_publish_http/test.js
mcandeia: you are right, I'm afraid that using the UTF-8 byte value won't work because of encoding. So, how about divide the input by 2?

ItalyPaleAle: Sure that works

tests/perf/pubsub_publish_http/test.js
ItalyPaleAle: Yeah, we normally use base 2 throughout Dapr.

yaron2 (approved):

Issue #5685, opened 2022-12-29
Actors mocks show compile errors outside `actors` package in VS Code


dazhao-msft:
## In what area(s)?

> /area runtime

> /area operator

> /area placement

> /area docs

/area test-and-release

## What version of Dapr?

edge: output of `git describe --dirty`

## Expected Behavior

After cloning a clean repo to local workspace and opening it in VS Code, the repo should just compile and no errors are shown in VS Code.

## Actual Behavior

MockActors and FailingActors are not recognized outside the `actors` package by the language service in VS Code. This is likely because the build tag `unit` is not set during the compilation by VS Code.

## Steps to Reproduce the Problem

Clone the repo to local workspace and open it in VS Code, inspect

- pkg/grpc/api_actor_test.go
- pkg/grpc/api_test.go
- pkg/http/api_test.go

VS Code shows ~30 build errors in those files.

## Release Note

The fix will not affect runtime, no release note is needed.

RELEASE NOTE:

ItalyPaleAle:
Although I support #5686 as a fix, I also wanted to share that it's possible to configure VS Code to automatically add tags by editing the `settings.json`.  https://stackoverflow.com/a/66816676/192024

Regardless of #5686, we should consider shipping a `.vscode/settings.json` that has these tags enabled, plus `e2e`, `perf`, and other tags we use in Dapr.

PR #5686, opened 2022-12-29
Remove build tag `unit` and move `MockActors` & `FailingActors` to a separate testing package under `pkg/actors` 

dazhao-msft:
# Description

The change is to get rid of build tag `unit` and instead move `MockActors` and `FailingActors` to a separate testing package under `pkg/actors`. This also follows the same convention by other implementations of mocks.

## Issue reference

Please reference the issue this PR will close: #5685

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Unit tests passing


mukundansundar (dismissed):
lgtm.

Thanks for the contribution!

ItalyPaleAle (dismissed):

berndverst (dismissed):

dazhao-msft:
@mukundansundar @ItalyPaleAle @berndverst thanks guys for reviewing the change. Please also help merge the PR (or let me know if any additional steps are required before merge).

berndverst:
@dazhao-msft we cannot merge as we are not maintainers of this repo. All maintainers are out of the office (at their respective businesses) until next week.

mukundansundar (approved):

dazhao-msft:
@ItalyPaleAle I'm removing `buildinfo_unit.go` which was added via PR #5677, as the file introduced a new occurrence of the `unit` tag, which is being removed by this PR. I think it makes more sense to remove `buildinfo_unit.go` because unit tests shouldn't depend on feature flags to run. Instead, it should be the integration tests that verify the impacts of feature flags. In addition, all the current unit tests can pass without `buildinfo_unit.go`.

BTW, what else needs to be done to merge this PR? I notice that it has been marked as autoupdate for a while.

ItalyPaleAle:
> In addition, all the current unit tests can pass without buildinfo_unit.go.

Yes, that's correct because Resiliency is now enabled everywhere.

However, we will need a way to hardcode feature flags for unit tests I'm afraid.

dazhao-msft:
> However, we will need a way to hardcode feature flags for unit tests I'm afraid.

Not really. Individual features should be unit tested independently. If a component depends on another with feature flag, ideally the dependency should be mocked up rather than turning on/off the actual feature flag with the code like `buildinfo_unit.go`.

ItalyPaleAle:
The problem is during “migrations”, before we enable a feature flag. That’s when I added the Resiliency one, a few days before enabling it everywhere. In this case, the goal was to test that _all_ unit tests would work even with Resiliency enabled.

dazhao-msft:
> The problem is during “migrations”, before we enable a feature flag. That’s when I added the Resiliency one, a few days before enabling it everywhere. In this case, the goal was to test that _all_ unit tests would work even with Resiliency enabled.

Actually would it make sense to also append `$(GCFLAGS) -ldflags=$(LDFLAGS)` to `go test` and `gotestsum` to keep consistent with `go build`?

pkg/buildinfo/buildinfo_unit.go
```diff
@@ -1,33 +0,0 @@
...
-*/
-
-package buildinfo
-
-import (
-	"strings"
-)
-
-// Comma-separated list of features to enable in unit tests
-const unitTestFeatures = "Resiliency"
```
ItalyPaleAle, line 24: Sorry, we really need this file to be able to hardcode certain features :(

dazhao-msft: How about adding the same arguments of `ldflags` to `go test`?

I made the changes in the PR a few days ago in the Makefile.

dazhao-msft: In general, I think it's better to keep the arguments of `go build` and `go test` as consistent as possible, so it tests the same behaviors as it ships.

PR #4758, opened 2022-06-13
Add components healthcheck endpoint

DeepanshuA:
# Description

## Issue reference

Please reference the issue this PR will close: #2167 

## Checklist

Please make sure you've  completed the relevant tasks for this PR, out of the following list:

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing
* [ ] End-to-end tests passing
* [ ] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_
* [ ] Specification has been updated / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_
* [ ] Provided sample for the feature / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_

PR #5688, opened 2022-12-29
Add Universal APIs

ItalyPaleAle:
Universal APIs allow using a gRPC handler for both the Dapr gRPC and HTTP servers. Whenever possible, developers working on Dapr can now write the business logic just once, in the gRPC handler, and with a wrapper method it can then be exposed as a HTTP API too. The benefits are:

- Implementing new APIs becomes simpler
- Less maintenance burden over time, especially when we make broad changes across all APIs (e.g. when we add Resiliency, or change how it works a lot like I did recently in #5538 as an example)
- For users, there's a stronger guarantee that the output will be consistent regardless of whether they're using HTTP or gRPC.

This commit includes the base code to support Universal APIs, including:

- Support for wrapping gRPC handlers as fasthttp ones.
   - Protobuf objects in HTTP handlers are converted to/from JSON using [protojson](https://pkg.go.dev/google.golang.org/protobuf/encoding/protojson), so it's consistent with the gRPC APIs.
   - The wrapper automatically parses the JSON request body (in non-GET requests) into a protobuf object.
   - Using the `InModifier` optional argument on the wrapper, it's possible to implement RESTful HTTP APIs by overriding properties in the input protobuf object (see examples in this PR).
   - An optional `OutModifier` exists too, to allow modifying the response object before it's serialized (this should only be used in case of migrations of existing APIs, when there are small differences in output - new APIs should try to stay close to the format used in the protos for consistency).
- A new APIError object that can be used to return errors with the correct status code for both HTTP and gRPC servers

~~It also converts the secret store APIs to "Universal APIs" as proof-of-concept. As you can see from the fact that tests haven't been modified, the behavior of the Universal APIs hasn't changed for both HTTP and gRPC endpoints.~~

I do not expect us to convert all APIs to Universal (and in some case, a conversion may not even be possible due to implementation differences), so this is primarily an option for new APIs that we'll build in the future... although with efforts such as #4979 , it may turn out to be simpler to do a conversion to Universal at least in some cases.

**How a Universal API is implemented**

First, implement the gRPC as usual, with the only difference that the code should go in `pkg/grpc/universalapi` (rather than `pkg/grpc`). Doing this is both required and sufficient to make the API available as gRPC.

To then expose the API as HTTP,  in the `pkg/http` use the `UniversalFastHTTPHandler` method to wrap the gRPC API. The method's options argument allows specifying an `InModifier` for modifying the input object, for example to add parameters from the URL (as is needed for RESTful APIs). Example:

https://github.com/dapr/dapr/blob/a1b7fac4b4fbd8e849bee93d67407500b9d209ef/pkg/http/api.go#L1398-L1409

The wrapped handler can be used with the FastHTTP router.

That's all :)


dapr/proto/runtime/v1/dapr.proto
```diff
@@ -380,7 +379,7 @@ message InvokeBindingResponse {
 // GetSecretRequest is the message to get secret from secret store.
 message GetSecretRequest {
   // The name of secret store.
-  string store_name = 1;
+  string store_name = 1 [json_name = "storeName"];
```
ItalyPaleAle, line 382: By using the `json_name` annotation in protos, it's possible to define what the key in the JSON objects will be, as in this example

> Although in this specific case, this property is set by the `inModifier` method and read from the URL, so it doesn't really matter here.

pkg/http/api.go
```diff
@@ -1367,123 +1372,19 @@ func (a *api) onDeleteState(reqCtx *fasthttp.RequestCtx) {
...
-	start := time.Now()
-	policyRunner := resiliency.NewRunner[*secretstores.GetSecretResponse](reqCtx,
-		a.resiliency.ComponentOutboundPolicy(secretStoreName, resiliency.Secretstore),
-	)
-	resp, err := policyRunner(func(ctx context.Context) (*secretstores.GetSecretResponse, error) {
-		rResp, rErr := store.GetSecret(ctx, req)
-		return &rResp, rErr
+func (a *api) onGetSecretHandler() fasthttp.RequestHandler {
+	return UniversalFastHTTPHandler(a.universal.GetSecret, func(reqCtx *fasthttp.RequestCtx, in *runtimev1pb.GetSecretRequest) {
+		in.StoreName = reqCtx.UserValue(secretStoreNameParam).(string)
```
ItalyPaleAle, line 1377: This is the (optional) `inModifier` handler. Its input is the proto object decoded from the request body (if any). In order for HTTP APIs to remain RESTful, the `inModifier` allows changing the proto object that was decoded to add properties read from the URL (for example)

yaron2:
How can a universal handler control the HTTP status code returned to HTTP1.1 callers? This is a required capability.

ItalyPaleAle:
> How can a universal handler control the HTTP status code returned to HTTP1.1 callers? This is a required capability.

- If the gRPC handler returns no error, the status code is:
  - `200` if the response is a non-nil object
  - `204` if the response is a nil object
- In case of errors, the handler can return a `messages.APIError` object which includes information for both HTTP and gRPC consumers on what status code to return

You can see an example of how errors are handled in the POC for secret stores. I converted the constants in `pkg/messages` from strings to `APIError` objects, which also include the status codes:

https://github.com/dapr/dapr/blob/8d0a7530c554a07e3a5cd88eecf35705a4cc250c/pkg/messages/api_errors.go#L107-L111

Then, gRPC handlers return them as in these examples (with and without formatting):

https://github.com/dapr/dapr/blob/8d0a7530c554a07e3a5cd88eecf35705a4cc250c/pkg/grpc/universalapi/api_secrets.go#L125-L141

The situations above should cover most cases in Dapr where we need to pass custom status codes. If there are more situations, we can either:

- Adjust the wrapper to support the new cases
- Just say that specific API cannot be made as "Universal" and will need to remain implemented twice (which is totally acceptable IMHO)

berndverst:
Cool, I'll definitely want to follow this for DocumentStore

ItalyPaleAle:
Moving to ready for review as it’s in a good state now. Can still make adjustments if needed

ItalyPaleAle:
@artursouza @yaron2 what are your thoughts on this? I think it could be already useful for things like #4758 and other APIs that are being worked on / will be worked on. Can this be considered for 1.10 or does it need to be postponed till 1.11?

ItalyPaleAle:
Reverted the change to the secret store APIs (which were used as POC) so the HTTP ones are unchanged.

(for gRPC, they have still been moved to a separate file, but their code and behavior are unchanged)

ItalyPaleAle:
@yaron2 @artursouza is this good to be merged for 1.11 now? Looking forward being able to use it

yaron2 (approved):

yaron2:
Looks good

PR #5689, opened 2022-12-29
Do not start Dapr Watchdog runnable unless it's enabled

ItalyPaleAle:
Currently, the Dapr Watchdog runnable is added to the manager whether the watchdog is enabled or not. This forces the Dapr Operator service to request leadership election in all cases, even if the disable-leader-election flag is set.

This follows reports from a user on Discord who's having issues with the Operator crashing due to failed leadership election. Although we cannot reproduce that error (and there's a change it may be due to an issue with their K8s setup), this PR should help with those situations too.

### DIFF
--- a/pkg/operator/operator.go
+++ b/pkg/operator/operator.go
@@ -87,7 +87,7 @@ func init() {
 func NewOperator(opts Options) Operator {
 	conf, err := ctrl.GetConfig()
 	if err != nil {
-		log.Fatalf("unable to get controller runtime configuration, err: %s", err)
+		log.Fatalf("Unable to get controller runtime configuration, err: %s", err)
 	}
 	mgr, err := ctrl.NewManager(conf, ctrl.Options{
 		Scheme:             scheme,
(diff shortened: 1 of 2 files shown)


pkg/operator/operator.go
```diff
@@ -87,7 +87,7 @@ func init() {
 func NewOperator(opts Options) Operator {
 	conf, err := ctrl.GetConfig()
 	if err != nil {
-		log.Fatalf("unable to get controller runtime configuration, err: %s", err)
+		log.Fatalf("Unable to get controller runtime configuration, err: %s", err)
```
ItalyPaleAle, line 90: Log messages are expected to start with uppercase letters

berndverst (approved):

berndverst:
Potentially 1.9.X hotfix?

ItalyPaleAle:
> @berndverst Potentially 1.9.X hotfix?

I would not consider it for a hotfix, personally.

yaron2:
Not a hotfix.

artursouza (approved):

PR #5695, opened 2023-01-03
Adding solace_amqp component

TKTheTechie:
Signed-off-by: TKTheTechie <thomas.kunnumpurath@solace.com>

# Description

Adding a PubSub+ AMQP Component

## Issue reference

Please reference the issue this PR will close: https://github.com/dapr/components-contrib/pull/2383

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Unit tests passing
* [x] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: https://github.com/dapr/docs/issues/3022
* [x] Specification has been updated / Created issue in the https://github.com/dapr/docs/ repo: https://github.com/dapr/docs/issues/3022
* [x] Provided sample for the feature / Created issue in the https://github.com/dapr/docs/ repo:https://github.com/dapr/docs/issues/3022


berndverst (dismissed):

berndverst:
I have to revert some dependency in the contrib repo which causes the build failure here.

berndverst:
@TKTheTechie please import the latest `components-contrib@master` into your PR at this time, and run `make modtidy-all` one more time.

This should downgrade two dependencies (most noteworthy wazero) which will allow this PR to compile on all platforms again. At that time we can merge this PR.

TKTheTechie:
> @TKTheTechie please import the latest `components-contrib@master` into your PR at this time, and run `make modtidy-all` one more time.
> 
> This should downgrade two dependencies (most noteworthy wazero) which will allow this PR to compile on all platforms again. At that time we can merge this PR.

This was done

TKTheTechie:
@berndverst can you merge this now?

TKTheTechie:
@berndverst  seems like there is an incompatiblity between the wazero lib and compiling to ARM. Can you advise?

yaron2:
> @berndverst seems like there is an incompatiblity between the wazero lib and compiling to ARM. Can you advise?

cc @codefromthecrypt

go.mod
```diff
@@ -241,7 +241,7 @@ require (
 	github.com/hashicorp/go-uuid v1.0.3 // indirect
 	github.com/hashicorp/serf v0.10.1 // indirect
 	github.com/hazelcast/hazelcast-go-client v0.0.0-20190530123621-6cf767c2f31a // indirect
-	github.com/http-wasm/http-wasm-host-go v0.2.0 // indirect
+	github.com/http-wasm/http-wasm-host-go v0.3.0 // indirect
```
codefromthecrypt, line 244: this should be v0.3.1

go.mod
```diff
@@ -347,7 +347,7 @@ require (
 	github.com/tchap/go-patricia/v2 v2.3.1 // indirect
 	github.com/tencentcloud/tencentcloud-sdk-go/tencentcloud/common v1.0.557 // indirect
 	github.com/tencentcloud/tencentcloud-sdk-go/tencentcloud/ssm v1.0.557 // indirect
-	github.com/tetratelabs/wazero v1.0.0-pre.4 // indirect
+	github.com/tetratelabs/wazero v1.0.0-pre.6 // indirect
```
codefromthecrypt, line 350: this should be v1.0.0-pre.7

codefromthecrypt (commented):
the wazero related versions need to be changed for arm (32-bit) to work

berndverst:
Will look tomorrow, but really need help from @codefromthecrypt. Newer versions wazero have different problems themselves so I don't think wazero should have been upgraded at all.

Wazero pre 4 is what we used previously.

I had to revert an upgrade that was made to wazero pre 6 and pre 7 that itself caused other problems.

I wasn't aware of any issues with wazero pre.4.

berndverst:
> the wazero related versions need to be changed for arm (32-bit) to work

@codefromthecrypt pre.7 was still broken in Dapr last time I tried. Do you have a newer release or a commit that will work??

codefromthecrypt:
sorry wires crossed. indeed for arm, it is either not upgrade (roll back) or use the versions I mentioned. there was a glitch in between which was much regretted.

codefromthecrypt:
@berndverst so back when I looked at this last time, I think the problem you had was version skew between http-wasm and wazero, so v0.3.1 + pre.7 works, but you have to change both. only updating wazero without updating http-wasm won't work. https://github.com/dapr/components-contrib/compare/master...codefromthecrypt:components-contrib:wazero-1.0.0-pre.7

berndverst:
> @berndverst so back when I looked at this last time, I think the problem you had was version skew between http-wasm and wazero, so v0.3.1 + pre.7 works, but you have to change both. only updating wazero without updating http-wasm won't work. https://github.com/dapr/components-contrib/compare/master...codefromthecrypt:components-contrib:wazero-1.0.0-pre.7

Ok will try that tomorrow. Today is a holiday at big companies and I'm not working

berndverst:
This must not have been the latest contrib in this PR by the way.

Contrib@master had contained 1.0.0-pre.4 which was completely fine. There was a brief period where it was 1.0.0-pre.6 but that was over a week ago. So had this PR really pinned the latest components-contrib when we discussed this it should have been ok.

It's important to always pin the latest components-contrib@master (`go get github.com/dapr/components-contrib@master`) and then run `make modtidy-all`

Either way, now we upgraded components-contrib to 1.0.0-pre.7 with http-wasm-go 0.3.1 which appears to be compiling fine on Linux ARM. (cc @codefromthecrypt )

Maybe we should wait for the following PR to be merged before revisiting this PR here.
https://github.com/dapr/dapr/pull/5769

yaron2:
@TKTheTechie please update with master branch.

berndverst (approved):

ItalyPaleAle:
Why is this called "solace.amqp"? Normally we reserve dots for grouping components into, for example, clouds, like "azure.foo" or "aws.bar". Could this just be "solace" or "solaceamqp"? Or will there be more "solace.xx" components?

yaron2:
> Why is this called "solace.amqp"? Normally we reserve dots for grouping components into, for example, clouds, like "azure.foo" or "aws.bar". Could this just be "solace" or "solaceamqp"? Or will there be more "solace.xx" components?

@TKTheTechie would be great to get a recommendation here :)

TKTheTechie:
Yes that is correct, Solace supports multiple protocols AMQP, REST, MQTT, WebSocket and a custom protocol called SMF (Solace Message Format). We have a Go Lang implementation of SMF but unfortunately it relies on CGO which isn't supported by DAPR. In the future, if we do support a native GoLang library I do forsee us building a native connector into Solace. I do forsee a REST integration as a possibility as well.

PR #5702, opened 2023-01-04
Worklows E2E test and updates to proto file

RyanLettieri:
# Description

Updated the proto file to have the input for "StartWorkflow" request to be a byte array rather than an "any" type.
Added in Workflows E2E test.

## Issue reference

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing


dapr/proto/runtime/v1/dapr.proto
```diff
@@ -697,7 +697,7 @@ message StartWorkflowRequest {
   string workflow_component = 2;
   string workflow_name = 3;
   map<string, string> options = 4;
-  google.protobuf.Any input = 5;
```
artursouza, line 700: This will affect the SDK changes too, not? Why was this needed?

tests/e2e/workflows/dist/workflows_test.go
```diff
@@ -0,0 +1,105 @@
...
+			DaprMemoryRequest: "100Mi",
+			AppMemoryLimit:    "200Mi",
+			AppMemoryRequest:  "100Mi",
+		},
+	}
+	tr = runner.NewTestRunner("workflowsapp", testApps, nil, nil)
+	os.Exit(tr.Start(m))
+}
+
+var workflowAppTests = []struct {
```
artursouza, line 69: No need to use this structure if there is only one combination.

.github/workflows/dapr-test.yml
```diff
@@ -507,6 +507,7 @@ jobs:
           make setup-test-env-redis
           make setup-test-env-kafka
           make setup-test-env-zipkin
+          make setup-test-env-temporal
```
ItalyPaleAle, line 510: Just FYI this seems to take about 3-4 mins for me on AKS. Although out of scope for this PR, it would be nice if we could find ways to either reduce the number of pods this deploys (can we do it without Cassandra for example?) and/or maybe deploy all 3rd party stuff (Redis, Kafka, MongoDB) in parallel.

tests/apps/workflowsapp/app.go
```diff
@@ -209,10 +212,21 @@ func main() {
 	w.RegisterWorkflow(HelloTemporalWF)
 	w.RegisterActivity(HelloTemporalAct)
 
+	for i := 0; i < 5; i++ {
+		err = w.Start()
```
ItalyPaleAle, line 216: Any idea on why this needs these retries? Although this works, just wondering what the underlying issue what could cause `w.Start()` to fail would be

tests/apps/workflowsapp/app.go
```diff
@@ -242,6 +256,6 @@ func HelloTemporalWF(ctx workflow.Context) (string, error) {
 }
 
 func HelloTemporalAct(ctx context.Context) (result string, err error) {
-	time.Sleep(8 * time.Second) // This 5s sleep is to allow for terminate to be called
+	time.Sleep(30 * time.Second) // This 5s sleep is to allow for terminate to be called
```
ItalyPaleAle, line 259: Comment says 5s; it was 8s and now it's 30s.

Is there a way for us to get away without sleeping here?

1. It makes tests slower
2. It introduces flakiness: pretty much every test we have that depends on waiting for a condition to happen is flaky at some point (and I presume that's the reason why it went from 5, to 8, to 30s)

dapr/proto/runtime/v1/dapr.proto
ItalyPaleAle: Not just the Dapr client SDKs, but perhaps also the workers SDKs. `google.protobuf.Any` in the Go code creates a struct which has the property `.Value` that is a byte slice. Changing to `bytes` only, there's no struct and it's a byte slice directly.

Looks like `input` is sent as-is to components-contrib, which send it as-is to the workflow "runner".

However, I do think that `bytes` is probably better here, if we expect users to pass arbitrary JSON-encoded data.

tests/apps/workflowsapp/app.go
RyanLettieri: I'll update the comment. The time is arbitrary since a terminate call is executed which will stop the activity all together. I'm using 30s here (I'll downgrade to 10s) in order to ensure that there is enough time for the activity to start and all the calls to be made. I do agree that 30s is excessive.

.github/workflows/dapr-test.yml
RyanLettieri: Unfortunately, this is with only required dependencies: https://artifacthub.io/packages/helm/wener/temporal

artursouza (approved):

.github/workflows/dapr-3rdparty-images.yaml
```diff
@@ -67,3 +67,31 @@ jobs:
           src: openzipkin/zipkin:latest
           dst: |
             ghcr.io/${{ github.repository_owner }}/3rdparty/zipkin:latest
+      - # copy 3rd party temporal admin-tools image from dockerhub to ghcr
+        name: Push Image (temporal admin-tools) to target registries
+        uses: akhilerm/tag-push-action@v2.0.0
+        with:
+          src: temporalio/admin-tools:1.15.1
+          dst: |
+            ghcr.io/${{ github.repository_owner }}/3rdparty/temporalio-admin-tools:1.15.1
```
artursouza, line 76: What about ../3rdparty/temporalio/admin-tools:1.15.1?

Same for the others.

dapr/proto/runtime/v1/dapr.proto
artursouza: OK.

PR #5706, opened 2023-01-05
HTTP service invocation streaming: almost-complete E2E

ItalyPaleAle:
# Description

This PR is the last part of adding support for streaming in HTTP service invocation, and it's the last spin-off from #5170 

With this PR, the E2E support for streaming is almost complete. It adds:

- A new gRPC method `CallLocalStream`, which is a variant of `CallLocal` that uses bi-directional streaming to send and receive the data in chunks. This is implemented very similar to how it was in #5170, but with a couple of changes for correctness (main ones are passing a sequence number to ensure all messages are received in order, and correctly closing the streams using the proper gRPC way, rather than passing a `complete=true` field)
- On the HTTP server, responses in the handler for service invocation now send data to clients in a streamed way

It's **almost** complete because while the HTTP server responds to clients by sending a stream (and the response is now never buffered in memory), the request is still read in memory first. This is due to a limitation of fasthttp and there's not much we can do until #4979 is done. However, that is the very last part that doesn't use streams in the E2E experience.

I've gated the changes above behind a new feature flag `ServiceInvocationStreaming`.

- When the feature flag is off, sidecars will behave as today and continue to make requests to other sidecars using `CallLocal`. The HTTP server responds to service invocation requests in the current way, which involves buffering the response in-memory first.
  - Even with the feature flag off, sidecars can still respond to incoming `CallLocalStream` messages. The feature flag controls what sidecars use when they're the callers, but as callees, they will accept both options (for forward-compatibility).
- When the feature flag is on, sidecars will make requests to other sidecars using `CallLocalStream` and in the HTTP server, responses to service invocation are sent as streams.

## Issue reference

#3103 (partially fixed by this PR)
#4866

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests
* [X] Unit tests passing
* [ ] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo:
  * [ ] TODO @ItalyPaleAle 
* [ ] Add new perf tests for service invocation with non-zero payloads
  * [ ] TODO @ItalyPaleAle

### DIFF
--- a/pkg/runtime/runtime.go
+++ b/pkg/runtime/runtime.go
@@ -3101,7 +3101,7 @@ func componentDependency(compCategory components.Category, name string) string {
 func (a *DaprRuntime) startSubscriptions() {
 	// Clean any previous state
 	if a.pubsubCancel != nil {
-		a.pubsubCancel()
+		a.stopSubscriptions() // Stop all subscriptions
 	}
 
 	// PubSub subscribers are stopped via cancellation of the main runtime's context
@@ -3116,12 +3116,12 @@ func (a *DaprRuntime) startSubscriptions() {
 
 // Stop subscriptions to all topics and cleans the cached topics
 func (a *DaprRuntime) stopSubscriptions() {
-	// Stop all subscriptions by canceling the subscription context
-	if a.pubsubCancel != nil {
-		a.pubsubCancel()
+	if a.pubsubCtx == nil || a.pubsubCtx.Err() != nil { // no pubsub to stop
+		return
 	}
-	a.pubsubCtx = nil
-	a.pubsubCancel = nil
+	if a.pubsubCancel != nil {
+		a.pubsubCancel() // Stop all subscriptions by canceling the subscription context
+	}
 
 	// Remove all contexts that are specific to each component (which have been canceled already by canceling pubsubCtx)
 	a.topicCtxCancels = nil
--- a/pkg/runtime/runtime_test.go
+++ b/pkg/runtime/runtime_test.go
@@ -5469,21 +5469,16 @@ func TestGracefulShutdownPubSub(t *testing.T) {
 	mockAppChannel := new(channelt.MockAppChannel)
 	rt.appChannel = mockAppChannel
 	mockAppChannel.On("InvokeMethod", mock.MatchedBy(matchContextInterface), matchDaprRequestMethod("dapr/subscribe")).Return(fakeResp, nil)
-
 	require.NoError(t, rt.initPubSub(cPubSub))
 	mockPubSub.AssertCalled(t, "Init", mock.Anything)
 	rt.startSubscriptions()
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -132,7 +132,8 @@ BASE_PACKAGE_NAME := github.com/dapr/dapr
 LOGGER_PACKAGE_NAME := github.com/dapr/kit/logger
 
 # Comma-separated list of features to enable
-ENABLED_FEATURES ?=
+# TODO: @ItalyPaleAle remove when ServiceInvocationStreaming is finalized
+ENABLED_FEATURES ?= ServiceInvocationStreaming
```
ItalyPaleAle, line 136: I need to remove this before the PR is merged - for now it's here so E2E tests run with this to validate the PR

pkg/messaging/direct_messaging.go
```diff
@@ -233,27 +243,189 @@ func (d *directMessaging) invokeRemote(ctx context.Context, appID, appNamespace,
...
+		diag.DefaultMonitoring.ServiceInvocationResponseReceived(appID, req.Message().Method, imr.Status().Code, start)
+	}
 
+	return imr, teardown, err
+}
+
+func (d *directMessaging) invokeRemoteUnary(ctx context.Context, clientV1 internalv1pb.ServiceInvocationClient, req *invokev1.InvokeMethodRequest, opts []grpc.CallOption) (*invokev1.InvokeMethodResponse, error) {
 	pd, err := req.ProtoWithData()
 	if err != nil {
-		return nil, teardown, fmt.Errorf("failed to read data from request object: %w", err)
```
yaron2, line 245: Why remove `teardown` here?

ItalyPaleAle: `invokeRemoteUnary` and `invokeRemoteStream` are both invoked by `invokeRemote`, which returns `teardown`. See: https://github.com/dapr/dapr/pull/5706/files/2aae9dabd9c916f0c8b177104f98ade516071af2#diff-e74cf36c876319f39e2be04a7514c9cff0058ce19cad497e3f41f3f182a3ad73R246-R258

pkg/http/api.go
```diff
@@ -1682,60 +1684,87 @@ func (a *api) onDirectMessage(reqCtx *fasthttp.RequestCtx) {
...
 	// Special case for timeouts/circuit breakers since they won't go through the rest of the logic.
 	if errors.Is(err, context.DeadlineExceeded) || breaker.IsErrorPermanent(err) {
 		respond(reqCtx, withError(fasthttp.StatusInternalServerError, NewErrorResponse("ERR_DIRECT_INVOKE", err.Error())))
+		cancel()
 		return
 	}
 
-	if dmpr != nil && len(dmpr.headers) > 0 {
-		invokev1.InternalMetadataToHTTPHeader(reqCtx, dmpr.headers, reqCtx.Response.Header.Set)
+	if resp != nil {
```
yaron2, line 1723: Shouldn't headers be set even if the response is empty?

ItalyPaleAle: Addressed on Teams

yaron2 (approved):

PR #5707, opened 2023-01-05
Fix profile port

ItalyPaleAle:
Fixes a regression (typo) introduced in #5648 . This is the reason for the test failures in components-contrib (see dapr/components-contrib#2405 )

### DIFF
--- a/pkg/http/server.go
+++ b/pkg/http/server.go
@@ -157,7 +157,7 @@ func (s *server) StartNonBlocking() error {
 
 	if s.config.EnableProfiling {
 		for _, apiListenAddress := range s.config.APIListenAddresses {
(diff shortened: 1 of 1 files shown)


berndverst (approved):
Thanks for fixing!

berndverst:
Currently dapr/dapr@master is broken and cannot be used in contrib for certification tests (causing a fatal error starting the profiling server). This fixes that.

Please merge as soon as possible @daixiang0 @yaron2 @artursouza 

This will unblock others like @DeepanshuA

DeepanshuA:
Thanks @ItalyPaleAle @berndverst @daixiang0  for quick investigation, fix and merge. It really helps to move forward.

Issue #5714, opened 2023-01-05
[Proposal] (/ok-to-perf) Support running performance tests on PRs using the source branch code


mcandeia:
## In what area(s)?

> /area runtime

> /area operator

> /area placement

> /area docs

/area test-and-release
## Describe the proposal

Currently, our perf tests executes when a slash command is issue in a given PR. The perf-tests uses the [dapr-perf](https://github.com/dapr/dapr/blob/master/.github/workflows/dapr-perf.yml#L439) workflow which deploys and runs the declared tests from `master` branch.

Therefore the **workflow file** is fetched from master, meaning that the local branch changes are not considered when running perf tests, so the newly added local tests are skipped and will run only when the PR gets merged, which makes difficult to properly set the right thresholds for the tests.

This proposal suggests to migrate the test list from dapr-perf to a make-like command (that already exists) and also migrate from environment variables to hardcoded values specified in the test file.

The only difference is that by default the perf test will run with production parameters that could be overwritten by the environment variables - when specified.

PR #5715, opened 2023-01-05
Migrate perf tests from github workflow to makefile

mcandeia:

# Description

As the title says.

## Issue reference

Please reference the issue this PR will close: #5714 

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing


tests/perf/actor_timer/actor_timer_test.go
```diff
@@ -92,8 +92,12 @@ func TestMain(m *testing.M) {
...
 
 func TestActorTimerWithStatePerformance(t *testing.T) {
-	p := perf.Params()
-	// Get the ingress external url of test app
+	p := perf.Params(
+		perf.WithQPS(220),
+		perf.WithConnections(10),
+		perf.WithDuration("1m"),
+		perf.WithPayload("{}"),
+	) // Get the ingress external url of test app
```
ItalyPaleAle, line 100: Nit: newline before comment please

.github/workflows/dapr-perf.yml
```diff
@@ -436,90 +436,9 @@ jobs:
       - name: Deploy test components
         if: env.TEST_PREFIX != ''
         run: make setup-test-components
-      - name: Run Perf test service_invocation_http
+      - name: Run all performance tests
         if: env.TEST_PREFIX != ''
-        env:
-          DAPR_PERF_QPS: 1000
```
ItalyPaleAle, line 442: Is there a way to still have the ability to manually set these env vars to override the values that are now hardcoded?

Use case is if one is testing perf tests locally and wants to set different targets for whatever reason (e.g. because the hardware is different)

ItalyPaleAle (changes requested):
I love this PR, but have a couple of comments

.github/workflows/dapr-perf.yml
mcandeia: Yes, I mentioned that in the related issue, the only difference is that the hardcoded values will be used but the environment variables yet have precendence

ItalyPaleAle: Oh alright. I missed that, thanks for clarifying

artursouza (dismissed):

ItalyPaleAle (approved):

ItalyPaleAle (approved):

yaron2:
@mcandeia the changes in this PR make it extremely difficult to reason about perf test result for separate tests. The consequence here is that you can no longer see the performance test results from GitHub actions. Currently considering reverting this PR. cc @ItalyPaleAle

yaron2:
So, does this mean the warning log wasn't the culprit?

mcandeia:
No the warning log was actually the culprit, there is another issue however with a single test that is receiving timeout without even execute. So I'm trying to find out when it has started @yaron2

mcandeia:
In this PR specifically seems that is passing, will try to run it again. But I don't know if the test is flaky or the problem was introduced after this PR.

PR #5720, opened 2023-01-07
gRPC invoke: Do not return error if message is nil

ItalyPaleAle:
Should fix https://github.com/dapr/java-sdk/pull/822


yaron2 (changes requested):
Please remove the added certs.

artursouza:
I have verified that this fixed the regression identified in ITs in Java SDK.

yaron2:
Tests are falling (e2e).

artursouza:
So, there is something else then. I wonder if we just need to change the IT for Java SDK to assert for a different error code now.

ItalyPaleAle:
I am not surprised this broke some E2E tests since it did change the behavior.

Let me know what you prefer. The previous code was returning "Unknown" and although "internal" isn't necessarily better, I am not sure if unknown is good either.

Otherwise, I will look into fixing the tests for this tomorrow, if quick, or Monday :)

ItalyPaleAle:
@artursouza 

> Right now, the HTTP API returns UNKNOWN error code too.

What do you mean? Are you talking about gRPC invoke to HTTP endpoint?

ItalyPaleAle:
After discussing with Artur, we're closing this as the tests will be updated instead

PR #5728, opened 2023-01-09
fix flaky test sync issue

yaron2:
The proper fix for https://github.com/dapr/dapr/pull/5723.


mcandeia:
trying to find out which is the last pr that actor id stress tests was passing

PR #5731, opened 2023-01-09
Enable Dapr-to-app communication when apps are behind a firewall

### LINKED ISSUE #5392: "Server-less" gRPC for applications
## Problem we're trying to solve

For Dapr runtimes to communicate with applications, the application needs to have a server running, listening on an open port. Although not an uncommon practice, if we could get rid of this need, the experience of developers using Dapr would be better.

## How can we solve that?

Dapr currently has 2 app channels: HTTP and gRPC. For HTTP, there's not much we can do today, so we'll leave it as-is

For gRPC, due to the duplex nature of the protocol, we (@artursouza and I) have been investigating possible approaches to remove the need for the users' applications to have a running server. The idea is that the Dapr SDKs could always manage connections from the application to Dapr as outbound connections, which would require no listener on the application side.

The application could always create a persistent connection to the runtime at startup. When the runtime needs to send messages to the application, it would re-use that connection.

@artursouza and I completed our initial discovery phase yesterday, and we identified two possible paths to implement this. Although gRPC allows two-way communication, its architecture has a limitation in that only the caller is allowed to initiate new RPCs (either unary or streaming). In this case, we need a solution in which both the caller (app) and callee (server) can send messages over the same connection. We identified 2 possible solutions:

### Using a multiplexer

The most promising one is [hashicorp/yamux](https://github.com/hashicorp/yamux), which is a multiplexer allowing two-way communications over an existing channel (TCP, UDS, or potentially even QUIC in the future).

We have seen at least 2 (open source) implementations that demonstrate how to use gRPC on top of yamux

- https://github.com/chetan/bidi-hello
- https://github.com/dustin-decker/grpc-firewall-bypass

We like this approach for its simplicity. Requiring just a change in the gRPC `Dial/DialWithContext` function to use a connection established outside and that leverages the muxer, it allows then using the existing gRPC invocations we use today, both on the app and on the server. Essentially, this is a change at the gRPC transport layer, which has no visible impacts for applications who continue to work as before. It also allows graceful fallbacks to using a server in the app if needed.

The downside of this approach is that the yamux library is only written in Go. There are community-contributed ports for [Python](https://github.com/xtqxk/pyamux), [JavaScript](https://github.com/th-ch/yamux-js), and [Rust](https://github.com/libp2p/rust-yamux) (_this last one comes from the IPFS project :)_). Because we have SDKs in other programming languages, we would have to implement at least the _client_ part of the [yamux](https://github.com/hashicorp/yamux/blob/master/spec.md) specs in the remaining languages (in theory, using the Rust implementation via language bindings/FFI's could work too, but then we'd have to manage native code which has its own downsides).

Another option for multiplexing could be using QUIC, which would also have the advantage of being quic...ker - but not sure how the support landscape across other programming languages currently is. QUIC is also not TCP-based, so that's a change that goes deeper down the network stack.

### Using a separate gRPC stream for messages from the runtime

In this approach, nothing changes when the app needs to invoke the runtime: it will continue to make outbound connections like it does today.

In addition, at startup the application creates a persistent connection to daprd and invokes a new stream RPC and starts listening for messages.

When daprd needs to send a message to the app (pubsub message, input binding, incoming service invocation/gRPC proxy message, or health probes), it will use this stream and send the request over the stream. The app would also respond in the same stream.

The biggest benefit of this approach is of course that we don't need to change the transport layer, and we continue to use gRPC as-is.

However, there are 2 things to keep in mind:

- We are using a single streaming RPC where the server sends all messages to the app, and the app is expected to respond. This means we will need to implement our own "router" on top of that. _When an app receives a message from the server, where should that be sent: service invocation, input binding, pubsub listener, app health callback?_ Second, we will need to maintain some sort of state correlation, so when the app responds to daprd, daprd knows what that message is in response to. _Is this message an ACK to the pubsub message, a response to the health probe, etc?_ This is going to be more complicated when we start to implement methods that use gRPC streams, as at that point there will be multiple requests and responses that need to be routed (and whose correlation needs to be preserved).
- Second, this is probably going to have a negative performance impact. Over a single HTTP/2 connection, gRPC uses the HTTP/2 multiplexing to have multiple concurrent streams (in this case, "streams" is in the sense that is used in the HTTP/2 specs: it means any active RPC made by gRPC, either unary or stream-based). The default is to allow "at least 100" streams per each HTTP/2 connection. This means that when we use gRPC "normally", there can easily be 100 concurrent requests on the same connection at the same time (e.g. an app could receive up to 100 messages from pubsub in parallel, while it responds to them). When we use a single gRPC stream like in this approach, however, we are limited to having a single message on the wire at the same time. daprd doesn't need to wait for the app to respond before sending another message on the stream, so it doesn't limit scalability to 1 active message at time, but it is likely going to have _some_ negative performance impact. (Of course, if we don't deprecate the current way Dapr works, apps that are sensitive to performance can continue to have their own server, and this would be a tradeoff in the decision of using a server or going "server-less").

ItalyPaleAle:
# Description

The problem this PR tries to solve is to enable apps to receive messages from Dapr without the need for a listener, so they can work even behind a firewall.

For an initial discussion, please see the issue this PR fixes: #5392

A full demo is available at: [ItalyPaleAle/dapr-firewall-poc](https://github.com/ItalyPaleAle/dapr-firewall-poc)

Please also see a description of how this works in that page: https://github.com/ItalyPaleAle/dapr-firewall-poc#implementation-details

In order for this PR to work, each Dapr SDK will need to be updated to support the new way of connecting to Dapr too.

## Issue reference

Fixes #5392

## SDKs

- Go: https://github.com/dapr/go-sdk/pull/350


ItalyPaleAle:
👋🤖

xckai:
Indeed, it is a potential issue that could exist in reality.

Issue #5520, opened 2022-11-15
Better error handling for StateStore pluggable components


mcandeia:
## In what area(s)?

/area runtime

> /area operator

> /area placement

> /area docs

> /area test-and-release

## Describe the feature

Returning semantic errors is also part of the components contracts, and for pluggable component specifically there is no error mapping from gRPC returned errors to semantic errors, as one example you can see the [etag mismatch](https://github.com/dapr/components-contrib/blob/master/state/errors.go#L32) error. There should be a way to return an etag mismatch error from a statestore pluggable component that will be used as a etag mistmatch go error in runtime. 

A possible solution would be having a mapping from gRPC code to the semantic error.

## Release Note

RELEASE NOTE: **ADD** Support e-TAG mismatch errors for StateStore pluggable components

PR #5734, opened 2023-01-10
Pluggable components error handling

mcandeia:
# Description

Added a new Error Converter which is responsible for taking a gRPC-returned error and map to a well-known business-level  error. The gRPC errors code are defined together with the pluggable component stubs, also a function that maps from the given statuscode to a business-level error should be provided.

Those functions are reusable across methods, which means that an error can be handled by the same grpc method multiple times as well as they can be composed together even sharing the same status code.

As mentioned, errors converters are ordinary functions linked to the respective gRPC status code, so a simple map `map[code]func(grpcError error) (businessError error)` is being used for that.

## Issue reference

Please reference the issue this PR will close: #5520 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [ ] Unit tests passing


pkg/components/pluggable/errors.go
```diff
@@ -0,0 +1,54 @@
+/*
+Copyright 2022 The Dapr Authors
```
mukundansundar, line 2: nitpick 2023 in all files in the PR

pkg/components/pluggable/errors_test.go
```diff
@@ -0,0 +1,89 @@
...
+		innerCalled := 0
+		inner := ErrorConverter(func(s status.Status) error {
+			innerCalled++
+			return errors.New("")
+		})
+		composed := outer.Compose(inner)
+		err := composed(*status.New(codes.Unknown, ""))
+		require.NotNil(t, err)
+		assert.Equal(t, outerCalled, 0)
+		assert.Equal(t, innerCalled, 1)
```
mukundansundar, line 42: flip the expected and actual.

pkg/components/pluggable/errors_test.go
```diff
@@ -0,0 +1,89 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+func TestComposeErrorsConverters(t *testing.T) {
+	t.Run("compose should call outer function when the error was not converted", func(t *testing.T) {
```
mukundansundar, line 27:
```suggestion
	t.Run("compose should not call outer function when the error was converted", func(t *testing.T) {
```

pkg/components/pluggable/errors_test.go
```diff
@@ -0,0 +1,89 @@
...
+			innerCalled++
+			return errors.New("")
+		})
+		composed := outer.Compose(inner)
+		err := composed(*status.New(codes.Unknown, ""))
+		require.NotNil(t, err)
+		assert.Equal(t, outerCalled, 0)
+		assert.Equal(t, innerCalled, 1)
+	})
+	t.Run("compose should not call outer function when the error was converted", func(t *testing.T) {
```
mukundansundar, line 44:
```suggestion
	t.Run("compose should call outer function when the error was not converted", func(t *testing.T) {
```

pkg/components/pluggable/grpc.go
```diff
@@ -128,7 +152,6 @@ func NewGRPCConnectorWithDialer[TClient GRPCClient](dialer GRPCConnectionDialer,
 	}
 }
 
-// NewGRPCConnector creates a new grpc connector for the given client factory and socket.
```
mukundansundar, line 131: please add relevant comment on what the function represents ....

pkg/components/pluggable/grpc.go
```diff
@@ -34,7 +35,30 @@ type GRPCClient interface {
...
+		}
+		convert, ok := errorsConverters[s.Code()]
+		if !ok {
+			return err
+		}
+		return convert(*s)
+	}
+}
+
+type GRPCConnectionDialer func(ctx context.Context, name string, opts ...grpc.DialOption) (*grpc.ClientConn, error)
```
philliphoff, line 54: Are these changes part of the error handling PR or just a separate branch/merge artifact?

mcandeia: I can remove that if necessary.

I was using it previously, because I tried to implement the error converter as a gRPC interceptor, but it ended up in being more complicated than it should be. 

Would you mind if I keep it? I think it still helpful

pkg/components/pluggable/errors.go
```diff
@@ -0,0 +1,54 @@
...
+	"google.golang.org/grpc/status"
+)
+
+type ErrorConverter func(status.Status) error
+
+// Compose together two errors converters by applying the inner first and if the error was not converted, then it applies to the outer.
+func (outer ErrorConverter) Compose(inner ErrorConverter) ErrorConverter {
+	return func(s status.Status) error {
+		err := inner(s)
+		st, ok := status.FromError(err)
```
tmacam, line 27: So an error is only _valid_ if it has an status message, right? That is what I got from here and `TestComposeErrorsConverters`

pkg/components/state/pluggable.go
```diff
@@ -35,6 +40,106 @@ var (
...
+	// affectedRowsMetadataKey is the metadata key used to return bulkdelete mismatch errors affected rows.
+	affectedRowsMetadataKey = "affected"
+	// expectedRowsMetadataKey is the metadata key used to return bulkdelete mismatch errors expected rows.
+	expectedRowsMetadataKey = "expected"
+)
+
+// etagErrFromStatus get the etag error from the given gRPC status, if the error is not an etag kind error the return is the original error.
+func etagErrFromStatus(s status.Status) (error, bool) {
+	details := s.Details()
+	if len(details) != 1 {
```
philliphoff, line 62: I assume it's intentional that an "ETag error" would only ever have a single details, with a single field violation as opposed to being potentially mixed with other kinds of invalid arguments.  That likely makes sense, as I'm not sure how the runtime would coalesce such errors into a single, semantically clear error, or respond to it.

pkg/components/pluggable/grpc.go
philliphoff: It's fine with me; I just wanted to be sure that I hadn't missed something.

pkg/components/state/pluggable.go
```diff
@@ -35,6 +40,106 @@ var (
...
+}
+
+var bulkDeleteErrors = pluggable.MethodErrorConverter{
+	GRPCCodeBulkDeleteRowMismatch: func(s status.Status) error {
+		details := s.Details()
+		if len(details) != 1 {
+			return s.Err()
+		}
+		errorInfoDetail, ok := details[0].(*errdetails.ErrorInfo)
+		if !ok {
```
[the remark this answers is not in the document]
mcandeia: check if my answer of the previous comment made sense for this as well, I guess yes

pkg/components/state/pluggable.go
philliphoff: I think it could go either way, just depending on how we want to classify ETag errors. If an ETag error is just a more specific type of "invalid argument" error, then I would tend to think it make sense that it could be just one of potentially several invalid argument errors returned as a bundle (for the caller to iterate through).  If an ETag error is considered a higher-level semantic error, considered and returned only after all lower-level errors (like invalid argument errors), then it would make sense to insist that it exist alone.

I'm not familiar enough with Go error handling to know its conventions with respect to bundles of returned errors.  .NET has its `AggregateException` where it's normal to look through and pull out specific kinds of exceptions, whereas with JS/TS, not so much.

mcandeia: I think my answer to Tiago might serve for this thread as well: https://github.com/dapr/dapr/pull/5734#discussion_r1084312145

Overall, I feel like we don't have any strong reason to support aggregating up errors, especially that they will end up of being transmitted over http/grpc protocol, so it's going to be serialized and having different treatment depending on the language runtime

philliphoff (approved):

artursouza (approved):

Issue #5537, opened 2022-11-17
Dapr Workflow Building Block / Engine


nyemade-uversky:
[Workflow proposal](https://github.com/dapr/dapr/issues/4576)

### Alpha (1.10 Release Target)

* Built-in workflow engine will be available as an initial preview feature
* API will support workflow management for 3 methods (START, STOP, GET STATUS)
* API will be released as alpha
* API Support for Temporal as a workflow engine
* API Support for Built-in engine 
* Authoring SDK will support: .NET 
* Management API SDK will support: .NET
* Documentation
* Code Samples

### Beta

* API Support for Logic Apps as a workflow engine
* Authoring SDK will support: Python, Java and Go SDKs
* Management API SDK will support: Python, Java and Go SDKs
* Workflow engine will be able to support a significant number of concurrent workflows 
* Performance tests will be used to set a baseline benchmark for concurrent workflow loads
* Stress testing in place based on realistic user scenarios 

### Stable

* No new functionality planned before stable release

### Alpha

- [ ] Minimum of one core SDK supports this feature (.NET / Python / Go / Java)
- [ ] Feature documentation added to `dapr/docs`
- [ ] Telemetry data (metrics) available for this feature
- [ ] Issue opened in `dapr/quickstarts` for quickstart examples to be created
- [ ] Both HTTP and gRPC protocols implemented
- [ ] HTTP API documentation added to the `Reference` section of Dapr documentation
- [ ] Interfaces to be used by `dapr/dapr` code defined and agreed upon
- [ ] New building block package is defined in `components-contrib` repo
- [ ] Conformance tests validating the components compliance added
- [ ] Minimum of _one_ implementation (preferably something already in-use such as Redis if possible to reduce complexity)

### Beta

- [ ] E2E tests are up-to-date and comprehensive
- [ ] SDK spec is updated
- [ ] No major changes to the API have occurred in the last XXX time period (releases? months?)
- [ ] Support in core SDKs
   - [ ] Python
   - [ ] Go
   - [ ] Java
   - [ ] .NET
- [ ] Documentation up-to-date with any new changes since Alpha
- [ ] Quickstarts have been created
- [ ] Performance tests exist but do not block builds
- [ ] Conformance tests updated to match any API changes that have been made
- [ ] Conformance tests exercise both positive and negative cases 
- [ ] Minimum of N (three?) implementations of this building block 
- [ ] Certification tests for implementations 
- [ ] APIs that are used in the building block also meet Beta criteria

[...]

PR #5736, opened 2023-01-10
Internal workflow engine feature flag

johnewart:
# Description

Adds a feature flag for workflow engine to ensure that it doesn't get enabled unless explicitly turned on.

## Issue reference

Part of the alpha release of workflows tracked in #5537

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Unit tests passing

### DIFF
--- a/pkg/actors/internal/placement.go
+++ b/pkg/actors/internal/placement.go
@@ -113,7 +113,8 @@ func NewActorPlacement(
 	}
 }
 
-// Force send an update to placement service (i.e late registration of internal actors)
+// Register an actor type by adding it to the list of known actor types (if it's not already registered)
+// The placement tables will get updated when the next heartbeat fires
 func (p *ActorPlacement) AddHostedActorType(actorType string) error {
 	for _, t := range p.actorTypes {
 		if t == actorType {
--- a/pkg/actors/internal/placement.go
+++ b/pkg/actors/internal/placement.go
@@ -14,6 +14,7 @@ limitations under the License.
 package internal
 
 import (
+	"fmt"
 	"net"
 	"sync"
 	"sync/atomic"
@@ -112,6 +113,18 @@ func NewActorPlacement(
 	}
 }
 
+// Force send an update to placement service (i.e late registration of internal actors)
+func (p *ActorPlacement) AddHostedActorType(actorType string) error {
+	for _, t := range p.actorTypes {
+		if t == actorType {
+			return fmt.Errorf("actor type %s already registered", actorType)
+		}
+	}
+
+	p.actorTypes = append(p.actorTypes, actorType)
+	return nil
+}
+
 // Start connects placement service to register to membership and send heartbeat
 // to report the current member status periodically.
 func (p *ActorPlacement) Start() {
@@ -198,7 +211,6 @@ func (p *ActorPlacement) Start() {
 				Load:     1, // Not used yet
 				// Port is redundant because Name should include port number
 			}
-
 			err := p.client.send(&host)
 			if err != nil {
 				diag.DefaultMonitoring.ActorStatusReportFailed("send", "status")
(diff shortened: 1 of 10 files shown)


yaron2:
We don't require a feature flag for alpha building blocks. Case in point, Configuration API and Distributed Lock.

ItalyPaleAle:
Wait, this is not a feature flag but rather just a CLI flag.

Why do we need a CLI flag to enable this? Does it alter the behavior of the runtime in some ways that it could cause conflicts with other parts of Dapr if it's enabled?

pkg/actors/internal/placement.go
```diff
@@ -112,6 +113,18 @@ func NewActorPlacement(
 	}
 }
 
+// Force send an update to placement service (i.e late registration of internal actors)
```
cgillum, line 116: Q: Which part of the implementation forces the update to the placement service?

cgillum (approved):
LGTM. Just had one question related to a comment in placement.go.

pkg/actors/internal/placement.go
johnewart: I will change that, it doesn't actually force an update, the update will happen with the heartbeat

pkg/runtime/cli.go
```diff
@@ -363,6 +365,11 @@ func FromFlags() (*DaprRuntime, error) {
 		globalConfig = daprGlobalConfig.LoadDefaultConfiguration()
 	}
 
+	if !daprGlobalConfig.IsFeatureEnabled(globalConfig.Spec.Features, daprGlobalConfig.WorkflowEngine) && *enableWorkflowEngine {
```
yaron2, line 368: @ItalyPaleAle your comment is correct about the CLI flag, but it seems to be that AND the feature preview.

johnewart:
> Why do we need a CLI flag to enable this? Does it alter the behavior of the runtime in some ways that it could cause conflicts with other parts of Dapr if it's enabled?

Since the engine will create and register internal actors that run constantly and do work with this enabled, the intention was to require that people turn it on (and you are right it looks like I forgot a change that checks `daprGlobalConfig` -- I must have stashed it while doing something else, I just pushed it after the tests finished). While the chances are low, it would mean that the blast radius of a bug in the workflow engine itself would be limited to those who explicitly opt-in to . 

Also I just realized the title should explicitly call out that it's for the _engine_, not the building block (also re: @yaron2's comment)

yaron2:
> Wait, this is not a feature flag but rather just a CLI flag.
> 
> Why do we need a CLI flag to enable this? Does it alter the behavior of the runtime in some ways that it could cause conflicts with other parts of Dapr if it's enabled?

Very good questions indeed. From what I observe, since Workflow Engine uses actors, then the runtime will try to initialize actors (and look for a compatible state store) by default unless instructed otherwise.

johnewart:
Changed the title to reflect that the engine is the part in question, not the building block itself

yaron2:
> Also I just realized the title should explicitly call out that it's for the engine, not the building block (also re: @yaron2's comment)

Thanks, but I'm not sure we want to separate the built-in Workflows from the management APIs. Essentially they both fall under the Workflows building block. Need to give this some thought.

johnewart:
The management API can be used, for example, to manage Temporal workflows so it could exist without the internal engine. As an alternative we can change it so that the workflow engine only gets started if a built-in workflow component exists, which is mostly the same logic as here only we execute it when the workflow component is initialized instead of based on a flag.

yaron2:
> As an alternative we can change it so that the workflow engine only gets started if a built-in workflow component exists

This sounds like a much better approach.

johnewart:
Sounds good, I will update the PR with the component-registration bits as well

cgillum:
> As an alternative we can change it so that the workflow engine only gets started if a built-in workflow component exists

I'd prefer we not go this route. The current design allows users to use the workflow engine even without using registering an API component. For example, a Durable Task SDK can interact with workflows directly on the workflow gRPC endpoint (this is how all my demos work today since we don't yet have the built-in API component). I'd hate to take that capability away if there isn't a really good reason to do so.  My thinking is that we keep the building block API and the engine decoupled.

yaron2:
> > As an alternative we can change it so that the workflow engine only gets started if a built-in workflow component exists
> 
> I'd prefer we not go this route. The current design allows users to use the workflow engine even without using registering an API component. For example, a Durable Task SDK can interact with workflows directly on the workflow gRPC endpoint (this is how all my demos work today since we don't yet have the built-in API component). I'd hate to take that capability away if there isn't a really good reason to do so. My thinking is that we keep the building block API and the engine decoupled.

I'm sure I'm missing something then - just so we're all clear: what would a "built-in workflow component" enable that using the workflow engine without an API component doesn't?

Issue #5735, opened 2023-01-10
Add support for xk6 disruptor on perf tests


mcandeia:
## In what area(s)?

> /area runtime

> /area operator

> /area placement

> /area docs

/area test-and-release
## Describe the proposal
[XK6 disruptor](https://github.com/grafana/xk6-disruptor) is a k6 extension for chaos testing, introducing intermittent failures and extended delays to the current APIs.

PR #5737, opened 2023-01-10
Support xk6 disruption extension on perf tests

mcandeia:
# Description

I've changed the double activation test to include a disruption agent introducing intermittent failures during the test.

Now the double activation test has two scenarios being executed, the `base` which is disruption free, and `faulted` which is using the k6 disruption agent.

## Issue reference

Please reference the issue this PR will close: #5735

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing


tests/apps/perf/k6-custom/Dockerfile
```diff
@@ -2,7 +2,7 @@
 FROM golang:1.18.1 as builder
```
ItalyPaleAle, line 2: 1.19?

tests/perf/actor_double_activation/actor_double_activation_test.go
```diff
@@ -2,7 +2,7 @@
 // +build perf
 
 /*
-Copyright 2021 The Dapr Authors
+Copyright 2022 The Dapr Authors
```
ItalyPaleAle, line 5: If you're changing this, then it should be 2023 :)

tests/perf/actor_double_activation/test.js
```diff
@@ -66,3 +85,26 @@ export function handleSummary(data) {
...
+// add disruption
+
+const selector = {
+  namespace: __ENV.TEST_NAMESPACE,
+  select: {
+    labels: {
+      testapp: __ENV.TEST_APP_NAME
+    }
+  }
+}
```
ItalyPaleAle, line 98: Semi here

Maybe we should add prettier to the codebase...

tests/perf/actor_double_activation/actor_double_activation_test.go
mcandeia: Actually it was wrong because I added this file in 2022 but the license was wrongly pointed to 2021. So I guess I can keep it, right?

ItalyPaleAle: It doesn't matter, it was just a nit :)

tests/perf/actor_double_activation/test.js
mcandeia: I think js formatting is less important here as this is used only for tests and generally speaking these files should not have more than 100 lines with no external imports/requires.

However, considering that the effort to add prettier config and a github action presumably is really small we should do that, can I do this later? Will create an issue to track this

ItalyPaleAle: Yes, can you do that in a separate PR?

It's just because we're now adding a lot of JS code and some consistency would not be bad.

daixiang0:
seems test fail :(

mcandeia:
> seems test fail :(
Yea but I could validate that my changes actually worked

tests/perf/actor_double_activation/test.js
mcandeia: Yes, I can do that, thanks!

Issue #5673, opened 2022-12-23
Running  unit tests for `go sendSigterm(rt)` to cause panic.


1046102779:
## In what area(s)?

> /area runtime

> /area test-and-release

## What version of Dapr?

master

## Affected PRs

https://github.com/dapr/dapr/pull/5388

## Which PR merge caused the problem

https://github.com/dapr/dapr/pull/5562

## Actual Behavior

https://github.com/dapr/dapr/actions/runs/3754035944/jobs/6377885100

```shell
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0x1900a25]

goroutine 891 [running]:
testing.tRunner.func1.2({0x1ab4c20, 0x2f76ed0})
	/opt/hostedtoolcache/go/1.19.4/x64/src/testing/testing.go:1396 +0x24e
testing.tRunner.func1()
	/opt/hostedtoolcache/go/1.19.4/x64/src/testing/testing.go:1399 +0x39f
panic({0x1ab4c20, 0x2f76ed0})
	/opt/hostedtoolcache/go/1.19.4/x64/src/runtime/panic.go:884 +0x212
github.com/dapr/dapr/pkg/runtime.TestGracefulShutdownPubSub(0x0?)
	/home/runner/work/dapr/dapr/pkg/runtime/runtime_test.go:5405 +0xac5
testing.tRunner(0xc000979d40, 0x1e4c640)
	/opt/hostedtoolcache/go/1.19.4/x64/src/testing/testing.go:1446 +0x10b
created by testing.(*T).Run
	/opt/hostedtoolcache/go/1.19.4/x64/src/testing/testing.go:1493 +0x35f
```

runtime/runtime_test.go
```golang
go sendSigterm(rt)

<-rt.pubsubCtx.Done()
```

```golang
 func (a *DaprRuntime) Shutdown(duration time.Duration) {
      ........
      a.stopSubscriptions()
      ........
}

func (a *DaprRuntime) stopSubscriptions() {
          ........
          a.pubsubCtx = nil
          a.pubsubCancel = nil
         ........
}
```

So there is a problem here that we cannot use asynchronous code

ItalyPaleAle:
[...]

PR #5739, opened 2023-01-11
fixed go test -race errors for runtime test

akhilac1:

# Description

As part of updating the graceful shutdown behaviour in dapr runtime, tests were added to validate the proper closure of pubsub, actors and bindings. These tests listen for context.Done() for respective contexts. In bindings, the context is set to nil after closure, to support a validation within runtime. This triggers a race condition as the context is simultaneously being set to nil and read in test for assertion. 

This PR mitigates this by listening for timeout after grace interval expiry and checking if the context is set to nil.

## Issue reference

Please reference the issue this PR will close: #5673 

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests
* [X] Unit tests passing


akhilac1:
@ItalyPaleAle @mukundansundar - tagging for review

pkg/runtime/runtime_test.go
```diff
@@ -5466,15 +5456,17 @@ func TestGracefulShutdownBindings(t *testing.T) {
...
 	assert.Equal(t, len(rt.outputBindings), 1)
-
 	rt.running.Store(true)
+	rt.inputBindingsCtx, rt.inputBindingsCancel = context.WithCancel(rt.ctx)
 	go sendSigterm(rt)
-	<-time.After(rt.runtimeConfig.GracefulShutdownDuration)
-	assert.Nil(t, rt.inputBindingsCancel)
-	assert.Nil(t, rt.inputBindingsCtx)
+	select {
+	case <-rt.inputBindingsCtx.Done():
```
mukundansundar, line 5465: does this cause any race condition? 
Can you run race check against this change?

pkg/runtime/runtime_test.go
```diff
@@ -5421,30 +5421,20 @@ func TestGracefulShutdownPubSub(t *testing.T) {
...
 	mockPubSub.AssertCalled(t, "Init", mock.Anything)
 	rt.startSubscriptions()
 	mockPubSub.AssertCalled(t, "Subscribe", mock.AnythingOfType("pubsub.SubscribeRequest"), mock.AnythingOfType("pubsub.Handler"))
-	assert.NotNil(t, rt.pubsubCtx)
-	assert.NotNil(t, rt.topicCtxCancels)
-	assert.NotNil(t, rt.topicRoutes)
+	assert.Nil(t, rt.pubsubCtx.Err())
 	rt.running.Store(true)
 	go sendSigterm(rt)
-	select {
```
ItalyPaleAle, line 5434: Rather than deleting this and adding a timeout, which makes tests slower, how about you run this `select` before invoking `sendSigterm`, in a goroutine?

```go
rt.runtimeConfig.GracefulShutdownDuration = 2 * time.Second
	testDoneCh := make(chan struct{}, 1)
	doneCh := rt.pubsubCtx.Done()
	go func() {
		select {
		case <-doneCh:
			// nop
		case <-time.After(rt.runtimeConfig.GracefulShutdownDuration + 2*time.Second):
			assert.Fail(t, "pubsub shutdown timed out")
		}
		testDoneCh <- struct{}{}
	}()
	sendSigterm(rt)
	<-testDoneCh
	assert.Nil(t, rt.pubsubCtx)
	assert.Nil(t, rt.topicCtxCancels)
	assert.Nil(t, rt.topicRoutes)
```

This passes with `go test -race`

akhilac1: [log.txt](https://github.com/dapr/dapr/files/10403210/log.txt)

Attached log for go test -race

pkg/runtime/runtime_test.go
akhilac1: [log.txt](https://github.com/dapr/dapr/files/10403234/log.txt)
Attached log for go test -race

pkg/runtime/runtime_test.go
akhilac1: @ItalyPaleAle - in runtime.go the pubsubctx is being set to nil -
https://github.com/dapr/dapr/blob/44dd89240e2c103a7f2e5ff5c768f89826168f7d/pkg/runtime/runtime.go#L3081-L3094

From blame I see that you have set pubsubCtx to nil, which is causing the race in the test. Any specific reason why the context needs to be set to nil? Can we rather use pubsubCtx.Err() to check if context is active or not?

akhilac1: A bit of digging helped understand that stopSubscriptions is called irrespective of whether the subscriptions have been initialised or not. hence the nil check helps conditionally invoke cancel. If we could replace
```code
	if a.pubsubCancel != nil {
	 	a.pubsubCancel()
	 }
         pubsubCtx = nil
         pubsubCancel = nil
```  
with 
```code
	if a.pubsubCtx == nil {
		return
	}
	if a.pubsubCtx.Err() == nil {
		a.pubsubCancel()
	}
        . . .
        . . .
```
we could mitigate the need to set pubsubctx to nil. This will eliminate race  condition in any code that uses the pubsubctx variable. WDYT?
Few assumptions : 
- Context can not be revived after cancelling. So not setting this to nil does not impact functionally
- As soon as the context is cancelled, all child contexts will be cancelled 
- every usage of the context should check for pubsubctx.Err() == nil to ensure context is active and not depend on nil check

akhilac1:
Updated the fix as follows:
<b>runtime.go  - stopSubscriptions </b>
pubsubCtx is set to nil after context Cancellation. This takes away the ability to validate if the context was cancelled and the reason for cancellation in the tests or any downstream. To validate the state of the context a nil check should not be used. Rather ctx.Err() should be validated - [Reference](https://pkg.go.dev/context)
```
// If Done is not yet closed, Err returns nil.
// If Done is closed, Err returns a non-nil error explaining why:
// Canceled if the context was canceled
// or DeadlineExceeded if the context's deadline passed.
// After Err returns a non-nil error, successive calls to Err return the same error.
Err() error
```
<b>runtime.go - startSubscriptions</b>
To reset the context and caches cancelContext is called. Changed this to invoke stopSubscriptions so that all the related state/variables is properly reset.

```code
	if a.pubsubCancel != nil {
		a.pubsubCancel()
	}
```
 is updated to
``` code
	if a.pubsubCancel != nil {
		a.stopSubscriptions() // Stop all subscriptions
	}
```
This ensures that aside from properly cancelling the context, topicCtxCancels and topicRoutes are also reset. These are again set to new values in subsequent code in startSubscription

<b>runtime_test.go</b>
Updated test to check for graceful shutdown of pubsub without triggering race condition.

mukundansundar (approved):
lgtm

mukundansundar:
@ItalyPaleAle any concerns with the `stopSubscriptions` call in the `startSubscriptions` before starting the subscriptions and not setting `pubsubContext` to nil?

pkg/runtime/runtime.go
```diff
@@ -3079,11 +3079,12 @@ func (a *DaprRuntime) startSubscriptions() {
 
 // Stop subscriptions to all topics and cleans the cached topics
 func (a *DaprRuntime) stopSubscriptions() {
-	// Stop all subscriptions by canceling the subscription context
+	if a.pubsubCtx == nil || a.pubsubCtx.Err() != nil { // no pubsub to stop
```
ItalyPaleAle, line 3082: Sadly this method is not concurrency-safe. If 2 goroutines call `stopSubscriptions` at the same time, one could set `pubsubCancel = nil` while the other one is trying to invoke `a.pubsubCancel`

I think it's fine to just call `pubsubCancel` multiple times and not set it to nil.

pkg/runtime/runtime_test.go
```diff
@@ -5469,30 +5469,25 @@ func TestGracefulShutdownPubSub(t *testing.T) {
...
 	mockAppChannel.On("InvokeMethod", mock.MatchedBy(matchContextInterface), matchDaprRequestMethod("dapr/subscribe")).Return(fakeResp, nil)
-
 	require.NoError(t, rt.initPubSub(cPubSub))
 	mockPubSub.AssertCalled(t, "Init", mock.Anything)
 	rt.startSubscriptions()
 	mockPubSub.AssertCalled(t, "Subscribe", mock.AnythingOfType("pubsub.SubscribeRequest"), mock.AnythingOfType("pubsub.Handler"))
-	assert.NotNil(t, rt.pubsubCtx)
-	assert.NotNil(t, rt.topicCtxCancels)
-	assert.NotNil(t, rt.topicRoutes)
+	assert.Nil(t, rt.pubsubCtx.Err())
```
ItalyPaleAle, line 5476:
```suggestion
	assert.NoError(t, rt.pubsubCtx.Err())
```

(using `Error`/`NoError` shows better logs if the assertion fails)

pkg/runtime/runtime_test.go
```diff
@@ -5469,30 +5469,25 @@ func TestGracefulShutdownPubSub(t *testing.T) {
...
 	rt.running.Store(true)
 	go sendSigterm(rt)
 	select {
 	case <-rt.pubsubCtx.Done():
-		assert.Nil(t, rt.pubsubCtx)
-		assert.Nil(t, rt.topicCtxCancels)
-		assert.Nil(t, rt.topicRoutes)
+		assert.NotNil(t, rt.pubsubCtx.Err()) // check that the context is not active
```
ItalyPaleAle, line 5481:
```suggestion
		assert.NoError(t, rt.pubsubCtx.Err()) // check that the context is not active
```

ItalyPaleAle (changes requested):

pkg/runtime/runtime.go
mukundansundar: Good point. The same issue as to setting pubsubCtx to nil also exists here.

pkg/runtime/runtime_test.go
akhilac1: @ItalyPaleAle - I am actually expecting an Error in this case as the context would have been cancelled during stopSubscription - here Err() method provides the cancellation reason if the context is not active and hence the check for notNil.
I have updated this to assert for the specific error - 
```code
	case <-rt.pubsubCtx.Done():
		assert.Error(t, rt.pubsubCtx.Err(), context.Canceled)
```

pkg/runtime/runtime.go
akhilac1: @ItalyPaleAle @mukundansundar - if this was a possibility with pubsubCancel, would go test -race not flag this?

ItalyPaleAle: Only if a test was built around those methods

mukundansundar (approved):
lgtm

akhilac1:
@ItalyPaleAle  - can you please mark the comment as resolved if appropriate.
@artursouza @yaron2 - can you please review this PR so that it can be merged. Thanks!

ItalyPaleAle (approved):

PR #5744, opened 2023-01-11
Set perf-tests timeout to 1hr

mcandeia:

# Description

1hr is been the value used for when running tests individually. We should keep the same number, there are tests failing due to timeouts which defaults to 10m.

## Issue reference

N/A

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing

### DIFF
--- a/tests/dapr_tests.mk
+++ b/tests/dapr_tests.mk
@@ -351,7 +351,7 @@ ifeq ($(DAPR_PERF_TEST),)
 		--junitfile $(TEST_OUTPUT_FILE_PREFIX)_perf.xml \
 		--format standard-quiet \
 		-- \
-			-p 1 -count=1 -v -tags=perf ./tests/perf/...
+			-timeout 1h -p 1 -count=1 -v -tags=perf ./tests/perf/...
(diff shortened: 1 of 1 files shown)


yaron2:
Isn't 1hr too long? What could take so long?

mcandeia:
> Isn't 1hr too long? What could take so long?

@yaron2 The tests generally took less than 15 minutes each, but this was the value that has been used previously. You can see [here](https://github.com/dapr/dapr/blob/master/tests/dapr_tests.mk#L328)

image

mcandeia:
Ok, I realized that this timeout is setting the binary test timeout. So It means that the whole test can run up to an hour long, which seems reasonable to me, WDYT?

yaron2:
> Ok, I realized that this timeout is setting the binary test timeout. So It means that the whole test can run up to an hour long, which seems reasonable to me, WDYT?
> 
> Previously it was 1hr per-test and now the entire binary is considered.

Ok that makes sense.

yaron2 (approved):

yaron2:
Approved, but lets be watchful and see if tests are timing out for over an hour and clogging the infra.

PR #5324, opened 2022-10-08
Deprecation notice for gRPC service invocation API

sunzhaochang:
Signed-off-by: sunzhaochang <zhchsun1992@gmail.com>

# Description

Deprecation notice for gRPC service invocation API

## Issue reference

Please reference the issue this PR will close: #4252 

## Checklist

Please make sure you've  completed the relevant tasks for this PR, out of the following list:

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing
* [ ] End-to-end tests passing
* [ ] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_
* [ ] Specification has been updated / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_
* [ ] Provided sample for the feature / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_

PR #5746, opened 2023-01-12
[DO NOT MERGE] Investigating performance regressions on service-invocation-grpc

mcandeia:

# Description

## Issue reference

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing


mcandeia:
@yaron2 seems that [this](https://github.com/dapr/dapr/pull/5324/files#diff-ce3f68305f0eb28b6b8911c281186536c80d8cac9dff3cece2c8bdf3dd955315R451) is the root cause of the lantency increases

[The test](https://github.com/dapr/dapr/blob/master/tests/perf/service_invocation_grpc/service_invocation_grpc_test.go#L204) is asserting that the latency on p90 is <=2.5ms.

Considering that our logs are allocating memory (thanks to @ItalyPaleAle to confirm that) and logging itself is not cheap (i.e more memory to flush, GC pressure, etc), that makes sense to me. Coincidentally, @ItalyPaleAle has already opened a PR to fix this (for other reasons), that should be sufficient. Let's wait for the tests results [there](https://github.com/dapr/dapr/pull/5740).

yaron2:
> @yaron2 seems that [this](https://github.com/dapr/dapr/pull/5324/files#diff-ce3f68305f0eb28b6b8911c281186536c80d8cac9dff3cece2c8bdf3dd955315R451) is the root cause of the lantency increases
> 
> [The test](https://github.com/dapr/dapr/blob/master/tests/perf/service_invocation_grpc/service_invocation_grpc_test.go#L204) is asserting that the latency on p90 is <=2.5ms.
> 
> Considering that our logs are allocating memory (thanks to @ItalyPaleAle to confirm that) and logging itself is not cheap (i.e more memory to flush, GC pressure, etc), that makes sense to me. Coincidentally, @ItalyPaleAle has already opened a PR to fix this (for other reasons), that should be sufficient. Let's wait for the tests results [there](https://github.com/dapr/dapr/pull/5740).

We are deprecating the method, and thus we must log an entry for it. I'm not sure removing the log is the correct solution.

mcandeia:
He's not removing the log however, he's just logging it only once instead of per-request, If I'm not wrong: https://github.com/dapr/dapr/pull/5740/files#diff-ce3f68305f0eb28b6b8911c281186536c80d8cac9dff3cece2c8bdf3dd955315R423

yaron2:
Yeah I saw it, will comment there.

yaron2:
> Perhaps a proper solution would be update the test to use the new method?

That needs to be done anyway, as it's a deprecated path now and will be removed in 1.12.

PR #5747, opened 2023-01-12
include publish metadata in cloud event extensions

berndverst:

Today we are not sending the custom PubSub Publish Metadata via GRPC Topic Event requests (events delivered to the user's app).

This PR adds this user provided metadata to the new `Extensions` struct on the proto. We use a a prefix `_metadata_` to avoid collisions with keys in the case where a user manually provided a cloud event instead of using the Dapr event envelope.

Here is a PR that shows how to consume this in Python SDK
https://github.com/dapr/python-sdk/pull/524


pkg/runtime/runtime.go
```diff
@@ -2318,16 +2318,20 @@ func (a *DaprRuntime) publishMessageGRPC(ctx context.Context, msg *pubsubSubscri
...
+	// Merge in the pubsub publish metadata with prefix _metadata_
 
 	extensions := map[string]interface{}{}
 	for key, value := range cloudEvent {
 		if !cloudEventDuplicateKeys.Has(key) {
 			extensions[key] = value
 		}
 	}
+	for key, value := range metadata {
+		extensions[fmt.Sprintf("_metadata_%s", key)] = value
```
berndverst, line 2333:
```suggestion
		extensions[fmt.Sprintf("_metadata_" + key)] = value
```

pkg/runtime/runtime.go
```diff
@@ -2318,16 +2318,20 @@ func (a *DaprRuntime) publishMessageGRPC(ctx context.Context, msg *pubsubSubscri
...
+	// Merge in the pubsub publish metadata with prefix _metadata_
 
 	extensions := map[string]interface{}{}
 	for key, value := range cloudEvent {
 		if !cloudEventDuplicateKeys.Has(key) {
 			extensions[key] = value
 		}
 	}
+	for key, value := range metadata {
+		extensions[fmt.Sprintf("_metadata_" + key)] = value
```
ItalyPaleAle, line 2333:
```suggestion
		extensions["_metadata_" + key] = value
```

ItalyPaleAle:
Please fix the linter issues, then LGTM

pkg/runtime/runtime.go
berndverst: I had already changed that before you left this comment :)

berndverst:
Hmm, actually I don't think this PR will be necessary as these things are sent via gRPC request metadata / HTTP headers. I can implement this differently in the Python SDK

Issue #5748, opened 2023-01-12
Resiliency, timeouts, and gRPC proxying streams


ItalyPaleAle:
While performing some tests, I ran into one potential issue that I think may be worth clarifying.

- `client` invokes `server` using gRPC proxying
- Resiliency policy:  
  
  ```yaml
  apiVersion: dapr.io/v1alpha1
  kind: Resiliency
  metadata:
    name: myresiliency
  scopes:
    - client
  spec:
    policies:
      timeouts:
        five: 5s
    targets:
      apps:
        server:
          timeout: five
  ```
- If the method `client` invokes in `server` is a streaming call (e.g. `rpc Pinger(stream Ping) returns (stream Ping) {}`), after 5s the timeout policy kicks in and forcefully terminates the connection

Is this an intended behavior? Which means that users should not set policies with a timeout when they are using streaming calls

- If so, possible downside: there's no way to set a custom timeout, for example to handle connection errors (there's a timeout of 30s for establishing connections, but that may be too long)

ItalyPaleAle:
CC: @halspang @yaron2

PR #5751, opened 2023-01-12
Multiple fixes to gRPC proxy including panics, data loss, resiliency not working

ItalyPaleAle:
# Description

This PR makes a number of fixes to the gRPC proxy package including fixing one panic during reconnecting, possible data losses, and resiliency policies not working.

Please see inline comments on changes for more precise description of what's changed.

# Issues with proxying of gRPC streamed requests

Although this PR fixes a bunch of issues, during investigation I found that there are a few more issues specific to handling gRPC streamed requests. These are **not** fixed in this PR, and I think they deserve further look (possibly even blockers for Dapr 1.10?):

1. Issues with timeouts set in resiliency policies. See: #5748
2. Issues with buffering: currently all incoming messages in a stream are kept in the ["bufferedFrames" slice](https://github.com/ItalyPaleAle/dapr/blob/09c9bb2d9ad55d5a3a1c565d41278692c3363a43/pkg/grpc/proxy/handler.go#L273-L274). This causes three problems:
   1. With streamed calls the buffer will keep every single message in memory, and it is never reset. For streams that are very long-lived, this can grow to consume A LOT of memory.
   2. When calls are restarted after a failure, the full buffer is replayed. For example, say that the client sent 10 messages, then the server crashed; when the connection is re-established, all 10 messages are re-sent to the client again (replaying the request).
   3. There's no backpressure on the client. For example, assume that a client sends a message every second, and the server crashes. The client's sidecar will continue to receive messages from the client app, and will keep them in the buffer which continues to grow (see first sub-bullet above), and are then sent to the server once it gets back online (see second sub-bullet above).

### DIFF
--- a/pkg/grpc/proxy/director.go
+++ b/pkg/grpc/proxy/director.go
@@ -4,7 +4,8 @@
 package proxy
 
 import (
-	"golang.org/x/net/context"
+	"context"
+
 	"google.golang.org/grpc"
 )
 
--- a/pkg/grpc/proxy/handler.go
+++ b/pkg/grpc/proxy/handler.go
@@ -4,17 +4,16 @@
 package proxy
 
 import (
-	"fmt"
+	"context"
 	"io"
 	"sync"
+	"sync/atomic"
 
-	"golang.org/x/net/context"
+	"github.com/google/uuid"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/metadata"
 	"google.golang.org/grpc/status"
-
-	"github.com/google/uuid"
 
 	"github.com/dapr/dapr/pkg/diagnostics"
 	"github.com/dapr/dapr/pkg/grpc/proxy/codec"
@@ -37,7 +36,7 @@ func RegisterService(server *grpc.Server, director StreamDirector, resiliency re
 	}
 	fakeDesc := &grpc.ServiceDesc{
 		ServiceName: serviceName,
-		HandlerType: (*interface{})(nil),
+		HandlerType: (*any)(nil),
 	}
 	for _, m := range methodNames {
 		streamDesc := grpc.StreamDesc{
@@ -58,11 +57,10 @@ func RegisterService(server *grpc.Server, director StreamDirector, resiliency re
 // This can *only* be used if the `server` also uses grpcproxy.CodecForServer() ServerOption.
 func TransparentHandler(director StreamDirector, resiliency resiliency.Provider, isLocalFn func(string) (bool, error), connFactory DirectorConnectionFactory) grpc.StreamHandler {
 	streamer := &handler{
-		director:      director,
-		resiliency:    resiliency,
-		isLocalFn:     isLocalFn,
-		bufferedCalls: sync.Map{},
-		connFactory:   connFactory,
+		director:    director,
+		resiliency:  resiliency,
+		isLocalFn:   isLocalFn,
+		connFactory: connFactory,
 	}
 	return streamer.handler
 }
@@ -72,27 +70,25 @@ type handler struct {
 	resiliency    resiliency.Provider
 	isLocalFn     func(string) (bool, error)
 	bufferedCalls sync.Map
-	headersSent   sync.Map
 	connFactory   DirectorConnectionFactory
 }
 
 // handler is where the real magic of proxying happens.
 // It is invoked like any gRPC server stream and uses the gRPC server framing to get and receive bytes from the wire,
 // forwarding it to a ClientStream established against the relevant ClientConn.
-func (s *handler) handler(srv interface{}, serverStream grpc.ServerStream) error {
+func (s *handler) handler(srv any, serverStream grpc.ServerStream) error {
 	// Create buffered calls for this request.
 	requestIDObj, err := uuid.NewRandom()
 	if err != nil {
 		return status.Errorf(codes.Internal, "failed to generate UUID: %v", err)
 	}
 	requestID := requestIDObj.String()
-	s.bufferedCalls.Store(requestID, []interface{}{})
-	s.headersSent.Store(requestID, false)
+	s.bufferedCalls.Store(requestID, []any{})
 
 	// little bit of gRPC internals never hurt anyone
 	fullMethodName, ok := grpc.MethodFromServerStream(serverStream)
 	if !ok {
-		return status.Errorf(codes.Internal, "lowLevelServerStream not exists in context")
+		return status.Errorf(codes.Internal, "full method name not found in stream")
 	}
 
 	// Fetch the AppId so we can reference it for resiliency.
@@ -107,8 +103,8 @@ func (s *handler) handler(srv interface{}, serverStream grpc.ServerStream) error
 		policyDef = noOp.EndpointPolicy("", "")
 	} else {
 		isLocal, err := s.isLocalFn(v[0])
-		if err == nil && isLocal {
-			policyDef = s.resiliency.EndpointPolicy(v[0], fmt.Sprintf("%s:%s", v[0], fullMethodName))
+		if err == nil && !isLocal {
+			policyDef = s.resiliency.EndpointPolicy(v[0], v[0]+":"+fullMethodName)
 		} else {
 			noOp := resiliency.NoOp{}
 			policyDef = noOp.EndpointPolicy("", "")
@@ -119,7 +115,12 @@ func (s *handler) handler(srv interface{}, serverStream grpc.ServerStream) error
 	clientStreamOpts := []grpc.CallOption{
 		grpc.CallContentSubtype((&codec.Proxy{}).Name()),
 	}
+	headersSent := &atomic.Bool{}
+	counter := atomic.Int32{}
 	_, cErr := policyRunner(func(ctx context.Context) (struct{}, error) {
+		// Get the current iteration count
+		iter := counter.Add(1)
+
 		// We require that the director's returned context inherits from the ctx.
 		outgoingCtx, backendConn, target, teardown, err := s.director(ctx, fullMethodName)
 		defer teardown(false)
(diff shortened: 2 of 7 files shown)


yaron2:
cc @halspang for bufferedFrames review.

pkg/grpc/proxy/director.go
```diff
@@ -4,7 +4,8 @@
 package proxy
 
 import (
-	"golang.org/x/net/context"
+	"context"
```
ItalyPaleAle, line 7: Reviewers: removed the old, deprecated `context` package and use the standard library (the old package was added as a denied one to the linter)

pkg/grpc/proxy/handler.go
```diff
@@ -119,7 +115,12 @@ func (s *handler) handler(srv interface{}, serverStream grpc.ServerStream) error
 	clientStreamOpts := []grpc.CallOption{
 		grpc.CallContentSubtype((&codec.Proxy{}).Name()),
 	}
+	headersSent := &atomic.Bool{}
```
ItalyPaleAle, line 118: Reviewers: perf improvement here: use a local variable rather than keeping in a sync.Map for better memory usage and perf.

pkg/grpc/proxy/handler.go
```diff
@@ -128,12 +129,26 @@ func (s *handler) handler(srv interface{}, serverStream grpc.ServerStream) error
...
+			fullMethodName,
+			clientStreamOpts...,
+		)
 		if err != nil {
 			code := status.Code(err)
 			if target != nil && (code == codes.Unavailable || code == codes.Unauthenticated) {
+				// It's possible that we get to this point while another goroutine is executing the same policy function.
+				// For example, this could happen if this iteration has timed out and "policyRunner" has triggered a new execution already.
+				// In this case, we should not teardown the connection because it could being used by the next execution. So just return and move on.
+				if counter.Load() != iter {
```
ItalyPaleAle, line 148: Reviewers: in case of a resiliency-induced timeout, there could be a race condition and the timed-out handler could otherwise teardown a connection that's being used by a new handler. This fixes it.

pkg/grpc/proxy/handler.go
```diff
@@ -148,29 +163,30 @@ func (s *handler) handler(srv interface{}, serverStream grpc.ServerStream) error
 				if err != nil {
 					return struct{}{}, err
 				}
+			} else {
+				return struct{}{}, err
```
ItalyPaleAle, line 167: Reviewer: this fixes a panic if `grpc.NewClientStream` returns an error that is not one of those handled above

pkg/grpc/proxy/handler.go
```diff
@@ -230,23 +248,31 @@ func (s *handler) forwardClientToServer(src grpc.ClientStream, dst grpc.ServerSt
 func (s *handler) forwardServerToClient(src grpc.ServerStream, dst grpc.ClientStream, requestID string) chan error {
 	ret := make(chan error, 1)
 	go func() {
-		f := &codec.Frame{}
```
ItalyPaleAle, line 233: Reviewers: this had to be moved within loop otherwise the replay buffer would capture stale data. Example:

```text
== APP == 2023/01/12 09:37:52 Received ping from client: 🦥 1
== APP == 2023/01/12 09:37:53 Received ping from client: 🦥 2
== APP == 2023/01/12 09:37:54 Received ping from client: 🦥 3
== APP == 2023/01/12 09:37:55 Received ping from client: 🦥 4
== APP == 2023/01/12 09:37:56 Received ping from client: 🦥 5
== APP == 2023/01/12 09:37:57 Received ping from client: 🦥 6
== APP == 2023/01/12 09:37:58 Received ping from client: 🦥 7
== APP == 2023/01/12 09:37:59 Received ping from client: 🦥 8
== APP == 2023/01/12 09:38:00 Received ping from client: 🦥 9
== APP == 2023/01/12 09:38:01 Received ping from client: 🦥 10
== APP == 2023/01/12 09:38:02 Received ping from client: 🦥 11
== APP == 2023/01/12 09:38:03 Received ping from client: 🦥 12
== APP == 2023/01/12 09:38:04 Received ping from client: 🦥 13
## ... (Restart server app)
== APP == 2023/01/12 09:38:24 Received ping from client: 🦥 14
== APP == 2023/01/12 09:38:24 Received ping from client: 🦥 14
== APP == 2023/01/12 09:38:24 Received ping from client: 🦥 14
== APP == 2023/01/12 09:38:24 Received ping from client: 🦥 14
== APP == 2023/01/12 09:38:24 Received ping from client: 🦥 14
== APP == 2023/01/12 09:38:24 Received ping from client: 🦥 14
== APP == 2023/01/12 09:38:24 Received ping from client: 🦥 14
== APP == 2023/01/12 09:38:24 Received ping from client: 🦥 14
== APP == 2023/01/12 09:38:24 Received ping from client: 🦥 14
== APP == 2023/01/12 09:38:24 Received ping from client: 🦥 14
== APP == 2023/01/12 09:38:24 Received ping from client: 🦥 14
== APP == 2023/01/12 09:38:24 Received ping from client: 🦥 14
== APP == 2023/01/12 09:38:24 Received ping from client: 🦥 14
== APP == 2023/01/12 09:38:24 Received ping from client: 🦥 14
== APP == 2023/01/12 09:38:24 Received ping from client: 🦥 15
== APP == 2023/01/12 09:38:24 Received ping from client: 🦥 16
== APP == 2023/01/12 09:38:24 Received ping from client: 🦥 17
== APP == 2023/01/12 09:38:24 Received ping from client: 🦥 18
== APP == 2023/01/12 09:38:24 Received ping from client: 🦥 19
== APP == 2023/01/12 09:38:24 Received ping from client: 🦥 20
== APP == 2023/01/12 09:38:24 Received ping from client: 🦥 21
```

As you can see, message "14" is repeated many times due to storing a pointer in the replay buffer

(Also this is an example of how, after a reconnection, the entire buffer is replayed back to the client every time and the target receives 14 messages that were already delivered)

pkg/grpc/proxy/handler.go
```diff
@@ -107,8 +103,8 @@ func (s *handler) handler(srv interface{}, serverStream grpc.ServerStream) error
 		policyDef = noOp.EndpointPolicy("", "")
 	} else {
 		isLocal, err := s.isLocalFn(v[0])
-		if err == nil && isLocal {
-			policyDef = s.resiliency.EndpointPolicy(v[0], fmt.Sprintf("%s:%s", v[0], fullMethodName))
+		if err == nil && !isLocal {
```
ItalyPaleAle, line 106: Reviewers: this should have been a `!isLocal` - otherwise resiliency policies are not applied.

ItalyPaleAle:
One possible solution I'd propose that would fix all the errors with streams:

1. Apply the resiliency (including retries, timeouts, etc) only on the first handshake.
2. Once the connection has been established, do not apply any more resiliency features and do not keep a buffer of messages. If the long-lived stream is dropped, just notify the client/server of the error and have them re-establish the connection.

One possible problem with the approach above is that we will need to find a way to handle unary vs streamed gRPC calls separately (need to investigate how to implement that):

- For unary gRPC calls, arguably we should make resiliency (timeouts, retries, etc) apply to the entire request + response lifecycle, as that's probably what users expect
- For streaming gRPC calls, we can apply the logic above

halspang (commented):
Overall, the changes seem fine to me. The question I have is why didn't the e2e catch this if it wasn't working? I'm guessing we just didn't check the exact cases but can we modify the e2e tests to ensure that this doesn't happen again?

Re: Doing resiliency at the handshake. I tried that originally but wasn't able to get it to work. My research led me to this approach. That was, however, my first time working with gRPC streams so it is possible I just missed the solution to make that approach work. I'm fine doing it that way if you can make it work, just giving context as to why it is the way it is atm.

ItalyPaleAle:
Thanks for the review, Hal.

I'm working on more tests (esp. Unit tests) and then wondering if this PR can be merged.

I will investigate the handshake issue in a separate PR.

> I did dig into the source code of gRPC and it does look like every invocation, internally, is a stream. The gRPC library then has a map of the calls that should be handled as unary ones, and "converts" them to streams internally. This means that without knowing the exact protos, it will be hard to tell if the invocation is a stream or unary. I will investigate to see if there's some other way to handle this problem.

ItalyPaleAle:
@yaron2 @halspang I added a bunch more unit tests to also validate the "un-happy" path (no changes to code since Hal's review).

If you can please merge this PR, I will work on a separate one to address the remaining issues with proxying of streamed calls.

yaron2:
Thanks @ItalyPaleAle I'll review. Can you please validate my favorite scenario, workload certificate rotations?

ItalyPaleAle:
@yaron2 

> Can you please validate my favorite scenario, workload certificate rotations?

Uhm, I'm not familiar with this. Do you have an example please?

yaron2:
> @yaron2
> 
> > Can you please validate my favorite scenario, workload certificate rotations?
> 
> Uhm, I'm not familiar with this. Do you have an example please?

You've verified this before (can't remember the PR this was tested for off the top of my head) - two apps doing service invocation (app A calls B) on Kubernetes, sending a message every 1s, with workload cert rotation set to ~1m to test that when a cert is replaces and the request gets rejected, retries kick in for the caller and recover. The logs should show no errors or interruptions.

ItalyPaleAle:
@yaron2  confirming the test passes:

image

However, worth pointing out that this is using a persistent connection, which is not re-created when the TLS certificate is rotated. So it is continuing to work regularly (every 1s) as expected without interruptions.

Underlying this is the same gRPC connection manager which we validated cert rotation with.

ItalyPaleAle:
Perf tests are failing:
- 3x due to actor ID scale - a new issue?
- 1x for gRPC service invocation (not proxy) - see #5740

In either case, seems unrelated

yaron2 (approved):

yaron2:
Merged so we can test further. It'd be great to add a test with multiple callers (multiple goroutines) and ensure no panics and consistency of data (correct response for the correct calling oepration)

ItalyPaleAle:
@yaron2 thanks.

Will do that in the next PR alongside tests for streamed calls. These will be unit tests (like the ones I've added to this PR) for better control and isolation.

Issue #5690, opened 2022-12-31
Allow component metadata value to use `{appID}`


artursouza:
## In what area(s)?

 /area runtime

> /area operator

> /area placement

> /area docs

> /area test-and-release
## Describe the proposal

Today, component metadata values can use template values like `{uuid}` or `{podName}`. In addition, components can have some attributes that can be bound to a particular application. So, this proposal is to also add `{appID}` as another possible template value.

VerstraeteBert:
Picking this up along with the Kafka component ticket.

PR #5752, opened 2023-01-12
Allow component metadata value to use {appID}

VerstraeteBert:
# Description

Today, component metadata values can use template values like `{uuid}` or `{podName}`. In addition, components can have some attributes that can be bound to a particular application. So, this PR adds `{appID} `as another possible template value. Which is relevant for, for example, https://github.com/dapr/components-contrib/issues/2397

## Issue reference

#5690

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Unit tests passing


artursouza (approved):

berndverst:
Please don't use this approach for the linked Kafka issue.

I don't like additional API calls in components.

Instead, the runtime should inject special purpose keys in metadata properties (see "consumerID" for such a special purpose metadata being handled by the library) ideally prefixed with "_dapr_" or something.

Then each component would only need to access these values from metadata.properties (available in the Init function of the component)

Do not add new API calls in components thanks!

artursouza:
> Please don't use this approach for the linked Kafka issue.
> 
> I don't like additional API calls in components.
> 
> Instead, the runtime should inject special purpose keys in metadata properties (see "consumerID" for such a special purpose metadata being handled by the library) ideally prefixed with `_dapr_` or something.
> 
> Then each component would only need to access these values from metadata.properties (available in the Init function of the component)
> 
> Do not add new API calls in components thanks!

This is beyond Kafka. Kafka is one of the scenarios and there is no extra API call since the info is already there.

berndverst:
Ah sorry, I thought this PR was something else (was reading from my phone)

We should additionally inject the appID into metadata properties but that doesn't have to be here.

artursouza:
@VerstraeteBert this PR will require documentation in dapr/docs, similar to the other properties.

VerstraeteBert:
Hab

> @VerstraeteBert this PR will require documentation in dapr/docs, similar to the other properties.

Updated docs here https://github.com/dapr/docs/pull/3062

PR #5761, opened 2023-01-13
Dapr Workflow Engine Part 1: Internal Actors

cgillum:
# Description

This is part 1 of the workflow engine described in #4576 (workflow building block).

### Context for maintainers

The full workflow engine for v1.10 is already implemented in the [`feature/workflows`](https://github.com/dapr/dapr/tree/feature/workflows) branch, and this PR is the first of two where we integrate those changes into `master`. @johnewart and I thought it would be easier for maintainers to review if we broke the integration into two smaller, independent PRs instead of one big PR.

Note that all the code in this PR was already reviewed before being checked into `feature/workflows` (mostly in https://github.com/dapr/dapr/pull/5301, https://github.com/dapr/dapr/pull/5344, and https://github.com/dapr/dapr/pull/5505). Also, this PR only includes the changes to introduce internal actors. It does not contain *any* workflow engine implementation code. That will come in a subsequent PR.

### Change description

The Dapr Workflow engine described in #4576 (workflow building block) introduces a new "internal actor" concept, where Dapr maintainers can build actors that run directly into the sidecar. The workflow engine uses these internal actors to manage state, scale-out, leader-election, etc. for workflows that run using the built-in engine. Internal actors cannot be implemented by app developers. 

Changes to support internal actors include:

* Created a new `InternalActor` interface that internal actors must implement
* Created a new `AppChannel` in the actor runtime specifically for routing requests to internal actors (named `internalActorChannel`)
* Updated reminder code to support conditionally deleting a reminder after execution (reminder cancellation)
* Created a `PlacementService` interface so that the placement service can be mocked for unit testing actors

No existing functionality should be impacted by these changes.

## Issue reference

This PR is related to the following GitHub issues:

* https://github.com/dapr/dapr/issues/4576
* https://github.com/dapr/dapr/issues/5537

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Unit tests passing


artursouza (dismissed):

pkg/actors/actors.go
```diff
@@ -514,6 +548,14 @@ func (a *actorsRuntime) callLocalActor(ctx context.Context, req *invokev1.Invoke
 	return resp, nil
 }
 
+func (a *actorsRuntime) getAppChannel(actorType string) channel.AppChannel {
+	if a.internalAppChannel.Contains(actorType) {
+		return a.internalAppChannel
+	} else {
+		return a.externalAppChannel
```
mcandeia, line 555: nit: typically this else is omitted following the go code style, so you can return straight

source: https://go.dev/doc/effective_go#if

pkg/actors/internal_actor.go
```diff
@@ -0,0 +1,198 @@
+/*
+Copyright 2022 The Dapr Authors
```
mcandeia, line 2: nit: 2023 in all new files

pkg/actors/actors.go
```diff
@@ -153,6 +165,10 @@ type ActorsOpts struct {
 	TracingSpec      configuration.TracingSpec
 	Resiliency       resiliency.Provider
 	StateStoreName   string
+	InternalActors   map[string]InternalActor
+
+	// MockPlacement is a placement service implementation used for testing
+	MockPlacement PlacementService
```
mcandeia, line 171: I would avoid to name it as a `Mock...` as this should be tied only for testing purposes. Why not just `PlacementService` ?
```suggestion
	PlacementService PlacementService
```

pkg/actors/actors.go
```diff
@@ -216,11 +246,13 @@ func (a *actorsRuntime) Init() error {
...
-		a.config.AppID, hostname, a.config.HostedActorTypes,
-		appHealthFn,
-		afterTableUpdateFn)
+	if a.placement == nil {
+		a.placement = internal.NewActorPlacement(
+			a.config.PlacementAddresses, a.certChain,
+			a.config.AppID, hostname, a.config.HostedActorTypes,
+			appHealthFn,
+			afterTableUpdateFn)
+	}
```
mcandeia, line 255: how about move this check to the `NewActors` function? 

Checking for nil gives the impression that placement can be nil in other places as well, that makes placement unsafe to call without a proper nil check. Overall I think that by unifying nil checks on the constructor would help the reader to understand which fields can be nil or not.

pkg/actors/actors.go
```diff
@@ -85,14 +85,22 @@ type Actors interface {
...
+	WaitUntilPlacementTableIsReady(ctx context.Context) error
+	LookupActor(actorType, actorID string) (host string, appID string)
+}
+
 // GRPCConnectionFn is the type of the function that returns a gRPC connection
 type GRPCConnectionFn func(ctx context.Context, address string, id string, namespace string, customOpts ...grpc.DialOption) (*grpc.ClientConn, func(destroy bool), error)
 
 type actorsRuntime struct {
-	appChannel             channel.AppChannel
+	externalAppChannel     channel.AppChannel
```
mcandeia, line 100: I feel that the names "external" and "internal" don't fit well with what we want to implement for actors. The Actor being internal doesn't mean we have an external app, does it? it seems to me that we need an ActorsChannel (which embeds an AppChannel) maybe? And it is able to decide whether to call an internal actor or an app actor. I don't know if the concept of "external" is the best one to use, but this is just an impression of someone who is reviewing a workflow PR for the first time.

I would avoid to create a new concept of "external app" because of the internal actors - which is kinda what is happening here.

mcandeia (changes requested):

pkg/actors/actors.go
cgillum: I don't think I'm understanding the nuance of your feedback, but I'll try to respond. "External" in this context means external to the sidecar (same with "Internal"). The reason to rename `appChannel` to `externalAppChannel` is simply to distinguish it from the newly added `internalAppChannel`. If "external" is confusing, I can certainly remove it, or perhaps name it `defaultAppChannel`?

pkg/actors/actors.go
cgillum: I named it this way because it's only intended to be used for unit testing. I didn't want to give the impression that this is the expected initialization path in production.

cgillum:
Thanks @mcandeia for the initial feedback. Take a look at my responses and my code updates and let me know if you have any other questions, concerns, or things you want me to follow up on.

pkg/actors/actors.go
mcandeia: Yes, I noticed that. My point is: It's being used on tests now but can be used for a different purpose later. That's why I think you should name as it is - which is good actually, meaning that we can provide a custom placement service when creating actors. The usage for tests is only one of possible, this is a way to not "change your code to be testable".

However this is a nit. You can keep it if you want.

pkg/actors/actors.go
mcandeia: fair enough

mcandeia (approved):

pkg/actors/actors.go
mcandeia: > Or, perhaps what I can do is rename "internalAppChannel" to "internalActorChannel" and revert "externalAppChannel" back to "appChannel"

That sounds better. Just to clarify: My concern is around the term"external/internal" for apps. There's no mention to an "external" app or "internal" app in our documentation, right? The usage of Internal actors doesn't mean that we have an internal app, does it ? You don't need to tie these concepts, we can have an internal actor without creating a concept of "internal app".

artursouza (approved):

pkg/actors/actors.go
cgillum: Correct, none of this is in our documentation because these are not concepts that are exposed to users. Rather, this is something that’s only relevant to Dapr maintainers.

We will mention that there is a thing called “internal actors” in the workflow operational docs, but only as a way to help explain the dependencies that are involved when running workflows.

I went ahead and made these naming changes in the previous commit. I appreciate the feedback and think that this will help reduce conceptual overhead for maintainers.

pkg/actors/actors.go
cgillum: My thinking is that we should rename it only when we’re ready to support this abstraction beyond testing. I worry that it would be more harmful to rename it now in anticipation of future use cases when we haven’t yet done the work to support those use cases.

cgillum:
Hi @artursouza @yaron2, what are the next steps for getting the PR merged? I see that the `darwin_amd64` build failed due to some test infra issue:

```
Warning: Aborting upload for /Users/runner/work/dapr/dapr/dist/archives/daprd_darwin_amd64.tar.gz due to failure
Error: aborting artifact upload
Total size of all the files uploaded is 62[645](https://github.com/dapr/dapr/actions/runs/3961401137/jobs/6786781074#step:18:646)920 bytes
File upload process has finished. Finalizing the artifact upload
Upload finished. There were 1 items that failed to upload

The raw size of all the files that were specified for upload is 90769695 bytes
The size of all the files that were uploaded is 62645920 bytes. This takes into account any gzip compression used to reduce the upload size, time and storage

Note: The size of downloaded zips can differ significantly from the reported size. For more information see: https://github.com/actions/upload-artifact#zipped-artifact-downloads 

Error: An error was encountered when uploading dapr_darwin_amd64. There were 1 items that failed to upload.
```

berndverst:
> Hi @artursouza @yaron2, what are the next steps for getting the PR merged? I see that the `darwin_amd64` build failed due to some test infra issue:
> 
> 
> 
> ```
> 
> Warning: Aborting upload for /Users/runner/work/dapr/dapr/dist/archives/daprd_darwin_amd64.tar.gz due to failure
> 
> Error: aborting artifact upload
> 
> Total size of all the files uploaded is 62[645](https://github.com/dapr/dapr/actions/runs/3961401137/jobs/6786781074#step:18:646)920 bytes
> 
> File upload process has finished. Finalizing the artifact upload
> 
> Upload finished. There were 1 items that failed to upload
> 
> 
> 
> The raw size of all the files that were specified for upload is 90769695 bytes
> 
> The size of all the files that were uploaded is 62645920 bytes. This takes into account any gzip compression used to reduce the upload size, time and storage
> 
> 
> 
> Note: The size of downloaded zips can differ significantly from the reported size. For more information see: https://github.com/actions/upload-artifact#zipped-artifact-downloads 
> 
> 
> 
> Error: An error was encountered when uploading dapr_darwin_amd64. There were 1 items that failed to upload.
> 
> ```

Restarted the failed job. We should not merge while any of those basic checks fail.

cgillum:
Thanks @berndverst, but unfortunately it looks like the re-run failed for an entirely different reason (flakey actor tests - these have bitten me before). Since these tests passed in the previous run and no changes were made, would it be okay to re-run this specific job again to see if we can get it green?

berndverst:
@cgillum I'll let one of the maintainers for this repo handle this. I'll trigger just this once but won't monitor.

Issue #4724, opened 2022-06-08
Proposal: Remove Dashboard from Dapr's Helm Chart


artursouza:
Dashboard does not seem to be widely used (feeling, no data yet - happy to be proven wrong here). It is also in a different maturity level compared to the rest of the project (still pre 1.0). I propose the following:

+ Remove Dashboard from Dapr's Helm Chart
+ Create a new subcommand in CLI: `dapr dashboard init` that also supports picking the dashboard version to be used.
+ Release Helm chart for old versions of the Dashboard to enable the command above.

berndverst:
I support this proposal.

@dapr/maintainers-dapr @dapr/approvers-dapr thoughts on this?

NickLarsenNZ:
> Remove Dashboard from Dapr's Helm Chart

Could it just be disabled by default in the chart?

_I haven't used it yet, but was going to take a look after resolving the issues I'm currently facing with the operator._

artursouza:
Yes. It can be disabled by default first, but the scope for this issue is to remove it from the dapr's chart and have its own chart with separate installation. Even the CLI could have `dapr dashboard init -k` to install the dashboard on K8s and `dapr dashboard init` to install in standalone mode.

PR #5767, opened 2023-01-16
WIP: Security Overhaul: Seperate out ServiceAccounts

JoshVanL:
Create separate ServiceAccount for each dapr controlplane component.
Break out and minimize the permissions for each component.


ItalyPaleAle (dismissed):

ItalyPaleAle:
Looks like the tests are failing with Dapr not being deployed. Can you please take a look?

JoshVanL:
@ItalyPaleAle I think the tests are ready to be run again. I'm currently struggling to run them locally on an arm machine.

artursouza:
Is this a breaking change when upgrading Dapr?

Just a reminder that we want to remove Dashboard from the main chart and create a separate one: #4724

JoshVanL:
> Is this a breaking change when upgrading Dapr?

It shouldn't be as this is at the kube API level- when upgrading, all but the operator deployments will be re-created to use their new own ServiceAccount. Each component _should_ still have enough permissions to function.

ItalyPaleAle:
@JoshVanL please add your name to the bot so you can run tests on AKS too. See example: https://github.com/dapr/dapr/pull/5770

JoshVanL:
Thanks @ItalyPaleAle, done https://github.com/dapr/dapr/pull/5774

ItalyPaleAle (approved):

ItalyPaleAle:
@JoshVanL please don't squash or rebase commits on a PR after it's being reviewed. It causes existing reviews to be dismissed (even if the commit is just to update the branch with "master"), and it breaks features like the ability to view changes since the last review. This makes our lives much harder as reviewers.
We normally squash PRs when we merge them anyways.

ItalyPaleAle (approved):

yaron2 (approved):

mukundansundar:
@JoshVanL @yaron2 Since ClusterRoleDefs are changing ... This needs a change in the dapr CLI upgrade Path tests as well ....

pravinpushkar:
> @JoshVanL @yaron2 Since ClusterRoleDefs are changing ... This needs a change in the dapr CLI upgrade Path tests as well ....

this might need a change to below files - 
https://github.com/dapr/cli/blob/master/tests/e2e/kubernetes/kubernetes_test.go
https://github.com/dapr/cli/blob/master/tests/e2e/upgrade/upgrade_test.go

PR #5776, opened 2023-01-17
[DO-NOT-MERGE] Correct metric names

yaron2:
Fix metric names in development doc. Thanks @alicejgibbons for spotting.


ItalyPaleAle (approved):

yaron2:
Holding off on merging, as although the metrics names outlined here do have `_bucket` added when querying the prom endpoint, the actual metric name used for other purposes does not have `_bucket` appended. I'm going to look into handling these special cases correctly in code.

mukundansundar:
@yaron2 was this planned for 1.10?

PR #5732, opened 2023-01-10
Make Resiliency stable

ItalyPaleAle:
# Description

Remove the "Resiliency" feature flag and all the code paths where we biforcated based on whether Resiliency was enabled or not

## Issue reference

Fixes #5440

Docs issue: dapr/docs#2873

PR #5777, opened 2023-01-18
gRPC proxy: fix streaming and resiliency

ItalyPaleAle:
# Description

This PR completes the work started in #5751 addressing a variety of issues with gRPC proxying, especially when proxying streamed RPCs and/or when Resiliency is enabled.

The most relevant changes include:

- Ensure that when resiliency is enabled, when proxying streamed RPCs, the policy (e.g. retries and/or timeouts) now applies to the handshake step only (see #5748 for a description of the issue)
- Reworked the replay buffer so at most one message is kept in the buffer and only if there's a resiliency policy with retries enabled. This fixes an issue discovered in #5748 where the replay buffer would grow indefinitely when proxying streamed requests, potentially using a lot of memory.
- Changed how proxying of streamed RPCs works, so they are not influenced by timeouts and/or retries in the resiliency policy. In case the streamed connection is dropped, the error is now immediately sent to the two apps so they can re-establish the connection. In this case, messages are not replied (by design).
- Added a lot more unit tests, including those validating resiliency and/or proxying of streamed RPCs

**There's one important user-facing change:** When proxying streamed RPCs, users now have to set an additional metadata property in the headers: `dapr-stream` set to a "truthy" value (`true`, `1`, `yes`, etc).

This was required, also as discussed on Discord, as we would have no other way to determine if a gRPC call is a stream or not without knowing the protos (which is not an option in this case).

PS: While working on this, I found that the `resiliency` package had a method that was not used anymore, and that caused some sub-optimal performance (that code was used in a very hot path). That's the primary reason for the changes in the resiliency package.

## Issue reference

Fixes: #5748
Fixes the remaining issues highlighted in: #5751
Relevant for: #5732 

Docs issue: dapr/docs#3065


pkg/grpc/proxy/handler_test.go
```diff
@@ -72,9 +82,28 @@ func (s *assertingService) PingEmpty(ctx context.Context, _ *pb.Empty) (*pb.Ping
...
+		if count%14 == 0 {
+			time.Sleep(800 * time.Millisecond)
+		} else if count%12 == 0 {
+			return nil, status.Errorf(codes.Internal, "Simulated random failure")
+		}
+	}
+
 	// Send user trailers and headers.
-	grpc.SendHeader(ctx, metadata.Pairs(serverHeaderMdKey, "I like turtles."))
-	grpc.SetTrailer(ctx, metadata.Pairs(serverTrailerMdKey, "I like ending turtles."))
```
ItalyPaleAle, line 77: Changed the message because "ending turtles" sounds really bad

yaron2:
Closed by mistake

yaron2:
Just to be clear: this has no impact on current proxying workloads that use streams and don't use resiliency, correct? users don't have to do anything except for setting the stream header *if* they are using resiliency and want to apply it.

pkg/grpc/proxy/handler_test.go
yaron2: The animal lover in me appreciates that.

ItalyPaleAle:
@yaron2 

> Just to be clear: this has no impact on current proxying workloads that use streams and don't use resiliency, correct? users don't have to do anything except for setting the stream header if they are using resiliency with gRPC streams and want to apply it.

Yes, that should be the case.

If Resiliency is enabled, or to be precise, there’s a resiliency policy **with retries**, then setting the “dapr-stream: true” metadata is required. Otherwise, users actually get a runtime error. There’s a new unit test to validate that specifically.

If there’s no resiliency policy with retries, I believe it should work even without the new metadata option because the replay buffer isn’t initialized anyways. I will run some tests tomorrow to confirm that. However, I would still strongly recommend users to set the “dapr-stream” metadata if they’re proxying a streamed RPC.

ItalyPaleAle:
@yaron2 just ran tests confirming that as long as there's no resiliency policy with retries, setting "dapr-stream=true" is not required for this to work, thus maintaining backwards-compatibility (mostly - users will notice that in case of errors, the entire stream will not be replayed anymore - but that behavior should have been considered a bug in the first place).

I also updated one of the unit tests to confirm the behavior above.

In any case, I think that even though "dapr-stream" is not strictly required, the docs should still say that when proxying streamed requests users should put "dapr-strem=true" regardless. That is: I don't think we should document this situation, when it's not strictly required, as an officially blessed situation, and we should encourage users to do the "safe" thing.

pkg/messaging/grpc_proxy.go
```diff
@@ -72,7 +72,15 @@ func NewProxy(opts ProxyOpts) Proxy {
 
 // Handler returns a Stream Handler for handling requests that arrive for services that are not recognized by the server.
 func (p *proxy) Handler() grpc.StreamHandler {
-	return grpcProxy.TransparentHandler(p.intercept, p.resiliency, p.IsLocal,
+	return grpcProxy.TransparentHandler(p.intercept,
+		func(appID, methodName string) *resiliency.PolicyDefinition {
+			_, isLocal, err := p.isLocal(appID)
+			if err == nil && !isLocal {
```
yaron2, line 78: Won't this stop any policies from being executed on a callee app when Dapr invokes the app?

ItalyPaleAle: My understanding is that policies are to be applied on the caller side only. I changed this because otherwise I wasn't able to get resiliency policies to be applied. But let me know if I'm wrong

yaron2 (approved):

PR #5783, opened 2023-01-20
Allow pubsub cloudevent override with special metadata

### LINKED ISSUE #5515: [PubSub] override cloud event attributes via metadata
> /area runtime

## Describe the feature
At present, there are two ways to publish a Cloud Event via PubSub

1. Supply a payload, which is then automatically wrapped by Dapr in a Cloud Event envelope.
2. Developer supplies the full [Cloud Event envelope](https://docs.dapr.io/developing-applications/building-blocks/pubsub/pubsub-cloudevents/#publish-your-own-cloudevent) as-is.

I propose a 3rd way, a developer can use option 1, however they can override any of the standard Cloud Event properties that dapr automatically sets, whilst retaining any other CE attributes automatically provided by Dapr.

I propose this is achieved by using a `cloudevent` reserved keyword, or similar, on `metadata`.

This then enables the convenience of option 1, without having to discover, decide, learn and use a 3rd party Cloud Event library, but also gives a super easy way to tailor the standard Cloud Event fields.

This proposal came off the back of an internal conversation in my org about how we need to control certain Cloud Event properties. It became clear to us that we need to adjust some of the standard CE attributes, but welcome the convenience of the Dapr supplied CE envelope.

```c#
var client = new DaprClientBuilder().Build();

var eventData = new { Id = "17", Amount = 10m, };

var overrides = new Dictionary<string,string>() {
    { "metadata.cloudevent.source", "my.custom.source" },
    { "metadata.cloudevent.id", "my-custom-id" }
}

await client.PublishEventAsync(pubsubName, "deposit", eventData, overrides, cancellationToken);
Console.WriteLine("Published deposit event!");
```

berndverst:

# Description

Allows overriding certain cloud event property via metadata / headers in pubsub publishing. The `cloudevent-` prefix must be used for the metadata key. For example `cloudevent-source` would manually override the `source` value.

Fixes #5515.


pkg/runtime/pubsub/cloudevents.go
```diff
@@ -14,28 +14,36 @@ limitations under the License.
...
+	SpecVersion     string `mapstructure:"specversion"`
+	Time            string `mapstructure:"time"`
 }
 
 // NewCloudEvent encapsulates the creation of a Dapr cloudevent from an existing cloudevent or a raw payload.
 func NewCloudEvent(req *CloudEvent) (map[string]interface{}, error) {
 	if contribContenttype.IsCloudEventContentType(req.DataContentType) {
 		return contribPubsub.FromCloudEvent(req.Data, req.Topic, req.Pubsub, req.TraceID, req.TraceState)
 	}
-	return contribPubsub.NewCloudEventsEnvelope(uuid.New().String(), req.ID, contribPubsub.DefaultCloudEventType,
```
berndverst, line 39: FYI the defaults removed here are already handled within the function itself if empty strings are passed, which is exactly what we are doing here assuming the given properties were not overridden.

berndverst:
PR looks good without this latest E2E test change. Let's see if this more sophisticated test also works...

pkg/grpc/api.go
```diff
@@ -347,15 +348,20 @@ func (a *api) PublishEvent(ctx context.Context, in *runtimev1pb.PublishEventRequ
...
 			DataContentType: in.DataContentType,
 			Data:            body,
 			TraceID:         corID,
 			TraceState:      traceState,
 			Pubsub:          in.PubsubName,
-		})
+		}
+
+		// metadata beginning with "cloudevent-" are considered overrides to the cloudevent envelope
+		mapstructure.WeakDecode(in.Metadata, &cloudevent) // allows ignoring of case
```
ItalyPaleAle, line 362: What if someone were to set properties like `data` or `pubsub` or `topic` in the metadata? Wouldn't that overwrite other properties we care about?

pkg/runtime/pubsub/cloudevents.go
```diff
@@ -14,28 +14,34 @@ limitations under the License.
...
-	ID              string
-	Data            []byte
-	Topic           string
-	Pubsub          string
-	DataContentType string
-	TraceID         string
-	TraceState      string
+	ID              string `mapstructure:"cloudevent-id"`
+	Data            []byte `mapstructure:"cloudevent-data"`
+	Topic           string `mapstructure:"cloudevent-topic"`
```
ItalyPaleAle, line 26: Not sure if we want people to override properties like data, topic, etc? Sounds dangerous as people could shoot themselves in the foot real bad (and then we'd have to support a variety of edge cases)

tests/apps/pubsub-bulk-subscriber/app.go
```diff
@@ -115,6 +116,7 @@ type AppBulkMessageEntry struct {
 type BulkSubscribeResponseEntry struct {
 	EntryId string `json:"entryId"` //nolint:stylecheck
 	Status  string `json:"status"`
+	Source  string `json:"strouce"`
```
ItalyPaleAle, line 119: Typo

ItalyPaleAle (changes requested):
See comments

pkg/grpc/api.go
berndverst: They would have to do so as "cloudevent-data" or "cloudevent-topic" in metadata. The topic is actually stored in another place and additionally in the cloud event. This topic that can be overridden here is not used for routing.

This is by design.

pkg/runtime/pubsub/cloudevents.go
berndverst: Overriding data is weird, but it would have to be done explicitly. There is no reason to prevent it.

Topic is fine for overriding - this topic here isn't the topic used for routing actually.

tests/apps/pubsub-bulk-subscriber/app.go
berndverst: Thanks!

ItalyPaleAle:
As discussed on discord, my recommendation would be to keep an allowlist of options that can be overridden, excluding things like topic and data. If users were to ask for that in the future we can add those, but otherwise it seems risky.

berndverst:
> As discussed on discord, my recommendation would be to keep an allowlist of options that can be overridden, excluding things like topic and data. If users were to ask for that in the future we can add those, but otherwise it seems risky.

There is no real risk here.

A user must explicitly specify metadata with hand "cloudevent-topic" or "cloudevent-data" to override those parts of the envelope.

Keep in mind that this would occur on the same request on which the user is literally specifying the data and the topic.
Sure it doesn't make a ton of sense, but there is no risk.

The request was to be able to override any of the cloudevent properties with metadata. That's what this PR does with the exception of time.

berndverst:
Reviewers - this PR works but I'm still struggling to figure out the E2E tests. There is a little bit more work required.

Maybe someone with knowledge of the PubSub E2E tests can see what I'm doing wrong there.

tests/apps/pubsub-bulk-subscriber/app.go
ItalyPaleAle: Still looks not fixed to me, maybe this is a reason for the test failures?

berndverst:
@ItalyPaleAle per your request Topic, Pubsub name, Data and Data content type cannot be overriden.

berndverst:
I restructured the code such that everything can now be tested in unit tests. I no longer need to modify E2E tests, so I removed those changes.

berndverst:
Removed the E2E test comments as the E2E test changes no longer apply.

pkg/runtime/pubsub/cloudevents.go
```diff
@@ -14,28 +14,40 @@ limitations under the License.
...
 	if contribContenttype.IsCloudEventContentType(req.DataContentType) {
 		return contribPubsub.FromCloudEvent(req.Data, req.Topic, req.Pubsub, req.TraceID, req.TraceState)
 	}
-	return contribPubsub.NewCloudEventsEnvelope(uuid.New().String(), req.ID, contribPubsub.DefaultCloudEventType,
+
+	// metadata beginning with "cloudevent-" are considered overrides to the cloudevent envelope
+	mapstructure.WeakDecode(metadata, &req) // allows ignoring of case
+
+	// ensures the trace ID data is overwritten correctly if desired
+	if req.TraceID == "" && req.TraceParent != "" {
```
ItalyPaleAle, line 48: Is it correct to set the TraceID to the TraceParent? Aren't those supposed to be 2 separate traces? Not an expert of tracing however (@johnewart  may know more)

pkg/runtime/pubsub/cloudevents.go
```diff
@@ -14,28 +14,40 @@ limitations under the License.
...
 // NewCloudEvent encapsulates the creation of a Dapr cloudevent from an existing cloudevent or a raw payload.
-func NewCloudEvent(req *CloudEvent) (map[string]interface{}, error) {
+func NewCloudEvent(req *CloudEvent, metadata map[string]string) (map[string]interface{}, error) {
 	if contribContenttype.IsCloudEventContentType(req.DataContentType) {
 		return contribPubsub.FromCloudEvent(req.Data, req.Topic, req.Pubsub, req.TraceID, req.TraceState)
 	}
-	return contribPubsub.NewCloudEventsEnvelope(uuid.New().String(), req.ID, contribPubsub.DefaultCloudEventType,
+
+	// metadata beginning with "cloudevent-" are considered overrides to the cloudevent envelope
+	mapstructure.WeakDecode(metadata, &req) // allows ignoring of case
```
ItalyPaleAle, line 45: Nit:

```suggestion
	_ = mapstructure.WeakDecode(metadata, &req) // allows ignoring of case
```

This way it's clear that you're ignoring the error on purpose

ItalyPaleAle (approved):

pkg/runtime/pubsub/cloudevents.go
berndverst: `TraceID` is a deprecated field and the current name is `TraceParent` however Dapr still uses `TraceID`. This allows someone to specify the value for `TraceID` with an override as either `cloudevent-traceid` or `cloudevent-traceparent`. They are the same thing. Or cloud event envelope that is generate does not include both of those fields if you follow the code.

pkg/runtime/pubsub/cloudevents.go
berndverst: Good call - let me see if I want to handle that error here.

ItalyPaleAle (approved):

artursouza:
Using "-" as separator is not intuitive. We should use "." instead. That is most common when dealing with nested properties in JSON or tree-like structures.

berndverst:
> Using "-" as separator is not intuitive. We should use "." instead. That is most common when dealing with nested properties in JSON or tree-like structures.

On the contrary, I don't find a period intuitive. Don't think of this as indexing into the event or JSON - that's not how this works at all.

Besides, in the original issue nobody suggested a period to be used for the prefix. The issue has been open for discussion for months and the PR has been sitting there for a week.

I personally don't feel this needs changing. The period would also give it the feeling you could alter every property (or add new ones) when you cannot and should not actually.

artursouza:
> > Using "-" as separator is not intuitive. We should use "." instead. That is most common when dealing with nested properties in JSON or tree-like structures.
> 
> On the contrary, I don't find a period intuitive. Don't think of this as indexing into the event or JSON - that's not how this works at all.
> 
> Besides, in the original issue nobody suggested a period to be used for the prefix. The issue has been open for discussion for months and the PR has been sitting there for a week.
> 
> I personally don't feel this needs changing. The period would also give it the feeling you could alter every property (or add new ones) when you cannot and should not actually.

The original issue starts with ".", using "-" was your idea.

PR #5784, opened 2023-01-20
Dapr Workflow Engine

cgillum:
# Description

This is the second of two PRs for the alpha version of Dapr Workflow, which we intend to ship as part of Dapr v1.10. The first PR was https://github.com/dapr/dapr/pull/5761, which introduced the concept of internal actors. This PR completes the Dapr Workflow feature work by adding the Dapr Workflow engine to daprd.

> **NOTE**: Most of the code in this PR was already reviewed before being checked into `feature/workflows`. However, it could use a second round of review now that it has been fully integrated with the latest changes from `master`. Also, there are a few new changes being introduced by @johnewart.

There are several important changes in this PR:

* A new dependency has been added to the [durabletask-go](https://github.com/microsoft/durabletask-go) library, which is the core of the Dapr workflow engine.
* A new package is introduced in the PR under `/pkg/runtime/wfengine` and implements a custom Dapr backend for the `durabletask-go` runtime using actors (the internal actors introduced in #5761).
* The `durabletask-go` library registers a new gRPC service on the existing Dapr gRPC server
* TODO: A default component will be auto-registered which routes workflow management API calls to this engine.

Documentation work is ongoing. For the time being, you can find many of the design details of this engine here: https://github.com/hhunter-ms/docs/blob/issue_2798b/daprdocs/content/en/operations/components/workflow-engine/workflow-engine.md

## Issue reference

This PR is related to the following GitHub issues:

* https://github.com/dapr/dapr/issues/4576
* https://github.com/dapr/dapr/issues/5537

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Unit tests passing

FYI @johnewart @artursouza @mcandeia


pkg/runtime/wfengine/README.md
```diff
@@ -0,0 +1,175 @@
...
+
+```bash
+DEBUG=1 make build
+```
+* `DEBUG=1` is required to attach debuggers. This should never be set for production or performance testing workloads.
+
+After building, the following bash command can be run from the project root to test the code:
+
+```bash
+./dist/linux_amd64/debug/daprd --app-id foo --dapr-grpc-port 4001 --placement-host-address :50005 --components-path ~/.dapr/components/ --config ~/.dapr/config.yaml
```
RyanLettieri, line 29: The "components-path" is completely optional here, correct? If so, do we want to keep it here? I ask because it may cause some confusion since all other engines, i.e. temporal need to have the component path specified whereas the built-in does not. It may lead to the users thinking that they need to have a yaml created for the built-in.

cgillum: Good points, but this is a slightly special case because we're running `daprd` directly from the enlistment root (something users won't normally do) rather than using the `dapr` CLI. I found that I did need to explicitly specify `--components-path`, not specifically because we needed it for a workflow component, but rather because we needed it for the _state store_ component (e.g. redis).

In the user-facing documentation, we won't explicitly include `--components-path`. We'll instead assume that they're just using the normal dapr initialization path, which includes a default redis state store setup. We can also mention explicitly that there is no component yaml required for the Dapr workflow engine.

pkg/runtime/wfengine/activity.go
```diff
@@ -0,0 +1,298 @@
...
+	taskEvent, err := backend.UnmarshalHistoryEvent(eventPayload)
+	if err != nil {
+		return err
+	}
+
+	endIndex := strings.LastIndex(actorID, "#")
+	if endIndex < 0 {
+		return fmt.Errorf("invalid activity actor ID: %s", actorID)
+	}
+	workflowID := actorID[0:endIndex]
```
mcandeia, line 149: qq: Do you think worth to have an ActorID type?

`type ActorID = string` 

so you can have custom operations like


```golang
func (id ActorID) WorkflowID() (string, error) {
...
}
```

mcandeia: plus: A custom constructor that can construct/deconstruct using the desired format

pkg/runtime/wfengine/activity.go
```diff
@@ -0,0 +1,298 @@
...
+		}
+		return newRecoverableError(fmt.Errorf("failed to schedule an activity execution: %w", err))
+	}
+
+loop:
+	for {
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case <-time.After(10 * time.Minute):
```
mcandeia, line 178: shouldn't this timeout be configurable?

pkg/runtime/wfengine/activity.go
```diff
@@ -0,0 +1,298 @@
...
+				wfLogger.Warnf("%s: '%s' is still running - will keep waiting until %v", actorID, name, deadline)
+			} else {
+				wfLogger.Warnf("%s: '%s' is still running - will keep waiting indefinitely", actorID, name)
+			}
+		case completed := <-callback:
+			if completed {
+				break loop
+			} else {
+				return newRecoverableError(errExecutionAborted)
+			}
```
mcandeia, line 189: nit: in case of the callback channel being closed this will return `executionAborted`, is this expected?
```suggestion
		case completed, ok := <-callback:
			if completed || !ok {
				break loop
			} else {
				return newRecoverableError(errExecutionAborted)
			}
```

pkg/runtime/wfengine/activity.go
```diff
@@ -0,0 +1,298 @@
...
+	if err != nil {
+		return newRecoverableError(fmt.Errorf("failed to invoke '%s' method on workflow actor: %w", AddWorkflowEventMethod, err))
+	}
+	defer resp.Close()
+	return nil
+}
+
+// InvokeTimer implements actors.InternalActor
+func (*activityActor) InvokeTimer(ctx context.Context, actorID string, timerName string, params []byte) error {
+	return nil // timers are not used
```
mcandeia, line 215: should we return an error instead? is it ok to give the "impression" that it actually works?

pkg/runtime/wfengine/backend.go
```diff
@@ -0,0 +1,267 @@
...
+	case <-ctx.Done():
+		return nil, ctx.Err()
+	case wi := <-be.orchestrationWorkItemChan:
+		return wi, nil
+	}
+}
+
+// Start implements backend.Backend
+func (be *actorBackend) Start(ctx context.Context) error {
+	if be.started {
```
mcandeia, line 242: If you expect to have this called more than once, how about using https://pkg.go.dev/sync#Once.Do to avoid concurrency as well

pkg/runtime/wfengine/backend.go
```diff
@@ -0,0 +1,267 @@
...
+	return nil, errors.New("not supported")
+}
+
+// GetOrchestrationWorkItem implements backend.Backend
+func (be *actorBackend) GetOrchestrationWorkItem(ctx context.Context) (*backend.OrchestrationWorkItem, error) {
+	// Wait for the workflow actor to signal us with some work to do
+	select {
+	case <-ctx.Done():
+		return nil, ctx.Err()
+	case wi := <-be.orchestrationWorkItemChan:
```
mcandeia, line 235: what if the channel got closed, is it ok to process a zero-value orchestrationworkitem?

pkg/runtime/wfengine/backend.go
```diff
@@ -0,0 +1,267 @@
...
+}
+
+// GetActivityWorkItem implements backend.Backend
+func (be *actorBackend) GetActivityWorkItem(ctx context.Context) (*backend.ActivityWorkItem, error) {
+	// Wait for the workflow actor to signal us with some work to do
+	select {
+	case <-ctx.Done():
+		return nil, ctx.Err()
+	case wi := <-be.activityWorkItemChan:
+		return wi, nil
```
mcandeia, line 220: what if the channel got closed, is it ok to process a zero-value activity work item?

pkg/runtime/runtime.go
```diff
@@ -583,6 +587,22 @@ func (a *DaprRuntime) appHealthReadyInit(opts *runtimeOpts) {
...
+				if a.workflowComponentRegistry != nil {
+					log.Infof("Registering component for dapr workflow engine...")
+					a.workflowComponentRegistry.RegisterComponent(wfComponentFactory, "dapr")
+					if componentInitErr := a.initWorkflowComponent(wfengine.ComponentDefinition); componentInitErr != nil {
+						log.Warnf("Failed to initialize Dapr workflow component: %v", componentInitErr)
+					}
+				} else {
+					log.Infof("No workflow registry available, not registering Dapr workflow component...")
+				}
+			}
```
mcandeia, line 605: can we move this logic to something like `initWfEngine()` method ? I guess that it would improve readability

pkg/runtime/wfengine/activity.go
```diff
@@ -0,0 +1,298 @@
...
+// DeactivateActor implements actors.InternalActor
+func (a *activityActor) DeactivateActor(ctx context.Context, actorID string) error {
+	wfLogger.Debugf("deactivating activity actor '%s'", actorID)
+	a.statesCache.Delete(actorID)
+	return nil
+}
+
+func (a *activityActor) loadActivityState(ctx context.Context, actorID string, generation uint64) (activityState, error) {
+	// See if the state for this actor is already cached in memory.
+	result, ok := a.statesCache.Load(actorID)
```
mcandeia, line 227: I'm assuming that due to the actor's nature of ensuring activation isolation across actors of the same ID you don't need to handle concurrent requests here - e.g using a RW mutex, is that correct?

pkg/runtime/wfengine/backend.go
```diff
@@ -0,0 +1,267 @@
...
+		}
+		if len(data) == 0 {
+			return nil, api.ErrInstanceNotFound
+		}
+		var metadata api.OrchestrationMetadata
+		if err := actors.DecodeInternalActorData(data, &metadata); err != nil {
+			return nil, fmt.Errorf("failed to decode the internal actor response: %w", err)
+		}
+		return &metadata, nil
+	}
```
mcandeia, line 139: nit: typically this else is omitted following the go code style

pkg/runtime/wfengine/component.go
```diff
@@ -0,0 +1,177 @@
...
+		} else if metadata.IsComplete() {
+			res.Metadata["dapr.workflow.output"] = metadata.SerializedOutput
+		}
+
+		return res, nil
+	}
+}
+
+// Status values are defined at: https://github.com/microsoft/durabletask-go/blob/119b361079c45e368f83b223888d56a436ac59b9/internal/protos/orchestrator_service.pb.go#L42-L64
+func getStatusString(status int32) string {
```
mcandeia, line 156: pure nit: use maps instead

```golang
var statusMap = map[int32]string {...}

if statusStr, ok := statusMap[status]; ok {
    return statusStr
}
return "UNKNOWN"
```

mcandeia (changes requested):
most of the comments are nitpicking,
needs to finish wfengine.go, workflowstate.go and tests.

awesome work.

pkg/runtime/wfengine/activity.go
```diff
@@ -0,0 +1,298 @@
...
+
+			// Returning nil signals that we want the execution to be retried in the next period interval
+			return nil
+		} else {
+			wfLogger.Errorf("%s: execution failed with a non-recoverable error: %v", actorID, err)
+			// TODO: Reply with a failure - this requires support from durabletask-go to produce TaskFailure results
+		}
+	}
+
+	// TODO: Purge actor state based on some data retention policy
```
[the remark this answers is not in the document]
cgillum: We have this tracked in our project board: https://github.com/orgs/dapr/projects/61 ("Workflow engine data governance")

pkg/runtime/wfengine/activity.go
cgillum: I think it's worth having an `ActorID` type if we want to enforce specific constraints or operations on those IDs. I don't really consider myself an expert on actors, though, so I might leave something like that up to an Actors SME. It may be best to consider this outside the scope of this PR, however.

pkg/runtime/wfengine/activity.go
cgillum: It's kind of configurable right now - you can see that we specify the timeout value in the `activityActor` struct, which influences the actual deadline for this `ctx` object. However, I don't think we expose it anywhere yet. Timeout handling in general is still a bit TODO at this point.

One thing I'll point out about this specific line of code, however, is that this 10-minute period is just for periodic logging - not for the actual timeout.

pkg/runtime/wfengine/activity.go
cgillum: I think so. If the channel is closed I assume that means something is shutting down, in which case aborting the execution is the right thing to do (it will be retried later). We definitely don't want to continue execution because then the workflow would receive an undefined response from the activity it scheduled. Activities must always be allowed to run until completion or until an explicit failure.

pkg/runtime/wfengine/activity.go
cgillum: Yeah, I can change this to an error.

pkg/runtime/wfengine/activity.go
cgillum: While it's true that we get the guarantees that any specific actor ID doesn't have concurrent execution, this `activityActor` struct is shared for all actor instances, so we do need to be careful about concurrency. The `a.statesCache` field is a `sync.Map`, though, so we should be safe for concurrent execution here.

pkg/runtime/wfengine/backend.go
cgillum: Will fix this.

pkg/runtime/wfengine/backend.go
cgillum: It's not, but I never close this channel. Is there a particular best practice that I should be following regarding managing channel lifetimes in this case?

pkg/runtime/wfengine/backend.go
cgillum: Same question as in my previous response.

pkg/runtime/wfengine/backend.go
cgillum: Thanks for the tip - I will look into this. I think it will be necessary.

pkg/runtime/wfengine/component.go
cgillum: I figured this would be more efficient, but if maps are preferred over switch in cases like this then I'm happy to do it.

cgillum (commented):
Thanks for the feedback! I'll work on an update to address these items.

pkg/runtime/wfengine/activity.go
mcandeia: Just to clarify, I guess the correct type would be `type internalActorID` as you've introduced the logic behind using a specific actor ID for workflows and activities.

The idea is to have a single place where the internal actor id are built to make clear that this is a special ID that contains the workflow id on it

pkg/runtime/wfengine/backend.go
mcandeia: Theoretically the channel lifecycle is tied to the receiver, the receiver should always be responsible for closing the channel, and you can't do any assumptions like "this channel should never be closed" as you can't express it compile time (e.g using a channel type that can't be closed), so I'd rather handling closed channels just in case

pkg/runtime/wfengine/activity.go
mcandeia: It means that we can have concurrently executions, even using a sync.Map you can have racing conditions where both tries to load from the map at exactly the same time, both returning false and proceeding with the method execution.

A https://pkg.go.dev/golang.org/x/sync/singleflight would be a good choice for this. Using a singleflight with the actor id as a key should solve the problem

pkg/runtime/wfengine/backend.go
cgillum: Personally, I like to be cautious about doing "just in case" development because it can confuse the actual intent of the code. For example, if I'm new to the code base and see handling for a channel being closed, then I will naturally assume that the channel is _expected_ to be closed at some point. However, that's not actually true, which creates confusion for me as the reader. For that reason, I'd prefer we add handling for closed channels at the point we decide to start closing channels.

pkg/runtime/runtime.go
cgillum: Agreed. Done.

pkg/runtime/wfengine/activity.go
cgillum: The guarantees of actors ensures that we'll never have two concurrent operations with the same `actorID` value, so we don't need to worry about to requests for the same actor racing with each other. The only thing we need to protect against is operations for two _different_ actors causing a panic due to concurrent map modifications. `sync.Map` should give us all the protection we need for that.

pkg/runtime/wfengine/activity.go
cgillum: We don't build internal actor IDs anywhere in the Dapr codebase. They are all generated externally, in this case either by consumers of the workflow API/SDK or via the `durabletask-go` dependency in cases where the workflow orchestration IDs are auto-generated UUIDs. I don't think having an `internalActorID` type will serve much purpose, at least not right now.

cgillum:
I found a couple bugs during additional manual testing. I plan to push another commit to resolve those bugs.

cgillum:
@mcandeia this is ready for another round of review. I made a few changes to handle issues related to how we handle shutdown to fix some issues I found while testing manually.

I'm not sure why the `TestProxySuite` failed in the CI. It's passing locally for me. It might need a re-run.

pkg/runtime/wfengine/activity.go
mcandeia: Even worse, in my opinion. We make an assumption based on a string derived from a library/sdk implementation. I believe that the ideal is to have this resolved on the sdk side, since it is responsible for generating the ID.

My general comment here is that the information on how the ID is formatted (which character should be used, what are the first part of the id and the second) is not unified in one place, but rather spread out between the creator and the consumer. Due to the lack of a well-defined type, this is passed as simply a string and its deconstruction is not clear to anyone reading the code. In case the sdk/library decides to have more information on that ID, it would be very difficult to modify it without a lot of effort.

For now, this comment doesn't mean much as this is the only place where the ID is being deconstructed, but if this starts to increase it could be a problem. In that case, I don't know if you and I would be here to warn about it.

mcandeia: just to make sure we are on the same page, I'm talking about [this](https://github.com/dapr/dapr/pull/5784/files/f7accc5308949f4c6d152e439e5dfb7495b47932#diff-092e8c8daaca7a6f737bf96a9b901f50a8f0f4a4fa2183ef14b8b4cfd726ed5aR145)

pkg/runtime/wfengine/activity.go
mcandeia: Thanks for the clarification.

pkg/runtime/wfengine/backend.go
mcandeia: I think it's ok to keep it that way as long as you have a unit test that can catch the problem if the channel is closed erroneously. Otherwise you didn't add code here because of "just in case" and also failed to prevent an erroneously closed channel.

cgillum: I don't see much value in writing a test for a condition that's never expected to happen. It seems like a lot more work and just as confusing. I will go ahead and just add some logic to handle the channel being unexpectedly closed.

pkg/runtime/wfengine/activity.go
cgillum: Ah, I think I was misunderstanding your root concern. You're specifically talking about the convention for activity actor IDs, which is {workflowID}#{taskID}. This ID is generated from a function [here](https://github.com/dapr/dapr/pull/5784/files/f7accc5308949f4c6d152e439e5dfb7495b47932#diff-9997677e8aa2146ca3f09e604ca0f932a17a85fddc758fcc785c7ebb9afbc08aR457-R460). Would you be okay with me encapsulating this deconstruction logic in a function as well?

mcandeia: Sorry, my comment was wrongly pointed the line where the actor id is used. But yes, I will be fine with that, go ahead and create the function

pkg/runtime/wfengine/backend.go
mcandeia: Ok, you can leave it the way it is if it takes all this time to do the test, even the test is preferable in this case.

mcandeia (approved):

artursouza (approved):

PR #5786, opened 2023-01-20
Fixed log message

ItalyPaleAle:
Tiny error


mukundansundar (approved):
lgtm. 
qq : doesn't `%s` also work?

ItalyPaleAle:
Yes `%s` also works but semantically `%v` seems more appropriate for errors. The problem however wasn't the usage of `%s` but that one of the referenced items was missing and logs were malformed

DeepanshuA (approved):

ItalyPaleAle:
Update: I found another instance where the wrong formatting was used and the log was not displayed correctly

yaron2 (approved):

PR #5791, opened 2023-01-23
Update dependencies, pin new contrib

berndverst:

Pins a new contrib which:
- Updates Event Hubs to Track 2 SDK
- Updates Event Grid management SDK
- Upgrades Aerospike to the latest major SDK (to deal with old dependencies)
- Updates all dependencies to* their latest minor version

Updates runtime dependencies* to their latest minor versions.

Overwrites `github.com/chzyer/logex` unlicensed indirect dependency to a licensed version.

* Does not upgrade Hashicorp dependencies as the upgrade path is a bit more involved.


berndverst:
No idea where `github.com/chzyer/logex` is used by the way. `go mod why` didn't reveal anything

berndverst:
Pprof and the aerospike components appear to import this.

We should also import a newer version of pprof.

I'll look into aerospike.

go.mod
```diff
@@ -70,6 +70,8 @@ require (
 	sigs.k8s.io/controller-runtime v0.14.1
 )
 
+require github.com/go-zookeeper/zk v1.0.3 // indirect
```
ItalyPaleAle, line 73: Nit: can you merge this into the same "require" block below?

berndverst: That was freaking mod tidy. Yeah

berndverst:
Found via `go mod graph | grep "chzyer"` that this comes from the Aerospike component and from pprof.

I am upgrading aerospike and pprof to address this also. Will pin a new contrib momentarily.

mukundansundar:
@berndverst is there any specific documentation needed for track 2 migration of EH? I am seeing that the upgrade migration forced restarts will happen only during the rollout and change of Dapr sidecar. I think we should at least have the information in release notes if not docs for 1.10 as a special section in AEH.

mukundansundar (approved):
lgtm

berndverst:
> @berndverst is there any specific documentation needed for track 2 migration of EH? I am seeing that the upgrade migration forced restarts will happen only during the rollout and change of Dapr sidecar. I think we should at least have the information in release notes if not docs for 1.10 as a special section in AEH.

@mukundansundar this seriously is not the right place to ask these questions.

You can ask that on the contrib issues for the EventHubs upgrade or on those PRs. You also no full well that @ItalyPaleAle worked on this, not I. Thanks.

berndverst:
@yaron2 @artursouza would be great to get this merged soon so we can include this in more longhaul tests also.

ItalyPaleAle:
@mukundansundar Thanks for reminding me I need to create a [docs issue](https://github.com/dapr/docs/issues/3083). I think we could mention this in the release notes, but not sure it needs to be in the docs.

yaron2:
> @yaron2 @artursouza would be great to get this merged soon so we can include this in more longhaul tests also.

Will merge once tests pass.

PR #5792, opened 2023-01-23
Add component field to logs from components

ItalyPaleAle:
This PR updates the loggers used by components to include the name of the component that emitted the log. For example:

```
DEBU[0004] Receiving messages for subscription checkout to topic topic1  app_id=checkout component="order-pub-sub (pubsub.azure.servicebus/v1)" instance=asegala-mbp.local scope=dapr.contrib type=log ver=dev
```

(in JSON, it would be another field like `"component": "order-pub-sub (pubsub.azure.servicebus/v1)"`)

This is particularly useful when having multiple components and trying to identify which one is emitting the log.

### DIFF
--- a/pkg/runtime/runtime.go
+++ b/pkg/runtime/runtime.go
@@ -661,13 +661,15 @@ func (a *DaprRuntime) buildHTTPPipelineForSpec(spec config.PipelineSpec, targetP
 			middlewareSpec := spec.Handlers[i]
 			component, exists := a.getComponent(middlewareSpec.Type, middlewareSpec.Name)
 			if !exists {
-				return httpMiddleware.Pipeline{}, fmt.Errorf("couldn't find middleware component with name %s and type %s/%s",
-					middlewareSpec.Name,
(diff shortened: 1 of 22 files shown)


pkg/runtime/runtime.go
```diff
@@ -1564,16 +1566,15 @@ func (a *DaprRuntime) isAppSubscribedToBinding(binding string) (bool, error) {
 }
 
 func (a *DaprRuntime) initInputBinding(c componentsV1alpha1.Component) error {
-	binding, err := a.bindingsRegistry.CreateInputBinding(c.Spec.Type, c.Spec.Version)
+	fName := fmt.Sprintf(componentFormat, c.ObjectMeta.Name, c.Spec.Type, c.Spec.Version)
```
yaron2, line 1569: Is it worth refactoring this line into it's own function?

ItalyPaleAle: Good suggestion. Changed

yaron2 (approved):

PR #5811, opened 2023-01-26
Register state.sqlite component

ItalyPaleAle:
Also updates pinned components-contrib and some deps (like `github.com/hashicorp/golang-lru` to `/v2`)


yaron2:
There are many changes in this PR unrelated to the SQLite component. I assume these are a mistake?

ItalyPaleAle:
@yaron2 pulling in the latest contrib pulled in also updated dependencies. Those are the changes you see there. I took that opportunity to also update github.com/hashicorp/golang-lru to a v2 since it was marked as outdated. If you want i can split that out (that does include a few lines of code changes)

PR #5812, opened 2023-01-26
Ignore context canceled errors in proxy test

ItalyPaleAle:
The stream can receive a context canceled error if it's the end of the test. It should not be considered a test failure.

Fixes some flakiness in the gRPC proxy test


yaron2:
Linter

yaron2 (dismissed):

ItalyPaleAle:
Fixed the linter

artursouza (approved):

PR #5814, opened 2023-01-26
Pinned components-contrib RC + Updated some deps

ItalyPaleAle:
Updated some dependencies, also to make dependabot happy

Two dependencies that were updated required some small code changes


berndverst:
Why not wait until new contrib is merged and then pin the new contrib?

ItalyPaleAle:
> Why not wait until new contrib is merged and then pin the new contrib?

Done

Some deps here are unrelated from components-contrib

pkg/placement/raft/logger.go
```diff
@@ -86,6 +86,10 @@ func (l *loggerAdapter) ResetNamed(name string) hclog.Logger { return l }
 
 func (l *loggerAdapter) SetLevel(level hclog.Level) {}
 
+func (l *loggerAdapter) GetLevel() hclog.Level {
+	return hclog.Info
```
mukundansundar, line 90: why is this always returning info?

pkg/placement/raft/logger.go
```diff
@@ -86,6 +86,10 @@ func (l *loggerAdapter) ResetNamed(name string) hclog.Logger { return l }
 
 func (l *loggerAdapter) SetLevel(level hclog.Level) {}
```
mukundansundar, line 87: should dapr debug level log also affect this?

pkg/placement/raft/logger.go
ItalyPaleAle: This logger is kind of a nop and info is the default level

pkg/placement/raft/logger.go
ItalyPaleAle: It's a nop

berndverst (approved):

PR #5809, opened 2023-01-25
remove unnecessary blank line in `go.mod`

berndverst:
A little OCD perhaps but this was a mistake :)

You can force merge this PR right away @yaron2 / @artursouza

Signed-off-by: Bernd Verst <github@bernd.dev>

# Description

## Issue reference

Please reference the issue this PR will close: #_[issue number]_

## Checklist

Please make sure you've  completed the relevant tasks for this PR, out of the following list:

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing
* [ ] End-to-end tests passing
* [ ] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_
* [ ] Specification has been updated / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_
* [ ] Provided sample for the feature / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_

PR #5816, opened 2023-01-26
Add CI step to detect disallowed changes to go.mod

ItalyPaleAle:
Makes sure that people don't accidentally make commits with changes in the go.mod around the code to replace components-contrib

Catches issues such as #5809 too


yaron2 (approved):

yaron2:
Good change.

Issue #2478, opened 2020-11-21
Graceful shutdown


jjcollinge:
/area runtime
> /area operator
> /area placement
> /area docs
> /area test-and-release

Currently daprd has a graceful shutdown mechanism by which it responds to SIGTERM signals. This invokes `runtime.Stop` but this only seems to clean up actors currently. It does not stop any more incoming requests to the gRPC or HTTP server whilst trying to cleanup existing inflight requests and it doesn't attempt to clean up other resources or ensure connections are closed. Additionally, I have read but we need to verify that although `signal.Notify(stop, syscall.SIGTERM, os.Interrupt)` works on Windows, [it does not work in Windows containers](https://github.com/moby/moby/issues/25982).

An additional thought is, do we look to provide lifecycle hooks from Dapr into the app? Could we abstract away the underlying platform lifecycle hooks and invoke them directly on the user's app to encourage apps to behave correctly?

jjcollinge:
Missed the duplicate: https://github.com/dapr/dapr/issues/2391 but I still think there's some valid questions we should consider above.

pkedy:
Related to #3088

PR #5819, opened 2023-01-26
Propagate context from caller to appropriate places in the code

JoshVanL:
As part of a [wider initiative](https://github.com/dapr/dapr/issues/2478) for dapr to gracefully shutdown (as well as callers having better control of client connections etc), this PR adds contexts to appropriate functions so a caller context can be propagated.

This PR is based on [this contrib PR](https://github.com/dapr/components-contrib/pull/2474).

The signals package has been updated to use the correct signal to notify on based on OS. The intention is that this signals package should be the root context for all components and propagated throughout the entire program.

This PR doesn't actually do much to help with graceful shutdown since most go routines started are completely self contained and not waited for, and there are multiple code paths to terminate the program.

In follow up PRs I'll be adding a manager package to manage these long running go routines and add logic to wait for all go routines to finish and return before exiting the program. There likely are also places in the code which will need updating to better respect/respond to closed contexts. My feeling is that all `Close` functions should be removed in favor of relying on contexts, for example.

Targeting  `v1.11`.


ItalyPaleAle:
Thanks! Marking this as draft until the contrib PR is merged (when the 1.11 merge window opens)

pkg/sentry/monitoring/metrics.go
```diff
@@ -44,36 +44,35 @@ var (
 )
 
 // CertSignRequestReceived counts when CSR received.
-func CertSignRequestReceived() {
-	stats.Record(context.Background(), csrReceivedTotal.M(1))
```
ItalyPaleAle, line 48: For things like telemetry and monitoring, IMHO it's more correct to use a background context than one specific to the request, since the recording should happen regardless of what happens to the request.

pkg/diagnostics/service_monitoring.go
```diff
@@ -175,8 +174,6 @@ func newServiceMetrics() *serviceMetrics {
 			"The latency of service invocation response.",
 			stats.UnitMilliseconds),
 
-		// TODO: use the correct context for each request
-		ctx:     context.Background(),
```
ItalyPaleAle, line 179: Here too I personally think diagnostics should use a context background. Maybe one with a per-operation timeout, but that is nevertheless un-tied from the context of the parent request. So if the parent request is canceled, we still get the metrics.

pkg/http/api.go
```diff
@@ -529,29 +529,29 @@ func (a *api) constructDistributedLockEndpoints() []Endpoint {
 	}
 }
 
-func (a *api) onOutputBindingMessage(reqCtx *fasthttp.RequestCtx) {
-	name := reqCtx.UserValue(nameParam).(string)
-	body := reqCtx.PostBody()
+func (a *api) onOutputBindingMessage(ctx *fasthttp.RequestCtx) {
```
ItalyPaleAle, line 532: Please don't rename `reqCtx` to `ctx`.

Also very important to point out that for as long as we're using fasthttp (hopefully not much longer? See #4979), the context is actually NOT tied to a request, but to the server's own lifecycle (see #4979).

ItalyPaleAle (changes requested):
Thanks for this, it's definitely something we needed and very good hygiene too :)

My advice based on personal experience with such large changes (and borderline ask as a reviewer), however, is to consider splitting this into separate PRs. For example, a PR for each "ancillary" binary (operator, sentry, placement), and as many PRs as needed for "daprd". For "daprd", ideally we should have the change to components happen in a separate PR that can be merged more quickly after the PR in components-contrib is merged (changing an interface in components-contrib is a hassle and we need to move quickly after that's merged).

Not only it will make it easier on the reviewers, but you'll have the PRs merged more quickly and less hassle on you to fix constant merge conflicts. Again, this is based on personal experience :)

pkg/sentry/monitoring/metrics.go
JoshVanL: I agree with you but in this case this context is actually used as key value store for supplementary info (as far as I understand). https://github.com/census-instrumentation/opencensus-go/blob/0028eeaf06735bceeb492e091b29fb43ad8e5d90/stats/record.go#L154

I guess we are not using any of the tag features anywho.

pkg/http/api.go
JoshVanL: Could you describe a bit more what you mean by the context is tied to the server's lifecycle and not the request? Aren't all request contexts tied to both the server _and_ request by virtue of the server's context being wrapped by the request's context?

JoshVanL:
Thanks @ItalyPaleAle! I hear you with it being a big PR- will split this up!

pkg/http/api.go
ItalyPaleAle: It’s a “feature” of fasthttp (“feature” according to the authors of the library). The context passed to each handler is canceled only when the server is stopped. If the request is canceled by the user, the context is not canceled. I’ve explained it more in #4979, there’s a section about that and a quote from the fasthttp maintainers.

JoshVanL:
Closing this PR since it is being split up by component.

Issue #5808, opened 2023-01-25
Configuration CRD has incorrect key for metrics configuration


NickLarsenNZ:
## In what area(s)?

/area operator

## What version of Dapr?

1.9.5

## Expected Behavior

I expected the `spec.metrics` to be valid as per documentation: https://docs.dapr.io/operations/configuration/configuration-overview/#metrics

## Actual Behavior

The key is actually `spec.metric` (without the `s`):

https://github.com/dapr/dapr/blob/8c70b2104a09803a392960ed00f6e68310399949/charts/dapr/crds/configuration.yaml#L218

_You could argue that the documentation is incorrect, but I think the key should be `spec.metrics`_.

## Steps to Reproduce the Problem

N/A

## Release Note

RELEASE NOTE:

ItalyPaleAle:
> You could argue that the documentation is incorrect, but I think the key should be spec.metrics.

I agree, but I'm concerned with modifying this now, as this seems to have been in place for a long time.

@yaron2 thoughts?

yaron2:
We want to support both and over time deprecate `metric`. So far the only thing you could do is disabled metrics alltogether, but now as we have additional functionality, might want to look at it sooner rather than later.

PR #5824, opened 2023-01-26
Add support for .metrics in addition to .metric

yaron2:
Closes https://github.com/dapr/dapr/issues/5808.

### DIFF
--- a/pkg/config/configuration.go
+++ b/pkg/config/configuration.go
@@ -345,7 +345,7 @@ func LoadKubernetesConfiguration(config, namespace string, podName string, opera
 	return conf, nil
 }
 
-// Apply .spec.metrics if set. If not, retain spec.metric.
+// Apply .metrics if set. If not, retain .metric.
 func sortMetricsSpec(conf *Configuration) {
 	if !conf.Spec.MetricsSpec.Enabled {
 		conf.Spec.MetricSpec.Enabled = false
--- a/pkg/config/configuration.go
+++ b/pkg/config/configuration.go
@@ -98,6 +98,7 @@ type ConfigurationSpec struct {
 	TracingSpec         TracingSpec        `json:"tracing,omitempty" yaml:"tracing,omitempty"`
 	MTLSSpec            MTLSSpec           `json:"mtls,omitempty" yaml:"mtls,omitempty"`
 	MetricSpec          MetricSpec         `json:"metric,omitempty" yaml:"metric,omitempty"`
+	MetricsSpec         MetricSpec         `json:"metrics,omitempty" yaml:"metrics,omitempty"`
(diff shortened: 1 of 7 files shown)


mukundansundar:
@yaron2 can you change existing e2e to use metrics instead of metric?

pkg/config/configuration.go
```diff
@@ -98,6 +98,7 @@ type ConfigurationSpec struct {
 	TracingSpec         TracingSpec        `json:"tracing,omitempty" yaml:"tracing,omitempty"`
 	MTLSSpec            MTLSSpec           `json:"mtls,omitempty" yaml:"mtls,omitempty"`
 	MetricSpec          MetricSpec         `json:"metric,omitempty" yaml:"metric,omitempty"`
```
ItalyPaleAle, line 100: Should this be made a pointer so we can check if it's nil?

pkg/config/configuration.go
```diff
@@ -336,9 +341,21 @@ func LoadKubernetesConfiguration(config, namespace string, podName string, opera
 		return nil, err
 	}
 
+	sortMetricsSpec(conf)
 	return conf, nil
 }
 
+// Apply .metrics if set. If not, retain .metric.
+func sortMetricsSpec(conf *Configuration) {
+	if !conf.Spec.MetricsSpec.Enabled {
```
ItalyPaleAle, line 350: I think that if we use .metrics for one thing, we should use it for all. With this code, a user could specify .metrics for one thing (for example, enabled), and .metric for the other (rules).

ItalyPaleAle (changes requested):
Could you also add a warning log for people who are using .metric that they need to switch to .metrics?

yaron2:
> @yaron2 can you change existing e2e to use metrics instead of metric?

I thought I did, with the testdata files. Can you point me to other locations?

pkg/config/configuration.go
yaron2: The issue is that both metric and metrics are set to true by default, so we can't know if one field was set unless it's false, in which case all other settings are ignored. It's unlikely someone would use both, but if they do, it should be legal for as long as we support both I think.

pkg/config/configuration.go
yaron2: In effect this can never be nil as we explicitly set it to true by default.

mukundansundar (approved):

artursouza (approved):

ItalyPaleAle (approved):

mukundansundar:
@yaron2 I think a doc change is needed for this as well. 
cc @msfussell

yaron2:
> @yaron2 I think a doc change is needed for this as well. cc @msfussell

Our docs have always mentioned `metrics`.

PR #5825, opened 2023-01-27
Load secret stores before other kinds of components

### LINKED ISSUE #5727: Components that reference secrets in secret stores show an ominous warning when secret store is loaded after component
## In what area(s)?

/area runtime

## What version of Dapr?

> 1.9

### Description

User-reported issue. Components that reference a secret in a secret store (not the built-in K8s one) sometimes fail to initialize if the secret store component is loaded after the component that needs the secrets.

Logs shared by user:

image

## Release Note

RELEASE NOTE: **FIX** Components that reference secrets from secret stores fail to init if the secret store component is loaded after

ItalyPaleAle:
This makes it so users won't see a warning like `component X references a secret store that isn't loaded` if the component was just loaded later 

> Note: users could still see the warning in case a secret store references secrets from another secret store that is loaded after. For example, if secret store A references a secret in store B (it will still work, but there'll be a warning). This doesn't apply to secrets that are loaded from the built-in K8s secret store, which is always loaded as the very first thing.

Fixes #5727

### DIFF
--- a/pkg/runtime/runtime.go
+++ b/pkg/runtime/runtime.go
@@ -2476,12 +2476,12 @@ func (a *DaprRuntime) loadComponents(opts *runtimeOpts) error {
 	// First, we look for secret stores and load those, then all other components
 	// Sure, we could sort the list of authorizedComps... but this is simpler and most certainly faster
 	for _, comp := range authorizedComps {
-		if strings.HasPrefix(comp.Spec.Type, "secretstores.") {
+		if strings.HasPrefix(comp.Spec.Type, string(components.CategorySecretStore)+".") {
 			a.pendingComponents <- comp
 		}
 	}
 	for _, comp := range authorizedComps {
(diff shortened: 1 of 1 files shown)


pkg/runtime/runtime.go
```diff
@@ -2472,8 +2472,18 @@ func (a *DaprRuntime) loadComponents(opts *runtimeOpts) error {
 	a.components = authorizedComps
```
mukundansundar, line 2472: With the change below should a.components also be changed to maintain the same order?

ItalyPaleAle: I looked into that but personally don't think it's needed. Do you think it would make a difference?

mukundansundar: looked into the code and I think you are right, the order in the components slice does not seem to make a difference.

artursouza (approved):

PR #5826, opened 2023-01-27
Fix workflow engine startup race

### LINKED ISSUE #5822: Workflow engine doesn't start when --app-port is configured
## In what area(s)?

> /area runtime

## What version of Dapr?

> v1.9.1-330-g81bc49c3-dirty

## Expected Behavior

Workflow engine should start up normally, regardless of whether `--app-port` is configured.

## Actual Behavior

Workflow engine startup fails with the following error output (note that this error output also needs to be fixed):

```
ERRO[0015] failed to auto-start the workflow engine: %!w(*errors.errorString=&{actor runtime is not configured})  app_id=foo instance=CGILLUM-DEV2 scope=dapr.runtime.wfengine type=log ver=edge
```

Use the workflow sample in the .NET SDK, but add `--app-port=10081` to the Dapr command line.

## Workaround

Don't use `--app-port` in your Dapr command-line - it's not needed for workflows.

cgillum:
# Description

Test automation discovered a race condition in the workflow engine startup that prevents initialization from succeeding when `--app-port` is specified in the dapr command line. The issue is that the Dapr workflow app may connect to the sidecar before the sidecar has finished initializing. When this happens, the sidecar gets into a bad state where the workflow engine never starts up.

The fix in this PR is twofold:

1. Update `durabletask-go` to allow the sidecar to reject connections by the client when the workflow engine isn't ready yet (https://github.com/microsoft/durabletask-go/pull/7) and update the workflow engine startup path to flow initialization errors back to the remote client so that it knows to retry later.
2. Fix the workflow engine startup code so that startup failures aren't permanent (replace the `Once` to a `Mutex`).

With these fixes, the client workflow app will retry connecting to the sidecar continuously until it's fully initialized, solving the race.

## Issue reference

Fixes #5822 

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests
* [x] Unit tests passing

FYI @RyanLettieri @johnewart @mcandeia @artursouza


go.mod
```diff
@@ -30,7 +30,7 @@ require (
 	github.com/hashicorp/raft-boltdb v0.0.0-20220329195025-15018e9b97e0
 	github.com/jhump/protoreflect v1.14.1
 	github.com/kelseyhightower/envconfig v1.4.0
-	github.com/microsoft/durabletask-go v0.1.2
+	github.com/microsoft/durabletask-go v0.1.3-0.20230127011117-67aad9acf4d1
```
ItalyPaleAle, line 33: Do you think you can make a new release with the patch before we release Dapr?

pkg/runtime/wfengine/wfengine.go
```diff
@@ -145,41 +148,40 @@ func (wfe *WorkflowEngine) SetActorReminderInterval(interval time.Duration) {
...
-		}
+	if wfe.IsRunning {
+		return nil
+	}
 
-		// TODO: Determine whether a more dynamic parallelism configuration is necessary.
-		parallelismOpts := backend.WithMaxParallelism(100)
+	if wfe.actorRuntime == nil {
+		return errors.New("actor runtime is not configured")
+	} else if wfe.executor == nil {
```
ItalyPaleAle, line 161: nit: just use another `if` without `else`

pkg/runtime/wfengine/wfengine.go
```diff
@@ -145,41 +148,40 @@ func (wfe *WorkflowEngine) SetActorReminderInterval(interval time.Duration) {
...
+	wfe.worker = backend.NewTaskHubWorker(wfe.backend, orchestrationWorker, activityWorker, wfLogger)
+	if err := wfe.worker.Start(ctx); err != nil {
+		return fmt.Errorf("failed to start workflow engine: %w", err)
+	}
 
-	return err
+	wfe.IsRunning = true
+	wfLogger.Info("workflow engine started")
+
+	return nil
```
ItalyPaleAle, line 184: (comment about L197, I can't comment there)

The method `Stop()` doesn't have a lock so there could be race conditions in modifying `IsRunning`

pkg/runtime/wfengine/wfengine.go
```diff
@@ -145,41 +148,40 @@ func (wfe *WorkflowEngine) SetActorReminderInterval(interval time.Duration) {
...
-	wfe.startedOnce.Do(func() {
-		if wfe.actorRuntime == nil {
-			err = errors.New("actor runtime is not configured")
-			return
-		} else if wfe.executor == nil {
-			err = errors.New("grpc executor is not yet configured")
-			return
-		}
+	// Start could theoretically get called by multiple goroutines concurrently
+	wfe.startMutex.Lock()
```
ItalyPaleAle, line 152: If the problem you're trying to solve is make sure this isn't started more than once, I'd use a simpler semaphore:

```go
// property m.isStarted is sync.Bool{}

if !m.isStarted.CompareAndSwap(false, true) {
    // If we're here, it's already running
    return nil
}
```

Then to stop this:

```go
m.isStarted.Store(false)
```

This doesn't require a lock, which means that concurrent callers will get a result (nil) right away rather than having to wait for the first caller's invocation of `Start()` to complete.

ItalyPaleAle (changes requested):

pkg/runtime/wfengine/wfengine.go
cgillum: Thanks for sharing this valuable tip! I was not aware of `sync.Bool`. However, I think I would actually prefer to keep the mutex in this case. That way I can ensure that the `Start()` method always returns only after the engine has actually started. This code path will be infrequently hit, so I'm not worried about the perf impact. :)

artursouza (approved):

pkg/runtime/wfengine/wfengine.go
mcandeia: this is a perfect fit for [singleflights](https://pkg.go.dev/golang.org/x/sync/singleflight), but this is a nit

mcandeia (approved):

ItalyPaleAle (approved):

RyanLettieri:
Confirmed that these changes do work with the .NET SDK changes. After pulling in this PR and testing locally, the issue is resolved. LGTM

PR #5845, opened 2023-01-28
Fix makefile: Docker build

shivamkm07:
Signed-off-by: Shivam Kumar <shivamkm07@gmail.com>

# Description

https://github.com/dapr/dapr/pull/5841 added `test -f` check in docker build step to check if binaries exist. However, if a binary doesn't exist, it will throw up error as `test` will return a non-zero error code, causing workflow step to fail.

This PR uses the BINARIES environmental variable (already-existing) instead.

## Issue reference

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing


ItalyPaleAle (changes requested):
No need to add || true at the end of commands, which would actually make the pipeline succeed in case of errors

docker/docker.mk
```diff
@@ -110,15 +110,20 @@ ifeq ($(ONLY_DAPR_IMAGE),true)
 else
 	$(DOCKER) build --build-arg PKG_FILES=* -f $(DOCKERFILE_DIR)/$(DOCKERFILE) $(BIN_PATH) -t $(DOCKER_IMAGE_TAG)-$(TARGET_OS)-$(TARGET_ARCH)
 	test -f "$(BIN_PATH)/daprd" && \
-		$(DOCKER) build --build-arg PKG_FILES=daprd -f $(DOCKERFILE_DIR)/$(DOCKERFILE) $(BIN_PATH) -t $(DAPR_RUNTIME_DOCKER_IMAGE_TAG)-$(TARGET_OS)-$(TARGET_ARCH)
+		$(DOCKER) build --build-arg PKG_FILES=daprd -f $(DOCKERFILE_DIR)/$(DOCKERFILE) $(BIN_PATH) -t $(DAPR_RUNTIME_DOCKER_IMAGE_TAG)-$(TARGET_OS)-$(TARGET_ARCH) || \
```
ItalyPaleAle, line 113: I just saw your message on Teams @shivamkm07 and I understand the problem now.

I am still not a fan of `|| true` because it would mask errors if the image fails to build.

How about:

```sh
if [ -f "$(BIN_PATH)/daprd" ]; then \
		$(DOCKER) build --build-arg PKG_FILES=daprd -f $(DOCKERFILE_DIR)/$(DOCKERFILE) $(BIN_PATH) -t $(DAPR_RUNTIME_DOCKER_IMAGE_TAG)-$(TARGET_OS)-$(TARGET_ARCH)
fi
```

This way it should not return a non-0 status code if the `test -f` fails.

shivamkm07: Done

shivamkm07:
Moving to draft as it needs some more changes

ItalyPaleAle (approved):
LGTM, thanks!

berndverst (approved):
I see @ItalyPaleAle edited the PR description in the middle of my review :p

Issue #5800, opened 2023-01-24
Improve perf tests summary output for resources usage


mcandeia:
## In what area(s)?

> /area runtime

> /area operator

> /area placement

> /area docs

/area test-and-release

## Describe the feature
Currently, when the performance test is passing the summary output does not contains any relevant information regarding compute resources like memory and CPU. Even when the test is passing these numbers should be monitored frequently.

The idea is to provide a simple markdown file or a table-like output for the compute resources used by the tests and write them as a workflow artifact to be used later.

## Release Note

RELEASE NOTE:

PR #5856, opened 2023-01-30
Enhanced github job summary for tests

mcandeia:
# Description

Motivated by @yaron2 I've created a simple job summary for tests that allow developers to output any amount of arbitrary data generated by tests instead of unstructured logs.

For each test there's an associated table of data that should be flushed when the test is finished, then, later on, the github action will use the summary to write [a enhanced github job summary](https://github.blog/2022-05-09-supercharging-github-actions-with-job-summaries/)

[Here](https://github.com/mcandeia/dapr/actions/runs/4047316765) you can see an example of it.

## Issue reference

Please reference the issue this PR will close: #5800 

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing


.github/scripts/dapr_tests_summary.js
```diff
@@ -0,0 +1,11 @@
+
+module.exports = async ({ glob, core }) => {
+    const globber = await glob.create(process.env["TEST_OUTPUT_FILE_PREFIX"] + "_summary_table_*.json")
+    for await (const file of globber.globGenerator()) {
+        const testSummary = require(file) // the file is a json file so we can import
```
ItalyPaleAle, line 5: Would prefer to see `await fs.readFile(file, 'utf8')` used here because using `require` to import JSON has some caveats (like with caching and possibly some other odd behaviors)

.github/workflows/dapr-perf.yml
```diff
@@ -462,11 +462,18 @@ jobs:
           #TODO: .json suffix can be removed from artifact name after test analytics scripts are updated
           name: test_perf.json
           path: ${{ env.TEST_OUTPUT_FILE_PREFIX }}_perf*.*
-      - name: Show test summary
+      - name: Add job test summary
         if: always()
         uses: test-summary/action@v2
         with:
           paths: ${{ env.TEST_OUTPUT_FILE_PREFIX }}_perf*.xml
+      - name: Add job test outputs
```
ItalyPaleAle, line 470: Please don't run JS files as GitHub scripts unless they need access to the GitHub token, also for security reasons...

May be enough to just add `#!/usr/bin/env node` and then use this as a script (don't forget to `chmod +x` before committing the script).

Example: https://github.com/dapr/dapr/blob/master/.github/workflows/dapr.yml#L140-L142

Bonus: this allows you to name your file `.mjs` and then use ES imports and top-level await.

tests/runner/summary/summary.go
```diff
@@ -0,0 +1,83 @@
...
+// the table output is typically used as a github enhanced job summary.
+// see more: https://github.blog/2022-05-09-supercharging-github-actions-with-job-summaries/
+type Table struct {
+	Test string      `json:"test"`
+	Data [][2]string `json:"data"`
+}
+
+// Output adds a pair to the table data pairs.
+func (t *Table) Output(header, value string) *Table {
+	t.Data = append(t.Data, [2]string{header, value})
```
ItalyPaleAle, line 42: You should be able to write to the output file as a stream (`os.Open` creates an `io.WriteCloser`) so you don't have to buffer the data in memory. Then, `Flush()` would just close the writable stream.

Not going to block the PR on this, but if you have a chance to do that it would be very nice.

ItalyPaleAle (changes requested):
Please also fix the mod tidy issue

tests/runner/summary/summary.go
mcandeia: I don't know if I understand your approach. Could you elaborate more?

I would understand if I was writing raw strings but I have to serialize to JSON first, unless you're suggesting that I use strings directly instead that would complicate things a bit

.github/workflows/dapr-perf.yml
ItalyPaleAle: The security implication is that scripts that run via the github-script action have access to the GitHub APIs pre-authorized and can make API calls that impact the repo. There are many cases where that's necessary, but if you don't (like in this case), it would be best to just run the script as a Node.js one with the `env node` hashbang.

Using `@actions/core` and `@actions/glob` would require github-script (likely), but I don't see them being used right now? Please correct me if I'm wrong.

tests/runner/summary/summary.go
ItalyPaleAle: Never mind, I missed the part where you were serializing the entire list as JSON.

.github/workflows/dapr-perf.yml
mcandeia: [Yes, I'm using them](https://github.com/dapr/dapr/pull/5856/files#diff-aed9855ec5ff916d6ced3a125bd9b76551a22febf05b7e6d4bafd9222eb26dfdR476)

.github/scripts/dapr_tests_summary.js
mcandeia: alright, that's why I prefer [deno](https://deno.land/) xD

ItalyPaleAle (dismissed):

artursouza (approved):

PR #5831, opened 2023-01-27
Graceful Shutdown: Updates function signatures to match new API, and add context.TODO()s

JoshVanL:
Part of https://github.com/dapr/dapr/issues/2478

This PR updates functions with contexts as implemented [here](https://github.com/dapr/components-contrib/pull/2474).

Callers of these functions use a `context.TODO()`, with the intention of implementing proper context propagation in later PRs.

/milestone v1.11
/cc @ItalyPaleAle

PR #5859, opened 2023-01-31
Graceful Shutdown: Operator

JoshVanL:
Part of https://github.com/dapr/dapr/issues/2478 and based on https://github.com/dapr/dapr/pull/5831

/milestone v1.11

Removed `webhook.go` in favor of moving that code into `operator.go`. I have also united two controller-runtime managers since there is no point running two.


pkg/components/state/pluggable.go
```diff
@@ -44,7 +44,7 @@ type grpcStateStore struct {
 
 // Init initializes the grpc state passing out the metadata to the grpc component.
 // It also fetches and set the current components features.
-func (ss *grpcStateStore) Init(metadata state.Metadata) error {
+func (ss *grpcStateStore) Init(ctx context.Context, metadata state.Metadata) error {
```
mcandeia, line 47: please replace the `ss.Context` with the ctx received as a parameter now

mcandeia (changes requested):

pkg/concurrency/runner_test.go
```diff
@@ -0,0 +1,217 @@
+package concurrency
```
mcandeia, line 1: missing copyright 2023

pkg/concurrency/runner.go
```diff
@@ -0,0 +1,89 @@
+package concurrency
```
mcandeia, line 1: missing copyright 2023

pkg/components/bindings/input_pluggable.go
```diff
@@ -33,7 +33,7 @@ type grpcInputBinding struct {
 }
 
 // Init initializes the grpc inputbinding passing out the metadata to the grpc component.
-func (b *grpcInputBinding) Init(metadata bindings.Metadata) error {
+func (b *grpcInputBinding) Init(ctx context.Context, metadata bindings.Metadata) error {
```
mcandeia, line 36: replace the ss.Context since you're receiving as a parameter

pkg/components/bindings/output_pluggable.go
```diff
@@ -30,7 +30,7 @@ type grpcOutputBinding struct {
 }
 
 // Init initializes the grpc outputbinding passing out the metadata to the grpc component.
-func (b *grpcOutputBinding) Init(metadata bindings.Metadata) error {
+func (b *grpcOutputBinding) Init(ctx context.Context, metadata bindings.Metadata) error {
```
mcandeia, line 33: replace the ss.Context since you're receiving as a parameter

pkg/components/pubsub/pluggable.go
```diff
@@ -35,7 +35,7 @@ type grpcPubSub struct {
 
 // Init initializes the grpc pubsub passing out the metadata to the grpc component.
 // It also fetches and set the component features.
-func (p *grpcPubSub) Init(metadata pubsub.Metadata) error {
+func (p *grpcPubSub) Init(ctx context.Context, metadata pubsub.Metadata) error {
```
mcandeia, line 38: replace the ss.Context since you are receiving as a parameter

pkg/concurrency/runner.go
```diff
@@ -0,0 +1,89 @@
...
+
+// RunnerManager is a manager for runners. It runs all runners in parallel and
+// waits for all runners to finish. If any runner returns, the RunnerManager
+// will stop all other runners and return any error.
+type RunnerManager struct {
+	errs    []string
+	lock    sync.Mutex
+	wg      sync.WaitGroup
+	runners []Runner
+	running chan struct{}
```
mcandeia, line 22: isn't an atomic bool enough for this scenario?

fyi:https://pkg.go.dev/sync/atomic

pkg/concurrency/runner.go
```diff
@@ -0,0 +1,89 @@
...
+	return nil
+}
+
+// Run runs all runners in parallel and waits for all runners to finish. If any
+// runner returns, the RunnerManager will stop all other runners and return any
+// error.
+func (r *RunnerManager) Run(ctx context.Context) error {
+	select {
+	case <-r.running:
+		return errors.New("manager already started")
```
mcandeia, line 53: could you please extract this and reuse as a variable?

`var ErrManagerAlreadyStarted = errors.New("manager already started")

pkg/signals/signals_posix.go
```diff
@@ -0,0 +1,24 @@
...
+*/
+
+package signals
+
+import (
+	"os"
+	"syscall"
+)
+
+var shutdownSignals = []os.Signal{os.Interrupt, syscall.SIGTERM}
```
mcandeia, line 24: qq: why the other signals are not considered as shutdown?

syscall.SIGHUP, syscall.SIGINT, syscall.SIGQUIT

pkg/concurrency/runner.go
```diff
@@ -0,0 +1,89 @@
...
+	"sync"
+)
+
+// Runner is a function that runs a task.
+type Runner func(context.Context) error
+
+// RunnerManager is a manager for runners. It runs all runners in parallel and
+// waits for all runners to finish. If any runner returns, the RunnerManager
+// will stop all other runners and return any error.
+type RunnerManager struct {
```
mcandeia, line 17: nit, I'd rather prefer the names used by js and .NET

async.All(...taskList)/async.Any(...taskList)

mcandeia (changes requested):
first pass

pkg/signals/signals_posix.go
JoshVanL: `os.Interrupt` is SIGINT on posix.

I left SIGHUP out as that is typically reserved for hot reloading, rather than terminating the program. It might be worth adding that at a later date.

Is SIGQUIT something we really want to be catching?

pkg/concurrency/runner.go
JoshVanL: Not too sure what you mean- do you mean renaming `RunnerManager` to `async`?

pkg/concurrency/runner.go
ItalyPaleAle: I agree. A Compare-and-swap may be simpler as semaphore

pkg/concurrency/runner.go
```diff
@@ -0,0 +1,89 @@
...
+// Run runs all runners in parallel and waits for all runners to finish. If any
+// runner returns, the RunnerManager will stop all other runners and return any
+// error.
+func (r *RunnerManager) Run(ctx context.Context) error {
+	select {
+	case <-r.running:
+		return errors.New("manager already started")
+	default:
+	}
+	close(r.running)
```
ItalyPaleAle, line 56: Why closing the channel here?

pkg/concurrency/runner.go
```diff
@@ -0,0 +1,89 @@
...
+				defer r.lock.Unlock()
+
+				r.errs = append(r.errs, err.Error())
+			}
+		}(runner)
+	}
+
+	r.wg.Wait()
+	if len(r.errs) > 0 {
+		return fmt.Errorf("%s", strings.Join(r.errs, "; "))
```
ItalyPaleAle, line 86: Doing this makes it so errors cannot be unwrapped however. How about using multierr (the Uber package we already use)?

Or... if you can wait 2 days for Go 1.20: https://tip.golang.org/doc/go1.20#errors

pkg/health/server.go
```diff
@@ -58,35 +60,32 @@ func (s *server) Run(ctx context.Context, port int) error {
...
 	//nolint:gosec
 	srv := &http.Server{
-		Addr:    fmt.Sprintf(":%d", port),
-		Handler: router,
+		Addr:        fmt.Sprintf(":%d", port),
+		Handler:     router,
+		BaseContext: func(_ net.Listener) context.Context { return ctx },
 	}
 
+	serveErr := make(chan error, 1)
```
ItalyPaleAle, line 68: I'm not sure I understand this change.

You're starting the server in background and capturing the error in serveErr. However, you're also blocking until the context is canceled.
So in order to be able to return the error, the context has to be canceled. Unlike the existing version, if the server fails to start there isn't even a log now.

pkg/operator/api/api.go
```diff
@@ -63,72 +62,81 @@ type apiServer struct {
...
 	allConnUpdateChan map[string]chan *componentsapi.Component
+	ready             chan struct{}
 }
 
 // NewAPIServer returns a new API server.
 func NewAPIServer(client client.Client) Server {
 	return &apiServer{
 		Client:            client,
 		allConnUpdateChan: make(map[string]chan *componentsapi.Component),
+		ready:             make(chan struct{}),
```
ItalyPaleAle, line 73: Here too IMHO a sync.Atomic.Bool and then using CompareAndSwap could make a better semaphore

pkg/operator/api/api.go
```diff
@@ -63,72 +62,81 @@ type apiServer struct {
...
 	lis, err := net.Listen("tcp", fmt.Sprintf(":%v", serverPort))
 	if err != nil {
-		log.Fatalf("error starting tcp listener: %v", err)
+		return fmt.Errorf("error starting tcp listener: %v", err)
 	}
 
-	if onReady != nil {
-		onReady()
-	}
+	close(a.ready)
```
ItalyPaleAle, line 99: Still unclear why the channel is closed if the goal is to use it as semaphore

pkg/operator/api/api.go
```diff
@@ -291,22 +299,24 @@ func (a *apiServer) ListResiliency(ctx context.Context, in *operatorv1pb.ListRes
 func (a *apiServer) ComponentUpdate(in *operatorv1pb.ComponentUpdateRequest, srv operatorv1pb.Operator_ComponentUpdateServer) error { //nolint:nosnakecase
 	log.Info("sidecar connected for component updates")
 	key := uuid.New().String()
```
ItalyPaleAle, line 301: (Unrelated from your PR but can you change this to uuid.NewRandom() which returns an error rather than panic if there's not enough entropy?)

pkg/operator/operator.go
```diff
@@ -85,20 +93,22 @@ func init() {
...
+		return nil, fmt.Errorf("Unable to get controller runtime configuration, err: %s", err)
 	}
 	mgr, err := ctrl.NewManager(conf, ctrl.Options{
-		Scheme:             scheme,
-		MetricsBindAddress: "0",
-		LeaderElection:     opts.LeaderElection,
-		LeaderElectionID:   "operator.dapr.io",
-		Namespace:          opts.WatchNamespace,
+		Scheme:                 scheme,
+		Port:                   19443,
```
ItalyPaleAle, line 103: Why is the port hardcoded?

pkg/operator/operator.go
```diff
@@ -111,18 +121,16 @@ func NewOperator(opts Options) Operator {
 			interval:          opts.WatchdogInterval,
 			maxRestartsPerMin: opts.WatchdogMaxRestartsPerMin,
 		}
-		err = mgr.Add(wd)
```
ItalyPaleAle, line 114: Nit, but I very much prefer not condensing the error into the `if`. More readable, and avoids allocating a new variable for holding the error :)

pkg/operator/operator.go
```diff
@@ -134,108 +142,246 @@ func NewOperator(opts Options) Operator {
...
-		log.Fatalf("Failed to wait for cache sync")
+		return nil
+	}}); err != nil {
+		return err
 	}
-	o.prepareConfig()
 
-	// load certs from disk
-	certChain := o.loadCertChain(ctx)
+	if err := o.mgr.Add(nonLeaderRunnable{func(ctx context.Context) error {
```
ItalyPaleAle, line 229: Like this one where there's a function inside an if block, seems pretty hard to read :(

ItalyPaleAle (changes requested):

pkg/health/server.go
JoshVanL: Thanks for the catch! This should be:


```go
	select {
	case <-serveErr:
		return serveErr
	case <-ctx.Done():
	}
```

instead of just

```go
<-ctx.Done()
```

pkg/concurrency/runner.go
JoshVanL: A closed running channel signals that the manager is running.

pkg/components/bindings/output_pluggable.go
JoshVanL: See above

pkg/components/pubsub/pluggable.go
JoshVanL: See above

pkg/concurrency/runner.go
JoshVanL: I guess we can just wait

pkg/concurrency/runner.go
JoshVanL: I've just gone ahead with the atomic strategy

pkg/operator/api/api.go
JoshVanL: If we change ready to an atomic bool, there is no way for `Ready` to block until the api is ready or the context is cancelled.

pkg/operator/api/api.go
JoshVanL: A closed ready channel signals that the apiServer is ready.

pkg/operator/operator.go
JoshVanL: The webhook port was already hardcoded.

https://github.com/dapr/dapr/pull/5859/files#diff-eb049014aed7f180e4ec0ec97c303124cc17ca08d97f9520bc0c4637a6fd1fdeL34

pkg/operator/api/api.go
ItalyPaleAle: But then you don’t have a semaphore anymore to prevent others from calling the method… or not?

pkg/concurrency/runner.go
```diff
@@ -48,12 +60,10 @@ func (r *RunnerManager) Add(runner Runner) error {
 // runner returns, the RunnerManager will stop all other runners and return any
 // error.
 func (r *RunnerManager) Run(ctx context.Context) error {
-	select {
-	case <-r.running:
-		return errors.New("manager already started")
-	default:
+	if r.running.Load() {
```
ItalyPaleAle, line 63: Between `r.running.Load()` and `r.running.Store(true)` there can be a race condition, so you can have 2 goroutines continuing with the init at the same time.

The nice thing about sync.Atomic is the compare-and-swap which is fully atomic.

So:

```go
if !r.running.CompareAndSwap(false, true) {
Return ErrManagerAlreadyStarted
}
```

This works because CompareAndSwap atomically (in a single operation) checks if the value was false, and only if it was false, it sets it to true. If the swap happened (i.e. it was false before and now it’s true), it returns true. If the value was already true (Run was already called), then no swap happens.

pkg/operator/api/api.go
JoshVanL: A new `apiServer` will have a `ready` channel which is not closed. This means the code block 

```go
	select {
	case <-a.ready:
		return errors.New("api server already running")
	default:
	}
```

will reach the default case and not error. When a listener has been created successfully, the ready channel will be closed. The above select statement will then entry the channel case and error.

pkg/concurrency/runner.go
mcandeia: basically, and use an approach more functional-like instead of object-like

async.Any(runners[]...) does all the thing you need to

pkg/signals/signals_posix.go
ItalyPaleAle: I'm ok with SIGINT and SIGTERM only.

ItalyPaleAle (approved):

pkg/health/server.go
```diff
@@ -77,7 +77,12 @@ func (s *server) Run(ctx context.Context, port int) error {
 		serveErr <- nil
 	}()
 
-	<-ctx.Done()
+	select {
+	case err := <-serveErr:
+		return err
```
mukundansundar, line 82: since serveErr is returned here, I think line 94 should be changed to return the error returned from Shutdown call. Also this call to `<-serveErr` in line 94 can block indefinitely, if for some reason `ListenAndServe` returns before the context is done?

pkg/operator/api/api.go
```diff
@@ -63,72 +63,81 @@ type apiServer struct {
...
+	if !a.running.CompareAndSwap(false, true) {
+		return errors.New("api server already running")
+	}
+
 	log.Infof("starting gRPC server on port %d", serverPort)
 
 	opts, err := daprCredentials.GetServerOptions(certChain)
 	if err != nil {
-		log.Fatalf("error creating gRPC options: %v", err)
+		return fmt.Errorf("error creating gRPC options: %w", err)
```
mukundansundar, line 89: nitpick
```suggestion
		return fmt.Errorf("error getting gRPC server options: %w", err)
```

pkg/operator/api/api.go
```diff
@@ -63,72 +63,81 @@ type apiServer struct {
...
 	err = lis.Close()
 	if err != nil {
-		log.Errorf("failed to close tcp listener: %v", err)
+		return fmt.Errorf("error closing listener: %w", err)
 	}
+	return nil
 }
 
-func (a *apiServer) OnComponentUpdated(component *componentsapi.Component) {
+func (a *apiServer) OnComponentUpdated(ctx context.Context, component *componentsapi.Component) {
```
mukundansundar, line 124:
```suggestion
func (a *apiServer) OnComponentUpdated(_ context.Context, component *componentsapi.Component) {
```

pkg/operator/api/api.go
```diff
@@ -63,72 +63,81 @@ type apiServer struct {
...
 	}
 	a.connLock.Unlock()
 }
 
+func (a *apiServer) Ready(ctx context.Context) error {
+	select {
+	case <-a.readyCh:
+		return nil
+	case <-ctx.Done():
+		return errors.New("timeout waiting for api server to be ready")
```
mukundansundar, line 137: please add UT if possible here.

JoshVanL:
Thanks @mukundansundar, please take another look

pkg/runtime/cli.go
```diff
@@ -342,8 +343,8 @@ func FromFlags() (*DaprRuntime, error) {
 	// Config and resiliency need the operator client
 	var operatorClient operatorV1.OperatorClient
 	if *mode == string(modes.KubernetesMode) {
-		log.Info("Initializing the operator client")
-		client, conn, clientErr := client.GetOperatorClient(*controlPlaneAddress, security.TLSServerName, runtimeConfig.CertChain)
+		log.Infof("Initializing the operator client (config: %s)", *config)
```
mukundansundar, line 346: why's is the whole config printed here? I think that is not needed.

JoshVanL: We removed in [this commit](https://github.com/dapr/dapr/commit/77136ebd1276d921d7dee5297ced4eab2f59e240#diff-713a1d05d1625fc206303695742169d2c70e5e5d326277965beee9b005b353adL345) on a seperate PR, and git has readded it during rebasing, or GitHub with "merge master".
Will remove.

PR #5876, opened 2023-02-01
Update to Go 1.20

ItalyPaleAle:
Also updates the pipelines to stop using deprecated Actions.

Note: this also changes the pipelines to read the version of Go to install from the go.mod file. This is useful because otherwise `/ok-to-test` would always install the version of Go defined in the pipeline.


berndverst:
Earliest date to consider merging this:
Feb 9

Need this page to list Go 1.20, then also verify the weekly GitHub Actions deployment has occurred including that change.
https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md#go

ItalyPaleAle:
Looks like Actions does support Go 1.20 now (it's [available](https://raw.githubusercontent.com/actions/go-versions/main/versions-manifest.json), just not cached so the setup step is a bit slower)

However, what I think we should wait for is this: https://github.com/Homebrew/homebrew-core/pull/122082 As without this, a lot of people won't be able to build Dapr locally :)

pkg/sentry/certs/certs.go
```diff
@@ -1,6 +1,7 @@
 package certs
 
 import (
+	"crypto"
```
mukundansundar, line 4: @ItalyPaleAle can this file change alone be moved to a separate PR? 
And only have at most tests and 1.20 change in this PR?

mukundansundar (dismissed):
lgtm.
Left one small comment.

pkg/sentry/certs/certs.go
ItalyPaleAle: Sure, done.

ItalyPaleAle:
@mukundansundar made the change you requested. Sorry for force-pushing but otherwise reverting a single file was a quite complex.

There's no code change now

berndverst (approved):

mukundansundar (approved):

Issue #5693, opened 2023-01-03
High memory consumption for Operator when there are lots of pods


ItalyPaleAle:
## In what area(s)?

/area operator

## What version of Dapr?

> 1.9.x
> master

## Actual Behavior

We are receiving user reports that the Dapr Operator service uses significant amounts of memory (700-800MB) when there are lots of pods in the K8s cluster. This happens even when there are no Dapr-enabled pods running.

Upon initial investigation, this seems to be due to the fact that the Dapr Operator implements a K8s reconciler that watches for all Deployment and StatefulSet objects, including those that don't use Dapr.

Per conversation with @yaron2, the reconciler is necessary so that Dapr can automatically provision K8s Service objects to allow app-to-app communication.

One possibility to reduce memory usage could be to make the reconciler only listen for objects with the Dapr annotation (or, if we cannot filter for annotations, we will need to add a label that is injected by the Injector). This will help reducing memory usage at least in the case when there are lots of pods without Dapr. 

## Release Note

RELEASE NOTE: **FIX** Dapr Operator uses lots of memory when there are lots of pods that aren't Dapr-enabled

akhilac1:
https://github.com/external-secrets/external-secrets/issues/721

ItalyPaleAle:
@akhilac1 are you suggesting disabling the cache? It may very well work, but I'm concerned with the impact on the K8s API server in terms of load.

@yaron2 wdyt?

akhilac1:
> @akhilac1 are you suggesting disabling the cache? It may very well work, but I'm concerned with the impact on the K8s API server in terms of load.
> 
> @yaron2 wdyt?

Oh no.. I was only suggesting that we focus on label based filtering. I missed submitting the updated comment. did so now

ItalyPaleAle:
Also worth pointing out that if the Watchdog is enabled, the Operator needs to request the full list of all Pods too. That may add to memory usage.

snoopy-dooog:
Hey, this is my first time trying to contribute to this repo, and am really not familiar with the codebase at all. However, you project sounds really cool and I'd like to start helping out. I have a WIP, which I'm not exactly sure how to test for and if it is correct, but happy to get some help from someone as I get more familiar with it :) 

I was also trying to go over how to run my own dapr in a Kube cluster to test this out better but couldn't find much guidelines on it. Looking forward to get some help on this as well. I usually develop locally so would appreciate to be able to run it locally, but if I have to go for the codespaces, will do that as well.

[...]

PR #5889, opened 2023-02-06
[WIP] Filter daprEnabled label to reduce operator's memory consumption

snoopy-dooog:
Signed-off-by: Alen Polakof <apolakof@Alens-MacBook-Pro.local>

# Description

Enable label based filtering for operator reconciler. An increase in the number of pods was producing high memory consumption since all the pods in the cluster with StatefulSets and Deployment were being watched for. This changes to only watch the pods that are part of the aforementioned set and also contain the dapr enabled label.   

NOTE: Currently, I'm watching for the daprEnabledAnnotationKey since it was already there and seems like it could do the job from reading the code? But I could also inject another label to it. 

## Issue reference 

Please reference the issue this PR will close: #5693

### DIFF
--- a/pkg/operator/handlers/dapr_handler.go
+++ b/pkg/operator/handlers/dapr_handler.go
@@ -13,8 +13,10 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/intstr"
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
 
 	"github.com/dapr/dapr/pkg/operator/monitoring"
 	"github.com/dapr/dapr/pkg/validation"
@@ -82,8 +84,17 @@ func (h *DaprHandler) Init() error {
 		return err
 	}
 
+	p, err := predicate.LabelSelectorPredicate(
+		metaV1.LabelSelector{
+			MatchLabels: map[string]string{daprEnabledAnnotationKey: "true"},
+		},
+	)
+	if err != nil {
+		return err
+	}
+
 	if err := ctrl.NewControllerManagedBy(h.mgr).
-		For(&appsv1.Deployment{}).
+		For(&appsv1.Deployment{}, builder.WithPredicates(p)).
 		Owns(&corev1.Service{}).
 		WithOptions(controller.Options{
 			MaxConcurrentReconciles: 100,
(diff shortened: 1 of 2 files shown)


daixiang0:
Good catch!

ItalyPaleAle:
As far as I know, `daprEnabledAnnotationKey` is an annotation, not a label, and K8s APIs cannot filter by annotations. The fact that [tests on K8s are failing](https://github.com/dapr/dapr/actions/runs/4099881314/jobs/7070247798) right now is an indicator that the Service is not created because K8s cannot filter by annotation.

I guess this _could_ work if you _also_ modified the Dapr injector to add a label to the Deployment in addition to an annotation? I am not sure if that's possible, but worth exploring.

CC: @yaron2

yaron2:
@snoopy-dooog the e2e tests are failing. Other than that, this is a good addition, thanks.

filintod:
you could create an annotation predicate: 
```
p := predicate.NewPredicateFuncs(func(object client.Object) bool {
		if v, ok := object.GetAnnotations()[daprEnabledAnnotationKey]; ok {
			return v == "true"
		}
		return false
	})
```

ItalyPaleAle:
@filintod although that would filter the invocations that hit the reconciler code, not sure if that would be effective at reducing memory usage, as to my understanding (and I could be wrong), the client still receives all messages and keeps the state in cache anyways.

However, worth an attempt.

filintod:
I think you are right, you would need to create a custom cache to pass to the manager (as mentioned here https://github.com/kubernetes-sigs/controller-runtime/blob/master/designs/use-selectors-at-cache.md)

The predicate above was to solve the issue of the label selector predicate using an annotation.

snoopy-dooog:
Thanks for the comments. I will take a look at this on Friday and update it accordingly!

filintod:
@snoopy-dooog I think the main problem, more than the operator deployment/sts, is from this line https://github.com/dapr/dapr/blob/master/pkg/operator/watchdog.go#L152. if as @ItalyPaleAle said we change the injector to add a label to the pod when it is patching it, then we can add an option to this get list of pods to filter the list to only retrieve those with the label set.  We might not need to have the custom cache if we never receive all the pods in the cluster.

ItalyPaleAle:
Correct! But what you change is not the operator, which doesn't patch pods, but the Dapr Injector: https://github.com/dapr/dapr/blob/master/pkg/injector/pod_patch.go

filintod:
yeah, but you still need to change the list operation in the operator

filintod:
actually was thinking on this and we could work on this before working on the injector label-addition patch change. We can use a non-cached clientset (non controller-runtime clientset) + list with pagination. We are not really watching those pods that would need a cache and this runs on a ticker period.  

Adding the injector change is still worthwhile as then this client can query just the pods that don't have the label added by the injector that will reduce the resource usage of the operation here.

ItalyPaleAle:
Moving to draft until #5937 is merged.

snoopy-dooog:
Just saw this: https://github.com/dapr/dapr/pull/5937. Will work on the filter part in the weekend! Sorry, had some busy weeks

filintod:
@snoopy-dooog I have created a [draft](https://github.com/dapr/dapr/pull/5944) that you could use/refactor, I think it is mostly ready but it does not have any tests nor I have tested it in a local environment.

PR #5892, opened 2023-02-06
Using <namespace:name> format for allowed service accounts

### LINKED ISSUE #5890: Fixing format for ALLOWED_SERVICE_ACCOUNTS
## In what area(s)?

/area runtime

## What version of Dapr?

1.9.6

## Expected Behavior

Initially introduced here: https://github.com/dapr/dapr/issues/4680
and in this PR: https://github.com/dapr/dapr/pull/4700 by @daixiang0 

The `ALLOWED_SERVICE_ACCOUNTS` variable enables you to specify a list of whitelisted service accounts that the sidecar injector can use. This list should follow the Kubernetes convention of describing service accounts in the `<namespace:service-account-name>` format. For example when using `kubectl` commands to create `clusterrolebindings`: 

```
kubectl create rolebinding myapp-view-binding --clusterrole=view --serviceaccount=acme:myapp --namespace=acme
```

The `--serviceaccount` flag uses this format `<namespace:service-account-name>`, where `acme` is the namespace and the name of the service account is `myapp`.

## Actual Behavior

The format used to add items into the `ALLOWED_SERVICE_ACCOUNTS` is `<service-account-name: namespace>`, which is confusing and not following the format used in Kubernetes tooling. 

## Release Note

RELEASE NOTE:

If you were using the `ALLOWED_SERVICE_ACCOUNTS` you should follow the format: `<namespace:service-account-name>`

salaboy:
# Description

Use `<namespace:sa-name>` to specify allowed service accounts into the sidecar injector by specifying the Environment Variable: `ALLOWED_SERVICE_ACCOUNTS`

## Issue reference

Fixes #5890 

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing

### DIFF
--- a/pkg/injector/config_test.go
+++ b/pkg/injector/config_test.go
@@ -28,7 +28,7 @@ func TestGetInjectorConfig(t *testing.T) {
 		t.Setenv("SIDECAR_IMAGE_PULL_POLICY", "Always")
 		t.Setenv("NAMESPACE", "test-namespace")
 		t.Setenv("KUBE_CLUSTER_DOMAIN", "cluster.local")
-		t.Setenv("ALLOWED_SERVICE_ACCOUNTS", "test-service-account1:test1,test-service-account2:test2")
+		t.Setenv("ALLOWED_SERVICE_ACCOUNTS", "test1:test-service-account1,test2:test-service-account2")
 
 		cfg, err := GetConfig()
 		assert.NoError(t, err)
@@ -38,7 +38,7 @@ func TestGetInjectorConfig(t *testing.T) {
 		assert.Equal(t, "Always", cfg.SidecarImagePullPolicy)
 		assert.Equal(t, "test-namespace", cfg.Namespace)
 		assert.Equal(t, "cluster.local", cfg.KubeClusterDomain)
-		assert.Equal(t, "test-service-account1:test1,test-service-account2:test2", cfg.AllowedServiceAccounts)
+		assert.Equal(t, "test1:test-service-account1,test2:test-service-account2", cfg.AllowedServiceAccounts)
 	})
 
 	t.Run("not set kube cluster domain env", func(t *testing.T) {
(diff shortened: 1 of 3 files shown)


pkg/injector/config_test.go
```diff
@@ -28,7 +28,7 @@ func TestGetInjectorConfig(t *testing.T) {
 		t.Setenv("SIDECAR_IMAGE_PULL_POLICY", "Always")
 		t.Setenv("NAMESPACE", "test-namespace")
 		t.Setenv("KUBE_CLUSTER_DOMAIN", "cluster.local")
-		t.Setenv("ALLOWED_SERVICE_ACCOUNTS", "test-service-account1:test1,test-service-account2:test2")
+		t.Setenv("ALLOWED_SERVICE_ACCOUNTS", "test1:test-service-account1,test2:test-service-account2")
```
ItalyPaleAle, line 31: Although this makes a lot of sense to me, it's also a breaking change.

Perhaps we should keep `ALLOWED_SERVICE_ACCOUNTS` behaving as it does today, mark it as deprecated (using it would show a warning log, but would still work), and introduce a new env var (for example `ALLOWED_SA` ? not sure about the name) that uses the new format

ItalyPaleAle (changes requested):

yaron2:
In other circumstances this might be considered a breaking change, but we can allow this into the 1.10 release because:

1. This environment variable was never documented nor configurable via our Helm chart or CLI
2. 1.10 is the first version that includes the Helm chart setting that exposes the env var
3. Considering #2, it makes sense to merge this into 1.10 as after it's documented it will become a true breaking change.

pkg/injector/config_test.go
yaron2: https://github.com/dapr/dapr/pull/5892#issuecomment-1419379947.

yaron2 (approved):

PR #5896, opened 2023-02-06
Add option to Operator to disable the reconciler for Services

ItalyPaleAle:
# Description

Adds a new CLI flag to the Dapr Operator service `--disable-service-reconciler` to disable the built-in reconciler (default=false - i.e. the reconciler is enabled)

The Operator has multiple roles in Dapr, and one of them is having a reconciler that creates Services for each Dapr-enabled Deployment and StatefulSet. In some cases, this reconciler may conflict with other reconcilers that are running on the cluster. Additionally, it uses a lot of memory, even in clusters that do not need it.

Also added an option to the Helm chart to control this flag.

### DIFF
--- a/charts/dapr/README.md
+++ b/charts/dapr/README.md
@@ -126,6 +126,7 @@ The Helm chart has the follow configuration options that can be supplied:
 | `dapr_operator.runAsNonRoot`              | Boolean value for `securityContext.runAsNonRoot`. You may have to set this to `false` when running in Minikube | `true` |
 | `dapr_operator.resources`                 | Value of `resources` attribute. Can be used to set memory/cpu resources/limits. See the section "Resource configuration" above. Defaults to empty | `{}` |
 | `dapr_operator.debug.enabled`             | Boolean value for enabling debug mode | `{}` |
+| `dapr_operator.serviceReconciler`         | Boolean value for enabling the reconciler that creates Services for Dapr-enabled Deployments and StatefulSets | `true` |
 | `dapr_operator.watchNamespace`            | The namespace to watch for annoated Dapr resources in | `""` |
 
 ### Dapr Placement options:
--- a/charts/dapr/README.md
+++ b/charts/dapr/README.md
@@ -133,7 +133,7 @@ The Helm chart has the follow configuration options that can be supplied:
 | `dapr_operator.runAsNonRoot`              | Boolean value for `securityContext.runAsNonRoot`. You may have to set this to `false` when running in Minikube | `true` |
 | `dapr_operator.resources`                 | Value of `resources` attribute. Can be used to set memory/cpu resources/limits. See the section "Resource configuration" above. Defaults to empty | `{}` |
 | `dapr_operator.debug.enabled`             | Boolean value for enabling debug mode | `{}` |
-| `dapr_operator.serviceReconciler.disabled`| Boolean value for disabling the reconciler that creates Services for Dapr-enabled Deployments and StatefulSets.<br>Note: disabling this reconciler could prevent Dapr service invocation from working. | `false` |
+| `dapr_operator.serviceReconciler.enabled`| If false, disables the reconciler that creates Services for Dapr-enabled Deployments and StatefulSets.<br>Note: disabling this reconciler could prevent Dapr service invocation from working. | `true` |
 | `dapr_operator.watchNamespace`            | The namespace to watch for annoated Dapr resources in | `""` |
 
 ### Dapr Placement options:
--- a/charts/dapr/README.md
+++ b/charts/dapr/README.md
@@ -126,7 +126,7 @@ The Helm chart has the follow configuration options that can be supplied:
 | `dapr_operator.runAsNonRoot`              | Boolean value for `securityContext.runAsNonRoot`. You may have to set this to `false` when running in Minikube | `true` |
 | `dapr_operator.resources`                 | Value of `resources` attribute. Can be used to set memory/cpu resources/limits. See the section "Resource configuration" above. Defaults to empty | `{}` |
 | `dapr_operator.debug.enabled`             | Boolean value for enabling debug mode | `{}` |
-| `dapr_operator.serviceReconciler`         | Boolean value for enabling the reconciler that creates Services for Dapr-enabled Deployments and StatefulSets.<br>Note: disabling this could prevent Dapr service invocation from working. | `true` |
+| `dapr_operator.serviceReconciler.disabled`| Boolean value for disabling the reconciler that creates Services for Dapr-enabled Deployments and StatefulSets.<br>Note: disabling this reconciler could prevent Dapr service invocation from working. | `false` |
 | `dapr_operator.watchNamespace`            | The namespace to watch for annoated Dapr resources in | `""` |
 
 ### Dapr Placement options:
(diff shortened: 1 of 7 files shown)


pkg/operator/handlers/dapr_handler.go
```diff
@@ -16,17 +15,14 @@ import (
...
 
+	"github.com/dapr/dapr/pkg/injector/annotations"
 	"github.com/dapr/dapr/pkg/operator/monitoring"
 	"github.com/dapr/dapr/pkg/validation"
 	"github.com/dapr/dapr/utils"
 	"github.com/dapr/kit/logger"
 )
 
 const (
-	daprEnabledAnnotationKey        = "dapr.io/enabled"
```
ItalyPaleAle, line 26: Reviewers: these constants were already defined in pkg/injector/annotations so importing them from there instead

yaron2:
It should be mentioned that enabling this will break/disable service to service invocation.

ItalyPaleAle:
@yaron2 Updated the README to add the warning, thanks.

pkg/operator/handlers/dapr_handler.go
```diff
@@ -212,23 +226,23 @@ func (h *DaprHandler) createDaprServiceValues(ctx context.Context, expectedServi
...
 
 	if enableMetrics {
-		annotations["prometheus.io/probe"] = "true"
-		annotations["prometheus.io/scrape"] = "true" // WARN: deprecated as of v1.7 please use prometheus.io/probe instead.
-		annotations["prometheus.io/port"] = strconv.Itoa(metricsPort)
-		annotations["prometheus.io/path"] = "/"
+		annotationsMap["prometheus.io/probe"] = "true"
+		annotationsMap["prometheus.io/scrape"] = "true" // WARN: deprecated as of v1.7 please use prometheus.io/probe instead.
+		annotationsMap["prometheus.io/port"] = strconv.Itoa(metricsPort)
+		annotationsMap["prometheus.io/path"] = "/"
```
mukundansundar, line 237: can these keys be defined as constants?

pkg/operator/handlers/dapr_handler.go
```diff
@@ -264,37 +278,32 @@ func (h *DaprHandler) createDaprServiceValues(ctx context.Context, expectedServi
...
+	return annotationsMap[annotations.KeyAppID]
 }
 
 func (h *DaprHandler) isAnnotatedForDapr(wrapper ObjectWrapper) bool {
-	annotations := wrapper.GetTemplateAnnotations()
-	enabled, ok := annotations[daprEnabledAnnotationKey]
-	if !ok {
+	annotationsMap := wrapper.GetTemplateAnnotations()
+	val := annotationsMap[annotations.KeyEnabled]
+	if val == "" {
```
mukundansundar, line 288: why was this changed to not use ok?

pkg/operator/handlers/dapr_handler.go
ItalyPaleAle: Done

pkg/operator/handlers/dapr_handler.go
ItalyPaleAle: Good question.

First, context: `val, ok := annotationsMap[annotations.KeyEnabled]`:

- If the key does not exist on the map, `val == ""`
- You can have `val == ""` also if the key exists, but it's empty. That's where the `ok` is useful: in this case, `ok == true`

Now let's take a look at the behavior of the previous code vs current:

**Previous**
```go
enabled, ok := annotations[daprEnabledAnnotationKey]
if !ok {
	return false
}
return utils.IsTruthy(enabled)
```

**Current**

```go
val := annotationsMap[annotations.KeyEnabled]
if val == "" {
	return false
}
return utils.IsTruthy(val)
```

As you can see, in both cases the if the value doesn't exist, the returned value is `val == ""` so the function returns false. The additional check for `ok` was redundant.

charts/dapr/README.md
```diff
@@ -126,6 +126,7 @@ The Helm chart has the follow configuration options that can be supplied:
 | `dapr_operator.runAsNonRoot`              | Boolean value for `securityContext.runAsNonRoot`. You may have to set this to `false` when running in Minikube | `true` |
 | `dapr_operator.resources`                 | Value of `resources` attribute. Can be used to set memory/cpu resources/limits. See the section "Resource configuration" above. Defaults to empty | `{}` |
 | `dapr_operator.debug.enabled`             | Boolean value for enabling debug mode | `{}` |
+| `dapr_operator.serviceReconciler`         | Boolean value for enabling the reconciler that creates Services for Dapr-enabled Deployments and StatefulSets.<br>Note: disabling this could prevent Dapr service invocation from working. | `true` |
```
mukundansundar, line 129: should this key be `dapr_operator.serviceReconciler.enabled` ?
`dapr_operator.runAsNonRoot` implicitly means that it is a bool value. But the above does not.
Having enabled or disabled would be better here IMO...

charts/dapr/charts/dapr_operator/values.yaml
```diff
@@ -16,6 +16,8 @@ fullnameOverride: ""
 
 runAsNonRoot: true
 
+serviceReconciler: true
```
mukundansundar, line 19: Would prefer to have a disable flag here instead of an enable flag.
Because the default behavior is enabled. So providing a flag to disable it makes it more consistent.

cmd/operator/main.go
```diff
@@ -114,8 +115,9 @@ func init() {
 
 	flag.StringVar(&watchInterval, "watch-interval", defaultWatchInterval, "Interval for polling pods' state, e.g. '2m'. Set to '0' to disable, or 'once' to only run once when the operator starts")
 	flag.IntVar(&maxPodRestartsPerMinute, "max-pod-restarts-per-minute", defaultMaxPodRestartsPerMinute, "Maximum number of pods in an invalid state that can be restarted per minute")
-	flag.BoolVar(&disableLeaderElection, "disable-leader-election", false, "Disable leader election for operator")
 
+	flag.BoolVar(&disableLeaderElection, "disable-leader-election", false, "Disable leader election for operator")
+	flag.BoolVar(&disableServiceReconciler, "disable-service-reconciler", false, "Disable the Service reconciler for Dapr-enabled Deployments and StatefulSets")
```
mukundansundar, line 120: nitpick: if grouping can group Bool vars together.

pkg/operator/handlers/dapr_handler.go
```diff
@@ -78,11 +79,13 @@ func (h *DaprHandler) Init() error {
 				return nil
 			}
 			return []string{owner.Name}
-		}); err != nil {
+		},
+	)
+	if err != nil {
```
mukundansundar, line 84: why is the short if changed to this format?

cmd/operator/main.go
```diff
@@ -66,6 +66,7 @@ func main() {
 		WatchdogInterval:          0,
 		WatchdogMaxRestartsPerMin: maxPodRestartsPerMinute,
 		WatchNamespace:            watchNamespace,
+		ServiceReconcilerEnabled:  !disableServiceReconciler,
```
mukundansundar, line 69: is it possible to add a unit test or e2e test for this?

pkg/operator/handlers/dapr_handler.go
ItalyPaleAle: IMHO a function inside an if was unreadable :)

cmd/operator/main.go
ItalyPaleAle: It is not, actually.... It would require creating an ad-hoc Dapr cluster so it doesn't fit with our E2E test framework.

charts/dapr/README.md
ItalyPaleAle: Done (using `disabled` as per below)

charts/dapr/charts/dapr_operator/values.yaml
ItalyPaleAle: Done

cmd/operator/main.go
ItalyPaleAle: They're currently grouped by kind. These 3 are all about configuring the reconciler

mukundansundar:
@ItalyPaleAle Can you change back to having the `enabled` config option as per helm [convention](https://helm.sh/docs/chart_best_practices/dependencies/ )

ItalyPaleAle:
@mukundansundar done

mukundansundar (approved):
lgtm

PR #5897, opened 2023-02-06
Reduce log verbosity for workflow execution

cgillum:
# Description

Currently there are three Info logs for each workflow activity execution, which is far too verbose since workflows can create lots of activity executions. This PR reduces some of those info logs to debug to avoid spamming Dapr Workflow users.

Note that this impacts existing logs from the Dapr runtime as well, and there's an assumption that these are safe to downgrade to `Debug`.

## Issue reference

This came up as part of a discussion on Discord, but I can open an issue if necessary.

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests
* [x] Unit tests passing


pkg/actors/actors.go
```diff
@@ -986,7 +986,7 @@ func (a *actorsRuntime) startReminder(reminder *Reminder, stopChannel chan bool)
 			if err = a.executeReminder(reminder); err != nil {
 				if errors.Is(err, ErrReminderCanceled) {
 					// The handler is explicitly canceling the timer
-					log.Infof("reminder %q was canceled by the actor", reminder.Name)
+					log.Debugf("reminder %q was canceled by the actor", reminder.Name)
```
cgillum, line 989: This is a new log introduced in v1.10, so there's no change of behavior here.

pkg/actors/actors.go
```diff
@@ -1796,7 +1796,7 @@ func (a *actorsRuntime) doDeleteReminder(ctx context.Context, req *DeleteReminde
 
 	stop, exists := a.activeReminders.Load(reminderKey)
 	if exists {
-		log.Infof("Found reminder with key: %v. Deleting reminder", reminderKey)
+		log.Debugf("Found reminder with key: %v. Deleting reminder", reminderKey)
```
cgillum, line 1799: This Info log predates my involvement in the project (and predates other people I've asked). I don't really understand why it was Info, but it feels like a good candidate for Debug given that it's written every time a reminder expires or gets deleted.

pkg/runtime/wfengine/activity.go
```diff
@@ -235,7 +235,7 @@ func (a *activityActor) loadActivityState(ctx context.Context, actorID string, g
 	}
 
 	// Loading from the state store is only expected in process failure recovery scenarios.
-	wfLogger.Infof("%s: loading activity state", actorID)
+	wfLogger.Debugf("%s: loading activity state", actorID)
```
cgillum, line 238: This one we need to investigate further. I was not expecting we'd ever see this log in the normal execution path (as hopefully can be understood from the comment above). I've taken a note to follow up on this to see if there is a performance issue that needs to be made here.

cgillum (commented):
Added some contextual inline comments.

pkg/actors/actors.go
ItalyPaleAle: Can the key contain any PII? Or is it always a random identifier?

pkg/runtime/wfengine/activity.go
ItalyPaleAle: It may be a leftover log that was used to debug this while programming. If you don't think it matters, feel free to remove them.

Our logs are not part of any API contract anyways, with very few exceptions (like API logs for the HTTP/gRPC servers)

pkg/actors/actors.go
cgillum: For normal actors, the reminder names are controlled by the developer and can be anything they want. For internal actors used by workflows, we control the reminder names and there is no risk of PII.

ItalyPaleAle: > For normal actors, the reminder names are controlled by the developer and can be anything they want.

Then consider not logging the name as it may contain PII.

cgillum: Logging reminder names is unrelated to Dapr Workflow. Logging of reminder names has been in the Dapr codebase since before I got involved. This is also not the only place we log reminder names. Can we make this bigger change as part of a separate PR? This PR is simply about reducing distracting log noise.

ItalyPaleAle (approved):

artursouza (approved):

PR #5898, opened 2023-02-07
dynamic allowed service accounts watcher for injector

filintod:
# Description

Added capability to dynamically add allowed UIDs by watching against namespace and/or service account name prefixes.   When a service account is created or deleted we add/delete the ID of the service account in the injector authUIDs list.  The watch can also be set against a string label selector.  

A user/operator can provide the extra value `allowedServiceAccountsPrefixNames` to the sidecar chart definition with star sign at the end to say whether the service account name or the namespace or both are prefixes. We separate each `namespace:serviceaccountname` tuple with commas (`,`). The predicates will be [ORed](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/predicate#Or).

### Service Account Name or Namespace Prefix watch examples

- Service Account name prefix with fixed namespace
  Watch for service account names prefix with `name1-x` or `name2-x` in namespace1, or for `name1-x` in namepsace-2.
  ```
  allowedServiceAccountsPrefixNames: "namespace1:name1-x*,namespace1:name2-x,namespace2:name1-x*"
  ```
- Namespace name prefix with fixed service account name
  Watch for service account names prefix with `name-x` in namespace1
  ```
  allowedServiceAccountsPrefixNames: "vc-team-*:name"
  ```
- Namespace name prefix and Service Account name prefix
    ```
  allowedServiceAccountsPrefixNames: "vc-team-*:team-*"
  ```

## Issue reference

Please reference the issue this PR will close: #5906 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Unit tests passing


yaron2 (changes requested):

yaron2:
From the linked issue:

> This will open the possibility of using dapr with vcluster.

I want to make it clear what scenario we're enabling here, as Dapr can in fact run by default inside of VClusters.
Does this enable injecting VCluster data plane sidecars from a control plane that sits outside of a VCluster?

salaboy:
@yaron2 yes this enables VCluster Clusters (Vcluster 0.14.0) to consume a single Dapr installation that is in the Host cluster. 
Each VCluster created will create a new service account, which needs to be added to the allowed list manually. 
This PR by @filintod  is trying to a way to bypass that manual intervention by using a convention and wildcards

salaboy:
@yaron2 I wonder.. can we make this into 1.10? or is it too late?

filintod:
looks like there was a recent commit where name:namespace was swap (to the more k8s common) namespace:name. I guess I should do the same. So expect another change

yaron2:
> @yaron2 I wonder.. can we make this into 1.10? or is it too late?

This will need to go in 1.11.

pkg/injector/allowedsawatcher/namenamespacepredicates.go
```diff
@@ -0,0 +1,60 @@
+package allowedsawatcher
```
ItalyPaleAle, line 1: Here too please

pkg/injector/allowedsawatcher/predutil_test.go
```diff
@@ -0,0 +1,144 @@
+package allowedsawatcher
```
ItalyPaleAle, line 1: Ditto

pkg/injector/interfaces/interfaces.go
```diff
@@ -0,0 +1,9 @@
+package interfaces
```
ItalyPaleAle, line 1: Ditto

utils/utils.go
```diff
@@ -164,6 +174,27 @@ func Filter[T any](items []T, test func(item T) bool) []T {
...
+	l := make([]T, len(m))
+	var i int
+	for uid := range m {
+		l[i] = uid
+		i++
+	}
+	return l
+}
+
+// SliceToMap simple utility to convert from a slice to a map generally to use for contains-like operations
```
ItalyPaleAle, line 188:
```suggestion
// SliceToMap converts from a slice to a map generally to use for contains-like operations
```

(It's best to avoid words like "simple" as they're [subjective](https://jameshfisher.com/2017/02/22/dont-use-simply/))

pkg/injector/injector.go
```diff
@@ -234,7 +249,12 @@ func (i *injector) handleRequest(w http.ResponseWriter, r *http.Request) {
 	if err != nil {
 		log.Errorf("Can't decode body: %v", err)
 	} else {
-		if !(utils.Contains(i.authUIDs, ar.Request.UserInfo.UID) || utils.Contains(ar.Request.UserInfo.Groups, systemGroup)) {
+		// get a snapshot of the authuids
+		i.authUIDsMu.RLock()
+		authUIDs := i.authUIDs
+		i.authUIDsMu.RUnlock()
+
+		if !(utils.Contains(authUIDs, ar.Request.UserInfo.UID) || utils.Contains(ar.Request.UserInfo.Groups, systemGroup)) {
```
ItalyPaleAle, line 257: My biggest question on the entire PR is here... Does this need a watcher that is constantly notified by the K8s API server if a new SA is available?

Couldn't we just lazily check if the SA exists when we try to use it here?

Yes, this will add a small latency to each operation, but on the other hand doesn't require creating a watcher which takes up a lot more memory.

ItalyPaleAle (changes requested):

pkg/injector/allowedsawatcher/predutil.go
```diff
@@ -0,0 +1,108 @@
...
+
+	"github.com/dapr/dapr/utils"
+)
+
+type equalPrefixLists struct {
+	equal  []string
+	prefix []string
+}
+
+const minPrefixLength = 4
```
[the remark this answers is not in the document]
filintod: at one moment was to avoid possibly hitting system namespaces but that was added already. I guess there is no need, as it could also be that a user wants all SAs matching a prefix on any namespace (`*:sa*`), os all SAs matching with a namespace matching a prefix (`ns*:*`). I'll take care. thanks

pkg/injector/injector.go
filintod: true, depending on the dynamic behavior of the creation/destruction of the virtual clusters, so it is kind of a balancing act, if it is very dynamic it makes sense but if not then not so much. Most of the code here could be repurposed for the evaluation, we would need a garbage collector to prune stale cached sa-ns entries (like a watchdog) if we want to keep a cache to avoid the sa query delay on each pod created for the same vcluster, and only hit the delay the first time (the others it would be cached). let me look into it and check with my peer.

I'll resolve the other comments and get back to this one after. thanks

ItalyPaleAle: I think a cache may be dangerous because we may be deploying pods with stale information.

IMHO the question here is whether we prioritize latency (removing one API call for each pod being deployed) vs memory usage. I'd personally lean on the second: we are seeing now with the Dapr Operator, for example, how watching for resources can cause significant memory issues on large clusters.

CC: @yaron2

filintod: I'm leaning on adding another option to default to extra latency, or opting for no latency but more memory usage, SAs are not as pervasive as pods but I understand the issue. The predicate methods can be reused by both after a refactor.

filintod: just realized that we might have an easy way out. service accounts userinfo should be in the format `system:serviceaccount:(NAMESPACE):(SERVICEACCOUNT)` per https://kubernetes.io/docs/reference/access-authn-authz/authentication/#service-account-tokens and that is the common one. So if we trust the api server that we do, we can just accept this value as a fact and don't query the api server and just try to match against this value, no need for cache nor query.

ItalyPaleAle: That sounds reasonable to me.

pkg/injector/namespacednamematcher/namenamespacematcher.go
```diff
@@ -0,0 +1,138 @@
...
+
+type EqualPrefixNameNamespaceMatcher struct {
+	prefixed map[string]*equalPrefixLists
+	equal    map[string]*equalPrefixLists
+}
+
+// CreateFromString from the csv provided by the user of sa:ns values, we create two maps
+// one with namespace prefixes and one with namespace exact values
+// inside each map we can have exact name or prefixed names
+// note there might be prefixes that cover other prefixes, but we are not filtering it for now
```
mukundansundar, line 46:
```suggestion
// CreateFromString creates two maps from the CSV provided by the user of ns:sa values,
// one with namespace prefixes and one with namespace exact values.
// Inside each map we can have exact name or prefixed names.
// Note there might be overlap in prefixes, but we are not filtering it for now.
```

pkg/injector/namespacednamematcher/namenamespacematcher.go
```diff
@@ -0,0 +1,138 @@
...
+	for nsPrefix, values := range m.prefixed {
+		if strings.HasPrefix(namespace, nsPrefix) {
+			return utils.ContainsPrefixed(values.prefix, name) || utils.Contains(values.equal, name)
+		}
+	}
+	for ns, values := range m.equal {
+		if ns == namespace {
+			return utils.ContainsPrefixed(values.prefix, name) || utils.Contains(values.equal, name)
+		}
+	}
```
mukundansundar, line 136: shouldn't equality have more precedence than prefix match?

pkg/injector/injector.go
salaboy: For the use case that this was originaly intended (vclusters reusing a single dapr installation) I don't think that it needs to be that dynamic. Creating a watcher for a defined pattern like `vcluster-*` would make total sense to me, instead of adding a check per request. But I might be missing some context about how Dapr work at that level.

pkg/injector/namespacednamematcher/namenamespacematcher.go
filintod: it is not important in this case, it would be more important if we were routing by longest prefix match, we are just allowing it or not, and having it this way we can capture earlier any overlapping (prefix covering other prefixes covering equal*) faster.

pkg/injector/injector.go
```diff
@@ -117,8 +121,25 @@ func NewInjector(authUIDs []string, config Config, daprClient scheme.Interface,
...
+	if err := createNamespaceNameMatcher(i, config); err != nil {
+		return nil, err
+	}
+
 	mux.HandleFunc("/mutate", i.handleRequest)
-	return i
+	return i, nil
+}
+
+func createNamespaceNameMatcher(i *injector, config Config) error {
```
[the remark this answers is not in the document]
filintod: make sense, will refactor

charts/dapr/README.md
```diff
@@ -172,24 +172,25 @@ The Helm chart has the follow configuration options that can be supplied:
...
+| `dapr_sidecar_injector.sidecarReadOnlyRootFilesystem` | When this boolean value is true (the default), the injected sidecar containers have `readOnlyRootFilesystem: true`                                                                                                                                                              | `true` |
+| `dapr_sidecar_injector.allowedServiceAccounts` | String value for extra allowed service accounts in the format of `serviceAccount1:namespace1,serviceAccount2:namespace2`                                                                                                                                                        | `""` |
+| `dapr_sidecar_injector.allowedServiceAccountsWatchNames` | String value for extra allowed service accounts to watch, even if not yet present, and with the possibility of using prefixes by suffixing the name or namespace prefix with a `*` sign, in the format of `namespacePrefix1*:serviceAccountPrefix1*,namespace2:serviceAccount2` | `""` |
```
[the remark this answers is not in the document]
ItalyPaleAle: Please take another look at this message, as it's fairly hard to understand. "using prefixes by suffixing the name" is quite the mouthful :)

pkg/injector/injector.go
```diff
@@ -117,8 +121,25 @@ func NewInjector(authUIDs []string, config Config, daprClient scheme.Interface,
...
+	}
+	i.namespaceNameMatcher = matcher
+
 	mux.HandleFunc("/mutate", i.handleRequest)
-	return i
+	return i, nil
+}
+
+func createNamespaceNameMatcher(allowedPrefix string) (matcher *namespacednamematcher.EqualPrefixNameNamespaceMatcher, err error) {
+	if prefixNames := allowedPrefix; prefixNames != "" {
```
ItalyPaleAle, line 135: Could this be just:

```suggestion
	if allowedPrefix != "" {
```

pkg/injector/injector_test.go
```diff
@@ -23,6 +23,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/dapr/dapr/pkg/injector/namespacednamematcher"
```
ItalyPaleAle, line 26: Please put this in the same group as the other `github.com/dapr*` dependencies. The linter will complain

ItalyPaleAle (changes requested):

pkg/injector/injector.go
filintod: sure, it must have done something before a refactor and was left dangling. thanks

filintod: done

charts/dapr/README.md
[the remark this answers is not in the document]
filintod: I did a change with your feedback, not sure if too long now. let me know

ItalyPaleAle:
@filintod Please don't rebase your PRs as that breaks our ability to use things like "see changes since your last review", and we need to review the entire PR from scratch every time!

ItalyPaleAle (dismissed):

ItalyPaleAle (approved):

PR #5908, opened 2023-02-08
Allow using --resources-path in Kubernetes mode

### LINKED ISSUE #5585: App version-specific component config files
If I understand the flow of things, the component name used by a Dapr-enabled application (in it's calls to the Dapr SDK or in the [URL used in an HTTP request](https://docs.dapr.io/reference/api/state_api/#http-request) to the Dapr sidecar) needs to match the name used in the [component file](https://docs.dapr.io/reference/api/state_api/#component-file). From looking at the samples, while it appears that some Dapr metadata (like the Dapr host and port) are injected into the app as env vars (e.g. `DAPR_HTTP_PORT`), the component name is not. This means that the component name is probably hardcoded into the app (or read from some app specific config info) and therefore fairly static.

There are cases where the static nature of the component name may be problematic and I'd like to use this issue to brainstorm possible solutions or to find out if the scenario is just nuts. :-)

Here's the scenarios:
- a component file is meant to be used by just one app - so no need to worry about sharing it between apps
- the hosting environment has the notion of versioning of the application - e.g. Knative, Container Apps, Cloud Run, Code Engine. So, as the dev makes config changes to the app a new version is created and traffic is slowly migrated from the old version to the new one. Each version's set of config options are totally independent from other version's.
- the use of Dapr is "version specific". Meaning, whether any particular version of the app is using Dapr or not, which components it is using, and the various config options used to configure how to talk with the backing service of the component, are all viewed as config options for that version of the application.
- this means that in the most extreme case we can think of the Dapr component files as being versioned at the same time as the application (ie. each version of the app has it's own version of the complete set of Dapr component files).
- This give the app owner the flexibility to roll-out changes to the Dapr component files just like any other configuration change. And they can do things like canary testing of these Dapr config changes to make sure everything works as expected.

From an abstract perspective this seems reasonable (to me anyway), however, in order for each version of the app to have it's own copy of the Dapr component files, each one needs to have a unique metadata.name value - and that "name" needs to be used by the application. Which means it can't be "static" as mentioned in the first paragraph.

I'm wondering if there's a way to solve this situation given the current state of Dapr? Is there a way for the app to "discover" the name? I'm guessing no.

Is there a way to enhance Dapr to support this type of dynamic naming or app-version specific Dapr config?

Some ideas:
- allow for extra metadata to be associated with each scopes.app in the component config file and then require not just the app's id to match the scope value but also some other info (e.g. an env var in the dapr sidecar) to also match the extra metadata. Then the scoping can be bound not just to a set of apps but versions of each app. Along with this we'd also need the ability to have an alias (or name mapping) so that the config file's metadata.name value can be different from the "name" used by the app.
- allow for the config file to be provided to the dapr sidecar directly and not stored in K8s - thus avoiding the metadata.name conflict issue. For example, make the component config files available as env vars, or perhaps from some dir in the dapr sidecar then people can mount the config files dynamically.

I kind of like the 2nd one since it just changes where we read the component config files from - no other magic is needed. And env vars are pretty easy to set up.

Anyway, any thoughts on this situation?

ItalyPaleAle:
# Description

**TLDR: fixes #5585** (Can't fully fit the scope of this PR in the title :) )

> In draft until #5909 is merged (included in this PR already) and pending further discussion

This PR includes a changes that are interrelated and have three business goals:

- Even in K8s mode, allow loading components, resiliency policies, and Dapr configurations from files (e.g. mounted via a ConfigMap) to allow tighter scoping with the with apps, including specific versions of an app (main ask of #5585)
- ~~Allows passing `--resources-path` multiple times to load from multiple folders, both in K8s and in self-hosted (secondary ask of #5585)~~  
  This has been spun out into its own PR #5909 (which is included in this PR)
- When the app is launched using `--resources-path` and not referencing any Configuration stored in K8s, the runtime dependency on a connection with the Dapr Operator is now removed (corollary of the issue above)

The relevant changes are:

- The daprd CLI can now accept multiple `--resources-path` flags. All files (Components, Resiliency policies, declarative pubsub) from all paths are loaded by Dapr.
- `--resources-path` can be specified in K8s mode too, and causes Dapr to load the resources (Components, Resiliency policies, declarative pubsub) from the local disk even in K8s mode (this generally requires mounting a ConfigMap within the container, as I'm doing in the E2E tests)
- The `--config` CLI flag has been deprecated in daprd.
   - Use `--config-file` to load a file from disk, on both self-hosted and K8s. In K8s mode, this caused daprd to load from file (e.g. a ConfigMap mounted in the daprd container) rather than requesting from the Operator
   - Use `--config-name` to load a named Configuration resource from K8s, through the Operator (this is a no-op in self-hosted mode)
   - The deprecated `--config` flag continues to work with the previous behavior: it corresponds to `--config-file` in self-hosted and `--config-name` in K8s (i.e. with `--config` you cannot load a local file in K8s mode)
- In K8s mode, if `--resources-path` is used and no `--config-name` (or the legacy `--config`) is passed, Dapr will not establish a connection with the Operator at runtime
- New annotations have been added for the above:
   - `dapr.io/config-name` (replaces `dapr.io/config`, which continues to work as fallback)
   - `dapr.io/config-file`
   - `dapr.io/resources-path`

Core E2E tests have been implemented by modifying the existing "middleware" E2E test, which was self-contained and was loading both a Configuration and Components. I need to work on a few more E2E tests to cover all scenarios.

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests
* [X] Unit tests passing
* [ ] Add E2E tests:
   * [X] Load Configurations from file - added to the "middleware" E2E test
   * [X] Load Components from file - added to the "middleware" E2E test
   * [ ] Load Resiliency policies from file
   * [ ] Load declarative pubsub from file
   * [ ] Load Components from file that require secrets from the K8s secret store
* [ ] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: TODO @ItalyPaleAle


yaron2:
I'm inclined to reject this PR due to #5585 not being accepted or triaged, and I have many concerns with this approach, namely the K8s anti-pattern that this would be introducing. As we want to remove backlog from our Pull Requests (we have more than 30 that need to be looked at, and clogging it has already caused some to be left behind) I ask that you close the PR until, if at all, we triage it.

yaron2:
Record for decision: https://github.com/dapr/dapr/issues/5585#issuecomment-1421902976.

artursouza:
I think it is worth the discussion of how to support component isolation per app instance, not just app-id.

yaron2:
> I think it is worth the discussion of how to support component isolation per app instance, not just app-id.

Indeed. there's a lot of discussion that needs to happen first on #5585. The PR is simply very premature at this point.

ItalyPaleAle:
Moved this PR to draft stage pending discussion on Monday.

Also rebased to be on top of #5909 which will be reviewed separately

ItalyPaleAle:
Closing as per discussion and we'll be looking at different approaches

PR #5909, opened 2023-02-08
Allow passing multiple `--resources-path` flags to daprd

ItalyPaleAle:
This is a spin-off from #5908 which takes one of the two parts of the PR (the largest one) as described in #5585

It allows starting daprd with multiple `--resources-path` flags, to specify multiple folders to load resources (components, resiliency policies, declarative subscriptions) from.

This allows, for example, maintaining components and resiliency policies in different folders, or maintain "shared" components in a separate folder from app-specific ones. More details can be found in #5585

It also includes some small tweaks and fixes encountered along the way.

## TODO

- [X] Dapr CLI issue: dapr/cli#1216
- [X] Docs issue: dapr/docs#3151


artursouza (changes requested):
Have not looked at the code yet, but I do miss E2E test for this. Is it possible with the current K8s centric approach? Is there a way to "manually" add the sidecar container in the test without relying on injector?

ItalyPaleAle:
Because this only works in self-hosted mode, we cannot have E2E tests for this (same reason as every other time, sadly). It is also not meant to be run on K8s (that's what #5908 was for). Perhaps they could be created in the CLI repo when the CLI adds supports for this.

yaron2:
> Have not looked at the code yet, but I do miss E2E test for this. Is it possible with the current K8s centric approach? Is there a way to "manually" add the sidecar container in the test without relying on injector?

We don't have E2E for self-hosted mode today, although technically it's possible add them by configuring a self-hosted sidecar on a Kubernetes deployment.

> it is also not meant to be run on K8s (that's what https://github.com/dapr/dapr/pull/5908 was for).

There are two separate issues here being mingled into one:

1. #5908 tried to make a self-hosted feature apply to k8s mode. this is wrong and shouldn't be allowed. It included additional features that have been moved into this PR, which is great and unblocks this functionality from being added

2. The changes in this PR are targeting self hosted mode, which we don't have E2E tests for (but would be great to design for those in the future as self-hosted is also a supported production mode of Dapr)

ItalyPaleAle:
> although technically it's possible add them by configuring a self-hosted sidecar on a Kubernetes deployment.

Albeit possible, this is not easy at all given our K8s test framework today. The test runner doesn't have code in place to deploy a daprd container (converting annotations into CLI flags etc), and some things would require large efforts such as being able to retrieve the TLS certificates to inject them into the Deployment.

I agree with you both there's a need for self-hosted E2E tests, but I kindly ask that be considered outside of the scope of this PR given the large effort.

In the meanwhile, I think this PR can be "E2E tested" in the Dapr CLI with relative ease.

yaron2:
> I agree with you both there's a need for self-hosted E2E tests, but I kindly ask that be considered outside of the scope of this PR given the large effort.

For sure, as we don't have an infrastructure for self-hosted E2E test, I don't expect this PR to be any different than any other self-hosted PR in terms of expectations.

pkg/components/disk_manifest_loader.go
```diff
@@ -118,7 +131,7 @@ func (m DiskManifestLoader[T]) decodeYaml(b []byte) ([]T, []error) {
 			continue
 		}
 
-		manifest := m.zv()
+		var manifest T
```
mcandeia, line 134: you must be thinking: why the hell did someone complicated a zero value instantiation in that way?

Not saying that you did the wrong thing, but let me tell the why: when I created this `DiskManifestLiader` I tried to kept the intention of the user that created the `newComponent` method, allowing folks to set default fields on the component "zero" value. Obviously, that given the current state of the `newComponent` method it can be completely replaced by what you did, but keep in mind that you have removed the possibility of setting default fields for whatever is unmarshalled

> There's workarounds to do the same like implementing a custom unmarshaller, however I'd prefer the old way :)

duglin: If we want to keep the old way of doing it then we should add a really wordy comment around it so people don't change it w/o really thinking hard about the implications. You won't always be around to add this commentary to every PR forever :-)

ItalyPaleAle: @mcandeia Thanks for explaining. I have re-added the ability to set a function that returns a zero value. In a slightly different way, to not alter the signature of `NewDiskManifestLoader` :)

artursouza (approved):

PR #5914, opened 2023-02-09
Return a "stable" port for internal gRPC server if unset (for better local DX)

ItalyPaleAle:
Fixes dapr/components-contrib#2413

# Description

When Dapr is running as self-hosted, the internal gRPC server (used for service invocation) is started on a random port every time, unless `--dapr-internal-grpc-port` is set explicitly. (In K8s mode, the injector adds that flag with a stable port automatically).

During local development, this can cause issues such as dapr/components-contrib#2413 and dapr/components-contrib#2489 . For example:

- App A calls App B. Both use mDNS (or Consul) as nameresolver. They do not have an explicit `--dapr-internal-grpc-port`
- App A invokes App B, which works.
- App B is restarted. When it comes back, the internal gRPC server listens on a different port.
- App A invokes App B. It looks up the address and port from the nameresolver, which uses the cache
- If the cache hasn't expired, App A receives the old port for App B, and service invocation fails until the cache is cleared

The above is a bad DX. It can be fixed by passing `--dapr-internal-grpc-port`, but then developers need to manually set non-conflicting ports and that's not great DX either.

This PR changes the way the "random" port is selected by first attempting to use a "stable random" port, which is computed by looking at 4 variables which should be stable when the app is restarted:

- app-id: app-id values should be unique within a solution
- Current working directory: allows separating different projects a developer is working on
- ID of the current user: allows separating different users
- Hostname: this is also different within each container

If the port computed that way is not available, only then Dapr uses a random port as fallback.

This improves the DX by making it so if the same app is restarted, its internal server will (likely) listen on the same port, so it doesn't break service invocation.

### DIFF
--- a/pkg/grpc/server.go
+++ b/pkg/grpc/server.go
@@ -63,7 +63,7 @@ type server struct {
 	metricSpec         config.MetricSpec
 	authenticator      auth.Authenticator
 	servers            []*grpcGo.Server
-	renewMutex         *sync.Mutex
+	renewMutex         sync.Mutex
 	signedCert         *auth.SignedCertificate
 	tlsCert            tls.Certificate
 	signedCertDuration time.Duration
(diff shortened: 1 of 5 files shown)


berndverst (approved):

mukundansundar (approved):
lgtm

PR #5926, opened 2023-02-13
Add `enabled` option to sidecar-injector to gate whether the sidecar injector chart is installed

JoshVanL:
# Description

For some use cases of dapr, users would like to manage their own daprd instances so installing the injector is a waste of bytes and cpu cycles in the cluster.

This PR adds a `dapr_sidecar_injector.enabled` helm option which gates whether the sidecar injector is installed. This value defaults to `true`.

/cc @famarting

### DIFF
--- a/charts/dapr/README.md
+++ b/charts/dapr/README.md
@@ -174,7 +174,7 @@ The Helm chart has the follow configuration options that can be supplied:
 ### Dapr Sidecar Injector options:
 | Parameter                                 | Description                                                             | Default                 |
 |-------------------------------------------|-------------------------------------------------------------------------|-------------------------|
-| `dapr_sidecar_injector.enable`            | Enable the sidecar injector                                             | `enable`                |
+| `dapr_sidecar_injector.enabled`           | Enable the sidecar injector                                             | `enable`                |
 | `dapr_sidecar_injector.sidecarImagePullPolicy`      | Dapr sidecar image pull policy                                | `IfNotPresent`                     |
 | `dapr_sidecar_injector.replicaCount`      | Number of replicas                                                      | `1`                     |
 | `dapr_sidecar_injector.logLevel`          | Log level                                                               | `info`                  |
--- a/charts/dapr/README.md
+++ b/charts/dapr/README.md
@@ -174,6 +174,7 @@ The Helm chart has the follow configuration options that can be supplied:
 ### Dapr Sidecar Injector options:
 | Parameter                                 | Description                                                             | Default                 |
 |-------------------------------------------|-------------------------------------------------------------------------|-------------------------|
+| `dapr_sidecar_injector.enable`            | Enable the sidecar injector                                             | `enable`                |
 | `dapr_sidecar_injector.sidecarImagePullPolicy`      | Dapr sidecar image pull policy                                | `IfNotPresent`                     |
 | `dapr_sidecar_injector.replicaCount`      | Number of replicas                                                      | `1`                     |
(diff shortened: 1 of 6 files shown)


famarting:
lgtm, thank you!

ItalyPaleAle (approved):

charts/dapr/README.md
```diff
@@ -174,6 +174,7 @@ The Helm chart has the follow configuration options that can be supplied:
 ### Dapr Sidecar Injector options:
 | Parameter                                 | Description                                                             | Default                 |
 |-------------------------------------------|-------------------------------------------------------------------------|-------------------------|
+| `dapr_sidecar_injector.enable`            | Enable the sidecar injector                                             | `enable`                |
```
mukundansundar, line 177: nitpick... 
`dapr_sidecar_injector.enabled`

charts/dapr/charts/dapr_sidecar_injector/values.yaml
```diff
@@ -1,3 +1,4 @@
+enabled: true
```
mukundansundar, line 1: going back to https://github.com/dapr/dapr/pull/5896, if the default value is going to be true, should we provide a flag to instead `disable` sidecar injector?

charts/dapr/README.md
JoshVanL: Thanks for the catch!

charts/dapr/charts/dapr_sidecar_injector/values.yaml
JoshVanL: Would this be a flag on the `dapr` cli? And if so, how would we want to handle the case where there is a conflict? `dapr init -k --disable-sidecar-injector --set dapr_sidecar_injector.enabled=true`

mukundansundar: sorry I meant have the config option as `dapr_sidecar_injector.disabled` instead of `enabled`.

JoshVanL: Ah gotcha. I'd much rather we use the [convention `enabled`](https://helm.sh/docs/chart_best_practices/dependencies/#conditions-and-tags) as I see this everywhere in the wild and consider it as a convention.

ItalyPaleAle: I don't think we should have a Dapr CLI flag `--disable-sidecar-injector`. Seems a niche-enough thing that could be kept as a Helm-chart only option

charts/dapr/README.md
```diff
@@ -174,6 +174,7 @@ The Helm chart has the follow configuration options that can be supplied:
 ### Dapr Sidecar Injector options:
 | Parameter                                 | Description                                                             | Default                 |
 |-------------------------------------------|-------------------------------------------------------------------------|-------------------------|
+| `dapr_sidecar_injector.enabled`           | Enable the sidecar injector                                             | `enable`                |
```
ItalyPaleAle, line 177:
```suggestion
| `dapr_sidecar_injector.enabled`           | Enable the sidecar injector                                             | `true`                |
```

ItalyPaleAle (approved):

charts/dapr/charts/dapr_sidecar_injector/values.yaml
yaron2: CLI doesn't need a flag.

yaron2 (approved):
LGTM

PR #5937, opened 2023-02-15
Add dapr label to injected sidecar pod

### LINKED ISSUE #5693: High memory consumption for Operator when there are lots of pods

filintod:
# Description

Add dapr injected label `dapr.io/sidecar-injected` when patching pod in MutatingWebHook.  This will help to filter pods injected with sidecar by the webhook, and this way it can be used in the operator to list pods filtering out those with the label present and thus reducing the memory issue mentioned in ticket below.

## Issue reference

This will not close but is one step to fix #5693 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Unit tests passing


pkg/injector/annotations/annotations.go
```diff
@@ -47,6 +47,7 @@ const (
 	KeyReadinessProbePeriodSeconds      = "dapr.io/sidecar-readiness-probe-period-seconds"
 	KeyReadinessProbeThreshold          = "dapr.io/sidecar-readiness-probe-threshold"
 	KeySidecarImage                     = "dapr.io/sidecar-image"
+	KeySidecarInjected                  = "dapr.io/sidecar-injected"
```
ItalyPaleAle, line 50: Since this is a label and not an annotation, can you move it outside of the `annotations` package? It's probably ok in `pkg/injector/sidecar/consts.go`. I'd also change the name to `SidecarInjectedLabel` or something like that

pkg/injector/sidecar/pod_patch_test.go
```diff
@@ -14,9 +14,13 @@ limitations under the License.
 package sidecar
 
 import (
+	"encoding/json"
 	"strconv"
 	"testing"
 
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	jsonpatch "github.com/evanphx/json-patch"
```
ItalyPaleAle, line 23:
```suggestion
	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
	jsonpatch "github.com/evanphx/json-patch"
```

Gotta make the linter happy...

ItalyPaleAle (changes requested):

ItalyPaleAle (dismissed):
LGTM if it works :)

artursouza (approved):

ItalyPaleAle:
Perfect. However there are some issues with the CI. Can you please take a look at the test failures?

Looks like you may need to run "make modtidy-all"

pkg/injector/sidecar/pod_patch_test.go
```diff
@@ -14,11 +14,14 @@ limitations under the License.
 package sidecar
 
 import (
+	"encoding/json"
 	"strconv"
 	"testing"
 
+	jsonpatch "github.com/evanphx/json-patch"
```
ItalyPaleAle, line 21: Please use `"github.com/evanphx/json-patch/v5"` (just realized there's the new version by looking at our go.mod)

ItalyPaleAle (changes requested):

pkg/injector/sidecar/pod_patch_test.go
filintod: done

ItalyPaleAle (approved):

artursouza (approved):

PR #5944, opened 2023-02-17
Reduce operator cache memory consumption

filintod:
# Description

This PR aims at reducing the cache memory footprint by using a series of memory reducing techniques.  The operator can consume large amounts of memory in large clusters as it will be watching for Deployments and Statefulsets, and if set with watchdog, it will also watch for all Pods in the cluster.  This can lead to large amounts of memory being used by it.

This PR changes the following:
- added an extra label patch added by the operator for cases where the pod does not have the injector label, but does have the daprd container. This is mainly to take into account the transition period between using old sidecar injector without the injected label, where the pods will have the sidecar but not the label. This way we can filter those by label and save the memory.
- added watch verb to operator rbac as it provides smoother operation, and an optional patch verb \when setup to patch pod labels. This way we can patch the pods and be notified via informers of the change so subsequent queries would not have it. 
- added custom cache to manager. This cache does the following:
  - limits what Pods are stored in the cache, only allowing pods that do not match the injected label, nor the patched label.
  - pods that are stored are reduced in size by removing managed fields and setting status to be an empty struct
  - other pods are sinkhole into a singleton reference empty pod
  - limits what Deployment/StatefulSets are stored in the cache, in the same way it is done with Pods, with an exception for the injector deployment

## Issue reference

Please reference the issue this PR will close: #5693

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Unit tests passing


filintod:
@artursouza fyi

filintod:
we could also use a non-cache back client to reduce memory usage further, maybe using the cache-back for the metadata list query and the non-cache for each of the pod query details

snoopy-dooog:
@filintod, I took a look at your code and I'm a little bit lost. I might have misunderstood the problem but to me it seemed like we wanted to look for the pods that had the sidecar injected label and then see if those pods had the sidecar container. Also, now that we have the sidecar, I think this could be simplified to not have the check for the annotations and do all in one go, something like this: https://github.com/dapr/dapr/pull/5955

I'm planning on writing the tests for this today but if it's wrong then I'll just abandon the draft and let you work on this here while picking another one since I'm just getting familiar with all of this and might have not understand it correctly. Thanks!

filintod:
@snoopy-dooog it is my understanding that the issue is to reduce memory utilization because how the watchdogs tries to find which pods that should have the injected sidecar don't. Currently, not using the new injected label, we are listing all the pods in the cluster to check which ones have dapr enabled annotation `dapr.io/enabled`, then check if the sidecar is present and if not delete it.  As we are using a cache-backed client (coming from the controller-runtime manager), when we do a list of all pods we are downloading all the pods information locally that can be substantial. 

The watchdog is just a safeguard for those cases where this sidecar is not injected but the annotation is present, maybe the webhook injector was not deployed or failed, or etc.  If you check the [default dapr injector webhook failure policy](https://github.com/dapr/dapr/blob/master/charts/dapr/charts/dapr_sidecar_injector/values.yaml#L20) in the chart, it is set to [ignore](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#failure-policy). So the pod will not failed to be deployed even if it failed to be patched.  So even with the new label we still have to check for those pods that have not been patched by the injector.  And why I have this negative check against the label being present in the pod (find which ones have not been patched for any reason).

ItalyPaleAle:
The watchdog is not the problem here, but the fact that the Operator has a reconciler for Deployment and StatefulSets.

We cannot reduce the cost of the watchdog, which cannot filter by pods with a label. This is because the label is injected by the injector, and the point of the watchdog is to check when the injector hasn't worked...

filintod:
@ItalyPaleAle 

this PR introduces a reduction of memory from watchdog by listing pods by metadata only:

image

On this image we have the operator running on 3 different scenarios (not really large scale but can give us some information):
- first (yellow) operator without watchdog (interval `0` that disables it and seems to be the default too).  I created 200 pods and did not see a difference and realized watchdog was not disabled.
- second (blueish) operator with 15s watchdog. It started without any pods created then I created 200 pods first, at around 14:53 mark, then increase it to 300 at around 14:57 mark, and left the pods running.
- third (greenish) operator, with 15s watchdog, where I replaced the image with this new image with reducing actions.  You can see that there is no much difference with the operator when there were no pods.

This still does not show what happen when we have dapr pods without the label where we will incur in some extra memory as we will be querying the whole pod and not just the metadata, but we will also delete it after to trigger injection. So we'll see on those later in subsequent checks.

I have not tested multiple deployments/sts, but my guess the issue is the number of pods more than the number of deploy/sts.

filintod:
the idea of this PR in connection with the injected label is to reduce the number of queries for whole pod, for those where the injector did not work.  As I said there will be a period of time where the reduction will not be as pronounced for those pods with a sidecar but without a label.  We could opt to use a non-cache back client or patching the pods with daprd sidecar but without the label with the injected label (that supposedly should not restart the pod - but I need to test this)

filintod:
@ItalyPaleAle just added the ability to patch pods when user has opt-in for this option in the helm charts `global` section, as it entails adding rbac permits as well as additional argument passed to watchdog.  This will probably be the most we can do at the moment, unless somebody think of some other clever actions.

cmd/operator/main.go
```diff
@@ -20,24 +20,26 @@ import (
 
 	"k8s.io/klog"
 
+	"github.com/dapr/kit/logger"
+
```
ItalyPaleAle, line 24: Remove newline here, or linter won't be happy

pkg/operator/operator.go
```diff
@@ -25,6 +25,9 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
+	"github.com/dapr/kit/fswatcher"
+	"github.com/dapr/kit/logger"
+
```
ItalyPaleAle, line 30: Please remove newline here too, the linter doesn't like that

pkg/operator/operator.go
```diff
@@ -108,8 +110,10 @@ func NewOperator(opts Options) Operator {
 		}
 		wd := &DaprWatchdog{
 			client:            mgrClient,
+			nonCachedClient:   mgr.GetAPIReader(),
```
ItalyPaleAle, line 113: Why removing the cache? Wouldn't this increase the load on the K8s APIs?

pkg/operator/watchdog.go
```diff
@@ -70,11 +87,11 @@ func (dw *DaprWatchdog) Start(parentCtx context.Context) error {
 				if !ok {
 					continue
 				}
-				ok = dw.listPods(ctx)
-				if firstCompleteCh != nil {
+				ok = dw.listPods(ctx, sel)
+				if !firstCompleted {
```
ItalyPaleAle, line 91: Why this change? Checking for channel being nil should have been enough

ItalyPaleAle (changes requested):
I'm all for filtering based on whether a label exists, but I think the label should be set by the injector and not by the watchdog...

pkg/operator/watchdog.go
filintod: there was a race condition

pkg/operator/operator.go
filintod: this is used with patch only, so the expectation is that the we hit the query once, patch, and it should not appear next time.

filintod:
@ItalyPaleAle the second label added by the watchdog was for those cases where, first the user has allow it as it requires extra rbac patch permits that is an understandable security concern, and second as the pod would not get deleted as it already has the daprd sidecar, it will always appear in the metadata list query, and thus will also be query after again.

Also this patch behavior is not the default, the user needs to opt-in for it. In most cases it might not matter but for large clusters it might.

filintod:
@ItalyPaleAle yeah, just checked and the cached client informer does the correct thing, so I removed the non cache backed client.

filintod:
one thing, without pod watch (that right now is only been added with the watchdog-can-path-pod-labels option), the pod list will return the error: 

```
dapr-operator-7ff4b44dd4-ffxjv dapr-operator E0222 20:31:18.851170       1 reflector.go:140] sigs.k8s.io/controller-runtime/pkg/cache/internal/informers_map.go:262: Failed to watch *v1.PartialObjectMetadata: unknown
```

I guess the same is happening today, when the pods try to start the informer watch and don't have permits.  So my question is, should I set it to have `watch` for pods all the time in the rbac?

ItalyPaleAle:
@filintod what's the status of this PR? Is it ready for review or are you waiting for something?

filintod:
@ItalyPaleAle sorry, I'm finishing some tests as I added a custom cache to reduce the cost of deployment/sts as well as pods that is not uploaded on this one.  I realized what you mentioned, and that the watchdog is optional, where deploy/sts costs are always there. Something I'm trying to figure out is why the memory usage goes up quickly when dapr applications are present. I'll put this back to draft in the meantime.

ItalyPaleAle (changes requested):

filintod:
btw, I'm thinking on enabling watch on Pods even if not allow to patch.  The resource behavior seems more stable when watch is enabled, I guess because there are not big swings when getting the list, but more progressive over time.

This is an example of the current operator in 0.10.2. (not the memory reducing one)

image

filintod:
note, on latest change I drop the metadata only query, at the end as we are always needing a watch against pods to check for running containers, we don't save much with the metadata only one.

filintod:
@ItalyPaleAle this is ready for review

filintod:
another graph of the savings with new operator on left and old one on right, with dapr apps

image

ItalyPaleAle (dismissed):

pkg/operator/watchdog.go
```diff
@@ -145,47 +181,59 @@ func (dw *DaprWatchdog) listPods(ctx context.Context) bool {
...
+	podList := &corev1.PodList{}
+	err = dw.client.List(ctx, podList, podListOpts)
 	if err != nil {
 		log.Errorf("Failed to list pods. Error: %v", err)
 		return false
 	}
 
-	for _, v := range pod.Items {
+	for _, pod := range podList.Items {
+		pod := pod
```
ItalyPaleAle, line 199:
```suggestion
	for i := range podList.Items {
		pod := podList.Items[i]
```

ItalyPaleAle (approved):

yaron2 (approved):

mukundansundar:
@filintod please do comment on https://github.com/dapr/dapr/issues/5693 so that it can be assigned to you

PR #5948, opened 2023-02-17
Add prettier to lint JavaScript code

ItalyPaleAle:
We have enough JS code in this repo now that it makes sense to add a linter to format it, like prettier.

New commands:

- `prettier-install` installs the prettier package from NPM
- `prettier-check` checks if the code is linted and prints a warning otherwise
- `prettier-format` lints all code and formats it in-place

I haven't added this as a step to our pipelines because I don't think it's that important yet. But it's nice to have a command to format all code at once.

I've implemented a similar change in a PR in components-contrib.

PS: No actual code changes in this PR. I just ran prettier and that's why there are changes


Makefile
```diff
@@ -464,6 +464,19 @@ ifeq (,$(wildcard $(BUILD_TOOLS)))
...
+# Prettier                                                                     #
+################################################################################
+.PHONY: prettier-install prettier-check prettier-format
+prettier-install:
+	npm install --global prettier
+
+prettier-check:
+	npx prettier --check "*/**/*.{ts,js,mjs,json}"
+
+prettier-format:
```
ItalyPaleAle, line 477: I just realized I missed a chance to name this "make me-prettier" 🤣

mukundansundar (dismissed):
lgtm

daixiang0:
Please add those to CI step (only trigger when JS code change) or update CONTRIBUTION.md, or contributors may ignore this.

Also, I want to confirm that we use same linter in website repo.

ItalyPaleAle:
@daixiang0 Updated CONTRIBUTING with this.

Also, I did add `make me prettier` 🤣

artursouza (approved):

PR #5952, opened 2023-02-18
Cherry-pick #5584 into release-1.10 [DO NOT SQUASH]

ItalyPaleAle:
#5584 was included in a hotfix for 1.9 but we never cherry-picked into master so it didn't make its way into 1.10. Hopefully it can be included in the next hotfix for 1.10

**Please do not squash this commit**

### DIFF
--- a/pkg/runtime/runtime.go
+++ b/pkg/runtime/runtime.go
@@ -2773,11 +2773,6 @@ func (a *DaprRuntime) processComponentSecrets(component componentsV1alpha1.Compo
 		}
 
 		secretStoreName := a.authSecretStoreOrDefault(component)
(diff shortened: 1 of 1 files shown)


berndverst (approved):

yaron2:
We don't have a hotfix planned for 1.10

ItalyPaleAle:
> We don't have a hotfix planned for 1.10

I am aware. But it will happen sooner or later :) Since this commit was in a hotfix for 1.9, it should at least be included in the next hotfix for 1.10 IMHO

yaron2 (approved):

artursouza (approved):

Issue #5959, opened 2023-02-20
Dapr 1.10.0 - CloudEvent Id is not a UUID


wtulloch:
## In what area(s)?

Area/Runtime

## What version of Dapr?
1.10.x

## Expected Behavior
When publishing a message the Id in the CloudEvent data should be a Guid.
In 1.9.6 this is an example of the cloudEvent data:

``` json
{
  "data": {
   "text": "data goes here"
  },
  "datacontenttype": "application/json",
  "id": "ffbe0263-dcee-4a28-88bf-f85691bbb32f",
  "pubsubname": "pubsub",
  "source": "ingestion",
  "specversion": "1.0",
  "time": "2023-02-20T10:03:07Z",
  "topic": "sentence-saved",
  "traceid": "00-38314ed375a9dc88f9159d60a5bb6678-af654446698a02ef-01",
  "traceparent": "00-38314ed375a9dc88f9159d60a5bb6678-af654446698a02ef-01",
  "tracestate": "",
  "type": "com.dapr.event.sent"
}
```

## Actual Behavior
With 1.10.0 the CloudEvent Id is a string and uses the source as the value.

``` json
{
    "data": {
       "text": "data goes here"
      },
      "timestampSaved": "2023-02-20T08:35:22.715943+00:00",
      "timestampStarted": "2023-02-20T08:35:18.553177+00:00"
    },
    "datacontenttype": "application/json",
    "id": "ingestion",
    "pubsubname": "pubsub",
    "source": "Dapr",
    "specversion": "1.0",
    "time": "2023-02-20T08:35:22Z",
    "topic": "sentence-saved",
    "traceid": "00-5a6f9181c395899122115e23971ffdd6-a21b72b39f01e919-01",
    "traceparent": "00-5a6f9181c395899122115e23971ffdd6-a21b72b39f01e919-01",
    "tracestate": "",
    "type": "com.dapr.event.sent"
}
```
** Note: **
If one of the example applications are run this doesn't cause an issue. In our code though we have defined a strongly typed Cloud Event Response. What this means is that the json deserialiser fails when the message is received by a subscriber.

```csharp
public class CloudEventResponse {
    [JsonPropertyName("id")]
    public Guid Id { get; set; }

    [JsonPropertyName("source")]
    public string Source { get; set; }

    [JsonPropertyName("topic")]
    public string Topic {get;set;}

    [JsonPropertyName("data")]
    public CloudEventData Data { get; set; }
}
```

## Release Note

[...]

PR #5960, opened 2023-02-21
1.10 Hotfix - Fix CloudEvent regression

yaron2:
Fixes https://github.com/dapr/dapr/issues/5959, adds test conditions to validate correctness.


yaron2:
Re-opening against correct branch.

PR #5967, opened 2023-02-21
Allow trace header propagation without trace exporting

berndverst:
This change allows trace ID / traceparent to be propagated or generated by the runtime without exporting full trace logs.

In configuration this can be done via

```
apiVersion: dapr.io/v1alpha1
kind: Configuration
metadata:
  name: daprConfig
spec:
  tracing:
    samplingRate: "1"
```

Previously it was necessary to choose one of the existing exporters like `otel`, `zipkin` or `stdout` (and configure an endpoint for the former two) even if all one was interested in was the forwarding of trace headers through Dapr.

Internally this is accomplished by creating a `NullExporter` which is registered if none of the other trace exporters have been registered.

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Unit tests passing

### DIFF
--- a/pkg/runtime/runtime.go
+++ b/pkg/runtime/runtime.go
@@ -408,6 +408,10 @@ func (a *DaprRuntime) setupTracing(hostAddress string, tpStore tracerProviderSto
 		tpStore.RegisterExporter(otelExporter)
 	}
 
+	if !tpStore.HasExporter() && tracingSpec.SamplingRate != "" {
+		tpStore.RegisterExporter(diagUtils.NewNullExporter())
+	}
+
 	// Register a resource
 	r := resource.NewWithAttributes(
 		semconv.SchemaURL,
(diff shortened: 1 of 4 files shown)


pkg/runtime/runtime.go
```diff
@@ -408,6 +408,10 @@ func (a *DaprRuntime) setupTracing(hostAddress string, tpStore tracerProviderSto
 		tpStore.RegisterExporter(otelExporter)
 	}
 
+	if !tpStore.HasExporter() && tracingSpec.SamplingRate != "" {
```
berndverst, line 411: Note that this entire code (and the code above) won't be used if Sampling Rate is 0 anyway

ItalyPaleAle, line 411: Shouldn't this check if the value of SamplingRate is also != "0" ?

ItalyPaleAle: Oh, I saw your comment posted a the same time as mine :)

ItalyPaleAle (approved):

yaron2 (approved):

PR #5969, opened 2023-02-21
Standardize postgres/postgresql component names

ItalyPaleAle:
We currently have 3 components for postgres, and they do not use the same nomenclature:

- `state.postgresql`
- `bindings.postgres`
- `configuration.postgres`

This PR changes all 3 to accept both "postgres" and "postgresql" (and we should probably change the docs to use "postgres" everywhere)

### DIFF
--- a/cmd/daprd/components/binding_postgres.go
+++ b/cmd/daprd/components/binding_postgres.go
@@ -19,5 +19,5 @@ import (
 )
 
 func init() {
-	bindingsLoader.DefaultRegistry.RegisterOutputBinding(postgres.NewPostgres, "postgres")
+	bindingsLoader.DefaultRegistry.RegisterOutputBinding(postgres.NewPostgres, "postgres", "postgresql")
 }
--- a/cmd/daprd/components/configuration_postgres.go
+++ b/cmd/daprd/components/configuration_postgres.go
@@ -19,5 +19,5 @@ import (
 )
 
 func init() {
-	configurationLoader.DefaultRegistry.RegisterComponent(configPostgres.NewPostgresConfigurationStore, "postgres")
+	configurationLoader.DefaultRegistry.RegisterComponent(configPostgres.NewPostgresConfigurationStore, "postgres", "postgresql")
 }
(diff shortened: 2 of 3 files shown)


berndverst (approved):

yaron2:
Looks like if we want to standardize on postgres, then this PR should only contain the change for the state store and update the docs?

ItalyPaleAle:
I'm a fan of erring on the side of caution, since it doesn't cost us anything to add the alias "postgresql" everywhere and may save users from a frustrating experience. Even if we update docs going forward, people may find old docs, for example in third-party articles.

berndverst:
> I'm a fan of erring on the side of caution, since it doesn't cost us anything to add the alias "postgresql" everywhere and may save users from a frustrating experience. Even if we update docs going forward, people may find old docs, for example in third-party articles.

I agree with Alessandro. It doesn't hurt - and I personally find it convenient.

yaron2:
> I'm a fan of erring on the side of caution, since it doesn't cost us anything to add the alias "postgresql" everywhere and may save users from a frustrating experience. Even if we update docs going forward, people may find old docs, for example in third-party articles.

Trying to understand what you mean by caution - If they find an old doc for state/postresql and choose to use that, it will work since the old name is still supported. We shouldn't have any docs that specify postgresql for configuration/binding, no?

ItalyPaleAle:
People may find docs that reference `state.postgresql` and then when they try to switch to a config store, would naturally be included to use `config.postgresql`.

Both "postgres" and "postgresql" are used to talk about the database, even though "postgres" is probably the most common one now, so it's not unthinkable that someone may try with either.

But also - it really doesn't cost us anything to add an alias for a component name, it's extremely unlikely there'll be another different database called "postgresql" in the future, so I don't really see a reason why we shouldn't?

yaron2:
It looks like the project community has decided to keep the official name as PostgreSQL while treating Postgres as an alias. While this information might invalidate making postres the name that appears in docs, it makes current changes more logical. While recognizing case by case basis validation of PRs, we have experienced all sorts of side effects from changes that were labeled as trivial. lgtm!

yaron2 (approved):

ItalyPaleAle:
Ok, we'll standardize the docs to use "PostgreSQL" then (sneaking in some "Postgres" for SEO :) )

PR #5977, opened 2023-02-22
Add API denylist to Dapr config, complementing API allowlist

ItalyPaleAle:
# Description

Dapr has included an API allowlist in the Configuration resource for a number of releases.

However, we've observed that most customers are using the allowlist more as a "denylist", i.e. they list all Dapr APIs except _some_. Feedback we received is that this complicates operations when a new version of Dapr is released.

This PR implements additional support for API denylisting. The denylist is complementary to the allowlist, and the two can be used simultaneously too:

- If only a denylist is defined, all Dapr APIs are allowed except those defined in the denylist
- If only an allowlist is defined, only the Dapr APIs listed in the allowlist are allowed
- If both an allowlist and a denylist are defined, the allowed APIs are those defined in the allowlist _unless_ they are also included in the denylist (i.e. the denylist overrides the allowlist for APIs that are defined in both places).
- If neither is defined, all APIs are allowed.

## Checklist

- [ ] Docs issue: TODO @ItalyPaleAle


pkg/http/server_test.go
```diff
@@ -58,413 +52,6 @@ func newServer() server {
 	}
 }
 
-func TestAllowedAPISpec(t *testing.T) {
```
ItalyPaleAle, line 61: Reviewers: these tests were moved to the endpoints_test  file (with new tests added)

yaron2:
@ItalyPaleAle please follow Dapr's contribution guidelines going forward and create an issue prior to opening a PR that adds a new feature and changes our specs. Thanks!

ItalyPaleAle:
My apologies, I'll be more mindful next time

mukundansundar (commented):
Please add E2E for this feature.

charts/dapr/crds/configuration.yaml
```diff
@@ -85,9 +85,25 @@ spec:
                 description: APISpec describes the configuration for Dapr APIs.
                 properties:
                   allowed:
+                    description: List of allowed APIs. Can be used in conjunction with Denied.
                     items:
-                      description: APIAccessRule describes an access rule for allowing
-                        a Dapr API to be enabled and accessible by an app.
+                      description: APIAccessRule describes an access rule for allowing or disallowing a Dapr API.
```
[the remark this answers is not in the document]
ItalyPaleAle: It comes from the Go struct that's used for both allow and deny so i can't have it different :(

ItalyPaleAle:
> Please add E2E for this feature.

There's currently no E2E test for API allowlisting or denylisting, and although we should create one, perhaps that could be a separate PR?

pkg/grpc/endpoints.go
```diff
@@ -102,33 +102,30 @@ var endpoints = map[string][]string{
...
+					_, ok := allowed[info.FullMethod]
+					if !ok {
+						return nil, invokev1.ErrorFromHTTPResponseCode(http.StatusNotImplemented, "requested endpoint is not available")
+					}
+				}
+				if len(denied) > 0 {
+					_, ok := denied[info.FullMethod]
+					if ok {
+						return nil, invokev1.ErrorFromHTTPResponseCode(http.StatusNotImplemented, "requested endpoint is not available")
+					}
```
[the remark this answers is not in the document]
ItalyPaleAle: It shouldn't matter. If the same API is both in allowlist and denylist, the method will return an error regardless of the order.

daixiang0:
> > Please add E2E for this feature.
> 
> There's currently no E2E test for API allowlisting or denylisting, and although we should create one, perhaps that could be a separate PR?

Please create an issue to track it.

ItalyPaleAle:
> Please create an issue to track it.

#6240

charts/dapr/crds/configuration.yaml
```diff
@@ -85,9 +85,25 @@ spec:
...
+                          type: string
+                        version:
+                          type: string
+                      required:
+                      - name
+                      - version
+                      type: object
+                    type: array
+                  denied:
+                    description: List of disallowed APIs. Can be used in conjunction with allowed.
```
daixiang0, line 104: typo.

charts/dapr/crds/configuration.yaml
```diff
@@ -85,9 +85,25 @@ spec:
...
+                    items:
+                      description: APIAccessRule describes an access rule for allowing or disallowing a Dapr API.
```
daixiang0, line 106: APIAccessRule describes an access rule to allow or deny one Dapr API.

pkg/apis/configuration/v1alpha1/types.go
```diff
@@ -65,10 +65,15 @@ type ConfigurationSpec struct {
...
+	// List of allowed APIs. Can be used in conjunction with denied.
+	// +optional
 	Allowed []APIAccessRule `json:"allowed,omitempty"`
+	// List of disallowed APIs. Can be used in conjunction with allowed.
+	// +optional
+	Denied []APIAccessRule `json:"denied,omitempty"`
 }
 
-// APIAccessRule describes an access rule for allowing a Dapr API to be enabled and accessible by an app.
+// APIAccessRule describes an access rule for allowing or disallowing a Dapr API.
```
daixiang0, line 76: allow and deny, let's keep sync

pkg/config/configuration.go
```diff
@@ -139,6 +142,22 @@ type APIAccessRule struct {
 	Protocol string `json:"protocol"`
 }
 
+// APIAccessRules is a list of API access rules (allowlist or denylist).
+type APIAccessRules []APIAccessRule
+
+// ForProtocol returns a list of APIAccessRule objects filtered by protocol
+func (r APIAccessRules) ForProtocol(protocol string) []APIAccessRule {
```
daixiang0, line 149: rename as `GetRulesByProtocol`?

pkg/config/configuration.go
```diff
@@ -139,6 +142,22 @@ type APIAccessRule struct {
...
+	res := make([]APIAccessRule, len(r))
```
daixiang0, line 150: how about `func (r APIAccessRules) ForProtocol(protocol string) (ret []APIAccessRule) {` to reduce `n`?

pkg/grpc/endpoints.go
```diff
@@ -137,11 +134,39 @@ func setAPIEndpointsMiddlewares(rules []config.APIAccessRule) (grpc.UnaryServerI
...
+// Converts a slice of config.APIAccessRule into a map where the key is the gRPC full endpoint
+func apiAccessRuleToMap(rules config.APIAccessRules) map[string]struct{} {
+	res := map[string]struct{}{}
+
+	for _, rule := range rules {
+		if rule.Protocol != "grpc" {
+			continue
+		}
+
+		if list, ok := endpoints[rule.Name+"."+rule.Version]; ok {
```
daixiang0, line 164: Please use const and fmt, also comment or document it for better maintain.

pkg/http/server.go
```diff
@@ -335,8 +335,13 @@ func (s *server) unescapeRequestParametersHandler(next fasthttp.RequestHandler)
 func (s *server) getRouter(endpoints []Endpoint) *routing.Router {
 	router := routing.New()
 	parameterFinder, _ := regexp.Compile("/{.*}")
+
+	// Build the API allowlist and denylist
+	allowedAPIs := s.apiSpec.Allowed.ForProtocol("http")
```
daixiang0, line 340: So it is better if we make all support protocols as const.

pkg/grpc/endpoints.go
ItalyPaleAle: I will document it, but concatenation is significantly faster than fmt.Sprintf for simple string concatenation:

```go
func BenchmarkFmt(b *testing.B) {
	const name = "foo"
	const version = "bar"
	for i := 0; i < b.N; i++ {
		_ = fmt.Sprintf("%s.%s", name, version)
	}
}
func BenchmarkConcat(b *testing.B) {
	const name = "foo"
	const version = "bar"
	for i := 0; i < b.N; i++ {
		_ = name + "." + version
	}
}
```

On my machine:

```text
goos: darwin
goarch: arm64
pkg: github.com/dapr/dapr
BenchmarkFmt
BenchmarkFmt-8      	20474700	        58.38 ns/op	       8 B/op	       1 allocs/op
BenchmarkConcat
BenchmarkConcat-8   	1000000000	         0.3238 ns/op	       0 B/op	       0 allocs/op
```

Not just concatenation is MUCH faster (99.95% faster), but fmt.Sprintf also causes heap allocations so more work on the GC.

pkg/apis/configuration/v1alpha1/types.go
ItalyPaleAle: Done

charts/dapr/crds/configuration.yaml
ItalyPaleAle: Done

pkg/config/configuration.go
ItalyPaleAle: Done

charts/dapr/crds/configuration.yaml
ItalyPaleAle: Done

pkg/http/server.go
ItalyPaleAle: Done

pkg/grpc/endpoints.go
```diff
@@ -137,11 +134,40 @@ func setAPIEndpointsMiddlewares(rules []config.APIAccessRule) (grpc.UnaryServerI
...
+					_, ok := allowed[info.FullMethod]
+					if !ok {
+						return invokev1.ErrorFromHTTPResponseCode(http.StatusNotImplemented, "requested endpoint is not available")
+					}
+				}
+				if len(denied) > 0 {
+					_, ok := denied[info.FullMethod]
+					if ok {
+						return invokev1.ErrorFromHTTPResponseCode(http.StatusNotImplemented, "requested endpoint is not available")
+					}
```
mukundansundar, line 147: @ItalyPaleAle The order here is opposite of the order seeing in `http/endpoint.go`. In the HTTP one, denied is checked first and then allowed is checked. Here the opposite is true.

pkg/config/configuration.go
```diff
@@ -129,14 +129,41 @@ type PipelineSpec struct {
...
+	APIAccessRuleProtocolHTTP APIAccessRuleProtocol = "http"
+	APIAccessRuleProtocolGRPC APIAccessRuleProtocol = "grpc"
+)
+
+// GetRulesByProtocol returns a list of APIAccessRule objects filtered by protocol
+func (r APIAccessRules) GetRulesByProtocol(protocol APIAccessRuleProtocol) []APIAccessRule {
+	res := make([]APIAccessRule, len(r))
+	n := 0
+	for _, v := range r {
+		if v.Protocol == protocol {
```
mukundansundar, line 161: what if the input given is mixed case or all upper case?

mukundansundar (approved):

yaron2:
Does this have a corresponding docs PR? cc @ItalyPaleAle

ItalyPaleAle:
> Does this have a corresponding docs PR? 

[Now there is.](https://github.com/dapr/docs/pull/3489) Thanks for the reminder.

PR #5992, opened 2023-02-25
Workflow: Fix generation handling for continue-as-new

### LINKED ISSUE #5986: Workflow Monitor issue with duplicate invocation

cgillum:
# Description

This PR fixes a bug in the Dapr Workflow engine that breaks basic scenarios involving the [continue-as-new workflow feature](https://docs.dapr.io/developing-applications/building-blocks/workflow/workflow-features-concepts/#infinite-loops-and-eternal-workflows).

Prior to this fix, workflow activity invocations would fail with a "duplicate invocation" error. This would cause the workflow to stall indefinitely. This PR does two things to improve this:

1. It fixes the root bug, which is that we forgot to increment the workflow "generation" counter when a continue-as-new is encountered (this counter is needed for detecting duplicate invocations).
2. It fixes the handling of duplicate invocations so that they aren't considered retriable. Instead, the correct behavior is to log a warning and skip that invocation (assuming it truly is a duplicate).

The existing functional test for continue-as-new missed this issue because it uses durable timers inside the workflow instead of activity invocations. The PR also improves the test to cover both timers and activity invocations.

The problem is easy to run into and it might be worth including as part of a planned hotfix, if possible (though this may go against normal policy for alpha features).

## Issue reference

Fixes #5986 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Unit tests passing

### DIFF
--- a/pkg/runtime/wfengine/workflow.go
+++ b/pkg/runtime/wfengine/workflow.go
@@ -365,13 +365,11 @@ func (wf *workflowActor) runWorkflow(ctx context.Context, actorID string, remind
 			if err != nil {
 				if errors.Is(err, ErrDuplicateInvocation) {
 					wfLogger.Warnf("%s: activity invocation %s#%d was flagged as a duplicate and will be skipped", actorID, ts.Name, e.EventId)
-				} else {
-					return newRecoverableError(fmt.Errorf("failed to invoke activity actor '%s' to execute '%s': %v", targetActorID, ts.Name, err))
+					continue
 				}
+				return newRecoverableError(fmt.Errorf("failed to invoke activity actor '%s' to execute '%s': %v", targetActorID, ts.Name, err))
 			}
-			if resp != nil {
-				resp.Close()
-			}
+			resp.Close()
 		} else {
 			wfLogger.Warn("don't know how to process task %v", e)
 		}
--- a/pkg/runtime/wfengine/workflow.go
+++ b/pkg/runtime/wfengine/workflow.go
@@ -298,6 +298,12 @@ func (wf *workflowActor) runWorkflow(ctx context.Context, actorID string, remind
 		}
 	}
 
+	// Increment the generation counter if the workflow used continue-as-new. Subsequent actions below
+	// will use this updated generation value for their duplication execution handling.
+	if runtimeState.ContinuedAsNew() {
+		state.Generation += 1
+	}
+
 	if !runtimeState.IsCompleted() {
 		// Create reminders for the durable timers. We only do this if the orchestration is still running.
 		for _, t := range runtimeState.PendingTimers() {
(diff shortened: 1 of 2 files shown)


pkg/runtime/wfengine/workflow.go
```diff
@@ -357,10 +363,15 @@ func (wf *workflowActor) runWorkflow(ctx context.Context, actorID string, remind
 
 			resp, err := wf.actors.Call(ctx, req)
 			if err != nil {
-				return newRecoverableError(
-					fmt.Errorf("failed to invoke activity actor '%s' to execute '%s': %v", targetActorID, ts.Name, err))
+				if errors.Is(err, ErrDuplicateInvocation) {
+					wfLogger.Warnf("%s: activity invocation %s#%d was flagged as a duplicate and will be skipped", actorID, ts.Name, e.EventId)
```
ItalyPaleAle, line 367: Suggest adding a `continue` after this call, for 3 reasons:

1. Safer in case someone decided to add more code in the loop in the future, after this block
1. Then you don't need the `if resp != nil` check that was added
1. It makes the flow more clear

pkg/runtime/wfengine/workflow.go
```diff
@@ -298,6 +298,12 @@ func (wf *workflowActor) runWorkflow(ctx context.Context, actorID string, remind
 		}
 	}
 
+	// Increment the generation counter if the workflow used continue-as-new. Subsequent actions below
+	// will use this updated generation value for their duplication execution handling.
+	if runtimeState.ContinuedAsNew() {
+		state.Generation += 1
```
ItalyPaleAle, line 304: I'm not super familiar with the flow of the code, but is this ever accessed by any other goroutine? (Maybe not yet, I see a comment "TODO: Parallelism"?) If it is, should use atomic.Add instead or this can cause race conditions

pkg/runtime/wfengine/workflow.go
cgillum: I considered that but then I have to worry about the `defer resp.Close()` call never getting executed, right? This seemed like a simpler approach that won't make the linter that tracks response-closing angry.

pkg/runtime/wfengine/workflow.go
cgillum: No, this will never be accessed by any other goroutine. We can rely on actor concurrency guarantees for this. The TODO: Parallelism refers to parallel invocation of the "activity" actors further below, which won't modify `state`.

pkg/runtime/wfengine/workflow.go
ItalyPaleAle: If err != nil, resp should be nil (or at least shouldn't need closing) per Go convention.

ItalyPaleAle: Also since you mention that, worth pointing out that because this line is in a loop, the defer will be executed at the end of the method. Not sure if that could be an issue and maybe an immediate `resp.Close()` (without defer) may be better

cgillum: Okay, I went ahead and switched to using `continue`. I might have been overthinking this. I do like that the logic is much simpler now.

ItalyPaleAle (approved):

mylona79:
Hey Dapr team , I see that this is scheduled for v.1.10.3 hot fix release , any estimation when we await that to be available ?

PR #5993, opened 2023-02-25
[release-1.10] Dapr fails to initialize when a middleware component failes to init

ItalyPaleAle:
# Description

When Dapr is configured with middleware components (both app HTTP pipeline and HTTP pipeline) and a component fails to initialize, a warning is shown and then errors are ignored and Dapr continues to start regularly, but without any middleware (not just the one(s) that failed to init).

This behavior can cause unexpected results and in some cases even, potentially, security issues, when the middlewares are used for security purposes (since they'd be totally bypassed if one of the middlewares fails to init).

The same thing happens if the Dapr configuration references a middleware component that has not been defined in a component YAML. In that case, no middleware is loaded (including those that have a matching component YAML)

This PR changes the behavior of the middleware init logic so:

- ~~If the Dapr configuration references a middleware that doesn't have a Component YAML loaded, Dapr fails to init and shuts down right away. This applies to both pipelines ("dapr" and app)~~
   - Per @artursouza 's feedback, in this case Dapr will show an error but will not crash. However, it will continue loading other middleware components (before, an error like this would cause _no_ middleware to be loaded)
- If a middleware component fails to init, we now respect the value of `ignoreErrors` like we do for all other components:
  - If `ignoreErrors: false` (default value if omitted), Dapr fails to start and shuts down right away
  - If `ignoreErrors: true`, Dapr continues to init but without that specific middleware component (other middlewares however can still be used)

Added unit tests to validate the behavior.

## Issue reference

Fixes the runtime part of https://github.com/dapr/components-contrib/issues/2599

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests
* [X] Unit tests passing

### DIFF
--- a/pkg/runtime/runtime.go
+++ b/pkg/runtime/runtime.go
@@ -687,6 +687,8 @@ func (a *DaprRuntime) buildHTTPPipelineForSpec(spec config.PipelineSpec, targetP
 				a.Shutdown(a.runtimeConfig.GracefulShutdownDuration)
 				log.Fatalf("couldn't find middleware component with name %s and type %s/%s",
 					middlewareSpec.Name, middlewareSpec.Type, middlewareSpec.Version)
+				// This error is only caught by tests, since during normal execution we panic
+				return pipeline, errors.New("dapr panicked")
 			}
 			md := middleware.Metadata{Base: a.toBaseMetadata(component)}
 			handler, err := a.httpMiddlewareRegistry.Create(middlewareSpec.Type, middlewareSpec.Version, md, middlewareSpec.LogName())
@@ -696,6 +698,8 @@ func (a *DaprRuntime) buildHTTPPipelineForSpec(spec config.PipelineSpec, targetP
 					log.Warn("error processing middleware component, daprd process will exit gracefully")
 					a.Shutdown(a.runtimeConfig.GracefulShutdownDuration)
 					log.Fatal(e)
+					// This error is only caught by tests, since during normal execution we panic
+					return pipeline, errors.New("dapr panicked")
 				}
 				log.Error(e)
 				continue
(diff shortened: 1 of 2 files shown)


pkg/runtime/runtime.go
```diff
@@ -676,30 +676,39 @@ func (a *DaprRuntime) populateSecretsConfiguration() {
...
 		for i := 0; i < len(spec.Handlers); i++ {
 			middlewareSpec := spec.Handlers[i]
 			component, exists := a.getComponent(middlewareSpec.Type, middlewareSpec.Name)
 			if !exists {
-				err := fmt.Errorf(
-					"couldn't find middleware component with name %s and type %s/%s",
-					middlewareSpec.Name, middlewareSpec.Type, middlewareSpec.Version,
-				)
-				return httpMiddleware.Pipeline{}, err
+				log.Warn("error processing middleware component, daprd process will exit gracefully")
```
artursouza, line 686: Is there a valid scenario that sidecar should continue if cannot find the component? For example, can there be an app that is trying to use a middleware that might not be configured yet due to multiple versions of the app running and the component is not created yet.

ItalyPaleAle: I was wondering about whether there could be a valid scenario, but eventually I thought I shouldn't allow this because it could lead to unexpected situations. My thinking was that a situation where the Configuration references a middleware that doesn't exist is more likely than not a mistake and allowing Dapr to init anyways could cause errors to go unnoticed. For multiple versions, one could use multiple Configuration resources.

However i am happy to change this if you prefer.

artursouza: Can we not change the behavior unless it is part of the hotix?

ItalyPaleAle: Ok, I will change that behavior tomorrow morning

ItalyPaleAle: Done

artursouza (approved):

PR #5994, opened 2023-02-25
[release-1.10] Cherry-pick #5976 to reduce unit test flakiness

ItalyPaleAle:
**Because this is a cherry-pick, please DO NOT SQUASH**

Cherry-picks a commit into release-1.10 as this should reduce a lot of flakiness in the unit tests and greatly reduce our build failures for 1.10 hotfixes.

The changes in the commit were primarily in unit tests, with a small code change in something that's gated behind a feature flag, so risk should be small.

### DIFF
--- a/pkg/grpc/api_daprinternal.go
+++ b/pkg/grpc/api_daprinternal.go
@@ -153,7 +153,7 @@ func (a *api) CallLocalStream(stream internalv1pb.ServiceInvocation_CallLocalStr
 
 			// Read the next chunk
 			readErr = stream.RecvMsg(chunk)
-			if readErr == io.EOF {
+			if errors.Is(readErr, io.EOF) {
 				// Receiving an io.EOF signifies that the client has stopped sending data over the pipe, so we can stop reading
 				break
 			} else if readErr != nil {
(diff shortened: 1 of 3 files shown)


yaron2 (approved):

artursouza:
Cherry-pick does not impact merging with or without squashing.

artursouza:
Cherry-pick does not impact merging with or without squashing.

ItalyPaleAle:
> Cherry-pick does not impact merging with or without squashing.

It does... Because then if we try to merge release-1.10 into master, the commits are different. Squashing is rebasing, essentially. See: https://www.atlassian.com/git/tutorials/merging-vs-rebasing (section "The Golden Rule of Rebasing")

artursouza:
> > Cherry-pick does not impact merging with or without squashing.
> 
> It does... Because then if we try to merge release-1.10 into master, the commits are different. Squashing is rebasing, essentially. See: https://www.atlassian.com/git/tutorials/merging-vs-rebasing (section "The Golden Rule of Rebasing")

The link above has nothing to do with the point being discussed. We all know the difference between merging and rebasing. The point here is about cherry-pick being as good as "making the change again", or applying the same patch file. When merging OR rebasing, there will be conflicts. Let's discuss this offline.

PR #6037, opened 2023-03-08
Draft: More Placment test improvments

JoshVanL:
- Adds context and block to placement `newTestClient`
- Always cleanup servers in tests
- Log when placement service stopped


JoshVanL:
@ItalyPaleAle @daixiang0 This should also get us further

PR #6039, opened 2023-03-08
[release-1.10] Actor Reminders improvements (fixes race conditions)

### LINKED ISSUE #6018: Create/delete reminder throws "possible etag mismatch" after enabling resiliency
**Note:** If you have a general support question and are looking for a quicker response, please checkout our discord channel for answers from the community:
https://aka.ms/dapr-discord

## In what area(s)?

/area runtime

## Ask your question here

We are using `dapr` actors and our application is very heavy on actor reminders. After enabling `resiliency` feature we started experiencing a lot of exceptions with `error deleting actor reminder: possible etag mismatch` when creating and deleting reminders. We also see that these operations write into a single key `actors||ActorName||metadata` and since we are triggering many reminders from many actors that could cause `ETag` mismatch if used on state requests:
```
Dapr.DaprApiException: error deleting actor reminder: possible etag mismatch. 
error from state store: ERR Error running script @user_script:14:  failed to set key actors||ActorName||metadata
 at Dapr.Actors.DaprHttpInteractor.SendAsyncHandleUnsuccessfulResponse(Func`1 requestFunc, String relativeUri, CancellationToken cancellationToken)\n 
```

Is there any way we could solve this problem and still keep the resiliency feature enabled? Thanks!

ItalyPaleAle:
This PR is part of the work to improve the performance of the actor reminders and timers subsystem, and it includes a fix to use transactions that should address #6018.

- Fixes #6018 (for the most part):
   - Makes operations that save reminder partitions and metadata in the state store atomic (using Dapr state store transactions). This should help improving issues with multiple sidecars creating reminders in the same partition, as long as the state store supports rollback (ie. [not Redis](https://github.com/dapr/components-contrib/issues/2071)).
      - Users may still get failures due to etag errors, but it should then be safe to retry saving the reminder as the consistency of the data is preserved. Resiliency could help doing this automatically
   - Includes #6050 to fix possible race conditions when saving reminder partitions when the operation times out
- Update actor reminders data structures
  - Both reminders and actors are now keeping their state in a Reminders struct. Aside from improving consistency and reducing code duplication, this is needed to be able to implement a priority queue for reminders and actors (see part 2: #6040)
- Switch data field to be of type `json.RawMessage`
  - This means that when the request from the user is parsed, or the reminders are serialized before being sent to storage or de-serialized when retrieved from storage, the `data` field is not parsed into a Go type. Dapr doesn't care about the value of the data (it's opaque to us), so this helps saving CPU cycles and memory.
- Reduce tech debt and code duplication

This PR doesn't change any user-facing behavior and it's fully backwards- and forwards-compatible (i.e. it doesn't include changes that would break "mixed" clusters where some Dapr sidecars are on an older version)


ItalyPaleAle:
The "/ok-to-test" seems broken in the release branch due to missing scripts

artursouza (approved):

PR #6040, opened 2023-03-08
[WIP] Manage reminders and actors in-memory using a priority queue

ItalyPaleAle:
# TODO

- [x] This PR is built on top of #6039  - that needs to be merged first
- [X] Implement the data structure
- [ ] Update the reminders and timers code to use the new data structure

# Description

This is part 2 of the work on improving actor reminders and timers, and changes how the queue of reminders and timers is managed.

Currently, each reminder or timer scheduled in the runtime causes a new goroutine which has multiple timers inside and is blocked until one of the events happens (e.g. the reminder is due to be executed, it's canceled, etc). This means that there's at least 1 goroutine for each timer/reminder that is blocked and waiting, requiring significant amounts of memory (each goroutine comes with a bunch of memory allocated) and more work on the garbage collector.

This PR implements a new data structure that is essentially a priority queue (heap-based). When reminders/timers are scheduled, they are added to the queue in order of their execution time. By maintaining the queue with reminders in order, we can make sure that there's at most one background goroutine blocked, which is waiting on the next reminder/timer only.

### DIFF
--- a/pkg/actors/reminders/processor.go
+++ b/pkg/actors/reminders/processor.go
@@ -18,7 +18,7 @@ import (
 	"sync"
 	"sync/atomic"
 
-	clocklib "github.com/benbjohnson/clock"
+	kclock "k8s.io/utils/clock"
 )
 
 var ErrProcessorStopped = errors.New("processor is stopped")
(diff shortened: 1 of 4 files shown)


ItalyPaleAle:
Closing in favor of #6716

PR #6046, opened 2023-03-08
Updated raft tests + fixed raft logger

ItalyPaleAle:
Some more attempts at stabilizing Raft unit tests which are still flaky

Also fixed the Raft logger adapter, so it stops printing out lots of `%EXTRA`. Their methods don't work like ours (which are like fmt.Sprintf), but instead are used to pass key-value pairs.


mukundansundar:
@ItalyPaleAle can you please fix the go mod check diff error?

ItalyPaleAle:
Done @mukundansundar

mukundansundar (approved):

PR #6050, opened 2023-03-09
Cascading contexts in actor reminder state store methods

ItalyPaleAle:
# Description

This PR alleviates some of the race conditions described in https://github.com/dapr/dapr/issues/6018#issuecomment-1461109891 , for the second kind (in-process).

TLDR: There are race conditions in the methods that store reminders in the state store.

It doesn't fully solve the issue, because:

- Calls to the state store are still in sequence and not in a transaction. We cannot make all state store operations happen in a transaction because we need to get the etag from the first operation and write that into the body of the second value.
- If the context is canceled mid-way while data is being stored, we rely on the state store to respect that and abort the operation. Although we do that as a best effort, that cannot be guaranteed.

Additionally, it expands the information returned in error logs to be able to pinpoint to issues better.

## Issue reference

Alleviates some issues in #6018


ItalyPaleAle:
In draft for now as it's included in #6039

ItalyPaleAle:
Closing as this PR is now part of #6039

PR #6054, opened 2023-03-09
Replaces `github.com/benbjohnson/clock` with `k8s.io/utils/clock`.

JoshVanL:
This PR does 2 things: replacing the use of `github.com/benbjohnson/clock` with `k8s.io/utils/clock`, and de-flaking the `pkgs/actors` unit tests.

When working on the actors unit tests, I found the k8s clock implementation to be more reliable. We can also assume that the k8s implementation to be maintained and exist on GH longer into the future given it's under the k8s project and used extensively there.

---

The actor tests now rely significantly less on wall time, and instead use the fake clock as much as possible. This means "eventually" style checks and assertions are more deterministic and reliable. We also get a big speed up, with the new tests taking about 10% of the time (when the tests actually pass!)

```
go test --race -v --count 1 ./pkg/actors  5.06s user 0.76s system 24% cpu 23.470 total
```

```
go test --race -v --count 1 ./pkg/actors  3.65s user 0.73s system 217% cpu 2.018 total
```

~Tests are now labelled for parallelism.~

Have removed the test `Parallel()`s since the Windows runners were having trouble with them.

/cc @ItalyPaleAle

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -132,6 +132,7 @@ linters-settings:
       - "github.com/cenkalti/backoff": "must use github.com/cenkalti/backoff/v4"
       - "github.com/cenkalti/backoff/v2": "must use github.com/cenkalti/backoff/v4"
       - "github.com/cenkalti/backoff/v3": "must use github.com/cenkalti/backoff/v4"
+      - "github.com/benbjohnson/clock": "must use k8s.io/utils/clock"
   misspell:
     # Correct spellings using locale preferences for US or UK.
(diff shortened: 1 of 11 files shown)


berndverst (changes requested):
Please add `github.com/benbjohnson/clock` to the `depguard` section of `.golanci.yml` to prevent anyone from using this package again. This way the linter will throw an error!

JoshVanL:
Thanks @berndverst, done!

artursouza (approved):

Issue #6031, opened 2023-03-07
Support Min TLS Version via Configuration


DeepanshuA:
## In what area(s)?

> /area runtime

## Describe the feature
There seems a possibility that the established TLS connection may be handshaked below TLS 1.2, at certain instances in Dapr.
Organizations trying to use Dapr in Production Code, usually verify services through some framework where-in it tries to breach some security requirements. These frameworks reports if able to breach some minimum supported connections etc.

We can provide a configuration in Dapr, so that at start-up, it can take a MinTLSVersion, which let's say if provided as `VersionTLS12`, then `MinVersion` can be supplied while handshaking for a tls connection via crypto/tls.

## Release Note

RELEASE NOTE:

tmacam:
Does it make sense extend this to not only limit the minimum TLS version but additionally to also limit the cipher suite available? Or should this be a distinct issue?

Issue #3565, opened 2021-08-19
How can I use HTTPS to communicate with the Dapr Sidecar?


fbridger:
## In what area(s)?

/area runtime

> /area operator

> /area placement

> /area docs

> /area test-and-release

## Ask your question here

How can I communicate with the Dapr Sidecar using a secure communication channel (HTTPS)?

I would like to be able to redirect all traffic coming from our API Gateway to the Dapr sidecars using a secure communication channel as described in the diagram. Is this possible? Or what is the correct way of doing this?

![Sidecar](https://user-images.githubusercontent.com/6483000/130057294-5ecea8ec-8855-4f21-abf4-1929a73b3c04.png)

CodeMonkeyLeet:
@fbridger Have you taken a look at the docs on [integration of APIM with Dapr](https://docs.dapr.io/developing-applications/integrations/azure-api-management/)? 

More generically dapr sidecar supports HTTPS communication with the app with `app-ssl` dapr CLI parameter or kubernetes annotation. See [Dapr arguments and annotations for daprd, CLI, and Kubernetes](https://docs.dapr.io/reference/arguments-annotations-overview/).

fbridger:
Hi @CodeMonkeyLeet thanks for your reply. 

I know DAPR Sidecar supports HTTPS communication with the Application. 

However, the question is related to why the Dapr Sidecar only supports HTTP for **INCOMING** communications. It seems this is not currently supported and I wanted to understand if this was by design.

My main goal is to be able to secure all in-transit communication using a secure channel. Does this makes sense?

jjcollinge:
@fbridge just so that I understand the context of the question - are you running Azure API Management in [self hosted mode](https://docs.microsoft.com/en-us/azure/api-management/how-to-deploy-self-hosted-gateway-kubernetes) on Kubernetes or are you trying to expose the Dapr HTTP API directly to a cloud hosted Azure API Management service?

fbridger:
Well, actually we are deploying our services to AWS. So we would be using AWS API Gateway and running them in ECS without Kubernetes. 

So my intial intent or thought is to route all incoming traffic (from AWS API Gateway) through the Dapr Sidecar => Application to get Distributed Tracing information.

jjcollinge:
[...]

PR #6078, opened 2023-03-15
Set Minimum TLS version to 1.2 for AppChannel's HTTP and gRPC clients

shubham1172:
# Description

Enforce minimum TLS version to 1.2 when Dapr talks to Application over HTTP or gRPC. This is enabled by-default and will break insecure scenarios. Users can still opt-out using a feature flag, however the feature flag will be removed in 2 releases.

## Issue reference

Related issues: https://github.com/dapr/dapr/issues/3565, https://github.com/dapr/dapr/issues/6031

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing


shubham1172:
Question: should be expose this behind a feature flag to allow users to opt-out of TLS 1.2 enforcement? We should remove the feature flag in a few releases then. Or since it's a security issue, are we ok breaking users and forcing them to either be secure (by upgrading to TLS 1.2), or be **intentionally** insecure (by removing --app-ssl flag)?

ItalyPaleAle:
We discussed yesterday in the maintainers call that:

1. We will make TLS >= 1.2 the default. This is a breaking change and we're ok with that.
2. We will add a **temporary** feature flag  (for example `AllowInsecureTLS` - I like that it says "insecure" :) ) for users who want to restore the behavior of allowing TLS 1.0 and 1.1
3. The feature flag will be removed in Dapr 1.13

shubham1172:
Perfect, thank you! I will make these changes @ItalyPaleAle.

pkg/channel/http/http_channel.go
```diff
@@ -68,13 +68,16 @@ type Channel struct {
 // CreateLocalChannel creates an HTTP AppChannel
 //
 //nolint:gosec
-func CreateLocalChannel(port, maxConcurrency int, pipeline httpMiddleware.Pipeline, spec config.TracingSpec, sslEnabled bool, maxRequestBodySizeMB, readBufferSizeKB int) (channel.AppChannel, error) {
+func CreateLocalChannel(port, maxConcurrency int, pipeline httpMiddleware.Pipeline, spec config.TracingSpec, sslEnabled bool, maxRequestBodySizeMB, readBufferSizeKB int, allowInsecureTLS bool) (channel.AppChannel, error) {
```
ItalyPaleAle, line 71: Not gonna block the PR due to this, but if you get a chance can you please convert this interface to accept a single struct with options, rather than 8 arguments? It becomes hard to even read what the invocations are doing.

shubham1172: Good idea, let me refactor in this PR itself.

ItalyPaleAle (approved):

ItalyPaleAle (approved):

mukundansundar (approved):
lgtm.
@dapr/maintainers-dapr Any concerns with merging this breaking change for min TLS for 1.11?

mukundansundar:
> lgtm.
> @dapr/maintainers-dapr Any concerns with merging this breaking change for min TLS for 1.11?

based on offline sync there is agreement on this PR being merged.

PR #6086, opened 2023-03-16
[release-1.10] API Allowlist should not apply to gRPC proxying

### LINKED ISSUE #6085: gRPC proxying broken in 1.10 when API allowlists are used
## In what area(s)?

/area runtime

## What version of Dapr?

1.10.x

## Description

When API allowlists are enabled, gRPC proxying doesn't work in Dapr 1.10 because all calls fail as not allowed. This is a regression introduced in Dapr 1.10

ItalyPaleAle:
Fixes #6085

Added unit tests

### DIFF
--- a/docs/release_notes/v1.10.4.md
+++ b/docs/release_notes/v1.10.4.md
@@ -1,0 +1,23 @@
+# Dapr 1.10.4
+
+This hotfix release contains two bug fixes.
+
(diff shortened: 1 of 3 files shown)


yaron2:
@ItalyPaleAle please add release notes.

ItalyPaleAle:
@yaron2 done

yaron2 (approved):

PR #6107, opened 2023-03-19
Fixed workflow APIs not working in master branch

ItalyPaleAle:
Fixes a regression introduced in a recent PR that converted the workflows APIs to universal. The map of workflow components was not set, so HTTP and gRPC APIs would not work.

### DIFF
--- a/pkg/grpc/api.go
+++ b/pkg/grpc/api.go
@@ -90,7 +90,6 @@ type api struct {
 	appChannel                 channel.AppChannel
 	resiliency                 resiliency.Provider
 	stateStores                map[string]state.Store
-	workflowComponents         map[string]workflows.Workflow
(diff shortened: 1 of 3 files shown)


ItalyPaleAle:
We probably need to enable some E2E tests for Workflows APIs as errors like these should have been caught in E2E tests.

daixiang0 (approved):

yaron2:
> Fixes a regression introduced in a recent PR that converted the workflows APIs to universal. The map of workflow components was not set, so HTTP and gRPC APIs would not work.

Can you link to the PR that introduced the regression please?

mukundansundar (approved):

yaron2:
> We probably need to enable some E2E tests for Workflows APIs as errors like these should have been caught in E2E tests.

cc @cgillum @RyanLettieri

PR #6108, opened 2023-03-19
Updated dependencies

ItalyPaleAle:
Updates a number of dependencies to their latest version, also to make Dependabot stop wasting CI minutes with all the open PRs (and most importantly, spam us every time something gets merged :) )

Also pins latest components-contrib and kit


ItalyPaleAle:
@mukundansundar it is not.
We were using 1.1.5 which has been withdrawn (that's how I discovered the error) and renamed to v2. Change log indicates that an upstream dependency of theirs had a breaking change. But it shouldn't alter the behavior for us because we were already using 1.1.5

daixiang0:
> Updates a number of dependencies to their latest version, also to make Dependabot stop wasting CI minutes with all the open PRs (and most importantly, spam us every time something gets merged :) )
> 
> Also pins latest components-contrib and kit

Maybe we can set Dependabot run every month or just upgrade when bigfix?

ItalyPaleAle:
> Maybe we can set Dependabot run every month or just upgrade when bigfix?

I am not a fan of dependabot at all, for various reasons, and I would always vote for manually updating deps every once in a while (like we do in components-contrib). But that's not my choice.

ItalyPaleAle:
@daixiang0 

> Maybe we can set Dependabot run every month or just upgrade when bigfix?

I am not a fan of dependabot at all, for various reasons, and I would always vote for manually updating deps every once in a while (like we do in components-contrib). But that's not my choice.

Anyways, if you don't have concerns would you mind merging this? So at least Dependabot stops spamming :)

mukundansundar (approved):

PR #6112, opened 2023-03-21
[release-1.10] Maintain all HTTP headers with the same name

### LINKED ISSUE #6104: Method invocation not forwarding multiple "Set-Cookie" headers from response
Hi,

I'm noticing this behavior with method invocation, when the response has multiple "Set-Cookie" headers, Dapr does not seem to propagate all of them but just one. 

Can anyone confirm this issue, as there's little chance this was not spotted before, so wondering if it's just my setup.

```
dapr --version
CLI version: 1.10.0
Runtime version: 1.10.4
OS: Linux
```

Running a simple service with an endpoint that sets multiple cookie values:
`dapr run --app-id myservice --dapr-http-port 7010 --app-port 7000  -- dotnet run -p src/Cookies.csproj --urls http://localhost:7000`

Making a request to the service directly. Multiple Set-Cookie headers:
```
curl -D - "http://localhost:7000/cookietest"
HTTP/1.1 200 OK
Content-Length: 0
Date: Sat, 18 Mar 2023 13:08:52 GMT
Server: Kestrel
Set-Cookie: cookie1=value1; path=/
Set-Cookie: cookie2a=value2a; path=/
Set-Cookie: cookie2b=value2b; path=/
X-PT-TRACEID: 0HMP7M76HM7I8:00000002
```

Dapr method invocation of same endpoint. Only 1 Set-Cookie header value propagated:
```
curl -D - 'http://localhost:7010/v1.0/invoke/myservice/method/cookietest'
HTTP/1.1 200 OK
Server: Kestrel
Date: Sat, 18 Mar 2023 13:08:19 GMT
Content-Type: text/plain; charset=utf-8
Content-Length: 0
X-Pt-Traceid: 00-0ab88c0b2557aed579cff3aa6c9f2f50-c64a02f6592f0081-01
Traceparent: 00-0ab88c0b2557aed579cff3aa6c9f2f50-c64a02f6592f0081-01
Set-Cookie: cookie1=value1; path=/
```

RFC6265 Set-Cookie 
https://datatracker.ietf.org/doc/rfc6265/
```
   An origin server can include multiple
   Set-Cookie header fields in a single response.  The presence of a
   Cookie or a Set-Cookie header field does not preclude HTTP caches
   from storing and reusing a response.

   Origin servers SHOULD NOT fold multiple Set-Cookie header fields into
   a single header field.  The usual mechanism for folding HTTP headers
   fields (i.e., as defined in [RFC2616]) might change the semantics of
   the Set-Cookie header field because the %x2C (",") character is used
   by Set-Cookie in a way that conflicts with such folding.
```

ItalyPaleAle:
# Description

The HTTP specs allow passing multiple headers with the same name, for example multiple `Set-Cookie` headers. However, for every header, Dapr was only preserving the first value.

This fixes the behavior of Dapr to preserve all HTTP headers, even when more than one have the same name

## Issue reference

Fixes #6104


ItalyPaleAle:
Rebased on release-1.10 per maintainers.

pkg/channel/http/http_channel.go
```diff
@@ -134,10 +134,8 @@ func (h *Channel) GetAppConfig() (*config.ApplicationConfig, error) {
 	// Get versioning info, currently only v1 is supported.
 	headers := resp.Headers()
 	var version string
-	if val, ok := headers["dapr-app-config-version"]; ok {
-		if len(val.Values) == 1 {
-			version = val.Values[0]
-		}
+	if val, ok := headers["dapr-app-config-version"]; ok && len(val.Values) > 0 {
+		version = val.Values[0]
```
mukundansundar, line 138: shouldn't the last value of the array be used here?

pkg/http/api.go
```diff
@@ -2098,7 +2093,7 @@ func (a *api) onDirectActorMessage(reqCtx *fasthttp.RequestCtx) {
 	}
 	defer resp.Close()
 
-	invokev1.InternalMetadataToHTTPHeader(reqCtx, resp.Headers(), reqCtx.Response.Header.Set)
+	invokev1.InternalMetadataToHTTPHeader(reqCtx, resp.Headers(), reqCtx.Response.Header.Add)
```
mukundansundar, line 2096: nitpick: add a comment why `Add` is being used...

utils/fasthttpadaptor/request.go
```diff
@@ -25,8 +25,12 @@ func ConvertRequest(ctx *fasthttp.RequestCtx, r *http.Request, forServer bool) e
...
 
 	r.Method = b2s(ctx.Method())
-	r.Proto = "HTTP/1.1"
-	r.ProtoMajor = 1
+	r.Proto = b2s(ctx.Request.Header.Protocol())
+	if r.Proto == "HTTP/2" {
+		r.ProtoMajor = 2
+	} else {
+		r.ProtoMajor = 1
+	}
```
mukundansundar, line 33: why is this needed?

ItalyPaleAle: Backported from https://github.com/valyala/fasthttp/blob/master/fasthttpadaptor/request.go which had changed since we took the code

pkg/http/api.go
ItalyPaleAle: Done

artursouza:
I would approve this PR but all E2E tests failed.

ItalyPaleAle:
@artursouza it's temporal that failed to set up (again). I'll re-trigger

PR #6115, opened 2023-03-21
Fix: add missing json_name to Workflow protos

ItalyPaleAle:
When the workflow API was changed to Universal, the `json_name` properties were not added to the fields in the protos, potentially causing inconsistent JSON serialization.


dapr/proto/runtime/v1/dapr.proto
```diff
@@ -680,13 +678,13 @@ message UnlockResponse {
...
-  string instance_id = 1;
+  string instance_id = 1 [json_name = "instanceId"];
 }
 
 message GetWorkflowRequest {
-  string instance_id  = 1;
-  string workflow_name = 2;
-  string workflow_component = 3;
+  string instance_id  = 1 [json_name = "instanceId"];
+  string workflow_component = 2 [json_name = "workflowComponent"];
```
ItalyPaleAle, line 686: Reviewers: I swapped workflow_component and workflow_name so in all messages, component is 2 and name is 3. This is technically a breaking change, but the entire API underwent significant breaking changes already.

mukundansundar: is there a reason for this switch?

ItalyPaleAle: Consistency because everywhere else name was 3 and component was 2. And this is a good time to add a breaking change since there's a bunch already.

yaron2 (approved):

yaron2:
The consumers of Workflows are SDKs and this is an Alpha API so I'm not concerned about frequent proto changes at this stage

Issue #3549, opened 2021-08-13
[Proposal] Distributed Lock API design


seeflood:
Note: this proposal has been updated (as v0.2) after the review and discusion on the first version(as v0.1). 

Here is the first version (as v0.1) of this proposal. There are some modifications in [v0.2](https://github.com/dapr/dapr/issues/3549#issuecomment-984755423)

# In what area(s)?

 /area runtime
# 0. tl;dr
This proposal try to add TryLock and Unlock API.

The Lock Renewal API might be controversial and will not be added into the first version

# 1. Why it is needed
Application developers need distributed locks to keep their data safe from race conditions,but implementing a distributed lock correctly is challenging,since you have to be careful to prevent deadlock or some incorrect corner cases from happening.

An easy-to-use distributed lock API provided by runtime sidecar can be helpful. Actually we have already implemented this API in [Layotto](https://github.com/mosn/layotto/blob/main/spec/proto/runtime/v1/runtime.proto) ,and hope to contribute it to Dapr. 

# 2. Evaluation of products on the market
| **System** | **tryLock(non-blocking lock)** | **Blocking lock(based on watch)** | **Availability** | **Write operations are linearizable** | **sequencer([chubby's feature](https://static.googleusercontent.com/media/research.google.com/zh-TW//archive/chubby-osdi06.pdf))** | **Lock renewal** |
| --- | --- | --- | --- | --- | --- | --- |
| Stand-alone redis | yes | x | unavailable when single failure | yes | yes(need poc) | yes |
| redis cluster | yes | x | yes | no. Locks will be unsafe when fail-over happens | yes(need poc) | yes |
| redis Redlock | yes | | | | | |
| consul | yes | | | | | |
| zookeeper | yes | yes | yes. [the election completes within 200 ms](https://pdos.csail.mit.edu/6.824/papers/zookeeper.pdf) | yes | yes use zxid as sequencer | yes |
| etcd | yes | yes | yes | yes | yes use revision | yes lease.KeepAlive |

There are some differences in feature supporting.

# 2. High-level design
### 2.1.0. Design principles
We are faced with many temptations. In fact, there are many lock related features that can be supported (blocking locks, reentrant locks, read-write locks, sequencer, etc.)

But after all, our goal is to design a general API specification, so we should be as conservative as possible in API definition.Start simple, abstract the simplest and most commonly used functions into API specifications, and wait for user feedback before considering adding more abstraction into API specification.

### 2.1.1. TryLock/Unlock API
[...]

PR #6116, opened 2023-03-21
Convert lock APIs to Universal and fix inconsistencies

ItalyPaleAle:
This is another API conversion to Universal (single implementation of business logic in gRPC, that is exposed as HTTP API too), with the usual goals:

- Have less APIs implemented in HTTP so the migration to net/http will be simpler (#4979)
- Have less code to maintain in the future
- Ensure gRPC and HTTP APIs are consistent
  - In this case, the HTTP APIs were using Resiliency, while gRPC ones were not. They both use Resiliency now.


pkg/grpc/universalapi/api_lock.go
```diff
@@ -0,0 +1,162 @@
...
+	"github.com/dapr/components-contrib/lock"
+	lockLoader "github.com/dapr/dapr/pkg/components/lock"
+	"github.com/dapr/dapr/pkg/messages"
+	runtimev1pb "github.com/dapr/dapr/pkg/proto/runtime/v1"
+	"github.com/dapr/dapr/pkg/resiliency"
+)
+
+func (a *UniversalAPI) TryLockAlpha1(ctx context.Context, req *runtimev1pb.TryLockRequest) (*runtimev1pb.TryLockResponse, error) {
+	// 1. validate and find lock component
+	if req.ExpiryInSeconds <= 0 {
```
mukundansundar, line 28: completely unrelated to this PR... ideally having data on the overhead i.e. perf numbers for this API is a must before making it stable.

pkg/http/api_test.go
```diff
@@ -2894,14 +2898,14 @@ func TestV1Alpha1DistributedLock(t *testing.T) {
 		b, _ := json.Marshal(&req)
 
 		resp := fakeServer.DoRequest("POST", apiPath, b, nil)
-		assert.Equal(t, 500, resp.StatusCode)
+		assert.Equal(t, 400, resp.StatusCode)
```
mukundansundar, line 2901: should this be `403 Forbidden` instead?

pkg/http/api_test.go
```diff
@@ -2876,14 +2880,14 @@ func TestV1Alpha1DistributedLock(t *testing.T) {
...
```
mukundansundar, line 2883: should this be `404` instead?

pkg/http/api_test.go
```diff
@@ -2948,17 +2952,15 @@ func TestV1Alpha1DistributedLock(t *testing.T) {
...
 		// assert
-		assert.NotNil(t, resp.JSONBody)
-		rspMap := resp.JSONBody.(map[string]interface{})
-		assert.NotNil(t, rspMap)
-		assert.Equal(t, float64(3), rspMap["status"])
+		assert.Contains(t, string(resp.RawBody), "ERR_MALFORMED_REQUEST")
+		assert.Contains(t, string(resp.RawBody), "ResourceId is empty in lock store store1")
 	})
 
 	t.Run("Unlock with invalid resource id that returns 500", func(t *testing.T) {
```
mukundansundar, line 2962: shouldn't this also return `404`?

pkg/messages/predefined.go
```diff
@@ -111,6 +104,15 @@ var (
...
 
+	// Lock.
+	ErrLockStoresNotConfigured    = APIError{"lock store is not configured", "ERR_LOCK_STORE_NOT_CONFIGURED", http.StatusInternalServerError, grpcCodes.FailedPrecondition}
+	ErrResourceIDEmpty            = APIError{"ResourceId is empty in lock store %s", "ERR_MALFORMED_REQUEST", http.StatusBadRequest, grpcCodes.InvalidArgument}
+	ErrLockOwnerEmpty             = APIError{"LockOwner is empty in lock store %s", "ERR_MALFORMED_REQUEST", http.StatusBadRequest, grpcCodes.InvalidArgument}
+	ErrExpiryInSecondsNotPositive = APIError{"ExpiryInSeconds is not positive in lock store %s", "ERR_MALFORMED_REQUEST", http.StatusBadRequest, grpcCodes.InvalidArgument}
+	ErrLockStoreNotFound          = APIError{"lock store %s not found", "ERR_LOCK_STORE_NOT_FOUND", http.StatusBadRequest, grpcCodes.InvalidArgument}
+	ErrTryLockFailed              = APIError{"failed to try acquiring lock: %s", "ERR_TRY_LOCK", http.StatusInternalServerError, grpcCodes.Internal}
+	ErrUnlockFailed               = APIError{"failed to release lock: %s", "ERR_UNLOCK", http.StatusInternalServerError, grpcCodes.Internal}
+
```
mukundansundar, line 115: why are all http status codes bad request or ISE?
same for gRPC, we can have corresponding codes?

mukundansundar:
Also a breaking change with the existing PR HTTP status code changes?

pkg/http/api_test.go
ItalyPaleAle: Agree not 400, but maybe 409 (Conflict)

pkg/http/api_test.go
```diff
@@ -2911,14 +2915,14 @@ func TestV1Alpha1DistributedLock(t *testing.T) {
 		b, _ := json.Marshal(&req)
 
 		resp := fakeServer.DoRequest("POST", apiPath, b, nil)
-		assert.Equal(t, 500, resp.StatusCode)
+		assert.Equal(t, 400, resp.StatusCode)
```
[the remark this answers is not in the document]
ItalyPaleAle: For invalid data, 400 seems appropriate.

pkg/http/api_test.go
mukundansundar: If it's a missing parameter I agree... Please do update the test name

pkg/http/api_test.go
mukundansundar: Isn't 409 more related to version conflicts? 
In this case it's someone else owning the lock right?

pkg/http/api_test.go
ItalyPaleAle: This was in the original API design, and I didn't change it.

I agree with you that this is not the correct status code, but changing that would require reworking the entire APIs down to the components. Right now, they return an error in all cases where a lock can't be released.

I'll open an issue to track that.

ItalyPaleAle: https://github.com/dapr/dapr/issues/3549#issuecomment-1487129276

pkg/http/api_test.go
```diff
@@ -2975,11 +2977,11 @@ func TestV1Alpha1DistributedLock(t *testing.T) {
 	})
 
 	t.Run("Unlock with invalid owner", func(t *testing.T) {
```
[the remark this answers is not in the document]
ItalyPaleAle: Same as above. I did not change this and doing this change is a fairly large breaking thing (and requires changing code down the stack).

ItalyPaleAle:
@mukundansundar Per comments above, I agree that the status codes right now aren't always perfect. For example, TryLock and Unlock do not distinguish between a lock that doesn't exist or an internal error. The original APIs returned an ISE in both cases. 

I have tracked that here: https://github.com/dapr/dapr/issues/3549#issuecomment-1487129276

I think that those should be addressed as part of the efforts to stabilize this API, but likely out of scope here as they even require changes in components-contrib to return the correct errors.

Most status codes here haven't changed, with some rare exceptions for bad requests. I don't think that's a breaking change we should be concerned with much because they're still error status codes, and these APIs aren't stable anyways.

mukundansundar:
> Most status codes here haven't changed, with some rare exceptions for bad requests. I don't think that's a breaking change we should be concerned with much because they're still error status codes, and these APIs aren't stable anyways.

Are there SDK changes needed for any specific request that have status code changes?

mukundansundar (approved):

mukundansundar:
> I believe most (all?) SDKs use gRPC, so this wouldn't impact them. But I'm not sure.

Seeing that no handcrafted SDK for now have HTTP APIs implemented.

yaron2:
> > I believe most (all?) SDKs use gRPC, so this wouldn't impact them. But I'm not sure.
> 
> Seeing that no handcrafted SDK for now have HTTP APIs implemented.

PHP does

mukundansundar:
> > > I believe most (all?) SDKs use gRPC, so this wouldn't impact them. But I'm not sure.
> > 
> > 
> > Seeing that no handcrafted SDK for now have HTTP APIs implemented.
> 
> PHP does

specifically I am not seeing a lock api implementation in PHP SDK.

Issue #5425, opened 2022-10-28
Add E2E tests for configuration building block APIs


pravinpushkar:
## In what area(s)?

/area runtime

PR #6128, opened 2023-03-24
E2e test for configuration api

shivamkm07:
# Description

This PR adds the e2e test for configuration API

## Issue reference

Please reference the issue this PR will close: #5425 

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing

### DIFF
--- a/tests/apps/configurationapp/app.go
+++ b/tests/apps/configurationapp/app.go
@@ -1,0 +1,274 @@
+/*
+Copyright 2021 The Dapr Authors
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package main
+
+import (
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"log"
+	"net/http"
+	"os"
+	"strconv"
+	"strings"
+	"sync"
+
+	"github.com/dapr/components-contrib/configuration"
+	"github.com/dapr/components-contrib/tests/utils/configupdater"
+	cu_redis "github.com/dapr/components-contrib/tests/utils/configupdater/redis"
+	"github.com/dapr/dapr/tests/apps/utils"
+	"github.com/dapr/kit/logger"
+	"github.com/gorilla/mux"
+)
+
+const (
+	daprHost        = "localhost"
+	daprHTTPPort    = "3500"
+	configStoreName = "configstore"
+	redisHost       = "dapr-redis-master.dapr-tests.svc.cluster.local:6379"
+)
+
+var (
+	appPort         = 3000
+	lock            sync.Mutex
+	receivedUpdates map[string][]string
+	updater         configupdater.Updater
+)
+
+type testCommandRequest struct {
+	Message string `json:"message,omitempty"`
+}
+
+type appResponse struct {
+	Message   string `json:"message,omitempty"`
+	StartTime int    `json:"start_time,omitempty"`
+	EndTime   int    `json:"end_time,omitempty"`
+}
+
+type receivedMessagesResponse struct {
+	ReceivedUpdates []string `json:"received-messages"`
+}
+
+func init() {
+	p := os.Getenv("PORT")
+	if p != "" && p != "0" {
+		appPort, _ = strconv.Atoi(p)
+	}
+	receivedUpdates = make(map[string][]string)
+}
+
+// indexHandler is the handler for root path
+func indexHandler(w http.ResponseWriter, r *http.Request) {
+	log.Println("indexHandler is called")
+
+	w.WriteHeader(http.StatusOK)
+	json.NewEncoder(w).Encode(appResponse{Message: "OK"})
+}
+
+// sendResponse returns response with status code and message
+func sendResponse(w http.ResponseWriter, statusCode int, message string) {
+	w.WriteHeader(statusCode)
+	json.NewEncoder(w).Encode(appResponse{Message: message})
+}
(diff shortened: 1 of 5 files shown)


tests/apps/configurationapp/app.go
```diff
@@ -0,0 +1,274 @@
...
+	"strconv"
+	"strings"
+	"sync"
+
+	"github.com/dapr/components-contrib/configuration"
+	"github.com/dapr/components-contrib/tests/utils/configupdater"
+	cu_redis "github.com/dapr/components-contrib/tests/utils/configupdater/redis"
+	"github.com/dapr/dapr/tests/apps/utils"
+	"github.com/dapr/kit/logger"
+	"github.com/gorilla/mux"
```
ItalyPaleAle, line 32: Since this is a new test, could you switch to Chi right away? gorilla/mux is unmaintained and we'll have to migrate to Chi in the not-too-distant future

tests/apps/configurationapp/app.go
```diff
@@ -0,0 +1,274 @@
...
+// getKeyValues is the handler for getting key-values from config store
+func getKeyValues(w http.ResponseWriter, r *http.Request) {
+	var keys []string
+	err := json.NewDecoder(r.Body).Decode(&keys)
+	if err != nil {
+		sendResponse(w, http.StatusBadRequest, err.Error())
+		return
+	}
+	url := "http://" + daprHost + ":" + daprHTTPPort + "/v1.0-alpha1/configuration/" + configStoreName + buildQueryParams(keys)
+	resp, err := http.Get(url)
```
ItalyPaleAle, line 94: Please don't use the default HTTP client, which has no timeout and could make the test hang. I belive we have some test utils that perform HTTP requests; otherwise, please just allocate a new, global http.Client

tests/apps/configurationapp/app.go
```diff
@@ -0,0 +1,274 @@
...
+		ret = ret + "&key=" + keys[i]
+	}
+	return ret
+}
+
+func subscribeToConfigUpdates(keys []string) (string, error) {
+	url := "http://" + daprHost + ":" + daprHTTPPort + "/v1.0-alpha1/configuration/" + configStoreName + "/subscribe" + buildQueryParams(keys)
+	resp, err := http.Get(url)
+	if err != nil {
+		return "", fmt.Errorf("Error subscribing config updates: %s", err)
```
ItalyPaleAle, line 119: In Go, errors always start lowercase (i'm surprised the linter hasn't complained!)


```suggestion
		return "", fmt.Errorf("error subscribing config updates: %w", err)
```

tests/apps/configurationapp/app.go
```diff
@@ -0,0 +1,274 @@
...
+	}
+	defer resp.Body.Close()
+	sub, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return "", fmt.Errorf("Error reading subscription Id: %s", err)
```
ItalyPaleAle, line 124: Same:

```suggestion
		return "", fmt.Errorf("error reading subscription Id: %w", err)
```

tests/apps/configurationapp/app.go
```diff
@@ -0,0 +1,274 @@
...
+	err := json.NewDecoder(r.Body).Decode(&component)
+	if err != nil {
+		sendResponse(w, http.StatusBadRequest, err.Error())
+		return
+	}
+	log.Printf("Initializing updater with component: %s\n", component)
+	switch component {
+	case "redis":
+		updater = cu_redis.NewRedisConfigUpdater(logger.NewLogger("configuration.redis"))
+		err := updater.Init(map[string]string{
```
ItalyPaleAle, line 219: I think E2E tests should not depend on Dapr code (incl. components-contrib) directly. The purpose of E2E tests is to validate the experience… E2E, which means going through the Dapr sidecar for everything.

ItalyPaleAle (changes requested):

tests/apps/configurationapp/app.go
shivamkm07: Done

tests/apps/configurationapp/app.go
shivamkm07: That is the ideal way agreed. But in configuration API, dapr sidecar just has read access to the config store. So there is no add/update/delete api exposed through dapr and that's why we need the updater interface for updating the key-value pairs.

ItalyPaleAle: Please, let's find another way to do that, even if that involves connecting to Redis directly.

I would really, really prefer to avoid adding another circular dependency, as that will cause headaches down the road :)

shivamkm07: Done. Please check @ItalyPaleAle. Thanks

tests/e2e/configuration/configuration_test.go
```diff
@@ -0,0 +1,287 @@
...
+	return keys
+}
+
+func testGet(t *testing.T, appExternalUrl string) {
+	updateUrl := fmt.Sprintf("http://%s/update-key-values", appExternalUrl)
+	items := generateKeyValues(10, v1)
+	itemsInBytes, _ := json.Marshal(items)
+	resp, statusCode, err := utils.HTTPPostWithStatus(updateUrl, itemsInBytes)
+	require.NoError(t, err, "error updating key values")
+	require.Equal(t, 200, statusCode, "expected statuscode 200, got %d. Error: %s", statusCode, string(resp))
```
ItalyPaleAle, line 170: When you put a `%` in the format and use `require/assert`, you must add `f` at the end of the function. For example: `require.Equalf`. Oterhwise it doesn't use sprintf internally and logs come out with some really weird formatting

ItalyPaleAle (commented):
Thanks for changing this to remove the cyclical dependency in the app. One more comment, but otherwise it's looking good.

Will gRPC be tested in a separate app?

shivamkm07:
> Thanks for changing this to remove the cyclical dependency in the app. One more comment, but otherwise it's looking good.
> 
> Will gRPC be tested in a separate app?

I have included the GRPC tests as well in the same PR. Please review. Thanks for noticing.

ItalyPaleAle (approved):

PR #6130, opened 2023-03-24
[release-1.10] Fixed large goroutine leak when connection to placement service fails

ItalyPaleAle:
This PR fixes an issue reported by users when daprd can't connect to the Placement service. If the placement service is defined via `--placement-host-address` but isn't available, or if a connection is dropped and then there's a transient failure in reconnecting, there's a very large goroutine leak. In our tests, we measured 5,000 goroutines leaked after ~10 minutes, eventually causing daprd to crash because it runs out of memory.

The root cause was identified as a missing `conn.Close()` after `connectToServer()` (which connects to the placement service) returns with an error.

Also includes two more tweaks:

1. Use an exponential backoff when reconnecting to the placement service rather than a linear one
2. Do not print a debug log on every single reconnection attempt, which were very frequent and causing noise


yaron2:
Why is this targeting release 1.10?

JoshVanL:
I have coincidentally also opened https://github.com/dapr/dapr/pull/6131

Should we merge the two? Or happy to rebase after this has merged.

ItalyPaleAle:
It depends.

@yaron2  do you want this in the release-1.10 branch (for whatever the next hotfix is), or master?

yaron2:
> It depends.
> 
> @yaron2 do you want this in the release-1.10 branch (for whatever the next hotfix is), or master?

Yes let's put it in 1.10

ItalyPaleAle:
@JoshVanL then I think your PR can be reviewed and merged independently from this.

However I'd like to ask that you please make sure this bug doesn't exist in your PR :)

Also please backport the incremental backoff while reconnecting and showing the log only once (until the connection is successful)

artursouza (approved):

PR #6131, opened 2023-03-24
Reflactor actor internal placment client. Use clock to keep time.

JoshVanL:
Part of https://github.com/dapr/dapr/issues/2478

Refactors the actor placement client to hopefully be a bit more simple, and ensures all go routines have returned on `Close()`.

Uses clock for time keeping so tests run quicker and are more reliable.

I found the tests to be slightly flaky before- this should fix that.

### DIFF
--- a/pkg/actors/internal/client.go
+++ b/pkg/actors/internal/client.go
@@ -16,8 +16,10 @@ package internal
 import (
 	"context"
 	"errors"
+	"fmt"
 	"sync"
 	"sync/atomic"
+	"time"
 
 	v1pb "github.com/dapr/dapr/pkg/proto/placement/v1"
 
@@ -59,16 +61,21 @@ func newPlacementClient(ctx context.Context, addr string, getGrpcOpts grpcOption
 		return nil, err
 	}
 
-	conn, err := grpc.DialContext(ctx, addr, opts...)
+	dialCtx, cancel := context.WithTimeout(ctx, time.Second*5)
+	defer cancel()
+	conn, err := grpc.DialContext(dialCtx, addr, opts...)
 	if err != nil {
 		if conn != nil {
 			conn.Close()
 		}
-		return nil, err
+		return nil, fmt.Errorf("error connecting to placement service at %q: %s", addr, err)
 	}
 
 	stream, err := v1pb.NewPlacementClient(conn).ReportDaprStatus(ctx)
 	if err != nil {
+		if conn != nil {
+			conn.Close()
+		}
 		return nil, err
 	}
 
--- a/pkg/actors/internal/client.go
+++ b/pkg/actors/internal/client.go
@@ -15,168 +15,108 @@ package internal
 
 import (
 	"context"
+	"errors"
 	"sync"
+	"sync/atomic"
 
 	v1pb "github.com/dapr/dapr/pkg/proto/placement/v1"
 
 	"google.golang.org/grpc"
 )
 
+type client interface {
+	Send(*v1pb.Host) error
+	Recv() (*v1pb.PlacementOrder, error)
+	Close()
+}
+
+type grpcOptionsFn func() ([]grpc.DialOption, error)
+
 // placementClient implements the best practices when handling grpc streams
 // using exclusion locks to prevent concurrent Recv and Send from channels.
-// properly handle stream closing by draining the renamining events until receives an error.
-// and broadcasts connection results based on new connections and disconnects.
+// properly handle stream closing by draining the renamining events until
+// receives an error.
 type placementClient struct {
-	// getGrpcOpts are the options that should be used to connect to the placement service
-	getGrpcOpts func() ([]grpc.DialOption, error)
-
 	// recvLock prevents concurrent access to the Recv() method.
-	recvLock *sync.Mutex
+	recvLock sync.Mutex
+
 	// sendLock prevents CloseSend and Send being used concurrently.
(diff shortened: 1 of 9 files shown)


pkg/actors/internal/client.go
```diff
@@ -15,162 +15,108 @@ package internal
...
 	}
 
-	ctx, cancel := context.WithCancel(context.Background())
-	client := v1pb.NewPlacementClient(conn)
-	stream, err := client.ReportDaprStatus(ctx)
+	stream, err := v1pb.NewPlacementClient(conn).ReportDaprStatus(ctx)
 	if err != nil {
-		cancel()
-		return err
+		return nil, err
```
ItalyPaleAle, line 72: This should fix #6130

```suggestion
        if conn != nil {
			conn.Close()
		}
		return nil, err
```

pkg/actors/internal/placement.go
```diff
@@ -129,70 +146,63 @@ func (p *ActorPlacement) AddHostedActorType(actorType string) error {
...
+	defer p.wg.Wait()
 
-	// Send the current host status to placement to register the member and
-	// maintain the status of member by placement.
-	p.shutdownConnLoop.Add(1)
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	serverAddr := p.serverAddr[p.serverIndex.Load()]
+	log.Infof("trying to connect to placement service: %s", serverAddr)
```
ItalyPaleAle, line 180: Please make sure that this log is only shown once and not on every attempt, as that can be very spammy (too many logs -> noise -> people miss out on important logs).

The counter should be reset after a successful connection.

pkg/actors/internal/placement.go
```diff
@@ -129,70 +146,63 @@ func (p *ActorPlacement) AddHostedActorType(actorType string) error {
...
+	client, err := newPlacementClient(ctx, serverAddr, p.grpcOpts)
+	if errors.Is(err, errEstablishingTLSConn) || errors.Is(err, context.Canceled) {
+		return nil
+	}
+	if err != nil {
+		p.serverIndex.Store((p.serverIndex.Load() + 1) % int32(len(p.serverAddr)))
+		return err
+	}
+
+	log.Infof("connected to placement service at address %s", serverAddr)
```
ItalyPaleAle, line 191: Nit (logs start uppercased usually):

```suggestion
	log.Info("Connected to placement service at address " + serverAddr)
```

pkg/actors/internal/placement.go
```diff
@@ -201,41 +211,57 @@ func (p *ActorPlacement) Start() {
 				// app is unresponsive, close the stream and disconnect from the placement service.
 				// Then Placement will remove this host from the member list.
 				log.Debug("disconnecting from placement service by the unhealthy app.")
```
ItalyPaleAle, line 213:
```suggestion
				log.Debug("Disconnecting from placement service by the unhealthy app")
```

pkg/actors/internal/placement.go
```diff
@@ -129,70 +146,63 @@ func (p *ActorPlacement) AddHostedActorType(actorType string) error {
...
-				})
-			} else {
-				p.onPlacementOrder(resp)
-			}
+		select {
+		case <-ctx.Done():
+			return nil
+		case <-p.closedCh:
+			return nil
+		case <-p.clock.Tick(placementReconnectInterval):
```
ItalyPaleAle, line 164: Please use an exponential backoff, which is reset when the connection is successful. See #6130

ItalyPaleAle (changes requested):
To test for the bug fixed by #6130:

1. Start daprd (`cd cmd/daprd && go run`) with `--metrics-port 9090` and `--placement-host-address localhost:9999` (this address is purposely wrong)
2. Make sure that the connection fails (and the logs should be shown only once even though attempts, with exponential backoff, are happening in background).
3. Do `curl localhost:9090/metrics -s | grep ^go_goroutines`
4. The number of goroutines should be fairly stable (if the bug is there, you'll see this growing to 100s or 1,000s in a matter of seconds).

JoshVanL:
Thanks @ItalyPaleAle, I've added your suggestions :slightly_smiling_face: 

I've tested that the go routine count is stable when running, and fluctuates in a consistent way when connecting/disconnecting.

artursouza (dismissed):

pkg/actors/internal/placement.go
```diff
@@ -129,113 +156,133 @@ func (p *ActorPlacement) AddHostedActorType(actorType string) error {
...
-	// Send the current host status to placement to register the member and
-	// maintain the status of member by placement.
-	p.shutdownConnLoop.Add(1)
+	serverAddr := p.serverAddr[p.serverIndex.Load()]
+	if firstAttempt {
+		log.Info("Trying to connect to placement service: " + serverAddr)
+	}
+
+	client, err := newPlacementClient(ctx, serverAddr, p.grpcOpts)
+	if errors.Is(err, errEstablishingTLSConn) || errors.Is(err, context.Canceled) {
```
ItalyPaleAle, line 203: What's the reason for `errEstablishingTLSConn` to be permanent? This would make the sidecar continue working without a connection to placement, so without actors, correct? Shouldn't this be retried?

ItalyPaleAle (dismissed):

mukundansundar:
@JoshVanL can you please fix the conflicts?

PR #6134, opened 2023-03-26
Fix multiple bugs in BulkGet APIs

ItalyPaleAle:
> In order of severity, decreasing

gRPC API:

1. Encrypted messages were not decrypted when a "nartive" BulkGet implementation is available.
2. When a "native" BulkGet implementation isn't available, the HTTP API returned items in the order they were requested (correctly), while the gRPC one returned items in a random order

Both HTTP and gRPC APIs:

1. Do not attempt to decrypt items that have no data or already have an error, which would otherwise cause the error to be swallowed
2. When there's an error, implementations could disclose the raw, encrypted data


mukundansundar (approved):
lgtm

PR #5429, opened 2022-10-29
Subtle Crypto APIs for gRPC and HTTP

ItalyPaleAle:
# Description

This PR contains the initial implementation of the **Subtle Crypto** APIs from the Crypto building block as described in: https://github.com/dapr/proposals/pull/3

This PR only implements the subtle crypto APIs, the "low level" ones. The higher-level encrypt/decrypt APIs will be built on top of this in a separate workstream.

It includes gRPC and HTTP APIs (implemented as "universal APIs").

The companion PR in components-contrib is: https://github.com/dapr/components-contrib/pull/2237 
That PR includes the first 4 components too:

- `crypto.kubernetes.secrets`: stores keys as Kubernetes secrets
- `crypto.jwks`: retrieves keys from a JWKS, either from a HTTP(S) URL or from the local disk
- `crypto.localstorage`: uses keys stored in files on a locally-mounted folder
- `crypto.azure.keyvault`: uses keys stored in Azure Key Vault (as keys)

## TODO

* [ ] Add more unit tests
* [ ] Add E2E tests
* [ ] Docs for the new building block and its APIs

PR #6136, opened 2023-03-27
[Crypto] Register local components

ItalyPaleAle:
Spin-off from #5429

Registers the local components:

- `crypto.localstorage`
- `crypto.jwks`
- `crypto.kubernetes.secrets`

They are all based on the same abstract "local crypto" base component


cmd/daprd/components/crypto_azure_keyvault.go
```diff
@@ -0,0 +1,23 @@
+/*
+Copyright 2021 The Dapr Authors
```
mukundansundar, line 2: nitpick 2023 in all new files.

cmd/daprd/components/crypto_kubernetes_secrets.go
```diff
@@ -0,0 +1,23 @@
...
+
+package components
+
+import (
+	"github.com/dapr/components-contrib/crypto/kubernetes/secrets"
+	cryptoLoader "github.com/dapr/dapr/pkg/components/crypto"
+)
+
+func init() {
+	cryptoLoader.DefaultRegistry.RegisterComponent(secrets.NewKubeSecretsCrypto, "kubernetes.secrets")
```
mukundansundar, line 22: do we need `k8s.secrets`? Can it just be `kubernetes`?

mukundansundar (commented):
small nitpicks ...

cmd/daprd/components/crypto_azure_keyvault.go
ItalyPaleAle: Done

ItalyPaleAle:
Proposed names:

`crypto.local.file`
`crypto.local.jwks`
`crypto.local.kubernetes.secrets`
`crypto.azure.keyvault`

JoshVanL:
I propose `local` -> `daprManagedKey`.

mukundansundar:
`local` -> `native`?

ItalyPaleAle:
I changed this PR to only have the "local" components while we agree on the naming. I'll move the Azure component to a separate PR so it can be merged in the meanwhile

mukundansundar:
For the naming part, can we change `local` to `inmemory`? Since all ops are performed in memory and keys are available in memory. 
cc @yaron2 @artursouza @msfussell @berndverst

ItalyPaleAle:
Let's try to get a consensus on naming!

Here are some proposed names from this thread and some other additions:

- `local`
- `inmemory`
- `builtin`
- `daprManagedKey`
- `native`
- `dapr`
- `in-dapr`
- `localkey`

yaron2:
+1 binding on dapr. My reasoning is that component names typically point to what the component is and not to where it stores/processes the data. Theoretically, an inmemory implementation can use an external component to implement the functionality.

ItalyPaleAle:
Ok so I've updated the PR and these are the current names as implemented:

- `crypto.dapr.localstorage` (I think `localstorage` is better than `file` because it actually watches a folder, and not a file - but I can revert)
- `crypto.dapr.kubernetes.secrets` (yes, it's a mouthful but I think `kubernetes.secrets` is required in case we need to support other K8s resources in the future)
- `crypto.dapr.jwks` (this actually watches a file, unlike `localstorage` :) - or a URL or a fixed value passed in the metadata)

mukundansundar:
> Ok so I've updated the PR and these are the current names as implemented:
> 
> * `crypto.dapr.localstorage` (I think `localstorage` is better than `file` because it actually watches a folder, and not a file - but I can revert)
> * `crypto.dapr.kubernetes.secrets` (yes, it's a mouthful but I think `kubernetes.secrets` is required in case we need to support other K8s resources in the future)
> * `crypto.dapr.jwks` (this actually watches a file, unlike `localstorage` :) - or a URL or a fixed value passed in the metadata)

These are needed for high level crypto APIs also right @ItalyPaleAle ?

mukundansundar (approved):
lgtm

ItalyPaleAle:
> These are needed for high level crypto APIs also right

Yes they are used by both low-level and high level. I had included this diagram in the proposal, hoping it clarifies:

https://github.com/dapr/proposals/blob/main/resources/0002-RCBS-Crypto-building-block/data-flow.png

PR #6137, opened 2023-03-27
[Crypto] Add SubtleCrypto APIs for HTTP and gRPC

ItalyPaleAle:
> Spin-off from #5429

Implements the Subtle Crypto APIs (low-level) as defined in https://github.com/dapr/proposals/pull/3

These APIs are implemented as universal APIs. They include support for resiliency and metrics.


mukundansundar (commented):
as discussed yesterday, please remove the public APIs alone for Subtle Crypto

ItalyPaleAle:
PR is now ready for review.

The subtlecrypto APIs are gated behind a build-time flag and are available only if you build Dapr with `-tags subtlecrypto`

Otherwise, all calls to those APIs will return an error indicating they are un-implemented.

dapr/proto/runtime/v1/dapr.proto
```diff
@@ -687,6 +708,174 @@ message UnlockResponse {
...
+message SubtleGetKeyAlpha1Request {
+  enum KeyFormat {
+    // PEM (PKIX) (default)
+    PEM = 0;
+    // JSON (JSON Web Key) as string
+    JSON = 1;
+  }
+ 
+  // Name of the component
+  string component_name = 1 [json_name="component"];
```
mukundansundar, line 721:
```suggestion
  string component_name = 1 [json_name="componentName"];
```
Lets have the same name across both proto and json
same for other scenarios.

pkg/grpc/universalapi/api_subtlecrypto_subtlecrypto.go
```diff
@@ -0,0 +1,351 @@
...
+		return &runtimev1pb.SubtleVerifyAlpha1Response{}, err
+	}
+
+	return &runtimev1pb.SubtleVerifyAlpha1Response{
+		Valid: valid,
+	}, nil
+}
+
+// Internal method that checks if the request is for a valid crypto component.
+func (a *UniversalAPI) cryptoValidateRequest(componentName string) (contribCrypto.SubtleCrypto, error) {
```
mukundansundar, line 336:
```suggestion
func (a *UniversalAPI) validateRequest(componentName string) (contribCrypto.SubtleCrypto, error) {
```

pkg/grpc/universalapi/api_subtlecrypto_subtlecrypto.go
```diff
@@ -0,0 +1,351 @@
...
+	policyRunner := resiliency.NewRunner[jwk.Key](ctx,
+		a.Resiliency.ComponentOutboundPolicy(in.ComponentName, resiliency.Crypto),
+	)
+	start := time.Now()
+	res, err := policyRunner(func(ctx context.Context) (jwk.Key, error) {
+		return component.GetKey(ctx, in.Name)
+	})
+	elapsed := diag.ElapsedSince(start)
+
+	diag.DefaultComponentMonitoring.CryptoInvoked(ctx, in.ComponentName, diag.Get, err == nil, elapsed)
```
mukundansundar, line 55: diagnostics wise, the key is got here ... but then it can fail to be marshalled leading to an error, shouldn't diagnostics be set only after it is marshalled, just be returning the value?

pkg/grpc/universalapi/api_subtlecrypto_subtlecrypto.go
```diff
@@ -0,0 +1,351 @@
...
+			err = fmt.Errorf("failed to marshal public key %s as JSON: %w", in.Name, err)
+			err = messages.ErrCryptoGetKey.WithFormat(in.Name, err)
+			a.Logger.Debug(err)
+			return &runtimev1pb.SubtleGetKeyAlpha1Response{}, err
+		}
+
+	default:
+		err = messages.ErrBadRequest.WithFormat("invalid key format")
+		a.Logger.Debug(err)
+		return &runtimev1pb.SubtleGetKeyAlpha1Response{}, err
```
mukundansundar, line 108: can be moved to before the component is invoked.

pkg/grpc/universalapi/api_subtlecrypto_subtlecrypto.go
```diff
@@ -0,0 +1,351 @@
...
+	if err != nil {
+		return &runtimev1pb.SubtleEncryptAlpha1Response{}, err
+	}
+
+	policyRunner := resiliency.NewRunner[subtleEncryptRes](ctx,
+		a.Resiliency.ComponentOutboundPolicy(in.ComponentName, resiliency.Crypto),
+	)
+	start := time.Now()
+	ser, err := policyRunner(func(ctx context.Context) (r subtleEncryptRes, rErr error) {
+		r.ciphertext, r.tag, rErr = component.Encrypt(ctx, in.Plaintext, in.Algorithm, in.Key, in.Nonce, in.AssociatedData)
```
mukundansundar, line 134: can the algorithm be validated before calling the component?

dapr/proto/runtime/v1/dapr.proto
```diff
@@ -687,6 +708,174 @@ message UnlockResponse {
...
+// SubtleEncryptAlpha1Request is the request for SubtleEncryptAlpha1.
+message SubtleEncryptAlpha1Request {
+  // Name of the component
+  string component_name = 1 [json_name="component"];
+  // Message to encrypt.
+  bytes plaintext = 2;
+  // Algorithm to use, as in the JWA standard.
+  string algorithm = 3;
+  // Name (or name/version) of the key.
+  string key = 4;
```
mukundansundar, line 746:
```suggestion
  string key_name = 4 [json_name="keyName"];
```
and similar changes in all other places. I believe we are calling it keyName in other places i.e. components-contrib

pkg/messages/predefined.go
```diff
@@ -95,7 +95,8 @@ const (
 
 var (
 	// Generic.
-	ErrBadRequest = APIError{"invalid request: %v", "ERR_BAD_REQUEST", http.StatusBadRequest, grpcCodes.InvalidArgument}
+	ErrBadRequest     = APIError{"invalid request: %v", "ERR_BAD_REQUEST", http.StatusBadRequest, grpcCodes.InvalidArgument}
+	ErrAPIUnavailable = APIError{"this API is currently not available", "ERR_API_UNAVAILABLE", http.StatusNotImplemented, grpcCodes.Unimplemented}
```
mukundansundar, line 99: should this more API not implemented?

pkg/proto/runtime/v1/dapr_additional.go
```diff
@@ -0,0 +1,64 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package runtime
+
+// This file contains additional, hand-written methods added to the generated objects.
```
mukundansundar, line 16: Why is this file needed?
Is `SetComponentName` not automatically generated?

pkg/grpc/universalapi/api_subtlecrypto_subtlecrypto.go
ItalyPaleAle: I'll leave it to `crypto` because the struct is used by a number of APIs, not just crypto, and other building blocks may have their own `validateRequest`.

dapr/proto/runtime/v1/dapr.proto
ItalyPaleAle: Changed

dapr/proto/runtime/v1/dapr.proto
ItalyPaleAle: Changed

(Although the JSON name here is unused)

pkg/grpc/universalapi/api_subtlecrypto_subtlecrypto.go
ItalyPaleAle: Looks kike this is consistent with what we do for other building blocks: https://github.com/ItalyPaleAle/dapr/blob/e07039b316f2ff8cfc0558d42c73e9054a539789/pkg/grpc/api.go

I think this is fine as what we want to measure here is the success of the operation from the component and the latency it adds

pkg/messages/predefined.go
ItalyPaleAle: Changed

pkg/proto/runtime/v1/dapr_additional.go
ItalyPaleAle: It is not.

I need these methods so I can have an interface that matches those structs, and allows avoiding repetition of a lot of code.

pkg/grpc/universalapi/api_subtlecrypto_subtlecrypto.go
ItalyPaleAle: Are you thinking of using the `SupportedEncryptionAlgorithms` method?

mukundansundar: Yes

ItalyPaleAle: IMHO that's redundant. All components right now synchronously check if the algorithm is supported at the very beginning. In the case of AKV, it's checked before invoking the cloud service: https://github.com/dapr/components-contrib/blob/master/crypto/azure/keyvault/component.go#L147

I'd let the component make that decision personally. There can also be edge cases the component can handle better than the runtime.

pkg/grpc/universalapi/api_subtlecrypto.go
```diff
@@ -0,0 +1,78 @@
...
+	return nil, messages.ErrAPIUnimplemented
+}
+
+// SubtleVerifyAlpha1 verifies the signature of a message using a key stored in the vault.
+func (a *UniversalAPI) SubtleVerifyAlpha1(ctx context.Context, in *runtimev1pb.SubtleVerifyAlpha1Request) (*runtimev1pb.SubtleVerifyAlpha1Response, error) {
+	return nil, messages.ErrAPIUnimplemented
+}
+
+// CryptoValidateRequest is an internal method that checks if the request is for a valid crypto component.
+func (a *UniversalAPI) CryptoValidateRequest(componentName string) (contribCrypto.SubtleCrypto, error) {
```
mukundansundar, line 63: nitpick: any reason to export the function here?

mukundansundar: it also seems to be repeated in the file `api_subtlecrypto_subtlecrypto.go` 
https://github.com/dapr/dapr/pull/6137/files#diff-d1155b1f22248cfd3a8bafe9904e5b41a702cbe4dca72c91d50f216b8f7caa7cR340

ItalyPaleAle: Yes It's needed in #6197 

It's repeated because `api_subtlecrypto_subtlecrypto.go` requires the build tag `subtlecrypto`, while `api_subtlecrypto.go` is built only if the build tag is NOT present.

mukundansundar (dismissed):
lgtm

pkg/http/api_subtlecrypto.go
```diff
@@ -0,0 +1,136 @@
...
+	)
+}
+
+func (a *api) onPostSubtleCryptoSign() fasthttp.RequestHandler {
+	return UniversalFastHTTPHandler(
+		a.universal.SubtleSignAlpha1,
+		UniversalFastHTTPHandlerOpts[*runtimev1pb.SubtleSignAlpha1Request, *runtimev1pb.SubtleSignAlpha1Response]{
+			InModifier: subtleCryptoInModifier[*runtimev1pb.SubtleSignAlpha1Request],
+		},
+	)
```
mukundansundar, line 120: @ItalyPaleAle adding unit tests for these apis would be good

ItalyPaleAle: When they get enabled, yes :) Right now, there's not much to test

mukundansundar (approved):

PR #6139, opened 2023-03-27
Add GetComponentMetadata methods due to component interface updates

berndverst:
# Description

Add GetComponentMetadata methods due to component interface updates

LockStore and ConfigurationStore now require `GetComponentMetadata` methods as their interface in contrib was updated to include this. This provides dummy implementations for our mock stores - which is sufficient.

This also proactively adds these methods to other mock components - even if they do not have the GetComponentMetadata interface update yet. This avoids future problems when attempting to pin a new components-contrib.


ItalyPaleAle (approved):

berndverst:
That GoRoutine test is meaningless here. We didn't add any new component.

yaron2 (approved):

PR #6159, opened 2023-03-30
[release-1.10] Use a single HTTP client for all connections to the app, which supports TLS [Fixes actors not working with app-ssl enabled]

### LINKED ISSUE #6141: (Actor) healthchecks not working when using https app port
## In what area(s)?

/area runtime

> /area operator

> /area placement

> /area docs

> /area test-and-release

## What version of Dapr?

> 1.10.3

## Expected Behavior

Calling /healthz without issues when using app-ssl with the https endpoint of an application keeping the application alive.

## Actual Behavior

/healthz is not called correctly causing the application to become unavailable with error:
```disconnecting from placement service by the unhealthy app```

I guess this is either cause fasthttp is not setup for SSL in general or that it's not skipping the certificate check here.
https://github.com/dapr/dapr/blob/a49efc961a3325b2a230496d9a0460dc8cc7edbb/pkg/health/health.go#L63-L67
while the application channel in general skips TLS checks:
https://github.com/dapr/dapr/blob/a49efc961a3325b2a230496d9a0460dc8cc7edbb/pkg/channel/http/http_channel.go#L86-L89
This results in an inconsistency.

Maybe it's just missing the same TLSConfig like this?

```go
              tlsConfig = &tls.Config{
			InsecureSkipVerify: true,
               }
               client := &fasthttp.Client{
			MaxConnsPerHost:           5, // Limit Keep-Alive connections
			ReadTimeout:               options.requestTimeout,
			MaxIdemponentCallAttempts: 1,
                        TLSConfig: tlsConfig
		}
```

Seems like fasthttp supports that:
https://github.com/valyala/fasthttp/blob/master/client.go#L208

Create an application with an actor.
Setup the application to listen via SSL, then use "--app-ssl" with that app port.

## Release Note

RELEASE NOTE: **FIX**: Health checks do not work when using HTTPS with `--app-ssl`

ItalyPaleAle:
Fixes #6141

This was implemented in a way that:

- Makes health checks work when `app-ssl` is enabled (#6141)
- Switches the health package's client to net/http rather than fasthttp. This allows taking advantage of HTTP/2 and multiplexing, among other things.
- Uses a single http.Client for health checks and app channel. This allows re-using TCP sockets and multiplexing when HTTP/2 is enabled, possibly offering much better efficienty

### DIFF
--- a/pkg/runtime/runtime.go
+++ b/pkg/runtime/runtime.go
@@ -15,6 +15,7 @@ package runtime
 
 import (
 	"context"
+	"crypto/tls"
 	"encoding/base64"
 	"encoding/json"
 	"errors"
@@ -212,6 +213,7 @@ type DaprRuntime struct {
 	appHealthReady            func() // Invoked the first time the app health becomes ready
 	appHealthLock             *sync.Mutex
 	bulkSubLock               *sync.Mutex
+	appHTTPClient             *nethttp.Client
 
 	secretsConfiguration map[string]config.SecretsScope
 
@@ -308,6 +310,8 @@ func NewDaprRuntime(runtimeConfig *Config, globalConfig *config.Configuration, a
 		dl := newComponentDenyList(globalConfig.Spec.ComponentsSpec.Deny)
 		rt.componentAuthorizers = append(rt.componentAuthorizers, dl.IsAllowed)
 	}
+
+	rt.initAppHTTPClient()
 
 	return rt
 }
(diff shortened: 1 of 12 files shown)


ItalyPaleAle:
Rebased onto release-1.10 per conversation with @artursouza

pkg/runtime/runtime.go
```diff
@@ -2951,6 +2958,64 @@ func (a *DaprRuntime) createAppChannel() (err error) {
...
+func (a *DaprRuntime) initAppHTTPClient() {
+	var tlsConfig *tls.Config
+	if a.runtimeConfig.AppSSL {
+		tlsConfig = &tls.Config{
+			//nolint:gosec
+			InsecureSkipVerify: true,
+			// For 1.11
+			// MinVersion:         channel.AppChannelMinTLSVersion,
+		}
+		// For 1.11
```
ItalyPaleAle, line 2981: Reviewers: when I rebased from master to 1.10, there was a merge conflict. Want to make sure this stays here for when we merge the release branch back into master, so we don't miss this.

Issue #6043, opened 2023-03-08
Support for PURGE workflow API


nyemade-uversky:
**Description**
Currently, actor state isn't removed from the user's state store. I would like the ability for users to remove actor state from their state store so that we can avoid bloating user state stores.

**Ask**
Support the PURGE workflow API to allow for state store cleanup 

**Requirements**
1. Workflow Actor state should be marked for deletion only via event:
   - timeframe
   - manual deletion
   - input received
2. Activity actor state automatically deleted after activity is completed. 
3. Removal of activity actor state is configurable

PR #6163, opened 2023-03-30
Workflow purge

RyanLettieri:
# Description

Addition of Purge method for dapr workflow.

## Issue reference

Note that this PR will not pass until the following PR is submitted into components-contrib: https://github.com/dapr/components-contrib/pull/2729

Please reference the issue this PR will close: #_[[6043](https://github.com/dapr/dapr/issues/6043)]_

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing


dapr/proto/runtime/v1/dapr.proto
```diff
@@ -711,6 +714,16 @@ message TerminateWorkflowRequest {
 }
 
 message TerminateWorkflowResponse {
+  // Empty
+}
+
+message PurgeWorkflowRequest {
+  string instance_id = 1;
+  string workflow_component = 2;
```
ItalyPaleAle, line 722: `json_name` tags are very important when the proto contains `_`. Otherwise, the default when serializing as JSON (like in Universal APIs) is to use the key in snake_case.

```suggestion
  string instance_id = 1 [json_name = "instanceId"];
  string workflow_component = 2 [json_name = "workflowComponent"];
```

pkg/actors/actors.go
```diff
@@ -399,7 +399,6 @@ func (a *actorsRuntime) Call(ctx context.Context, req *invokev1.InvokeMethodRequ
 	if lar == nil {
 		lar = &lookupActorRes{}
 	}
-
 	var resp *invokev1.InvokeMethodResponse
```
ItalyPaleAle, line 402:
```suggestion
	}

	var resp *invokev1.InvokeMethodResponse
```

Because there'll be a change on this file getting merged very soon and I want to spare a merge conflict :)

pkg/runtime/wfengine/activity.go
```diff
@@ -78,19 +78,30 @@ func (a *activityActor) InvokeMethod(ctx context.Context, actorID string, method
 	}
 
 	// Try to load activity state. If we find any, that means the activity invocation is a duplicate.
-	if state, err := a.loadActivityState(ctx, actorID, ar.Generation); err != nil {
+	var state activityState
```
ItalyPaleAle, line 81: Uhm something feels odd here. `state` is not read from `a.loadActivityState` anymore, so it will always be empty? and `if state.Generation> 0` will always be false? I'm not sure what this should have been.

pkg/runtime/wfengine/activity.go
```diff
@@ -78,19 +78,30 @@ func (a *activityActor) InvokeMethod(ctx context.Context, actorID string, method
...
+	if _, err := a.loadActivityState(ctx, actorID, ar.Generation); err != nil {
 		return nil, err
 	} else if state.Generation > 0 {
 		return nil, ErrDuplicateInvocation
 	}
 
+	if methodName == "PurgeWorkflowState" {
+		if err := a.purgeActivityState(ctx, actorID, state); err != nil {
+			return nil, err
+		}
```
ItalyPaleAle, line 91: Should this method return after processing this? We don't need to schedule a reminder, do we?

pkg/runtime/wfengine/activity.go
```diff
@@ -280,8 +291,30 @@ func (a *activityActor) saveActivityState(ctx context.Context, actorID string, s
...
+
+	if !a.cachingDisabled {
+		a.statesCache.Store(actorID, state)
+	}
+	return nil
+}
+
 func getActivityInvocationKey(generation uint64) string {
-	return fmt.Sprintf("activityreq-%d", generation)
+	// return fmt.Sprintf("activityreq-%d", generation)
```
ItalyPaleAle, line 316: Why removing the generation?

pkg/runtime/wfengine/component.go
```diff
@@ -25,7 +25,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"github.com/dapr/components-contrib/workflows"
-	componentsV1alpha1 "github.com/dapr/dapr/pkg/apis/components/v1alpha1"
+	componentsV1alpha1 "github.com/dapr/dapr/pkg/apis/components/v1alpha1" // This will be removed
```
ItalyPaleAle, line 28: Forgot to remove this and `ComponentDefinition`? :)

pkg/runtime/wfengine/component.go
```diff
@@ -130,6 +130,18 @@ func (c *workflowEngineComponent) Terminate(ctx context.Context, req *workflows.
 	return nil
 }
 
+func (c *workflowEngineComponent) Purge(ctx context.Context, req *workflows.PurgeRequest) error {
+	if req.InstanceID == "" {
+		return fmt.Errorf("a workflow instance ID is required")
```
ItalyPaleAle, line 135:
```suggestion
		return errors.New("a workflow instance ID is required")
```

If no format is included, please use `errors.New` (it's faster and makes less allocs)

pkg/runtime/wfengine/workflow.go
```diff
@@ -213,6 +216,70 @@ func (wf *workflowActor) getWorkflowMetadata(ctx context.Context, actorID string
...
+	}
+
+	runtimeState := getRuntimeState(actorID, state)
+	if !runtimeState.IsCompleted() {
+		return api.ErrNotCompleted
+	}
+
+	// Loop through the history of the loaded workflow state and schedule purge calls for all the activities that have been completed
+
+	// RRL TODO: Looping through history and invoking each actor once. They are not in the same transaction. Eqch transaction happens as its own call
```
ItalyPaleAle, line 235: As per the initial review during our 1:1 earlier today, this needs to change - although I admit I don't _fully_ understand how this works so I can't tell you exactly how to fix this yet 🙃

However, the problem is that the "purge" needs to happen in an atomic way. We discussed this and you explained that you need to delete 3 things:

- Each activity (this can be more than one)
- The history object
- The inbox object

These all have to be removed _at once_. We do that in Dapr by using the transactional state stores and performing a "transaction" (which are just operations performed atomically as a batch), but here, each activity is removed separately, in its own "transaction" (with a single opeation). Each activity is removed by the actor that you're invoking within the loop, and an error in the loop causes the entire method to exit, which means that _some_ state will have been removed, but not all.

ItalyPaleAle (dismissed):

dapr/proto/runtime/v1/dapr.proto
cgillum: nit: use `instanceID` instead of `instanceId` for the JSON name? This is what I've been trying to do in my other PRs.

dapr/proto/runtime/v1/dapr.proto
```diff
@@ -148,7 +148,10 @@ service Dapr {
   rpc ResumeWorkflowAlpha1 (WorkflowActivityRequest) returns (WorkflowActivityResponse) {}
 
   // Raise Event for a Workflow
-  rpc RaiseEventWorkflowAlpha1 (RaiseEventWorkflowRequest) returns (RaiseEventWorkflowResponse) {}
+  rpc RaiseEventWorkflowAlpha1 (RaiseEventWorkflowRequest) returns (WorkflowActivityResponse) {}
+
+  // Purge Workflow
+  rpc PurgeWorkflowAlpha1 (PurgeWorkflowRequest) returns (WorkflowActivityResponse) {}
```
cgillum, line 154: Standard practice for APIs that don't return anything is to use `google.protobuf.Empty`. I suggest we do the same here.

pkg/grpc/universalapi/api_workflow.go
```diff
@@ -125,30 +125,30 @@ func (a *UniversalAPI) TerminateWorkflowAlpha1(ctx context.Context, in *runtimev
 	return a.workflowActivity(ctx, in, method, messages.ErrTerminateWorkflow)
 }
 
-func (a *UniversalAPI) RaiseEventWorkflowAlpha1(ctx context.Context, in *runtimev1pb.RaiseEventWorkflowRequest) (*runtimev1pb.RaiseEventWorkflowResponse, error) {
+func (a *UniversalAPI) RaiseEventWorkflowAlpha1(ctx context.Context, in *runtimev1pb.RaiseEventWorkflowRequest) (*runtimev1pb.WorkflowActivityResponse, error) {
```
cgillum, line 128: I suggest reverting these changes because I'm making additional changes to these signatures in a nother PR.

pkg/messages/predefined.go
```diff
@@ -136,6 +136,7 @@ var (
 	ErrTerminateWorkflow             = APIError{"error terminating workflow %s", "ERR_TERMINATE_WORKFLOW", http.StatusInternalServerError, grpcCodes.Internal}
 	ErrMissingWorkflowEventName      = APIError{"missing workflow event name", "ERR_WORKFLOW_EVENT_NAME_MISSING", http.StatusBadRequest, grpcCodes.InvalidArgument}
 	ErrRaiseEventWorkflow            = APIError{"error raising event on workflow %s", "ERR_RAISE_EVENT_WORKFLOW", http.StatusInternalServerError, grpcCodes.Internal}
+	ErrPurgeWorkflow                 = APIError{"error purging workflow %s", "ERR_PURGE_WORKFLOW", http.StatusInternalServerError, grpcCodes.Internal}
```
cgillum, line 139: Surround the `%s` with quotes, like this:
```suggestion
	ErrPurgeWorkflow                 = APIError{"error purging workflow '%s'", "ERR_PURGE_WORKFLOW", http.StatusInternalServerError, grpcCodes.Internal}
```

I'm already making this change to all the other messages in another PR.

pkg/runtime/wfengine/activity.go
```diff
@@ -78,19 +78,30 @@ func (a *activityActor) InvokeMethod(ctx context.Context, actorID string, method
...
-	state := activityState{
+	state = activityState{
 		Generation:   ar.Generation,
 		EventPayload: ar.HistoryEvent,
 	}
-	if err := a.saveActivityState(ctx, actorID, state); err != nil {
-		return nil, err
+
+	// This conditional block prevents the statestore from re-saving the purged item
+	if methodName != "PurgeWorkflowState" {
```
cgillum, line 101: This can (and should) be removed if you add a `return` further up like Alessandro suggested.

pkg/runtime/wfengine/activity.go
```diff
@@ -280,8 +291,30 @@ func (a *activityActor) saveActivityState(ctx context.Context, actorID string, s
...
+		ActorID:   actorID,
+		Operations: []actors.TransactionalOperation{{
+			Operation: actors.Delete,
+			Request: actors.TransactionalDelete{
+				Key: getActivityInvocationKey(state.Generation),
+			},
+		}},
+	}
+	if err := a.actorRuntime.TransactionalStateOperation(ctx, &req); err != nil {
+		return fmt.Errorf("failed to save activity state: %w", err)
```
cgillum, line 306:
```suggestion
		return fmt.Errorf("failed to delete activity state: %w", err)
```

Also, let's add the key in this error message to make debugging easier/possible.

pkg/runtime/wfengine/wfengine_test.go
```diff
@@ -714,6 +715,60 @@ func TestRaiseEvent(t *testing.T) {
...
+		return input + 1, nil
+	})
+
+	ctx := context.Background()
+	client, engine, stateStore := startEngineAndGetStore(ctx, r)
+	for _, opt := range GetTestOptions() {
+		t.Run(opt(engine), func(t *testing.T) {
+			id, err := client.ScheduleNewOrchestration(ctx, "ActivityChainToPurge")
+			if assert.NoError(t, err) {
+				metadata, err := client.WaitForOrchestrationStart(ctx, id)
```
cgillum, line 745: Rather than waiting for the orchestration to start, wait for it to complete.

```suggestion
				metadata, err := client.WaitForOrchestrationCompletion(ctx, id)
```

pkg/runtime/wfengine/wfengine_test.go
```diff
@@ -714,6 +715,60 @@ func TestRaiseEvent(t *testing.T) {
...
+				if assert.NoError(t, err) {
+					assert.Equal(t, id, metadata.InstanceID)
+					metadata, err = client.FetchOrchestrationMetadata(ctx, id)
+					time.Sleep(5 * time.Second) // The sleep is here to avoid purging an on-going activity
```
cgillum, line 749: Delete this sleep and the call to `FetchOrchestrationMetadata` after making the above suggested change.

pkg/runtime/wfengine/wfengine_test.go
```diff
@@ -714,6 +715,60 @@ func TestRaiseEvent(t *testing.T) {
...
+
+					// Check that no key from the statestore that was expected to be purged is still present in the statestore
+					keysPostPurge := []string{}
+					for key := range stateStore.(*fakeStateStore).items {
+						keysPostPurge = append(keysPostPurge, key)
+					}
+
+					for _, item := range keysPostPurge {
+						// Note that "activityState" comes from "getActivityInvocationKey" inside activity.go
+						if strings.Contains(item, "history") || strings.Contains(item, "activityState") || strings.Contains(item, "inbox") {
```
cgillum, line 762: Alternatively, can we make sure there aren't any keys that contain the workflow instance ID? That might be less likely to result in false positives. Otherwise, if we add new fields to the stored workflow state we have to remember to also update this code, which would be really easy to forget.

pkg/runtime/wfengine/workflow.go
```diff
@@ -214,6 +217,70 @@ func (wf *workflowActor) getWorkflowMetadata(ctx context.Context, actorID string
...
+	// We are deleting one by one in sequence and not in builk.
+	// In case of each actor returning an error, the entire function returns an error
+	// Everything before error is deleted, and erverything afyter is not.
+	for _, e := range state.History {
+		if ts := e.GetTaskScheduled(); ts != nil {
+			targetActorID := getActivityActorID(actorID, e.EventId)
+			eventData, err := backend.MarshalHistoryEvent(e)
```
cgillum, line 243: We shouldn't be including the event data in the purge request.

cgillum (changes requested):

pkg/runtime/wfengine/workflow.go
cgillum: Hi @ItalyPaleAle. The state exists across potentially multiple actors, and AFAIK, we don't support transactions across multiple actors. We instead go for the next best thing, which is to do independent purge operations in a particular order so that failures that happen midway can be retried until there is finally a successful completion. This is why, for example, we delete state for each activity first before deleting the history (without the history, we wouldn't know which actors need to be deleted). Does that make sense? Let me know if you feel we need to meet and discuss this more in depth.

ItalyPaleAle: I see. But what happens if there's a crash while the process is not complete?

cgillum: The crash will leave you in a partially deleted state. However, the workflow will still be visible from the GetWorkflow management APIs, and re-issuing the purge call should succeed.

pkg/runtime/wfengine/backend.go
```diff
@@ -241,6 +241,22 @@ func (be *actorBackend) GetOrchestrationWorkItem(ctx context.Context) (*backend.
 	}
 }
 
+// PurgeOrchestrationState deletes all saved state for the specific orchestration instance.
+func (be *actorBackend) PurgeOrchestrationState(ctx context.Context, id api.InstanceID) error {
+	req := invokev1.
+		NewInvokeMethodRequest(PurgeWorkflowStateMethod).
+		WithActor(be.config.workflowActorType, string(id)).
+		WithContentType(invokev1.OctetStreamContentType)
```
cgillum, line 249: You can remove the `WithContentType` line since we're not sending any content.

dapr/proto/runtime/v1/dapr.proto
cgillum: Looks like this got missed?

pkg/runtime/wfengine/wfengine_test.go
```diff
@@ -564,6 +565,68 @@ func TestRaiseEvent(t *testing.T) {
...
+		}
+		return input + 1, nil
+	})
+
+	ctx := context.Background()
+	client, engine, stateStore := startEngineAndGetStore(ctx, r)
+	for _, opt := range GetTestOptions() {
+		t.Run(opt(engine), func(t *testing.T) {
+			id, err := client.ScheduleNewOrchestration(ctx, "ActivityChainToPurge")
+			if assert.NoError(t, err) {
```
cgillum, line 594: To reduce nesting, let's replace `assert.NoError` throughout this test with `require.NoError` if remove the `if`-blocks

pkg/runtime/wfengine/workflow.go
cgillum: This still needs to be addressed (though the problem has moved to a new line now)

pkg/runtime/wfengine/wfengine_test.go
```diff
@@ -564,6 +565,68 @@ func TestRaiseEvent(t *testing.T) {
...
+					for _, item := range keysPostPurge {
+						if strings.Contains(item, string(id)) {
+							assert.True(t, false)
+						}
+					}
+				}
+			}
+		})
+	}
+}
```
cgillum, line 628: Let's add one more test in order to ensure we're covering the multi-generation case. Basically, copy the `TestContinueAsNewWorkflow` logic and do a purge against that workflow state, validating that all state store content was successfully removed.

cgillum (commented):
A few more things.

pkg/runtime/wfengine/activity.go
```diff
@@ -229,11 +234,7 @@ func (a *activityActor) loadActivityState(ctx context.Context, actorID string, g
 	result, ok := a.statesCache.Load(actorID)
 	if ok {
 		cachedState := result.(activityState)
-
-		// Make sure the cached state is for the same generation of the workflow.
-		if cachedState.Generation == generation {
```
cgillum, line 234: Now that we're not using the `Generation` field, let's delete it from `activityState`.

pkg/runtime/wfengine/wfengine_test.go
```diff
@@ -564,6 +565,134 @@ func TestRaiseEvent(t *testing.T) {
...
+			assert.NoError(t, err)
+
+			// Check that no key from the statestore containing the actor id is still present in the statestore
+			keysPostPurge := []string{}
+			for key := range stateStore.(*daprt.FakeStateStore).Items {
+				keysPostPurge = append(keysPostPurge, key)
+			}
+
+			for _, item := range keysPostPurge {
+				if strings.Contains(item, string(id)) {
```
cgillum, line 620: Should we use `assert.True(t, strings.Contains(item, string(id)))` here?

pkg/runtime/wfengine/wfengine_test.go
```diff
@@ -564,6 +565,134 @@ func TestRaiseEvent(t *testing.T) {
...
+		if err := ctx.GetInput(&input); err != nil {
+			return nil, err
+		}
+		return input + 1, nil
+	})
+	ctx := context.Background()
+	client, engine, stateStore := startEngineAndGetStore(ctx, r)
+	for _, opt := range GetTestOptions() {
+		t.Run(opt(engine), func(t *testing.T) {
+			// Second Test
```
cgillum, line 659: Is this comment still relevant or can it be removed?

pkg/runtime/wfengine/workflow.go
```diff
@@ -216,6 +219,64 @@ func (wf *workflowActor) getWorkflowMetadata(ctx context.Context, actorID string
...
+	}
+
+	runtimeState := getRuntimeState(actorID, state)
+	if !runtimeState.IsCompleted() {
+		return api.ErrNotCompleted
+	}
+
+	// Loop through the history of the loaded workflow state and schedule purge calls for all the activities that have been completed
+
+	// RRL TODO: Looping through history and invoking each actor once. They are not in the same transaction. Each transaction happens as its own call
```
cgillum, line 238: This doesn't seem like a TODO, but more like a description of the code.

pkg/runtime/wfengine/workflow.go
```diff
@@ -216,6 +219,64 @@ func (wf *workflowActor) getWorkflowMetadata(ctx context.Context, actorID string
...
+	// We are deleting one by one in sequence and not in builk.
```
cgillum, line 239: This comment seems redundant with the previous one (and has a typo). I suggest we clean up these comments.

pkg/runtime/wfengine/workflow.go
```diff
@@ -216,6 +219,64 @@ func (wf *workflowActor) getWorkflowMetadata(ctx context.Context, actorID string
...
+			activityRequestBytes, err := actors.EncodeInternalActorData(ActivityRequest{
+				Generation: state.Generation,
+			})
+			if err != nil {
+				return err
+			}
+			req := invokev1.
+				NewInvokeMethodRequest(PurgeWorkflowStateMethod).
+				WithActor(wf.config.activityActorType, targetActorID).
+				WithRawDataBytes(activityRequestBytes).
```
cgillum, line 254: Now that we don't use the generation anymore, can we remove the payload completely from the actor invocation?

pkg/runtime/wfengine/workflow.go
```diff
@@ -271,6 +332,32 @@ func (wf *workflowActor) runWorkflow(ctx context.Context, actorID string, remind
 		return nil
 	}
 
+	// The logic/for loop below purges/removes any leftover state from a completed or failed activity
+	// TODO: for optimization make multiple go routines and run them in parallel
+	for _, e := range state.Inbox {
```
cgillum, line 337: Let's move this logic into its own method.

pkg/runtime/wfengine/workflowstate.go
```diff
@@ -152,6 +152,19 @@ func addStateOperations(req *actors.TransactionalRequest, keyPrefix string, even
 	return nil
 }
 
+func addPurgeStateOperations(req *actors.TransactionalRequest, keyPrefix string, events []*backend.HistoryEvent) error {
+	// TODO: Investigate whether Dapr state stores put limits on batch sizes. It seems some storage
+	//       providers have limits and we need to know if that impacts this algorithm:
+	//       https://learn.microsoft.com/azure/cosmos-db/nosql/transactional-batch#limitations
```
cgillum, line 158: I think we should open a bug to track doing purges in smaller batches to avoid running into transactional batch limitations.

RyanLettieri: https://github.com/dapr/dapr/issues/6289

pkg/runtime/wfengine/workflow.go
```diff
@@ -233,34 +233,14 @@ func (wf *workflowActor) purgeWorkflowState(ctx context.Context, actorID string)
 		return api.ErrNotCompleted
 	}
 
-	// Loop through the history of the loaded workflow state and schedule purge calls for all the activities that have been completed
-
-	// RRL TODO: Looping through history and invoking each actor once. They are not in the same transaction. Each transaction happens as its own call
-	// We are deleting one by one in sequence and not in builk.
+	// Looping through history and invoking each actor once. They are not in the same transaction. Each transaction happens as its own call
+	// We are deleting one by one in sequence and not in bulk.
```
cgillum, line 237: This comment needs to be deleted or moved.

pkg/runtime/wfengine/activity.go
```diff
@@ -52,7 +52,6 @@ type ActivityRequest struct {
 
 type activityState struct {
 	EventPayload []byte
-	Generation   uint64
```
cgillum, line 53: There's also a Generation field on the `ActivityRequest` struct. Can that be removed too?

cgillum (commented):
Note to other reviewers that we decided to remove the de-dupe detection for activities since it was turning out to be a problematic (undocumented) optimization. We'll revisit whether it makes sense to add it back in the future.

A couple more things before signing off.

cgillum (approved):

berndverst:
Once this PR is merged, @RyanLettieri you'll need to generate the protos in the various SDK PRs you have open for workflow management.

Implement the purge operation too, please.

artursouza (approved):

Issue #5756, opened 2023-01-13
[PROPOSAL] Identity Management Overhaul


JoshVanL:
### PRs
- (main) https://github.com/dapr/dapr/pull/6171
- https://github.com/dapr/dapr/pull/6179
- https://github.com/dapr/dapr/pull/6193
- https://github.com/dapr/dapr/pull/6761
- https://github.com/dapr/dapr/pull/6791
- https://github.com/dapr/dapr/pull/6902
- https://github.com/dapr/dapr/pull/6895
- https://github.com/dapr/dapr/pull/7214
- https://github.com/dapr/proposals/pull/42
- https://github.com/dapr/dapr/pull/7111
- https://github.com/dapr/dapr/pull/7037
- https://github.com/dapr/dapr/pull/7041

/area runtime
/area operator
/area placement

This proposal describes a strategy for overhauling the identity story in Dapr. It involves several distinct points of interest, but have been put into the same proposal as they all serve the same goal and together will have a greater sum than their parts.

# Motivation

In Kubernetes mode, Dapr is made up from a set of privileged control plane components (sentry, operator, injector, dashboard, placement (also self hosted)), and lower privileged workload daprd sidecars. In order for Dapr to maintain a strong security posture to ensure malicious, or otherwise, actors cannot read/write/get/post data it should not have access to, Dapr must have a strong authentication and authorization story. Today, Dapr uses the sentry control plane component as an identity provider and certificate authority to the rest of the Dapr components and workloads.

I've identified a number of things Dapr can improve which should give us:

- Ensure mTLS cannot be compromised,
- Locked down and Principle Of Least Privilege (POLP) prevents bad actors accessing data they shouldn’t,
- Stronger identity improves reliability of metrics/tracing,
- Stronger identity allows for developing richer authorization features in future,
- The current identity management is inherently weak so no reliance on workload identity can be made.

# Goals

- Overhaul identity management in Dapr so workload identities are secure and mTLS between dapr components can be trusted.
- Upgrades to this new identity system should be seamless for users, with no manual intervention required.

## Non-Goals

- Implement a mechanism for an external CA system (though don't write the new system in a way which wouldn't allow this to be added in future).

# Current Shortfalls

Dapr currently has a number of shortcomings which prevent it from having a strong identity story, and undermines the mTLS of Dapr components.

### 1. Issuer Key Sharing

[...]

PR #6171, opened 2023-03-31
[5756 Identity Management Overhaul] Rewrite sentry to authenticate clients before signing

JoshVanL:

This is the first in a chain or PRs to complete this issue. The following PRs will implement the security changes for each of the other components. 

This rewrites the sentry service to now authenticate clients before signing for their identity in a backwards compatible way.

It introduces a new `pkg/security` package. This package is designed to manage the components SPIFFE ID (fetched from sentry), and expose connection options for authentication and authorizing servers.

The sentry flag `--token-audience` has now been deprecated. The audience of the kubernetes token used by clients should always be the SPIFFE ID of sentry (not implemented in this PR). For backwards compatibility, sentry will still accept `dapr.io/sentry` which should be remove in `v1.13`.

Sentry will ensure clients request their app ID. Sentry will still accept clients sending a CSR request with the ID set to `$namespace:$sa`, however this exception should be removed in `1.13`.

Sentry will still send the trust anchors in the sign response, but should be removed in `1.13` as trust distribution should happen out of band of signing.

Sentry will continue to include `cluster.local` as a DNS SAN on its serving cert, since pre `v1.12` client authorize on this. This should be removed in `v1.13`

Sentry will also continue to include `cluster.local` as DNS SAN on the _issuer certificate_ since this is used by the control plane components as serving certificates. Also to be removed in `v1.13`.

`global.mtls.controlPlaneTrustDomain` is a new top level helm chart to set the control plane trust domain. Currently only read by sentry.

The root CA is now created by sentry (no longer helm) and written to the `dapr-trust-bundle` secret. It also write the trust anchors to the config map `dapr-trust-bundle`. The control plane components will read the trust anchors from this ConfigMap as a volume mount.

Updates sentry validators:

1. Renamed the "selfhosted" validator to "insecure", since it doesn't do any validation at all.
1. Added a new `mtls.tokenValidators` option to the "daprsentry" Configuration resource to configure *additional* validators. This is an array that contains that contains:
   - `mtls.tokenValidators[*].name`: name of the validator.
   - `mtls.tokenValidators[*].options`: optional dictionary of key-values that are options that can be passed to validator.
1. In Kubernetes mode, the built-in "kubernetes" validator is always enabled.  In self-hosted mode, the built-in "insecure" validator is enabled only if there's no additional validator configured.
1. Added the `jwks` validator that validates . 
   - This accepts 1 option: `source` is the location of a JWKS that contains the **public** keys to validate the keys. This can be one of:
     - A path to a file on disk - this is watched for changes with fsnotify
     - A HTTP(S) URL - this is periodically refreshed if needed
     - The actual JWKS in the string value, optionally base64-encoded
   - This validator requires JWT tokens that contain the 2 claims that SPIFFE's JWT-SIV use too:
     - `sub` must be the SPIFFE ID of the app, for example `spiffe://localhost/ns/default/myapp`
     - `aud` must contain the SPIFFE ID of Sentry, for example `spiffe://localhost/ns/dapr-system/dapr-sentry`
     - The validator will validate time validity bounds if present, but it will not enforce their presence (so tokens can live as long as the issuer wants them to)
1. `daprd` can load a token to send to Sentry that is signed with an arbitrary key. These are defined using:
   - A file pointed by the env var `DAPR_SENTRY_TOKEN_FILE`
   - (Note that the env var must start with `DAPR_` for security reasons, since those are not accessible in the localenv secret store)
1. Token requests that want to use a non-default validator must pass the "token_validator" option in the gRPC request.


pkg/sentry/server/validator/internal/common.go
```diff
@@ -0,0 +1,57 @@
...
+// Validate validates the common rules for all requests.
+func Validate(_ context.Context, req *sentryv1pb.SignCertificateRequest) (spiffeid.TrustDomain, error) {
+	var errs []error
+
+	// TODO: we should also validate that the app ID is alpha numeric, and
+	// doesn't contain things like spaces etc.
+	if len(req.Id) == 0 {
+		errs = append(errs, errors.New("app ID is required"))
+	}
+	if len(req.Id) > 64 {
```
ItalyPaleAle, line 35: Why 64? Arbitrary numnber, or is there a reason?

pkg/sentry/server/validator/internal/common.go
```diff
@@ -0,0 +1,57 @@
...
+)
+
+// Validate validates the common rules for all requests.
+func Validate(_ context.Context, req *sentryv1pb.SignCertificateRequest) (spiffeid.TrustDomain, error) {
+	var errs []error
+
+	// TODO: we should also validate that the app ID is alpha numeric, and
+	// doesn't contain things like spaces etc.
+	if len(req.Id) == 0 {
+		errs = append(errs, errors.New("app ID is required"))
```
ItalyPaleAle, line 33: If you use errors.Join, all of these can be changed to:

```go
var err error
if len(req.Id) == 0 {
	err = errors.Join(err, errors.New("app ID is required"))
}
```

It's ok to call `errors.Join` and "join" a nil error too, which will not be added

pkg/sentry/server/ca/ca.go
```diff
@@ -0,0 +1,231 @@
...
+	if err != nil {
+		return caBundle{}, fmt.Errorf("failed to sign root certificate: %w", err)
+	}
+	trustAnchors := pem.EncodeToMemory(&pem.Block{Type: "CERTIFICATE", Bytes: rootCertDER})
+
+	issKey, err := ecdsa.GenerateKey(elliptic.P256(), rand.Reader)
+	if err != nil {
+		return caBundle{}, err
+	}
+	issKeyDer, err := x509.MarshalECPrivateKey(issKey)
```
ItalyPaleAle, line 203: Why not PKCS#8? Which would work with other private key types too

pkg/sentry/server/ca/ca.go
```diff
@@ -0,0 +1,231 @@
...
+	if err != nil {
+		return caBundle{}, err
+	}
+	issKeyPEM := pem.EncodeToMemory(&pem.Block{Type: "EC PRIVATE KEY", Bytes: issKeyDer})
```
ItalyPaleAle, line 207: For PKCS#8 this would be "PRIVATE KEY"

pkg/sentry/server/ca/ca.go
```diff
@@ -0,0 +1,231 @@
...
+	return append([]*x509.Certificate{cert}, c.bundle.issChain...), nil
+}
+
+// TODO: Remove this method in v1.12 since it is not used any more.
+func (c *ca) TrustAnchors() []byte {
+	return c.bundle.trustAnchors
+}
+
+func generateCABundle(conf config.Config) (caBundle, error) {
+	rootKey, err := ecdsa.GenerateKey(elliptic.P256(), rand.Reader)
```
ItalyPaleAle, line 183: More like a wishlist item (for a separate PR), but would be great to have support for other signing algorithms like good ol' RSA or EdDSA (Ed25519) too…

pkg/sentry/server/ca/cert.go
```diff
@@ -0,0 +1,123 @@
...
+
+	"github.com/spiffe/go-spiffe/v2/spiffeid"
+
+	"github.com/dapr/dapr/pkg/security"
+	"github.com/dapr/dapr/pkg/security/spiffe"
+)
+
+const (
+	// caTTL is the CA certificate TTL.
+	caTTL = 10 * 365 * 24 * time.Hour
```
ItalyPaleAle, line 33: Isn't this a lot, 10 years? Could this have a shorter lifetime?

pkg/sentry/server/ca/cert.go
```diff
@@ -0,0 +1,123 @@
...
+	cert.KeyUsage |= x509.KeyUsageCertSign
+	cert.Subject = pkix.Name{Organization: []string{trustDomain}}
+	cert.IsCA = true
+	cert.BasicConstraintsValid = true
+	cert.SignatureAlgorithm = x509.ECDSAWithSHA256
+	return cert, nil
+}
+
+// generateIssuerCert returns a CA issuing x509 Certificate.
+func generateIssuerCert(trustDomain string, skew time.Duration) (*x509.Certificate, error) {
```
ItalyPaleAle, line 83: What's the difference between a root CA and issuer CA cert?

pkg/security/token.go
```diff
@@ -0,0 +1,74 @@
...
+// getKubernetesIdentityToken returns the value of the Kubernetes identity
+// token.
+// If a audience bound token for sentry does not exist, we use the Pods API
+// Server token. After all supported Dapr control plane versions support
+// audience bound tokens, we can remove this function and use the audience
+// bound token.
+func getKubernetesIdentityToken() (string, error) {
+	b, err := os.ReadFile(filepath.Join(rootFS, kubeTknPath))
+	if os.IsNotExist(err) {
+		log.Warn("⚠️ daprd is initializing using the legacy service account token with access to Kubernetes APIs, which is discouraged. This usually happens when daprd is running against an older version of the Dapr control plane.")
```
ItalyPaleAle, line 65: I think this can be removed, it was meant to support upgrading from 1.9 to 1.10.

pkg/security/pem/pem.go
```diff
@@ -0,0 +1,192 @@
...
+func EncodePrivateKey(key any) ([]byte, error) {
+	var (
+		keyBytes  []byte
+		err       error
+		blockType string
+	)
+
+	switch key := key.(type) {
+	case *ecdsa.PrivateKey:
+		keyBytes, err = x509.MarshalECPrivateKey(key)
```
ItalyPaleAle, line 118: PKCS#8? :)

pkg/security/pem/pem.go
```diff
@@ -0,0 +1,192 @@
...
+		if err != nil {
+			return nil, err
+		}
+		blockType = "EC PRIVATE KEY"
+	case *rsa.PrivateKey:
```
ItalyPaleAle, line 123: PKCS#8? :)

cmd/sentry/main.go
```diff
@@ -69,6 +68,11 @@ func main() {
 		kubeconfig = flag.String("kubeconfig", "", "absolute path to the kubeconfig file")
 	}
 	flag.Parse()
+
+	if len(*depTA) > 0 {
+		log.Warn("--token-audience is deprecated and will be removed in v1.12")
```
ItalyPaleAle, line 73: Please also add a link to docs or somewhere that explains how to update

ItalyPaleAle (changes requested):
This is my first round of review, although there are still a few files missing.

One question is on why we are still using both a root CA cert and and intermediate CA cert. Can't we remove one from the chain, and just use the root CA?

Also, since this is a major rewrite and reviewing this PR is not super easy (because the diff isn't much helpful), would you be available for a review during a call?

mukundansundar (commented):
first set of comments.

JoshVanL:
Thanks for the review @ItalyPaleAle!

> This is my first round of review, although there are still a few files missing.

> One question is on why we are still using both a root CA cert and and intermediate CA cert. Can't we remove one from the chain, and just use the root CA?

Could do in future, but since this is what we are doing already today we should keep doing that until we are in a future version, past needing to rely on any old PKI. Remember that all components are currently consuming the issuing certificate and private key and using that for serving and clients. Changing up the chain structure will add to the backwards/forwards compat complexity. Ideally, we should not care what PKI setup Dapr is using so long as it follows correct PKI.

> Also, since this is a major rewrite and reviewing this PR is not super easy (because the diff isn't much helpful), would you be available for a review during a call?

Yes, definitely makes sense to do this! :slightly_smiling_face:

pkg/sentry/server/validator/kubernetes/kubernetes_test.go
```diff
@@ -0,0 +1,878 @@
...
+*/
+
+package kubernetes
+
+import (
+	"context"
+	"fmt"
+	"testing"
+
+	"github.com/golang-jwt/jwt"
```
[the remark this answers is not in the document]
JoshVanL: Good catch. Will add everything up to `v4` to the CI block list.

pkg/sentry/server/validator/kubernetes/kubernetes.go
```diff
@@ -0,0 +1,281 @@
...
+
+package kubernetes
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"strings"
+
+	"github.com/golang-jwt/jwt/v4"
```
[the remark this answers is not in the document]
JoshVanL: Sure thing. Any particular reason to use this library over `golang-jwt` since that seems more popular?

mukundansundar:
@JoshVanL what will be the effect of these sentry changes wrt the CLI renew certificate command?

pkg/sentry/server/ca/cert.go
JoshVanL: `Root CA` is the self signed root certificate authority. `Issuer CA` is a certificate authority which has been signed by the `Root CA`, also known as `subordinate`, or `intermediate` CA. It was called `issuer CA` in the code base already, so have kept the same wording.

`Root CA` -> `Issuer CA` -> `Workload CA`

charts/dapr/charts/dapr_sentry/templates/dapr_sentry_deployment.yaml
```diff
@@ -115,7 +113,7 @@ spec:
 {{ toYaml .Values.resources | indent 10 }}
         volumeMounts:
           - name: credentials
-            mountPath: /var/run/dapr/credentials
+            mountPath: /var/run/secrets/dapr.io/credentials
```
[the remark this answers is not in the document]
JoshVanL: To be consistent with Kubernetes who uses `kubernetes.io`.

cmd/sentry/main.go
```diff
@@ -50,11 +49,11 @@ const (
...
-	flag.StringVar(&credentials.RootCertFilename, "issuer-ca-filename", credentials.RootCertFilename, "Certificate Authority certificate filename")
-	flag.StringVar(&credentials.IssuerCertFilename, "issuer-certificate-filename", credentials.IssuerCertFilename, "Issuer certificate filename")
-	flag.StringVar(&credentials.IssuerKeyFilename, "issuer-key-filename", credentials.IssuerKeyFilename, "Issuer private key filename")
 	trustDomain := flag.String("trust-domain", "localhost", "The CA trust domain")
-	tokenAudience := flag.String("token-audience", "", "Expected audience for tokens; multiple values can be separated by a comma")
+	rootCertFilename := flag.String("issuer-ca-filename", config.DefaultRootCertFilename, "Certificate Authority certificate filename")
+	issuerCertFilename := flag.String("issuer-certificate-filename", config.DefaultIssuerCertFilename, "Issuer certificate filename")
+	issuerKeyFilename := flag.String("issuer-key-filename", config.DefaultIssuerKeyFilename, "Issuer private key filename")
+	depTA := flag.String("token-audience", "", "DEPRECATED")
```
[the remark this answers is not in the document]
JoshVanL: The audience should always be the spiffe ID of sentry.

pkg/sentry/server/validator/validator.go
```diff
@@ -0,0 +1,33 @@
...
+
+	"github.com/spiffe/go-spiffe/v2/spiffeid"
+
+	sentryv1pb "github.com/dapr/dapr/pkg/proto/sentry/v1"
+)
+
+// Interface is used to validate the identity of a certificate requester by
+// using an ID and token.
+// Returns the trust domain of the certificate requester.
+type Interface interface {
```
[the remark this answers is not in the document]
JoshVanL: I preferred to have `Interface` to avoid stuttering.

pkg/sentry/server/validator/internal/common_test.go
```diff
@@ -0,0 +1,101 @@
...
+	"fmt"
+	"testing"
+
+	"github.com/spiffe/go-spiffe/v2/spiffeid"
+	"github.com/stretchr/testify/assert"
+
+	sentryv1pb "github.com/dapr/dapr/pkg/proto/sentry/v1"
+)
+
+func Test_Validate(t *testing.T) {
```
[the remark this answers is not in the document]
JoshVanL: Why?

I use `_` as a separator so that when testing un-exported functions, we can maintain the case of the tested function. It also improves the readability of test run output.

i.e. `Testfoo` vs `Test_foo`.

pkg/sentry/server/validator/kubernetes/kubernetes.go
```diff
@@ -0,0 +1,281 @@
...
+	log    = logger.NewLogger("dapr.sentry.identity.kubernetes")
+	scheme = runtime.NewScheme()
+)
+
+func init() {
+	if err := configv1alpha1.AddToScheme(scheme); err != nil {
+		panic(err)
+	}
+	if err := corev1.AddToScheme(scheme); err != nil {
+		panic(err)
```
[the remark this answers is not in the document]
JoshVanL: An error here would signal that something is fundamentally broken in our APIs, or those imported. We should catch these.

pkg/security/spiffe/spiffe_test.go
```diff
@@ -0,0 +1,107 @@
...
+import (
+	"errors"
+	"fmt"
+	"testing"
+
+	"github.com/spiffe/go-spiffe/v2/spiffeid"
+	"github.com/stretchr/testify/assert"
+)
+
+func Test_FromID(t *testing.T) {
```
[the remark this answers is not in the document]
JoshVanL: See above

pkg/sentry/certs/certs.go
```diff
@@ -1,172 +0,0 @@
...
-	}
-	csr, err := x509.ParseCertificateRequest(block.Bytes)
-	if err != nil {
-		return nil, fmt.Errorf("failed to parse X.509 certificate signing request: %w", err)
-	}
-	return csr, nil
-}
-
-// GenerateECPrivateKey returns a new EC Private Key.
-func GenerateECPrivateKey() (*ecdsa.PrivateKey, error) {
```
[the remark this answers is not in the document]
JoshVanL: What is the intention for the dapr CLI to use the same certificate generation code as sentry?

pkg/sentry/server/validator/internal/common_test.go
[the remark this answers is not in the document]
mukundansundar: by convention the name starts with `TestXyz` https://pkg.go.dev/testing

pkg/sentry/certs/certs.go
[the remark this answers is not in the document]
mukundansundar: to generate a new certificate for renewal if the old certificate expires.

pkg/sentry/server/validator/validator.go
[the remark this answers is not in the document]
mukundansundar: again this is by convention https://go.dev/doc/effective_go#interface-names 
generally having interface names based on the action that they perform is better.

pkg/sentry/server/validator/kubernetes/kubernetes.go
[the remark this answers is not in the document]
mukundansundar: This is still run time panic, this is not going to be during dev alone right? Or is it expected that this will help catch issues during dev?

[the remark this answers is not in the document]
JoshVanL: Correct, this will catch issuing during dev. If this errors, the only thing we can ever do it terminate the program with an error anyway.

pkg/sentry/server/validator/kubernetes/kubernetes.go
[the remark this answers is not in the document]
JoshVanL: Makes sense :+1:

cmd/sentry/main.go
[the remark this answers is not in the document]
ItalyPaleAle: This is a matter of Dapr communicating with the control plane, so this change doesn't have an impact on users and doesn't require code changes from them. The only reason to support the deprecated option is so clusters can be upgraded doing a rollout and things will continue to work.

IMHO this doesn't need to adhere to the 2 releases policy since it's an internal thing.

ItalyPaleAle:
> > One question is on why we are still using both a root CA cert and and intermediate CA cert. Can't we remove one from the chain, and just use the root CA?
> 
> Could do in future, but since this is what we are doing already today we should keep doing that until we are in a future version, past needing to rely on any old PKI. Remember that all components are currently consuming the issuing certificate and private key and using that for serving and clients. Changing up the chain structure will add to the backwards/forwards compat complexity. Ideally, we should not care what PKI setup Dapr is using so long as it follows correct PKI.

The problem with the current approach is that because we throw away the root key, we can not renew the root CA certs. 10 years is a long time (and is probably too long), and yet that's how we got the Y2K and Y2K38 problems :) 

Since we're making so drastic changes to Sentry, I think this would be the perfect time to remove this choice, which IMHO doesn't even make much sense (I get protecting the root CA key, but if the intermediate CA can sign certificates, this makes no difference at the end of the day).

cmd/sentry/main.go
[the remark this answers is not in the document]
mukundansundar: But the users have to make changes to the `--token-audience` flag right? At the end of the day it is a user facing change.

JoshVanL:
> 

@ItalyPaleAle Completely agree- lets do this in a followup PR.

I would like to see a better option set for sentry so the user can tune these things (with more sane defaults). This is also where we can introduce external CAs.

cmd/sentry/main.go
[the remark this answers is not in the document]
ItalyPaleAle: The token-audience flag is normally set by Helm automatically with a default value, although users can override that (new thing added in 1.10)

But after this, the flag would become useless anyways.

pkg/sentry/certs/certs.go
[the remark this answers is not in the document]
JoshVanL: I have exported `GenerateCABundle`

pkg/sentry/server/ca/cert.go
DeepanshuA: As there are already major overhaul changes going on, should we also rename Issuer CA to Intermediate CA? Intermediate CA is quickly understandable, whereas all this while I too have been thinking that what is this Issuer CA.

charts/dapr/charts/dapr_sentry/templates/dapr_sentry_deployment.yaml
```diff
@@ -9,16 +7,16 @@ metadata:
...
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: dapr-trust-bundle
+  labels:
+    app: dapr-sentry
+    {{- range $key, $value := .Values.global.k8sLabels }}
+    {{ $key }}: {{ tpl $value $ }}
+    {{- end }}
```
DeepanshuA, line 19: Do we also need to enable encryption or provide an option for that? https://kubernetes.io/docs/tasks/administer-cluster/encrypt-data/#configuration-and-determining-whether-encryption-at-rest-is-already-enabled

pkg/security/security.go
```diff
@@ -0,0 +1,207 @@
...
+func CurrentNamespace() string {
+	namespace, ok := os.LookupEnv("NAMESPACE")
+	if !ok {
+		return "default"
+	}
+	return namespace
+}
+
+// SentryID returns the SPIFFE ID of the sentry server.
+func SentryID(sentryTrustDomain, sentryNamespace string) (spiffeid.ID, error) {
```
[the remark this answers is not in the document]
JoshVanL: Yep, that is the plan :slightly_smiling_face:

You can see [here](https://github.com/dapr/dapr/compare/master...JoshVanL:dapr:sentry-impl-wip-8) in my "north star" diff which I am working towards constructs the expected SPIFFE ID in each control plane component client.

I am tackling each component in turn. Here are the drafts for the operator and placement PRs
https://github.com/dapr/dapr/pull/6193
https://github.com/dapr/dapr/pull/6179

JoshVanL:
> As there are already major overhaul changes going on, should we also rename Issuer CA to Intermediate CA? Intermediate CA is quickly understandable, whereas all this while I too have been thinking that what is this Issuer CA.

Sure thing, more  than happy to make this change!

charts/dapr/charts/dapr_sentry/templates/dapr_sentry_deployment.yaml
DeepanshuA: Thanks for the explanation. It helps.
Just for my understanding: So, when we say that Trust Anchor can be accessed from ConfigMap using volume mount, so can any non-Dapr process inside Pod be also able to access this volume and use this Trust Anchor to get itself Authenticated from Sentry?

JoshVanL: This trust anchor is used by clients (operator, placement, injector) to verify the server (sentry) connection. We use a ConfigMap here because:
1. we don't want any component, apart from Sentry, to be mounting the intermediate certificate and private key,
2. ConfigMap is a better resource type in this scenario because the TrustAnchor is not secret data,
3. In future when we implement other CA signing sources (such as cert-manager), the intermediate Secret may never exist, so using dedicated ConfigMap for the TrustAnchor means that will always exist regardless of the signing/intermediate CA provider.

---

Daprd gets this TrustAnchor from the injector, set as a environment variable, which itself gets it from this ConfigMap. There may be an argument for implementing a controller to distribute the TrustAnchor to all (dapr enabled) namespaces as well so that `daprd`s are able to update their trust store without being restarted (environment variables cannot be updated in place, whereas file mounts can).

---

We likely want to move to using `ClusterTrustBundles` once all supported Kubernetes versions serve this API.

https://github.com/kubernetes/enhancements/issues/3257

pkg/sentry/server/validator/kubernetes/kubernetes.go
```diff
@@ -0,0 +1,262 @@
...
+	if req.Namespace+":"+pod.Spec.ServiceAccountName == req.Id {
+		req.Id = expID
+	}
+
+	if expID != req.Id {
+		return spiffeid.TrustDomain{}, fmt.Errorf("%s: app-id mismatch. expected: %s, received: %s",
+			errPrefix, expID, req.Id)
+	}
+
+	configName, ok := pod.GetAnnotations()[annotations.KeyConfig]
```
ItalyPaleAle, line 191: When #6101 is merged, there could be more than 1 config referenced. The annotation in this case will be a comma-separated list.

pkg/sentry/server/validator/kubernetes/kubernetes.go
```diff
@@ -0,0 +1,262 @@
...
+		return spiffeid.TrustDomain{}, errMissingPodClaim
+	}
+
+	var pod corev1.Pod
+	err = k.client.Get(ctx, types.NamespacedName{Namespace: req.Namespace, Name: claims.Pod.Name}, &pod)
+	if err != nil {
+		log.Errorf("failed to get pod %s/%s for requested identity: %s", req.Namespace, claims.Pod.Name, err)
+		return spiffeid.TrustDomain{}, fmt.Errorf("%s: failed to get pod of identity", errPrefix)
+	}
+	expID, ok := pod.GetAnnotations()[annotations.KeyAppID]
```
ItalyPaleAle, line 168: Please don't read from the annotations, but inspect the flags passed to the `daprd` binary directly (also for the configurations below).

Although uncommon, it's entirely possible for someone to deploy an app on K8s skipping the Dapr Injector, by adding the `daprd` container directly. In this case, there'd be no annotations.

ItalyPaleAle (changes requested):

pkg/sentry/server/validator/kubernetes/kubernetes.go
JoshVanL: This is wrong- whilst also fragile, the Kubernetes validator should be using Kubernetes primitives to evaluate the identity. Users who are not using the injector are still expected add appropriate annotations to pod templates.

ItalyPaleAle: We should make sure this is clearly documented then. Right now I don't believe that's documented.

pkg/sentry/server/validator/kubernetes/kubernetes.go
JoshVanL: What trust domain will the app belong to if multiple configs are assigned? I don't think the current trust domain system is compatible with having multiple config per app

ItalyPaleAle: Multiple configs works in last-one wins fashion, if more than one config sets the same option.

ItalyPaleAle:
@JoshVanL Good news :)

- The E2E tests here confirm that we can run dapr 1.10 and 1.11 sidecars against "vNext" (with this PR)
   - Required a bit of a "hack" because current Dapr versions request a TLS certificate from Sentry by sending a CSR that has type "CERTIFICATE" in the PEM block, instead of "CERTIFICATE REQUEST". See: https://github.com/dapr/dapr/pull/6171/commits/4fcdf9317399ab883f348d6a47cc80172565a03f
- Although we cannot write an E2E test for that, I manually verified that we can deploy a Dapr 1.11 control plane and use "vNext" sidecars. The sidecars do get a TLS certificate correctly from v1.11 Sentry, and things like service invocation works.
   - Although this scenario is not officially supported, it's something we still make an effort to allow, especially because there can be a version mismatch during upgrades.

ItalyPaleAle (approved):

ItalyPaleAle (approved):

artursouza (commented):
@yaron2 I think we need to merge this after tests pass. We need additional testing after but this is a (needed) rewrite of sentry.

yaron2:
> > > One question is on why we are still using both a root CA cert and and intermediate CA cert. Can't we remove one from the chain, and just use the root CA?
> > 
> > 
> > Could do in future, but since this is what we are doing already today we should keep doing that until we are in a future version, past needing to rely on any old PKI. Remember that all components are currently consuming the issuing certificate and private key and using that for serving and clients. Changing up the chain structure will add to the backwards/forwards compat complexity. Ideally, we should not care what PKI setup Dapr is using so long as it follows correct PKI.
> 
> The problem with the current approach is that because we throw away the root key, we can not renew the root CA certs. 10 years is a long time (and is probably too long), and yet that's how we got the Y2K and Y2K38 problems :)
> 
> Since we're making so drastic changes to Sentry, I think this would be the perfect time to remove this choice, which IMHO doesn't even make much sense (I get protecting the root CA key, but if the intermediate CA can sign certificates, this makes no difference at the end of the day).

Dapr root and issuer certs are valid for 1 year, not 10.

artursouza:
This PR is breaking service invocation and tracing ITs in Java SDK.

JoshVanL:
@artursouza those Java failing tests look unrelated to this PR. Here is a similar failure which was executed from the SDK test schedule https://github.com/dapr/dapr/actions/runs/5686772228

artursouza (approved):

artursouza:
@JoshVanL This will be a release blocker.

pkg/runtime/security/auth.go
```diff
@@ -109,14 +108,22 @@ func (a *authenticator) CreateSignedWorkloadCert(id, namespace, trustDomain stri
...
 
+	token, tokenValidator, err := securityToken.GetSentryToken(true)
+	if err != nil {
+		return nil, fmt.Errorf("error obtaining token: %w", err)
+	}
 	resp, err := c.SignCertificate(context.Background(),
 		&sentryv1pb.SignCertificateRequest{
-			CertificateSigningRequest: certPem,
+			CertificateSigningRequest: csrPem,
 			Id:                        getSentryIdentifier(id),
```
a-shoemaker, line 118: This intentionally not just the app ID?  (Instead, being SENTRY_LOCAL_IDENTITY which can be > 64 characters)

pkg/sentry/server/validator/kubernetes/kubernetes.go
```diff
@@ -0,0 +1,241 @@
...
+		log.Errorf("Service account on pod %s/%s does not match token", req.Namespace, claims.Pod.Name)
+		return spiffeid.TrustDomain{}, errors.New("pod service account mismatch")
+	}
+
+	// TODO: @joshvanl: Before v1.11, the injector instructed daprd to request
+	// for the ID containing their namespace and service account (ns:sa). This
+	// is wrong- dapr identities are based on daprd namespace + _app ID_.
+	// Remove this allowance in v1.12.
+	if req.Namespace+":"+pod.Spec.ServiceAccountName == req.Id {
+		req.Id = expID
```
a-shoemaker, line 181: But you validated the req.ID is < 64 characters before this even though it wasn't the app ID.  Which subsequently seems to make 1.12 unusable if you have any namespaces + app ID's of much length at ll.

pkg/runtime/security/auth.go
JoshVanL: Hi @a-shoemaker, indeed it has been done to ensure backwards compatibility with a 1.11 sentry server. The intention is to use the app ID in 1.13.

pkg/sentry/server/validator/kubernetes/kubernetes.go
JoshVanL: Thanks @a-shoemaker, indeed this looks like a bug in 1.12 which needs to be fixed.

JoshVanL: @a-shoemaker I have opened up an issue [here](https://github.com/dapr/dapr/issues/7079) to track.

PR #6172, opened 2023-03-31
Disable actor state e2e for cosmos db

JoshVanL:
Currently, e2e test runs are failing on actor state TTL https://github.com/dapr/dapr/actions/workflows/dapr-test.yml

This looks to be because we are using the CosmosDB without TTL being enabled at the container level.

PR enables TTL, but set _no_ default value.

See: https://learn.microsoft.com/en-us/azure/cosmos-db/nosql/how-to-time-to-live?tabs=dotnet-sdk-v3


ItalyPaleAle (approved):

mukundansundar (approved):

mukundansundar:
@JoshVanL the tests seem to fail with a similar error.

ItalyPaleAle (changes requested):

ItalyPaleAle (approved):

mukundansundar:
For now the newly added E2E actor state test has been disabled for cosmos's

ItalyPaleAle (changes requested):
I think we figured out the issue.

Can you merge this commit?

```sh
curl -L "https://github.com/dapr/dapr/commit/ab73a6dee28c79cbf2b5a355b07c6aae0a49c57e.patch" | git am
```

Then if you re-enable the test, it should work

It was my mistake, I accidentally corrupted the partition key for cosmos DB while storing actor state. We'll need this PR ported into the release branch then

JoshVanL:
Thanks @ItalyPaleAle, done!

ItalyPaleAle:
Fixed now in #6199, so I'll close this

Additional issues included:

- In the Bicep file, the TTL property had to be put inside `resources`, otherwise it wasn't recongized
- There was a race condition in `assert.Eventually`. Because the function defined inside is executed in a background goroutine, storing state in variables defined outside of the lambda was causing data race conditions (our GitHub Actions agents seem to be very slow...

PR #6179, opened 2023-04-03
[5756 Identity Management Overhaul] Placement

JoshVanL:


charts/dapr/charts/dapr_placement/templates/dapr_placement_statefulset.yaml
```diff
@@ -177,9 +175,16 @@ spec:
                 fieldPath: metadata.namespace
       serviceAccountName: dapr-placement
       volumes:
-        - name: credentials
-          secret:
-            secretName: dapr-trust-bundle
+      - name: dapr-trust-bundle
+        configMap:
+          name: dapr-trust-bundle
+      - name: dapr-identity-token
```
ItalyPaleAle, line 181: What about kubelets that don't supported projected volumes? Is there a way to make this optional (so users can configure Sentry to use JWTs signed somewhere else instead)?

cmd/placement/options/options.go
```diff
@@ -87,6 +97,10 @@ func New() *Options {
 
 	flag.Parse()
 
+	if len(*depRCF) > 0 || len(*depICF) > 0 || len(*depIKF) > 0 || len(*depCC) > 0 {
+		log.Warn("--certchain, --issuer-ca-filename, --issuer-certificate-filename and --issuer-key-filename are deprecated and will be removed in v1.12.")
```
ItalyPaleAle, line 101: 1.13? Or 1.14?

pkg/security/security.go
```diff
@@ -148,12 +160,46 @@ func (p *provider) Start(ctx context.Context) error {
...
+				log.Infof("watching trust anchors file %q for changes", p.trustAnchorsFile)
+				return fswatcher.Watch(ctx, p.trustAnchorsFile, caEvent)
+			},
+			func(ctx context.Context) error {
+				for {
+					select {
+					case <-ctx.Done():
+						return nil
+					case <-caEvent:
+						log.Info("trust anchors file changed, reloading trust anchors")
```
ItalyPaleAle, line 182: Nit:
```suggestion
						log.Info("Trust anchors file changed, reloading trust anchors")
```

pkg/security/security.go
```diff
@@ -205,3 +259,19 @@ func SentryID(sentryTrustDomain, sentryNamespace string) (spiffeid.ID, error) {
...
+
+func (x *x509source) updateTrustAnchorFromFile(filepath string) error {
+	rootPEMs, err := os.ReadFile(filepath)
+	if err != nil {
+		return fmt.Errorf("failed to read trust anchors file %q: %s", filepath, err)
+	}
+
+	trustAnchorCerts, err := secpem.DecodePEMCertificates(rootPEMs)
+	if err != nil {
+		return fmt.Errorf("failed to decode trust anchors: %s", err)
```
ItalyPaleAle, line 271:
```suggestion
		return fmt.Errorf("failed to decode trust anchors: %w", err)
```

pkg/security/security.go
```diff
@@ -148,12 +160,46 @@ func (p *provider) Start(ctx context.Context) error {
...
+				for {
+					select {
+					case <-ctx.Done():
+						return nil
+					case <-caEvent:
+						log.Info("trust anchors file changed, reloading trust anchors")
+
+						p.sec.source.lock.Lock()
+						if uErr := p.sec.source.updateTrustAnchorFromFile(p.trustAnchorsFile); uErr != nil {
+							log.Errorf("failed to read trust anchors file %q: %s", p.trustAnchorsFile, uErr)
```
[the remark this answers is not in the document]
JoshVanL: What's wrong with '%q'?

charts/dapr/charts/dapr_placement/templates/dapr_placement_statefulset.yaml
ItalyPaleAle: We have at least one large customer who depends on a kubelet without support for projected volumes.

pkg/security/security.go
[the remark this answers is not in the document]
ItalyPaleAle: Will encode slashes so it becomes harder to read. https://go.dev/play/p/vdBrYBVp2wO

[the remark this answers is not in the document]
JoshVanL: Gotcha, makes sense. I agree :+1:

charts/dapr/charts/dapr_placement/templates/dapr_placement_statefulset.yaml
JoshVanL: OK, can we do this in another PR as this is going to effect all the other components anyway so will be cleaner to do all of them at once in a dedicated PR?

ItalyPaleAle: Ok, can you please open an issue so we don't forget this is a P0?

JoshVanL: I've added it to the list of the parent issue https://github.com/dapr/dapr/issues/5756

ItalyPaleAle:
Can you please test this to make sure that a Dapr 1.11 sidecar can connect to the placement service that uses the new security? We have an E2E test that tests for upgrades, but not for actors

JoshVanL:
@ItalyPaleAle is this not the same as what the test-version-skew workflow is doing?

JoshVanL:
> > @ItalyPaleAle is this not the same as what the test-version-skew workflow is doing?
> 
> thinking when was test-version-skew added, I didn't know about it. Yes seems like it should be ok

@ItalyPaleAle  https://github.com/dapr/dapr/pull/6374 :smile:

JoshVanL:
Version skew daprd-sidecar-master e2e tests are failing because of a change in an error string from [this PR](https://github.com/dapr/dapr/pull/6703/files#diff-0c175408a25c2ea4a2568014b413734c677c9ab72972a6227e9b75c8b1d782e2), but otherwise this PR is passing.

artursouza (approved):

MregXN:
Hi @JoshVanL and @artursouza. This PR seems to directly cause the E2E test `actor_metadata` to fail.

 Before this PR was merged, test `actor_metadata` will fail with a certain probability due to network delay. But after the PR was merged, the test has been failing.

This can be observed in [E2E test log](https://github.com/dapr/dapr/actions/workflows/dapr-test.yml) after August 15th (the date when this PR was merged) . It can be found that all `actor_metadata` tests failed with the same reason and have not succeeded since.

Could you please take a look at this issue? Thank you.

Issue #5873, opened 2022-03-03
Verify performance regression with nacos binding upgrade and add it back


shubham1172:
## Describe the feature
Dapr introduced the Alibaba Nacos binding with dapr/components-contrib#814 in v1.2 which was then reverted back in the endgame with https://github.com/dapr/dapr/pull/3212, after a performance regression https://github.com/dapr/dapr/issues/2479.

Nacos was upgraded to a newer version with dapr/components-contrib#1478 (and many other PRs since), this issue is to verify if the performance regression is still happening and add it back to daprd if it's fixed.

## Release Note

RELEASE NOTE: ADD binding for Nacos configuration service

berndverst:
I'm making this a P1 for v1.11 in runtime just so someone will actually go through the effort of adding this component back (after many dependency upgrades occurred) to evaluate whether adding this back will cause a performance regression or not.

PR #3212, opened 2021-05-22
Remove nacos and rocketmq after performance regression investigation

wcs1only:
# Description

Screen Shot 2021-05-21 at 1 34 05 PM

Screen Shot 2021-05-21 at 8 33 31 PM

CPU usage regression was discovered in 1.2.0-rc.4 during longhaul tests. After some profiling, we discovered that these two components both spawn background goroutines that actively use CPU, even if the components themselves are not loaded. We'll remove them to unblock the 1.2.0 release and re-add them once the CPU usage regression is addressed.

## Checklist

Please make sure you've  completed the relevant tasks for this PR, out of the following list:

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing
* [ ] End-to-end tests passing
* [ ] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_
* [ ] Specification has been updated / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_
* [ ] Provided sample for the feature / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_

PR #6181, opened 2023-04-04
Register alicloud nacos binding component

liu-song:
# Description
It seems that  forgot to add,  Add alicloud nacos binding component
https://github.com/dapr/components-contrib/tree/master/bindings/nacos

## Issue reference

Please reference the issue this PR will close: #_[https://github.com/dapr/dapr/issues/5873]_

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing


mukundansundar (dismissed):

liu-song:
I just realized there is an issue with the nacos-go-sdk and I will try to confirm  it.https://github.com/dapr/dapr/issues/5873

liu-song:
First, let's review the previous issue and the code fragment that was fixed.
https://github.com/dapr/dapr/pull/3212

nacos-sdk-go issue
https://github.com/nacos-group/nacos-sdk-go/issues/250

nacos-sdk-go fix
https://github.com/nacos-group/nacos-sdk-go/pull/251/files

Before the issue was fixed, the nacos-go-sdk would initialize a map and establish a goroutine during the init() function. 
https://github.com/nacos-group/nacos-sdk-go/pull/251/files#diff-3f155c99c6b9cfd0acfbb7bcc7a22c3384d0c5400130670f6966ef899de0a7fdL79-R81
After the fix, the map and goroutine are only created when they are actually used, which ensures that the nacos component does not consume additional resources in Dapr when it is not being used.

liu-song:
At the same time, I also monitored the differences between running the nacos-go-sdk with the same code before the fix and with the latest version.
before fix
code
https://github.com/liu-song/registry-nacos/tree/beforefix/example/basic/server
nacos-go-sdk version
https://github.com/liu-song/registry-nacos/blob/beforefix/go.mod#L9
latest  versioin
code 
https://github.com/liu-song/registry-nacos/blob/latest/example/basic/server/main.go
nacos-go-sdk version
https://github.com/liu-song/registry-nacos/blob/latest/go.mod#L10

image

liu-song:
The above code is a simple example using Kitex as an RPC framework and Nacos for service registration and discovery. The section before the blue dotted line shows the code running with the "beforefix" version, and the section after the line shows the code running with the latest version.

My computer has a configuration of 16GB RAM and 8 cores. From the graph, we can observe that the memory usage of the program ranges from 0.1% to 0.13%, the number of goroutines ranges from 11 to 16, and the CPU usage ranges from 0% to 2%, with most of the time spent at 1%.

One noteworthy point is that the latest version has reduced the number of times the CPU usage reaches 2%, indicating that the new version of nacos-go-sdk can reduce some CPU consumption. Overall, using nacos-go-sdk correctly does not result in excessive resource consumption.

Finally, if the binding  nacos is actually applied and brings additional resource consumption, providing more code information, conditions for reproduction, or pprof data can help us address the issue and make nacos-go-sdk more robust.

liu-song:
Could you please review this issue again?   :grin:
@shubham1172 @yaron2 @berndverst @mukundansundar

liu-song:
Do I need to provide any other additional information?

ItalyPaleAle:
I think we still need this?

ItalyPaleAle:
Are we still interested in this? Otherwise, I think we should just remove the component from components-contrib

liu-song:
There may not be a significant demand, so temporarily close it.

ItalyPaleAle:
Opened https://github.com/dapr/components-contrib/pull/3118 to delete the component's code

Issue #6073, opened 2023-03-14
[Proposal] Map registered workflow types to unique actor types


cgillum:
## In what area(s)?

/area runtime
/area docs

### Background

In Dapr v1.10, only one workflow app was allowed to be registered in a cluster. This was _not_ an intentional design decision and resulted in hard-to-debug errors about workflows not existing. More discussion on this problem is tracked in https://github.com/dapr/dapr/issues/6013.

The root of the problem is that all workflow and activity actor types are the same, either `dapr.internal.wfengine.workflow` or `dapr.internal.wfengine.activity`, which made it impossible to correctly route workflow messages to the correct app.

### Change proposal summary

Instead of having two fixed workflow and activity actor types, we propose to have a unique workflow/activity actor type for each workflow/activity registered in the app.

As an example, if an app defines a workflow named _myworkflow1_ and an activity named _myactivity1_, the corresponding internal actor types will be registered as `dapr.internal.workflow.myworkflow1` and `dapr.internal.activity.myactivity1`.

> Note: we leave out "wfengine" from the type name since it isn't really useful and only adds more characters to the type name.

One downside of this change is that it results in a larger number of actor types registered in a cluster. A new proposal will be written and linked to this proposal that explains how we can work around known problems related to registering too many actor types in a cluster.

### Low-level engine changes

The `durabletask-go` engine will need to be updated to support registration of orchestration (workflow) and activity types since it doesn't require any such registration today. This change only affects the `GetWorkItems` API. A new parameter will be added that includes the list of orchestration and activity type names. These names will be made available to the `wfengine` Dapr package when the client SDK connects, and the workflow engine will then register the new actor types on-demand.

### Benefits of this change

This has the following key benefits over the current design:

1. It removes the restriction of having a single workflow app per cluster.
2. It allows workflows to schedule activities and child-workflows defined in other apps.

This is a natural change that arguably should have been the original design for workflows in Dapr. It's also more closely aligned with the design for actors.

### Workflow and activity name restrictions

Because workflow and activity type names are now included in the internal actor type name, we will put some restrictions on the workflow type names that users can register. Specifically, workflow and activity type names:

* Must be a valid identifier, consisting of only letters, numbers, or underscores (note, we can use [`token.IsIdentifier(str)`](https://go.googlesource.com/go/+/60a8dbf3b4e10627b9f5c3a0a0bf4462247270c8/src/go/token/token.go#333) as a starter implementation for this)
* Must be between 1 and 50 characters in length
[...]

Issue #6013, opened 2023-03-01
Workflow: Support for multiple workflow apps in a single cluster


cgillum:
## In what area(s)?

> /area runtime

## Describe the feature

Dapr Workflow only supports a single workflow app in a cluster. This is a known limitation of the workflow engine and is documented [here](https://docs.dapr.io/developing-applications/building-blocks/workflow/workflow-architecture/#workflow-scalability).

> the Dapr Workflow engine requires that all instances of each workflow app register the exact same set of workflows and activities. In other words, it’s not possible to scale certain workflows or activities independently. All workflows and activities within an app must be scaled together.

However, the burden is on the user to remember to abide by this rule. Trying to use multiple workflow apps in the same cluster results in the following runtime error, randomly:

> The workflow failed - OrchestratorTaskNotFound: No orchestrator task named 'XXX' was found.

This happens because the underlying workflow actor may get load balanced to a Dapr instance for an app that hasn't registered a particular workflow or activity.

Given that the current design is both limiting and easy to get wrong. We should improve the workflow engine to remove this limitation.

## Release Note

RELEASE NOTE: **ADD** Support for multiple workflow apps in a single cluster

olitomlinson:
> This happens because the underlying workflow actor may get load balanced to a Dapr instance for an app that hasn't registered a particular workflow or activity.

How is it possible to have a _daprized_ App which registers a `Workflow` and `Activity` (Lets call it `App A`) but have certain _instances_ of `App A` not have both `Workflow` and `Activity` Registered?

Surely the only way to achieve this is to have some non-deterministic logic in how an App registers its Workflows and Activities, to generate inconsistencies across the App Instances? But that would be crazy, right?

<img width="507" alt="image" src="https://user-images.githubusercontent.com/4224880/222290312-e7ccc6de-4f18-438a-b5ea-5f5619758fa0.png">

cgillum:
I should have been clearer in my description. The problem isn't multiple instances of a workflow app being inconsistent (as illustrated in your code snippet), but rather multiple workflow apps that have different workflows registered. For example:

* App1: WorkflowA, WorkflowB
* App2: WorkflowC, WorkflowD

[...]

PR #6192, opened 2023-04-05
Modify Actor type with details

DeepanshuA:
# Description

Modify Internal Actor Types, getting used in Workflows and Activities.

## Issue reference

Please reference the issue this PR is linked to: #6013 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Unit tests passing
* [x] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs https://github.com/dapr/docs/issues/3346


pkg/runtime/wfengine/wfengine.go
```diff
@@ -49,15 +49,32 @@ type WorkflowEngine struct {
...
+}
+
 func IsWorkflowRequest(path string) bool {
 	return backend.IsDurableTaskGrpcRequest(path)
 }
 
-func NewWorkflowEngine() *WorkflowEngine {
+func NewWorkflowEngine(config *WFConfig) *WorkflowEngine {
+	WorkflowActorType = WorkflowActorType + "." + config.AppID + "." + getNamespace()
+	ActivityActorType = ActivityActorType + "." + config.AppID + "." + getNamespace()
```
cgillum, line 77: nit: namespace should come before the app ID.

pkg/runtime/wfengine/wfengine.go
```diff
@@ -49,15 +49,32 @@ type WorkflowEngine struct {
...
 	errExecutionAborted = errors.New("execution aborted")
 )
 
+// WFConfig is the configuration for the workflow engine
+type WFConfig struct {
+	AppID string
+}
+
+func getNamespace() string {
+	namespace := os.Getenv("NAMESPACE")
```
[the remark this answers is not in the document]
DeepanshuA: I tried to look for ways to fetch namespace, and it seems we are doing similar at some other places too in runtime: https://github.com/dapr/dapr/blob/e988ea9b92f715a59aa3e8bacba8863b97ef4ce6/pkg/runtime/runtime.go#L356

pkg/runtime/wfengine/wfengine.go
DeepanshuA: Done

pkg/runtime/wfengine/wfengine.go
```diff
@@ -49,15 +49,32 @@ type WorkflowEngine struct {
...
+}
+
 func IsWorkflowRequest(path string) bool {
 	return backend.IsDurableTaskGrpcRequest(path)
 }
 
-func NewWorkflowEngine() *WorkflowEngine {
+func NewWorkflowEngine(config *WFConfig) *WorkflowEngine {
+	WorkflowActorType = WorkflowActorType + "." + getNamespace() + "." + config.AppID
+	ActivityActorType = ActivityActorType + "." + getNamespace() + "." + config.AppID
```
mukundansundar, line 77: ideally should be `ns.appID.<var>` namespace should be a prefix?

mukundansundar: Please add UTs for the same...

pkg/runtime/wfengine/wfengine_test.go
```diff
@@ -760,7 +760,9 @@ func startEngine(ctx context.Context, r *task.TaskRegistry) (backend.TaskHubClie
 }
 
 func getEngine() *wfengine.WorkflowEngine {
-	engine := wfengine.NewWorkflowEngine()
+	engine := wfengine.NewWorkflowEngine(&wfengine.WFConfig{
+		AppID: "test",
+	})
```
mukundansundar, line 765: in the line below https://github.com/dapr/dapr/pull/6192/files#diff-acb2ead1c43373f11db66d9821d6bfa8b1e16ee5566bc89fb0936cedefcd4019R781
Please do not panic, fail the test after planning in `t` to this method.

pkg/runtime/wfengine/wfengine.go
DeepanshuA: > ideally should be `ns.appID.<var>`. `namespace` should be a prefix?

Actually, the purpose is just to differentiate two Apps and if namespace available, then specifying that too, for better differentiation and recognition by underlying WF engine. 
Right now, many entries are formed in state Stores for a Workflow and Activities, with Internal Actor having representation like:

dapr.internal.wfengine.workflow.default.wfapp

dapr.internal.wfengine.activity.default.wfapp

I feel it would just elongate the Actor type naming, with `namespace` and `appId`, without serving any actual purpose.
WDYT?

DeepanshuA: > Please add UTs for the same...

Now, asserting WorkflowActorType and ActivityActorType in `TestStartWorkflowEngine`.

pkg/runtime/wfengine/wfengine.go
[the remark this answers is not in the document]
cgillum: Instead of copying that logic, let's instead make a call into the existing API to referenced to get the namespace. That way if they update it to support more than just environment variables, our logic will continue to be consistent with the rest of the system.

pkg/runtime/wfengine/wfengine.go
```diff
@@ -49,15 +47,32 @@ type WorkflowEngine struct {
 }
 
 var (
+	WorkflowActorType   = ""
+	ActivityActorType   = ""
```
cgillum, line 51: I don't feel great about having global variables that are mutable. Can we instead make these fields (or getter methods, like `func (wfe *WorkflowEngine) GetWorkflowActorType() string`) on the `WorkflowEngine` struct? I see that we initialize these in `NewWorkflowEngine` so I assume it would be easy to make this change. Or if we don't even need these at all, we can just remove them.

pkg/runtime/wfengine/wfengine_test.go
```diff
@@ -779,7 +783,7 @@ func getEngine() *wfengine.WorkflowEngine {
 	})
 
 	if err := actors.Init(); err != nil {
-		panic(err)
+		assert.FailNow(t, "failed to init actors: %s", err)
```
cgillum, line 786: It's better to use `assert.NoError(t, err)` for cases like this.

cgillum (commented):
This is looking good - just a few small things.

pkg/runtime/wfengine/wfengine.go
cgillum: @DeepanshuA I think Mukandan is just suggesting a rearrangement of the existing components, and not asking that we literally add "namespace" and "appId". How about if we want to rearrange to something like `dapr.internal.mynamespace.myapp.workflow` and `dapr.internal.mynamespace.myapp.activity`? That might make for a slightly cleaner hierarchy.

pkg/runtime/wfengine/wfengine_test.go
ItalyPaleAle: Or actually, `require.NoError` in this case. The difference is that `require` aborts the current test.

cgillum: > Or actually, `require.NoError` in this case. The difference is that `require` aborts the current test.

Oh wow, I didn't know about this. I have a lot of test code that I need to go back and clean up now. 😅

pkg/runtime/wfengine/wfengine.go
mukundansundar: yes that is what I meant.

pkg/runtime/wfengine/wfengine.go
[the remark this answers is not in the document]
DeepanshuA: Found one place in sentry, where they had similar use of using `default`, if nothing found.
I have placed a common logic in utils.

pkg/runtime/wfengine/backend.go
```diff
@@ -172,10 +182,14 @@ func (be *actorBackend) AddNewOrchestrationEvent(ctx context.Context, id api.Ins
 		return err
 	}
 
+	var actorType string
+	if be.config != nil {
+		actorType = be.config.WorkflowActorType
```
mukundansundar, line 187: is empty string a valid actor type? 
Should an error be returned if config is nil?

pkg/runtime/wfengine/wfengine.go
```diff
@@ -46,37 +42,50 @@ type WorkflowEngine struct {
...
 
 	return engine
 }
 
 // InternalActors returns a map of internal actors that are used to implement workflows
 func (wfe *WorkflowEngine) InternalActors() map[string]actors.InternalActor {
 	internalActors := make(map[string]actors.InternalActor)
-	internalActors[WorkflowActorType] = wfe.workflowActor
-	internalActors[ActivityActorType] = wfe.activityActor
+	if wfe.config != nil {
```
mukundansundar, line 85: when can the config here be nil? 
Please add UTs for the same as necessary.

utils/utils.go
```diff
@@ -198,3 +200,12 @@ func ComponentLogName(name, typ, version string) string {
 
 	return fmt.Sprintf(logNameVersionFmt, name, typ, version)
 }
+
+// GetNamespace returns the namespace for Dapr, or the default namespace if it is not set.
+func GetNamespaceOrDefault() string {
```
mukundansundar, line 205: Please add UTs for this.

pkg/runtime/wfengine/wfengine.go
```diff
@@ -46,37 +42,50 @@ type WorkflowEngine struct {
...
+}
+
 func IsWorkflowRequest(path string) bool {
 	return backend.IsDurableTaskGrpcRequest(path)
 }
 
-func NewWorkflowEngine() *WorkflowEngine {
+func NewWorkflowEngine(config *WFConfig) *WorkflowEngine {
+	config.WorkflowActorType = actors.InternalActorTypePrefix + utils.GetNamespaceOrDefault() + "." + config.AppID + ".workflow"
+	config.ActivityActorType = actors.InternalActorTypePrefix + utils.GetNamespaceOrDefault() + "." + config.AppID + ".activity"
```
mukundansundar, line 66: have `.` as a separator and const and have workflow and activity as consts as well.

mukundansundar, line 66: This is going to be a breaking change. Is that expected?
cc @cgillum

mukundansundar (commented):
Please add additional code paths after config nil checks, as in does it make sense to have actorType empty or ActivityActorType empty?

pkg/runtime/wfengine/backend.go
DeepanshuA: Was Doing that only, haven't yet requested re-review

DeepanshuA:
> Please add additional code paths after config nil checks, as in does it make sense to have actorType empty or ActivityActorType empty?

Have been doing that only, haven't yet requested re-review

pkg/runtime/wfengine/wfengine.go
cgillum: @mukundansundar yes, we're intentionally doing this as a breaking change as part of v1.11.

cgillum:
> > Please reference the issue this PR will close: #6073
> 
> @DeepanshuA this is referring an issue that has been withdrawn and closed without work. What is the right issue to link to this PR?

@DeepanshuA please link this PR instead directly to https://github.com/dapr/dapr/issues/6013.

DeepanshuA:
> > Please add additional code paths after config nil checks, as in does it make sense to have actorType empty or ActivityActorType empty?
> 
> Have been doing that only, haven't yet requested re-review :)

Alright, I am done with changes that I was intending to make. I will take care of rest of comments now. Thanks for the quick feedback.

pkg/runtime/wfengine/backend.go
DeepanshuA: I am done with the changes, hope this should be fine now

pkg/runtime/wfengine/wfengine.go
DeepanshuA: Ideally, config here won't be null as per the way current code is written, but I have placed a check accordingly.

pkg/runtime/wfengine/activity_test.go
```diff
@@ -12,28 +12,31 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 */
-package wfengine_test
+package wfengine
```
cgillum, line 15: Is it necessary to change the package name from `wfengine_test` to `wfengine`? I was hoping to avoid this to keep the tests more blackbox friendly.

DeepanshuA: Actually, will have to expose `config` outside wfengine, for testing purpose then. And, didn't want to do so, just for testing. Hence, used same package as source code.
Is there something specific that we would be able to achieve with black-box testing? What I could understand that both strategies are fine (ref: https://stackoverflow.com/questions/19998250/proper-package-naming-for-testing-with-the-go-language).

pkg/runtime/wfengine/wfengine.go
DeepanshuA: Added UT.

utils/utils.go
DeepanshuA: Added.

pkg/runtime/wfengine/wfengine.go
DeepanshuA: Please check now.

pkg/runtime/wfengine/wfengine.go
DeepanshuA: Please comment if this approach looks fine, or you think something better can be done to propagate this info.

cgillum:
Hi @mukundansundar - I noticed you removed this from the v1.11 GitHub milestone. Are we really removing this from the v1.11 release?

mukundansundar:
> Hi @mukundansundar - I noticed you removed this from the v1.11 GitHub milestone. Are we really removing this from the v1.11 release?

I added back the underlying issue into 1.11. Ideally we have either only the PR or the underlying issue tracked in the milestone.

pkg/runtime/wfengine/wfengine.go
```diff
@@ -46,38 +43,61 @@ type WorkflowEngine struct {
...
 
-func NewWorkflowEngine() *WorkflowEngine {
+func NewWorkflowEngine(config *WFConfig) *WorkflowEngine {
+	actorTypePerAppPrefix := strings.Builder{}
+	actorTypePerAppPrefix.WriteString(actors.InternalActorTypePrefix)
+	actorTypePerAppPrefix.WriteString(utils.GetNamespaceOrDefault())
+	actorTypePerAppPrefix.WriteString(utils.DotDelimiter)
+	actorTypePerAppPrefix.WriteString(config.AppID)
+	actorTypePerAppPrefix.WriteString(utils.DotDelimiter)
+	actorTypePerAppPrefixStr := actorTypePerAppPrefix.String()
```
cgillum, line 72: This manual string-building seems like unnecessary complexity. I would expect something like this for extremely performance-sensitive code. However, this is *not* performance sensitive (we initialize the engine exactly once) so I find this kind of code to be distracting and unnecessarily hard to understand.

pkg/runtime/wfengine/wfengine.go
```diff
@@ -46,38 +43,61 @@ type WorkflowEngine struct {
...
+
+	config.WorkflowActorType = actorTypePerAppPrefixStr + utils.WorkflowNameLabelKey
+	config.ActivityActorType = actorTypePerAppPrefixStr + utils.ActivityNameLabelKey
```
cgillum, line 75: It doesn't make sense to initialize the config object that's passed in as a constructor. I think of this as a bit of an anti-pattern since config objects should generally be immutable. It seems to me that you should simply pass compute the actor type value using the app ID and namespace and then just pass the actor type value into the `NewActivityActor` and `NewWorkflowActor` functions.

pkg/runtime/wfengine/activity_test.go
cgillum: `WFConfig` is already a public type, so it's already exposed outside of `wfengine`. But regardless, I think you should instead have a `engine.GetActivityActorType()` and `engine.GetWorkflowActorType()` rather than `engine.config.XXXActorType`.

Generally speaking, I want to minimize the churn in this PR. I feel like it can be much smaller than it is right now.

pkg/runtime/wfengine/wfengine.go
cgillum: Alternatively, consider initializing the workflow and activity actor types when constructing the `WFConfig` object (perhaps using a `NewWorkflowConfig` function) and exposing those types via read-only methods or fields.

cgillum:
> I added back the underlying issue 6013 into 1.11. Ideally we have either only the PR or the underlying issue tracked in the milestone.

Got it - thanks for the clarification.

pkg/runtime/wfengine/wfengine.go
DeepanshuA: Removed

pkg/runtime/wfengine/wfengine.go
DeepanshuA: Changed as discussed

pkg/runtime/wfengine/activity_test.go
DeepanshuA: Changed as discussed

utils/utils.go
```diff
@@ -26,6 +26,15 @@ import (
 	"k8s.io/client-go/tools/clientcmd"
 )
 
+const (
+	defaultNamespace = "default"
+	DotDelimiter     = "."
+
+	// Workflow
+	WorkflowNameLabelKey = "workflow"
+	ActivityNameLabelKey = "activity"
```
cgillum, line 35: I think it's probably better to define the workflow and activity const values in the wfengine package since they're really only applicable in the context of the workflow engine.

cgillum (approved):
This looks great - I'm much happier with this newest iteration. Thanks for your patience with applying the feedback!

pkg/runtime/wfengine/activity_test.go
```diff
@@ -30,10 +30,10 @@ import (
 func TestDedupeActivityInvocation(t *testing.T) {
 	// We do just enough to get the activity actor properly initialized
 	ctx := context.Background()
-	_, engine := startEngine(ctx, task.NewTaskRegistry())
+	_, engine := startEngine(ctx, t, task.NewTaskRegistry())
```
mukundansundar, line 33: every occurrence of `startEngine` seems to be preceded with a `context.Background` call ... Do you really need context to be passed in, can it just be init inside the `startEngine` function itself?

pkg/runtime/wfengine/wfengine.go
```diff
@@ -46,37 +42,56 @@ type WorkflowEngine struct {
...
 
+// wfConfig is the configuration for the workflow engine
+type wfConfig struct {
+	AppID             string
+	workflowActorType string
+	activityActorType string
+}
+
+// NewWorkflowConfig creates a new workflow engine configuration
+func NewWorkflowConfig(appID string) *wfConfig {
```
mukundansundar, line 61: why is a pointer to this struct returned? Just a normal struct is enough here?
Am I correct in saying that the value is written once and accessed multiple times, so in that case why is a pointer needed here?

utils/utils_test.go
```diff
@@ -270,3 +270,16 @@ func TestMapToSlice(t *testing.T) {
...
 }
+
+func TestGetNamespaceOrDefault(t *testing.T) {
+	t.Run("namespace is empty", func(t *testing.T) {
+		ns := GetNamespaceOrDefault()
+		assert.Equal(t, "default", ns)
+	})
+
+	t.Run("namespace is not empty", func(t *testing.T) {
+		os.Setenv("NAMESPACE", "testNs")
```
mukundansundar, line 281: use `t.Setenv` instead

pkg/runtime/wfengine/activity_test.go
DeepanshuA: Can do so but let's take a case like `TestEmptyWorkflow`: https://github.com/dapr/dapr/blob/3fb850d12c5306f1dd424cfa15b41d27f0c4e0e7/pkg/runtime/wfengine/wfengine_test.go#L107

There, we also need to pass `context` to `ScheduleNewOrchestration` and `WaitForOrchestrationCompletion`, and if I understand correctly, a Prod scenario will have same context being passed in there after starting Engine and then to schedule and wait for an orchestration. So, current test mimics almost similar scenario with same context being passed to them.
The tests would ideally also work with `context.TODO()` in all these places. But, I feel it is better to use a common context, which is being done currently.
WDYT?

pkg/runtime/wfengine/wfengine.go
DeepanshuA: Nice point. It doesn't need to be mutable. And, with a pointer, it would also need to do `escape analysis` and it might also land up in heap, where it is not needed actually. I will change this one.

pkg/runtime/wfengine/wfengine_test.go
```diff
@@ -600,8 +609,8 @@ func startEngine(ctx context.Context, r *task.TaskRegistry) (backend.TaskHubClie
 	return client, engine
```
mukundansundar, line 609: remove panic in line 607.

pkg/runtime/wfengine/wfengine_test.go
```diff
@@ -587,9 +596,9 @@ func TestPauseResumeWorkflow(t *testing.T) {
 	}
 }
 
-func startEngine(ctx context.Context, r *task.TaskRegistry) (backend.TaskHubClient, *wfengine.WorkflowEngine) {
+func startEngine(ctx context.Context, t *testing.T, r *task.TaskRegistry) (backend.TaskHubClient, *wfengine.WorkflowEngine) {
```
mukundansundar, line 599: is there a proper way to stop engine? If so please add code for that too. 
Are there race conditions in these tests anywhere? If not add them to the list of packages that do not have race conditions.

mukundansundar: if this requires more work create an issue and create a separate PR for the same.

DeepanshuA: I have created https://github.com/dapr/dapr/issues/6264 as it is a wider issue, separate from current PR.

pkg/sentry/certs/store.go
```diff
@@ -12,10 +12,7 @@ import (
 	"github.com/dapr/dapr/pkg/sentry/config"
 	"github.com/dapr/dapr/pkg/sentry/consts"
 	"github.com/dapr/dapr/pkg/sentry/kubernetes"
-)
-
-const (
-	defaultSecretNamespace = "default"
+	"github.com/dapr/dapr/utils"
```
mukundansundar, line 15: this package is already used in `sentry/config/config.go`. It's not newly added to this binary. 
https://github.com/dapr/dapr/blob/master/pkg/sentry/config/config.go#L15

pkg/sentry/certs/store.go
```diff
@@ -12,10 +12,7 @@ import (
 	"github.com/dapr/dapr/pkg/sentry/config"
 	"github.com/dapr/dapr/pkg/sentry/consts"
 	"github.com/dapr/dapr/pkg/sentry/kubernetes"
-)
-
-const (
```
mukundansundar, line 17: @DeepanshuA Please revert the changes to this file alone. 
Have the utils function ... but use the local var and local function defined in this file and not the utils one.

mukundansundar (changes requested):

pkg/sentry/certs/store.go
DeepanshuA: Accepting a param as defaultNamespace now in utils.

pkg/sentry/certs/store.go
```diff
@@ -12,6 +12,7 @@ import (
 	"github.com/dapr/dapr/pkg/sentry/config"
 	"github.com/dapr/dapr/pkg/sentry/consts"
 	"github.com/dapr/dapr/pkg/sentry/kubernetes"
+	"github.com/dapr/dapr/utils"
```
mukundansundar, line 15: utils package is already used in `sentry/config/config.go`

mukundansundar (approved):

PR #6193, opened 2023-04-05
[5756 Identity Management Overhaul] Operator

JoshVanL:


charts/dapr/charts/dapr_operator/templates/dapr_operator_deployment.yaml
```diff
@@ -75,6 +32,7 @@ spec:
         prometheus.io/scrape: "{{ .Values.global.prometheus.enabled }}"
         prometheus.io/port: "{{ .Values.global.prometheus.port }}"
         prometheus.io/path: "/"
+        dapr.io/app-id: dapr-operator
```
ItalyPaleAle, line 35: I would not set app-id in a control plane service. We use these for sidecars only

charts/dapr/charts/dapr_operator/templates/dapr_operator_deployment.yaml
```diff
@@ -208,12 +156,16 @@ spec:
...
-        - name: credentials
-          secret:
-            secretName: dapr-trust-bundle
-        - name: webhook-creds
-          secret:
-            secretName: dapr-webhook-cert
+        - name: dapr-trust-bundle
+          configMap:
+            name: dapr-trust-bundle
+        - name: dapr-identity-token
```
ItalyPaleAle, line 162: Leaving a comment here as a reminder that this too needs (in a separate PR) to have an option to support non-projected volumes

cmd/operator/options/options.go
```diff
@@ -89,6 +87,10 @@ func New() *Options {
 
 	flag.Parse()
 
+	if len(*depRCF) > 0 || len(*depICF) > 0 || len(*depIKF) > 0 || len(*depCCP) > 0 {
+		log.Warn("--certchain, --issuer-ca-filename, --issuer-certificate-filename and --issuer-key-filename are deprecated and will be removed in v1.12")
```
ItalyPaleAle, line 91: 1.14 here too

charts/dapr/charts/dapr_operator/templates/dapr_operator_deployment.yaml
JoshVanL: Control plane components will now all also receive SVIDs, and have SPIFFE IDs in the Dapr trust domain scheme. The `dapr.io/app-id` annotation is the mechanism which the Kubernetes sentry validator uses to check the requested app ID. This means that the annotation is now required for control plane components.

ItalyPaleAle: Should we use a different annotation for control-plane components however? Maintainers may have an opinion

JoshVanL: That would be fine by me.

JoshVanL:
/ok-to-test

Flake?

ItalyPaleAle:
1. Looks like "/test-version-skew" isn't working
2. For AKS, seems that we need to upgrade the version of K8s used in the tests

ItalyPaleAle:
This should fix the AKS issues: https://github.com/dapr/dapr/pull/6776

JoshVanL:
@ItalyPaleAle I think they became broken after https://github.com/dapr/dapr/pull/6765.

ItalyPaleAle:
> @ItalyPaleAle I think they became broken after #6765.

I doubt it. The bot isn’t even publishing a message so the failure is earlier.

by the way our bot doesn’t support multiple commands per message. Must be one message per command

JoshVanL:
Version skew daprd-sidecar-master e2e tests are failing because of a change in an error string from [this PR](https://github.com/dapr/dapr/pull/6703/files#diff-0c175408a25c2ea4a2568014b413734c677c9ab72972a6227e9b75c8b1d782e2), but otherwise this PR is passing.

charts/dapr/charts/dapr_operator/templates/dapr_operator_deployment.yaml
artursouza: We need to document which app-ids are reserved. I would even claim that "dapr-" prefix is reserved for internal services.

JoshVanL:
@artursouza Done.

pkg/operator/operator.go
```diff
@@ -105,16 +104,49 @@ func NewOperator(ctx context.Context, opts Options) (Operator, error) {
...
+
+	secProvider, err := security.New(ctx, security.Options{
+		SentryAddress:           config.SentryAddress,
+		ControlPlaneTrustDomain: config.ControlPlaneTrustDomain,
+		ControlPlaneNamespace:   security.CurrentNamespace(),
+		TrustAnchorsFile:        opts.TrustAnchorsFile,
+		AppID:                   "dapr-operator",
+		// mTLS is always enabled for the operator.
+		MTLSEnabled:    true,
+		WriteSVIDToDir: ptr.Of(filepath.Join(os.TempDir(), "/var/run/secrets/dapr.io/webhook/tls")),
```
ItalyPaleAle, line 121: Should this use a process-specific tempdir? https://pkg.go.dev/os#MkdirTemp

pkg/operator/config.go
```diff
@@ -2,37 +2,41 @@ package operator
...
+	ControlPlaneTrustDomain string
+	SentryAddress           string
 }
 
 // LoadConfiguration loads the Kubernetes configuration and returns an Operator Config.
-func LoadConfiguration(name string, client client.Client) (*Config, error) {
+func LoadConfiguration(ctx context.Context, name string, restConfig *rest.Config) (*Config, error) {
+	client, err := client.New(restConfig, client.Options{Scheme: scheme})
+	if err != nil {
+		return nil, fmt.Errorf("could not get Kubernetes API client: %v", err)
```
ItalyPaleAle, line 26:
```suggestion
		return nil, fmt.Errorf("could not get Kubernetes API client: %w", err)
```

pkg/operator/operator.go
```diff
@@ -187,216 +206,157 @@ func (o *operator) syncHTTPEndpoint(ctx context.Context) func(obj interface{}) {
...
 	healthzServer := health.NewServer(log)
 
-	err := o.mgr.Add(nonLeaderRunnable{func(ctx context.Context) error {
-		// start healthz server
-		if rErr := healthzServer.Run(ctx, healthzPort); rErr != nil {
-			return fmt.Errorf("failed to start healthz server: %w", rErr)
+	/*
+		Make sure to set `ENABLE_WEBHOOKS=false` when we run locally.
+	*/
+	if !strings.EqualFold(os.Getenv("ENABLE_WEBHOOKS"), "false") {
```
ItalyPaleAle, line 216: The current behavior is anti-pattern. An unset variable should be interpreted as false (so as ENABLE_WEBHOOK=false), while in this case it's interpreted as true.

This change impacts local development only so it's non-breaking.

```suggestion
	// Make sure to set `DISABLE_WEBHOOKS=true` when we run locally.
	if !utils.IsTruthy(os.Getenv("DISABLE_WEBHOOKS")) {
```

pkg/operator/operator.go
```diff
@@ -187,216 +206,157 @@ func (o *operator) syncHTTPEndpoint(ctx context.Context) func(obj interface{}) {
...
-		if rErr != nil {
-			return fmt.Errorf("unable to add components informer event handler: %w", rErr)
-		}
-		<-ctx.Done()
-		return nil
-	}})
-	if err != nil {
-		return err
-	}
+			log.Info("Starting api server")
```
ItalyPaleAle, line 297: Nit :)

"api" = "bees" in italian

```suggestion
			log.Info("Starting API server")
```

pkg/operator/operator.go
```diff
@@ -187,216 +206,157 @@ func (o *operator) syncHTTPEndpoint(ctx context.Context) func(obj interface{}) {
...
-func (o *operator) patchCRDs(ctx context.Context, conf *rest.Config, crdNames ...string) error {
-	client, err := kubernetes.NewForConfig(conf)
-	if err != nil {
-		return fmt.Errorf("could not get Kubernetes API client: %v", err)
-	}
-
+func (o *operator) patchCRDs(ctx context.Context, caBundle []byte, conf *rest.Config, crdNames ...string) error {
 	clientSet, err := apiextensionsclient.NewForConfig(conf)
 	if err != nil {
 		return fmt.Errorf("could not get API extension client: %v", err)
```
ItalyPaleAle, line 356:
```suggestion
		return fmt.Errorf("could not get API extension client: %w", err)
```

pkg/security/x509source.go
```diff
@@ -312,6 +323,115 @@ func (x *x509source) requestFromSentry(ctx context.Context, csrDER []byte) ([]*x
...
+
+	if err := atomicWrite(x.clock, *x.writeToDiskDir, map[string][]byte{
+		"tls.crt": certs,
+		"tls.key": pk,
+		"ca.crt":  ta,
+	}); err != nil {
+		return fmt.Errorf("unable to write SVID to disk: %w", err)
+	}
+
+	log.Infof("Wrote SVID to %q", *x.writeToDiskDir)
```
ItalyPaleAle, line 351: To avoid ugly stuff on Windows

```suggestion
	log.Infof("Wrote SVID to '%s'", *x.writeToDiskDir)
```

pkg/security/x509source.go
```diff
@@ -312,6 +323,115 @@ func (x *x509source) requestFromSentry(ctx context.Context, csrDER []byte) ([]*x
...
+	return nil
+}
+
+func (x *x509source) updateTrustAnchorFromFile(ctx context.Context, filepath string) error {
+	x.lock.RLock()
+	defer x.lock.RUnlock()
+
+	rootPEMs, err := os.ReadFile(filepath)
+	if err != nil {
+		return fmt.Errorf("failed to read trust anchors file %q: %s", filepath, err)
```
ItalyPaleAle, line 362:
```suggestion
		return fmt.Errorf("failed to read trust anchors file '%s': %w", filepath, err)
```

pkg/security/x509source.go
```diff
@@ -312,6 +323,115 @@ func (x *x509source) requestFromSentry(ctx context.Context, csrDER []byte) ([]*x
...
+	}
+
+	trustAnchorCerts, err := secpem.DecodePEMCertificates(rootPEMs)
+	if err != nil {
+		return fmt.Errorf("failed to decode trust anchors: %s", err)
```
ItalyPaleAle, line 367:
```suggestion
		return fmt.Errorf("failed to decode trust anchors: %w", err)
```

pkg/security/x509source.go
```diff
@@ -312,6 +323,115 @@ func (x *x509source) requestFromSentry(ctx context.Context, csrDER []byte) ([]*x
...
+func atomicWrite(clock clock.Clock, dir string, data map[string][]byte) error {
+	dir = filepath.Clean(dir)
+	if err := os.MkdirAll(filepath.Dir(dir), 0o700); err != nil {
+		return err
+	}
+
+	// Replace ':' with '-' in the timestamp to avoid issues with Windows
+	// filesystems.
+	newDir := dir + "-" + strings.ReplaceAll(clock.Now().Format(time.RFC3339), ":", "-")
+	if err := os.MkdirAll(newDir, 0o700); err != nil {
```
ItalyPaleAle, line 402:
```suggestion
	newDir := dir + "-" + clock.Now().Format("20060102-150405-999999999")
	if err := os.MkdirAll(newDir, 0o700); err != nil {
```

ItalyPaleAle (changes requested):

pkg/operator/operator.go
JoshVanL: I _think_ we do actually want to have a deterministic path for the SVID as this seems more consistent with other var run secrets. IIRC, the reason we are using temp here is that the operator process doesn't have permissions in the container to create a dir under `/var/run`.

pkg/security/x509source.go
JoshVanL: Thanks!

JoshVanL: The nanosecond exponent of this format wasn't working for me, so have left it at millisecond precision. Think this is enough anyway :slightly_smiling_face:

ItalyPaleAle: Oh yes for sure

pkg/operator/operator.go
artursouza: I don't find any reference to ENABLE_WEBHOOKS in dapr/dapr or dapr/docs. My recommendation is not to bundle unrelated changes to the same PR. It allows easier troubleshooting if it adds issues. I am OK keeping the non-ideal handling of this value and have it as a separate PR.

cmd/operator/options/options.go
```diff
@@ -77,9 +72,12 @@ func New() *Options {
 	flag.BoolVar(&opts.EnableArgoRolloutServiceReconciler, "enable-argo-rollout-service-reconciler", false, "Enable the service reconciler for Dapr-enabled Argo Rollouts")
 	flag.BoolVar(&opts.WatchdogCanPatchPodLabels, "watchdog-can-patch-pod-labels", false, "Allow watchdog to patch pod labels to set pods with sidecar present")
 
-	flag.StringVar(&credentials.RootCertFilename, "issuer-ca-filename", credentials.RootCertFilename, "Certificate Authority certificate filename")
-	flag.StringVar(&credentials.IssuerCertFilename, "issuer-certificate-filename", credentials.IssuerCertFilename, "Issuer certificate filename")
-	flag.StringVar(&credentials.IssuerKeyFilename, "issuer-key-filename", credentials.IssuerKeyFilename, "Issuer private key filename")
+	flag.StringVar(&opts.TrustAnchorsFile, "trust-anchors-file", "/var/run/secrets/dapr.io/tls/ca.crt", "Path to trust anchors file")
+
+	depCCP := flag.String("certchain", "", "DEPRECATED")
```
artursouza, line 77: Thanks for not making this a breaking change - at least in regard of removing the cmd args.

pkg/operator/operator.go
```diff
@@ -424,7 +389,7 @@ func (o *operator) patchCRDs(ctx context.Context, conf *rest.Config, crdNames ..
 		payload := []patchValue{{
 			Op:    "replace",
 			Path:  "/spec/conversion/webhook/clientConfig/service/namespace",
-			Value: namespace,
+			Value: security.CurrentNamespace(),
```
artursouza, line 392: Thanks for keeping this namespace-aware.

artursouza (approved):

PR #6196, opened 2023-04-05
Changed stream sequence counter from uint32 to uint64 + Enabled E2E tests with ServiceInvocationStreaming

ItalyPaleAle:
In operations that send data using chuks over a gRPC stream (e.g. stream-based service invocation, and the high-level crypto I'm working on), each chunk contains a counter that needs to be increasing on each invocation.

That counter was defined as uint32, which meant that, with the default chunk size of 2KB, the max message that could be sent was 8TB. Changing to uint64 increases the max message size to much more than we could ever need.

The good news is that this is not a breaking change because protobuf encodes both uint32 and uint64 in a variable-length format so they're compatible on the wire. This change is thus forwards- and backwards-compatible (although messages larger than 8TB will fail when sent to a Dapr 1.10 sidecar).

Additionally, this PR enables E2E tests with the `ServiceInvocationStreaming` feature on, so tests for service invocation run with both streaming and without.


tests/e2e/service_invocation/service_invocation_test.go
```diff
@@ -319,113 +321,120 @@ var grpcProxyTests = []struct {
 }
 
 func TestServiceInvocation(t *testing.T) {
```
ItalyPaleAle, line 323: Reviewers: the reason why this PR appears to be so large is because these code blocks were indented more :) However not much has changed inside them, besides replacing `"serviceinvocation-caller"` with `targetApp`

mukundansundar:
> In operations that send data using chuks over a gRPC stream (e.g. stream-based service invocation, and the high-level crypto I'm working on), each chunk contains a counter that needs to be increasing on each invocation.

> That counter was defined as uint32, which meant that, with the default chunk size of 2KB, the max message that could be sent was 8TB. Changing to uint64 increases the max message size to much more than we could ever need.

> The good news is that this is not a breaking change because protobuf encodes both uint32 and uint64 in a variable-length format so they're compatible on the wire. This change is thus forwards- and backwards-compatible (although messages larger than 8TB will fail when sent to a Dapr 1.10 sidecar).

https://protobuf.dev/programming-guides/proto3/#updating

mukundansundar (approved):

Issue #6147, opened 2023-03-28
[Tracking issue] Crypto(graphy) building block


ItalyPaleAle:
This issue tracks the work needed to release the crypto(graphy) building block.

- kit:
  - [X] dapr/kit#38
  - [X] dapr/kit#40
  - [ ] Implementation of `dapr.io/enc/v1` scheme for high-level APIs
- Components-contrib:
  - [X] dapr/components-contrib#2690
  - [X] dapr/components-contrib#2691
  - [X] dapr/components-contrib#2692
  - [ ] dapr/components-contrib#2693
- Runtime:
  - [ ] #6135
  - [ ] #6136
  - [ ] #6137 (likely pushed to 1.12)
  - [ ] High-level APIs

PR #6197, opened 2023-04-05
[Crypto] Implement high-level crypto APIs for gRPC and HTTP

ItalyPaleAle:
This PR implements the high-level crypto APIs for gRPC and HTTP:

- EncryptAlpha1
- DecryptAlpha1

These APIs process data as a stream and encrypt/decrypt messages using the dapr.io/enc/v1 scheme implemented in dapr/kit#42

These APIs cannot be implemented as "universal" because the gRPC ones handle data as a stream.

Includes unit tests for Subtle APIs too, even though they are not executed because the build flag is missing.

Next steps:

- E2E tests will be added in a separate PR, once the Go SDK adds the required methods: dapr/go-sdk#387

## Related issues

Tracking issue: #6147 
Proposal: dapr/proposals#3


ItalyPaleAle:
PR updated with unit tests for both high-level and subtle. Should provide sufficient unit test coverage; given how this works with streams, E2E tests (which also include the Go SDK) will be key in this case (coming on a follow-up PR)

pkg/grpc/api_crypto.go
```diff
@@ -0,0 +1,353 @@
...
+			if errors.Is(readErr, io.EOF) {
+				// Receiving an io.EOF signifies that the client has stopped sending data over the pipe, so we can stop reading
+				inWriter.Close()
+				break
+			} else if readErr != nil {
+				inWriter.CloseWithError(fmt.Errorf("error receiving message: %w", readErr))
+				return
+			}
+
+			if reqProto.HasOptions() {
```
mukundansundar, line 185: add comment that this is executed only after the second chunk of data is read.

pkg/grpc/api_crypto_test.go
```diff
@@ -0,0 +1,565 @@
...
+					},
+					Payload: &commonv1pb.StreamPayload{
+						Seq:  0,
+						Data: enc[0:50],
+					},
+				},
+				&runtimev1pb.DecryptAlpha1Request{
+					Payload: &commonv1pb.StreamPayload{
+						Seq:  1,
+						Data: enc[50:180],
```
mukundansundar, line 222: probably define header size and then set the magic numbers as vars or constants :)

pkg/grpc/universalapi/api_secrets_test.go
```diff
@@ -24,60 +24,14 @@ import (
...
 	"github.com/dapr/dapr/pkg/runtime/compstore"
 	daprt "github.com/dapr/dapr/pkg/testing"
-	"github.com/dapr/kit/logger"
-	"github.com/dapr/kit/ptr"
 )
 
-var testLogger = logger.NewLogger("testlogger")
-
-var testResiliency = &v1alpha1.Resiliency{
-	Spec: v1alpha1.ResiliencySpec{
```
mukundansundar, line 41: why is this removed?

mukundansundar (changes requested):
why are there changes to the subtle crypto files in this PR? 
My understanding is that the high level APIs do not use the subtle crypto ones at all but rather the crypto lib in kit. 
For adding UTs and changes to subtle crypto API files please create a separate PR.

mukundansundar:
Are there plans to add E2E for this API?

dapr/proto/runtime/v1/dapr.proto
```diff
@@ -876,6 +882,66 @@ message SubtleVerifyAlpha1Response {
...
+}
+
+// DecryptAlpha1RequestOptions contains options for the first message in the DecryptAlpha1 request.
+message DecryptAlpha1RequestOptions {
+  // Name of the component
+  string component_name = 1 [json_name="componentName"];
+  // Name (or name/version) of the key to decrypt the message.
+  // Overrides any key reference included in the message if present.
+  // This is required if the message doesn't include a key reference (i.e. was created with omit_decryption_key_name set to true).
+  string key_name = 12 [json_name="keyName"];
```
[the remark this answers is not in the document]
ItalyPaleAle: Same reason as here: https://github.com/dapr/dapr/pull/6197#discussion_r1166951315
Consistency in the props

pkg/proto/runtime/v1/dapr_additional.go
```diff
@@ -62,3 +67,87 @@ func (x *SubtleVerifyAlpha1Request) SetComponentName(name string) {
...
+	GetPayload() *commonv1pb.StreamPayload
+	// Reset the object.
+	Reset()
+	// SetOptions sets the Options property.
+	SetOptions(opts proto.Message)
+	// HasOptions returns true if the Options property is not empty.
+	HasOptions() bool
+}
+
+func (x *EncryptAlpha1Request) SetPayload(payload *commonv1pb.StreamPayload) {
```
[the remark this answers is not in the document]
ItalyPaleAle: Would cause compile-time issues if fields are renamed. However renaming a field in protobuf is a breaking change so that should not happen. Adding new fields is ok.

pkg/grpc/universalapi/api_secrets_test.go
ItalyPaleAle: Was moved to `universalapi_test.go` since it's not only used by secrets: https://github.com/dapr/dapr/pull/6197/files/82d581e4ca6b02857799b4eba8d7f5b60e926aca#diff-f3396dd779d4e21a044dcf5b996f144be2c7911131da8405c5678522356b9070R22

ItalyPaleAle:
@mukundansundar 

Thanks for the review, I'll make the requested changes shortly

> why are there changes to the subtle crypto files in this PR? My understanding is that the high level APIs do not use the subtle crypto ones at all but rather the crypto lib in kit.

They need this method from the subtlecrypto file: https://github.com/dapr/dapr/pull/6197/files#diff-c28fbdd7d5274b9739fa23e35394cfbfaa69ebb24217b218610746e4b8584bd2R63 Aside from that, you're correct that they don't use the subtle gRPC APIs.

> For adding UTs and changes to subtle crypto API files please create a separate PR.

Ok. The UTs were included here because they both depended on the same mock: https://github.com/dapr/dapr/pull/6197/files#diff-d07a8238af477b1286bb1a5fce65623a4337c1fefebbcabd23c8e9fed5f8d1eeR1
I will move them to a separate PR

> Are there plans to add E2E tests for this API in this milestone?

Yes. They'll be in a separate PR but they're blocked by dapr/go-sdk#387

dapr/proto/runtime/v1/dapr.proto
```diff
@@ -876,6 +882,66 @@ message SubtleVerifyAlpha1Response {
   bool valid = 1;
 }
 
+// EncryptAlpha1Request is the request for EncryptAlpha1.
+message EncryptAlpha1Request {
```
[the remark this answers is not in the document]
ItalyPaleAle: Done

dapr/proto/runtime/v1/dapr.proto
```diff
@@ -876,6 +882,66 @@ message SubtleVerifyAlpha1Response {
...
+// EncryptAlpha1RequestOptions contains options for the first message in the EncryptAlpha1 request.
+message EncryptAlpha1RequestOptions {
+  // Name of the component. Required.
+  string component_name = 1 [json_name="componentName"];
+  // Name (or name/version) of the key. Required.
+  string key_name = 2 [json_name="keyName"];
+  // Key wrapping algorithm to use. Required.
+  // Supported options include: A256KW (alias: AES), A128CBC, A192CBC, A256CBC, RSA-OAEP-256 (alias: RSA).
+  string algorithm = 3;
+  // Force cipher to use to encrypt data: "aes-gcm" or "chacha20-poly1305" (optional)
```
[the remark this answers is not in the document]
ItalyPaleAle: Done

dapr/proto/runtime/v1/dapr.proto
```diff
@@ -876,6 +882,66 @@ message SubtleVerifyAlpha1Response {
...
+  string cipher = 10;
+  // If true, the encrypted document does not contain a key reference.
+  // In that case, calls to the Decrypt method must provide a key reference (name or name/version).
+  // Defaults to false.
+  bool omit_decryption_key_name = 11 [json_name="omitDecryptionKeyName"]; 
+  // Key reference to embed in the encrypted document (name or name/version).
+  // This is helpful if the reference of the key used to decrypt the document is different from the one used to encrypt it.
+  // If unset, uses the reference of the key used to encrypt the document (this is the default behavior).
+  // This option is ignored if omit_decryption_key_name is true.
+  string decryption_key = 12 [json_name="decryptionKey"];
```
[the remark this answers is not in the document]
ItalyPaleAle: Done (we use "name" everywhere else)

pkg/grpc/api_crypto.go
```diff
@@ -0,0 +1,353 @@
...
+const cryptoFirstChunkTimeout = 5 * time.Second
+
+// EncryptAlpha1 encrypts a message using the Dapr encryption scheme and a key stored in the vault.
+func (a *api) EncryptAlpha1(stream runtimev1pb.Dapr_EncryptAlpha1Server) (err error) { //nolint:nosnakecase
+	// Get the first message from the caller containing the options
+	reqProto := &runtimev1pb.EncryptAlpha1Request{}
+	err = cryptoGetFirstChunk(stream, reqProto)
+	if err != nil {
+		a.Logger.Debug(err)
+		return err
```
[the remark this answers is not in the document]
ItalyPaleAle: `cryptoGetFirstChunk` returns an APIError that is ready to be sent to users and should not be wrapped. I'm adding a comment explaining this.

pkg/grpc/api_crypto.go
```diff
@@ -0,0 +1,353 @@
...
+}
+
+// DecryptAlpha1 decrypts a message using the Dapr encryption scheme and a key stored in the vault.
+func (a *api) DecryptAlpha1(stream runtimev1pb.Dapr_DecryptAlpha1Server) (err error) { //nolint:nosnakecase
+	// Get the first message from the caller containing the options
+	reqProto := &runtimev1pb.DecryptAlpha1Request{}
+	err = cryptoGetFirstChunk(stream, reqProto)
+	if err != nil {
+		a.Logger.Debug(err)
+		return err
```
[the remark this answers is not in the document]
ItalyPaleAle: Same answer

pkg/grpc/api_crypto.go
```diff
@@ -0,0 +1,353 @@
...
+	}
+
+	// Validate required options
+	if reqProto.Options == nil {
+		err = messages.ErrBadRequest.WithFormat("first message does not contain the required options")
+		a.Logger.Debug(err)
+		return err
+	}
+	if reqProto.Options.KeyName == "" {
+		err = messages.ErrBadRequest.WithFormat("missing property 'key' in the options message")
```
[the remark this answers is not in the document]
ItalyPaleAle: Changed (however not putting `key_name` too because that's the gRPC name and in the APIs, it depends on the programming language used).

pkg/grpc/api_crypto.go
ItalyPaleAle: The comment above says "Read next chunk"

pkg/grpc/api_crypto.go
```diff
@@ -0,0 +1,353 @@
...
+		}
+
+		// Stop with the last chunk
+		// This will make the method return and close the stream
+		if done {
+			break
+		}
+
+		// Reset the object so we can re-use it
+		// Use `resProto.Reset()` if more properties are added
```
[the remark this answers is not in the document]
ItalyPaleAle: Expanded the comment to explain. It essentially means that `resProto.SetPayload(nil)` is fine only if the only property in `resProto` is `payload`.

pkg/grpc/api_crypto.go
```diff
@@ -0,0 +1,353 @@
...
+	select {
+	case <-firstChunkCtx.Done():
+		return messages.ErrBadRequest.WithFormat(fmt.Errorf("error waiting for first message: %w", firstChunkCtx.Err()))
+	case err = <-firstMsgCh:
+		// Nop
+	}
+
+	if err != nil {
+		return messages.ErrCryptoOperation.WithFormat(fmt.Errorf("error receiving the first message: %w", err))
+	}
```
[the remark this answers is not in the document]
ItalyPaleAle: Done

pkg/grpc/api_crypto_test.go
```diff
@@ -0,0 +1,565 @@
...
+			send := []runtimev1pb.CryptoRequests{
+				&runtimev1pb.DecryptAlpha1Request{
+					Options: &runtimev1pb.DecryptAlpha1RequestOptions{
+						ComponentName: "myvault",
+					},
+				},
+				&runtimev1pb.DecryptAlpha1Request{
+					Payload: &commonv1pb.StreamPayload{
+						Seq:  0,
+						Data: enc[0:180],
```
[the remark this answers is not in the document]
ItalyPaleAle: It's not really a constant. The header is variable in size, but it's around 150-160 bytes. I put 180 because that should be right in the middle of the first chunk. I'll add a comment explaining.

pkg/grpc/api_crypto_test.go
ItalyPaleAle: See above. Also note header size isn't constant.

pkg/grpc/api_crypto_test.go
```diff
@@ -0,0 +1,565 @@
...
+	res := &bytes.Buffer{}
+	for !done {
+		recv.Reset()
+		err = stream.RecvMsg(recv)
+		if errors.Is(err, io.EOF) {
+			err = nil
+			done = true
+		}
+		if err != nil {
+			return nil, fmt.Errorf("failed to receive message: %w", err)
```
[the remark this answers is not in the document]
ItalyPaleAle: Added `defer stream.CloseSend()` where the stream is created. Returning from this method will cause the defer'd call to be invoked. (Also note this only impacts tests)

pkg/http/api_crypto.go
```diff
@@ -0,0 +1,194 @@
...
+
+	// Get the required properties from the query string
+	keyName := string(reqCtx.QueryArgs().Peek("keyName"))
+	if keyName == "" {
+		err = messages.ErrBadRequest.WithFormat("missing query string parameter 'keyName'")
+		log.Debug(err)
+		universalFastHTTPErrorResponder(reqCtx, err)
+		return
+	}
+	algorithm := string(reqCtx.QueryArgs().Peek("algorithm"))
```
mukundansundar, line 73: are query Args the best place to have these values? 
Should it not be part of the POST request?

pkg/http/api_crypto.go
```diff
@@ -0,0 +1,194 @@
...
+	}
+
+	// Ensure we have the required query string args
+	encOpts := encv1.EncryptOptions{
+		KeyName:   keyName,
+		Algorithm: encv1.KeyAlgorithm(strings.ToUpper(algorithm)),
+		WrapKeyFn: a.universal.CryptoGetWrapKeyFn(reqCtx, componentName, component),
+
+		// The next values are optional and could be empty
+		OmitKeyName:       utils.IsTruthy(string(reqCtx.QueryArgs().Peek("omitKeyName"))),
```
mukundansundar, line 88: think this should be `omitDecryptionKeyName`

dapr/proto/runtime/v1/dapr.proto
```diff
@@ -870,12 +876,72 @@ message SubtleVerifyAlpha1Request {
...
+message EncryptRequestOptions {
+  // Name of the component. Required.
+  string component_name = 1 [json_name="componentName"];
+  // Name (or name/version) of the key. Required.
+  string key_name = 2 [json_name="keyName"];
+  // Key wrapping algorithm to use. Required.
+  // Supported options include: A256KW (alias: AES), A128CBC, A192CBC, A256CBC, RSA-OAEP-256 (alias: RSA).
+  string algorithm = 3;
+  // Cipher used to encrypt data (optional): "aes-gcm" (default) or "chacha20-poly1305"
+  string cipher = 10;
```
mukundansundar, line 903: @ItalyPaleAle I think having these two might confuse the user. 
Should we rename the fields to make it clear? 
`keyWrappingAlgorithm` and `dataEncryptionCipher`?

pkg/http/api_crypto.go
ItalyPaleAle: This API doesn't have a POST body :( Because it's (eventually, when #6248 is merged) supposed to work with streams, it treats the body as the message to encrypt, entirely. The body is not encoded as JSON (it's can be a large file and forcing the use of JSON would require a lot of memory and additional processing time, and would make it impossible to work with the data as a stream), nor it uses multipart/form-data (because if the other fields, besides the file, are placed after the file, then we need to buffer the entire file in memory before we can start processing it).

dapr/proto/runtime/v1/dapr.proto
ItalyPaleAle: Yeah I can see that. I'll change this.

pkg/http/api_crypto.go
mukundansundar: My concern is that query args are directly available to be cached and used. Can we have the key names and algorithms in some other place?

ItalyPaleAle: Done

pkg/grpc/api_crypto.go
```diff
@@ -0,0 +1,292 @@
...
+const cryptoFirstChunkTimeout = 5 * time.Second
+
+// EncryptAlpha1 encrypts a message using the Dapr encryption scheme and a key stored in the vault.
+func (a *api) EncryptAlpha1(stream runtimev1pb.Dapr_EncryptAlpha1Server) (err error) { //nolint:nosnakecase
+	// Get the first message from the caller containing the options
+	reqProto := &runtimev1pb.EncryptRequest{}
+	err = cryptoGetFirstChunk(stream, reqProto)
+	if err != nil {
+		// This is already an APIError object.
+		a.Logger.Debug(err)
```
daixiang0, line 44: Can users see the returned error? If so, no need to print debug error.

pkg/grpc/api_crypto.go
```diff
@@ -0,0 +1,292 @@
...
+	if reqProto.Options.KeyWrapAlgorithm == "" {
+		err = messages.ErrBadRequest.WithFormat("missing property 'keyWrapAlgorithm' in the options message")
+		a.Logger.Debug(err)
+		return err
+	}
+
+	// Validate the request and get the component
+	component, err := a.CryptoValidateRequest(reqProto.Options.ComponentName)
+	if err != nil {
+		// Error is already logged
```
daixiang0, line 68: ?

pkg/grpc/api_crypto.go
```diff
@@ -0,0 +1,292 @@
...
+	if reqProto.Options == nil {
+		err = messages.ErrBadRequest.WithFormat("first message does not contain the required options")
+		a.Logger.Debug(err)
+		return err
+	}
+
+	// Validate the request and get the component
+	component, err := a.CryptoValidateRequest(reqProto.Options.ComponentName)
+	if err != nil {
+		// Error is already logged
```
daixiang0, line 113: ?

go.mod
```diff
@@ -425,3 +425,6 @@ replace (
 //
 // Then, run `make modtidy` in this repository.
 // This ensures that go.mod and go.sum are up-to-date.
+
+// TEMP UNTIL https://github.com/dapr/kit/pull/42 IS MERGED
+replace github.com/dapr/kit => github.com/italypaleale/dapr-kit v0.0.2-0.20230412203039-0210051428b8
```
mukundansundar, line 430: should be removed.

mukundansundar (commented):
small nits

pkg/grpc/api_crypto.go
```diff
@@ -0,0 +1,292 @@
+/*
+Copyright 2021 The Dapr Authors
```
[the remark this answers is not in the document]
ItalyPaleAle: Done

Taction:
LGTM, please resolve conflicts.

artursouza (dismissed):

PR #6198, opened 2023-04-05
[release-1.10] Backport changes to test infra

ItalyPaleAle:
The pipelines that run tests, especially perf ones, have changed in master, so AKS tests in the release-1.10 branch are failing.

Example: https://github.com/dapr/dapr/actions/runs/4621492216

This contains cherry-picked commits that should restore our ability to run perf tests in the release-1.10 branch. Does not include any change that would alter the code of Dapr when built.

> Since this contains cherry-picked commits, DCO will fail

### DIFF
--- a/.github/scripts/dapr_tests_summary.js
+++ b/.github/scripts/dapr_tests_summary.js
@@ -1,0 +1,12 @@
+const fs = require('fs')
+
+module.exports = async ({ glob, core }) => {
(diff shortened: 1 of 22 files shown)


ItalyPaleAle:
Looks like backporting these fixes is harder than I thought due to other changes such as switching Windows to LTSC 2022.

I think I will close this PR. We will be able to run tests on the release branch by manually triggering the workflow instead from the Actions page.

This PR can be reopened if needed

CC: @mukundansundar

PR #6199, opened 2023-04-06
Fix actor_state test

ItalyPaleAle:
Includes the Bicep fix from #6172 and some additional fixes


mukundansundar:
the same error seems to be happening.

ItalyPaleAle:
Still not working :( Will debug more tomorrow

ItalyPaleAle:
We're almost there. Looks like one more issue is that the Cosmos DB collection is still deployed without support for TTLs. Here's what it looks like in the Portal:
image

ItalyPaleAle:
Looks like everything is working now!! 🙌

In the first 2 tests above, I updated the CosmosDB used by Linux but not Windows. In the 3rd test Windows has been updated too.

mukundansundar (approved):

PR #6201, opened 2023-04-06
Perf tests: allow running specific tests with workflow_dispatch

ItalyPaleAle:
Changes our perf test pipelines to add an (optional) input to the "workflow_dispatch" trigger, which allows specifying a subset of tests to run when triggering the workflow manually.

> This PR is opened from a branch in dapr/dapr and not a fork or there's no way to test this


ItalyPaleAle:
Confirmed it works: 

https://github.com/dapr/dapr/actions/runs/4630805196

The test failed because I typed the wrong test name. Which is good, it means that the pipeline is working as intended :)

yaron2 (approved):

PR #6206, opened 2023-04-07
Case-insensitive parsing of metadata properties `actorStateStore` and `keyPrefix`

ItalyPaleAle:
In components-contrib, we've been making some efforts to make all component metadata parsing case-insensitive (thanks especially to @berndverst ) and use `utils.IsTruthy` to parse booleans.

The runtime itself parses the component metadata to extract certain fields, including: `actorStateStore` (state stores) and `keyPrefix` (state stores and locks).

This PR changes matching for `actorStateStore` and `keyPrefix` to be case-insensitive. Additionally, it uses `utils.IsTruthy` to convert `actorStateStore` to a boolean (so it's case-insensitive too and values like "1" map to true).

> Also includes some fixes to `pkg/components/lock/lock_config.go` that were applied in `pkg/components/state/state_config.go` (whose code looks almost identical and perhaps should be merged into a single file)


pkg/components/lock/lock_config.go
```diff
@@ -17,24 +18,32 @@ const (
...
-	strategy := strings.ToLower(metadata[strategyKey])
-	if strategy == "" {
-		strategy = strategyDefault
-	} else {
-		err := checkKeyIllegal(metadata[strategyKey])
-		if err != nil {
-			return err
+	strategy := strategyDefault
+	for k, v := range metadata {
+		if strings.ToLower(k) == strategyKey { //nolint:gocritic
```
mukundansundar, line 33: any reason not to use `EqualFold`?

ItalyPaleAle: They're both guaranteed to be "simple" ASCII strings, so `EqualFold` is not needed (because it considers Unicode runes, it's more complex and slow, and unnecessary here; see docs: https://pkg.go.dev/strings#EqualFold )

shubham1172 (dismissed):

pkg/runtime/runtime.go
```diff
@@ -1796,14 +1796,20 @@ func (a *DaprRuntime) initState(s componentsV1alpha1.Component) error {
...
+				if strings.ToLower(k) == actorStateStore { //nolint:gocritic
+					actorStoreSpecified = utils.IsTruthy(v)
+				}
+			}
+
+			if actorStoreSpecified {
 				a.actorStateStoreLock.Lock()
 				if a.actorStateStoreName == "" {
-					log.Infof("detected actor state store: %s", s.ObjectMeta.Name)
+					log.Info("Using actor state store: " + s.ObjectMeta.Name)
```
mukundansundar, line 1809: nitpick:
```suggestion
					log.Infof("Using %s as actor state store", s.ObjectMeta.Name)
```

pkg/runtime/runtime_test.go
```diff
@@ -5651,7 +5651,7 @@ func TestGracefulShutdownActors(t *testing.T) {
 			Version: "v1",
 			Metadata: []componentsV1alpha1.MetadataItem{
 				{
-					Name: "actorStateStore",
+					Name: actorStateStore,
```
mukundansundar, line 5654: have this as string.ToUpper? just to test the case insensitivity?

mukundansundar (dismissed):
small nitpick comments.

berndverst (dismissed):

pkg/runtime/runtime.go
ItalyPaleAle: Changed thanks

pkg/runtime/runtime_test.go
ItalyPaleAle: Done

shubham1172 (approved):

mukundansundar:
@ItalyPaleAle docs need to be added for this.

PR #6209, opened 2023-04-07
Actor reminder perf test: reduce wait time

ItalyPaleAle:
The current actor_reminder perf test has a `time.Sleep(10 * time.Minute)` which is meant to let reminders be executed and capture errors.

That is a very long wait time that makes the tests take a lot of time, but has no impact on perf. In fact, perf is measured only while reminders are being created or updated before that.

Its purpose seems only to make sure that reminders execute without error, and that's something that should be done in an E2E test and not perf.

So, this PR reduces the wait time from 10m to 90s, to make the tests quicker to run.

### DIFF
--- a/tests/perf/actor_reminder/actor_reminder_test.go
+++ b/tests/perf/actor_reminder/actor_reminder_test.go
@@ -29,6 +29,7 @@ import (
 	kube "github.com/dapr/dapr/tests/platforms/kubernetes"
 	"github.com/dapr/dapr/tests/runner"
 	"github.com/dapr/dapr/tests/runner/summary"
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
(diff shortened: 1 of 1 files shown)


mukundansundar:
The perf test seems to be failing @ItalyPaleAle

ItalyPaleAle:
@mukundansundar yes but it’s unrelated. You can see if fails on master too.

mukundansundar:
Currently the actor reminder registration perf is continuously failing. Once that is fixed, this PR can be reassessed and then merged.

ItalyPaleAle:
Uhm something's off. Now the test is expecting a QPS of 500. 🤔

mukundansundar:
@ItalyPaleAle both seem to be failing in line 197 which is 
> assert.Equal(t, 0, daprResult.RetCodes.Num400)

ItalyPaleAle:
@mukundansundar looks like GHA was just "drunk". Because the test said it failed with condition not true, but that line did not have an "assert.True".

Anyways, tests are passing now

mukundansundar (approved):
the perf tests succeeds even with reduced wait time. LGTM

berndverst (approved):

PR #6210, opened 2023-04-07
Use a lock around all operations to store reminders

### LINKED ISSUE #6018: Create/delete reminder throws "possible etag mismatch" after enabling resiliency

ItalyPaleAle:
> Rebased into master

Should fix the remaining race conditions causing operations to fail (even if without data loss). This will help at least when the calls are made to the same Dapr runtime; if calls are hitting different runtimes, the data races will still cause write operations to fail (without data loss).

Fixes #6018 (whatever is left)

### Perf impact

As for impact on perf, it doesn't look like there's any significant impact on the actors_reminder perf test. Although the lock does limit one operation at a time, before we were still using etags to _essentially_ limit to one operation at a time too.

These are the QPS:

```text
BEFORE:
39.43, 39.36, 39.37, 39.51

AFTER:
39.36, 39.40, 39.37, 38.98
```

Note that these tests were run on a different Kubernetes cluster than the one we use in our Actions, so the results are not comparable. The numbers should be looked at as relative to before and after.


mukundansundar:
Currently the main perf test that is failing is actor reminders perf test 
From [scheduled run](https://github.com/dapr/dapr/actions/runs/4715985268)
```
actor_reminder_test.go:152: target dapr app consumed 1m CPU and 4.964352Mb of Memory
actor_reminder_test.go:153: target dapr sidecar consumed 2m CPU and 31.531008Mb of Memory
actor_reminder_test.go:154: target dapr app or sidecar restarted 0 times
actor_reminder_test.go:164: 90th percentile: 274.95ms
actor_reminder_test.go:164: 99th percentile: 314.51ms
actor_reminder_test.go:166: Actual QPS: 52.43, expected QPS: 500
actor_reminder_test.go:196:
Error Trace:	/home/runner/work/dapr/dapr/tests/perf/actor_reminder/actor_reminder_test.go:196
Error:      	Should be true
Test:       	TestActorReminderRegistrationPerformance
--- FAIL: TestActorReminderRegistrationPerformance (676.64s)
```

From `/ok-to-perf` run [above](https://github.com/dapr/dapr/pull/6210#issuecomment-1510718899)
```
actor_reminder_test.go:152: target dapr app consumed 1m CPU and 4.984832Mb of Memory
actor_reminder_test.go:153: target dapr sidecar consumed 2m CPU and 28.184576Mb of Memory
actor_reminder_test.go:154: target dapr app or sidecar restarted 0 times
actor_reminder_test.go:164: 90th percentile: 273.69ms
actor_reminder_test.go:164: 99th percentile: 299.90ms
actor_reminder_test.go:166: Actual QPS: 51.59, expected QPS: 500
actor_reminder_test.go:196:
Error Trace:	/home/runner/work/dapr/dapr/tests/perf/actor_reminder/actor_reminder_test.go:196
Error:      	Should be true
Test:       	TestActorReminderRegistrationPerformance
--- FAIL: TestActorReminderRegistrationPerformance (702.88s)
```

QPS seems to have been reduced by 1 point potentially, not much of a change ideally.

artursouza (changes requested):
We need an E2E test that fails without this change but passes afterwards.

ItalyPaleAle:
@artursouza I implemented the tests for race conditions as unit tests. This is done by changing the `FakeStateStore` to have an optional `NoLock` mode, so it doesn't use locks internally. This means that concurrent access by multiple goroutines will cause a race condition. When locks are used in the actor methods, race conditions should not happen.

You can now validate that race conditions do not happen by running the tests with `go test -race`. Right now the unit tests pass successfully. If you remove one of the locks, they fail with a race condition.

This should allow preventing (not just detecting) race conditions when requests are made to the same sidecar. With etags, we can then detect race conditions across multiple sidecars.

pkg/actors/actors.go
```diff
@@ -1202,8 +1201,12 @@ func (a *actorsRuntime) CreateReminder(ctx context.Context, req *CreateReminderR
 		return err
 	}
 
-	a.activeRemindersLock.Lock()
-	defer a.activeRemindersLock.Unlock()
+	if !a.waitForEvaluationChan() {
```
ItalyPaleAle, line 1204: This was moved to before the lock because internally it uses a lock too and could otherwise cause a deadlock. Same in a couple more spots.

ItalyPaleAle:
Closing this as it was included in #6325

PR #6213, opened 2023-04-09
Add `https` and `h2c` to `app-protocol` for better HTTP/2 support, and deprecate `app-ssl`

### LINKED ISSUE #6158: Rename `app-ssl` flag to a more appropriate name
## In what area(s)?

/area runtime

## Describe the proposal

SSL has been deprecated since 2015. The name of the CLI flag should be changed to better reflect what it does, which is making the HTTP app channel use TLS/HTTPS (does not impact gRPC). Examples include:

- `app-channel-https`
- `app-https`
- `app-channel-tls`
- `app-tls`

Similarly, we should rename the annotation `dapr.io/app-ssl`.

The old flag/annotation can be kept for backwards-compatibility, but deprecated.

### LINKED ISSUE #6205: Add `https` and `h2c` to `app-protocol` for better HTTP/2 support, and deprecate `app-ssl`
## In what area(s)?

/area runtime

## Describe the proposal

Currently, Dapr can be configured to use 2 values for `app-protcol`:

- `grpc`
- `http`
  - Uses HTTPS if `app-ssl` is true

The `app-ssl` flag, which is probably misnamed (#6158), only applies to the HTTP app channel, and is ignored if the protocol is gRPC.

We can simplify this and instead add `https` as a protocol.

Why:

- This will make it easier for people to discover that we support HTTPS, which is something we should encourage since it enables using HTTP/2 (which can lead to much better perf)
- It's clearer since gRPC isn't impacted by `app-ssl`
- I have recently talked to a user who explained that now that we support HTTP/2, they were able to use HTTP for their app channel and at the same time use gRPC proxying to target the same app. This would be possible by setting `app-channel` to `https`. Fixes #4291

We can also support `h2c` for the app channel, which is HTTP/2 Cleartext. This needs to be explicitly set for the protocol negotiation to work correctly.

> We will maintain the `app-ssl` flag for backwards compatibility, but deprecated. When `app-channel` is set to `http` and `app-ssl` is enabled, we will switch the protocol to `https`

ItalyPaleAle:
# Description

Fixes #6205 and #6158 by:

- Adding support for H2C (HTTP/2 Cleartext - i.e. without TLS) as app-protocol
- Adding 3 new values to the `app-protocol` CLI flag
   - `grpcs` is gRPC + TLS (without validating TLS certificates)
   - `https` is HTTP + TLS (without validating TLS certificates)
   - `h2c` is HTTP/2 Cleartext
- Deprecates the `app-ssl` CLI option, which however continues to work for backwards compatibility. Setting `app-ssl` to true transforms `app-protocol`: `http -> https` and `grpc -> grpcs` (and shows a warning)

Also includes E2E tests for the new `app-protocol` feature by adding a test with `h2c` to the existing healthcheck test.

## Issue reference

Fixes #6205
Fixes #6158

## TODO

- [X] Update docs: dapr/docs#3328
- [X] Update Dapr CLI: dapr/cli#1267


pkg/runtime/config.go
```diff
@@ -52,6 +59,16 @@ const (
 	DefaultAppHealthCheckPath = "/health"
 )
 
+// IsHTTP returns true if the protocol is HTTP.
+func (p Protocol) IsHTTP() bool {
```
shubham1172, line 63: nit:
The function sounds like it's checking for protocol == 'HTTP' only.

```suggestion
func (p Protocol) IsGRPCBased() bool {
```

```suggestion
func (p Protocol) IsOverGRPC() bool {
```
Since gRPC also works over HTTP, to avoid confusions maybe change to above and invert the if condition everywhere?

pkg/messaging/grpc_proxy.go
```diff
@@ -112,7 +112,7 @@ func (p *proxy) intercept(ctx context.Context, fullName string) (context.Context
 	if isLocal {
 		// proxy locally to the app
 		if p.acl != nil {
-			ok, authError := acl.ApplyAccessControlPolicies(ctx, fullName, common.HTTPExtension_NONE, config.GRPCProtocol, p.acl) //nolint:nosnakecase
+			ok, authError := acl.ApplyAccessControlPolicies(ctx, fullName, common.HTTPExtension_NONE, false, p.acl) //nolint:nosnakecase
```
shubham1172, line 115: Always false?

pkg/runtime/cli.go
```diff
@@ -235,9 +235,26 @@ func FromFlags() (*DaprRuntime, error) {
...
+			// TODO: Remove in a feature Dapr version
+			if *appSSL {
+				log.Warn("The 'app-ssl' flag is deprecated; use 'app-protocol=https' instead")
+				appProtocol = string(HTTPSProtocol)
+			} else {
+				appProtocol = string(HTTPProtocol)
+			}
+		case string(GRPCProtocol):
+			// For backwards compatibility, when protocol is GRPC and --app-ssl is set, use "grpcs"
+			// TODO: Remove in a feature Dapr version
```
shubham1172, line 251:
```suggestion
			// TODO: Remove in a future Dapr version
```

pkg/runtime/runtime.go
```diff
@@ -2986,42 +2981,61 @@ func (a *DaprRuntime) createAppChannel() (err error) {
...
+	var transport nethttp.RoundTripper
+	if a.runtimeConfig.ApplicationProtocol == H2CProtocol {
+		// Enable HTTP/2 Cleartext transport
+		transport = &http2.Transport{
+			AllowHTTP: true, // To enable using "http" as protocol
+			DialTLS: func(network, addr string, _ *tls.Config) (net.Conn, error) {
+				// Return the TCP socket without TLS
+				return net.Dial(network, addr)
+			},
+			// TODO: This may not be exactly the same as "MaxResponseHeaderBytes" so check before enabling this
```
shubham1172, line 3007: Does this include restrictions for both response and request? https://datatracker.ietf.org/doc/html/rfc9113#section-10.5.1

pkg/runtime/config.go
ItalyPaleAle: I will expand the comment to clarify that it includes HTTP variants. I am not worried about the fact that gRPC is based on HTTP as in this context it should not be cause of confusion.

pkg/messaging/grpc_proxy.go
ItalyPaleAle: The argument has changed. Before it was the protocol name, either "http" or "grpc", and we were passing "grpc". Now the argument is "isHTTP" so in this case it's always false.

pkg/runtime/runtime.go
```diff
@@ -2986,42 +2981,61 @@ func (a *DaprRuntime) createAppChannel() (err error) {
...
+			// MaxHeaderListSize: uint32(a.runtimeConfig.ReadBufferSize << 10),
 		}
-		// TODO: Remove when the feature is finalized
-		if a.globalConfig.IsFeatureEnabled(config.AppChannelAllowInsecureTLS) {
-			tlsConfig.MinVersion = 0
+	} else {
```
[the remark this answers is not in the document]
ItalyPaleAle: This is the `initAppHTTPClient` method, so in here the protocol is either `http`, `https`, or `h2c`.

pkg/runtime/runtime.go
ItalyPaleAle: I am actually not sure at all

pkg/runtime/cli.go
```diff
@@ -235,9 +235,26 @@ func FromFlags() (*DaprRuntime, error) {
 	{
 		p := strings.ToLower(*appProtocolPtr)
 		switch p {
-		case string(HTTPProtocol),
-			string(GRPCProtocol):
+		case string(GRPCSProtocol), string(HTTPSProtocol), string(H2CProtocol):
 			appProtocol = p
+		case string(HTTPProtocol):
+			// For backwards compatibility, when protocol is HTTP and --app-ssl is set, use "https"
+			// TODO: Remove in a feature Dapr version
```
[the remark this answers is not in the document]
ItalyPaleAle: Thanks

shubham1172 (approved):

ItalyPaleAle:
Docs issue is here: https://github.com/dapr/docs/issues/3328

CLI issue: https://github.com/dapr/cli/issues/1267

mukundansundar:
@ItalyPaleAle the related issues are part of 1.11 milestone so PR need not be part of the milestone.

daixiang0 (approved):
LGTM

PR #6218, opened 2023-04-10
Workflow management API breaking changes for v1.11

### LINKED ISSUE #5911: [Workflow] Improvements to existing workflow management APIs
## In what area(s)?

> /area runtime

## Describe the feature

The initial release of the Workflow management API in v1.10 included APIs for starting workflows, querying their status, and terminating workflows. However, there are some problems with the implementation that can make them a bit hard to use.

### Start Workflow

The URL template is as follows: `POST workflows/{workflowComponent}/{workflowType}/{instanceId}/start`

#### Proposed improvements
* **Make the instance ID optional**  
Today it's required, but we shouldn't force users to come up with instance IDs. If one isn't provided, we should let the component generate a random ID. This will likely require a new HTTP route.

* **Return `Location` and `Retry-After` response headers from the HTTP handler**  
Today we return a HTTP 202 response but without a `Location` header. We should add this header to the response and have it point to the GetWorkflowStatus API. The goal is to natively support the [asynchronous request-reply pattern](https://learn.microsoft.com/azure/architecture/patterns/async-request-reply) out of the box when invoking workflows. The `Retry-After` header can be mainly to ensure that consumers of this pattern do not try to poll too frequently.

* **Return more information in the response**  
The current response payload is very sparse: `{"instance_id":"..."}` and doesn't follow our existing JSON field casing conventions. We should enrich this with more information and use consistent casing for JSON fields. Here's an example:

```json
{
  "workflow": {
    "instanceID": "{instanceID}"
  },
  "httpManagement": {
    "status": {"method":"GET","url":"http://localhost:3500/v1.0-alpha1/workflows/{componentName}/{instanceID}"},
    "terminate": {"method":"POST","url":"http://localhost:3500/v1.0-alpha1/workflows/{componentName}/{instanceID}"},
    "suspend": {"method":"POST","url":"http://localhost:3500/v1.0-alpha1/workflows/{componentName}/{instanceID}/suspend"},
    "resume": {"method":"POST","url":"http://localhost:3500/v1.0-alpha1/workflows/{componentName}/{instanceID}/resume"}
  }
}
```  

  Note that `{componentName}` and `{instanceID}` would be filled in with the relevant values. Also, the HTTP scheme and hostname will match the schema and hostname of the incoming request (with overrides allowed by using `X-Forwarded-*` request headers). Users (or code) should be able to simply use the URLs as-is to manage the created workflow instance.

* **Don't parse the input in the implementation**
Today we parse the input in the API handler, assuming that the payload is JSON. This API should actually just pass-through the payload to the component without trying to interpret it. Otherwise, we're adding unnecessary performance overhead and restricting all implementations to only support JSON.

### Get Workflow

The URL template is as follows: `GET workflows/{workflowComponent}/{workflowType}/{instanceId}`

#### Proposed improvements

* **Remove the `{workflowType]` segment from the URL**  
This segment is not needed. The only thing that's needed is {workflowComponent} and {instanceID}.

* **Rename "WFInfo" to "workflow"**  
This is to 1) make the payload similar to the proposed "Start Workflow" response payload and 2) follow are more conventional model for JSON payloads.

* **Change timestamp** fields in protobuf contract from string/int to `google.protobuf.Timestamp`**
This is a more proper way to represent timestamp information and will help avoid compatibility issues between SDKs and components.

* **Surface more metadata to the top-level `"workflow"` JSON object**  
Today the response payload returns an instance ID, a created timestamp, and a bag of metadata values. Many of these metadata values should be promoted to be 1st-class fields in the `workflow` object. This will make the JSON easier to parse and deserialize into rich types by the SDKs.  Here's an example:

```json
{
  "workflow": {
    "instanceID": "order123",
    "createdAt": "2023-02-02T07:00:00Z",
    "lastUpdatedAt": "2023-02-02T12:00:00Z",
    "runtimeStatus": "RUNNING"
  }
}
```

As an aside, the internal implementation should remove the `WorkflowReference` struct and just use `instanceID` (string) to identify workflow instances.

## Release Note

RELEASE NOTE: **UPDATE** workflow management APIs

/cc @nyemade-uversky

### LINKED ISSUE #6156: Workflows: actor ID should not use the `#` character (Workflows not working with Cosmos DB)
## In what area(s)?

/area runtime

## What version of Dapr?

> 1.10.x
> master

## Description

This issue was originally reported by @mPyKen here:

When Workflows creates the actor ID, it includes a `#` in the name: https://github.com/dapr/components-contrib/issues/2071#issuecomment-1480914079

https://github.com/dapr/dapr/blob/beef33d04f9528e312fe5d71ab6318a9c0728830/pkg/runtime/wfengine/workflow.go#L466-L470

This makes Workflows not usable with certain state stores such as Cosmos DB.

After discussing with @cgillum offline, we think the best approach may be to change how actor IDs are named by workflows, using a different character. This will be a breaking change (on an alpha building block), but because the `#` is not commonly used, we are unsure what other issues may appear.

cgillum:
# Description

This PR contains a set of breaking changes to improve the usability of the workflow management APIs. These changes are expected to go out with Dapr v1.11.

**Corresponding components-contrib PR: https://github.com/dapr/components-contrib/pull/2751**

## Changes

There are several changes in this PR, many of them breaking.

### Start Workflow API
* **[Breaking from v1.10]** Removed `{instanceID}` from the path and made it an optional query string parameter.
* **[Breaking from v1.10]** Changed response type and format
* **[Breaking from v1.10]** Instance IDs must now only have letters, numbers, dashes, and underscores and be < 64 characters
* **[Breaking from v1.10]** Workflow input now comes from raw HTTP request payload - no parsing
* If instance ID is not specified, a random GUID is chosen

### Get Workflow API
* **[Breaking from v1.10]** Changed format of JSON response
* **[Breaking from v1.10]** Moved several "metadata" fields to be first-class in the JSON response to make it more usable

### Raise Event API
* Event payload now comes from raw HTTP request payload - no parsing

### Terminate / Pause / Resume Workflow APIs (new in v1.11)
* Changed gRPC contracts to be distinct (previously they all shared the same contract)

### Test updates
* Minor refactoring / simplification
* Added validation of HTTP response payloads
* Workflow API tests now have 99% code coverage

### Misc. updates
* Added additional clarity in API error messages

## Issue reference

Resolves #5911
Resolves #6156

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Unit tests passing
* [x] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#3352
* [x] Specification has been updated / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#3352
* [ ] ~Provided sample for the feature / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_~


dapr/proto/runtime/v1/dapr.proto
```diff
@@ -876,46 +877,50 @@ message SubtleVerifyAlpha1Response {
...
 }
 
-message WorkflowActivityResponse {
-  // Empty
+message PauseWorkflowRequest {
+  string instance_id = 1 [json_name = "instanceID"];
+  string workflow_component = 2 [json_name = "workflowComponent"];
+}
+
+message ResumeWorkflowRequest {
```
DeepanshuA, line 916: As the attributes for Terminate, Pause and Resume Request seems to be same, should we not keep a common request?

cgillum: No, I want to better future proof these APIs by giving each their own request message. That way we can evolve them independently over time.

cgillum:
NOTE: this PR is currently blocked waiting for https://github.com/dapr/dapr/pull/6186 to be merged.

cgillum:
This PR is unblocked now and should be good to go. :)  Would appreciate if an approver/maintainer could approve the PR validation workflows to run, and provide any PR feedback.

dapr/proto/runtime/v1/dapr.proto
```diff
@@ -876,46 +877,50 @@ message SubtleVerifyAlpha1Response {
...
-  string workflow_name = 3 [json_name = "workflowName"];
 }
 
 message GetWorkflowResponse {
-  string instance_id = 1;
-  int64 start_time = 2;
-  map<string, string> metadata = 3;
+  string instance_id = 1 [json_name = "instanceID"];
+  string workflow_name = 2 [json_name = "workflowName"];
+  google.protobuf.Timestamp start_time = 3 [json_name = "createdAt"];
```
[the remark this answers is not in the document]
mukundansundar: Does this specifically have to be timestamp? 
cc @dapr/maintainers-dapr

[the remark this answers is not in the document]
cgillum: Timestamp is the natural choice. Choosing `string` creates a maintainability problem since you rely on both sides of every timestamp field to agree on a particular format (e.g. RFC3339). It's also more efficient over the wire. I recommend the Dapr project use timestamps instead of strings in new protobuf contracts moving forward.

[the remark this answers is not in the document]
ItalyPaleAle: I second @cgillum. If not `google.protobuf.Timestamp`, then perhaps a uint64 with a UNIX timestamp as another option.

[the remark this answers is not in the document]
cgillum: I went ahead and updated the protobuf field name to be `created_at` to match the JSON name. I also renamed  `last_updated_time` to `last_updated_at` to match the corresponding JSON field (apologies that this change got included in my commit to merge latest from master).

pkg/grpc/universalapi/api_workflow.go
```diff
@@ -50,175 +47,196 @@ func (a *UniversalAPI) GetWorkflowAlpha1(ctx context.Context, in *runtimev1pb.Ge
...
+		// Limit the length of the instance ID to avoid potential conflicts with state stores that have restrictive key limits.
+		maxInstanceIdLength := 64
+		if len(instanceID) > maxInstanceIdLength {
+			return messages.ErrInstanceIDTooLong.WithFormat(maxInstanceIdLength)
+		}
+
+		// Check to see if the instance ID contains invalid characters. Valid characters are letters, digits, dashes, and underscores.
+		// See https://github.com/dapr/dapr/issues/6156 for more context on why we check this.
+		for _, c := range instanceID {
+			if !unicode.IsLetter(c) && c != '_' && c != '-' && !unicode.IsDigit(c) {
```
DeepanshuA, line 222: Do we want to convert letters to all upper/lower - so that instance ids are not just differentiated in some cases on the basis of case sensitivity?

go.mod
```diff
@@ -425,3 +425,6 @@ replace (
 //
 // Then, run `make modtidy` in this repository.
 // This ensures that go.mod and go.sum are up-to-date.
+
+// DO NOT MERGE THIS CHANGE
+replace github.com/dapr/components-contrib => github.com/cgillum/dapr-components-contrib v0.0.0-20230410173729-f4047d1e15e3
```
DeepanshuA, line 430: I know we have a code comment there, but just commenting here explicitly, so that we remember to update it before merging this PR.

pkg/grpc/universalapi/api_workflow.go
cgillum: I like the idea of having workflow IDs be case-insensitive but I'm not sure if I'd want to enable that through a transformation in the API layer. I would prefer something like that be done at the engine layer - probably not part of this PR.

ItalyPaleAle, line 222: Does this need to support non-ASCII characters? I am not sure if that's a risk. The current code iterates through runes, and methods like `unicode.IsLetter` match the entire "letter" plane in the Unicode specs and not just ASCII letters.

```go
	var chr byte
// Must not use `for .. range` because that iterates through runes (i.e. UTF-8 characters), while a regular for loop iterates through bytes
	for i := 0; i < len(instanceID); i++ {
		chr = instanceID[i]
		if (chr < 'A' || chr > 'Z') && (chr < 'a' || chr > 'z') && (chr < '0' || chr > '9') && chr != '_' && chr != '-' {
			return messages.ErrInvalidInstanceID.WithFormat(instanceID)
		}
	}
```

pkg/http/api.go
```diff
@@ -692,24 +694,41 @@ func (a *api) getStateStoreWithRequestValidation(reqCtx *fasthttp.RequestCtx) (s
 	return state, storeName, nil
 }
 
-// Route:   "workflows/{workflowComponent}/{workflowName}/{instanceId}",
+// Route:   "workflows/{workflowComponent}/{workflowName}/start?instanceID={instanceID}",
```
ItalyPaleAle, line 697: Should this be passed in the body?

pkg/grpc/universalapi/api_workflow.go
```diff
@@ -50,175 +47,196 @@ func (a *UniversalAPI) GetWorkflowAlpha1(ctx context.Context, in *runtimev1pb.Ge
...
 
 // TerminateWorkflowAlpha1 is the API handler for terminating a workflow
-func (a *UniversalAPI) TerminateWorkflowAlpha1(ctx context.Context, in *runtimev1pb.WorkflowActivityRequest) (*runtimev1pb.WorkflowActivityResponse, error) {
-	var method func(context.Context, *workflows.WorkflowReference) error
-	if in.WorkflowComponent != "" {
-		if wf, ok := a.CompStore.GetWorkflow(in.WorkflowComponent); ok {
-			method = wf.Terminate
-		}
+func (a *UniversalAPI) TerminateWorkflowAlpha1(ctx context.Context, in *runtimev1pb.TerminateWorkflowRequest) (*emptypb.Empty, error) {
+	emptyResponse := &emptypb.Empty{}
```
ItalyPaleAle, line 100: Nit: this can cause an unnecessary heap allocation (and so more work for the GC) when it's not needed. I'd recommend:

```
return &emptypb.Empty{}, err
```

pkg/grpc/universalapi/api_workflow.go
```diff
@@ -50,175 +47,196 @@ func (a *UniversalAPI) GetWorkflowAlpha1(ctx context.Context, in *runtimev1pb.Ge
...
-	if err != nil {
-		err = methodErr.WithFormat(in.InstanceId)
+func (a *UniversalAPI) validateInstanceId(instanceID string, isCreate bool) error {
+	if instanceID == "" {
+		return messages.ErrMissingOrEmptyInstance
+	}
+
+	if isCreate {
+		// Limit the length of the instance ID to avoid potential conflicts with state stores that have restrictive key limits.
+		maxInstanceIdLength := 64
```
ItalyPaleAle, line 214:
```suggestion
		const maxInstanceIdLength = 64
```

pkg/http/api.go
```diff
@@ -722,14 +741,12 @@ func (a *api) onGetWorkflowHandler() fasthttp.RequestHandler {
 		})
 }
 
-type workflowActivityHandler = func(ctx context.Context, in *runtimev1pb.WorkflowActivityRequest) (*runtimev1pb.WorkflowActivityResponse, error)
-
-// This is a common handler for TerminateWorkflow, PauseWorkflow and ResumeWorkflow
-func (a *api) onWorkflowActivityHandler(handler workflowActivityHandler) fasthttp.RequestHandler {
```
ItalyPaleAle, line 728: Why was this undone?

pkg/runtime/wfengine/component.go
```diff
@@ -156,32 +162,32 @@ func (c *workflowEngineComponent) Get(ctx context.Context, req *workflows.Workfl
...
-				"dapr.workflow.custom_status":  metadata.SerializedCustomStatus,
-				"dapr.workflow.last_updated":   metadata.LastUpdatedAt.Format(time.RFC3339),
+			Workflow: &workflows.WorkflowState{
+				InstanceID:    req.InstanceID,
+				WorkflowName:  metadata.Name,
+				CreatedAt:     metadata.CreatedAt,
+				LastUpdatedAt: metadata.LastUpdatedAt,
+				RuntimeStatus: getStatusString(int32(metadata.RuntimeStatus)),
+				Properties: map[string]string{
+					"dapr.workflow.input":         metadata.SerializedInput,
```
ItalyPaleAle, line 172: Can these be constants?

pkg/grpc/universalapi/api_workflow.go
cgillum: There's no requirement to support non-ASCII characters. I figured we can start out as being a bit strict and then loosen up the requirements over time based on user feedback.

pkg/http/api.go
cgillum: No, I'm intentionally trying to use the request body exclusively for the raw workflow input without any Dapr interpretation.

pkg/grpc/universalapi/api_workflow.go
cgillum: There shouldn't be any unnecessary allocations because every code path returns the same `emptyResponse` value, whether there is an error or a success. Or am I misunderstanding?

pkg/http/api.go
cgillum: Yes, this was removed intentionally. I wanted to make the terminate, pause, and resume APIs more future-proof by giving them their own distinct contract and implementation, which means we don't use a common API handler for all of them.

pkg/runtime/wfengine/component.go
cgillum: They can, but is that what we want when they're only used in one place? I didn't see an advantage to using constants for these, only disadvantages.

pkg/http/api.go
ItalyPaleAle: Ok the same situation was just addressed here: https://github.com/dapr/dapr/pull/6197#discussion_r1170857312

@mukundansundar  and @artursouza recommended to use HTTP headers instead

pkg/grpc/universalapi/api_workflow.go
ItalyPaleAle: Oh I'm sorry, I missed that it was returned also in case of success.

pkg/runtime/wfengine/component.go
ItalyPaleAle: The advantage would be if they were used in unit tests too. Otherwise, ok

pkg/grpc/universalapi/api_workflow.go
ItalyPaleAle: I agree. Can you change to use the loop as in my example above then? The current implementation does support unicode broadly

pkg/http/api.go
cgillum: That's interesting. I'd love to get more context on this statement, which I don't understand:
> My concern is that query args are directly available to be cached and used.

Happy to add support for headers. Is supporting both an option? I do like the convenience of query string parameters.

pkg/runtime/wfengine/component.go
cgillum: My counter to that is duplicating these values in unit tests helps prevent accidental breaking changes (a unit test is a bit more representative of customer code this way). :)

pkg/grpc/universalapi/api_workflow.go
cgillum: Oh, I missed that detail. That's interesting because I actually started by borrowing logic from [token.IsIdentifier](https://github.com/golang/go/blob/598cf5e6ac1deb87f4931af195dad66f2b3f1672/src/go/token/token.go#L335) for this code. I'm not an expert on this topic by any means, but any idea why the go maintainers were okay with using `range` on a string instead of treating it as an array like you're suggesting?

pkg/http/api.go
DeepanshuA: @mukundansundar @artursouza What is your viewpoint here?

go.mod
cgillum: This has been removed. We're now getting the latest from `dapr/components-contrib`.

cgillum:
@ItalyPaleAle @artursouza @mukundansundar there are a couple threads still unresolved, but this PR is otherwise ready to go.

Unfortunately, it looks like GitHub is having a partial outage, which is causing some CI failures. Independent of that, I'd love to get a sign-off. The corresponding components-contrib PR has been merged: https://github.com/dapr/components-contrib/pull/2751.

pkg/grpc/universalapi/api_workflow.go
cgillum: Resolving for now as I don't think this should be a blocker. I'd love to hear your opinion on this when you get back from vacation, though!

cgillum:
The two failing tests in the most recent run appear to be flakey unit tests related to actors. They aren't related in any way to this PR and can likely be scheduled for retry by a maintainer. This PR is otherwise good to be merged, pending any blocking PR comments.

pkg/grpc/universalapi/api_workflow.go
```diff
@@ -50,175 +47,196 @@ func (a *UniversalAPI) GetWorkflowAlpha1(ctx context.Context, in *runtimev1pb.Ge
...
+	if err != nil {
+		a.Logger.Debug(err)
+		return emptyResponse, err
 	}
 
 	req := workflows.RaiseEventRequest{
 		InstanceID: in.InstanceId,
 		EventName:  in.EventName,
-		Input:      in.Input,
+		EventData:  in.Input,
```
mukundansundar, line 145: should this be changed to `EventData` in the incoming request as well?

pkg/grpc/universalapi/api_workflow.go
```diff
@@ -50,175 +47,196 @@ func (a *UniversalAPI) GetWorkflowAlpha1(ctx context.Context, in *runtimev1pb.Ge
...
 	}
-	return a.workflowActivity(ctx, in, method, messages.ErrResumeWorkflow)
-}
 
-// workflowActivity is a helper function to handle workflow requests for pause, resume, and terminate
-func (a *UniversalAPI) workflowActivity(ctx context.Context, in *runtimev1pb.WorkflowActivityRequest, method func(context.Context, *workflows.WorkflowReference) error, methodErr messages.APIError) (*runtimev1pb.WorkflowActivityResponse, error) {
-	if in.InstanceId == "" {
-		err := messages.ErrMissingOrEmptyInstance
+	req := &workflows.PauseRequest{
+		InstanceID: in.InstanceId,
```
mukundansundar, line 172: would prefer consistency in casing.

pkg/grpc/universalapi/api_workflow.go
```diff
@@ -50,175 +47,196 @@ func (a *UniversalAPI) GetWorkflowAlpha1(ctx context.Context, in *runtimev1pb.Ge
...
+		for _, c := range instanceID {
+			if !unicode.IsLetter(c) && c != '_' && c != '-' && !unicode.IsDigit(c) {
+				return messages.ErrInvalidInstanceID.WithFormat(instanceID)
+			}
+		}
+	}
+	return nil
+}
+
+func (a *UniversalAPI) getComponent(componentName string) (workflows.Workflow, error) {
```
mukundansundar, line 230:
```suggestion
func (a *UniversalAPI) getWorkflowComponent(componentName string) (workflows.Workflow, error) {
```
There can be other methods with similar names later on for other components.

pkg/grpc/universalapi/api_workflow.go
```diff
@@ -50,175 +47,196 @@ func (a *UniversalAPI) GetWorkflowAlpha1(ctx context.Context, in *runtimev1pb.Ge
...
-	if err != nil {
-		err = methodErr.WithFormat(in.InstanceId)
+func (a *UniversalAPI) validateInstanceID(instanceID string, isCreate bool) error {
+	if instanceID == "" {
+		return messages.ErrMissingOrEmptyInstance
+	}
+
+	if isCreate {
+		// Limit the length of the instance ID to avoid potential conflicts with state stores that have restrictive key limits.
+		const maxInstanceIDLength = 64
```
mukundansundar, line 214: Will the instance Id be prefixed with `dapr`? So that it is known that it is dapr specific.

pkg/http/api.go
mukundansundar: In this case I agree that queryArgs will be better, as this is specific for an instance and optional.

pkg/grpc/universalapi/api_workflow.go
```diff
@@ -50,175 +47,196 @@ func (a *UniversalAPI) GetWorkflowAlpha1(ctx context.Context, in *runtimev1pb.Ge
...
-		err := messages.ErrTimerParse.WithFormat(err)
-		a.Logger.Debug(err)
-		return &runtimev1pb.GetWorkflowResponse{}, err
-	}
-
 	res := &runtimev1pb.GetWorkflowResponse{
-		InstanceId: id.InstanceId,
-		StartTime:  t.Unix(),
-		Metadata:   response.Metadata,
+		InstanceId:    response.Workflow.InstanceID,
```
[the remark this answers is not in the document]
cgillum: A few things about this:

- Dapr has linting rules that flag "Id" and require us to change to "ID".
- "Id" is a .NET convention whereas all other languages use "ID".
- The `InstanceId` in this line comes from the durabletask gRPC proto definitions, which are external to Dapr and can't be changed right now.
- The `InstanceID` in the workflow struct comes from components-contrib, which has already been merged for 1.11. Changing this now would require another round of changes in components-contrib.

With this context, do you still feel strongly that we should change this now?

pkg/grpc/universalapi/api_workflow.go
cgillum: See my previous response on casing.

pkg/grpc/universalapi/api_workflow.go
cgillum: No, these IDs are provided by the user. We don't apply any transformations to them.

pkg/grpc/universalapi/api_workflow.go
cgillum: Done.

pkg/grpc/universalapi/api_workflow.go
cgillum: Good call. Done.

pkg/grpc/universalapi/api_workflow.go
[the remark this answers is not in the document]
artursouza: Everything inside Dapr should be ID. So I am fine with external things being as-is.

artursouza (approved):

Issue #5939, opened 2023-02-15
Improve granularity in resiliency metrics


alicejgibbons:
## In what area?

 /area runtime

## Describe the feature

As a Dapr user, I desire additional granularity for resiliency metrics. It can be difficult to monitor if a resiliency policy is working as expected with the metrics that are provided today. Today I can only see the following (as per [docs](https://github.com/dapr/dapr/blob/master/docs/development/dapr-metrics.md#resiliency)):
- dapr_runtime_resiliency_loaded: The number of resiliency policies loaded.
- dapr_runtime_resiliency_count: The number of times a resiliency policy has been executed.

I would like to have:
- the target the resiliency policy is being executed on, for component whether this is inbound or outbound
- the circuit breaker state: open, half-open, closed
- the number of times a *default* resiliency policy has been executed

## Release Note

RELEASE NOTE: **UPDATE** Improved granularity on resiliency metrics

filintod:
in regards to the default resiliency policy, couldn't we already filtered them by using the names or even prefixes?:
- [built-in](https://docs.dapr.io/operations/resiliency/policies/#overriding-default-retries) `DaprBuiltInServiceRetries`, `DaprBuiltInActorRetries`, `DaprBuiltInActorReminderRetries` and `DaprBuiltInInitializationRetries`
- [Global defaults](https://docs.dapr.io/operations/resiliency/policies/#setting-default-policies): `DefaultRetryPolicy`, `DefaultTimeoutPolicy` and `DefaultCircuitBreakerPolicy`

Just trying to see if we really need another label

PR #6221, opened 2023-04-11
Add granular metrics information to resiliency

filintod:
# Description

Added labels to incorporate extra information to resiliency counters:
- `flow_direction`: inbound or outbound
-  `target`: that depends on the building block can be `app_<app name>`, `actor_<actor type>` or for components `<component type>_<name>`
-  `status`: that is so far mainly used by circuit breaker type metrics with possible values `open`, `close` and `half-open`

It also adds a new metric `resiliency/activations_total` to record when the different resiliencies actually get engaged (stops after timeout, retry execution or when circuit breaker trips from close to open). It only records the first activation, as the retry mechanism uses a one time notification I added similar behavior to timeout, cb state might get activated more than once if the retry is longer than the cb timeout and it could go to half-open and then to open from close.

## Issue reference

Please reference the issue this PR will close: #5939 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Unit tests passing


filintod:
found a bug in default CB for endpoints and actors, added the fix but would like some feedback.

Also need feedback if we want to have that extra configuration to make adding metrics to default resiliency optional, or should I remove the config and always add metrics for default resiliency actions.

yaron2:
> found a bug in default CB for endpoints and actors, added the fix but would like some feedback.
> 
> Also need feedback if we want to have that extra configuration to make adding metrics to default resiliency optional, or should I remove the config and always add metrics for default resiliency actions.

cc @halspang

filintod:
another thing I noticed is that we count the number of times the resiliency wraps a function call, so even if the function works the first time (no timeout/retry/etc) we count it as using the resiliency that might lead to incorrect assumptions if we are alerting on resiliency rate. 

To be able to count at the moment of usage of the resiliency we would need to move the value increment to the resiliency runner that might require more work though, we could leave for another PR on the next release or I can try to incorporate but no assurance as the code freeze is near.

filintod:
I still have a couple questions pending to incorporate or not:
- should default resiliency have their metrics usage counted by default or optional via the config added in this PR? that I can remove and make it non optional. Making it default makes sense, and it is part of the linked issue request, but it is a new behavior.
- should I try to trigger increment of the counter at the moment of failure or keep as it is today at the moment when creating the PolicyDefinition before actually starting the resiliency runner?  If on failure, should I record each failure (mainly each retry/timeout and on CB each state change) or only the first failure.

artursouza:
> I still have a couple questions pending to incorporate or not:
> 
> * should default resiliency have their metrics usage counted by default or optional via the config added in this PR? that I can remove and make it non optional. Making it default makes sense, and it is part of the linked issue request, but it is a new behavior.
Unless you think this will generate too much data, I would make it non optional and not have another flag for users to enable.

> * should I try to trigger increment of the counter at the moment of failure or keep as it is today at the moment when creating the PolicyDefinition before actually starting the resiliency runner?  If on failure, should I record each failure (mainly each retry/timeout and on CB each state change) or only the first failure.
I think that we should track resiliency failures too and make it distinct metrics, so users can alert on those. Regarding CB, I think a metric for each change of state is good too, as I imagine users might want to alert in some conditions.

artursouza (commented):
Is there a way to have E2E tests to cover these? Meaning, we validate the metrics are present in the scenarios we expect.

artursouza (approved):

PR #6222, opened 2023-04-11
Reduce flakiness in gRPC tests by not using TCP

ItalyPaleAle:
The unit tests in pkg/grpc frequently fail due to flakiness. It seems that it's at least in part related to the fact that they create lots of TCP servers that are short-lived.

This PR changes the unit tests to use ["bufconn"](https://google.golang.org/grpc/test/bufconn) (part of the official gRPC package) to create a "virtual listener" which is entirely in-memory. This doesn't require creating TCP sockets or listeners so it's less prone to failures in unit tests.

### DIFF
--- a/pkg/grpc/api_actor_test.go
+++ b/pkg/grpc/api_actor_test.go
@@ -17,7 +17,6 @@ import (
 	"context"
 	"testing"
 
-	"github.com/phayes/freeport"
 	"github.com/stretchr/testify/assert"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
@@ -56,13 +55,12 @@ var testActorResiliency = &v1alpha1.Resiliency{
 
 func TestRegisterActorReminder(t *testing.T) {
 	t.Run("actors not initialized", func(t *testing.T) {
-		port, _ := freeport.GetFreePort()
-		server := startDaprAPIServer(port, &api{
+		server, lis := startDaprAPIServer(&api{
 			id: "fakeAPI",
 		}, "")
 		defer server.Stop()
 
-		clientConn := createTestClient(port)
+		clientConn := createTestClient(lis)
 		defer clientConn.Close()
 
 		client := runtimev1pb.NewDaprClient(clientConn)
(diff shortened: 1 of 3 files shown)


pkg/grpc/server.go
```diff
@@ -124,7 +124,9 @@ func getDefaultMaxAgeDuration() *time.Duration {
 // StartNonBlocking starts a new server in a goroutine.
 func (s *server) StartNonBlocking() error {
 	var listeners []net.Listener
-	if s.config.UnixDomainSocket != "" && s.kind == apiServer {
+	if s.config.listener != nil {
+		listeners = append(listeners, s.config.listener)
+	} else if s.config.UnixDomainSocket != "" && s.kind == apiServer {
```
mukundansundar, line 129: is there a way to implement this without introducing the if else clause here?

filintod: you can create a helper function `getListeners` and just return `listeners` slice at each if step

mukundansundar: That would move the if-else clause from here into the helper function.

filintod: in some way but without the deeper nesting
```
func getListeners() ([]net.Listener, error) {
	if s.config.listener != nil {
		return []net.Listener{s.config.listener}, nil
	if s.config.UnixDomainSocket != "" && s.kind == apiServer {
		socket := fmt.Sprintf("%s/dapr-%s-grpc.socket", s.config.UnixDomainSocket, s.config.AppID)
		l, err := net.Listen("unix", socket)
		if err != nil {
			return nil, err
		}
		s.logger.Infof("gRPC server listening on UNIX socket: %s", socket)
		return []net.Listener{l}, nil
	listeners := make([]net.Listener,0,len(s.config.APIListenAddresses))
	for _, apiListenAddress := range s.config.APIListenAddresses {
		addr := apiListenAddress + ":" + strconv.Itoa(s.config.Port)
		l, err := net.Listen("tcp", addr)
		if err != nil {
			s.logger.Debugf("Failed to listen for gRPC server on TCP address %s with error: %v", addr, err)
			continue
		} 
		s.logger.Infof("gRPC server listening on TCP address: %s", addr)
		listeners = append(listeners, l)
	}
	return listeners, nil
}
```

ItalyPaleAle: I have reverted the changes to server.go and server_test.go so:

1. Still tests the creation of the TCP sockets per @mukundansundar 's concerns
2. Does not have this `if` anymore

mukundansundar (approved):

Issue #1172, opened 2020-02-28
Dapr should need a pod restart when a component is applied


amanbha:
Currently in k8s, Dapr needs a pod restart to pick new configuration after they have been applied.
This will cause inconsistency in scenarios in which few pods of an app are restarted by platform. Restarted pods will use new config whereas pods which are not restarted will continue to use old config.

msfussell:
There is a discussion here on the overlap between dynamic updating of components/config by dapr and when the K8s operator manages the rollout of a new application. We need to get options from both sides for this proposal.

amanbha:
Adding notes from maintainers meeting:
Pros and cons of dynamic updating the components::
1. dynamic updating of components will solve a "split-brain" problem due to app scale out, failover.
2. dynamic updating of components can also lead to issues as application code might not be ready for the change.

Suggested approaches:
1. Keep the current behavior and document best practise that after applying the new component yaml, user should do upgrade of the application (or do pod restarts) to pick up updated components.
2. Do dynamic upload but give user an option to opt out as an annotation.
3. Add a dynamic upload property to each component, dapr sidecar will do "hot-reload" of the component when it receives the notification from operator only when this property is set to true.

AaronCrawfis:
## Proposal

Given the above what it we:
1. Keep current behavior as default (pods do not pickup changes to components until restarted) to address Pro/Con #1
1. Write documentation to address Approach #1
1. Add metadata to components and configuration yaml files to opt-in to hot reloading for all pods/applications to pick up this component instantly.
    - This will allow customers to rollout hot-patches or errors to applications without restarting entire application
1. (optional) Add ability for applications/pod specs to be annotated/flagged to always pick up components instantly upon change, no matter what above metadata says

This will allow 3 cases:

## Default Behavior

1. Application running locally or kubernetes
1. New configuration applied
1. Application doesn't pick up new configuration until restarted/scaled

## Apply Hot Reload to Component/Configuration

1. Application running locally or kubernetes
1. New configuration/component applied with metadata set to opt into hot reload
1. All applications pick up new configuration/component

## Apply Hot Reload to Application/Pod

1. Application running locally or kubernetes with flag or annotation set to hot reload
1. New configuration/component applied (with or without metadata for hot reload, doesn't matter)
1. Application picks up new configuration/component

We could brand this as a new feature that users opt into if they want to.

RadoslavGatev:
[...]

PR #6224, opened 2023-04-11
Adds `enable-hot-reloading` flag to daprd. Set with `dapr.io/enable-hot-reloading` annotation.

JoshVanL:
Fixes https://github.com/dapr/dapr/issues/1172

Adds CLI flag to daprd `--enable-hot-reloading`, which when enabled, will watch for component updates from the operator and restart that component with the new spec.

Can only be enabled in `Kubernetes` mode.

Flag will be set by the injector when the pod has the `dapr.io/enable-hot-reloading` annotation set to `"true"`.

If and when a component is updated  _and_ the spec has been changed, the component will be closed, and then init with the new spec. Daprd will not roll back to the old spec in the event of an error when initing the new component spec.


yaron2:
> The last time we removed hot-reloading from k8s, we also discussed to bring standalone and k8s to parity. #4432
> 
> > As discussed in #4218, the k8s components updates were buggy and should be disabled till the hot-reloading feature is completed (for both k8s and standalone).
> 
> There is an open PR today for standalone mode #5197 (which also seems unattended), do we plan on merging the two and bringing support to both k8s and standalone?

The biggest quality of life improvement for users is on K8s, so parity is nice but self hosted IMO is much lower priority than k8s.

Issue #6216, opened 2023-04-10
Actors health checks should use the app health subsystem


ItalyPaleAle:
## In what area(s)?

/area runtime

## Describe the proposal

Currently, we have 2 separate subsystems that perform health checks for the app:

- When using actors, pkg/health is used to ping the app at the /healthz endpoint to determine if it's ready to run actors, using HTTP (the same package contains code for the health check server too, in addition to the health check client used by actors)
- When app health checks are enabled, pkg/apphealth is used to probe the app and, when it's in an un-healthy state, stop delivering pubsub messages and incoming service invocation requests. This supports both HTTP and gRPC.

The two subsystems are performing very similar tasks, and it would make sense to merge them by making actors use pkg/apphealth too, which has the benefit of supporting gRPC too.

ItalyPaleAle:
🤖

PR #6225, opened 2023-04-11
Make AppHealthChecks stable and change default path to /healthz

### LINKED ISSUE #6217: Make App Health Checks stable (and change the default HTTP healthcheck path)
[App Health Checks](https://docs.dapr.io/developing-applications/building-blocks/observability/app-health/) were introduced in Dapr 1.9 and are behind a feature flag.

In Dapr 1.11 we should consider removing the feature flag to make the feature stable. Note that this won't enable them for everyone, as they still need to be explicitly enabled by the user with the `enable-app-health-check` CLI flag or annotation.

As part of the stabilization effort, I also propose we make a **breaking change** and switch the default path for HTTP healthchecks from `/health` to `/healthz`. This is the same paths that actor health checks use, so it will make it easier to converge them (see #6216)

ItalyPaleAle:
# Description

Removes the "AppHealthChecks" feature flag to make app health checks stable. The feature is still not enabled by default and requires the user to opt-in with the `enable-app-health-checks` CLI flag/annotation.

**Breaking change:** The default path for HTTP health checks is changed from `/health` to `/healthz` so it's aligned with the path used by actors health checks. See: #6216 

## Issue reference

Fixes #6217


mukundansundar:
@ItalyPaleAle what is the impact of this breaking change?

mukundansundar:
> @mukundansundar if users were not specifically passing a value for app-health-check-path, the default was /health - see: https://docs.dapr.io/developing-applications/building-blocks/observability/app-health/#configuring-app-health-checks
> 
> I would like to change the default to /healthz for the reasons explained in the linked issue, to harmonize with how actor health checks work today (with the eventual goal of merging the two)
> 
> This will impact users that were relying on the default value being /health so they may need to either change the healthcheck endpoint in their apps, or explicitly set app-health-check-path=/health

ok ... is there a change required in the Dapr SDKs?
And for the time being can we have it query both `/healthz` and if not available `/health`? With a deprecation set for `/health` endpoint?

mukundansundar:
Are there breaking changes needed in SDKs?

mukundansundar (dismissed):

yaron2:
> Are there breaking changes needed in SDKs?

Seconding this question

yaron2:
> And for the time being can we have it query both /healthz and if not available /health

That would be much preferable, even if not required

ItalyPaleAle:
> > Are there breaking changes needed in SDKs?
> 
> Seconding this question

I am not sure. I don't know which SDKs implement AppHealth, and whether they use HTTP or gRPC (not impacted), and whether they rely on the built-in health path. How would we go about finding this out?

> > And for the time being can we have it query both /healthz and if not available /health
> 
> That would be much preferable, even if not required

Is this additional complexity warranted in this case? Since we're talking about code that runs on init, it will be challenging to detect if /healthz responds with an error becuase it's the wrong endpoint, or because the app is just not yet ready.

Since what we're talking about is an alpha feature (so per contract we can make breaking changes), and users who depend on the old behavior only need to set a CLI flag.

I'm also concerned is that if we add this second option, users will begin depending on that and then deprecating this will be much harder.

mukundansundar:
cc @yaron2 @artursouza I think this can be merged .... I agree that with moving this feature out of preview, we can have a breaking change and it will be better to have consistency in health endpoints across Dapr
Wdyt?

artursouza (approved):
I am fine merging as-is but can postpone to next release if people want to discuss the default health end point further.

Issue #5631, opened 2022-12-12
github.com/gorilla has been archived


robertojrojas:
The [Gorilla Web Toolkit](https://github.com/gorilla) has been **archived**. I think it'll be a good thing to look into how this will affect Dapr.

ItalyPaleAle:
Usually when these things happen (see the JWT library for example), the community will take that up. A few forks will pop up, and one will establish itself as the new one.

It's also possible that the gorilla libraries are in a "done" state and won't really need much maintenance either.

robertojrojas:
@ItalyPaleAle good point!
On the other hand,  thinking about how quickly this project would react if in the future a major bug or security issue is found in the gorilla libraries. 
 I guess instead of creating this issue as **`bug`**, I should've created it as a **`question`** issue.

ItalyPaleAle:
So far I haven't seen a single "strong" fork for gorilla/mux.

A lot of people seem to be migrating to other routers, actually. The most popular choices (aside from those based on fasthttp like fiber, which we should steer clear of) seem to be Gin and Chi.

Gin is more powerful, but Chi is simpler and perhaps a better choice for us?

https://github.com/go-chi/chi

yaron2:
Chi would be a good choice.

kolinfluence:
@ItalyPaleAle @robertojrojas @yaron2 @davidkarlsen 
pls just use this github.com/cloudwego/hertz
or allow an option to switch default http server

shubham1172:
GWT has been revived again https://github.com/gorilla/mux/issues/715

PR #6226, opened 2023-04-12
Convert Metadata and Shutdown APIs to universal

ItalyPaleAle:
More API conversion to Universal, to help with #4979

Fixes:

- Before, extended metadata fields set using gRPC and HTTP were kept separate, so setting an extended metadata field using gRPC and reading it using HTTP, or vice-versa, was not possible
- The "httpEndpoints" property was added to the HTTP metadata APIs [only](https://github.com/dapr/dapr/commit/7a0fdf9f016b0c3a3f48447bc4169838e6026ec2#diff-237f098772824bb7f3c25e36c2565bf147e4dfedb1077bb351595b0ba4b881bdR128), missing gRPC

Also adds E2E tests for put metadata which did not exist before.

The changes are fully backwards-compatible


pkg/http/api_test.go
```diff
@@ -2249,136 +2246,101 @@ func TestV1ActorEndpoints(t *testing.T) {
...
 	}
-	expectedBodyBytes, _ := json.Marshal(expectedBody)
 
-	t.Run("Metadata - 200 OK", func(t *testing.T) {
-		apiPath := "v1.0/metadata"
-		mockActors := new(actors.MockActors)
+	fakeServer.StartServer(testAPI.constructMetadataEndpoints())
 
-		mockActors.On("GetActiveActorsCount")
+	t.Run("Set Metadata", func(t *testing.T) {
```
ItalyPaleAle, line 2324: New test, since there was none for this method before :)

pkg/grpc/api.go
```diff
@@ -139,31 +129,32 @@ func NewAPI(opts APIOpts) API {
...
-		accessControlList:          opts.AccessControlList,
-		appProtocol:                opts.AppProtocol,
-		shutdown:                   opts.Shutdown,
-		getComponentsFn:            opts.GetComponentsFn,
-		getComponentsCapabilitesFn: opts.GetComponentsCapabilitiesFn,
-		getSubscriptionsFn:         opts.GetSubscriptionsFn,
-		daprRunTimeVersion:         buildinfo.Version(),
+		directMessaging:        opts.DirectMessaging,
+		actor:                  opts.Actor,
+		id:                     opts.AppID,
```
daixiang0, line 147: still `appid` to make it easier to understand?

pkg/grpc/universalapi/universalapi.go
```diff
@@ -16,25 +16,38 @@ limitations under the License.
...
-	CryptoProviders      map[string]contribCrypto.SubtleCrypto
-	StateStores          map[string]state.Store
-	SecretStores         map[string]secretstores.SecretStore
-	SecretsConfiguration map[string]config.SecretsScope
-	LockStores           map[string]lock.Store
-	WorkflowComponents   map[string]workflows.Workflow
+	AppID                      string
+	Logger                     logger.Logger
+	Resiliency                 resiliency.Provider
+	Actors                     actors.Actors
```
daixiang0, line 39: I see sometimes we use `actor` represent `actors.Actors`, sometime `actors`, could we unify them in this time?

ItalyPaleAle: `actor` is the current name. I agree we should change to `actors`, but seems to be a bigger change than this PR right now.

pkg/grpc/api.go
ItalyPaleAle: `id` is the current name. I would rather not change it in this PR, although I agree `appID` is better.

dapr/proto/runtime/v1/dapr.proto
```diff
@@ -580,11 +580,11 @@ message RegisteredComponents {
 }
 
 message PubsubSubscription {
-  string pubsub_name = 1;
-  string topic = 2;
-  map<string,string> metadata = 3;
-  PubsubSubscriptionRules rules = 4;
-  string dead_letter_topic = 5;
+  string pubsub_name = 1 [json_name = "pubsubname"];
+  string topic = 2 [json_name = "topic"];
```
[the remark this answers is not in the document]
ItalyPaleAle: When using `protojson` to marshal protobuf objects to JSON, the default name is the name of the field. In this case, `pubsub_name`.

However our existing APIs use `pubsubname` when marshalling to JSON, so this is added for backwards-compatibiity.

I don't think we need to update all protos today. When updating APIs to universal, we may need to update them as needed to preserve backwards compatibility.

pkg/grpc/universalapi/universalapi.go
mukundansundar: please create an issue to follow this up.

pkg/grpc/api.go
ItalyPaleAle: I have made the change in this PR, at least for pkg/grpc/api

pkg/grpc/universalapi/universalapi.go
ItalyPaleAle: I have made the change in this PR, at least for pkg/grpc/api

ItalyPaleAle:
Added the E2E tests for PUT metadata. Also changed the properties with name `id` and `actor`.

pkg/http/api_metadata.go
```diff
@@ -0,0 +1,135 @@
...
+			Handler: a.onPutMetadata(),
+		},
+	}
+}
+
+func (a *api) onGetMetadata() fasthttp.RequestHandler {
+	return UniversalFastHTTPHandler(
+		a.universal.GetMetadata,
+		UniversalFastHTTPHandlerOpts[*emptypb.Empty, *runtimev1pb.GetMetadataResponse]{
+			OutModifier: func(out *runtimev1pb.GetMetadataResponse) (any, error) {
```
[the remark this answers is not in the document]
mukundansundar: I think it will be clearer to unmarshal the data into structs and then finally marshalling the struct into json. 
The part of modifying an already encoded json object is something that I think should be avoided.

Then if we are marshalling into different structs, should the metadata API be moved into universalAPI?

mukundansundar (commented):
Overall looks good, but the question is do we really need to move the metadata API into universal API? And if so can we avoid the rewriting json encoded objects instead work directly with structs?

mukundansundar:
Is there a reason to change to use Chi in tests inside this PR? Can it be a separate change?

ItalyPaleAle:
> I think it will be clearer to unmarshal the data into structs and then finally marshalling the struct into json.
The part of modifying an already encoded json object is something that I think should be avoided.

I was trying to avoid copying into other structs as it would require duplicating all data structures (and not just the current 3), and the multiple rounds of marshalling/unmarshalling would be a lot slower. I agree it's not the nicest but it should be safe as this just adds something before the last character.

However if you think readability is more important here, I can change it. I am always in favor of readability.

> Then if we are marshalling into different structs, should the metadata API be moved into universalAPI?

I think there's still value:

- The metadata now can be read and written using both gRPC and HTTP. As explained in the first post, currently if you write a metadata using gRPC and try to read it using HTTP, it's not there.
- The business logic now is now implemented just once across all APIs, so it's easier to maintain
- Will be easier to migrate to net/http

> Is there a reason to change to use Chi in tests inside this PR? Can it be a separate change?

Yes: #5631 

Since I was working on that test, I thought I may as well do that

mukundansundar:
> I was trying to avoid copying into other structs as it would require duplicating all data structures (and not just the current 3), and the multiple rounds of marshalling/unmarshalling would be a lot slower. I agree it's not the nicest but it should be safe as this just adds something before the last character.

> However if you think readability is more important here, I can change it. I am always in favor of readability.

I think yes this will help making changes to the API in the future as well. Please do make this change.

> Yes: https://github.com/dapr/dapr/issues/5631

> Since I was working on that test, I thought I may as well do that

Would prefer having it as a separate PR if possible.

Please fix the linter error also in this PR.

ItalyPaleAle:
@mukundansundar done both things (and fixed the merge conflicts)

mukundansundar (approved):

ItalyPaleAle:
Fixed the merge conflicts

yaron2:
Why is this PR needed? I understand why starting a new API as universal would make sense, but why convert existing ones that incur almost no maintenance?

ItalyPaleAle:
@yaron2 because of #4979 and what I wrote above, including the fact that right now this doesn't work as expected:

>     * The metadata now can be read and written using both gRPC and HTTP. As explained in the first post, currently if you write a metadata using gRPC and try to read it using HTTP, it's not there.
> 
>     * The business logic now is now implemented just once across all APIs, so it's easier to maintain
> 
>     * Will be easier to migrate to net/http
>

mukundansundar:
> @yaron2 because of #4979 and what I wrote above, including the fact that right now this doesn't work as expected:
> 
> > * The metadata now can be read and written using both gRPC and HTTP. As explained in the first post, currently if you write a metadata using gRPC and try to read it using HTTP, it's not there.
> > * The business logic now is now implemented just once across all APIs, so it's easier to maintain
> > * Will be easier to migrate to net/http

@yaron2 is there any reason this PR should not be merged?
Since the work is already done and E2E tests pass for this API?

mukundansundar:
> > @yaron2 because of #4979 and what I wrote above, including the fact that right now this doesn't work as expected:
> > > * The metadata now can be read and written using both gRPC and HTTP. As explained in the first post, currently if you write a metadata using gRPC and try to read it using HTTP, it's not there.
> > > * The business logic now is now implemented just once across all APIs, so it's easier to maintain
> > > * Will be easier to migrate to net/http
> 
> @yaron2 is there any reason this PR should not be merged? Since the work is already done and E2E tests pass for this API?

@yaron2 ?

Issue #6230, opened 2023-04-12
Add postgres component in e2e test


shivamkm07:
## In what area(s)?

> /area runtime

> /area operator

> /area placement

> /area docs

> /area test-and-release
## Describe the proposal

This is to add postgres component in E2E test for configuration API

shivamkm07:
ping

PR #6231, opened 2023-04-12
Adding postgres  configuration component in E2E test

shivamkm07:
# Description

This PR adds postgres configuration component to e2e test

## Issue reference

Please reference the issue this PR will close: #6230 

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing


shivamkm07:
Note that test here for postgres component fails because it requires a fix in the component included in PR https://github.com/dapr/components-contrib/pull/2740. So it will only pass after the above PR is merged and go.mod points to the updated master in contrib.

yaron2:
Why is this needed? We do not test individual components in `dapr/dapr`.

shivamkm07:
@yaron2 As per https://github.com/dapr/proposals/blob/main/guides/api-design.md#beta-to-stable, `
E2E tests must exercise at least two different implementations of the building block's API`. Currently only `redis` is being tested in E2E test. This PR is to add postgres as well to the same.

yaron2:
> @yaron2 As per https://github.com/dapr/proposals/blob/main/guides/api-design.md#beta-to-stable, ` E2E tests must exercise at least two different implementations of the building block's API`. Currently only `redis` is being tested in E2E test. This PR is to add postgres as well to the same. cc @mukundansundar

I'll work to remove this requirement, thanks for bringing it up. It makes no sense to test multiple components from the API perspective: if one component test fails, it means there's an implementation issue with that component and we have conformance/certification tests to ensure the validity from the components side.

Can you think of any practical example where testing a minimum of 2 components (out of possibly dozens) would yield real benefits from testing them in dapr/dapr e2e?

mukundansundar:
> > @yaron2 As per https://github.com/dapr/proposals/blob/main/guides/api-design.md#beta-to-stable, ` E2E tests must exercise at least two different implementations of the building block's API`. Currently only `redis` is being tested in E2E test. This PR is to add postgres as well to the same. cc @mukundansundar
> 
> I'll work to remove this requirement, thanks for bringing it up. It makes no sense to test multiple components from the API perspective: if one component test fails, it means there's an implementation issue with that component and we have conformance/certification tests to ensure the validity from the components side.
> 
> Can you think of any practical example where testing a minimum of 2 components (out of possibly dozens) would yield real benefits from testing them in dapr/dapr e2e?

@yaron2 We have had scenarios where sometimes having a second E2E component has helped. For example with respect to CosmosDB and state store building block. 

For the question of whether it would help in this particular scenario, it might.
One particular thing here is should we have only Redis for most of the building blocks or should we have other components also?

yaron2:
> > > @yaron2 As per https://github.com/dapr/proposals/blob/main/guides/api-design.md#beta-to-stable, ` E2E tests must exercise at least two different implementations of the building block's API`. Currently only `redis` is being tested in E2E test. This PR is to add postgres as well to the same. cc @mukundansundar
> > 
> > 
> > I'll work to remove this requirement, thanks for bringing it up. It makes no sense to test multiple components from the API perspective: if one component test fails, it means there's an implementation issue with that component and we have conformance/certification tests to ensure the validity from the components side.
> > Can you think of any practical example where testing a minimum of 2 components (out of possibly dozens) would yield real benefits from testing them in dapr/dapr e2e?
> 
> @yaron2 We have had scenarios where sometimes having a second E2E component has helped. For example with respect to CosmosDB and state store building block.
> 
> For the question of whether it would help in this particular scenario, it might. One particular thing here is should we have only Redis for most of the building blocks or should we have other components also?

I'm interested in knowing how this can help detect runtime API issues instead of component implementation issues. If the purpose is to help find issues in components implementations, then `dapr/dapr` is not the place as this responsibility falls directly under contrib conformance and certification tests. More so, this feels heavily non-deterministic - if you have N components but only pick 2 as this is the minimum, you have no established way of choosing "the right component".

tests/apps/configurationapp/app.go
```diff
@@ -36,13 +37,17 @@ import (
...
 const (
 	daprGRPCPort               = 50001
-	configStore                = "configstore"
 	daprConfigurationURLStable = "http://localhost:3500/v1.0/configuration/"
 	daprConfigurationURLAlpha1 = "http://localhost:3500/v1.0-alpha1/configuration/"
 	separator                  = "||"
 	redisHost                  = "dapr-redis-master.dapr-tests.svc.cluster.local:6379"
 	writeTimeout               = 5 * time.Second
 	alpha1Endpoint             = "alpha1"
+	postgresComponent          = "postgres"
```
artursouza, line 46: Please, keep the test unchanged and switch between stores in the workflow config. See how we do it for Cosmosdb in state store.

artursouza (changes requested):

go.mod
```diff
@@ -245,7 +245,7 @@ require (
 	github.com/influxdata/line-protocol v0.0.0-20210922203350-b1ad95c89adf // indirect
 	github.com/jackc/pgpassfile v1.0.0 // indirect
 	github.com/jackc/pgservicefile v0.0.0-20221227161230-091c0ba34f0a // indirect
-	github.com/jackc/pgx/v5 v5.3.1 // indirect
+	github.com/jackc/pgx/v5 v5.3.1
```
ItalyPaleAle, line 248: Make sure that all non-indirect dependencies are in the same block please

tests/apps/configurationapp/app.go
shivamkm07: Done.

shivamkm07:
@mukundansundar @artursouza ping for reviews.

go.mod
mukundansundar: @shivamkm07 small change

.github/workflows/kind-e2e.yaml
```diff
@@ -75,12 +75,15 @@ jobs:
...
           kind-image-sha: sha256:ef453bb7c79f0e3caba88d2067d4196f427794086a7d0df8df4f019d5e336b61
+          DAPR_TEST_CONFIG_STORE: "postgres"
         - k8s-version: v1.24.7
           kind-version: v0.17.0
           kind-image-sha: sha256:577c630ce8e509131eab1aea12c022190978dd2f745aac5eb1fe65c0807eb315
+          DAPR_TEST_CONFIG_STORE: "redis"
         - k8s-version: v1.25.3
           kind-version: v0.17.0
           kind-image-sha: sha256:f52781bc0d7a19fb6c405c2af83abfeb311f130707a0e219175677e366cc45d1
+          DAPR_TEST_CONFIG_STORE: "redis"
```
artursouza, line 86: nit: please, keep the same naming convention. No all caps here.

artursouza (changes requested):

go.mod
shivamkm07: Done

.github/workflows/kind-e2e.yaml
shivamkm07: Done

ItalyPaleAle (approved):

PR #6232, opened 2023-04-12
Adds `runtime/compstore` runtime component store

JoshVanL:
compstore is a new runtime package which dynamically loads and unloads runtime components. It is used as a consistent store for runtime services to ensure they all have the same view of the world.

This is a predicate for [hot reloading](https://github.com/dapr/dapr/issues/1172), whereby all runtime services will need to use the same store of components loaded, which may be removed, added and updated, multiple times during runtime.


pkg/http/api.go
```diff
@@ -76,17 +73,14 @@ type API interface {
 }
 
 type api struct {
-	universal                  *universalapi.UniversalAPI
+	*universalapi.UniversalAPI
```
ItalyPaleAle, line 76: Why this change? Not sure we should have `api` depend on this, since all methods would automatically be exported and if we have a method with the same name, but different arguments, there'd be a conflict IIRC

pkg/runtime/runtime.go
```diff
@@ -913,10 +867,13 @@ func (a *DaprRuntime) subscribeTopic(parentCtx context.Context, name string, top
 		policyRunner := resiliency.NewRunner[any](ctx, policyDef)
 		_, err = policyRunner(func(ctx context.Context) (any, error) {
 			var pErr error
-			if a.runtimeConfig.ApplicationProtocol.IsHTTP() {
+			switch a.runtimeConfig.ApplicationProtocol {
```
ItalyPaleAle, line 870: Please revert this and use `IsHTTP`, because we now have `https` and `h2c` too, and for gRPC (`!IsHTTP`) we have `grpcs`

pkg/runtime/runtime.go
```diff
@@ -1264,12 +1224,12 @@ func (a *DaprRuntime) sendBindingEventToApp(bindingName string, data []byte, met
...
 	var appResponseBody []byte
-	path := a.inputBindingRoutes[bindingName]
-	if path == "" {
+	path, ok := a.compStore.GetInputBindingRoute(bindingName)
+	if !ok || path == "" {
 		path = bindingName
 	}
 
-	if !a.runtimeConfig.ApplicationProtocol.IsHTTP() {
+	if a.runtimeConfig.ApplicationProtocol == GRPCProtocol {
```
ItalyPaleAle, line 1232: Here too, please revert this

pkg/runtime/runtime.go
```diff
@@ -1344,7 +1304,7 @@ func (a *DaprRuntime) sendBindingEventToApp(bindingName string, data []byte, met
 				}
 			}
 		}
-	} else {
+	} else if a.runtimeConfig.ApplicationProtocol == HTTPProtocol {
```
ItalyPaleAle, line 1307: Ditto

pkg/runtime/runtime.go
```diff
@@ -1589,7 +1533,7 @@ func (a *DaprRuntime) getSubscribedBindingsGRPC() ([]string, error) {
 
 func (a *DaprRuntime) isAppSubscribedToBinding(binding string) (bool, error) {
 	// if gRPC, looks for the binding in the list of bindings returned from the app
-	if !a.runtimeConfig.ApplicationProtocol.IsHTTP() {
+	if a.runtimeConfig.ApplicationProtocol == GRPCProtocol {
```
ItalyPaleAle, line 1536: Ditto

pkg/runtime/runtime.go
```diff
@@ -1602,9 +1546,9 @@ func (a *DaprRuntime) isAppSubscribedToBinding(binding string) (bool, error) {
 				return true, nil
 			}
 		}
-	} else {
+	} else if a.runtimeConfig.ApplicationProtocol == HTTPProtocol {
```
ItalyPaleAle, line 1549: Ditto

pkg/runtime/runtime.go
```diff
@@ -1794,20 +1739,14 @@ func (a *DaprRuntime) initState(s componentsV1alpha1.Component) error {
...
-			actorStoreSpecified := false
-			for k, v := range props {
-				if strings.ToLower(k) == actorStateStore { //nolint:gocritic
-					actorStoreSpecified = utils.IsTruthy(v)
-				}
-			}
-
-			if actorStoreSpecified {
+			actorStoreSpecified := props[actorStateStore]
+			if actorStoreSpecified == "true" {
```
ItalyPaleAle, line 1743: Please revert this change too, which was made on purpose to make `actorstatestore` case-insensitive and match with IsTruthy

pkg/runtime/runtime.go
```diff
@@ -1794,20 +1739,14 @@ func (a *DaprRuntime) initState(s componentsV1alpha1.Component) error {
...
 				a.actorStateStoreLock.Lock()
 				if a.actorStateStoreName == "" {
-					log.Info("Using '" + s.ObjectMeta.Name + "' as actor state store")
+					log.Infof("detected actor state store: %s", s.ObjectMeta.Name)
```
ItalyPaleAle, line 1746: Why were these logs reverted too?

pkg/runtime/runtime.go
```diff
@@ -1871,9 +1806,9 @@ func (a *DaprRuntime) getSubscriptions() ([]runtimePubsub.Subscription, error) {
 	}
 
 	// handle app subscriptions
-	if a.runtimeConfig.ApplicationProtocol.IsHTTP() {
+	if a.runtimeConfig.ApplicationProtocol == HTTPProtocol {
```
ItalyPaleAle, line 1809: Here too, and same 2 lines below

pkg/runtime/runtime.go
```diff
@@ -2962,12 +2887,13 @@ func (a *DaprRuntime) createAppChannel() (err error) {
 	}
 
 	var ch channel.AppChannel
-	if !a.runtimeConfig.ApplicationProtocol.IsHTTP() {
+	switch a.runtimeConfig.ApplicationProtocol {
```
ItalyPaleAle, line 2890: Please don't change these too, same reason

ItalyPaleAle (changes requested):
Overall LGTM but you've reverted some changes introduced in recent PRs that need to be kept please

pkg/http/api.go
JoshVanL: This was to keep it consistent with `grpc`, but happy to revert.

pkg/http/api.go
```diff
@@ -1815,7 +1793,7 @@ func (a *api) onGetActorState(reqCtx *fasthttp.RequestCtx) {
 		respond(reqCtx, withError(fasthttp.StatusInternalServerError, msg))
 		log.Debug(msg)
 	} else {
-		if resp == nil || len(resp.Data) == 0 {
+		if resp == nil || resp.Data == nil {
```
ItalyPaleAle, line 1796: Should not be reverted

```suggestion
		if resp == nil || len(resp.Data) == 0 {
```

JoshVanL:
Thanks for the review @ItalyPaleAle. Apologies for the bad reverts- was a bad rebase+copy&paste PBKAC on my end.

pkg/runtime/runtime.go
```diff
@@ -1264,8 +1213,8 @@ func (a *DaprRuntime) sendBindingEventToApp(bindingName string, data []byte, met
 	ctx, span := diag.StartInternalCallbackSpan(a.ctx, spanName, spanContext, a.globalConfig.Spec.TracingSpec)
 
 	var appResponseBody []byte
-	path := a.inputBindingRoutes[bindingName]
-	if path == "" {
+	path, ok := a.compStore.GetInputBindingRoute(bindingName)
+	if !ok || path == "" {
```
ItalyPaleAle, line 1217: Nit: if the property doesn't exist, it returns the zero value which is `""`. So checking `ok` is redundant 

```suggestion
	path := a.compStore.GetInputBindingRoute(bindingName)
	if path == "" {
```

ItalyPaleAle (commented):
Overall LGTM but you've reverted some changes introduced in recent PRs that need to be kept please

pkg/http/api.go
ItalyPaleAle: Btw this is the reason why the E2E tests are failing I think. We made that change because the tests were not working properly.

ItalyPaleAle (dismissed):

ItalyPaleAle (approved):

yaron2 (approved):

PR #6237, opened 2023-04-13
Allow disabling Placement and Actors in Helm chart

### LINKED ISSUE #6173: DAPR Placement service - ability to disable it if not in use
DAPR Placement service is one of the services in DAPR Control plane. This service is required for Actors runtime and if the Actors is not enabled in the DAPR setup then this service is not required and an ability to shutdown the service will help save system resources.

## In what area(s)?

 Placement Service

## Describe the feature
We can turn off in Helm  to not load Dashboards / injector service in control plane ; but I cannot turn off the placement service to not load it.  Placement service is needed for Actors building block and if Actor is not used, we should allow to switch off placement service too.

Add a flag in HELM like dapr_placement.enabled = true/ false

ex command: helm install dapr dapr/dapr --namespace dapr-system --set  dapr_placement.enabled = false --wait

ItalyPaleAle:
Fixes #6173

Adds option `global.actors.enabled` to Helm chart (defaults to true).

When false, the Dapr Placement service is not deployed, and the injector sets an empty `--placement-host-address` CLI flag in injected daprd containers. However, if Deployments are created with an explicit annotation `dapr.io/placement-host-address`, that still has priority.


mukundansundar:
Just a clarification, placement service if disabled won't it affect workflow APIs also?

ItalyPaleAle:
> Just a clarification, placement service if disabled won't it affect workflow APIs also?

Correct. At least for the internal workflow engine.

Taction (approved):
LGTM

mukundansundar (approved):
lgtm

ItalyPaleAle:
@mukundansundar 

> does a later requirement of placement service in the cluster, need a full re-install of dapr or can the placement stateful set be installed separately in an already existing cluster. Do we have a documentation for that? And how will that interact with the existing dapr installation?

- Because this is set as an option in the Helm chart, the simplest thing is to re-run "helm upgrade" with the changed flag.
- We use an injector and not a reconciler, so when you enable Placement if it was previously-disabled, pods will not get the connection to placement. Pods will have to be recreated. Likewise, if the placement becomes disabled, pods that had the `--placement-host` flag will attempt to connect anyways (but this is not the end of the world - it won't cause production issues (unless we have bugs like #6130  ))
- A user is always welcome to manually deploy Placement. in this case they can use the `dapr.io/placement-host-address` annotation to have sidecars connect to Placement even if it was disabled in Helm.

mukundansundar:
> Because this is set as an option in the Helm chart, the simplest thing is to re-run "helm upgrade" with the changed flag.
In this case I assume even `dapr upgrade` with required `set` flags will work?

ItalyPaleAle:
> In this case I assume even dapr upgrade with required set flags will work?

What do you mean?

mukundansundar:
> > In this case I assume even dapr upgrade with required set flags will work?
> 
> 
> 
> What do you mean?

Will the dapr upgrade command with set parameters work or should it only be helm upgrade?

ItalyPaleAle:
> Will the dapr upgrade command with set parameters work or should it only be helm upgrade?

I am not familiar with how "dapr upgrade" works :( I know this is "normal" Helm.

artursouza (approved):

pravinpushkar:
`dapr upgrade` uses helm upgrade behind the scene, so IMO it should work.
1. If the placement service was disabled and some performs `dapr upgrade` it should enable the placement service as well as the default value is true for `global.actors.enabled`
2. For disabling it while installing or on a existing cluster, we might need to provide another flag in cli which can be consumed by helm.

pravinpushkar (approved):

mukundansundar (commented):
lgtm

mukundansundar (approved):
lgtm

Issue #6176, opened 2023-04-03
Allow to set OTEL endpoint with the OTEL_EXPORTER_OTLP_ENDPOINT env var


yaron2:
Today Dapr will only initialize OTEL if an `endpointAddress` is explicitly set in the configuration CRD. The OTEL SDK can also read the `OTEL_EXPORTER_OTLP_ENDPOINT` environment variable. We should allow setting the environment variable if it's present on the `daprd` process to allow for advanced scenarios like sending traces to a node IP more easily.

PR #6242, opened 2023-04-14
Allow to set OTEL endpoint with the OTEL_EXPORTER_OTLP_ENDPOINT env var

drewby:
# Description

This allows the use of environment variables injected by the OpenTelemetry Operator or set by other central configuration. There are two advantages to using the environment variables. 

1) Users of the OpenTelemetry Operator can have configuration in one place and use same configuration for primary containers and the sidecar.
2) In some scenarios (OtelCollector in DaemonSet mode), the endpoint address is unknown and cannot be added to the Dapr config. For example, it might be the Node IP address.

This implementation focuses on updating the configuration system to apply the environment variables if Otel is not already defined in the Dapr configuration. The Dapr configuration takes precedent over the environment variables.

There are some alternative approaches:
1) I'm calling `SetTracingSpecFromEnv` from the `FromFlags` function in `cli.go`. This is so the environment variables are checked regardless of the presence of a configuration file (ie. default configuration). This could be moved to `LoadDefaultConfiguration`.
2) Another approach would be to detect the presence of the environment variable so we know to create tracer, but leave it to the Otel SDK to actually load the variables (don't pass an endpoint address to the SDK). 

## Issue reference

Closes: #6176

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests
* [X] Unit tests passing


yaron2 (approved):

NickLarsenNZ:
> In some scenarios (OtelCollector in DaemonSet mode), the endpoint address is unknown and cannot be added to the Dapr config. For example, it might be the Node IP address.

Isn't this solved by having a `Service` like any other pod (where the IP address is unknown until it is running)? The fact that a DaemonSet created it shouldn't change anything from the Service's point of view (just likw how Services can work for `Deployment` and `StatefulSet` resources.

drewby:
> > In some scenarios (OtelCollector in DaemonSet mode), the endpoint address is unknown and cannot be added to the Dapr config. For example, it might be the Node IP address.
> 
> Isn't this solved by having a `Service` like any other pod (where the IP address is unknown until it is running)? The fact that a DaemonSet created it shouldn't change anything from the Service's point of view (just like how Services can work for `Deployment` and `StatefulSet` resources.

You are correct, and the Operator even configures the Service for the DeamonSet. But the downside to using a Service is that it doesn't guarantee pods on the same node will use to the local instance of the collector, which is why some configure the endpoint using the node name or node IP.

yaron2:
@drewby now this is merged, please open a docs PR to update with this new functionality

drewby:
> @drewby now this is merged, please open a docs PR to update with this new functionality

Done

PR #6244, opened 2023-04-14
Propagate content type properties in state store requests

### LINKED ISSUE #6208: Propagate content type [metadata] to content type request property
## In what area(s)?

/area runtime

> /area operator

> /area placement

> /area docs

> /area test-and-release

## What version of Dapr?

1.10

## Expected Behavior

Historically, the Dapr APIs (e.g. for setting state store values) expect gRPC callers to pass the content type as a property in the request metadata, which is then passed on to the component as request metadata. HTTP callers are expected to pass the content type as a query parameter, which is then copied into the request metadata for passage to the component.  Later on, content type was made "first class" and added as a direct property of such component requests.  The content type metadata, if provided, should be propagated to the content type property (see https://github.com/dapr/dapr/issues/2892#issuecomment-989105035)

It happens that HTTP calls unmarshal the JSON body into a component request object, so callers *can* set the content type property (which could potentially conflict with the content type set via the query parameter). gRPC calls have no similar mechanism today.

Existing components *can* continue using the component content type property, but new components *should* prefer the content type property, if set.  

## Actual Behavior

gRPC calls (e.g. state store) do not propagate content type metadata to the content type property. Likewise, if not otherwise provided, HTTP calls also do not propagate the content type.  As a complication, the component implementation request types are more loose than the equivalent gRPC types; they allow *any* type of value to be passed to a component.  (In practice, it appears only one of types is expected, either an opaque byte array or a JSON marshal-able object. HTTP calls will always pass a the latter.  gRPC calls will, if the content type metadata indicates JSON, decode and un-marshal the value, else it will pass the value unmodified.  This behavior affords components some implicit knowledge of "opaque" vs. "JSON", even when the caller does not explicitly declare the content type; existing components use this to determine how to send data to their backing services.)

This ambiguity is an issue, in particular, for pluggable components which are adapted from the components' types to more strict gRPC types.  Pluggable components only receive a byte array so, without an explicit content type, they cannot make any assumptions about the type of data and so cannot exactly replicate the behavior of "built in" components.

## Release Note

RELEASE NOTE:

philliphoff:
# Description

- Updates the gRPC state store (bulk) set behavior to propagate any content type set in the metadata to the request content type property.
- Updates the HTTP state store (bulk) set behavior to:
   - Propagate any content type set in the metadata to the request content type property
   - Propagate any set content type property to the metadata
   - If both content type properties are set, overwrites the metadata property with the request property
- Updates the state store pluggable component adaptor to change the content type properties if it serializes a value to a JSON-formatted encoded string. This enables pluggable components the ability to differentiate between opaque arrays and implicitly JSON-based values.  (Built-in components can be passed both opaque arrays and object instances, whereas pluggable components can be passed only opaque arrays.)

## Issue reference

Please reference the issue this PR will close: #6208 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Unit tests passing


pkg/grpc/api.go
```diff
@@ -661,7 +661,9 @@ func (a *api) SaveState(ctx context.Context, in *runtimev1pb.SaveStateRequest) (
 			Metadata: s.Metadata,
 		}
 
-		if req.Metadata[contribMetadata.ContentType] == contenttype.JSONContentType {
+		contentType, hasContentType := s.Metadata[contribMetadata.ContentType]
+
+		if hasContentType && contentType == contenttype.JSONContentType {
```
ItalyPaleAle, line 666: Nit: `contentType` can't be the JSON one if there's no content type in the dictionary

```suggestion
		if contentType == contenttype.JSONContentType {
```

pkg/grpc/api_test.go
```diff
@@ -1206,6 +1206,117 @@ func TestSaveState(t *testing.T) {
 	}
 }
 
+func TestSaveStateContentType(t *testing.T) {
+	fakeStore := &daprt.MockStateStore{}
+
+	contentTypeProperty := "contentType"
+
+	var expectedContentType *string
```
ItalyPaleAle, line 1214: Please don't use variables like these that are modified in the scope outside of the function checking them. It makes the code much harder to read.

Would recommend instead attaching a `fakeStore.On()` in each test, so it's clearer what's being tested, and it'll be easier to modify this in the future.

pkg/http/api_test.go
```diff
@@ -4873,6 +4873,171 @@ func TestV1StateEndpoints(t *testing.T) {
...
+		}
+		b, _ := json.Marshal(reqs)
+
+		fakeServer.DoRequest("POST", apiPath, b, nil)
+	})
+
+	t.Run("state save with metadata content type", func(t *testing.T) {
+		contentTypeStore := &daprt.MockStateStore{}
+		
+		applicationJson := "application/json"
```
ItalyPaleAle, line 4919: This package has a constant `jsonContentTypeHeader` that can be used

philliphoff: Used that package's constant instead.

pkg/grpc/api_test.go
philliphoff: Reworked the tests to avoid that.

pkg/http/api.go
```diff
@@ -979,6 +979,23 @@ func (a *api) onPostState(reqCtx *fasthttp.RequestCtx) {
 			}
 		}
 
+		contentType := reqs[i].ContentType
+		metadataContentType, ok := reqs[i].Metadata[contribMetadata.ContentType]
+
+		hasShownWarning := false
```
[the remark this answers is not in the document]
philliphoff: I believe so; updated.

pkg/components/state/pluggable.go
```diff
@@ -27,6 +27,8 @@ import (
 	proto "github.com/dapr/dapr/pkg/proto/components/v1"
 	"github.com/dapr/kit/logger"
 
+	"golang.org/x/exp/maps"
+
```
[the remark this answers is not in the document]
philliphoff: Updated.

pkg/components/state/pluggable.go
```diff
@@ -425,13 +431,23 @@ func toSetRequest(req *state.SetRequest) (*proto.SetRequest, error) {
 		if dataBytes, err = utils.Marshal(reqValue, json.Marshal); err != nil {
 			return nil, err
 		}
+
+		// Add or overwrite metadata with content type
```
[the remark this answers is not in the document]
philliphoff: It would *probably* be ok to modify the request's metadata instance, but I was concerned about changing assumptions about its immutability that other components might have in the request pipeline.  It seemed safer to make a copy.

yaron2:
@philliphoff can you please check why e2e tests are failing? It'd be great to get this merged

philliphoff:
I've given it another go to setup/run the E2E tests myself but, again, without success.  I may need to work with someone to help me understand what I'm doing wrong.

ItalyPaleAle:
> I've given it another go to setup/run the E2E tests myself but, again, without success. I may need to work with someone to help me understand what I'm doing wrong.

Ping me on Teams later today!

olitomlinson:
@philliphoff can this be re-opened pls?

olitomlinson:
bump

olitomlinson:
@yaron2 do you see any value in keeping this PR alive?

tthophan (approved):

PR #6245, opened 2023-04-14
Add alias for Azure Service Bus Queues binding

berndverst:
# Description

Add alias for Azure Service Bus Queues binding

Now both `bindings.azure.servicesbusqueues` and `bindings.azure.servicebus.queues` can be used.


berndverst:
Apparently this was already in master

Issue #6246, opened 2023-04-15
`ServiceInvocationStreaming` isn't effective in responses from service invocation


ItalyPaleAle:
## In what area(s)?

/area runtime

## What version of Dapr?

> 1.10.x
> current master

## Description

The feature flag `ServiceInvocationStreaming` was added in 1.10 and should enable:

- Using `CallLocalStream` for sidecar-to-sidecar communication
- Responses from `onDirectMessage` return data as a stream

It appears that the second thing isn't actually working.

This seems to be related to the fact that when the HTTP API object is initialized ([here](https://github.com/dapr/dapr/blob/640a0cdfdaa6c3f2fbb0b992f55ab41375bf6601/pkg/runtime/runtime.go#L1390)), this line was not included:

```go
IsStreamingEnabled: a.globalConfig.IsFeatureEnabled(config.ServiceInvocationStreaming),
```

Sadly, adding the line above isn't the end of the issue. Because HTTP requests go through a conversion from fasthttp to net/http (for applying the middlewares), and then back to fasthttp, when a handler calls `reqCtx.Hijack()` (which is necessary to use streaming), that is then lost. So setting `IsStreamingEnabled` would actually cause no data to be returned to the client.

PR #6248, opened 2023-04-16
Migrate HTTP server to net/http

### LINKED ISSUE #4979: Migrate HTTP server from fasthttp to net/http

ItalyPaleAle:
- Fixes #4979 (although more work will be needed to completely remove fasthttp, see below)
- Necessary (but not quite sufficient) to fix also:
  - #5392
  - #6246
  - #6364
  - microsoft/azure-container-apps#760

This PR changes the HTTP server's listeners from fasthttp to net/http and fixes (or enables fixing) a lot of the scenarios that are blocked by #4979. Immediately, this should enable the web server to use HTTP/2 when TLS is enabled.

## Included and not included

What this PR includes:

- Changes the listeners for the 3 HTTP servers (app, healthz, pprof) from using fasthttp to using the standard library (net/http)
- Changes all built-in middlewares used by the servers above to be based on net/http

What this PR does NOT include:

- No handler has been updated from fasthttp to net/http, and handlers continue to work thanks to the "adapters"
  - This is not something new as in Dapr 1.10, all requests go through adapters twice to be converted fasthttp(listener + built-in middlewares)->net/http (middleware components)->fasthttp (handlers). Now there's only one conversion: net/http(everything but handlers)->fasthttp(handlers)
- Does not change the router, which is still the fasthttp one. This is blocking updating handlers from fasthttp to net/http, and needs to be done in a separate PR.
- Does not enable H2C. Should be done in a separate PR. 
- Some unit tests, including many (not all) in `pkg/http` still use fasthttp to create the test server. This isn't a concern at least until the handlers are migrated, but it should be changed in the future.

All features continue to work as expected, as testified by the fact that tests have not changed. Only one exception to this rule is that when performing HTTP->gRPC invocations, the "dapr-host" metadata is not included anymore (this is probably ok also because HTTP->gRPC service invocation is being deprecated)

## Next steps

- [ ] Update the router to one based on net/http, and apply the adapters only on individual handlers rather than on the router itself.
- [ ] Migrate the Universal API adapter from fasthttp to net/http
- [ ] Migrate APIs that can use streaming (such as service invocation) to be based on net/http and leverage streaming for input and output
- [ ] Migrate all other APIs from fasthttp to either net/http handlers or Universal APIs.


pkg/http/api.go
```diff
@@ -1383,32 +1372,20 @@ func (a *api) onDirectMessage(reqCtx *fasthttp.RequestCtx) {
...
 		respond(reqCtx, withError(fasthttp.StatusInternalServerError, NewErrorResponse("ERR_DIRECT_INVOKE", fmt.Sprintf(messages.ErrDirectInvoke, targetID, "response object is nil"))))
-		cancel()
 		return
 	}
+	defer resp.Close()
 
 	statusCode := int(resp.Status().Code)
 
-	// TODO @ItalyPaleAle: Make this the only path once streaming is finalized
-	if a.isStreamingEnabled {
```
ItalyPaleAle, line 1393: Reviewers: this was removed for now since it's not effective (see #6246 )

pkg/http/server.go
```diff
@@ -259,47 +281,39 @@ func (s *server) usePublicRouter() fasthttp.RequestHandler {
...
-			nethttpadaptor.NewNetHTTPHandlerFunc(next),
-		),
-	)
+func (s *server) useComponents(next http.Handler) http.Handler {
+	return s.pipeline.Apply(next)
 }
 
-func (s *server) useCors(next fasthttp.RequestHandler) fasthttp.RequestHandler {
+func (s *server) useCors(next http.Handler) http.Handler {
+	// TODO: Technically, if "AllowedOrigins" is "*", all origins should be allowed
```
ItalyPaleAle, line 289: Reviewers: Dapr is likely behaving incorrectly here. Setting "*" in CORS should mean "allow all origins", but Dapr <=1.10 was actually omitting the CORS headers entirely. Since this is a backwards-incompatible change, I am not touching it now.

utils/responsewriter/README.md
```diff
@@ -0,0 +1,29 @@
+# responsewriter
+
+This package contains code from [`github.com/urfave/negroni`](https://github.com/urfave/negroni).
```
daixiang0, line 3: Why can not import directly? Do we change some codes? If so, please comment changes.

ItalyPaleAle: Yes there are significant changes, such as support for user values (which we need when moving from fasthttp - may be removed in the future) and removal of "features" (that negroni only needs): https://github.com/dapr/dapr/pull/6248/files#diff-83f49a9138e9ad18fbee03f55f1035fa5ca1a626e1777ec037752212db0e11b7R74

Also, I would prefer not import the entire negroni framework if we just need a single struct. Importing the entire package would cause a lot more code to be brought in that we don't need.

daixiang0: Got, please comment changes in the README.

ItalyPaleAle: Done

yaron2:
> /ok-to-perf

Long have we waited for this change... Very curious about the perf results

ItalyPaleAle:
@yaron2 all perf tests are passing, and looks like perf may actually be **improving**

Looking at the service invocation perf test (which is particularly useful here as it is among those more sensitive to the web server's own perf):

[This PR](https://github.com/dapr/dapr/actions/runs/4722655475)
- P50: 1.38ms
- P90: 1.98ms

Three recent perf tests from master ([1]https://github.com/dapr/dapr/actions/runs/4722541970) [2](https://github.com/dapr/dapr/actions/runs/4686319469) [3](https://github.com/dapr/dapr/actions/runs/4690592444))
- Added latency avg: 1.76ms / 1.50ms / 1.47ms
- P90: 2.86ms / 2.13ms /2.09ms

yaron2:
This is a horrible regression in general, at the very least for 1.9 we had ~1.10ms for P90 and I am quite certain for 1.10 as well (will check later for 1.10.0). The regression in master started before this PR

ItalyPaleAle:
@yaron2 Not sure about 1.9 because logs have expired for those runs.

This is what I could find from the release-1.10 branch:

[First](https://github.com/dapr/dapr/actions/runs/4235229688/jobs/7358708370)
image

[Second](https://github.com/dapr/dapr/actions/runs/4318079447/jobs/7536068522)
image

yaron2:
> @yaron2 Not sure about 1.9 because logs have expired for those runs.
> 
> This is what I could find from the release-1.10 branch:
> 
> [First](https://github.com/dapr/dapr/actions/runs/4235229688/jobs/7358708370) image
> 
> [Second](https://github.com/dapr/dapr/actions/runs/4318079447/jobs/7536068522) image

Right, the first run is an outlier, the 2nd was the more common perf numbers we've seen. Now we need to check what caused the regression

ItalyPaleAle:
In any case, this should not block this PR, correct? As it seems at the very least this isn't causing further regression (and may be having a positive impact actually)

PS: I wonder if it's also possible the perf regression may not be our code but some infra-related thing. Since between the two runs linked above on release-1.10, the biggest difference seems to be time. Just wondering.

yaron2:
We will want to merge this at the beginning of the 1.12 milestone to make sure we can release 1.11 compatible versions to test thoroughly.

ItalyPaleAle:
This is disappointing as we had this scheduled for 1.11. We have work that depends on this right now that was scheduled for 1.11: complete streaming (still in preview), high-level crypto, among others.

ItalyPaleAle:
This is sad as we had this scheduled for 1.11. We have work that depends on this right now that was scheduled for 1.11: complete streaming (still in preview), high-level crypto, among others.

Any chance we can get this into 1.11 and unblock these scenarios?

daixiang0:
Good things are never too late. I agree that merge it later then we have enough time to ensure it is good enough.

Not only for this PR, for other big PRs, I do not want to merge them near the end of the milestone.

daixiang0:
BTW, do we also need to migrate in all components? It looks a big job.

yaron2:
> BTW, do we also need to migrate in all components? It looks a big job.

Middleware components support the net/http signatures

ItalyPaleAle:
@artursouza 

Fixed all the issues reported by the SDK tests:

- Java: issue was with the shutdown sequence. Stopping a net/http server returns an error always, which normally is ErrServerClosed, and had to be ignored.
- Python: issue was related to trace headers not being converted between net/http and fasthttp. The routes are still implemented as fasthttp, but the middleware that sets the tracing span now uses net/http. I added a temporary compatibility layer to make it work (until the routes are converted to net/http)

yaron2 (approved):

PR #6249, opened 2023-04-17
Adding ok-to-perf-components command

ASHIQUEMD:
# Description

Adding a unique command for components perf test, so that, it doesn't get executed when normal perf test command (ok-to-perf) is executed.

### DIFF
--- a/.github/scripts/dapr_bot.js
+++ b/.github/scripts/dapr_bot.js
@@ -347,7 +347,7 @@ async function cmdOkToPerfComponents(github, issue, isFromPulls, args) {
         await github.repos.createDispatchEvent({
             owner: issue.owner,
             repo: issue.repo,
-            event_type: 'perf-test-components',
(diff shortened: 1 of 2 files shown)


.github/scripts/dapr_bot.js
```diff
@@ -303,3 +311,50 @@ async function cmdOkToPerf(github, issue, isFromPulls, args) {
...
+            command: 'ok-to-perf-components',
+            args,
+            issue: issue,
+        }
+
+        // Fire repository_dispatch event to trigger e2e test
+        await github.repos.createDispatchEvent({
+            owner: issue.owner,
+            repo: issue.repo,
+            event_type: 'perf-test-components',
```
mukundansundar, line 350:
```suggestion
            event_type: 'components-perf-test',
```

mukundansundar (changes requested):

mukundansundar (approved):

PR #6253, opened 2023-04-17
Reduce QPS target for actor_reminder perf test

ItalyPaleAle:
Our perf tests are failing pretty much 100% of times due to the actor_reminder test: https://github.com/dapr/dapr/actions/workflows/dapr-perf.yml

We know we have perf issues there, but addressing them is a long-term project which may not even do too much until a "v2" comes out.

In the meanwhile, this PR reduces the QPS target to 50 (from 55) which should make the tests pass.


artursouza (changes requested):
This is a regression. Do we know where did this start?

ItalyPaleAle:
> This is a regression. Do we know where did this start?

It's been going on for months at this point. Some changes that were made with the goal of correctness may have had an impact too.

mukundansundar:
> @artursouza looks like #6186 may help a bit. The actor_reminder perf test passed 2/3 times in that PR, which is a lot more than on master right now.

Is this due to state changes in actors as well?

artursouza:
OK. I will accept this PR.

artursouza (approved):

Issue #6177, opened 2023-04-03
Allow app channel address override


yaron2:
We currently hard code the app channel address to 127.0.0.1. We should allow to override this as this is blocking many scenarios where 127.0.0.1 doesn't resolve, such as running Dapr in Dockerized environments.

olitomlinson:
Can confirm this is blocking the utilisation of TestContainer SDKs with Dapr.

![image](https://user-images.githubusercontent.com/4224880/229487742-ee4c0994-64b8-4267-b7ae-bd79abbbb4fc.png)

https://dotnet.testcontainers.org/api/best_practices/

olitomlinson:
Is this still on track for 1.11? Many thanks!

yaron2:
> Is this still on track for 1.11? Many thanks!

Indeed

olitomlinson:
@yaron2 tested this successfully on the edge build, works great :

`docker run daprio/daprd:edge-mariner ./daprd -app-id 123 -app-channel-address host.docker.internal -app-port 80 -log-level debug`

`docker run -p 80:80 -d nginx`

Lauriy:
Would have loved to have seen -app-channel-address here 2 days ago : )

https://docs.dapr.io/reference/arguments-annotations-overview/

PR #6281, opened 2023-04-24
Make app channel address configurable instead of hard coded

yaron2:
Closes https://github.com/dapr/dapr/issues/6177.


pkg/actors/actors.go
```diff
@@ -598,7 +598,7 @@ func (a *actorsRuntime) callRemoteActor(
 }
 
 func (a *actorsRuntime) isActorLocal(targetActorAddress, hostAddress string, grpcPort int) bool {
-	return strings.Contains(targetActorAddress, "localhost") || strings.Contains(targetActorAddress, "127.0.0.1") ||
+	return strings.Contains(targetActorAddress, "localhost") || strings.Contains(targetActorAddress, a.config.AppChannelAddress) ||
```
ItalyPaleAle, line 601: Any reason why localhost is hardcoded but 127.0.0.1 may not be present? If app channel isn't on localhost shouldn't that be missing too

ItalyPaleAle (changes requested):
Please add an E2E test for this. Modifying an existing test (app health could be a good one since it uses a bunch of features of the app channel) may be enough

pkg/actors/actors.go
yaron2: Further investigation has shown this section checks the `daprd` address and not the app channel, so this was unneeded. Changed it back

mukundansundar:
@yaron2 this requires a change in CLI and the run template also ...

pkg/actors/actors.go
yaron2: Resolving this now as the code has been reverted

yaron2:
As FYI, our e2e test infrastructure didn't really support apps with the sidecar disabled, so fixed that along the way.

pkg/injector/annotations/annotations.go
```diff
@@ -67,6 +67,7 @@ const (
 	KeyPluggableComponentsSocketsFolder = "dapr.io/pluggable-components-sockets-folder"
 	KeyPluggableComponentContainer      = "dapr.io/component-container"
 	KeyPluggableComponentsInjection     = "dapr.io/inject-pluggable-components"
+	KeyAppChannel                       = "dapr.io/app-channel-address"
```
artursouza, line 70: Does this have any real life application? If this is added, then the app has a Deployment in K8s to be talked via localhost, not?

yaron2: Yes, this tells Dapr to send events and messages to this configured address which is not localhost

daixiang0 (approved):

Issue #6283, opened 2023-04-24
1.10.6 HotFix


artursouza:
- [X] Dapr maintainer(s) communicate out the discovery of issues that require a patch release and the reason why to the Discord Maintainers and Release channels. Patch releases are made for one of categories below:
   - [ ] Security vulnerability
   - [X] Regression that does not have a work around:
     - #6268  
     - https://github.com/dapr/components-contrib/issues/2786
   - [ ] Broken mainline scenario that has a missing test case:
- [x] Create v1.10.6-rc.1 Tag
- [x] Performance tests passing: 
- [x] End to End tests passing on Linux and Windows:
- [x] New test case written to catch future occurrences.  (Including in PRs above)
- [x] Notify users to try RC1
- [x] Update the longhaul tests to use 1.10.6-rc.1
- [x] Write release notes [@artursouza ]: 
- [x] Review release notes [@dapr/maintainers-dapr ]
- [x] Create v1.10.6 Tag [@dapr/maintainers-dapr ]
- [x] Backport fixes into master branch [@dapr/maintainers-dapr ]
- [x] Update the documentation: Latest version & versions in supported releases [@nyemade-uversky ]
- [x] Push new tag in installer-bundle repo. [@artursouza or @shivamkm07]
- [x] Announce the hotfix release on Discord Announcements channel [@nyemade-uversky or @greenie-msft ]

PR #6306, opened 2023-05-04
Fix error propagation regression detected by Java SDK.

artursouza:
# Description

Fix error propagation regression detected by Java SDK.

## Issue reference

Part of: https://github.com/dapr/dapr/issues/6283

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing


yaron2 (dismissed):

daixiang0 (approved):

artursouza:
The driving factor for this is a test in Java SDK. Here is what I found:
- The current behavior in 1.10.6 (prior to this PR) fails that test BUT it passes if the test is changed to assert the behavior in 1.9
- There was a change in 1.10 runtime that failed this same test BUT we agreed to change the expectation of the test since the message was better.

Since we have users complaining about 1.10 because of this behavior change, I recommend we DO NOT MERGE this PR and change the Java SDK test for the same expectation as in 1.9. There is a missing test to reproduce the issue pointed by users.

Issue #6301, opened 2023-05-02
v1.11 Endgame


artursouza:
## Release team
Release lead: @artursouza
Release lead shadow(s): @nyemade-uversky @robertojrojas
Perf test lead: @JoshVanL @robertojrojas
Long haul test: @halspang
Build manager: @JoshVanL

Read about the [Dapr release lifecycle](https://github.com/dapr/community/blob/master/release-process.md)

## Milestone dates
Code freeze: ~05/01/2023~ - 05/18/2023
Endgame starts: 05/19/2023
Endgame ends: 06/05/2023
1.11.0 release: 06/06/2023

Release Tasks:

------------------
ETA for all below: TBD

- [x] All P0 issues below are resolved PRIOR to code freeze:
  - [x] Actor reminder & timer serialization regressions are fixed in 1.10.6 and merged into master - @artursouza
  - [x] ~gRPC service invocation error message regression (detected by Java SDK) is fixed in 1.10.6 and merged into master~ Not a regression, it is due to a legit hotfix and error message is not how it was in 1.9 - @artursouza
  - [x] Perf regressions in master branch - @yaron & @JoshVanL 
  - [ ] ~Flakiness of placement unit tests https://github.com/dapr/dapr/issues/6265~ - @JoshVanL 
  - [ ] ~Flakiness of external service invocation E2E tests on Windows~ - @yaron2
  - [x] Eventgrid component regression on metadata change - @berndverst 
  - [x] Bulk secret regression detected by Java SDK - @mukundansundar
  - [x] Add stable_components build flavor tag for newly stable components - @berndverst 
  - [x] ~https://github.com/dapr/dapr/issues/6316~ https://github.com/dapr/dapr/pull/6325 - @artursouza 
  - [x] https://github.com/dapr/dapr/pull/6342 - @ItalyPaleAle @berndverst 
  - [x] https://github.com/dapr/dapr/issues/6324 - @JoshVanL 
- [x] All MUST HAVE feature and improvements planned for 1.11 are merged:
  - [x] https://github.com/dapr/java-sdk/pull/846 - @artursouza @mukundansundar 
  - [x] https://github.com/dapr/dapr/pull/6280
  - [x] https://github.com/dapr/dapr/pull/6225
  - [x] https://github.com/dapr/dapr/issues/6099 - @ItalyPaleAle
- [x] CLI changes blocked by runtime RC
  - [x] https://github.com/dapr/cli/pull/1276
[...]

Issue #6265, opened 2023-04-19
Investigate and fix flaky UTs for placement package


mukundansundar:
## In what area(s)?

/area runtime

## What version of Dapr?

master
## Expected Behavior

Placement package UTs are not flaky.

## Actual Behavior

Placement package UTs fail more than 50% of the time mostly in Windows followed sometime in Intel Mac tests.

Example runs where they fail. But generally any PR that is currently open might have a failure in windows or Intel Mac. 

https://github.com/dapr/dapr/actions/runs/4744388572/jobs/8426648380?pr=5977
https://github.com/dapr/dapr/actions/runs/4744467349/jobs/8426792502?pr=6192

## Steps to Reproduce the Problem

Run the placement package UTs with debug logs enabled multiple times in a row. >10 with `go test -count 20` etc.

## Release Note

RELEASE NOTE:

artursouza:
This test should be moved to an E2E or integration test, so we can also refactor or make changes to placement service.

ItalyPaleAle:
I fixed this in master 2 weeks ago I believe.

PR #6308, opened 2023-05-04
Add `time.Sleep`s to placement tests to improve flakes

JoshVanL:
Part of the [`v1.11` Endgame](https://github.com/dapr/dapr/issues/6301).

Although painful to do so, because of the nature of the raft library we are using being nondeterministic in its state machine and being reliant on wall clock time, I have added `time.Sleep`s to the placement tests to improve the reliability of the placement tests on slower test runners (windows & mac). This obviously increases the run time of the tests, but when testing on this PR, has greatly reduced the runs being flaky.

We can revisit these tests if we still find them to be flaky, but this should at least improve things somewhat.

Improves (not fix): https://github.com/dapr/dapr/issues/6265

### DIFF
--- a/pkg/placement/ha_test.go
+++ b/pkg/placement/ha_test.go
@@ -75,6 +75,13 @@ func TestPlacementHA(t *testing.T) {
 
 	// Run tests
 	t.Run("elects leader with 3 nodes", func(t *testing.T) {
+		// It is painful that we have to include a `time.Sleep` here, but due to
+		// the non-deterministic behaviour of the raft library we are using we will
+		// later fail on slower test runner machines. A clock timer wait means we
+		// have a _better_ chance of being in the right spot in the state machine
+		// and the network has died down. Ideally we should move to a different
+		// raft library that is more deterministic and reliable for our use case.
+		time.Sleep(time.Second * 3)
 		require.NotEqual(t, -1, findLeader(t, raftServers))
 	})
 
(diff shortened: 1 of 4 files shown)


pkg/placement/ha_test.go
```diff
@@ -94,17 +101,27 @@ func TestPlacementHA(t *testing.T) {
...
+			// later fail on slower test runner machines. A clock timer wait means we
+			// have a _better_ chance of being in the right spot in the state machine
+			// and the network has died down. Ideally we should move to a different
+			// raft library that is more deterministic and reliable for our use case.
+			time.Sleep(time.Second * 3)
+
+			require.Eventually(t, func() bool {
+				return oldLeader != findLeader(t, raftServers)
+			}, time.Second*10, time.Millisecond*100)
+		})
```
mukundansundar, line 124: for any `findLeader` calls, is `require.Eventually` needed?
Since `findLeader` itself has that for 10s with step being 1s. 
Here the wrapping `require.Eventually` is 10s with step as 100ms. Is this needed?

JoshVanL: It is needed in the case when we are waiting for the leader to change.

JoshVanL:
@artursouza

pkg/placement/ha_test.go
```diff
@@ -75,6 +75,13 @@ func TestPlacementHA(t *testing.T) {
 
 	// Run tests
 	t.Run("elects leader with 3 nodes", func(t *testing.T) {
+		// It is painful that we have to include a `time.Sleep` here, but due to
+		// the non-deterministic behaviour of the raft library we are using we will
+		// later fail on slower test runner machines. A clock timer wait means we
+		// have a _better_ chance of being in the right spot in the state machine
+		// and the network has died down. Ideally we should move to a different
+		// raft library that is more deterministic and reliable for our use case.
+		time.Sleep(time.Second * 3)
```
daixiang0, line 84: Is it possible to use loop to check rather than hardcode sleep?

artursouza (approved):

PR #6304, opened 2023-05-03
Adds scheduled, and manual test runner, for the Dapr Java + Python SDK test validation.

JoshVanL:
Part of #5797.

Adds GitHub action workflow for running SDK tests against Dapr HEAD. These are run on an infrequent schedule, as well as on pull requests when someone comments `/test-sdk-all`.

Tests  the Java SDK, which can also be triggered with `/test-sdk-java`.

Tests  the Python SDK, which can also be triggered with `/test-sdk-python`.

Not fully tested e2e just because of how GH actions work, but hopefully should be close to working.

/cc @artursouza

Issue #5797, opened 2023-01-24
Run Dapr's Java SDK tests on dapr/dapr PRs


artursouza:
We had situations where breaking changes were only detected by SDKs, mainly by Java SDK. This issue is to create a new GitHub Workflow that will run the Java SDK integration tests and allow other SDK tests to also be included in the future. This way, PRs are also validated against Java SDK.

UPDATE: The workflow should be generic enough to allow other SDKs to be added as validation steps too.

mukundansundar:
IMO this should be a manually triggered workflow similar to ok-to-test since there can be cases where there are breaking changes in dapr which necessitates a change in sdk eg: config api changes

mukundansundar:
Ideally one additional step or job should be to trigger the example validations if any. As some sdks have examples instead of integration tests.

JoshVanL:
/close

Done in https://github.com/dapr/dapr/pull/6304

PR #6309, opened 2023-05-04
Adds `go-sdk` to sdk test github workflow with corresponding commands

JoshVanL:
Branched from #6304, part of https://github.com/dapr/dapr/issues/5797 

Job can be manually triggered on a PR with the commands `/test-sdk-all` or `test-sdk-go`.


.github/workflows/dapr-test-sdk.yml
```diff
@@ -0,0 +1,220 @@
...
+          ./dist/linux_amd64/release/placement --healthz-port 9091 &
+      - name: Install dependencies
+        run: |
+          cd python-sdk
+          python -m pip install --upgrade pip
+          pip install setuptools wheel twine tox
+      - name: Check Python Examples
+        run: |
+          cd python-sdk || true
+          tox -e examples
```
artursouza, line 88: Can you add reporting on test progress as PR comment? This was done in E2E tests and it looks super nice and easy to follow progress.

JoshVanL: Hi @artursouza, I have updated the non-draft Java and Python PR https://github.com/dapr/dapr/pull/6304/commits/b1f496f9eaaaceb9b59690e247067452501131ea

mukundansundar (approved):

mukundansundar:
Not for this PR, but I believe we need to enhance running tests against SDKs. 
As it exists right now, this workflow tests the PR code against master branch of the SDK. But in certain scenarios that is not the right approach, for example running against a PR against a release branch, ideally we would need to run against the latest stable version of the SDK. 
I propose we add another parameter for the individual commands `/test-sdk-java release-1.9` so that it tests the PR against the release branch specified. For the `/test-sdk-all` the current behavior is good. 

cc @artursouza @yaron2 WDYT?

daixiang0:
> Not for this PR, but I believe we need to enhance running tests against SDKs. As it exists right now, this workflow tests the PR code against master branch of the SDK. But in certain scenarios that is not the right approach, for example running against a PR against a release branch, ideally we would need to run against the latest stable version of the SDK. I propose we add another parameter for the individual commands `/test-sdk-java release-1.9` so that it tests the PR against the release branch specified. For the `/test-sdk-all` the current behavior is good.
> 
> cc @artursouza @yaron2 WDYT?

Agree.

ItalyPaleAle:
This doesn't seem to be working: https://github.com/dapr/dapr/actions/runs/5293791337/jobs/9583394466

.github/workflows/dapr-test-sdk.yml
```diff
@@ -464,3 +466,124 @@ jobs:
...
+              );
+            }
+      - name: Create PR comment
+        if: env.PR_NUMBER != ''
+        uses: artursouza/sticky-pull-request-comment@v2.2.0
+        with:
+          header: ${{ github.run_id }}-go
+          number: ${{ env.PR_NUMBER }}
+          hide: true
+          hide_classify: OUTDATED
```
ItalyPaleAle, line 511: Please delete these or re-running the test will create a new comment without links.

PR #6310, opened 2023-05-04
Adds js-sdk to sdk test github workflow with corresponding commands

JoshVanL:
Branched from #6304, part of https://github.com/dapr/dapr/issues/5797 

Job can be manually triggered on a PR with the commands `/test-sdk-all` or `test-sdk-go`.


.github/workflows/dapr-test-sdk.yml
```diff
@@ -314,3 +319,133 @@ jobs:
...
+      ( github.event_name == 'repository_dispatch' &&
+        (
+          github.event.action == 'test-sdk-all' ||
+          github.event.action == 'test-sdk-js'
+        )
+      )
+    name: "JS SDK verification tests"
+    runs-on: ubuntu-latest
+    env:
+      NODE_VER: 16.14.0
```
ItalyPaleAle, line 335: Should this use a matrix to test different Node.js versions? 16 is going to reach EOL in 3 months

JoshVanL: Do we care so much here? I'd say that is more a `dapr/js-sdk` concern, we just want more of a signal on `dapr/dapr`.

Def shouldn't use `16` though since if that's case. Looks like `18` is more appropriate?

p.s. pls don't push to my PR branches :sweat_smile: I can see this is going to fix the PR comment issue though, so thank you

ItalyPaleAle: 16 is fine

I wanted to fix the git clone issue so this PR could be merged into master so we can validate the bulk state store.

artursouza (approved):

PR #6323, opened 2023-05-09
Move `runtime/{init_error.go,retriable_error.go}` into separate shared package

JoshVanL:
As part of a larger effort in breaking up runtime.go to be more modular so to eventually implement features like [hot reloading](https://github.com/dapr/dapr/issues/1172).

Moves the custom runtime init and retriable error types into a shared `runtime/errors` package.


pkg/runtime/errors/init.go
```diff
@@ -43,8 +43,8 @@ func (e *InitError) Unwrap() error {
 	return e.err
 }
 
-// NewInitError returns an InitError wrapping an existing context error.
-func NewInitError(kind InitErrorKind, entity string, err error) *InitError {
+// NewInit returns an InitError wrapping an existing context error.
+func NewInit(kind InitErrorKind, entity string, err error) *InitError {
```
daixiang0, line 47: I would like the old one since it is more clear, why change it?

pkg/runtime/errors/retriable.go
```diff
@@ -32,8 +32,8 @@ func (e *RetriableError) Unwrap() error {
 	return e.err
 }
 
-// NewRetriableError returns a RetriableError wrapping an existing context error.
-func NewRetriableError(err error) *RetriableError {
+// NewRetriable returns a RetriableError wrapping an existing context error.
+func NewRetriable(err error) *RetriableError {
```
daixiang0, line 36: Same as above.

pkg/runtime/errors/init.go
JoshVanL: This causes a stutter in the package function call:

```go
errors.NewInit
```

vs


```go
errors.NewInitError
```

ItalyPaleAle (approved):

mukundansundar (approved):

PR #6327, opened 2023-05-10
Adds initial integration testing scaffolding

JoshVanL:
PR adds initial scaffolding for running integration tests against the daprd binary.

These tests are intended to run particular scenarios and code paths of daprd, where the test suite is lightweight and should be simple for developers to add to the test cases add-hock and regularly.

The daprd binary is built on the fly during test runtime at init. The daprd binary path can be overridden with the env var `DAPR_INTEGRATION_DAPRD_PATH`.

Integration tests are run in parallel, with a maximum of 3 test cases being run at the same time. This might be premature optimization, but the mac and windows test runners are quite slow.

Integration tests are run with the `make test-integration` Make target.

We use a global env var registry strategy for registering integration test cases. Blank imports of integration test packages invoke `init` functions that register the test case.

The daprd process is also being signaled an interrupt twice because it is not gracefully shutting down correctly on the first, so we force terminate. See #2478

We can assume that other binaries such as placement and sentry will be optionally run in later added tests.

/cc @artursouza


tests/docs/writing-integration-test.md
```diff
@@ -0,0 +1,53 @@
...
+Integration tests are run against a live running daprd binary locally. Each test
+scenario is run against a new instance of daprd, where each scenario modifies
+the daprd configuration to suit the test. Tests are expected to complete within
+seconds, ideally less than 5, and should not take longer than 30. Daprd is
+always built from source within the test.
+
+
+## Invoking the test
+
+`go test -v --race ./tests/integration`
```
ItalyPaleAle, line 14: Using `--race` is not ideal in "normal testing" because it uses CGo.

JoshVanL: Why is that a bad thing?

ItalyPaleAle: We build and ship dapr with CGo disabled. Running tests with CGo enabled could cause some issues we haven't yet figured out.
Also tests with `--race` are slower and they require rebuilding Dapr and all deps with CGo support :)

JoshVanL: > could cause some issues

What issues would this lead to?

> slower and they require rebuilding Dapr

You're going to be building dapr to run these tests as is stands, and currently, this test takes the same amount of time with or without `--race`- about 2 seconds.

 ---

Regardless of all this, developers should always be testing their software with a race checker.

JoshVanL: I have added a commit to ensure that CGO is disabled when building the daprd binary.

https://github.com/dapr/dapr/pull/6327/commits/540a2429d85f6c588b016db4257116b63b36d49d

tests/integration/framework/framework.go
```diff
@@ -0,0 +1,221 @@
...
+	appID   string
+
+	appPort                 int
+	grpcPort                int
+	httpPort                int
+	internalGRPCPort        int
+	publicPort              int
+	metricsPort             int
+	profilePort             int
+	runError                func(error)
```
daixiang0, line 52: It looks like an error not a func, please rename it.

tests/integration/suite/ports/ports.go
```diff
@@ -0,0 +1,54 @@
...
+		"grpc":          cmd.GRPCPort,
+		"http":          cmd.HTTPPort,
+		"metrics":       cmd.MetricsPort,
+		"internal-grpc": cmd.InternalGRPCPort,
+		"public":        cmd.PublicPort,
+	} {
+		assert.Eventuallyf(t, func() bool {
+			_, err := net.Dial("tcp", fmt.Sprintf("localhost:%d", port))
+			return err == nil
+		}, time.Second*5, time.Millisecond, "port %s (:%d) was not available in time", name, port)
```
daixiang0, line 52: Is it enough? Could we use a loop to detect it?

tests/integration/integration_test.go
```diff
@@ -0,0 +1,59 @@
...
+
+		// Use a consistent temp dir for the binary so that the binary is cached on
+		// subsequent runs.
+		daprdPath := filepath.Join(os.TempDir(), "dapr_integration_tests/daprd")
+		if runtime.GOOS == "windows" {
+			daprdPath += ".exe"
+		}
+
+		// Ensure CGO is disabled to avoid linking against system libraries.
+		os.Setenv("CGO_ENABLED", "0")
```
daixiang0, line 45: The testing has its setenv func, and exec.Command has env params either, if we use the os library, we need to clean it when test ends.

JoshVanL: It doesn't actually matter in this case since the integrations tests will only be run by themselves, but will use `testing.Setenv` for correctness.

JoshVanL:
Thanks @daixiang0, please take another look.

tests/integration/framework/kill/kill_windows.go
```diff
@@ -0,0 +1,55 @@
...
+import (
+	"os/exec"
+	"syscall"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"golang.org/x/sys/windows"
+)
+
+func interrupt(t *testing.T, cmd *exec.Cmd) {
```
mukundansundar, line 28: just a question. does this work as expected, has this been validated?

does `taskkill` work? https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/taskkill

JoshVanL: I don't really have access to a windows machine so difficult for me to get quick round trip validations done. Very happy to use taskkill instead if that is the more canonical way of doing this.

tests/integration/framework/kill/kill_posix.go
```diff
@@ -0,0 +1,34 @@
...
+	"os"
+	"os/exec"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func interrupt(t *testing.T, cmd *exec.Cmd) {
+	// TODO: daprd does not currently gracefully exit on a single interrupt
```
ItalyPaleAle, line 29: Make sure you check the value of graceful-shutdown.

A single kill signal will start the graceful shutdown, and IIRC the default is 5s.

To force a shutdown, send a second kill signal.

tests/integration/framework/framework.go
```diff
@@ -0,0 +1,221 @@
...
+	ProfilePort      int
+}
+
+func RunDaprd(t *testing.T, ctx context.Context, opts ...RunDaprdOption) *Command {
+	t.Helper()
+
+	uid, err := uuid.NewUUID()
+	require.NoError(t, err)
+
+	appListener, err := net.Listen("tcp", "127.0.0.1:0")
```
ItalyPaleAle, line 89: Can we use a UNIX socket as listener? It seems that every time we use an actual TCP socket in tests, we get flakiness.

ItalyPaleAle (changes requested):

tests/integration/framework/iowriter/iowriter.go
```diff
@@ -0,0 +1,36 @@
...
+type stdwriter struct {
+	t *testing.T
+}
+
+func New(t *testing.T) io.WriteCloser {
+	return &stdwriter{t}
+}
+
+func (w *stdwriter) Write(p []byte) (n int, err error) {
+	w.t.Log(w.t.Name() + ": " + string(p))
```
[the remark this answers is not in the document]
JoshVanL: This was actually fixed in a further PR. Will cherry pick this in.

tests/integration/framework/kill/kill_posix.go
JoshVanL: We are checking the exit code  [here](https://github.com/dapr/dapr/pull/6327/files#diff-5c299386208e2c1d7b4b5257eb6fb95e522ea4f94d6bb0d6a8f67be94f0c416bR213).

It is not really a "graceful" shutdown today, because we are not propagating a shutdown signal to everything properly and waiting for all resources to be released. 5s wait time is not appropriate for these tests, especially when nothing is happening in that 5s time.

When [graceful shutdown](https://github.com/dapr/dapr/issues/2478) is finished, we can get rid of this "hack". This integration testing framework is actually one of the requirements for that work anyway :sweat_smile:

tests/integration/framework/framework.go
```diff
@@ -0,0 +1,221 @@
...
+	t.Helper()
+
+	t.Log("waiting for daprd process to exit")
+
+	c.runErrorFnFn(c.cmd.Wait())
+	assert.NotNil(t, c.cmd.ProcessState, "process state should not be nil")
+	assert.Equalf(t, c.exitCode, c.cmd.ProcessState.ExitCode(), "expected exit code to be %d", c.exitCode)
+}
+
+func freeport(t *testing.T) int {
```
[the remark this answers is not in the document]
JoshVanL: Alleviating the race conditions was also done in a [future PR](https://github.com/dapr/dapr/pull/6340/files#diff-69da0c9a305e8f9283c452ce429a865f03fc2799f5a92bae950b944d46570947R1). I can cherry pick this in also.

tests/integration/framework/framework.go
ItalyPaleAle: Actually.... *no* 😱

You can configure Dapr to listen on a UDS (`--unix-domain-socket`) for its APIs. But you cannot make Dapr use a UDS to talk to an app.

This seems to be a missing piece.

tests/integration/framework/kill/kill_windows.go
JoshVanL: After playing around with `taskill`, it seems that it does not gracefully shutdown the processes.

I'll revert back to the `CTRL_BREAK_EVENT` approach.

tests/integration/framework/iowriter/iowriter.go
```diff
@@ -0,0 +1,42 @@
...
+	t *testing.T
+}
+
+func New(t *testing.T) io.WriteCloser {
+	return &stdwriter{t}
+}
+
+func (w *stdwriter) Write(p []byte) (n int, err error) {
+	for _, line := range bytes.Split(p, []byte("\n")) {
+		if len(line) == 0 {
```
ItalyPaleAle, line 32: Should probably still log this as an empty line?

tests/integration/framework/iowriter/iowriter.go
```diff
@@ -0,0 +1,42 @@
...
+type stdwriter struct {
+	t *testing.T
+}
+
+func New(t *testing.T) io.WriteCloser {
+	return &stdwriter{t}
+}
+
+func (w *stdwriter) Write(p []byte) (n int, err error) {
+	for _, line := range bytes.Split(p, []byte("\n")) {
```
ItalyPaleAle, line 31: This may have issues if the data is flushed before the line is complete.

For example, assume the producer wants to log `supercalifragilisticexpialidocious` but because of internal buffering, it is sent to `Write()` in 2 separate calls:

- `supercalifragil`
- `isticexpialidocious`

This will log that in 2 separate lines

Better would be to:

- When data comes in, rather than splitting for `\n`, just iterate byte-by-byte, storing everything in an internal buffer.
- If you encounter `\n` or EOF, then print out whatever is in the internal buffer.
- At the end of the loop in the Write method you may have data left into the internal buffer
- That data should be flushed out in the Close method.

tests/integration/suite/healthz/apphealthz.go
```diff
@@ -0,0 +1,117 @@
...
+	done    chan struct{}
+}
+
+func (a *AppHealthz) Setup(t *testing.T, _ context.Context) []framework.RunDaprdOption {
+	a.healthy.Store(true)
+	a.done = make(chan struct{})
+
+	mux := http.NewServeMux()
+	mux.HandleFunc("/foo", func(w http.ResponseWriter, r *http.Request) {
+		require.Equal(t, http.MethodGet, r.Method)
```
ItalyPaleAle, line 49: Note that code inside a handler is executed in a separate goroutine, so we shouldn't use test assertions here.

Instead, make the handler respond with the desired output, and then check the output in the caller.


```go
	mux.HandleFunc("/foo", func(w http.ResponseWriter, r *http.Request) {
		if a.healthy.Load() {
			w.WriteHeader(http.StatusOK)
		} else {
			w.WriteHeader(http.StatusServiceUnavailable)
		}

        fmt.Fprintf(w, "%s %s", r.Method, r.URL.Path)
	})
```

Then in the caller check that the body is the desired one

tests/integration/suite/healthz/apphealthz.go
```diff
@@ -0,0 +1,117 @@
...
+func (a *AppHealthz) Run(t *testing.T, ctx context.Context, cmd *framework.Command) {
+	assert.Eventually(t, func() bool {
+		_, err := net.Dial("tcp", fmt.Sprintf("localhost:%d", cmd.InternalGRPCPort))
+		return err == nil
+	}, time.Second*5, time.Millisecond)
+
+	a.healthy.Store(true)
+
+	assert.Eventually(t, func() bool {
+		resp, err := http.DefaultClient.Get(fmt.Sprintf("http://localhost:%d/v1.0/invoke/%s/method/myfunc", cmd.HTTPPort, cmd.AppID))
```
ItalyPaleAle, line 95: Please either use a request with a context with a timeout, or use a custom http.Client with a timeout. Otherwise, if the request hangs (not unusual if the server isn't fully up yet), it will just make this call hang forever, while it should fail and be retried.

tests/integration/suite/ports/ports.go
```diff
@@ -0,0 +1,54 @@
...
+	}
```
ItalyPaleAle, line 53:
```suggestion
		}, time.Second*5, 100*time.Millisecond, "port %s (:%d) was not available in time", name, port)
	}
```

Let's reduce the load a little bit on our slow runners :)

ItalyPaleAle (changes requested):

tests/integration/framework/iowriter/iowriter.go
JoshVanL: Makes sense :+1: Thanks @ItalyPaleAle :slightly_smiling_face:

tests/integration/suite/ports/ports.go
JoshVanL: :smile:

tests/integration/suite/healthz/apphealthz.go
JoshVanL: I have passed through the test context as is to the HTTP request. Each test scenario has a 30 sec timeout for the full case runtime.

JoshVanL:
Thanks @ItalyPaleAle, please take another look!

Also note I have moved building dapr into `integration.go`- vim-go has a bit of a wobbly building/editing build tagged files (`+build`) so minimizing code in those is good for me. I'm assuming it's also a bit painful in other editors.

Also added `DAPR_INTEGRATION_CONCURRENCY` to configure test concurrency. This would help debugging a test by isolating the process outputs. Later we may also want to add a flag which only prints the process output all at once at the end of the test so that multiple processes in the same test scenario aren't muddled. Maybe we want to do that anyway as default?

ItalyPaleAle:
> Also note I have moved building dapr into `integration.go`- vim-go has a bit of a wobbly building/editing build tagged files (`+build`) so minimizing code in those is good for me. I'm assuming it's also a bit painful in other editors.

Assuming vim-go uses gopls like VS Code, you can configure gopls to add the tags. See what we do in the dev container: https://github.com/dapr/dapr/blob/0c8292c590208bd38886245de7ee1b66b84e4779/.devcontainer/devcontainer.json#L73

> Later we may also want to add a flag which only prints the process output all at once at the end of the test so that multiple processes in the same test scenario aren't muddled. Maybe we want to do that anyway as default?

I think this actually makes a lot of sense. If we assume that the primary use of this is CI (yes, people can run the tests locally, but that's secondary), then having logs streamed out as they come may not matter much.

tests/integration/framework/framework.go
```diff
@@ -0,0 +1,217 @@
...
+				return
+			}
+			conn.Close()
+		}
+	}()
+
+	defaultExitCode := 0
+	if runtime.GOOS == "windows" {
+		// Windows returns 1 when we kill the process.
+		defaultExitCode = 1
```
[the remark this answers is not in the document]
ItalyPaleAle: What if we send an os.Interrupt signal?

ItalyPaleAle (dismissed):
Approved but I do like the idea of collecting all logs before writing them out.

Or maybe, it can be configured via an env var, to determined if logs are streamed out or collected all at once.

tests/integration/framework/framework.go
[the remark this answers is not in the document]
JoshVanL: As I understand it, the `CTRL_BREAK_EVENT` is the equivalent in windows 

https://github.com/dapr/dapr/pull/6327/files#diff-ff04f5380377d12324fcfe65bf31948e2356f889ceeb251c23cc686e40c22c67

JoshVanL:
Thanks @ItalyPaleAle, did the trick :slightly_smiling_face:  https://github.com/JoshVanL/nixos/blob/6b343ca5fe122a00e15e45119376e26e23747ecd/modules/dev/go.nix#L79

ItalyPaleAle (approved):

yaron2 (approved):

PR #6331, opened 2023-05-10
[1.11] Actually downgrade sarama to the same version as in contrib

berndverst:
# Description

Components-contrib has sarama downgraded to 1.37.2, but due to a previous `go mod tidy` dapr/dapr go.mod still had the indirect dependency pinned at 1.38.2 (go mod tidy does now downgrade dependencies).

This PR downgrades the indirect dependency in runtime to the correct version.


artursouza:
This PR can be closed as we merge release-1.10 into master after the hotfix.

mukundansundar:
closed since change incorporated once release-1.10 was merged into master.

PR #6332, opened 2023-05-11
Normalize URI Path in in nethttpadaptor

### LINKED ISSUE #6324: Actor invocation should not require dapr-app-id when using "//" over HTTP
## In what area(s)?

 /area runtime

> /area operator

> /area placement

> /area docs

> /area test-and-release

## What version of Dapr?

1.10.6

## Expected Behavior

When invoking an actor method over HTTP, it should not require app-id to be passed in. This was reproduced with HTTP but should be verified over gRPC too.

## Actual Behavior

Starting in Dapr 1.10.6 (maybe 1.10.0), invoking an actor method requires the app-id to be passed in, if "//" is included in the HTTP path.

Run any actor example. Here, we will use the one from Python SDK:

```sh
git clone git@github.com:dapr/python-sdk.git
cd python-sdk/examples/demo_actor/demo_actor
pip3 install -r requirements.txt
dapr run --app-id demo-actor --dapr-http-port=3500  --app-port 3000 -- uvicorn --port 3000 demo_actor_service:app
```

Then make the following PUT request:
```txt
PUT http://localhost:3500//v1.0/actors/DemoActor/1234/method/SetReminder
Content-Type: application.json
{
  "enabled": true
}
```

See it fail, but IT SHOULD WORK.
```json
{
    "errorCode": "ERR_DIRECT_INVOKE",
    "message": "failed getting app id either from the URL path or the header dapr-app-id"
}
```

Now, change it to this and see it work:
```txt
PUT http://dapr-app-id:demo-actor@localhost:3500//v1.0/actors/DemoActor/1234/method/SetReminder
Content-Type: application.json
{
  "enabled": true
}
```

This also works:
```txt
PUT http://localhost:3500/v1.0/actors/DemoActor/1234/method/SetReminder
Content-Type: application.json
{
  "enabled": true
}
```

## Release Note

RELEASE NOTE:

JoshVanL:
fixes #6324 

In the base nethttpadapter, we are currently using `SetRequestURI` for incoming requests. This results in fasthttp _not_ normalizing the HTTP request path which eventually leads to requests not being routed to their intended handler (see issue).

PR changes the adapter to use `SetQueryString`, and importantly, `SetPath` which _does_ normalize the path. This results in fixing downstream problems like the one in the issue.

/assign @artursouza

### DIFF
--- a/utils/nethttpadaptor/nethttpadaptor.go
+++ b/utils/nethttpadaptor/nethttpadaptor.go
@@ -46,7 +46,8 @@ func NewNetHTTPHandlerFunc(h fasthttp.RequestHandler) http.HandlerFunc {
 			}
 			c.Request.SetBody(reqBody)
 		}
-		c.Request.SetRequestURI(r.URL.RequestURI())
+		c.Request.URI().SetQueryString(r.URL.RawQuery)
+		c.Request.URI().SetPath(r.URL.Path)
 		c.Request.URI().SetScheme(r.URL.Scheme)
 		c.Request.SetHost(r.Host)
 		c.Request.Header.SetMethod(r.Method)
--- a/utils/nethttpadaptor/nethttpadaptor_test.go
+++ b/utils/nethttpadaptor/nethttpadaptor_test.go
@@ -344,6 +344,22 @@ func TestNewNetHTTPHandlerFuncRequests(t *testing.T) {
 				}
 			},
 		},
+		{
+			"http request path is normalized",
+			func() *http.Request {
+				req, _ := http.NewRequestWithContext(context.Background(), http.MethodGet, "https://localhost:8080//foo//bar/123/?hello=world", nil)
+				return req
+			},
+			func(t *testing.T) func(ctx *fasthttp.RequestCtx) {
+				return func(ctx *fasthttp.RequestCtx) {
+					assert.Equal(t, "/foo/bar/123/", string(ctx.Request.URI().Path()))
+					assert.Equal(t, "localhost:8080", string(ctx.Request.URI().Host()))
+					assert.Equal(t, "https", string(ctx.Request.URI().Scheme()))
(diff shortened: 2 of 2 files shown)


ItalyPaleAle:
Would this cause any issue with using service invocation?

JoshVanL:
> Would this cause any issue with using service invocation?

@ItalyPaleAle can you expand a bit on what you mean by this?

ItalyPaleAle:
> > Would this cause any issue with using service invocation?
> 
> @ItalyPaleAle can you expand a bit on what you mean by this?

If a user invokes Dapr with a path that doesn't match any Dapr API, we assume they're doing service invocation. For example:

```
curl -H 'dapr-app-id: checkout' 'http://localhost:3602/checkout/100' -X POST
```

Do you think this could cause any issue if we "normalize" the URL?

JoshVanL:
> > > Would this cause any issue with using service invocation?
> > 
> > 
> > @ItalyPaleAle can you expand a bit on what you mean by this?
> 
> If a user invokes Dapr with a path that doesn't match any Dapr API, we assume they're doing service invocation. For example:
> 
> ```
> curl -H 'dapr-app-id: checkout' 'http://localhost:3602/checkout/100' -X POST
> ```
> 
> Do you think this could cause any issue if we "normalize" the URL?

I don't expect so. By the time the request has got to the invoke handler (by the catch all `{method:*}` matcher), the path is normalized. The problem here is that the path is not normalized _before_ routing, causes some paths to match to that catch all invoke route when they actually belong to another.

ItalyPaleAle (approved):

artursouza:
Please, move this change to release-1.10 and we can add to 1.10.7 release with proper release notes.

artursouza:
We have a PR merged in release-1.10 already for this, which means it will be merged into master. Closing this PR.

PR #6340, opened 2023-05-12
Integration Tests: placement

JoshVanL:
Based on #6327 

Modulise runnable binaries into a `process` package.

Add support for building and running the placement binary, built from current source.

Squat free ports until the binary is executed to ensure the best chance that the port if free when used.

Add healthz and port test cases for the placement binary.

Use the test case scenario package name as part of Go test name.


artursouza (dismissed):

tests/integration/framework/process/daprd/daprd.go
```diff
@@ -0,0 +1,145 @@
...
+				return
+			}
+			conn.Close()
+		}
+	}()
+
+	fp := freeport.New(t, 6)
+	opts := options{
+		appID:            uid.String(),
+		appPort:          appListener.Addr().(*net.TCPAddr).Port,
```
ItalyPaleAle, line 93: Why are you using this method for `appPort`, but using `freeport` for all other ports?

tests/integration/suite/healthz/placement.go
```diff
@@ -0,0 +1,68 @@
...
+			return false
+		}
+		require.NoError(t, conn.Close())
+		return true
+	}, time.Second*5, 100*time.Millisecond, "healthz port %d not ready", d.proc.HealthzPort)
+
+	reqURL := fmt.Sprintf("http://127.0.0.1:%d/healthz", d.proc.HealthzPort)
+
+	assert.Eventually(t, func() bool {
+		req, err := http.NewRequestWithContext(ctx, http.MethodGet, reqURL, nil)
```
ItalyPaleAle, line 61: Context with a timeout please, since using the default client

ItalyPaleAle (commented):
See comments

ItalyPaleAle (changes requested):
See comments

tests/integration/framework/process/daprd/daprd.go
JoshVanL: Because this is a TCP listener we are instantiating in the test. The others are released and given to the dapr process to use.

ItalyPaleAle: Ok, makes sense now!

tests/integration/suite/healthz/placement.go
JoshVanL: The context is scoped to the test case which has a timeout. It don't see the need to wrap it again with another timeout.

JoshVanL:
@ItalyPaleAle replied

tests/integration/suite/healthz/placement.go
ItalyPaleAle: I responded to Shubham here too: https://github.com/dapr/dapr/pull/6524#discussion_r1232501908

The timeout is the timeout for the entire test. If a request hanged for any reason, it would take up the entire timeout for the test, and it would never retry in the `Eventually` block.

JoshVanL: Is this not the right behavior? I want the test to fail if the connection doesn't work. I'm retrying here only to allow placement to return with a successful healthy=false response, and then eventually a healthy=true response.

ItalyPaleAle: Ok then if this the behavior you want, sounds good :)

ItalyPaleAle (approved):

artursouza (approved):

PR #6342, opened 2023-05-12
State store: make BulkSet/BulkDelete not transactional

ItalyPaleAle:
This is a follow-up from #6186 that addresses the requested change of making BulkSet/BulkDelete operations in state stores not use transactions, and instead revert to perform operations one-by-one.

It is needed also to make sure that the correct HTTP status codes are returned to users when state operations fail with etag errors in bulk operations, per: https://github.com/dapr/dapr/pull/6186#issuecomment-1539262923 (which is currently a P0 blocker  for #6301 )

The main concern behind changing to using transactions in #6186 was that the previous way of running bulk operations did not play well with resiliency. For example, assume that the user wanted to set 3 keys: A, B, and C; all of them have an etag. Previously, operations were executed in sequence and errors would have caused to abort the sequence. If, for example, B failed with an etag error and resiliency was enabled, Dapr would have retried all operations, including A which would have now failed with an etag failure (because A was updated).

This PR changes how the bulk operations are performed:

1. Operations are now performed in parallel and not in sequence. There is support for limiting concurrency, although that's not exposed to end users right now as it requires changes to user-facing APIs which I thought not appropriate this late in the endgame. However, exposing this should be relatively easy.
2. When an operation fails, resiliency will retry only failed operations that are retriable. Operations that fail because of an etag error are not retried.

It fixes the error reported in https://github.com/dapr/dapr/pull/6186#issuecomment-1539262923 so the correct status codes are returned. New (unit) tests have been added to make sure that status codes will not change again, even for bulk operations, for both HTTP and GRPC.

Changes are to both bulk set and bulk delete operations.

## Future plans

There are a few things that are not included in this PR which could (should?) be done in the future:

- When a bulk set/delete operation fails, we should probably return details on what operation failed specifically. Right now we just return an internal error if at least one operation failed with any error that's not an etag error. Otherwise, we return a conflict error if at least one operation failed with an etag error. However, users have no way to know which operation failed.
- It also occurred to me that Dapr doesn't implement a Bulk Delete API in HTTP. It does exist in GRPC, but not in HTTP.
- As mentioned, bulk set operations at the moment are all executed in parallel. The internal APIs have an option to limit concurrency, but that's not exposed to end users at the moment.


artursouza (commented):
Please, add an E2E test to reproduce the original issue.

ItalyPaleAle:
Added E2E tests (had to be created _ex-novo_ because there were no E2E tests for state store etags). PR should be ready.

Contrib PR will need to be merged before this however.

go.mod
```diff
@@ -425,3 +425,6 @@ replace (
 //
 // Then, run `make modtidy` in this repository.
 // This ensures that go.mod and go.sum are up-to-date.
+
+// TODO: REMOVE WHEN PR IN COMPONENTS-CONTRIB IS MERGED
+replace github.com/dapr/components-contrib => github.com/dapr/components-contrib v1.11.0-rc.2.0.20230516184520-41303a9b0c00
```
ItalyPaleAle, line 430: Leaving this comment to remember to remove this

berndverst:
Sorry I pushed a bad contrib tag. I will repush as rc4 and update this PR

artursouza (changes requested):

.github/workflows/dapr-test-sdk.yml
```diff
@@ -84,7 +84,11 @@ jobs:
 
             Commit ref: ${{ env.CHECKOUT_REF }}
 
-      - uses: actions/checkout@v3
+      - name: Check out code
```
ItalyPaleAle, line 87: The SDK tests were failing because it was cloning master rather than the PR. This should fix it.

.github/workflows/dapr-test-sdk.yml
```diff
@@ -72,7 +72,7 @@ jobs:
         if: env.PR_NUMBER != ''
         uses: artursouza/sticky-pull-request-comment@v2.2.0
         with:
-          header: ${{ github.run_id }}
+          header: ${{ github.run_id }}-python
```
ItalyPaleAle, line 75: This allows having a separate comment for python and java, so updates should work now (hopefully)

.github/workflows/dapr-test-sdk.yml
```diff
@@ -19,8 +19,8 @@ name: dapr-test-sdk
 on:
   # Run every 12 hours on weekdays, and every 24 hours on weekends.
   schedule:
-    - cron: "0 */12 * * 1-5"
-    - cron: "30 0 * * 0,6"
+    - cron: "16 */11 * * 1-5"
```
ItalyPaleAle, line 22: Changed to not be on top of the hour because that's when Actions is busier :)

ItalyPaleAle:
SDK tests are still expected to fail until the updated workflows are merged into master.

I have confirmed the Python SDK examples pass on my machine against this PR.

artursouza (approved):

PR #6343, opened 2023-05-12
raft: fix index out of range

AdamKorcz:
# Description

Fix index out of range in case log.Data has too little data.

If log.Data has less than 2 entries, then several IoR can happen on these lines:

https://github.com/dapr/dapr/blob/770d4e51604f1264d8bb25cedf16ea9f77539394/pkg/placement/raft/fsm.go#L156C35-L163

Found by OSS-Fuzz: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=58799

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing

### DIFF
--- a/pkg/placement/raft/fsm.go
+++ b/pkg/placement/raft/fsm.go
@@ -154,7 +154,7 @@ func (c *FSM) Apply(log *raft.Log) interface{} {
 	}
 
 	if len(log.Data) < 2 {
-		logging.Infof("too short log data")
+		logging.Warnf("too short log data in raft logs")
 		return false
 	}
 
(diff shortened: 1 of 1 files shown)


ItalyPaleAle (dismissed):
Makes sense

Has this caused any crash in a live system? Just trying to understand how severe this is

AdamKorcz:
> Makes sense
> 
> Has this caused any crash in a live system? Just trying to understand how severe this is

Not for me.

pkg/placement/raft/fsm.go
```diff
@@ -153,6 +153,11 @@ func (c *FSM) Apply(log *raft.Log) interface{} {
 		return false
 	}
 
+	if len(log.Data) < 2 {
+		logging.Infof("too short log data")
```
daixiang0, line 157: How about print the log data here, may be helpful.

artursouza:
@yaron2 Please, merge PR once validation completes.

Issue #4012, opened 2021-11-04
mTLS settings from config.yaml are not used for Dapr instances


ionut-balazs-lateral:
## Expected Behavior

When configuring dapr instance yaml for mTLS, I expect mTLS to be configured at runtime.
Also, it would help if the configuration would accept a _sentryAddress_ property in the _mtls_ section.

## Actual Behavior

Updates to _config.yaml_ doesn't seem to affect the Dapr instance runtime, _mTLS_ is still disabled. The only way it works now is by supplying the flags _--enable-mtls_ and _--sentry-address_.

## Steps to Reproduce the Problem

1. Dapr 1.4.3
2. I am using _daprd_ directly to start the instance, not the CLI (not sure if this is relevant)
3. Configure mtls in config.yaml
4. Start Dapr instance
5. Monitor logs for the mTLS entry
6. sentry-address can only be set via the CLI, it would be nice to be able to do it via config

Taction:
The codebase has been changed/refactored a lot. for now, if mtls is enabled, the sentry address is needed by the security provider for certificate rotation. In k8s mode connecting to the operator to get global config needs the cert from the security provider. Supporting this needs to parse global config earlier than init the security provider in standalone mode, or change the security provider to expose setting sentry address and get workload cert then.

It would take me some more time to catch up with the code. I am afraid I don't have enough time recently to get this done, unassign myself so anyone who is interested in this can proceed on.

yaron2:
Changes here arre addressed.

PR #3951, opened 2021-11-25
Fix use mTLS settings from config.yaml when flag not defined

Taction:
# Description

Add `sentryAddress` for mtls config. Use enableMTLS and sentryAddress from config.yaml when not supplying  from flags.

## Issue reference

Please reference the issue this PR will close: dapr/dapr#4012

## Checklist

Please make sure you've  completed the relevant tasks for this PR, out of the following list:

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Unit tests passing
* [ ] End-to-end tests passing
* [ ] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_
* [ ] Specification has been updated / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_
* [ ] Provided sample for the feature / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_

PR #6347, opened 2023-05-15
Fix use mTLS settings from config.yaml when flag not defined

Taction:
# Description

Add sentryAddress for mtls config. Use enableMTLS and sentryAddress from config.yaml when not supplying from flags.

Original code pr #3951

## Issue reference

Please reference the issue this PR will close: #4012

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests
* [x] Unit tests passing


cmd/daprd/main.go
```diff
@@ -43,7 +45,7 @@ func main() {
 	// set GOMAXPROCS
 	_, _ = maxprocs.Set()
 
-	rt, err := runtime.FromFlags()
+	rt, err := runtime.FromFlags(os.Args[1:])
```
JoshVanL, line 48: Why are we passing in arguments in this way?

JoshVanL (changes requested):

pkg/runtime/cli.go
```diff
@@ -92,7 +93,8 @@ func FromFlags() (*DaprRuntime, error) {
 	metricsExporter.Options().AttachCmdFlags(flag.StringVar, flag.BoolVar)
 
 	// Finally parse the CLI flags!
-	flag.Parse()
+	// Ignore errors; CommandLine is set for ExitOnError.
+	_ = flag.CommandLine.Parse(args)
```
ItalyPaleAle, line 97: Can you move this to a separate PR? I can see how this could be useful but it's probably worth a separate, deeper conversation.

pkg/runtime/cli.go
```diff
@@ -295,6 +297,64 @@ func FromFlags() (*DaprRuntime, error) {
...
+
+	// Config and resiliency need the operator client
+	var operatorClient operatorV1.OperatorClient
+	if *mode == string(modes.KubernetesMode) {
+		log.Info("Initializing the operator client")
+		certChain, err = security.GetCertChain()
+		if err != nil {
+			return nil, err
+		}
+		client, conn, clientErr := client.GetOperatorClient(context.TODO(), *controlPlaneAddress, security.TLSServerName, certChain)
```
ItalyPaleAle, line 310: Shouldn't this context have a timeout - even a long one like 1 minute

pkg/runtime/cli.go
```diff
@@ -425,6 +442,7 @@ func FromFlags() (*DaprRuntime, error) {
 	}
 	log.Info("Resiliency configuration loaded")
 
+	var accessControlList *daprGlobalConfig.AccessControlList
 	accessControlList, err = acl.ParseAccessControlSpec(
```
ItalyPaleAle, line 446:
```suggestion
	accessControlList, err := acl.ParseAccessControlSpec(
```

ItalyPaleAle (changes requested):

cmd/daprd/main.go
Taction: This makes it easier to write tests for `runtime.FromFlags`. As @ItalyPaleAle suggested, I'll move it to a separate pr.

pkg/runtime/cli.go
Taction: Timeout has been set to 30 seconds inside the `GetOperatorClient` function to avoid blocking the app. So it is safe to leave context todo here, I'll add a comment for this.

ItalyPaleAle: In that case please use `context.Background()` or any parent context that may be defined in the code.

`context.TODO()` is meant to be a ... TODO to be fixed in the future.

pkg/runtime/cli.go
Taction: Thanks for your review, done.

pkg/runtime/cli.go
Taction: Sure, I'll move it to a separate pr.

pkg/runtime/cli.go
JoshVanL: Ok to leave as TODO?

this will make it easier to pick up when we finish daprd graceful shutdown

Taction: I am ok with both `context.Background()` and `context.TODO()`, since the original code uses `context.TODO()` and with JoshVanL's comment, I am inclined to leave it as `context.TODO()` here.

pkg/runtime/cli.go
Taction: Moved to #6372

ItalyPaleAle (approved):

PR #6350, opened 2023-05-15
Add `reinvocationPolicy: IfNeeded` to Dapr Sidecar injector

berndverst:
# Description

This PR updates the Dapr sidecar injector to use the core Kubernetes feature `reinvocationPolicy` to retrigger the sidecar injector in case another mutating admissions webhook modified the deployment.

This can reduce another source or flakiness where the sidecar was not injector because of competition with another sidecar injector.

While no other admission webhook injector should remove an already injected Dapr sidecar - it is possible. In that case, Kubernetes would retrigger the Dapr sidecar injector to inject the Dapr Sidecar.

See for reference:
https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#reinvocation-policy


ItalyPaleAle (approved):

yaron2:
Good for 1.11.

Issue #6349, opened 2023-05-15
Configuration to increase the max response size above the 4 MB limit


cyRawlings:
## In what area(s)?

/area runtime

## Describe the feature

Java gRPC calls between services can increase the size of max response message size by configuring the ManagedChannel.
However, if these requests are sent via dapr sidecars the channel configuration is ignored and the limit cannot be raised above the 4 MB limit.

The following response was received:
```
   io.grpc.StatusRuntimeException: RESOURCE_EXHAUSTED: gRPC message exceeds maximum size 4194304: 4603681
        at io.grpc.stub.ClientCalls.toStatusRuntimeException(ClientCalls.java:271)
        at io.grpc.stub.ClientCalls.getUnchecked(ClientCalls.java:252)
        at io.grpc.stub.ClientCalls.blockingUnaryCall(ClientCalls.java:165)
```

### Test setup

A python server and a java client.
Both are deployed locally with a dapr sidecar via docker compose.

The Java used the following channel:
```
ManagedChannel channel = ManagedChannelBuilder.forAddress(HOST, Integer.valueOf(PORT))
.maxInboundMessageSize(1000000000)
.intercept(interceptor)
.usePlaintext()
.build();
```

`maxInboundMessageSize()` increases the size of the response message if the PORT is set to the app_port of the python service, but not if the dapr sidecar is used.

The dapr sidecars used were `daprio/daprd:1.9.5 ` (I don't see any changes to indicate limits have increased in more recent images).
We have also been using version 1.10.4 in our azure deployments.

Java grpc packages: version 1.51.1
(io.grpc.grpc-core, io.grpc.grpc-stub, io.grpc.grpc-protobuf, io.grpc.grpc-netty-shaded)

Python service package:
grpcio==1.51.1

Note:
- We have tried launching the dapr sidecars with `--dapr-http-max-request-size` which some sources indicated might also increase the max size of the gRPC response, however this did not affect the outcome.
- We are not using the dapr java-sdk. We have not seen additional configuration options in the SDK, but maybe we overlooked something here.

## Release Note

RELEASE NOTE: **ADD** New feature in Dapr.

yaron2:
The max-request-size setting does not apply to gRPC proxying. We'll add it to 1.12.

yaron2:
[...]

PR #6353, opened 2023-05-16
Passing gRPC max msg size to proxy client stream

Flern:
# Description

Updated the `handler.go` file to use a passed-in argument of `maxMessageBodySizeMB` that is provided from the `runtime.go` instantiation of the proxy down through the `TransparentHandler` by using the configuration item `--dapr-http-max-request-size` parameter.  If there is no parameter configured, a zero value is propagated, so any value of zero (or less) is ignored and defaults are used.

## Issue reference

https://github.com/dapr/dapr/issues/6349

Please reference the issue this PR will close: #_6349_

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Unit tests passing

### DIFF
--- a/tests/e2e/service_invocation/service_invocation_test.go
+++ b/tests/e2e/service_invocation/service_invocation_test.go
@@ -370,6 +370,12 @@ var grpcProxyTests = []struct {
 		"",
 		"success",
 	},
+	{
+		"Test grpc proxy",
+		"grpcproxyclient",
+		"maxsize",
+		"success",
+	},
 }
 
 func TestServiceInvocation(t *testing.T) {
--- a/tests/e2e/service_invocation/service_invocation_test.go
+++ b/tests/e2e/service_invocation/service_invocation_test.go
@@ -146,12 +146,13 @@ func TestMain(m *testing.M) {
 			AppProtocol:    "grpc",
 		},
 		{
-			AppName:        "grpcproxyclient",
-			DaprEnabled:    true,
-			ImageName:      "e2e-service_invocation_grpc_proxy_client",
-			Replicas:       1,
-			IngressEnabled: true,
-			MetricsEnabled: true,
+			AppName:          "grpcproxyclient",
+			DaprEnabled:      true,
+			ImageName:        "e2e-service_invocation_grpc_proxy_client",
+			Replicas:         1,
+			IngressEnabled:   true,
+			MetricsEnabled:   true,
+			MaxRequestSizeMB: 6,
 		},
 		{
 			AppName:        "grpcproxyserverexternal",
@@ -171,6 +172,7 @@ func TestMain(m *testing.M) {
 			AppProtocol:       "grpc",
 			AppPort:           50051,
 			AppChannelAddress: "grpcproxyserver-app",
+			MaxRequestSizeMB:  6,
 		},
 		{
 			AppName:        "serviceinvocation-callee-external",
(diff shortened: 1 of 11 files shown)


ItalyPaleAle:
Overall I think this is fine. 2 comments:

1. Using `dapr-http-max-request-size` doesn't feel right for something that is related to gRPC. Should we have a separate CLI flag for this maybe? Also, whatever flag should also apply to our gRPC server too perhaps (not just for gRPC proxying).
2. Could you please update one of the E2E tests, such as the gRPC proxying one, to validate that with this flag we can send messages larger than 4MB?

yaron2:
> 1. Using `dapr-http-max-request-size` doesn't feel right for something that is related to gRPC. Should we have a separate CLI flag for this maybe? Also, whatever flag should also apply to our gRPC server too perhaps (not just for gRPC proxying).

We should reuse `dapr-http-max-request-size` here as this is already applied to our gRPC servers.

ItalyPaleAle:
> > 1. Using `dapr-http-max-request-size` doesn't feel right for something that is related to gRPC. Should we have a separate CLI flag for this maybe? Also, whatever flag should also apply to our gRPC server too perhaps (not just for gRPC proxying).
> 
> We should reuse `dapr-http-max-request-size` here as this is already applied to our gRPC servers.

I see. Can we deprecate it and create `dapr-max-request-size` and use the old one as alias?

yaron2:
> > > 1. Using `dapr-http-max-request-size` doesn't feel right for something that is related to gRPC. Should we have a separate CLI flag for this maybe? Also, whatever flag should also apply to our gRPC server too perhaps (not just for gRPC proxying).
> > 
> > 
> > We should reuse `dapr-http-max-request-size` here as this is already applied to our gRPC servers.
> 
> I see. Can we deprecate it and create `dapr-max-request-size` and use the old one as alias?

I don't see a reason to do that but we can consider a contribution (separate from this PR) in the future.

Flern:
@yaron2 - update pushed.  I also missed some significant infrastructure updates needed to allow the `dapr-http-max-request-size` to be included in service configurations for e2e deployments.  Having trouble running e2e locally, though, to verify that this will work.

tests/e2e/service_invocation/service_invocation_test.go
```diff
@@ -146,12 +146,13 @@ func TestMain(m *testing.M) {
...
-			Replicas:       1,
-			IngressEnabled: true,
-			MetricsEnabled: true,
+			AppName:          "grpcproxyclient",
+			DaprEnabled:      true,
+			ImageName:        "e2e-service_invocation_grpc_proxy_client",
+			Replicas:         1,
+			IngressEnabled:   true,
+			MetricsEnabled:   true,
+			MaxRequestSizeMB: 6,
```
yaron2, line 155: You need to set `MaxRequestSizeMB` on `grpcproxyserver` as well. The YAML files under `tests/apps/` are not applied during the tests.

yaron2 (changes requested):

yaron2:
@Flern you need to add gRPC options to the proxy server app [here](https://github.com/dapr/dapr/blob/master/tests/apps/service_invocation_grpc_proxy_server/app.go#L45) for `MaxRecvMsgSize` and `MaxSendMsgSize`.

yaron2 (approved):

PR #6355, opened 2023-05-16
Fix injector to handle empty app-channel-address.

artursouza:
Fix injector to handle empty app-channel-address.
Allow sidecar image override on E2E tests.

# Description

## Issue reference

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing

### DIFF
--- a/.github/workflows/kind-e2e.yaml
+++ b/.github/workflows/kind-e2e.yaml
@@ -50,6 +50,11 @@ jobs:
       DAPR_REGISTRY: localhost:5000/dapr
       DAPR_TAG: dev
       DAPR_NAMESPACE: dapr-tests
+      # Useful for upgrade/downgrade/compatibility tests
+      # 1.9.6 does not work on newest Windows version for E2E tests
+      # Add this env var to Windows runs too once we have an old version that works with latest win version
+      # TODO: Make this auto-populated based on GitHub's releases.
+      DAPR_TEST_N_MINUS_2_IMAGE: "ghcr.io/dapr/daprd:1.9.6"
       # Container registry where to cache e2e test images
       DAPR_CACHE_REGISTRY: "dapre2eacr.azurecr.io"
       PULL_POLICY: IfNotPresent
--- a/pkg/injector/sidecar/container_test.go
+++ b/pkg/injector/sidecar/container_test.go
@@ -172,7 +172,6 @@ func TestGetSidecarContainer(t *testing.T) {
 			"--dapr-http-read-buffer-size", "-1",
 			"--dapr-graceful-shutdown-seconds", "-1",
 			"--disable-builtin-k8s-secret-store=false",
-			"--app-channel-address", "",
 			"--log-as-json",
 			"--enable-mtls",
 		}
@@ -247,7 +246,6 @@ func TestGetSidecarContainer(t *testing.T) {
 			"--dapr-http-read-buffer-size", "-1",
 			"--dapr-graceful-shutdown-seconds", "-1",
 			"--disable-builtin-k8s-secret-store=false",
-			"--app-channel-address", "",
 			"--log-as-json",
 			"--enable-mtls",
 		}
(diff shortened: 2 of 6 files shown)


.github/workflows/kind-e2e.yaml
```diff
@@ -50,6 +50,10 @@ jobs:
       DAPR_REGISTRY: localhost:5000/dapr
       DAPR_TAG: dev
       DAPR_NAMESPACE: dapr-tests
+      # Useful for upgrade/downgrade/compatibility tests
+      # 1.9.6 does not work on newest Windows version for E2E tests
+      # Add this env var to Windows runs too once we have an old version that works with latest win version
+      DAPR_TEST_OLD_IMAGE: "ghcr.io/dapr/daprd:1.9.6"
```
ItalyPaleAle, line 56: Thoughts on calling this something like `DAPR_TEST_N_MINUS_2_IMAGE` and then we can also have `DAPR_TEST_N_MINUS_1_IMAGE`?

tests/e2e/hellodapr/hellodapr_test.go
```diff
@@ -56,6 +57,7 @@ func TestMain(m *testing.M) {
 			AppName:           "hellobluedapr",
 			DaprEnabled:       true,
 			ImageName:         "e2e-hellodapr",
+			SidecarImage:      runtime_utils.GetEnvOrElse("DAPR_TEST_OLD_IMAGE", ""), // old image to avoid regression in injector.
```
ItalyPaleAle, line 60: Can you add a check at the beginning of the test to make sure that `DAPR_TEST_OLD_IMAGE` is defined?

.github/workflows/kind-e2e.yaml
```diff
@@ -50,6 +50,10 @@ jobs:
       DAPR_REGISTRY: localhost:5000/dapr
       DAPR_TAG: dev
       DAPR_NAMESPACE: dapr-tests
+      # Useful for upgrade/downgrade/compatibility tests
```
ItalyPaleAle, line 53: Please update dapr-test.yaml too (for AKS)

ItalyPaleAle (changes requested):

tests/e2e/hellodapr/hellodapr_test.go
artursouza: I can't because it does not work on Windows. Remember that 1.9.6 or 1.10.7 don't have the latest windows images.

.github/workflows/kind-e2e.yaml
artursouza: I am not adding it there because it does not work on Windows' latest version since the old image does not have an image for that.

ItalyPaleAle: How about adding it for the Linux tests only?

tests/e2e/hellodapr/hellodapr_test.go
ItalyPaleAle: Makes sense

ItalyPaleAle (approved):

.github/workflows/kind-e2e.yaml
artursouza: This does the same. See the comments next to it. The goal is to add it to work with both once we have the new os build available in the old images.

pkg/injector/sidecar/container_test.go
```diff
@@ -172,7 +172,6 @@ func TestGetSidecarContainer(t *testing.T) {
 			"--dapr-http-read-buffer-size", "-1",
 			"--dapr-graceful-shutdown-seconds", "-1",
 			"--disable-builtin-k8s-secret-store=false",
-			"--app-channel-address", "",
```
daixiang0, line 175: Look around the code, we do not have a test for no-empty app-channel-address.

artursouza: Let me merge this and will add the new test scenario. It is blocking another PR that requires this fix to have E2E tests with old images.

artursouza:
Sorry, I did not notice Alessandro's review was dismissed.

PR #6360, opened 2023-05-16
Fix: Bulk pubsub subscriber panics when tracing is disabled

### LINKED ISSUE #6099: Bulk pubsub subscriber panics when tracing is disabled
## In what area(s)?

/area runtime

## What version of Dapr?

> 1.10.x
> master

## Description

Using bulk pubsub to subscribe to a topic, Daprd panics if tracing is disabled.

The bug is in these lines (and in another similar spot in the same file):

https://github.com/dapr/dapr/blob/585da2feb90d5ddbc5a94efca5ba0702a9e43808/pkg/runtime/bulk_subscriber.go#L587-L588

When tracing is disabled, `diag.StartInternalCallbackSpan` returns `nil` for `span`, so the next line panics.

## Release Note

RELEASE NOTE: **FIX** Bulk pubsub subscriber panics when tracing is disabled

ItalyPaleAle:
Fixes #6099 by backporting changes from #6101

### DIFF
--- a/pkg/runtime/bulk_subscriber.go
+++ b/pkg/runtime/bulk_subscriber.go
@@ -368,8 +368,10 @@ func (a *DaprRuntime) publishBulkMessageHTTP(ctx context.Context, bulkSubCallDat
 			sc, _ := diag.SpanContextFromW3CString(traceID)
 			var span trace.Span
 			ctx, span = diag.StartInternalCallbackSpan(ctx, "pubsub/"+psm.topic, sc, a.globalConfig.Spec.TracingSpec)
-			spans[n] = span
-			n++
+			if span != nil {
+				spans[n] = span
+				n++
+			}
 		}
 	}
 	spans = spans[:n]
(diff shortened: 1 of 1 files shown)


pkg/runtime/bulk_subscriber.go
```diff
@@ -561,8 +561,12 @@ func (a *DaprRuntime) publishBulkMessageGRPC(ctx context.Context, bulkSubCallDat
 		items[i] = item
 	}
 
+	uuidObj, err := uuid.NewRandom()
+	if err != nil {
+		return fmt.Errorf("failed to generate UUID: %w", err)
+	}
 	envelope := &runtimev1pb.TopicEventBulkRequest{
-		Id:         uuid.New().String(),
+		Id:         uuidObj.String(),
```
ItalyPaleAle, line 569: This is not related to tracing but was fixed in #6101 too. In short, `uuid.New()` can panic (if there's not enough entropy in the kernel), while `uuid.NewRandom()` allows retrieving an error.

yaron2 (approved):

PR #6361, opened 2023-05-16
Fix /ok-to-test not working for some users

ItalyPaleAle:
#6304 broke `/ok-to-test` for some users (those who had an uppercase letter in the `owners` array)


tests/perf/configuration/test.js
ItalyPaleAle, line 1: Prettier did this

artursouza (approved):

PR #6367, opened 2023-05-18
Update Dapr Bot Action and respond to users that can't invoke commands

ItalyPaleAle:
Two small fixes to the Dapr Bot Action:

1. Update the Action to v6. The current v4 is using an older Node.js version and those are deprecated by GitHub
2. Changed the bot so if a user that is not in the allowlist tries to use a command they're not allowed to (like /ok-to-test), the bot will respond with a message. This is to provide feedback to users - for example, when the bot stopped working for me the other day due to a bug, I was confused as to what was going on, and debugging these issues for contributors is hard.


.github/scripts/dapr_bot.js
```diff
@@ -157,7 +158,7 @@ async function handleIssueLabeled({ github, context }) {
     // Only authorized users can add labels to issues.
     if (label == 'docs-needed') {
         // Open a new issue
-        await github.issues.create({
+        await github.rest.issues.create({
```
JoshVanL, line 161: What is the significance of `github.issues` -> `github.rest.issues`, or is this the new path in v6?

JoshVanL (changes requested):

.github/scripts/dapr_bot.js
ItalyPaleAle: That’s how v5+ work. They changed the APIs.

tests/perf/pubsub_bulk_publish_http/test.js
ItalyPaleAle, line 1: Changes to this file are due to `make prettier`

mukundansundar (approved):

PR #6372, opened 2023-05-19
make FromFlags function testable

Taction:
# Description

Runtime FromFlags function prases flags and loads configurations, and in some cases, it may merge the configurations from both sides. Passing the args to this function makes it easier to write unit tests.

## Issue reference

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Unit tests passing


pkg/runtime/cli_test.go
```diff
@@ -44,3 +52,20 @@ func TestParsePlacementAddr(t *testing.T) {
 		})
 	}
 }
+
+func TestAppFlag(t *testing.T) {
+	// run test
```
daixiang0, line 57: Remove this useless comment.

pkg/runtime/cli_test.go
```diff
@@ -44,3 +52,20 @@ func TestParsePlacementAddr(t *testing.T) {
...
+	runtime, err := FromFlags([]string{"--app-id", testAppID, "--app-port", testAppPort, "--app-protocol", testAppProtocol})
+	assert.NoError(t, err)
+	assert.EqualValues(t, testAppID, runtime.runtimeConfig.ID)
+	assert.EqualValues(t, 80, runtime.runtimeConfig.ApplicationPort)
+	assert.EqualValues(t, testAppProtocol, runtime.runtimeConfig.ApplicationProtocol)
+}
+
+func TestStandaloneGlobalConfig(t *testing.T) {
+	// run test
```
daixiang0, line 66: Same as above.

pkg/runtime/cli.go
```diff
@@ -92,14 +92,15 @@ func FromFlags() (*DaprRuntime, error) {
 	metricsExporter.Options().AttachCmdFlags(flag.StringVar, flag.BoolVar)
 
 	// Finally parse the CLI flags!
-	flag.Parse()
+	// Ignore errors; CommandLine is set for ExitOnError.
```
daixiang0, line 95: In which case, cmdline raise an error but do not affect?

Taction: This is exactly the code and comment under the `flag.Parse()` function.  The comment is to show that if any error occurs, the code inside `CommandLine` will call exit so we do not need to handle errors here.

pkg/runtime/cli_test.go
Taction: Done.

pkg/runtime/cli_test.go
Taction: Done.

pkg/runtime/cli.go
```diff
@@ -92,14 +92,15 @@ func FromFlags() (*DaprRuntime, error) {
...
+	flag.CommandLine.Parse(args)
 
 	// flag.Parse() will always set a value to "enableAPILogging", and it will be false whether it's explicitly set to false or unset
 	// For this flag, we need the third state (unset) so we need to do a bit more work here to check if it's unset, then mark "enableAPILogging" as nil
 	// It's not the prettiest approach, but…
 	if !*enableAPILogging {
 		enableAPILogging = nil
-		for _, v := range os.Args {
+		for _, v := range args {
```
daixiang0, line 103: Here use all os.Args before, now use Args[1:], what cause the difference?

pkg/runtime/cli_test.go
```diff
@@ -44,3 +52,18 @@ func TestParsePlacementAddr(t *testing.T) {
...
+	assert.NoError(t, err)
+	assert.EqualValues(t, testAppID, runtime.runtimeConfig.ID)
+	assert.EqualValues(t, 80, runtime.runtimeConfig.ApplicationPort)
+	assert.EqualValues(t, testAppProtocol, runtime.runtimeConfig.ApplicationProtocol)
+}
+
+func TestStandaloneGlobalConfig(t *testing.T) {
+	runtime, err := FromFlags([]string{"--app-id", testAppID, "--mode", string(modes.StandaloneMode), "--config", "../config/testdata/metric_disabled.yaml"})
+	assert.NoError(t, err)
+	assert.EqualValues(t, testAppID, runtime.runtimeConfig.ID)
```
daixiang0, line 67: Please check all args.

pkg/runtime/cli.go
Taction: Args[0] is the name of the command that it is running, which is `dapr` here. The aim of this for loop is to find `--enable-api-logging` flag is presented or not, so it is safe to replace with args here.

Taction:
ping @ItalyPaleAle as you may have thoughts on this.

pkg/runtime/cli_test.go
Taction: Done. notes: `metrics-port` parsed as a local variable, so could not be checked.

ItalyPaleAle (approved):

ItalyPaleAle:
@Taction LGTM :)

cmd/daprd/main.go
```diff
@@ -43,7 +45,7 @@ func main() {
 	// set GOMAXPROCS
 	_, _ = maxprocs.Set()
 
-	rt, err := runtime.FromFlags()
+	rt, err := runtime.FromFlags(os.Args[1:])
```
mukundansundar, line 48: I think we can pass in args as is here, in the unlikely scenario that `daprd` is run directly, it will fail right now(master code), but won't it fail with index out of bounds in this code path?

daixiang0 (approved):
LGTM

PR #6374, opened 2023-05-19
Version Skew GitHub action test

JoshVanL:
PR adds a new workflow: version skew tests which run on push to master, or manually on a bot command.

1 job in the workflow will deploy the master build version of the control plane with the daprd sidecar set to the previous release version. The e2e tests of the previous release will be run against this deployment. The second job runs the same tests, but with the control plane on the previous release version, daprd on master.

The intention of this workflow is to verify that the current master branch of dapr is compatible to upgrading, according to the [upgrade path policy](https://docs.dapr.io/operations/support/support-release-policy/#upgrade-paths).

The previous release is automatically discovered from the GitHub releases page.

You can find the results to this workflow on my personal fork [here](https://github.com/JoshVanL/dapr/actions/runs/5034252901).

The bot command `/test-version-skew` will trigger the workflow on a PR. You can also optionally supply a specific version to test against as an argument to this command, e.g. `/test-version-skew 1.10.5`. Note that this version needs to be released and its artifacts be publicly available.

Example of this working [here](https://github.com/JoshVanL/dapr/pull/20#issuecomment-1556043761).

---

Another option was to consider running the master e2e tests, however this would result in a bunch of tests being marked to skip running on this workflow & it doesn't make much sense to expect master branch features/tests to be successful on older software versions.. we do expect all previous release tests to pass on master though.

---

There is also room to extend this job so that the control plane and/or the daprd sidecar is upgraded/downgraded "live", after deployment, before running the e2e tests.

/cc @artursouza


.github/workflows/upgrade-downgrade-e2e.yaml
```diff
@@ -0,0 +1,214 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+name: Upgrade/Downgrade E2E tests on KinD
+
+on:
+  # Run every 12 hours on weekdays, and every 24 hours on weekends.
+  schedule:
```
ItalyPaleAle, line 18: Thoughts on running this as part of a bot command? Could be the same "/ok-to-test" or a separate command specific to this.

I feel this workflow is more useful during PR review, and after a PR is merged, it shouldn't over time.

.github/workflows/upgrade-downgrade-e2e.yaml
```diff
@@ -0,0 +1,214 @@
...
+    name: e2e-upgrade-downgrade
+    runs-on: ubuntu-latest
+    env:
+      REGISTRY_PORT: 5000
+      REGISTRY_NAME: kind-registry
+      DAPR_REGISTRY: localhost:5000/dapr
+      DAPR_TAG: dev
+      DAPR_NAMESPACE: dapr-tests
+      DAPR_CACHE_REGISTRY: "dapre2eacr.azurecr.io"
+      PULL_POLICY: IfNotPresent
```
ItalyPaleAle, line 47:
```suggestion
      PULL_POLICY: Always
```

.github/workflows/upgrade-downgrade-e2e.yaml
```diff
@@ -0,0 +1,214 @@
...
+      matrix:
+        mode:
+        - control-plane-master
+        - dapr-sidecar-master
+    steps:
+    - name: Check out code
+      uses: actions/checkout@v3
+    - name: Get last version of dapr
+      run: |
+        echo "DAPR_LATEST_VERSION=$(curl -s https://api.github.com/repos/dapr/dapr/releases/latest | jq -r '.tag_name' | cut -c 2-)" >> $GITHUB_ENV
```
ItalyPaleAle, line 59: For `workflow_dispatch` could it be possible to make this a variable that is passed as input? So manually we can test specific upgrade paths.

.github/workflows/upgrade-downgrade-e2e.yaml
```diff
@@ -0,0 +1,214 @@
...
+    - name: Checkout last release repo
+      uses: actions/checkout@v3
+      with:
+        repository: dapr/dapr
+        path: latest-release
+        ref: v${{ env.DAPR_LATEST_VERSION }}
+    - name: Set up Go
+      id: setup-go
+      uses: actions/setup-go@v3
```
ItalyPaleAle, line 68:
```suggestion
      uses: actions/setup-go@v4
```

.github/workflows/upgrade-downgrade-e2e.yaml
```diff
@@ -0,0 +1,214 @@
...
+    - name: Setup local registry
+      run: |
+        # Run a registry.
+        docker run -d --restart=always \
+          -p $REGISTRY_PORT:$REGISTRY_PORT --name $REGISTRY_NAME registry:2
+        # Connect the registry to the KinD network.
+        docker network connect "kind" $REGISTRY_NAME
+
+    - name: Setup Helm
+      uses: azure/setup-helm@v1
```
ItalyPaleAle, line 128:
```suggestion
      uses: azure/setup-helm@v3
```

.github/workflows/upgrade-downgrade-e2e.yaml
```diff
@@ -0,0 +1,214 @@
...
+    - name: Build and push test apps
+      run: |
+        cd latest-release && make build-push-e2e-app-all
+
+    - name: Setup Dapr - control-plane master
+      if: matrix.mode == 'control-plane-master'
+      run: |
+        make setup-helm-init
+        make create-test-namespace
+        export ADDITIONAL_HELM_SET="dapr_operator.logLevel=debug,dapr_operator.watchInterval=20s,dapr_sidecar_injector.image.name=ghcr.io/dapr/daprd:${{env.DAPR_LATEST_VERSION}}"
```
ItalyPaleAle, line 147:
```suggestion
        export ADDITIONAL_HELM_SET="dapr_operator.logLevel=debug,dapr_sidecar_injector.image.name=ghcr.io/dapr/daprd:${{env.DAPR_LATEST_VERSION}}"
```

.github/workflows/upgrade-downgrade-e2e.yaml
```diff
@@ -0,0 +1,214 @@
...
+        make setup-test-env-zipkin
+        make setup-test-components
+
+    - name: Free up some diskspace
+      run: |
+        docker image prune -a -f
+
+    - name: Run tests
+      run: |
+        cd latest-release && make test-e2e-all
```
ItalyPaleAle, line 180: Do we need to run the entire test suite? Can we just run a subset of tests that are related to confirming the upgrade works?

.github/workflows/upgrade-downgrade-e2e.yaml
```diff
@@ -0,0 +1,214 @@
...
+    - name: Setup Dapr - control-plane previous release
+      if: matrix.mode == 'dapr-sidecar-master'
+      run: |
+        make setup-helm-init
+        make create-test-namespace
+        export DAPR_CLI_LATEST_VERSION=$(curl -s https://api.github.com/repos/dapr/cli/releases/latest | jq -r '.tag_name' | cut -c 2-)
+        wget -q https://raw.githubusercontent.com/dapr/cli/master/install/install.sh -O - | /bin/bash -s $DAPR_CLI_LATEST_VERSION
+        dapr uninstall --all
+        dapr init -k --runtime-version ${{ env.DAPR_LATEST_VERSION }} \
+          --set dapr_operator.logLevel=debug,dapr_operator.watchInterval=20s,dapr_sidecar_injector.image.name=ghcr.io/dapr/daprd:dev \
```
ItalyPaleAle, line 160:
```suggestion
          --set dapr_operator.logLevel=debug,dapr_sidecar_injector.image.name=ghcr.io/dapr/daprd:dev \
```

.github/workflows/upgrade-downgrade-e2e.yaml
JoshVanL: Do you mean running the job control plane master vs daprd master, or do you mean overriding `DAPR_LATEST_VERSION`? If the latter, I'm not sure it makes sense to override this, since the dapr project makes no guarantees about version skew above +1 so seems like wasted effort/distracting to allow that to be run.

.github/workflows/upgrade-downgrade-e2e.yaml
JoshVanL: I copied this from `kind-e2e.yaml`, but yeah, prob best to remove. Can add it back if we think it is appropriate for whatever reason.

.github/workflows/upgrade-downgrade-e2e.yaml
JoshVanL: I've thought about this and originally experimented with running a subset of the current master branch's test suite, adding Go test skips or some other method to run a subset to ensure all appropriate tests would pass. I came to the conclusion that 1. if we run a subset of tests, we will not be testing against the full feature set and will therefore miss incompatibilities, breaking dapr's version skew guarantee, and 2. that guarantee means we _should_ be passing _all_ e2e tests from the previous version, so I see no need not to run them all. It obviously increases test time etc, but not much considering we are already running the full suite 3 times at least once on every PR anyway.

.github/workflows/upgrade-downgrade-e2e.yaml
ItalyPaleAle: The runners are free to us (but not free for the environment 🌲).

My question is more on whether we think that things can change over time. The reason why AKS tests run on a schedule, for example, is to catch things that change due to K8s changing, or maybe catch errors we do not handle with external services, etc. In this case, do we think there's a chance that once a commit is merged and it doesn't alter upgradeability, could that change?

Alternatively, how about we run this when a commit is merged on master? (In addition to /ok-to-something)

.github/workflows/upgrade-downgrade-e2e.yaml
ItalyPaleAle: I meant being able to override DAPR_LATEST_VERSION.

Although we don't have guarantees beyond N-1, I can see us wanting to test edge cases in some situations. Or maybe test a different patch release.

.github/workflows/upgrade-downgrade-e2e.yaml
ItalyPaleAle: Ok, you convinced me :)

.github/workflows/upgrade-downgrade-e2e.yaml
JoshVanL: Gotcha, yeah testing a specific patch def makes sense. I'll update the workflow with this capability and add a comment about what should and shouldn't be tested against

.github/workflows/upgrade-downgrade-e2e.yaml
JoshVanL: Yeah, so the intention here is that this test probably isn't appropriate to run on every PR as a blocking check however, running as a schedule means that we would still get a signal on code changes which _do_ effect version skew compatibility- particularly when the code change that breaks skew is non-obvious during review.

Running on the merged master commit is a good candidate too- I'll change to that for now.

JoshVanL:
@daixiang0 the command won't be available until this is merged to master. You can try it out on my fork though  https://github.com/JoshVanL/dapr/actions/runs/5034252901

daixiang0 (approved):

PR #6384, opened 2023-05-22
Update devcontainer for 1.11

ItalyPaleAle:
1. Add `all_components` to the `go.buildTags` settings
   - Without this, users in VS Code / Codespaces will have issues reported by the Go language server, for example trying to use the debugger
2. Move `extensions` and `settings` into `customizations/vscode` per the updated Dev Container specs

Note: this needs to be merged into master and not release-1.11


daixiang0 (approved):
LGTM

PR #6400, opened 2023-05-24
Actor State TTL: Put feature being feature gate

JoshVanL:
Due to the fact that actor client SDKs cache state, when a client has a cold cache (like when restarting), client's would have to either cache data past their TTL time or always query the DB on gets. Until https://github.com/dapr/components-contrib/issues/2857 is implements, and this value passed to the actor SDK clients, we should actor state TTLs behind a feature preview flag.

PR adds the `ActorStateTTL` feature gate. Unless enabled, the actor transactional state operator API will return an error whenever an `Upsert` with a Metadata key `ttlInSeconds` is given. 

/cc @artursouza


ItalyPaleAle (changes requested):

ItalyPaleAle (approved):

artursouza:
Don't merge this. It needs to go to the release-1.11 branch first.

Issue #6418, opened 2023-05-29
Decoding slash in url


Skreet2k:
/area runtime

## Expected Behavior

dapr doesnt decode request before invoke.
`curl --header "dapr-app-id: appId" http://localhost:3500/api/v1/files/Icon/icons%2F24%2FHU.png
`

## Actual Behavior

I think dapr decode request and make it to 
`http://localhost:3500/api/v1/files/Icon/icons/24/HU.png
`

## Steps to Reproduce the Problem

After update from 1.10.4 to 1.10.7 all requests, with encoded slash return 404.

1.10.6 doesnt have issue.

## Release Note

RELEASE NOTE:

FIX decode url before invoke

berndverst:
Appears to be related to https://github.com/dapr/dapr/pull/6335

yaron2:
> Should result in the same call into the app as calling it without Dapr:
curl http://appId.svc:3000api/v1/files/Icon/icons%2F24%2FHU.png 

Agree with this. Dapr should not decode on behalf of the app.

JoshVanL:
Since fasthttp [normalizes slashes together with segment contents](https://github.com/valyala/fasthttp/blob/master/uri.go#LL575C3-L575C3), we need to duplicate that logic in dapr and disable normalization in fasthttp (assuming we still want to slash normalization to address https://github.com/dapr/dapr/issues/6324).

PR #6430, opened 2023-06-01
Daprd router normalize only seperator

JoshVanL:
Fixes https://github.com/dapr/dapr/issues/6418

Currently, daprd will decode incoming path segments before routing and forwarding on requests during service invocation. This breaks users attempting to encode data inside URL segments without that data interfering with routing.

In order to preserve normalization of URI separators (https://github.com/dapr/dapr/issues/6324), this PR copies across the normalization code from `valyala/fasthttp`, removes the segment normalization, and disables the normalization done in the `Request.URI()`. We use `EncodedPath` when setting the path in the URI to preserve segment encoding.

/cc @artursouza

### DIFF
--- a/tests/apps/service_invocation/app.go
+++ b/tests/apps/service_invocation/app.go
@@ -392,6 +392,12 @@ func appRouter() http.Handler {
 	router.HandleFunc("/badservicecalltesthttp", badServiceCallTestHTTP).Methods("POST")
 	router.HandleFunc("/badservicecalltestgrpc", badServiceCallTestGrpc).Methods("POST")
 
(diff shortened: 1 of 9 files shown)


ItalyPaleAle (changes requested):
Can you apply the fix in the 1.10 and 1.11 branches and not master? At this point master is very, very different: as of yesterday the server uses net/http (while 1.10 and 1.11 use fasthttp). Additionally, service invocation endpoints will not use the nethttpwrapper anymore.

(but the test would still be good to have in master too)

JoshVanL:
@ItalyPaleAle sure, I'll cherry pick these commits and open PRs to those branches linking back to this.

PR #6439, opened 2023-06-03
Fix wasm binding register

Taction:
# Description

wasm is a supported arch by go, according to go [build constraints](https://pkg.go.dev/cmd/go#hdr-Build_constraints) the file with _wasm.go suffix should only be compiled with `GOARCH=wasm` which is not actually the behavior we want.

So I renamed the file and added a check for the component that is registered when initializing a binding component.

## Issue reference

Please reference the issue this PR will close: #6440

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Unit tests passing


ItalyPaleAle (approved):

daixiang0 (approved):
LGTM

PR #6455, opened 2023-06-07
[Workflow] Add workflow generation to activity actor IDs

### LINKED ISSUE #6435: [Workflow] Calls to GetWorkflowAsync or TerminateWorkflowAsync don't return
## In what area(s)?

/area runtime

> /area operator

> /area placement

> /area docs

> /area test-and-release

## What version of Dapr?

dapr runtime: 1.11.0-rc.7
dotnet SDK: 1.11.0-rc01

## Expected Behavior

When calling either the `GetWorkflowAsync` or the `TerminateWorkflowAsync ` it is expected that they execute and immediately return.

## Actual Behavior

I regularly get into a situation where calls to either of those methods enter but never return.   There doesn't appear to be any indication in the daprd logs stating what is happening, from what I can see.

## Release Note

RELEASE NOTE:
[dapr-workflow-dapr.txt](https://github.com/dapr/dapr/files/11636756/dapr-workflow-dapr.txt)
[dapr-workflow-dotnet.txt](https://github.com/dapr/dapr/files/11636757/dapr-workflow-dotnet.txt)

cgillum:
# Description

This change ensures uniqueness for activity invocations across multiple workflow generations. This change will prevent actor deadlocks caused by multiple workflow generations that schedule concurrently executing activities with the same task ID.

NOTE: workflow "generations" are monotonically increasing integer values for a given workflow ID. If you create a workflow with ID = `abc123`, it will start with a generation of `1`. If you terminate and recreate a workflow with the same ID, its generation will be `2`.

This change is intended to be released as part of v1.11.

See the linked issue discussion for more details.

## Issue reference

Closes #6435

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing


clintsinger:
I would like some clarification on "concurrently executing".   The intention of a workflow ID is that there should only be one running at a time.  Is it my understanding that when a workflow is terminated it will continue to execute it's activities but won't return the values?   

What happens if one of those activities should make a change to a database under normal execution but shouldn't have based on the timing of the termination?   Will that activity run still and make the modification or will the workflow not continue with its execution for subsequent activities after termination?

cgillum:
> What happens if one of those activities should make a change to a database under normal execution but shouldn't have based on the timing of the termination? Will that activity still run and make the modification or will the workflow not continue with its execution for subsequent activities after termination?

Once an activity has been scheduled, any external side effects, like updating a database, are going to happen regardless of the timing of the workflow termination. We can improve things and implement activity termination, but it’s no guarantee that you’ll be able to stop the activity’s side effecting behavior since termination is inherently a race.

clintsinger:
I guess I'm still on the "activities scheduled".   If given the following workflow.

WF 
  Activity1
  Activity2 : Executing
  Activity3

If a termination is sent while "Activity2" is executing I assume it will complete but it won't do anything with the result.  What happens with Activity3?  Does it get scheduled as well?

clintsinger:
Now that you could potentially have two workflows executing with the same workflow id, what does it mean when you call the "Get" and "Terminate" methods?  Does it return the highest generation workflow or something else?

pkg/runtime/wfengine/workflow.go
```diff
@@ -534,9 +534,9 @@ func getRuntimeState(actorID string, state workflowState) *backend.Orchestration
 	return backend.NewOrchestrationRuntimeState(api.InstanceID(actorID), state.History)
 }
 
-func getActivityActorID(workflowActorID string, taskID int32) string {
-	// An activity can be identified by it's name followed by it's task ID. Example: SayHello::0, SayHello::1, etc.
-	return fmt.Sprintf("%s::%d", workflowActorID, taskID)
+func getActivityActorID(workflowActorID string, taskID int32, generation uint64) string {
+	// An activity can be identified by it's name followed by it's task ID and generation. Example: SayHello::0::1, SayHello::1::1, etc.
```
ItalyPaleAle, line 538: Nit
```suggestion
	// An activity can be identified by its name followed by its task ID and generation. Example: SayHello::0::1, SayHello::1::1, etc.
```

ItalyPaleAle (approved):

cgillum:
> I guess I'm still confused about the "activities scheduled". If given the following workflow.
> 
> WF
> Activity1
> Activity2 : Executing
> Activity3
> 
> If a termination is sent while "Activity2" is executing I assume it will complete but it won't do anything with the result. What happens with Activity3? Does it get scheduled as well?

If you terminate a workflow while Activity 2 is executing, Activity 3 will NOT execute (otherwise, terminate isn't really doing anything at all).

> Now that you could potentially have two workflows executing with the same workflow id, what does it mean when you call the "Get" and "Terminate" methods? Does it return the highest generation workflow or something else?

It's never possible to have two workflows executing with the same workflow ID. The singleton nature of actors guarantees this. In the extremely unlikely event that there's a duplicate actor activation for a particular workflow, a single terminate operation would effectively terminate both of them.

mukundansundar (approved):

PR #6458, opened 2023-06-07
Unify HTTP channels implementations: pkg/channel/http and pkg/channel/external

### LINKED ISSUE #6515: service invocation for non-dapr endpoints requires app channel
## In what area(s)?

/area runtime

## What version of Dapr?

> 1.11

## Expected Behavior

Created a naked daprd with the CLI and requested an invocation with curl with an explicit URL

Expected the content of the website at the URL to be returned

## Actual Behavior

received an error

> {
   "errorCode":"ERR_DIRECT_INVOKE","
   message":"fail to invoke, id: https://www.iana.org, err: cannot invoke http endpoint: http endpoints app channel not initialized"
}

Reported on Discord in '#service-invocation' (as discord user  hcguy) and @sicoyle advised to create this issue to track

`dapr run -H 3500`
`curl http://localhost:3500/v1.0/invoke/https://www.iana.org/method/domains/reserved`

## Release Note

RELEASE NOTE:

ItalyPaleAle:
The package `pkg/channel/external` contained a lot of duplication with `pkg/channel/http`, and the two had already started to drift apart ([_mea culpa_](https://github.com/dapr/dapr/pull/6236)).

This PR makes `pkg/channel/http` be usable for both internal and external invocation, reducing duplicate code.

Also fixes some bugs:

- External HTTP channel doesn't work if app channel isn't started. Fixes #6515
- Setting `-app-protocol h2c` would have previously caused the external HTTP channel to use H2C too, which would have likely caused failures


pkg/actors/internal_actor.go
```diff
@@ -88,7 +88,7 @@ func (internalActorChannel) HealthProbe(ctx context.Context) (bool, error) {
 }
 
 // InvokeMethod implements channel.AppChannel
-func (c *internalActorChannel) InvokeMethod(ctx context.Context, req *invokev1.InvokeMethodRequest) (*invokev1.InvokeMethodResponse, error) {
+func (c *internalActorChannel) InvokeMethod(ctx context.Context, req *invokev1.InvokeMethodRequest, appID string) (*invokev1.InvokeMethodResponse, error) {
```
JoshVanL, line 91:
```suggestion
func (c *internalActorChannel) InvokeMethod(ctx context.Context, req *invokev1.InvokeMethodRequest, _ string) (*invokev1.InvokeMethodResponse, error) {
```

JoshVanL (dismissed):
nit

pkg/http/api_test.go
```diff
@@ -1408,9 +1392,7 @@ func TestV1DirectMessagingEndpoints(t *testing.T) {
 				mock.MatchedBy(func(b string) bool {
 					return b == "fakeAppID"
 				}),
-				mock.MatchedBy(func(c *invokev1.InvokeMethodRequest) bool {
-					return true
-				}),
+				mock.AnythingOfType("*v1.InvokeMethodRequest"),
```
mukundansundar, line 1395: is there a reason to switch this to this format?

ItalyPaleAle: Uhm, I made a bunch of changes back-and-forth in this code (for a while, I had removed the 2nd argument `appId`), when I reverted them I didn't notice this was changed.

Shouldn't make much of a difference however, the effect is the same.

artursouza (approved):

Issue #6364, opened 2023-05-17
Runtime awareness of memory consumption to prevent OOM


berndverst:
**What:** The runtime should be self-aware of its memory consumption and gracefully pause or refuse additional requests.

**Why:** This is to avoid the sidecar container from being killed due to being Out of Memory (OOM).

**Who benefits from this?** This is especially important when a memory limit has been defined on the Dapr sidecar, or when Dapr is run on systems with small available memory. That could be in an IoT / embedded scenario, on a Raspberry Pi Kubernetes Cluster - or simply a small Kubernetes cluster with very small VM sizes.

**How can this be done?**

Conceptually, I recommend that the Dapr HTTP and gRPC server upon each incoming request immediately check memory utilization. If the utilization is beyond the current allowable limit (say Kubernetes container resource limit - 50MB):
1. Pause incoming connections for a certain period of time, frequently checking if the request can be processed because the memory utilization has been reduced.
2. If after a certain period of time the memory utilization is not in the allowable range, the request should be denied with HTTP 503 (temporarily unavailable). This is far better than killing the sidecar pod.

Here is an example of how this could work conceptually:

Imagine this where my sidecar definition (the memory limit is exposed to the sidecar using the Kubernetes Downward API):

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: fasthttp-example
spec:
  replicas: 1
  selector:
    matchLabels:
      app: fasthttp-example
  template:
    metadata:
      labels:
        app: fasthttp-example
    spec:
      containers:
      - name: daprd
        image: daprio/daprd:latest
        env:
        - name: MAX_MEMORY
          valueFrom:
            resourceFieldRef:
              containerName: daprd
              resource: limits.memory
              divisor: "1Mi"
```

And for example the HTTP request handler of the Dapr API server could work like so:

```go
package main

import (
	"fmt"
	"log"
	"runtime"
	"time"

	"github.com/valyala/fasthttp"
)

const holdDuration = 3 * time.Second
const checkInterval = 100 * time.Millisecond

func main() {
	maxMemoryStr := os.Getenv("MAX_MEMORY")
	if maxMemoryStr == "" {
		log.Fatal("MAX_MEMORY environment variable not set")
	}
	maxMemory, err := strconv.ParseInt(maxMemoryStr, 10, 64)
	if err != nil {
		log.Fatalf("Invalid MAX_MEMORY value: %v", err)
	}
	maxMemory *= 1024 * 1024

[...]

PR #6461, opened 2023-06-07
Migrate HTTP router to Chi

ItalyPaleAle:
# Description

This is a follow-up PR form #6248 including the next step in migrating Dapr away from fasthttp.

This PR includes:

- Migrates the router from fasthttprouter to Chi (based on net/http). Migrating the router is necessary to be able to write handlers that are based on net/http, and thus leverage things like streaming (although we do have adapters that convert handlers from net/http to fasthttp, during the conversion we lose the ability to do streaming, and it adds overhead)
- Implements a UniversalHTTPHandler which adapts Universal APIs into net/http handlers 
  - Migrates _some_ (not all) universal APIs from UniversalFastHTTPHandler to UniversalHTTPHandler

In the current master branch, this is the stack:

Current

After this PR, the stack changes to:

After

Separate PRs will follow-up to convert the remaining universal handlers to net/http, and to convert existing fasthttp handlers into either net/http or universal ones (this last conversion is a longer-term project and is not strictly required for handlers that do not need streaming or other similar features; however, eliminating the fasthttp adapter will improve performance)

Some additional notes:

- The router's behavior remains the same, with one exception only:
  - Take for example the route `v1.0/state/{storeName}/{key}`. Before, passing an empty `{storeName}`, such as `v1.0/state//somekey` would have still invoked the handler, which would have received an empty "storeName" (and likely returned 400 - bad request). Because the router now normalizes paths and removes double slashes (see #6324), the path is converted to `v1.0/state/somekey`, so it now returns a 404 because it can't match any route.
- The invoke method continues to work as before, but now _also_ supports external endpoints that are URL-encoded, such as `v1.0/invoke/http:%2F%2Fexample.com/method/foo` (which are not working in Dapr 1.11) 

## Issue reference

- Related to #4979 (addresses some of the follow-ups, but not all)
- Necessary (but not quite sufficient) to fix also:
  - #5392
  - #6246
  - #6364
  - microsoft/azure-container-apps#760


docs/decision_records/architecture/ARC-004-http-server.md
```diff
@@ -1,18 +0,0 @@
-# ARC-004: HTTP API server
```
JoshVanL, line 1: Should we document the decision to go to `net/http` + `chi`?

pkg/http/universalapi.go
```diff
@@ -51,10 +59,109 @@ type UniversalFastHTTPHandlerOpts[T proto.Message, U proto.Message] struct {
...
+			if len(body) > 0 {
+				err = pjsonDec.Unmarshal(body, in)
+				if err != nil {
+					msg := messages.ErrMalformedRequest.WithFormat(err)
+					respondWithError(w, msg)
+					log.Debug(msg)
+					return
+				}
+			}
+		}
```
JoshVanL, line 101: Is there no way to marshal the body reader directly? Pain to read the entire body first.

JoshVanL (dismissed):

pkg/http/universalapi.go
JoshVanL: :+1: sadge

pkg/http/responses_fasthttp.go
```diff
@@ -0,0 +1,60 @@
+/*
+Copyright 2021 The Dapr Authors
```
daixiang0, line 2: 2023?

ItalyPaleAle:
```suggestion
Copyright 2023 The Dapr Authors
```

pkg/diagnostics/http_monitoring.go
```diff
@@ -211,8 +211,8 @@ func (h *httpMetrics) Init(appID string) error {
 }
 
 // HTTPMiddleware is the middleware to track HTTP server-side requests.
-func (h *httpMetrics) HTTPMiddleware(next http.HandlerFunc) http.HandlerFunc {
-	return func(w http.ResponseWriter, r *http.Request) {
+func (h *httpMetrics) HTTPMiddleware(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
mukundansundar, line 215: qq, why is this change needed?

pkg/diagnostics/http_monitoring.go
```diff
@@ -230,13 +230,13 @@ func (h *httpMetrics) HTTPMiddleware(next http.HandlerFunc) http.HandlerFunc {
 
 		start := time.Now()
 
-		next(w, r)
+		next.ServeHTTP(w, r)
```
mukundansundar, line 233: Is this the standard way to call next with Chi router?

pkg/grpc/universalapi/api_metadata.go
```diff
@@ -89,6 +89,11 @@ func (a *UniversalAPI) GetMetadata(ctx context.Context, in *emptypb.Empty) (*run
 
 // SetMetadata Sets value in extended metadata of the sidecar.
 func (a *UniversalAPI) SetMetadata(ctx context.Context, in *runtimev1pb.SetMetadataRequest) (*emptypb.Empty, error) {
+	// Nop if the key is empty
+	if in.Key == "" {
+		return &emptypb.Empty{}, nil
+	}
```
mukundansundar, line 95: please add a separate PR... Let this PR be only chi router change.

pkg/diagnostics/http_monitoring.go
ItalyPaleAle: Chi is just a router, this is the standard way of creating a middleware with the std library. https://drstearns.github.io/tutorials/gomiddleware/

pkg/diagnostics/http_monitoring.go
ItalyPaleAle: Consistency with the other middlewares

pkg/grpc/universalapi/api_metadata.go
ItalyPaleAle: I _think_ this was needed to fix a test that was failing. Due to how the new router works, it may have been needed.

ItalyPaleAle:
The go-sdk test is not working due to a test issue.

halspang (commented):
Lots of changes, but overall seems fine. Just a few questions for my own understanding :)

pkg/http/api.go
```diff
@@ -208,249 +208,251 @@ func (a *api) MarkStatusAsOutboundReady() {
...
 		{
-			Methods:           []string{router.MethodWild},
-			Route:             "invoke/{id}/method/{method:*}",
-			Alias:             "{method:*}",
-			Version:           apiVersionV1,
-			KeepParamUnescape: true,
-			Handler:           a.onDirectMessage,
+			// No method is defined here to match any method
+			Methods: []string{},
+			Route:   "invoke/*",
```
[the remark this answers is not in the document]
ItalyPaleAle: Before the handler could be invoked in 2 ways:

- `invoke/{id}/method/{method:*}`
- `{method:*}`

Because of that and especially with the introduction of external service invocation, which makes it so `method` can include `/` even in variant 1 (and those can, but don't necessarily have to, be encoded as `%2F`), it made the parsing of the arguments a lot more complex. You can see how even in the current master, the matching is not straightfoward, and we've had bugs in there already.

Because the 1st option is essentially a subset of the 2nd, and because of changes in the router and how it handles %2F etc, the best way to make this work was to remove all path matching decisions from the router and move that to the handler itself.

I did discuss this with @artursouza and @yaron2 on discord, as the possible solution to make everything work, including making URL-encoding of `/` for external service invocation optional.

I'm not saying this is "simple" by all means, as it involves even a state machine (which however has full unit test coverage), but it was either this or a regular expression, which may be simpler or not (depends who you ask :) ) but it's certainly slower.

pkg/http/api_crypto_test.go
```diff
@@ -96,25 +96,6 @@ func TestCryptoEndpoints(t *testing.T) {
 		}
 	})
 
-	t.Run("Missing component name - 400", func(t *testing.T) {
```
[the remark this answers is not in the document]
ItalyPaleAle: With the updated router this case is impossible because `//` gets simplified to `/`. In any case, this will still cause an error, just a not specific one.

pkg/http/api_test.go
```diff
@@ -696,7 +696,7 @@ func TestBulkPubSubEndpoints(t *testing.T) {
 	})
 
 	t.Run("Bulk Publish with topic name '/' - 204", func(t *testing.T) {
-		apiPath := fmt.Sprintf("%s/publish/bulk/pubsubname//", apiVersionV1alpha1)
+		apiPath := fmt.Sprintf("%s/publish/bulk/pubsubname/%%2F", apiVersionV1alpha1)
```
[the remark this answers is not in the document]
ItalyPaleAle: Yes, because in `fmt.Sprintf` the `%` is otherwise interpreted as a formatting one (like in `%f`). So to have an actual `%`, it needs to be doubled as `%%`

pkg/http/api_test.go
```diff
@@ -820,12 +820,27 @@ func TestGetStatusCodeFromMetadata(t *testing.T) {
 }
 
 func TestGetMetadataFromRequest(t *testing.T) {
+	// set
+	r, _ := gohttp.NewRequest(gohttp.MethodGet, "http://test.example.com/resource?metadata.test=test&&other=other", nil)
```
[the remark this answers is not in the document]
ItalyPaleAle: Uhm I just copied this from the fasthttp test below which did have `&&`. I assume it's intended for testing purposes. Just like `other` is included because it should not be returned.

pkg/http/api_test.go
```diff
@@ -3509,46 +3536,11 @@ func TestV1Alpha1Workflow(t *testing.T) {
 		},
 	}
 
-	fakeServer.StartServer(testAPI.constructWorkflowEndpoints())
+	fakeServer.StartServer(testAPI.constructWorkflowEndpoints(), nil)
 
 	/////////////////////
 	// START API TESTS //
 	/////////////////////
-	t.Run("Start with no workflow type", func(t *testing.T) {
```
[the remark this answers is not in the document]
ItalyPaleAle: Correct, they all have `//` which is simplified in the new router. Essentially this becomes `/` so the path doesn't have any match. They still do throw an error, but it's not a specific error anymore.

halspang (approved):

yaron2:
While there is a slight performance hit in regards to per-call latency and increase in memory consumption under load, it's not significant and we can merge this PR. When end-to-end streaming is supported it might offset any penalties introduced here.

ItalyPaleAle:
Thanks for merging @yaron2 !

I agree that the next work is very likely to offset some of the hit we are seeing here, to start thanks to the removal of the fasthttp conversion layer:

image

Will have the work done in the next days

kolinfluence:
use https://github.com/cloudwego/hertz instead? net/http http2 is too much slower

PR #6463, opened 2023-06-07
Integration testing: sentry

JoshVanL:
Branched from https://github.com/dapr/dapr/pull/6340

PR allows the port and healthz port to be configured on sentry.

Adds integration tests covering healthz and port for sentry.

### DIFF
--- a/pkg/sentry/config/config.go
+++ b/pkg/sentry/config/config.go
@@ -20,12 +20,13 @@ const (
 	kubernetesServiceHostEnvVar = "KUBERNETES_SERVICE_HOST"
 	kubernetesConfig            = "kubernetes"
 	selfHostedConfig            = "selfhosted"
-	defaultPort                 = 50001
 	defaultWorkloadCertTTL      = time.Hour * 24
 	defaultAllowedClockSkew     = time.Minute * 15
 
 	// defaultDaprSystemConfigName is the default resource object name for Dapr System Config.
 	defaultDaprSystemConfigName = "daprsystem"
+
+	DefaultPort = 50001
 )
 
 var log = logger.NewLogger("dapr.sentry.config")
@@ -58,8 +59,8 @@ func (c SentryConfig) String() string {
 		caStore = c.CAStore
 	}
 
-	return fmt.Sprintf("Configuration: port:'%v' ca store:'%s', allowed clock skew:'%s', workload cert ttl:'%s'",
-		c.Port, caStore, c.AllowedClockSkew.String(), c.WorkloadCertTTL.String())
+	return fmt.Sprintf("Configuration: ca store:'%s', allowed clock skew:'%s', workload cert ttl:'%s'",
+		caStore, c.AllowedClockSkew.String(), c.WorkloadCertTTL.String())
 }
 
 var configGetters = map[string]func(string) (SentryConfig, error){
@@ -94,7 +95,7 @@ func IsKubernetesHosted() bool {
 
 func getDefaultConfig() SentryConfig {
 	return SentryConfig{
-		Port:             defaultPort,
+		Port:             DefaultPort,
 		WorkloadCertTTL:  defaultWorkloadCertTTL,
 		AllowedClockSkew: defaultAllowedClockSkew,
 	}
--- a/pkg/sentry/config/config.go
+++ b/pkg/sentry/config/config.go
@@ -59,8 +59,8 @@ func (c SentryConfig) String() string {
 		caStore = c.CAStore
 	}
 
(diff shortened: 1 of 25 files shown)


tests/integration/framework/binary/binary.go
```diff
@@ -0,0 +1,82 @@
...
+		if runtime.GOOS == "windows" {
+			binPath += ".exe"
+		}
+
+		// Ensure CGO is disabled to avoid linking against system libraries.
+		require.NoError(t, os.Setenv("CGO_ENABLED", "0"))
+
+		t.Logf("Root dir: %q", rootDir)
+		t.Logf("Compiling %q binary to: %q", name, binPath)
+		cmd := exec.Command("go", "build", "-tags=allcomponents", "-v", "-o", binPath, filepath.Join(rootDir, "cmd/"+name))
```
filintod, line 66: is it worth removing the debug symbols (`-ldflags="-s -w"`)?

tests/integration/framework/process/exec/exec.go
```diff
@@ -0,0 +1,129 @@
...
+
+	e.cmd.Stdout = e.stdoutpipe
+	e.cmd.Stderr = e.stderrpipe
+
+	require.NoError(t, e.cmd.Start())
+}
+
+func (e *exec) Cleanup(t *testing.T) {
+	t.Helper()
+	e.lock.Lock()
```
filintod, line 111: can this be executed more than once? just thinking that if you kill the command on the first iteration, all the next ones will fail checkExit failing the test. or maybe it should be protected to run once only?

JoshVanL: `exec` is a process for executing a program file, and is more of a helper for the other processes which want to run binaries, as of this PR, daprd, placement, and sentry. For each test case, the framework will  `Run` and `Cleanup` each process exactly once, and the binary processes are going to do the same with their exec. Processes are not intended to be run more than once.. I can add a check for this if you think this will be beneficial?

filintod: I meant the Cleanup method, just wondering if there was a path where this would be called more than once and maybe failing the test at the end even if it was fine.  If not then don't worry

JoshVanL: Sure, I can refactor processes a bit to ensure these functions only get called once. This may be beneficial for future developers.

filintod (commented):
lgtm

pkg/sentry/config/config.go
```diff
@@ -58,8 +59,8 @@ func (c SentryConfig) String() string {
 		caStore = c.CAStore
 	}
 
-	return fmt.Sprintf("Configuration: port:'%v' ca store:'%s', allowed clock skew:'%s', workload cert ttl:'%s'",
-		c.Port, caStore, c.AllowedClockSkew.String(), c.WorkloadCertTTL.String())
+	return fmt.Sprintf("Configuration: ca store:'%s', allowed clock skew:'%s', workload cert ttl:'%s'",
```
ItalyPaleAle, line 62: Why was the port removed?

tests/integration/framework/binary/binary.go
ItalyPaleAle, line 66: Please consider including an option to add other build tags.

tests/integration/framework/process/daprd/daprd.go
```diff
@@ -0,0 +1,177 @@
...
+	internalGRPCPort int
+	publicPort       int
+	metricsPort      int
+	profilePort      int
+}
+
+func New(t *testing.T, fopts ...Option) *Daprd {
+	t.Helper()
+
+	uid, err := uuid.NewUUID()
```
ItalyPaleAle, line 71: Is there a reason for using `NewUUID` (UUID v1) rather than `New` (UUID v4)?

tests/integration/framework/binary/binary.go
JoshVanL: What build flags do you foresee us needing to add in future?

tests/integration/framework/process/daprd/daprd.go
JoshVanL: Fixed

tests/integration/framework/binary/binary.go
JoshVanL: I've left this like it is for now. If we need more build flags in future we can add them here, or if we need a unique list of flags for different binaries we can add this functionality in future.

JoshVanL:
@artursouza @ItalyPaleAle PTAL

ItalyPaleAle (approved):

Issue #6422, opened 2023-05-31
Add Strict-Sandbox Configuration CRD to Dapr for Enhanced WASM Security


Taction:
## In what area(s)?

 /area runtime

> /area operator

> /area placement

> /area docs

> /area test-and-release

### Background
Dapr supports WASM middleware and output binding features, allowing users to run custom WASM code to extend the functionality of Dapr without modifying the Dapr runtime. However, this introduces potential security threats, such as having access to the current time accurately, which could be used to perform timing attacks on the rest of Dapr, similar to the Spectre/Meltdown vulnerabilities. 

### Proposal
Adding a new configuration CRD to Dapr that enables a strict sandbox mode for all WASM components. This will enforce limits on time and file access to prevent potential attacks.

Strict Time Access
Set walltime to 2022-01-01 UTC and increment it by 1ms on each reading, set nanotime to 0 and increment it by 1ms on each reading, nanosleep returns immediately(These are the behaviours wazero runtime already supports).

Limited File Access
The strict sandbox will provide an actual mount point inside the /tmp directory for each WASM component, and mount to wasm's root file system. The floder name could be `dapr-wasm-${random}`.

An example of configuration:
```yaml
apiVersion: dapr.io/v1alpha1
kind: Configuration
metadata:
  name: wasm
spec:
  wasm:
    strictSandbox: true
```

#### Component Implementation
Adding `wasmStrictSandbox` metadata for every wasm component, and if `strictSandbox` is setted to true in configuration CRD, the dapr runtime should override this metadata.

## Release Note

RELEASE NOTE: Add Strict-Sandbox Configuration CRD to Dapr for Enhanced WASM Security

//cc @codefromthecrypt @ItalyPaleAle The basic idea is from you, I open this issue to have further discussion.

ItalyPaleAle:
Thanks for opening this issue.

I recommended making this option into the Configuration CRD (rather than in each component's metadata) for 2 reasons:

1. This way it applies to all and any WASM component that may be loaded.
2. To separate concerns, so the people that operate the cluster (which may be different from the people who deploy components) have control over this.

We currently don't pass options from the Configuration CRD to components, so some wiring will need to happen in [pkg/components](https://github.com/dapr/dapr/tree/master/pkg/components) to change what's passed to components' `Init()` methods.

As for this:

> Strict Time Access
[...]

PR #6464, opened 2023-06-08
Add wasm strict sandbox

Taction:
# Description

Add wasm strict sandbox configuration.
related contrib pr: https://github.com/dapr/components-contrib/pull/2884

## Issue reference

Please reference the issue this PR will close: #6422

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Unit tests passing


charts/dapr/crds/configuration.yaml
```diff
@@ -397,6 +397,14 @@ spec:
                 required:
                 - samplingRate
                 type: object
+              wasm:
+                description: WasmSpec describes the configuration for Wasm across all Dapr components.
```
ItalyPaleAle, line 401:
```suggestion
                description: WasmSpec describes the security profile for all Dapr Wasm components.
```

pkg/config/configuration.go
```diff
@@ -279,6 +280,11 @@ type ComponentsSpec struct {
 	Deny []string `json:"deny,omitempty" yaml:"deny,omitempty"`
 }
 
+// WasmSpec describes the configuration for Wasm across all Dapr components.
+type WasmSpec struct {
+	StrictSandbox *bool `json:"strictSandbox,omitempty" yaml:"strictSandbox,omitempty"`
```
ItalyPaleAle, line 285: If the default value is "false" (i.e. strict sandbox not enabled), consider making this non-nullable, i.e.:


```suggestion
	StrictSandbox bool `json:"strictSandbox,omitempty" yaml:"strictSandbox,omitempty"`
```

pkg/runtime/component/wasm.go
```diff
@@ -0,0 +1,22 @@
+// This needs to be moved to contrib repo if accepted.
```
ItalyPaleAle, line 1: +1

pkg/runtime/runtime.go
```diff
@@ -253,6 +254,10 @@ type pubsubSubscribedMessage struct {
 // NewDaprRuntime returns a new runtime with the given runtime config and global config.
 func NewDaprRuntime(runtimeConfig *Config, globalConfig *config.Configuration, accessControlList *config.AccessControlList, resiliencyProvider resiliency.Provider) *DaprRuntime {
 	ctx, cancel := context.WithCancel(context.Background())
+	if globalConfig.Spec.WasmSpec.StrictSandbox != nil {
```
ItalyPaleAle, line 257: Does this need to be added to the global context? How about adding it just to the context when we init a middleware component?

charts/dapr/crds/configuration.yaml
Taction: Thanks! Done.

pkg/config/configuration.go
Taction: Done. Make it non-nullable.

pkg/runtime/runtime.go
Taction: Only add this config to context when init a middleware component and an output binding component(Currently only these two types have wasm components).

pkg/runtime/component/wasm.go
```diff
@@ -0,0 +1,22 @@
...
+// Put it here just for disscuss how to pass configuration to components.
```
daixiang0, line 2: I would like that we have options in metadata then we can control it finer.

Taction: Thanks for review. Both ways is ok with me, Changed to pass option via metadata.

pkg/runtime/meta/meta.go
```diff
@@ -96,3 +97,12 @@ func ContainsNamespace(items []common.NameValuePair) bool {
 	}
 	return false
 }
+
+// AddWasmStrictSandbox adds wasm strict sandbox configuration to metadata.
+func AddWasmStrictSandbox(comp *compapi.Component, enable bool) {
+	sandbox := &common.NameValuePair{
+		Name: "strictSandbox",
```
daixiang0, line 104: Please make it as const.

Taction: Done.

pkg/runtime/runtime.go
```diff
@@ -2409,6 +2411,11 @@ func (a *DaprRuntime) processComponentAndDependents(comp componentsV1alpha1.Comp
 		timeout = defaultComponentInitTimeout
 	}
 
+	if compCategory == components.CategoryBindings && comp.Spec.Type == string(components.CategoryBindings)+".wasm" {
```
ItalyPaleAle, line 2414: I am not a fan of this TBH. I think we should modify the registry to indicate "this is a wasm component" without looking at the name.

What if we add more Wasm components in the future?

ItalyPaleAle (changes requested):

yaron2:
@Taction please see comments and conflicts

pkg/apis/configuration/v1alpha1/types.go
```diff
@@ -73,6 +75,11 @@ type APISpec struct {
 	Denied []APIAccessRule `json:"denied,omitempty"`
 }
 
+// WasmSpec describes the security profile for all Dapr Wasm components.
+type WasmSpec struct {
+	StrictSandbox bool `json:"strictSandbox,omitempty"`
```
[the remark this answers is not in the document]
Taction: Removed `omitempty` make  `StrictSandbox` as required. Mark the `WasmSpec` as optional. In this way, we can tell whether it is defined as false or undefined.

pkg/runtime/meta/meta.go
```diff
@@ -96,3 +99,12 @@ func ContainsNamespace(items []common.NameValuePair) bool {
...
 	return false
 }
+
+// AddWasmStrictSandbox adds wasm strict sandbox configuration to metadata.
+func AddWasmStrictSandbox(comp *compapi.Component, enable bool) {
+	sandbox := &common.NameValuePair{
+		Name: WasmStrictSandboxMetadataKey,
+	}
+	sandbox.SetValue([]byte(strconv.FormatBool(enable)))
+	comp.Spec.Metadata = append(comp.Spec.Metadata, *sandbox)
```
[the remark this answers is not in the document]
Taction: Done.

pkg/runtime/meta/meta.go
```diff
@@ -96,3 +99,12 @@ func ContainsNamespace(items []common.NameValuePair) bool {
 	}
 	return false
 }
+
+// AddWasmStrictSandbox adds wasm strict sandbox configuration to metadata.
+func AddWasmStrictSandbox(comp *compapi.Component, enable bool) {
```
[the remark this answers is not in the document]
Taction: Done.
1. if `enable` is undefined, do nothing, keep the user's configuration. Otherwise:
2. scan `comp.Spec.Metadata` if `WasmStrictSandboxMetadataKey` exists update, else add.

pkg/runtime/runtime.go
Taction: Done, Add to those components marked as wasm. And wasm component should register themself.

pkg/apis/configuration/v1alpha1/types.go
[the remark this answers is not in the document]
ItalyPaleAle: Please keep it optional in both places.

1. Unless you use pointers, it is not possible to distinguish between an explicit false or undefined. I also don't think we need to know an explicit false vs undefined.
2. This would be requires if `wasm` is set, but if `wasm` is empty then it's not required. It doesn't seem right.

pkg/runtime/runtime.go
```diff
@@ -163,11 +164,16 @@ func newDaprRuntime(ctx context.Context,
 	resiliencyProvider resiliency.Provider,
 ) (*DaprRuntime, error) {
 	compStore := compstore.New()
+	wasmStrictSandbox := (*bool)(nil)
```
ItalyPaleAle, line 167: Why is this a pointer? I don't think the "nil" case matters here:

CRD can force the strict sandbox to be enabled, but it cannot force it to be disabled. If for a specific component the developers/ops teams want it to be enabled, it should be enabled.

Taction: Done. changed to bool.

pkg/apis/configuration/v1alpha1/types.go
```diff
@@ -73,6 +75,14 @@ type APISpec struct {
 	Denied []APIAccessRule `json:"denied,omitempty"`
 }
 
+// WasmSpec describes the security profile for all Dapr Wasm components.
+type WasmSpec struct {
+	// Force enabling strict sandbox mode for all WASM components.
+	// When this is enabled, WASM components always run in strict mode regardless of their configuration.
+	// Strict mode enhances security of the WASM sandbox by limiting access to certain capabilities such as real-time clocks and random number generators.
+	StrictSandbox bool `json:"strictSandbox"`
```
ItalyPaleAle, line 83:
```suggestion
	StrictSandbox bool `json:"strictSandbox,omitempty"`
```

charts/dapr/crds/configuration.yaml
```diff
@@ -397,6 +397,14 @@ spec:
                 required:
                 - samplingRate
                 type: object
+              wasm:
+                description: WasmSpec describes the security profile for all Dapr Wasm components.
+                properties:
+                  strictSandbox:
+                    type: boolean
+                required:
+                  - strictSandbox
```
ItalyPaleAle, line 406:
```suggestion
```

pkg/config/configuration.go
```diff
@@ -297,6 +298,14 @@ type ComponentsSpec struct {
 	Deny []string `json:"deny,omitempty" yaml:"deny,omitempty"`
 }
 
+// WasmSpec describes the security profile for all Dapr Wasm components.
+type WasmSpec struct {
+	// Force enabling strict sandbox mode for all WASM components.
+	// When this is enabled, WASM components always run in strict mode regardless of their configuration.
+	// Strict mode enhances security of the WASM sandbox by limiting access to certain capabilities such as real-time clocks and random number generators.
+	StrictSandbox bool `json:"strictSandbox" yaml:"strictSandbox"`
```
ItalyPaleAle, line 306:
```suggestion
	StrictSandbox bool `json:"strictSandbox,omitEmpty" yaml:"strictSandbox,omitEmpty"`
```

pkg/runtime/meta/meta.go
```diff
@@ -96,3 +108,28 @@ func ContainsNamespace(items []common.NameValuePair) bool {
...
 }
+
+// AddWasmStrictSandbox adds global wasm strict sandbox configuration to component metadata.
+// When strict sandbox is enabled, WASM components always run in strict mode regardless of their configuration.
+// When strict sandbox is disabled or unset, keep the original component configuration.
+func (m *Meta) AddWasmStrictSandbox(comp *compapi.Component) {
+	// If the global strict sandbox is disabled(or unset), do nothing.
+	if m.strictSandbox == false {
+		return
+	}
```
ItalyPaleAle, line 119:
```suggestion
	// If the global strict sandbox is disabled (or unset), it is not enforced.
	if !m.strictSandbox {
		return
	}
```

pkg/runtime/meta/meta.go
```diff
@@ -96,3 +108,28 @@ func ContainsNamespace(items []common.NameValuePair) bool {
...
+
+	// If the metadata already contains the strict sandbox key, update the value to global strict sandbox config.
+	for i, c := range comp.Spec.Metadata {
+		if strings.EqualFold(c.Name, WasmStrictSandboxMetadataKey) {
+			comp.Spec.Metadata[i].SetValue([]byte(strconv.FormatBool(m.strictSandbox)))
```
ItalyPaleAle, line 124: If we're here, the value can only be "true"

```suggestion
			comp.Spec.Metadata[i].SetValue([]byte("true"))
```

pkg/runtime/meta/meta.go
```diff
@@ -96,3 +108,28 @@ func ContainsNamespace(items []common.NameValuePair) bool {
...
+			return
+		}
+	}
+
+	// If the metadata does not contain the strict sandbox key, add it.
+	sandbox := common.NameValuePair{
+		Name: WasmStrictSandboxMetadataKey,
+	}
+	sandbox.SetValue([]byte(strconv.FormatBool(m.strictSandbox)))
```
ItalyPaleAle, line 133:
```suggestion
	sandbox.SetValue([]byte("true"))
```

pkg/apis/configuration/v1alpha1/types.go
Taction: Done.

pkg/runtime/meta/meta.go
Taction: Done.

pkg/runtime/meta/meta.go
Taction: Done.

pkg/runtime/meta/meta.go
Taction: Done.

pkg/apis/configuration/v1alpha1/types.go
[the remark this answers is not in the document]
Taction: Done.

ItalyPaleAle (approved):

daixiang0:
@Taction please fix conflicts then we can make it land.

Taction:
@daixiang0 Thanks, conflicts are fixed.

daixiang0 (approved):
LGTM

daixiang0:
@yaron2 today code freeze, do you think it worth land in 1.12 release?

yaron2:
> @yaron2 today code freeze, do you think it worth land in 1.12 release?

Code freeze is now Sept 5. Merged. Thanks @Taction for this

PR #6474, opened 2023-06-08
Fixes in AKS test pipelines

ItalyPaleAle:
Two fixes in the AKS test pipeline:

1. Disables deploying the ARM64 cluster when not in use. We haven't been running tests on ARM64 in months, but we still deploy a cluster (and all resources) every time. This is unnecessarily expensive and helps burning through our Azure credits.
2. We've been able to trigger the workflow manually for a while, but when triggered from a branch (e.g. release-1.11) it was still cloning the code from master. Now, if the workflow is run manually from a branch, it clones the correct branch. This will allow us to run tests on the release branches at any time (and it's a better option than opening a PR against that branch because in this case the workflow YAML comes from the same branch)
  - This was fixed in the perf test YAML too
3. Updates some soon-to-be deprecated Actions that were triggering a warning

Tests:

- E2E: https://github.com/dapr/dapr/actions/runs/5214807072
- Perf: https://github.com/dapr/dapr/actions/runs/5214581409

PS: This PR was opened from a branch in dapr/dapr and not from my fork because otherwise there's no way to test the pipeline (using /ok-to-test here wouldn't work)

### DIFF
--- a/.github/workflows/dapr-test.yml
+++ b/.github/workflows/dapr-test.yml
@@ -225,13 +225,19 @@ jobs:
             target_arch: amd64
             windows_version: ltsc2022
     steps:
+      - name: Set up for manual runs
+        if: github.event_name == 'workflow_dispatch'
+        run: |
+          echo "CHECKOUT_REPO=${{ github.repository }}" >> $GITHUB_ENV
+          echo "CHECKOUT_REF=${{ github.ref }}" >> $GITHUB_ENV
+        shell: bash
       - name: Set up for scheduled test
-        if: github.event_name != 'repository_dispatch'
+        if: github.event_name == 'schedule'
         run: |
           echo "CHECKOUT_REPO=${{ github.repository }}" >> $GITHUB_ENV
           echo "CHECKOUT_REF=refs/heads/master" >> $GITHUB_ENV
         shell: bash
-      - name: Parse test payload
+      - name: Set up for dispatched events
         if: github.event_name == 'repository_dispatch'
         uses: actions/github-script@v6.2.0
         with:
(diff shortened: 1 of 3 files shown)


.github/workflows/dapr-test.yml
```diff
@@ -397,9 +409,9 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        target_os: [linux]
-        target_arch: [amd64, arm64]
```
daixiang0, line 401: Same as above, is it possible to remove arm64 here directly?

artursouza:
This change was somehow included in the merge from release-1.11 into master and the PR got closed because of that.

.github/workflows/dapr-test.yml
artursouza: @ItalyPaleAle Please, follow up on these.

.github/workflows/dapr-test.yml
```diff
@@ -207,25 +216,28 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-22.04, windows-2022]
-        target_arch: [amd64]
```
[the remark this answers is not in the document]
artursouza: @ItalyPaleAle Please, follow up on these.

ItalyPaleAle:
> This change was somehow included in the merge from release-1.11 into master and this PR is merged because of that.

I’m not sure how that happened but i can open a PR reverting this

.github/workflows/dapr-test.yml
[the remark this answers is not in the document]
ItalyPaleAle: It’s equivalent, the net effect is the same. I do like the clarity that a list, rather than a matrix, provides.

.github/workflows/dapr-test.yml
ItalyPaleAle: Same as above. The net effect is the same, this is just more clear because windows was not in the matrix. Now it’s easier to tell what gets executed.

ItalyPaleAle:
I've re-opened this in #6488 please review that @daixiang0

PR #6490, opened 2023-06-09
Fix panic in service invocation when connection fails

ItalyPaleAle:
Fixes error reported by user on Discord: https://discord.com/channels/778680217417809931/1055270900125138975/1109309802435321896

### DIFF
--- a/pkg/messaging/direct_messaging.go
+++ b/pkg/messaging/direct_messaging.go
@@ -267,10 +267,6 @@ func (d *directMessaging) isHTTPEndpoint(appID string) bool {
 	return ok
 }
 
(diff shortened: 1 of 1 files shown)


pkg/messaging/direct_messaging.go
```diff
@@ -267,10 +267,6 @@ func (d *directMessaging) isHTTPEndpoint(appID string) bool {
 	return ok
 }
 
-func noopTeardown(destroy bool) {
```
ItalyPaleAle, line 270: There was a `noopTeardown` and a `nopTeardown` so removed the duplicated

yaron2 (approved):

PR #6493, opened 2023-06-09
Fixed: operations on reminders should not be allowed if the sidecar isn't hosting actors of that type

### LINKED ISSUE #6150: Actor reminder replacement silently fails if the sidecar doesn't host actors of the given type
## In what area(s)?

/area runtime

## What version of Dapr?

> master
> v1.10.x

## Description

Start Dapr without an app:

```sh
cd cmd/daprd
go run . \
  --app-id testreminders \
  --resources-path resources \
  --placement-host-address localhost:50005 \
  --dapr-http-port 3500
```

(You will need to have a a Component YAML in the `resources` folder that is configured as actorStateStore).

On another terminal:

```sh
curl -l --request POST 'http://localhost:3500/v1.0/actors/myactor/actor1/reminders/testreminder3' \
  --header 'Content-Type: application/json' \
  --data-raw '{"dueTime": "2h","period": "1h"}'
curl 'http://localhost:3500/v1.0/actors/myactor/actor1/reminders/testreminder3'

# Output: {"period":"1h","dueTime":"2h"}
```

Then **stop and restart** the sidecar in the first terminal.

Then run in the second terminal:

```sh
curl -l --request POST 'http://localhost:3500/v1.0/actors/myactor/actor1/reminders/testreminder3' \
  --header 'Content-Type: application/json' \
  --data-raw '{"dueTime": "14h","period": "2h"}'
curl 'http://localhost:3500/v1.0/actors/myactor/actor1/reminders/testreminder3'
# Output is wrong: {"period":"1h","dueTime":"2h"}
```

Repeat:

```sh
curl -l --request POST 'http://localhost:3500/v1.0/actors/myactor/actor1/reminders/testreminder3' \
  --header 'Content-Type: application/json' \
  --data-raw '{"dueTime": "14h","period": "2h"}'
curl 'http://localhost:3500/v1.0/actors/myactor/actor1/reminders/testreminder3'
# Output is now right: {"period":"2h","dueTime":"14h"}
```

Confirmed with Dapr 1.10.4 (`git checkout v1.10.4`).

## Release Note

RELEASE NOTE: **FIX** Updating a reminder silently fails if the sidecar is not an actor host

ItalyPaleAle:
Fixes #6150

# Description

Although endpoints to modify actor reminders are meant to be invoked by running actors only, there's currently nothing preventing them from being invoked by anyone. This causes interesting issues such as #6150, where operations on reminders can fail if they are being executed by a sidecar that doesn't host actors of the given type.

This implements the fix discussed on the issue (https://github.com/dapr/dapr/issues/6150#issuecomment-1489665935), disabling operations on actor reminders if the sidecar isn't hosting actors of the given type.

This should be considered a **breaking change**, even if (hopefully) of small impact. It prevents behaviors that were never intended to be used (but may have been used by someone, per Hyrum's law) and that were buggy and not always working as intended.

## Release Note

RELEASE NOTE: **FIXED** Disallow operations on reminders for actor types that are not currently being hosted

### DIFF
--- a/pkg/actors/actors.go
+++ b/pkg/actors/actors.go
@@ -154,11 +154,12 @@ type actorReminderReference struct {
 	reminder                  reminders.Reminder
 }
 
-var ErrIncompatibleStateStore = errors.New("actor state store does not exist, or does not support transactions which are required to save state - please see https://docs.dapr.io/operations/components/setup-state-store/supported-state-stores/")
-
-var ErrDaprResponseHeader = errors.New("error indicated via actor header response")
-
-var ErrReminderCanceled = errors.New("reminder has been canceled")
+var (
+	ErrIncompatibleStateStore   = errors.New("actor state store does not exist, or does not support transactions which are required to save state - please see https://docs.dapr.io/operations/components/setup-state-store/supported-state-stores/")
+	ErrDaprResponseHeader       = errors.New("error indicated via actor header response")
+	ErrReminderCanceled         = errors.New("reminder has been canceled")
+	ErrReminderOpActorNotHosted = errors.New("operations on actor reminders are only possible on hosted actor types")
+)
 
 // ActorsOpts contains options for NewActors.
 type ActorsOpts struct {
@@ -247,9 +248,11 @@ func (a *actorsRuntime) Init() error {
 	if a.placement == nil {
 		a.placement = internal.NewActorPlacement(
 			a.config.PlacementAddresses, a.certChain,
-			a.config.AppID, hostname, a.config.PodName, a.config.HostedActorTypes,
+			a.config.AppID, hostname, a.config.PodName,
+			a.config.HostedActorTypes.ListActorTypes(),
 			appHealthFn,
-			afterTableUpdateFn)
+			afterTableUpdateFn,
+		)
 	}
 
 	go a.placement.Start()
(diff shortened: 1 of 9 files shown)


pkg/actors/actors.go
```diff
@@ -1861,6 +1874,10 @@ func (a *actorsRuntime) storeReminder(ctx context.Context, store transactionalSt
 }
 
 func (a *actorsRuntime) GetReminder(ctx context.Context, req *GetReminderRequest) (*reminders.Reminder, error) {
+	if !a.config.HostedActorTypes.IsActorTypeHosted(req.ActorType) {
+		return nil, ErrReminderOpActorNotHosted
```
daixiang0, line 1878: Could add a test to cover this case?

ItalyPaleAle: There are unit tests in pkg/http that cover this scenario, as well as E2E tests.

Additionally the unit tests in pkg/actors have been updated to make the actor type hosted, so the positive case is included.

I think that should be enough?

daixiang0 (approved):
LGTM, wait for CI happy

ItalyPaleAle:
Per conversation with @yaron2 over Discord this change should not be considered a major breaking change, although it has a chance of breaking a small number of users. However:

- It seems that SDKs (or at least the ones I've looked at) don't allow creating reminders for actors that aren't hosted anyways, so this _may_ only impact users that directly invoke the Actor APIs - arguably a small minority
- Users who did end up invoking this method to create reminders for actors that aren't hosted would have run into a number of issues. Even when the operation succeeded, it could have caused issues downstreams.

I have added the "breaking change" label so we remember to document this change clearly. However, maintainers don't seem concerned with it particulary.

PR #6494, opened 2023-06-09
In self-hosted mode, filter declarative subscriptions by namespace if set

### LINKED ISSUE #6468: Component of type kind: Subscription not respecting metadata/namespace.
## In what area(s)?

/area runtime

## What version of Dapr?

1.10

## Expected Behavior

When I have multiple subscription definitions `kind: Subscription` I expect only those with-in the `metadata/namespace` to be loaded.  This behavior works with components, I was hoping it would work with subscription as well.

The use case is we have a shared service bus, with all of our queues named per env.
ie: my-dev-queue, my-qa-queue, my-prod-queue.

This means we cannot use the `Topic` attribute to designate what the endpoint should be triggered by.
We are currently segregating our components by using the namespace attribute in all of their respective metadata settings.  This way we can easily separate the components per env.  This has not stopped with subscriptions.

## Actual Behavior

Observe the following two subscription definitions
```
apiVersion: dapr.io/v1alpha1
kind: Subscription
metadata:
  name: submit
  namespace: local
spec:
  topic: local-submit-queue
  route: /submit
  pubsubname: messagebus-primary
scopes:
  - processor-service
```
```
apiVersion: dapr.io/v1alpha1
kind: Subscription
metadata:
  name: submit
  namespace: dev
spec:
  topic: development-submit-queue
  route: /submit
  pubsubname: messagebus-primary
scopes:
  - processor-service
```

If I look into the dapr sidecar logs I see both of the subscriptions being loaded.  Only the one corresponding to the env namespace should be.

I've looked into the code and it appears that we we're running in local mode, we don't even send the namespace to the method to load subscriptions from the component folder.

Use the pub/sub QuickStart.  Add the two subscriptions above and set the env variable `namespace` to `dev`.

Observe both subscriptions are loaded.

## Release Note

RELEASE NOTE:

ItalyPaleAle:
Fixes #6468

### DIFF
--- a/pkg/runtime/pubsub/subscriptions.go
+++ b/pkg/runtime/pubsub/subscriptions.go
@@ -229,9 +229,9 @@ func GetSubscriptionsGRPC(channel runtimev1pb.AppCallbackClient, log logger.Logg
 }
 
 // DeclarativeLocal loads subscriptions from the given local resources path.
-func DeclarativeLocal(resourcesPaths []string, log logger.Logger) (subs []Subscription) {
+func DeclarativeLocal(resourcesPaths []string, namespace string, log logger.Logger) (subs []Subscription) {
 	for _, path := range resourcesPaths {
-		res := declarativeFile(path, log)
+		res := declarativeFile(path, namespace, log)
 		if len(res) > 0 {
 			subs = append(subs, res...)
 		}
@@ -240,7 +240,7 @@ func DeclarativeLocal(resourcesPaths []string, log logger.Logger) (subs []Subscr
 }
 
 // Used by DeclarativeLocal to load a single path.
-func declarativeFile(resourcesPath string, log logger.Logger) (subs []Subscription) {
+func declarativeFile(resourcesPath string, namespace string, log logger.Logger) (subs []Subscription) {
 	if _, err := os.Stat(resourcesPath); os.IsNotExist(err) {
 		return subs
 	}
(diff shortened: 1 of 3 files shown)


pkg/runtime/pubsub/subscriptions.go
```diff
@@ -270,7 +270,13 @@ func declarativeFile(resourcesPath string, log logger.Logger) (subs []Subscripti
 
 		bytesArray := bytes.Split(b, []byte("\n---"))
 		for _, item := range bytesArray {
-			subs, err = appendSubscription(subs, item)
+			// Skip empty items
+			item = bytes.TrimSpace(item)
+			if len(item) == 0 {
+				continue
```
mukundansundar, line 276: @ItalyPaleAle Can you add tests to cover this scenario?

pkg/runtime/pubsub/subscriptions.go
```diff
@@ -304,6 +310,9 @@ func marshalSubscription(b []byte) (*Subscription, error) {
 		if err := yaml.Unmarshal(b, &sub); err != nil {
 			return nil, err
 		}
+		if namespace != "" && sub.Namespace != "" && sub.Namespace != namespace {
+			return nil, nil
```
mukundansundar, line 314: Also this. Wonder why this is not covered by tests already written.

mukundansundar (commented):
small changes

pkg/runtime/pubsub/subscriptions.go
ItalyPaleAle: This is covered by E2E tests (all cases, where `namespace` is empty, `sub.Namespace` is empty and when `sub.Namespace != namespace`)

pkg/runtime/pubsub/subscriptions.go
ItalyPaleAle: Added

pkg/runtime/pubsub/subscriptions.go
ItalyPaleAle: Added

ItalyPaleAle: The reason why it wasn't already covered was because the test used the v1 CRD but that line was for the v2. I copied the test for v2

mukundansundar (approved):
lgtm

mukundansundar:
ideally once the integration test initial PR is merged, #6327 this PR should have an integration test for this change.

PR #6496, opened 2023-06-10
Add `envRef` in addition to `secretKeyRef` in metadata for Components and HTTPEndpoints

### LINKED ISSUE #6495: Component CRD: add `envRef` in addition to `secretRef` in metadata
## In what area(s)?

/area runtime

## Describe the feature

Following up from #6214

We would like to add a new property `envRef` to the metadata dictionary in the Component CRD, to allow daprd to grab the value of a secret from an env var while initializing a component.

This would work similarly to `secretRef`, but doesn't require a secret store to be initialized, and has priority over `secretRef` (but not over `value`).

Note that daprd will keep a denylist of env vars that can't be retrieved, just like the [local.env secret store](https://github.com/dapr/components-contrib/blob/f125940490d146284b7608a526f5f79d811f973e/secretstores/local/env/envstore.go#L124-L134): it will not allow access to `APP_API_TOKEN` or env vars whose name begins with `DAPR_` (case-insensitive).

## Release Note

RELEASE NOTE: **ADDED** Component metadata can be initialized with values from env vars using `envRef`

ItalyPaleAle:
Updates the Component and HTTPEndpoint CRDs to add a new `envRef` property which allows populating the value of a metadata (components) or header (HTTP endpoints) from an env var.

Fixes #6495

Includes unit tests and E2E tests (as part of the existing "healthcheck" test)


pkg/apis/shared/namevalue.go
ItalyPaleAle, line 1: The reason for this `pkg/apis/shared` code is because when working on this PR I found that there was A LOT of code duplication between Component and HTTPEndpoint, and that also caused a lot of code duplication in `pkg/runtime`.

pkg/runtime/runtime.go
```diff
@@ -2955,17 +2951,51 @@ func (a *DaprRuntime) WaitUntilShutdown() error {
...
+	key = strings.ToUpper(key)
+	switch {
+	case key == "APP_API_TOKEN":
+		return false
+	case strings.HasPrefix(key, "DAPR_"):
+		return false
+	default:
+		return true
+	}
+}
```
ItalyPaleAle, line 2973: This is the same allowlist [implemented in the local.env secret store](https://github.com/dapr/components-contrib/pull/2511). It denies access to env vars in the sidecar container that could be dangerous to expose, such as `DAPR_API_TOKEN`, `DAPR_CERT_KEY`, etc

tools/codegen.mk
```diff
@@ -9,7 +9,9 @@ endif
 # Generate code
 code-generate: controller-gen
```
ItalyPaleAle, line 10: Updated these commands:

- The first command was not working with more recent versions of controller-gen
- The second command was necessary to (re-)generate the `*.deepcopy.go` files

pkg/runtime/runtime.go
artursouza: This is an denylist, not allowlist. The sidecar injector can pass the list of keys from `env` as `DAPR_ALLOWED_ENV_KEYS` that this code can parse and use as an allowlist if it is set.

artursouza (changes requested):

pkg/runtime/runtime.go
ItalyPaleAle: I can add an allowlist but this denylist is a requirement for us for security reasons.

ItalyPaleAle: Should have clarified that the denylist is needed for standalone mode where we can't have an injector adding an allowlist.

pkg/apis/shared/namevalue.go
```diff
@@ -0,0 +1,76 @@
...
+	Name string `json:"name"`
+	// Value of the property, in plaintext.
+	//+optional
+	Value DynamicValue `json:"value,omitempty"`
+	// SecretKeyRef is the reference of a value in a secret store component.
+	//+optional
+	SecretKeyRef SecretKeyRef `json:"secretKeyRef,omitempty"`
+	// EnvRef is the name of an environmental variable to read the value from.
+	//+optional
+	EnvRef string `json:"envRef,omitempty"`
```
JoshVanL, line 36: These 3 should be pointers, which are used to determine which has been selected. We should have validation the 1 and only 1 is non nil.

pkg/apis/shared/namevalue.go
```diff
@@ -0,0 +1,76 @@
...
+}
+
+// HasValue returns true if the NameValuePair has a non-empty value.
+func (nvp NameValuePair) HasValue() bool {
+	return len(nvp.Value.JSON.Raw) > 0
+}
```
JoshVanL, line 42: This should be checking a nil pointer, rather than the presence of string. An empty string should be considered a value.

pkg/runtime/runtime.go
```diff
@@ -2955,17 +2951,51 @@ func (a *DaprRuntime) WaitUntilShutdown() error {
...
+		if !m.HasValue() && m.EnvRef != "" {
+			if isEnvVarAllowed(m.EnvRef) {
+				metadata[i].SetValue([]byte(os.Getenv(m.EnvRef)))
+			} else {
+				log.Warnf("%s %s references an env variable that isn't allowed: %s", resource.Kind(), resource.GetName(), m.EnvRef)
+			}
+			metadata[i].EnvRef = ""
+			updated = true
 			continue
 		}
```
JoshVanL, line 2994: I strongly disagree of an order of operation behavior in our API. We should enforce exactly 1 is defined.

JoshVanL (changes requested):
Added some comments :slightly_smiling_face: 

I'm strongly against adding environment variable support here- env vars are inherently evil in Kube environments imo, as they encourage users to leak secrets (defining secret values inside config), or produce unnecessary churn (a process needs to restart in order to pick up a new environment variable value).

It should _at least_ be made clear in documentation that users should always prefer secret references over any other form of secret data delivery.

pkg/apis/components/v1alpha1/types.go
```diff
@@ -14,11 +14,9 @@ limitations under the License.
 package v1alpha1
 
 import (
-	"strconv"
-
-	v1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
+	"github.com/dapr/dapr/pkg/apis/shared"
```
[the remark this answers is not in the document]
ItalyPaleAle: sure, I can rename that

pkg/apis/shared/namevalue.go
```diff
@@ -0,0 +1,76 @@
...
+type SecretKeyRef struct {
+	// Secret name.
+	Name string `json:"name"`
+	// Field in the secret.
+	//+optional
+	Key string `json:"key"`
+}
+
+// DynamicValue is a dynamic value struct for the component.metadata pair value.
+type DynamicValue struct {
```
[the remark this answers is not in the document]
ItalyPaleAle: I have no idea, but this is what it has always been called, I just moved it to the shared/common file.

pkg/apis/shared/namevalue.go
```diff
@@ -0,0 +1,76 @@
...
+	apiextensionsV1.JSON `json:",inline"`
```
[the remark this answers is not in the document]
ItalyPaleAle: Ditto: I have no idea, just keeping what the code was before.

pkg/runtime/runtime.go
```diff
@@ -2955,17 +2951,51 @@ func (a *DaprRuntime) WaitUntilShutdown() error {
...
+	GetSecretStore() string
+	NameValuePairs() []sharedapi.NameValuePair
+}
+
+func isEnvVarAllowed(key string) bool {
+	key = strings.ToUpper(key)
+	switch {
+	case key == "APP_API_TOKEN":
+		return false
+	case strings.HasPrefix(key, "DAPR_"):
```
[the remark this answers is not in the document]
ItalyPaleAle: I think DAPR_ covers all present cases (aside from APP_API_TOKEN) and possibile future ones for now.

pkg/runtime/runtime.go
```diff
@@ -2955,17 +2951,51 @@ func (a *DaprRuntime) WaitUntilShutdown() error {
...
+	GetName() string
+	GetNamespace() string
+	GetSecretStore() string
+	NameValuePairs() []sharedapi.NameValuePair
+}
+
+func isEnvVarAllowed(key string) bool {
+	key = strings.ToUpper(key)
+	switch {
+	case key == "APP_API_TOKEN":
```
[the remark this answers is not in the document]
ItalyPaleAle: Can you open a separate issue for this?

pkg/apis/shared/namevalue.go
ItalyPaleAle: I don't agree with the use of pointers here because these are all string values (or structs containig string values), so we don't need to distinguish between empty strings and non-present values (we treat empty strings as non-defined because we do not allow empty values).

Also, in K8s mode, both `secretKeyRef` and `value` can be non-empty: when using the K8s secret store, the Operator fetches the value of the secret and sends it to the runtime as well as `secretKeyRef`, so the runtime knows that the value was already fetched from the Operator.

pkg/apis/shared/namevalue.go
ItalyPaleAle: `len()` does a nil-check built-in. With our CRDs (also Configuration etc), we consider empty strings the same as non-present values and we don't have a need to distinguish between the two.

ItalyPaleAle:
Thanks for the review @JoshVanL , I'll address the feedback that can be addressed.

As per this:

> I'm strongly against adding environment variable support here- env vars are inherently evil in Kube environments imo, as they encourage users to leak secrets (defining secret values inside config), or produce unnecessary churn (a process needs to restart in order to pick up a new environment variable value).

Please remember that Dapr can be executed in separate contexts than K8s. This includes (but is not limited to) a standalone binary (including local development), as well as Docker Compose. Those scenarios are currently disadvantages as referencing secrets in a component YAML requires manually initializing a secret store: the complexity actually encourages the even worse practice of including secrets in plain-text.

pkg/apis/shared/namevalue.go
ItalyPaleAle: An empty string isn't a valid `value` either right now. Components expect all strings to be non-empty, and it's a convention used commonly in dapr/dapr and components-contrib.

Although I hear you, I don't think this change is needed now, or at least not in this PR since I'm not changing the existing behavior here.

JoshVanL: :+1: 

FWIW, an empty string value is permissible in Kubernetes Secrets, so there is some inconsistencies in API there.

artursouza (approved):

mukundansundar (approved):

Issue #6035, opened 2023-03-08
Support configuration for SDKs to talk to remote Dapr APIs


artursouza:
Today, all SDKs assume sidecar to be running locally. In addition, they should all allow connection to a remote API with TLS over HTTP or gRPC. SDKs should agree on some common configurations:

0. Validate the SDK can connect to a remove Dapr API endpoint over TLS via the app hardcoding the endpoint and passing it via a constructor method or builder class.
1. ~`DAPR_HTTP_ADDRESS`~ `DAPR_HTTP_ENDPOINT` environment variable should be accepted and parsed by the factory method/class on each SDK to use that endpoint in case of HTTP protocol (e.g. https://dapr.mycompany.com)
2. ~`DAPR_GRPC_ADDRESS`~ `DAPR_GRPC_ENDPOINT` environment variable should be accepted and parsed by the factory method/class on each SDK to use that endpoint in case of GRPC protocol. (e.g. https://dapr-grpc.mycompany.com)
3. ~`DAPR_SSL_CA_PINNING_PATH` environment variable containing the CA certs accepted for SSL pinning.~

SDKs should accept (1) or (2) (both if both protocols are supported). Also, (1) and (2) should take precedence over `DAPR_SIDECAR_IP` + (`DAPR_HTTP_PORT` or `DAPR_GRPC_PORT`) when `DAPR_SIDECAR_IP` is supported.

Assessment on each SDK:

* Java SDK
  * Does not provide a way to override the endpoint apart from IP/hostname and port.
  * `DAPR_SIDECAR_IP` env var already exists and is used in conjunction with `DAPR_HTTP_PORT` or `DAPR_GRPC_PORT`. See https://github.com/dapr/java-sdk/blob/806ddc00bd4e5fb1531c3746ccc2ae295043b59b/sdk/src/main/java/io/dapr/config/Properties.java
* .Net SDK
  * Offers a way to override grpc and http endpoints in builder class. See https://github.com/dapr/dotnet-sdk/blob/f42b690f4c67e6bb4209932f660c46a96d0b0457/src/Dapr.Client/DaprClientBuilder.cs#L33
  * Defaults to 127.0.0.1 and user can override via code. See https://github.com/search?q=repo%3Adapr%2Fdotnet-sdk%20httpEndpoint&type=code

Update: 
SSL pinning is out of scope for this issue.

yaron2:
[...]

PR #6512, opened 2023-06-14
Inject APP_PROTOCOL in app containers

ItalyPaleAle:
Changes the Dapr Injector to inject the APP_PROTOCOL variable into the app containers. This will contain the value of the (computed) `--app-protocol` flag, such as "http", "grpc", "https", "grpcs", or "h2c".

This can be used by SDKs (and users) to determine if they need to start a server with TLS or with H2C support, for example. It may be possible for the SDKs to use this to determine if they should use HTTP or gRPC too, possibly.

See: https://github.com/dapr/dapr/issues/6035#issuecomment-1590624276

### DIFF
--- a/pkg/injector/sidecar/consts.go
+++ b/pkg/injector/sidecar/consts.go
@@ -31,9 +31,16 @@ const (
 	SidecarMetricsEnabledLabel     = "dapr.io/metrics-enabled"
 	APIVersionV1                   = "v1.0"
 	UnixDomainSocketVolume         = "dapr-unix-domain-socket" // Name of the Unix domain socket volume.
+	UserContainerAppProtocolName   = "APP_PROTOCOL"            // Name of the variable exposed to the app containing the app protocol.
 	UserContainerDaprHTTPPortName  = "DAPR_HTTP_PORT"          // Name of the variable exposed to the app containing the Dapr HTTP port.
 	UserContainerDaprGRPCPortName  = "DAPR_GRPC_PORT"          // Name of the variable exposed to the app containing the Dapr gRPC port.
 	PatchPathLabels                = "/metadata/labels"
 	TokenVolumeKubernetesMountPath = "/var/run/secrets/dapr.io/sentrytoken" /* #nosec */ // Mount path for the Kubernetes service account volume with the sentry token.
 	TokenVolumeName                = "dapr-identity-token"                  /* #nosec */ // Name of the volume with the service account token for daprd.
+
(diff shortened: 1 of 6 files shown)


pkg/injector/sidecar/consts.go
```diff
@@ -31,9 +31,16 @@ const (
 	SidecarMetricsEnabledLabel     = "dapr.io/metrics-enabled"
 	APIVersionV1                   = "v1.0"
 	UnixDomainSocketVolume         = "dapr-unix-domain-socket" // Name of the Unix domain socket volume.
+	UserContainerAppProtocolName   = "APP_PROTOCOL"            // Name of the variable exposed to the app containing the app protocol.
```
JoshVanL, line 34:
```suggestion
	UserContainerAppProtocolName   = "DAPR_APP_PROTOCOL"            // Name of the variable exposed to the app containing the app protocol.
```

JoshVanL (changes requested):

pkg/injector/sidecar/consts.go
ItalyPaleAle: We use `DAPR_*` for env vars that are about the Dapr container, such as "DAPR_HTTP_PORT" -> the port Dapr's HTTP server is listening on. This instead is about the app. We already have "APP_CHANNEL_ADDRESS", "APP_ID", "APP_PORT".

JoshVanL: All of these are still Dapr related, and therefore must have a `DAPR` prefix. Makes sense to have the `DAPR_APP` namespace for apps using Dapr. We should deprecate these other `APP_*` env vars, in favor of `DAPR_APP_*` imo.

In my opinion, not namespacing env vars is quite hostile/bad mannered to users and generally breaks convention, especially with `APP` being so generic.

JoshVanL: Sounds good to me :+1:

yaron2 (approved):

ItalyPaleAle:
@yaron2 why did you change the title to `DAPR_APP_PROTOCOL`? The env var is `APP_PROTOCOL` because `DAPR_*` things are read by the sidecar (like `APP_PORT`, `APP_API_TOKEN`, etc)

yaron2:
> @yaron2 why did you change the title to `DAPR_APP_PROTOCOL`? The env var is `APP_PROTOCOL` because `DAPR_*` things are read by the sidecar (like `APP_PORT`, `APP_API_TOKEN`, etc)

Misread your discussion with @JoshVanL

Issue #6399, opened 2023-05-24
[Workflows] Potential memory leak


olitomlinson:
cc @yaron2 @cgillum 

As per [discord](https://discord.com/channels/778680217417809931/1075836407156850698/1110886297221992508), Memory hits between 300-400mb and never releases after all workflows have completed.

---

**my local environment**
- Mac OS - M1
- Docker Desktop 4.19.0 (106363)

**Can reproduce this on the following dapr sidecar versions**

`1.11.0-rc.3`
(not tried earlier versions)

**client libraries**

    <PackageReference Include="Dapr.Client" Version="1.10.0" />
    <PackageReference Include="Dapr.Workflow" Version="1.10.0" />`

**Repro steps**

1. pull repo https://github.com/olitomlinson/dapr-workflow-examples
2. `docker compose build`
3. `docker compose up`
4. Use insomnia/postman/whatever :
`POST http://localhost:5112/start?count=2000&async=true`<- make this request just once to start 2000 workflows
5. Takes approx 20 minutes to start and complete all workflows.
6. Observe the dapr sidecar consume lots of memory, most of which doesn't get released

PR #6523, opened 2023-06-15
Fixed goroutine leak in reminders and timers

### LINKED ISSUE #6503: [Actor Reminders] Potential memory leak
cc @yaron2 @cgillum 

As per our chats on Discord, I believe there is another memory leak in dapr workflows or the underlying subsystems (reminders  subsystem has been mentioned as a possible root cause)

For example, on start-up, the idling sidecar is taking up 26mb, however after running the below repro, the mem sits around 104mb.

After 5 hours of observation, this has not dropped, which leads me to believe there is another memory leak.

---

**my local environment**
- Mac OS - M1
- Docker Desktop 4.19.0 (106363)

**Can reproduce this on the following dapr sidecar versions**

`1.11.2`

**client libraries**

    <PackageReference Include="Dapr.Client" Version="1.11.0" />
    <PackageReference Include="Dapr.Workflow" Version="1.11.0" />`

**Repro steps**

1. pull repo https://github.com/olitomlinson/dapr-workflow-examples
2. `docker compose build`
3. `docker compose up`
4. Use insomnia/postman/whatever :
`POST http://localhost:5112/start?count=2000&async=true&runId={runId}`<- make this request just once to start 2000 workflows. **Replace {runId} with a unique string for each time you run this request, i.e. `001`, `002,`, `003` etc**
5. Takes approx 10-20 minutes to start and complete all workflows.
6. Observe the dapr sidecar consume memory, most of which doesn't get released.

### LINKED ISSUE #6517: Runtime memory leak using Actor timers
## In what area(s)?
/area runtime

## What version of Dapr?
1.10.7

## Expected Behavior

Actor timers should not put pressure on the Dapr sidecar.

## Actual Behavior

The actor app's Dapr sidecar continuously grows in memory (see graph below at 23.5 GiB) until it is eventually killed. There is only one actor per app with a single timer however the `RegisterTimerAsync` method is called periodically with the same actor name in order to achieve the effect of a timer with a "floating window".

image

Instantiate an actor and then continuously call `RegisterTimerAsync` method with the actor name to watch how its memory behaves.

## Release Note

RELEASE NOTE: **FIX** Runtime Memory leak with Actor reminders.

ItalyPaleAle:
Fixes #6517 
Fixes #6503 (assuming there aren't other leaks)

Fixes the goroutine leak that was also causing memory to leak when using actor reminders and timers


artursouza (commented):
Can we have a test to repro this issue?

ItalyPaleAle:
A test would be nice. I am not sure how to approach writing a test for this, do you have any suggestion?

pkg/actors/actors.go
```diff
@@ -979,9 +979,6 @@ func (a *actorsRuntime) startReminder(reminder *reminders.Reminder, stopChannel
 				break L
 			}
 
-			if nextTimer.Stop() {
```
ItalyPaleAle, line 982: Reviewers: this was deleted because when we hit this stage, the reminder has already fired. Trying to drain it again would cause a deadlock.

JoshVanL:
> A test would be nice. I am not sure how to approach writing a test for this, do you have any suggestion?

@ItalyPaleAle We can have an integration test that does similar to what Artur did here https://github.com/dapr/dapr/issues/6517#issuecomment-1592262158

ItalyPaleAle:
> > A test would be nice. I am not sure how to approach writing a test for this, do you have any suggestion?
> 
> @ItalyPaleAle We can have an integration test that does similar to what Artur did here [#6517 (comment)](https://github.com/dapr/dapr/issues/6517#issuecomment-1592262158)

My biggest problem is not how to run the test, but how to write a test that _reliably_ determines the _absence of the effect_. I can't just have a reminder or timer execute and count the goroutines #, because that can be affected by other things (there are background tasks that create goroutines randomly)

artursouza:
> > > A test would be nice. I am not sure how to approach writing a test for this, do you have any suggestion?
> > 
> > 
> > @ItalyPaleAle We can have an integration test that does similar to what Artur did here [#6517 (comment)](https://github.com/dapr/dapr/issues/6517#issuecomment-1592262158)
> 
> My biggest problem is not how to run the test, but how to write a test that _reliably_ determines the _absence of the effect_. I can't just have a reminder or timer execute and count the goroutines #, because that can be affected by other things (there are background tasks that create goroutines randomly)

It is OK. Tests can fail due to another root cause later on. It is testing the symptom of go routines leaks by re-registering the same timer over and over again. If the root cause later is from somewhere else, the test can fail and it is OK. Every end-to-end test works that way - it checks for correct behavior not one particular root cause.

artursouza:
> > > > A test would be nice. I am not sure how to approach writing a test for this, do you have any suggestion?
> > > 
> > > 
> > > @ItalyPaleAle We can have an integration test that does similar to what Artur did here [#6517 (comment)](https://github.com/dapr/dapr/issues/6517#issuecomment-1592262158)
> > 
> > 
> > My biggest problem is not how to run the test, but how to write a test that _reliably_ determines the _absence of the effect_. I can't just have a reminder or timer execute and count the goroutines #, because that can be affected by other things (there are background tasks that create goroutines randomly)
> 
> It is OK. Tests can fail due to another root cause later on. It is testing the symptom of go routines leaks by re-registering the same timer over and over again. If the root cause later is from somewhere else, the test can fail and it is OK. Every end-to-end test works that way - it checks for correct behavior not one particular root cause.

Also, in this case the number of go routines grows linear with the number of requests - it is hard to miss that in the test. Keep running the requests for some time and see if there are 1K go routines - it is a leak any time of the day when it happens.

yaron2:
I can confirm this PR does not fix the memory leak issue. Running the steps outlined [here] (https://github.com/dapr/dapr/issues/6399#issue-1724542106) which helped drive [this](https://github.com/microsoft/durabletask-go/pull/20) memory leak fix shows the Dapr process starting at ~30Mb when idle and stays at ~85mb - ~95mb after the execution is completed.

ItalyPaleAle:
Just pushed some more changes:

1. There was still some stuff leaking. The original patch fixed leaks when a reminder/timer was deleted, but not when it fired. Fixed all situations, including reminders/timers firing or reaching TTL.
2. Added a unit test that checks goroutines before and after, validating reminders that are deleted, that fire (with no more repeats), and that reach TTL.

Hopefully the test doesn't turn out to be flaky... I wrote it as unit test because it allows having more control over what goroutines are created (E2E tests, besides being a lot more complex due to the need of parsing prometheus output, have a lot more going on and number of goroutines there is less stable) and because integration tests don't have placement service and don't have apps running.

ItalyPaleAle:
> I can confirm this PR does not fix the memory leak issue. Running the steps outlined [here] ([#6399 (comment)](https://github.com/dapr/dapr/issues/6399#issue-1724542106)) which helped drive [this](https://github.com/microsoft/durabletask-go/pull/20) memory leak fix shows the Dapr process starting at ~30Mb when idle and stays at ~85mb - ~95mb after the execution is completed.

@yaron2 please try with the latest fix, i don't think it was complete.

However please note that measuring memory that way (from Docker) isn't accurate because of the GC:

1. Memory could be in use even if the objects are not in use - those objects will stick around until the next time the GC runs
2. The GC doesn't return all memory to the OS after it runs

See: https://povilasv.me/prometheus-go-metrics/

Best would be to just check the number of active goroutines, since memory can't really be leaked in Go anyways.

yaron2:
> > I can confirm this PR does not fix the memory leak issue. Running the steps outlined [here] ([#6399 (comment)](https://github.com/dapr/dapr/issues/6399#issue-1724542106)) which helped drive [this](https://github.com/microsoft/durabletask-go/pull/20) memory leak fix shows the Dapr process starting at ~30Mb when idle and stays at ~85mb - ~95mb after the execution is completed.
> 
> @yaron2 please try with the latest fix, i don't think it was complete.
> 
> However please note that measuring memory that way (from Docker) isn't accurate because of the GC:
> 
> 1. Memory could be in use even if the objects are not in use - those objects will stick around until the next time the GC runs
> 2. The GC doesn't return all memory to the OS after it runs
> 
> See: https://povilasv.me/prometheus-go-metrics/
> 
> Best would be to just check the number of active goroutines, since memory can't really be leaked in Go anyways.

I will check again with changes you're making to this PR, but surely I can't agree with the statement that Go can't have memory leaks, and I can certainly see memory being returned. With the fix mentioned earlier, the memory reported by Docker hanged around at ~300Mb after the same test run, and after the durabletask-go fix it consistently dropped to ~100mb. Furthermore, memory snapshots showed that objects were indeed released from memory (timer Tickers) after the durabletask-go fix was applied, as opposed to snapshots before the fix that showed them still being in use.

ItalyPaleAle:
Memory leaks in Go are only possible in very limited cases, including incorrect use of global variables, or goroutine leaks. The issue with durabletask was a goroutine leak too.

As for how and when memory is returned, I recommend reading these:
- https://tip.golang.org/doc/gc-guide
- https://blog.twitch.tv/en/2019/04/10/go-memory-ballast-how-i-learnt-to-stop-worrying-and-love-the-heap/

yaron2:
> Memory leaks in Go are only possible in very limited cases, including incorrect use of global variables, or goroutine leaks. The issue with durabletask was a goroutine leak too.
> 
> As for how and when memory is returned, I recommend reading these:
> 
> * https://tip.golang.org/doc/gc-guide
> * https://blog.twitch.tv/en/2019/04/10/go-memory-ballast-how-i-learnt-to-stop-worrying-and-love-the-heap/

I read these resources while fixing the durabletask-go issue. Whether this memory leak is caused by a goroutine leak or something else, it is clear based on the tests that this PR doesn't fix the issue, so we need to continue looking.

Here are some good reads of a memory leak example with `time`:

* https://www.arangodb.com/2020/09/a-story-of-a-memory-leak-in-go-how-to-properly-use-time-after/
* https://www.arangodb.com/2020/09/a-story-of-a-memory-leak-in-go-how-to-properly-use-time-after/

artursouza:
> Memory leaks in Go are only possible in very limited cases, including incorrect use of global variables, or goroutine leaks. The issue with durabletask was a goroutine leak too.
> 
> As for how and when memory is returned, I recommend reading these:
> 
> * https://tip.golang.org/doc/gc-guide
> * https://blog.twitch.tv/en/2019/04/10/go-memory-ballast-how-i-learnt-to-stop-worrying-and-love-the-heap/

Leak does not mean memory not used by the program, it means that there is a reference using the memory and that is stopping GC from returning it. This is similar problem in Java where "Java does not have memory leak" but it does if your program keeps references to object that will never be accessed again.

ItalyPaleAle:
> I read these resources while fixing the durabletask-go issue. Whether this memory leak is caused by a goroutine leak or something else, it is clear based on the tests that this PR doesn't fix the issue, so we need to continue looking.

Please test with the latest changes. I reckon it wasn't complete when you posted your comment above (only fixed goroutine leaks when reminders were deleted, but not when they fired). It should be fine now, as validated by the tests.

artursouza:
@ItalyPaleAle can you have a E2E test to repro the mem leak? This way, it is easy to test with and without this change.

yaron2:
> > I read these resources while fixing the durabletask-go issue. Whether this memory leak is caused by a goroutine leak or something else, it is clear based on the tests that this PR doesn't fix the issue, so we need to continue looking.
> 
> Please test with the latest changes. I reckon it wasn't complete when you posted your comment above (only fixed goroutine leaks when reminders were deleted, but not when they fired). It should be fine now, as validated by the tests.

Tested with latest changes, the memory leak still exists to the same effect.

ItalyPaleAle:
Could you share the code used for testing please?

ItalyPaleAle:
Could you share the code used for testing please? This way I can look into what's still going on.

yaron2:
> Could you share the code used for testing please? This way I can look into what's still going on.

Sure, already did [here](https://github.com/dapr/dapr/pull/6523#issuecomment-1595346820)

ItalyPaleAle:
Ok, it's fixed now. Timers were good, but reminders needed a `nextTimer = nil` in a few more places. For some reason, calling `Stop()` on a timer that has already fired returns `false`, which normally means the channel needs to be drained, but it then blocks forever causing a goroutine leak. So the variable has to be nullified instead.

I used the sample above, but also enabled prometheus and pprof. I can see now that the # of goroutines goes back to where it was before it started.

artursouza:
Awesome. Is it possible to have an E2E test for this?

yaron2:
After the latest changes, the memory is still held at > 105Mb when all reminders finish firing. This may have reduced the number of goroutines, but there are still unclaimed objects in memory.

artursouza (approved):

Issue #6534, opened 2023-06-17
Actor Heartbeat Metric


antontroshin:
## In what area(s)?

> /area runtime

> /area operator

/area placement

> /area docs

> /area test-and-release

## Problem
The lack of visibility into actors' placements, reduces the ability to diagnose resource utilization, hotspot, and other issues in k8s clusters.
Without knowing where actors or specific actor types are running, it's hard to monitor the allocation and manage/tune resources in dynamic environments.
Dapr Placement service holds the placement data where actors are hosted, but the table is not accessible for outside use.

The information in the `Host` object includes the following:
```
name: 10.244.1.149:50002
load: 1
entities: "MyActor"
id: "my-actor-server"
```
While host information may point to the pod, it makes it more complex to map to the pod name when most of the metrics use pod names.
Having a pod name helps to troubleshoot as it can be seen almost anywhere unlike host IP.

## Possible use cases
- See when and where actors and/or actor types reporting and track if they are alive
- Mapping of actor types to pods and resource usage in metric dashboards
- Alerts on specific actor types exceeding the resources allowed
- Find hotspot issues for actor types and infrastructure bottlenecks hosting these actors
- Tracking actor's hearbeat on the timeline and be able to map where actors were running

Additional benefit, aggregating a metric will allow building placement table without complexity of exposing API and doesn't require use of internal gRPC service that requires authentication as was discovered in PR https://github.com/dapr/dapr/pull/6212 which was reverted.
Provides alternative fix for https://github.com/dapr/dapr/issues/6033

## Proposed solution
Create the Actor Heartbeat metric when receiving Host in PlacementServer and report for each actor type reported by the actor.

Metric name: `dapr_placement_actor_heartbeat_timestamp`
Tags: `app_id`, `actor_type`, `host_name`, `pod_name`
Value: timestamp in seconds since epoch

## POC
Example of metric reported by placement service
```
dapr_placement_actor_heartbeat_timestamp{actor_type="PingPong",app_id="server-ping-pong",host_name="10.244.1.165:50002",pod_name="server-ping-pong-648f467c5b-6bbm8"} 1686792312
```
Example Grafana panel with actor placements
<img width="1306" alt="Screen Shot 2023-06-16 at 2 29 10 PM" src="https://github.com/dapr/dapr/assets/18515134/9e660fb2-56df-435a-b527-f3e8faaa8959">

## Release Note

[...]

PR #6535, opened 2023-06-17
Add actor heartbeat metric in the placement report for each actor type

antontroshin:
# Description

Add a new heartbeat metric for actors. See the reference issue below.

## Issue reference

Please reference the issue this PR will close: #6534

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [ ] Unit tests passing


dapr/proto/placement/v1/placement.proto
```diff
@@ -46,4 +46,5 @@ message Host {
   int64 load = 3;
   repeated string entities = 4;
   string id = 5;
+  string pod = 6;
```
artursouza, line 49: Is this optional? Can this be k8s agnostic, maybe hostInfo?

pkg/placement/placement.go
```diff
@@ -215,6 +216,10 @@ func (p *Service) ReportDaprStatus(stream placementv1pb.Placement_ReportDaprStat
 				continue
 			}
 
+			for _, entity := range req.Entities {
+				monitoring.RecordActorHeartbeat(req.Id, entity, req.Name, req.Pod, p.clock.Now())
```
artursouza, line 220: How would this work in standalone mode?

artursouza (changes requested):

pkg/placement/placement.go
yaron2: In self-hosted mode the pod name should just be empty, but the metric should still be tracked.

antontroshin: Tested this in self-hosted mode and the metric reported like this in `placement-service:9090` metrics API
```
dapr_placement_actor_heartbeat_timestamp{actor_type="OrderActor",app_id="actor-service",host_name="10.1.2.10:54656",pod_name=""} 1.687829089e+09
```

dapr/proto/placement/v1/placement.proto
antontroshin: Yes, this is an optional field and will be filled only in k8s mode.
Since this will be empty in self-hosted mode, maybe it makes sense to leave the `pod` as a name making it a bit more obvious that it is only expected in k8s? WDYT?

pkg/placement/placement.go
artursouza: OK

dapr/proto/placement/v1/placement.proto
artursouza: OK.

artursouza (dismissed):

artursouza (dismissed):

yaron2 (approved):

Issue #6033, opened 2023-03-07
API to retrieve placement table from placement service


artursouza:
## In what area(s)?

> /area runtime

> /area operator

> /area placement

> /area docs

> /area test-and-release

## Describe the feature
Offer HTTP or gRPC API for placement service to expose the placement table - it can be helpful for debugging but also allow other tools (Dashboard) to extract and present for information about Actors.

## Release Note

RELEASE NOTE:

artursouza:
Reopened the issue since the authenticated endpoint cannot be used by users. We need to decide if Placement service should offer an open metadata endpoint, like the sidecar does.

artursouza:
@Taction Can you look into the docs issue next? https://github.com/dapr/docs/issues/3364

Taction:
@artursouza Sure, I'll start to finish the doc.

PR #6536, opened 2023-06-17
[placement] Add table endpoint

Taction:
# Description

Add the `tables` endpoint to get placement tables to the health check http server(the sidecar metadata endpoint is also added to the health check server).

As this behavior has not been discussed, I'll backfill tests if acceptable.

## Issue reference

Please reference the issue this PR will close: #6033 

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing


artursouza:
That is a fine solution. I would have an environment variable to disable that endpoint for users that think it exposes information in an unauthenticated endpoint.

pkg/health/server.go
```diff
@@ -33,16 +34,53 @@ type Server interface {
...
+type RouterOptions func() (string, http.Handler)
+
+func NewRouterOptions(path string, handler http.Handler) RouterOptions {
+	return func() (string, http.Handler) {
+		return path, handler
+	}
+}
+
+func NewJsonDataRouterOptions[T any](path string, getter func() (T, error)) RouterOptions {
+	return NewRouterOptions(path, http.HandlerFunc(func(writer http.ResponseWriter, request *http.Request) {
```
ItalyPaleAle, line 51: IMHO the /healthz endpoint should not have any data in it, because it's meant to just respond with a 2xx status code indicating everything is ok. It is invoked multiple times per minute on K8s, so it needs to be as fast as possible.

This is especially important in this case, as placement tables can have a lot of data, so this endpoint is very slow (plus, it allocates a lot of memory due to copying the table).

Can this be implemented as a separate endpoint rather than /healthz? For example `GET /placement/state` or anything else.

pkg/health/server.go
```diff
@@ -33,16 +34,53 @@ type Server interface {
...
+		data, err := getter()
+		if err != nil {
+			writer.WriteHeader(http.StatusInternalServerError)
+			writer.Write([]byte(err.Error()))
+			return
+		}
+		bytes, err := json.Marshal(data)
```
ItalyPaleAle, line 58: Let's use streams please to avoid buffering (large) data in memory.

```go
writer.Header().Set("Content-Type", "application/json")
err := json.NewEncoder(writer).Encode(data)
// Handle error
```

ItalyPaleAle (changes requested):

pkg/health/server.go
Taction: > IMHO the /healthz endpoint should not have any data in it, because it's meant to just respond with a 2xx status code indicating everything is ok. It is invoked multiple times per minute on K8s, so it needs to be as fast as possible.

Yes, this totally makes sense! And in fact, it is indeed implemented as a separate endpoint, the variable `path` in this function is the endpoint for the handler. It just uses the same port as health check.
The endpoint currently is set to `/tables`, as it is included in the placement server, I haven't add a `/placement` prefix.

pkg/health/server.go
Taction: Thanks, Done.

pkg/health/server.go
ItalyPaleAle: Ohh, I missed that it was a separate endpoint.

Ok, makes sense. However I would like one of @dapr/maintainers-dapr to confirm what the API endpoint should be. I think a prefix is needed.

ItalyPaleAle:
@Taction LGTM, but:

1. Looks like there's a linter issue 🥲
2. Just want confirmation from a maintainer on what the endpoint should be

pkg/health/server.go
Taction: I am ok with `/placement/state`, changed,  use this endpoint unless there are further ideas on that. Tips, I am also ok with something like `/placement/actor` or `/actor/state`.

Taction:
> That is a fine solution. I would have an environment variable to disable that endpoint for users that think it exposes information in an unauthenticated endpoint.

Add `metadata-enable` command line flag and `DAPR_PLACEMENT_METADATA_ENABLED` env to define if expose this endpoint(flags take priority over env variables).

cmd/placement/config.go
```diff
@@ -76,6 +80,7 @@ func newConfig() *config {
 	flag.IntVar(&cfg.healthzPort, "healthz-port", cfg.healthzPort, "sets the HTTP port for the healthz server")
 	flag.StringVar(&cfg.certChainPath, "certchain", cfg.certChainPath, "Path to the credentials directory holding the cert chain")
 	flag.BoolVar(&cfg.tlsEnabled, "tls-enabled", cfg.tlsEnabled, "Should TLS be enabled for the placement gRPC server")
+	flag.BoolVar(&cfg.metadataEnabled, "metadata-enable", cfg.metadataEnabled, "Expose the placement tables, using the same HTTP port as healthz")
```
artursouza, line 83: nit: metadata-enabled

Taction: Sorry for the delay. Done, also resolved the conflict.

pkg/placement/tables.go
```diff
@@ -0,0 +1,46 @@
...
+type PlacementTables struct {
+	HostList     []*HostInfo `json:"host_list,omitempty"`
+	TableVersion uint64      `json:"table_version,omitempty"`
+}
+type HostInfo struct {
+	Name      string   `json:"name,omitempty"`
+	AppID     string   `json:"app_id,omitempty"`
+	Entities  []string `json:"entities,omitempty"`
+	UpdatedAt int64    `json:"updated_at,omitempty"`
+}
```
ItalyPaleAle, line 25: Please use camelCase for JSON keys. We always do this in Dapr.

```suggestion
type PlacementTables struct {
	HostList     []*HostInfo `json:"hostList,omitempty"`
	TableVersion uint64      `json:"tableVersion,omitempty"`
}
type HostInfo struct {
	Name      string   `json:"name,omitempty"`
	AppID     string   `json:"appId,omitempty"`
	Entities  []string `json:"entities,omitempty"`
	UpdatedAt int64    `json:"updatedAt,omitempty"`
}
```

pkg/placement/tables.go
```diff
@@ -0,0 +1,46 @@
...
+
+// GetPlacementTables returns the current placement host infos.
+func (p *Service) GetPlacementTables() (*PlacementTables, error) {
+	m := p.raftNode.FSM().State().Members()
+	version := p.raftNode.FSM().State().TableGeneration()
+	response := &PlacementTables{
+		TableVersion: version,
+	}
+	members := make([]*HostInfo, 0, len(m))
```
ItalyPaleAle, line 34: Can you change this to not be a pointer? Since we don't need null values here, I believe it should be more efficient


```suggestion
	members := make([]HostInfo, 0, len(m))
```

pkg/health/server.go
```diff
@@ -33,16 +34,56 @@ type Server interface {
...
+
+type RouterOptions func(log logger.Logger) (string, http.Handler)
+
+func NewRouterOptions(path string, handler http.Handler) RouterOptions {
+	return func(log logger.Logger) (string, http.Handler) {
+		return path, handler
+	}
+}
+
+func NewJSONDataRouterOptions[T any](path string, getter func() (T, error)) RouterOptions {
```
ItalyPaleAle, line 50: I don't know if a generic function is really needed here, since we are not treating the data any differently depending on the type. It may be more efficient to not use generics:


```suggestion
func NewJSONDataRouterOptions(path string, getter func() (any, error)) RouterOptions {
```

pkg/health/server.go
```diff
@@ -33,16 +34,56 @@ type Server interface {
...
+	return func(log logger.Logger) (string, http.Handler) {
+		return path, http.HandlerFunc(func(writer http.ResponseWriter, request *http.Request) {
+			data, err := getter()
+			if err != nil {
+				writer.WriteHeader(http.StatusInternalServerError)
+				writer.Write([]byte(err.Error()))
+				return
+			}
+			writer.Header().Set("Content-Type", "application/json")
+			writer.WriteHeader(http.StatusOK)
```
ItalyPaleAle, line 60: Remove this, which is implied when the first byte is written. Otherwise, in case of errors, the `StatusInternalError` will never be sent


```suggestion
```

cmd/placement/options/options.go
```diff
@@ -65,6 +69,7 @@ func New() *Options {
 	flag.IntVar(&opts.HealthzPort, "healthz-port", defaultHealthzPort, "sets the HTTP port for the healthz server")
 	flag.StringVar(&opts.CertChainPath, "certchain", defaultCredentialsPath, "Path to the credentials directory holding the cert chain")
 	flag.BoolVar(&opts.TLSEnabled, "tls-enabled", false, "Should TLS be enabled for the placement gRPC server")
+	flag.BoolVar(&opts.MetadataEnabled, "metadata-enabled", opts.MetadataEnabled, "Expose the placement tables, using the same HTTP port as healthz")
```
ItalyPaleAle, line 72:
```suggestion
	flag.BoolVar(&opts.MetadataEnabled, "metadata-enabled", opts.MetadataEnabled, "Expose the placement tables on the healthz server")
```

cmd/placement/options/options.go
```diff
@@ -77,6 +82,12 @@ func New() *Options {
 	opts.Metrics = metrics.DefaultMetricOptions()
 	opts.Metrics.AttachCmdFlags(flag.StringVar, flag.BoolVar)
 
+	// parse env variables before parsing flags, so the flags takes priority over env variables
+	metadataEnable := os.Getenv(envMetadataEnabled)
+	if len(metadataEnable) > 0 {
+		opts.MetadataEnabled = utils.IsTruthy(metadataEnable)
+	}
```
ItalyPaleAle, line 89: An empty value is considered as false, and the default value for `opts.MetadataEnabled` is false.

```suggestion
	opts.MetadataEnabled = utils.IsTruthy(os.Getenv(envMetadataEnabled))
```

ItalyPaleAle (changes requested):

pkg/health/server.go
Taction: IMHO, generic function is needed here. `func() Any-specific-type` have a different signature than `func() any`. For example, this couldn't be compiled `var f func() any = func() int { return 1 }`.

For performance, we are not using interface constrain here, I think the go compiler would be clever enough monomorphizing the code -- generates a copy of the generic function for the specific data type it is invoked with.

ItalyPaleAle: Ok I see what you mean now.

ItalyPaleAle (approved):

artursouza (approved):

PR #6537, opened 2023-06-17
Add Dapr runtime version and enabled features to Metadata API

shubham1172:
# Description

Co-authored-by: @ItalyPaleAle 

Add a runtime version to metadata root. It is kept in extended metadata for back-compat but should be deprecated later.
This also adds an 'enabledFeatures' key with list of preview features enabled.

```bash
daprd --app-id test --app-port 3001 --app-protocol grpc --dapr-http-port 3501 -- sleep 1000
curl localhost:3501/v1.0/metadata | jq .
```
```json
{
  "id": "test",
  "runtimeVersion": "edge",
  "enabledFeatures": [
    "ServiceInvocationStreaming"
  ],
  "extended": {
    "daprRuntimeVersion": "edge"
  },
  "appConnectionProperties": {
    "port": 3001,
    "protocol": "grpc",
    "channelAddress": "127.0.0.1"
  },
}
```

## Issue reference

Please reference the issue this PR will close: #6528

## Checklist

Please make sure you've  completed the relevant tasks for this PR, out of the following list:

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Unit tests passing
* [x] End-to-end tests passing
* [x] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: https://github.com/dapr/docs/issues/3557
* [ ] Specification has been updated / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_
* [ ] Provided sample for the feature / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_

PR #6546, opened 2023-06-19
E2E tests for the crypto building block

### LINKED ISSUE #6259: [Crypto] Add E2E tests for Crypto building block
Needed to complete the building block.

Blocked until dapr/go-sdk#387 is merged

ItalyPaleAle:
Fixes #6259 

This PR adds E2E tests for the crypto building block, for both high-level and subtle APIs.

1. Tests for the subtle APIs run only if the binary was compiled with the `subtlecrypto` Go build tag. This is made possible by adding a "fake" feature flag when the binary is built with that tag, which is then read from the runtime's metadata APIs
2. Tests on KIND run with subtle crypto APIs disabled; they are enabled on AKS
3. The reason why this PR was opened from dapr/dapr and not my fork is to be able to test the changes to the AKS pipeline, which otherwise would not be testable
4. Includes an updated components-contrib to fix an issue with Azure Key Vault

**Tests on AKS with the Azure Key Vault component and subtle-crypto APIs enabled**

Please do not run /ok-to-test in this PR as it wouldn't work due to changes on the pipeline

- https://github.com/dapr/dapr/actions/runs/5336118282 (Linux passed; Windows failed on unrelated tests)
- https://github.com/dapr/dapr/actions/runs/5336363952 (Linux passed; Windows failed on unrelated tests)


Makefile
```diff
@@ -27,11 +27,17 @@ BINARIES    ?= daprd placement operator injector sentry
 HA_MODE     ?= false
 # Force in-memory log for placement
 FORCE_INMEM ?= true
-# Go's build tags:
+# Dapr sidecar "flavor" build tag:
 # allcomponents - (default) includes all components in Dapr sidecar
 # stablecomponents - includes all stable components in Dapr sidecar
 DAPR_SIDECAR_FLAVOR ?= allcomponents
-DAPR_GO_BUILD_TAGS = $(DAPR_SIDECAR_FLAVOR)
+# Additional build tags
```
ItalyPaleAle, line 34: This change allows adding additional build flags besides the sidecar flavor

.github/workflows/dapr-test.yml
```diff
@@ -55,8 +55,13 @@ env:
   AZURE_ARM_REGIONS: "eastus"
   # Container registry where to cache e2e test images
   DAPR_CACHE_REGISTRY: "dapre2eacr.azurecr.io"
+  # Name of the Azure Key Vault resource used in tests
+  # The credentials defined in AZURE_CREDENTIALS must have permissions to perform operations in this vault
+  AZURE_KEY_VAULT_NAME: "dapre2ekv"
```
ItalyPaleAle, line 60: For Azure, I have pre-provisioned a Key Vault in our test subscription, which is accessible using the same Azure AD application we use to deploy resources. Deploying a new KV on each test is complex due to the need to also generate keys and manage permissions via Azure AD, so there's a single one that all tests on AKS use.

This should not be a problem, since capacity isn't a concern even if we run multiple tests in parallel. Additionally, Key Vault's pricing is essentially just per API call, so keeping a Key Vault pre-provisioned incurs only marginal costs.

pkg/grpc/universalapi/api_subtlecrypto_subtlecrypto.go
```diff
@@ -28,13 +28,21 @@ import (
...
 	runtimev1pb "github.com/dapr/dapr/pkg/proto/runtime/v1"
 	"github.com/dapr/dapr/pkg/resiliency"
 	kitCrypto "github.com/dapr/kit/crypto"
 )
 
+// TODO: Remove this when the build tag is removed
+func init() {
+	// Add a fake feature "_SubtleCrypto" which is not a real preview feature
+	// This is then exposed in the sidecar's metadata endpoint so tests (and users) can query for this.
+	buildinfo.AddFeature("_SubtleCrypto")
```
ItalyPaleAle, line 43: Exposing `_SubtleCrypto` as a feature flag makes it visible through the metadata APIs (thanks to #6537 ), so the E2E tests can decide if they should skip the subtle crypto tests

tests/apps/crypto/Dockerfile
ItalyPaleAle, line 1: This test app requires a Dockerfile because it needs to import the Go SDK

tests/apps/crypto/utils.go
```diff
@@ -0,0 +1,120 @@
...
+	"os/signal"
+	"strconv"
+	"syscall"
+	"time"
+
+	"github.com/google/uuid"
+)
+
+/**
+This file contains code from tests/e2e/utils (with some simplifications).
```
ItalyPaleAle, line 32: See this comment here. Copying this code was required because this app has a Dockerfile (because it needs the SDK).

Importing `github.com/dapr/dapr` in the Dockerfile was causing issues due to protos being registered again.

tests/apps/perf/actor-activation-locker/app.go
```diff
@@ -22,15 +22,16 @@ import (
...
+	redis "github.com/go-redis/redis/v9"
+
 	"github.com/dapr/go-sdk/actor"
 	dapr "github.com/dapr/go-sdk/client"
 	daprd "github.com/dapr/go-sdk/service/http"
-	redis "github.com/go-redis/redis/v9"
 )
 
-func testActorFactory(client dapr.Client, redisClient *redis.Client) func() actor.Server {
+func testActorFactory(client dapr.Client, redisClient *redis.Client) func() actor.ServerContext {
```
ItalyPaleAle, line 32: These changes are due to updating the Go SDK for this app

go.mod
```diff
@@ -8,8 +8,8 @@ require (
 	github.com/PuerkitoBio/purell v1.2.0
 	github.com/argoproj/argo-rollouts v1.4.1
 	github.com/cenkalti/backoff/v4 v4.2.1
-	github.com/dapr/components-contrib v1.11.0-rc.11
-	github.com/dapr/kit v0.11.2
+	github.com/dapr/components-contrib v1.11.3-0.20230621082614-31485cc1e772
```
ItalyPaleAle, line 11: Updates components-contrib to fix an issue with Azure Key Vault (which impacts subtle APIs only)

pkg/runtime/runtime.go
```diff
@@ -3004,7 +3004,7 @@ func isEnvVarAllowed(key string) bool {
 	}
 
 	// If we have a `DAPR_ENV_KEYS` env var (which is added by the Dapr Injector in Kubernetes mode), use that as allowlist too
-	allowlist := os.Getenv(authConsts.EnvKeysEnvVar)
+	allowlist := strings.ToUpper(os.Getenv(authConsts.EnvKeysEnvVar))
```
mukundansundar, line 3007: move this and its tests to another PR if possible with associated issue if any.

ItalyPaleAle: Tests don't pass without this. It's a fixup needed or env vars that aren't all uppercase would not work on Linux.

mukundansundar:
@dapr/maintainers-dapr I believe we can remove the `subtlecrypto` build tag and have those APIs in 1.12, Since E2E tests are also being added in this PR. Are there any concerns with that?

ItalyPaleAle:
> @dapr/maintainers-dapr I believe we can remove the `subtlecrypto` build tag and have those APIs in 1.12, Since E2E tests are also being added in this PR. Are there any concerns with that?

Tracking that in #6593

mukundansundar (approved):

tests/test-infra/setup_azure.sh
```diff
@@ -77,5 +78,27 @@ if $USE_SERVICEBUS ; then
...
+
+  AzureKeyVaultTenantId=$(echo $AZURE_CREDENTIALS | jq -r '.tenantId')
+  AzureKeyVaultClientId=$(echo $AZURE_CREDENTIALS | jq -r '.clientId')
+  AzureKeyVaultClientSecret=$(echo $AZURE_CREDENTIALS | jq -r '.clientSecret')
+
+  kubectl create secret generic azurekeyvault-secret \
+    --namespace=$DAPR_NAMESPACE \
+    --from-literal="tenant-id=${AzureKeyVaultTenantId}" \
+    --from-literal="client-id=${AzureKeyVaultClientId}" \
+    --from-literal="client-secret=${AzureKeyVaultClientSecret}"
```
mukundansundar, line 101: @ItalyPaleAle Why are the credentials being stored here?

tests/config/dapr_crypto_azurekeyvault.yaml
```diff
@@ -0,0 +1,24 @@
...
+        name: azurekeyvault-secret
+        key: tenant-id
+    - name: azureClientId
+      secretKeyRef:
+        name: azurekeyvault-secret
+        key: client-id
+    - name: azureClientSecret
+      secretKeyRef:
+        name: azurekeyvault-secret
+        key: client-secret
```
mukundansundar, line 22: Does this component require actual Azure creds? If so can a specific one be created for the test alone? And not the general Azure Creds for the test also.

mukundansundar (changes requested):

tests/test-infra/setup_azure.sh
ItalyPaleAle: We need a service principal to authenticate with Azure Key Vault

tests/config/dapr_crypto_azurekeyvault.yaml
ItalyPaleAle: The problem is that we need a service principal. We could create a separate one, but then we need to store those credentials somewhere and we need to rotate them.

Maybe I can look into using MI or workload identity. Which would be nice to have too. That would allow us to avoid the need for the client secret?

ItalyPaleAle: As discussed on discord, it seems that neither MI nor WI are convenient in a situation like ours, where clusters are ephemeral, as they require significant setup.

The alternative is to create a separate Azure AD application, however that comes with complexities too as it would be another set of client credentials, which need to be stored as secrets in the repo by an admin and need to be rotated every 6 months too. The complexity is not small either.

As mentioned on Discord, I am not convinced using the same credentials in this specific case is necessarily a problem. While what you say is indeed a best practice for production, in this case I am not sure there is any increased risk in doing what this PR does. The credentials are only available to GitHub Actions that are executed through "/ok-to-test", so only available to trusted users. There's also no _additional_ risk of exfiltration besides what exists today (a malicious actor could modify the pipeline or the `setup_azure.sh` script to make a cURL request to steal the token). Lastly, it should be noted that the subscription associated contains only resources used for tests and no other data.

artursouza: I am fine sharing the creds here. The creds are accessible in the workflow no matter what. What protects the credentials is the requirement to review the PR before sensing /ok-to-test. User can access the secrets from github during workflow after that no matter what.

mukundansundar (approved):

ItalyPaleAle:
Closing as per discussion on the call

Issue #6557, opened 2023-06-20
When subscribing to configuration, clients are not notified if the connection to the configuration store fails


ItalyPaleAle:
## In what area(s)?

/area runtime

## What version of Dapr?

> 1.11.x
> master

## Details

When a client subscribes to a configuration, in case of errors from the components while processing the configuration updates (e.g. connection severed with the configuration store service), there is no automatic recovery and clients are never notified. The only effect is a log message, which apps can't react to.

- In the case of gRPC, Dapr doesn't stop the gRPC stream so clients continue to see the stream alive, but it will not receive messages anymore
- In the case of HTTP, the app stops receiving messages over the callback channel

In case of errors, components are generally simply stopping to listen for messages, printing an error but not notifying the runtime:

- Redis ([relevant code](https://github.com/dapr/components-contrib/blob/bf1f983305c042d04208474981d2b7e9ccfd594e/internal/component/redis/v8client.go#L75)): the SDK does retry to reconnect in case of errors, but:
  - The reconnection can fail with a persistent error, which we do not handle. This should be handled (looking at the code of the SDK) by checking if [`p.Channel()`](https://github.com/dapr/components-contrib/blob/bf1f983305c042d04208474981d2b7e9ccfd594e/internal/component/redis/v8client.go#L93) is closed.
- Postgres ([relevant code](https://github.com/dapr/components-contrib/blob/bf1f983305c042d04208474981d2b7e9ccfd594e/configuration/postgres/postgres.go#L226-L235)): in case of errors, it logs an error and returns
- Azure App Configuration ([relevant code](https://github.com/dapr/components-contrib/blob/bf1f983305c042d04208474981d2b7e9ccfd594e/configuration/azure/appconfig/appconfig.go#L264)): this handles errors by retrying indefinitely, but doesn't distinguish between persistent errors that cannot be retried (e.g. if the key doesn't exist at all).

I am opening this issue in the runtime repo, however, because even if the components did handle errors properly in all cases, they'd have no way to notify the runtime if a persistent error had happened, so the runtime can't inform the app.

The app could then be left hanging waiting for messages, unaware that they are not going to receive any.

PR #6558, opened 2023-06-20
Fixed race conditions in gRPC Configuration APIs

ItalyPaleAle:
This PR fixes some race conditions in the Configuration gRPC API.

It should also fix flaky tests such as https://github.com/dapr/dapr/actions/runs/5326029072/jobs/9647616443?pr=6546

1. Per the gRPC docs, calling `Send()` on a stream from multiple goroutines isn't supported. Added some locks in the `Subscribe` method to ensure concurrency-safety: both to make sure that no more than 1 message is sent at the same time, and also to make sending messages from the configuration store wait for the "hello world" message to be sent first (where only the ID is sent and no configuration updates)
2. In `Unsubscribe`, the channel used to stop a gRPC stream sending subscriptions was deleted before the operation from the component returned. In case the component couldn't unsubscribe, then, there was no way for the client to retry unsubscribing.

PS: There were a bunch of bugs in the unit tests too that required changes (e.g. invoking the handler synchronously instead of async). To avoid having to maintain the code as duplicate for alpha1 and stable, I've merged the two tests.


pkg/grpc/api.go
```diff
@@ -1306,12 +1307,29 @@ func (a *api) GetConfigurationAlpha1(ctx context.Context, in *runtimev1pb.GetCon
 }
 
 type configurationEventHandler struct {
+	readyCh      chan struct{}
+	readyClosed  bool
```
JoshVanL, line 1311:
```suggestion
	readyClosed  atomic.Bool
```

pkg/grpc/api.go
```diff
@@ -1306,12 +1307,29 @@ func (a *api) GetConfigurationAlpha1(ctx context.Context, in *runtimev1pb.GetCon
...
 	api          *api
 	storeName    string
 	serverStream runtimev1pb.Dapr_SubscribeConfigurationAlpha1Server //nolint:nosnakecase
 }
 
+func (h *configurationEventHandler) ready() {
+	if !h.readyClosed {
+		close(h.readyCh)
+		h.readyClosed = true
+	}
```
JoshVanL, line 1322:
```suggestion
	if h.CompareAndSwap(false, true) {
		close(h.readyCh)
	}
```

pkg/grpc/api.go
```diff
@@ -1306,12 +1307,29 @@ func (a *api) GetConfigurationAlpha1(ctx context.Context, in *runtimev1pb.GetCon
...
+}
+
 func (h *configurationEventHandler) updateEventHandler(ctx context.Context, e *configuration.UpdateEvent) error {
+	// Blocks until the first message is sent
+	<-h.readyCh
```
JoshVanL, line 1327: Important for us to exit early here if we are shutting down or something.

```suggestion
        select {
	   case <-h.readyCh:
	   case <-ctx.Done:
	     return ctx.Error()
	   }
```

pkg/grpc/api.go
```diff
@@ -1337,28 +1356,24 @@ func (a *api) SubscribeConfiguration(request *runtimev1pb.SubscribeConfiguration
...
-	})
 
-	subscribeKeys := make([]string, 0)
+	sort.Strings(request.Keys)
 
 	// TODO(@halspang) provide a switch to use just resiliency or this.
 
-	if len(request.Keys) > 0 {
-		subscribeKeys = append(subscribeKeys, request.Keys...)
-	}
```
JoshVanL, line 1350: Why are we no longer removing empty keys? This seems important.

pkg/grpc/api.go
```diff
@@ -1377,15 +1392,22 @@ func (a *api) SubscribeConfiguration(request *runtimev1pb.SubscribeConfiguration
...
 	}
 
 	stop := make(chan struct{})
 	a.CompStore.AddConfigurationSubscribe(subscribeID, stop)
+
+	// We have sent the first message, so signal that we're ready to send messages in the stream
+	handler.ready()
+
+	// Wait until the channel is stopped
 	<-stop
```
JoshVanL, line 1411: Does this close in shutdown scenarios? It doesn't look like it does when we delete the configuration subscription at least.

pkg/grpc/api_test.go
```diff
@@ -1583,68 +1583,60 @@ func TestSubscribeConfiguration(t *testing.T) {
...
-				retry := 3
-				count := 0
-				_, err := resp.Recv()
-				for {
-					if err != nil {
-						break
-					}
-					if count > retry {
-						break
+				resp, _ := subscribeFn(context.Background(), req)
```
JoshVanL, line 1593: Checking the error here seems important.

JoshVanL (changes requested):
Thanks for the code around here.

pkg/grpc/api.go
ItalyPaleAle: This isn't needed because the variable is accessed only in the same goroutine. Confirmed with go test -race

pkg/grpc/api.go
ItalyPaleAle: I has made this change but then realized the issue is much bigger. I'll track that in #6560

pkg/grpc/api.go
ItalyPaleAle: See #6560. I'd like to tackle that separately

pkg/grpc/api_test.go
ItalyPaleAle: Agree and I'm tracking that in #6557

mukundansundar:
Integration tests have failed a few times in this PR but there are no config API integration tests as yet.

mukundansundar (approved):

pkg/grpc/api.go
yaron2: Why is this block of code removed?

ItalyPaleAle: I am putting request.keys below already, this was redundant.

PR #6562, opened 2023-06-20
Remove app port requirement for http app channel

yaron2:
Fixes https://github.com/dapr/dapr/issues/6515 for v1.11


artursouza (approved):

ItalyPaleAle (commented):
Without having tested this, this should be ok.

However I also recommend updating `getAppHTTPChannelForHTTPEndpointsConfig` and passing as `Client: http.DefaultClient`. This is because the client that is created by the runtime is optimized for the internal HTTP channel and:

- If app-protocol is H2C, only supports HTTP/2 Cleartext
- Is configured to not verify TLS certs, which makes sense for a local channel but probably not a remote one

yaron2:
> Without having tested this, this should be ok.
> 
> However I also recommend updating `getAppHTTPChannelForHTTPEndpointsConfig` and passing as `Client: http.DefaultClient`. This is because the client that is created by the runtime is optimized for the internal HTTP channel and:
> 
> * If app-protocol is H2C, only supports HTTP/2 Cleartext
> * Is configured to not verify TLS certs, which makes sense for a local channel but probably not a remote one

Discussed on Discord, we'll keep the H2C changes for 1.12 as part of https://github.com/dapr/dapr/pull/6458.

PR #6565, opened 2023-06-21
adding support in the pubsub component for pluggable component and th…

arturotrenard:
…e bulkpublish function

# Description
The idea of this PR is to allow the custom implementation of the BulkPublish functionality in pubsub pluggable components.
This PR depends on this other PR https://github.com/dapr/components-contrib/pull/2931
This PR is a blocker for this other PR https://github.com/dapr-sandbox/components-go-sdk/pull/26

## Issue reference

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing


go.mod
```diff
@@ -437,7 +437,8 @@ replace github.com/lestrrat-go/jwx/v2 => github.com/lestrrat-go/jwx/v2 v2.0.9-0.
 // Uncomment for local development for testing with changes in the components-contrib repository.
 // Don't commit with this uncommented!
 //
-// replace github.com/dapr/components-contrib => ../components-contrib
+//replace github.com/dapr/components-contrib => ../components-contrib
+
```
ItalyPaleAle, line 441: Tests will fail unless this is unchanged

```suggestion
// replace github.com/dapr/components-contrib => ../components-contrib
```

dapr/proto/components/v1/pubsub.proto
```diff
@@ -82,6 +84,29 @@ message PublishRequest {
...
+
+message BulkMessageEntry {
+  string   entry_id = 1;
+  bytes    event = 2;
+  string   content_type = 3;
+  map<string, string> metadata = 4;
+}
+
+message BulkPublishResponse {
+  repeated BulkPublishResponseFailedEntry FailedEntries = 1;
```
ItalyPaleAle, line 102: Convention is to use snake_case in protos

```suggestion
  repeated BulkPublishResponseFailedEntry failed_entries = 1;
```

pkg/runtime/runtime.go
```diff
@@ -2138,13 +2138,27 @@ func (a *DaprRuntime) BulkPublish(req *pubsub.BulkPublishRequest) (pubsub.BulkPu
...
-	if bulkPublisher, ok := ps.Component.(pubsub.BulkPublisher); ok {
+	bulkPublisher, ok := ps.Component.(pubsub.BulkPublisher)
+
+	if !ok {
+		log.Debugf("pubsub %s does not implement the BulkPublish API; falling back to publishing messages individually", req.PubsubName)
+		defaultBulkPublisher := runtimePubsub.NewDefaultBulkPublisher(ps.Component)
+		return runtimePubsub.ApplyBulkPublishResiliency(context.TODO(), req, policyDef, defaultBulkPublisher)
+	}
+
+	useBulkDefaultPublisher, err := bulkPublisher.UseBulkDefaultPublisher(a.ctx)
```
ItalyPaleAle, line 2151: This makes a 2nd network call in each call which is quite a bit of overhead.

Instead of implementing it this way, which also requires changing all components in components-contrib, how about:

1. We implement a new method in the proto for pubsub pluggable components called "GetMetadata" which returns metadata about the component, including capabilities
2. We list if the component supports native bulk publish
3. The GetMetadata method is invoked just once when the component is initialized. Its status is kept in the component registry, since it's not going to change in the future.

dapr/proto/components/v1/pubsub.proto
```diff
@@ -30,6 +30,10 @@ service PubSub {
   // Publish publishes a new message for the given topic.
   rpc Publish(PublishRequest) returns (PublishResponse) {}
 
+  rpc BulkPublish(BulkPublishRequest) returns (BulkPublishResponse) {}
+
+  rpc UseDefaultBulkPublish(Empty) returns (UseBulkDefaultPublisher) {}
```
artursouza, line 35: You can use the Features() list to return BULK_PUBSUB instead.

pkg/runtime/runtime.go
artursouza: I agree. There should be another way to tag the pubsub as a bulkPubsub without relying on an API call.

artursouza (changes requested):

pkg/runtime/runtime.go
```diff
@@ -2135,13 +2136,24 @@ func (a *DaprRuntime) BulkPublish(req *pubsub.BulkPublishRequest) (pubsub.BulkPu
...
-	if allowed := a.isPubSubOperationAllowed(req.PubsubName, req.Topic, ps.ScopedPublishings); !allowed {
+	allowed := a.isPubSubOperationAllowed(req.PubsubName, req.Topic, ps.ScopedPublishings)
+	if !allowed {
 		return pubsub.BulkPublishResponse{}, runtimePubsub.NotAllowedError{Topic: req.Topic, ID: a.runtimeConfig.ID}
 	}
+
 	policyDef := a.resiliency.ComponentOutboundPolicy(req.PubsubName, resiliency.Pubsub)
-	if bulkPublisher, ok := ps.Component.(pubsub.BulkPublisher); ok {
+
+	bulkPublisher, ok := ps.Component.(pubsub.BulkPublisher)
```
artursouza, line 2146: This should not be here. Instead, use the Features() method prior to this to determine at init time if bulk pubsub is supported natively.

artursouza (changes requested):

pkg/runtime/runtime.go
arturotrenard: ready thanks

artursouza (dismissed):

ItalyPaleAle (dismissed):

ItalyPaleAle (approved):

ItalyPaleAle (approved):

artursouza (approved):

PR #6569, opened 2023-06-21
fuzzing: add fuzzers from cncf-fuzzing

AdamKorcz:
# Description

## Issue reference

Moves the fuzzers to Daprs repository from [cncf-fuzzing](https://github.com/cncf/cncf-fuzzing/tree/main/projects/dapr).

cc @ItalyPaleAle 

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing


ItalyPaleAle:
@AdamKorcz  this is great, thanks :) How are these being used? How can we trigger fuzz tests ourselves?

JoshVanL:
Thanks from me as well @AdamKorcz :smile: 

I'm also been doing some fuzz testing with our integration tests [here](https://github.com/dapr/dapr/pull/6573). I think it's definitely a good idea for us to have fuzz testing at both the unit and integration level.

As a project, I think we should be using the same underlying fuzz testing library, so we need to decide on either `github.com/AdaLogics/go-fuzz-headers` or `github.com/google/gofuzz`. The reason I picked `gofuzz` was that this is what Kubernetes is using upstream, which the project is already heavily dependent on anyway. Does `go-fuzz-headers` provide us with functionality that `gofuzz` cannot?

Also notice you are using a personal fork of the upstream `go-fuzz-headers`, is there a particular reason for that @AdamKorcz? As a project, I don't think we should be relying on personal projects- particularly forks unless there is good reason for it.

AdamKorcz:
> I'm also been doing some fuzz testing with our integration tests

That is great!

> Does go-fuzz-headers provide us with functionality that gofuzz cannot?

I have not used gofuzz extensively. When I tried it (a couple of years ago) I found it to be ineffective. The project is more or less unmaintained, and I am not able to improve it. All projects here use go-fuzz-headers (along with containerd, [Istio](https://github.com/istio/istio/blob/e932c583f226b12856e6a3257456c8968ebc536d/go.mod#L18) and non-CNCF projects as well): https://github.com/cncf/cncf-fuzzing/tree/main/projects.

> Also notice you are using a personal fork of the upstream go-fuzz-headers, is there a particular reason for that @AdamKorcz?

This is because my branch has breaking changes that I want to upstream. I should do that within 1-2 weeks.

AdamKorcz:
> How are these being used?

They are running continuously on OSS-Fuzz

> How can we trigger fuzz tests ourselves?

You can run them locally with `go test -fuzz=`

JoshVanL:
> I have not used gofuzz extensively. When I tried it (a couple of years ago) I found it to be ineffective.

Thanks, I am very interested to know what was ineffective. This package is now heavily used in our integration tests and we should change that now to avoid a larger rewrite in future if it is not good enough.

> This is because my branch has breaking changes that I want to upstream. I should do that within 1-2 weeks.

:+1:

pkg/acl/fuzz_test.go
```diff
@@ -0,0 +1,75 @@
+/*
```
ItalyPaleAle, line 1: Because fuzz tests can be compute-intensive, should we add a build tag so they're only run when those tests are invoked?

for example, build tag `fuzztesting` and then we add a command to the makefile `make test-fuzz`?

AdamKorcz:
The failing CI tests look unrelated to me.

PR #6382, opened 2023-05-22
Standardise flag definitions in CLI commands

JoshVanL:
A general house cleaning PR that helps to break up the overloading of the `/pkg/runtime` package. See also https://github.com/dapr/dapr/pull/6323 https://github.com/dapr/dapr/issues/1172.

PR moves all flag option definitions into dedicated `/options` packages for each respective CLIs. For Daprd, this results in removing `cli.go` completely, and creating a `Config` and `internalConfig` type, with `config.go` handling defaulting.

PR #6584, opened 2023-06-27
Adds Runtime Component Processor and Registry

JoshVanL:
Branched from #6382 which should be merged first.

Part of #2478 and #1172 

PR adds a new runtime `registry` and `processor`. `registry` is a simple wrapper around the different component registries to clean things up a bit. `processor` is designed to manage the lifecycle of each component. Currently it only `init`s the components, but will be expanded to shutting them down too in a future PR. Functionally this PR does nothing but refactor things, but is a necessary step to abstracting/breaking up runtime enough for us to have graceful shutdown/hotreloading.


daixiang0:
Please make PR title more meaningful.

pkg/runtime/registry/options.go
```diff
@@ -0,0 +1,122 @@
...
+	"github.com/dapr/dapr/pkg/components/middleware/http"
+	"github.com/dapr/dapr/pkg/components/nameresolution"
+	"github.com/dapr/dapr/pkg/components/pubsub"
+	"github.com/dapr/dapr/pkg/components/secretstores"
+	"github.com/dapr/dapr/pkg/components/state"
+	"github.com/dapr/dapr/pkg/components/workflows"
+)
+
+// Options is the options to configure the registries
+type Options struct {
```
daixiang0, line 30: The `components` may be more meaningful.

JoshVanL: >  The components may be more meaningful.

I don't know what this means.

> I prefer to NewRegistry register all default registry, and WithXXX in Registry, no need to create Options.

Options are very important here for ensuring that the registry pointers cannot be changed after `NewRegistry` is called, i.e. changing a registry store mid lifecycle of a runtime.

pkg/runtime/meta/meta.go
```diff
@@ -0,0 +1,102 @@
...
+				log.Fatalf("failed to parse metadata: property %s refers to {podName} but podName is not set", c.Name)
+			}
+			val = strings.Replace(val, "{podName}", m.podName, 1)
+		}
+		for strings.Contains(val, "{namespace}") {
+			val = strings.Replace(val, "{namespace}", fmt.Sprintf("%s.%s", m.namespace, m.id), 1)
+		}
+		for strings.Contains(val, "{appID}") {
+			val = strings.Replace(val, "{appID}", m.id, 1)
+		}
```
ItalyPaleAle, line 77: This for loop makes sense when replacing the UUID since we want that to be unique every time. However, `{podName}`, `{namespace}`, `{appID}` are constants, so a more simple `strings.ReplaceAll` may suffice


```suggestion
		if strings.Contains(val, "{podName}") {
			if m.podName == "" {
				// TODO: @joshvanl: return error here rather than panicing.
				log.Fatalf("failed to parse metadata: property %s refers to {podName} but podName is not set", c.Name)
			}
			val = strings.ReplaceAll(val, "{podName}", m.podName)
		}
		val = strings.ReplaceAll(val, "{namespace}", fmt.Sprintf("%s.%s", m.namespace, m.id))
		val = strings.ReplaceAll(val, "{appID}", m.id)
```

pkg/runtime/mock/mock.go
```diff
@@ -0,0 +1,43 @@
+/*
```
ItalyPaleAle, line 1:
```suggestion
//go:build unit
// +build unit

/*
```

pkg/runtime/processor/binding.go
```diff
@@ -0,0 +1,135 @@
...
+		diag.DefaultMonitoring.ComponentInitFailed(comp.Spec.Type, "creation", comp.ObjectMeta.Name)
+		return rterrors.NewInit(rterrors.CreateComponentFailure, fName, err)
+	}
+	err = binding.Init(ctx, bindings.Metadata{Base: b.meta.ToBaseMetadata(comp)})
+	if err != nil {
+		diag.DefaultMonitoring.ComponentInitFailed(comp.Spec.Type, "init", comp.ObjectMeta.Name)
+		return rterrors.NewInit(rterrors.InitComponentFailure, fName, err)
+	}
+
+	log.Infof("successful init for input binding %s (%s/%s)", comp.ObjectMeta.Name, comp.Spec.Type, comp.Spec.Version)
```
ItalyPaleAle, line 80: In all these situations (also in other files), you can use `LogName()`:


```suggestion
	log.Infof("successful init for input binding %s", comp.LogName())
```

pkg/runtime/processor/binding.go
```diff
@@ -0,0 +1,135 @@
...
+
+	fName := comp.LogName()
+	binding, err := b.registry.CreateOutputBinding(comp.Spec.Type, comp.Spec.Version, fName)
+	if err != nil {
+		diag.DefaultMonitoring.ComponentInitFailed(comp.Spec.Type, "creation", comp.ObjectMeta.Name)
+		return rterrors.NewInit(rterrors.CreateComponentFailure, fName, err)
+	}
+
+	if binding != nil {
+		err := binding.Init(context.TODO(), bindings.Metadata{Base: b.meta.ToBaseMetadata(comp)})
```
ItalyPaleAle, line 106: Was this supposed to be `ctx`?

```suggestion
		err := binding.Init(ctx, bindings.Metadata{Base: b.meta.ToBaseMetadata(comp)})
```

pkg/runtime/processor/binding.go
```diff
@@ -0,0 +1,135 @@
...
+			return rterrors.NewInit(rterrors.InitComponentFailure, fName, err)
+		}
+		log.Infof("successful init for output binding %s (%s/%s)", comp.ObjectMeta.Name, comp.Spec.Type, comp.Spec.Version)
+		b.compStore.AddOutputBinding(comp.ObjectMeta.Name, binding)
+		diag.DefaultMonitoring.ComponentInitialized(comp.Spec.Type)
+	}
+	return nil
+}
+
+func (b *binding) isBindingOfDirection(direction string, metadata []common.NameValuePair) bool {
```
ItalyPaleAle, line 118: This method works by looking at the component's metadata only. Shouldn't we also check if the component struct actually implements the InputBinding or OutputBinding interfaces (in addition to the metadata value)?

pkg/runtime/processor/binding.go
```diff
@@ -0,0 +1,135 @@
...
+		return rterrors.NewInit(rterrors.CreateComponentFailure, fName, err)
+	}
+
+	if binding != nil {
+		err := binding.Init(context.TODO(), bindings.Metadata{Base: b.meta.ToBaseMetadata(comp)})
+		if err != nil {
+			diag.DefaultMonitoring.ComponentInitFailed(comp.Spec.Type, "init", comp.ObjectMeta.Name)
+			return rterrors.NewInit(rterrors.InitComponentFailure, fName, err)
+		}
+		log.Infof("successful init for output binding %s (%s/%s)", comp.ObjectMeta.Name, comp.Spec.Type, comp.Spec.Version)
```
ItalyPaleAle, line 111:
```suggestion
		log.Infof("successful init for output binding %s", comp.LogName())
```

pkg/runtime/processor/state.go
```diff
@@ -0,0 +1,116 @@
...
+		}
+
+		// when placement address list is not empty, set specified actor store.
+		if s.placementEnabled {
+			// set specified actor store if "actorStateStore" is true in the spec.
+			actorStoreSpecified := false
+			for k, v := range props {
+				//nolint:gocritic
+				if strings.ToLower(k) == propertyKeyActorStateStore {
+					actorStoreSpecified = utils.IsTruthy(v)
```
ItalyPaleAle, line 95: Nit:

```suggestion
					actorStoreSpecified = utils.IsTruthy(v)
                    break
```

pkg/runtime/runtime.go
```diff
@@ -2538,7 +2197,9 @@ func (a *DaprRuntime) initActors() error {
 	}
 	a.actorStateStoreLock.Lock()
 	defer a.actorStateStoreLock.Unlock()
-	if a.actorStateStoreName == "" {
+
+	actorStateStoreName, ok := a.processor.ActorStateStore()
```
ItalyPaleAle, line 2201: Nit: does this need to return an `ok`? a component can't have an empty name anwyays

ItalyPaleAle (changes requested):
This tidies things up a lot and I am a fan :)

pkg/runtime/meta/meta.go
JoshVanL: :+1:

pkg/runtime/mock/mock.go
JoshVanL: Will begrudgingly include! :slightly_smiling_face:

pkg/runtime/processor/binding.go
JoshVanL: It needs to be `ctx` eventually yes. Context propagation is part of this runtime refactoring effort to eventually achieve graceful shutdown + hot reloading.

pkg/runtime/processor/binding.go
JoshVanL: I didn't write this func myself and am apprehensive about changing it's behavior here. I copied it as is. I think this should be addressed in a separate PR.

pkg/runtime/runtime.go
JoshVanL: I like having `ok` checks here as it is idiomatic in Go, and gives us a firm type to check against if we want to extend the logic which determines whether the name is valid.

pkg/runtime/processor/binding.go
```diff
@@ -108,7 +108,7 @@ func (b *binding) initOutputBinding(ctx context.Context, comp compapi.Component)
 			diag.DefaultMonitoring.ComponentInitFailed(comp.Spec.Type, "init", comp.ObjectMeta.Name)
 			return rterrors.NewInit(rterrors.InitComponentFailure, fName, err)
 		}
-		log.Infof("successful init for output binding %s (%s/%s)", comp.ObjectMeta.Name, comp.Spec.Type, comp.Spec.Version)
+		log.Infof("successful init for output binding %s (%s)", comp.ObjectMeta.Name, comp.LogName())
```
ItalyPaleAle, line 111:
```suggestion
		log.Infof("successful init for output binding %s", comp.LogName())
```

pkg/runtime/processor/workflow.go
```diff
@@ -36,7 +36,7 @@ func (w *workflow) init(ctx context.Context, comp compapi.Component) error {
 	fName := comp.LogName()
 	workflowComp, err := w.registry.Create(comp.Spec.Type, comp.Spec.Version, fName)
 	if err != nil {
-		log.Warnf("error creating workflow component %s (%s/%s): %s", comp.ObjectMeta.Name, comp.Spec.Type, comp.Spec.Version, err)
+		log.Warnf("error creating workflow component %s (%s): %s", comp.ObjectMeta.Name, comp.LogName(), err)
```
ItalyPaleAle, line 39:
```suggestion
		log.Warnf("error creating workflow component %s: %s", comp.LogName(), err)
```

ItalyPaleAle (commented):
LogName includes the name too, not just type

ItalyPaleAle (approved):

ItalyPaleAle:
@JoshVanL looks like E2E tests are failing consistently on the bindings test

JoshVanL:
> @JoshVanL looks like E2E tests are failing consistently on the bindings test

@ItalyPaleAle Thanks, yeah I've just pushed a change which should resolve it.

ItalyPaleAle (dismissed):

pkg/runtime/meta/meta.go
```diff
@@ -0,0 +1,98 @@
...
+	properties := map[string]string{}
+	for _, c := range items {
+		val := c.Value.String()
+		for strings.Contains(val, "{uuid}") {
+			val = strings.Replace(val, "{uuid}", uuid.New().String(), 1)
+		}
+		if strings.Contains(val, "{podName}") {
+			if m.podName == "" {
+				// TODO: @joshvanl: return error here rather than panicing.
+				log.Fatalf("failed to parse metadata: property %s refers to {podName} but podName is not set", c.Name)
```
artursouza, line 68: Please, return error :)

pkg/runtime/processor/binding.go
artursouza: Thanks for that. We should limit behavior change in refactor PRs.

pkg/runtime/processor/state.go
```diff
@@ -0,0 +1,117 @@
...
+
+			if actorStoreSpecified {
+				s.lock.Lock()
+				defer s.lock.Unlock()
+				if s.actorStateStoreName == "" {
+					log.Info("Using '" + comp.ObjectMeta.Name + "' as actor state store")
+					s.actorStateStoreName = comp.ObjectMeta.Name
+				} else if s.actorStateStoreName != comp.ObjectMeta.Name {
+					// TODO: @joshvanl: Don't fatal here. Return error.
+					log.Fatalf("Detected duplicate actor state store: %s and %s", s.actorStateStoreName, comp.ObjectMeta.Name)
```
artursouza, line 108: Please, no fatal.

artursouza (approved):

pkg/runtime/meta/meta.go
JoshVanL: @artursouza I didn't write these Fatals and copied them as is. Happy to change these now, but planned to address all Fatals in daprd in a separate PR.

artursouza (approved):

PR #6586, opened 2023-06-27
Updates go modules and register both etcd v1 and v2

JoshVanL:
With https://github.com/dapr/components-contrib/pull/2932, we now have `etcd` V2.

PR registers both `etcd` v1 and v2, making available:

```yaml
  type: state.etcd
  version: v2
```


go.mod
```diff
@@ -30,7 +30,7 @@ require (
 	github.com/jhump/protoreflect v1.15.1
 	github.com/kelseyhightower/envconfig v1.4.0
 	github.com/labstack/gommon v0.4.0
-	github.com/lestrrat-go/jwx/v2 v2.0.9
+	github.com/lestrrat-go/jwx/v2 v2.0.11
```
ItalyPaleAle, line 33: Please remove the replace on L434

ItalyPaleAle (changes requested):

cmd/daprd/components/state_etcd.go
```diff
@@ -21,5 +21,6 @@ import (
 )
 
 func init() {
-	stateLoader.DefaultRegistry.RegisterComponent(etcd.NewEtcdStateStore, "etcd")
+	stateLoader.DefaultRegistry.RegisterComponent(etcd.NewEtcdStateStoreV1, "etcd/v1")
```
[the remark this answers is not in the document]
JoshVanL: Added a test [here](https://github.com/dapr/dapr/pull/6586/files#diff-fb06f885e9b18b42ed117263facfc21add5c70d50d141573f95931941844077e) to ensure dapr can parse v1 and v2 components.

ItalyPaleAle (approved):

tests/integration/suite/components/versionv2.go
```diff
@@ -0,0 +1,73 @@
...
+
+func (v *versionv2) Setup(t *testing.T) []framework.Option {
+	v.proc = procdaprd.New(t, procdaprd.WithComponentFiles(`
+apiVersion: dapr.io/v1alpha1
+kind: Component
+metadata:
+  name: etcd-v1
+spec:
+  type: state.etcd
+  version: v1
```
artursouza, line 42: What if version is omitted? Can you add a scenario for that too?

JoshVanL: @artursouza It failed unless we register with no version (which I've added). I've made the default version `v1` for backwards compat. Let me know if you want me to change to `v2`/

cmd/daprd/components/state_etcd.go
```diff
@@ -21,5 +21,7 @@ import (
 )
 
 func init() {
-	stateLoader.DefaultRegistry.RegisterComponent(etcd.NewEtcdStateStore, "etcd")
+	stateLoader.DefaultRegistry.RegisterComponent(etcd.NewEtcdStateStoreV1, "etcd")
```
ItalyPaleAle, line 24: Rather than doing this, should change the component registry to default to `v1` if omitted

tests/integration/suite/components/versionv2.go
```diff
@@ -0,0 +1,87 @@
...
+kind: Component
+metadata:
+  name: etcd-omitted
+spec:
+  type: state.etcd
+  metadata:
+    - name: endpoints
+      value: "localhost:12379"
+    - name: keyPrefixPath
+      value: "daprv2"
```
ItalyPaleAle, line 76:
```suggestion
      value: "daprv1"
```

tests/integration/framework/process/daprd/daprd.go
```diff
@@ -120,6 +130,9 @@ func New(t *testing.T, fopts ...Option) *Daprd {
 	if opts.appHealthCheckPath != "" {
 		args = append(args, "--app-health-check-path="+opts.appHealthCheckPath)
 	}
+	if len(opts.componentFiles) > 0 {
+		args = append(args, "-components-path="+dir)
```
ItalyPaleAle, line 134:
```suggestion
		args = append(args, "-resources-path="+dir)
```

tests/integration/framework/process/daprd/daprd.go
```diff
@@ -120,6 +130,9 @@ func New(t *testing.T, fopts ...Option) *Daprd {
 	if opts.appHealthCheckPath != "" {
 		args = append(args, "--app-health-check-path="+opts.appHealthCheckPath)
 	}
+	if len(opts.componentFiles) > 0 {
```
ItalyPaleAle, line 133: Nit: Consider renaming to `resourceFiles` to support all kind of resources

tests/integration/suite/components/versionv2.go
ItalyPaleAle: No, default should be `v1` for backwards-compat. Otherwise all users who did not specify a version (and there's probably a lot) would get upgraded.

ItalyPaleAle (changes requested):

tests/integration/suite/components/versionv2.go
daixiang0: > No, default should be `v1` for backwards-compat. Otherwise all users who did not specify a version (and there's probably a lot) would get upgraded.

Agree.

cmd/daprd/components/state_etcd.go
JoshVanL: Done

tests/integration/framework/process/daprd/daprd.go
JoshVanL: Done

cmd/daprd/components/state_etcd.go
JoshVanL: There is now a mechanism for defining versioning behavior of component types which have multiple version.

pkg/components/state/registry.go
```diff
@@ -35,6 +38,13 @@ var DefaultRegistry *Registry = NewRegistry()
 func NewRegistry() *Registry {
 	return &Registry{
 		stateStores: map[string]func(logger.Logger) state.Store{},
+		verSet: map[string]components.Versioning{
```
ItalyPaleAle, line 41: Can this be done in the code that registers the components? in this case, `cmd/daprd/components/state_etcd.go`?

Two reasons:

1. Keeping all component registration code tidy and in the same place
2. The etcd component may not be registered depending on the build tags and custom configurations

JoshVanL: But this, do you mean that when a component is registered, it also passes version information? I.e. Add a new `RegisterComponent` func which accepts version mapping.

ItalyPaleAle: Yes. I am not sure how to implement this in practice, i.e. what the API should look like... But ideally we should keep all information on a component self-contained in the file in `cmd/daprd/components`, rather than in multiple places

JoshVanL: No worries! I'll have a go at that. 

I agree with the reasoning for sure.

cmd/daprd/components/state_etcd.go
```diff
@@ -16,10 +16,19 @@ limitations under the License.
...
 	stateLoader "github.com/dapr/dapr/pkg/components/state"
 )
 
 func init() {
-	stateLoader.DefaultRegistry.RegisterComponent(etcd.NewEtcdStateStore, "etcd")
+	stateLoader.DefaultRegistry.RegisterComponentWithVersions("etcd", components.Versioning{
+		Preferred: components.VersionConstructor{
+			Version: "v2", Constructor: etcd.NewEtcdStateStoreV2,
+		},
+		Deprecated: []components.VersionConstructor{
```
ItalyPaleAle, line 29: Thanks for this, this allows us to add a v2 without "deprecating" v1.

ItalyPaleAle (approved):

PR #6587, opened 2023-06-27
Windows: Use taskkill to kill windows processes

JoshVanL:
In an effort to improve the reliability of windows tests, use taskkill to kill the process, rather than simulating terminal events.


ItalyPaleAle (dismissed):

ItalyPaleAle (commented):
Thoughts on setting for daprd (in integration tests) graceful shutdown duration to 0 (which makes it 0 seconds)?

ItalyPaleAle (approved):

JoshVanL:
> Thoughts on setting for daprd (in integration tests) graceful shutdown duration to 0 (which makes it 0 seconds)?

I don't think we achieve much doing that. We will have proper graceful shutdown in 1.12 anyway.

yaron2 (approved):

PR #6477, opened 2023-06-08
Enable flag `ServiceInvocationStreaming` in all builds by default

ItalyPaleAle:
As we work on enabling streaming everywhere, this PR changes the Makefile to enable the flag `ServiceInvocationStreaming`. This will allow us to test Dapr with the feature flag enabled more actively in the next weeks (just like we did before making Resiliency stable).

PR #6594, opened 2023-06-27
Streaming: convert HTTP service invocation handler to net/http

ItalyPaleAle:
> 9 months later, this **finally** completes the project!

Converts the handler for HTTP service invocation to net/http, with full support for streaming.

Together with #6477, this will enable E2E support for streaming in HTTP service invocation.


Makefile
```diff
@@ -348,7 +348,7 @@ test-race:
 ################################################################################
 .PHONY: test-integration
 test-integration: test-deps
-		gotestsum \
+		CGO_ENABLED=1 gotestsum \
```
ItalyPaleAle, line 351: Running `test-integration` requires CGo, so this enables it in case the shell's environment has `CGO_ENABLED=0` (like mine)

pkg/messaging/v1/replayable_request_test.go
```diff
@@ -215,3 +217,29 @@ func TestReplayableRequest(t *testing.T) {
 		})
 	})
 }
+
+// readerCloser is a io.Reader that can be closed. Once the stream is closed, reading from it returns an error.
+type readerCloser struct {
```
ItalyPaleAle, line 222: This is added here to simulate reading from a `http.Response.Body` which has already been closed.

ItalyPaleAle:
It seems that this improves perf a bit for HTTP service invocation. In both tests, P90 is around 1.95ms

Makefile
ItalyPaleAle: It’s because it uses `-race`

mukundansundar (commented):
LGTM pending tests pass

ItalyPaleAle:
The tests seem to be passing. The failure in the Python tests is related to workflows and it's in the master and release-1.11 branch too. See #6661 too.

This should be good to be merged now @mukundansundar

mukundansundar (approved):

PR #6596, opened 2023-06-27
Streaming: Convert crypto APIs to net/http

ItalyPaleAle:
Follow-up from #6461 

Converts the high-level crypto APIs to net/http, enabling the handling of data with streams which significantly reduces memory usage (although these APIs are still less efficient than the gRPC ones, as the data is still buffered in memory once)


ItalyPaleAle:
Closing as per discussion on the call

PR #6601, opened 2023-06-28
Enable HTTP/2 Cleartext in HTTP Server

### LINKED ISSUE #6570: Enable HTTP/2 Cleartext support in Dapr HTTP Server
## In what area(s)?

/area runtime

## Describe the proposal

After #6461 is merged, we can enable HTTP/2 Cleartext support in Dapr's HTTP server. This will be implemented by supporting "upgrades" from HTTP/1.1, allowing both protocols to be used at the same time.

### LINKED ISSUE #6767: why Proto is HTTP/1.1 when dapr.io/app-protocol is h2c
## In what area(s)?

> runtime: 1.11.0 kubernetes

## Ask your question here
 I had added  `dapr.io/app-protocol: h2c` in annotations
But when i invoke by dapr, `Proto` still is `HTTP/1.1` in http header and the actual request is HTTP2

image

image

image

ItalyPaleAle:
Fixes #6570
Fixes #6767

Enables support for HTTP/2 Cleartext in the HTTP server. This should allow apps to talk to Dapr using HTTP/2 with perf improvements.

Apps making a HTTP/1 connection can upgrade to HTTP/2 if the client supports that, transparently. E.g.:

```
curl http://localhost:3500/v1.0/healthz -v --http2               
*   Trying ::1:3500...
* Connected to localhost (::1) port 3500 (#0)
> GET /v1.0/healthz HTTP/1.1
> Host: localhost:3500
> User-Agent: curl/7.74.0
> Accept: */*
> Connection: Upgrade, HTTP2-Settings
> Upgrade: h2c
> HTTP2-Settings: AAMAAABkAAQCAAAAAAIAAAAA
> 
* Mark bundle as not supporting multiuse
< HTTP/1.1 101 Switching Protocols
< Connection: Upgrade
< Upgrade: h2c
* Received 101
* Using HTTP2, server supports multi-use
* Connection state changed (HTTP/2 confirmed)
* Copying HTTP/2 data in stream buffer to connection buffer after upgrade: len=0
* Connection state changed (MAX_CONCURRENT_STREAMS == 250)!
< HTTP/2 204 
< content-type: text/plain; charset=utf-8
< date: Wed, 28 Jun 2023 15:48:04 GMT
< 
* Connection #0 to host localhost left intact
```

Clients that know that the server (daprd) supports HTTP/2 Cleartext can also make a H2C connection right away (this is called "with prior knowledge"):

```
curl http://localhost:3500/v1.0/healthz -v --http2-prior-knowledge 
*   Trying ::1:3500...
* Connected to localhost (::1) port 3500 (#0)
* Using HTTP2, server supports multi-use
* Connection state changed (HTTP/2 confirmed)
* Copying HTTP/2 data in stream buffer to connection buffer after upgrade: len=0
* Using Stream ID: 1 (easy handle 0x55778816fad0)
> GET /v1.0/healthz HTTP/2
> Host: localhost:3500
> user-agent: curl/7.74.0
> accept: */*
> 
* Connection state changed (MAX_CONCURRENT_STREAMS == 250)!
< HTTP/2 204 
< content-type: text/plain; charset=utf-8
< date: Wed, 28 Jun 2023 15:49:05 GMT
< 
* Connection #0 to host localhost left intact
```

A few notes:

- H2C support is enabled by default but can be disabled with the `DAPR_HTTP_DISABLE_H2C=1` env var.
  - The main reason why we would want to disable this is testing. For example, to write a perf test that measures the perf with both HTTP/1 and HTTP/2, since perf testers like K6 automatically upgrade to HTTP/2 if that's offered.
  - I am not really aware of any other reason why we'd turn off H2C, so that's why it's an env var rather than a CLI flag.
- We **cannot** get a perf test for this **yet**. I've opened an issue (#6602) to track that, since it involves a lot more work than is in scope for this PR.
  - Our service invocation perf tests currently use fortio
  - We use a fork of fortio 1.38 where we added support for gRPC
  - Fortio added support for testing with HTTP/2 only at a later release
  - I was able to port our patches for gRPC to [fortio 1.55](https://github.com/dapr/fortio/tree/v1.55.2-dapr) which does have HTTP/2 support, but...
  - When fortio runs with HTTP/1, it uses fasthttp as a HTTP client internally. As we know well, fasthttp doesn't support HTTP/2, so when running tests against HTTP/2, fortio uses a different HTTP client. That lead to results that were not comparable


mukundansundar:
> We cannot get a perf test for this yet. I've opened an issue (https://github.com/dapr/dapr/issues/6602) to track that, since it involves a lot more work than is in scope for this PR.
Our service invocation perf tests currently use fortio
We use a fork of fortio 1.38 where we added support for gRPC
Fortio added support for testing with HTTP/2 only at a later release
I was able to port our patches for gRPC to [fortio 1.55](https://github.com/dapr/fortio/tree/v1.55.2-dapr) which does have HTTP/2 support, but...
When fortio runs with HTTP/1, it uses fasthttp as a HTTP client internally. As we know well, fasthttp doesn't support HTTP/2, so when running tests against HTTP/2, fortio uses a different HTTP client. That lead to results that were not comparable

I think we should move the service invocation perf tests to k6 if possible.

ItalyPaleAle:
> 
> I think we should move the service invocation perf tests to k6 if possible.

Tracking that in #6602 as it’s out of scope of this PR

tests/integration/suite/httpserver/httpserver.go
```diff
@@ -0,0 +1,159 @@
...
+
+		reqCtx, reqCancel := context.WithTimeout(ctx, 150*time.Millisecond)
+		defer reqCancel()
+
+		req, err := http.NewRequestWithContext(reqCtx, http.MethodGet, fmt.Sprintf("http://localhost:%d/v1.0/healthz", h.proc.HTTPPort()), nil)
+		require.NoError(t, err)
+
+		req.Header.Set("Connection", "Upgrade, HTTP2-Settings")
+		req.Header.Set("Upgrade", "h2c")
+		req.Header.Set("HTTP2-Settings", "AAMAAABkAARAAAAAAAIAAAAA")
```
mukundansundar, line 135: any specific magic string here?

pkg/http/server.go
```diff
@@ -128,11 +132,17 @@ func (s *server) StartNonBlocking() error {
 		return errors.New("could not listen on any endpoint")
 	}
 
+	// Create a handler with support for HTTP/2 Cleartext
+	var handler http.Handler = r
+	if !utils.IsTruthy(os.Getenv("DAPR_HTTP_DISABLE_H2C")) {
+		handler = h2c.NewHandler(r, &http2.Server{})
```
mukundansundar, line 138: qq can TLS still be set with h2c handler?

ItalyPaleAle: H2C is HTTP/2 Cleartext so it cannot use TLS by definition :)

However our server (now) does support HTTP/2 automatically when TLS is enabled. But that only work when using HTTPS.

tests/integration/suite/httpserver/httpserver.go
ItalyPaleAle: That is from the specs: https://book.hacktricks.xyz/pentesting-web/h2c-smuggling

Users will never need to worry about this stuff. HTTP clients should be smart to do these things automatically. In this case, we are doing things manually to confirm stuff works as expected.

mukundansundar (dismissed):

daixiang0 (dismissed):
LGTM

daixiang0:
@ItalyPaleAle @mukundansundar pgs setup fail :(

shivamkm07:
> > @ItalyPaleAle @mukundansundar postgres setup fail :(
> 
> Yes, it hasn’t been working since #6230 has been merged.
> 
> CC: @shivamkm07

Yes. This is because of an issue in upstream bitnami charts for postgresql: https://github.com/bitnami/charts/issues/17488

mukundansundar (dismissed):

mukundansundar (approved):

Issue #6582, opened 2023-06-26
Change Component/Configuration/Subscription CRDs apiVersion to v1


yaron2:
The API version for the Component/Configuration/Subscription CRD is stable and should be changed to support `v1` while keeping `v1alpha1` for backwards compatibility.

aharonYK:
hello i would like to fix this issue can you give me the schema for v1 so i can update it?

yaron2:
Thanks for picking this up! The current CRDs schemas for Component, Configuration and Subscription (v2alpha1) remain the same and should be made `v1` while still supporting the existing api versions for backward compatbility.

cicoyle:
@aharonYK this would be great to get in next release if you are able :)

PR #6604, opened 2023-06-28
Add v1 apiVersion to CRDs

aharonYK:
# Description

Change Component/Configuration/Subscription CRDs apiVersion to v1 #6582
let me know if this is what you meant. it was a very good experience working on this i didn't came across it before
## Issue reference
#6582

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing


aharonYK:
please let me know if it need fixing

charts/dapr/crds/subscription.yaml
```diff
@@ -21,7 +21,7 @@ spec:
       - v1
       - v2alpha1
   versions:
-  - name: v1alpha1
+  - name: v1
```
yaron2, line 24: For Subscriptions, we need to add `v1` with the same schema as `v2alpha1` and leave `v1alpha` unchanged.

yaron2 (changes requested):

ItalyPaleAle (changes requested):
These changes need to happen in the `pkg/api` structs, which are used to generate the CRDs with `make code-generate`. Also, please do not delete old versions because that's a backwards-incompatible change. We need to continue supporting old versions and use a webhook to convert to `v1`.

daixiang0:
@aharonYK please resolve the comments :)

charts/dapr/crds/subscription.yaml
berndverst: And we also need to keep `v2alpha1` around because that is also in use.

cicoyle:
Hey @aharonYK - Will you be able to pick this back up?

PR #6605, opened 2023-06-28
Fixes from #6573

### LINKED ISSUE #6588: Change in API behaviour/breakage.
After https://github.com/dapr/dapr/pull/6461 it looks like we have some breakages in the API. After rebasing the [integration tests PR](https://github.com/dapr/dapr/pull/6573), they are consistently failing when before they would consistently pass.

- Service invocation HTTP looks to no longer pass-through the HTTP method from the caller to the app.
- Secret HTTP GET responses now include a newline at the end of the message.
- State store HTTP request parsing seems totally broken on fuzzed inputs

This is only the stuff which the [integration tests](https://github.com/dapr/dapr/pull/6573) have caught, though they don't cover everything so there well may be more.

/cc @yaron2

ItalyPaleAle:
Fixes the errors reported by #6573
Fixes #6588

Among the 3 issues reported in #6588:

1. **Service invocation HTTP looks to no longer pass-through the HTTP method from the caller to the app.**
   - This was determined to be not an issue in Dapr.
   - As explained in #6588, the service invocation handler is the only one that, by design, doesn't always remove double slashes. So, when the invocation was made for `localhost:port////foo`, Dapr would have invoked the app at `app:port///foo`. The request was made with the correct HTTP verb. The test app uses the Go  mux from the standard library, which detected the extra `/` at the beginning and responded with a 301 redirect pointing Dapr to retry with `/foo`. However, 301 redirects do not respect the verb [per specs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/301) and retries used the GET verb. 
   - The issue was fixed by changing the test to not test for `localhost:port////foo` and use a single slash
   - In #6594 this has been changed to allow using `///foo` again. The extra slashes are cleaned there.
2. **Secret HTTP GET responses now include a newline at the end of the message.**
   - Maintainers agree this is a non-issue, since we fulfill the contract of returning valid JSON.
3. **State store HTTP request parsing seems totally broken on fuzzed inputs**, or to be more precise, issues with path encoding (in state stores and secret stores).
   - This was a combination of issues in the runtime and in tests:
   - In the runtime, we had to switch the way we grab the path. This was fixed by using a custom version of the CleanPath middleware See go-chi/chi#641 
   - In the test framework, we're using `url.QueryEscape` rather than `url.PathEscape` which is more complete
     - Filter keys that are only dots

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -348,7 +348,7 @@ test-race:
 ################################################################################
 .PHONY: test-integration
 test-integration: test-deps
-		gotestsum \
+		CGO_ENABLED=1 gotestsum \
 			--jsonfile $(TEST_OUTPUT_FILE_PREFIX)_integration.json \
 			--format testname \
 			-- \
--- a/tests/integration/suite/daprd/secret/http/fuzz.go
+++ b/tests/integration/suite/daprd/secret/http/fuzz.go
@@ -62,7 +62,8 @@ func (f *fuzzsecret) Setup(t *testing.T) []framework.Option {
 		for *s == "" ||
 			takenNames[*s] ||
 			len(path.IsValidPathSegmentName(*s)) > 0 ||
-			!reg.MatchString(*s) {
+			!reg.MatchString(*s) ||
+			strings.HasSuffix(*s, ".") {
 			*s = c.RandString()
 		}
 		takenNames[*s] = true
--- a/tests/integration/suite/daprd/secret/http/fuzz.go
+++ b/tests/integration/suite/daprd/secret/http/fuzz.go
@@ -24,7 +24,6 @@ import (
 	"os"
 	"path/filepath"
 	"regexp"
-	"strconv"
 	"strings"
 	"testing"
 
@@ -129,9 +128,9 @@ func (f *fuzzsecret) Run(t *testing.T, ctx context.Context) {
 		t.Run(key+":"+value, func(t *testing.T) {
 			t.Parallel()
 			getURL := fmt.Sprintf("http://localhost:%d/v1.0/secrets/%s/%s", f.daprd.HTTPPort(), url.QueryEscape(f.secretStoreName), url.QueryEscape(key))
-			t.Log("URL", getURL)
-			t.Log("Secret store name", f.secretStoreName, printRunes(f.secretStoreName))
-			t.Log("Key", key, printRunes(key))
+			// t.Log("URL", getURL)
+			// t.Log("Secret store name", f.secretStoreName, printRunes(f.secretStoreName))
+			// t.Log("Key", key, printRunes(key))
 			req, err := http.NewRequestWithContext(ctx, http.MethodGet, getURL, nil)
 			require.NoError(t, err)
 			resp, err := http.DefaultClient.Do(req)
@@ -147,6 +146,7 @@ func (f *fuzzsecret) Run(t *testing.T, ctx context.Context) {
 	// TODO: Bulk APIs, nesting, multi-valued
 }
 
+/*
 func printRunes(str string) []string {
 	result := make([]string, 0, len(str))
 	for _, r := range str {
@@ -154,3 +154,4 @@ func printRunes(str string) []string {
 	}
 	return result
 }
+*/
--- a/tests/integration/suite/daprd/secret/http/fuzz.go
+++ b/tests/integration/suite/daprd/secret/http/fuzz.go
@@ -63,7 +63,7 @@ func (f *fuzzsecret) Setup(t *testing.T) []framework.Option {
 			takenNames[*s] ||
 			len(path.IsValidPathSegmentName(*s)) > 0 ||
 			!reg.MatchString(*s) ||
-			strings.HasSuffix(*s, ".") {
+			*s == "." {
 			*s = c.RandString()
 		}
 		takenNames[*s] = true
--- a/tests/integration/suite/daprd/secret/http/fuzz.go
+++ b/tests/integration/suite/daprd/secret/http/fuzz.go
@@ -24,6 +24,7 @@ import (
 	"os"
 	"path/filepath"
 	"regexp"
+	"strconv"
 	"strings"
 	"testing"
 
@@ -126,7 +127,10 @@ func (f *fuzzsecret) Run(t *testing.T, ctx context.Context) {
 		value := value
 		t.Run(key+":"+value, func(t *testing.T) {
 			t.Parallel()
-			getURL := fmt.Sprintf("http://localhost:%d/v1.0/secrets/%s/%s", f.daprd.HTTPPort(), url.PathEscape(f.secretStoreName), url.PathEscape(key))
+			getURL := fmt.Sprintf("http://localhost:%d/v1.0/secrets/%s/%s", f.daprd.HTTPPort(), url.QueryEscape(f.secretStoreName), url.QueryEscape(key))
+			t.Log("URL", getURL)
+			t.Log("Secret store name", f.secretStoreName, printRunes(f.secretStoreName))
+			t.Log("Key", key, printRunes(key))
 			req, err := http.NewRequestWithContext(ctx, http.MethodGet, getURL, nil)
 			require.NoError(t, err)
 			resp, err := http.DefaultClient.Do(req)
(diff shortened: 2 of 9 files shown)


tests/integration/suite/daprd/serviceinvocation/http/basic.go
```diff
@@ -206,16 +219,18 @@ func (b *basic) Run(t *testing.T, ctx context.Context) {
 	for i := 0; i < 100; i++ {
 		t.Run("parallel requests", func(t *testing.T) {
 			t.Parallel()
-			reqURL := fmt.Sprintf("http://localhost:%d/v1.0/invoke/%s/method/foo", b.daprd1.HTTPPort(), b.daprd2.AppID())
-			req, err := http.NewRequestWithContext(ctx, http.MethodPost, reqURL, nil)
+			u := uuid.New().String()
```
ItalyPaleAle, line 222: This is a modified test that sends a body and validates the response too. Since we're testing parallel things, this makes sure a response is correlated to a request.

Makefile
```diff
@@ -348,7 +348,7 @@ test-race:
 ################################################################################
 .PHONY: test-integration
 test-integration: test-deps
-		gotestsum \
+		CGO_ENABLED=1 gotestsum \
```
ItalyPaleAle, line 351: Using `-race` requires CGo and my shell's env has `CGO_ENABLED=0`

tests/integration/suite/daprd/serviceinvocation/http/basic.go
```diff
@@ -114,38 +122,43 @@ func (b *basic) Run(t *testing.T, ctx context.Context) {
...
 		}{
 			{url: fmt.Sprintf("http://localhost:%d/v1.0/invoke/%s/method/foo", b.daprd1.HTTPPort(), b.daprd2.AppID())},
 			{url: fmt.Sprintf("http://localhost:%d/v1.0/invoke/%s/method/foo", b.daprd2.HTTPPort(), b.daprd1.AppID())},
 			{url: fmt.Sprintf("http://localhost:%d/v1.0////invoke/%s/method/foo", b.daprd2.HTTPPort(), b.daprd1.AppID())},
 			{url: fmt.Sprintf("http://localhost:%d/v1.0//invoke//%s/method//foo", b.daprd1.HTTPPort(), b.daprd2.AppID())},
-			{url: fmt.Sprintf("http://localhost:%d///foo", b.daprd1.HTTPPort()), headers: map[string]string{
+			// We cannot use `///foo` here because the test app uses the standard Go mux which responds with a 301 status code if the invocation is for `///foo`
+			// This makes Dapr retry with a GET request in all cases, as per specs
+			// See: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/301
+			{url: fmt.Sprintf("http://localhost:%d/foo", b.daprd1.HTTPPort()), headers: map[string]string{
```
JoshVanL, line 136: Rather than change the test, why can we not just have the previous behavior back?

At _minimum_ we need this test case to be covered, please don't delete it. It's an edge case in which we are interested to know the result of, and we can't have the behavior keep changing on a whim. It also needs to be documented that this is a breaking change if we are not going to maintain the previous behavior of Service Invocation.

tests/integration/suite/daprd/secret/http/fuzz.go
```diff
@@ -61,7 +62,8 @@ func (f *fuzzsecret) Setup(t *testing.T) []framework.Option {
 		for *s == "" ||
 			takenNames[*s] ||
 			len(path.IsValidPathSegmentName(*s)) > 0 ||
-			!reg.MatchString(*s) {
+			!reg.MatchString(*s) ||
+			strings.HasSuffix(*s, ".") {
```
JoshVanL, line 66: It previously worked to have dots as a suffix for both state and secret keys. What's changed that this no longer works?

tests/integration/suite/daprd/state/http/compname.go
```diff
@@ -92,7 +92,7 @@ func (c *componentName) Run(t *testing.T, ctx context.Context) {
 		storeName := storeName
 		t.Run(storeName, func(t *testing.T) {
 			t.Parallel()
-			reqURL := fmt.Sprintf("http://localhost:%d/v1.0/state/%s", c.daprd.HTTPPort(), url.PathEscape(storeName))
+			reqURL := fmt.Sprintf("http://localhost:%d/v1.0/state/%s", c.daprd.HTTPPort(), url.QueryEscape(storeName))
```
JoshVanL, line 95: Thanks- I thought  I had it to `QueryEscape` for a reason :smile:

tests/integration/suite/daprd/state/http/fuzz.go
```diff
@@ -74,7 +75,7 @@ func (f *fuzzstate) Setup(t *testing.T) []framework.Option {
 	fuzzFuncs := []any{
 		func(s *saveReqBinary, c fuzz.Continue) {
 			var ok bool
-			for len(s.Key) == 0 || strings.Contains(s.Key, "||") || ok {
+			for len(s.Key) == 0 || strings.Contains(s.Key, "||") || strings.HasSuffix(s.Key, ".") || ok {
```
JoshVanL, line 78: Again re dots as a suffix. No longer supporting dots at the end of keys would be a breaking change to users. If this is no longer supported, we should have a test case specifically for it, showing it produces a user readable error, and is documented.

Please not this ran hundreds of thousands of times previously with no issue.

tests/integration/suite/daprd/state/http/fuzz.go
```diff
@@ -153,7 +155,10 @@ func (f *fuzzstate) Run(t *testing.T, ctx context.Context) {
 	t.Run("get", func(t *testing.T) {
 		t.Parallel()
 		for i := range f.getFuzzKeys {
-			getURL := fmt.Sprintf("http://localhost:%d/v1.0/state/%s/%s", f.daprd.HTTPPort(), url.PathEscape(f.storeName), url.PathEscape(f.getFuzzKeys[i]))
+			getURL := fmt.Sprintf("http://localhost:%d/v1.0/state/%s/%s", f.daprd.HTTPPort(), url.QueryEscape(f.storeName), url.QueryEscape(f.getFuzzKeys[i]))
+			t.Log("URL", getURL)
+			t.Log("State store name", f.storeName, hex.EncodeToString([]byte(f.storeName)), printRunes(f.storeName))
+			t.Log("Key", f.getFuzzKeys[i], printRunes(f.getFuzzKeys[i]))
```
JoshVanL, line 161: I don't think we should add logs here. This massively balloons the output. Ideally if something errors, all information will be included in the `assert`/`require` output.

JoshVanL (changes requested):
Thank you for creating the PR for the issues raised @ItalyPaleAle 

I'm still concerned about the 2 issues raised though- no longer allowing dot prefix is key names, and the change in behavior to Service invocation, by which the router no longer strips slashes before sending the request to the app.

tests/integration/suite/daprd/serviceinvocation/http/basic.go
ItalyPaleAle: I guess this could be restored by removing extra slashes at the beginning. Let me see if that can work

tests/integration/suite/daprd/secret/http/fuzz.go
ItalyPaleAle: What doesn't work is when a path component is only a dot, such as `state/./.` which happened sometimes in fuzzed tests. This doesn't work because we clean the path using `path.Clean`, which simplifies the dots.

I guess allowing a dot at the end may still work, as long as it's not only a dot. I can change this

tests/integration/suite/daprd/state/http/fuzz.go
ItalyPaleAle: I responded above.

tests/integration/suite/daprd/state/http/fuzz.go
ItalyPaleAle: These logs should appear only if the test fails? I can comment them out if needed.

tests/integration/suite/daprd/serviceinvocation/http/basic.go
ItalyPaleAle: This has been fixed in #6594. Will update the tests in that PR once this is merged

tests/integration/suite/daprd/state/http/fuzz.go
ItalyPaleAle: Changed

tests/integration/suite/daprd/state/http/fuzz.go
JoshVanL: It's just because if something significant is wrong it balloons the logs and makes text editors/terminals sluggish.

JoshVanL (approved):

tests/integration/suite/daprd/state/http/fuzz.go
ItalyPaleAle: Ok I can comment those out. I would like to leave them in case they are useful in the future to debug more issues.

ItalyPaleAle: Done

tests/integration/suite/daprd/serviceinvocation/http/basic.go
ItalyPaleAle: I added a unit test in there. I would rather not make changes to these tests there as once this PR is merged, it will cause more merge conflicts.

artursouza (approved):

PR #6509, opened 2023-06-14
Add direction for bindings

yaron2:
Closes https://github.com/dapr/dapr/issues/6449.

PR #6608, opened 2023-06-29
Skip OPTIONS request for input binding if direction is specified

yaron2:
Follow up for https://github.com/dapr/dapr/pull/6509, this completes the feature by not probing the app with an OPTIONS call if the user explicitly set the binding as input.

### DIFF
--- a/pkg/runtime/runtime.go
+++ b/pkg/runtime/runtime.go
@@ -1721,6 +1721,22 @@ func isBindingOfDirection(direction string, metadata []commonapi.NameValuePair) 
 	return !directionFound
 }
 
+func isBindingOfExplicitDirection(direction string, metadata map[string]string) bool {
+	for k, v := range metadata {
+		if strings.EqualFold(k, bindingDirection) {
+
+			directions := strings.Split(v, ",")
+			for _, d := range directions {
+				if strings.TrimSpace(strings.ToLower(d)) == direction {
+					return true
+				}
+			}
+		}
+	}
+
+	return false
+}
+
 func (a *DaprRuntime) initInputBinding(c componentsV1alpha1.Component) error {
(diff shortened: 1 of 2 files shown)


pkg/runtime/runtime.go
```diff
@@ -3476,10 +3491,18 @@ func (a *DaprRuntime) startReadingFromBindings() (err error) {
...
 
 	for name, binding := range a.compStore.ListInputBindings() {
-		isSubscribed, err := a.isAppSubscribedToBinding(name)
-		if err != nil {
-			return err
+		var isSubscribed bool
+		m := binding.GetComponentMetadata()
+
+		if isBindingOfExplicitDirection(inputBinding, m) {
+			isSubscribed = true
```
ItalyPaleAle, line 3498: Is this a fair assumption to make? What if the user is using a Component YAML that is shared among multiple apps?

What I thought this PR was about, was skipping the OPTIONS request if an input binding is only used as output.

yaron2: Yeah, the whole point of the OPTIONS request to begin with was to test if a binding is an input by expecting the app to response successfully to the OPTIONS request. If the user explicitly sets the direction, the call is no longer needed..

ItalyPaleAle: Ok. Do we also need the opposite? If the binding is marked as output explicitly, omit the OPTIONS request and say `isSubscribed=false`?

yaron2: > What I thought this PR was about, was skipping the OPTIONS request if an input binding is only used as output.

Yep that was achieved with https://github.com/dapr/dapr/pull/6509

ItalyPaleAle (approved):

Issue #6611, opened 2023-06-30
Dapr control plane components do not adapt to run in a single namespace


jjcollinge:
## In what area(s)?

/area runtime

## What version of Dapr?

> master

## Expected Behavior

Dapr components respect the limited RBAC permissions when using `global.rbac.namespaced` and work within a single namespace.

## Actual Behavior

The Dapr control plane components do not appear to respect the limited RBAC permissions and still require cluster level permissions to operate successfully leading to crash loop backoff. For instance the operators attempts to read cluster level CRDs to patch them even though it does not have permission to do so.

- https://github.com/dapr/dapr/blob/83d1727204b6554e302199ab1693033b54bdd6a6/pkg/sentry/config/config.go#L113
- https://github.com/dapr/dapr/blob/83d1727204b6554e302199ab1693033b54bdd6a6/pkg/injector/pod_patch.go#L167
- also in the operator code

## Steps to Reproduce the Problem
Use the helm value `global.rbac.namespaced=true` and deploy the Dapr control plane to a namespace. The control plane components will all go into crash loop backoff.

I'll look to fix this in a follow up PR.

## Release Note

RELEASE NOTE: FIX dapr control plane crashing when `global.rbac.namespaced` enabled.

PR #6613, opened 2023-06-30
Tweaks to respect namespace when using global.rbac.namespaced

jjcollinge:
Signed-off-by: Joni Collinge <jonathancollinge@live.com>

# Description

> NOTE: This is still WIP whilst I test various things. Also this is a hack aimed at getting a Dapr control plane deployment working within a namespace - this may bastardize things not originally intended for this use case so we can add new settings where required.

Once you enable `global.rbac.namespaced` the Dapr control plane gets RBAC permissions limited to a single namespace. Currently the Dapr control plane still tries to access resources outside of that namespace and then crashes.
Changes required to make this work:
- Update sentry to work the same as the operator and use the Kubernetes downward API to get the namespace it is deployed in and get the dapr configuration from there. (is this a breaking change?)
- Update the sidecar injector to do the same when loading the configuration. (is this a breaking change?)
- Update the mutating webhook to use a namespace selector limited to the deployment namespace.
- Add a new flag to the operator `no-crd-patch` which stops the operator trying to patch the cluster level CRD resources with the expectation that the CRDs will be installed and patched separately. This is needed as the operator will never have permission to access the cluster level CRD resources with the current RBAC restrictions. This needs further investigation to understand how to handle the cabundle in the conversion webhook properly externally.
- Updates the RBAC to NOT deploy cluster role or cluster role binding when running in `namespaced` mode and move permissions down to role. This needs further reviewing to see if it'll cause any issues.

## Issue reference

Please reference the issue this PR will close: https://github.com/dapr/dapr/issues/6611

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing


charts/dapr/charts/dapr_rbac/templates/sentry.yaml
```diff
@@ -61,6 +61,9 @@ rules:
   - apiGroups: ["dapr.io"]
     resources: ["configurations"]
     verbs: ["list"]
+  - apiGroups: ["authentication.k8s.io"]
+    resources: ["tokenreviews"]
+    verbs: ["create"]
```
JoshVanL, line 66: Token reviews are not namespaced and must be cluster role bound.

JoshVanL (changes requested):
How can we test a namespaced deployment of Dapr in CI?

charts/dapr/charts/dapr_rbac/templates/sentry.yaml
jjcollinge: Yes and I was waiting for this to cause me an issue with validation... not seen anything yet!

JoshVanL: Yes, today this won't do much but will cause issues in 1.12 when clients begin requesting their certs from sentry for real.

lguianu:
Hey @jjcollinge, have you been able to find a fix for when we enable global.rbac.namespaced and it's not working?

Issue #6544, opened 2023-06-19
Workflow save internal states in single attempt failed with Azure Cosmos DB backed State Store due to Cosmos DB's Maximum Single Operation items limit


softnado:
## In what area(s)?

/area runtime

> /area operator

> /area placement

> /area docs

> /area test-and-release

## What version of Dapr?

> 1.1.x
> 1.0.x

1.11.x

> edge: output of `git describe --dirty`

## Expected Behavior

Expect Dapr Workflow working with Azure Cosmos DB backed state store for save the internal states. The same logic working with the Azure Redis Cache backed state store.

## Actual Behavior

Dapr Workflow failed to save internal states due to it attempt to save more than 100 states and blocked by Azure Cosmos DB's single operation maximum items limit.

## Steps to Reproduce the Problem

We have one workflow with around 12 stages in sequence, 8 of them fan out 20 parallel activities. When running with Redis backed state store, everything working as expected. When running with Azure Cosmos DB backed state store, it stuck with failing to save internal states and some Azure Cosmos DB error indicate it try to save more than 100 states in single request.

## Release Note

RELEASE NOTE:

berndverst:
This was discussed in the Discord workflow channel.

@cgillum can you assign this one to yourself or someone working on workflows please?

softnado:
With 1.11.1, the behavior changed and still not working. Here is the snapshot of the log.

<img width="906" alt="image" src="https://github.com/dapr/dapr/assets/2173696/6c49a124-90fd-4d80-9073-6bbd9d22f804">

It failed to process activities that call custom actor and report errors as above.

cgillum:
I can assign this to myself for now. Or @DeepanshuA if you have cycles, this might be a good one to load balance your way.

There are potentially two different issues that need to be addressed as part of resolving this bug:

1. Why are users seeing large batch sizes for simple workflows? This shouldn't be happening unless the workflow fans-out to multiple operations in parallel.
2. For the cases where large batch sizes are legitimate, we should save them in smaller, configurable batch sizes. To ensure compatibility with Cosmos DB, the default batch size should be 100.

Let's first use the steps above to try and reproduce the problem locally. With a repro, we should be able to investigate and fix (1). Separately, we should do the work for (2). Depending on the size of each fix, this could be done in a single PR or in two separate PRs.

cgillum:
Regarding issue (1) mentioned above (why so many history events being saved), I was able to reproduce the issue using the following C# workflow code.

[...]

PR #6516, opened 2023-06-14
Workflow e2e test in .NET

RyanLettieri:
# Description

Created an E2E test for workflow using the .NET SDK

## Issue reference

Please reference the issue this PR will close: [#_6114_](https://github.com/dapr/dapr/issues/6114)

## Checklist

Please make sure you've  completed the relevant tasks for this PR, out of the following list:

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing
* [ ] End-to-end tests passing
* [ ] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_
* [ ] Specification has been updated / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_
* [ ] Provided sample for the feature / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_

PR #6617, opened 2023-07-03
Updating batch size for workflow

RyanLettieri:
# Description

Currently CosmosDB has a limitation on the amount of items that can be stored in a single transaction (100). This PR changes the amount of items that are saved in a single request to account for this limitation.

It is also worth noting that this PR should not be approved/closed until an E2E test is created and this test case is implemented inside the test. The referenced PR for that E2E test: #6516 

## Issue reference

This PR will partially close: #6544 
 Since the issue referenced above contains 2 separate issues, it will only partially close the issue and the other portion will still need to be addressed.

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing


pkg/runtime/wfengine/workflow.go
```diff
@@ -515,8 +515,35 @@ func (wf *workflowActor) saveInternalState(ctx context.Context, actorID string,
 	}
 
 	wfLogger.Debugf("%s: saving %d keys to actor state store", actorID, len(req.Operations))
-	if err = wf.actors.TransactionalStateOperation(ctx, req); err != nil {
-		return err
+
+	// Break down the operations into smaller groups since cosmosDB can only handle 100 at a time
+	maxOp := 100 // TODO: Make this into a user configurable parameter
```
cgillum, line 520: Let's make this configurable as part of the PR.

cgillum (changes requested):
This is a good start but it's vulnerable to crashes that may occur in between batches. To create a more robust fix, we need a more comprehensive refactoring of the workflow state management code, including code that schedules actions. Here's a quick summary of what needs to be done:

* State store records need to be batched into groups of N (N = 100 by default but must be configurable)
* Each batch MUST include an appropriately updated `metadata` record that has the correct values for the saved history length (this will be different for each batch since you're gradually adding more and more).
* To minimize possible duplicate executions due to failures, we need to interleave scheduling of actions, like creating timers, calling activities, and sub-workflows, with the saving of history batches. For example, if we are trying to save 500 history records corresponding to 500 parallel activity executions, we should also schedule the activity calls, etc. in batches of N. This may require more significant refactoring.

Also, at a minimum for testing, please add a unit test in wfengine_test.go that fans-out to a large number of parallel activity calls (N >= 100).

PR #6619, opened 2023-07-03
Register updated OAuth2 middlewares

ItalyPaleAle:
> Builds will fail until dapr/components-contrib#2964 is merged and a new contrib is pinned

See dapr/components-contrib#2964 for context


ItalyPaleAle:
Closing as we changed it to a single component

PR #6626, opened 2023-07-06
Some fixes for E2E tests

ItalyPaleAle:
Our E2E tests on AKS are currently failing basically 100% of times, with flaky tests.  Here are the last 2 pages:

image
image

Failures are varied, and some are things we've known for a long time. This PR tries to make some small improvements:

1. Increase the tolerance in healthcheck tests. Fixes things like https://github.com/dapr/dapr/actions/runs/5458991068/jobs/9934667511
2. Reduces the scope of many components. Right now, most components are scoped to all apps, and that may be causing some issues.
3. Lately, jobs (and also deployments, but mostly jobs) have been failing to deploy very frequently. It's unclear why, but I've added the ability to fetch container logs from failed pods too, so we can better inspect what's going on (and see if for example point 2 above helped). Example: https://github.com/dapr/dapr/actions/runs/5461475945/jobs/9939942323

This is not going to fix all issues. I am hoping it's a start however...


tests/config/app_pubsub_routing.yaml
ItalyPaleAle, line 1: This was deleted as the feature is not in preview anymore

mukundansundar (approved):
lgtm

Issue #6647, opened 2023-07-11
Workflow giving 404 when using cosmosdb as Actor state store


DeepanshuA:
## In what area(s)?

Had a conversation with @cgillum, in which he pointed out that using Cosmos db may be an issue right now with workflows.
Tried using cosmosdb and redis as actor state stores respectively for exactly same scenarios.
1. With redis, 12 keys were persisted and no error was received.

Keys:
```
 1) "actors||dapr.internal.default.orderapp.workflow"
 2) "orderapp||dapr.internal.default.orderapp.workflow||f276f281dd1a48fb859e37484919ff0f||metadata"
 3) "orderapp||dapr.internal.default.orderapp.workflow||f276f281dd1a48fb859e37484919ff0f||history-000004"
 4) "orderapp||dapr.internal.default.orderapp.workflow||f276f281dd1a48fb859e37484919ff0f||history-000005" 
 5) "orderapp||dapr.internal.default.orderapp.workflow||f276f281dd1a48fb859e37484919ff0f||history-000003"
 6) "actors||dapr.internal.default.orderapp.activity||metadata"
 7) "orderapp||dapr.internal.default.orderapp.workflow||f276f281dd1a48fb859e37484919ff0f||history-000002"
 8) "orderapp||dapr.internal.default.orderapp.workflow||f276f281dd1a48fb859e37484919ff0f||history-000000"
 9) "orderapp||dapr.internal.default.orderapp.workflow||f276f281dd1a48fb859e37484919ff0f||history-000001"
10) "orderapp||dapr.internal.default.orderapp.workflow||f276f281dd1a48fb859e37484919ff0f||customStatus"
11) "actors||dapr.internal.default.orderapp.workflow||metadata"
12) "actors||dapr.internal.default.orderapp.activity"
```
2. While with cosmosdb, 10 keys were persisted and some 404 errors were noticed.
Keys:
```
 1) "actors||dapr.internal.default.orderapp.workflow"
 2) "actors||dapr.internal.default.orderapp.workflow||metadata"
 3) "orderapp||dapr.internal.default.orderapp.workflow||01f36664c64d4f48802d985a42dfb202||customStatus"
 4) "orderapp||dapr.internal.default.orderapp.workflow||01f36664c64d4f48802d985a42dfb202||metadata"
 5) "actors||dapr.internal.default.orderapp.activity"
 6) "actors||dapr.internal.default.orderapp.activity||metadata"
[...]

PR #6650, opened 2023-07-11
Update pinned components-contrib

ItalyPaleAle:
Updates pinned components-contrib to include:

- dapr/components-contrib#2984 which makes the GetComponentMetadata method optional.
- dapr/components-contrib#3015 which adds `MultiMaxSize` to transactional state stores. This is useful with components like Cosmos DB which have a limit in the max number of transactions. Needed by #6647

PS: Closes #6701, closes #6718, closes #6719 too


mukundansundar:
> Updates pinned components-contrib to include [dapr/components-contrib#2984](https://github.com/dapr/components-contrib/pull/2984) which makes the GetComponentMetadata method optional.

Just a clarity, this method was originally added for metadata builder only and not for dapr runtime to use. Now a different option is there with the above PR merge and thus this is not needed to be implemented anymore. 
Is my understanding correct?

mukundansundar (approved):

ItalyPaleAle:
JS tests are currently broken in master. See https://github.com/dapr/dapr/pull/6726

mukundansundar (approved):

PR #6659, opened 2023-07-13
Update DurableTask-Go to fix issues with workflows hanging

### LINKED ISSUE #6648: Problems with Workflows on AKS
## In what area(s)?
/area runtime
> /area operator
> /area placement
> /area docs
> /area test-and-release

## What version of Dapr?
1.11.x

## Expected Behavior
A simple workflow app in an AKS (k8s generically?) should function.

## Actual Behavior
The workflow fails to start and remains in pending forever. The pending error is a timeout when trying to start/communicate with the orchestrator. The app itself claims to have a stable connection to the workflow engine and we can see logs in Dapr for the workflow engine starting.

Failure:
```json
{"app_id":"testbed-workflow","instance":"testbed-workflow-app-5bdfcf687-84h2g","level":"info","msg":"1234: starting new 'OrderProcessingWorkflow' instance with ID = '1234'.","scope":"dapr.runtime.wfengine","time":"2023-07-10T21:34:33.00047503Z","type":"log","ver":"1.11.1"}
{"app_id":"testbed-workflow","instance":"testbed-workflow-app-5bdfcf687-84h2g","level":"warning","msg":"1234: execution timed-out and will be retried later","scope":"dapr.runtime.wfengine","time":"2023-07-10T21:35:03.001311696Z","type":"log","ver":"1.11.1"}
```

Logs after ~2 hours of failure:
```json
{"app_id":"testbed-workflow","instance":"testbed-workflow-app-5bdfcf687-84h2g","level":"warning","msg":"1234: execution failed with a recoverable error and will be retried later: timed-out trying to schedule a workflow execution - this can happen if there are too many in-flight workflows or if the workflow engine isn't running","scope":"dapr.runtime.wfengine","time":"2023-07-10T21:37:03.001908384Z","type":"log","ver":"1.11.1"}
```

The app in question works locally and in docker desktop (docker-compose). Interestingly enough, it also works in _some_ AKS clusters, but not all. KIND is also having this same problem. The issue has been reproduced on with both Redis and CosmosDB as the underlying statestore.

1. Create an AKS cluster.
2. Install Dapr 1.11.1 on the cluster.
3. Install Redis on the cluster.
4. Add component (actor statestore)/app.
5. Start workflow.

## Release Note

RELEASE NOTE: **FIX**: Intermittent issues in Dapr Workflows on AKS.

ItalyPaleAle:
Fixes #6648

The issue was with the fact that the Dapr server, on K8s, by default creates two listeners, on `127.0.0.1` and on `[::1]`. Each listener was getting its own instance of the workflow backend, so depending on "what localhost" each app was communicating with, it could hit a separate workflow backend, often causing a deadlock.


pkg/runtime/wfengine/wfengine.go
```diff
@@ -168,7 +172,7 @@ func (wfe *WorkflowEngine) SetActorReminderInterval(interval time.Duration) {
 	wfe.activityActor.reminderInterval = interval
 }
 
-func (wfe *WorkflowEngine) Start(ctx context.Context) error {
+func (wfe *WorkflowEngine) Start(ctx context.Context) (err error) {
```
cgillum, line 175: What's the reason for this change?

pkg/runtime/wfengine/wfengine.go
```diff
@@ -53,6 +54,7 @@ const (
 
 var (
 	wfLogger            = logger.NewLogger("dapr.runtime.wfengine")
+	wfBackendLogger     = logger.NewLogger("wfengine.backend")
```
cgillum, line 57: Good idea

pkg/runtime/runtime.go
```diff
@@ -258,6 +258,9 @@ type pubsubSubscribedMessage struct {
 func newDaprRuntime(runtimeConfig *internalConfig, globalConfig *config.Configuration, accessControlList *config.AccessControlList, resiliencyProvider resiliency.Provider) *DaprRuntime {
 	ctx, cancel := context.WithCancel(context.Background())
 
+	wfe := wfengine.NewWorkflowEngine(wfengine.NewWorkflowConfig(runtimeConfig.id))
+	wfe.ConfigureGrpcExecutor()
```
cgillum, line 262: It looks like in all cases we call `ConfigureGrpcExecutor()` immediately after creating the engine. Any reason not to just put this logic inside `NewWorkflowEngine(...)' to avoid the extra explicit step?

pkg/runtime/wfengine/wfengine.go
```diff
@@ -210,6 +214,10 @@ func (wfe *WorkflowEngine) Stop(ctx context.Context) error {
 	wfe.startMutex.Lock()
 	defer wfe.startMutex.Unlock()
 
+	if !wfe.IsRunning {
```
cgillum, line 217: It looks to me like we removed all assignments to `wfe.IsRunning`, but we still check it here and in `Start(...)`. Did I miss something?

pkg/runtime/runtime.go
cgillum: Ah, or is it because you need to configure an alternate executor in the test code? In that case, I think this separation makes sense.

cgillum (commented):
Thanks for making this change (and for taking the hit of incorporating the breaking changes in the test code)! A few comments/questions, but overall this looks good to me.

pkg/runtime/runtime.go
ItalyPaleAle: Correct, I originally put it inside NewWorkfowEngine but then realized I needed a different one for unit tests

pkg/runtime/wfengine/wfengine.go
ItalyPaleAle: (For other reviewers) This helped figuring out where the logs were coming from, since the dtf-go library uses this logger too

pkg/runtime/wfengine/wfengine.go
ItalyPaleAle: Oh it must have been something I did for some changes and then didn’t need it anymore. I’ll revert this

ItalyPaleAle: Just so the `err` inside the loop doesn’t need to be instantiated in the loop. A bit like a small nit of mine :)

pkg/runtime/wfengine/wfengine.go
ItalyPaleAle: Whoops that was a mistake. WHile working on this I made some changes and then didn’t revert them correctly. I’ll fix this tomorrow. Thanks for catching it

ItalyPaleAle: Fixed

pkg/runtime/wfengine/activity.go
```diff
@@ -168,24 +168,30 @@ func (a *activityActor) executeActivity(ctx context.Context, actorID string, nam
...
 		return newRecoverableError(fmt.Errorf("failed to schedule an activity execution: %w", err))
 	}
 
 loop:
 	for {
+		t := time.NewTimer(10 * time.Minute)
 		select {
 		case <-ctx.Done():
+			if !t.Stop() {
+				<-t.C
```
cgillum, line 182: Just curious: it's my understanding that `t.Stop()` returns false if the timer has already fired. One thing I didn't quite grasp is why we still need to read from the channel in such cases. Is it because leaving the "message" in the timer channel could cause a memory leak?

cgillum (approved):
LGTM! Thanks again for making these fixes (and all the additional small improvements)!

pkg/runtime/wfengine/activity.go
ItalyPaleAle: That's my understanding too. (*Insert note on how Go can't have "leaks" in the proper sense (unless there's a goroutine leak), more about delayed collection by the GC*)

ItalyPaleAle: In this case, the timer should not have fired because that's a separate `case` in the switch

go.mod
```diff
@@ -25,20 +25,20 @@ require (
...
-	github.com/hashicorp/golang-lru/v2 v2.0.2
+	github.com/hashicorp/golang-lru/v2 v2.0.4
 	github.com/hashicorp/raft v1.4.0
 	github.com/hashicorp/raft-boltdb v0.0.0-20230125174641-2a8082862702
 	github.com/jhump/protoreflect v1.15.1
 	github.com/kelseyhightower/envconfig v1.4.0
 	github.com/labstack/gommon v0.4.0
 	github.com/lestrrat-go/jwx/v2 v2.0.11
-	github.com/microsoft/durabletask-go v0.2.4
+	github.com/microsoft/durabletask-go v0.2.5-0.20230713052746-460a8a64a386
```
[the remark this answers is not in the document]
cgillum: I've published the [v0.3.0 release tag for durabletask-go](https://github.com/microsoft/durabletask-go/releases/tag/v0.3.0). Can you update this reference to use it?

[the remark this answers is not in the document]
ItalyPaleAle: Done

cgillum (approved):
LGTM!

yaron2 (approved):

PR #6674, opened 2023-07-15
reduce allocations on hot path for metrics

filintod:
# Description

While troubleshooting a user memory problem, and checking pprof output it was notice that we have a good amount of these allocations that could be improved.

This is not the main contributor to the memory allocation issue but a small improvement

## Issue reference

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing

### DIFF
--- a/pkg/diagnostics/grpc_monitoring.go
+++ b/pkg/diagnostics/grpc_monitoring.go
@@ -230,11 +230,11 @@ func (g *grpcMetrics) AppHealthProbeCompleted(ctx context.Context, status string
 		elapsed := float64(time.Since(start) / time.Millisecond)
 		stats.RecordWithTags(
 			ctx,
-			diagUtils.WithTags(g.healthProbeCompletedCount.Name(), g.appID, KeyClientStatus, status),
+			diagUtils.WithTags(g.healthProbeCompletedCount.Name(), appIDKey, g.appID, KeyClientStatus, status),
 			g.healthProbeCompletedCount.M(1))
 		stats.RecordWithTags(
 			ctx,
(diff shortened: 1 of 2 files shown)


ItalyPaleAle (dismissed):
LGTM

Do you think this is a place where using a sync.Pool may help?

filintod:
maybe not a bad idea, the max number of elements I saw from a grep in pkg is 10. Let me run a benchmark to see if the added logic improves this without much or any performance hit

filintod:
@ItalyPaleAle I did not see a big difference adding the sync.Pool as there is still another allocation that would require another kind of cache for the tag.Upsert that is adding allocations each time in https://github.com/dapr/dapr/pull/6674/files#diff-bb2a89c3078a94caa5e92dcb3e74a19afa60e05a5fb4b5fd29a52d65628fb66cR69. Maybe this can be added to an lru cache

Something else I did find a bug on grpc metrics where it is missing the appIDKey when adding the tags

ItalyPaleAle (approved):

ItalyPaleAle:
@filintod good find on the metrics stuff

As for the LRU cache, i think that may be an overkill, at least for now :)

filintod:
some interesting info that we should probably look for precomputed tags https://blog.howardjohn.info/posts/zero-alloc-otel/
that also hit Istio in the past

ItalyPaleAle:
@filintod that seems a very good idea. We sure do allocate a lot of memory in hot paths due to metrics. Can you open a separate issue for that?

filintod:
@ItalyPaleAle I created this issue to follow up https://github.com/dapr/dapr/issues/6680

Issue #6459, opened 2023-06-07
Add new Service Invocation metrics to proto-based Dapr gRPC proxy


antontroshin:
## In what area(s)?

/area runtime

> /area operator

> /area placement

> /area docs

> /area test-and-release

## Describe the feature

Add new Service Invocation metrics to proto-based Dapr's gRPC proxy.
The metrics were added in this issue https://github.com/dapr/dapr/issues/5484

Currently, the metrics are missing, to replicate, follow the guide: https://docs.dapr.io/developing-applications/building-blocks/service-invocation/howto-invoke-services-grpc/
After running the server or client as instructed and making sure requests went through, access the metrics port logged in `metrics server started on: <port number>` line, the metrics won't be there.

The only usage found in `direct_messaging.go` https://github.com/dapr/dapr/blob/0c8292c590208bd38886245de7ee1b66b84e4779/pkg/messaging/direct_messaging.go#L274
https://github.com/dapr/dapr/blob/0c8292c590208bd38886245de7ee1b66b84e4779/pkg/messaging/direct_messaging.go#L290

## Release Note

RELEASE NOTE: **ADD** Service Invocation metrics when using proto-based Dapr's gRPC proxying

artursouza:
This should also be validated in E2E tests to make sure it is working and no regression is added in the future.

AnnuCode:
Hi, I tried adding service invocation metrics in `pkg/messaging/grpc_proxy.go` in the `intercept` method, but it didn't work. The metrics do not appear on the metrics server address. I also tried adding the invocation metrics in `pkg/grpc/proxy/handler.go` file in the `handler` method, and it still didn't work. Am I missing some fundamental concept in the way proxying of grpc requests is happening? I thought the `grpc_proxy.go` file of `messaging` package is where the proxying is being executed. Any suggestions would be helpful, thank you.

AnnuCode:
@artursouza Sure, I made these changes(as shown in the draft pull request) to test out but it didn't work. Am I passing the wrong parameters in the service invocation methods?

PR #6676, opened 2023-07-15
adding service invocation metrics in grpc proxy

AnnuCode:
# Description 
### Steps performed:
- Made below mentioned changes in the code
- Compiled the new code to create a new binary of `dapr` and replaced it with the downloaded binary.
- Tested the code by running the grpc proxy example as mentioned in the docs.
- I couldn't get the required service invocation metrics in the metrics server.

 Even if I have placed these snippets at the wrong place, why does it not increment the invocation metrics counter and display them in the started metrics server(for the server side only, can't accsess client side metrics server location as the client side app closes after invoking the server)? 

### Changes:
- added ServiceInvocationRequestSent  metrics in `grpc_proxy.go` file in the `pkg/messaging` directory
- added ServiceInvocationRequestReceived in `handler.go` file in the `grpc/proxy` directory.  

## Issue reference
Please reference the issue this PR will close: #6459 

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing


yaron2:
@AnnuCode are you actively working on this?

yaron2:
> > @AnnuCode are you actively working on this?
> 
> Hi @yaron2 , I couldn't progress because of the doubt as mentioned in the description. I have also documented the approach which I took.

ok i'll take a look.

filintod:
@AnnuCode I have not tested but this is what I think:  the metrics should be incremented in the handler only but in a different place, just before the definition of the policyRunner in line 139 like:
```
// get current time to calculate elapsed
requestStartedAt := time.Now()
// increment metric counter
diagnostics.DefaultMonitoring.ServiceInvocationRequestReceived(v[0], fullMethodName)

policyRunner := resiliency.NewRunnerWithOptions(ctx, policyDef, resiliency.RunnerOpts[*proxyRunner]{
		Disposer: func(pr *proxyRunner) {
			pr.dispose()
		},
	})
pr, cErr := policyRunner(func(ctx context.Context) (*proxyRunner, error) {
    // connect to endpoint (inside resiliency)
    // if stream return proxy runner (pr) to execute outside resiliency
    // unary operation, execute inside resiliency
}

// calculate elapsed time to response on return 
defer func() {
  diagnostics.DefaultMonitoring.ServiceInvocationResponseReceived(v[0], fullMethodName, int32(status.Convert(cErr).Code()), requestStartedAt)
}

// if stream execute (currently line 248)
// If the policy function returned a proxy runner, execute it here, outside of the policy function so it's not influenced by timeouts
if pr != nil {
	cErr = pr.run()
	if cErr != nil {
		return cErr
	}
}
```

let me know if this helps, and if not I'll debug locally.  Where you have it now might not execute if it returns for certain error, or if using streaming

AnnuCode:
@filintod Thank you for explaining this issue here and also in your pull request 💡🙂

PR #6688, opened 2023-07-17
Refactor `runtime.go` to move component lifecycle managment into the runtime processor

JoshVanL:
PR moves the component lifecycle management to the relevant runtime processor. The intention of this PR is to further abstract each component type so that we can achieve graceful shutdown and hot reloading. This refactor should also help to improve the maintainability and test-ability of runtime.

Closing components are now done concurrently. Daprd will return exit 0 as soon as all components, servers, topics etc. have closed, and will fatal if the graceful timeout is exceeded during shutdown.


yaron2:
@JoshVanL does this PR contain behavioral changes as part of the refactor, like changes to shutdown order / component close sequence? (I assume this is all to support hot-reloading) - If so, it'd be helpful for reviewers to call these out on the description

pkg/runtime/runtime.go
```diff
@@ -2591,78 +1344,41 @@ func (a *DaprRuntime) stopActor() {
...
 	}
+	for _, comp := range a.compStore.ListComponents() {
+		closers = append(closers, closeComponent(comp))
+	}
 
-	return nil
+	return concurrency.NewRunnerManager(closers...).Run(ctx)
 }
 
 // ShutdownWithWait will gracefully stop runtime and wait outstanding operations.
```
yaron2, line 1384: Can you add tests specifically for `ShutdownWithWait` to ensure the order of operations and graceful shutdown period are executed correctly?

yaron2 (changes requested):

pkg/runtime/runtime.go
JoshVanL: Sure thing, I'll add those tests! I would like to point out though that the [follow-up daprd PR](https://github.com/dapr/dapr/pull/6720) addresses graceful shutdown specifically. Since this PR is already pretty out of hand as is, I didn't want to also include the wait group and channel changes which are added in that branch which will significantly improve the reliability of everything being fully closed before exit (as well as how that is tracked in runtime.go). Although we obviously strive to make the dapr master branch as correct/stable as possible- please keep in mind that the daprd shutdown handling here is going significantly change in the next PR.

yaron2: Ok, I'll accept the tests in the follow up PR.

yaron2 (approved):

tests/integration/suite/daprd/pubsub/grpc/compname.go
```diff
@@ -45,7 +45,7 @@ type componentName struct {
 }
 
 func (c *componentName) Setup(t *testing.T) []framework.Option {
-	const numTests = 1000
```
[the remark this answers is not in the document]
JoshVanL: @ItalyPaleAle, ergh. This was a mistake. I'll add a revert.

PR #6689, opened 2023-07-17
Integration tests: Wait for healthz endpoint to return ready for processes

JoshVanL:
Ensure dapr processes return healthy on their healthz endpoint before firing off the suite test cases.

This should ensure, particularly for daprd, that everything has been loaded in ready for the test execution.


tests/integration/framework/process/daprd/daprd.go
```diff
@@ -129,14 +129,16 @@ func (d *Daprd) Cleanup(t *testing.T) {
 }
 
 func (d *Daprd) WaitUntilRunning(t *testing.T, ctx context.Context) {
-	dialer := &net.Dialer{Timeout: time.Second * 5}
+	client := http.Client{Timeout: time.Second}
 	assert.Eventually(t, func() bool {
-		conn, err := dialer.DialContext(ctx, "tcp", fmt.Sprintf("localhost:%d", d.publicPort))
+		req, err := http.NewRequestWithContext(ctx, http.MethodGet, fmt.Sprintf("http://localhost:%d/v1.0/healthz", d.httpPort), nil)
+		require.NoError(t, err)
```
ItalyPaleAle, line 135: I don't think we are supposed to use assertions inside an `assert.Eventually` block. This is because this code runs in a separate goroutine.

Looks like there's an ["EventuallyWithT"](https://pkg.go.dev/github.com/stretchr/testify/assert#EventuallyWithT) we should use instead

JoshVanL: Thanks :+1:

ItalyPaleAle (approved):

PR #6703, opened 2023-07-20
Update HTTPEndpoint tls tests with generated PKI, with failure cases and 1 length chains.

JoshVanL:
In a post LLM world and general safety we should avoid committing private key material at all costs. Updates the tests so that PKI is ephemerally generated on the fly.

Uses cert-manager to generate the PKI for the service invocation TLS e2e tests.

Fixes the integration test's http server to correctly serve TLS when a TLS config has been provided. Call https endpoint in service invocation integration test so that the client certificate is actually used - prove with a failure case which involves a bad PKI setup.

Question: Currently, you can provide client TLS config in HTTPEndpoint for a non-HTTPS endpoint. This doesn't really make much sense and the client certificate won't even be used. I suggest that we should fail `Init` with a HTTPEndpoint with a client TLS config defined on a non-HTTPS base URL.

/cc @artursouza @yaron2


pkg/runtime/runtime_test.go
```diff
@@ -6096,6 +6102,31 @@ func TestStartReadingFromBindings(t *testing.T) {
 }
 
 func TestGetHTTPEndpointAppChannel(t *testing.T) {
+	testPK, err := ecdsa.GenerateKey(elliptic.P256(), rand.Reader)
+	require.NoError(t, err)
+
+	testPKBytes, err := x509.MarshalECPrivateKey(testPK)
```
ItalyPaleAle, line 6108: Nit: why not PKCS8 :)

tests/dapr_tests.mk
```diff
@@ -253,7 +253,7 @@ create-test-namespace:
 delete-test-namespace:
 	kubectl delete namespace $(DAPR_TEST_NAMESPACE)
 
-setup-3rd-party: setup-helm-init setup-test-env-redis setup-test-env-kafka setup-test-env-mongodb setup-test-env-zipkin
+setup-3rd-party: setup-helm-init setup-test-env-redis setup-test-env-kafka setup-test-env-mongodb setup-test-env-zipkin setup-test-env-cert-manager
```
ItalyPaleAle, line 256: I would really prefer if we didn't deploy yet another Helm chart in our test K8s clusters... Makes tests slower, requires more CPU, and can add to flakiness when a chart fails to deploy due to network issues (happens quite frequently).

Can we find a way that doesn't involve using cert-manager? How about we create the certs in a script and load them as secrets in K8s before we start the tests?

JoshVanL: @ItalyPaleAle No worries. We can generate them before the tests is executed as a setup step. Is it possible to get a Kubernetes client configured in the e2e test runner?

JoshVanL: Yes- we can pass the Secret resources as a `ComponentDescription` to the test runner.

yaron2:
> Question: Currently, you can provide client TLS config in HTTPEndpoint for a non-HTTPS endpoint. This doesn't really make much sense and the client certificate won't even be used. I suggest that we should fail Init with a HTTPEndpoint with a client TLS config defined on a non-HTTPS base URL.

Will this break use cases where the endpoint supports automatic https redirection?

JoshVanL:
> Will this break use cases where the endpoint supports automatic https redirection?

@yaron2 Thanks, that makes sense. We want to support HTTP endpoints which redirect to HTTPS, so we should keep it as is.

tests/platforms/kubernetes/secret.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+}
+
+func (s *Secret) Init(ctx context.Context) error {
+	log.Printf("Adding secret %q ...", s.name)
+	if err := s.Dispose(false); err != nil {
+		return err
+	}
+
+	_, err := s.kubeClient.ClientSet.CoreV1().Secrets(s.namespace).Create(context.TODO(), &corev1.Secret{
```
ItalyPaleAle, line 55: Would it make sense to set a timeout on this context?

Also, you could use `ctx` to cascade a context. Not that it matters much, it's test code anyways :)

tests/platforms/kubernetes/secret.go
```diff
@@ -0,0 +1,80 @@
...
+	return err
+}
+
+func (s *Secret) Name() string {
+	return s.name
+}
+
+func (s *Secret) Dispose(wait bool) error {
+	log.Printf("Delete secret %q ...", s.name)
+	_, err := s.kubeClient.ClientSet.CoreV1().Secrets(s.namespace).Get(context.TODO(), s.name, metav1.GetOptions{})
```
ItalyPaleAle, line 72: Maybe put a timeout on the context here too?

tests/platforms/kubernetes/secret.go
JoshVanL: Sounds good, done

tests/platforms/kubernetes/secret.go
JoshVanL: :+1:

ItalyPaleAle (dismissed):

PR #6704, opened 2023-07-20
Prevent github workflow from failing if windows integration tests fail

JoshVanL:
/cc @artursouza

### DIFF
--- a/.github/workflows/dapr.yml
+++ b/.github/workflows/dapr.yml
@@ -246,6 +246,7 @@ jobs:
             ${{ matrix.target_os }}-${{ matrix.target_arch }}-go-${{ steps.setup-go.outputs.go-version }}-build-
       - name: Run make test-integration
         run: make test-integration
+        continue-on-error: ${{ matrix.target_os == 'windows' }}
   build:
     name: "Build artifacts on ${{ matrix.job_name }} - ${{ matrix.sidecar_flavor }}"
     runs-on: "${{ matrix.os }}"


ItalyPaleAle (approved):
LGTM but can we also try to make the Windows tests work?

artursouza (approved):

JoshVanL:
> LGTM to unblock builds, but can we also try to make the Windows tests work?

I agree. I would like to know why daprd seems to completely break down in serving on CPU constrained environments- assuming that is the issue. It doesn't seem windows OS specific to me.

ItalyPaleAle:
> > LGTM to unblock builds, but can we also try to make the Windows tests work?
> 
> I agree. I would like to know why daprd seems to completely break down in serving on CPU constrained environments- assuming that is the issue. It doesn't seem windows OS specific to me.

We know from prior experience that the Windows agents are slower. Not sure exactly why, but it may be related to the OS being heavier and the fact that it runs more complex anti-malware. Even I/O is [significantly slower](https://github.com/actions/runner-images/issues/3577).

I also mentioned on Discord that I think we should change the ITs to run without the `-race` flag, since it makes tests much slower (and using more CPU), and it doesn't really help much in the case of ITs: since all binaries are pre-compiled, the flag can't help detecting race conditions there, and it can only detect race conditions in the test suite (which is of very little use IMHO).

JoshVanL:
> I also mentioned on Discord that I think we should change the ITs to run without the -race flag, since it makes tests much slower (and using more CPU), and it doesn't really help much in the case of ITs: since all binaries are pre-compiled, the flag can't help detecting race conditions there, and it can only detect race conditions in the test suite (which is of very little use IMHO).

I'll open a PR to remove the `-race` flag and we can see what the result is.

PR #6709, opened 2023-07-21
Remove `--race` flag from `make test-integration` to reduce resource constraint related flakes in CI.

JoshVanL:
See https://github.com/dapr/dapr/pull/6704#issuecomment-1644664791


ItalyPaleAle:
Is it still failing on building Dapr binaries?

I have an idea… how about we switch the order of the CI tasks and we first build Dapr, then run ITs?

this way, in ITs we can download the pre-compiled binaries.

JoshVanL:
> Is it still failing on building Dapr binaries?

@ItalyPaleAle no and I don't think that has ever been the case. [This](https://github.com/dapr/dapr/pull/6673) PR was created to address this failure.

artursouza (approved):

PR #6698, opened 2023-07-19
[WIP] Refactor injector to be used as a library

ItalyPaleAle:
**TODO:**

- [ ] Add more tests
- [ ] Perform additional validation

Fixes #6572 

Refactoring of `pkg/injector` so it can be used as a library in 3rd-party projects.

1. The code from the previous `pkg/injector` has been split into 2 packages (excluding "accessory" ones):
   - `pkg/injector/patcher` contains the utilities that patch a Pod resource to inject the Daprd container and all other relevant properties (this code was scattered a bit all over the place in multiple packages)
   - `pkg/injector/service` contains the code for the Injector service, i.e. the app that has the HTTP server responding to the webhook
2. `pkg/injector/patcher` can now be conveniently imported as a library. It works by:
   - The `SidecarConfig` struct can accept options to configure how Dapr has to be inejcted in a pod.
   - In addition to passing options "statically", `SidecarConfig` can parse options from a list of annotations
   - `SidecarConfig.GetPatch()` returns the patch that is to be applied to a Pod resource, using `github.com/evanphx/json-patch` (this is the same library that Kubernetes uses in its admission webhook)
   - This package has minimal dependencies on the rest of Dapr

There are no changes in functionality, and the injector continues to support all and only the same annotations. The injected Daprd has the same properties as before (although some CLI flags are now omitted if they had the default value).

PR #6712, opened 2023-07-21
Initialize UniversalAPI object only once

ItalyPaleAle:
When the HTTP and gRPC APIs are created, they now receive the same instance of the UniversalAPI object.

This should:

- Offer some (small) memory improvements
- Make it so the metadata API allows setting custom metadata using gRPC and read it using HTTP, and vice versa.
   - Once this is merged, I'll update the E2E test that is being added in #6698 to validate this behavior


artursouza (approved):

pkg/grpc/api.go
```diff
@@ -87,38 +85,20 @@ type api struct {
...
-	DirectMessaging             messaging.DirectMessaging
-	Actors                      actors.Actors
-	SendToOutputBindingFn       func(name string, req *bindings.InvokeRequest) (*bindings.InvokeResponse, error)
-	TracingSpec                 config.TracingSpec
-	AccessControlList           *config.AccessControlList
-	Shutdown                    func()
-	GetComponentsCapabilitiesFn func() map[string][]string
-	AppConnectionConfig         config.AppConnectionConfig
-	GlobalConfig                *config.Configuration
+	UniversalAPI          *universalapi.UniversalAPI
```
JoshVanL, line 88: Please revert so that these options are passed by value, and _not_ a pointer to an instantiated `*universalapi.UniversalAPI`. I don't think the code needs to be changed here at all, and is going against the recent efforts to decouple packages.

Ideally, `universalapi.UniversalAPI` itself should not have any exported fields, take in an `Options`, and use `funcs` to return fields needed by consumers.

This is part of the larger effort to ensure there is proper separation of concerns between packages WRT memory manipulation,

JoshVanL (changes requested):

pkg/grpc/api.go
ItalyPaleAle: I'm afraid I cannot do what you're asking.

`universalapi.UniversalAPI` must have exported fields because `grpc.api` inherits from that, and it's in a separate package.

Having the same struct (a pointer) is needed to make it so things like metadata API access the same metadata properties.

JoshVanL: > universalapi.UniversalAPI must have exported fields because grpc.api inherits from that, and it's in a separate package.

ItalyPaleAle: Can we do that in a separate PR? That seems to be a rather large change

JoshVanL: I had a go at doing this [here](https://github.com/ItalyPaleAle/dapr/pull/97), PTAL

ItalyPaleAle: Although I do like the changes, please let’s do that in a separate PR in Dapr/dapr so this bug fix can be merged more quickly.

pkg/runtime/runtime.go
```diff
@@ -409,38 +410,40 @@ func (a *DaprRuntime) initRuntime(ctx context.Context) error {
...
 
+	// Start internal gRPC server (used for sidecar-to-sidecar communication)
 	err = a.startGRPCInternalServer(a.daprGRPCAPI, a.runtimeConfig.internalGRPCPort)
 	if err != nil {
 		log.Fatalf("failed to start internal gRPC server: %s", err)
 	}
-	log.Infof("internal gRPC server is running on port %v", a.runtimeConfig.internalGRPCPort)
+	log.Infof("Internal gRPC server is running on port %v", a.runtimeConfig.internalGRPCPort)
+
+	a.initDirectMessaging(a.nameResolver)
```
yaron2, line 446: Any reason `initDirectMessaging` was moved here? Otherwise LGTM

ItalyPaleAle: I don't think it makes a significant difference, but the previous flow seemed weird:

1. Init gRPC API server
2. Init direct messaging
3. Init HTTP API server
4. Init internal gRPC server

It seemed best to move it to either before or after, so all servers are initialized at the same time. I landed on after for no specific reason.

mukundansundar (approved):

PR #6713, opened 2023-07-21
Remove `continue-on-error` from Windows integration test step

JoshVanL:
After https://github.com/dapr/dapr/pull/6709 and https://github.com/dapr/dapr/pull/6673, `continue-on-error` should no longer be necessary for the windows integration tests.


JoshVanL:
@artursouza The windows integration tests are now successfully passing reliably on this PR so we can re-enable them.

daixiang0 (approved):

PR #6714, opened 2023-07-22
Added protected topics to pub/sub

### LINKED ISSUE #5843: Proposal: protected topics for pub/sub
## In what areas?

> /area runtime

## Proposal

In #1296, @yaron2 has introduced the scopes for pub/sub. As described in [the docs](https://docs.dapr.io/developing-applications/building-blocks/pubsub/pubsub-scopes/), one can limit the scopes like this:
```yaml
apiVersion: dapr.io/v1alpha1
kind: Component
metadata:
  name: pubsub
spec:
  type: pubsub.redis
  version: v1
  metadata:
  - name: redisHost
    value: "localhost:6379"
  - name: redisPassword
    value: ""
  - name: allowedTopics
    value: "A,B"
  - name: publishingScopes
    value: "app1=A"
  - name: subscriptionScopes
    value: "app1=;app2=A"
```
so that `app1` can only publish to topic `A`, but subscribe to no topic, because it is explicitly forbidden to subscribe to any topic in `subscriptionScopes`.

Futhermore, `app3` is not listed in any of the scopes an can do whatever it wants with the allowed topics:
| subscribe to: | A | B | C |
| --- | --- | --- | --- |
| app1 | | | |
| app2 | X | | |
| app3 | X | X | |

If the data published on one of the topics is sensitive, every new app has to be explicitly listed in `subscriptionScopes` to prevent it from subscribing to that topic. This makes it impossible to protect a topic.

So I suggest adding a new field to the metadata:
```yaml
  - name: protectedTopics
    value: "A"
```
when this is added to the configuration above, `app3` would not be allowed to subscribe to topic `A` any more, because it is not explicitly allowed in the `subscriptionScopes` (but `app2` is):
| subscribe to: | A | B | C |
| --- | --- | --- | --- |
| app1 | | | |
| app2 | X | | |
| app3 | | X | |

---

### Note
This would leave the publishing permissions untouched:
| publish to: | A | B | C |
| --- | --- | --- | --- |
| app1 | X | | |
| app2 | X | X | |
| app3 | X | X | |

kovalromank:
# Description

Pub/sub topics can be marked as protected. Apps without explicit publish/subscribe scopes for a protected topic will not be able to publish/subscribe to it. Previous behaviour of scopes was maintained.

Updated tests to include all protected scope functionality.

## Issue reference

Closes #5843.

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Unit tests passing


kovalromank:
Plan on updating the documentation once the new configuration here is approved.

pkg/scopes/scopes.go
```diff
@@ -47,14 +48,13 @@ func GetScopedTopics(scope, appID string, metadata map[string]string) []string {
...
-// GetAllowedTopics return the all topics list of params allowedTopics.
-func GetAllowedTopics(metadata map[string]string) []string {
+func getParamTopics(param string, metadata map[string]string) []string {
 	var (
 		existM = map[string]struct{}{}
 		topics = []string{}
 	)
 
-	if val, ok := metadata[AllowedTopics]; ok && val != "" {
+	if val, ok := metadata[param]; ok && val != "" {
```
ItalyPaleAle, line 57: Please, make the match case-insensitive. We are trying to make all metadata properties match case-insensitive.

pkg/runtime/runtime.go
```diff
@@ -1948,7 +1948,7 @@ func (a *DaprRuntime) GetPubSub(pubsubName string) pubsub.PubSub {
 }
 
 func (a *DaprRuntime) isPubSubOperationAllowed(pubsubName string, topic string, scopedTopics []string) bool {
-	inAllowedTopics := false
+	inAllowedTopics, inProtectedTopics := false, false
```
ItalyPaleAle, line 1951: Nit, for readability:

```suggestion
	var inAllowedTopics, inProtectedTopics bool
```

ItalyPaleAle (commented):
Thanks for this PR!

pkg/scopes/scopes.go
kovalromank: Sounds good, updated in c0718be3

pkg/runtime/runtime.go
kovalromank: That does read better, fixed in c0718be3

ItalyPaleAle (approved):
Thank you!

pkg/runtime/runtime.go
```diff
@@ -1967,7 +1967,19 @@ func (a *DaprRuntime) isPubSubOperationAllowed(pubsubName string, topic string,
...
+		for _, t := range pubSub.ProtectedTopics {
+			if t == topic {
+				inProtectedTopics = true
+				break
+			}
+		}
+	}
+
+	// if topic is protected then a scope must be applied
+	if !inProtectedTopics && len(scopedTopics) == 0 {
```
yaron2, line 1982: Shouldn't this be `if inProtectedTopics && len(scopedTopics) == 0`?

yaron2 (approved):

Issue #6592, opened 2023-06-27
[Tracking issue] Actor subsystem improvements


ItalyPaleAle:
## Actor reminders and timers

Some ideas are collected in #6121

- [ ] #6040
- [ ] #6591
- [ ] #6503
- [ ] #6311
- [ ] #6150

## Placement

- Make Placement methods into an interface
- #6536
- #6265

## Related issues

- #5671
- #5403
- #6551

PR #6715, opened 2023-07-22
Remove `ActorScanInterval`

ItalyPaleAle:
Small refactoring to the actors subsystem to remove the "ActorScanInterval" configuration. Idle actors are now deactivated automatically shortly after the idle timeout for their actor type.

> The priority queue implemented in this PR in `pkg/actors/internal` is the same code used in #6716 . It should be possible to review and merge the 2 PRs independently.

Benefits:

### DIFF
--- a/pkg/actors/actor.go
+++ b/pkg/actors/actor.go
@@ -22,6 +22,7 @@ import (
 	"k8s.io/utils/clock"
 
 	diag "github.com/dapr/dapr/pkg/diagnostics"
+	"github.com/dapr/kit/ptr"
 )
 
 // ErrActorDisposed is the error when runtime tries to hold the lock of the disposed actor.
@@ -39,11 +40,12 @@ type actor struct {
 	// pendingActorCalls is the number of the current pending actor calls by turn-based concurrency.
 	pendingActorCalls atomic.Int32
 
-	// When consistent hashing tables are updated, actor runtime drains actor to rebalance actors
-	// across actor hosts after drainOngoingCallTimeout or until all pending actor calls are completed.
-	// lastUsedTime is the time when the last actor call holds lock. This is used to calculate
-	// the duration of ongoing calls to time out.
-	lastUsedTime time.Time
+	// idleTimeout is the configured max idle time for actors of this kind.
+	idleTimeout time.Duration
+
+	// idleAt is the time after which this actor is considered to be idle.
+	// When the actor is locked, idleAt is updated by adding the idleTimeout to the current time.
+	idleAt atomic.Pointer[time.Time]
 
 	// disposeLock guards disposed and disposeCh.
 	disposeLock sync.RWMutex
@@ -56,17 +58,33 @@ type actor struct {
 	clock clock.Clock
 }
 
-func newActor(actorType, actorID string, maxReentrancyDepth *int, cl clock.Clock) *actor {
+func newActor(actorType, actorID string, maxReentrancyDepth *int, idleTimeout time.Duration, cl clock.Clock) *actor {
 	if cl == nil {
 		cl = &clock.RealClock{}
 	}
-	return &actor{
-		actorType:    actorType,
-		actorID:      actorID,
-		actorLock:    NewActorLock(int32(*maxReentrancyDepth)),
-		clock:        cl,
-		lastUsedTime: cl.Now().UTC(),
+
+	a := &actor{
+		actorType:   actorType,
+		actorID:     actorID,
+		actorLock:   NewActorLock(int32(*maxReentrancyDepth)),
+		clock:       cl,
+		idleTimeout: idleTimeout,
 	}
+	a.idleAt.Store(ptr.Of(cl.Now().Add(idleTimeout)))
+
+	return a
+}
+
+// Key returns the key for this unique actor.
+// This is implemented to comply with the queueable interface.
+func (a *actor) Key() string {
+	return a.actorType + daprSeparator + a.actorID
+}
+
+// ScheduledTime returns the time the actor becomes idle at.
+// This is implemented to comply with the queueable interface.
+func (a *actor) ScheduledTime() time.Time {
+	return *a.idleAt.Load()
 }
 
 // isBusy returns true when pending actor calls are ongoing.
@@ -115,7 +133,8 @@ func (a *actor) lock(reentrancyID *string) error {
 		a.unlock()
 		return ErrActorDisposed
 	}
-	a.lastUsedTime = a.clock.Now().UTC()
+
+	a.idleAt.Store(ptr.Of(a.clock.Now().Add(a.idleTimeout)))
 	return nil
 }
 
--- a/pkg/actors/actor_test.go
+++ b/pkg/actors/actor_test.go
@@ -26,7 +26,7 @@ import (
 var reentrancyStackDepth = 32
 
 func TestIsBusy(t *testing.T) {
-	testActor := newActor("testType", "testID", &reentrancyStackDepth, nil)
+	testActor := newActor("testType", "testID", &reentrancyStackDepth, time.Second, nil)
 
 	testActor.lock(nil)
 	assert.Equal(t, true, testActor.isBusy())
@@ -34,12 +34,12 @@ func TestIsBusy(t *testing.T) {
 }
 
 func TestTurnBasedConcurrencyLocks(t *testing.T) {
-	testActor := newActor("testType", "testID", &reentrancyStackDepth, nil)
+	testActor := newActor("testType", "testID", &reentrancyStackDepth, time.Second, nil)
 
 	// first lock
 	testActor.lock(nil)
 	assert.Equal(t, true, testActor.isBusy())
-	firstLockTime := testActor.lastUsedTime
+	firstIdleAt := *testActor.idleAt.Load()
 
 	waitCh := make(chan bool)
 
@@ -57,7 +57,7 @@ func TestTurnBasedConcurrencyLocks(t *testing.T) {
 	time.Sleep(10 * time.Millisecond)
 	assert.Equal(t, int32(2), testActor.pendingActorCalls.Load())
 	assert.True(t, testActor.isBusy())
-	assert.Equal(t, firstLockTime, testActor.lastUsedTime)
+	assert.Equal(t, firstIdleAt, *testActor.idleAt.Load())
 
 	// unlock the first lock
 	testActor.unlock()
@@ -70,12 +70,12 @@ func TestTurnBasedConcurrencyLocks(t *testing.T) {
 
 	assert.Equal(t, int32(0), testActor.pendingActorCalls.Load())
 	assert.False(t, testActor.isBusy())
-	assert.True(t, testActor.lastUsedTime.Sub(firstLockTime) >= 10*time.Millisecond)
+	assert.True(t, testActor.idleAt.Load().Sub(firstIdleAt) >= 10*time.Millisecond)
 }
 
 func TestDisposedActor(t *testing.T) {
 	t.Run("not disposed", func(t *testing.T) {
-		testActor := newActor("testType", "testID", &reentrancyStackDepth, nil)
+		testActor := newActor("testType", "testID", &reentrancyStackDepth, time.Second, nil)
 
 		testActor.lock(nil)
 		testActor.unlock()
@@ -86,7 +86,7 @@ func TestDisposedActor(t *testing.T) {
 	})
 
 	t.Run("disposed", func(t *testing.T) {
-		testActor := newActor("testType", "testID", &reentrancyStackDepth, nil)
+		testActor := newActor("testType", "testID", &reentrancyStackDepth, time.Second, nil)
 
 		testActor.lock(nil)
 		ch := testActor.channel()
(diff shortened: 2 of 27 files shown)


artursouza (changes requested):
Please, start a discussion. This is not minor and this setting has been key to balance timely GC and CPU usage.

ItalyPaleAle:
> This is not minor and this setting has been key to balance timely GC and CPU usage.

Yes, that's why it's been removed :) Before there was a goroutine that on an interval was scanning all actors, thus with the concerns around GC and CPU. Now, we block the goroutine until we know an actor becomes idle.

You can see the single commit here: https://github.com/dapr/dapr/pull/6715/commits/c75700f48c63ca4f056979f1a2cf0909cce813d4

yaron2:
We can't remove this for this release as it's a breaking chang. We can mark this as deprecated in 1.12 (if there's consensus for it) and then remove in 1.14.

ItalyPaleAle:
> We can't remove this for this release as it's a breaking change. We can mark this as deprecated in 1.12 (if there's consensus for it) and then remove in 1.14.

There's no breaking change and it's a pretty important perf improvement (see #6592)

yaron2:
> > We can't remove this for this release as it's a breaking change. We can mark this as deprecated in 1.12 (if there's consensus for it) and then remove in 1.14.
> 
> There's no breaking change and it's a pretty important perf improvement (see #6592)

It's documented as part of the API and contract between Dapr and the app. Removing this will result in a breaking behavioral change.

yaron2:
> > > We can't remove this for this release as it's a breaking change. We can mark this as deprecated in 1.12 (if there's consensus for it) and then remove in 1.14.
> > 
> > 
> > There's no breaking change and it's a pretty important perf improvement (see #6592)
> 
> It's documented as part of the API and contract between Dapr and the app. Removing this will result in a breaking behavioral change.

cc @daixiang0 @mukundansundar

ItalyPaleAle:
> > > We can't remove this for this release as it's a breaking change. We can mark this as deprecated in 1.12 (if there's consensus for it) and then remove in 1.14.
> > 
> > 
> > There's no breaking change and it's a pretty important perf improvement (see #6592)
> 
> It's documented as part of the API and contract between Dapr and the app. Removing this will result in a breaking behavioral change.

It doesn't alter the behavior that actors are deactivated after they become idle for a certain amount of time. If the scan interval option is passed, it is ignored.

This just makes it more efficient and more predictable. Before, if scan interval was 5s, an idle actor could be deactivated right away or after 5s. Now, idle actors are deactivated "as soon as" their timeout expires.

There's no user-facing behavior change. Per the contract, users should have assumed that an actor could be deactivated any time after its idle timeout passed. This doesn't change that at all.

yaron2:
> 

Ok, so to be clear:

If a user passes an ActorScanInterval value of 1m, can they expect an actor to become deactivated after 1m?

ItalyPaleAle:
> > 
> 
> Ok, so to be clear:
> 
> If a user passes an ActorScanInterval value of 1m, can they expect an actor to become deactivated after 1m?

If a user passes "ActorIdleTimeout" of 1m, they can expect an actor to be deactivated after 1m of inactivity. This was the case before and is the case after

What changed is:

- Before: if users passed "ActorScanInterval" of 1m, the (idle) actor could be deactivated after 1 min of becoming idle in the "best case" scenario, and 2 mins in the "worst case" scenario. This was not guaranteed and it depended on where in the scan loop the actor runtime was. Users should have assumed the actor's lifetime was not to extend 1m after becoming idle.
- After: there's no "ActorScanInterval", so an actor that becomes idle is deactivated "right after". This means that an idle actor gets deactivated after 1 min of becoming idle in all cases.

https://docs.dapr.io/developing-applications/building-blocks/actors/actors-runtime-config/

yaron2:
> > > 
> > 
> > 
> > Ok, so to be clear:
> > If a user passes an ActorScanInterval value of 1m, can they expect an actor to become deactivated after 1m?
> 
> If a user passes "ActorIdleTimeout" of 1m, they can expect an actor to be deactivated after 1m of inactivity. This was the case before and is the case after
> 
> What changed is:
> 
> * Before: if users passed "ActorScanInterval" of 1m, the (idle) actor could be deactivated after 1 min of becoming idle in the "best case" scenario, and 2 mins in the "worst case" scenario. This was not guaranteed and it depended on where in the scan loop the actor runtime was. Users should have assumed the actor's lifetime was not to extend 1m after becoming idle.
> * After: there's no "ActorScanInterval", so an actor that becomes idle is deactivated "right after". This means that an idle actor gets deactivated after 1 min of becoming idle in all cases.
> 
> https://docs.dapr.io/developing-applications/building-blocks/actors/actors-runtime-config/

I'm not debating the benefits or correctness of this change, simply this is a clear breaking change and should be treated as such. Will wait for feedback of other maintainers on this.

ItalyPaleAle:
> I'm not debating the benefits or correctness of this change, simply this is a clear breaking change and should be treated as such. Will wait for feedback of other maintainers on this.

I am not sure I follow. Why do you consider this a breaking change?

artursouza:
Can this be enabled via a configuration. If ActorScanInterval <= 0 then it is the new behavior, otherwise it is the old behavior.

ItalyPaleAle:
I didn't think this would be controversial as it shouldn't be a breaking change at all.

The current behavior is that there's an optimistic case, which is that actors are deactivated as soon as the idle timeout passes, and a pessimistic case, which is that actors stay alive for longer, as long as the scan interval. There's obviously a trade-off between the two things: keeping actors alive for longer uses more resources, while a shorter scan interval means more frequent scans which use more resources too. I believe @artursouza when he says that the scan interval option was added to find the right balance with GC and CPU usage.

The point of this change is to remove the make the optimistic case the only case. It's quite literally the "have your cake and eat it too": you get actors that are deactivated right away after the idle timeout (so not wasting more resources and acting very predictably), and you don't pay the cost of the periodic scan interval.

Although it's true this PR does remove a configuration option, it does that by retaining the same functionality, so users are not missing out on anything. I would be surprised to hear from anyone complaining about this.

> The only situation where users would be impacted here is if they were counting on actor scan interval as a way to "extend" the lifetime of an actor beyond its idle timeout. I am not concerned about these users much because they shouldn't have done this to begin with: in some cases the actors were alive for an extra "scan interval" time, but in others they'd be terminated right away, and users should not have taken a dependency on this random behavior.

yaron2:
> I would be surprised to hear from anyone complaining about this

Personally knowing how many users are using it, I wouldn't be.

> Although it's true this PR does remove a configuration option, it does that by retaining the same functionality

Yes, but that same functionality is retained by utilizing a different configuration option. If you had said that users who previously set `ActorScanInterval` to `7h` and expected that actors get deactivated after 7 hours (at minimum) or later would expect the exact same behavior when moving to 1.12 without code changes, then I would have agreed this should ot be considered a breaking change. To my understanding, this is not what's being said.

yaron2:
> > Yes, but that same functionality is retained by utilizing a different configuration option. If you had said that users who previously set `ActorScanInterval` to `7h` and expected that actors get deactivated after 7 hours (at minimum) or later would expect the exact same behavior when moving to 1.12 without code changes, then I would have agreed this should not be considered a breaking change. To my understanding, this is not what's being said.
> 
> Hold on, I think I understand where the misunderstanding is. Are you maybe thinking of `ActorIdleTimeout`?
> 
> * `ActorIdleTimeout` is what determines how long after the last message the actor is considered idle.
> * `ActorScanInterval` indicates how often the loop runs to check which actors are idle and are past `ActorIdleTimeout`.
> 
> If you want actors to have a 7h lifetime minimum, you set `ActorIdleTimeout` to 7h. This will not change in this PR.

I will make this clearer - If users set `ActorScanInterval` to any value in 1.11.x and expected the scan to elapse at the configured interval to check which actors are idle after `ActorIdleTimeout`, and after having upgraded to 1.12.x have not changed anything in their code or configuration and are observing a different behavior, then this is a breaking change.

ItalyPaleAle:
> if users set ActorScanInterval to any value in 1.11.x and expected the scan to elapse at the configured interval to check which actors are idle after ActorIdleTimeout

Users must not have this expectation. Actors in 1.11 can be terminated any time after ActorIdleTimeout lapses. In the "optimistic case", the loop happens immediately after so actors are terminated right away.

For users to rely on the pessimistic case to determine the lifecycle of an actor is a bug even in 1.11 because it is not deterministic.

yaron2:
> > if users set ActorScanInterval to any value in 1.11.x and expected the scan to elapse at the configured interval to check which actors are idle after ActorIdleTimeout
> 
> Users must not have this expectation. Actors in 1.11 can be terminated any time after ActorIdleTimeout lapses. In the "optimistic case", the loop happens immediately after so actors are terminated right away.
> 
> For users to rely on the pessimistic case to determine the lifecycle of an actor is a bug even in 1.11 because it is not deterministic.

Bug or not, expectation or not - are all subjective. What's not subjective is that we are removing an API field that will no longer do what it was supposed to do, and that is objectively a breaking change that we must, and will, treat as one.

yaron2:
I think the suggestion made by @artursouza to retain backwards-compat is solid and should allow this PR to be merged once implemented.

yaron2:
> PS: please don't "merge master" into draft PRs. When the underlying PR is merged, I'll have to cherry-pick the second commit and having master merged in makes that harder! 🙏

👍

yaron2:
> The current behavior is that a 0 value means "accept the default", which is 30s. So that could be a breaking change too per your definition.

If a user isn't required to perform any changes to their code/config from 1.11 to 1.12 and will observe the exact same behavior as they did before, then this isn't a breaking change.

ItalyPaleAle:
This PR is now ready for review and it's been updated to be even more efficient thanks to using a priority queue, which allows avoiding all scans of the active actors table. (The Queue and Processor structs used by the priority queue are the same used for #6716).

I have expanded the description and also included an explanation on why this does not include a breaking change for users, and at the same time offers very welcome improvements on perf and ease of use.

I ask maintainers to please give this a reconsideration.

yaron2:
> I have expanded the description and also included an explanation on why this does not include a breaking change for users, and at the same time offers very welcome improvements on perf and ease of use.

Thanks for the update. The `ActorScanInterval` field exists as a public contract in all Dapr SDKs. If we remove it from the runtime, we must also remove it from the SDKs. In that case, applications that update their dependencies to a new Dapr SDK version as part of upgrading to Dapr 1.12 will break at compilation time, forcing users to change their code to adapt. Do you agree this is a breaking change for the SDKs, and in general how do you suggest we tackle that issue if we remove `ActorScanInterval`?

ItalyPaleAle:
> > I have expanded the description and also included an explanation on why this does not include a breaking change for users, and at the same time offers very welcome improvements on perf and ease of use.
> 
> Thanks for the update. The `ActorScanInterval` field exists as a public contract in all Dapr SDKs. If we remove it from the runtime, we must also remove it from the SDKs. In that case, applications that update their dependencies to a new Dapr SDK version as part of upgrading to Dapr 1.12 will break at compilation time, forcing users to change their code to adapt. Do you agree this is a breaking change for the SDKs, and in general how do you suggest we tackle that issue if we remove `ActorScanInterval`?

The field can continue to be provided it is just ignored. SDKs can manage the deprecation nicely, and make the property a no-op to avoid breaking users. So outdated SDKs will send the field which is ignored by the runtime. Updated SDKs can continue to have the method which either sends the field (for compat with old Dapr versions), or becomes a no-op.

yaron2:
> > > I have expanded the description and also included an explanation on why this does not include a breaking change for users, and at the same time offers very welcome improvements on perf and ease of use.
> > 
> > 
> > Thanks for the update. The `ActorScanInterval` field exists as a public contract in all Dapr SDKs. If we remove it from the runtime, we must also remove it from the SDKs. In that case, applications that update their dependencies to a new Dapr SDK version as part of upgrading to Dapr 1.12 will break at compilation time, forcing users to change their code to adapt. Do you agree this is a breaking change for the SDKs, and in general how do you suggest we tackle that issue if we remove `ActorScanInterval`?
> 
> The field can continue to be provided it is just ignored. SDKs can manage the deprecation nicely, and make the property a no-op to avoid breaking users. So outdated SDKs will send the field which is ignored by the runtime. Updated SDKs can continue to have the method which either sends the field (for compat with old Dapr versions), or becomes a no-op.

As an SDK maintainer I feel uncomfortable providing a field that has no use, because this not only affects existing users but more so it means developers who are getting started with the SDK have an option that is seemingly provided for them but in fact does nothing. It creates a drift between the API spec of the runtime and the different SDKs. So far changes /removals made to the API spec were followed by SDK changes in the same release, the Configuration API is a recent example. At the very least we need to communicate to SDK maintainers that a runtime flag present as a typed entity in their SDK is going to be removed.

ItalyPaleAle:
Alright, if that’s the consensus among maintainers we’ll merge this PR in our distribution. Actor perf is important to us and we want to provide our users the best experience.

Thanks for the feedback.

yaron2:
> Alright, if that’s the consensus among maintainers we’ll merge this PR in our distribution. Actor perf is important to us and we want to provide our users the best experience.
> 
> Thanks for the feedback.

I don't know if that's the consensus among maintainers.. we're still waiting for feedback here. There's no reason to close this PR.

yaron2:
@mukundansundar @daixiang0 your input on this matter is required as we want to arrive at a resolution of this soon.

PR #6716, opened 2023-07-23
Manage actor timers using a priority queue in memory

ItalyPaleAle:
Part of: #6592
Based on original work on: #6040
Related issue: #6503

Currently, the Dapr sidecar maintains one goroutine for each active timer (and reminder), each one blocked until it's time for the timer to be executed. This adds significant pressure on memory as well as additional work on the GC.

This PR implements a new data structure that is essentially a priority queue (heap-based). When timers (and in the future reminders) are scheduled, they are added to the queue in order of their execution time. By maintaining the queue with timers/reminders in order, we can make sure that there's at most one background goroutine, which is waiting on the next timer/reminder only.

This PR implements the updated code on timers only. Reminders are of a different level of complexity and will be addressed in a future PR (based on the data structures introduced in this).


pkg/operator/watchdog_test.go
```diff
@@ -12,13 +12,13 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/utils/pointer"
```
ItalyPaleAle, line 15: Linter was complaining this package was deprecated. The correct replacement seems to be dapr/kit/ptr

go.mod
```diff
@@ -10,7 +10,7 @@ require (
 	github.com/cenkalti/backoff/v4 v4.2.1
 	github.com/cloudevents/sdk-go/v2 v2.14.0
 	github.com/dapr/components-contrib v1.11.3-0.20230725200615-9ca07d9bca9d
-	github.com/dapr/kit v0.11.3
+	github.com/dapr/kit v0.11.4-0.20230802200247-90b1d0d906e0
```
ItalyPaleAle, line 13: See dapr/kit#54

artursouza (approved):

PR #6720, opened 2023-07-24
Daprd: graceful shutdown

JoshVanL:
closes https://github.com/dapr/dapr/issues/2478

~Branched from https://github.com/dapr/dapr/pull/6688 which should be merged first.~

PR updates daprd so that it is gracefully shutdown. In practice this means that once daprd is initialized, if daprd receives an interrupt signal or an error occurred during initialization then the runtime will shutdown all servers and components and waits for all go routines to exit. A graceful shutdown timeout is still in effect. All long running processes are closed concurrently.

Updates the metrics exporter server to gracefully shutdown.

Adds waitgroups to `actors`, `http`, `grpc` which are waited on on `Close`.


cmd/injector/main.go
```diff
@@ -52,12 +52,6 @@ func main() {
 		log.Fatalf("Error set env: %v", err)
 	}
 
-	// Initialize dapr metrics exporter
-	err = metricsExporter.Init()
-	if err != nil {
-		log.Fatal(err)
-	}
-
```
JoshVanL, line 60: `metricsExporter.Init` renamed to `metricsExporter.Run`, and is a now a long running process which closes on context cancellation.

pkg/runtime/meta/meta.go
```diff
@@ -48,14 +47,19 @@ func New(options Options) *Meta {
 	}
 }
 
-func (m *Meta) ToBaseMetadata(comp compapi.Component) metadata.Base {
+func (m *Meta) ToBaseMetadata(comp compapi.Component) (metadata.Base, error) {
+	props, err := m.convertItemsToProps(comp.Spec.Metadata)
+	if err != nil {
+		return metadata.Base{}, err
```
JoshVanL, line 53: Avoid `Fatal`ing here, in favor of bubbling up the error.

pkg/runtime/runtime.go
```diff
@@ -460,9 +584,7 @@ func (a *DaprRuntime) initRuntime(ctx context.Context) error {
 		log.Infof("app max concurrency set to %v", a.runtimeConfig.appConnectionConfig.MaxConcurrency)
 	}
 
-	a.appHealthReady = func() {
-		a.appHealthReadyInit(ctx)
-	}
+	a.appHealthReady = a.appHealthReadyInit
```
ItalyPaleAle, line 587: This was a lambda on purpose, otherwise `a` isn't updated. You can add the context but it needs to remain a lambda

pkg/runtime/runtime.go
```diff
@@ -1331,138 +1467,58 @@ func (a *DaprRuntime) loadHTTPEndpoints() error {
...
+			return nil
+		}
 	}
 
 	return nil
 }
 
-func (a *DaprRuntime) stopActor() {
+func (a *DaprRuntime) stopActor(_ context.Context) error {
+	log.Info("Initiating actor shutdown")
```
ItalyPaleAle, line 1479: Redundant log?

pkg/runtime/runtime.go
```diff
@@ -555,7 +679,9 @@ func (a *DaprRuntime) appHealthChanged(ctx context.Context, status uint8) {
 	case apphealth.AppStatusHealthy:
 		// First time the app becomes healthy, complete the init process
 		if a.appHealthReady != nil {
-			a.appHealthReady()
+			if err := a.appHealthReady(ctx); err != nil {
+				log.Warnf("failed to complete app init: %s ", err)
```
ItalyPaleAle, line 683:
```suggestion
				log.Warnf("Failed to complete app init: %v ", err)
```

pkg/runtime/runtime.go
```diff
@@ -511,9 +633,11 @@ func (a *DaprRuntime) appHealthReadyInit(ctx context.Context) {
 			DirectMessaging: a.directMessaging,
 			CompStore:       a.compStore,
 		}); err != nil {
-			log.Fatalf("failed to register components with callback: %s", err)
+			return fmt.Errorf("failed to register components with callback: %s", err)
```
ItalyPaleAle, line 636:
```suggestion
			return fmt.Errorf("failed to register components with callback: %w", err)
```

pkg/runtime/runtime.go
```diff
@@ -1261,25 +1400,22 @@ func (a *DaprRuntime) processComponentAndDependents(ctx context.Context, comp co
...
-	select {
-	case err := <-ch:
-		if err != nil {
-			return err
-		}
-	case <-time.After(timeout):
+	err = a.processor.Init(ctx, comp)
+	// If the context is canceled, we want  to return an init error.
+	if errors.Is(ctx.Err(), context.DeadlineExceeded) {
+		err = fmt.Errorf("init timeout for component %s exceeded after %s", comp.Name, timeout.String())
```
ItalyPaleAle, line 1414:
```suggestion
		err = fmt.Errorf("init timeout for component %s exceeded after %s", comp.LogName(), timeout.String())
```

pkg/runtime/runtime.go
```diff
@@ -1331,138 +1467,58 @@ func (a *DaprRuntime) loadHTTPEndpoints() error {
...
-				log.Fatal("Graceful shutdown timeout exceeded")
-			case <-ctx.Done():
-				return
+func (a *DaprRuntime) cleanSockets() error {
+	var errs []error
+	if a.runtimeConfig.unixDomainSocket != "" {
+		for _, s := range []string{"http", "grpc"} {
+			err := os.Remove(fmt.Sprintf("%s/dapr-%s-%s.socket", a.runtimeConfig.unixDomainSocket, a.runtimeConfig.id, s))
+			if err != nil {
+				errs = append(errs, fmt.Errorf("error removing socket file: %s", err))
```
ItalyPaleAle, line 1517:
```suggestion
				errs = append(errs, fmt.Errorf("error removing socket file: %w", err))
```

pkg/runtime/runtime.go
```diff
@@ -1331,138 +1467,58 @@ func (a *DaprRuntime) loadHTTPEndpoints() error {
...
-		go func() {
-			select {
-			case <-time.After(duration):
-				log.Fatal("Graceful shutdown timeout exceeded")
-			case <-ctx.Done():
-				return
+func (a *DaprRuntime) cleanSockets() error {
+	var errs []error
+	if a.runtimeConfig.unixDomainSocket != "" {
+		for _, s := range []string{"http", "grpc"} {
```
ItalyPaleAle, line 1514: Should remove sockets for pluggable components too? Or not?

pkg/runtime/runtime.go
```diff
@@ -1592,28 +1648,22 @@ func (a *DaprRuntime) authSecretStoreOrDefault(resource meta.Resource) string {
...
-			// Cause a shutdown
-			a.ShutdownWithWait()
-			return
 		case <-ctx.Done():
 			// Return
-			return
-		default:
-			// nop - continue execution
+			return ctx.Err()
+			// prevents overwhelming the OS with open connections
```
ItalyPaleAle, line 1665: Nit: comment’s indentation

pkg/runtime/runtime.go
```diff
@@ -1768,25 +1836,19 @@ func createGRPCManager(runtimeConfig *internalConfig, globalConfig *config.Confi
...
+		return nil
 	}
 	// Flush and shutdown the tracing provider.
 	if err := a.tracerProvider.ForceFlush(ctx); err != nil && !errors.Is(err, context.Canceled) {
 		log.Warnf("error flushing tracing provider: %v", err)
 	}
 
 	if err := a.tracerProvider.Shutdown(ctx); err != nil && !errors.Is(err, context.Canceled) {
-		log.Warnf("error shutting down tracing provider: %v", err)
+		return fmt.Errorf("error shutting down tracing provider: %v", err)
```
ItalyPaleAle, line 1849:
```suggestion
		return fmt.Errorf("error shutting down tracing provider: %w", err)
```

pkg/runtime/runtime.go
```diff
@@ -438,14 +560,16 @@ func (a *DaprRuntime) initRuntime(ctx context.Context) error {
 
 	err = a.startGRPCInternalServer(a.daprGRPCAPI, a.runtimeConfig.internalGRPCPort)
 	if err != nil {
-		log.Fatalf("failed to start internal gRPC server: %s", err)
+		return fmt.Errorf("failed to start internal gRPC server: %s", err)
```
ItalyPaleAle, line 563:
```suggestion
		return fmt.Errorf("failed to start internal gRPC server: %w", err)
```

pkg/runtime/runtime.go
```diff
@@ -427,7 +549,7 @@ func (a *DaprRuntime) initRuntime(ctx context.Context) error {
 	// Start HTTP Server
 	err = a.startHTTPServer(a.runtimeConfig.httpPort, a.runtimeConfig.publicPort, a.runtimeConfig.profilePort, a.runtimeConfig.allowedOrigins, pipeline)
 	if err != nil {
-		log.Fatalf("failed to start HTTP server: %s", err)
+		return fmt.Errorf("failed to start HTTP server: %s", err)
```
ItalyPaleAle, line 552:
```suggestion
		return fmt.Errorf("failed to start HTTP server: %w", err)
```

pkg/runtime/runtime.go
```diff
@@ -414,7 +536,7 @@ func (a *DaprRuntime) initRuntime(ctx context.Context) error {
 
 	err = a.startGRPCAPIServer(a.daprGRPCAPI, a.runtimeConfig.apiGRPCPort)
 	if err != nil {
-		log.Fatalf("failed to start API gRPC server: %s", err)
+		return fmt.Errorf("failed to start API gRPC server: %s", err)
```
ItalyPaleAle, line 539:
```suggestion
		return fmt.Errorf("failed to start API gRPC server: %w", err)
```

ItalyPaleAle: All of these changes also allow detecting if the error is context.Canceled for example

pkg/runtime/runtime.go
```diff
@@ -228,25 +234,144 @@ func newDaprRuntime(ctx context.Context,
...
+		},
+		func(ctx context.Context) error {
+			log.Info("Stopping Dapr APIs")
+			errs := make([]error, len(a.apiClosers))
+			var wg sync.WaitGroup
+			wg.Add(len(a.apiClosers))
+			for i, closer := range a.apiClosers {
+				go func(i int, c io.Closer) {
+					defer wg.Done()
+					if err := closer.Close(); err != nil {
```
ItalyPaleAle, line 337:
```suggestion
					if err := c.Close(); err != nil {
```

pkg/runtime/runtime.go
```diff
@@ -228,25 +234,144 @@ func newDaprRuntime(ctx context.Context,
...
+						errs[i] = fmt.Errorf("error closing API: %v", err)
```
ItalyPaleAle, line 338:
```suggestion
						errs[i] = fmt.Errorf("error closing API: %w", err)
```

ItalyPaleAle (commented):
First pass

pkg/runtime/runtime.go
JoshVanL: Not sure I follow- this is still a lambda, it just now has the same signature so doesn't need to be wrapped.

pkg/runtime/runtime.go
JoshVanL: This is the behavior today and don't think we should do that in this PR.

pkg/runtime/runtime.go
ItalyPaleAle: I am not 100% sure of why that was even happening, but the original code had something like `a.appHealthReady = a.appHealthReadyInit` and it "captured" `a`, so if a property of `a` changed, it was not reflected in the method.

Issue #5230, opened 2022-09-22
Correlate actor reminder creation and execution events


fabistb:
## In what area(s)?

 /area runtime

> /area operator

> /area placement

> /area docs

> /area test-and-release

## Describe the feature

If an actor creates a reminder which is later executed these operations are currently treated as individual events on the observability side. 
In application insights this is displayed by different operation ids for the reminder creation and the reminder execution.

Correlating these would improve the end to end visibility. 

## Release Note

RELEASE NOTE:
**ADD** Correlate actor reminder creation and execution

fabistb:
Cc: @artursouza

fabistb:
keep alive

fabistb:
Hi @DeepanshuA , yes I'm referring to the Zipkin or application insights trace correlation.
I face a similar issue you mentioned that the reminder triggered wasn't shown in Zipkin.
In my test I just added some state operation which is shown in Zipkin and works as a substitute.

Overview:
![image](https://github.com/dapr/dapr/assets/63794579/60699682-4110-4451-a2b0-1b80b096b040)

Initial actor operation:
![image](https://github.com/dapr/dapr/assets/63794579/4f1316a8-8581-463d-9e74-9d5ff4a87444)

Reminder triggered:
![image](https://github.com/dapr/dapr/assets/63794579/9d9a8e5f-c239-4db9-b02c-7ba9e1f3da6c)

The _reminder triggered_ operation isn't correlated with the initial _reminder create_ operation.

PR #6724, opened 2023-07-25
Correlate Reminders Creation and Execution

DeepanshuA:
# Description

Correlate Actor Reminders creation and execution events.
With this change, reminder track would be stored at the time of reminder creation.
It is done so that traceState of creation time can be saved - it will be used so that this traceState can be linked to at the time of execution.
At the creation time, Last Fired Time will be time.Time{} i.e. `0001-01-01 00:00:00 +0000 UTC`.
Last Fired Time will be updated when reminder is actually fired.

## Issue reference

Please reference the issue this PR will close: #5230 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Unit tests passing


pkg/actors/reminders/reminders.go
```diff
@@ -149,7 +156,12 @@ func (r *reminders) CreateReminder(ctx context.Context, reminder *internal.Remin
 	if err != nil {
 		return fmt.Errorf("error storing reminder: %w", err)
 	}
-	return r.startReminder(reminder, stop)
+	err = r.startReminder(reminder, stop)
+	if err != nil {
+		return err
+	}
+	w3cString := diag.SpanContextToW3CString(span.SpanContext())
+	return r.updateReminderTrack(ctx, reminder.Key(), reminder.RepeatsLeft(), time.Time{}, nil, w3cString)
```
mukundansundar, line 164: is there a scenario possible where the reminder is fired immediately and the track is not updated with the trace state info?

pkg/actors/reminders/reminders.go
```diff
@@ -899,6 +920,13 @@ func (r *reminders) startReminder(reminder *internal.Reminder, stopChannel chan
 			}
 
 			nextTimer.Reset(nextTick.Sub(r.clock.Now()))
+			if err != nil {
+				diag.UpdateSpanStatusFromHTTPStatus(span, http.StatusInternalServerError)
```
mukundansundar, line 924: span nil check is missing

mukundansundar (commented):
please update/add unit tests.

pkg/actors/reminders/reminders.go
DeepanshuA: Changed to update reminder track before starting reminders.

pkg/actors/reminders/reminders.go
DeepanshuA: Done

pkg/actors/reminders/reminders.go
```diff
@@ -149,6 +156,11 @@ func (r *reminders) CreateReminder(ctx context.Context, reminder *internal.Remin
 	if err != nil {
 		return fmt.Errorf("error storing reminder: %w", err)
 	}
+	w3cString := diag.SpanContextToW3CString(span.SpanContext())
```
ItalyPaleAle, line 159: Invoking `updateReminderTrack` performs more I/O operations with storage. It will make this a lot slower (than it already is).

Could the tracing stuff be stored with the reminder, rather than with the track, to avoid an additional I/O?

Otherwise, could we update the reminder track only if tracing is enabled for reminders?

pkg/actors/reminders/reminders.go
```diff
@@ -854,6 +866,15 @@ func (r *reminders) startReminder(reminder *internal.Reminder, stopChannel chan
 				nextTimer = nil
 				return
 			}
+			track, gErr := r.getReminderTrack(context.TODO(), reminderKey)
```
ItalyPaleAle, line 869: Why getting the track again? Don't we have it already at the start of the method?

ItalyPaleAle (commented):
Concerned about the perf impact of this, which seems to be significant.

ItalyPaleAle (commented):
Concerned about the perf impact of this, which seems to be significant.

pkg/actors/reminders/reminders.go
ItalyPaleAle: @DeepanshuA here's an idea, which is similar to how we store the execution time:

- The initial span is stored with the reminder itself
- The track _can_ store a span, which overrides the one from the reminder data if present

This way we avoid adding more storage operations

mukundansundar:
@DeepanshuA Do we know the perf impact of this change?

pkg/actors/reminders/reminders.go
DeepanshuA: Yeah, correct. Deleted it.

pkg/actors/reminders/reminders.go
ItalyPaleAle: > What I understand right now is that more than number of operations, bigger concern is the data of reminder(s) itself getting stored in key "actors||<actor_type>". If I store traceState in reminder as well, my concern is that reminder data will be more corrupted.

That’s true, but the tracing data is relatively small.

The number of operations is a concern too however, as in many databases it requires an extra round-trip to the DB.

cicoyle:
@DeepanshuA What was the status of this PR, were you planning to finish and get this merged?

cicoyle:
@DeepanshuA mentioned not taking this PR forward. Going to close it out again

PR #6726, opened 2023-07-25
tests/integration/suite/daprd/pubsub/grpc/compname: revert numTests `1` -> `1000`

JoshVanL:
The number of pubsub grpc component loader fuzz test cases was mistakenly reduced from `1000` to `1` in https://github.com/dapr/dapr/pull/6688. This reverts that change.

/cc @ItalyPaleAle


ItalyPaleAle (commented):
Thanks! Will approve when JS SDK tests pass

ItalyPaleAle (approved):
Approved since #6731 fixes the JS SDK issue

Issue #6725, opened 2023-07-25
[Docs] Complete schema for components CRD 


philliphoff:
## In what area(s)?

> /area runtime

> /area operator

> /area placement

/area docs

> /area test-and-release

## Describe the feature
Tooling intends to use the schema in the [components CRD](https://raw.githubusercontent.com/dapr/dapr/master/charts/dapr/crds/components.yaml) to help developers writing Dapr component definition YAML. Several properties of the schema are missing `description` and related properties, which limits the schemas usefulness. We should review the schema and add (at a minimum) a `description` property but also any other helpful information (such as type, default, etc.).

## Release Note

RELEASE NOTE:

PR #6728, opened 2023-07-25
[Draft] Description changes for components CRD.

philliphoff:
# Description

Updates the components CRD with additional/updated schema property descriptions.

## Issue reference

Please reference the issue this PR will close: #6725 

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing
* [x] Specification has been updated / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#3638

### DIFF
--- a/charts/dapr/crds/components.yaml
+++ b/charts/dapr/crds/components.yaml
@@ -15,56 +15,65 @@ spec:
         description: Component describes an Dapr component type.
         properties:
           apiVersion:
-            description: 'APIVersion defines the versioned schema of this representation
-              of an object. Servers should convert recognized schemas to the latest
-              internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources'
+            description: The version of the Dapr (and Kubernetes if applicable) API you are calling.
             type: string
           auth:
             description: Auth represents authentication details for the component.
             properties:
               secretStore:
+                description: Name of a configured secret store.
                 type: string
             required:
             - secretStore
             type: object
           kind:
-            description: 'Kind is a string value representing the REST resource this
-              object represents. Servers may infer this from the endpoint the client
-              submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds'
+            description: The type of CRD. For components is must always be Component.
             type: string
           metadata:
+            description: Information about the component registration.
+            properties:
+              name:
+                description: The name of the component.
+                type: string
+              namespace:
+                description: The namespace for the component for hosting environments with namespaces.
+                type: string
+            required:
+            - name
             type: object
           scopes:
+            description: Specific application IDs which can use this component.
             items:
               type: string
             type: array
           spec:
-            description: ComponentSpec is the spec for a component.
+            description: Detailed information on the component resource.
             properties:
               ignoreErrors:
+                description: Tells the Dapr sidecar to continue initialization if the component fails to load. Default is false.
                 type: boolean
               initTimeout:
+                description: The timeout duration for the initialization of the component. Default is 5s.
                 type: string
               metadata:
+                description: Key/value pairs of component specific configuration.
                 items:
-                  description: NameValuePair is a name/value pair.
+                  description: Key/value pair of component specific configuration.
                   properties:
                     envRef:
-                      description: EnvRef is the name of an environmental variable
-                        to read the value from.
+                      description: Reference to an environmental variable from which to read the value.
                       type: string
                     name:
                       description: Name of the property.
                       type: string
                     secretKeyRef:
-                      description: SecretKeyRef is the reference of a value in a secret
-                        store component.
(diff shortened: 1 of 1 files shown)


charts/dapr/crds/components.yaml
```diff
@@ -15,56 +15,65 @@ spec:
         description: Component describes an Dapr component type.
         properties:
           apiVersion:
-            description: 'APIVersion defines the versioned schema of this representation
-              of an object. Servers should convert recognized schemas to the latest
-              internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources'
+            description: The version of the Dapr (and Kubernetes if applicable) API you are calling.
```
ItalyPaleAle, line 18: The CRDs are auto-generated from what is in `pkg/api` (then you run “make code-generate” and copy from the “staging” files into the charts folder). That’s why they all begin with “APIVersion” etc, because of the Go convention

philliphoff: Ah, ok, I see.  Playing with this a bit, I see a couple issues:

- While I suppose *some* description is better than none, the Go convention for docs meant for developers *of* Dapr doesn't translate well into descriptions exposed for developers *usingt* Dapr (e.g. `ComponentSpec is the spec for a component.`).
- Some of the schema properties are "inherited" from k8s common types, so we don't control their descriptions and they are somehow both lengthy *and* vague at the same time (e.g. `Kind is a string value representing the REST resource this object represents. Servers may infer this from the endpoint the client submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds'`).

If it were just the first issue, I might argue we bend the Go docs conventions to allow generation of more user-friendly descriptions.  It would be nice if `controller-gen` supported a marker that overrrode the description, but looking at the  source, I'm not sure I can justify the time for my purposes and, given the second issue, it wouldn't completely solve the problem (unless there was a way to override the validation metadata of inherited properties).  Updating descriptions of the generated schema as a second automated step is a possibility; I'd have to look to see what that'd take and whether the time would be justified vs. doing it by hand (e.g. in a copy of the CRD) given the (un)likelyhood of the schema dramatically changing over time.

ItalyPaleAle: For the first issue, I think we can maybe ignore the Go convention here? We'd need to add the proper `//nolint` tags or the linter would complain too much.

For the "inherited" properties, I don't think there's much we can do. I also don't know if we need to worry about those? Not sure what your use case is.

> Updating descriptions of the generated schema as a second automated step is a possibility; I'd have to look to see what that'd take and whether the time would be justified vs. doing it by hand (e.g. in a copy of the CRD) given the (un)likelyhood of the schema dramatically changing over time.

As for this,please make sure it's automated. We cannot rely on OSS contributors not removing that in the future.

yaron2, line 18: For Dapr this doesn't apply only to K8s and we should remove K8s specific verbiage.

philliphoff: > Not sure what your use case is.

The use case is using this schema information to help developers writing component definition YAML in an editor (e.g. provide descriptions of properties on hover, suggest properties, warn when properties are omitted or invalid, etc.).

> As for this,please make sure it's automated. We cannot rely on OSS contributors not removing that in the future.

If an overlay were done manually, it'd be done separate from this file (e.g. and shipped with the tooling).  If it made sense to automate, the question is whether it'd make sense to do it in this repo or as part of the separate tooling. It'd probably easier to do that as part of the tooling or its build process, but it's theoretically more useful here, being available to all potential users.

philliphoff: The intent would be to run any new descriptions past our docs experts as we'd want them to align with what's already in the docs (which they do not, currently, due to the difference in audience between Dapr developers and Dapr users).

msfussell: This can be referenced from here https://docs.dapr.io/reference/resource-specs/component-schema/ which is where this is defined in the docs (certainly may not be complete). Given that Dapr is platform agnostic, we stay away from the term CRD in the docs ( there are places this needs to be removed) and use the term "resource spec" or just "resource". We have to be attentive to ensuring that nothing ties Dapr to Kubernetes as the only platform.

ItalyPaleAle: We have a `.build-tools` folder. You could use that to write some additional tooling. Which would not be bad: our tooling right now puts the generated CRDs in the wrong place too.

philliphoff: Yeah, but then I'd have to write Go code or, worse, shell script. :-)

berndverst:
Can this PR be closed @philliphoff ? I do not think this is the right place to accomplish what you wish to accomplish. I agree using `.build-tools` in components contrib can be of use to you. There are tools in Go and Bash that can be expanded to get you what you need.

philliphoff:
Closing as I've found an alternate approach.

Issue #6723, opened 2023-07-25
Reduce cardinality in Dapr metrics and add more information to API logs


ItalyPaleAle:
## In what area(s)?

/area runtime

## What version of Dapr?

1.1.x
master

## Description

Related: #6581

The Dapr HTTP server collects metrics with very high cardinality by default. Each path invoked in the HTTP server is used to create a "bucket" in the metrics, and this causes significant memory usage when using Dapr to invoke RESTful APIs (where the path is usually different on every request).

This happens in service invocation metrics as well as in the general HTTP server metrics.

Users can manually configure Dapr to [reduce the cardinality of metrics,](https://docs.dapr.io/operations/monitoring/metrics/metrics-overview/#high-cardinality-metrics) but that is a manual process that is not immediately clear, and it requires writing regular expressions.

@yaron2 and I have been discussing how to address this issue and our proposal is:

1. By default, do not report metrics per each endpoint/path in the HTTP server. Aggregation will stop at the app ID for service invocation.
2. Add an option to the Configuration CRD `metrics.highCardinality`, disabled by default, to retain the current behavior.
3. Change the API logging middleware to include additional information on each request, including latency and returned status code. These are things that were before included in the metrics. Although API logging aren't a 1:1 replacement for metrics, users can leverage them in tools like Grafana to gather similar insights.

## Release Note

RELEASE NOTE: **CHANGED** Reduced default cardinality of metrics collected by HTTP server

Hsuwen:
About this issue, I found that the docs has been updated.
[https://docs.dapr.io/operations/monitoring/metrics/metrics-overview/#high-cardinality-metrics](url)

Several details are still not very clear:
1. What specific metrics will be affected by this will determine which metrics I should set
2. How to filter by appid in the configuration of ```metric.rules``` 

Regarding point **2**:
If there are paths like ```/users/<id>``` in both app1 and app2, but only app1 needs to be set, how can we do this? Of course, it is possible to configure each app separately, but this will increase complexity.
What I mean is, can we do this:
```
  metric:
    enabled: false
    rules:
    - name: dapr_runtime_service_invocation_req_sent_total
      labels:
      - name: method
        appid: app1
        regex:
          "users/": "users/.+"
```

msfussell:
Tracking doc issue https://github.com/dapr/docs/issues/3987

PR #6729, opened 2023-07-26
[Fix-6723] Part 1: add more data to API logs + improve completness of tracing logs

ItalyPaleAle:
This PR is part 1 of the fix for #6723

It lays the groundwork that will be necessary to complete the change in the next PR - the work was split in 2 parts to make reviews easier.

Part 1 contains:

1. Add more information to API logs for both HTTP and gRPC, including things that are not going to be included in metrics by default anymore:
   - Response status code
   - Response size (if possible; for HTTP only)
   - Latency
2. When `obfuscateURLs` is enabled for API logging, in API logs we now include a more descriptive name for the method rather than the path that was matched in the router. The descriptive names map to the names of the methods in gRPC, for example `SaveState` in place of `POST /state/{storeName}/{name}`. Since gRPC API logs are always, only "obfuscated" (because the params are in the body and not in the "URL"), this makes HTTP and gRPC API logs more consistent too
3. Refactors how tracing and API logging (and the API token auth middleware) get the data from the handler/router:
   - The new approach is a lot more declarative, and less based on heuristics (such as parsing the path from the URL again)
   - The new approach also reduces the number of maps that are allocated and used in each request, which generally contained duplicate information generated in multiple parts of the code
4. For both HTTP and gRPC, the way metadata is added to a tracing span has changed and it's now more declarative:
   - When looking at how tracing spans were composed, the metadata was added in `pkg/diagnostics`, separately from the endpoints. Only a VERY SMALL number of APIs had proper tracing configured (as a matter of fact, see the number of "TODO"'s in this PR), in large part due to the fact that this was in a separate package. The approach also relied a lot on heuristics.
   - For HTTP, now the `Endpoint` struct contains a property to add a function that adds properties to a span for tracing purposes. This lives right next to the handler.
   - For gRPC, messages defined in the protos whose name ends in "Request" now must define an `AppendSpanAttribute` method (this is enforced by a unit test)
5. Update API Allowlisting/Denylisting for HTTP to:
   - Make sure that the same constants can be used for both HTTP and gRPC, especially versions. Right now, versions are in the format "v1.0-alpha1" for HTTP, and "v1alpha1" for gRPC. This PR changes the HTTP versions to be in the format "v1alpha1" too ("v1.0-alpha1" is preserved for bacwkards-compatibility)
   - Improved perf of the HTTP Allowlist/Denylist, especially when using the "new" format (e.g. versions with "v1alpha1"): checking the allowlist is now a simple map lookup rather than an iteration over the entire allowlist.


dapr/proto/runtime/v1/dapr.proto
```diff
@@ -122,7 +122,7 @@ service Dapr {
   rpc DecryptAlpha1(stream DecryptRequest) returns (stream DecryptResponse);
 
   // Gets metadata of the sidecar
-  rpc GetMetadata (google.protobuf.Empty) returns (GetMetadataResponse) {}
+  rpc GetMetadata (GetMetadataRequest) returns (GetMetadataResponse) {}
```
ItalyPaleAle, line 125: Reviewers: this needed to have a message ending in "Request". This change is backwards compatible in the wire (i.e. v1.11 SDKs can talk to this without issues), but it may require a change in the name in the SDK if the protos are updated.

pkg/http/api_directmessaging.go
```diff
@@ -0,0 +1,324 @@
...
+				Name:                  "InvokeService",
+				IsFallback:            true,
+				KeepWildcardUnescaped: true,
+			},
+		},
+	}
+}
+
+// onDirectMessage is the handler for HTTP service invocation
+func (a *api) onDirectMessage(w http.ResponseWriter, r *http.Request) {
```
ItalyPaleAle, line 83: This method, and many more in this file, has been moved from `api.go` without changes, to keep things more organized

pkg/http/api_healthz.go
```diff
@@ -16,26 +16,43 @@ package http
...
+	"github.com/dapr/dapr/pkg/http/endpoints"
 	"github.com/dapr/dapr/pkg/messages"
 )
 
-func (a *api) constructHealthzEndpoints() []Endpoint {
-	return []Endpoint{
+var endpointGroupHealthzV1 = &endpoints.EndpointGroup{
+	Name:                 endpoints.EndpointGroupHealth,
+	Version:              endpoints.EndpointGroupVersion1,
+	AppendSpanAttributes: nil, // TODO
```
ItalyPaleAle, line 26: All of these `TODO`'s are because these endpoints did not have anything defined in their configuration for tracing in the old methods in `pkg/diagnostics`.

Completing these TODOs should be done in separate PRs.

pkg/http/middlewares.go
```diff
@@ -43,8 +42,12 @@ func APITokenAuthMiddleware(token string) func(next http.Handler) http.Handler {
 		}
 
 		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			// Check if the route is excluded from API token auth
+			endpointData, _ := r.Context().Value(endpoints.EndpointCtxKey{}).(*endpoints.EndpointCtxData)
+			excluded := endpointData != nil && endpointData.Settings.BypassAPITokenAuth
+
 			v := r.Header.Get(authConsts.APITokenHeader)
-			if v != token && !isRouteExcludedFromAPITokenAuth(r.Method, r.URL) {
+			if v != token && !excluded {
```
ItalyPaleAle, line 50: This method caused a CVE. It's been changed now so the route itself defines if it can be excluded by API token auth

utils/responsewriter/response_writer.go
```diff
@@ -154,26 +144,3 @@ func (rw *responseWriter) callBefore() {
 		rw.beforeFuncs[i](rw)
 	}
 }
-
-func (rw *responseWriter) SetUserValue(key any, value any) {
```
ItalyPaleAle, line 158: Using a property in the context allows us to get rid of some maps and improve allocations and perf.

pkg/proto/runtime/v1/dapr_tracing_test.go
```diff
@@ -0,0 +1,135 @@
...
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"golang.org/x/exp/slices"
+)
+
+// Tests that all requests messages include the "AppendSpanAttribute" method.
+func TestRequestsIncludeAppendSpanAttribute(t *testing.T) {
```
ItalyPaleAle, line 29: The purpose of this test is to make sure that all messages defined in the proto whose name ends in "Request" include the `AppendSpanAttribute`. Unit tests will fail if that's not included.

pkg/proto/runtime/v1/dapr_tracing.go
```diff
@@ -0,0 +1,192 @@
...
+
+package runtime
+
+import (
+	diagConsts "github.com/dapr/dapr/pkg/diagnostics/consts"
+)
+
+// This file contains additional, hand-written methods added to the generated objects.
+
+func (x *InvokeServiceRequest) AppendSpanAttributes(rpcMethod string, m map[string]string) {
```
ItalyPaleAle, line 22: The `AppendSpanAttribute` methods must be hand-written in this file. It's not as nice as an approach as we have for HTTP, but it's consistent. The new test ensures that developers cannot add a message without the `AppendSpanAttribute` method

Again, the large number of `TODO`'s is due to the fact that these routes did not have proper tracing before. Completing this work should be done in separate PRs.

yaron2:
Including high cardinality disabled by default is important to get in 1.12. However, this PR contains many changes that are out of scope of the issue that this PR is fixing (ie. the "improve completeness of tracing logs" part and other changes). 

Here are the relevant parts for 1.12 (pasted from the [issue](https://github.com/dapr/dapr/issues/6723)):

```
1. By default, do not report metrics per each endpoint/path in the HTTP server. Aggregation will stop at the app ID for service invocation.
2. Add an option to the Configuration CRD metrics.highCardinality, disabled by default, to retain the current behavior.
3. Change the API logging middleware to include additional information on each request, including latency and returned status code. These are things that were before included in the metrics. Although API logging aren't a 1:1 replacement for metrics, users can leverage them in tools like Grafana to gather similar insights.
```

Please revert the changes that do not deal with disabling high cardinality by default and adding latency and status code for API logs. The additional improvements and changes introduced here can easily go into 1.13 with no issues.

ItalyPaleAle:
@yaron2 Although it's true that this PR includes other things, they are all related.

The things mentioned above are a natural consequence of having expanded the metadata included in the "Endpoints" struct. Those changes are very small. I'd rather not have a separate PR given how if it has to wait for 1-2 months till 1.13, it'd drown in merge conflicts...

yaron2:
> @yaron2 Although it's true that this PR includes other things, they are all related.
> 
> The things mentioned above are a natural consequence of having expanded the metadata included in the "Endpoints" struct. They help keeping the code well-organized so the next step of the PR is simpler.
> 
> Those changes are very small. I'd rather not have a separate PR given how if it has to wait for 1-2 months till 1.13, it'd drown in merge conflicts...

The changes related to tracing can be made on a PR that can be opened at any time during the 1.13 milestone, doesn't even have to now, which I agree will drown in merge conflicts.

ItalyPaleAle:
The point is that the changes are done and they helped making this PR and tidying things up. Opening it later doesn't change the fact that there'll be significant merge conflicts.

I do want to stress the usefulness of having things tidier. We were checking the route in multiple places and each one had different heuristics. This PR's main contribution was aggregating everything in a struct such as this:

https://github.com/dapr/dapr/pull/6729/files#diff-037644126d1c1589e20eaf7f82ec9c28a40952d9e29344149567ee8af589b8f3R59-R80

This required changing some middlewares to make sure all route metadata is kept in the context, rather than being split in the context and in the "responsewriter" object (and in a bunch of other places). This is what then naturally enabled the other changes to the tracing and API allowlist middlewares, which are really small.

The next step will be to build upon this expanded struct to improve the diagnostics collection. The data sent to the diag methods will be fetched from the same centralized sources.

yaron2:
Alright, keep the PR as is. We'll triage #6723 to 1.13 and can revisit all the changes in this PR as a whole in the next milestone.

ItalyPaleAle:
Sounds good. We’ll backport this in our distribution for customers who are blocked by the issue and can’t use the OSS version

yaron2:
@ItalyPaleAle the changes in this PR are great, but this release already saw a large amount of different refactors to major parts of the codebase and these risk the release. Our prior experience shows these types of changes are many times the number one cause of patch releases. Changes that fix the original issue alone will be actually prioritized above other PRs (at least as far as I'm concerned).

artursouza:
Thanks, Alessandro and Yaron. I do agree to focus on the original impact on our users prior to refactoring the code. We struggled with that in the past and this is an opportunity to do different. The refactorings should not be bundled with behavior or improvements because it might put the reviewer on the spot to accept the PR because a welcomed change is bundled with changes that will require further discussions.

I think the OSS users also deserve the improvements regarding the original impact described in the issue.

PR #6743, opened 2023-07-28
Fix validate-sidecar-resources workflow.

JoshVanL:
The validate sidecar resources has been broken since #6688 

Use `wait` rather than `lsof` to wait for the sidecar process to shutdown. `wait $pid` can be used as the daprd processes are spawned as a child of the script process.

Echo all issues with resource usage, not just the first and error exit if any of them are over.

Default variables used in script so `./.github/scripts/validate-sidecar-resources.sh` can be run from the repo.

Make curl silent.

Use `/usr/bin/env bash` rather than `/bin/bash` as this path is not available on all systems.


filintod:
lgtm. not needed but not sure if you want to update/remove the defaults in https://github.com/dapr/dapr/blob/f12a49598770abfadc015a0bd4d4b20e816df5a1/.github/workflows/dapr-standalone-validation.yml#L77 as most are now the same as defaults.

JoshVanL:
@filintod thanks, I've removed the duplicate value declarations from the workflow file.

daixiang0 (approved):

PR #6750, opened 2023-07-31
[fix #6739] Do not send duplicate querystring on HTTP invocation

### LINKED ISSUE #6739: Duplicated query strings when using service invocations
## In what area(s)?

/area runtime

> /area operator

> /area placement

> /area docs

> /area test-and-release

## What version of Dapr?

> 1.1.x
> 1.0.x
 edge: from fresh pull

## Expected Behavior

Invoking a method which has query strings should send the query to the service correctly.

## Actual Behavior

Query string is duplicated when sent to the service

- Clone this [repo](https://github.com/Arcalise08/DaprDuplicatedQueryString)
- Verify port 8080 is open, if not, change exposed port in docker-compose.yaml file
- Run Project
- Use browser or postman to navigate to `http://localhost:8080/api/controllerOne/fireMethodWithDapr?code=123` This endpoint calls the other service and echos back what it got as a query string. Notice when you go there it says `123?code=123`
- Use browser or postman to navigate to `http://localhost:8080/api/controllerOne/fireMethodWithoutDapr?code=123`. Notice this endpoint returns the expected result of `"123"`

If you inspect deeper you'll see asp is receiving the query string duplicated. You can also intercept the request before its sent and see the query string has not been duplicated, So the duplication is coming from the dapr sidecar for sure.

## Release Note

RELEASE NOTE: FIX Bug with setting multiple query strings.

ItalyPaleAle:
Fixes #6739

The issue was due to passing the full URL to `findTargetIDAndMethod`; this now passes `r.URL.Path` only

Behavior is now validated by unit tests


pkg/http/api_directmessaging_test.go
```diff
@@ -0,0 +1,1078 @@
...
+					if msg.Method != "fakeMethod" {
+						return false
+					}
+
+					if imr.EncodeHTTPQueryString() != "param1=val1&param2=val2" {
+						return false
+					}
+
+					return true
+				}),
```
ItalyPaleAle, line 521: This is the test that validates the fix (and it's present in a couple of other places)

ItalyPaleAle:
The failing SDK tests are failing in master too. Here's a scheduled run: https://github.com/dapr/dapr/actions/runs/5709199497

halspang (approved):

yaron2 (approved):

PR #6754, opened 2023-08-01
Fixes to perf and E2E tests (fixes perf tests failing 100% of times)

ItalyPaleAle:
Another round of fixes to perf and E2E tests.

- Perf tests have been failing 100% of times since #6493 was merged, because the perf test depended on that behavior/bug. The actor_reminder perf test has been fixed to use a single app now.
- Certain E2E tests now always run with debug logging enabled, which will help debugging issues such as with actor reminders not firing, which are very hard to reproduce outside of our CI. The information contained in the debug logs is crucial to debug the issue, as it shows for example when reminders are evaluated.
- Tightened the scope of a lot of components (still, not all). Aside from reducing noise in logs, this makes it so perf test apps do not spend resources connecting, for example, to Kafka. It may make perf tests better.
- The external service invocation test now performs a healthcheck of the app used for external service invocation, which should hopefully reduce flakiness
- A few other fixes in the test framework itself:
  - In certain cases logs from containers were not saved. This seemed to happen when there was a race condition between us requesting the logs and the pod being not yet ready. This has been fixed by adding automatic retries, so logs are always collected.
  - (Finally) fixed the fact that all of our E2E tests printed the warning `spec.template.spec.imagePullSecrets[0].name: invalid empty name ""`


ItalyPaleAle:
Tests (E2E and perf) are now passing. Ready for review

halspang (approved):
Just some nits about commented logs.

yaron2:
Great to see perf tests running again

Issue #4233, opened 2022-02-10
Outbox pattern for statestore + pubsub in Dapr


gdlcf88:
/area runtime

## Overview

Hi guys. Maybe Dapr could provide an outbox & inbox infrastructure since any distributed app development framework needs it to make sure domain events are published and subscribed.

This issue only discusses outbox, I will continue to think about the inbox later and create another [issue for the inbox](https://github.com/dapr/dapr/issues/4247).

## Why the Outbox Pattern?

Given we want to publish an `OrderPaidEvent` domain event after an order is paid. Normally this will work:

[...]

PR #6755, opened 2023-08-01
Add outbox support

yaron2:
Closes https://github.com/dapr/dapr/issues/4233.

This will be added as a preview feature [here](https://docs.dapr.io/operations/support/support-preview-features/#current-preview-features).


pkg/runtime/pubsub/outbox.go
```diff
@@ -0,0 +1,294 @@
...
+	defer o.lock.RUnlock()
+
+	_, ok := o.outboxStores[stateStore]
+	return ok
+}
+
+func transaction() state.TransactionalStateOperation {
+	return state.SetRequest{
+		Key:   outboxStatePrefix + "-" + uuid.New().String(),
+		Value: "0",
```
yaron2, line 125: Technically a value isn't needed here at all, but there isn't a deterministic way in Dapr yet to distinguish between key not found and empty value.

pkg/runtime/pubsub/outbox.go
```diff
@@ -0,0 +1,292 @@
...
+	o.lock.RLock()
+	defer o.lock.RUnlock()
+
+	_, ok := o.outboxStores[stateStore]
+	return ok
+}
+
+func transaction() state.TransactionalStateOperation {
+	return state.SetRequest{
+		Key:   outboxStatePrefix + "-" + uuid.New().String(),
```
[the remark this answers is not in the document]
yaron2: Indeed. Changed

pkg/runtime/pubsub/outbox.go
```diff
@@ -0,0 +1,292 @@
...
+			time.Sleep(d)
+			policyRunner := resiliency.NewRunner[any](ctx,
+				resiliency.NewPolicyDefinition(outboxLogger, "outbox", d, &retry.Config{
+					Policy:      retry.PolicyExponential,
+					MaxInterval: time.Second * 15,
+					MaxRetries:  3,
+				}, nil),
+			)
+
+			_, err = policyRunner(func(ctx context.Context) (any, error) {
```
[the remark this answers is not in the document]
yaron2: Good to know, changed. Might want to track changing this signature for changing in other places in the code to help set a standard for this.

pkg/runtime/processor/pubsub/pubsub.go
```diff
@@ -238,6 +248,10 @@ func matchRoutingRule(rules []*rtpubsub.Rule, data map[string]interface{}) (*rtp
 	return nil, nil
 }
 
+func ExtractCloudEventProperty(cloudEvent map[string]any, property string) string {
```
artursouza, line 251: nit: is this being used?

pkg/runtime/processor/pubsub/pubsub.go
```diff
@@ -238,6 +248,10 @@ func matchRoutingRule(rules []*rtpubsub.Rule, data map[string]interface{}) (*rtp
...
+	return extractCloudEventProperty(cloudEvent, property)
```
artursouza, line 252: Why not just rename it to ExtraCloudProperty and avoid the wrapper?

artursouza (commented):
LGTM overall. Maybe E2E tests to follow.

pkg/runtime/processor/pubsub/pubsub.go
yaron2: Renamed it. Originally wanted to avoid multiple changes.

pkg/runtime/processor/pubsub/pubsub.go
yaron2: Changed below

yaron2:
> LGTM overall. Maybe E2E tests to follow.

Yes, I'll add E2E most likely near the end of the 1.13 milestone once we're confident the underlying implementation is stable. It's definitely an exit criteria out of preview.

pkg/runtime/pubsub/outbox.go
```diff
@@ -0,0 +1,305 @@
...
+			if !ok {
+				return fmt.Errorf("cannot get outbox state: state store %s not found", stateStore)
+			}
+
+			time.Sleep(d)
+			policyRunner := resiliency.NewRunner[struct{}](ctx,
+				resiliency.NewPolicyDefinition(outboxLogger, "outbox", d, &retry.Config{
+					Policy:      retry.PolicyExponential,
+					MaxInterval: time.Second * 15,
+					MaxRetries:  3,
```
famarting, line 239: the fact that this does finite retries, makes it not providing a strong guarantee that the final event will be delivered... with the limitations that exist there is little things you can do about this... one possible improvement is to make the max retries configurable. Still does not provide a guarantee of strong consistency but should help for most of the cases

pkg/runtime/pubsub/outbox.go
```diff
@@ -0,0 +1,305 @@
...
+	for _, op := range operations {
+		sr, ok := op.(state.SetRequest)
+		if ok {
+			tr, err := transaction()
+			if err != nil {
+				return nil, err
+			}
+
+			var ceData []byte
+			bt, ok := sr.Value.([]byte)
```
famarting, line 152: the original key of the user state request is not included in the outbox event, users may want it

pkg/runtime/pubsub/outbox.go
yaron2: This has been changed in a follow up PR, there's now a strong guarantee.

pkg/runtime/pubsub/outbox.go
yaron2: That might be useful in some contexts and has come up by users already

PR #6756, opened 2023-08-01
Misc fixes and tweaks from #6729

ItalyPaleAle:
Includes some minor tweaks and fixes from the closed PR #6729

Nothing of substance or that changes behavior. Most relevant changes (and "most" is relative):

- Avoid unnecessary compStore property in gRPC and HTTP API structs
- Remove now-unused KeepWildcardUnescaped property (after #6750)
- Misc improvements to a few unit tests

### DIFF
--- a/pkg/acl/acl.go
+++ b/pkg/acl/acl.go
@@ -247,7 +247,7 @@ func ApplyAccessControlPolicies(ctx context.Context, operation string, httpVerb 
 	spiffeID, err := GetAndParseSpiffeID(ctx)
 	if err != nil {
 		// Apply the default action
-		log.Debugf("error while reading spiffe id from client cert: %v. applying default global policy action", err.Error())
+		log.Debugf("Error while reading spiffe id from client cert: %v. applying default global policy action", err)
 	}
 	var appID, trustDomain, namespace string
 	if spiffeID != nil {
@@ -260,7 +260,7 @@ func ApplyAccessControlPolicies(ctx context.Context, operation string, httpVerb 
 	var errMessage string
 
 	if err != nil {
-		errMessage = fmt.Sprintf("error in method normalization: %s", err)
+		errMessage = fmt.Sprintf("error in method normalization: %v", err)
 		log.Debugf(errMessage)
 		return false, errMessage
 	}
--- a/pkg/diagnostics/metrics.go
+++ b/pkg/diagnostics/metrics.go
@@ -68,5 +68,11 @@ func InitMetrics(appID, namespace string, rules []config.MetricsRule) error {
 
 	// Set reporting period of views
 	view.SetReportingPeriod(DefaultReportingPeriod)
-	return utils.CreateRulesMap(rules)
+
+	// Creates a rules map
+	err := utils.CreateRulesMap(rules)
+	if err != nil {
+		return err
+	}
+	return nil
 }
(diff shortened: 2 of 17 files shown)


pkg/http/server.go
```diff
@@ -257,7 +257,7 @@ func (s *server) useMaxBodySize(r chi.Router) {
 	}
 
 	maxSize := int64(s.config.MaxRequestBodySizeMB) << 20 // To bytes
-	log.Infof("Enabled max body size HTTP middleware with size %d MB", maxSize)
+	log.Infof("Enabled max body size HTTP middleware with size %d bytes", maxSize)
```
JoshVanL, line 260: Please print the size in MB for logging. Bytes is essentially useless to humans without getting the calculator out at this order of magnitude.

JoshVanL (dismissed):
Small nit from me

pkg/http/server.go
ItalyPaleAle: Yeah that's a good callout. Feel free to make a suggestion too!
```suggestion
	log.Infof("Enabled max body size HTTP middleware with size %d MB", s.config.MaxRequestBodySizeMB)
```

pkg/diagnostics/metrics.go
```diff
@@ -68,5 +68,11 @@ func InitMetrics(appID, namespace string, rules []config.MetricsRule) error {
 
 	// Set reporting period of views
 	view.SetReportingPeriod(DefaultReportingPeriod)
-	return utils.CreateRulesMap(rules)
```
shubham1172, line 71: Isn't this change functionally same and more verbose than a direct return statement?

pkg/diagnostics/metrics.go
```diff
@@ -68,5 +68,11 @@ func InitMetrics(appID, namespace string, rules []config.MetricsRule) error {
...
+
+	// Creates a rules map
+	err := utils.CreateRulesMap(rules)
+	if err != nil {
+		return err
+	}
+	return nil
```
ItalyPaleAle, line 77:
```suggestion
	return utils.CreateRulesMap(rules)
```

pkg/diagnostics/metrics.go
ItalyPaleAle: Reverted it. I do find that a bit clearer to read but not worth the diff.

shubham1172 (approved):

pkg/acl/acl.go
```diff
@@ -260,7 +260,7 @@ func ApplyAccessControlPolicies(ctx context.Context, operation string, httpVerb
 	var errMessage string
 
 	if err != nil {
-		errMessage = fmt.Sprintf("error in method normalization: %s", err)
+		errMessage = fmt.Sprintf("error in method normalization: %v", err)
```
DeepanshuA, line 263: I know there are couple of `fmt.Sprintf` in our code, and I feel at some places out of those, a string concatenation would have been better probably.
Do we want to do a string concatenation here?

ItalyPaleAle: Yeah, although for error handling it's probably ok since it should not be hot paths

DeepanshuA (approved):

pkg/http/api_directmessaging.go
```diff
@@ -41,10 +41,9 @@ func (a *api) constructDirectMessagingEndpoints() []Endpoint {
 			Methods: []string{},
 			Route:   "invoke/*",
 			// This is the fallback route for when no other method is matched by the router
-			IsFallback:            true,
-			Version:               apiVersionV1,
-			KeepWildcardUnescaped: true,
```
mukundansundar, line 46: is true the only value for this bool in all code paths??

ItalyPaleAle: KeepWildcardUnescaped was already a no-op since #6750 and it's "false" for all routes

mukundansundar (approved):

PR #6759, opened 2023-08-02
Grpc Proxy service invocation metrics

filintod:
# Description

Started from discussion in https://github.com/dapr/dapr/pull/6676 where metrics were still not showing up.

This will allow for metrics to show up as in:

```
# TYPE dapr_runtime_service_invocation_req_sent_total counter
dapr_runtime_service_invocation_req_sent_total{app_id="client",dst_app_id="server",method="/helloworld.Greeter/SayHello",type="unary"} 3
# HELP dapr_runtime_service_invocation_res_recv_latency_ms The latency of service invocation response.
# TYPE dapr_runtime_service_invocation_res_recv_latency_ms histogram
dapr_runtime_service_invocation_res_recv_latency_ms_bucket{app_id="client",method="/helloworld.Greeter/SayHello",src_app_id="server",status="0",le="1"} 0
dapr_runtime_service_invocation_res_recv_latency_ms_bucket{app_id="client",method="/helloworld.Greeter/SayHello",src_app_id="server",status="0",le="2"} 0
dapr_runtime_service_invocation_res_recv_latency_ms_bucket{app_id="client",method="/helloworld.Greeter/SayHello",src_app_id="server",status="0",le="3"} 0
```

And also for streaming. On streaming we don't create a histogram and only capture sent total and recv total if error but without the recv histogram.

Service Invocation Totals Request Sent and Response Receive will have a new label `type` with possible values `unary` or `streaming`.   The status on response received entries is the grpc status code. The histogram does not have the type as it is only calculated for `unary` operations and streaming histograms do not make much sense.

## Issue reference

Please reference the issue this PR will close: #6459 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Unit tests passing


pkg/grpc/proxy/handler.go
```diff
@@ -161,6 +165,17 @@ func (s *handler) handler(srv any, serverStream grpc.ServerStream) error {
 		// Do not "defer clientCancel()" yet, in case we need to proxy a stream
 		clientCtx, clientCancel := context.WithCancel(outgoingCtx)
 
+		// should we count each retry as a new request from client for metrics?? is so we'll need to move this inside the
```
ItalyPaleAle, line 168: Retries should be considered as part of the same request, and in fact they increase the latency.

PS: Leave this as comment in the PR rather than in the code please :)

ItalyPaleAle: Oh ok. Then that's probably fine, we should be consistent

filintod:
@ItalyPaleAle I could add a label to indicate that the request is a retry?

filintod:
I will leave as is to keep the current behavior (count metrics inside resiliency loop). We can add label in another PR to indicate the execution is part of the resiliency loop to indicate flakiness/overloading of the system.

artursouza (approved):

PR #6760, opened 2023-08-02
Remove `ActorScanInterval`

ItalyPaleAle:
Small refactoring to the actors subsystem to remove the "ActorScanInterval" configuration. Idle actors are now deactivated automatically shortly after the idle timeout for their actor type.

Benefits:


ItalyPaleAle:
My apologies, this was re-opened by accident

PR #6761, opened 2023-08-03
[5756 Identity Management Overhaul] Injector & daprd 

JoshVanL:


pkg/security/x509source.go
```diff
@@ -277,6 +316,91 @@ func (x *x509source) requestFromSentry(ctx context.Context, csrDER []byte) ([]*x
...
+	return nil
+}
+
+func (x *x509source) updateTrustAnchorFromFile(filepath string) error {
+	x.lock.RLock()
+	defer x.lock.RUnlock()
+
+	rootPEMs, err := os.ReadFile(filepath)
+	if err != nil {
+		return fmt.Errorf("failed to read trust anchors file %q: %s", filepath, err)
```
filintod, line 355: any reason to not wrap the error `%w`?

pkg/security/x509source.go
```diff
@@ -277,6 +316,91 @@ func (x *x509source) requestFromSentry(ctx context.Context, csrDER []byte) ([]*x
...
+	return nil
+}
+
+func atomicWrite(clock clock.Clock, dir string, data map[string][]byte) error {
+	dir = filepath.Clean(dir)
+	if err := os.MkdirAll(filepath.Dir(dir), 0o700); err != nil {
+		return err
+	}
+
+	newDir := dir + "-" + clock.Now().Format(time.RFC3339)
```
filintod, line 382: maybe not too important in the container, but should we remove the dir on error?

cmd/injector/main.go
```diff
@@ -105,6 +124,43 @@ func main() {
...
+			caBundle, rErr := sec.CurrentTrustAnchors()
+			if rErr != nil {
+				return rErr
+			}
+
+			for {
+				_, rErr = kubeClient.AdmissionregistrationV1().MutatingWebhookConfigurations().Patch(ctx,
+					"dapr-sidecar-injector",
+					types.JSONPatchType,
+					[]byte(`[{"op":"replace","path":"/webhooks/0/clientConfig/caBundle","value":"`+base64.StdEncoding.EncodeToString(caBundle)+`"}]`),
```
ItalyPaleAle, line 150: What does this do?

pkg/grpc/grpc.go
```diff
@@ -207,29 +203,7 @@ func (g *Manager) connectRemote(
...
-		ta := credentials.NewTLS(&tls.Config{
-			ServerName:   serverName,
-			Certificates: []tls.Certificate{cert},
-			RootCAs:      signedCert.TrustChain,
-		})
-		opts = append(opts, grpc.WithTransportCredentials(ta))
-	} else {
-		opts = append(opts, grpc.WithTransportCredentials(insecure.NewCredentials()))
-	}
+	opts = append(opts, g.sec.GRPCDialOptionUnknownTrustDomain(namespace, id))
```
ItalyPaleAle, line 206: Would this cause issues if a sidecar was trying to connect to a v1.11 sidecar?

pkg/injector/patcher/sidecar.go
```diff
@@ -22,6 +22,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 
 	injectorConsts "github.com/dapr/dapr/pkg/injector/consts"
+	"github.com/dapr/dapr/pkg/security"
```
ItalyPaleAle, line 25: pkg/injector just went through a refactoring to make it have minimal dependencies on the rest of Dapr. THis is because it can be used as a library outside of Dapr. Looks like pulling in pkg/security brings in a lot, including transitive dependencies such as pkg/diagnostics. What is this package used for in the injector, and can an alternative be found please?

pkg/injector/patcher/sidecar_container.go
```diff
@@ -70,6 +70,9 @@ func (c *SidecarConfig) getSidecarContainer(opts getSidecarContainerOpts) (*core
 		"--app-protocol", c.AppProtocol,
 		"--log-level", c.LogLevel,
 		"--dapr-graceful-shutdown-seconds", strconv.Itoa(c.GracefulShutdownSeconds),
+		"--control-plane-namespace", c.sec.ControlPlaneNamespace(),
```
ItalyPaleAle, line 73: This will make the injector unable to inject Dapr sidecars running older versions of Dapr, since they'll see an unsupported CLI flag and crash. The E2E tests will fail with this.

pkg/injector/patcher/sidecar_container.go
```diff
@@ -84,10 +87,6 @@ func (c *SidecarConfig) getSidecarContainer(opts getSidecarContainerOpts) (*core
 		args = append(args, "--control-plane-address", c.OperatorAddress)
 	}
 
-	if c.SentryAddress != "" {
```
ItalyPaleAle, line 87: Why was this removed? Can't SentryAddress be empty?

charts/dapr/charts/dapr_sidecar_injector/templates/dapr_sidecar_injector_deployment.yaml
```diff
@@ -191,13 +185,23 @@ spec:
...
           readOnly: true
       volumes:
-      - name: cert
-        secret:
-          secretName: dapr-sidecar-injector-cert
+      - name: dapr-trust-bundle
+        configMap:
+          name: dapr-trust-bundle
+      - name: dapr-identity-token
+        projected:
```
ItalyPaleAle, line 199: Comment here too as a reminder that we must support non-projected volumes too (with a JWKS token)

pkg/runtime/config.go
```diff
@@ -122,7 +123,6 @@ type internalConfig struct {
 	allowedOrigins               string
 	standalone                   configmodes.StandaloneConfig
 	kubernetes                   configmodes.KubernetesConfig
-	mTLSEnabled                  bool
```
ItalyPaleAle, line 125: Please don't just move properties around as it risks causing merge conflicts unnecessarily.

pkg/security/security.go
```diff
@@ -82,32 +102,42 @@ type Options struct {
...
 
+	var source *x509source
 	if opts.MTLSEnabled {
-		if len(opts.TrustAnchors) == 0 {
+		if len(opts.TrustAnchors) > 0 && len(opts.TrustAnchorsFile) > 0 {
+			return nil, errors.New("trust anchors cannot be specified in both TrustAnchors and TrustAnchorsFile")
+		}
+
+		if len(opts.TrustAnchors) == 0 && len(opts.TrustAnchorsFile) == 0 {
 			return nil, errors.New("trust anchors are required")
```
ItalyPaleAle, line 136:
```suggestion
			return nil, errors.New("trust anchors are required: either one of TrustAnchors or TrustAnchorsFile")
```

pkg/security/security.go
```diff
@@ -140,20 +172,52 @@ func (p *provider) Start(ctx context.Context) error {
...
+				log.Infof("Watching trust anchors file '%s' for changes", p.trustAnchorsFile)
+				return fswatcher.Watch(ctx, p.trustAnchorsFile, caEvent)
+			},
+			func(ctx context.Context) error {
+				for {
+					select {
+					case <-ctx.Done():
+						return nil
+					case <-caEvent:
+						log.Info("trust anchors file changed, reloading trust anchors")
```
ItalyPaleAle, line 202: Nit:
```suggestion
						log.Info("Trust anchors file changed, reloading trust anchors")
```

pkg/security/security.go
```diff
@@ -167,21 +231,100 @@ func (p *provider) Handler(ctx context.Context) (Handler, error) {
...
+// instruments client authentication using the current signed client
+// certificate. Doesn't verify the servers trust domain, but does authorize the
+// SPIFFE ID path.
+// Used for clients which don't know the servers Trust Domain.
+func (s *security) GRPCDialOptionUnknownTrustDomain(ns, appID string) grpc.DialOption {
+	if !s.mtls {
+		return grpc.WithTransportCredentials(insecure.NewCredentials())
+	}
+
+	expID := fmt.Sprintf("/ns/%s/%s", ns, appID)
```
ItalyPaleAle, line 268: Nit, since I believe this is used in a hot path:


```suggestion
	expID := "/ns/" + ns + "/" + appID
```

pkg/security/security.go
```diff
@@ -167,21 +231,100 @@ func (p *provider) Handler(ctx context.Context) (Handler, error) {
...
+	s.source.lock.Unlock()
+
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case <-sub:
+			caBundle, err := s.CurrentTrustAnchors()
+			if err != nil {
+				log.Errorf("failed to marshal trust anchors: %s", err)
```
ItalyPaleAle, line 297: failed to marshal?

pkg/security/x509source.go
```diff
@@ -84,31 +82,68 @@ type x509source struct {
...
+				return nil, ctx.Err()
+			case <-clock.After(time.Second):
+				log.Warnf("trust anchors file '%s' not found, waiting...", opts.TrustAnchorsFile)
+			}
+		}
+
+		defer f.Close()
+		rootPEMs, err = io.ReadAll(f)
+		if err != nil {
+			return nil, fmt.Errorf("failed to read trust anchors file %q: %w", opts.TrustAnchorsFile, err)
```
ItalyPaleAle, line 123:
```suggestion
			return nil, fmt.Errorf("failed to read trust anchors file '%s': %w", opts.TrustAnchorsFile, err)
```

pkg/security/x509source.go
```diff
@@ -277,6 +316,91 @@ func (x *x509source) requestFromSentry(ctx context.Context, csrDER []byte) ([]*x
...
+
+	if err := atomicWrite(x.clock, *x.writeToDiskDir, map[string][]byte{
+		"tls.crt": certs,
+		"tls.key": pk,
+		"ca.crt":  ta,
+	}); err != nil {
+		return fmt.Errorf("unable to write SVID to disk: %w", err)
+	}
+
+	log.Infof("Wrote SVID to %q", *x.writeToDiskDir)
```
ItalyPaleAle, line 344:
```suggestion
	log.Infof("Wrote SVID to '%s'", *x.writeToDiskDir)
```

pkg/security/x509source.go
```diff
@@ -277,6 +316,91 @@ func (x *x509source) requestFromSentry(ctx context.Context, csrDER []byte) ([]*x
...
+		return err
+	}
+
+	for file, contents := range data {
+		if err := os.WriteFile(filepath.Join(newDir, file), contents, 0o600); err != nil {
+			return err
+		}
+	}
+
+	if err := os.Symlink(newDir, dir+"-new"); err != nil {
```
ItalyPaleAle, line 393: Will this work on Windows without issues?

ItalyPaleAle (changes requested):

cmd/injector/main.go
JoshVanL: Patches the injector's mutating webhook with the latest trust bundle.

[Number 4 of current shortfalls in the parent issue](https://github.com/dapr/dapr/issues/5756).

pkg/grpc/grpc.go
JoshVanL: Yes, we are missing a legacy wrapper for `GRPCDialOptionUnknownTrustDomain`.

pkg/injector/patcher/sidecar_container.go
JoshVanL: How have flags been added in the past?

pkg/injector/patcher/sidecar_container.go
JoshVanL: No- why would the sentry address ever be empty?

pkg/security/security.go
JoshVanL: Yes:

```go
// CurrentTrustAnchors returns the current trust anchors for this Dapr
// installation.
func (s *security) CurrentTrustAnchors() ([]byte, error) {
	if s.source == nil {
		return nil, nil
	}
	return s.source.trustAnchors.Marshal()
}
```

pkg/security/x509source.go
JoshVanL: Fixed

cmd/injector/main.go
ItalyPaleAle: Got it. Please add a comment explaining what this is doing for future contributors :)

pkg/injector/patcher/sidecar_container.go
ItalyPaleAle: It can be empty though, right? In standalone mode daprd runs without sentry. So it's _possible_ even if unlikely.

pkg/security/security.go
ItalyPaleAle: Ok. A nit would be to wrap the error in `CurrentTrustAnchors` since the name of the method doesn't imply a connection with marshalling

pkg/injector/patcher/sidecar_container.go
JoshVanL: Right, but this injector patching would not be happening in standalone mode.

ItalyPaleAle: Correct. But it's possible to use daprd without sentry. So when someone uses the injector as a library, they have the choice.

pkg/injector/patcher/sidecar_container.go
ItalyPaleAle: This will help for the future, although can't help for today (we need to wait at least 2 more Dapr releases before we can leverage it): https://github.com/dapr/dapr/pull/6766

JoshVanL: I'm not a fan of ignoring unknown flags. Exiting error on unknown flags is standard for all programs, and makes things super confusing for users to know whether a CLI flag was even picked up. I don't want to think about the human hours wasted by not noticing a spelling mistake in a flag :grimacing: 


We can use environment variables for backwards compat until all supported versions have the CLI flag.

pkg/runtime/runtime.go
```diff
@@ -1459,7 +1457,7 @@ func isEnvVarAllowed(key string) bool {
 	}
 
 	// If we have a `DAPR_ENV_KEYS` env var (which is added by the Dapr Injector in Kubernetes mode), use that as allowlist too
-	allowlist := strings.ToUpper(os.Getenv(securityConsts.EnvKeysEnvVar))
+	allowlist := os.Getenv(securityConsts.EnvKeysEnvVar)
```
ItalyPaleAle, line 1460: THis really needs not to be reverted please

```suggestion
	allowlist := strings.ToUpper(os.Getenv(securityConsts.EnvKeysEnvVar))
```

pkg/security/consts/consts.go
```diff
@@ -13,13 +13,15 @@ limitations under the License.
 
 package consts
 
-/* #nosec */
```
ItalyPaleAle, line 16: Why was this removed in place of a less ergonomic sequence of `//nolint:gosec`?

pkg/security/consts/consts.go
```diff
@@ -41,9 +43,20 @@ const (
...
 	// SentryTokenFileEnvVar is the environment variable for the Sentry token file.
+	//nolint:gosec
 	SentryTokenFileEnvVar = "DAPR_SENTRY_TOKEN_FILE"
 
 	// AnnotationKeyControlPlane is the annotation to mark a control plane
 	// component. The value is the name of the control plane service.
 	AnnotationKeyControlPlane = "dapr.io/control-plane"
 )
+
+const (
```
ItalyPaleAle, line 54:
```suggestion

```

ItalyPaleAle (changes requested):

pkg/injector/service/config.go
```diff
@@ -38,6 +36,10 @@ type Config struct {
 	ReadOnlyRootFilesystem            string `envconfig:"SIDECAR_READ_ONLY_ROOT_FILESYSTEM"`
 	SidecarDropALLCapabilities        string `envconfig:"SIDECAR_DROP_ALL_CAPABILITIES"`
 
+	TrustAnchorsFile        string `envconfig:"DAPR_TRUST_ANCHORS_FILE"`
```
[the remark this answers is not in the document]
JoshVanL: A filename containing PEM encoded root Certificate Authorities

tests/e2e/utils/http.go
```diff
@@ -41,11 +41,6 @@ func InitHTTPClient(allowHTTP2 bool) {
 	httpClient = NewHTTPClient(allowHTTP2)
 }
```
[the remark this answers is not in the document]
JoshVanL: It's not being used.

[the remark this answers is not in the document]
ItalyPaleAle: It was used by a test that was removed, and perhaps could be re-added one day. It still doesn't hurt to keep it.

cmd/injector/options/options.go
```diff
@@ -31,14 +30,16 @@ type Options struct {
...
 	var opts Options
 
 	flag.IntVar(&opts.HealthzPort, "healthz-port", 8080, "The port used for health checks")
 
-	flag.StringVar(&credentials.RootCertFilename, "issuer-ca-secret-key", credentials.RootCertFilename, "Certificate Authority certificate secret key")
-	flag.StringVar(&credentials.IssuerCertFilename, "issuer-certificate-secret-key", credentials.IssuerCertFilename, "Issuer certificate secret key")
-	flag.StringVar(&credentials.IssuerKeyFilename, "issuer-key-secret-key", credentials.IssuerKeyFilename, "Issuer private key secret key")
+	depCAFlag := flag.String("issuer-ca-secret-key", "", "DEPRECATED")
+	depCertFlag := flag.String("issuer-certificate-secret-key", "", "DEPRECATED")
+	depKeyFlag := flag.String("issuer-key-secret-key", "", "DEPRECATED")
```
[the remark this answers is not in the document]
JoshVanL: Done.

cmd/injector/options/options.go
```diff
@@ -31,14 +30,16 @@ type Options struct {
...
+
 func New() *Options {
 	var opts Options
 
 	flag.IntVar(&opts.HealthzPort, "healthz-port", 8080, "The port used for health checks")
 
-	flag.StringVar(&credentials.RootCertFilename, "issuer-ca-secret-key", credentials.RootCertFilename, "Certificate Authority certificate secret key")
-	flag.StringVar(&credentials.IssuerCertFilename, "issuer-certificate-secret-key", credentials.IssuerCertFilename, "Issuer certificate secret key")
-	flag.StringVar(&credentials.IssuerKeyFilename, "issuer-key-secret-key", credentials.IssuerKeyFilename, "Issuer private key secret key")
+	depCAFlag := flag.String("issuer-ca-secret-key", "", "DEPRECATED; Certificate Authority certificate secret key")
```
ItalyPaleAle, line 40: I think these flags can be deleted right away.

This is a control plane service that is only used in K8s. There's no risk that someone accidentally passes these flags, since the injector is only deployed by Helm.

pkg/acl/acl.go
```diff
@@ -244,16 +151,15 @@ func normalizeOperation(operation string) (string, error) {
 
 func ApplyAccessControlPolicies(ctx context.Context, operation string, httpVerb commonv1pb.HTTPExtension_Verb, isHTTP bool, acl *config.AccessControlList) (bool, string) {
 	// Apply access control list filter
-	spiffeID, err := GetAndParseSpiffeID(ctx)
+	spiffeID, ok, err := spiffe.FromGRPCContext(ctx)
 	if err != nil {
-		// Apply the default action
-		log.Debugf("Error while reading spiffe id from client cert: %v. applying default global policy action", err)
+		log.Debugf(err.Error())
```
ItalyPaleAle, line 156: Leftover log? Or should this at least have some message around it?

pkg/acl/acl.go
```diff
@@ -244,16 +151,15 @@ func normalizeOperation(operation string) (string, error) {
...
 	}
-	var appID, trustDomain, namespace string
-	if spiffeID != nil {
-		appID = spiffeID.AppID
-		namespace = spiffeID.Namespace
-		trustDomain = spiffeID.TrustDomain
+
+	if !ok {
+		// Apply the default action
+		log.Debugf("error while reading spiffe id from client cert. applying default global policy action")
```
ItalyPaleAle, line 162: We uppercase logs
```suggestion
		log.Debugf("Error while reading SPIFFE ID from client cert; applying default global policy action")
```

pkg/acl/acl.go
```diff
@@ -265,37 +171,37 @@ func ApplyAccessControlPolicies(ctx context.Context, operation string, httpVerb
...
 
-	action, actionPolicy := IsOperationAllowedByAccessControlPolicy(spiffeID, appID, operation, httpVerb, isHTTP, acl)
-	emitACLMetrics(actionPolicy, appID, trustDomain, namespace, operation, httpVerb.String(), action)
+	action, actionPolicy := isOperationAllowedByAccessControlPolicy(spiffeID, operation, httpVerb, isHTTP, acl)
+	emitACLMetrics(spiffeID, actionPolicy, operation, httpVerb.String(), action)
 
 	if !action {
-		errMessage = fmt.Sprintf("access control policy has denied access to appid: %s operation: %s verb: %s", appID, operation, httpVerb)
+		errMessage = fmt.Sprintf("access control policy has denied access to id: %s operation: %s verb: %s", spiffeID.URL(), operation, httpVerb)
 		log.Debugf(errMessage)
```
ItalyPaleAle, line 179:
```suggestion
		log.Debug(errMessage)
```

pkg/actors/placement/grpc_opts.go
```diff
@@ -45,17 +46,19 @@ func getGrpcOptsGetter(servers []string, clientCert *daprCredentials.CertChain)
...
 			return nil, errEstablishingTLSConn
 		}
 
+		opts = append(opts, sec.GRPCDialOptionMTLS(placementID))
+
 		if diag.DefaultGRPCMonitoring.IsEnabled() {
 			opts = append(
 				opts,
 				grpc.WithUnaryInterceptor(diag.DefaultGRPCMonitoring.UnaryClientInterceptor()),
-				grpc.WithBlock(),
```
ItalyPaleAle, line 58: Why was this removed? This is quite important

pkg/grpc/server.go
```diff
@@ -289,87 +264,25 @@ func (s *server) getGRPCServer() (*grpc.Server, error) {
...
-		//nolint:gosec
-		tlsConfig := tls.Config{
-			ClientCAs:  s.signedCert.TrustChain,
-			ClientAuth: tls.RequireAndVerifyClientCert,
-			GetCertificate: func(*tls.ClientHelloInfo) (*tls.Certificate, error) {
-				return &s.tlsCert, nil
-			},
-		}
-
-		// In the internal server, enforce minimum version TLS 1.2
```
ItalyPaleAle, line 307: Looks like this is not enforced anymore in the new code? We need to make sure we are enforcing TLS 1.2+

pkg/grpc/server.go
```diff
@@ -289,87 +264,25 @@ func (s *server) getGRPCServer() (*grpc.Server, error) {
...
+	}
+
 	if s.proxy != nil {
 		opts = append(opts, grpc.UnknownServiceHandler(s.proxy.Handler()))
 	}
 
 	return grpc.NewServer(opts...), nil
 }
 
-func (s *server) startWorkloadCertRotation() {
```
ItalyPaleAle, line 335: Happy to see this gone :) So long, and thanks for all the fish!

pkg/injector/service/injector.go
```diff
@@ -56,12 +55,27 @@ var AllowedServiceAccountInfos = []string{
 	"tekton-pipelines:tekton-pipelines-controller",
 }
 
+type (
+	signDaprdCertificateFn func(context.Context, string) ([]byte, []byte, error)
+	currentTrustAnchorsFn  func() ([]byte, error)
+)
```
ItalyPaleAle, line 61: Please, use named arguments and returned value, so we know what this function's expected and returned values are. Just seeing "string" or "[]byte" is not clear

pkg/injector/service/injector.go
```diff
@@ -110,29 +129,28 @@ func getAppIDFromRequest(req *admissionv1.AdmissionRequest) (appID string) {
...
-		config: config,
+		config: opts.Config,
 		deserializer: serializer.NewCodecFactory(
 			runtime.NewScheme(),
 		).UniversalDeserializer(),
 		server: &http.Server{
-			Addr:    fmt.Sprintf(":%d", port),
-			Handler: mux,
-			TLSConfig: &tls.Config{
-				MinVersion: tls.VersionTLS12,
```
ItalyPaleAle, line 125: This needs to be kept for all control plane services

pkg/security/spiffe/spiffe.go
```diff
@@ -24,42 +26,77 @@ import (
...
+	id spiffeid.ID
+
+	namespace string
+	appID     string
 }
 
-// FromContext parses a SPIFFE ID from a gRPC context.
-func FromContext(ctx context.Context) (Parsed, bool, error) {
+// FromGRPCContext parses a SPIFFE ID from a gRPC context.
+func FromGRPCContext(ctx context.Context) (*Parsed, bool, error) {
```
ItalyPaleAle, line 36: Question, why was the Parsed returned value changed to a pointer (heap-allocated)?

pkg/security/spiffe/spiffe.go
```diff
@@ -24,42 +26,77 @@ import (
...
 
-	return spiffeid.FromSegments(td, "ns", p.Namespace, p.AppID)
+	return &Parsed{
+		id:        id,
+		namespace: namespace,
+		appID:     appID,
+	}, nil
+}
+
+func (p *Parsed) TrustDomain() spiffeid.TrustDomain {
```
ItalyPaleAle, line 76: All of these methods would probably be not necessary if we did not change From* functions to return a pointer. Which would also cause less allocations on the heap.

pkg/acl/acl.go
JoshVanL: Yeah, the message describing the action has been moved to the log line below. I'll add context to the error here :+1:

cmd/injector/options/options.go
JoshVanL: I'm happy to do this! I'll make the change.

pkg/actors/placement/grpc_opts.go
JoshVanL: It is explicitly called out as not recommended in [the docs](https://pkg.go.dev/google.golang.org/grpc#WithBlock), but can add it back if you think we should?

pkg/grpc/server.go
JoshVanL: It is in the spiffe library https://github.com/spiffe/go-spiffe/blob/3000ff09ead100134b2828fec1151b0419a988b0/v2/spiffetls/tlsconfig/config.go#L239

pkg/injector/service/injector.go
JoshVanL: Done in the spiffe library.

pkg/security/spiffe/spiffe.go
JoshVanL: I thought this would be preferred as it is being passed around a lot.

pkg/security/spiffe/spiffe.go
JoshVanL: We don't want to make `Parsed` mutable once it has been created is the reason for creating these functions.

pkg/actors/placement/grpc_opts.go
ItalyPaleAle: It is a fairly significant change actually. We use DialContext to connect to the gRPC server (placement), and the behavior of DialContext [changes](https://pkg.go.dev/google.golang.org/grpc#DialContext) depending on whether WithBlock is set or not:

> In the non-blocking case, the ctx does not act against the connection. It only controls the setup steps.
>
> In the blocking case, ctx can be used to cancel or expire the pending connection. Once this function returns, the cancellation and expiration of ctx will be noop. Users should call ClientConn.Close to terminate all the pending operations after this function returns. 

However, now that I see that... this `grpc.WithBlock` should be added to the options in all cases, and not just if monitoring is enabled (so should be outside of the if block)

JoshVanL:
@ItalyPaleAle Thanks for the review! I have implemented your comments or responded :slightly_smiling_face:

pkg/acl/acl.go
ItalyPaleAle: Ok. Also change `Debugf` to `Debug` if there's no formatting. Otherwise, `%` chars inside the error message are interpreted as formatting things

pkg/security/spiffe/spiffe.go
ItalyPaleAle: Ok

pkg/security/spiffe/spiffe.go
ItalyPaleAle: Ok

pkg/actors/placement/grpc_opts.go
JoshVanL: I added `grpc.WithBlock()` to the actor placement client by default.

mukundansundar:
> # Dapr Version Skew test (dapr-sidecar-master - 1.11.3)
> 🔗 **[Link to Action run](https://github.com/dapr/dapr/actions/runs/6132792434)**
> 
> Commit ref: [fb64dfc](https://github.com/dapr/dapr/commit/fb64dfcfe6e51840d9b1c018e2b84423a59e6c29)
> 
> ## ❌ Version Skew tests failed
> Please check the logs for details on the error.

There seems to be a panic seg fault in actor tests in this run.

pkg/acl/acl.go
```diff
@@ -265,37 +171,37 @@ func ApplyAccessControlPolicies(ctx context.Context, operation string, httpVerb
...
-	action, actionPolicy := IsOperationAllowedByAccessControlPolicy(spiffeID, appID, operation, httpVerb, isHTTP, acl)
-	emitACLMetrics(actionPolicy, appID, trustDomain, namespace, operation, httpVerb.String(), action)
+	action, actionPolicy := isOperationAllowedByAccessControlPolicy(spiffeID, operation, httpVerb, isHTTP, acl)
+	emitACLMetrics(spiffeID, actionPolicy, operation, httpVerb.String(), action)
 
 	if !action {
-		errMessage = fmt.Sprintf("access control policy has denied access to appid: %s operation: %s verb: %s", appID, operation, httpVerb)
-		log.Debugf(errMessage)
+		errMessage = fmt.Sprintf("access control policy has denied access to id: %s operation: %s verb: %s", spiffeID.URL(), operation, httpVerb)
+		log.Debug(errMessage)
```
daixiang0, line 179: Why L170 use `log.Debugf` but use `log.Debug` here?

ItalyPaleAle: `log.Debugf` is like fmt.Sprintf. In this case we are passing the exact message so we must use `log.Debug`.

daixiang0:
@JoshVanL thanks for great job! It is very important for security and I am happy to see it land in 1.12 release.

Please make CI happy.

JoshVanL:
@daixiang0 CI should be more happy now 🙂

artursouza (approved):

PR #6762, opened 2023-08-03
Hot reloading: Dynamic Channels

JoshVanL:
Branched from #2478 

In order to achieve [hot reloading](#1172), channel callers need to fetch the current channel dynamically. 

PR updates all callers to fetch the current channel type from `channels` during invocation to ensure they are using the latest configured.

/cc @yaron2

### DIFF
--- a/pkg/grpc/api_daprinternal.go
+++ b/pkg/grpc/api_daprinternal.go
@@ -35,7 +35,8 @@ import (
 
 // CallLocal is used for internal dapr to dapr calls. It is invoked by another Dapr instance with a request to the local app.
 func (a *api) CallLocal(ctx context.Context, in *internalv1pb.InternalInvokeRequest) (*internalv1pb.InternalInvokeResponse, error) {
-	if a.appChannel == nil {
+	appChannel := a.channels.AppChannel()
+	if appChannel == nil {
 		return nil, status.Error(codes.Internal, messages.ErrChannelNotFound)
 	}
 
@@ -60,7 +61,7 @@ func (a *api) CallLocal(ctx context.Context, in *internalv1pb.InternalInvokeRequ
 	}()
 
 	// stausCode will be read by the deferred method above
-	res, err := a.appChannel.InvokeMethod(ctx, req, "")
+	res, err := appChannel.InvokeMethod(ctx, req, "")
 	if err != nil {
 		statusCode = int32(codes.Internal)
 		return nil, status.Errorf(codes.Internal, messages.ErrChannelInvoke, err)
@@ -75,7 +76,8 @@ func (a *api) CallLocal(ctx context.Context, in *internalv1pb.InternalInvokeRequ
 // CallLocalStream is a variant of CallLocal that uses gRPC streams to send data in chunks, rather than in an unary RPC.
 // It is invoked by another Dapr instance with a request to the local app.
 func (a *api) CallLocalStream(stream internalv1pb.ServiceInvocation_CallLocalStreamServer) error { //nolint:nosnakecase
-	if a.appChannel == nil {
+	appChannel := a.channels.AppChannel()
+	if appChannel == nil {
 		return status.Error(codes.Internal, messages.ErrChannelNotFound)
 	}
 
@@ -187,7 +189,7 @@ func (a *api) CallLocalStream(stream internalv1pb.ServiceInvocation_CallLocalStr
 	}()
 
 	// Submit the request to the app
-	res, err := a.appChannel.InvokeMethod(ctx, req, "")
+	res, err := appChannel.InvokeMethod(ctx, req, "")
 	if err != nil {
 		statusCode = int32(codes.Internal)
 		return status.Errorf(codes.Internal, messages.ErrChannelInvoke, err)
--- a/pkg/grpc/manager/dial_test.go
+++ b/pkg/grpc/manager/dial_test.go
@@ -1,3 +1,16 @@
+/*
+Copyright 2023 The Dapr Authors
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
 package manager
 
 import (
--- a/pkg/grpc/manager/dial_test.go
+++ b/pkg/grpc/manager/dial_test.go
@@ -1,4 +1,4 @@
-package grpc
+package manager
 
 import (
 	"runtime"
--- a/pkg/http/api_directmessaging.go
+++ b/pkg/http/api_directmessaging.go
@@ -197,7 +197,7 @@ func (a *api) onDirectMessage(w http.ResponseWriter, r *http.Request) {
 // 1. HTTP header 'dapr-app-id' (path is method)
 // 2. Basic auth header: `http://dapr-app-id:<service-id>@localhost:3500/<method>`
 // 3. URL parameter: `http://localhost:3500/v1.0/invoke/<app-id>/method/<method>`
-func findTargetIDAndMethod(reqPath string, headers http.Header) (string, string) {
+func findTargetIDAndMethod(reqPath string, headers http.Header) (targetID string, method string) {
 	if appID := headers.Get(daprAppID); appID != "" {
 		return appID, strings.TrimPrefix(path.Clean(reqPath), "/")
 	}
--- a/pkg/http/api_directmessaging.go
+++ b/pkg/http/api_directmessaging.go
@@ -197,7 +197,7 @@ func (a *api) onDirectMessage(w http.ResponseWriter, r *http.Request) {
 // 1. HTTP header 'dapr-app-id' (path is method)
 // 2. Basic auth header: `http://dapr-app-id:<service-id>@localhost:3500/<method>`
 // 3. URL parameter: `http://localhost:3500/v1.0/invoke/<app-id>/method/<method>`
-func findTargetIDAndMethod(reqPath string, headers http.Header) (targetID string, method string) {
+func findTargetIDAndMethod(reqPath string, headers http.Header) (string, string) {
 	if appID := headers.Get(daprAppID); appID != "" {
 		return appID, strings.TrimPrefix(path.Clean(reqPath), "/")
 	}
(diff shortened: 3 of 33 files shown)


pkg/actors/internal/placement_service.go
```diff
@@ -22,7 +22,7 @@ import (
 type PlacementService interface {
 	io.Closer
 
-	Start(ctx context.Context) error
+	Start(context.Context)
```
ItalyPaleAle, line 25: Please leave the error. This was added recently and it helps with future expansion

```suggestion
	Start(ctx context.Context) error
```

pkg/actors/placement/client.go
```diff
@@ -44,43 +44,34 @@ type placementClient struct {
...
 // stream with the connected stream.
-func (c *placementClient) connectToServer(serverAddr string) error {
+func (c *placementClient) connectToServer(ctx context.Context, serverAddr string) error {
 	opts, err := c.getGrpcOpts()
 	if err != nil {
 		return err
 	}
 
-	conn, err := grpc.Dial(serverAddr, opts...)
+	conn, err := grpc.DialContext(ctx, serverAddr, opts...)
```
ItalyPaleAle, line 57: If you use DialContext, it's important that you add the `WithBlock` option, since DialContext is non-blocking: https://pkg.go.dev/google.golang.org/grpc#DialContext

pkg/actors/placement/placement.go
```diff
@@ -142,12 +142,12 @@ func (p *actorPlacement) AddHostedActorType(actorType string) error {
 
 // Start connects placement service to register to membership and send heartbeat
 // to report the current member status periodically.
-func (p *actorPlacement) Start(ctx context.Context) error {
+func (p *actorPlacement) Start(ctx context.Context) {
```
ItalyPaleAle, line 145: Let's keep the errors here so the interface can return errors. It was added recently to help with future expansion.


```suggestion
func (p *actorPlacement) Start(ctx context.Context) error {
```

pkg/actors/placement/placement.go
```diff
@@ -142,12 +142,12 @@ func (p *actorPlacement) AddHostedActorType(actorType string) error {
...
 	p.serverIndex.Store(0)
 	p.shutdown.Store(false)
 
-	if !p.establishStreamConn() {
-		return nil
+	if !p.establishStreamConn(ctx) {
+		return
```
ItalyPaleAle, line 150:
```suggestion
		return nil
```

pkg/actors/placement/placement.go
```diff
@@ -241,8 +241,6 @@ func (p *actorPlacement) Start(ctx context.Context) error {
 			}
 		}
 	}()
-
-	return nil
 }
```
ItalyPaleAle, line 244:
```suggestion
	}()

    return nil
}
```

pkg/actors/placement/placement_test.go
```diff
@@ -137,7 +137,7 @@ func TestAppHealthyStatus(t *testing.T) {
 	}).(*actorPlacement)
 
 	// act
-	require.NoError(t, testPlacement.Start(context.Background()))
+	testPlacement.Start(context.Background())
```
ItalyPaleAle, line 140:
```suggestion
	require.NoError(t, testPlacement.Start(context.Background()))
```

pkg/actors/placement/placement_test.go
```diff
@@ -84,7 +84,7 @@ func TestPlacementStream_RoundRobin(t *testing.T) {
 		testSrv[leaderServer[0]].setLeader(true)
 
 		// act
-		require.NoError(t, testPlacement.Start(context.Background()))
+		testPlacement.Start(context.Background())
```
ItalyPaleAle, line 87:
```suggestion
		require.NoError(t, testPlacement.Start(context.Background()))
```

pkg/actors/actors.go
```diff
@@ -77,8 +77,8 @@ var (
 //nolint:interfacebloat
 type Actors interface {
 	Call(ctx context.Context, req *invokev1.InvokeMethodRequest) (*invokev1.InvokeMethodResponse, error)
-	Init() error
-	Stop()
+	Init(context.Context) error
+	Stop() error
```
ItalyPaleAle, line 81: Sae, please keep errors here

```suggestion
	Stop() error
```

pkg/actors/actors.go
```diff
@@ -239,13 +241,20 @@ func (a *actorsRuntime) Init() error {
 			},
 			AfterTableUpdateFn: func() {
 				a.drainRebalancedActors()
-				a.actorsReminders.OnPlacementTablesUpdated(context.TODO())
+				a.actorsReminders.OnPlacementTablesUpdated(ctx)
```
ItalyPaleAle, line 244: This cannot be `ctx` and must be a background context, because this callback is invoked periodically in background by the placement service (when placement tables are updated)

For now, maybe context.TODO()?


```suggestion
				a.actorsReminders.OnPlacementTablesUpdated(context.TODO())
```

ItalyPaleAle: Oh, looks like the context is not just for init but for the entire lifecycle of actors? Please confirm if that's the case

pkg/actors/actors_test.go
```diff
@@ -966,7 +966,7 @@ func TestActorsAppHealthCheck(t *testing.T) {
 	clock := testActorsRuntime.clock.(*clocktesting.FakeClock)
 
 	testActorsRuntime.actorsConfig.Config.HostedActorTypes = internal.NewHostedActors([]string{"actor1"})
-	go testActorsRuntime.startAppHealthCheck(
+	go testActorsRuntime.startAppHealthCheck(context.Background(),
```
ItalyPaleAle, line 969: Should this context be canceled when tests end?

pkg/grpc/server.go
```diff
@@ -179,14 +187,19 @@ func (s *server) StartNonBlocking() error {
...
 	for _, server := range s.servers {
 		// This calls `Close()` on the underlying listener.
-		server.GracefulStop()
+		go server.GracefulStop()
 	}
 
 	if s.api != nil {
 		if closer, ok := s.api.(io.Closer); ok {
-			closer.Close()
+			return closer.Close()
```
ItalyPaleAle, line 202: I would add an error check like "if err != nil"...

I can see a day when someone adds something to the Close() method after this, and doesn't realize that there's a `return` above.

pkg/grpc/server.go
```diff
@@ -313,22 +330,28 @@ func (s *server) startWorkloadCertRotation() {
 	s.logger.Infof("starting workload cert expiry watcher. current cert expires on: %s", s.signedCert.Expiry.String())
 
 	ticker := time.NewTicker(certWatchInterval)
```
ItalyPaleAle, line 332: Unrelated from this PR but why are we checking if the cert is expired on a regular interval? We know when the cert expires, so we should simply wait until that needs to be renewed, no? That avoids getting locks for no reason.

pkg/http/api_directmessaging.go
```diff
@@ -198,7 +198,7 @@ func (a *api) onDirectMessage(w http.ResponseWriter, r *http.Request) {
 // 1. HTTP header 'dapr-app-id' (path is method)
 // 2. Basic auth header: `http://dapr-app-id:<service-id>@localhost:3500/<method>`
 // 3. URL parameter: `http://localhost:3500/v1.0/invoke/<app-id>/method/<method>`
-func findTargetIDAndMethod(reqPath string, headers http.Header) (targetID string, method string) {
+func findTargetIDAndMethod(reqPath string, headers http.Header) (string, string) {
```
ItalyPaleAle, line 201: Why was this changed? It seems functionally equivalent in all ways?

pkg/runtime/wfengine/wfengine_test.go
```diff
@@ -64,8 +64,7 @@ func (*mockPlacement) LookupActor(actorType string, actorID string) (name string
 }
 
 // Start implements internal.PlacementService
-func (*mockPlacement) Start(ctx context.Context) error {
-	return nil
+func (*mockPlacement) Start(context.Context) {
```
ItalyPaleAle, line 67:
```suggestion
func (*mockPlacement) Start(ctx context.Context) error {
	return nil
```

pkg/runtime/runtime.go
```diff
@@ -1597,28 +1641,22 @@ func (a *DaprRuntime) authSecretStoreOrDefault(resource meta.Resource) string {
...
-			return
+		// Return
 		case <-ctx.Done():
-			// Return
-			return
-		default:
-			// nop - continue execution
+			return ctx.Err()
+		// prevents overwhelming the OS with open connections
+		case <-time.After(time.Millisecond * 100):
```
ItalyPaleAle, line 1659: The difference is that now the first attempt is delayed 100ms. Maybe ove this `select` at the end of the loop?

ItalyPaleAle (changes requested):

pkg/actors/actors.go
JoshVanL: That's right- the context coming into `Init` will be the runner manager context which originates from the base signal context.

pkg/grpc/server.go
JoshVanL: Yes, good shout

pkg/http/api_directmessaging.go
JoshVanL: We should avoid named return variables at all costs (imo), and _especially_ never do naked returns. The linter was complaining to me about the naked returns.

pkg/runtime/runtime.go
JoshVanL: :+1:

pkg/grpc/server.go
ItalyPaleAle: Yeah it was just an unrelated thing. Good that it's getting ripped anyways

JoshVanL:
@ItalyPaleAle I have address your comments in the [base PR](https://github.com/dapr/dapr/pull/6720)

ItalyPaleAle:
Thank you!

pkg/grpc/manager/dial_test.go
```diff
@@ -1,4 +1,4 @@
-package grpc
+package manager
```
ItalyPaleAle, line 1: Missing copyright header

pkg/http/api_directmessaging.go
ItalyPaleAle: Ok on not using naked returns (agree). I do like named returns here because they're both strings and using named returns helps documenting the signature of the method.

ItalyPaleAle (changes requested):

pkg/grpc/manager/dial_test.go
JoshVanL: :+1:

pkg/grpc/api_daprinternal.go
```diff
@@ -35,7 +35,8 @@ import (
 
 // CallLocal is used for internal dapr to dapr calls. It is invoked by another Dapr instance with a request to the local app.
 func (a *api) CallLocal(ctx context.Context, in *internalv1pb.InternalInvokeRequest) (*internalv1pb.InternalInvokeResponse, error) {
-	if a.appChannel == nil {
+	appChannel := a.channels.AppChannel()
```
yaron2, line 38: Is `a.channels` always guaranteed to be non-empty? Question also applies to other places that make use of it

JoshVanL: Correct, `channels` will always be non-nil (passed through `Options` or equivalent`), however the underlying channel returned from `AppChannel()` may or may not be nil.

JoshVanL: We create it as one of the first things we do when creating the `DaprRuntime` https://github.com/dapr/dapr/pull/6762/files#diff-e56f9bb947923c6f8bb4c995bf4b9469a0ed117276223d16b177dcb04399a4bcR183

yaron2 (approved):

PR #6766, opened 2023-08-04
Do not panic when daprd receives unknown CLI flags

ItalyPaleAle:
Changes how CLI flags are parsed in daprd to avoid panicking when a flag that is undefined is passed.

This is needed to allow smoother upgrades in the future. For reference, see conversation here where we would need to add new CLI flags passed by the injector, but which would cause older versions of the sidecar to crash: https://github.com/dapr/dapr/pull/6761#discussion_r1282581479

Although this change is applied to master only and it's not retroactive, so the problem highlighted in the conversation above can't be solved (at least not for 2 more Dapr releases), it will allow making these changes safer in the future.

Flags that are unknown will cause a warning to be printed, but will not cause a panic.

This change is possible by switching the CLI flag parser to spf13/pflag, which also comes with some additional benefits such as being able to hide deprecated flags from the help message (we have 2 at the moment), or having built-in support for multiple flags (like for `--config` and `--resources-path`).

This PR includes unit tests for the added feature and for other "tricky" parts of the package.

> This change is applied to daprd only. I don't think it's necessary to apply it to sentry, operator, injector, placement. This is because these are all deployed via Helm and flags are controlled by the Helm chart for the given version.


JoshVanL:
As mentioned in my [comment](https://github.com/dapr/dapr/pull/6761#discussion_r1284451987), I don't think not exiting on error is a good idea. It's not standard and is going to causing UX issues. An operator must know that the CLI flags passed to a program are actually being picked up.

We can use environment variables for introducing new options until all supported versions implement the flag.

ItalyPaleAle:
> As mentioned in my [comment](https://github.com/dapr/dapr/pull/6761#discussion_r1284451987), I don't think not exiting on error is a good idea. It's not standard and is going to causing UX issues. An operator must know that the CLI flags passed to a program are actually being picked up.
> 
> We can use environment variables for introducing new options until all supported versions implement the flag.

You're right on the fact that panicking makes it easier to see typos. This PR shows a large warning at the beginning however which should help a bit.

But if the alternative is to use env vars, that's a much bigger evil:

1. They are in the environment so users set them with `export` in the shell, forget about them, and then see the app behaving incorrectly. Env vars are like global vars in code, and developers are taught to avoid them :)
2. There's no way to detect typos at all; at least with CLI flags we can show a warning.
3. There's no autocompletion in the shell.
4. Env vars do not appear in the `--help` page and are less discoverable.
5. Users will be confused by what is set via env vars and what is set via CLI flags
6. Our docs will be more confusing. The [table](https://docs.dapr.io/reference/arguments-annotations-overview/) already has many columns: `dapr run` flags, `daprd` flags, annotation. Adding a 4th column will not help.

JoshVanL:
> They are in the environment so users set them with export in the shell, forget about them, and then see the app behaving incorrectly. Env vars are like global vars in code, and developers are taught to avoid them :)

These are only meant to be used in kube where we need to support new flags for backwards compat. We are already using environment variable extensively.

> There's no way to detect typos at all; at least with CLI flags we can show a warning.

Warning do not give immediate signals in large deployments and are effectively silent until an outage happens and logs are investigated.

> There's no autocompletion in the shell.

We still have auto-completion of the flags. The point here is to have both the environment variable _and_ flags be available.

> Env vars do not appear in the --help page and are less discoverable.

^ flags do as above.

> Users will be confused by what is set via env vars and what is set via CLI flags

Flags take precedence over environment variables. This is typical  behavior of all programs.

ItalyPaleAle:
> These are only meant to be used in kube where we need to support new flags for backwards compat. We are already using environment variable extensively.

There's standalone mode too for Daprd.

> > Users will be confused by what is set via env vars and what is set via CLI flags
>
> Flags take precedence over environment variables. This is typical behavior of all programs.

What I meant was, users will need to remember which option can be set via env vars and which via flags. That is confusing.

ItalyPaleAle:
This is currently blocked due to some weird stuff done by the Kubernetes components in components-contrib. See: https://github.com/dapr/components-contrib/pull/2992#issuecomment-1638552086

artursouza:
Can we discuss this in the next maintainer's call? This is a breaking change.

ItalyPaleAle:
> Can we discuss this in the next maintainer's call?

Yes.

> This is a breaking change.

How so?

berndverst (approved):
This change looks good to me. `pflag` is a better library for parameter parsing. This change here is fully backwards compatible.

The question is whether `daprd` when given an unknown flag (assume this is in addition to required flags) should launch (with warning) or exit / panic.

I too prefer the former option (which is what this PR is about). It provides a lot more flexibility for a greater combination of CLI and daprd versions, throughout upgrade/downgrades and testing.

I am approving this PR - but of course it's up to the runtime maintainers to decide whether to merge or provide additional review.

yaron2:
Using pflags as a library is an ok change. @berndverst you're correct in assessing the question is about how Dapr behaves with unknown flags. My opinion is that allowing unknown flags to pass without validation is an anti pattern in most software and just an overall bad user experience for Dapr, which is why I don't think we should accept this change as-is. The tooling around Dapr is diverse - CLI, multi app run, Docker Compose, the sidecar injector and any form of way to run Dapr processes. When users provide a wrongly configured flag today, they would immediately receive feedback before Dapr starts. This lends itself  naturally to most testing validations, including the ones we use for the sidecar injector. With this change, users are left to encounter errors during runtime, or even worse - Dapr continues to operate without the desired behavior silently and never informs the user that something is wrong. In addition, this opens the door to many injector bugs related to malformed flag names, that cannot be caught with e2e, leading to greater chances of a hotfixes.

artursouza:
I agree with Yaron here. Every cmd tool I use fails with invalid commands, moving that to a runtime issue is a problem. You are just moving the problem down the road. So, people running Dapr will have the illusion a given flag is being considered while it is not. I think the problem of upgrading and handling new params is a legit concern but we can fix it in other ways. One is to have ENV variables for the optional params.

ItalyPaleAle:
1. Most users don't run `daprd` directly, but use the Dapr CLI or the injector. So they don't risk making typos.
2. Env vars have literally the same issues you've described above, such as inability to catch typos etc, with the additional ones of being global, defined in the environment, and undocumented when running `--help`. And usually defined in places scattered all over the codebase. See: https://github.com/dapr/dapr/pull/6766#issuecomment-1665714022
3. The reason for this change is not because pflag is nicer (albeit it is), but because this boxes us adding required flags in the future.

Anyways, you two will decide what you want as usual.

Closing

berndverst:
I don't think `daprd` is really a command line tool - yes it can be used that way, but the command line tool is the Dapr CLI. `daprd` is only called by CLI, of course our sidecar injector, and by a few select very advanced users. In the CLI I would agree with erroring with unknown flags. In `daprd` I do not agree erroring is necessary, and it makes things harder.

I will also state that ENV vars are a much more error prone pattern.

Fortunately as I am not a runtime or CLI maintainer I do not have to deal with the CLI / runtime upgrade downgrade mess. I just advocated for a change that I believe to be good for the project and evaluate its merit without ganging up on the contributor.

PR #6769, opened 2023-08-04
Stop SubscribeConfiguration when client disconnects to prevent leaks

### LINKED ISSUE #6560: Configuration API: anti-pattern behavior for gRPC APIs can lead to leaks
## In what area(s)?

/area runtime

## What version of Dapr?

> 1.11
> master

## Description

The Configuration APIs use 2 methods for subscribing to configuration changes, for both HTTP and gRPC:

- `Subscribe` creates a new subscription, returning the subscription ID
- `Unsubscribe` stops the subscription (which usually involves doing work on the component too)

With HTTP, configuration changes are sent using the app channel to invoke callbacks, so this is fine.

With gRPC, however, the `Subscribe` method is a streaming RPC that sends messages to the caller as soon as they come in. The requirement of calling `Unsubscribe` in this scenario is anti-pattern and has some unintended consequences:

1. When using server-stream gRPC, it's assumed that clients are the ones that control how long they're interested in receiving the data for, and they can communicate their lack of interest by closing the RPC.
   - Servers are not supposed to close the RPC unless they do not have more data to send
   - This pattern is also what enables proper resiliency: if the connection between Dapr and the app is temporarily severed, the client should simply re-instantiate the RPC to resume receiving data.
2. In this case, if the client closes the RPC, but doesn't call `Unsubscribe`, the handler doesn't return and causes goroutine leaks:
   - This is because the only way to make the `Subscribe` handler to return is to [close the channel](https://github.com/dapr/dapr/blob/ca08aa97deebf93306a66af7c31c9b1309f7a9b3/pkg/grpc/api.go#L1389) like `Unsubscribe` does (and that's the only way)
   - If new messages were to come in after the client has disconnected, the runtime [still tries to send them](https://github.com/dapr/dapr/blob/ca08aa97deebf93306a66af7c31c9b1309f7a9b3/pkg/grpc/api.go#L1314), even though the channel is closed, so flooding the logs with errors

Although fixing the goroutine leak is simple (just needs a `select` that unblocks when `stop` is closed OR when the stream's context is done), an actual fix would also involve deprecating the `Unsubscribe` method and allowing clients using gRPC to stop subscriptions by simply closing the RPC channel.

ItalyPaleAle:
Fixes #6560

Clients can continue to call UnsubscribeConfiguration for backwards-compatibility. However, doing so will show a warning encouraging users to simply close the connection to the SubscribeConfiguration RPC.

### DIFF
--- a/pkg/grpc/api.go
+++ b/pkg/grpc/api.go
@@ -18,13 +18,13 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"sort"
 	"strconv"
 	"sync"
 	"sync/atomic"
 	"time"
 
 	otelTrace "go.opentelemetry.io/otel/trace"
+	"golang.org/x/exp/slices"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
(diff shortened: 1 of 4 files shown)


JoshVanL (approved):

mukundansundar (approved):

ItalyPaleAle:
> @ItalyPaleAle Let's call out deprecation of gRPC unsubscribe API in the release notes and then remove it later.

Sounds good! I'll add the "Breaking change" label here so we don't forget (although this is not a breaking change)

PR #6772, opened 2023-08-04
Return error on failing grpc app health probe

yaron2:
This PR changes the gRPC probe to return the error for logging instead of silently dropping it.

Additional context: I encountered a situation where a large online retailer experienced production downtime due to a version of their app deployed to production not implementing the HealthCheck method in AppCallBack. Hours were spent trying to figure out the root cause, and the only way to find out the problem was for them to build a custom version of Dapr that prints the error. The error was not a network error, but a gRPC Not Implemented error.

### DIFF
--- a/pkg/apphealth/health.go
+++ b/pkg/apphealth/health.go
@@ -197,6 +197,7 @@ func (h *AppHealth) doProbe(parentCtx context.Context) {
 	successful, err := h.probeFn(ctx)
 	// In case of errors, we do not record the failed probe because this is generally an internal error
 	if err != nil {
+		h.setResult(parentCtx, false)
 		log.Errorf("App health probe could not complete with error: %v", err)
 		return
 	}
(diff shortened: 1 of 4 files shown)


JoshVanL (dismissed):

pkg/apphealth/health.go
```diff
@@ -197,6 +197,7 @@ func (h *AppHealth) doProbe(parentCtx context.Context) {
 	successful, err := h.probeFn(ctx)
 	// In case of errors, we do not record the failed probe because this is generally an internal error
```
JoshVanL, line 198: This comment needs to be updated (or removed).

JoshVanL (approved):

pkg/apphealth/health.go
yaron2: Removed

Issue #6734, opened 2023-07-27
Dapr GRPC proxy connection leak


jsheetzmt:
## In what area(s)?

> /area runtime 
## What version of Dapr?

> 1.1.x

## Expected Behavior

Proxying GRPC calls through dapr sidecar doesn't open many connections to destination app during load

## Actual Behavior

During load, dapr sidecar is opening many connections to the destination app causing high TCP usage and memory leak in our .NET Core application. When I remove the dapr sidecar and make a direct connection between the server and client, I only have a single connection open on port 8081 (HTTP2 traffic).

**dotnet-monitor**
![image](https://github.com/dapr/dapr/assets/103272899/85125f6e-f3ba-410a-bb29-dd7216e363b5)

**ss**
![image](https://github.com/dapr/dapr/assets/103272899/744477ea-8781-41a4-bf81-b40c27022d5b)

## Steps to Reproduce the Problem

Proxy GRPC calls through dapr sidecar with high load

## Possible related issue
https://github.com/dapr/dapr/issues/4937

yaron2:
So to be clear, in the following call chain: App 1 --> Dapr 1 -------> Dapr 2 --> App 2

You see connections being opened and not re-used in the path of Dapr 2 --> App 2?

jsheetzmt:
Yep, that is exactly right @yaron2 but there is some connection reuse between Dapr 2 --> App 2. If I am doing a single request at a time in something like Postman, connection count stays at 1. As soon as I throw JMeter with 10 threads at it, connection count grows steadily, but it doesn't increase by 1 for every request.

yaron2:
Does the connection count continue growing after you've finished opening the 10 threads with JMeter?

jsheetzmt:
It stops growing but the connections remain open.

![image](https://github.com/dapr/dapr/assets/103272899/5384f918-2a8b-4577-a974-d18c44be5aca)

yaron2:
Ok thanks a lot of the info, will check this out. Do you have any special connection max age/timeout settings on App 2's server?

jsheetzmt:
I have this configured to keep alive the HTTP2 connection so it doesn't die while idle. I tried without this logic and still saw the same behavior with dapr sidecar --> App 2

```
private static readonly SocketsHttpHandler _handler = new SocketsHttpHandler
{
    PooledConnectionIdleTimeout = Timeout.InfiniteTimeSpan,
[...]

PR #6774, opened 2023-08-04
Fix connection leak with app channel

yaron2:
Fixes https://github.com/dapr/dapr/issues/6734.

### DIFF
--- a/pkg/grpc/grpc.go
+++ b/pkg/grpc/grpc.go
@@ -58,14 +58,13 @@ type AppChannelConfig struct {
 
 // Manager is a wrapper around gRPC connection pooling.
 type Manager struct {
-	remoteConns       *RemoteConnectionPool
-	auth              security.Authenticator
-	mode              modes.DaprMode
-	channelConfig     *AppChannelConfig
-	localConn         *ConnectionPool
-	localConnCreateFn ConnCreatorFn
-	localConnLock     sync.RWMutex
-	appClientConn     grpc.ClientConnInterface
+	remoteConns   *RemoteConnectionPool
+	auth          security.Authenticator
+	mode          modes.DaprMode
+	channelConfig *AppChannelConfig
+	localConn     *ConnectionPool
+	localConnLock sync.RWMutex
+	appClientConn grpc.ClientConnInterface
 }
 
 // NewGRPCManager returns a new grpc manager.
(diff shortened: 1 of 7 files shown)


pkg/grpc/grpc.go
```diff
@@ -108,31 +109,21 @@ func (g *Manager) GetAppChannel() (channel.AppChannel, error) {
 // GetAppClient returns the gRPC connection to the local app.
 // If there's no active connection to the app, it creates one.
 func (g *Manager) GetAppClient() (grpc.ClientConnInterface, error) {
-	if g.localConnCreateFn != nil {
-		return g.localConn.Get(g.localConnCreateFn)
-	}
-	return g.localConn.Get(g.defaultLocalConnCreateFn)
-}
+	if g.appClientConn == nil {
```
yaron2, line 112: This section of the code cannot have concurrent access, thus no usage of locks.

ItalyPaleAle: I think the reason for the lock was because otherwise `go test -race` would fail otherwise, either here or in the runtime. Can you please confirm that both this package (pkg/grpc) and pkg/runtime pass with `go test -race` without the lock?

ItalyPaleAle: I do see this method invoked in multiple places, so a lock may be needed to avoid race conditions

pkg/grpc/grpc.go
```diff
@@ -108,31 +109,21 @@ func (g *Manager) GetAppChannel() (channel.AppChannel, error) {
...
+		c, err := g.defaultLocalConnCreateFn()
```
yaron2, line 113: Its this line that cannot be invoked concurrently, as it's called once on init and any subsequent calls will not satisfy the nil check. @ItalyPaleAle

pkg/grpc/grpc.go
yaron2: See comment below

pkg/grpc/grpc.go
ItalyPaleAle: But if 2 goroutines call `GetAppClient` at the exact same time, `g.appClient` will be nil for both, so both will get here and both will try to write  `g.appClientConn = c`?

yaron2: When these goroutines get there, it will not be nil. Before any code that can reach this line is executed, it would already have been initialized during init in runtime

ItalyPaleAle: Ok, I see

ItalyPaleAle:
Please fix the linter issue and then LGTM

ItalyPaleAle (approved):

PR #6775, opened 2023-08-04
Flaky unit tests: fix many race conditions

ItalyPaleAle:
Also increases the timeout on resiliency tests which seem to sometimes take longer on our Actions runners

Should reduce flakiness in our unit tests.

> Every package should now pass tests with `go test -race` 🎉


pkg/actors/internal/config.go
```diff
@@ -61,30 +62,41 @@ func (c *Config) GetRemindersPartitionCountForType(actorType string) int {
 	return c.RemindersStoragePartitions
 }
 
-type hostedActors map[string]struct{}
+type hostedActors struct {
+	actors map[string]struct{}
+	lock   sync.Mutex
```
JoshVanL, line 67: Consider using a read write mutex

JoshVanL (changes requested):

pkg/actors/internal/config.go
ItalyPaleAle: Done

JoshVanL (approved):

mukundansundar (approved):

Issue #6784, opened 2023-08-08
Add soft memory limit to daprd sidecar


filintod:
## In what area(s)?

> /area runtime

This is mainly in the injector to allow adding soft memory limit via GOMEMLIMIT to sidecar to avoid potential oom kill

## Describe the feature

As Go now [has soft memory limit](https://pkg.go.dev/runtime) it will be good to have an option to add to the sidecar to avoid oom. 

## Release Note

RELEASE NOTE:

filintod:
The Fix will be mostly a documentation entry instead of a PR.

filintod:
Docs added

PR #6785, opened 2023-08-08
Add soft memory limit via GOMEMLIMIT to daprd sidecar

filintod:
# Description

Allow to set soft memory limit to daprd sidecar via a couple of annotations:
```
	KeySoftMemoryLimit                  = "dapr.io/sidecar-soft-memory-limit"
	KeySoftMemoryLimitPercentage        = "dapr.io/sidecar-soft-memory-limit-percentage"
```

It follows certain rules:
- If no user limit set:
  - and no soft limit set, there is no change to current behavior
  - and soft limit set, use soft limit
  - and soft limit percentage, no change to current behavior
- if user limit set:
  -  and no soft limit set, we set a default soft limit equal to 90% of the limit set by the user following [Go GC doc recommendations](https://tip.golang.org/doc/gc-guide#Suggested_uses).
  - and soft limit set, set it to the soft limit as long as it is between 25 and 90% of the limit set, if outside raise an error (this could be changed to not raise an error and just hard limit the values to these ranges)
  - and soft limit percentage set, set it to perc * hard limit. If perc outside of 25-90% range we raise an error (again we can change this behavior to cut at these values instead of raising error)

When soft limit is set depending on the rules above, we add the environment variable GOMEMLIMIT with value equal to the defined soft limit or percentage soft limit.  We try to use provided soft limit from user if possible but sometimes we need to change to a format allowed by GOMEMLIMIT (for example if they use 100Mi we need to use 100MiB).

## Issue reference

Please reference the issue this PR will close: #6784

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Unit tests passing


pkg/injector/patcher/sidecar.go
```diff
@@ -25,6 +25,12 @@ import (
 	"github.com/dapr/dapr/utils"
 )
 
+const (
+	defaultSoftLimitPercentage    = 80
```
artursouza, line 29: Shouldn't the default just be no soft limit?

artursouza (changes requested):

pkg/injector/patcher/sidecar.go
filintod: it could be, just wanted to have some best practice to avoid OOM kill. I've increased it to 90% to follow recs in [golang gc doc](https://tip.golang.org/doc/gc-guide#Memory_limit).  This is only applied if there is a memory limit set by the user if there is no limit assigned by the user there is no percentage limit.

filintod:
@yaron2 just pointed me to a potential simpler solution by reusing an annotation I did not know about nicely called `env` https://docs.dapr.io/reference/arguments-annotations-overview/.

I'll check and test locally and if everything checks out instead of this PR just create a documentation on how to add soft limits to the sidecar using that.

filintod:
I'll be closing this in favor of adding documentation to handle this.

Issue #6454, opened 2023-06-07
[Workflow] Activity execution gets stuck when parent workflow is terminated


cgillum:
## In what area(s)?

> /area runtime

## What version of Dapr?

> 1.11

## Description

As part of investigating https://github.com/dapr/dapr/issues/6435, an issue was discovered where an activity execution is orphaned with the following warning message logged repeatedly:

```
WARN[1598] TestWorkflow_a7f10a37-54ec-43ae-89c2-60630765e7da::1::2: 'run-activity' is still running - will keep waiting until XXX scope=dapr.runtime.wfengine type=log ver=edge
```

The issue seems to occur after the following error is observed in the SDK:

```
== APP == fail: Microsoft.DurableTask[20]
== APP ==       Unexpected error in handling of instance ID 'TestWorkflow_a7f10a37-54ec-43ae-89c2-60630765e7da'.
== APP ==       Grpc.Core.RpcException: Status(StatusCode="Unknown", Detail="unknown instance ID/task ID combo: TestWorkflow_a7f10a37-54ec-43ae-89c2-60630765e7da/1")
== APP ==          at Microsoft.DurableTask.Worker.Grpc.GrpcDurableTaskWorker.Processor.OnRunActivityAsync(ActivityRequest request)
== APP ==          at Microsoft.DurableTask.Worker.Grpc.GrpcDurableTaskWorker.Processor.<>c__DisplayClass12_0.<<RunBackgroundTask>b__0>d.MoveNext()
```

The problem is that the `durabletask-go` engine doesn't work with concurrent, duplicate activity executions. As a result, the Dapr Workflow activity actor is never notified that the duplicate activity invocations complete. This results in the repeated warning.

## Impact

This should recover automatically after an hour because Dapr Workflow stops waiting for activity completion after 1 hour. During this time, the above warning will be repeated. There may also be a memory impact during this hour since the activity actor will remain active during this time.

## Expected Behavior

Results for duplicate activity executions should be discarded with a single warning clearly indicating that an execution was abandoned.

## Actual Behavior

Duplicate/abandoned activity executions are not detected and are not correctly handled, resulting in repeated warnings.

## Steps to Reproduce the Problem

1. Start a workflow that starts an activity that delays for 30 seconds
2. Terminate the workflow after waiting for about 10 seconds
3. Start a new instance of the workflow using the same ID

## Release Note

RELEASE NOTE: **FIX** Repeated warnings about abandoned workflow activities executions not completing

PR #6786, opened 2023-08-08
Add Pending Activity check

DeepanshuA:
# Description

This PR addresses issue occurring while re-scheduling a workflow with same instance id, while previous run was terminated but still an activity left un-finished.

I have tried to enlist the detailed description of issue and thought process behind fix in following comment: https://github.com/dapr/dapr/issues/6454#issuecomment-1670041771

Another workflow with same instance Id is allowed to Run.
That's why the concept of Generation in there. And also, in https://github.com/dapr/dapr/pull/6455#issue-1745212239 , @cgillum specifically talks about it, so that means that it is Allowed.

To keep behavior consistent, I strongly feel that thus any Workflow terminated once, should be allowed to be scheduled with same id again.
It is just that the definition of Complete needs to be defined correctly i.e. Workflow with same ID should not be allowed just because previous Workflow run was Terminated. It should be allowed only if No Activity is remaining to be returned back as Completed or Failed, in addition to Workflow Completed.

So, let's analyze following 2 cases:

Case 1:
Scheduled a workflow W1 with activities A1, A2, A3 and A4. Let's Assume, workflow id is ID1.
In A2, put a sleep of 30 seconds.
After 10 seconds of scheduling workflow, Terminate it.
And, schedule this workflow again with same id ID1.

Here, re-scheduling of this workflow should NOT be allowed as A2 has not yet returned, even though Workflow itself has been Terminated.

Case 2:
Scheduled a workflow W1 with activities A1, A2, A3 and A4. Let's Assume, workflow id is ID1.
In A2, put a sleep of 30 seconds.
After 10 seconds of scheduling workflow, Terminate it.
Wait for 25 more seconds. By this time, A2 has also Completed. A3 and A4 will not be called.
At this time, schedule this workflow again with same id ID1.

Here, re-scheduling of this workflow SHOULD be allowed as A2 has returned, and Workflow itself has been truly Terminated and Completed.

## Issue reference

Please reference the issue this PR will close: #6454 

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing


pkg/runtime/wfengine/workflow.go
```diff
@@ -175,12 +176,17 @@ func (wf *workflowActor) createWorkflowInstance(ctx context.Context, actorID str
 		return errors.New("invalid execution start event")
 	}
 
-	// We block (re)creation of existing workflows unless they are in a completed state.
+	// We block (re)creation of existing workflows unless they are in a completed state
+	// Or if they still have any pending activity result awaited.
 	if !created {
```
mukundansundar, line 181: change flow as such:
```go
  if created {
     return error
  }
  // unnest the rest 
  runtimeState := getRuntimeState(actorID, state)
  if !runtimeState.IsCompleted() {
     return error
  }
  if wf.activityResultAwaited {
     return fmt.Errorf("a terminated workflow with ID '%s' is already awaiting an activity result", actorID)
  }
  // no else needed
  wfLogger.Infof("%s: workflow was previously completed and is being recreated", actorID)
  state.Reset()
```

pkg/runtime/wfengine/workflow.go
```diff
@@ -440,6 +449,8 @@ func (wf *workflowActor) runWorkflow(ctx context.Context, actorID string, remind
 				WithContentType(invokev1.OctetStreamContentType)
 			defer req.Close()
 
+			wf.activityResultAwaited = true
```
mukundansundar, line 452: will this state be accessed by multiple goroutines? Does this need to be an atomic boolean?

pkg/runtime/wfengine/workflow.go
DeepanshuA: Could not take the first check from suggestion:
```
  if created {
     return error
  }
```
I have taken the later part of suggestion, that sounds good.

mukundansundar (approved):
lgtm

PR #6791, opened 2023-08-09
Sentry Kubernetes Validator ensures control-plane clients have the control plane annotation set.

JoshVanL:
Introduces a new `dapr.io/control-plane` annotation which must be set on the control plane service's Pod. Sentry uses this annotation as part verifying a control plane client's identity.

This annotation cannot be used with `dapr.io/app-id`, and sentry only accepts the set of well-known control plane services.

This is being added so as to not have to use the `dapr.io/app-id` annotation on control plane services which would overload its meaning.

/cc @yaron2 @artursouza


pkg/sentry/server/validator/kubernetes/kubernetes.go
```diff
@@ -209,6 +212,37 @@ func (k *kubernetes) Validate(ctx context.Context, req *sentryv1pb.SignCertifica
...
+	ctrlPlane, ctrlOK := pod.Annotations[consts.AnnotationKeyControlPlane]
+	appID, appOK := pod.Annotations[annotations.KeyAppID]
+
+	if ctrlOK {
+		if pod.Namespace != k.controlPlaneNS {
+			return "", false, fmt.Errorf("control plane service in namespace '%s' is not allowed", pod.Namespace)
+		}
+
+		if !isControlPlaneService(ctrlPlane) {
+			return "", false, fmt.Errorf("unknown control plane service '%s'", pod.Name)
```
daixiang0, line 229: Could we cover this case in the test?

pkg/sentry/server/validator/kubernetes/kubernetes_test.go
```diff
@@ -795,3 +945,37 @@ func TestValidate(t *testing.T) {
 		})
 	}
 }
+
+func Test_isControlPlaneService(t *testing.T) {
```
daixiang0, line 949:
```suggestion
func isControlPlaneService(t *testing.T) {
```

IMO, a `test` prefix is duplicate in a test file.

JoshVanL: We need the `Test` prefix, otherwise the test won’t execute on `go test`

pkg/sentry/server/validator/kubernetes/kubernetes.go
JoshVanL: Done

artursouza (approved):

Issue #6800, opened 2023-08-13
[PubSub] CloudEvent trace* fields can't be customised when using custom CloudEvent


olitomlinson:
_This feels like a bug... ?_

Given a user specifies a custom CE envelope (`Content-Type: application/cloudevents+json`), any customised `trace*` headers in the CE envelope are ignored and overwritten by Dapr.

```
POST /v1.0/publish/redis-pubsub/orders HTTP/1.1
> Host: localhost:3501
> User-Agent: insomnia/2023.4.0
> dapr-app-id: myapp3
> Content-Type: application/cloudevents+json
> Accept: */*
> Content-Length: 317

{
     "id": "my-custom-id",
     "specversion": "my-custom-specversion",
     "type": "my-custom-type",
     "source": "my-custom-source",
     "traceid": "my-custom-traceid",
     "traceparent": "my-custom-traceparent",
     "tracestate": "my-custom-tracestate",
     "mycustomcefield": "my-custom-ce-field-value",
     "data": {
         "some": "body"
     }
}
```

**Actual behaviour**

```json
{
  "data": {
    "some": "body"
  },
  "id": "my-custom-id",
  "mycustomcefield": "my-custom-ce-field-value",
  "pubsubname": "redis-pubsub",
  "source": "my-custom-source",
  "specversion": "my-custom-specversion",
  "time": "2023-08-12T09:40:36Z",
  "topic": "orders",
  "traceid": "00-00000000000000000000000000000000-0000000000000000-00",
  "traceparent": "00-00000000000000000000000000000000-0000000000000000-00",
  "tracestate": "",
  "type": "my-custom-type"
}
```

**Expected behaviour**

```json
{
  "data": {
    "some": "body"
  },
  "id": "my-custom-id",
  "mycustomcefield": "my-custom-ce-field-value",
  "pubsubname": "redis-pubsub",
  "source": "my-custom-source",
  "specversion": "my-custom-specversion",
  "time": "2023-08-12T09:40:36Z",
  "topic": "orders",
  "traceid": "my-custom-traceid",
  "traceparent": "my-custom-traceparent",
  "tracestate": "my-custom-tracestate",
  "type": "my-custom-type"
}
```

berndverst:
This is intended behavior.

Cloudevents sent via `application/cloudevents+json` content type cannot be modified with other Dapr features, e.g. the metadata overwrite functionality. This is by design. You must provide the tracing information or other info via the cloudevent you are sending in this case.

[...]

PR #6805, opened 2023-08-14
Fix trace headers for existing cloud events

yaron2:
Closes https://github.com/dapr/dapr/issues/6800.


daixiang0:
Codes looks good, could we add tests to cover it?

ItalyPaleAle (dismissed):
LGTM but It would be nice if this code (that adds trace headers) could be moved to a function, for example in the universal package, so it’s not repeated twice (and we don’t risk it drifts in the future)

yaron2:
> LGTM but It would be nice if this code (that adds trace headers) could be moved to a function, for example in the universal package, so it’s not repeated twice (and we don’t risk it drifts in the future)

Changed, addressed two feedbacks with one commit.

ItalyPaleAle (approved):

ItalyPaleAle (approved):

artursouza (approved):

PR #6809, opened 2023-08-15
Remove preview feature flag ServiceInvocationStreaming

### LINKED ISSUE #6684: Remove feature flag `ServiceInvocationStreaming`
With #6651 the `ServiceInvocationStreaming` feature flag has been enabled by default in all builds.

If we are confident that this feature flag does not cause issues, we should remove the feature flag entirely, dropping all paths that are now-unused (and dropping the E2E test)

ItalyPaleAle:
Fixes #6684 

Flag was already "always-on" since #6651 ; this PR just removes it entirely (because it was already hardcoded as enabled, there shouldn't be any behavior change)


mukundansundar (approved):

mukundansundar:
@artursouza There seems to be flaky tests in java sdk master branch 
MethodInvokeIT
TracingIT
SdkResiliencyIT

All three seem to be flaky and the gRPC TracingIT seems to be failing most of the time

PR #6813, opened 2023-08-15
Small fixes from closed E2E test branch

ItalyPaleAle:
From the now-closed PR #6546 whose purpose was to add E2E tests.

Most notably, includes a fix in how DAPR_ENV_KEYS is handled, so it's case-insensitive


yaron2 (dismissed):

tests/e2e/utils/http.go
```diff
@@ -41,6 +41,11 @@ func InitHTTPClient(allowHTTP2 bool) {
 	httpClient = NewHTTPClient(allowHTTP2)
 }
 
+// GetHTTPClient returns the shared httpClient object.
+func GetHTTPClient() *http.Client {
```
yaron2, line 45: Is this method referenced from anywhere?

berndverst (approved):

tests/apps/utils/helpers.go
```diff
@@ -27,3 +29,15 @@ func IsTruthy(val string) bool {
 		return false
 	}
 }
+
+// PortFromEnv returns a port from an env var, or the default value.
+func PortFromEnv(envName string, defaultPort int) (res int) {
```
yaron2, line 34: Same question as https://github.com/dapr/dapr/pull/6813/files#r1296388182.

berndverst: Doesn't seem that way, but the intention probably is to use this helper method throughout E2E tests in the future. The refactor of existing E2E tests wasn't added here.

PR #6819, opened 2023-08-16
Various improvements in workflow engine performance + E2E test fixes

ItalyPaleAle:
Various improvements that came out of a debugging session. It includes a number of perf improvements too:

1. Use state bulk get operations when loading the workflow state.
   - This replaces the current loops that load each key individually, performing possibly hundreds of roundtrips to the state store
   - It is implemented by adding a GetBulkState method to the actors interface. This method is only used by internal actors and not exposed to end-users (e.g. not callable by regular actors). Should that be desirable, it can be done in the future.
1. Allow running E2E tests outside of Kubernetes for a faster debug loop experience. How to do that is explained in a comment; TLDR:
   - Run the E2E app with the Dapr CLI: `ASPNETCORE_URLS=http://*:3000 dapr run --app-id workflowsapp --resources-path ./resources -- dotnet run`
   - Run the test: `WORKFLOW_APP_ENDPOINT=http://localhost:3000 DAPR_E2E_TEST="workflows" make test-clean test-e2e-all`
   - Locally, tests use SQLite
3. Changed various log levels: debug logs are moved to "debug", and warnings are moved to "warn"
4. When deleting leftover state, use a single transaction per each actor ID. This can help reducing roundtrips to the state store.
5. `DecodeInternalActorData` processes data as a stream directly without first buffering in-memory
6. Some cleanups in the E2E test code


pkg/actors/internal_actor_test.go
```diff
@@ -159,11 +160,10 @@ func TestInternalActorCall(t *testing.T) {
 		// Verify the response metadata matches what we expect
 		assert.Equal(t, int32(200), resp.Status().Code)
 		contentType := resp.ContentType()
-		data, _ := resp.RawDataFull()
 		assert.Equal(t, invokev1.OctetStreamContentType, contentType)
 
 		// Verify the actor got all the expected inputs (which are echoed back to us)
-		info, err := decodeTestResponse(data)
+		info, err := decodeTestResponse(resp.RawData())
```
RyanLettieri, line 166: What was the purpose behind changing this from RawDataFull() to RawData()?

pkg/runtime/wfengine/backend.go
```diff
@@ -129,13 +129,7 @@ func (be *actorBackend) GetOrchestrationMetadata(ctx context.Context, id api.Ins
...
 
 	defer res.Close()
-	data, err := res.RawDataFull()
-	if err != nil {
-		return nil, fmt.Errorf("failed to read the internal actor response: %w", err)
-	}
-	if len(data) == 0 {
-		return nil, api.ErrInstanceNotFound
-	}
+	data := res.RawData()
```
RyanLettieri, line 132: Same as previous comment, why are we now using RawData() and no longer error checking here?

ItalyPaleAle: `DecodeInternalActorData` which is what this value is passed to internally uses `gob.NewDecoder` which accepts a stream.

`res.RawDataFull()` returns a byte slice, reading the stream fully in-memory first (internally, InvokeMethodResponse manages data as a stream).

`res.RawData()` just returns the stream.

Before, we were using `RawDataFull` to buffer the data in-memory, then internally in `DecodeInternalActorData` we were getting a readable stream from the data. So we were reading a stream, buffering in-memory, and then re-creating a stream from that.

Errors are now checked by `DecodeInternalActorData` while gob reads the stream.

pkg/actors/internal_actor_test.go
ItalyPaleAle: See https://github.com/dapr/dapr/pull/6819#discussion_r1296555262

pkg/runtime/wfengine/component.go
```diff
@@ -95,22 +95,19 @@ func (c *workflowEngineComponent) Start(ctx context.Context, req *workflows.Star
...
-	var err error
-
-	workflowID, err = c.client.ScheduleNewOrchestration(ctx, req.WorkflowName, opts...)
+	workflowID, err := c.client.ScheduleNewOrchestration(ctx, req.WorkflowName, opts...)
 	if err != nil {
 		return nil, fmt.Errorf("unable to start workflow: %w", err)
 	}
 
-	c.logger.Infof("created new workflow instance with ID '%s'", workflowID)
+	c.logger.Debugf("Created new workflow instance with ID '%s'", workflowID)
```
RyanLettieri, line 110: Why are you changing all the statements for starting, stopping, raising event, etc.. from info to debug?

ItalyPaleAle: We normally treat these kinds of logs as debug logs in APIs, since they can add a lot of noise for heavy workflow users. But I can revert this if needed.

pkg/runtime/wfengine/workflow.go
```diff
@@ -418,40 +427,43 @@ func (wf *workflowActor) runWorkflow(ctx context.Context, actorID string, remind
...
-			activityRequestBytes, err := actors.EncodeInternalActorData(ActivityRequest{
-				HistoryEvent: eventData,
-			})
-			if err != nil {
-				return err
-			}
-			targetActorID := getActivityActorID(actorID, e.EventId, state.Generation)
+		ts := e.GetTaskScheduled()
+		if ts == nil {
+			wfLogger.Warn("don't know how to process task %v", e)
```
RyanLettieri, line 435: Can we change this to "unable to process task: %v"?

pkg/runtime/wfengine/workflowstate.go
```diff
@@ -203,47 +235,64 @@ func LoadWorkflowState(ctx context.Context, actorRuntime actors.Actors, actorID
...
 	state := NewWorkflowState(config)
 	state.Generation = metadata.Generation
-	// CONSIDER: Do some of these loads in parallel
+	state.Inbox = make([]*backend.HistoryEvent, metadata.InboxLength)
+	state.History = make([]*backend.HistoryEvent, metadata.HistoryLength)
+
+	bulkReq := &actors.GetBulkStateRequest{
+		ActorType: config.workflowActorType,
+		ActorID:   actorID,
+		Keys:      make([]string, metadata.InboxLength+metadata.HistoryLength+1),
```
RyanLettieri, line 248: Nit: I assume the +1 is for custom status? If so, can we add a little comment stating this?

pkg/runtime/wfengine/workflowstate.go
```diff
@@ -256,7 +305,7 @@ func (s *workflowState) GetPurgeRequest(actorID string) (*actors.TransactionalRe
 	req := &actors.TransactionalRequest{
 		ActorType:  s.config.workflowActorType,
 		ActorID:    actorID,
-		Operations: make([]actors.TransactionalOperation, 0, 100),
+		Operations: make([]actors.TransactionalOperation, 0, len(s.Inbox)+len(s.History)+2),
```
RyanLettieri, line 308: Add a comment stating what the +2 is for. I assume CustomStatus + MetaData?

RyanLettieri (changes requested):
Looks good! I just had a few minor nit-pick comments

pkg/runtime/wfengine/workflowstate.go
ItalyPaleAle:
```suggestion
		// Initial capacity should be enough to contain the entire inbox, history, and custom status + metadata
		Operations: make([]actors.TransactionalOperation, 0, len(s.Inbox)+len(s.History)+2),
```

tests/e2e/workflows/workflow_test.go
```diff
@@ -155,132 +168,101 @@ func raiseEventTest(url string, instanceID string) string {
...
+	require.NoError(t, utils.HealthCheckApps(externalURL))
+
+	// Generate a unique test suffix for this test
+	suffixBytes := make([]byte, 7)
+	_, err := io.ReadFull(rand.Reader, suffixBytes)
+	require.NoError(t, err)
+	suffix := hex.EncodeToString(suffixBytes)
+
+	// Run tests
+	t.Run("Start", func(t *testing.T) {
```
RyanLettieri, line 253: One thing I failed to realize with the previous Run structure of this test was that when a failure occurred inside the function/test, we didn't get the line number to trace exactly where the failure was. With this new method of passing in the "testing" pointer, will this provide the line number in the instance of failure? If not, how difficult would it be to add that? (definitely not a blocker for this review)

ItalyPaleAle: This will help a bit because at least you can see the line that failed in the test (like L254). But you won't see the exact line that failed in the test function, because that doesn't have the `t` and doesn't do assertions. We can change that if needed (maybe a separate PR)

RyanLettieri (approved):

mukundansundar (commented):
overall looks good. just a few comments and questions. Please add E2E for the Get Bulk state for actors.

pkg/components/state/state_config.go
```diff
@@ -93,7 +93,7 @@ func GetModifiedStateKey(key, storeName, appID string) (string, error) {
 }
 
 func GetOriginalStateKey(modifiedStateKey string) string {
-	splits := strings.Split(modifiedStateKey, daprSeparator)
+	splits := strings.SplitN(modifiedStateKey, daprSeparator, 2)
```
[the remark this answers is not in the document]
ItalyPaleAle: Yes this is correct and it does change the key...

However, outside of actors, `a|b|c` is not a valid key. [`checkIllegalKey`](https://github.com/dapr/dapr/blob/c1522bc66ffd7937c2f01f057744d179722a128f/pkg/components/state/state_config.go#L128) enforces no `|` in the key name when users perform state store operations.

pkg/actors/actors.go
```diff
@@ -81,6 +82,7 @@ type Actors interface {
 	Call(ctx context.Context, req *invokev1.InvokeMethodRequest) (*invokev1.InvokeMethodResponse, error)
 	Init(context.Context) error
 	GetState(ctx context.Context, req *GetStateRequest) (*StateResponse, error)
+	GetBulkState(ctx context.Context, req *GetBulkStateRequest) (BulkStateResponse, error)
```
cgillum, line 85: This is exactly the type of API I was hoping we could add to improve efficiency. Glad to see it here!

pkg/runtime/wfengine/activity.go
```diff
@@ -104,7 +105,7 @@ func (a *activityActor) InvokeMethod(ctx context.Context, actorID string, method
 
 // InvokeReminder implements actors.InternalActor and executes the activity logic.
 func (a *activityActor) InvokeReminder(ctx context.Context, actorID string, reminderName string, data []byte, dueTime string, period string) error {
-	wfLogger.Debugf("invoking reminder '%s' on activity actor '%s'", reminderName, actorID)
+	wfLogger.Debugf("Invoking reminder '%s' on activity actor '%s'", reminderName, actorID)
```
cgillum, line 108: Curious about this change - I've recently been seeing the VS Code linter for Go give me warnings about starting logs (or maybe it was just error messages?) with capital letters - i.e., it wanted me to start them with lowercase. Do we have separate guidance for capitalization in the Dapr project?

pkg/runtime/wfengine/component.go
cgillum: I'd love to better understand how we think about logging in Dapr more generally. I tried to be really selective about which logs were promoted to Info, figuring that starting a workflow may be something that an operator looking through production logs might need to know about. But happy to learn if we think there are better ways to help provide these kinds of operational insights (OTel traces?).

pkg/runtime/wfengine/component.go
```diff
@@ -141,7 +138,7 @@ func (c *workflowEngineComponent) Purge(ctx context.Context, req *workflows.Purg
 
 	if err := c.client.PurgeOrchestrationState(ctx, api.InstanceID(req.InstanceID)); err != nil {
 		if errors.Is(err, api.ErrInstanceNotFound) {
-			c.logger.Infof("Unable to purge the instance: '%s', no such instance exists", req.InstanceID)
+			c.logger.Errorf("Unable to purge the instance: '%s', no such instance exists", req.InstanceID)
```
cgillum, line 141: At maximum, I think this should just be a warning. Error to me feels like something with Dapr is broken, which is definitely not the case here.

pkg/runtime/wfengine/component.go
```diff
@@ -244,13 +241,13 @@ func (c *workflowEngineComponent) PurgeWorkflow(ctx context.Context, req *workfl
 
 	if err := c.client.PurgeOrchestrationState(ctx, api.InstanceID(req.InstanceID)); err != nil {
 		if errors.Is(err, api.ErrInstanceNotFound) {
-			c.logger.Infof("The requested instance: '%s' does not exist or has already been purged", req.InstanceID)
+			c.logger.Errorf("The requested instance: '%s' does not exist or has already been purged", req.InstanceID)
```
cgillum, line 244: Same comment about warning vs. error

pkg/runtime/wfengine/workflow.go
```diff
@@ -331,19 +328,28 @@ func (wf *workflowActor) runWorkflow(ctx context.Context, actorID string, remind
...
+		activityActorID := getActivityActorID(actorID, taskID, state.Generation)
+		if transactionalRequests[activityActorID] == nil {
+			transactionalRequests[activityActorID] = []actors.TransactionalOperation{op}
+		} else {
+			transactionalRequests[activityActorID] = append(transactionalRequests[activityActorID], op)
+		}
+	}
+	// TODO: for optimization make multiple go routines and run them in parallel
+	for activityActorID, operations := range transactionalRequests {
+		err = wf.actors.TransactionalStateOperation(ctx, &actors.TransactionalRequest{
```
cgillum, line 346: With this change to put all the deletes into a single transaction, do we run the risk of exceeding Cosmos DB transaction batch size limitations?

pkg/runtime/wfengine/workflowstate_test.go
```diff
@@ -73,45 +75,43 @@ func TestClearingInbox(t *testing.T) {
 	state.ClearInbox()
 
 	req, err := state.GetSaveRequest("wf1")
-	if assert.NoError(t, err) {
-		assert.Equal(t, "wf1", req.ActorID)
-		assert.Equal(t, workflowActorType, req.ActorType)
+	require.NoError(t, err)
```
cgillum, line 78: I wish I'd known about the `require` package sooner. 😅 Thanks for taking the time to clean this up.

tests/e2e/workflows/workflow_test.go
```diff
@@ -28,123 +31,133 @@ import (
...
-			IngressEnabled:    true,
-			MetricsEnabled:    true,
-			DaprMemoryLimit:   "200Mi",
-			DaprMemoryRequest: "100Mi",
-			AppMemoryLimit:    "200Mi",
-			AppMemoryRequest:  "100Mi",
-			AppPort:           3000,
-		},
+	// This test can be run outside of Kubernetes too
+	// Run the workflow e2e app using, for example, the Dapr CLI:
```
cgillum, line 41: 😍

pkg/runtime/wfengine/activity.go
ItalyPaleAle: Error messages always start lowercase per Go convention: https://github.com/golang/go/wiki/CodeReviewComments#error-strings

For logs, we generally try to make them start with an uppercase letter.

pkg/runtime/wfengine/workflow.go
ItalyPaleAle: This doesn't put _all_ deletes in a single transaction, but only those for the same actor (because of partition keys). But yes, we do.

- We have been talking about dropping support for Cosmos DB anyways, since it has issues also when saving state
- If absolutely needed, we can split this into multiple calls to TransactionalStateOperation. Since these are just deletes, it's _probably_ ok if we don't have atomicity here.

pkg/runtime/wfengine/component.go
ItalyPaleAle: Will change

pkg/runtime/wfengine/component.go
ItalyPaleAle: Will change

pkg/runtime/wfengine/component.go
ItalyPaleAle:
```suggestion
			c.logger.Warnf("Unable to purge the instance: '%s', no such instance exists", req.InstanceID)
```

pkg/runtime/wfengine/component.go
ItalyPaleAle:
```suggestion
			c.logger.Warnf("The requested instance: '%s' does not exist or has already been purged", req.InstanceID)
```

pkg/components/state/state_config.go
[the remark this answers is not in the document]
mukundansundar: won't this be a breaking change for existing actors?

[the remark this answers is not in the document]
ItalyPaleAle: Aside from this addition (pkg/actors.GetBulkState), the method `GetOriginalStateKey` is only used in the user-facing gRPC and HTTP APIs, so no.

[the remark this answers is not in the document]
mukundansundar: let me try to understand this correctly. 
> the method GetOriginalStateKey is only used in the user-facing gRPC and HTTP APIs, so no

For existing actors when users query over gRPC and HTTP, the `GetOriginalStateKey` response will be different between 1.11 and 1.12. 
Isn't that a breaking change?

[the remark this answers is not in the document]
ItalyPaleAle: So here's the current master.

GetOriginalStateKey is used in 4 places (aside from unit tests):

- `pkg/grpc.GetBulkState`: https://github.dev/dapr/dapr/blob/c1522bc66ffd7937c2f01f057744d179722a128f/pkg/grpc/api.go#L535
- `pkg/grpc/universal.QueryStateAlpha1`: https://github.dev/dapr/dapr/blob/c1522bc66ffd7937c2f01f057744d179722a128f/pkg/grpc/universalapi/api_state_query.go#L105
- `pkg/http.onBulkGetState`: https://github.dev/dapr/dapr/blob/c1522bc66ffd7937c2f01f057744d179722a128f/pkg/http/api.go#L480
- `pkg/http.onQueryStateHandler`: https://github.dev/dapr/dapr/blob/c1522bc66ffd7937c2f01f057744d179722a128f/pkg/http/api.go#L1964

All of the above are unrelated from actors. Actor state operations are performed in pkg/actors and not in pkg/http or pkg/grpc.

The state management APIs do not allow users to create keys that have `|` in there. For example, this is `pkg/grpc.SaveState`: https://github.dev/dapr/dapr/blob/c1522bc66ffd7937c2f01f057744d179722a128f/pkg/grpc/api.go#L645 which uses `GetModifiedStateKey`. Inside that function, you see that it uses `checkIllegalKey`: https://github.dev/dapr/dapr/blob/c1522bc66ffd7937c2f01f057744d179722a128f/pkg/components/state/state_config.go#L65-L68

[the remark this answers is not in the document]
ItalyPaleAle: I figured a way to do this without changing GetOriginalStateKey.

pkg/runtime/wfengine/component.go
halspang: I'm fine with the logs either way, but I think in this case debug does make sense as you could get a very similar log to this with API logging, right? If you're just looking for the calls, that's what I would do. If I have was having issues with my workflow, that's when I'd want this log and would turn on debug logging to see the extra info.

halspang (dismissed):

halspang:
After I hit approve, I noticed you have a conflict @ItalyPaleAle. Let me know when you take care of it so I can look at that file again :)

ItalyPaleAle:
> After I hit approve, I noticed you have a conflict @ItalyPaleAle. Let me know when you take care of it so I can look at that file again :)

Should be fixed now, thanks!

pkg/runtime/wfengine/wfengine.go
```diff
@@ -74,6 +74,14 @@ func NewWorkflowConfig(appID string) wfConfig {
 	}
 }
 
+// String implements fmt.Stringer and is primarily used for debugging purposes.
+func (c *wfConfig) String() string {
+	if c == nil {
+		return "(nil)"
```
yaron2, line 80: Why not return empty string here?

pkg/components/state/state_config.go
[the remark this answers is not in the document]
mukundansundar: This is still modifying the key.

PR #6826, opened 2023-08-17
Fix for 404 error in workflow timer expiration

RyanLettieri:
# Description

Currently, timers in workflow are not being stored into the statestore. The object is created, added to the inbox, and then removed. Later, an attempt is made to remove the timer from the statestore, but the timer was never stored so a 404 is returned by cosmosDB (other statestores mask this).

This PR now checks if the inbox item to be added is a timer and if it has fired, and if so, does not clear it out of the inbox.

This fix was verified by making the changes and re-running the workflow quickstart in the python sdk. No errors occurred and the correct number of keys were stored.

## Issue reference

Please reference the issue this PR will close: #_[[6647](https://github.com/dapr/dapr/issues/6647)]_

## Checklist

* [ x] Code compiles correctly
* [ ] Created/updated tests
* [ x] Unit tests passing
* [ x] End-to-end tests passing

### DIFF
--- a/pkg/runtime/wfengine/workflow.go
+++ b/pkg/runtime/wfengine/workflow.go
@@ -322,13 +322,31 @@ func (wf *workflowActor) runWorkflow(ctx context.Context, actorID string, remind
 
 	// The logic/for loop below purges/removes any leftover state from a completed or failed activity
 	// TODO: for optimization make multiple go routines and run them in parallel
-	for _, e := range state.Inbox {
+	for num, e := range state.Inbox {
 		var taskID int32
 		if ts := e.GetTaskCompleted(); ts != nil {
 			taskID = ts.TaskScheduledId
 		} else if tf := e.GetTaskFailed(); tf != nil {
 			taskID = tf.TaskScheduledId
 		} else {
+			if strings.Contains(e.String(), "timerFired") {
+				numZero := 6 - len(fmt.Sprint(num))
+				req := actors.TransactionalRequest{
+					ActorType: wf.config.workflowActorType,
+					ActorID:   actorID,
+					Operations: []actors.TransactionalOperation{{
+						Operation: actors.Upsert,
+						Request: actors.TransactionalUpsert{
+							Key: inboxKeyPrefix + "-" + strings.Repeat("0", numZero) + fmt.Sprint(num),
+						},
+					}},
+				}
+				err = wf.actors.TransactionalStateOperation(ctx, &req)
+				if err != nil {
+					wfLogger.Infof("RRL failed to save timer with error: %w, and req: %v", err, req)
+					return fmt.Errorf("failed to save timer with error: %w", err)
+				}
+			}
 			continue
 		}
 		req := actors.TransactionalRequest{
@@ -528,8 +546,13 @@ func (wf *workflowActor) saveInternalState(ctx context.Context, actorID string, 
 
 func (wf *workflowActor) createReliableReminder(ctx context.Context, actorID string, namePrefix string, data any, delay time.Duration) (string, error) {
 	// Reminders need to have unique names or else they may not fire in certain race conditions.
-	reminderName := fmt.Sprintf("%s-%s", namePrefix, uuid.NewString()[:8])
-	wfLogger.Debugf("%s: creating '%s' reminder with DueTime = %s", actorID, reminderName, delay)
+	var reminderName string
+	if strings.Contains(namePrefix, "timer") {
+		reminderName = fmt.Sprintf("timer-%s", uuid.NewString()[:8])
+	} else {
+		reminderName = fmt.Sprintf("%s-%s", namePrefix, uuid.NewString()[:8])
+	}
+
 	dataEnc, err := json.Marshal(data)
 	if err != nil {
 		return reminderName, fmt.Errorf("failed to encode data as JSON: %w", err)
--- a/pkg/runtime/wfengine/workflow.go
+++ b/pkg/runtime/wfengine/workflow.go
@@ -322,31 +322,13 @@ func (wf *workflowActor) runWorkflow(ctx context.Context, actorID string, remind
 
 	// The logic/for loop below purges/removes any leftover state from a completed or failed activity
 	// TODO: for optimization make multiple go routines and run them in parallel
-	for num, e := range state.Inbox {
+	for _, e := range state.Inbox {
 		var taskID int32
 		if ts := e.GetTaskCompleted(); ts != nil {
 			taskID = ts.TaskScheduledId
 		} else if tf := e.GetTaskFailed(); tf != nil {
 			taskID = tf.TaskScheduledId
 		} else {
-			if strings.Contains(e.String(), "timerFired") {
-				numZero := 6 - len(fmt.Sprint(num))
-				req := actors.TransactionalRequest{
-					ActorType: wf.config.workflowActorType,
-					ActorID:   actorID,
-					Operations: []actors.TransactionalOperation{{
-						Operation: actors.Upsert,
-						Request: actors.TransactionalUpsert{
-							Key: inboxKeyPrefix + "-" + strings.Repeat("0", numZero) + fmt.Sprint(num),
-						},
-					}},
-				}
-				err = wf.actors.TransactionalStateOperation(ctx, &req)
-				if err != nil {
-					wfLogger.Infof("RRL failed to save timer with error: %w, and req: %v", err, req)
-					return fmt.Errorf("failed to save timer with error: %w", err)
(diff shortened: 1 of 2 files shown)


cgillum:
Hey @RyanLettieri, can you elaborate a bit more in the PR description on the testing you did to confirm the fix?

RyanLettieri:
> Hey @RyanLettieri, can you elaborate a bit more in the PR description on the testing you did to confirm the fix?

Updated

pkg/runtime/wfengine/workflow.go
```diff
@@ -322,13 +322,31 @@ func (wf *workflowActor) runWorkflow(ctx context.Context, actorID string, remind
...
+					ActorType: wf.config.workflowActorType,
+					ActorID:   actorID,
+					Operations: []actors.TransactionalOperation{{
+						Operation: actors.Upsert,
+						Request: actors.TransactionalUpsert{
+							Key: inboxKeyPrefix + "-" + strings.Repeat("0", numZero) + fmt.Sprint(num),
+						},
+					}},
+				}
+				err = wf.actors.TransactionalStateOperation(ctx, &req)
```
cgillum, line 344: Inbox updates are supposed to happen at the very end of this function:

```go
	state.ApplyRuntimeStateChanges(runtimeState)
	state.ClearInbox()

```

If the inbox isn't getting properly cleared there, then we should do the work to understand why. This loop here is not intended to be used for touching workflow state. Workflow state should only be altered as part of `saveInternalState`.

RyanLettieri: The inbox is getting cleared, however, this line inside ClearInbox() is where the problem resides:
https://github.com/dapr/dapr/blob/c1522bc66ffd7937c2f01f057744d179722a128f/pkg/runtime/wfengine/workflowstate.go#L101

When we create a durable timer, we add it to the inbox: https://github.com/dapr/dapr/blob/c1522bc66ffd7937c2f01f057744d179722a128f/pkg/runtime/wfengine/workflow.go#L312

We then call ClearInbox() later in that runWorkflow() function. This removes the timer from the inbox and increments the s.InboxRemovedCount.

Inside workflowstate.go, we then add state operations: https://github.com/dapr/dapr/blob/c1522bc66ffd7937c2f01f057744d179722a128f/pkg/runtime/wfengine/workflowstate.go#L114

This adds 1 delete operation into the state request since the s.InboxRemovedCount is now 1 and the inbox size is 0: https://github.com/dapr/dapr/blob/c1522bc66ffd7937c2f01f057744d179722a128f/pkg/runtime/wfengine/workflowstate.go#L162

This request is then executed which leads to the 404 since there was never anything actually stored inside the statestore for the timer

pkg/runtime/wfengine/workflow.go
```diff
@@ -322,13 +322,31 @@ func (wf *workflowActor) runWorkflow(ctx context.Context, actorID string, remind
...
-	for _, e := range state.Inbox {
+	for num, e := range state.Inbox {
 		var taskID int32
 		if ts := e.GetTaskCompleted(); ts != nil {
 			taskID = ts.TaskScheduledId
 		} else if tf := e.GetTaskFailed(); tf != nil {
 			taskID = tf.TaskScheduledId
 		} else {
+			if strings.Contains(e.String(), "timerFired") {
+				numZero := 6 - len(fmt.Sprint(num))
```
ItalyPaleAle, line 333: To avoid expensive conversions to string
```suggestion
				numZero := 5 - int(math.Log10(float64(num)))
```

pkg/runtime/wfengine/workflow.go
```diff
@@ -528,8 +546,13 @@ func (wf *workflowActor) saveInternalState(ctx context.Context, actorID string,
 
 func (wf *workflowActor) createReliableReminder(ctx context.Context, actorID string, namePrefix string, data any, delay time.Duration) (string, error) {
 	// Reminders need to have unique names or else they may not fire in certain race conditions.
-	reminderName := fmt.Sprintf("%s-%s", namePrefix, uuid.NewString()[:8])
-	wfLogger.Debugf("%s: creating '%s' reminder with DueTime = %s", actorID, reminderName, delay)
+	var reminderName string
+	if strings.Contains(namePrefix, "timer") {
+		reminderName = fmt.Sprintf("timer-%s", uuid.NewString()[:8])
```
ItalyPaleAle, line 551: Please don't use uuid.NewString() which can panic. Instead, use uuid.NewRandom() (and handle errors)

pkg/runtime/wfengine/workflow.go
mukundansundar: Please add a constant for the magic number `5` or `6`

cgillum:
@RyanLettieri one more thing, can we add another unit test, ideally in workflowstate_test.go, that can be used for regression testing? Ideally if we comment out your fix, this test will fail.

pkg/runtime/wfengine/workflow.go
cgillum: Wait, does this mean you're adding arbitrary data into the state store so that the delete can succeed? This feels like the wrong direction to me. I understand now why timers are special, and it's because we're adding something to the inbox on line 312 which has no representation in the state store. But if we know that something isn't in the state store, instead of adding it there, we should simply *not try to delete it* - right?

If you agree with the approach, I think we should instead just update `ClearInbox()` to not count `timerFired` events when updating the value of `s.historyRemovedCount`. That way we don't try to delete something that isn't there.
Here's an example (haven't tested it):

```go
func (s *workflowState) ClearInbox() {
	for _, e := range s.Inbox {
		if tf := e.GetTimerFired(); tf != nil {
			// ignore timer events since those aren't saved into the state store
			continue
		}
		s.inboxRemovedCount--
	}
	s.Inbox = nil
	s.inboxAddedCount = 0
}
```

pkg/runtime/wfengine/workflow.go
mukundansundar: Please make magic number 8 a constant.

mukundansundar, line 551: please consider adding the structure of the `"timer"` key as a comment.

pkg/runtime/wfengine/workflow.go
```diff
@@ -322,13 +322,31 @@ func (wf *workflowActor) runWorkflow(ctx context.Context, actorID string, remind
...
 		} else {
+			if strings.Contains(e.String(), "timerFired") {
+				numZero := 6 - len(fmt.Sprint(num))
+				req := actors.TransactionalRequest{
+					ActorType: wf.config.workflowActorType,
+					ActorID:   actorID,
+					Operations: []actors.TransactionalOperation{{
+						Operation: actors.Upsert,
+						Request: actors.TransactionalUpsert{
+							Key: inboxKeyPrefix + "-" + strings.Repeat("0", numZero) + fmt.Sprint(num),
```
mukundansundar, line 340: have `-` defined as a constant separator for history/inbox?

pkg/runtime/wfengine/workflow.go
mukundansundar: qq Why are we adding a `"timer"` record here after the `"timer"` has been fired?

pkg/runtime/wfengine/workflow.go
mukundansundar: Why is it that this namePrefix is a special case which requires the reminder name to just timer instead of the incoming name prefix?

RyanLettieri:
@mukundansundar and @ItalyPaleAle The fix changed significantly so your comments are no longer applicable since workflow.go is no longer altered

pkg/runtime/wfengine/workflowstate.go
```diff
@@ -98,7 +98,14 @@ func (s *workflowState) AddToInbox(e *backend.HistoryEvent) {
 }
 
 func (s *workflowState) ClearInbox() {
-	s.inboxRemovedCount += len(s.Inbox)
+	// s.inboxRemovedCount += len(s.Inbox)
```
ItalyPaleAle, line 101: Let's not have commented-out code

pkg/runtime/wfengine/workflowstate.go
```diff
@@ -98,7 +98,14 @@ func (s *workflowState) AddToInbox(e *backend.HistoryEvent) {
...
+	for _, e := range s.Inbox {
+		if tf := e.GetTimerFired(); tf != nil {
```
ItalyPaleAle, line 103: Not_

```suggestion
		if e.GetTimerFired() != nil {
```

mukundansundar (approved):

Issue #5899, opened 2023-02-07
[Proposal]: Extend Actor SDK with state garbage collection through reminders


JoshVanL:
## In what area(s)?

/area runtime

(I've used the Go SDK in the below examples).

Currently, there is no way for actor state to be deleted automatically after a period of time. This means that the amount of data that can be written to the store over time is unbounded. The state store will eventually fill up unless users implement clean up logic, out of band of dapr.

I propose that we extend the dapr SDKs to implement a function:

```go
func SetFor(stateName string, value any, duration time.Duration) 
```

This function would create an actor reminder for the given type, with a reminder name following the schema (suggestions very welcome!):

`io.dapr.setFor.<stateName>`

The SDK will implement a middleware to the existing reminder handler which looks for this well known schema. The middleware will delete the relevant state and then delete the reminder. The middleware will prevent `ReminderCall` from being invoked.

Dapr will recommend users to _always_ use `SetFor` with some appropriate duration (e.g. 2 weeks) to ensure state is always cleaned up, unless they already have state management in place.

This method is only best effort since an application may no longer implement that type when a reminder is invoked.

---

We could implement this logic in the dapr runtime and save some hops, however this would require dapr API changes.

cgillum:
FYI @ItalyPaleAle this could be a valuable feature for us to use to help with workflow actor garbage collection.

JoshVanL:
Note: schema definition pending- we should make it clear in the documentation that dapr reserves the `"io.dapr."` namespace and cannot be used by users. This should be added to docs, and added as validation to the SDKs.

JoshVanL:
[Here](https://github.com/dapr/go-sdk/compare/main...JoshVanL:go-sdk:actor-state-manager-setfor) is a poc implementation for the go-sdk. It still needs tests.

ItalyPaleAle:
I've been thinking about this issue since you mentioned it, @cgillum 

Some food for thought:

- I think the first thing that we should do is document the need to perform cleanups as needed as a best practice. Right now our docs don't make a great job at explaining the lifecycle of the persistent state stored by actors, and we should explain how it's persisted and recommend developers to create their own reminders to invoke a "Dispose" method as they see fit. This also allows developers to have the ability to "push back" a deadline.
[...]

PR #6827, opened 2023-08-17
Actor State `ttlExpireTime`: Pipe `ttlExpireTime` to state responses

JoshVanL:
Part of https://github.com/dapr/dapr/issues/5899

PR pipes `ttlExpireTime` to get state responses, including actors.

Adds integration tests for state and actor state TTLs. Updates the existing actor state e2e test to check for `ttlExpireTime`.

Fixed an issue where an actors go routine was not returning on close.


tests/dapr_tests.mk
```diff
@@ -90,7 +90,7 @@ DAPR_CONTAINER_LOG_PATH?=./dist/container_logs
 DAPR_TEST_LOG_PATH?=./dist/logs
 
 ifeq ($(DAPR_TEST_STATE_STORE),)
-DAPR_TEST_STATE_STORE=redis
+DAPR_TEST_STATE_STORE=postgres
```
ItalyPaleAle, line 93: Let's keep this unrelated change in a different PR please.

ItalyPaleAle (dismissed):

tests/dapr_tests.mk
JoshVanL: I still need to finalize this PR, but this is because redis doesn't support `ttlExpireTime`.

pkg/actors/placement/placement.go
```diff
@@ -314,7 +315,7 @@ func (p *actorPlacement) establishStreamConn(ctx context.Context) (established b
 		}
 
 		err := p.client.connectToServer(ctx, serverAddr)
-		if err == errEstablishingTLSConn {
```
ItalyPaleAle, line 317: Changes to this file seem unrelated

pkg/grpc/api_actor_test.go
```diff
@@ -134,6 +134,9 @@ func TestGetActorState(t *testing.T) {
 			Key:       "key1",
 		}).Return(&actors.StateResponse{
 			Data: data,
+			Metadata: map[string]string{
+				"ttlExpireTime": "2020-10-02T22:00:00Z",
```
artursouza, line 138: Sorry to be that person, but why not just "expiration"? Is "ttlExpireTime" used in other systems and we want to mimic?

pkg/actors/actors.go
```diff
@@ -287,9 +287,9 @@ func (a *actorsRuntime) startAppHealthCheck(ctx context.Context, opts ...health.
 	for {
 		select {
 		case <-ctx.Done():
-			break
+			return
```
artursouza, line 290: Why is this needed?

tests/config/dapr_postgres_state.yaml
```diff
@@ -20,10 +20,10 @@ spec:
...
   initTimeout: 1m
   metadata:
-  - name: connectionString
-    value: "host=dapr-postgres-postgresql.dapr-tests.svc.cluster.local user=postgres password=example port=5432 connect_timeout=10 database=dapr_test"
-  - name: table
-    value: querytable
+    - name: connectionString
+      value: "host=dapr-postgres-postgresql.dapr-tests.svc.cluster.local user=postgres password=example port=5432 connect_timeout=10 database=dapr_test"
+    - name: table
+      value: statetable
```
artursouza, line 26: Why this change is needed?

artursouza (changes requested):
LGTM overall but have some questions, mainly about naming. This is a contract after being merged.

pkg/grpc/api_actor_test.go
JoshVanL: `expiration` could be ambiguous. It is clearer that `ttlExpireTime` refers to and pairs with the `ttl` field.

pkg/actors/actors.go
JoshVanL: The logic is currently bugged. The `break` statement will break from the `select` block and not the `for` loop.

tests/config/dapr_postgres_state.yaml
JoshVanL: It doesn't matter all that much, but the table name is conflicting with the [query state store](https://github.com/dapr/dapr/blob/e57a91fc8d2930eca71c8fdbf20084ed18161fc2/tests/config/dapr_postgres_query_state.yaml#L26).

JoshVanL:
Flake?

pkg/grpc/api_actor_test.go
artursouza: "ttl" is a duration: "1min" or "3 hours". In this case, it means the expiration timestamp, like the date we see stamped at food labels.

pkg/actors/actors.go
```diff
@@ -660,7 +660,8 @@ func (a *actorsRuntime) GetState(ctx context.Context, req *GetStateRequest) (*St
 	}
 
 	return &StateResponse{
-		Data: resp.Data,
+		Data:     resp.Data,
+		Metadata: resp.Metadata,
```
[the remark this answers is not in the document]
JoshVanL: It is currently only `ttlExpireTime`.

pkg/grpc/api_actor_test.go
JoshVanL: Yes, `ttl` is what is given in the metadata when saving and is a duration. `ttlExpireTime` is on the get response and is the timestamp of when that `ttl` duration will take effedct.

mukundansundar: Can we shorten it? `ttlExpires` similar to HTTP `Expires` header but more qualified to `ttl` expiry. Wdyt?

JoshVanL: It will need to be changed on the contrib side, as [that is where it is coming from](https://github.com/dapr/components-contrib/pull/2863). It is used here in the unit test as a realistic example.

yaron2 (approved):

PR #6830, opened 2023-08-18
E2E tests: Use Postgres over Redis as default state and actor state store.

JoshVanL:
Redis is not a truly transactional database, and as such is not supported as an actor state store in production. It _also_ does not support the new [`ttlExpireTime`](https://github.com/dapr/components-contrib/issues/2857) response metadata field, and as such, is not compatible with e2e tests for that feature.

This PR switches out the state and actor state stores in e2e tests to use Postgres, which is both a proper transactional DB and supports the new `ttlExpireTime` metadata field. We are also already installing Postgres in our e2e tests so this is an easier switch.

I have kept the original Redis configs committed for prosperity.


olitomlinson:
I just want to make aware that there is a problem in that the current postgres state store implementation isn't compatible with `pgbouncer` -- a common connection pooler, used in production scenarios.

https://github.com/dapr/components-contrib/issues/3002

It is my opinion that the E2E tests should also run with pgbouncer in the chain (and not just directly against postgres) to give a more realistic, production-like E2E test scenario.

JoshVanL:
@olitomlinson would testing against pgbouncer be more appropriate in dapr/componnts-contrib? There we are more concerned with compatibility with a particular component implementation (and its flavors), whereas dapr/dapr is more concerned with the correctness of runtime.

olitomlinson:
@JoshVanL great point, equally I don't have the insights to know where is best as I'm not a maintainer.

As long as its done _somewhere_, is all I care about :)

mukundansundar:
> @olitomlinson would testing against pgbouncer be more appropriate in dapr/componnts-contrib? There we are more concerned with compatibility with a particular component implementation (and its flavors), whereas dapr/dapr is more concerned with the correctness of runtime.

I agree testing for a particular scenario/configuration should ideally be part of certification tests for Postgres in dapr/components-contrib repo.

ItalyPaleAle:
> > @olitomlinson would testing against pgbouncer be more appropriate in dapr/componnts-contrib? There we are more concerned with compatibility with a particular component implementation (and its flavors), whereas dapr/dapr is more concerned with the correctness of runtime.
> 
> I agree testing for a particular scenario/configuration should ideally be part of certification tests for Postgres in dapr/components-contrib repo.

I agree

ItalyPaleAle (commented):
Can you completely delete Redis from here?

Also, could we use Postgres as query state store too?

JoshVanL:
@ItalyPaleAle I have removed redis state, actor state, and configuration. We are currently using mongodb as the query state store- what's the reason for changing that?

ItalyPaleAle:
I think we are using MongoDB just for query state store, if I'm not mistaken. And Redis is the other query state store.

If we now have Postgres, which does support queries, we could remove MongoDB entirely. This is one less Helm chart to deploy and 3 less pods (because it requires 3 pods and they use quite a bit of CPU even when idle)

JoshVanL:
@ItalyPaleAle to be clear, you mean getting rid of both redis and mongodb entirely? 

redis is also being used as a pubsub and a pluggable component. Mongodb also looks to be used in the state and actor php apps.

ItalyPaleAle:
> to be clear, you mean getting rid of both redis and mongodb entirely?
>
> redis is also being used as a pubsub and a pluggable component. Mongodb also looks to be used in the state and actor php apps.

The general rule is that we need to test each building block with 2 services.

I don't think we can remove Redis (as the application installed via Helm) as we use that for PubSub, Configuration, and other things.

But if we switch the state query test from MongoDB to Postgres (and change the PHP test too), then we may be able to stop installing MongoDB entirely. This should make tests a bit faster (one less Helm chart to install) and should reduce some load from the K8s clusters, since MongoDB on my cluster uses 1/4 CPU even when idle.

tests/dapr_tests.mk
```diff
@@ -90,19 +90,19 @@ DAPR_CONTAINER_LOG_PATH?=./dist/container_logs
...
+DAPR_TEST_QUERY_STATE_STORE=postgres
 endif
 
 ifeq ($(DAPR_TEST_PUBSUB),)
 DAPR_TEST_PUBSUB=redis
 endif
 
 ifeq ($(DAPR_TEST_CONFIG_STORE),)
-DAPR_TEST_CONFIG_STORE=redis
+DAPR_TEST_CONFIG_STORE=postgres
```
ItalyPaleAle, line 105: Wait, we need this to be Redis. We need to test the config store with Redis, and there's another pipeline that tests with Postgres

tests/config/dapr_redis_state_badpass.yaml
ItalyPaleAle, line 1: Why was this removed? I think we need it, it was to simulate component init failure. Same for "badhost"

tests/dapr_tests.mk
JoshVanL: Done :+1:

.github/workflows/dapr-test.yml
```diff
@@ -535,9 +535,6 @@ jobs:
           make setup-test-env-kafka
           make setup-test-env-zipkin
           make setup-test-env-postgres
-          if [ "${{ env.DAPR_TEST_STATE_STORE }}" == "" ] ; then
-            make setup-test-env-mongodb
```
daixiang0, line 539: Would we clean all mongo setup:
```
root:[dapr]$ git grep setup-test-env-mongodb
.github/workflows/dapr-test.yml:539:            make setup-test-env-mongodb
.github/workflows/kind-e2e.yaml:197:        make setup-test-env-mongodb
.github/workflows/version-skew-e2e.yaml:222:        make setup-test-env-mongodb
tests/dapr_tests.mk:257:setup-3rd-party: setup-helm-init setup-test-env-redis setup-test-env-kafka setup-test-env-mongodb setup-test-env-zipkin setup-test-env-postgres
tests/dapr_tests.mk:501:setup-test-env-mongodb:
tests/dapr_tests.mk:535:setup-test-env: setup-test-env-kafka setup-test-env-redis setup-test-env-mongodb setup-test-env-postgres setup-test-env-k6 setup-test-env-zipkin setup-test-env-postgres
tests/docs/running-e2e-test.md:116:make setup-test-env-mongodb
```

tests/config/dapr_postgres_query_state.yaml
```diff
@@ -16,12 +16,14 @@ kind: Component
 metadata:
```
daixiang0, line 16: Rename to `dapr_postgres_state` for same style?

daixiang0, line 16: Why we still need a query state? Seems we have two postgres backend.

.github/workflows/dapr-test.yml
JoshVanL: We do not set this to mongodb anymore so this is not needed.

tests/config/dapr_postgres_query_state.yaml
JoshVanL: This is what we had before, and these component names are referenced in tests. I do not want to extend this PR further by also updating the test code to change the component names used.

PR #6835, opened 2023-08-21
Encrypt and Authenticate Placement Raft Network

JoshVanL:
Updates the placement service to use a SPIFFE TLS listener and dialer for the Placement raft network. Previously it was using bare TCP connections.

Placement will continue to use bare TCP connections when `--tls-enabled=false`.

Part of #5756


cmd/placement/options/options.go
```diff
@@ -77,7 +77,7 @@ func New() *Options {
 	flag.BoolVar(&opts.MetadataEnabled, "metadata-enabled", opts.MetadataEnabled, "Expose the placement tables on the healthz server")
 	flag.IntVar(&opts.ReplicationFactor, "replicationFactor", defaultReplicationFactor, "sets the replication factor for actor distribution on vnodes")
 
-	flag.StringVar(&opts.TrustDomain, "trust-domain", "cluster.local", "Trust domain for the Dapr control plane")
+	flag.StringVar(&opts.TrustDomain, "trust-domain", "localhost", "Trust domain for the Dapr control plane")
```
JoshVanL, line 80: This is to be consistent with the sentry default. It is also a more logical default as the default will be used more in self-hosted mode.

yaron2: Will this potentially break self-hosted users when upgrading to a newer version of placement that's running with older `daprd` instances?

JoshVanL: No, this flag is being introduced in v1.12 there is no available release with this option.

artursouza (approved):

Issue #6838, opened 2023-08-22
Include runtime version in Dapr placement data to ease with upgrades


ItalyPaleAle:
As we continue to make improvements to actors (especially reminders) and workflow, we are sometimes faced with the need to make changes that, while non-breaking (in the sense that don't alter the behavior observed by end users), are not backwards-compatible.

For example, assume that we wanted to change the format used to store reminders in the actor state store (today it's JSON, and we could use a more efficient format like protobuf). This change would make it impossible for older versions of the Dapr sidecar to be able to parse the reminders data, and would cause issues in clusters with heterogeneous Dapr versions. In fact, we can make Dapr "vNext" be able to read data created from "vCurrent", but there's no way for us to make "vCurrent" read data created by "vNext".

This scenario, where "vCurrent" and "vNext" run side-by-side, is very common while performing rolling upgrades, as there are always sidecars running alongside older versions. It generally is short-lived, as upgrades are eventually completed in minutes/hours, but if we don't manage this situation we will observe a variety of issues.

Because of this problem, it's really hard to make changes to actors/reminders, and workflow too once it reaches beta in the next release, as we'll have to manage clusters with heterogeneous versions.

A possible solution to this problem could be making each Dapr sidecar report its version to the Placement service and use that to ease with upgrades.

1. Every time a sidecar registers itself with Placement, it includes a constant indicating the "Actor API level". For example, sidecar A is running 1.12.0 and reports API level "10". Sidecar B is running 1.13.0 and reports API level "20". (API level "0" is reserved for Dapr <= 1.11 which does not report the version)
   - API levels can be incremented whenever needed. It may not be necessary to increment them on every Dapr release
   - Major Dapr releases increment API levels by adding 10, so we can make hotfixes if needed
3. When Placement disseminates its state, it also includes the minimum API version observed among all sidecars. In this example, it would be "10". This instructs all sidecars to behave in "compatibility mode", so when they store data they must store it in the format that can be read by 1.12 (API level "10").
4. Once all sidecars are updated, Placement will report that minimum version is "20". Sidecars can now begin storing data in the updated format and can be confident no older sidecar will try to read that.
5. At that point, Placement will refuse to accept connections from sidecars running API versions < 20 This way, there's no risk that an older sidecar will try to access actor data using an old format.

Pros:

- Allows us to make changes that are needed to improve actors and reminders, including their performance, without worrying about having an inconsistent state when older sidecars are running.
- Can be used by actors, actor reminders, and workflow.

[...]

PR #6839, opened 2023-08-22
Report Dapr Actor API level (runtime version) to Placement

ItalyPaleAle:
Implements the first bullet point of #6838 (and only the first bullet point)

- The Dapr runtime reports the actor API level to the Placement service when it connects
   - This is currently set to "10"
- This data point is currently unused and ignored by the Placement service
- In Dapr 1.13 we can use that information to perform decisions as described in #6838
- Having the data submitted by 1.12 already will be helpful to ensure smoother upgrades


daixiang0 (approved):
LGTM

Issue #6351, opened 2023-05-15
[Proposal] PubSub canary capability


olitomlinson:
## In what area(s)?

/area runtime

## Problem space

Diagnosing pubsub issues takes time, particularly when teams/individuals are unfamiliar/new to Dapr.

When diagnosing an issue with pubsub. The first thing I want to know is:

> "Is it an infrastructure / dapr configuration / dapr component issue ?"

or

>"is it a user-code issue ?"

if I could simply port-forward onto a sidecar, and submit a canary message, I can very quickly exercise the infrastructure/configuration of the sidecars and the pubsub component, to see if the canary is received on the receiving sidecar.

![Untitled (7)](https://github.com/dapr/dapr/assets/4224880/e73bf07f-9947-4048-a444-f5d22c01b9bf)

## How to publish the Canary message?

Couple of options : 

~~**1. New path segment**~~

~~Create a new path on the PubSub publish endpoint specifically for triggering a canary message. For example :~~

~~`POST http://localhost/:3500/v1.0/publish/general-purpose-pubsub/TopicA/**canary**`~~

~~_This is my least-favoured approach, when compared to option 2_~~

~~**2. Metadata property**~~

~~`POST http://localhost:3500/v1.0/publish/general-purpose-pubsub/TopicA?metadata.canary=true`~~

~~or~~

~~`POST http://localhost:3500/v1.0/publish/general-purpose-pubsub/TopicA?metadata.canary-Id=123456`~~

~~_I prefer `canary-id`, its easier to explicitly identify a single canary message among many canaries._~~

**3. Metadata `CloudEvent.Type` field**

`POST http://localhost:3500/v1.0/publish/general-purpose-pubsub/TopicA?metadata.cloudevent.type=dapr-canary`

and

`POST http://localhost:3500/v1.0/publish/general-purpose-pubsub/TopicA?metadata.cloudevent.type=dapr-canary&metadata.cloudevent.id=123456`

## How would the canary message be observed?

My gut here is to simply write to `stdout` on the receiving sidecar with enough info to identify the canary among among the other log output

`level=info msg="Canary received. Id : '23456'. component: 'general-purpose-pubsub', topic: 'TopicA'`

There may be other ways to observe the canary, such as writing to a state store, or calling a webhook, but I feel this defeats the point as it would implicate other moving parts that may or may not be in their own state of failure. Keeping it simple with less moving parts wins here IMO.

## Scope / Assumptions

[...]

PR #6851, opened 2023-08-25
[PubSub] Canary message capability PoC

olitomlinson:
# Description

This is a PoC implementation for my [Canary PubSub proposal](https://github.com/dapr/dapr/issues/6351), hence the draft PR.

Introduced the concept of a `canary message` and complimentary `canary subscriber` into PubSub. See linked issue for more details.

Areas that need to be addressed as part of the draft PR.

* [ ] General Go idiomatic syntax & patterns (This is my first ever attempt at writing Go and my background is dotnet, so I suspect there is lots to improve)
* [x] Existing unit tests pass, therefor no regression in existing behaviour [at unit test level]
* [ ] updated unit tests to cover the new behaviour
* [ ] Updated injector to accommodate canary subscribers
* [x] Update metadata endpoint to accommodate canary subscribers
* [ ] Any impact on pluggable components is accomodated

## Issue reference

https://github.com/dapr/dapr/issues/6351#issuecomment-1672268024

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing


pkg/runtime/processor/pubsub/subscribe.go
```diff
@@ -133,35 +148,72 @@ func (p *pubsub) subscriptions(ctx context.Context) ([]rtpubsub.Subscription, er
...
-	if err != nil {
-		return nil, err
+
+	// inject auto-generated canary subscriber
+	cs := p.autoGenerateCanarySubscriber
+	if cs != "" {
+		parts := strings.Split(cs, ":")
+		if len(parts) != 2 {
+			// TODO: Move this logic as early as possible in the init phase
+			panic("'autoGenerateCanarySubscriber' Should have had two parts")
```
mikeee, line 181: Should this panic here or could we potentially handle this gracefully with an error?

pkg/runtime/processor/pubsub/topics.go
```diff
@@ -167,6 +170,17 @@ func (p *pubsub) subscribeTopic(ctx context.Context, name, topic string, route c
 			return nil
 		}
 
+		if route.Canary {
+			if t, ok := cloudEvent[contribpubsub.TypeField]; ok && t == "dapr.canary" {
+				log.Infof("🐦 Canary received! -- CloudEvent Id %s", cloudEvent[contribpubsub.IDField])
```
mikeee, line 175:
```suggestion
				log.Infof("🐦 Canary received! -- CloudEvent ID %s", cloudEvent[contribpubsub.IDField])
```

mikeee (changes requested):

pkg/runtime/processor/pubsub/subscribe.go
olitomlinson: If this code path is reached, the user has misconfigured dapr, so we must instruct dapr to exit (this is why I used panic) - I'm very new to Go, so if there is a better way to do that please do let me know :)

olitomlinson:
I'm looking to get some feedback on before I start adding new Unit Tests etc

Particularly looking for some architectural feedback on this new flag that I've added https://github.com/dapr/dapr/pull/6851/files#diff-086878b4da37581eef5d27a27298349f4cc12564fdf79a9e487ca624a5f0c943R121

The reason why I'm asking is this flag is pretty niche, and wanted to check that is ok to put such a niche thing at this level (as it does create a ripple of changes throughout the app to support it, not big changes, but changes nevertheless...) there might be a better way, but I can't see it

cmd/daprd/options/options.go
```diff
@@ -117,6 +118,7 @@ func New(args []string) *Options {
 	flag.IntVar(&opts.AppHealthProbeTimeout, "app-health-probe-timeout", int(config.AppHealthConfigDefaultProbeTimeout/time.Millisecond), "Timeout for app health probes in milliseconds")
 	flag.IntVar(&opts.AppHealthThreshold, "app-health-threshold", int(config.AppHealthConfigDefaultThreshold), "Number of consecutive failures for the app to be considered unhealthy")
 	flag.StringVar(&opts.AppChannelAddress, "app-channel-address", runtime.DefaultChannelAddress, "The network address the application listens on")
+	flag.StringVar(&opts.CanarySubscriber, "canary-subscriber", "", "Auto-generates a canary subscriber. use the pattern {PubSubName}:{TopicName}")
```
yaron2, line 121: I would like to find another way to create tell the runtime to create a canary subscriber which isn't a flag. Since this is a special string format anyway (Pubsub:Topic) can we not instead use the Subscription CRD and add an explicit, strongly typed `canary: true/false` field to it?

olitomlinson: @yaron2 thanks for taking the time to check out the PR!

> can we not instead use the Subscription CRD and add an explicit, strongly typed canary: true/false field to it?

This PR implementation already exactly does this (`canary: true` on the subscription yaml)

Additionally, I've also gone a step further and provided the CLI flag `canary-subscriber` which has the same effect as someone creating an entire subscription yaml and setting its values manually. It's a pure convenience, my thinking is this provides the most accessible and least error prone method of setting up a subscription which is ready to send a canary. But it is a convenience, so technically it can be removed. Happy to remove if you would prefer!

yaron2: It's best to remove the flag and make the configuration experience simple. We also generally prefer to keep CLI flags for global runtime configs and not for individual features as much as possible.

olitomlinson: Removed the convenience CLI flag.

PR #6854, opened 2023-08-25
Change the default sentry serving port from 80 to 443

JoshVanL:
Previously in Kubernetes, the sentry service was serving on port 80. Port 443 is more appropriate, given it is a TLS server.

The service will continue to serve on port 80 to ensure backward compatibility on older clients. We should remove also serving port 80 by default in 2 releases.


mukundansundar:
hopefully once #6863 is merged and 1.11.3 is released, the version skew tests should be unblocked in terms of component setup failures

PR #6869, opened 2023-08-30
Pin redis version in version-skew setup workflow test

JoshVanL:
Fixes redis installation which is currently broken in the version-skew workflow because of an error in the latest chart. Pins to the same version as in master https://github.com/dapr/dapr/blob/cc25c181bc45733c9cefae15d551ad25b868e886/tests/dapr_tests.mk#L448

https://github.com/dapr/dapr/actions/runs/6025664924

/cc @mukundansundar

### DIFF
--- a/.github/workflows/version-skew-e2e.yaml
+++ b/.github/workflows/version-skew-e2e.yaml
@@ -217,7 +217,7 @@ jobs:
         cd latest-release
 
         make setup-helm-init
-        make setup-test-env-redis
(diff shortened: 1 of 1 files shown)


mukundansundar:
@JoshVanL ideally would prefer to use only make commands in workflow files ... 

See my comment in the PR #6854 Once the branch release-1.11 is fixed as part of the 1.11.3 patch release, the test version skew should also be fixed.


Would like opinions of other maintainers regarding this.

mukundansundar:
closing PR as pipeline in release-1.11 is fixed with release of 1.11.3 as seen in 
https://github.com/dapr/dapr/actions/runs/6041393545/job/16394265664

PR #6879, opened 2023-09-04
Docs/update

cicoyle:
Fix some grammatical errors and add READMEs to the dapr concepts folders under `cmd/<concept>`, so that newer developers have easier access to the relevant docs.


daixiang0 (approved):
LGTM

Issue #6874, opened 2023-08-31
[Workflows] 2 errors when testing workflow concurrency


olitomlinson:
**runtime version :** only tested with `1.11.2`, `1.11.3`
**workflow sdk:**  only tested with dotnet `1.11.0`

I have a docker compose set up with **2 copies of the same App deployed**.

Pretty much every time I schedule a bunch of workflows (of the same type) at the same time, I get multiples of this error : 

> dapr-workflow-examples-workflow-dapr2-1  | time="2023-08-31T20:23:31.101396718Z" level=warning msg="0-6::2::1: execution failed with a recoverable error and will be retried later: failed to invoke 'AddWorkflowEvent' method on workflow actor: rpc error: code = Internal desc = error invoke actor method: error storing reminder: error saving reminders partition and metadata: ERR user_script:14: failed to set key actors||dapr.internal.default.workflow.workflow script: d908b9553add63a82e03589b5c6d01a7654ef0f2, on @user_script:14." app_id=workflow instance=0bc74a8eb8c5 scope=dapr.runtime.wfengine type=log ver=1.11.3

For example, if I schedule 4 instances, 1 or 2 will complete instantly (expected) then the remaining will take 1-2 minutes to complete eventually (unexpected) sometimes one may never complete and just be stuck in the running state.

_Occasionally_ I get this error too, perhaps 5-10% of the time : 

> dapr-workflow-examples-workflow-dapr-1   | time="2023-08-31T20:07:45.006831377Z" level=warning msg="0-2014: execution failed with a recoverable error and will be retried later: failed to invoke activity actor '0-2014::3::1' to execute 'NotifyCompensateActivity': rpc error: code = Internal desc = error invoke actor method: error storing reminder: failed to store reminder dapr.internal.default.workflow.activity||0-2014::3::1||run-activity: reminder was created concurrently by another goroutine" app_id=workflow instance=da464a10fb2e scope=dapr.runtime.wfengine type=log ver=edge

--- 

### Repro

1. pull repo https://github.com/olitomlinson/dapr-workflow-examples
2. **switch to `kafka2` branch**
3. `docker compose build`
4. `docker compose up`
5. Use insomnia/postman/whatever :
`POST http://localhost:5112/saga?count=4&async=true&runId={runId}`<- make this request just once to start 4 workflows. **Replace {runId} with a unique string for each time you run this request, i.e. `001`, `002,`, `003` etc**
[...]

PR #6886, opened 2023-09-05
Keep retrying storing/deletng reminders

DeepanshuA:
# Description

When trying to run multiple Workflow instances (I used 25 WF instance per App Instance) in multiple App instances (used 2 App instance as of now), then even after change from https://github.com/dapr/dapr/pull/6885, workflows were getting 1. completely stuck AND 2. there were erroneous errors.

This PR on top of https://github.com/dapr/dapr/pull/6885 will ensure that Creation or Deletion of reminders happens definitely. And, thus resolves both issues:
1. Workflows were getting stuck, if Reminder was not created to schedule a workflow itself. Ensuring that reminder is created on retrying resolved this case.
2. The errors were coming as some reminders could not be deleted but the related workflow activities got completed. So, the errors like following were coming:
```
level=error msg="activity-processor: failed to process work item: invalid TaskScheduled event"
```
As deletion of reminder is ensured, this error is also resolved.

Confirmed on multiple runs that finally ALL workflow instances were getting Successfully Completed.

## Issue reference

Please reference the issue this PR (along with https://github.com/dapr/dapr/pull/6885) will close: #6874

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing


pkg/actors/reminders/reminders.go
```diff
@@ -159,9 +160,19 @@ func (r *reminders) CreateReminder(ctx context.Context, reminder *internal.Remin
 
 	stop := make(chan struct{})
 
-	err = r.storeReminder(ctx, store, reminder, stop)
+	config := retry.DefaultConfig()
```
ItalyPaleAle, line 163: Does this have a maximum attempt count?

pkg/actors/reminders/reminders.go
```diff
@@ -159,9 +160,19 @@ func (r *reminders) CreateReminder(ctx context.Context, reminder *internal.Remin
...
+	config.Multiplier = 1.0
+	b := config.NewBackOffWithContext(ctx)
+
+	err = retry.NotifyRecover(func() error {
+		return r.storeReminder(ctx, store, reminder, stop)
```
ItalyPaleAle, line 168: I wonder if we should retry only if the error is related to etags? We have a way to see if the error is an etag error (which is implemented in components-contrib and has conformance tests): https://vscode.dev/github/ItalyPaleAle/dapr/blob/master/pkg/grpc/api.go#L703-L707

pkg/actors/reminders/reminders.go
DeepanshuA: no, I skipped it intentionally to ensure that we definitely store/delete reminder

pkg/actors/reminders/reminders.go
DeepanshuA: makes sense - let me try that

pkg/actors/reminders/reminders.go
DeepanshuA: And, I kept it Multiplier as 1.0 intentionally, to ensure that workflows execution time doesn't exceed too much.

pkg/actors/reminders/reminders.go
DeepanshuA: done

pkg/actors/reminders/reminders.go
```diff
@@ -159,9 +160,37 @@ func (r *reminders) CreateReminder(ctx context.Context, reminder *internal.Remin
...
+		innerErr := r.storeReminder(ctx, store, reminder, stop)
+		if innerErr != nil {
+			if errors.As(innerErr, &etagErr) {
+				switch etagErr.Kind() {
+				case state.ETagMismatch:
+					// If the etag is mismatched, we can retry the operation.
+					return innerErr
+				default:
+					log.Errorf("error storing reminder: %w", innerErr)
+					err = innerErr
```
ItalyPaleAle, line 178: Please don't do this: it's a side effect that will cause race conditions.

Instead, please use `NotifyRecoverWIthData` to return a second error if needed.

pkg/actors/reminders/reminders.go
```diff
@@ -205,7 +234,40 @@ func (r *reminders) DeleteReminder(ctx context.Context, req internal.DeleteRemin
...
+					// we can't retry the operation.
+					return nil
+				}
+			}
+		}
+		log.Errorf("error storing reminder: %w", err)
+		err = innerErr
```
ItalyPaleAle, line 259: Here too. Please, no side effects in goroutines. Use `NotifyRecoverWIthData`

ItalyPaleAle (changes requested):

pkg/actors/reminders/reminders.go
ItalyPaleAle: Actually even better... wrap the non-etag error in a permanent error: https://pkg.go.dev/github.com/cenkalti/backoff/v4#PermanentError we do that in other parts of Dapr.

mukundansundar:
created a separate issue for tracking E2E for this change https://github.com/dapr/dapr/issues/6889

pkg/actors/reminders/reminders.go
DeepanshuA: done

pkg/actors/reminders/reminders.go
DeepanshuA: done

pkg/actors/reminders/reminders.go
```diff
@@ -159,9 +161,33 @@ func (r *reminders) CreateReminder(ctx context.Context, reminder *internal.Remin
 
 	stop := make(chan struct{})
 
-	err = r.storeReminder(ctx, store, reminder, stop)
+	config := retry.DefaultConfig()
+	config.Multiplier = 1.0
+	b := config.NewBackOffWithContext(ctx)
+
+	var etagErr *state.ETagError
```
ItalyPaleAle, line 168: This is a side effect too

pkg/actors/reminders/reminders.go
```diff
@@ -159,9 +161,33 @@ func (r *reminders) CreateReminder(ctx context.Context, reminder *internal.Remin
...
+			if errors.As(innerErr, &etagErr) {
+				switch etagErr.Kind() {
+				case state.ETagMismatch:
+					// If the etag is mismatched, we can retry the operation.
+					return innerErr
+				default:
+					log.Errorf("error storing reminder: %w", innerErr)
+					return backoff.Permanent(innerErr)
+				}
+			}
```
ItalyPaleAle, line 181: Missing a`return innerErr` here

ItalyPaleAle (approved):

artursouza (approved):

ItalyPaleAle:
@DeepanshuA we should consider changing the log from warning to info. We are seeing on Discord users that tried the nightly and are concerned with this. Maybe even mask the error message and just say that we'll retry?

Issue #6538, opened 2023-06-18
Allow components to define internal actors


ItalyPaleAle:
Note this is about extending the capabilities of existing components, not creating a new component class.

Internal actors have been added in Dapr 1.10 to enable workflows. They allow the runtime itself to act as actor host for some actor types.

It would be helpful to extend the ability to register internal actor types to components, so that components can host actors when needed to perform certain tasks.

The most immediate need for this is to enable the Rate Limiter middleware to perform rate-limiting across multiple instances of the same service. Right now, the state is kept in-memory, so if the service is scaled horizontally, each instance has to manage its own rate-limiting.

From an implementation standpoint, this could be enabled by exposing the method [`actors.RegisterInternalActor`](https://github.com/dapr/dapr/blob/6d90a8576891b7586923147338bc4d511a2fd257/pkg/actors/actors.go#L87) to components, for example adding it to the context that’s passed to the components’ `Init` method

ItalyPaleAle:
Another issue in components-contrib that could benefit from internal actors: dapr/components-contrib#2635

yaron2:
Why not extend this to all component types and also to other parts of the runtime? For example, making the entire Resiliency feature distributed (global counts for circuits, retries) is a very lucrative feature that could benefit from this (It may be that internal actors are abstracted well enough to be used already by other parts of the runtime).

ItalyPaleAle:
> Why not extend this to all component types and also to other parts of the runtime? For example, making the entire Resiliency feature distributed (global counts for circuits, retries) is a very lucrative feature that could benefit from this (It may be that internal actors are abstracted well enough to be used already by other parts of the runtime).

I think that's a good idea for things like circuits (maybe less for retries since that would require sharing the entire request's state).

The downside of this approach of relying on actors (and why I am limiting it to middleware components for now) is that it requires having an actor runtime set up. So, a compatible actor state store, a placement service running, etc.

Perhaps there could be a separate issue for other places where the actor runtime could be leveraged more?

ItalyPaleAle:
I am very interested in this issue to enable more internal actors.

- Components, especially middleware, can greatly benefit from them
   - Examples include rate limiting, OAuth2
- In the runtime, we can use them for other background tasks too, so they are coordinated even when multiple apps are running. Some examples _could_ include:
   - Workflow: cleaning up old workflow history
   - Migrations, for example if we need to change how things like reminders are stored (hypothetically). This can be done in conjunction with #6838
 
[...]

PR #6892, opened 2023-09-06
Split Actors interface betweeen runtime and internal

ItalyPaleAle:
Related to #6538

Splits the Actors interface into two parts:

- `Actors` includes methods that actors can invoke only, such as calling actors, working with timers and reminders
- `ActorsRuntime` is the main actor runtime. It embeds `Actors` and it offers additional methods that are only used internally

This change makes it possible to expose to internal actors (such as workflow) a subset of methods. It will make it easier to create more internal actors (see #6538).

Otherwise, this PR doesn't include any changes in functionality. Changes can be considered mostly cosmetic.

### DIFF
--- a/pkg/runtime/registry/registry.go
+++ b/pkg/runtime/registry/registry.go
@@ -14,7 +14,6 @@ limitations under the License.
 package registry
 
 import (
-	"github.com/dapr/dapr/pkg/actors"
 	"github.com/dapr/dapr/pkg/components/bindings"
 	"github.com/dapr/dapr/pkg/components/configuration"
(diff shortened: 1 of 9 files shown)


pkg/runtime/registry/registry.go
```diff
@@ -32,7 +31,6 @@ import (
 type ComponentsCallback func(components ComponentRegistry) error
 
 type ComponentRegistry struct {
-	Actors          actors.Actors
```
ItalyPaleAle, line 35: This was completely unused, so it was removed

mukundansundar (approved):

PR #6895, opened 2023-09-07
Adds `mtls.mountSentryVolume`, `extraVolumes`, `extraVolumeMounts` Helm Chart options

JoshVanL:
To support users who are unable to mount bound service account tokens, this PR adds the `mtls.mountSentryVolume` variable to gate creating and mounting the volume.

The injector will not mount the sentry bound service account volume on daprds if it itself does not have the token available.

To support alternative sentry authentication strategies, the `global.extraVolumes` and `global.extraVolumeMounts` helm chart options have been added to add volumes to the control plane pods. The `dapr.io/volume-mounts` annotation can be used to mount alternative credentials to dapr sidecars.


ItalyPaleAle (commented):
Thanks for this! Although LGTM so far, this is only a partial solution.

Unless I'm missing something...

The current Helm charts are offering the option to disable mounting the volume on /var/run/secrets/dapr.io/sentrytoken, but without that the control plane won't work, because there isn't a token.

We need to make it so the Helm charts can still produce a working control plane, by mounting _some_ volume.

JoshVanL:
> 

Gotcha, so if I understand correctly, you would like to see a helm values option to mount a generic volume[mount] for the control plane (and daprd ?) ?

ItalyPaleAle:
I am not sure actually. We need to provide _some_ way for users to pass their own tokens. At least that's my understanding as I am not sure the control plane would work without? (Correct me if I'm wrong here, there's been too many changes lately)

Mounting an empty volume wouldn't help much either, if we don't provide a way for them to write to that volume?

ItalyPaleAle:
Yeah I agree users are on their own, but we need to provide the "tools" for them.

How about we allow passing the name of a volume to mount? Would that work?

JoshVanL:
@ItalyPaleAle I've added options to allow mounting extra volumes and volume mounts to control plane components to expose alternative sentry authentication credentials. Do you think we should have a separate volume and volume mount definition for each control plane component, or do you think a single one for all components is fine?

ItalyPaleAle:
Since each control plane service needs a token with a different subject, it should be a different volume (or there should be sub-folders mounted for each service)

charts/dapr/values.yaml
```diff
@@ -39,8 +39,24 @@ global:
     workloadCertTTL: 24h
     allowedClockSkew: 15m
     controlPlaneTrustDomain: "cluster.local"
+    # If set to true, a bound service account token will be mounted and used to
+    # authenticate to Sentry.
+    mountSentryVolume: true
```
ItalyPaleAle, line 44: Please update the README too with all the possible values

ItalyPaleAle (commented):
LGTM but please update the README

charts/dapr/values.yaml
JoshVanL: Done, please take a look. Ideally we should be generating that README.md from the values file.

ItalyPaleAle (approved):

PR #6897, opened 2023-09-07
add to clientID tests to test both k8s and standalone mode

cicoyle:
# Description

Made sure that both the standalone and kubernetes modes are being tested for clientID. 

## Issue reference

Will close this issue: [validating Kafka](https://github.com/dapr/components-contrib/issues/2397)

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing

### DIFF
--- a/pkg/runtime/runtime_test.go
+++ b/pkg/runtime/runtime_test.go
@@ -1290,7 +1290,7 @@ func TestMetadataClientID(t *testing.T) {
 		},
 	}
 
-	// ClientID should be namespace for k8s
+	// ClientID should be namespace.AppID for Kubernetes
 	t.Run("Kubernetes Mode AppID", func(t *testing.T) {
 		t.Setenv("NAMESPACE", "test")
 		pubsubComponent.Spec.Metadata = append(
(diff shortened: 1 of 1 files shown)


pkg/runtime/runtime_test.go
```diff
@@ -1290,40 +1290,84 @@ func TestMetadataAppID(t *testing.T) {
...
-	rt.runtimeConfig.registry.PubSubs().RegisterComponent(
-		func(_ logger.Logger) pubsub.PubSub {
-			return mockPubSub
-		},
-		"mockPubSub",
-	)
+		mockPubSub.On("Init", mock.Anything).Return(nil).Run(func(args mock.Arguments) {
+			metadata := args.Get(0).(pubsub.Metadata)
+			clientID := metadata.Properties["clientID"]
+			assert.Equal(t, "test.myApp", clientID)
```
ItalyPaleAle, line 1324: Because this method is invoked on a background goroutine, we shouldn't do assertions in here. Assertions in Go should only be on the goroutine where the test is running. Additionally, the test function may return before the callback is executed.

Instead, you should:

1. Create a channel like `clientIDChan := make(chan string)` outside of this `mockPubSub.On`
2. In the callback, write the client ID to the channel: `clientIDChan <- clientID`
3. After `assert.NoError` on L1328, read from the channel and ensure the value matches, such as `assert.Equal(t, "test.myApp", <-clientIDChan)`

ItalyPaleAle (changes requested):

pkg/runtime/runtime_test.go
```diff
@@ -1290,40 +1290,103 @@ func TestMetadataAppID(t *testing.T) {
...
 			assert.Equal(t, daprt.TestRuntimeConfigID, appID)
 		}
-	})
 
-	err := rt.processComponentAndDependents(context.Background(), pubsubComponent)
-	assert.NoError(t, err)
+		select {
+		case clientID := <-clientIDChan:
+			assert.Equal(t, standAloneClientID, clientID)
+		case <-time.After(2 * time.Second):
```
ItalyPaleAle, line 1386:
```suggestion
		case <-time.After(20 * time.Second):
```

pkg/runtime/runtime_test.go
```diff
@@ -1290,40 +1290,103 @@ func TestMetadataAppID(t *testing.T) {
...
+		assert.NoError(t, err)
 
-	mockPubSub.On("Init", mock.Anything).Return(nil).Run(func(args mock.Arguments) {
-		metadata := args.Get(0).(pubsub.Metadata)
-		clientID := metadata.Properties["clientID"]
-		appIds := strings.Split(clientID, " ")
+		select {
+		case clientID := <-clientIDChan:
+			assert.Equal(t, "test.myApp", clientID)
+		case <-time.After(2 * time.Second):
```
ItalyPaleAle, line 1335:
```suggestion
		case <-time.After(20 * time.Second):
```

ItalyPaleAle (commented):
Increasing the timeout because our GitHub Actions runners are SLOW. Otherwise LGTM

ItalyPaleAle (approved):

Issue #6339, opened 2023-05-12
Error in logs: error invoke actor method: failed to cast response


IvanILanekassen:
## In what area(s)?
/area runtime

## What version of Dapr?
1.10.5

## Expected Behavior

No errors of this kind should happen

## Actual Behavior

We see a lot of Dapr.DaprApiException excptions in our logs after upgrading from 1.9.5 to 1.10.5. The exceptions have the following error message: "error invoke actor method: failed to cast response". This exact error message returns only 1 hit on google, namely the [dapr source code that returns it](https://github.com/dapr/dapr/blob/770d4e51604f1264d8bb25cedf16ea9f77539394/pkg/http/api.go#L1863). We have investigated some of the occurences, and see that the issue occurs when there is some kind of error in the actor. The error results in an exception thrown from the actor. I tried to browse the docs to see what is expected/acceptable return types from the actor, but could not find anything. Right now, our actor methods return Task (but might throw an exception). We are on v1.10.0 of the .NET SDK (Dapr.Actors.AspNetCore). So, should we not throw exceptions? How should the actors behave in the case of an error?

## Steps to Reproduce the Problem

Unsure if it is possible, but create an actor that throws an exception, and invoke this actor using the .NET SDK. This might trigger the error. 

## Release Note

RELEASE NOTE: **ADD** New documentation on error handling in actors.

nissbran:
This seems to be related to the changes made in the #5645 pull request. The policy runner sets the response to zero at https://github.com/dapr/dapr/blob/1a79c6f2de62a20290f5a399199a3fb5e31dcf2d/pkg/resiliency/policy.go#L195

This only affects the .NET SDK implementation as I understand it. Because it sends errors in the response as 200 and with a header.

akkie:
We faced the same issue and for us the root cause was that we had two actor implementations with the same name in two different services. It seems that when calling the remote actor, that Dapr calls the actor from a different service and then the above error occurs.

I know it's an error on our side. But maybe the info is usefull for others as well.

artursouza:
Does this issue also happen with 1.10.9?

onionhammer:
Yes

onionhammer:
Just got it in production today, wonder what the error was...  guess I'll never know because failed to cast response.

tmacam:
[...]

PR #6912, opened 2023-09-12
[release-1.12] Fix + E2E test to verify actor resiliency + x-error header.

artursouza:
# Description

Fix + E2E test to verify actor resiliency + x-error header.

## Issue reference

Please reference the issue this PR will close: #6339

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing

### DIFF
--- a/pkg/actors/actor_error.go
+++ b/pkg/actors/actor_error.go
@@ -1,0 +1,101 @@
+/*
+Copyright 2023 The Dapr Authors
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package actors
+
+import (
+	"errors"
+	"fmt"
+
+	"google.golang.org/grpc/codes"
+
+	invokev1 "github.com/dapr/dapr/pkg/messaging/v1"
+)
+
+type ActorError struct {
+	body        []byte
+	headers     invokev1.DaprInternalMetadata
+	contentType string
+	statusCode  int
+	message     string
+}
+
+func NewActorError(invokeResponse *invokev1.InvokeMethodResponse) *ActorError {
+	if invokeResponse == nil {
+		return &ActorError{
+			body:        []byte{},
+			headers:     nil,
+			contentType: "text/plain",
+			statusCode:  500,
+			message:     "actor error without response body",
+		}
+	}
+
+	body, err := invokeResponse.RawDataFull()
+	if err != nil {
+		return &ActorError{
+			body:        []byte{},
+			headers:     nil,
+			contentType: "text/plain",
+			statusCode:  500,
+			message:     fmt.Sprintf("could not read error's body: %s", err),
+		}
+	}
+
+	statusCode := int(invokeResponse.Status().Code)
+	if !invokeResponse.IsHTTPResponse() {
+		statusCode = invokev1.HTTPStatusFromCode(codes.Code(statusCode))
+	}
+
+	return &ActorError{
+		body:        body,
+		headers:     invokeResponse.Headers(),
+		contentType: invokeResponse.ContentType(),
+		statusCode:  statusCode,
+		message:     "actor error with details in body",
+	}
+}
+
+func (e *ActorError) Error() string {
+	return e.message
+}
+
+func (e *ActorError) Headers() invokev1.DaprInternalMetadata {
+	return e.headers
+}
+
+func (e *ActorError) ContentType() string {
+	return e.contentType
+}
+
+func (e *ActorError) StatusCode() int {
+	return e.statusCode
+}
+
+func (e *ActorError) Body() []byte {
+	return e.body
+}
+
+func AsActorError(err error) (*ActorError, bool) {
+	var actorError *ActorError
+	if errors.As(err, &actorError) {
+		return actorError, true
+	}
+
+	return nil, false
+}
+
+func IsActorError(err error) bool {
+	var actorError *ActorError
+	return errors.As(err, &actorError)
+}
--- a/pkg/actors/actor_error.go
+++ b/pkg/actors/actor_error.go
@@ -22,6 +22,7 @@ import (
 	invokev1 "github.com/dapr/dapr/pkg/messaging/v1"
 )
 
+// ActorError is an error returned by an Actor via a header + body in the method's response.
 type ActorError struct {
 	body        []byte
 	headers     invokev1.DaprInternalMetadata
@@ -32,12 +33,12 @@ type ActorError struct {
 
 func NewActorError(invokeResponse *invokev1.InvokeMethodResponse) error {
 	if invokeResponse == nil {
-		return fmt.Errorf("coult not parse actor error: no response object")
+		return fmt.Errorf("could not parse actor error: no response object")
 	}
 
 	body, err := invokeResponse.RawDataFull()
 	if err != nil {
-		return fmt.Errorf("coult not read actor error: %s", err)
+		return fmt.Errorf("could not read actor error: %s", err)
 	}
 
 	statusCode := int(invokeResponse.Status().Code)
--- a/pkg/actors/actor_error.go
+++ b/pkg/actors/actor_error.go
@@ -30,26 +30,14 @@ type ActorError struct {
 	message     string
 }
 
-func NewActorError(invokeResponse *invokev1.InvokeMethodResponse) *ActorError {
+func NewActorError(invokeResponse *invokev1.InvokeMethodResponse) error {
 	if invokeResponse == nil {
-		return &ActorError{
-			body:        []byte{},
-			headers:     nil,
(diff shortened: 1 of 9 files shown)


artursouza:
First, validating the E2E test actually reproduces the error.

pkg/http/api.go
```diff
@@ -1323,42 +1329,54 @@ func (a *api) onDirectActorMessage(reqCtx *fasthttp.RequestCtx) {
 	policyRunner := resiliency.NewRunnerWithOptions(reqCtx, policyDef,
 		resiliency.RunnerOpts[*invokev1.InvokeMethodResponse]{
 			Disposer: resiliency.DisposerCloser[*invokev1.InvokeMethodResponse],
+			Accumulator: func(r *invokev1.InvokeMethodResponse) {
+				if r == nil {
+					return
+				}
```
ItalyPaleAle, line 1335: This is not necessary because the accumulator is invoked only if r is not zero (in this case, non-nil)

pkg/grpc/api.go
```diff
@@ -1219,26 +1221,32 @@ func (a *api) InvokeActor(ctx context.Context, in *runtimev1pb.InvokeActorReques
...
+		defer resp.Close()
+	}
+
 	if err != nil && !errors.Is(err, actors.ErrDaprResponseHeader) {
 		err = status.Errorf(codes.Internal, messages.ErrActorInvoke, err)
 		apiServerLogger.Debug(err)
 		return response, err
 	}
 
-	if resp == nil {
```
ItalyPaleAle, line 1233: The problem is that here you are changing the behavior in case the response object is actually nil. This can happen when using the .NET SDK, but can happen in other cases too, and now the error is swallowed.

IMHO this fix should be applied in `pkg/actors` where we change how the actor is invoked. Sometime over the course of the last months the code that checked for the header included by the .NET SDK was changed, and now isn't checked anymore. Instead, it should return `ErrDaprResponseHeader` like you see being checked above

artursouza:
As expected, both new E2E test cases now fail, remote and local: https://github.com/dapr/dapr/actions/runs/6178949540/job/16773055827?pr=6912

pkg/actors/actor_error.go
```diff
@@ -0,0 +1,101 @@
...
+
+type ActorError struct {
+	body        []byte
+	headers     invokev1.DaprInternalMetadata
+	contentType string
+	statusCode  int
+	message     string
+}
+
+func NewActorError(invokeResponse *invokev1.InvokeMethodResponse) *ActorError {
```
ItalyPaleAle, line 33: I think this should return `error` so you can return other errors in case they are not ActorError. See my other comments

pkg/actors/actor_error.go
```diff
@@ -0,0 +1,101 @@
...
+
+	body, err := invokeResponse.RawDataFull()
+	if err != nil {
+		return &ActorError{
+			body:        []byte{},
+			headers:     nil,
+			contentType: "text/plain",
+			statusCode:  500,
+			message:     fmt.Sprintf("could not read error's body: %s", err),
+		}
```
ItalyPaleAle, line 52: If the method returns an `error`, here you can return a non-ActorError to indicate an ISE.

```suggestion
		return fmt.Errorf("could not read actor error's body: %w", err)
```

pkg/actors/actor_error.go
artursouza: I like the idea. Done.

pkg/actors/actor_error.go
artursouza: Fixed.

pkg/actors/actor_error.go
```diff
@@ -0,0 +1,89 @@
...
+	body        []byte
+	headers     invokev1.DaprInternalMetadata
+	contentType string
+	statusCode  int
+	message     string
+}
+
+func NewActorError(invokeResponse *invokev1.InvokeMethodResponse) error {
+	if invokeResponse == nil {
+		return fmt.Errorf("coult not parse actor error: no response object")
```
JoshVanL, line 35:
```suggestion
		return errors.New("could not parse actor error: no response object")
```

pkg/actors/actor_error.go
```diff
@@ -0,0 +1,89 @@
...
+
+import (
+	"errors"
+	"fmt"
+
+	"google.golang.org/grpc/codes"
+
+	invokev1 "github.com/dapr/dapr/pkg/messaging/v1"
+)
+
```
JoshVanL, line 24: Consider adding a comment describing what this type is.

Also consider removing the `Actor` prefix from all type and func names since we have context that it is in the`actors` package. This also prevents stutter (`actors.ActorError` vs `actors.Error`). Even better to move this file into a new package `pkg/actors/errors` to separate concerns and prevent future circular imports.

pkg/actors/actor_error.go
```diff
@@ -0,0 +1,89 @@
...
+}
+
+func NewActorError(invokeResponse *invokev1.InvokeMethodResponse) error {
+	if invokeResponse == nil {
+		return fmt.Errorf("coult not parse actor error: no response object")
+	}
+
+	body, err := invokeResponse.RawDataFull()
+	if err != nil {
+		return fmt.Errorf("coult not read actor error: %s", err)
```
JoshVanL, line 40:
```suggestion
		return fmt.Errorf("could not read actor error: %w", err)
```

pkg/actors/actor_error.go
```diff
@@ -0,0 +1,89 @@
...
+	if !invokeResponse.IsHTTPResponse() {
+		statusCode = invokev1.HTTPStatusFromCode(codes.Code(statusCode))
+	}
+
+	return &ActorError{
+		body:        body,
+		headers:     invokeResponse.Headers(),
+		contentType: invokeResponse.ContentType(),
+		statusCode:  statusCode,
+		message:     "actor error with details in body",
```
JoshVanL, line 53: This reads somewhat confusing to me. Does this mean that the body contains more information about why there is an error? If so, it might read better `see body for details`. We know this is an actor error from the context.

pkg/actors/actor_error.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+}
+
+func (e *ActorError) Error() string {
+	return e.message
+}
```
JoshVanL, line 59: Consider adding a unit test file with the below as a sanity check to ensure that `*ActorError` implements the error interface.

```go
func TestActorError(*testing.T) {
  var _ error = new(ActorError)
}
```

pkg/actors/actor_error.go
```diff
@@ -0,0 +1,89 @@
...
+
+func AsActorError(err error) (*ActorError, bool) {
+	var actorError *ActorError
+	if errors.As(err, &actorError) {
+		return actorError, true
+	}
+
+	return nil, false
+}
```
JoshVanL, line 84: If moved to a dedicated `actors/errors`package, this function could be renamed to read more naturally to the stdlib, i.e. 

```go
import (
  actorserrors "github.com/dapr/dapr/pkg/actors/errors"
)
....
actorErr, ok := actorserrors.As(err)
```

pkg/actors/actor_error.go
```diff
@@ -0,0 +1,89 @@
...
+
+func IsActorError(err error) bool {
```
JoshVanL, line 86: Same as above, re: moving to a dedicated package.

pkg/http/api.go
```diff
@@ -1341,7 +1359,6 @@ func (a *api) onDirectActorMessage(reqCtx *fasthttp.RequestCtx) {
 		log.Debug(msg)
 		return
 	}
-	defer resp.Close()
```
JoshVanL, line 1344: Why is this line removed?

JoshVanL (changes requested):
Some comments from me :slightly_smiling_face:

pkg/actors/actor_error.go
artursouza: We need a message to be printed at the Error() function to comply with the Error interface.

pkg/http/api.go
artursouza: It was moved up, to make sure it is closed even in case of error.

pkg/actors/actor_error.go
artursouza: It is not needed anymore because NewActorError returns `error` and the compiler will check for that.

pkg/actors/actor_error.go
artursouza: I would like to keep ActorError. I am moving to the new package and renaming other things.

pkg/actors/actor_error.go
artursouza: done.

pkg/actors/actor_error.go
artursouza: done.

pkg/actors/errors/actor_error.go
```diff
@@ -0,0 +1,90 @@
...
+}
+
+func NewActorError(invokeResponse *invokev1.InvokeMethodResponse) error {
+	if invokeResponse == nil {
+		return fmt.Errorf("could not parse actor error: no response object")
+	}
+
+	body, err := invokeResponse.RawDataFull()
+	if err != nil {
+		return fmt.Errorf("could not read actor error: %s", err)
```
ItalyPaleAle, line 41: Nit:
```suggestion
		return fmt.Errorf("could not read actor error: %w", err)
```

ItalyPaleAle (approved):

mukundansundar (approved):

PR #6917, opened 2023-09-13
Placement: reduce verbosity of a debug log

ItalyPaleAle:
When running placement with debug logs, line `Process Raft state command: nothing happened...` is getting logged every 10s, causing a lot of unnecessary noise that makes debugging harder.


pkg/placement/membership.go
```diff
@@ -264,7 +264,7 @@ func (p *Service) processRaftStateCommand(ctx context.Context) {
 			return
 
 		case <-processingTicker.C():
-			log.Debugf("Process Raft state command: nothing happened...")
+			// log.Debugf("Process Raft state command: nothing happened...")
```
JoshVanL, line 267: I agree with the change though I think this line should be deleted completely, or replaced with a proper code comment.

JoshVanL (changes requested):

pkg/placement/membership.go
ItalyPaleAle: Thinking about it, i don't think we even need the `processingTicker`. Its sole purpose was to show this log...

ItalyPaleAle: Removed the ticker entirely

JoshVanL (approved):

mukundansundar (approved):
we had a ticker only for logging?

ItalyPaleAle:
> we had a ticker only for logging?

Looks like that was the case, yes

PR #6904, opened 2023-09-08
workflows HTTP - obfuscate metrics

famarting:
# Description

Implementation for https://github.com/dapr/dapr/issues/6903

Adds logic to `convertPathToMetricLabel` to handle the workflows http API

NOTE: I decided to not obfuscate `workflowComponentName` which will be `dapr` in all cases for some time and also I haven't obfuscated `workflowName` because I think the workflow name is useful for observability and I think likely users may not use a huge amount of different workflow names but if there are different opinions I'm happy to change this.

## Issue reference

Please reference the issue this PR will close: #6903 

## Checklist

Please make sure you've  completed the relevant tasks for this PR, out of the following list:

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Unit tests passing
* [ ] End-to-end tests passing
* [ ] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_
* [ ] Specification has been updated / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_
* [ ] Provided sample for the feature / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_

PR #6919, opened 2023-09-13
Reduce cardinality in Dapr metrics and add more information to API logs

### LINKED ISSUE #6723: Reduce cardinality in Dapr metrics and add more information to API logs

ItalyPaleAle:
Fixes #6723

As requested by the maintainers, re-opening #6729 now that the merge window for 1.13 is open. It also complies with the "rule" of having refactoring PRs merged early in the release cycle.

The main goal of this PR is to fix #6723 and do that in a "tidy" way.

- Things are currently "scattered" a bit all over the place. For example, code that extracts metrics needs to have specific rules defined for each endpoint in pkg/diagnostics, and that means that very few building blocks have "proper" metrics.
- The main refactoring part is changing the `Endpoint` struct used by the HTTP server to include more fields, so things are done more declaratively in the pkg/http package, alongside the rule definition. 
- For example, this PR would have made #6904 redundant as it would have included that already, but without the need to define specific rules for the building block in pkg/diagnostics.

The main refactoring done above is the reason why the PR is on the larger side, as the changes are all interrelated.

### Details of changes

1. Drastically reduce cardinality of metrics, especially those emitted by the HTTP server:
   - Also removes the possibility of PII being included in the metrics endpoint.
1. Add more information to API logs for both HTTP and gRPC, including things that are not going to be included in metrics by default anymore:
   - Response status code
   - Response size (for HTTP only & if possible)
   - Latency
3. When `obfuscateURLs` is enabled for API logging, in API logs we now include a more descriptive name for the method rather than the path that was matched in the router. The descriptive names map to the names of the methods in gRPC, for example `SaveState` in place of `POST /state/{storeName}/{name}`. Since gRPC API logs are always, only "obfuscated" (because the params are in the body and not in the "URL"), this makes HTTP and gRPC API logs more consistent too
4. Refactors how tracing and API logging (and the API token auth middleware) get the data from the handler/router:
   - The new approach is a lot more declarative, and less based on heuristics (such as parsing the path from the URL again)
   - The new approach also reduces the number of maps that are allocated and used in each request, which generally contained duplicate information generated in multiple parts of the code
5. For both HTTP and gRPC, the way metadata is added to a tracing span has changed and it's now more declarative:
   - When looking at how tracing spans were composed, the metadata was added in `pkg/diagnostics`, separately from the endpoints. Only a VERY SMALL number of APIs had proper tracing configured (as a matter of fact, see the number of "TODO"'s in this PR), in large part due to the fact that this was in a separate package. The approach also relied a lot on heuristics.
   - For HTTP, now the `Endpoint` struct contains a property to add a function that adds properties to a span for tracing purposes. This lives right next to the handler.
   - For gRPC, messages defined in the protos whose name ends in "Request" now must define an `AppendSpanAttribute` method (this is enforced by a unit test)
      - Note that the SDKs importing the updated protos will need to define new types. However, what matters is that the messages are wire-compatible, so clients using old protos (old SDKs) won't see any change.
6. Update API Allowlisting/Denylisting for HTTP to:
   - Make sure that the same constants can be used for both HTTP and gRPC, especially versions. Right now, versions are in the format "v1.0-alpha1" for HTTP, and "v1alpha1" for gRPC. This PR changes the HTTP versions to be in the format "v1alpha1" too ("v1.0-alpha1" is preserved for bacwkards-compatibility)
   - Improved perf of the HTTP Allowlist/Denylist, especially when using the "new" format (e.g. versions with "v1alpha1"): checking the allowlist is now a simple map lookup rather than an iteration over the entire allowlist.


pkg/diagnostics/consts/consts.go
```diff
@@ -0,0 +1,68 @@
...
+	DBNameSpanAttributeKey                   = string(semconv.DBNameKey)
+	DBStatementSpanAttributeKey              = string(semconv.DBStatementKey)
+	DBConnectionStringSpanAttributeKey       = string(semconv.DBConnectionStringKey)
+	MessagingSystemSpanAttributeKey          = string(semconv.MessagingSystemKey)
+	MessagingDestinationSpanAttributeKey     = string(semconv.MessagingDestinationKey)
+	MessagingDestinationKindSpanAttributeKey = string(semconv.MessagingDestinationKindKey)
+	GrpcServiceSpanAttributeKey              = string(semconv.RPCServiceKey)
+	NetPeerNameSpanAttributeKey              = string(semconv.NetPeerNameKey)
+
+	DaprAPISpanAttributeKey           = "dapr.api"
```
JoshVanL, line 39: Question: Why are these values not in the form of `io.dapr.x` or `x.dapr.io`

JoshVanL (changes requested):

pkg/channel/http/http_channel.go
```diff
@@ -228,7 +228,7 @@ func (h *Channel) invokeMethodV1(ctx context.Context, req *invokev1.InvokeMethod
 	}()
 
 	// Emit metric when request is sent
-	diag.DefaultHTTPMonitoring.ClientRequestStarted(ctx, channelReq.Method, req.Message().Method, int64(len(req.Message().Data.GetValue())))
+	diag.DefaultHTTPMonitoring.ClientRequestStarted(ctx, int64(len(req.Message().Data.GetValue())))
```
artursouza, line 231: These are breaking changes since it removes information others might already be depending on for their dashboards and alerts. Is there a way to make this configurable?

artursouza (changes requested):
I did not review the entire thing since I think there is a more fundamental decision to be made here. Do we accept breaking changes to metrics? I believe those are part of our contract and people depend on data being published in those.

pkg/channel/http/http_channel.go
yaron2: We should remain consistent with industry standard golden metrics - meaning, latency and status codes should be reported on response metrics. Size I think we can remove safely

pkg/diagnostics/consts/consts.go
ItalyPaleAle: I don’t know - this didn’t change

pkg/channel/http/http_channel.go
ItalyPaleAle: Latency and status codes are included. Size is included when available, but if it can be omitted it would allow for a lot of simplifications.

yaron2: Are latency and status codes included by default? I want to clarify due to this:

```
including things that are not going to be included in metrics by default anymore:
* Response status code
* Response size (for HTTP only & if possible)
* Latency
```

ItalyPaleAle: I wrote that a while ago so I had to go double-check :)

Yes, they are still included (size is included only if available):

https://github.com/ItalyPaleAle/dapr/blob/38b09613667e88e5fe84f48f9f0188b2ce090219/pkg/diagnostics/http_monitoring.go#L208-L240

What I think I meant with that is that they are not included anymore with the same granularity. Users can rely on the API logs to get more specific data points, such as per-endpoint (rather than just per target app)

pkg/http/server.go
```diff
@@ -163,6 +167,7 @@ func (s *server) StartNonBlocking() error {
 	// Start the public HTTP server
 	if s.config.PublicPort != nil {
 		publicR := s.getRouter()
+		s.useContextSetup(publicR)
 		s.useTracing(publicR)
 		s.useMetrics(publicR)
```
ItalyPaleAle, line 172: This has been kept for backwards-compat, and it was checked by an E2E test, but should the public API server (used by K8s healthchecks primarily) still use metrics and tracing?

Not sure if it's needed to collect metrics on that. And it may be odd to have metrics for the "regular" HTTP server intermixed with metrics from the public HTTP server

pkg/channel/http/http_channel.go
mukundansundar: Is there a way to include back the original granular metrics if needed? I.e. can either this change be the new normal and the old way be behind an opt-in flag or the other way around?

pkg/config/configuration.go
```diff
@@ -156,18 +156,18 @@ const (
...
+// The result is a map where the key is in the format "<version>/<endpoint>"
+func (r APIAccessRules) GetRulesByProtocol(protocol APIAccessRuleProtocol) map[string]struct{} {
+	res := make(map[string]struct{}, len(r))
 	for _, v := range r {
 		//nolint:gocritic
 		if strings.ToLower(string(v.Protocol)) == string(protocol) {
-			res[n] = v
-			n++
+			key := v.Version + "/" + v.Name
+			res[key] = struct{}{}
```
mukundansundar, line 167: please specifically add unit tests for these changes.

mukundansundar (commented):
I have not completely reviewed the PR yet. But left some initial comments.

dapr/proto/runtime/v1/dapr.proto
```diff
@@ -600,7 +598,12 @@ message InvokeActorResponse {
   bytes data = 1;
 }
 
-// GetMetadataResponse is a message that is returned on GetMetadata rpc call
+// GetMetadataRequest is the message for the GetMetadata request.
+message GetMetadataRequest {
```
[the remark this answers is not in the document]
ItalyPaleAle: Yes, but:

1. This will not break apps that are currently running, which is the most important thing. Users may need to change a struct name if they import the protos directly and are updating the protos.
2. Using the gRPC protos directly is not an officially supported scenario.

pkg/channel/http/http_channel.go
ItalyPaleAle: I can look into that, but before we invest engineering effort, is there any demand for that? The current behavior is causing massive memory consumption (hundreds of MBs), causing live site incidents due to pods OOM'ing (we actually had users opening support tickets with Azure too... and seems Diagrid heard users having issues too). It is also costing users a lot of money when metrics are ingested in analytics tools. To the point it may even be considered a bug?

ItalyPaleAle: Per conversation on Discord, we can move forward on this PR without the flag to add the old behavior back as opt-in, as there's already enough "substance" here. When this is merged I'll open an issue to discuss that need (see if folks from the community have opinions) and also how that can be implemented (e.g. should it be a permanent option in the Configuration CRD, a feature flag, or...?). That issue will be completed before 1.13 (as a release blocker, but we have lots of time)

dapr/proto/runtime/v1/dapr.proto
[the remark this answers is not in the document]
mukundansundar: > Using the gRPC protos directly is not an officially supported scenario.

Still something that users might use. Again, my point being here is that it must be communicated as a breaking change. We can expand on it saying that running apps will not be affected, just struct change as of now for new proto imports.

pkg/config/configuration.go
ItalyPaleAle: Done

RyanLettieri (changes requested):
First pass through, just some syntax and tidying up suggestions

pkg/diagnostics/http_tracing.go
```diff
@@ -182,104 +184,20 @@ func tracestateToHeader(sc trace.SpanContext, setHeader func(string, string)) {
...
-		targetID := rw.UserValueString("id")
-		m[netPeerNameSpanAttributeKey] = targetID
-		m[daprAPISpanNameInternal] = "CallLocal/" + targetID + "/" + rw.UserValueString("method")
-
-	case "publish":
-		m[messagingSystemSpanAttributeKey] = pubsubBuildingBlockType
-		m[messagingDestinationSpanAttributeKey] = rw.UserValueString("topic")
-		m[messagingDestinationKindSpanAttributeKey] = messagingDestinationTopicKind
+	// Init with a worst-case initial capacity
+	m := make(map[string]string, 7)
```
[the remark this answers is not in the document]
ItalyPaleAle: I don't think it needs a variable because it's something used by the compiler only, but it does need a clearer comment.

Essentially, by doing `make(map[string]string, 7)` we are telling the compiler to allocate the map with an initial capacity for (up to) 7 items. If we don't pass it explicitly, the compiler will allocate an initial capacity as it sees fit. Because in the code below we do not store more than 7 unique keys in the map, this makes it possible to avoid re-allocations (which happen when the map was allocated with a smaller capacity than needed). Re-allocations take time and also add more work on the garbage collector.

The number 7 is just counting the number of unique keys used. Nothing really "magic" about it, and setting it to 6 or 8 would not cause anything to break (although could cause inefficiencies).

I'll add a comment in a bit to clarify.

pkg/grpc/server.go
```diff
@@ -206,9 +208,9 @@ func (s *server) Close() error {
 	return nil
 }
 
-func (s *server) getMiddlewareOptions() []grpc.ServerOption {
-	intr := make([]grpc.UnaryServerInterceptor, 0, 6)
-	intrStream := make([]grpc.StreamServerInterceptor, 0, 5)
+func (s *server) getMiddlewareOptions() []grpcGo.ServerOption {
+	intr := make([]grpcGo.UnaryServerInterceptor, 0, 6)
```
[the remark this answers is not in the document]
ItalyPaleAle: See my comment on the other thing. I don't think a variable would help because this should ideally be interpreted by the compiler. But I will add a comment to explain.

pkg/http/api_lock.go
```diff
@@ -18,25 +18,41 @@ import (
...
 			Route:   "lock/{storeName}",
 			Version: apiVersionV1alpha1,
+			Group: &endpoints.EndpointGroup{
+				Name:                 endpoints.EndpointGroupLock,
+				Version:              endpoints.EndpointGroupVersion1alpha1,
+				AppendSpanAttributes: nil, // TODO
+			},
 			Handler: a.onTryLockAlpha1(),
+			Settings: endpoints.EndpointSettings{
+				Name: "TrytLock",
```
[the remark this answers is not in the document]
ItalyPaleAle: Good catch, will fix

pkg/diagnostics/http_tracing.go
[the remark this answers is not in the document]
RyanLettieri: Okay, adding a comment is sufficient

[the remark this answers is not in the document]
ItalyPaleAle: Done

dapr/proto/runtime/v1/dapr.proto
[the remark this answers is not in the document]
yaron2: This is a breaking change, however this will be breaking at compilation time and not at runtime. I recommend we announce the breaking change with no need to create a v2 of this method.

yaron2:
Since it is difficult to know at this point if some labels need to be reinstated by default, I suggest we merge this PR so we can audit the new metrics carefully and then revert if/whatever is needed (by rolling forward, not reverting the PR).

As mentioned, an option will be introduced later to enable the more granular metrics we had until now, and that option might be used to choose if we want non-granular or granular metrics by default.

cc @mukundansundar @artursouza @daixiang0

artursouza:
> Since it is difficult to know at this point if some labels need to be reinstated by default, I suggest we merge this PR so we can audit the new metrics carefully and then revert if/whatever is needed (by rolling forward, not reverting the PR).
> 
> 
> 
> As mentioned, an option will be introduced later to enable the more granular metrics we had until now, and that option might be used to choose if we want non-granular or granular metrics by default.
> 
> 
> 
> cc @mukundansundar @artursouza @daixiang0 

Yes. We need the nightly build to work too so we can test with an immutable artifact.

dapr/proto/runtime/v1/dapr.proto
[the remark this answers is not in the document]
ItalyPaleAle: >  This is a breaking change, however this will be breaking at compilation time and not at runtime. I recommend we announce the breaking change with no need to create a v2 of this method.

This is a fair assessment.

To recap, and to be fully clear, from a practical effect this change is similar to what happened with #6945 (landing in 1.13 too) which removed a method from the gRPC proto (of course, that was done for different reasons).

- No impact on apps currently running
- Dapr SDKs maintainers will need to rename the structs/objects when they upgrade the protos
- Users of the SDKs should not see any direct impact.
  - AFAIK the SDKs always abstract the gRPC objects, so users who just depend on Dapr SDKs shouldn't need any code change if/when they upgrade the SDKs.
  - At runtime, old SDKs will continue to work alongside new SDKs.
- For users who import the protos directly:
   - No change needed if they are using existing protos
   - If they upgrade the protos, they will need to rename the structs/objects

artursouza:
I was thinking about this and it is an opportunity for us to increase test coverage. Before merging this PR, I think there should be a PR that adds functional tests for metrics from the current implementation. Then this PR can be merged with higher guarantees of not adding a regression. 

This was done by Josh when he added upgrade/downgrade tests and integration tests to increase confidence in the refactoring and changes he wanted to merge.

ItalyPaleAle:
> Before merging this PR, I think there should be a PR that adds functional tests for metrics from the current implementation.

What test would you like to see?

We currently have E2E tests that check that metrics are sent to Prometheus. Because of the dependency on an external service (Prometheus), testing metrics is probably something that is not well suited for an IT.

If you have suggestions on additional E2E tests, let me know.

ItalyPaleAle:
@artursouza I've added an IT that validates metrics are sent correctly.

For tracing, we have an existing E2E test. For traces, it's not going to be possible to create an IT as that does require a service (Zipkin). In fact, for metrics we can scrape them in a "pull" way, but traces are pushed to Zipkin so they require a service.

Should cover the tests?

tests/integration/framework/binary/binary.go
```diff
@@ -60,7 +60,7 @@ func Build(t *testing.T, name string) {
 
 		t.Logf("Root dir: %q", rootDir)
 		t.Logf("Compiling %q binary to: %q", name, binPath)
-		cmd := exec.Command("go", "build", "-tags=allcomponents", "-v", "-o", binPath, "./cmd/"+name)
+		cmd := exec.Command("go", "build", "-tags=stablecomponents", "-v", "-o", binPath, "./cmd/"+name)
```
ItalyPaleAle, line 63:
```suggestion
		cmd := exec.Command("go", "build", "-tags=allcomponents", "-v", "-o", binPath, "./cmd/"+name)
```

Revert this

daixiang0:
> # Dapr SDK JS test
> 🔗 **[Link to Action run](https://github.com/dapr/dapr/actions/runs/6592674982)**
> 
> Commit ref: [7da569d](https://github.com/dapr/dapr/commit/7da569de8d32e38339872285df01749b1c8b4f11)
> 
> ## ❌ JS SDK tests failed
> Please check the logs for details on the error.

Is it still a network issue?

ItalyPaleAle:
> > # Dapr SDK JS test
> > 🔗 **[Link to Action run](https://github.com/dapr/dapr/actions/runs/6592674982)**
> > Commit ref: [7da569d](https://github.com/dapr/dapr/commit/7da569de8d32e38339872285df01749b1c8b4f11)
> > ## ❌ JS SDK tests failed
> > Please check the logs for details on the error.
> 
> Is it still a network issue?

Looks like the tests are broken. Here's the run from master: https://github.com/dapr/dapr/actions/runs/6620638088

yaron2 (approved):

pkg/channel/http/http_channel.go
mukundansundar: @ItalyPaleAle is there an issue open to track the addition of flag to include the original granular metrics.

filintod: any idea where this issue or flag is implemented?

PR #6921, opened 2023-09-14
add unit test to cmd/placement/options

MregXN:
# Description
Test whether each variable is filled with default value. This test is not covered so the coverage of this file is only 25% currently.


JoshVanL:
Thanks @MregXN!

JoshVanL (approved):

mukundansundar (approved):

PR #6925, opened 2023-09-14
Fixes in control plane services CLI flags

ItalyPaleAle:
- Removed flags that are deprecated and have no effect
   - Unlike daprd where we need to keep flags or older injectors can cause daprd to crash, this isn't a problem for control plane services since they're deployed by Helm
- Remove `tokenAudience` flag from Helm chart since it's now unused
- Fixed description of some flags
- Use constants for default trust anchors file path


cmd/placement/options/options.go
```diff
@@ -62,8 +63,6 @@ type Options struct {
 	Metrics *metrics.Options
 }
 
-var log = logger.NewLogger("dapr.placement.options")
```
daixiang0, line 65: No log prefix any more?

ItalyPaleAle: There’s nothing writing logs now

daixiang0:
Close this in favor of #7029

ItalyPaleAle:
Per @mukundansundar this should be merged first

daixiang0 (dismissed):
LGTM

cmd/sentry/main.go
```diff
@@ -48,10 +48,6 @@ func main() {
 
 	metricsExporter := metrics.NewExporterWithOptions(log, metrics.DefaultMetricNamespace, opts.Metrics)
 
-	if len(opts.TokenAudience) > 0 {
-		log.Warn("--token-audience is deprecated and will be removed in Dapr v1.14")
```
mukundansundar, line 52: Why is `token-audience` being removed now if it is planned to be removed in 1.14?

cmd/operator/options/options.go
```diff
@@ -72,12 +73,7 @@ func New() *Options {
 	flag.BoolVar(&opts.EnableArgoRolloutServiceReconciler, "enable-argo-rollout-service-reconciler", false, "Enable the service reconciler for Dapr-enabled Argo Rollouts")
 	flag.BoolVar(&opts.WatchdogCanPatchPodLabels, "watchdog-can-patch-pod-labels", false, "Allow watchdog to patch pod labels to set pods with sidecar present")
 
-	flag.StringVar(&opts.TrustAnchorsFile, "trust-anchors-file", "/var/run/secrets/dapr.io/tls/ca.crt", "Path to trust anchors file")
-
-	depCCP := flag.String("certchain", "", "DEPRECATED")
-	depRCF := flag.String("issuer-ca-filename", "", "DEPRECATED")
-	depICF := flag.String("issuer-certificate-filename", "", "DEPRECATED")
-	depIKF := flag.String("issuer-key-filename", "", "DEPRECATED")
+	flag.StringVar(&opts.TrustAnchorsFile, "trust-anchors-file", securityConsts.ControlPlaneDefaultTrustAnchorsPath, "Filepath to the trust anchors for the Dapr control plane")
```
mukundansundar, line 77: similarly shouldn't this be removed only in 1.14 following the dapr deprecation policy?

mukundansundar (changes requested):

cmd/operator/options/options.go
ItalyPaleAle: These flags are unused today. This is a K8s service that is only deployed via Helm, so there's no risk of removing flags.

cmd/sentry/main.go
ItalyPaleAle: It's completely unused in this service

ItalyPaleAle: What I mean is, setting this option today just prints a warning and is a no-op. Should never have said that will be removed, and should have just been removed. This is a control plane service so removing CLI flags is not an issue

ItalyPaleAle:
@mukundansundar @daixiang0 is this ready to be merged?

JoshVanL (approved):

mukundansundar (approved):

PR #6937, opened 2023-09-19
Use Namespace from AdmissionReview rather than Pod

JoshVanL:
It is not always the case that the `metadata.Nameapace` of a submitted object to an admission review is populated, and so the injector must use the namespace of the admission request itself when requesting for the Pod's identity certificate from sentry.

See https://github.com/kubernetes/website/issues/30574


JoshVanL:
Closed in favor of https://github.com/dapr/dapr/pull/6938

PR #6943, opened 2023-09-19
Use Redis as actor store in tests [Do-Not-Merge]

yaron2:
This is to investigate the performance regressions observed in master and the 1.12 release branch with regards to actor reminder registration.


ItalyPaleAle:
I believe we have the data we need now. @yaron2  can we close this?

Issue #6949, opened 2023-09-20
Sidecar not working when mtls is false


aaguilartablada:
## In what area(s)?

/area runtime

## What version of Dapr?

1.12.0-rc.3

## Expected Behavior

We have LinkerD and Dapr in our platform. We decided to use LinkerD to perform MTLS, so we deployed Dapr 1.10.5 disabling mtls and it worked fine. 

Now, when deploying dapr operator 1.12.0-rc.3 with mtls disabled it does not work:

```
global:
  mtls:
    enabled: false
```

I expect DAPR sidecar containers starting correctly the same way it started when using version 1.10.5.

## Actual Behavior

DAPR sidecar containers never start. Instead, they report the next log:

```
time="2023-09-20T14:49:27.187840991Z" level=info msg="Starting Dapr Runtime -- version 1.12.0-rc.3 -- commit af7053834669b3355fc01ce64c2a372e5787b015" app_id=product-catalog-ingestion-bystore instance=product-catalog-ingestion-bystore-6dcb9bf545-74zsm scope=dapr.runtime type=log ver=1.12.0-rc.3
time="2023-09-20T14:49:27.187872692Z" level=info msg="Log level set to: info" app_id=product-catalog-ingestion-bystore instance=product-catalog-ingestion-bystore-6dcb9bf545-74zsm scope=dapr.runtime type=log ver=1.12.0-rc.3
time="2023-09-20T14:49:27.188030494Z" level=warning msg="mTLS is disabled. Skipping certificate request and tls validation" app_id=product-catalog-ingestion-bystore instance=product-catalog-ingestion-bystore-6dcb9bf545-74zsm scope=dapr.runtime.security type=log ver=1.12.0-rc.3
time="2023-09-20T14:49:27.188298197Z" level=info msg="Initializing the operator client" app_id=product-catalog-ingestion-bystore instance=product-catalog-ingestion-bystore-6dcb9bf545-74zsm scope=dapr.runtime type=log ver=1.12.0-rc.3
time="2023-09-20T14:49:42.816327862Z" level=info msg="Received signal 'terminated'; beginning shutdown" app_id=product-catalog-ingestion-bystore instance=product-catalog-ingestion-bystore-6dcb9bf545-74zsm scope=dapr.signals type=log ver=1.12.0-rc.3
[...]

PR #6951, opened 2023-09-20
Daprd: always use mTLS with control plane services when in kubernetes mode

JoshVanL:
Closes https://github.com/dapr/dapr/issues/6949

When running in Kubernetes mode, the control plane services (correctly) always network with [m]TLS, even when `mtls` has been
disabled in configuration.

Currently, daprd will (wrongly) attempt to plain text connect to control plane services when `mtls` has been disabled in configuration. This PR updates the `security` package so that it will always request it's identity certificate when running in Kubernetes mode, regardless of of the `mtls` config option. It has also been updated to that if the identity cert is available (`mtls=true` or `kubernetes=true`) mTLS Dial (which is used by daprd to connect to control plane services) is always mTLS.  

`mtls=false` will continue to allow Daprds to use insecure connections between each other.


ItalyPaleAle (dismissed):
Sorry but I don't think this is a solution to the problem. if mTLS is disabled, it means users want it to be disabled everywhere, not just between sidecars. Having mTLS enabled comes with some additional complexities that may not be desirable, such as the need to rotate root certs...

I agree (m)TLS is much better in general, but we should honor what the user wants in this case.

ItalyPaleAle:
> @ItalyPaleAle As I understand, it has always been the case that disabling `mTLS`, disables `mTLS` between daprd sidecards. `daprd => control plane` and `control plane => control plane` always enforces mTLS regardless (in Kubernetes).

I think you may be right. In 1.11 the only control plane service that had a `tls-enabled` CLI flag was Placement, but in the Helm chart it was always enabled. Ok, i'll dismiss my review

tests/integration/suite/daprd/mtls/kubernetes/operator.go
```diff
@@ -0,0 +1,105 @@
...
+	"google.golang.org/grpc"
+	"google.golang.org/protobuf/types/known/emptypb"
+
+	operatorv1pb "github.com/dapr/dapr/pkg/proto/operator/v1"
+	"github.com/dapr/dapr/pkg/security"
+	procgrpc "github.com/dapr/dapr/tests/integration/framework/process/grpc"
+)
+
+// operator is a mocked implementation of the operator service.
+type operator struct{}
```
JoshVanL, line 32: This stubs the operator sever with always enforced mTLS

JoshVanL:
The PR with only the integrations tests has failed CI because of the following test. This is expected as this is the issue this PR is trying to address.

```
done
2023-09-21T13:44:36.8442449Z     --- FAIL: Test_Integration/daprd/mtls/kubernetes/disable (180.01s)
```

As you can see, this test is now passing with the implementation changes on this PR.

```
2023-09-21T13:42:52.5301368Z         --- PASS: Test_Integration/daprd/mtls/kubernetes/disable/run (2.31s)
```

pkg/security/x509source.go
```diff
@@ -156,7 +157,7 @@ func newX509Source(ctx context.Context, clock clock.Clock, cptd spiffeid.TrustDo
 		appID:          opts.AppID,
 		appNamespace:   ns,
 		trustDomain:    trustDomain,
-		kubernetesMode: os.Getenv("KUBERNETES_SERVICE_HOST") != "",
+		kubernetesMode: opts.Mode == modes.KubernetesMode,
```
artursouza, line 160: Is this a breaking change? Shouldn't we use both flags to enable this (via an OR)?

JoshVanL: As discussed in the call yesterday, we agreed that the `KUBERNETES_SERVICE_HOST` environment variable is fallible so instead we are determining the mode via a passed option. For Daprd, this option is coming from the existing `--mode` CLI flag. For placement, this is coming from a new CLI flag `--mode`.

it's not a breaking change.

pkg/security/security.go
```diff
@@ -135,8 +141,11 @@ func New(ctx context.Context, opts Options) (Provider, error) {
 		return nil, fmt.Errorf("invalid control plane trust domain: %w", err)
 	}
 
+	// Always request certificates from Sentry if mTLS is enabled or running in
+	// Kubernetes. In Kubernetes, Daprd always communicates mTLS with the control
+	// plane.
 	var source *x509source
-	if opts.MTLSEnabled {
+	if opts.MTLSEnabled || opts.Mode == modes.KubernetesMode {
```
yaron2, line 148: Thanks for making that change, this is much better.

yaron2 (approved):

PR #6954, opened 2023-09-21
[DO NOT MERGE] Testing with in-memory SQLite

ItalyPaleAle:
Just like #6943 but uses an in-memory SQLite


ItalyPaleAle:
Using SQLite with in-memory shows QPS:

- 63.46
- 63.74

Closing this as we have the data now

PR #6959, opened 2023-09-24
Resiliency: add `GetAttempt` method for policy function

ItalyPaleAle:
Quick change that makes it easier for a resiliency policy function to get the current attempt number (safely also for concurrent use by multiple goroutines).

Inside a policy runner function, it's now possible to call `resiliency.GetAttempt` with the context to get the current attempt number, for example:

```go
val, err := policyRunner(func(ctx context.Context) (any, error) {
	attempt := resiliency.GetAttempt(ctx)
})
```

There were already 2 instances in the codebase where this could be used. I have also expanded the logs to return the number of attempts before the operation was successful, to improve debugging.

Unit tests are included.

### DIFF
--- a/pkg/actors/actors.go
+++ b/pkg/actors/actors.go
@@ -456,9 +456,8 @@ func (a *actorsRuntime) callRemoteActorWithRetry(
 				Disposer: resiliency.DisposerCloser[*invokev1.InvokeMethodResponse],
 			},
 		)
-		attempts := atomic.Int32{}
 		return policyRunner(func(ctx context.Context) (*invokev1.InvokeMethodResponse, error) {
-			attempt := attempts.Add(1)
+			attempt := resiliency.GetAttempt(ctx)
 			rResp, teardown, rErr := fn(ctx, targetAddress, targetID, req)
 			if rErr == nil {
 				teardown(false)
--- a/pkg/messaging/direct_messaging.go
+++ b/pkg/messaging/direct_messaging.go
@@ -20,7 +20,6 @@ import (
 	"io"
 	"os"
 	"strings"
-	"sync/atomic"
 	"time"
 
 	"github.com/cenkalti/backoff/v4"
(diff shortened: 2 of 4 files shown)


JoshVanL (approved):

pkg/actors/actors.go
```diff
@@ -456,9 +456,8 @@ func (a *actorsRuntime) callRemoteActorWithRetry(
 				Disposer: resiliency.DisposerCloser[*invokev1.InvokeMethodResponse],
 			},
 		)
-		attempts := atomic.Int32{}
 		return policyRunner(func(ctx context.Context) (*invokev1.InvokeMethodResponse, error) {
-			attempt := attempts.Add(1)
+			attempt := resiliency.GetAttempt(ctx)
```
mukundansundar, line 460: I think this line should be after the function call....

pkg/messaging/direct_messaging.go
```diff
@@ -196,9 +195,8 @@ func (d *directMessaging) invokeWithRetry(
 				Disposer: resiliency.DisposerCloser[*invokev1.InvokeMethodResponse],
 			},
 		)
-		attempts := atomic.Int32{}
 		return policyRunner(func(ctx context.Context) (*invokev1.InvokeMethodResponse, error) {
-			attempt := attempts.Add(1)
+			attempt := resiliency.GetAttempt(ctx)
```
mukundansundar, line 199: Same here

pkg/actors/actors.go
ItalyPaleAle: It doesn't matter. The attempt is set in the context so the value would be the same.

pkg/messaging/direct_messaging.go
ItalyPaleAle: Same as other. There's no difference since the value is set in the context.

pkg/actors/actors.go
mukundansundar: Only after the function is run would the context be updated?

ItalyPaleAle: The context is set before the function runs. So the attempt is the same throughout the policy function

Also, changing this to increment `attempts` after (which would be done in the `resiliency` pkg) would cause incorrect numbers in case the function call takes longer than the allowed timeout (a sort of race condition)

mukundansundar (approved):
lgtm

PR #6960, opened 2023-09-25
Fix: make sure actors is locally hosted before executing reminder or timer

ItalyPaleAle:
There's currently a race condition that may cause an actor to be activated on two separate hosts when a reminder or timer is being executed.

1. When the placement state changes, [`evaluateReminders`](https://github.com/ItalyPaleAle/dapr/blob/0762af919d768c7de96b9deecdb0e9179c8df793/pkg/actors/reminders/reminders.go#L268) begins its loop. (Placement state changes every time a Dapr sidecar that can host actors of a given type comes online or offline)
2. One of the first steps it to [load](https://github.com/ItalyPaleAle/dapr/blob/0762af919d768c7de96b9deecdb0e9179c8df793/pkg/actors/reminders/reminders.go#L294) the list of reminders from persistent storage. This is a very costly I/O operation which can take some non-insignificant amount of time.
3. Now, assume that after the placement state changes, the sidecar needs to deactivate the actor "type1/id1". `evaluateReminders` will stop all reminders currently active on the sidecar for "type1/id1". However, this happens after the costly I/O operation.
4. If in the meanwhile a reminder for actor "type1/id1" reaches its execution time, that could [trigger](https://github.com/ItalyPaleAle/dapr/blob/0762af919d768c7de96b9deecdb0e9179c8df793/pkg/actors/reminders/reminders.go#L887-L889) the execution of the reminder (`<-nextTimer.C()`) before the reminder is stopped (`<-stopChannel`).
5. This means that the reminder is executed on the current sidecar even though the actor is supposed to be running on a separate sidecar. This could cause the actor to be active in two sidecars at the same time.

The fix involves:

1. Changing `doExecuteReminderOrTimer` so it performs a lookup in the placement tables to make sure the actor is actually hosted locally before invoking it.
   - Actor invocation goes through a different path and [was already doing that](https://github.com/ItalyPaleAle/dapr/blob/0762af919d768c7de96b9deecdb0e9179c8df793/pkg/actors/actors.go#L406-L418)
2. Changing `placement.LookupActor` to perform retries with resiliency within the method itself. Currently resiliency is performed by the caller, which meant more duplicate code in this case.

The remaining changes are due to changes needed in unit tests


pkg/actors/internal/placement_service.go
```diff
@@ -24,6 +25,22 @@ type PlacementService interface {
...
+	AddHostedActorType(actorType string, idleTimeout time.Duration) error
+}
+
+// LookupActorRequest is the request for LookupActor.
+type LookupActorRequest struct {
+	ActorType string
+	ActorID   string
+}
+
+func (lar LookupActorRequest) ActorKey() string {
```
shubham1172, line 38: nit: doc comment for exported method.

shubham1172 (approved):
LGTM.

pkg/actors/internal/placement_service.go
```diff
@@ -24,6 +25,22 @@ type PlacementService interface {
...
+	return lar.ActorType + "/" + lar.ActorType
```
ItalyPaleAle, line 39:
```suggestion
// ActorKey returns the key for the actor, which is "type/id".
func (lar LookupActorRequest) ActorKey() string {
	return lar.ActorType + "/" + lar.ActorID
```

pkg/actors/internal/placement_service.go
ItalyPaleAle: Thanks, fixed

pkg/actors/actors.go
```diff
@@ -388,42 +396,25 @@ func (a *actorsRuntime) deactivationTicker(configuration Config, deactivateFn de
...
-		return &lookupActorRes{
-			targetActorAddress: rAddr,
-			appID:              rAppID,
-		}, nil
+	lar, err := a.placement.LookupActor(ctx, internal.LookupActorRequest{
+		ActorType: actor.GetActorType(),
+		ActorID:   actor.GetActorId(),
 	})
 	if err != nil {
 		return nil, err
```
RyanLettieri: Can we log the error here?

pkg/actors/placement/placement.go
```diff
@@ -288,23 +294,40 @@ func (p *actorPlacement) WaitUntilReady(ctx context.Context) error {
...
 
 	t := p.placementTables.Entries[actorType]
 	if t == nil {
-		return "", ""
+		return "", "", nil
 	}
 	host, err := t.GetHost(actorID)
 	if err != nil || host == nil {
-		return "", ""
+		return "", "", nil //nolint:nilerr
```
RyanLettieri, line 328: Any reason we don't check for these separately and log + return the error if there is one?

RyanLettieri (commented):
First pass through, just some error+logging suggestions

pkg/actors/placement/placement.go
ItalyPaleAle: This is what the previous code was doing. I am letting the empty address "bubble up" so the caller can eal with it

pkg/actors/actors.go
ItalyPaleAle: I believe the caller will log it

RyanLettieri:
LGTM

ItalyPaleAle:
This has been rebased into release-1.12 in #6968

ItalyPaleAle:
Closing as this has been merged into release-1.12

PR #6972, opened 2023-09-27
[release-1.12] Fix race condition in app health and actors/workflow initialization

ItalyPaleAle:
Fixes the issue discovered in https://github.com/dapr/dapr/pull/6968#issuecomment-1736563355

Dapr reports in /healthz (and /healthz/outbound) a ready status when the API servers are ready, but the actor runtime (and thus workflow) is initialized asynchronously.

This means that apps that try to invoke actors or use workflow fail. We've experienced this error in E2E tests in #6968, where a refactoring on unrelated things made the race condition appear.

This PR fixes the issue by making all actor and workflow APIs (including those managed by the WF engine) block while the actor runtime is initialized (successfully or not).

I validated this fix by adding a `time.Sleep(10 * time.Second)` at the start of `runtime.appHealthReadyInit`

### DIFF
--- a/pkg/grpc/api.go
+++ b/pkg/grpc/api.go
@@ -100,6 +100,8 @@ type APIOpts struct {
 
 // NewAPI returns a new gRPC API.
 func NewAPI(opts APIOpts) API {
+	opts.UniversalAPI.InitUniversalAPI()
+
 	return &api{
 		UniversalAPI:          opts.UniversalAPI,
 		directMessaging:       opts.DirectMessaging,
@@ -952,9 +954,7 @@ func (a *api) ExecuteStateTransaction(ctx context.Context, in *runtimev1pb.Execu
 }
 
 func (a *api) RegisterActorTimer(ctx context.Context, in *runtimev1pb.RegisterActorTimerRequest) (*emptypb.Empty, error) {
-	if a.UniversalAPI.Actors == nil {
-		err := status.Errorf(codes.Internal, messages.ErrActorRuntimeNotFound)
-		apiServerLogger.Debug(err)
+	if err := a.actorReadinessCheck(ctx); err != nil {
 		return &emptypb.Empty{}, err
 	}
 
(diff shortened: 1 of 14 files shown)


JoshVanL (changes requested):
@ItalyPaleAle Can you add an integration test to prove this works? We can test by running daprd (with a placement address configured), testing that healthz fails, then starting placement and wait for the healthz endpoint to return healthy.

ItalyPaleAle:
> @ItalyPaleAle Can you add an integration test to prove this works? We can test by running daprd (with a placement address configured), testing that healthz fails, then starting placement and wait for the healthz endpoint to return healthy.

No, this is not what this PR does.

1. Healthz hasn't changed
2. Also this isn't impacted by whether placement is configured or not. If the actor runtime can't start (because there's no placement configured, or because there's no actor state store, or because the app doens't respond to /dapr/configuration), healthz still returns true and the actors runtime is considered "initialized" (even though is inactive)

The issue this fixes is different:

1. Apps that use actors (and workflow) are requested to wait for /healthz to report success. See https://github.com/dapr/dapr/blob/e4f89012f56022a3e0dec18725493fb0c549ed2b/tests/apps/workflowsapp/Controller.cs#L40
2. Actors are initialized async in the meanwhile
3. The app tries to invoke actors or workflow (example: https://github.com/dapr/dapr/blob/e4f89012f56022a3e0dec18725493fb0c549ed2b/tests/apps/workflowsapp/Controller.cs#L43 )
   - If the initialization of actors is done, this succeeds
   - If not, it fails

The fix makes it so calls like the one above will block until actors is initialized, whether successfully or not.

I am not sure how to even write a test for this, since testing for absence of race conditions is always a hell of a problem.

JoshVanL:
@ItalyPaleAle Please add the integration test added in https://github.com/dapr/dapr/pull/6974 which will pass on this branch. It doesn't test workflows as we don't have any infra for them in integration tests yet but at least covers actors.

ItalyPaleAle:
> @ItalyPaleAle Please cherry pick #6974 which will pass on this branch. It doesn't test workflows as we don't have any infra for them in integration tests yet but at least covers actors.

Thank you, merged this

artursouza (commented):
LGTM

PR #6974, opened 2023-09-27
[1.12] Don't merge; Adds integration test to prove actors respond before initilization is complete

JoshVanL:
This integration should fail, and is used to prove that https://github.com/dapr/dapr/pull/6972 fixes this issue.

Only tests actors as we don't have an infra setup for workflows in integration tests yet.


ItalyPaleAle:
Thanks, merged into #6972

PR #6977, opened 2023-09-28
[release-1.12] Fix issue where gRPC app channel cannot recover from no connection.

artursouza:
# Description

I modified the MethodInvokeIT in Java SDK to use a slow gRPC app startup combined with App Health Check and it failed consistently after runtime commit 0fdab70eb32a63dd2630c39c4b1bc623212cd099. This change allows the gRPC app
channel to recover. The modified integration test mentioned here shows that the channel never recovers from a bad connection without this change, even with a parallel app probe coming from the test code to succeed calling the app's health endpoint.

Related to https://github.com/dapr/java-sdk/pull/917

This issue is in 1.12 only and does not reproduce in 1.11. I used the modified version of MethodInvokeIT for grpc in the Java SDK PR above.

## Issue reference

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing


ItalyPaleAle (approved):

mukundansundar (approved):

yaron2:
@artursouza it looks like unit tests are failing consistently

artursouza:
OK. I will look into it tomorrow.

pkg/grpc/manager/manager.go
```diff
@@ -153,6 +154,13 @@ func (g *Manager) createLocalConnection(parentCtx context.Context, port int, ena
 	} else {
 		opts = append(opts, grpc.WithTransportCredentials(insecure.NewCredentials()))
 	}
+	opts = append(opts, grpc.WithConnectParams(grpc.ConnectParams{
+		Backoff: backoff.Config{
+			MaxDelay:  5 * time.Second,
```
JoshVanL, line 159: Question: Where are these numbers selected from?

JoshVanL (changes requested):

pkg/grpc/manager/manager.go
JoshVanL: Please also add a test for what this is fixing.

artursouza: These numbers came from my intuition :) We can eventually have entries in the Configuration CRD to allow this to be modified.

I am relying on your integration test to validate this, in addition to the Java SDK ITs. This is not unit-testable.

tests/integration/suite/daprd/serviceinvocation/grpc/slowappstartup.go
```diff
@@ -0,0 +1,113 @@
...
+	return []framework.Option{
+		framework.WithProcesses(s.srv, s.daprd),
+	}
+}
+
+func (s *slowappstartup) Run(t *testing.T, ctx context.Context) {
+	s.daprd.WaitUntilRunning(t, ctx)
+	s.daprd.WaitUntilAppHealth(t, ctx)
+	// Waiting is needed to compensate delay from runtime's app health check and the test's app health check.
+	time.Sleep(3 * time.Second)
```
artursouza, line 91: We can only replace this after the runtime supports a new health endpoint, where we can wait for the app channel to be healthy.

tests/integration/framework/process/daprd/daprd.go
```diff
@@ -24,15 +24,19 @@ import (
 	"testing"
 	"time"
 
+	"github.com/dapr/dapr/tests/integration/framework/binary"
```
JoshVanL: Linting

tests/integration/framework/process/daprd/daprd.go
```diff
@@ -24,15 +24,19 @@ import (
...
+	"google.golang.org/grpc"
+	"google.golang.org/protobuf/types/known/emptypb"
 
-	"github.com/dapr/dapr/tests/integration/framework/binary"
+	runtimev1pb "github.com/dapr/dapr/pkg/proto/runtime/v1"
 	"github.com/dapr/dapr/tests/integration/framework/process"
 	"github.com/dapr/dapr/tests/integration/framework/process/exec"
 	prochttp "github.com/dapr/dapr/tests/integration/framework/process/http"
 	"github.com/dapr/dapr/tests/integration/framework/util"
+	"google.golang.org/grpc/credentials/insecure"
```
JoshVanL: Linting

tests/integration/suite/daprd/serviceinvocation/grpc/slowappstartup.go
```diff
@@ -0,0 +1,113 @@
...
+		Id: s.daprd.AppID(),
+		Message: &commonv1.InvokeRequest{
+			Method: "Ping",
+			Data:   req,
+		},
+	})
+	require.NoError(t, err)
+	var pingResp testpb.PingResponse
+	assert.NoError(t, resp.Data.UnmarshalTo(&pingResp))
+}
```
JoshVanL, line 113: We should avoid using `time.Sleep`s at all costs as they begin to add up and greatly increase test execution time (this is a massive problem in unit tests).

The below achieves the same thing but will return as soon as the app is healthy. Having 3 seconds be the timeout for the `Eventually` is semantically the same as the `time.Sleep`. 

```suggestion
func (s *slowappstartup) Run(t *testing.T, ctx context.Context) {
	s.daprd.WaitUntilRunning(t, ctx)
	s.daprd.WaitUntilAppHealth(t, ctx)

	conn, err := grpc.DialContext(ctx, "localhost:"+strconv.Itoa(s.daprd.GRPCPort()),
		grpc.WithTransportCredentials(insecure.NewCredentials()),
		grpc.WithBlock(),
	)
	require.NoError(t, err)
	client := rtv1.NewDaprClient(conn)

	req, err := anypb.New(new(testpb.PingRequest))
	require.NoError(t, err)

	var pingResp testpb.PingResponse
	var resp *commonv1.InvokeResponse
	require.EventuallyWithT(t, func(c *assert.CollectT) {
		resp, err = client.InvokeService(ctx, &rtv1.InvokeServiceRequest{
			Id: s.daprd.AppID(),
			Message: &commonv1.InvokeRequest{
				Method: "Ping",
				Data:   req,
			},
		})
		// This function must only return that the app is not in a healthy state
		// until the app is in a healthy state.
		if !assert.NoError(c, err) {
			require.ErrorContains(c, err, "app is not in a healthy state")
		}
	}, time.Second*3, time.Millisecond*100)
	assert.NoError(t, resp.Data.UnmarshalTo(&pingResp))
}
```

JoshVanL (changes requested):

tests/integration/framework/process/daprd/daprd.go
artursouza: Fixed

tests/integration/framework/process/daprd/daprd.go
artursouza: Fixed.

JoshVanL (approved):

PR #6982, opened 2023-09-29
Integration test optimization 115s -> 80s

JoshVanL:
Adds ParallelTest test helper. This can be used by integration tests as
an alternative to Go's `t.Parallel()` which requires spawning a new test
func which can make test result output noisy.

Updates iowriter to print exec output all at once at end of test
execution, rather than line by line.

Change placement and sentry integration processes to log at info level
rather than debug by default. This is to reduce the amount of logs
generated by the integration tests.

Updates the integration test suite to use the ParallelTest
rather than `t.Parallel()` to reduce test case number.

### DIFF
--- a/tests/integration/framework/util/parallel.go
+++ b/tests/integration/framework/util/parallel.go
@@ -14,10 +14,13 @@ limitations under the License.
 package util
 
 import (
+	"os"
+	"strconv"
 	"sync"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 // ParallelTest is a helper for running tests in parallel without having to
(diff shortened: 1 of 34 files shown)


ItalyPaleAle:
This should probably go into master and not 1.12 at this stage.

artursouza:
> This should probably go into master and not 1.12 at this stage.

Yes. I agree.

tests/integration/framework/util/parallel.go
```diff
@@ -0,0 +1,89 @@
...
+	p := &ParallelTest{
+		fns: fns,
+	}
+	t.Cleanup(func() {
+		p.lock.Lock()
+		defer p.lock.Unlock()
+		t.Helper()
+
+		jobs := make(chan func(), len(p.fns))
+		for i := 0; i < 5; i++ {
```
filintod, line 42: probably this `5` could be an env var or some other configurable variable

filintod: other than that lgtm

ItalyPaleAle (dismissed):

daixiang0:
Start from https://github.com/dapr/dapr/actions/runs/6434413380/job/17473804904?pr=6982#step:7:15396, seems they are package list, do we really need them in log?

tests/integration/framework/util/parallel.go
JoshVanL: Done :+1:

JoshVanL:
> Start from https://github.com/dapr/dapr/actions/runs/6434413380/job/17473804904?pr=6982#step:7:15396, seems they are package list, do we really need them in log?

That is the output from the integration tests running `go build`, but not sure why it is outputting that command here. Will need more investigation.

artursouza (approved):

PR #6985, opened 2023-09-29
Integration: Add pluggable state store test case

JoshVanL:
Adds a pluggable state store integration test process, with a basic state store test.


tests/integration/framework/util/rand.go
```diff
@@ -0,0 +1,29 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package util
+
+import (
+	"crypto/rand"
+)
+
```
daixiang0, line 19:
```suggestion
func init() {
    rand.Seed(time.Now().UnixNano())
}
```

tests/integration/framework/util/rand.go
```diff
@@ -0,0 +1,29 @@
...
+// RandomString generates a random string of length n.
+func RandomString(n int) string {
+	const letters = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789"
```
daixiang0, line 22: Put it at top level, not in function.

tests/integration/framework/util/rand.go
```diff
@@ -0,0 +1,29 @@
...
+	bytes := make([]byte, n)
+	rand.Read(bytes)
+	for i, b := range bytes {
+		bytes[i] = letters[b%byte(len(letters))]
```
daixiang0, line 26:
```suggestion
		bytes[i] = letters[rand.Intn(len(letters))]
```

tests/integration/framework/util/rand.go
```diff
@@ -0,0 +1,29 @@
...
+
+import (
+	"crypto/rand"
+)
+
+// RandomString generates a random string of length n.
+func RandomString(n int) string {
+	const letters = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789"
+	bytes := make([]byte, n)
+	rand.Read(bytes)
```
daixiang0, line 24:
```suggestion
	
```

tests/integration/suite/actors/healthz/initerror.go_
```diff
@@ -0,0 +1,128 @@
+/*
```
daixiang0, line 1: Why need `.go_`

tests/integration/suite/daprd/pluggable/basic.go
```diff
@@ -0,0 +1,191 @@
...
+func init() {
+	suite.Register(new(basic))
+}
+
+type basic struct {
+	daprd *daprd.Daprd
+}
+
+func (b *basic) Setup(t *testing.T) []framework.Option {
+	if runtime.GOOS == "windows" {
```
daixiang0, line 48: Does inmemory state store can not run in Win?

tests/integration/suite/actors/healthz/reminderrace.go_
```diff
@@ -0,0 +1,244 @@
...
+	"github.com/dapr/dapr/tests/integration/framework/process/placement"
+	procstatestore "github.com/dapr/dapr/tests/integration/framework/process/statestore"
+	"github.com/dapr/dapr/tests/integration/suite"
+)
+
+func init() {
+	suite.Register(new(reminderrace))
+}
+
+// TODO: @joshvanl
```
daixiang0, line 46: Please the details about TODO and remove all comment codes.

tests/integration/framework/util/rand.go
ItalyPaleAle: `crypto/rand` is a CSPRNG that uses entropy from the kernel, so `rand.Seed`  isn't needed (it should not even work)

tests/integration/framework/util/rand.go
ItalyPaleAle: `b` contains random bytes so should be ok?

Now, we could argue that using a `char` and doing the modulus [doesn't offer perfect uniformly-distributed random numbers](https://stackoverflow.com/questions/10984974/why-do-people-say-there-is-modulo-bias-when-using-a-random-number-generator), but this is just a test :)

tests/integration/framework/util/rand.go
JoshVanL: It is here on purpose because I didn't want to have a test specific constant be at the `util` package namespace level.

tests/integration/framework/util/rand.go
JoshVanL: I think it should be fine, but have changed it to use `rand.Int` anyway.

tests/integration/suite/actors/healthz/reminderrace.go_
JoshVanL: Removed .

tests/integration/framework/util/rand.go
daixiang0: The util is for test and we may use it in more tests.

JoshVanL: Until then, I’d rather we keep it inside the func to reduce noise in the util package namespace.

cicoyle:
LGTM

yaron2 (approved):

PR #6988, opened 2023-09-29
Fixed race conditions in Workflow E2E test

ItalyPaleAle:
Fixes the Workflow E2E test which had a significant number of race conditions - in the test itself.

Operations on Workflow management, such as start, pause, resume, etc, are all asynchronous, so checking their status immediately after invoking them can lead to race conditions.

The test has been updated to use a number of `require.Eventually`'s which poll until the condition is met.

Additionally, `AppPort` has been set to -1 because the workflow app doesn't use an app channel, and doing so causes a  number of issues with Dapr actors healthchecks (which are not implemented by the app).

### DIFF
--- a/tests/e2e/workflows/workflow_test.go
+++ b/tests/e2e/workflows/workflow_test.go
@@ -25,10 +25,12 @@ import (
 	"testing"
 	"time"
 
+	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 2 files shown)


JoshVanL (approved):

RyanLettieri:
LGTM!

PR #7001, opened 2023-10-06
[1.12] Fix invoke remote stream from closing stream before EOF

### LINKED ISSUE #6997: unexpected resiliency policy behavior with v1.12
## What version of Dapr?

1.12 rc 5

## Expected Behavior

When I set a resiliency policy for my caller appid with the yaml manifest (omitting metadata for brevity):
```
apiVersion: dapr.io/v1alpha1
kind: Resiliency
scopes:
- ignore-me
spec:
  policies:
    circuitBreakers:
      DefaultAppCircuitBreakerPolicy:
        interval: 0s
        maxRequests: 1
        timeout: 60s
        trip: consecutiveFailures > 5
    retries:
      DefaultAppRetryPolicy:
        duration: 5s
        maxInterval: 60s
        maxRetries: -1
        policy: constant
    timeouts:
      DefaultAppTimeoutPolicy: 10s
  targets: {}
```
I expect no issue performing service invocation.

## Actual Behavior

Then my caller sidecar gives me logs (some things omitted with ...):
```
"app_id":"sam-invokerrr","error_message":"{\"errorCode\":\"ERR_DIRECT_INVOKE\",\"message\":\"failed to invoke, id: ignore-me, err: context canceled\"}","execution_time":"180.883958ms","instance":"...","level":"info" ... "method":"GET /v1.0/invoke/ignore-me/method/todos","msg":"Inbound App ID call received (http)",...,"protocol":"http","scope":....,"start_time":"2023-10-05T16:12:55.509103626Z","status":"500","time":"2023-10-05T16:12:55.690156084Z","type":"request",...}
```

However, if I update my resiliency yaml from:
```
  targets: {}

```
to 👇 
```
  targets:
    apps: {}
```

THEN it works 🎉 .

Looking at the resiliency structs, there is an omitempty json tag on apps:
```
	Apps       map[string]EndpointPolicyNames  `json:"apps,omitempty" yaml:"apps,omitempty"`
```

That I do not think we should have. Otherwise, there is other weirdness going on in the underlying resiliency logic causing my sidecar to _immediately_ give me the context cancel error.

### Other weirdness

Furthermore, if I update my invocation URL from `https://jsonplaceholder.typicode.com/todos` to `https://jsonplaceholder.typicode.com/todos/1`, THEN it also works. But, it does not work for `https://jsonplaceholder.typicode.com/todos`, showing the context cancel error. I use this URL as it easily allows me to do basic crud operations without me doing any underlying logic just to test the resiliency :) Again, the context cancel is given almost immediately irrespective of my specified timeout.

1. Create 2 appids using invoker resiliency policy listed above.
2. Perform service invocation. NOTE: you will see an okay status code; however, if you check the caller sidecar logs, then you see a status 500 with the context cancel error.

JoshVanL:
Closes #6997

The invoke remote stream go routine data handler will exit early and write the context's error in the event that the context is cancelled. Importantly, it will react to the context being cancelled, before the actual connection has an EOF or graceful close.
This scenario happens specifically when a timeout resiliency policy is used for service invocation (though there could be other scenarios where this happens I didn't find). With resiliency, regardless of whether a timeout is reached or not, the context is always (correctly) cancelled after the operation is finished, but since the stream data routine would still be sending data or before an EOF, it catches the context cancelled and sends that instead.

https://github.com/dapr/dapr/blob/5d8f6de1b6affb78e393e634f2b9cdf36d7ad254/pkg/resiliency/policy.go#L118-L128

https://github.com/dapr/dapr/blob/master/pkg/messaging/direct_messaging.go#L344-L346

This PR updates this behaviour so that the go routine will continue to attempt to read and write data on the line until there is a error such as in the event of a graceful close of the sender (because the stream is complete or because the context was cancelled!). This ensures the entire message is sent over the stream.

This error is also present in v1.11 and should be back ported.

### DIFF
--- a/pkg/messaging/direct_messaging.go
+++ b/pkg/messaging/direct_messaging.go
@@ -440,11 +440,6 @@ func (d *directMessaging) invokeRemoteStream(ctx context.Context, clientV1 inter
 			readErr   error
 		)
 		for {
-			if ctx.Err() != nil {
-				pw.CloseWithError(ctx.Err())
-				return
-			}
-
 			// Get the payload from the chunk that was previously read
 			payload = chunk.GetPayload()
 			if payload != nil {
--- a/pkg/messaging/direct_messaging_test.go
+++ b/pkg/messaging/direct_messaging_test.go
@@ -21,6 +21,7 @@ import (
 	"io"
 	"net"
 	"path/filepath"
+	"runtime"
 	"testing"
 	"time"
 
@@ -186,6 +187,8 @@ func TestInvokeRemote(t *testing.T) {
 	}
 
 	t.Run("streaming with no data", func(t *testing.T) {
+		routines := runtime.NumGoroutine()
+
 		messaging, teardown := prepareEnvironment(t, true, nil)
 		defer teardown()
 
@@ -199,11 +202,16 @@ func TestInvokeRemote(t *testing.T) {
 
 		pd, err := res.ProtoWithData()
 		require.NoError(t, err)
-
+		if err != nil {
+			return
+		}
 		assert.True(t, pd.Message.Data == nil || len(pd.Message.Data.Value) == 0)
+		assert.Equal(t, routines, runtime.NumGoroutine())
 	})
 
 	t.Run("streaming with single chunk", func(t *testing.T) {
+		routines := runtime.NumGoroutine()
+
 		messaging, teardown := prepareEnvironment(t, true, []string{"🐱"})
 		defer teardown()
 
@@ -219,9 +227,11 @@ func TestInvokeRemote(t *testing.T) {
 		require.NoError(t, err)
 
 		assert.Equal(t, "🐱", string(pd.Message.Data.Value))
+		assert.Equal(t, routines, runtime.NumGoroutine())
 	})
 
 	t.Run("streaming with multiple chunks", func(t *testing.T) {
+		routines := runtime.NumGoroutine()
 		chunks := []string{
 			"Sempre caro mi fu quest'ermo colle ",
 			"e questa siepe, che da tanta parte ",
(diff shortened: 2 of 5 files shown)


pkg/messaging/direct_messaging.go
```diff
@@ -438,11 +438,6 @@ func (d *directMessaging) invokeRemoteStream(ctx context.Context, clientV1 inter
 			readErr   error
 		)
 		for {
-			if ctx.Err() != nil {
-				pw.CloseWithError(ctx.Err())
-				return
-			}
-
```
JoshVanL, line 445: This bit is the problem.

pkg/messaging/direct_messaging_test.go
```diff
@@ -219,9 +227,11 @@ func TestInvokeRemote(t *testing.T) {
 		require.NoError(t, err)
 
 		assert.Equal(t, "🐱", string(pd.Message.Data.Value))
+		assert.Equal(t, routines, runtime.NumGoroutine())
```
ItalyPaleAle, line 230: I wonder how reliable this test will be, since there could perhaps be other background goroutines started at random?

pkg/messaging/direct_messaging.go
ItalyPaleAle: If there's a context error, should we close the reader however? This way we are still consuming the stream even though we know that the caller isn't interested in that anymore? Unless I'm missing something

JoshVanL: When that is the case, this will surface in the block below. The reader should be controlled by the same context as here.

pkg/messaging/direct_messaging_test.go
JoshVanL: Indeed, let's remove them.

daixiang0 (approved):
LGTM

ItalyPaleAle (approved):

mukundansundar (approved):

PR #7010, opened 2023-10-09
ensure clientID is properly set & add test for pubsub.kafka && bindings.kafka

cicoyle:
Added a metadata_helper func to update the `ClientID` value that is shared btw both `bindings.kafka` && `pubsub.kafka`. I added to the tests to test both standalone and k8s mode. Since the value depends on the mode: self hosted or kubernetes, I put the logic here.

## Issue reference

Will close: https://github.com/dapr/components-contrib/issues/2397

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing
* [ ] Extended the documentation / https://github.com/dapr/docs/pull/3713

### DIFF
--- a/pkg/runtime/metadata_helper/metadata_helper.go
+++ b/pkg/runtime/metadata_helper/metadata_helper.go
@@ -1,0 +1,35 @@
+/*
+Copyright 2023 The Dapr Authors
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package metadatahelper
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/dapr/dapr/pkg/modes"
+)
+
+// Shared helper func for metadata on pubsub.kafka && bindings.kafka to properly set the ClientID
+func SetKafkaClientID(properties map[string]string, id string, namespace string, mode modes.DaprMode) {
+	clientID := strings.TrimSpace(properties["clientID"])
+	if clientID == "" || clientID == "sarama" {
+		switch mode {
+		case modes.KubernetesMode:
+			clientID = fmt.Sprintf("%s.%s", namespace, id)
+		case modes.StandaloneMode:
+			clientID = id
+		}
+		properties["clientID"] = clientID
+	}
+}
--- a/pkg/runtime/processor/binding/binding.go
+++ b/pkg/runtime/processor/binding/binding.go
@@ -28,11 +28,13 @@ import (
 	"github.com/dapr/dapr/pkg/config"
 	diag "github.com/dapr/dapr/pkg/diagnostics"
 	"github.com/dapr/dapr/pkg/grpc/manager"
+	"github.com/dapr/dapr/pkg/modes"
 	"github.com/dapr/dapr/pkg/resiliency"
 	"github.com/dapr/dapr/pkg/runtime/channels"
 	"github.com/dapr/dapr/pkg/runtime/compstore"
 	rterrors "github.com/dapr/dapr/pkg/runtime/errors"
 	"github.com/dapr/dapr/pkg/runtime/meta"
+	metadatahelper "github.com/dapr/dapr/pkg/runtime/metadata_helper"
 	"github.com/dapr/kit/logger"
 )
 
@@ -61,7 +63,10 @@ type Options struct {
 }
 
 type binding struct {
-	isHTTP bool
+	id        string
+	namespace string
+	isHTTP    bool
+	mode      modes.DaprMode
 
 	registry    *compbindings.Registry
 	resiliency  resiliency.Provider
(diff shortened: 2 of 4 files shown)


pkg/runtime/metadata_helper/metadata_helper.go
```diff
@@ -0,0 +1,35 @@
...
+	if clientID == "" || clientID == "sarama" {
+		switch mode {
+		case modes.KubernetesMode:
+			clientID = fmt.Sprintf("%s.%s", namespace, id)
+		case modes.StandaloneMode:
+			clientID = id
+		}
+		properties["clientID"] = clientID
+	}
+}
```
JoshVanL, line 35: We should refrain from putting component specific logic in runtime as much as we can help it, and instead make this generic for all components or a component type. I think we should always be passing in the name and namespace as metadata to components `Init` funcs (`dapr.io/id` & `dapr.io/namespace`) and the Kafka implementations in components-contrib will be able to use that as they need.

pkg/runtime/metadata_helper/metadata_helper.go
```diff
@@ -0,0 +1,35 @@
...
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/dapr/dapr/pkg/modes"
+)
+
+// Shared helper func for metadata on pubsub.kafka && bindings.kafka to properly set the ClientID
+func SetKafkaClientID(properties map[string]string, id string, namespace string, mode modes.DaprMode) {
```
JoshVanL, line 24: Setting the `name` and `namespece` probably belongs in `pkg/runtime/meta`.

pkg/runtime/runtime_test.go
```diff
@@ -1267,70 +1267,266 @@ func TestMetadataNamespace(t *testing.T) {
 	assert.NoError(t, err)
 }
 
-func TestMetadataClientID(t *testing.T) {
```
JoshVanL, line 1270: Why did this test get deleted?

JoshVanL (changes requested):
Few comments from me :slightly_smiling_face:

ItalyPaleAle (changes requested):
I do not support making the runtime implement logic specific for some component types. All component-specific code should live in components-contrib.

The runtime should not be concerned with what component is being initialized and should not change its behavior accordingly.

If you need to pass more things to a component, perhaps you could change the Metadata base struct for that component type (e.g. [here for pubsub](https://github.com/dapr/components-contrib/blob/master/pubsub/metadata.go)) so every component receives the properties and then uses them accordingly?

pkg/runtime/metadata_helper/metadata_helper.go
mukundansundar: Agreed, I believe this should move into components-contrib probably internal utils. And the appID and namespace should always be passed in on Init.

pkg/runtime/metadata_helper/metadata_helper.go
```diff
@@ -0,0 +1,35 @@
...
+	clientID := strings.TrimSpace(properties["clientID"])
+	if clientID == "" || clientID == "sarama" {
+		switch mode {
+		case modes.KubernetesMode:
+			clientID = fmt.Sprintf("%s.%s", namespace, id)
+		case modes.StandaloneMode:
+			clientID = id
```
mukundansundar, line 31: we do support NAMESPACE env var for standalone mode, here we are not taking that into account?

pkg/runtime/processor/binding/binding.go
```diff
@@ -166,13 +171,18 @@ func (b *binding) initInputBinding(ctx context.Context, comp compapi.Component)
...
-	meta, err := b.meta.ToBaseMetadata(comp)
+	baseMetadata, err := b.meta.ToBaseMetadata(comp)
 	if err != nil {
 		diag.DefaultMonitoring.ComponentInitFailed(comp.Spec.Type, "init", comp.ObjectMeta.Name)
 		return rterrors.NewInit(rterrors.CreateComponentFailure, fName, err)
 	}
+	properties := baseMetadata.Properties
 
-	err = binding.Init(ctx, bindings.Metadata{Base: meta})
+	if comp.Spec.Type == "bindings.kafka" {
```
mukundansundar, line 181: prefer not to have comp specific checks here.

pkg/runtime/processor/pubsub/pubsub.go
```diff
@@ -159,6 +160,10 @@ func (p *pubsub) Init(ctx context.Context, comp compapi.Component) error {
 	}
 	properties["consumerID"] = consumerID
 
+	if comp.Spec.Type == "pubsub.kafka" {
```
mukundansundar, line 163: same as above.

mukundansundar (changes requested):

artursouza:
> I do not support making the runtime implement logic specific for some component types. All component-specific code should live in components-contrib.
> 
> The runtime should not be concerned with what component is being initialized and should not change its behavior accordingly.
> 
> If you need to pass more things to a component, perhaps you could change the Metadata base struct for that component type (e.g. [here for pubsub](https://github.com/dapr/components-contrib/blob/master/pubsub/metadata.go)) so every component receives the properties and then uses them accordingly?

I agree that runtime should not have component-specific code. Although there is exceptions, like how actor runtime needs to pass a metadata specific to CosmosDb for transactions to work in that component.

pkg/runtime/runtime_test.go
cicoyle: it was renamed to the below: `TestKafkaMetadataClientID`

cicoyle:
I see. I'll close this and make the component-contrib changes

PR #7013, opened 2023-10-09
Update pinned components-contrib

ItalyPaleAle:
Also deletes the registration for the pubsub.natsstreaming component that was deleted


daixiang0 (approved):

daixiang0 (approved):
LGTM

PR #7022, opened 2023-10-11
Tune actor app health check to become healthy sooner

JoshVanL:
The current implementation of the actor health check means that it takes a significant amount of time before an app is checked for its healthy status (5s) and the actor sub system becomes ready. This amount of time is not appropriate for actor apps which are generally going to be stateless unto themselves will little dependencies- besides Dapr(!).

This PR updates the implementation to have two modes of health check- a unhealthy status check and a healthy status check. On startup and after the initial delay of 1 second, the checker enters the unhealthy check mode which checks the app health every half second. On success, the checker reports a healthy status and moves onto the healthy status mode. The healthy status mode checks the health every 3 seconds and requires 4 consecutive unhealthy status checks before reporting the unhealthy status and moving to the unhealthy status mode again.

This dramatically speeds up actor app startup times which can be seen from the integration tests. There are more improvements we can do to the actor subsystem within placemenet, but that will be followup PRs.

The actor placement client handler now responds immediately to a shutdown signal, rather than waited idling for a static `time.Sleep` duration, speeding up shutdown times of actor apps.


ItalyPaleAle:
Please, let's wait for release-1.12 to be merged into master before making more changes into pkg/actors: https://github.com/dapr/dapr/pull/7017

pkg/actors/actors.go
```diff
@@ -270,26 +288,27 @@ func (a *actorsRuntime) Init(ctx context.Context) error {
...
 	// Be careful to configure healthz endpoint option. If app healthz returns unhealthy status, Dapr will
 	// disconnect from placement to remove the node from consistent hashing ring.
 	// i.e if app is busy state, the healthz status would be flaky, which leads to frequent
 	// actor rebalancing. It will impact the entire service.
-	return a.getAppHealthCheckChanWithOptions(ctx,
+	return a.getAppHealthCheckerWithOptions(
 		health.WithFailureThreshold(4),
-		health.WithInterval(5*time.Second),
+		health.WithHealthyStateInterval(5*time.Second),
+		health.WithUnHealthyStateInterval(time.Second/2),
```
ItalyPaleAle, line 303: Sometimes the apps could become un-healthy when they are overloaded. In this case, I think that an interval of .5s can add to an overloaded app too much?

```suggestion
		health.WithUnHealthyStateInterval(1500*time.Millisecond),
```

pkg/actors/health/healthunit.go
```diff
@@ -0,0 +1,26 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package health
+
+import "k8s.io/utils/clock"
+
+// WithClock sets a custom clock (for mocking time).
+func WithClock(clock clock.WithTicker) Option {
```
ItalyPaleAle, line 22: Let's keep this without the build tag. The overhead is basically none. We had to recently un-do a similar thing in dapr/kit as it made writing unit tests for other packages harder (https://github.com/dapr/kit/pull/62)

pkg/actors/health/health.go
```diff
@@ -0,0 +1,297 @@
...
+
+	"github.com/dapr/kit/logger"
+)
+
+const (
+	defaultInitialDelay           = time.Second * 1
+	defaultFailureThreshold       = int32(2)
+	defaultRequestTimeout         = time.Second * 2
+	defaultHealthyStateInterval   = time.Second * 3
+	defaultUnHealthyStateInterval = time.Second / 2
```
ItalyPaleAle, line 33: See my other comment
```suggestion
	defaultUnHealthyStateInterval = time.Millisecond * 1500
```

pkg/actors/health/health.go
```diff
@@ -0,0 +1,297 @@
...
+		closeCh:                make(chan struct{}),
+	}, nil
+}
+
+func (c *Checker) Run(ctx context.Context) {
+	c.wg.Add(2)
+	defer c.Close()
+	defer c.wg.Done()
+
+	ctx, cancel := context.WithCancel(ctx)
```
ItalyPaleAle, line 118: Let's keep each wg.Add next to the wg.Done so:

1. It's clearer when they are tied to each other
2. It saves errors when we make changes in the future (a similar pattern in pkg/actors has caused a bunch of issues over the last months)

```suggestion
	defer c.Close()
	c.wg.Add(1)
	defer c.wg.Done()

	ctx, cancel := context.WithCancel(ctx)
    c.wg.Add(1)
```

pkg/actors/health/health.go
```diff
@@ -0,0 +1,297 @@
...
+			return
+
+		case <-ticker.C():
+			if c.getStateHealth(ctx) {
+				failureCount = 0
+			} else {
+				failureCount++
+			}
+
+			if failureCount >= c.failureThreshold {
```
ItalyPaleAle, line 207: This is to prevent overflows

```suggestion
			if failureCount >= c.failureThreshold {
                failureCount = c.failureThreshold
```

pkg/actors/health/health.go
```diff
@@ -0,0 +1,297 @@
...
+				healthLogger.Warnf("Actor health check failed %d times, marking unhealthy", failureCount)
```
ItalyPaleAle, line 208: This should only be displayed after the first time the actor becomes unhealthy, or it could cause a lot of noise in the logs

pkg/actors/health/health.go
```diff
@@ -0,0 +1,297 @@
...
+// endpoint is in a healthy state, according to the configured status code.
+type Checker struct {
+	client *http.Client
+
+	ch chan bool
+
+	address                string
+	initialDelay           time.Duration
+	requestTimeout         time.Duration
+	failureThreshold       int32
```
ItalyPaleAle, line 64: Why not `int`? Or if it has to be 32-bit, why not `uint32`?

ItalyPaleAle (changes requested):

pkg/actors/actors.go
JoshVanL: > I think that an interval of .5s can add to an overloaded app too much?

I don't think so- 500ms is a really long time for a  HTTP endpoint that isn't doing any work. We have the request timeout to handle overloaded networks.

pkg/actors/health/healthunit.go
JoshVanL: No, this we should not pollute implementation code with test code. The `unit` build tag is a standard test tag used throughout the Dapr project where any testing code is able to make use of the `WithClock` option.

> We had to recently un-do a similar thing in dapr/kit as it made writing unit tests for other packages harder

Why was this is the case? Surely the same argument can be made for interface mocks as well.

pkg/actors/health/health.go
JoshVanL: As explained above, 500ms is already a long time for an HTTP endpoint over localhost that does no work, and request timeout covers an overloaded network.

pkg/actors/health/healthunit.go
JoshVanL: In the linked PR is says?

>  There are use cases where having those methods available even outside of a unit test is helpful, such as when the objects are instantiated with a clock that could be mocked in the unit test for the parent method.

Why would you want to use a mocked clock outside of a unit test?

pkg/actors/health/health.go
JoshVanL: Why would this prevent an overflow?

pkg/actors/health/health.go
JoshVanL: > This should only be displayed after the first time the actor becomes unhealthy, or it could cause a lot of noise in the logs

A log message is only sent on health state changes, i.e. healthy -> non-healthy and non-healthy to healthy. I think this is a reasonable as a Dapr operator would want to see the last state change from the logs else they would be guessing if they were printed once.

pkg/actors/health/health.go
JoshVanL: Just because that was what was there before. I agree that we should be using `int`.

pkg/actors/health/health.go
ItalyPaleAle: Please, this HAS caused merge conflicts that were hard to debug

JoshVanL: This has been implemented https://github.com/dapr/dapr/pull/7022/files#diff-b51da868ff9b911968be455196e2898445e06852e44a7a89882b06e1734d36ceR114

pkg/actors/health/health.go
JoshVanL: Sorry, still not quite sure how this would be problematic. `failureCount` is a variable local to the function which we use to track how many failures were encountered doing the health check. This begins at 0 and once it hits the threshold, we report the unhealthy state and return from the function to enter the unhealthy state. This throws away the variable, where it will be 0 again when we re-enter.

It is actually impossible for `failureCount` to be any larger than `failureThreshold` (we could change `>=` to just `==`), and so it's impossible for it to overflow, since  `failureThreshold` _has_ to be at most the maximum value that `failureCount` could be.

ItalyPaleAle: Ok, I misunderstood this code

artursouza (approved):

elena-kolevska (approved):

pkg/actors/actors.go
```diff
@@ -1101,15 +1121,15 @@ func (a *actorsRuntime) Close() error {
...
 		if a.placement != nil {
-			err := a.placement.Close()
-			if err != nil {
-				errs = append(errs, fmt.Errorf("failed to close placement service: %w", err))
+			if err := a.placement.Close(); err != nil {
+				return fmt.Errorf("failed to close placement service: %w", err)
 			}
 		}
-		return errors.Join(errs...)
 	}
```
ItalyPaleAle, line 1133: Please don't remove the errors.Join, there can be other things to close

```suggestion
	if a.closed.CompareAndSwap(false, true) {
		defer close(a.closeCh)
		errs := []error{}
        if a.checker != nil {
			a.checker.Close()
		}
		if a.placement != nil {
			err := a.placement.Close()
			if err != nil {
				errs = append(errs, fmt.Errorf("failed to close placement service: %w", err))
			}
		}
		return errors.Join(errs...)
	}
```

JoshVanL: Sure thing, thanks @ItalyPaleAle. Done

ItalyPaleAle (approved):

PR #7023, opened 2023-10-11
[release-1.10] Pins Go version

elena-kolevska:
Pinning Go version to the version in go.mod

### DIFF
--- a/.github/workflows/dapr.yml
+++ b/.github/workflows/dapr.yml
@@ -69,7 +69,7 @@ jobs:
         with:
           go-version-file: "go.mod"
       - name: Set Go Version to Env Variable
-        run: echo "GOVER=$(go version | grep -oP 'go\K[0-9]+\.[0-9]+(\.[0-9]+)?')" >> $GITHUB_ENV
+        run: echo "GOVER=$(go version | awk '{print substr($3,3)}')" >> $GITHUB_ENV
       - name: Set up QEMU
         uses: docker/setup-qemu-action@v1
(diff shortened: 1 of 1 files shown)


JoshVanL (approved):

.github/workflows/dapr.yml
```diff
@@ -34,7 +34,7 @@ jobs:
     name: Build ${{ matrix.target_os }}_${{ matrix.target_arch }} binaries
     runs-on: ${{ matrix.os }}
     env:
-      GOVER: "^1.19.2"
+      GOVER: "1.19.2"
```
ItalyPaleAle, line 37: If you pin a version, should at least be 1.19.13 which is the latest version. There have been security fixes too in-between: https://go.dev/doc/devel/release#go1.19.minor

But even better would be to change the setup-go action to read from the go.mod file: https://github.com/dapr/dapr/blob/5ac2f3d601dedc49913d7a848c49c1b8df660893/.github/workflows/dapr.yml#L138-L142

ItalyPaleAle (changes requested):

.github/workflows/dapr.yml
elena-kolevska: Good point! I didn't know about this option; it's definitely much cleaner.

ItalyPaleAle (approved):

mukundansundar (approved):

mukundansundar:
closing in favor of #7165

Issue #7028, opened 2023-10-12
Inconsistencies in daprd CLI flags and annotations


ItalyPaleAle:
The CLI flags for the `daprd` command, and corresponding annotations, have grown organically over time and now have a bunch of inconsistencies.

Aside from being "not nice to the eye", these inconsistencies cause confusion to users and overall a poor user experience.

For example:

- All addresses for Dapr control plane follow the `<service>-address` scheme (e.g. `sentry-address`), but for placement it is `placement-host-address`
- The name for the flag for the Dapr operator is misleading: `control-plane-address`, while it should be `operator-address`. The operator is only one of the control plane services.
- Units are unclear. For example:
   - `dapr-http-max-request-size` -> this accepts an integer and interprets it as **MBs**
   - `dapr-http-read-buffer-size` -> this accepts an integer and interprets it as **KBs**
   - `app-health-probe-interval` -> this is in seconds
   - `app-health-probe-timeout` -> this is in milliseconds
   - `dapr-graceful-shutdown-seconds` -> this has seconds in the name (and doesn't allow setting values in milliseconds, for example)
- `dapr-http-max-request-size` and `dapr-http-read-buffer-size` -> although these have "HTTP" in the name, they also apply to gRPC

Similarly to efforts that have been underway in components-contrib for a while to standardize and simplify component metadata,  I suggest we clean up the inconsistencies and standardize on common patterns:

| Current option | Proposed change |
|---|---|
| `control-plane-address` | Rename to `operator-address`<br>Preserve `control-plane-address` for backwards-compat as a deprecated alias |
| `placement-host-address` | Rename to `placement-address`<br>Preserve `placement-host-address` for backwards-compat as a deprecated alias |
| `dapr-http-max-request-size` | Create new option `max-body-size` (name is generic as it's used for gRPC too) which accepts a _quantity_<br>Preserve `dapr-http-max-request-size` for backwards comaptibility as deprecated alias, in MB (binary unit, <<20) |
| `dapr-http-read-buffer-size` | Create new option `read-buffer-size` (name is generic as it's used for gRPC too) which accepts a _quantity_<br>Preserve `dapr-http-read-buffer-size` for backwards comaptibility as deprecated alias, in KB (binary unit, <<10). |
[...]

PR #7029, opened 2023-10-12
Switch parsing of CLI flags to pflag

ItalyPaleAle:
> Includes #6925 and expands that

Minor refactoring of the CLI flag parsers to be based on pflag rather than the stdlib's flag. Applies to daprd and control plane services.

There are 3 reasons for this change:

- The stdlib's flag package defines all CLI flags in a global state. This means that if we import a library that defines a flag, it gets added to our CLIs. This is not a hypothetical scenario, as we had an issue fixed in Dapr 1.12 where a component was adding a flag to the daprd CLI without us realizing it for a long time.
- pflag has better support for deprecated flags, which can be hidden from the `--help` command. We have 2 deprecated flags in daprd currently, and there may be more if #7028 is worked on
- Lastly, pflag has some very nice features, including a nicer `--help` message and helpers for example for accepting slices. We can define a flag with `StringSliceVar` and it parses automatically multiple instances of the same flag (e.g. `--config a.yaml --config b.yaml`) as well as comma-separated values (`--config a.yaml,b.yaml`) and a combination of the two.

> Yes, this is partly based on #6766, but unlike that it doesn't change the behavior. Apps will continue to crash if unknown flags are passed.


daixiang0 (approved):

mukundansundar (changes requested):
please have token audience removal as a separate PR.

DeepanshuA (commented):
Do we need to showcase any manual tests here as well, to show the usage of pflag makes no difference in user experience?

ItalyPaleAle:
> Do we need to showcase any manual tests here as well, to show the usage of pflag makes no difference in user experience?

The E2E and IT test should be validating this already, as they invoke the commands using the CLI flags

DeepanshuA (approved):

mukundansundar:
@ItalyPaleAle Was this change to `daprd` tested with existing `dapr` CLI?

ItalyPaleAle:
> @ItalyPaleAle Was this change to `daprd` tested with existing `dapr` CLI?

Yes it works with the CLI

mukundansundar (approved):

PR #7037, opened 2023-10-13
Removes legacy SPIFFE TLS clients and servers in favour of the new SPIRE TLS clients and servers.

JoshVanL:
E2E tests are currently failing on N-2 version skew, though it was my understanding that this was best effort and not guaranteed. Perhaps we need to be skipping this test for now.

Part of #5756


ItalyPaleAle:
> E2E tests are currently failing on N-2 version skew, though it was my understanding that this was best effort and not guaranteed. Perhaps we need to be skipping this test for now.

The failures seem related to some setup steps, not code changes?

pkg/injector/patcher/sidecar_container.go
```diff
@@ -260,24 +260,6 @@ func (c *SidecarConfig) getSidecarContainer(opts getSidecarContainerOpts) (*core
 		},
 	}
 
-	// TODO: @joshvanl: included for backwards compatibility with v1.11 daprd's
-	// which request these environment variables to be present when running in
-	// Kubernetes. Should be removed in v1.13.
-	container.Env = append(container.Env,
```
ItalyPaleAle, line 266: Are we ok dropping support for 1.11 in 1.13? That is not the usual "N-2" compatibility we offer.

artursouza: We support N-1 compatibility, N-2 support is for users running with N-2 and not mixing N-2 with N. Anyway, 1.11 is N-3 from 1.14, so we are OK to remove it either way.

artursouza (approved):

PR #7038, opened 2023-10-13
Local file-based name resolver with SQLite + Update components-contrib to register Blob Storage v2

### LINKED ISSUE #3256: [Proposal] Provide alternative service resolution mechanism besides mdns
## In what area(s)?

/area runtime

## Describe the feature

We commonly get reports that service invocation does not work with dapr standalone mode:
https://github.com/dapr/dapr/issues/2665
https://github.com/dapr/quickstarts/issues/361
https://github.com/dapr/quickstarts/issues/388
https://github.com/dapr/quickstarts/issues/399 (May also be a general failure to start daprd?)
https://github.com/dapr/python-sdk/issues/146
... and a bunch more across various SDKs/docs repos

We've root caused most of these to security/VPN configurations blocking mdns requests. The fix varies heavily based on which OS/security software the user is using. It would be lovely to have a generic solution to this, either by falling back to another service discovery mechanism if mdns doesn't work, or replacing mdns outright with something else.

## Release Note

RELEASE NOTE:

ItalyPaleAle:
Fixes #3256

> This PR updates components-contrib to the latest commit, so it also requires registering the new Azure Blob Storage v2 components

Addresses a longstanding request (> 2 years!) to have an alternative name resolver component for local development, for situations where mDNS can't be used (and even in those situations, a local name resolver is more convenient for local development).

As discussed in dapr/components-contrib#1380, this is implemented with SQLite, which is fast and efficient.


go.mod
```diff
@@ -453,3 +453,6 @@ replace github.com/microcosm-cc/bluemonday => github.com/microcosm-cc/bluemonday
 //
 // Then, run `make modtidy` in this repository.
 // This ensures that go.mod and go.sum are up-to-date.
+
+// TODO: REMOVE
+replace github.com/dapr/components-contrib => github.com/italypaleale/dapr-components-contrib v1.10.2-0.20231013225904-b55509fe9b3a
```
ItalyPaleAle, line 458: Remove when dapr/components-contrib#3178 is merged

pkg/messaging/direct_messaging.go
```diff
@@ -566,7 +566,7 @@ func (d *directMessaging) getRemoteApp(appID string) (remoteApp, error) {
 		address = d.checkHTTPEndpoints(id)
 	} else {
 		request := nr.ResolveRequest{ID: id, Namespace: namespace, Port: d.grpcPort}
-		address, err = d.resolver.ResolveID(request)
+		address, err = d.resolver.ResolveID(context.TODO(), request)
```
JoshVanL, line 569: Can we update `getRemoteApp` to accept a context?

JoshVanL (changes requested):
One thing, otherwise LGTM :+1:

pkg/messaging/direct_messaging.go
ItalyPaleAle: I was thinking this would be a separate PR (hence the TODO).

ItalyPaleAle:
The PR in contrib was merged so this is good to go!! 🚀

> Could you add a simple service invocation integration test using SQLite?

No, we should not use tests in the runtime to validate a component. There are _some_ “ITs” for this component in components-contrib (implemented as unit tests because we don’t have an infra to do otherwise). We should add an infra in components-contrib to test nameresolvers, also for the other nameresolvers that don’t have a test (Consul)

pkg/runtime/runtime.go
```diff
@@ -1096,33 +1098,36 @@ func (a *DaprRuntime) initNameResolution() error {
...
 	if a.globalConfig.Spec.NameResolutionSpec != nil {
 		resolverMetadata.Configuration = a.globalConfig.Spec.NameResolutionSpec.Configuration
 	}
-	resolverMetadata.Properties = map[string]string{
-		nr.DaprHTTPPort: strconv.Itoa(a.runtimeConfig.httpPort),
-		nr.DaprPort:     strconv.Itoa(a.runtimeConfig.internalGRPCPort),
-		nr.AppPort:      strconv.Itoa(a.runtimeConfig.appConnectionConfig.Port),
-		nr.HostAddress:  a.hostAddress,
-		nr.AppID:        a.runtimeConfig.id,
+	resolverMetadata.Instance = nr.Instance{
```
ItalyPaleAle, line 1112: A small API change in contrib. Now properties such as address and port are passed in a strongly-typed struct rather than in a weakly-typed map.

go.mod
```diff
@@ -453,3 +453,6 @@ replace github.com/microcosm-cc/bluemonday => github.com/microcosm-cc/bluemonday
 //
 // Then, run `make modtidy` in this repository.
 // This ensures that go.mod and go.sum are up-to-date.
+
+// TODO: REMOVE WHEN https://github.com/dapr/components-contrib/pull/3178 IS MERGED
+replace github.com/dapr/components-contrib => github.com/italypaleale/dapr-components-contrib v1.10.2-0.20231015174010-5746c8f67dcc
```
yaron2, line 458: https://github.com/dapr/components-contrib/pull/3178 is merged, lets remove the replace command.

yaron2 (dismissed):

go.mod
ItalyPaleAle: Sorry, I forgot to "git push"! Fixing now

yaron2:
Integration tests seem to be consistently failing. I reran the tests 3 times. Are there known issues in master branch with integration tests? cc @JoshVanL

ItalyPaleAle:
> Integration tests seem to be consistently failing. I reran the tests 3 times. Are there known issues in master branch with integration tests? cc @JoshVanL

Looks like the test that's failing is the reminders rebalancing, and it's failing consistently in this PR and also when I run it locally. It appears the issue is that placement table dissemination doesn't happen in time.

It's really weird as this PR doesn't touch anything related to placement 😮 

I am investigating

ItalyPaleAle:
The issue seems to be an unrelated bug introduced by a separate commit in components-contrib. I'm going to fix or revert that commit.

yaron2 (approved):

PR #7040, opened 2023-10-14
Metadata API: report actor runtime status

ItalyPaleAle:
> I promise this PR is not _actually_ 1,600 LOCs! Almost all of them are due to changes in the generated proto files, which are huge for some reason

Updates the metadata API to include additional information on the actor runtime and its status. This is helpful to understand if the actor runtime has completed the initialization, whether the actor runtime is an actor host or just an actor client, and whether the connection with the placement service is active.

Examples (only added data is shown):

1. While the (actor) runtime is initializing (i.e. before the request for /dapr/config has been completed):

   ```json
   {
     "actorRuntime": {
        "runtimeStatus": "INITIALIZING",
        "hostReady": false
      }
   }
   ```

2. Actor runtime has initialized and the app is hosting actors:

   ```json
   {
     "actorRuntime": {
       "runtimeStatus": "RUNNING",
       "activeActors": [
         {
           "type": "actor-1",
           "count": 1
         },
         {
           "type": "actor-2"
         }
       ],
       "hostReady": true,
       "placement": "placement: connected"
     }
   }
   ```

3. Actor runtime has initialized as actor client only (e.g. there's no "actorStateStore" configured):

   ```json
   {
     "actorRuntime": {
       "runtimeStatus": "RUNNING",
       "hostReady": false,
       "placement": "placement: connected"
     }
   }
   ```

4. Actor runtime has initialized but connection to placement is down:

   ```json
   {
     "actorRuntime": {
       "runtimeStatus": "RUNNING",
       "activeActors": [
         {
           "type": "actor-1",
           "count": 1
         },
         {
           "type": "actor-2"
         }
       ],
       "hostReady": false,
       "placement": "placement: disconnected"
     }
   }
   ```

5. Actor runtime is disabled (no `placement-host-address` option)

   ```json
   {
     "actorRuntime": {
       "runtimeStatus": "DISABLED",
       "hostReady": false
     }
   }
   ```

The `actorRuntime.activeActors` property is an alias of the pre-existing `actors` property, which contains a count of active actors. The old property is deprecated but not removed for backwards-compatibility (so the data is included twice, at least for now).


dapr/proto/runtime/v1/dapr.proto
```diff
@@ -592,14 +592,33 @@ message InvokeActorResponse {
...
   AppConnectionProperties app_connection_properties = 7 [json_name = "appConnectionProperties"];
   string runtime_version = 8 [json_name = "runtimeVersion"];
   repeated string enabled_features = 9 [json_name = "enabledFeatures"];
+  ActorRuntime actor_runtime = 10 [json_name = "actorRuntime"];
+}
+
+message ActorRuntime {
+  enum ActorRuntimeStatus {
+    INITIALIZING = 0;
+    DISABLED = 1;
```
daixiang0, line 610: Does actor have completed status, not familiar with it.

ItalyPaleAle: This is about the actor runtime, not individual actors. It is just meant to indicate whether actors can be used (because they are initialized asynchronously in background).

tests/integration/suite/actors/metadata/actor_client.go
```diff
@@ -0,0 +1,101 @@
...
+	assert.False(t, res.ActorRuntime.HostReady)
+	assert.Empty(t, res.ActorRuntime.Placement)
+	assert.Empty(t, res.ActorRuntime.ActiveActors)
+
+	// Complete init
+	close(m.blockConfig)
+	assert.EventuallyWithT(t, func(t *assert.CollectT) {
+		res := getMetadata(t, ctx, client, m.daprd.HTTPPort())
+		assert.Equal(t, "RUNNING", res.ActorRuntime.RuntimeStatus)
+		assert.False(t, res.ActorRuntime.HostReady)
```
mukundansundar, line 97: clarification question: Based on https://github.com/dapr/dapr/pull/7040/files#diff-ab620a0df90b230287b642dd2222bc40d6092a80d11092cd654da649de77d85fR137 

When app is healthy and placement is connected, host must be ready ... 
So why is that not the case here?

pkg/actors/actors.go
```diff
@@ -972,7 +972,21 @@ func (a *actorsRuntime) RegisterInternalActor(ctx context.Context, actorType str
...
 
-func (a *actorsRuntime) GetActiveActorsCount(ctx context.Context) []*runtimev1pb.ActiveActorsCount {
+func (a *actorsRuntime) GetRuntimeStatus(ctx context.Context) *runtimev1pb.ActorRuntime {
+	// Do not populate RuntimeStatus, which will be populated by the runtime
+	res := &runtimev1pb.ActorRuntime{
+		ActiveActors: a.getActiveActorsCount(ctx),
+	}
+
+	if a.placement != nil {
+		res.HostReady = a.placement.ActorHostReady() && a.haveCompatibleStorage()
```
mukundansundar, line 982: either the name `a.placement.ActorHostReady()` is misleading or a host is considered ready only when there is compatible storage. 
In that case, a separate function should be there in this file calling both `a.placement.ActorHostReady()` and the `a.haveCompatibleStorage()` methods. 
Additionally, `a.placement.ActorHostReady()` should be renamed to something different.

mukundansundar (changes requested):

dapr/proto/runtime/v1/dapr.proto
```diff
@@ -592,14 +592,33 @@ message InvokeActorResponse {
 // GetMetadataResponse is a message that is returned on GetMetadata rpc call
 message GetMetadataResponse {
   string id = 1;
-  repeated ActiveActorsCount active_actors_count = 2 [json_name = "actors"];
+  // Deprecated alias for actor_system.active_actors.
```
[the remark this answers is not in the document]
ItalyPaleAle: Thanks for catching this

```suggestion
  // Deprecated alias for actor_runtime.active_actors.
```

dapr/proto/runtime/v1/dapr.proto
```diff
@@ -592,14 +592,33 @@ message InvokeActorResponse {
...
+  }
+
+  // Contains an enum indicating whether the actor runtime has been initialized.
+  ActorRuntimeStatus runtime_status = 1 [json_name = "runtimeStatus"];
+  // Count of active actors per type.
+  repeated ActiveActorsCount active_actors = 2 [json_name = "activeActors"];
+  // Indicates whether the actor runtime is ready to host actors.
+  bool host_ready = 3 [json_name = "hostReady"];
+  // Custom message from the placement provider.
+  string placement = 4 [json_name = "placement"];
```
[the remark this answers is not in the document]
ItalyPaleAle: this was left as a string on purpose, as implementations of placement can define their own status which can depend on the implementation.

tests/integration/suite/actors/metadata/actor_client.go
ItalyPaleAle: See my response on the other comment below

pkg/actors/actors.go
ItalyPaleAle: This is a good point. I think I overloaded the expression "actor host ready" indeed.

- For placement, the host is ready when the connection with placement is up and the app is healthy. In this case, "actor host ready" was probably not the correct name
- For the actor runtime, "actor host ready" was meant to indicate that the runtime can host actors.

I will rename the method in placement.

ItalyPaleAle: Done

tests/integration/suite/actors/metadata/actor_client.go
mukundansundar: Ok I see now, that a state store is not configured for this test which I see in the `actor_host.go`. That would then make the host not ready.

mukundansundar (approved):
lgtm

mukundansundar:
@ItalyPaleAle please fix the conflicts in the PR.

ItalyPaleAle:
> @ItalyPaleAle please fix the conflicts in the PR.

Done

dapr/proto/runtime/v1/dapr.proto
```diff
@@ -595,14 +595,33 @@ message GetMetadataRequest {
...
 }
```
mukundansundar, line 625: should the actor api level also be included in this ?

ItalyPaleAle: I think so, but it’s currently not used. When the placement does report the min actor level, then we can add it.

tests/integration/suite/actors/metadata/actor_client.go
```diff
@@ -0,0 +1,101 @@
+/*
```
JoshVanL, line 1: Remove `actor_` prefix to files in directory as they are superfluous being in the `actors` top level suite category.

tests/integration/suite/actors/metadata/shared.go
```diff
@@ -0,0 +1,71 @@
...
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+)
+
+// In-memory state store component.
+const stateStore = `apiVersion: dapr.io/v1alpha1
```
JoshVanL, line 27: Consider creating a `WithActorInMemStore()` daprd process options func to prevent duplicating this here (and in other integration actor tests which all write this out).

tests/integration/framework/process/daprd/daprd.go
```diff
@@ -186,7 +199,7 @@ func (d *Daprd) WaitUntilAppHealth(t *testing.T, ctx context.Context) {
 			}
 			defer resp.Body.Close()
 			return http.StatusNoContent == resp.StatusCode
-		}, time.Second*10, 100*time.Millisecond)
+		}, 10*time.Second, 100*time.Millisecond)
```
JoshVanL, line 202: Big nit: if we want to be consistent on precision and magnitude ordering, can we have magnitude on the right? This can read better on multiple lines as the column matches up on precious regardless of magnitude, i.e. 

```
12934*time.Second
12*time.Second
```

vs

```
time.Second*12934
time.Second*12
```

tests/integration/suite/actors/metadata/actor_disabled.go
```diff
@@ -0,0 +1,76 @@
...
+	"github.com/dapr/dapr/tests/integration/framework/util"
+	"github.com/dapr/dapr/tests/integration/suite"
+)
+
+func init() {
+	suite.Register(new(metadataActorDisabled))
+}
+
+// metadataActorDisabled tests the response of the metadata API when the actor runtime is disabled
+type metadataActorDisabled struct {
```
JoshVanL, line 36:
```suggestion
type disabled struct {
```

JoshVanL (changes requested):

tests/integration/suite/actors/metadata/actor_client.go
ItalyPaleAle: Done

pkg/http/api_metadata.go
```diff
@@ -155,6 +165,14 @@ type metadataResponse struct {
...
 	HTTPEndpoints           []*runtimev1pb.MetadataHTTPEndpoint     `json:"httpEndpoints,omitempty"`
 	AppConnectionProperties metadataResponseAppConnectionProperties `json:"appConnectionProperties,omitempty"`
+	ActorRuntime            metadataActorRuntime                    `json:"actorRuntime,omitempty"`
+}
+
+type metadataActorRuntime struct {
+	Status       string                           `json:"runtimeStatus"`
+	ActiveActors []*runtimev1pb.ActiveActorsCount `json:"activeActors,omitempty"`
+	HostReady    bool                             `json:"hostReady"`
+	Placement    string                           `json:"placement,omitempty"`
```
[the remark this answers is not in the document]
ItalyPaleAle: This isn't a new change, but regardless:

- The main struct is in the proto.
- When the API was converted to universal, the gRPC and HTTP responses were different, so we couldn't just return the gRPC response encoded with protojson
- So we had to create this struct which mimics what's in the proto

I don't love it, but we can't change this much.

tests/integration/framework/process/daprd/daprd.go
ItalyPaleAle: TBH I prefer the opposite. In natural language, we say "10 seconds". it's also common in the SI to put units after the number, for everything except for currencies in North America ($10 but it's 10€)

tests/integration/suite/actors/metadata/actor_client.go
```diff
@@ -0,0 +1,101 @@
...
+	"github.com/dapr/dapr/tests/integration/framework/util"
+	"github.com/dapr/dapr/tests/integration/suite"
+)
+
+func init() {
+	suite.Register(new(metadataActorClient))
+}
+
+// metadataActorClient tests the response of the metadata API for a healthy actor client.
+type metadataActorClient struct {
```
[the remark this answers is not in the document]
ItalyPaleAle: Done

tests/integration/suite/actors/metadata/shared.go
ItalyPaleAle: Done

mukundansundar:
@JoshVanL can you review this again?

tests/integration/framework/process/daprd/options.go
```diff
@@ -136,6 +136,21 @@ func WithResourceFiles(files ...string) Option {
 	}
 }
 
+// WithInMemoryStore adds an in-memory state store component, which is also enabled as actor state store.
+func WithInMemoryStore(storeName string) Option {
```
JoshVanL, line 140: Sorry, my fault when I suggested the name. Can we include `ActorState` to be explicit?

```suggestion
func WithInMemoryActorStateStore(storeName string) Option {
```

JoshVanL (changes requested):

dapr/proto/runtime/v1/dapr.proto
```diff
@@ -595,14 +595,33 @@ message GetMetadataRequest {
...
+  }
+
+  // Contains an enum indicating whether the actor runtime has been initialized.
+  ActorRuntimeStatus runtime_status = 1 [json_name = "runtimeStatus"];
+  // Count of active actors per type.
+  repeated ActiveActorsCount active_actors = 2 [json_name = "activeActors"];
+  // Indicates whether the actor runtime is ready to host actors.
+  bool host_ready = 3 [json_name = "hostReady"];
+  // Custom message from the placement provider.
+  string placement = 4 [json_name = "placement"];
```
DeepanshuA, line 624: As it is a message received from placement, should the name be more explicit? `placement_msg`

dapr/proto/runtime/v1/dapr.proto
```diff
@@ -595,14 +595,33 @@ message GetMetadataRequest {
...
+  enum ActorRuntimeStatus {
+    INITIALIZING = 0;
+    DISABLED = 1;
+    RUNNING = 2;
+  }
+
+  // Contains an enum indicating whether the actor runtime has been initialized.
+  ActorRuntimeStatus runtime_status = 1 [json_name = "runtimeStatus"];
+  // Count of active actors per type.
+  repeated ActiveActorsCount active_actors = 2 [json_name = "activeActors"];
```
DeepanshuA, line 620: Just trying to understand the context: Why knowing count would be useful here? Some debugging purpose to know of load?
I see that it is already there but just trying to understand the relevance of it.

dapr/proto/runtime/v1/dapr.proto
```diff
@@ -595,14 +595,33 @@ message GetMetadataRequest {
...
   AppConnectionProperties app_connection_properties = 7 [json_name = "appConnectionProperties"];
   string runtime_version = 8 [json_name = "runtimeVersion"];
   repeated string enabled_features = 9 [json_name = "enabledFeatures"];
+  ActorRuntime actor_runtime = 10 [json_name = "actorRuntime"];
+}
+
+message ActorRuntime {
+  enum ActorRuntimeStatus {
+    INITIALIZING = 0;
+    DISABLED = 1;
```
DeepanshuA, line 613: Do you think a comment here would help, which can explain a bit that in what cases ActorRuntime can be Disabled?

dapr/proto/runtime/v1/dapr.proto
ItalyPaleAle: I think the "msg" is redundant IMHO. Just sending more bytes over the wire (when using JSON).

dapr/proto/runtime/v1/dapr.proto
ItalyPaleAle: I don't fully understand it either. But it was there before, and I didn't add it - I just added it here too for consistency.

dapr/proto/runtime/v1/dapr.proto
ItalyPaleAle: Yes comments would help. I'll add them

ItalyPaleAle: Done

mukundansundar (approved):

mukundansundar (approved):

PR #7041, opened 2023-10-15
Injector: Change daprd projected token audience to sentry SPIFFE ID

JoshVanL:
Inject the SPIFFE ID of sentry as the daprd sentry token audience over the legacy `dapr.io/sentry`. More secure as it is more specific to the dapr cluster installation control plane trust domain and control plane namespace. The control plane components are already using the sentry SPIFFE ID of sentry as their identity token audience.

The legacy `dapr.io/sentry` audience continues to be accepted by sentry for backwards compatibility. 

Part of #5756

### DIFF
--- a/pkg/injector/patcher/sidecar_volumes.go
+++ b/pkg/injector/patcher/sidecar_volumes.go
@@ -20,7 +20,6 @@ import (
 	corev1 "k8s.io/api/core/v1"
 
 	injectorConsts "github.com/dapr/dapr/pkg/injector/consts"
-	securityConsts "github.com/dapr/dapr/pkg/security/consts"
 	"github.com/dapr/kit/ptr"
 )
 
@@ -78,7 +77,7 @@ func (c *SidecarConfig) getTokenVolume() corev1.Volume {
 				DefaultMode: ptr.Of(int32(420)),
 				Sources: []corev1.VolumeProjection{{
 					ServiceAccountToken: &corev1.ServiceAccountTokenProjection{
-						Audience:          securityConsts.ServiceAccountTokenAudience,
+						Audience:          c.SentrySPIFFEID.String(),
 						ExpirationSeconds: ptr.Of(int64(7200)),
 						Path:              "token",
 					},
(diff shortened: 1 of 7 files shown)


pkg/injector/patcher/sidecar_volumes.go
```diff
@@ -78,7 +77,7 @@ func (c *SidecarConfig) getTokenVolume() corev1.Volume {
 				DefaultMode: ptr.Of(int32(420)),
 				Sources: []corev1.VolumeProjection{{
 					ServiceAccountToken: &corev1.ServiceAccountTokenProjection{
-						Audience:          securityConsts.ServiceAccountTokenAudience,
+						Audience:          c.SentrySPIFFEID.String(),
```
ItalyPaleAle, line 80: Can you make this an option passed to SidecarConfig as a string? For example just "Audience" and then the caller will pass a SPIFFE ID. Since the injector can be used as a library

ItalyPaleAle (changes requested):

pkg/injector/patcher/sidecar_volumes.go
JoshVanL: We need to use the spiffe library as it is important that the Sentry SPIFFE ID is validated as a well formed string. It makes little difference to the go modules of the injector when used as a library. The SPIFFE library itself is small.

```
$ diff go.mod go.mod_
12a13
> 	github.com/spiffe/go-spiffe/v2 v2.1.6
$ diff go.sum go.sum_
131a132,133
> github.com/spiffe/go-spiffe/v2 v2.1.6 h1:4SdizuQieFyL9eNU+SPiCArH4kynzaKOOj0VvM8R7Xo=
> github.com/spiffe/go-spiffe/v2 v2.1.6/go.mod h1:eVDqm9xFvyqao6C+eQensb9ZPkyNEeaUbqbBpOhBnNk=
```

ItalyPaleAle: Yes I am not saying don't pass the SPIFFE ID, I am just saying make `c.SentrySPIFFEID` of type `String()`. Then do the `.String()` call before passing the value to SidecarConfig.

JoshVanL: Gotcha, done

yaron2 (approved):

PR #7042, opened 2023-10-15
Update GitHub verison-skew workflow to use 1.12 setup components

JoshVanL:
Updates version skew workflow to use `1.12` setup components targets.

This unblocks running version skew tests on master which is currently broken because we are using setup components for `1.11`.

Demo of PR working on my fork: https://github.com/JoshVanL/dapr/actions/runs/6526229680
Looks like we have a failing backwards compat with Dapr in master


JoshVanL:
Demo of PR working on my fork: https://github.com/JoshVanL/dapr/actions/runs/6526229680
Looks like we have a failing backwards compat with Dapr in master

artursouza (approved):

Issue #6840, opened 2023-08-22
Resiliency - user-defined retry policy does not honor http status code 450


sol-lee:
While using the user-defined retry policy, its behavior and logic do not align with the default retry policy. It does not honor the HTTP status code 450. Please find the gap in "Expected Behavior" and "Actual Behavior" sections for details. 
I'm wondering is it possible to - 
1. Make the user-defined retry behavior the same as the default one?
2. In a future version, is it possible to add a feature that can disable the resiliency feature (turn off the DaprBuiltInServiceRetries default policy)? The sole reason I'm using a user-fined retry policy is to make the maxRetries=0 that turns off the retry feature.  For reference, please see:  @https://docs.dapr.io/operations/resiliency/policies/

## In what area(s)?

/area runtime

## What version of Dapr?

1.10.4

## Expected Behavior

Given there're 2 microservices: (A)personal and (B)bank, we make a service invocation from (A)personal to (B)bank. So it's like 
`(A)personal -> personal sidecar -> bank sidecar -> (B)bank`
The (B) bank is expected to return a response with HTTP status 450. The "bank sidecar" and "personal sidecar" return HTTP status 450 as well accordingly, which is expected. 

**"bank sidecar: HTTP code 450"**
![image](https://github.com/dapr/dapr/assets/13745612/7598e6e4-072a-40b1-af70-7823a5c97aa4)

**"personal sidecar: HTTP code 450"**
![image](https://github.com/dapr/dapr/assets/13745612/f7657343-f55d-4bf4-9cac-878ff00cc6bb)

## Actual Behavior

Given there're 2 microservices: (A)personal and (B)bank, we make a service invocation from (A)personal to (B)bank. So it's like 
`(A)personal -> personal sidecar -> bank sidecar -> (B)bank`
We configure a [user-defined retry policy](https://docs.dapr.io/operations/resiliency/policies/#retries) as below:

```
apiVersion: dapr.io/v1alpha1
kind: Resiliency
metadata:
  name: myresiliency
scopes:
spec:
  policies:
    retries:
      DefaultAppRetryPolicy:
        policy: constant
        duration: 1s
        maxRetries: 3
  targets:
    apps:
```

[...]

Issue #5872, opened 2023-02-01
Resiliency: respect "Retry-After" delays set in 429 responses


ItalyPaleAle:
## In what area(s)?

/area runtime

## Describe the feature

(Spin-off from #5862)

When resiliency is enabled and has retries configured, if the endpoint responds with a 429 ("too many requests") status code and it contains a Retry-After header, it should take that into account when deciding the interval to retry.

Assuming both a delay is set in the retry policy, and a Retry-After header is present, the amount of time to wait should be the maximum of the two.

## Release Note

RELEASE NOTE: **ADD** Respect "Retry-After" headers in 429 responses when retrying a request with Resiliency

ItalyPaleAle:
This is pending on https://github.com/cenkalti/backoff/issues/134

PR #7052, opened 2023-10-17
Fix: return HTTP status code in HTTP service invocation with resiliency enabled

shivamkm07:
# Description

Currently With Resiliency enabled, in case of http service invocation, if one application sends error status codes (http code <200 or >399), Dapr instead of forwarding the received status code, converts it to 500(See #6840). This doesn't seem to be intended behavior as well(See the comment here: https://github.com/dapr/dapr/blob/725e7d6865e35726104b856208e2d60f1a583c75/pkg/http/api_directmessaging.go#L147). Also it makes dapr behave differently with and without resiliency enabled(#6840)

This PR adds a custom error object to handle the scenario for incorrect status codes. It also adds the unit test as well as integration test for the same.

## Issue reference

Please reference the issue this PR will close: #6840 

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing

### DIFF
--- a/pkg/http/api_directmessaging.go
+++ b/pkg/http/api_directmessaging.go
@@ -322,5 +322,5 @@ type codeError struct {
 }
 
 func (ce codeError) Error() string {
-	return fmt.Sprintf("received non-successful status code in response: statusCode='%d' msg='%v'", ce.statusCode, string(ce.msg))
+	return fmt.Sprintf("received non-successful status code in response: %d", ce.statusCode)
 }
--- a/pkg/http/api_directmessaging.go
+++ b/pkg/http/api_directmessaging.go
@@ -322,5 +322,5 @@ type codeError struct {
 }
 
 func (ce codeError) Error() string {
-	return fmt.Sprintf("CodeError: received non-successful status code (statusCode='%d') msg='%v'", ce.statusCode, string(ce.msg))
+	return fmt.Sprintf("received non-successful status code in response: statusCode='%d' msg='%v'", ce.statusCode, string(ce.msg))
 }
--- a/pkg/http/api_directmessaging.go
+++ b/pkg/http/api_directmessaging.go
@@ -322,5 +322,5 @@ type codeError struct {
 }
 
 func (ce codeError) Error() string {
-	return fmt.Sprintf("codeError (statusCode='%d') msg='%v'", ce.statusCode, string(ce.msg))
+	return fmt.Sprintf("CodeError: received non-successful status code (statusCode='%d') msg='%v'", ce.statusCode, string(ce.msg))
 }
--- a/pkg/http/api_directmessaging.go
+++ b/pkg/http/api_directmessaging.go
@@ -98,11 +98,8 @@ func (a *api) onDirectMessage(w http.ResponseWriter, r *http.Request) {
 	}
 	defer req.Close()
 
-	policyRunner := resiliency.NewRunnerWithOptions(
+	policyRunner := resiliency.NewRunner[*invokev1.InvokeMethodResponse](
 		r.Context(), policyDef,
-		resiliency.RunnerOpts[*invokev1.InvokeMethodResponse]{
-			Disposer: resiliency.DisposerCloser[*invokev1.InvokeMethodResponse],
-		},
 	)
 	// Since we don't want to return the actual error, we have to extract several things in order to construct our response.
 	resp, err := policyRunner(func(ctx context.Context) (*invokev1.InvokeMethodResponse, error) {
--- a/pkg/http/api_directmessaging.go
+++ b/pkg/http/api_directmessaging.go
@@ -98,8 +98,11 @@ func (a *api) onDirectMessage(w http.ResponseWriter, r *http.Request) {
 	}
 	defer req.Close()
 
-	policyRunner := resiliency.NewRunner[*invokev1.InvokeMethodResponse](
+	policyRunner := resiliency.NewRunnerWithOptions(
 		r.Context(), policyDef,
+		resiliency.RunnerOpts[*invokev1.InvokeMethodResponse]{
+			Disposer: resiliency.DisposerCloser[*invokev1.InvokeMethodResponse],
+		},
 	)
 	// Since we don't want to return the actual error, we have to extract several things in order to construct our response.
 	resp, err := policyRunner(func(ctx context.Context) (*invokev1.InvokeMethodResponse, error) {
@@ -140,9 +143,14 @@ func (a *api) onDirectMessage(w http.ResponseWriter, r *http.Request) {
 				resStatus.Code = statusCode
 			}
 		} else if resStatus.Code < 200 || resStatus.Code > 399 {
-			// We are not returning an `invokeError` here on purpose.
-			// Returning an error that is not an `invokeError` will cause Resiliency to retry the request (if retries are enabled), but if the request continues to fail, the response is sent to the user with whatever status code the app returned so the "received non-successful status code" is "swallowed" (will appear in logs but won't be returned to the app).
-			return rResp, fmt.Errorf("received non-successful status code: %d", resStatus.Code)
+			msg, _ := rResp.RawDataFull()
+			// Returning a `codeError` here will cause Resiliency to retry the request (if retries are enabled), but if the request continues to fail, the response is sent to the user with whatever status code the app returned.
+			return rResp, codeError{
+				headers:     rResp.Headers(),
+				statusCode:  int(resStatus.Code),
+				msg:         msg,
+				contentType: rResp.ContentType(),
+			}
 		}
 		return rResp, nil
 	})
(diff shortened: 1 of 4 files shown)


pkg/http/api_directmessaging.go
```diff
@@ -98,11 +98,8 @@ func (a *api) onDirectMessage(w http.ResponseWriter, r *http.Request) {
 	}
 	defer req.Close()
 
-	policyRunner := resiliency.NewRunnerWithOptions(
+	policyRunner := resiliency.NewRunner[*invokev1.InvokeMethodResponse](
 		r.Context(), policyDef,
-		resiliency.RunnerOpts[*invokev1.InvokeMethodResponse]{
-			Disposer: resiliency.DisposerCloser[*invokev1.InvokeMethodResponse],
```
ItalyPaleAle, line 104: Sadly we do need this disposer, it's very important because otherwise we would not clean up HTTP responses (the disposer invokes `.Close()` on the response objects)

I'm not sure why this does the trick in your case, we can chat if you'd like! But the disposer has got to remain.

ItalyPaleAle (changes requested):

pkg/http/api_directmessaging.go
shivamkm07: We are invoking `.Close()`  lateron on the recieved response message: https://github.com/dapr/dapr/blob/725e7d6865e35726104b856208e2d60f1a583c75/pkg/http/api_directmessaging.go#L179
I understand still it won't close the other resp objects created during retries, but that is the only cleaner solution I can think of. Basically we need the response object in the end, to forward the received message and statuscode and disposer cleans that only. Do you have any other solution? We can chat for sure!

ItalyPaleAle: Yes, we are invoking `Close()` later, but that's only invoked on the successful response.

The point of the disposer is that if a request fails with a timeout, the disposer is the only thing we have that performs a cleanup of the resources, which otherwise would leak.

skyao: can we update this disposer to let it forward the http status code?

ItalyPaleAle: No, the disposer isn’t aware of what the response is. Also the disposer is invoked on failed (like timed out) responses too

But the way to do this could be creating a custom Error object that contains the status code we need.

shivamkm07: Pushed the alternate implementation using a custom error object. Please review

shivamkm07:
Although this is a buggy behavior and doesn't look like what we intended to do (See the description), but still this PR would change behavior of resiliency in case of error http codes. In the sense that if there are any users who are relying on generic `500` code for all errors, including error codes, it might break them right? Would this be considered as a breaking change then? IMO it shouldn't since it's a bug in the first place. Also, we already have user who are blocked because of this bug and aren't able to upgrade. But still adding it here for others' views. cc @ItalyPaleAle @mukundansundar @skyao

daixiang0:
Is it possible to add a field to detail the error?

shivamkm07:
> Is it possible to add a field to detail the error?

Are you asking about the `codeError` added in this PR? The details for that can be found in the `error()` method. I can update the returned string to give some more context about the error like `unsuccessful statuscode received`. But I don't think we should send the error detail to the user. This is the same behavior that is there right now.

shivamkm07:
> > Is it possible to add a field to detail the error?
> 
> Are you asking about the `codeError` added in this PR? The details for that can be found in the `error()` method. I can update the returned string to give some more context about the error like `unsuccessful statuscode received`. But I don't think we should send the error detail to the user. This is the same behavior that is there right now.

Done

pkg/http/api_directmessaging.go
```diff
@@ -292,3 +313,14 @@ type invokeError struct {
...
+
+type codeError struct {
+	statusCode  int
+	msg         []byte
+	headers     invokev1.DaprInternalMetadata
+	contentType string
+}
+
+func (ce codeError) Error() string {
+	return fmt.Sprintf("CodeError: received non-successful status code (statusCode='%d') msg='%v'", ce.statusCode, string(ce.msg))
```
ItalyPaleAle, line 325: Errors normally start with lowercased characters in Go.

```suggestion
	return fmt.Sprintf("received non-successful status code in response: statusCode='%d' msg='%v'", ce.statusCode, string(ce.msg))
```

ItalyPaleAle:
Interestingly, this behavior seems to have been present since 1.10. It wasn't in 1.9, but which did implement this incorrectly (there were race conditions in the code). The actual error message is returned if the response comes from gRPC, but not if it comes from HTTP.

IMHO, sounds like a bug. Possibly even a patch to backport?

pkg/testing/directmessaging_mock.go
```diff
@@ -85,9 +85,18 @@ func (_m *FailingDirectMessaging) Invoke(ctx context.Context, targetAppID string
 	if statusCode == 0 {
 		statusCode = http.StatusOK
 	}
+	//Setting up the headers passed in request
```
ItalyPaleAle, line 88: Linter
```suggestion
	// Setting up the headers passed in request
```

pkg/http/api_directmessaging.go
```diff
@@ -292,3 +313,14 @@ type invokeError struct {
...
+
+type codeError struct {
+	statusCode  int
+	msg         []byte
+	headers     invokev1.DaprInternalMetadata
+	contentType string
+}
+
+func (ce codeError) Error() string {
+	return fmt.Sprintf("received non-successful status code in response: statusCode='%d' msg='%v'", ce.statusCode, string(ce.msg))
```
ItalyPaleAle, line 325: On a second thought... Let's not log the message as it may contain PII. Sorry for the back-and-forth.

```suggestion
	return fmt.Sprintf("received non-successful status code in response: %d", ce.statusCode)
```

shivamkm07:
> Is it possible to have an integration test for this too?

@ItalyPaleAle Added the integration test as well.

tests/integration/suite/daprd/resiliency/apps/defaultretry.go
```diff
@@ -0,0 +1,138 @@
...
+      DefaultAppRetryPolicy:
+        policy: constant
+        duration: 100ms
+        maxRetries: 3
+  targets: {}
+`
+	d.daprd1 = daprd.New(t,
+		daprd.WithAppPort(srv.Port()),
+		daprd.WithResourceFiles(resiliency),
+		daprd.WithLogLevel("debug"),
```
JoshVanL, line 88: This probably isn't necessary 
```suggestion

```

JoshVanL (changes requested):
LGTM, just a small nit from me

tests/integration/suite/daprd/resiliency/apps/defaultretry.go
```diff
@@ -0,0 +1,137 @@
...
+	})
+
+	handler.HandleFunc("/retry", func(w http.ResponseWriter, r *http.Request) {
+		body, err := io.ReadAll(r.Body)
+		if err != nil {
+			w.WriteHeader(http.StatusInternalServerError)
+			return
+		}
+		key := string(body)
+		d.callCount[key]++
```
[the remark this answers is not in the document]
shivamkm07: Done

tests/integration/suite/daprd/resiliency/apps/defaultretry.go
```diff
@@ -0,0 +1,137 @@
...
+)
+
+func init() {
+	suite.Register(new(defaultretry))
+}
+
+type defaultretry struct {
+	daprd1    *daprd.Daprd
+	daprd2    *daprd.Daprd
+	callCount map[string]int
```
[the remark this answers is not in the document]
shivamkm07: Done

ItalyPaleAle (approved):
LGTM, thanks for this!

DeepanshuA (approved):

pkg/http/api_directmessaging.go
```diff
@@ -143,9 +143,14 @@ func (a *api) onDirectMessage(w http.ResponseWriter, r *http.Request) {
 				resStatus.Code = statusCode
 			}
 		} else if resStatus.Code < 200 || resStatus.Code > 399 {
```
DeepanshuA, line 145: Not a comment on the PR particularly: But, I want to understand dapr resiliency behavior in general: Here, we see that 4xx error are retried, wheras most of them are non-retriable (1 quick ref: https://stackoverflow.com/a/47841642/18848251). Is it correct to retry errors like 404?

mukundansundar: I think this is something that needs to be discussed separate from this PR. Can you create an issue regarding this?

mukundansundar (approved):
lgtm.
@shivamkm07 Can you update the description to match the change being made in the PR.

mukundansundar:
> Although this is a buggy behavior and doesn't look like what we intended to do (See the description), but still this PR would change behavior of resiliency in case of error http codes. In the sense that if there are any users who are relying on generic `500` code for all errors, including error codes, it might break them right? Would this be considered as a breaking change then? IMO it shouldn't since it's a bug in the first place. Also, we already have user who are blocked because of this bug and aren't able to upgrade. Also the documentation says the status code is honored: https://docs.dapr.io/reference/api/service_invocation_api/#http-response-codes. So it should be fine ig But still adding it here for others' views. cc @ItalyPaleAle @mukundansundar @skyao

I think this should be considered a bug as in essence we are swallowing up the actual error code and returning `500` in response instead. Specifically for service invocation, I believe this should not be the case.

pkg/http/api_directmessaging.go
ItalyPaleAle: I agree with @DeepanshuA this may be worth a conversation. There is also an issue that involves being "smarter" about retries depending on status codes (such as respecting Retry-after in 427 responses: #5872 )

Also agree with @mukundansundar it should be done in a separate issue.

shivamkm07:
> lgtm. @shivamkm07 Can you update the description to match the change being made in the PR.

Done

mukundansundar:
@shivamkm07 Please cherry pick this for 1.12 and 1.11 branches ...

Issue #7055, opened 2023-10-18
[P0] Dapr downgrade from 1.12.0 to 1.11.2 doesn't work


ASHIQUEMD:
## In what area(s)?

> /area runtime

## What version of Dapr?

> 1.12.0

## Expected Behavior

Downgrade to 1.11.2 should work seamlessly.

## Actual Behavior
When Dapr is downgraded from 1.12.0 to 1.11.2, dapr sidecar injector fails to inject the dapr sidecar due to tls issue.

```bash
# k logs dapr-sidecar-injector-59b4f68f6d-p6l28 -n dapr-system                                                     
time="2023-10-18T13:25:55.20080834Z" level=info msg="log level set to: info" instance=dapr-sidecar-injector-59b4f68f6d-p6l28 scope=dapr.injector type=log ver=1.11.2
time="2023-10-18T13:25:55.201074423Z" level=info msg="metrics server started on :9090/" instance=dapr-sidecar-injector-59b4f68f6d-p6l28 scope=dapr.metrics type=log ver=1.11.2
time="2023-10-18T13:25:55.201202215Z" level=info msg="starting Dapr Sidecar Injector -- version 1.11.2 -- commit 83ca1abb11ffe34211db55dcd36d96b94252827a" instance=dapr-sidecar-injector-59b4f68f6d-p6l28 scope=dapr.injector type=log ver=1.11.2
time="2023-10-18T13:25:55.207855631Z" level=warning msg="Unable to get SA tekton-pipelines:tekton-pipelines-controller UID" instance=dapr-sidecar-injector-59b4f68f6d-p6l28 scope=dapr.injector type=log ver=1.11.2
time="2023-10-18T13:25:55.207948173Z" level=info msg="Healthz server is listening on :8080" instance=dapr-sidecar-injector-59b4f68f6d-p6l28 scope=dapr.injector type=log ver=1.11.2
time="2023-10-18T13:25:55.208014465Z" level=info msg="Sidecar injector is listening on :4000, patching Dapr-enabled pods" instance=dapr-sidecar-injector-59b4f68f6d-p6l28 scope=dapr.injector type=log ver=1.11.2
2023/10/18 13:26:28 http: TLS handshake error from 10.244.0.1:58110: remote error: tls: bad certificate
```

** Note: Upgrading Dapr to 1.12.0 and app pods rollout restart, starts working again **

## Steps to Reproduce the Problem
```bash
# install latest dapr version
dapr init -k --runtime-version 1.12.0
[...]

PR #7057, opened 2023-10-18
[1.11] Change injector webhook config template to source caBundle from Secret

JoshVanL:
Source injector webhook CA bundle from webhook certificate secret, rather than existing webhook configuration to allow downgrades from 1.12.

The CA Bundle will change from the root CA to the injector serving certificate on subsequent runs of helm install, which doesn't have any impact on functionality or security.

See: https://github.com/dapr/dapr/issues/7055#issuecomment-1768573741


yaron2:
@JoshVanL please update once an upgrade from 1.11.4 to this version of 1.11 is verified, then we can merge.

JoshVanL:
> @JoshVanL please update once an upgrade from 1.11.4 to this version of 1.11 is verified, then we can merge.

I can confirm that upgrading from 1.11.4 to "1.114" with this patch works.

yaron2 (approved):

PR #7062, opened 2023-10-19
[1.10] Change injector webhook config template to source caBundle from Secret 

JoshVanL:
We may not consider merging and release a patch release for this, since downgrading from `1.12` directly to `1.10` is not fully supported, and considered best effort.

https://docs.dapr.io/operations/support/support-release-policy/#upgrade-paths

See: https://github.com/dapr/dapr/pull/7057


yaron2:
This upgrade path from 1.12 to 1.10  isn't needed nor recommended

PR #7069, opened 2023-10-21
Placement: Disseminate rate limiting

JoshVanL:
Currently, placement is using a [static 500ms](https://github.com/dapr/dapr/blob/a2270ea9baa1bcc8e9ea0773decf18a8d3ac5c1f/pkg/placement/placement.go#L64) interval to check whether it needs to update tables in the event that membership has changed. This is problematic because 1. continually checking every 500ms is wasteful whilst also slowing down a single instance joining placement, and 2. in the event of a large amount of churn in the Daprd cluster, results in a larger number of table writes which in itself is expensive.

PR updates the dissemination table channel to use a coalescing rate limiter. This has the benefit of both immediately joining new members if they join in isolation, as well as reducing the number of table updates in the event that there is a burst of members churning. The rate limit is exponential, starting at 200ms, maximum 4s. To prevent the table from ever being updated in the event of a large throughput, the maximum number of pending member updates is 5. If there are 5 pending membership updates the table is force updated regardless of the rate limiting status.

Adds actor invoke integration tests using 3 unique actor types on 3 daprds, in both HTTP and gRPC.

PR replaces kit with my personal fork which has the implementation of the rate limiter https://github.com/dapr/kit/pull/64. This PR should hold until that has been merged to `dapr/kit`.

### DIFF
--- a/cmd/placement/app/app.go
+++ b/cmd/placement/app/app.go
@@ -93,7 +93,10 @@ func Run() {
 	if opts.MaxAPILevel >= 0 && opts.MaxAPILevel < math.MaxInt32 {
 		placementOpts.MaxAPILevel = ptr.Of(uint32(opts.MaxAPILevel))
 	}
-	apiServer := placement.NewPlacementService(placementOpts)
+	apiServer, err := placement.NewPlacementService(placementOpts)
+	if err != nil {
+		log.Fatal(err)
+	}
 
 	err = concurrency.NewRunnerManager(
 		func(ctx context.Context) error {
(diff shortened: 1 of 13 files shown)


ItalyPaleAle:
Moving to draft until dapr/kit#64 is merged

elena-kolevska (dismissed):

yaron2:
@JoshVanL please resolve conflicts and see failing tests

yaron2 (dismissed):

PR #7074, opened 2023-10-21
Integration: Only write process logs on failure

JoshVanL:
PR updates the integration test io writer to by default, only write the exec process logs on test failure. This is in an effort to reduce the amount of IO done by the test runners. If a test fails, the processes logs will be output at the end. Logs can be enabled for all tests by setting the `DAPR_INTEGRATION_LOGS` environment variable to `true`.

### DIFF
--- a/tests/docs/writing-integration-test.md
+++ b/tests/docs/writing-integration-test.md
@@ -15,9 +15,7 @@ always built from source within the test.
 go test -v -race -count -tags="integration" ./tests/integration` -run="Test_Integration/daprd/pubsub/http/fuzzpubsubNoRaw"
 ```
 
-Rather than building from source, you can also set a custom daprd, sentry, or
-placement binary path
-with the environment variables
+Rather than building from source, you can also set a custom daprd, sentry, or placement binary path with the environment variables:
 - `DAPR_INTEGRATION_DAPRD_PATH`
 - `DAPR_INTEGRATION_PLACEMENT_PATH`
 - `DAPR_INTEGRATION_SENTRY_PATH`
--- a/tests/docs/writing-integration-test.md
+++ b/tests/docs/writing-integration-test.md
@@ -15,8 +15,16 @@ always built from source within the test.
 go test -v -race -count -tags="integration" ./tests/integration` -run="Test_Integration/daprd/pubsub/http/fuzzpubsubNoRaw"
 ```
 
-Rather than building from source, you can also set a custom daprd binary path
-with the environment variable `DAPR_INTEGRATION_DAPRD_PATH`.
+Rather than building from source, you can also set a custom daprd, sentry, or
+placement binary path
+with the environment variables
+- `DAPR_INTEGRATION_DAPRD_PATH`
+- `DAPR_INTEGRATION_PLACEMENT_PATH`
+- `DAPR_INTEGRATION_SENTRY_PATH`
+
+By default, binary logs will not be printed unless the test fails. You can force
+logs to always be printed with the environment variable
+`DAPR_INTEGRATION_LOGS=true`.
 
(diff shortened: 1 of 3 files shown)


tests/docs/writing-integration-test.md
```diff
@@ -15,8 +15,16 @@ always built from source within the test.
 go test -v -race -count -tags="integration" ./tests/integration` -run="Test_Integration/daprd/pubsub/http/fuzzpubsubNoRaw"
 ```
 
-Rather than building from source, you can also set a custom daprd binary path
-with the environment variable `DAPR_INTEGRATION_DAPRD_PATH`.
+Rather than building from source, you can also set a custom daprd, sentry, or
+placement binary path
+with the environment variables
```
daixiang0, line 20: No need a linebreak here.

tests/integration/framework/process/exec/iowriter/iowriter.go
```diff
@@ -69,9 +73,25 @@ func (w *stdwriter) flush() {
...
+	if !w.t.Failed() &&
+		strings.ToLower(os.Getenv("DAPR_INTEGRATION_LOGS")) != "true" {
+		return
+	}
+
+	for {
+		line, err := w.buf.ReadBytes('\n')
 		if len(line) > 0 {
-			w.t.Log(w.t.Name() + "/" + w.procName + ": " + string(line))
+			w.t.Log(w.t.Name() + "/" + w.procName + ": " +
```
daixiang0, line 87: Use `fmt.Sprintf`

tests/integration/framework/process/exec/iowriter/iowriter.go
```diff
@@ -69,9 +73,25 @@ func (w *stdwriter) flush() {
...
+				strings.TrimSuffix(string(line), "\n"))
+		}
+		if err != nil {
+			if !errors.Is(err, io.EOF) {
+				w.t.Log(w.t.Name() + "/" + w.procName + ": " + err.Error())
```
daixiang0, line 92: Same as above.

tests/integration/framework/process/exec/iowriter/iowriter_test.go
```diff
@@ -86,19 +91,56 @@ func TestWrite(t *testing.T) {
...
 func TestClose(t *testing.T) {
 	t.Run("should flush and close", func(t *testing.T) {
-		logger := &mockLogger{t: t}
+		logger := &mockLogger{t: t, failed: true}
+		writer := New(logger, "proc").(*stdwriter)
+		writer.Write([]byte("test"))
+		writer.Close()
+
+		assert.Equal(t, 0, writer.buf.Len())
+		_ = assert.Equal(t, 1, len(logger.msgs)) &&
```
daixiang0, line 100: Can not use two single assert lines? Why need `&&`

ItalyPaleAle: I presume it's to avoid a panic if len == 0. If so, trying to access `logger.msgs[0]` would panic

tests/docs/writing-integration-test.md
JoshVanL: :+1: done

tests/integration/framework/process/exec/iowriter/iowriter_test.go
JoshVanL: Yep exactly, it's to prevent the panic but don't want to use `require` necessarily because we want to continue the test. ^

mukundansundar (dismissed):

ItalyPaleAle (approved):

mukundansundar (approved):

Issue #6955, opened 2023-09-21
Verify local Golangci-lint version with GitHub actions


sicoyle:
## In what area(s)?

> /area runtime

**Is your feature request related to a problem? Please describe.**
We should have tooling to verify that the linter installed locally matches that of the Github actions to prevent confusion for contributors and issues showing up in CI that do not occur locally due to mismatching versions.

**Describe the solution you'd like**
A similar solution to that of the components contrib repo solution here: https://github.com/dapr/components-contrib/pull/1861/files

**Describe alternatives you've considered**
None

**Additional context**
Reference issue: https://github.com/dapr/cli/issues/1052
Inspiration for solution: https://github.com/dapr/components-contrib/pull/1861/files

## Release Note

RELEASE NOTE:

PR #7077, opened 2023-10-22
feat: implement local golangci-lint version checking

mikeee:
# Description

- Adds the check-lint-version tool (accessible through the build-tools - and associated lint workflow).
- Adds testing for the check-linter tool.
- Defines LF-style endings for the repo on checkout via `.gitattributes`

## Issue reference

Please reference the issue this PR will close: #6955

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Unit tests passing


tools/check-lint-version/Readme.md
```diff
@@ -0,0 +1,12 @@
+# Check Lint Version
```
ItalyPaleAle, line 1: Could you add this tool to the existing build tools in `.build-tools`? Could be another command in the same binary.

.github/workflows/test-tooling.yml
```diff
@@ -0,0 +1,59 @@
...
+jobs:
+  lint:
+    name: Test (${{ matrix.os}}) go ${{ matrix.gover }}
+    
+    strategy:
+      fail-fast: false
+      matrix:
+        gover:
+          - "1.21"
+        os:
```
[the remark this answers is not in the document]
mikeee: I think it should run on as many OSes as possible with this workflow already catching that LF-style endings were not applied on Windows - especially since this workflow is not resource intensive and only blocks ~1min of action runner time.

Makefile
```diff
@@ -409,7 +420,7 @@ modtidy-all: $(TIDY_MODFILES)
 modtidy:
 	go mod tidy
 
-################################################################################
+################################################################################f
```
ItalyPaleAle, line 423:
```suggestion
################################################################################
```

.github/workflows/test-tooling.yml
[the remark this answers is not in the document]
mikeee: I think it should run on as many OSes as possible with this workflow already catching that LF-style endings were not applied on Windows - especially since this workflow is not resource intensive and only blocks ~1min of action runner time. I believe this should be kept but have no overly strong feelings on it either way; I'm happy to remove if this is the consensus reached.

Makefile
```diff
@@ -378,9 +378,20 @@ test-integration: test-deps
...
+
+################################################################################
+# Target: check-linter                                                         #
+################################################################################
+.SILENT: check-linter # Silence output other than the application run
+.PHONY: check-linter
+check-linter:
+	cd ./.build-tools && \
+	go mod tidy && \
+	go run . check-linter
```
ItalyPaleAle, line 393: See tests/dapr_tests.mk for an example

```suggestion
	$(RUN_BUILD_TOOLS) check-linter
```

.build-tools/cmd/check-lint-version.go
```diff
@@ -0,0 +1,97 @@
...
+	res := semver.MajorMinor(workflowVersion) == semver.MajorMinor(currentVersion)
+	return res
+}
+
+func compareVersions(path string) string {
+	workflowVersion, err := parseWorkflowVersionFromFile(path)
+	if err != nil {
+		return fmt.Sprintf("Error parsing workflow version: %v", err)
+	}
+	currentVersion, err := getCurrentVersion()
```
ItalyPaleAle, line 68: If the command doesn't exist, should print a message indicating how to install the linter?

.build-tools/cmd/check-lint-version.go
```diff
@@ -0,0 +1,97 @@
...
+}
+
+func getCmdCheckLint(cmdType string) *cobra.Command {
+	// Base command
+	cmd := &cobra.Command{
+		Use:   cmdType,
+		Short: "Compare local golangci-lint version against workflow version",
+		Run: func(cmd *cobra.Command, args []string) {
+			path := cmd.Flag("path").Value.String()
+			fmt.Println(compareVersions(path))
```
ItalyPaleAle, line 86: Make the binary exit with non-0 status code if the comparison fails. You may need to change the signature of `compareVersions` to return an error, and if there's an error, exit with `os.Exit(1)`

.build-tools/cmd/check-lint-version_test.go
```diff
@@ -0,0 +1,88 @@
...
+
+	t.Run("parse an invalid workflow file", func(t *testing.T) {
+		parsedVersion, err := parseWorkflowVersionFromFile("../testdata/check-lint-version/invalid-yaml.yml")
+		assert.Equal(t, "", parsedVersion)
+		assert.Error(t, err)
+	})
+
+	t.Run("parse testing workflow file", func(t *testing.T) {
+		parsedVersion, err := parseWorkflowVersionFromFile("../../.github/workflows/test-tooling.yml")
+		assert.Equal(t, "v1.51.2", parsedVersion)
```
ItalyPaleAle, line 30: I would remove this test because it is tightly coupled with the version in the YAML

ItalyPaleAle (changes requested):

daixiang0:
Could we have a file to set golangci-lint version as general control?

.build-tools/cmd/check-lint-version_test.go
mikeee: I've migrated the version to the testdata folder to still validate parsing a workflow file

mikeee:
> Could we have a file to set golangci-lint version as general control?

We could do but I can't think of a way that doesn't add more complexity when having to pull the version back into the workflow. I've updated the error to include where to look if there is a version mismatch so it's not ambiguous which version it's being compared to if that helps.

.github/workflows/test-tooling.yml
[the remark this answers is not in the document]
ItalyPaleAle: This only runs if you change `.build-tools` and `.github/workflows/test-tooling.yml`, right?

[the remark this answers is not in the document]
ItalyPaleAle: Ok i think we can live with 3 OS's then IMHO

artursouza (approved):

daixiang0 (approved):
LGTM

Issue #7078, opened 2023-10-22
daprd:1.12.0 Fails to Retrieve the Initial Identity Certificate due to SENTRY_LOCAL_IDENTITY length > 64 characters


a-shoemaker:
## In what area(s)?

> /area runtime

## What version of Dapr?

> 1.12.0

## Expected Behavior

Ability to use dapr version 1.12 when SENTRY_LOCAL_IDENTITY is greater than 64 characters in length.

Sentry should be making adjustments to the request ID to turn it into an app ID before validating or alternatively adjust validation knowing of this possibility.  Potentially should change the code so the SignCertificateRequest always provides the app ID as the request's Id.

## Actual Behavior

daprd fails to startup when the SENTRY_LOCAL_IDENTITY is greater than 64 characters in length.

```
time="2023-10-22T14:55:46.984217502Z" level=fatal msg="Fatal error from runtime: failed to retrieve the initial identity certificate: error from sentry SignCertificate: rpc error: code = PermissionDenied desc = invalid request: app ID must be 64 characters or less" app_id={app ID} instance={instance name} scope=dapr.runtime type=log ver=1.12.0
```

## Steps to Reproduce the Problem

Upgrade to dapr 1.12.0 and restart a deployment with a SENTRY_LOCAL_IDENTITY (namespace + service account name) greater than 64 characters in length.

x509source.go provides ID as SENTRY_LOCAL_IDENTITY in requestFromSentry.  In pkg/sentry/server/validator/kubernetes/kuberentes.go validates the request at the beginning of Validate, which checks that the request Id is <= 64 characters, assuming it's the app ID.  Later in the method, it changes the request Id to the app ID when it was provided in the form of the SENTRY_LOCAL_IDENTITY:
```
	// TODO: @joshvanl: Before v1.12, the injector instructed daprd to request
	// for the ID containing their namespace and service account (ns:sa). This
	// is wrong- dapr identities are based on daprd namespace + _app ID_.
	// Remove this allowance in v1.13.
	if req.Namespace+":"+pod.Spec.ServiceAccountName == req.Id {
		req.Id = expID
	}
```

## Release Note

RELEASE NOTE:

JoshVanL:
To add a little bit more context from the the other issue: https://github.com/dapr/dapr/issues/7079

[...]

PR #7080, opened 2023-10-22
Fix sentry rejecting legacy request IDs of length larger than 64

JoshVanL:
Fixes https://github.com/dapr/dapr/issues/7078
Proof of fix: https://github.com/dapr/dapr/pull/7081

Fixes an issue whereby sentry would reject SignCertificate requests that have an identifier which was longer than 64 characters. Because sentry still accepts the "legacy" identifier format of `<namespace>:<service account>` (which v1.12 is currently using), Daprd's which are in a namespace and/or using a Service Account with a sufficiently long name will erroneously be rejected.

PR updates the Kubernetes sentry validator so that it will augment the request identifier to the app ID before evaluating whether the identifier is over 64 characters long.

Fix should be back-ported to v1.12 and patch released.

Adds sentry Kubernetes validator integration tests to cover this scenario.


pkg/sentry/sentry.go
```diff
@@ -158,16 +158,12 @@ func (s *sentry) getValidators(ctx context.Context) (map[sentryv1pb.SignCertific
...
-			if err != nil {
-				return nil, err
-			}
 			sentryID, err := security.SentryID(td, security.CurrentNamespace())
 			if err != nil {
 				return nil, err
 			}
 			val, err := validatorKube.New(ctx, validatorKube.Options{
-				RestConfig:     config,
+				RestConfig:     utils.GetConfig(),
```
JoshVanL, line 166: Non-functional change to enable Kubernetes validator integration tests.

pkg/sentry/server/ca/ca.go
```diff
@@ -89,11 +89,7 @@ func New(ctx context.Context, conf config.Config) (Signer, error) {
 	if config.IsKubernetesHosted() {
 		log.Info("Using kubernetes secret store for trust bundle storage")
 
-		restConfig, err := rest.InClusterConfig()
-		if err != nil {
-			return nil, err
-		}
-		client, err := kubernetes.NewForConfig(restConfig)
+		client, err := kubernetes.NewForConfig(utils.GetConfig())
 		if err != nil {
```
JoshVanL, line 93: Non-functional change to enable Kubernetes validator integration tests.

pkg/sentry/server/validator/kubernetes/kubernetes.go
```diff
@@ -192,10 +187,18 @@ func (k *kubernetes) Validate(ctx context.Context, req *sentryv1pb.SignCertifica
...
 		req.Id = expID
 	}
 
+	// The TrustDomain field is ignored by the Kubernetes validator. We must
+	// validate the request _after_ performing the token review so that in the
+	// event the client is uing the "legacy" <ns>:<sa> ID, we can override it
+	// with the expected app ID.
+	if _, _, err = internal.Validate(ctx, req); err != nil {
+		return spiffeid.TrustDomain{}, false, err
+	}
```
JoshVanL, line 200: The fix

pkg/sentry/server/validator/kubernetes/kubernetes.go
```diff
@@ -192,10 +187,18 @@ func (k *kubernetes) Validate(ctx context.Context, req *sentryv1pb.SignCertifica
 	// for the ID containing their namespace and service account (ns:sa). This
 	// is wrong- dapr identities are based on daprd namespace + _app ID_.
 	// Remove this allowance in v1.13.
-	if req.Namespace+":"+pod.Spec.ServiceAccountName == req.Id {
+	if pod.Namespace+":"+pod.Spec.ServiceAccountName == req.Id {
```
daixiang0, line 190: Better to use fmt.Sprintf

tests/integration/suite/sentry/validator/insecure/insecure.go
```diff
@@ -169,3 +171,45 @@ func (m *insecure) Run(t *testing.T, parentCtx context.Context) {
...
+
+func validateCertificateResponse(t *testing.T, res *sentrypbv1.SignCertificateResponse, sentryBundle ca.Bundle, expectSPIFFEID, expectDNSName string) {
+	t.Helper()
+
+	require.NotEmpty(t, res.WorkloadCertificate)
+
+	rest := res.WorkloadCertificate
+
+	// First block should contain the issued workload certificate
+	{
```
daixiang0, line 183: No need block, the comment is enough.

pkg/sentry/server/validator/kubernetes/kubernetes.go
JoshVanL: We don't want to use `fmt.Sprintf` here as this is a very hot path for Dapr and bare string concatenation `+` is more performant.

tests/integration/suite/sentry/validator/insecure/insecure.go
JoshVanL: Done

yaron2 (approved):
LGTM

PR #7083, opened 2023-10-23
Change operator default port to 443.

JoshVanL:
The operator service should serve on 443 as this is the more proper TLS port over the current 80. The injector has been updated to now inject port 443 to sidecars. Port 80 is still being served for backwards compatibility, but should be removed in v1.14.


charts/dapr/dapr-0.0.0.tgz
daixiang0, line 1: Please remove it, if it is not in gitignore, update as well.

JoshVanL: Thanks, fixed!

mukundansundar (approved):

mukundansundar:
> The operator service should serve on 443 as this is the more proper TLS port over the current 80. The injector has been updated to now inject port 443 to sidecars. Port 80 is still being served for backwards compatibility, but should be removed in v1.14.

It should ideally be only removed in 1.15, 2 releases after deprecation.

JoshVanL:
@mukundansundar It was my understanding that this was not the case for non-API related control plane configuration options. For example we do not do this for removing CLI flags.

daixiang0 (approved):
LGTM

PR #7084, opened 2023-10-23
Remove the legacy port 80 from the sentry service

JoshVanL:
Before v1.12, Daprd's would connect to the sentry service on port 80. This was changed to 443 in v1.12, but we kept the port 80 open for backwards compatibility. This PR removes the port 80 from the sentry service.

Part of #5756


mukundansundar (changes requested):
@JoshVanL ideally we have a deprecation policy of 2 releases .... I think this should not be removed in this release.

JoshVanL:
@mukundansundar What is the reason for doing it over 2 releases?

yaron2 (approved):

mukundansundar:
> @mukundansundar What is the reason for doing it over 2 releases?

This is the dapr deprecation policy.

mukundansundar:
> @mukundansundar What is the reason for doing it over 2 releases?

This is the dapr deprecation policy.

JoshVanL:
> > @mukundansundar What is the reason for doing it over 2 releases?
> 
> This is the dapr deprecation policy.

This applies to features/APIs, however this is a control plane internal option which doesn't come with the same requirements w.r.t backward compat.

ItalyPaleAle (approved):

PR #7086, opened 2023-10-23
Split up runtime resource handler into processor manager

JoshVanL:
PR splits out the runtime authorizer and resource processor into a dedicated authorizer package, and a manager in the processor package. This de-couples the main runtime package from the processor so that hot reloading (not implemented here) can also make use of the processor manager.

Components are no longer indexed in the ComponentStore by both their name _and_ type. This is because the hot reloader would not be able to determine "which" Component it should delete on delete events. No two Components with the same type but different names can co-exist anyway.

Removes the existing hot reloading functionality as it is being replaced with the more robust implementation in a follow-up PR.

Part of #1172 
Spun out of https://github.com/dapr/dapr/pull/6845 to have separation of concerns and make reviewing easier.

### DIFF
--- a/pkg/runtime/authorizer/authorizer.go
+++ b/pkg/runtime/authorizer/authorizer.go
@@ -1,0 +1,113 @@
+/*
+Copyright 2023 The Dapr Authors
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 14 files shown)


artursouza:
integration tests seem to be flakier with this implementation.

pkg/runtime/authorizer/authorizer.go
```diff
@@ -0,0 +1,113 @@
+/*
```
yaron2, line 1: Can unit tests be added to test the functionality here?

JoshVanL: Done :+1:

yaron2 (approved):

Issue #7089, opened 2023-10-24
[Workflow] Make the Dapr workflow engine configurable


cgillum:
## In what area(s)?

> /area runtime

## Describe the feature

There's currently no way to configure the workflow engine. Some values, such as maximum concurrency thresholds, are hardcoded. We should make it possible to configure the workflow engine in config.yml.

## Release Note

RELEASE NOTE: **ADD** Ability to configure Dapr workflow engine in config.yml.

cgillum:
This was completed in https://github.com/dapr/dapr/pull/7090

PR #7090, opened 2023-10-24
[Workflow] Make workflow engine configurable (and other improvements)

cgillum:
# Description

Until now, there was no way to configure the workflow engine. Some values, such as maximum concurrency thresholds, had been hardcoded. With this PR, however, it's now possible to configure the workflow engine. In particular, this PR enables configuring the maximum workflow and activity execution concurrency (defaults to 100, as it did previously).

This PR also contains other misc. fixes and code refactoring. Specifically:

- Reorganized some of the workflow and activity configuration to make it more cleanly separated from workflow engine configuration.
- Added some additional debug logging based on issues I had while debugging my changes.
- Updated workflow test code to output debug logs (previously was only outputting info logs)

I've added comments to the PR directly to help explain some of the specific changes.

## Issue reference

https://github.com/dapr/dapr/issues/7089

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Unit tests passing
* [x] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: https://github.com/dapr/docs/pull/3844
* [ ] ~Specification has been updated / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]~
* [ ] ~Provided sample for the feature / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]~


pkg/runtime/wfengine/activity.go
```diff
@@ -130,8 +138,6 @@ func (a *activityActor) InvokeReminder(ctx context.Context, actorID string, remi
 		}
 	}
 
-	// TODO: Purge actor state based on some data retention policy
```
cgillum, line 133: Removed this TODO since this work was already done in v1.11.

pkg/runtime/wfengine/backend.go
```diff
@@ -25,52 +25,99 @@ import (
...
 	invokev1 "github.com/dapr/dapr/pkg/messaging/v1"
+	"github.com/dapr/dapr/utils"
 )
 
-// workflowScheduler is an interface for pushing work items into the backend
-type workflowScheduler interface {
-	ScheduleWorkflow(ctx context.Context, wi *backend.OrchestrationWorkItem) error
-	ScheduleActivity(ctx context.Context, wi *backend.ActivityWorkItem) error
+// actorsBackendConfig is the configuration for the workflow engine's actors backend
+type actorsBackendConfig struct {
```
cgillum, line 32: This struct and the functions that follow were moved from wfengine.go to here since they are specific to the workflow backend. No changes were made to these.

pkg/runtime/wfengine/backend.go
```diff
@@ -25,52 +25,99 @@ import (
 
 	"github.com/dapr/dapr/pkg/actors"
 	invokev1 "github.com/dapr/dapr/pkg/messaging/v1"
+	"github.com/dapr/dapr/utils"
 )
 
-// workflowScheduler is an interface for pushing work items into the backend
-type workflowScheduler interface {
-	ScheduleWorkflow(ctx context.Context, wi *backend.OrchestrationWorkItem) error
-	ScheduleActivity(ctx context.Context, wi *backend.ActivityWorkItem) error
```
cgillum, line 33: I replaced the `workflowScheduler` interface with two separate func definitions: `workflowScheduler` and `activityScheduler`, defined in workflow.go and activity.go respectively. This change made it easier to refactor some of the configuration. No changes were made to the required method signatures.

pkg/runtime/wfengine/backend.go
```diff
@@ -210,12 +257,18 @@ func (*actorBackend) DeleteTaskHub(context.Context) error {
 
 // GetActivityWorkItem implements backend.Backend
 func (be *actorBackend) GetActivityWorkItem(ctx context.Context) (*backend.ActivityWorkItem, error) {
-	// Wait for the workflow actor to signal us with some work to do
+	// Wait for the activity actor to signal us with some work to do
+	wfLogger.Debug("Actor backend is waiting for an activity actor to schedule an invocation.")
```
cgillum, line 261: These debug logs help make it easier to understand what's going on if a workflow or activity doesn't run when we expect it to (a problem that we've often seen but lacked clarity on what's actually stuck).

pkg/runtime/wfengine/wfengine.go
```diff
@@ -217,19 +187,27 @@ func (wfe *WorkflowEngine) Start(ctx context.Context) (err error) {
...
-	parallelismOpts := backend.WithMaxParallelism(100)
-
-	orchestrationWorker := backend.NewOrchestrationWorker(wfe.backend, wfe.executor, wfBackendLogger, parallelismOpts)
-	activityWorker := backend.NewActivityTaskWorker(wfe.backend, wfe.executor, wfBackendLogger, parallelismOpts)
+	// There are separate "workers" for executing orchestrations (workflows) and activities
+	orchestrationWorker := backend.NewOrchestrationWorker(
+		wfe.backend,
+		wfe.executor,
+		wfBackendLogger,
+		backend.WithMaxParallelism(wfe.spec.MaxConcurrentWorkflows))
```
cgillum, line 202: This is where we actually reference the new max concurrency configuration settings. Previously these values were hardcoded to 100 (see the deleted lines above here).

cgillum (commented):
Adding explanatory comments

yaron2:
Adding configuration options are useful only if developers know when they should tweak them. For workflow concurrency, what is the guidelines for developers? When should they tweak the number below and above 100, and what are the tradeoffs of tweaking that number?

pkg/config/configuration.go
```diff
@@ -112,6 +112,17 @@ type ConfigurationSpec struct {
 	ComponentsSpec      *ComponentsSpec     `json:"components,omitempty" yaml:"components,omitempty"`
 	LoggingSpec         *LoggingSpec        `json:"logging,omitempty" yaml:"logging,omitempty"`
 	WasmSpec            *WasmSpec           `json:"wasm,omitempty" yaml:"wasm,omitempty"`
+	WorkflowSpec        *WorkflowSpec       `json:"workflow,omitempty" yaml:"workflow,omitempty"`
+}
+
+// WorkflowSpec defines the configuration for Dapr workflows.
+type WorkflowSpec struct {
```
ItalyPaleAle, line 119: Please make the changes in `pkg/apis/configuration` too. You will then need to re-generate the CRDs with `make code-generate`

pkg/config/configuration.go
```diff
@@ -582,6 +593,18 @@ func (c Configuration) GetAPILoggingSpec() APILoggingSpec {
 	return *c.Spec.LoggingSpec.APILogging
 }
 
+// GetWorkflowSpec returns the Workflow spec.
+// It's a short-hand that includes nil-checks for safety.
+func (c *Configuration) GetWorkflowSpec() WorkflowSpec {
+	if c == nil || c.Spec.WorkflowSpec == nil {
+		return WorkflowSpec{
+			MaxConcurrentWorkflows:  100,
```
ItalyPaleAle, line 601: Default configuration should go in https://github.dev/cgillum/dapr/blob/60401b1cbd009ee20dfc61aa87bfa27f1f18b6a8/pkg/config/configuration.go#L362

pkg/runtime/wfengine/activity.go
```diff
@@ -117,6 +120,11 @@ func (a *activityActor) InvokeReminder(ctx context.Context, actorID string, remi
 		if errors.Is(err, context.DeadlineExceeded) {
```
ItalyPaleAle, line 120: Nit: please convert these `if..else if..else if..else if..else` to a switch for readability

pkg/runtime/wfengine/backend.go
```diff
@@ -266,7 +321,7 @@ func (*actorBackend) Stop(context.Context) error {
 
 // String displays the type information
 func (be *actorBackend) String() string {
-	return fmt.Sprintf("dapr.actors/v1-alpha")
+	return "dapr.actors/v1-beta"
```
ItalyPaleAle, line 324: Should this be backported into 1.12?

ItalyPaleAle (changes requested):

cgillum:
> Adding configuration options are useful only if developers know when they should tweak them. For workflow concurrency, what is the guidelines for developers? When should they tweak the number below and above 100, and what are the tradeoffs of tweaking that number?

@yaron2 Does the linked docs PR help answer this question? I can add more guidance if necessary but may need direction in terms of where the best place to document such guidance is.

pkg/runtime/wfengine/backend.go
cgillum: Ideally yes but I don’t think it’s important.

yaron2:
> > Adding configuration options are useful only if developers know when they should tweak them. For workflow concurrency, what is the guidelines for developers? When should they tweak the number below and above 100, and what are the tradeoffs of tweaking that number?
> 
> @yaron2 Does the linked docs PR help answer this question? I can add more guidance if necessary but may need direction in terms of where the best place to document such guidance is.

Additional context needs to be added, and I'll make comments on the docs PR based on the discussion here.

What can a user expect when they schedule a workflow or activity that passes the threshold? Is it rejected or queued until a workflow/activity complete?

cgillum:
> What can a user expect when they schedule a workflow or activity that passes the threshold? Is it rejected or queued until a workflow/activity complete?

A request that passes the threshold will be queued until a workflow/activity completes.

Note that this is primarily intended to be used as a safety mechanism to prevent runaway execution by workflows. For example, if a workflow were to schedule 1,000 concurrent CPU-intensive activities, this feature can be used to help ensure that we don't try to execute all of that work at the same time on any given machine.

cgillum:
> When should they tweak the number below and above 100

Below 100 when the developer/operator needs to run workflows in resource-limited environments, or if it is known in advance that activity/workflow execution can be very CPU/memory intensive.

Above 100 when that value is too restrictive given the amount of compute resources, limiting their overall workflow throughput.

> and what are the tradeoffs of tweaking that number?

Increasing the value allows for greater concurrency (throughput) at the cost of more resource consumption. Lowering the value leads to more conservative resource consumption at the cost of reduced throughput.

Note that we can debate whether 100 is the right number. I'm beginning to think setting the default higher (perhaps 1,000) might be reasonable. However, as of 1.12, the limit is hardcoded to 100.

pkg/runtime/wfengine/activity.go
cgillum: Novice Go question: how can I convert `else if _, ok := err.(recoverableError); ok` into a valid `case` statement?

ItalyPaleAle: Use errors.As: https://pkg.go.dev/errors#As

cgillum: Thanks - Done

pkg/config/configuration.go
```diff
@@ -582,6 +595,18 @@ func (c Configuration) GetAPILoggingSpec() APILoggingSpec {
 	return *c.Spec.LoggingSpec.APILogging
 }
 
+// GetWorkflowSpec returns the Workflow spec.
+// It's a short-hand that includes nil-checks for safety.
+func (c *Configuration) GetWorkflowSpec() WorkflowSpec {
+	if c == nil || c.Spec.WorkflowSpec == nil {
+		return WorkflowSpec{
```
ItalyPaleAle, line 602: I think default values should be handled by the consumer and not here.

The problem with this is that if you pass a configuration like:

```yaml
spec:
  workflow:
```

Or:

```yaml
spec:
  workflow:
    maxConcurrentWorkflowInvocations: 100
```

Or:

```yaml
spec:
  workflow:
    maxConcurrentActivityInvocations: 100
```

In all of the cases above, `c.Spec.WorkflowSpec` will not be nil, so _at least one of_ `MaxConcurrentWorkflowInvocations` and `MaxConcurrentActivityInvocations` will have a value of 0, because this `if` will not be invoked

ItalyPaleAle: If you do want to keep the defaults in this file, perhaps consider creating a getter for each one of those properties.

ItalyPaleAle (changes requested):

ItalyPaleAle (approved):

ItalyPaleAle (approved):

yaron2:
> Below 100 when the developer/operator needs to run workflows in resource-limited environments, or if it is known in advance that activity/workflow execution can be very CPU/memory intensive.

But what is resource-limited? It's very subjective. I'm extremely worried about using magic numbers with no proper guidance. The fact you wrote this could be made 1000 instead of 100 strengthens this.

> Increasing the value allows for greater concurrency (throughput) at the cost of more resource consumption. Lowering the value leads to more conservative resource consumption at the cost of reduced throughput.

This might be a proper way to describe this to users, so they know to observe and tweak if needed. Please carry out this language to the associated docs issue.

yaron2:
> A request that passes the threshold will be queued until a workflow/activity completes

When queued, will users get an ACK from the Dapr runtime immediately after queuing or will the client hang until its picked up?

cgillum:
> But what is resource-limited? It's very subjective. I'm extremely worried about using magic numbers with no proper guidance.

Can you clarify what specifically you're extremely worried about? People configuring it wrong in a particular way?

Backing up to talk a little bit more about the motivation for this work: In my experience, the most common resource-related issue that users run into is running out of memory. It's really easy to run out of memory because the workflow programming model makes it _really_ easy to schedule tons of work to run in parallel, which is obviously a double-edged sword. If those activities need to allocate a non-trivial amount of memory, then apps may start getting ugly OOM failures that create massive app stability issues. Next up would be overwhelming some downstream dependency, like a database, with too many connections because you have too many activities running concurrently. High CPU is another one, which should be self-explanatory. These are all app-specific issues which will apply to some workloads but not to others. The fastest and most effective way to mitigate most of these problems when they impact production is to make a config change to reduce concurrency. Dapr Workflow users today have no such capability since we currently hardcode these throttles - hence this PR.

> This might be a proper way to describe this to users, so they know to observe and tweak if needed. Please carry out this language to the associated docs issue.

Yes, this is absolutely an "observe and tweak if needed" configuration knob. I can make this clearer in the documentation.

> When queued, will users get an ACK from the Dapr runtime immediately after queuing or will the client hang until its picked up?

Regardless of whether you're above or below the threshold, the client always gets an immediate ACK. Clients are never blocked on workflow operations. Similarly, workflows are never blocked on scheduling activity calls. These existing throttles are expected only to add latency between the time you schedule some work and the time when that work starts executing.

pkg/config/configuration.go
```diff
@@ -60,6 +60,9 @@ const (
 	DefaultNamespace    = "default"
 	ActionPolicyApp     = "app"
 	ActionPolicyGlobal  = "global"
+
+	defaultMaxWorkflowConcurrentInvocations = 1000
+	defaultMaxActivityConcurrentInvocations = 1000
```
mukundansundar, line 65: have we tested this? Or should we for now set it back to 100? and later increase it once we have enough tests.

mukundansundar: @cgillum is there a reason for the increase in concurrent invocations?

cgillum: I'm fine with reverting it back and increasing it later.

artursouza (dismissed):

pkg/config/configuration.go
```diff
@@ -112,6 +115,33 @@ type ConfigurationSpec struct {
...
 	LoggingSpec         *LoggingSpec        `json:"logging,omitempty" yaml:"logging,omitempty"`
 	WasmSpec            *WasmSpec           `json:"wasm,omitempty" yaml:"wasm,omitempty"`
+	WorkflowSpec        *WorkflowSpec       `json:"workflow,omitempty" yaml:"workflow,omitempty"`
+}
+
+// WorkflowSpec defines the configuration for Dapr workflows.
+type WorkflowSpec struct {
+	// maxConcurrentWorkflowInvocations is the maximum number of concurrent workflow invocations that can be scheduled by a single Dapr instance.
+	// Attempted invocations beyond this will be queued until the number of concurrent invocations drops below this value.
+	// If omitted, the default value of 1000 will be used.
```
DeepanshuA, line 125: Default value to be changed in doc

pkg/config/configuration.go
```diff
@@ -112,6 +115,33 @@ type ConfigurationSpec struct {
...
+	MaxConcurrentWorkflowInvocations int32 `json:"maxConcurrentWorkflowInvocations,omitempty" yaml:"maxConcurrentWorkflowInvocations,omitempty"`
+	// maxConcurrentActivityInvocations is the maximum number of concurrent activities that can be processed by a single Dapr instance.
+	// Attempted invocations beyond this will be queued until the number of concurrent invocations drops below this value.
+	// If omitted, the default value of 1000 will be used.
```
DeepanshuA, line 129: Same

pkg/apis/configuration/v1alpha1/types.go
```diff
@@ -63,6 +63,22 @@ type ConfigurationSpec struct {
...
 	WasmSpec *WasmSpec `json:"wasm,omitempty"`
+	// +optional
+	WorkflowSpec *WorkflowSpec `json:"workflow,omitempty"`
+}
+
+// WorkflowSpec defines the configuration for Dapr workflows.
+type WorkflowSpec struct {
+	// maxConcurrentWorkflowInvocations is the maximum number of concurrent workflow invocations that can be scheduled by a single Dapr instance.
+	// Attempted invocations beyond this will be queued until the number of concurrent invocations drops below this value.
+	// If omitted, the default value of 1000 will be used.
```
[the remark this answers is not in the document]
cgillum: Thanks - looks like I need to work on my find/replace skills 😅

pkg/runtime/wfengine/workflow.go
```diff
@@ -521,6 +526,10 @@ func (wf *workflowActor) loadInternalState(ctx context.Context, actorID string)
 		// No such state exists in the state store
 		return nil, nil
 	}
+	if !wf.cachingDisabled {
+		// update cached state
+		wf.states.Store(actorID, state)
+	}
```
mukundansundar, line 532: do we need the same check at the start of the function, at line 514?

mukundansundar: Also in `saveInternalState` the cache is being updated even if the TransactionStateOperation fails lines 539 and 549?

cgillum: Not strictly necessary for 514 since the cache should be empty.

For 539, 549, yes this needs to be fixed. That issue is why some Cosmos DB compatibility problems weren’t more quickly detected. Could do it now or in another PR.

ItalyPaleAle (approved):

mukundansundar (approved):

cgillum:
Tests are looking good. Are we good to merge? @mukundansundar @artursouza @yaron2 

I see that there's an `automerge` label added, but I'm not sure what the criteria for that is.

PR #7096, opened 2023-10-25
Tiny tweaks in `utils`

ItalyPaleAle:
1. Moved Kubernetes functions to a more specific `kubernetes.go` file (without code changes)
2. Small perf improvements for `Filter`, about 8% faster (305.5ns/op -> 281.5ns/op)
3. Even smaller perf improvement for IsTruthy (~2%; bigger improvement on larger strings)


utils/utils.go
```diff
@@ -114,7 +68,11 @@ func GetEnvOrElse(name, orElse string) string {
 // IsTruthy returns true if a string is a truthy value.
 // Truthy values are "y", "yes", "true", "t", "on", "1" (case-insensitive); everything else is false.
 func IsTruthy(val string) bool {
-	switch strings.ToLower(strings.TrimSpace(val)) {
+	val = strings.TrimSpace(val)
+	if len(val) > 4 {
+		val = val[:4]
+	}
```
mukundansundar, line 74: doesn't this open up the possibility of accepting strings which are more than 4 characters but contain random trailing characters? eg: `true1234` would still pass since it is truncated at 4 characters.

ItalyPaleAle: That's really true. I'll change to 5 and that will not be ambiguous.

utils/utils.go
```diff
@@ -114,7 +68,11 @@ func GetEnvOrElse(name, orElse string) string {
 // IsTruthy returns true if a string is a truthy value.
 // Truthy values are "y", "yes", "true", "t", "on", "1" (case-insensitive); everything else is false.
 func IsTruthy(val string) bool {
-	switch strings.ToLower(strings.TrimSpace(val)) {
+	val = strings.TrimSpace(val)
+	if len(val) > 4 {
+		val = val[:4]
```
ItalyPaleAle, line 73:
```suggestion
	if len(val) > 5 {
		val = val[:5]
```

utils/utils.go
```diff
@@ -114,7 +68,11 @@ func GetEnvOrElse(name, orElse string) string {
 // IsTruthy returns true if a string is a truthy value.
 // Truthy values are "y", "yes", "true", "t", "on", "1" (case-insensitive); everything else is false.
 func IsTruthy(val string) bool {
-	switch strings.ToLower(strings.TrimSpace(val)) {
+	val = strings.TrimSpace(val)
+	if len(val) > 5 {
+		val = val[:5]
```
ItalyPaleAle, line 73:
```suggestion
	if len(val) > 4 {
		// Short-circuit, this can never be a truthy value
		return false
```

mukundansundar (approved):

PR #7100, opened 2023-10-26
Integration: Adds `-focus` CLI flag

JoshVanL:
Adds a `-focus` CLI flag to the integration test runner. This flag allows the user to specify a regex to filter the tests that are run.

This is to allow easier focusing on a specific test when debugging or developing a new test.

Even if `-focus` is enabled, binaries will always be built.

``` bash
go test -v -race -tags integration ./tests/integration -focus sentry
go test -v -race -tags integration ./tests/integration -test.skip Test_Integration/sentry/validator/jwks -focus sentry
```


tests/integration/integration.go
```diff
@@ -33,20 +35,30 @@ import (
 	_ "github.com/dapr/dapr/tests/integration/suite/sentry"
 )
 
+var focusF = flag.String("focus", ".*", "Focus on specific test cases. Accepts regex.")
```
daixiang0, line 38: Seems `filter` would be better.

JoshVanL: I like focus because it is consistent with [GinkGo](https://onsi.github.io/ginkgo/#focused-specs) which is a popular Go test library/program.

tests/docs/writing-integration-test.md
```diff
@@ -5,14 +5,21 @@
 Integration tests are run against a live running daprd binary locally. Each test
 scenario is run against a new instance of daprd, where each scenario modifies
 the daprd configuration to suit the test. Tests are expected to complete within
-seconds, ideally less than 5, and should not take longer than 30. Daprd is
+seconds, ideally less than 5, and should not take longer than 30. Binaries
```
ItalyPaleAle, line 8: Nit, grammar:

```suggestion
seconds, ideally less than 5, and should not take longer than 30. Binaries are
```

tests/integration/integration.go
```diff
@@ -33,20 +35,30 @@ import (
...
+
 func RunIntegrationTests(t *testing.T) {
-	var binFailed bool
+	flag.Parse()
+
+	focus, err := regexp.Compile(*focusF)
+	require.NoError(t, err)
```
ItalyPaleAle, line 44: Nit:

```suggestion
	require.NoError(t, err, "Invalid parameter focus")
```

tests/docs/writing-integration-test.md
```diff
@@ -5,14 +5,21 @@
...
 ```bash
-go test -v -race -count -tags="integration" ./tests/integration` -run="Test_Integration/daprd/pubsub/http/fuzzpubsubNoRaw"
+go test -v -race -tags integration ./tests/integration
+```
+
+You can also run a subset of tests by specifying the `-focus` flag, which takes a [Go regular expression](https://github.com/google/re2/wiki/Syntax).
+
+```bash
+go test -v -race -tags integration ./tests/integration -focus sentry
+go test -v -race -tags integration ./tests/integration -test.skip Test_Integration/sentry/validator/jwks -focus sentry
```
ItalyPaleAle, line 22: I am confused by this example, as there's a `-test.skip` and `-focus`

How should focus be used?

JoshVanL: I have updated this page with comments. The skip example here means to run all sentry related tests, but skip this specific jwks validator tes .

tests/docs/writing-integration-test.md
```diff
@@ -5,14 +5,21 @@
...
 
 
 ## Invoking the test
 
 ```bash
-go test -v -race -count -tags="integration" ./tests/integration` -run="Test_Integration/daprd/pubsub/http/fuzzpubsubNoRaw"
+go test -v -race -tags integration ./tests/integration
+```
+
+You can also run a subset of tests by specifying the `-focus` flag, which takes a [Go regular expression](https://github.com/google/re2/wiki/Syntax).
```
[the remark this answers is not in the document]
JoshVanL: Test names are comprised of the suite package tree + the struct name which was registered for that suite case. It should be quite natural for a developer to use the `focus` on a specific area they are interested in, for example `-focus secret` will run all secret related tests, `-focus actor` for actor related tests etc. Focus would pick up on tests with `actor` in the name/path so doesn't limit to a specific package.

When a developer is working on a specific feature or integration test, they will likely know the specific path they want to use, e.g. `operator/api/authz`.

ItalyPaleAle (approved):

mukundansundar (approved):

PR #7104, opened 2023-10-28
Integration: Adds healthz and port tests for operator

JoshVanL:
PR adds the basic healthz and port integration tests for the operator binary. PR updates the operator so that the API, healthz, and metrics port can be configured via CLI flags. Expands the integration test kubernetes process to implement more resource handlers.

The Kubernetes process now serves the CRDs as defined in the helm chart at the Kubernetes CRD endpoint.

The operator still serves the same ports by default and in the helm chart.

Moves the integration top level test import to dedicated `import.go` file.


pkg/operator/api/api.go
```diff
@@ -73,27 +79,34 @@ type apiServer struct {
...
+	sec, err := a.sec.Handler(ctx)
+	if err != nil {
+		return err
+	}
 
 	s := grpc.NewServer(sec.GRPCServerOptionMTLS())
 	operatorv1pb.RegisterOperatorServer(s, a)
 
-	lis, err := net.Listen("tcp", fmt.Sprintf(":%v", serverPort))
+	lis, err := net.Listen("tcp", fmt.Sprintf(":%v", a.port))
```
mukundansundar, line 109: can we avoid sprintf in hot path here?

pkg/operator/api/api.go
```diff
@@ -117,7 +130,7 @@ func (a *apiServer) Run(ctx context.Context, sec security.Handler) error {
 		return err
 	}
 	err = lis.Close()
-	if err != nil {
+	if err != nil && !errors.Is(err, net.ErrClosed) {
```
mukundansundar, line 133: are there any other errors being swallowed here with this new check?

JoshVanL: Not sure I follow, we are only ignoring the case when the listener errors because it has been closed.

pkg/operator/api/api.go
JoshVanL: Done.

yaron2 (approved):

PR #7111, opened 2023-10-30
Operator: Adds authorization to the operator API

JoshVanL:
Today, the operator does no authorization on clients making requests for
resources. This means that it is currently possible for any
authenticated client to get any Component/HTTPEndpoint/Resiliency in any
namespace. This includes the ability for any client is able to get
Dapr secrets in any namespace.

PR updates the operator to add authorization to the operator API. The
API server will ensure that the client's identity (SPIFFE ID) resides in
the same namespace as the requested resource namespace. This is done by
parsing the clients SPIFFE SVID of the mTLS connection.

If the client does not reside in the same namespace, then the request is
rejected with a permission denied.

---

The health server has been updated so that a target number can be given
where the health server will only return healthy once the consumer has
given the target number of healthy calls. This is useful for the
operator so that the operator only reports healthy once all informers
and servers have started. Without this change, clients would attempt to
connect to the API server when it wasn't actually ready to serve
requests. This was surfaced in the integration tests.

The PR adds an authz operator integration test which proves that the API
server will only honour requests with clients in the appropriate
namespace.

The Kubernetes intergration process now has the ability to serve
HTTP/2.0 informer requests from the API server controller runtime
informers- added so that informer reliant functions (`ComponentUpdate`,
`HTTPEndpointUpdate`, etc.) can be handled.

Updates `tests/util/pki` so that it can respond with a context which
includes the SPIFFE ID of the client, used for operator authz unit
tests.

  Signed-off-by: joshvanl <me@joshvanl.dev>

---


yaron2:
LGTM but need to make sure this is backward compatible, ie. 1.12 sidecars connecting to edge/1.13 control plane.

yaron2 (approved):

PR #7115, opened 2023-10-30
Change injector and sentry to GET daprsystem Configuration

### LINKED ISSUE #7114: Vulnerability in Global Dapr Configuration for Sentry & Injector
In Kubernetes mode, when the Injector is patching a Pod to determine whether mTLS is enabled[0] and Sentry on startup[1] fetches the global daprsystem Configuration, they do so by listing all Configurations in all namespaces and then match on the first Configuration with the name daprsystem. As Namespaces are sorted alphabetically when listed, the Configuration chosen by these services may not be the one located in the Dapr System namespace. This means that a malicious actor, or by accident a user of a Dapr enabled Kubernetes cluster, with write permissions to Configurations in a namespace which is alphabetically higher than the Dapr system namespace is able to override the global config for Sentry and Injector. We can expect that users of Dapr in Kubernertes would be able to have permissions to Configurations in order for them to control their Dapr deployment configuration.

The impact of this vulnerability is that an actor would be able to do things such as disable mTLS Dapr cluster wide; re-write the sentry address meaning no new Dapr instances can join the cluster or existing Daprds renew their certificate bringing the Dapr cluster down; or change the sentry auth so that they can request for whatever identity they want. These are some examples of the impact, but there are likely others. This vulnerability impacts v1.12, v1.11 and v1.10, but impacts versions before that.

To reproduce this issue, create a Namespace called "a", create a Dapr Configuration with some values you want, and finally install Dapr. Observe those values being honored by the Injector and Sentry. You can install Dapr first, but Sentry needs to be restarted as it pulls the Configuration on startup. For example[2].

To mitigate this vulnerability, we need to update Injector and Sentry to only get the daprsystem Configuration in the Dapr System namespace. I think we should apply the fix and release patch versions for v1.12, v1.11, and v1.10.

- 0: https://github.com/dapr/dapr/blob/e09bd301ad54aa2e6e6ddb1a837f74577b68fbec/pkg/injector/service/pod_patch.go#L108
- 1: https://github.com/dapr/dapr/blob/e09bd301ad54aa2e6e6ddb1a837f74577b68fbec/pkg/sentry/config/config.go#L109
- 2:

```
apiVersion: v1
kind: Namespace
metadata:
  name: a
---
apiVersion: [dapr.io/v1alpha1](http://dapr.io/v1alpha1)
kind: Configuration
metadata:
  name: daprsystem
  namespace: a
spec:
  metric:
    enabled: true
  metrics:
    enabled: true
  mtls:
    allowedClockSkew: 0m
    controlPlaneTrustDomain: cluster.local
    enabled: false
    sentryAddress: bad-address:1234
    workloadCertTTL: 1ms
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: sleep
  namespace: default
spec:
  replicas: 1
  selector:
    matchLabels:
      testapp: sleep
  template:
    metadata:
      labels:
        testapp: sleep
      annotations:
        dapr.io/enabled: "true"
        dapr.io/app-id: "sleep"
        dapr.io/log-level: "debug"
    spec:
      containers:
      - name: sleep
        image: alpine
        command: ["/bin/sh"]
        args: ["-c", "sleep 1000"]
```

JoshVanL:
In Kubernetes mode, when the Injector is patching a Pod to determine whether mTLS is enabled and Sentry on startup] fetches the global daprsystem Configuration, they do so by listing all Configurations in all namespaces and then match on the first Configuration with the name daprsystem. As Namespaces are sorted alphabetically when listed, the Configuration chosen by these services may not be the one located in the Dapr System namespace. This means that a malicious actor, or by accident a user of a Dapr enabled Kubernetes cluster, with write permissions to Configurations in a namespace which is alphabetically higher than the Dapr system namespace is able to override the global config for Sentry and Injector. We can expect that users of Dapr in Kubernertes would be able to have permissions to Configurations in order for them to control their Dapr deployment configuration.

PR updates the injector and sentry services to GET the daprsystem Configuration in the Dapr control plane namespace.

Fixes #7114


artursouza (dismissed):

pkg/security/security.go
```diff
@@ -402,6 +402,16 @@ func CurrentNamespace() string {
 	return namespace
 }
 
+// CurrentNamespaceOrError returns the namespace of this workload. If current
+// Namespace is not found, error.
+func CurrentNamespaceOrError() (string, error) {
+	namespace, ok := os.LookupEnv("NAMESPACE")
+	if !ok {
```
ItalyPaleAle, line 409: What if NAMESPACE is set but empty?

JoshVanL: In this case, we would end up with an error downstream. As you point out, it is probably better to error in this case. I think a separate error message would be more proper.

ItalyPaleAle (approved):

yaron2 (approved):

mukundansundar:
@JoshVanL Do we need this PR, Can't we merge 1.12 back into master instead?

mukundansundar:
@JoshVanL can 1.12 branch be merged back into master for this?

JoshVanL:
@mukundansundar sure thing.

PR #7118, opened 2023-10-30
[1.10] Change injector and sentry to GET daprsystem Configuration

JoshVanL:
In Kubernetes mode, when the Injector is patching a Pod to determine whether mTLS is enabled and Sentry on startup] fetches the global daprsystem Configuration, they do so by listing all Configurations in all namespaces and then match on the first Configuration with the name daprsystem. As Namespaces are sorted alphabetically when listed, the Configuration chosen by these services may not be the one located in the Dapr System namespace. This means that a malicious actor, or by accident a user of a Dapr enabled Kubernetes cluster, with write permissions to Configurations in a namespace which is alphabetically higher than the Dapr system namespace is able to override the global config for Sentry and Injector. We can expect that users of Dapr in Kubernertes would be able to have permissions to Configurations in order for them to control their Dapr deployment configuration.

PR updates the injector and sentry services to GET the daprsystem Configuration in the Dapr control plane namespace.

See: https://github.com/dapr/dapr/issues/7114

### DIFF
--- a/cmd/injector/main.go
+++ b/cmd/injector/main.go
@@ -62,7 +62,10 @@ func main() {
 		log.Fatalf("failed to get authentication uids from services accounts: %s", err)
 	}
 
-	inj := injector.NewInjector(uids, cfg, daprClient, kubeClient)
+	inj, err := injector.NewInjector(uids, cfg, daprClient, kubeClient)
+	if err != nil {
+		log.Fatal(err)
+	}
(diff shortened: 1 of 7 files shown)


artursouza (approved):

artursouza:
Build fails.

artursouza (approved):

Issue #6121, opened 2023-03-22
Incremental improvements to Actor Reminders subsystem


ItalyPaleAle:
## Describe the proposal

This proposal contains a number of possible incremental improvements aimed at increasing the performance and scalability of the Actor Reminders subsystem. 

We understand that the current architecture of the Actor Reminders subsystem has some limitations that will not allow us to get the performance and/or scalability that we desire. Addressing that is a longter-term issue that will involve implementing a "v2" of the subssystem, which we're currently working on a design for (_not yet ready as of writing this!_).

In parallel, there are some ways to provide incremental improvements to the subsystem in its "v1" form, to ease (a bit) the pain of users that make extensive use of actor reminders and workflows.

> I welcome any other suggestion on how we can make other improvements

## Prerequisite

All of the improvements mentioned below depend on #6039 being merged, which (besides fixing some bugs with race conditions and cleaning up some code), introduces some data structures needed for the points below.

### Use a priority queue for keeping scheduled reminders (and timers) in-memory

**Status:** WIP, with initial code in #6040 

> This applies to timers too

Currently, when reminders are loaded, a new goroutine is started for each reminder that blocks until the reminder's due time hits (or until it's canceled). This means that there's a goroutine for each reminder (and for each timer), which put significant pressure on memory and cause more work for the GC.

This can be improved by changing the internal implementation to use a priority queue to maintain the list of reminders that are scheduled to be executed (for timers, there'll be a separate but identical queue). A heap-based priority queue is efficient as it allows for peeking with O(1) complexity, and adding or removing an arbitrary item has complexity O(log N).

- Reminders in the queue are sorted in order of execution time, with the first reminder at the front of the queue.
- If there's at least one reminder in the queue, the first one is grabbed from the queue and a background goroutine is started. This goroutine blocks until it's time for the reminder to be executed.
    - When a reminder is executed, the same goroutine gets the next reminder and blocks until that is executed too. This means that there's at most one background goroutine that is blocked, regardless of the size of the reminders queue.
    - If the reminder at the top of the queue is deleted or modified, the waiting goroutine is stopped.

## Changes that require a feature flag

These changes need to be behind a feature flag because, while they're backwards-compatible, they're not forward-compatible. That is: the next Dapr version will be able to read data created by the current one (backwards-compatibility), but the current Dapr can't read data created by the new one (lack of forward-compatibility). This causes issues with clusters with mixed versions, so needs to be handled smoothly.

[...]

PR #7129, opened 2023-10-31
Serialize reminders as protobuf in state store

ItalyPaleAle:
Part of #6121 

Changes how reminders are serialized in the state store, switching from the current JSON serialization to protobuf. Using protobuf is more efficient, not just because the serialization itself is faster, but because the payload is more "compressed" and so the I/O cost (the biggest issue today) is reduced.

Results from the actor_reminder perf test (using Postgres) (higher is better):

- [JSON](https://github.com/dapr/dapr/actions/runs/6708202961): 37.80 QPS
- [Protobuf](https://github.com/dapr/dapr/actions/runs/6710588956): 58.96 QPS (+56%)

Workflow perf test (TestSeriesWorkflowWithMaxVUs - Req Duration P90) (lower is better):

- [JSON](https://github.com/dapr/dapr/actions/runs/6708202961): 123.4s
- [Protobuf](https://github.com/dapr/dapr/actions/runs/6710609050): 54.5s (-55%)

This is the first change that leverages #6838 for real to preserve backwards-compatibility and to be able to function in a cluster with mixed Dapr versions

- Actor API level has been bumped to 20 (was 10 in 1.12)
- If the cluster API level is 20 or higher, reminders are serialized as protobuf. This means that if all actor hosts in the cluster are running on "vNext" (assuming 1.13), then reminders are serialized to protobuf.
- If the actors API level in the cluster is < 20 (either because there's at least 1 host running 1.12 or lower, or because the admin explicitly set the maximum API version to < 20 in the Helm chart), then reminders are serialized as JSON
- It's always possible to parse reminders that are serialized as JSON, regardless of the current API level

TODO:

- [ ] Figure out how to write a test for ensuring that this can read/write JSON depending on the API level in the cluster


dapr/proto/internals/v1/reminders.proto
```diff
@@ -0,0 +1,37 @@
...
+
+// Reminder represents a reminder that is stored in the Dapr actor state store.
+message Reminder {
+  string actor_id = 1;
+  string actor_type = 2;
+  string name = 3;
+  bytes data = 4;
+  string period = 5;
+  google.protobuf.Timestamp registered_time = 6;
+  string due_time = 7;
```
cgillum, line 30: Why `string` and not `google.protobuf.Timestamp` for `due_time`?

ItalyPaleAle: it's the value as sent by the user, and it could be a duration such as "2m". A bit of a legacy thing.

yaron2:
Can attest personally changing JSON to proto several years back has improved Redis throughput x10 on a project I worked on

artursouza:
I really like this change. It is a quick life improvement to users before we can offer a v2 experience for reminders.

mukundansundar (commented):
the change is something that will be extremely useful. Please add version skew tests for this feature.

ItalyPaleAle:
> the change is something that will be extremely useful. Please add version skew tests for this feature.

I'll see what I can do after I'm back next week.

The current setup for version skew tests doesn't allow testing for something like this. We may need some creative thinking.

ItalyPaleAle:
I have added integration tests that verify the behavior. In particular:

1. That daprd can read reminders serialized as protobuf or JSON (this works as "upgrade test")
2. That when the Actors API level in the cluster is >= 20, reminders are stored as protobuf
3. That when the Actors API level in the cluster is < 20, reminders are stored as JSON

@mukundansundar PTAL

JoshVanL:
@ItalyPaleAle @mukundansundar  
friendly reminder that actor/placement PRs are blocked until https://github.com/dapr/dapr/issues/7250 is resolved

ItalyPaleAle:
This PR actually includes a fix for the error in 7250. Not sure why it does appear only on MacOS. But i fixed it here

tests/integration/framework/process/daprd/options.go
```diff
@@ -136,6 +146,37 @@ func WithResourceFiles(files ...string) Option {
...
+    - name: connectionString
+      value: 'file:` + dbPath + `'
+    - name: actorStateStore
+      value: 'true'
+`)
+
+	// Connect to the database
+	conn, err := sql.Open("sqlite", "file://"+dbPath+"?_txlock=immediate&_pragma=busy_timeout(5000)&_pragma=journal_mode(WAL)")
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to open connection to SQLite database: %w", err)
```
JoshVanL, line 175: Don't return error, and use `t` instead.

tests/integration/suite/actors/reminders/serialization.go
```diff
@@ -0,0 +1,259 @@
...
+	"github.com/dapr/dapr/tests/integration/framework/process/daprd"
+	prochttp "github.com/dapr/dapr/tests/integration/framework/process/http"
+	"github.com/dapr/dapr/tests/integration/framework/process/placement"
+	"github.com/dapr/dapr/tests/integration/framework/util"
+	"github.com/dapr/dapr/tests/integration/suite"
+)
+
+func init() {
+	suite.Register(new(serializationProtobuf))
+	suite.Register(new(serializationJSON))
```
JoshVanL, line 43: Put these into separate files `reminders/serialisation/json.go` `reminders/serialisation/protobuf`

tests/integration/framework/process/daprd/options.go
```diff
@@ -136,6 +146,37 @@ func WithResourceFiles(files ...string) Option {
...
+  version: v1
+  metadata:
+    - name: connectionString
+      value: 'file:` + dbPath + `'
+    - name: actorStateStore
+      value: 'true'
+`)
+
+	// Connect to the database
+	conn, err := sql.Open("sqlite", "file://"+dbPath+"?_txlock=immediate&_pragma=busy_timeout(5000)&_pragma=journal_mode(WAL)")
```
JoshVanL, line 173: Connection should be closed when test is cleaned up.

Also consider moving SQLite to a separate framework process. We should only ever be retuning `*Options` from option funcs, so this suggests this setup isn't quite right.

tests/perf/actor_reminder/actor_reminder_test.go
```diff
@@ -39,7 +39,7 @@ const (
 	appName         = "perf-actor-reminder-service"
 
 	// Target for the QPS
-	targetQPS = 50
+	targetQPS = 56
```
JoshVanL, line 42: Why?

tests/integration/suite/actors/reminders/serialization.go
```diff
@@ -0,0 +1,259 @@
...
+	place   *placement.Placement
+	db      *sql.DB
+}
+
+func (i *serializationProtobuf) Setup(t *testing.T) []framework.Option {
+	// Init placement with no maximum API level
+	i.place = placement.New(t)
+
+	// Create a SQLite database
+	sqliteOpt, db, err := daprd.WithSQLiteStateStore(t)
```
JoshVanL, line 62: `WithSQLightActorStateStore` should be receiving an SQLite process.

tests/integration/suite/actors/reminders/serialization.go
ItalyPaleAle: A lot of the code is shared

ItalyPaleAle:
> @ItalyPaleAle it's not clear to me why this should be fixing the actor tests, can you explain further?

I can't link right now, on my phone. It's L73 in rebalancing.go that fixes it by setting a maximum version for actors api level.

I stumbled upon that by accident and fixed it in this PR before I saw your issue with the log indicating that failure.

It may appear just on Mac due to timings.

tests/integration/framework/process/daprd/options.go
ItalyPaleAle: No need to close, the connection is automatically closed when the object is deallocated. It's also a temporary file so doesn't matter anyways.

As for a separate process, SQLite is not a process that is spawned, so not sure I understand what you mean.

tests/perf/actor_reminder/actor_reminder_test.go
ItalyPaleAle: Because we can now :) the perf has improved a lot so setting a new baseline.

tests/integration/suite/actors/reminders/serialization.go
ItalyPaleAle: There's no process being spawned.

tests/integration/framework/process/daprd/options.go
JoshVanL: We want to make this a separate process so that it itself can be passed to other options. The SQLite process would then expose a `DBFile() string` method which can be used by receiving Option funcs.

I'd rather we did close the connection for completeness. We will then know for sure all resources are cleaned up on test completion.

JoshVanL: To be clear here- an Integration framework process doesn't refer to a Linux process (although it can), it instead refers to a long running test dependency which has a Setup, Run, and exposes runtime config which can be used by other processes or the test itself. In this case, the SQLite will expose the db file created, and/or a connection to that file.

ItalyPaleAle:
@JoshVanL i've updated the tests per your feedback

tests/integration/suite/actors/reminders/rebalancing.go
```diff
@@ -241,15 +229,15 @@ func (i *rebalancing) Run(t *testing.T, ctx context.Context) {
 	}
 }
 
-type httpServer struct {
+type rebalancingHTTPServer struct {
```
JoshVanL, line 232: We don't need this change because `rebalancing` is inferred through the test (package) name.

tests/integration/suite/actors/reminders/serialization_json.go
```diff
@@ -0,0 +1,105 @@
...
+
+	// Check the data in the SQLite database
+	// The value must begin with `[{`, which indicates it was serialized as JSON
+	storedVal := serializationLoadRemindersFromDB(t, ctx, i.db.GetConnection())
+	assert.Truef(t, strings.HasPrefix(storedVal, "[{"), "Prefix not found in value: '%v'", storedVal)
+
+	// Ensure the reminder was invoked at least once
+	assert.Eventually(t, func() bool {
+		return i.handler.remindersInvokeCount.Load() > 0
+	}, time.Second, 10*time.Millisecond, "Reminder was not invoked at least once")
```
JoshVanL, line 104: A second seems very low, are we worried about this?

JoshVanL (changes requested):
@ItalyPaleAle the actor integration tests still don't seem happy.

tests/integration/suite/actors/reminders/rebalancing.go
ItalyPaleAle: There is another HTTP server in the package.

tests/integration/framework/process/sqlite/sqlite.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+
+	for _, fopt := range fopts {
+		fopt(&s.opts)
+	}
+
+	// Create a SQLite database in the test's temporary directory
+	tmpDir := t.TempDir()
+	s.dbPath = filepath.Join(tmpDir, "test-data.db")
+	// t.Logf("Storing SQLite database at %s", s.dbPath)
```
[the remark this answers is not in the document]
ItalyPaleAle: It's very helpful for debugging the test.

tests/integration/framework/process/daprd/options.go
```diff
@@ -13,7 +13,13 @@ limitations under the License.
 
 package daprd
 
-import "github.com/dapr/dapr/tests/integration/framework/process/exec"
+import (
+
+	// Blank import for the sqlite driver
+	_ "modernc.org/sqlite"
+
```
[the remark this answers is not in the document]
ItalyPaleAle: Without this it won't work

tests/integration/suite/actors/reminders/serialization_json.go
ItalyPaleAle:
```suggestion
	}, 5*time.Second, 10*time.Millisecond, "Reminder was not invoked at least once")
```

tests/integration/framework/process/sqlite/sqlite.go
```diff
@@ -0,0 +1,147 @@
...
+
+	// Connect to the database
+	// We do this here because the test's init code may need to access the database before starting the test
+	if s.opts.connect {
+		var err error
+		s.conn, err = sql.Open("sqlite", "file://"+s.dbPath+"?_txlock=immediate&_pragma=busy_timeout(5000)&_pragma=journal_mode(WAL)")
+		require.NoError(t, err, "Failed to connect to SQLite database")
```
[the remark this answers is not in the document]
ItalyPaleAle: That's why there's `s.opts.connect`. It either needs the connection for the entire test, or not. It should be fine.

tests/integration/framework/process/sqlite/sqlite.go
```diff
@@ -0,0 +1,147 @@
...
+INSERT INTO metadata VALUES('migrations','1');
+CREATE TABLE state (
+  key TEXT NOT NULL PRIMARY KEY,
+  value TEXT NOT NULL,
+  is_binary BOOLEAN NOT NULL,
+  etag TEXT NOT NULL,
+  expiration_time TIMESTAMP DEFAULT NULL,
+  update_time TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP
+);`)
+	require.NoError(t, err)
```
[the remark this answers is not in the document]
ItalyPaleAle: It should only be used in some edge cases, such as when the database needs to be pre-populated with data.

tests/integration/framework/process/sqlite/sqlite.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+
+	for k, v := range s.opts.metadata {
+		c.Spec.Metadata = append(c.Spec.Metadata, commonapi.NameValuePair{
+			Name:  k,
+			Value: toDynamicValue(v),
+		})
+	}
+
+	enc, _ := json.Marshal(c)
```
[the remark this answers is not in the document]
ItalyPaleAle: This method doesn't accept a `t`. And this can't really fail because we only pass values that can be serialized as JSON.

tests/integration/framework/process/sqlite/sqlite.go
```diff
@@ -0,0 +1,147 @@
...
+	return string(enc)
+}
+
+func toDynamicValue(val string) commonapi.DynamicValue {
+	j, _ := json.Marshal(val)
```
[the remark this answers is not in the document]
ItalyPaleAle: The caller of this method doesn't have `t`. And strings never fail to be serialized to JSON

tests/integration/framework/process/sqlite/sqlite.go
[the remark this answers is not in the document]
JoshVanL: I prefer not to include these comment out debug lines as it increases noise in the code.

tests/integration/framework/process/sqlite/sqlite.go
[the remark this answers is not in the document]
JoshVanL: The `conn` here is a runtime test dependency and does not belong in the `New` function. Please look at other framework processes to see how `New` is used to instantiate options and configs. Runtime dependencies are created through `Run`. Ports are the only special case I can think of because of the nature of needing to both retrieve a reserved port from `0` and needing to squat that port till test execution. We don't have that requirement here.

tests/integration/framework/process/sqlite/sqlite.go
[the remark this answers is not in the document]
JoshVanL: The method can be updated to accept a `t *testing.T` and is proper to always check errors. The contents may be updated in future, error, and cause headaches for the programmer.

tests/integration/framework/process/sqlite/sqlite.go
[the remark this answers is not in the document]
JoshVanL: Same as above, please check all errors.

tests/integration/suite/actors/reminders/serialization_json.go
```diff
@@ -0,0 +1,105 @@
+/*
```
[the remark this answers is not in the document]
JoshVanL: This is not a resolved comment.

tests/integration/suite/actors/reminders/serialization_protobuf.go
```diff
@@ -0,0 +1,116 @@
+/*
```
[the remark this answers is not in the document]
JoshVanL: Same here.

tests/integration/suite/actors/reminders/rebalancing.go
JoshVanL: This will not be the case with `serialization` being made a subpackage.

tests/integration/framework/process/sqlite/sqlite.go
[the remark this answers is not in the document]
ItalyPaleAle: No, we can't do that in `Run` because it's executed in parallel with daprd starting. This must be done before daprd starts up

JoshVanL:
@ItalyPaleAle it shouldn't be the case that we need _more_ resources for the runners when things were working before. This suggests to me that resources are not being cleaned up between test cases, or a test case is completely de-funked.

tests/integration/suite/actors/reminders/serialization_json.go
ItalyPaleAle:
```suggestion
	}, 5*time.Second, 10*time.Millisecond, "Reminder was not invoked at least once")
```

tests/integration/framework/process/sqlite/sqlite.go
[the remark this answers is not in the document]
ItalyPaleAle: Our metadata only accepts strings so really this can never fail. Let's keep it simple :)

tests/integration/suite/actors/reminders/serialization_protobuf.go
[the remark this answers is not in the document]
ItalyPaleAle: Why does it need a sub-package? It's still reminders stuff

tests/integration/framework/process/sqlite/sqlite.go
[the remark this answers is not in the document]
ItalyPaleAle: I heavily relied on this log during development, to access the DB and see the data inside...

tests/integration/suite/actors/reminders/serialization_protobuf.go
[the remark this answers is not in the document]
JoshVanL: Yes, `serialization` is a sub topic of reminders- like `reminders` being a sub topic of `actors`. Package namespacing ensures better code organization and removes code smells like `serializationHTTPServer` & `rebalancingHTTPServer`.

tests/integration/framework/process/sqlite/sqlite.go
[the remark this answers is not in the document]
ItalyPaleAle: Ok

tests/integration/framework/process/sqlite/sqlite.go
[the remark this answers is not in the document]
ItalyPaleAle: Sorry I misspoke. Now that I've more time I went back and looked again at what the issue was.

We must create the tables before calling `Run`, because we must write data in the database before calling `Run`. Otherwise, we'd have to populate the database in the **test's** `Run` method, but at that point daprd is already running.

tests/integration/suite/actors/reminders/serialization/json.go
```diff
@@ -0,0 +1,105 @@
...
+	"github.com/dapr/dapr/tests/integration/framework/process/daprd"
+	prochttp "github.com/dapr/dapr/tests/integration/framework/process/http"
+	"github.com/dapr/dapr/tests/integration/framework/process/placement"
+	"github.com/dapr/dapr/tests/integration/framework/process/sqlite"
+	"github.com/dapr/dapr/tests/integration/framework/util"
+	"github.com/dapr/dapr/tests/integration/suite"
+)
+
+func init() {
+	suite.Register(new(serializationJSON))
```
JoshVanL, line 37: `serialization` is part of the package (and test name) so inferred. 

```suggestion
	suite.Register(new(json)
```

tests/integration/suite/actors/reminders/serialization/json.go
```diff
@@ -0,0 +1,105 @@
...
+}
+
+// serializationJSON tests:
+// - That reminders are serialized to JSON when the Actors API level in the cluster is < 20
+type serializationJSON struct {
```
JoshVanL, line 42:
```suggestion
type json struct {
```

tests/integration/suite/actors/reminders/serialization/json.go
```diff
@@ -0,0 +1,105 @@
...
+	"github.com/dapr/dapr/tests/integration/framework/process/sqlite"
+	"github.com/dapr/dapr/tests/integration/framework/util"
+	"github.com/dapr/dapr/tests/integration/suite"
+)
+
+func init() {
+	suite.Register(new(serializationJSON))
+}
+
+// serializationJSON tests:
```
JoshVanL, line 40:
```suggestion
// json tests:
```

tests/integration/suite/actors/reminders/serialization/json.go
```diff
@@ -0,0 +1,105 @@
...
+// - That reminders are serialized to JSON when the Actors API level in the cluster is < 20
+type serializationJSON struct {
+	daprd   *daprd.Daprd
+	srv     *prochttp.HTTP
+	handler *httpServer
+	place   *placement.Placement
+	db      *sqlite.SQLite
+}
+
+func (i *serializationJSON) Setup(t *testing.T) []framework.Option {
```
JoshVanL, line 50:
```suggestion
func (j *json) Setup(t *testing.T) []framework.Option {
```

tests/integration/suite/actors/reminders/serialization/protobuf.go
```diff
@@ -0,0 +1,115 @@
...
+	"github.com/dapr/dapr/tests/integration/framework/process/daprd"
+	prochttp "github.com/dapr/dapr/tests/integration/framework/process/http"
+	"github.com/dapr/dapr/tests/integration/framework/process/placement"
+	"github.com/dapr/dapr/tests/integration/framework/process/sqlite"
+	"github.com/dapr/dapr/tests/integration/framework/util"
+	"github.com/dapr/dapr/tests/integration/suite"
+)
+
+func init() {
+	suite.Register(new(serializationProtobuf))
```
JoshVanL, line 38: Also here.

```suggestion
	suite.Register(new(protobuf))
```

tests/integration/suite/actors/reminders/serialization/protobuf.go
```diff
@@ -0,0 +1,115 @@
...
+	// This will assert the ability to "upgrade" from JSON to Protobuf
+	queryCtx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+	defer cancel()
+	now := time.Now().UTC().Format(time.RFC3339)
+	_, err := i.db.GetConnection(t).ExecContext(queryCtx, fmt.Sprintf(`
+INSERT INTO state VALUES
+  ('actors||myactortype','[{"registeredTime":"%[1]s","period":"2m","actorID":"myactorid","actorType":"myactortype","name":"oldreminder","dueTime":"0"}]',0,'e467f810-4e93-45ed-85d9-e68d9fc7af4a',NULL,'%[1]s'),
+  ('actors||myactortype||metadata','{"id":"00000000-0000-0000-0000-000000000000","actorRemindersMetadata":{"partitionCount":0}}',0,'e82c5496-ae32-40a6-9578-6a7bd84ff331',NULL,'%[1]s');
+`, now))
+	require.NoError(t, err)
```
JoshVanL, line 81: This doesn't belong here- the code smell is the `context.Background()`.

This should be moved to a `WithRunExec` option which is executed on the process `Run`. This will then also have the proper test context a available.

tests/integration/framework/process/sqlite/sqlite.go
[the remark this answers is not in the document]
JoshVanL: Daprd is not running at this point.

tests/integration/framework/process/sqlite/sqlite.go
```diff
@@ -0,0 +1,156 @@
...
+	// Create a SQLite database in the test's temporary directory
+	tmpDir := t.TempDir()
+	s.dbPath = filepath.Join(tmpDir, "test-data.db")
+	t.Logf("Storing SQLite database at %s", s.dbPath)
+
+	return s
+}
+
+func (s *SQLite) Run(t *testing.T, ctx context.Context) {
+	// Nothing here
```
JoshVanL, line 70: This is where the SQLite Exec setup belongs.

JoshVanL (changes requested):

JoshVanL:
> > @ItalyPaleAle it shouldn't be the case that we need _more_ resources for the runners when things were working before. This suggests to me that resources are not being cleaned up between test cases, or a test case is completely de-funked.
> 
> I don't know what to say... Maybe more tests now are using more CPU and we're exhausting resources?

The error is because of the actor rebalancing test. You can see this here https://github.com/JoshVanL/dapr/actions/runs/7060286640/job/19219678365?pr=28

tests/integration/suite/actors/reminders/serialization/json.go
ItalyPaleAle: That would conflict with the json package

tests/integration/framework/process/sqlite/sqlite.go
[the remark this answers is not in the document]
ItalyPaleAle: When the test is Run-ing, daprd is not??

yaron2:
@ItalyPaleAle once you address existing comments I think we'll be good to merge, brief review LGTM.

tests/integration/suite/actors/reminders/serialization/json.go
ItalyPaleAle: that would conflict with the `json` package

tests/integration/suite/actors/reminders/serialization/protobuf.go
ItalyPaleAle: Because the other test is called `serializationJSON` (and can't be `json` as it would conflict with the serialization package), this should stay as `serializationProtobuf`

tests/integration/framework/process/sqlite/sqlite.go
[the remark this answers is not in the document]
ItalyPaleAle: I don't think there's anything I can do here as we need to run the queries before daprd starts.

Regardless of whether processes are started in sequence or parallel, the test's `Run` method is invoked after all processes have started, so we cannot execute queries to write data in that method.

Please confirm and I'll resolve the conversation.

ItalyPaleAle:
The second bug in the reminders rebalancing IT should be fixed by d25832b

tests/integration/suite/actors/reminders/serialization/json.go
JoshVanL: We can use an identifier for the `encoding/json` package such as `encodingjson encoding/json`. I think this is better than having stuttering in the test name.

tests/integration/suite/actors/reminders/serialization/protobuf.go
JoshVanL: Please see above.

tests/integration/framework/process/sqlite/sqlite.go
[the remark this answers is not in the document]
JoshVanL: Sorry, I don't know what you mean and think there is some confusion on the intention of a `suite.Case`, the framework `Process` and the order things are executed.

The test `Case` is the main test definition, whose struct name and package tree derive the Go test name. Its `Setup` function is always run first, and is where test processes (test dependencies) are defined. Next, the framework will execute the `Run` function of the test case process dependencies in order. This is where the process executes the actual dependency- this could be a binary process, in memory server, writing to a database w.e.
The test `Case` Run is then executed where the test invocation and checking code lives. Finally the processes are then `Cleanup`ed in reverse order to the `Run` order- it is important that processes ensure all resources are released here.

Below is the patch file for these changes showing how the integration test process is designed to be structured.

[sqlite.patch.txt](https://github.com/dapr/dapr/files/13599591/sqlite.patch.txt)

ItalyPaleAle:
Looks like the race condition is another bug in that test case (only in test code) that predates this PR. Fix doesn't seem immediate, but I can look into fixing that in a separate PR.

The remaining comments are about nits on test code, including naming, and can be fixed separately if needed. Looking forward getting this PR merged to fix the perf and integration tests failing in the master branch for many weeks :)

tests/integration/suite/actors/reminders/rebalancing.go
```diff
@@ -419,14 +406,19 @@ func (i *rebalancing) reportStatusToPlacement(ctx context.Context, stream placem
...
+				}
+				return
 			}
-			if o.GetOperation() == "update" {
+			if o.GetOperation() == "update" && !sent {
 				errCh <- nil
-				return
+				sent = true
 			}
 		}
```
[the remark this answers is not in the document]
JoshVanL: I do not get a race condition for this test on master.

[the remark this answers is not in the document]
ItalyPaleAle: The race condition is due to concurrent `recv` and `send` in the gRPC channel. Not sure why it doesn't appear on master (maybe because the background goroutine was terminated on the first message?) but that's not new code. Fixing that requires creating 2 separate streams and it's a moderately-sized code change. Can be fixed later

artursouza:
> Looks like the race condition is another bug in that test case (only in test code) that predates this PR. Fix doesn't seem immediate, but I can look into fixing that in a separate PR.
> 
> The remaining comments are about nits on test code, including naming, and can be fixed separately if needed. Looking forward getting this PR merged to fix the perf and integration tests failing in the master branch for many weeks :)

There might be 2 different issues here, so I don't want to dismiss the race condition in this PR or a bug in the test code.

On the other hand, we should raise our bar in pull requests because our users (and us) are paying a price for that. I suggest we move slower but get the PRs to a mergeable state and avoid "fix later" as much as possible. It is better to move slower in the right direction than fast in the wrong one.

yaron2 (approved):

yaron2:
Merging this as the race condition is scoped to test code and we're getting close to code freeze. We have a plan with @JoshVanL to follow up with a fix.

Issue #6683, opened 2023-07-17
Resiliency - Service Invocation allow to ignore resiliency policy based on response http status code


fabistb:
## In what area(s)?

/area runtime

## Describe the feature
In a business case it might not be required to always execute the resiliency policy if a http status code indicates a business error and it isn't expected that a retry will solve this.

In these scenarios it is often not required or desired to run the same operation multiple times.
However in case the http status code indicates a technical issue the resiliency is required.

Since the resiliency policies also apply for invoking non dapr endpoints this would provide additional benefits.

## Release Note

RELEASE NOTE: **ADD** Status filter for service invocation resiliency

iamzhaoxu:
I believe what the author means is if we apply the configuration below

```
apiVersion: dapr.io/v1alpha1
kind: Resiliency
metadata:
  name: slotting-pubsub-resiliency
  namespace: apps
scopes:
spec:
  policies:
    timeouts:
      general: 5s
      important: 60s
      largeResponse: 10s
    retries:
      linerRetry:
        policy: constant
        duration: 5s
        maxRetries: 3
        **httpCodes:
          - 500
          - 503**
       ```
The retry will only happen if the client app return response with 500 or 503. So we can have the ability define when to retry and when the message should be pushed to DLQ directedly.

fabistb:
@iamzhaoxu yes, that matches really well.

Hsuwen:
This is also what I am looking forward to.

EmealFox:
looking forward to this feature

fabistb:
keep alive

yaron2:
keepalive

fabistb:
keep alive

prashantsengar-cl:
I understand that the PR #8128 is still in progress and this issue is assigned P0 in the [project - 1.15 release](https://github.com/orgs/dapr/projects/87).
Just wanted to understand if we have any timelines for the release of this change. 

We are using dapr in production and it would be a great thing to be able to filter resilience policies based on http status codes. Since we are unable to do it via dapr right now, we have to do it using code in our clients. We saw the PR #7132 about 5 months ago and have been waiting for it since then :)

@mtuppad

Issue #7124, opened 2023-10-31
1.12.1 Patch Release Checklist


JoshVanL:
- [ ] Dapr maintainer(s) communicate out the discovery of issues that require a patch release and the reason why to the Discord Maintainers and Release channels. Patch releases are made for one of categories below:
   - [ ] Security vulnerability
     - https://github.com/dapr/dapr/pull/7116
   - [ ] Regression that does not have a work around:
     - https://github.com/dapr/components-contrib/pull/3199
     - https://github.com/dapr/dapr/pull/7066
     - https://github.com/dapr/dapr/pull/7082
     - https://github.com/dapr/dapr/pull/7094
     - https://github.com/dapr/components-contrib/issues/3204
     - https://github.com/dapr/dapr/pull/7172
   - [ ] Broken mainline scenario that has a missing test case:
- [ ] Create v1.12.1-rc.2 Tag
- [ ] Performance tests passing: 
- [ ] End to End tests passing on Linux and Windows: 
- [ ] New test case written to catch future occurrences 
- [ ] Notify users to try RC1
- [ ] Update the longhaul tests to use v1.12.1-rc.2
- [ ] Write release notes :  https://github.com/dapr/dapr/pull/7150
- [ ] Review release notes [@dapr/maintainers-dapr ]
- [ ] Create v1.12.1 Tag [@dapr/maintainers-dapr ]
- [ ] Backport fixes into master branch [@dapr/maintainers-dapr ]
- [ ] Update the documentation: Latest version & versions in supported releases
- [ ] Push new tag in installer-bundle repo.
- [ ] Announce the patch release on Discord Announcements channel

artursouza:
Pending the Consul fix.

PR #7132, opened 2023-11-01
Resiliency: add filter based on response http status code for retry

Taction:
# Description

ADD Status filter for retry resiliency. An example config:
```yaml
apiVersion: dapr.io/v1alpha1
kind: Resiliency
metadata:
  name: slotting-pubsub-resiliency
  namespace: apps
scopes:
spec:
  policies:
    retries:
      linerRetry:
        policy: constant
        duration: 5s
        maxRetries: 3
        retryOnCodes:
          - "500"
          - "503"
        ignoredOnCodes:
          - "400-499"
```

Each item is a code range or a specific code or a combination separated by comma, followings are valid "200-299", "300-309,315", "400,402,404,", "500".

This currently working for input binding, http service call, grpc proxy.

Updated design:
```yaml
apiVersion: dapr.io/v1alpha1
kind: Resiliency
metadata:
  name: slotting-pubsub-resiliency
  namespace: apps
scopes:
spec:
  policies:
    retries:
      linerRetry:
        policy: constant
        duration: 5s
        maxRetries: 3
        matching:
          httpStatusCodes: "500,501,503-510"
          grpcStatusCodes: "1-5,8"
```

## Issue reference

Please reference the issue this PR will close: #6683

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Unit tests passing
* [x] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#3859


pkg/resiliency/retry.go
```diff
@@ -0,0 +1,152 @@
...
+	}
+	f.ignoreOnPatterns = regexpPatterns
+	f.defaultRetry = true
+	return nil
+}
+
+func (f *StatusCodeFilter) StatusCodeNeedRetry(statusCode int32) bool {
+	statusCodeStr := strconv.Itoa(int(statusCode))
+
+	// Check whitelist patterns
```
ItalyPaleAle, line 123:
```suggestion
	// Check retriable codes (allowlist)
```

pkg/resiliency/retry.go
```diff
@@ -0,0 +1,152 @@
...
+	for _, pattern := range f.retryOnPatterns {
+		if pattern.MatchString(statusCodeStr) {
+			return true
+		}
+	}
+
+	// Check blacklist patterns
```
ItalyPaleAle, line 130:
```suggestion
	// Check ignored codes (denylist)
```

pkg/http/api_directmessaging.go
```diff
@@ -191,12 +191,12 @@ func (a *api) onDirectMessage(w http.ResponseWriter, r *http.Request) {
 		} else if resStatus.Code < 200 || resStatus.Code > 399 {
 			msg, _ := rResp.RawDataFull()
 			// Returning a `codeError` here will cause Resiliency to retry the request (if retries are enabled), but if the request continues to fail, the response is sent to the user with whatever status code the app returned.
-			return rResp, codeError{
+			return rResp, resiliency.NewCodeError(resStatus.Code, codeError{
```
ItalyPaleAle, line 194: We need this change for gRPC too?

pkg/resiliency/retry.go
```diff
@@ -0,0 +1,152 @@
...
+
+func NewCodeError(statusCode int32, err error) CodeError {
+	return CodeError{
+		StatusCode: statusCode,
+		err:        err,
+	}
+}
+
+type Retry struct {
+	*retry.Config
```
ItalyPaleAle, line 47: Make these 2 not pointers please

pkg/resiliency/retry.go
```diff
@@ -0,0 +1,152 @@
...
+
+func NewRetry(retryConfig *retry.Config, statusCodeFilter *StatusCodeFilter) *Retry {
+	return &Retry{
+		Config:           retryConfig,
+		StatusCodeFilter: statusCodeFilter,
+	}
+}
+
+type StatusCodeFilter struct {
+	retryOnPatterns  []*regexp.Regexp
```
ItalyPaleAle, line 59: Make these not pointers: `[]regexp.Regexp`

pkg/resiliency/retry.go
```diff
@@ -0,0 +1,152 @@
...
+	f.defaultRetry = true
+	return nil
+}
+
+func (f *StatusCodeFilter) StatusCodeNeedRetry(statusCode int32) bool {
+	statusCodeStr := strconv.Itoa(int(statusCode))
+
+	// Check whitelist patterns
+	for _, pattern := range f.retryOnPatterns {
+		if pattern.MatchString(statusCodeStr) {
```
ItalyPaleAle, line 125: These are using regexps which are kind of slow. Could we use ranges rather than regexps? Since they are numbers?

ItalyPaleAle (changes requested):

pkg/resiliency/retry.go
Taction: Yes, this makes sense. How about using the following patterns:
- a specific http code, such as "500"
- http code range, such as "500-509", 500 and 509 are included.
- *now can be removed, since 5** can be defined as "500-599"

ItalyPaleAle: I think ranges are even clearer and more powerful. So, it could be:

- Precise status code: `500`
- Multiple status codes: `404,412` (comma-separated)
- Ranges: `400-499`
- A combination of the two above: `404,500-599`

Please let's do the parsing of the rule only once during initialization of the resiliency policies, and not on each request.

pkg/resiliency/retry.go
```diff
@@ -0,0 +1,165 @@
...
+		filter.defaultRetry = true
+	}
+	return filter, nil
+}
+
+func (f StatusCodeFilter) parsePatterns(patterns []string) ([]codeRange, error) {
+	parsedPatterns := make([]codeRange, 0, len(patterns))
+	for _, item := range patterns {
+		// one item can contain multiple patterns separated by comma
+		splited := strings.Split(item, ",")
```
ItalyPaleAle, line 107: Nit: `splitted`

pkg/resiliency/retry.go
```diff
@@ -0,0 +1,165 @@
...
+		for _, p := range splited {
+			parsedPattern, err := compilePattern(p)
+			if err != nil {
+				return nil, fmt.Errorf("failed to compile item %s, %w", item, err)
```
ItalyPaleAle, line 111: Nit:

```suggestion
				return nil, fmt.Errorf("failed to compile pattern '%s': %w", item, err)
```

pkg/resiliency/retry.go
```diff
@@ -0,0 +1,165 @@
...
+			return false
+		}
+	}
+
+	return f.defaultRetry
+}
+
+func compilePattern(pattern string) (codeRange, error) {
+	// parse code range pattern
+	if strings.Contains(pattern, "-") {
```
ItalyPaleAle, line 140: Two things.

First, please check to make sure that `pattern != ""`

Second, this could be simplified a bit:

1. Always do `strings.Split(pattern, "-")`
2. `if len(patterns) > 2` then error
2. Parse `start, err := strconv.Atoi(patterns[0])`
3. Set `end := start`
4. `if len(patterns) == 2` then also do `end, err = strconv.Atoi(patterns[1])` 
5. **Also** need to make sure that `start <= end` or error

pkg/resiliency/retry_test.go
```diff
@@ -0,0 +1,56 @@
+package resiliency
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestResiliencyFunctions(t *testing.T) {
```
ItalyPaleAle, line 9: Please add test cases for gRPC status codes (which are < 100)

pkg/resiliency/retry.go
Taction: done

pkg/resiliency/retry.go
Taction: done.

pkg/resiliency/retry.go
Taction: done

pkg/resiliency/retry_test.go
Taction: done.

pkg/http/api_directmessaging.go
Taction: Sorry for the late reply, it takes me some time to look into the code.

Added for grpc_proxy mode, along with the e2e tests.

For http directmessaging, if the application uses the grpc protocol, when the application returns a grpc codes error, it is parsed into the rResp and does not treat it as an error(so would not retry). see: https://github.com/dapr/dapr/blob/master/pkg/channel/grpc/grpc_channel.go#L126-L133
https://github.com/dapr/dapr/blob/master/pkg/http/api_directmessaging.go#L173-L190
So currently not added, If we want to add it for this case, I think it is worth opening a new issue.

charts/dapr/crds/resiliency.yaml
```diff
@@ -64,6 +64,14 @@ spec:
...
                         policy:
                           type: string
+                        retryOnCodes:
+                          type: array
+                          items:
+                            type: string
+                        ignoreOnCodes:
+                          type: array
+                          items:
+                            type: string
```
JoshVanL, line 74: It is, at least to me, not immediately obvious what "ignore" means in this context. `noRetry` might be clearer to users. 

```suggestion
                        noRetryOnCodes:
                          type: array
                          items:
                            type: string
```

pkg/resiliency/retry_test.go
```diff
@@ -0,0 +1,80 @@
...
+		{
+			name:        "TestInvalidStartCodeBiggerThanEndCode",
+			retryOn:     []string{"399-299"},
+			expectError: true,
+		},
+		{
+			name:        "TestInvalidCodeRange",
+			retryOn:     []string{"199-299-300"},
+			expectError: true,
+		},
```
JoshVanL, line 61: Can we add some fuzz tests here https://github.com/google/gofuzz ?

What happens when retry and ignore cover the same range?

JoshVanL (changes requested):
Please can you create a documentation PR detailing the features added here? 

> This currently working for input binding, http service call, grpc proxy.

Can you be explicit what _isn't_ covered by this PR?

---

Please can you also include [integration tests](https://github.com/dapr/dapr/blob/master/tests/docs/writing-integration-test.md) for this PR? Not only good practice and coverage sorely needed in this are, both resiliency and handling of HTTP status codes has led to backports for 1.12 https://github.com/dapr/dapr/issues/7124

pkg/http/api_directmessaging.go
ItalyPaleAle: We need to cover the case of gRPC -> gRPC and gRPC -> HTTP too. These are legacy methods but still supported.

pkg/resiliency/retry_test.go
Taction: Sure, I'll add fuzz tests. 
When retry and ignore are used together, an error will occur.

Taction:
> Please can you create a documentation PR detailing the features added here?
> > This currently working for input binding, http service call, grpc proxy.
> Can you be explicit what isn't covered by this PR?

I'll start work on the doc recently. also with the integration tests.

In the resiliency e2e tests, it covers binding、service call(http/grpc)、pubsub、actor. pubsub(subscription) has not been added this feature(plan to add maybe not in this pr), and I am not sure if actor is needed.

JoshVanL:
I have a WIP local branch working through this which I'm happy to keep going with if you want @Taction :slightly_smiling_face:

Taction:
Sorry for the late reply. @JoshVanL if you have already worked on this, you can take over this and proceed. Otherwise, I would start continuing this work this weekend.

charts/dapr/crds/resiliency.yaml
Taction: Done.

pkg/resiliency/retry_test.go
Taction: @JoshVanL I may not have enough time to complete this test recently, since you mentioned you have a WIP local branch, can you help with this test as well as the integration test?

Taction:
@yaron2 added this feature to pubsub subscription call with e2e tests. Basically ready for review except for the fuzz and integration tests mentioned above.

yaron2:
@Taction this looks good to merge, can you please resolve the conflict? I can't seem to do it on your branch.

JoshVanL:
Before merging, please can we make sure we have a corresponding docs PR, as well as integration tests for both HTTP and gRPC protocols

tests/integration/suite/daprd/resiliency/apps/retryfilter.go
```diff
@@ -0,0 +1,241 @@
...
+			return
+		}
+
+		c, _ := d.counters.LoadOrStore(key, &atomic.Int32{})
+		counter := c.(*atomic.Int32)
+		counter.Add(1)
+
+		respStatusCode, err := strconv.Atoi(string(body))
+		if (err != nil) || (respStatusCode < 100) || (respStatusCode >= 600) {
+			// Trying to write a bad status code will cause app to crash.
```
JoshVanL, line 73: To be clear what is crashing here? Nothing should ever be crashing due to this.

tests/integration/suite/daprd/resiliency/apps/noretryfilter.go
```diff
@@ -0,0 +1,241 @@
...
+          - "546,547"
+`
+	d.daprd1 = daprd.New(t,
+		daprd.WithAppPort(srv.Port()),
+		daprd.WithResourceFiles(resiliency),
+	)
+	d.daprd2 = daprd.New(t,
+		daprd.WithAppPort(srv.Port()),
+		daprd.WithResourceFiles(resiliency),
+	)
```
JoshVanL, line 111: Please can we also include tests for the gRPC protocol.

tests/integration/suite/daprd/resiliency/apps/noretryfilter.go
```diff
@@ -0,0 +1,241 @@
...
+      DefaultAppRetryPolicy:
+        policy: constant
+        duration: 100ms
+        maxRetries: 3
+        noRetryOnCodes:
+          - "505"
+          - "2"
+          - "invalid"
+          - "501-503,509"
+          - "546,547"
```
JoshVanL, line 102: What happens when this is defined along with `retryOnCodes`, and what happens when they overlap?

tests/integration/suite/daprd/resiliency/apps/noretryfilter.go
```diff
@@ -0,0 +1,241 @@
...
+  policies:
+    retries:
+      DefaultAppRetryPolicy:
+        policy: constant
+        duration: 100ms
+        maxRetries: 3
+        noRetryOnCodes:
+          - "505"
+          - "2"
+          - "invalid"
```
JoshVanL, line 100: I would expect the entire Resiliency policy to be rejected on invalid config like this. Without that, it's impossible for the user to know if the config has actually been successfully consumed by Dapr as we have no status fields.

pkg/resiliency/retry_test.go
```diff
@@ -0,0 +1,81 @@
+package resiliency
```
JoshVanL, line 1: Copyright header.

JoshVanL (changes requested):
We still need some fuzz tests and a docs PR.

tests/integration/suite/daprd/resiliency/apps/noretryfilter.go
artursouza: My expectation was the opposite. If there is one invalid entry, invalidating the entire policy is overkill. Invalidating the entire policy does not do anything more than add a log entry, just like what happens with a single invalid record. The user will then be surprised about why retries suddenly stopped working.

What do others expect? Having some quantitative data on what users expect will help settle this.

tests/integration/suite/daprd/resiliency/apps/noretryfilter.go
artursouza: There is an error and the policy is invalid.

tests/integration/suite/daprd/resiliency/apps/noretryfilter.go
artursouza: Yes.

tests/integration/suite/daprd/resiliency/apps/retryfilter.go
artursouza: There is a stacktrace. The app does not crash per se. Updating comment.

pkg/http/api_directmessaging.go
artursouza: We don't support gRPC to HTTP.

tests/integration/suite/daprd/resiliency/apps/noretryfilter.go
bibryam: When there is overlap, can there be a convention like "the new entry overrides the old one"...so the rule is still valid and not ignored silently

charts/dapr/crds/resiliency.yaml
```diff
@@ -64,6 +64,12 @@ spec:
                           type: integer
                         policy:
                           type: string
+                        conditions:
```
JoshVanL, line 67: Do we want to use `filter` or `match` instead? We are using the word filter elsewhere, so we should at least be consistent.

pkg/apis/resiliency/v1alpha1/types.go
```diff
@@ -51,10 +51,16 @@ type Policies struct {
...
+	Policy      string           `json:"policy,omitempty" yaml:"policy,omitempty"`
+	Duration    string           `json:"duration,omitempty" yaml:"duration,omitempty"`
+	MaxInterval string           `json:"maxInterval,omitempty" yaml:"maxInterval,omitempty"`
+	MaxRetries  *int             `json:"maxRetries,omitempty" yaml:"maxRetries,omitempty"`
+	Conditions  *RetryConditions `json:"conditions,omitempty" yaml:"conditions,omitempty"`
+}
+
+type RetryConditions struct {
+	HTTPStatusCodes string `json:"httpStatusCodes,omitempty" yaml:"httpStatusCodes,omitempty"`
+	GRPCStatusCodes string `json:"gRPCStatusCodes,omitempty" yaml:"gRPCStatusCodes,omitempty"`
```
JoshVanL, line 63:
```suggestion
        // +optional
	HTTPStatusCodes *string `json:"httpStatusCodes,omitempty" yaml:"httpStatusCodes,omitempty"`
	 // +optional
	GRPCStatusCodes *string `json:"gRPCStatusCodes,omitempty" yaml:"gRPCStatusCodes,omitempty"`
```

JoshVanL, line 63: Please add API comments. These also appear in the kubectl describe output.

pkg/resiliency/retry.go
```diff
@@ -0,0 +1,208 @@
+/*
+Copyright 2023 The Dapr Authors
```
JoshVanL, line 2:
```suggestion
Copyright 2024 The Dapr Authors
```

pkg/resiliency/retry.go
```diff
@@ -0,0 +1,208 @@
...
+	// Reference: https://grpc.io/docs/guides/status-codes/
+	validGRPCStatusCodeRange = codeRange{0, 16}
+)
+
+type retryScenario int
+
+const (
+	retryScenarioHTTPInvoke retryScenario = iota
+	retryScenarioGRPCInvoke
+	// add more retry scenarios here.
```
JoshVanL, line 37:
```suggestion
```

JoshVanL (changes requested):

charts/dapr/crds/resiliency.yaml
artursouza: I don't like "filter" because of the ambiguity: is it filtering in or filtering out.

artursouza: I will consider `match`. Fixing E2E test first.

charts/dapr/crds/resiliency.yaml
```diff
@@ -64,6 +64,12 @@ spec:
...
+                          httpStatusCodes:
+                            type: string
+                          gRPCStatusCodes:
+                            type: string
+                          type: object
```
antontroshin, line 72:
```suggestion
                        conditions:
                          properties:
                            httpStatusCodes:
                              type: string
                            gRPCStatusCodes:
                              type: string
                          type: object
```

charts/dapr/crds/resiliency.yaml
artursouza: Changing to `matching`

pkg/apis/resiliency/v1alpha1/types.go
artursouza: nil and empty string are semantically the same here. Also, keeping it as `string` for consistency with the rest of the API definition here.

artursouza: Done.

pkg/resiliency/retry.go
artursouza: Done.

artursouza (approved):
Approved but I am also a co-author. Waiting on @yaron2 @JoshVanL @daixiang0 to approve it too. Changed it to 1.14 release.

ulf-melin-sveasolar:
Hi there, time to finish this up? Been longing for it to come into being! Keep up the good work, really appreciate the things you all do.

antontroshin:
Closing, superseded by https://github.com/dapr/dapr/pull/8252

PR #7134, opened 2023-11-01
Placement: disseminate min Actor API version + reject nodes with version lower than minium

### LINKED ISSUE #6838: Include runtime version in Dapr placement data to ease with upgrades

ItalyPaleAle:
Fixes #6838

Completes the work that was started in 1.12:

- The placement service collects all API versions from dapr (which were already reported starting with Dapr 1.12, but unused)
- If a node tries to join a cluster and its actor API version is lower than the minimum observed in the cluster, it's rejected
- It's possible to manually set a "ceiling" using `-max-api-version` when starting the placement service
- The cluster API version is disseminated to all sidecars connected
- Once a cluster reaches a given API level, it can never be downgraded (unless users pass `-max-api-version`). This is to prevent issues such as clusters with mixed versions, in which Placement is restarted, and the minimum version depends on the order the apps come back online at.

The behavior is validated with new ITs, as well as updated unit tests.

Note:

- Although daprd now receives the API level for the cluster, it is currently unused in Daprd - meaning, it doesn't yet use it to make any decision (a decision will be #7129)


charts/dapr/README.md
```diff
@@ -141,6 +141,7 @@ The Helm chart has the follow configuration options that can be supplied:
 | `dapr_placement.cluster.logStoreWinPath`       | Mount path for persistent volume for log store in windows when HA is true                                                                                                 | `C:\\raft-log`          |
 | `dapr_placement.volumeclaims.storageSize`      | Attached volume size                                                                                                                                                      | `1Gi`                   |
 | `dapr_placement.volumeclaims.storageClassName` | storage class name                                                                                                                                                        |                         |
+| `dapr_placement.maxApiLevel`                  | Sets the `max-api-level` flag which prevents the Actor API level for going above this value. The default value of -1 means no cap.  | `-1` |
```
mukundansundar, line 144: we know that the placement server is only for actors, but should the parameter name still be made clear? `dapr_placement.maxActorApiLevel`

pkg/placement/placement.go
```diff
@@ -234,16 +258,18 @@ func (p *Service) ReportDaprStatus(stream placementv1pb.Placement_ReportDaprStat
 				continue
 			}
 
+			now := p.clock.Now()
+
 			for _, entity := range req.Entities {
-				monitoring.RecordActorHeartbeat(req.Id, entity, req.Name, req.Pod, p.clock.Now())
+				monitoring.RecordActorHeartbeat(req.Id, entity, req.Name, req.Pod, now)
```
mukundansundar, line 264: is this a blocking call which takes some time, and if so should `now` be calculated again after the loop?

mukundansundar (commented):
should a `minActorApiLevel` also be a configurable parameter?

charts/dapr/README.md
ItalyPaleAle: Makes sense, changing that.

ItalyPaleAle:
> should a `minActorApiLevel` also be a configurable parameter?

I didn't think of that at the beginning, but I now can see a use case for this. I will add it.

> Do we consider the APILevel there in placement service as the min API level or should we have a separate value

The API level reported by placement is the minimum observed API level in the cluster

>  i.e. let's say that there is complete backwards compatibility between levels 10, 11 and 12 ..

The design doc #6838 states that versions are never fully backwards-compatible. The version _does not_ track the daprd version however, so if no change is made that requires updating the version in a Dapr release, it won't be updated.

yaron2:
`Sets the `max-api-level` flag which prevents the Actor API level for going above this value. The default value of -1 means no cap` - this description isn't clear to users. lets make this very clear so users know why they're given this option, otherwise this Helm chart settings needs to be removed.

charts/dapr/README.md
ItalyPaleAle: Done

ItalyPaleAle:
Updated per review feedback. Added "min" option and clarified message on Helm chart.

tests/integration/suite/placement/apilevel/no_max.go
```diff
@@ -0,0 +1,141 @@
...
+	require.EventuallyWithT(t, func(t *assert.CollectT) {
+		tableVersion = checkAPILevelInState(t, n.place.HealthzPort(), 10)
+	}, 5*time.Second, 100*time.Millisecond)
+
+	// Register the second host with API level 20
+	stopCh2 := make(chan struct{})
+	registerHost(ctx, conn, "myapp2", 20, placementMessageCh, stopCh2)
+
+	// After 3s, we should not receive an update
+	// This can take a while as disseination happens on intervals
```
cicoyle, line 107: nit, otherwise great! I appreciate the comments - its really nice to follow 👍🏻 
```suggestion
	// This can take a while as dissemination happens on intervals
```

ItalyPaleAle: Thanks!

tests/integration/suite/placement/apilevel/with_max.go
```diff
@@ -0,0 +1,124 @@
...
+
+	require.EventuallyWithT(t, func(t *assert.CollectT) {
+		checkAPILevelInState(t, n.place.HealthzPort(), 10)
+	}, 5*time.Second, 100*time.Millisecond)
+
+	// Register the second host with API level 20
+	registerHost(ctx, conn, "myapp2", 20, placementMessageCh, nil)
+
+	// After 3s, we should not receive an update
+	// This can take a while as disseination happens on intervals
```
cicoyle, line 106: I think here too
```suggestion
	// This can take a while as dissemination happens on intervals
```

cicoyle:
LGTM -Nice job! 🎉

daixiang0 (dismissed):
LGTM

JoshVanL (changes requested):

tests/integration/suite/placement/apilevel/no_max.go
```diff
@@ -0,0 +1,141 @@
...
+	time.Sleep(3 * time.Second)
+	require.Equal(t, lastUpdate, lastVersionUpdate.Load())
+
+	// API level should still be 10 (but table version should have increased)
+	require.EventuallyWithT(t, func(t *assert.CollectT) {
+		newTableVersion := checkAPILevelInState(t, n.place.HealthzPort(), 10)
+		assert.Greater(t, newTableVersion, tableVersion)
+	}, 10*time.Second, 100*time.Millisecond)
```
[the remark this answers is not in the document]
ItalyPaleAle: Why did you remove the check above? We need to make sure that we don't receive a signal, and the only way to do that is to sleep and then check we did _not_ get a signal.

Checking for absence of something is not easy.

tests/integration/suite/placement/apilevel/shared.go
```diff
@@ -0,0 +1,178 @@
...
+		if rErr != nil {
+			log.Printf("Failed to receive message; will retry: %v", rErr)
+			_ = stream.CloseSend()
+			// Sleep before retrying
+			time.Sleep(1500 * time.Millisecond)
+			continue
+		}
+
+		placementStream = stream
+	}
```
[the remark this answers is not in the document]
ItalyPaleAle: I don't want to pass a `t` to this method, makes handling some errors harder (I'd have to put a lot of `if !assert.NoError`)

charts/dapr/README.md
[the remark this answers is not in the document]
ItalyPaleAle: How about I add the more detailed explanation in the docs? I'd rather not write a long paragraph here.

Most users will not need to set this, but it can be necessary in some rare situations.

By default this is automatically set to the minimum observed value in the cluster. For example, say you have 3 apps, with these actor API levels: 10, 20, 20. The minimum observed is 10, so that will be the level used in the cluster.

Next, assume that the app with level 10 goes away, so the minimum level now becomes 20.

The level can never be decreased, only increased. Which means that if now the app with level 10 comes back, it will not be able to connect to Placement (and use actors).

In this very specific case, users can set `maxApiLevel` to `10` to enforce the reported level is never higher than 10.

I can see this possibly being the case during upgrades or if users have the need to run older versions of a sidecar alongside new ones. It should not be a common case, however.

tests/integration/suite/placement/apilevel/shared.go
```diff
@@ -0,0 +1,178 @@
...
+	"google.golang.org/grpc/codes"
+	grpcinsecure "google.golang.org/grpc/credentials/insecure"
+	"google.golang.org/grpc/status"
+
+	placementv1pb "github.com/dapr/dapr/pkg/proto/placement/v1"
+)
+
+var client = http.Client{
+	Timeout: 5 * time.Second,
+}
```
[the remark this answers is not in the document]
ItalyPaleAle: Can't use that because:

1. It returns a new client on every call, so not leveraging TCP keepalives
2. It requires `t *testing.T` and not all callers have it.

pkg/placement/membership.go
```diff
@@ -353,6 +353,16 @@ func (p *Service) performTablesUpdate(ctx context.Context, hosts []placementGRPC
 	// Otherwise, each Dapr runtime will have inconsistent hashing table.
 	startedAt := p.clock.Now()
 
+	// Enforce maximum API level
+	if newTable != nil {
+		if newTable.ApiLevel < uint32(p.minAPILevel) {
+			newTable.ApiLevel = uint32(p.minAPILevel)
+		}
+		if p.maxAPILevel >= 0 && int(newTable.ApiLevel) > p.maxAPILevel {
+			newTable.ApiLevel = uint32(p.maxAPILevel)
```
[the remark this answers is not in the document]
ItalyPaleAle: Using a pointer with a nil-value seems like a good idea

tests/integration/suite/placement/apilevel/shared.go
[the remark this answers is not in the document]
JoshVanL: It's very important that we do not share state between our tests in this way. We also need to ensure to kill idle connection on test cleanup https://github.com/dapr/dapr/blob/master/tests/integration/framework/util/http.go#L32C1-L32C40

A single client can be created which is shared, only calling `HTTPClient` once for each test.

> It requires t *testing.T and not all callers have it.

Can't we add `client *http.Client` as a parameter to callers?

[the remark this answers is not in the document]
ItalyPaleAle: Done

tests/integration/suite/placement/apilevel/with_max.go
ItalyPaleAle: The "m" on my keyboard sometimes fails -.-"

JoshVanL (approved):

PR #7141, opened 2023-11-02
Treat `duration` as `initialInterval` for exponential policies

berndverst:
Temporary workaround because `initialInterval` is missing in CRDs.


ItalyPaleAle (approved):
LGTM although this actually does add  “initialInterval” to the CRD, with a fallback on “duration”

yaron2:
What is the motivation for this PR?

Issue #7127, opened 2023-10-31
Proposal: Alternate backends for Dapr Workflow


cgillum:
## In what area(s)?

> /area runtime

## Describe the proposal

This is a proposal for allowing users to configure alternate backend implementations for Dapr Workflow, beyond the default Actors backend.

## Background

Dapr Workflow relies on the Durable Task Framework for Go (a.k.a. [durabletask-go](https://github.com/microsoft/durabletask-go)) as the core engine for executing workflows. This engine is designed to support multiple backend implementations. For example, the durabletask-go repo includes a SQLite implementation and the Dapr repo includes an Actors implementation. For Dapr Workflow users today, the Actors implementation is the only option.

The backend implementation is largely decoupled from the core engine or the programming model that developers see. Rather, the backend primarily impacts how workflow state is stored and how workflows execution is coordinated across replicas. In that sense, it is similar to Dapr's state store abstraction, except designed specifically for workflow. All APIs and programming model features are the same regardless of which backend is used.

## Proposal

This proposal is about opening up the possibility of using alternate `durabletask-go` *backend* implementations for Dapr Workflow. For example, the [existing (POC) SQLite implementation](https://github.com/microsoft/durabletask-go/tree/main/backend/sqlite) could be used for local development and testing, or a custom, vendor-specific implementation could be used for specialized scenarios. Dapr Workflow would continue to use the Actors implementation by default, and there is currently no desire to change this.

To be clear, this proposal does *not* propose making the Dapr Workflow backend into a full-fledged building block. Rather, it would be limited to backends that are either compiled into the `durabletask-go` project or into the Dapr runtime, like how the Actors backend is today. This is a scoping decision designed to keep the proposal simple and focused, but could be revisited if there's sufficient interest in community-contributed workflow backends.

Also, this is *not* a proposal to make the APIs for Dapr Workflow pluggable like we did in the start of the project - e.g., with a Temporal API implementation. The workflow management APIs would continue to target only the built-in Dapr Workflow implementation (and would not need to be updated to support alternate backends).

Lastly, this proposal doesn't recommend any specific alternate implementations. SQLite is mentioned as a practical example that could be made available for all Dapr users since a (POC) implementation already exists. But there are other possibilities, such as a Postgres implementation, or even a gRPC-based implementation that could target a highly customized remote workflow backend implementation.

### Configuration options

[...]

PR #7152, opened 2023-11-06
Defining Workflow metrics

### LINKED ISSUE #7109: [Workflow] Adding metrics for Dapr Workflow
## In what area(s)?

/area runtime

## Describe the feature
Workflow should emit metrics that help Dapr users with tracking the overall traffic and health, also improving overall diagnosability.

### Dapr Runtime – current code 

The following metrics can be extracted from the Dapr sidecars. 

#### workflow/operations 

These metrics cover the total successful and failed requests to create, get, purge workflows, and add events. It also covers the overall latency to execute these requests. Note, in case of `create workflow` and `add event`, it only measures the time taken to create the reliable reminder. 

#### workflow/reminders 

These metrics cover the total number of internal-actor reminder requests created, which includes reminders to start a new workflow, run an activity, raise a new workflow event, or create durable timers. 

#### workflow/execution 

These metrics cover the total successful and failed workflow/activity execution requests. When failing, it also contains if the failure was recoverable (i.e., will be retried via actor reminders) or not. 

#### workflow/scheduling	 

By adding request-time in the reminder payload, we can measure the time between (1) a workflow being scheduled and the workflow’s actual start, and (2) an activity being created and the activity’s actual start. 

#### workflow/execution 

Using the same payload as above, when a workflow or activity finally finishes (with success or non-recoverable error), capture the total time taken to run a workflow or activity. 

### Dapr SDKs 

Dapr SDKs can configure Workflow resiliency in the underlying implementation of Durable Task Framework (DTF). For example, [durabletask-dotnet](https://github.com/microsoft/durabletask-dotnet/blob/25f4220052ea2302c2f2b2ea11892f8f0a88503d/src/Worker/Core/Shims/TaskOrchestrationContextWrapper.cs#L116) and [durabletask-python](https://github.com/microsoft/durabletask-python/pull/11) have retry policies that take effect while calling activities, which are then configurable in the Dapr SDKs like [dotnet-sdk](https://github.com/dapr/dotnet-sdk/blob/3b979e6bdb1d779563f1656fa684183b2bfecd08/examples/Workflow/WorkflowConsoleApp/Workflows/OrderProcessingWorkflow.cs#L12). However, not all Dapr SDKs (e.g., Python or Java) allow the same today. 

When Dapr SDKs start supporting retry policies consistently, we can bubble the retry details from DTF clients to Dapr SDKs, and have them shared with the Dapr sidecars to be emitted as a metric. This can be like the [existing sidecar resiliency metrics](https://github.com/dapr/dapr/blob/master/docs/development/dapr-metrics.md#resiliency). For this issue, **this can be considered out of scope**.

## Summary

Metric name | Description | Tags | Status
-- | -- | -- | --
runtime/workflow/operations/total | The number of successful workflow operation requests. | operation | Ready for implementation
runtime/workflow/operations/failed_total | The number of failed workflow operation requests. | operation | Ready for implementation
runtime/workflow/operations/latencies | The latencies of responses for workflow operation requests. | elapsed | Ready for implementation
runtime/workflow/reminders/total | The number of workflow/activity reminders created. | reminder-type | Ready for implementation
runtime/workflow/execution/total | The number of successful workflow/activity executions. | execution-type | Ready for implementation
runtime/workflow/execution/failed_total | The number of failed workflow/activity executions. | execution-type, is-retriable | Ready for implementation
runtime/workflow/execution/latencies | The total time taken to run a workflow/activity to completion. | execution-type, status | Needs runtime changes
runtime/workflow/scheduling/latencies | The latencies between execution request and actual execution | execution-type | Needs runtime changes

## Release Note

RELEASE NOTE: **ADD** metrics for Dapr Workflow

prateek041:
# Description

This pull request adds workflow metrics.

## Issue reference

Will fix #7109 

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,8 @@
 module github.com/dapr/dapr
 
-go 1.20
+go 1.21
+
+toolchain go1.21.3
 
 require (
 	contrib.go.opencensus.io/exporter/prometheus v0.4.2
--- a/pkg/diagnostics/workflow_metrics.go
+++ b/pkg/diagnostics/workflow_metrics.go
@@ -1,0 +1,120 @@
+/*
+Copyright 2021 The Dapr Authors
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package diagnostics
+
+import (
+	diagUtils "github.com/dapr/dapr/pkg/diagnostics/utils"
+	"go.opencensus.io/stats"
+	"go.opencensus.io/stats/view"
+	"go.opencensus.io/tag"
+)
+
+var (
+	elapsedKey = tag.MustNewKey("elapsed")
+
+	reminderTypeKey = tag.MustNewKey("reminder_type")
+
+	executionTypeKey = tag.MustNewKey("execution_type")
+
+	isRetryableKey = tag.MustNewKey("is_retryable")
+)
+
+type workflowMetrics struct {
+	// Total successful workflow operation requests
+	workflowOperationsTotal *stats.Int64Measure
+	// Total failed workflow operation requests
+	workflowOperationsFailedTotal *stats.Int64Measure
+	// Workflow operation request's Response latency
+	workflowOperationsLatency *stats.Float64Measure
+
+	// Total workflow/activity reminders created
+	workflowRemindersTotal *stats.Int64Measure
+
+	// Total successful workflow/activity executions
+	workflowExecutionTotal *stats.Int64Measure
+	// Total failed workflow/activity executions
+	workflowExecutionFailedTotal *stats.Int64Measure
+	// Time taken to run a workflow to completion
+	workflowExecutionLatency *stats.Float64Measure
+
+	// Latency between execution request and actual execution
+	workflowSchedulingLatency *stats.Float64Measure
+
+	appID   string
+	enabled bool
+}
+
+func newWorkflowMetrics() *workflowMetrics {
+	return &workflowMetrics{
+		workflowOperationsTotal: stats.Int64(
+			"runtime/workflow/operations/total",
+			"The number of successful workflow operation requests.",
+			stats.UnitDimensionless),
+
+		workflowOperationsFailedTotal: stats.Int64(
+			"runtime/workflow/operations/failed_total",
+			"The number of failed workflow operations requests.",
+			stats.UnitDimensionless),
+		workflowOperationsLatency: stats.Float64(
+			"runtime/workflow/operations/latency",
+			"The latencies of responses for workflow operation requests.",
+			stats.UnitMilliseconds),
+		workflowRemindersTotal: stats.Int64(
+			"runtime/workflows/reminders/total",
+			"The number of workflows/activity executions.",
+			stats.UnitDimensionless),
+		workflowExecutionTotal: stats.Int64(
+			"runtime/workflow/execution/total",
+			"The number of successful workflow/activity executions.",
+			stats.UnitDimensionless),
+		workflowExecutionFailedTotal: stats.Int64(
+			"runtime/workflow/execution/failed_total",
+			"The number of failed workflow/activity executions.",
+			stats.UnitDimensionless),
+		workflowExecutionLatency: stats.Float64(
+			"runtime/workflow/execution/latency",
+			"The total time taken to run a workflow/activity to completion.",
+			stats.UnitMilliseconds),
+		workflowSchedulingLatency: stats.Float64(
+			"runtime/workflow/scheduling/latency",
+			"The latency between execution request and actual execution.",
+			stats.UnitMilliseconds),
+	}
+}
+
+func (w *workflowMetrics) IsEnabled() bool {
+	return w != nil && w.enabled
+}
+
+// Init registers the workflow metrics views.
+func (w *workflowMetrics) Init(appID string) error {
+	w.appID = appID
+	w.enabled = true
+
+	return view.Register(
+		diagUtils.NewMeasureView(w.workflowOperationsTotal, []tag.Key{appIDKey, operationKey}, view.Count()),
+		diagUtils.NewMeasureView(w.workflowOperationsFailedTotal, []tag.Key{appIDKey, operationKey}, view.Count()),
+		diagUtils.NewMeasureView(w.workflowOperationsLatency, []tag.Key{appIDKey, elapsedKey}, defaultLatencyDistribution),
+		diagUtils.NewMeasureView(w.workflowRemindersTotal, []tag.Key{appIDKey, reminderTypeKey}, view.Count()),
+		diagUtils.NewMeasureView(w.workflowExecutionTotal, []tag.Key{appIDKey, executionTypeKey}, view.Count()),
+		diagUtils.NewMeasureView(w.workflowExecutionFailedTotal, []tag.Key{appIDKey, executionTypeKey, isRetryableKey}, view.Count()),
+		diagUtils.NewMeasureView(w.workflowExecutionLatency, []tag.Key{appIDKey, executionTypeKey, statusKey}, defaultLatencyDistribution),
+		diagUtils.NewMeasureView(w.workflowSchedulingLatency, []tag.Key{appIDKey, executionTypeKey}, defaultLatencyDistribution))
+}
+
+// TODO: Functions to handle metrics for:
+// - Operations
+// - Reminders
+// - Execution
+// - Scheduling
--- a/pkg/runtime/wfengine/workflow.go
+++ b/pkg/runtime/wfengine/workflow.go
@@ -62,6 +62,8 @@ type recoverableError struct {
 	cause error
 }
 
+// [Suggestion] we can elaborate more about durable timers ?
+
 func NewDurableTimer(bytes []byte, generation uint64) durableTimer {
 	return durableTimer{bytes, generation}
 }
@@ -266,6 +268,7 @@ func (wf *workflowActor) purgeWorkflowState(ctx context.Context, actorID string)
 	return nil
 }
 
+// [Question] What could be the use-case of calling this function ?
 func (wf *workflowActor) addWorkflowEvent(ctx context.Context, actorID string, historyEventBytes []byte) error {
 	state, err := wf.loadInternalState(ctx, actorID)
 	if err != nil {
@@ -290,6 +293,7 @@ func (wf *workflowActor) addWorkflowEvent(ctx context.Context, actorID string, h
 	return wf.saveInternalState(ctx, actorID, state)
 }
 
+// [Question] Does this invoke ? or schedules a reminder for the workflow ?
 func (wf *workflowActor) runWorkflow(ctx context.Context, actorID string, reminderName string, reminderData []byte) error {
 	state, err := wf.loadInternalState(ctx, actorID)
 	if err != nil {
@@ -387,11 +391,13 @@ func (wf *workflowActor) runWorkflow(ctx context.Context, actorID string, remind
 	case <-ctx.Done(): // caller is responsible for timeout management
 		return ctx.Err()
 	case completed := <-callback:
+		// [Question] Is this rescheduling the workflow to be completed ?
 		if !completed {
 			return newRecoverableError(errExecutionAborted)
 		}
 	}
 
+	// [Question] Not quite sure why this below logic is written.
 	// Increment the generation counter if the workflow used continue-as-new. Subsequent actions below
 	// will use this updated generation value for their duplication execution handling.
 	if runtimeState.ContinuedAsNew() {
@@ -498,6 +504,7 @@ func (wf *workflowActor) runWorkflow(ctx context.Context, actorID string, remind
 		}
 	}
 
+	// [Question] Is this workflow instance completed at this stage, since we are clearing the inbox ?
 	state.ApplyRuntimeStateChanges(runtimeState)
 	state.ClearInbox()
 
@@ -546,6 +553,7 @@ func (wf *workflowActor) saveInternalState(ctx context.Context, actorID string, 
 	return nil
 }
 
+// [Question] creating reminders is how we schedule workflows/Activities ?
 func (wf *workflowActor) createReliableReminder(ctx context.Context, actorID string, namePrefix string, data any, delay time.Duration) (string, error) {
 	// Reminders need to have unique names or else they may not fire in certain race conditions.
 	reminderName := fmt.Sprintf("%s-%s", namePrefix, uuid.NewString()[:8])
--- a/tests/apps/resiliencyapp/go.mod
+++ b/tests/apps/resiliencyapp/go.mod
@@ -1,6 +1,8 @@
 module github.com/dapr/dapr/tests/apps/resiliencyapp
 
-go 1.20
+go 1.21
+
+toolchain go1.21.3
 
 require (
 	github.com/dapr/dapr v0.0.0


prateek041:
I have mentioned a few `[Questions]` to solidify my understanding of the current code. The idea is to have a proper understanding of `workflow engine` and `workflows` and eventually write docs for future contributors.

cc @shubham1172

pkg/runtime/wfengine/workflow.go
```diff
@@ -266,6 +268,7 @@ func (wf *workflowActor) purgeWorkflowState(ctx context.Context, actorID string)
 	return nil
 }
 
+// [Question] What could be the use-case of calling this function ?
 func (wf *workflowActor) addWorkflowEvent(ctx context.Context, actorID string, historyEventBytes []byte) error {
```
shivamkm07, line 272: `addWorkflowEvent` is called to trigger further processing in workflow. The event could be like TerminateEvent, SuspendEvent, RaiseEvent etc. See this: https://github.com/dapr/dapr/tree/master/pkg/runtime/wfengine#resiliency.

pkg/runtime/wfengine/workflow.go
```diff
@@ -290,6 +293,7 @@ func (wf *workflowActor) addWorkflowEvent(ctx context.Context, actorID string, h
 	return wf.saveInternalState(ctx, actorID, state)
 }
 
+// [Question] Does this invoke ? or schedules a reminder for the workflow ?
```
shivamkm07, line 296: `runWorkflow` schedules the workflow(https://github.com/dapr/dapr/blob/HEAD/pkg/runtime/wfengine/workflow.go#L378-L379). The reminder would have been already invoked, which is what calls this method.

pkg/runtime/wfengine/workflow.go
```diff
@@ -387,11 +391,13 @@ func (wf *workflowActor) runWorkflow(ctx context.Context, actorID string, remind
 	case <-ctx.Done(): // caller is responsible for timeout management
 		return ctx.Err()
 	case completed := <-callback:
+		// [Question] Is this rescheduling the workflow to be completed ?
```
shivamkm07, line 394: If the workflow execution has completed successfully, then `true` is returned on the channel. In case of any errors, `false` is returned, and then recoverable error returned makes the engine retry it again(https://github.com/dapr/dapr/blob/HEAD/pkg/runtime/wfengine/workflow.go#L118-L119)

pkg/runtime/wfengine/workflow.go
```diff
@@ -387,11 +391,13 @@ func (wf *workflowActor) runWorkflow(ctx context.Context, actorID string, remind
...
 		if !completed {
 			return newRecoverableError(errExecutionAborted)
 		}
 	}
 
+	// [Question] Not quite sure why this below logic is written.
```
shivamkm07, line 400: The comment below explains the logic. when `continueAsNew` is invoked, then workflow resets it's history etc. and runs further. It's like a new workflow. So the previous `generation` shouldn't impact this one. e.g. Timers of previous generation don't get triggered in new one: https://github.com/dapr/dapr/blob/HEAD/pkg/runtime/wfengine/workflow.go#L309-L310

pkg/runtime/wfengine/workflow.go
```diff
@@ -498,6 +504,7 @@ func (wf *workflowActor) runWorkflow(ctx context.Context, actorID string, remind
 		}
 	}
 
+	// [Question] Is this workflow instance completed at this stage, since we are clearing the inbox ?
```
shivamkm07, line 507: yeah seems everything completed correctly and so clearing the inbox for this state.

pkg/runtime/wfengine/workflow.go
```diff
@@ -546,6 +553,7 @@ func (wf *workflowActor) saveInternalState(ctx context.Context, actorID string,
 	return nil
 }
 
+// [Question] creating reminders is how we schedule workflows/Activities ?
```
shivamkm07, line 556: yes!

shivamkm07:
> I have mentioned a few `[Questions]` to solidify my understanding of the current code. The idea is to have a proper understanding of `workflow engine` and `workflows` and eventually write docs for future contributors.
> 
> cc @shubham1172

Replied over the `Questions`. 
Btw PR won't be a good place to understand code logic, it might get cluttered. Feel free to ping me on Discord (same Id as github) and we can chat if you have any more queries!
And yes, Thanks for the contribution. Really appreciated!

pkg/runtime/wfengine/workflow.go
prateek041: This makes total sense now.

pkg/runtime/wfengine/workflow.go
prateek041: Oh, I get it now. Things are like this.

- `createReliableReminder` creates a Reminder.
- `invokeReminder` in Internal Actor Interface, is implemented by Activities and Workflow. 
- They call the `runWorkflow` and `executeActivity` methods, which are implementing the internal Actor interface.

Thanks for explaining this.

pkg/runtime/wfengine/workflow.go
prateek041: Understood, what will be the use-case for it though ?

prateek041:
Thank you for being patient with me @shivamkm07.

Sure, for any more conceptual doubts, I will ping you on discord.

prateek041:
@shivamkm07 please review.

Also, there are many needless comments which I added for extra context, will remove them when the PR is ready.

pkg/diagnostics/metrics.go
```diff
@@ -41,7 +41,8 @@ var (
 	// DefaultResiliencyMonitoring holds resiliency specific metrics.
 	DefaultResiliencyMonitoring = newResiliencyMetrics()
 	// Rules holds regex expressions for metrics labels
-	Rules map[string]string
+	DefaultWorkflowMonitoring = newWorkflowMetrics()
+	Rules                     map[string]string
```
shivamkm07, line 45: Seems this is not getting used anywhere. We can remove this.

pkg/diagnostics/workflow_metrics.go
```diff
@@ -0,0 +1,162 @@
...
+	reminderTypeKey = tag.MustNewKey("reminder_type")
+
+	executionTypeKey = tag.MustNewKey("execution_type")
+
+	isRetryableKey = tag.MustNewKey("is_retryable")
+)
+
+type workflowMetrics struct {
+	// Total successful workflow operation requests
+	workflowOperationsTotal *stats.Int64Measure
```
shivamkm07, line 36: Maybe rename to `workflowOperationsCount` similar to other metrics being defined(See component metrics for e.g.)

pkg/diagnostics/workflow_metrics.go
```diff
@@ -0,0 +1,162 @@
...
+	// Total failed workflow operation requests
+	workflowOperationsFailedTotal *stats.Int64Measure
```
shivamkm07, line 38: We can just add another key in `workflowOperationsCount` above to record Success/Failure Count. No need for another metric here. e.g. `processStatusKey` here: https://github.com/dapr/dapr/blob/474eab437329bccac9f5e37553413e103c46da84/pkg/diagnostics/component_monitoring.go#L173

Similar to this, we can add a `Status` Key with values as `Success`, `Retry` and `Failed`. That would also handle the usecase of `isRetryable` key being defined above

pkg/diagnostics/workflow_metrics.go
```diff
@@ -0,0 +1,162 @@
...
+	// Workflow operation request's Response latency
+	workflowOperationsLatency *stats.Float64Measure
+
+	// Total workflow/activity reminders created
+	workflowRemindersTotal *stats.Int64Measure
+
+	// Total successful workflow/activity executions
+	workflowExecutionTotal *stats.Int64Measure
+	// Total failed workflow/activity executions
+	workflowExecutionFailedTotal *stats.Int64Measure
```
shivamkm07, line 48: Same with this. Both Success and Failure can be combined into one

pkg/diagnostics/workflow_metrics.go
```diff
@@ -0,0 +1,162 @@
...
+	workflowSchedulingLatency *stats.Float64Measure
+
+	appID   string
+	enabled bool
+}
+
+func newWorkflowMetrics() *workflowMetrics {
+	return &workflowMetrics{
+		workflowOperationsTotal: stats.Int64(
+			"runtime/workflow/operations/total",
```
shivamkm07, line 62: Any reason behind adding `runtime` prefix? That doesn't seem to be adding much information

pkg/diagnostics/workflow_metrics.go
```diff
@@ -0,0 +1,162 @@
...
+	return w != nil && w.enabled
+}
+
+// Init registers the workflow metrics views.
+func (w *workflowMetrics) Init(appID string) error {
+	w.appID = appID
+	w.enabled = true
+
+	return view.Register(
+		diagUtils.NewMeasureView(w.workflowOperationsTotal, []tag.Key{appIDKey, operationKey}, view.Count()),
```
shivamkm07, line 107: Add namespace Key as well similar to other component metrics. Add a componentKey which would point to which component is being used(`dapr` for now). 

We also have a proposal to support different backends in Dapr(#7127). So adding a `Backendkey` would be useful as well(Currently the only supported value would be `ActorBackend`, but more would come later). 

And like said above a `Success` Key would help combining 2 metrics together

pkg/diagnostics/workflow_metrics.go
```diff
@@ -0,0 +1,162 @@
...
+		diagUtils.NewMeasureView(w.workflowOperationsFailedTotal, []tag.Key{appIDKey, operationKey}, view.Count()),
+		diagUtils.NewMeasureView(w.workflowOperationsLatency, []tag.Key{appIDKey, elapsedKey}, defaultLatencyDistribution),
```
shivamkm07, line 109: Similar to Count metrics above, the same Keys can be added here

pkg/diagnostics/workflow_metrics.go
```diff
@@ -0,0 +1,162 @@
...
+		diagUtils.NewMeasureView(w.workflowRemindersTotal, []tag.Key{appIDKey, reminderTypeKey}, view.Count()),
```
shivamkm07, line 110: Namespace Key can be added here as well maybe? (Although namespace info is there in reminderType as well, but still...)

pkg/runtime/wfengine/activity.go
```diff
@@ -100,6 +101,10 @@ func (a *activityActor) InvokeMethod(ctx context.Context, actorID string, method
 
 	// The actual execution is triggered by a reminder
 	err := a.createReliableReminder(ctx, actorID, nil)
+	if err == nil {
+		// Activity reminder created, record metrics
+		diag.DefaultWorkflowMonitoring.RemindersTotalCreated(ctx, "Activity")
```
shivamkm07, line 106: Don't use the string "Activity" directly. A constant can be defined in `pkg/diagnostics/workflow_metrics.go` with this value and that can be used here.

pkg/runtime/wfengine/activity.go
```diff
@@ -164,8 +169,12 @@ func (a *activityActor) executeActivity(ctx context.Context, actorID string, nam
 	wi.Properties[CallbackChannelProperty] = callback
 	if err = a.scheduler.ScheduleActivity(ctx, wi); err != nil {
 		if errors.Is(err, context.DeadlineExceeded) {
+			// Activity execution failed with recoverable error, record metrics.
+			diag.DefaultWorkflowMonitoring.ExecutionFailed(ctx, "Activity", true)
```
shivamkm07, line 173: Same here

pkg/runtime/wfengine/backend.go
```diff
@@ -108,8 +109,12 @@ func (be *actorBackend) CreateOrchestrationInstance(ctx context.Context, e *back
 
 	resp, err := be.actors.Call(ctx, req)
 	if err != nil {
+		// Failed request to create workflow, record metrics.
+		diag.DefaultWorkflowMonitoring.WorkflowOperationsFailed(ctx, "Create Workflow")
```
shivamkm07, line 113: Similarly define another constant in workflow_metrics and use that. And maybe `create_workflow` instead of `Create Workflow`?

pkg/runtime/wfengine/backend.go
```diff
@@ -125,8 +130,12 @@ func (be *actorBackend) GetOrchestrationMetadata(ctx context.Context, id api.Ins
 
 	res, err := be.actors.Call(ctx, req)
 	if err != nil {
+		// failed request to GET workflow Information, record metrics
+		diag.DefaultWorkflowMonitoring.WorkflowOperationsFailed(ctx, "Get Workflow")
```
shivamkm07, line 134: Same here

shivamkm07 (changes requested):

pkg/diagnostics/workflow_metrics.go
prateek041: yes, actually I tried to name them as suggested in the issue itself here

https://github.com/dapr/dapr/issues/7109

pkg/diagnostics/workflow_metrics.go
prateek041: Understood, also what would be the values for `namespace keys` ?

pkg/diagnostics/workflow_metrics.go
prateek041: Fixed

pkg/diagnostics/workflow_metrics.go
prateek041: fixed

pkg/diagnostics/workflow_metrics.go
prateek041: Combined into one.

pkg/runtime/wfengine/activity.go
prateek041: fixed

pkg/runtime/wfengine/activity.go
prateek041: fixed

pkg/runtime/wfengine/backend.go
prateek041: fixed

pkg/diagnostics/workflow_metrics.go
shivamkm07: The namespace can be added while initializing the metric. See this for example: https://github.com/dapr/dapr/blob/5050df5bf10835e9890b6b6519f5fa22e4736b28/pkg/diagnostics/metrics.go#L61

pkg/diagnostics/workflow_metrics.go
shivamkm07: Oh okay. Yeah we can keep it as well. No issues

pkg/diagnostics/workflow_metrics.go
```diff
@@ -0,0 +1,156 @@
...
+	Timer           = "timer"
+)
+
+type workflowMetrics struct {
+	// Total Successful and Failed workflow operation requests
+	workflowOperationsCount *stats.Int64Measure
+	// Workflow operation request's Response latency
+	workflowOperationsLatency *stats.Float64Measure
+	// Total workflow/activity reminders created
+	workflowRemindersTotal *stats.Int64Measure
```
shivamkm07, line 53:
```suggestion
	workflowRemindersCount *stats.Int64Measure
```

pkg/diagnostics/workflow_metrics.go
```diff
@@ -0,0 +1,156 @@
...
+}
+
+// Init registers the workflow metrics views.
+func (w *workflowMetrics) Init(appID string) error {
+	w.appID = appID
+	w.enabled = true
+
+	return view.Register(
+		diagUtils.NewMeasureView(w.workflowOperationsCount, []tag.Key{appIDKey, operationKey, statusKey}, view.Count()),
+		diagUtils.NewMeasureView(w.workflowOperationsLatency, []tag.Key{appIDKey}, defaultLatencyDistribution),
```
shivamkm07, line 105: Missing `operationKey` here?

pkg/runtime/wfengine/activity.go
```diff
@@ -189,8 +201,16 @@ loop:
 				<-t.C
 			}
 			if completed {
+				elapsed := diag.ElapsedSince(start)
+				// Activity completed, record metrics
+				diag.DefaultWorkflowMonitoring.ExecutionCount(ctx, diag.Activity, diag.StatusSuccess)
+
+				// record the time taken to complete Activity execution
+				diag.DefaultWorkflowMonitoring.ExecutionLatency(ctx, diag.Activity, diag.StatusSuccess, elapsed)
```
shivamkm07, line 209: You can combine both into a common Record function, similar to this: https://github.com/dapr/dapr/blob/5050df5bf10835e9890b6b6519f5fa22e4736b28/pkg/diagnostics/component_monitoring.go#L210.

pkg/runtime/wfengine/activity.go
```diff
@@ -189,8 +201,16 @@ loop:
...
 				break loop
 			} else {
+				// Activity execution failed with recoverable error, record metrics
+				diag.DefaultWorkflowMonitoring.ExecutionCount(ctx, diag.Activity, diag.StatusRetryable)
```
shivamkm07, line 213: Latency should ideally be recorded even in case of failures. e.g. If an activity completed half it's execution and errored out, it should be added in latency.

pkg/runtime/wfengine/backend.go
```diff
@@ -106,10 +108,21 @@ func (be *actorBackend) CreateOrchestrationInstance(ctx context.Context, e *back
...
 
+	start := time.Now()
 	resp, err := be.actors.Call(ctx, req)
+	elapsed := diag.ElapsedSince(start)
 	if err != nil {
+		// Failed request to create workflow, record metrics.
+		diag.DefaultWorkflowMonitoring.WorkflowOperationsCount(ctx, diag.CreateWorkflow, diag.StatusFailed)
+
+		// record latency metrics.
+		diag.DefaultWorkflowMonitoring.WorkflowOperationsLatency(ctx, diag.CreateWorkflow, diag.StatusFailed, elapsed)
```
shivamkm07, line 119: Similarly you can combine both into a common event function: https://github.com/dapr/dapr/blob/5050df5bf10835e9890b6b6519f5fa22e4736b28/pkg/diagnostics/component_monitoring.go#L210

pkg/runtime/wfengine/workflow.go
```diff
@@ -153,6 +159,7 @@ func (wf *workflowActor) DeactivateActor(ctx context.Context, actorID string) er
 	return nil
 }
 
+// NOTE: A workflow Instance is made here, check if it failed or not.
```
shivamkm07, line 162: Note to remove these `//NOTE: ` later on from the code

pkg/runtime/wfengine/workflow.go
```diff
@@ -375,21 +390,34 @@ func (wf *workflowActor) runWorkflow(ctx context.Context, actorID string, remind
...
+	start := time.Now()
 	err = wf.scheduler.ScheduleWorkflow(ctx, wi)
 	if err != nil {
 		if errors.Is(err, context.DeadlineExceeded) {
+			// Workflow execution scheduling request failed with recoverable error, record metrics.
+			diag.DefaultWorkflowMonitoring.ExecutionCount(ctx, diag.Workflow, diag.StatusRetryable)
 			return newRecoverableError(fmt.Errorf("timed-out trying to schedule a workflow execution - this can happen if there are too many in-flight workflows or if the workflow engine isn't running: %w", err))
 		}
+		// Workflow execution scheduling request failed with recoverable error, record metrics.
+		diag.DefaultWorkflowMonitoring.ExecutionCount(ctx, diag.Workflow, diag.StatusRetryable)
```
shivamkm07, line 403: If the scheduler has failed, then that would mean that workflow execution has not even started. It couldn't be scheduled only. So probably no need to record metrics in case scheduler returns error

pkg/runtime/wfengine/workflow.go
```diff
@@ -556,6 +590,7 @@ func (wf *workflowActor) createReliableReminder(ctx context.Context, actorID str
 		return reminderName, fmt.Errorf("failed to encode data as JSON: %w", err)
 	}
 
+	// Do I have to add some sort of time instance here, which can be checked in "run-workflow" function for latency calculations ?
```
shivamkm07, line 593: Which latency calculation?

pkg/diagnostics/workflow_metrics.go
prateek041: Sure.

pkg/runtime/wfengine/workflow.go
prateek041: I was wondering if the current method is correct for recording latency.

pkg/diagnostics/workflow_metrics.go
```diff
@@ -0,0 +1,145 @@
...
+			"runtime/workflow/operations/total",
+			"The number of successful/failed workflow operation requests.",
+			stats.UnitDimensionless),
+
+		workflowOperationsLatency: stats.Float64(
+			"runtime/workflow/operations/latency",
+			"The latencies of responses for workflow operation requests.",
+			stats.UnitMilliseconds),
+		workflowRemindersCount: stats.Int64(
+			"runtime/workflows/reminders/total",
```
RyanLettieri, line 76: Should we change "workflows" here and below to "workflow" (no 's') to maintain consistency with the other elements of the struct?

RyanLettieri:
One more question, will the metrics be correlated to each instance of a workflow/activity or do they only provide a generalized set of data for workflows and activities?

pkg/diagnostics/workflow_metrics.go
shivamkm07: Agree.

shivamkm07:
> One more question, will the metrics be correlated to each instance of a workflow/activity or do they only provide a generalized set of data for workflows and activities?

These are generalized metrics e.g. Total number of activities executed. Not correlated to a single instance of workflow/activity

pkg/diagnostics/workflow_metrics.go
shivamkm07: @DeepanshuA wdyt? Since workflow actors have the namespace in their name, I believe adding it might be useful. It can be added to all the workflow metrics ideally

pkg/diagnostics/workflow_metrics.go
```diff
@@ -0,0 +1,145 @@
...
+	// Workflow operation request's Response latency
+	workflowOperationsLatency *stats.Float64Measure
+	// Total workflow/activity reminders created
+	workflowRemindersCount *stats.Int64Measure
+	// Total successful/failed workflow/activity executions
+	workflowExecutionCount *stats.Int64Measure
+	// Time taken to run a workflow to completion
+	workflowExecutionLatency *stats.Float64Measure
+	// Latency between execution request and actual execution
+	workflowSchedulingLatency *stats.Float64Measure
```
shivamkm07, line 59: This we are not implementing in the current PR right?

shivamkm07: We can probably add this. When a workflow creation request comes up, a historyEvent is created and sent: https://github.com/dapr/dapr/blob/ed3417223c5ea4654f38df1fc5abee93e52cea6e/pkg/runtime/wfengine/workflow.go#L173 which contains the timeStamp. And when it is actually scheduled i.e. reminders gets invoked, the same historyEvent is there in workflow inbox: https://github.com/dapr/dapr/blob/ed3417223c5ea4654f38df1fc5abee93e52cea6e/pkg/runtime/wfengine/workflow.go#L332. We can use this to calculate the Scheduling latency. 

I saw it marked as requiring runtime changes in the issue: https://github.com/dapr/dapr/issues/7109. @shubham1172 Which runtime changes you meant?

shivamkm07: It would actually be the difference in creating a reminder and invoking that reminder. Idk if there is a better way to do that...

shubham1172: Hi @shivamkm07, I meant the runtime changes mentioned in the issue.
image

The idea was to use Actors reminder payload to calculate when the reminder was fired vs when it was executed.
We can alternatively explore the other way too, by looking into the `historyEvent`. Does it also work for activities along with workflow?

pkg/diagnostics/workflow_metrics.go
prateek041: fixed

go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/dapr/dapr
 
-go 1.20
+go 1.21
```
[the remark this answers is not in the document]
prateek041: I should rebase ?

prateek041:
Hey @shivamkm07 @RyanLettieri thanks for reviewing the PR.

I have added tests for the following metrics:

- [x] `runtime/workflow/operations/count`
- [x] `runtime/workflow/operations/latency`
- [x] `runtime/workflow/reminders/count`
- [x] `runtime/workflow/execution/count`
- [x] `runtime/workflow/execution/latency`
- [ ] `runtime/workflow/scheduling/latency`

I don't thing all those unit tests are needed but I tried to write most of the tests, feel free to point the ones that are not needed.

pkg/diagnostics/workflow_metrics.go
shivamkm07: That makes sense. HistoryEvent approach might not be the correct one as I pointed out above. We can explore this later ig. It should be out of scope of this PR then.

go.mod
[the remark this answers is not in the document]
shivamkm07: Better just merge and resolve the conflicts..

pkg/diagnostics/workflow_metrics.go
prateek041: sure. I will remove it for now.

pkg/runtime/wfengine/workflow.go
```diff
@@ -498,6 +521,9 @@ func (wf *workflowActor) runWorkflow(ctx context.Context, actorID string, remind
 		}
 	}
 
+	// record time taken to complete the workflow
+	elapsed := diag.ElapsedSince(start)
+	diag.DefaultWorkflowMonitoring.ExecutionEvent(ctx, "dapr", diag.Workflow, diag.StatusSuccess, elapsed)
```
shivamkm07, line 526: @prateek041 This is actually not the correct time for workflow completion. The time calculated here just measures the time for One `Schedule` of workflow. There would be different schedules and the workflow completion time should be the sum of all that. As mentioned here: https://github.com/dapr/dapr/issues/7109 It requires some runtime changes. So for now we can just pass `0` in elapsed time to not record and add a comment that workflow execution latency is not supported yet.

pkg/diagnostics/metrics.go
shivamkm07: @prateek041 Reminder to remove this

pkg/diagnostics/workflow_metrics.go
prateek041: added.

pkg/runtime/wfengine/workflow.go
prateek041: sure, fixed

pkg/diagnostics/workflow_metrics.go
prateek041: removed `workflowSchedulingLatency` metric.

pkg/runtime/wfengine/activity.go
```diff
@@ -184,6 +194,8 @@ loop:
 			if !t.Stop() {
 				<-t.C
 			}
+			// Activity execution failed with non-recoverable error. Record metrics
+			diag.DefaultWorkflowMonitoring.ExecutionEvent(ctx, "dapr", diag.Activity, diag.StatusFailed, 0)
```
shivamkm07, line 198: It's a retriable error as mentioned in the comment. Also add the time elapsed in the event

pkg/runtime/wfengine/activity.go
```diff
@@ -316,6 +334,7 @@ func (a *activityActor) createReliableReminder(ctx context.Context, actorID stri
 	if err != nil {
 		return fmt.Errorf("failed to encode data as JSON: %w", err)
 	}
+	// Similar to Workflow, should I add a time stamp here ? which can be checking in execute Activity to determin time taken for complete execution ?
```
shivamkm07, line 337: Remove this comment

pkg/runtime/wfengine/workflow.go
```diff
@@ -385,23 +393,33 @@ func (wf *workflowActor) runWorkflow(ctx context.Context, actorID string, remind
...
+			// Workflow execution scheduling request failed with recoverable error, record metrics.
+			diag.DefaultWorkflowMonitoring.ExecutionEvent(ctx, "dapr", diag.Workflow, diag.StatusRetryable, 0)
 			return newRecoverableError(fmt.Errorf("timed-out trying to schedule a workflow execution - this can happen if there are too many in-flight workflows or if the workflow engine isn't running: %w", err))
 		}
 		return newRecoverableError(fmt.Errorf("failed to schedule a workflow execution: %w", err))
 	}
 
 	select {
 	case <-ctx.Done(): // caller is responsible for timeout management
+		// Workflow execution failed with non-recoverable error, record metrics.
```
shivamkm07, line 411: This is retried as well. Mark it as recoverable

pkg/runtime/wfengine/workflow.go
```diff
@@ -385,23 +393,33 @@ func (wf *workflowActor) runWorkflow(ctx context.Context, actorID string, remind
...
+		diag.DefaultWorkflowMonitoring.ExecutionEvent(ctx, "dapr", diag.Workflow, diag.StatusFailed, 0)
 		return ctx.Err()
 	case completed := <-callback:
 		if !completed {
+			// Workflow execution failed with Recoverable Error, record metrics
+			diag.DefaultWorkflowMonitoring.ExecutionEvent(ctx, "dapr", diag.Workflow, diag.StatusRetryable, 0)
+
 			return newRecoverableError(errExecutionAborted)
 		}
+		// workflow execution completed, record metrics
```
shivamkm07, line 421: Better add a comment here: `execution latency for workflow is not supported yet`

pkg/runtime/wfengine/activity.go
```diff
@@ -168,11 +173,16 @@ func (a *activityActor) executeActivity(ctx context.Context, actorID string, nam
 	//       introduce some kind of heartbeat protocol to help identify such cases.
 	callback := make(chan bool)
 	wi.Properties[CallbackChannelProperty] = callback
+	start := time.Now()
 	wfLogger.Debugf("Activity actor '%s': scheduling activity '%s' for workflow with instanceId '%s'", actorID, name, wi.InstanceID)
 	if err = a.scheduler(ctx, wi); err != nil {
 		if errors.Is(err, context.DeadlineExceeded) {
+			// Activity execution failed with recoverable error, record metrics.
+			diag.DefaultWorkflowMonitoring.ExecutionEvent(ctx, "dapr", diag.Activity, diag.StatusRetryable, 0)
```
shivamkm07, line 181: Define another constant in workflow_metrics.go `ComponentName: "dapr"` and use that everywhere instead of using "dapr" string at all places for componentKey

pkg/runtime/wfengine/workflow.go
```diff
@@ -579,6 +601,7 @@ func (wf *workflowActor) createReliableReminder(ctx context.Context, actorID str
 		return reminderName, fmt.Errorf("failed to encode data as JSON: %w", err)
 	}
 
+	// Do I have to add some sort of time instance here, which can be checked in "run-workflow" function for latency calculations ?
```
shivamkm07, line 604: Remove this comment

pkg/runtime/wfengine/workflow.go
```diff
@@ -515,6 +535,8 @@ func (wf *workflowActor) runWorkflow(ctx context.Context, actorID string, remind
 		}
 	}
 
+	// Workflow Execution latency is not supported yet.
+	diag.DefaultWorkflowMonitoring.ExecutionEvent(ctx, "dapr", diag.Workflow, diag.StatusSuccess, 0)
```
shivamkm07, line 539: Successful workflow execution has already been recorded above. Remove this. Also add the comment line above at the time of successful execution event only.

pkg/diagnostics/workflow_metrics.go
```diff
@@ -0,0 +1,143 @@
...
```
shivamkm07, line 60: remove this

prateek041: I have removed the scheduling latency, since nothing from it is implemented, Execution latency on the other hand is being recorded in a few places, so leaving it as it is.

pkg/diagnostics/workflow_metrics_test.go
```diff
@@ -0,0 +1,326 @@
...
+
+import (
+	"context"
+	"github.com/stretchr/testify/assert"
+	"go.opencensus.io/stats/view"
+	"testing"
+)
+
+const (
+	workflowName = componentName
```
shivamkm07, line 11: Why use this const `WorkflowName` in the tests. You can simply use `componentName`. Also the name `workflowName` would be incorrect as it's actually the component name only, not the workflow name.

shivamkm07:
LGTM. just one nit suggested

pkg/diagnostics/workflow_metrics_test.go
prateek041: Sure, changing that.

shivamkm07 (approved):

pkg/runtime/wfengine/activity.go
```diff
@@ -168,11 +173,16 @@ func (a *activityActor) executeActivity(ctx context.Context, actorID string, nam
 	//       introduce some kind of heartbeat protocol to help identify such cases.
 	callback := make(chan bool)
 	wi.Properties[CallbackChannelProperty] = callback
+	start := time.Now()
 	wfLogger.Debugf("Activity actor '%s': scheduling activity '%s' for workflow with instanceId '%s'", actorID, name, wi.InstanceID)
 	if err = a.scheduler(ctx, wi); err != nil {
 		if errors.Is(err, context.DeadlineExceeded) {
+			// Activity execution failed with recoverable error, record metrics.
+			diag.DefaultWorkflowMonitoring.ExecutionEvent(ctx, diag.ComponentName, diag.Activity, diag.StatusRetryable, 0)
```
mukundansundar, line 181: change to elapsed ...

pkg/runtime/wfengine/activity.go
```diff
@@ -168,11 +173,16 @@ func (a *activityActor) executeActivity(ctx context.Context, actorID string, nam
...
 			return newRecoverableError(fmt.Errorf("timed-out trying to schedule an activity execution - this can happen if too many activities are running in parallel or if the workflow engine isn't running: %w", err))
 		}
+		// Activity execution failed with recoverable error, record metrics.
+		diag.DefaultWorkflowMonitoring.ExecutionEvent(ctx, diag.ComponentName, diag.Activity, diag.StatusRetryable, 0)
```
mukundansundar, line 185: same here.

pkg/runtime/wfengine/activity.go
```diff
@@ -184,6 +194,9 @@ loop:
 			if !t.Stop() {
 				<-t.C
 			}
+			// Activity execution failed with recoverable error. Record metrics
+			elapsed := diag.ElapsedSince(start)
```
mukundansundar, line 198: @shivamkm07 Please confirm that the activity execution completes even when a workflow is paused, just that a new activity in that workflow is not started when in pause state.

pkg/diagnostics/workflow_metrics.go
```diff
@@ -0,0 +1,143 @@
...
+	workflowExecutionLatency *stats.Float64Measure
+	appID                    string
+	enabled                  bool
+	namespace                string
+}
+
+func newWorkflowMetrics() *workflowMetrics {
+	return &workflowMetrics{
+		workflowOperationsCount: stats.Int64(
+			"runtime/workflow/operations/count",
```
mukundansundar, line 69: remove `operations` from the name to be similar to other metrics being reported.

mukundansundar (changes requested):
Please add the metrics being added here to https://github.com/dapr/dapr/blob/master/docs/development/dapr-metrics.md
file.

pkg/runtime/wfengine/activity.go
shivamkm07: @mukundansundar Confirmed that activity execution which has already started will complete even if a workflow is paused.

e.g. in Python-sdk, once workflow is paused, all future workflow events are buffered till the execution is resumed: https://github.com/microsoft/durabletask-python/blob/cbd07a8f2eae94ba82de798341d7c14f53b34693/durabletask/worker.py#L507
But this doesn't stop an activity from executing: https://github.com/microsoft/durabletask-python/blob/cbd07a8f2eae94ba82de798341d7c14f53b34693/durabletask/worker.py#L746

pkg/diagnostics/workflow_metrics.go
```diff
@@ -0,0 +1,143 @@
...
+	"go.opencensus.io/tag"
+)
+
+var (
+	elapsedKey = tag.MustNewKey("elapsed")
+
+	reminderTypeKey = tag.MustNewKey("reminder_type")
+
+	executionTypeKey = tag.MustNewKey("execution_type")
+)
```
ItalyPaleAle, line 31: Nit:


```suggestion
	elapsedKey = tag.MustNewKey("elapsed")
	reminderTypeKey = tag.MustNewKey("reminder_type")
	executionTypeKey = tag.MustNewKey("execution_type")
)
```

pkg/diagnostics/workflow_metrics.go
```diff
@@ -0,0 +1,143 @@
+/*
+Copyright 2021 The Dapr Authors
```
ItalyPaleAle, line 2: Nit
```suggestion
Copyright 2023 The Dapr Authors
```

pkg/diagnostics/workflow_metrics.go
```diff
@@ -0,0 +1,143 @@
...
+// Init registers the workflow metrics views.
+func (w *workflowMetrics) Init(appID, namespace string) error {
+	w.appID = appID
+	w.enabled = true
+	w.namespace = namespace
+
+	return view.Register(
+		diagUtils.NewMeasureView(w.workflowOperationsCount, []tag.Key{appIDKey, componentKey, namespaceKey, operationKey, statusKey}, view.Count()),
+		diagUtils.NewMeasureView(w.workflowOperationsLatency, []tag.Key{appIDKey, componentKey, namespaceKey, operationKey, statusKey}, defaultLatencyDistribution),
+		diagUtils.NewMeasureView(w.workflowRemindersCount, []tag.Key{appIDKey, componentKey, namespaceKey, reminderTypeKey}, view.Count()),
```
ItalyPaleAle, line 105: What is the reason for this metric? Isn't this linearly correlated to the number of operations (1 reminder per each op)? And could users just look at the actor metrics?

pkg/runtime/wfengine/activity.go
```diff
@@ -168,11 +173,16 @@ func (a *activityActor) executeActivity(ctx context.Context, actorID string, nam
...
 	wi.Properties[CallbackChannelProperty] = callback
+	start := time.Now()
 	wfLogger.Debugf("Activity actor '%s': scheduling activity '%s' for workflow with instanceId '%s'", actorID, name, wi.InstanceID)
 	if err = a.scheduler(ctx, wi); err != nil {
 		if errors.Is(err, context.DeadlineExceeded) {
+			// Activity execution failed with recoverable error, record metrics.
+			diag.DefaultWorkflowMonitoring.ExecutionEvent(ctx, diag.ComponentName, diag.Activity, diag.StatusRetryable, 0)
 			return newRecoverableError(fmt.Errorf("timed-out trying to schedule an activity execution - this can happen if too many activities are running in parallel or if the workflow engine isn't running: %w", err))
 		}
+		// Activity execution failed with recoverable error, record metrics.
```
ItalyPaleAle, line 184: This is the same metric so do it just once before the `errors.Is` on L179?

pkg/diagnostics/workflow_metrics.go
shivamkm07: `workflowRemindersCount` represents the number of internal reminders getting created. It is not linearly correlated to the number of operations, but instead to the reminderName prefixes for the Reminders: https://github.com/dapr/dapr/blob/9e2bcdc3d8a64b02466a9274e9259f166eedc95a/pkg/runtime/wfengine/workflow.go#L632
which basically corresponds to a set of reminders for same work. In total, there are 4 types:1) Workflow 2) Activity 3) Adding Event 4) Timers

It basically gives the total number of reminders getting created, which helps in correlating how the perf etc. is being impacted with remindersCount in workflow.

shubham1172 (changes requested):
Hello @prateek041, please also add the metrics to https://github.com/dapr/dapr/blob/master/docs/development/dapr-metrics.md

pkg/diagnostics/metrics.go
```diff
@@ -41,7 +41,7 @@ var (
 	// DefaultResiliencyMonitoring holds resiliency specific metrics.
 	DefaultResiliencyMonitoring = newResiliencyMetrics()
 	// Rules holds regex expressions for metrics labels
```
DeepanshuA, line 43: Please correct the comment here

pkg/diagnostics/workflow_metrics.go
```diff
@@ -0,0 +1,141 @@
...
+	Workflow        = "workflow"
+
+	WorkflowEvent = "event"
+	Timer         = "timer"
+
+	ComponentName = "dapr"
+)
+
+type workflowMetrics struct {
+	// Total Successful and Failed workflow operation requests
```
DeepanshuA, line 49: What all constitutes workflow operation?
Create/ get/ Delete etc. on workflow? If yes, can we please add to the comment here?

pkg/diagnostics/workflow_metrics.go
```diff
@@ -0,0 +1,141 @@
...
+	workflowOperationsCount *stats.Int64Measure
```
DeepanshuA, line 50: So, it's total number of requests.
Is there any metric for `failed_total` as well, just like mentioned in https://github.com/dapr/dapr/issues/7109#issue-1967760866

pkg/diagnostics/workflow_metrics.go
```diff
@@ -0,0 +1,141 @@
...
+	// Workflow operation request's Response latency
```
DeepanshuA, line 51: Not clear from the comment that whether it's addition of all latencies OR an average of latencies OR latency per operation

pkg/runtime/wfengine/activity.go
```diff
@@ -196,8 +207,14 @@ loop:
...
 			}
 			if completed {
+				elapsed := diag.ElapsedSince(start)
+				// activity completed, record count and latency metrics
+				diag.DefaultWorkflowMonitoring.ExecutionEvent(ctx, diag.ComponentName, diag.Activity, diag.StatusSuccess, elapsed)
 				break loop
 			} else {
+				elapsed := diag.ElapsedSince(start)
+				// Activity execution failed with recoverable error, record metrics
+				diag.DefaultWorkflowMonitoring.ExecutionEvent(ctx, diag.ComponentName, diag.Activity, diag.StatusRetryable, elapsed)
```
DeepanshuA, line 217: For my understanding: How do we conclude here that it is for `retry` and not for `fail`?

pkg/runtime/wfengine/activity.go
prateek041: I did not understand what you mean by this.

pkg/diagnostics/workflow_metrics.go
prateek041: Workflow Operations Requests consist of:
- Create workflow
- Get workflow
- Purge workflow
- Add Events
I have added the description of them, should I provide full description for every metric ?

pkg/diagnostics/workflow_metrics.go
prateek041: It is a collective count of requests created, if a request fails due to some reasons, a key-value pair of `status: failed` is attached to record. like [here](https://github.com/dapr/dapr/pull/7152/files#diff-9d8a4ceb7b0a3e0bb6d0ec2d8527ea7771d223ba416c55e727fc950b52b68f10R200)

pkg/diagnostics/workflow_metrics.go
prateek041: From what I understand, it is "latency per operation" because that is how it is being calculated, wdys @shivamkm07

pkg/runtime/wfengine/activity.go
prateek041: because we are returning a `recoverable error` which means Dapr will retry the activity execution.

pkg/diagnostics/metrics.go
prateek041: fixed that.

pkg/runtime/wfengine/activity.go
prateek041: should I resolve this conversation then ?

prateek041:
I have added the metrics to the `dapr-metrics.md` file. @mukundansundar @shubham1172

pkg/diagnostics/workflow_metrics.go
shivamkm07: Yeah, it's calculating latency per operation, very similar to latency being calculated for state operations: https://github.com/dapr/dapr/blob/9e2bcdc3d8a64b02466a9274e9259f166eedc95a/pkg/diagnostics/component_monitoring.go#L132

pkg/diagnostics/workflow_metrics.go
shivamkm07: Actually since we already have the reminders Metric: https://github.com/dapr/dapr/blob/9e2bcdc3d8a64b02466a9274e9259f166eedc95a/pkg/diagnostics/service_monitoring.go#L387, this metric doesn't seem to be adding much value. @prateek041 Please remove the reminders metric then...

prateek041: sure

prateek041:
Removed the Reminders Count metrics @shivamkm07

pkg/diagnostics/workflow_metrics.go
DeepanshuA: Should it be `workflowOperationLatency` then?

mukundansundar:
@shivamkm07 or @prateek041 Can the linter errors be fixed in the PR?

shivamkm07:
> @shivamkm07 or @prateek041 Can the linter errors be fixed in the PR?

Fixed

pkg/diagnostics/workflow_monitoring.go
```diff
@@ -0,0 +1,121 @@
...
+	GetWorkflow       = "get_workflow"
+	AddEvent          = "add_event"
+	PurgeWorkflow     = "purge_workflow"
+	Activity          = "activity"
+	Workflow          = "workflow"
+
+	WorkflowEvent = "event"
+	Timer         = "timer"
+
+	ComponentName = "dapr"
```
mukundansundar, line 42:
```suggestion
	WorkflowComponentName = "dapr"
```

pkg/runtime/wfengine/activity.go
```diff
@@ -175,7 +176,16 @@ func (a *activityActor) executeActivity(ctx context.Context, actorID string, nam
...
+	// Activity execution started
+	start := time.Now()
+	executionStatus := ""
+	elapsed := float64(0)
+	// Record metrics on exit
+	defer func() {
+		if executionStatus != "" {
+			diag.DefaultWorkflowMonitoring.ExecutionEvent(ctx, diag.ComponentName, diag.Activity, executionStatus, elapsed)
+		}
+	}()
```
mukundansundar, line 188: adding a test verifying the different execution status values on return would be good. It can be UT or IT.

pkg/runtime/wfengine/workflow.go
```diff
@@ -454,12 +466,23 @@ func (wf *workflowActor) runWorkflow(ctx context.Context, actorID string, remind
 		}
 		return newRecoverableError(fmt.Errorf("failed to schedule a workflow execution: %w", err))
 	}
```
mukundansundar, line 468: if a workflow fails to be scheduled, shouldn't the metric be recorded as execution failed? or do we have another metrics for schedule failed?

pkg/runtime/wfengine/workflow.go
```diff
@@ -578,7 +604,24 @@ func (wf *workflowActor) runWorkflow(ctx context.Context, actorID string, remind
 	state.ApplyRuntimeStateChanges(runtimeState)
 	state.ClearInbox()
 
-	return wf.saveInternalState(ctx, actorID, state)
+	err = wf.saveInternalState(ctx, actorID, state)
+	if err != nil {
+		return err
+	}
```
mukundansundar, line 610: in this flow, if the internal state is not saved correctly, then the status is non-recoverable failure in terms of metrics, correct?

pkg/runtime/wfengine/workflow.go
```diff
@@ -578,7 +604,24 @@ func (wf *workflowActor) runWorkflow(ctx context.Context, actorID string, remind
...
+	if executionStatus != "" {
```
mukundansundar, line 611: isn't the only possible option here is that executionStatus is diag.StatusFailed here? Can we check on that? Or do we have other possible values which are then being reset in line below when it is set back to empty string?

pkg/diagnostics/workflow_monitoring.go
```diff
@@ -0,0 +1,121 @@
...
+func (w *workflowMetrics) Init(appID, namespace string) error {
+	w.appID = appID
+	w.enabled = true
+	w.namespace = namespace
+
+	return view.Register(
+		diagUtils.NewMeasureView(w.workflowOperationCount, []tag.Key{appIDKey, componentKey, namespaceKey, operationKey, statusKey}, view.Count()),
+		diagUtils.NewMeasureView(w.workflowOperationLatency, []tag.Key{appIDKey, componentKey, namespaceKey, operationKey, statusKey}, defaultLatencyDistribution),
+		diagUtils.NewMeasureView(w.workflowExecutionCount, []tag.Key{appIDKey, componentKey, namespaceKey, executionTypeKey, statusKey}, view.Count()),
+		diagUtils.NewMeasureView(w.workflowExecutionLatency, []tag.Key{appIDKey, componentKey, namespaceKey, executionTypeKey, statusKey}, defaultLatencyDistribution))
```
DeepanshuA, line 94: I think we should have workflowName and activityName as well in this metric; else a common plotting won't make much sense - as different activities mayhave different expected latencies.

pkg/diagnostics/workflow_monitoring.go
```diff
@@ -0,0 +1,121 @@
...
+		workflowOperationLatency: stats.Float64(
+			"runtime/workflow/operation/latency",
+			"The latencies of responses for workflow operation requests.",
+			stats.UnitMilliseconds),
+		workflowExecutionCount: stats.Int64(
+			"runtime/workflow/execution/count",
+			"The number of successful/failed/recoverable workflow/activity executions.",
+			stats.UnitDimensionless),
+		workflowExecutionLatency: stats.Float64(
+			"runtime/workflow/execution/latency",
```
DeepanshuA, line 74: Should it be `runtime/workflow/execution/latency` and `runtime/workflow/activity/execution/latency` - differentiation would be there - and cardinality can also be reduced by removing executionTypeKey.

DeepanshuA: Similarly, for workflow/execution/count metric as well

DeepanshuA (commented):
Sorry, for these comments at this time of this PR; but had just noticed these points.

pkg/runtime/wfengine/workflow.go
shivamkm07: Correct

pkg/runtime/wfengine/workflow.go
shivamkm07: The metric being added here is for execution. For scheduling, another metric will be added later on: https://github.com/dapr/dapr/issues/7109#issuecomment-1868047411

pkg/diagnostics/workflow_monitoring.go
shivamkm07: Sounds good to me

pkg/diagnostics/workflow_monitoring.go
shivamkm07: Sure

pkg/runtime/wfengine/workflow.go
shivamkm07: No, executionStatus can be empty as well, which happens when workflow has already completed beforehand: https://github.com/dapr/dapr/pull/7152/files#diff-9997677e8aa2146ca3f09e604ca0f932a17a85fddc758fcc785c7ebb9afbc08aR457

for e.g. If a terminate signal is sent to a workflow and it has already completed, we don't want to re-record the metrics and so executionStatus would be empty here and so it won't be overwritten in the logic below and therefore in the defer statement, the metrics won't be recorded.

pkg/diagnostics/workflow_monitoring.go
shivamkm07: Sounds good to me.

pkg/diagnostics/workflow_monitoring.go
shivamkm07: Done

pkg/runtime/wfengine/activity.go
shivamkm07: The general UTs are there as similar to other metrics. IT for workflow is not there yet, more tests can be added there as required. Can that be a separate item to add IT for workflow?

shivamkm07: Added ITs as well

DeepanshuA (approved):

docs/development/dapr-metrics.md
```diff
@@ -101,6 +101,16 @@ Dapr uses prometheus process and go collectors by default.
...
 * dapr_resiliency_activations_total: Number of times a resiliency policy has been activated in a building block after a failure or after a state change.
 
+#### Workflow metrics
+
+[workflow metrics](../../pkg/diagnostics/workflow_monitoring.go)
+
+* dapr_runtime_workflow_operation_count: The number of successful/failed workflow operation requests.
+* dapr_runtime_workflow_operation_latency: The latencies of responses for workflow operation requests.
+* dapr_runtime_workflow_execution_count: The number of successful/failed/recoverable workflow/activity executions.
+* dapr_runtime_workflow_execution_latency: The total time taken to run a workflow/activity to completion.
```
mukundansundar, line 111: nitpick: change this metric name

shivamkm07: Done

mukundansundar (dismissed):

mukundansundar (approved):

PR #7153, opened 2023-11-06
Helm: add options to configure the Service resources

ItalyPaleAle:
Adds options to configure, for each Service that is created:

- The type (`ClusterIP`, the default, `LoadBalancer`, etc) _(not for placement, which doesn't allocate any IP for the service)_
- Custom annotations - there was already an option to set annotations on deployments, but not on services

Default values don't change, but just gives a few more options for users who have special environments.


daixiang0 (approved):
LGTM

shubham1172 (approved):

mukundansundar (approved):

PR #7154, opened 2023-11-07
Operator: do not patch the Subscription CRDs if the conversion webhooks are disabled

ItalyPaleAle:
When running the operator with `ENABLE_WEBHOOKS=false`, which disables listening for the conversion webhooks, the operator should not be patching the CRDs


ItalyPaleAle:
> Do we have a test to cover this case?

We do not, and we do not have the infrastructure for creating tests for this (we do not have the ability to test different Helm options in our CI, for example)

daixiang0 (approved):

yaron2 (approved):

PR #7167, opened 2023-11-09
Fix placement and CallLocalStream unit test

ItalyPaleAle:
The placement test is failing almost 100% of times lately.

The CallLocalStream test is another one that fails frequently

### DIFF
--- a/pkg/actors/placement/placement_test.go
+++ b/pkg/actors/placement/placement_test.go
@@ -15,6 +15,7 @@ package placement
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"io"
 	"net"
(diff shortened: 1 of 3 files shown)


pkg/actors/placement/placement_test.go
```diff
@@ -425,17 +425,15 @@ func (s *testServer) ReportDaprStatus(srv placementv1pb.Placement_ReportDaprStat
 			return status.Error(codes.FailedPrecondition, "only leader can serve the request")
 		}
 
-		req, err := srv.Recv()
-		if err == io.EOF {
```
ItalyPaleAle, line 429: This was the core of the issue. Sometimes in the test server (only in test code) the shutdown was reported as a gRPC "Context canceled" error, rather than just EOF. This seems to be a bug in the test code only, as the app code handles this case already.

mukundansundar (approved):

Issue #5141, opened 2022-09-07
sidecar injector - GetAppID does not work as expected


famarting:
## In what area(s)?

/area k8s-injector

## What version of Dapr?

> 1.8.x

## Expected Behavior

The function used by the sidecar injector to get the app id behaves as expected
https://github.com/dapr/dapr/blob/master/pkg/injector/sidecar/utils.go#L30

## Actual Behavior
The function used by the sidecar injector to get the app id it's coded to pick the pod's name as the default app id but in reality that's impossible and will never occur making the sidecar injector code confusing.

Because that logic runs when creating a pod, the pod name is not available, because it's not defined yet, however the function in the sidecar injector to get the app id fallback to the pod name for choosing an app id, in reality it will never be possible that the pod name is used as the dapr app id.

## Steps to Reproduce the Problem

- Create an application in k8s without the annotation "dapr.io/app-id"
- Check the logs for the sidecar injector

```
{"instance":"dapr-sidecar-injector-57979f5f95-rk78n","level":"info","msg":"AdmissionReview for Kind=/v1, Kind=Pod, Namespace=crud-app Name= () UID=92038e80-f46d-48e8-b532-74cf96f12f14 patchOperation=CREATE UserInfo={system:serviceaccount:kube-system:replicaset-controller f47b6ec6-8498-4784-aa42-0be8b128696e [system:serviceaccounts system:serviceaccounts:kube-system system:authenticated] map[]}","scope":"dapr.injector","time":"2022-09-07T05:52:16.332681919Z","type":"log","ver":"1.8.2"}
{"instance":"dapr-sidecar-injector-57979f5f95-rk78n","level":"error","msg":"Sidecar injector failed to inject for app ''. Error: value for the dapr.io/app-id annotation is empty","scope":"dapr.injector","time":"2022-09-07T05:52:16.333064633Z","type":"log","ver":"1.8.2"}
{"instance":"dapr-sidecar-injector-57979f5f95-rk78n","level":"error","msg":"Admission succeeded, but pod was not patched. No sidecar injected for ''","scope":"dapr.injector","time":"2022-09-07T05:52:16.333197325Z","type":"log","ver":"1.8.2"}
```

## Release Note

RELEASE NOTE:

madhavsikka:
Hi @artursouza. Does defaulting it to an empty string sound reasonable?

artursouza:
First, it cannot default to pod's name. That logic should be gone.

Then, we do the following:
[...]

PR #7177, opened 2023-11-13
logging: sidecar injector - GetAppID

sadath-12:
# Description

sidecar injector - GetAppID will not default to any other name if not set and will fail the injection if not appID not set

## Issue reference

Please reference the issue this PR will close: #5141 
## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Unit tests passing

### DIFF
--- a/pkg/injector/patcher/sidecar_container.go
+++ b/pkg/injector/patcher/sidecar_container.go
@@ -38,6 +38,9 @@ type getSidecarContainerOpts struct {
 
 // getSidecarContainer returns the Container object for the sidecar.
 func (c *SidecarConfig) getSidecarContainer(opts getSidecarContainerOpts) (*corev1.Container, error) {
+	if c.GetAppID() == "" {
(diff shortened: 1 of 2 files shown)


artursouza (dismissed):

ItalyPaleAle (commented):
Is this a breaking change?

yaron2:
> Is this a breaking change?

Yes. @sadath-12 we actually want to allow users to default a pod name if an ID isn't specified and I know at least one company using Dapr like that. Could you please change this PR to only log a warning that an app-id isn't set and that a pod name is used instead?

artursouza:
Can this PR also add a end-to-end or functional test scenario to catch the breaking change? Unit-test usually don't catch those.

sadath-12:
> Can this PR also add a end-to-end or functional test scenario to catch the breaking change? Unit-test usually don't catch those.

Not making a breaking change

Issue #7182, opened 2023-11-14
Add ability to add custom environment variable to dapr control plane


filintod:
## In what area(s)?

> /area runtime

Need a way to specify soft memory limit (ie GOMEMLIMIT) to dapr system. This comes handy when setting memory limit on container so the Go runtime can try to release memory before OOMKilled happens.  Another options would be to add it in code https://pkg.go.dev/runtime/debug#SetMemoryLimit but this way is configurable by the customer directly as adding it in code will still require some kind of user interaction (ie set it at 90% or 95%, etc).

## What version of Dapr?

> 1.1.x
> 1.0.x
> edge: output of `git describe --dirty`

## Expected Behavior

I should be able to say, in the corresponding chart:
```
extraEnvVars:
- name: GOMEMLIMIT
  value: 900MiB
```

and this is set to the corresponding deployment/sts for the container in the chart.

## Actual Behavior

There is no option for this

## Release Note

RELEASE NOTE:

PR #7183, opened 2023-11-14
Add extra env vars to charts

filintod:
# Description

Add optional extra env vars list to allow adding custom environment variables to dapr control plane system.  
## Issue reference 

Please reference the issue this PR will close: #7182 

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing


charts/dapr/charts/dapr_operator/values.yaml
```diff
@@ -35,6 +35,8 @@ ports:
 
 resources: {}
 
+extraEnvVars: []
```
ItalyPaleAle, line 38: Make this a dictionary of key->value

ItalyPaleAle (changes requested):
Please update the Helm chart README too

charts/dapr/charts/dapr_operator/values.yaml
filintod: usually this is an array, at least in bitnami charts. I can see the benefit of using a dictionary, and thought about it, but that also includes adding a range to parse it into an array of dictionary in the template. I did not see env vars added in a chart in dapr that I looked for to follow so decided to follow bitnami's standard that is easier to implement.

I mentioned the format here https://github.com/dapr/dapr/issues/7182

ItalyPaleAle: We have other fields in the chart that allow setting annotations and are dictionaries.

Using an array allows doing things like referencing secrets. However secrets can't be mounted in the current Helm chart so it doesn't matter in this case.

Using a dictionary is a lot simpler to pass via `--set` in Helm too.

My .02.

filintod: annotations are dictionaries by default so it kind of make sense. true about `--set` but now helm also has `--set-json` that makes life easier for most of previous hard cases

filintod: I could add the dictionary. let me think of a way to not make it to bad in the template

filintod: @ItalyPaleAle could you do another check

filintod:
@ItalyPaleAle could you take another look please

artursouza:
cc @ItalyPaleAle

Issue #7180, opened 2023-11-14
fail to run make setup-test-components due to crd version conversion not work


lucus-sun:
## In what area(s)?

 /area docs

 /area test-and-release

## Ask your question here

When setup the e2e environment following [this doc](https://github.com/dapr/dapr/blob/master/tests/docs/running-e2e-test.md) by executing the command:

```
make setup-test-components
```

It reports error:

```
kubectl apply -f ./tests/config/app_topic_subscription_pluggable_pubsub.yaml --namespace dapr-tests
Error from server: error when creating "./tests/config/app_topic_subscription_pluggable_pubsub.yaml": conversion webhook for dapr.io/v1alpha1, Kind=Subscription failed: Post "https://dapr-webhook.replaceme.svc:443/convert?timeout=30s": service "dapr-webhook" not found
```

It seems we're using the v1alpha1 subscription in `./tests/config/app_topic_subscription_pluggable_pubsub.yaml`:

```
apiVersion: dapr.io/v1alpha1
kind: Subscription
metadata:
  name: pluggable-c-topic-subscription
spec:
  pubsubname: pluggable-messagebus
  topic: pubsub-c-topic-http
  route: /pubsub-c-topic-http
scopes:
  - pubsub-subscriber-pluggable
```

It needs version conversion but we don't have the dapr service installed yet in the e2e environment preparation.

PR #7187, opened 2023-11-15
Fix subscription version to v2alpha1 format

lucus-sun:
# Description

Fixes https://github.com/dapr/dapr/issues/7180

## Issue reference

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing

### DIFF
--- a/tests/config/app_topic_subscription_pluggable_pubsub.yaml
+++ b/tests/config/app_topic_subscription_pluggable_pubsub.yaml
@@ -11,13 +11,14 @@
 # limitations under the License.
 #
 
-apiVersion: dapr.io/v1alpha1
+apiVersion: dapr.io/v2alpha1
 kind: Subscription
 metadata:
   name: pluggable-c-topic-subscription
 spec:
   pubsubname: pluggable-messagebus
   topic: pubsub-c-topic-http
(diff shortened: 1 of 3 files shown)


tests/config/app_topic_subscription_pluggable_pubsub.yaml
```diff
@@ -11,13 +11,14 @@
 # limitations under the License.
 #
 
-apiVersion: dapr.io/v1alpha1
```
ItalyPaleAle, line 14: It’s important to test with at least one on version v1alpha1 so we test the upgrading. If you change them all, we are removing testing some functionality.

the reason it fails for you is generally due to not having the CRDs installed or not having dapr running correctly on the cluster.

ItalyPaleAle (changes requested):

tests/config/app_topic_subscription_pluggable_pubsub.yaml
lucus-sun: Thanks @ItalyPaleAle , I realized you're right that is because I didn't have dapr runtime running before the e2e.

PR #7195, opened 2023-11-15
Enforce Component names to be unique

JoshVanL:
Currently in standalone mode, it is possible to load multiple components with the same name but of different types. This is problematic as it can produce undefined behaviour in Daprd where global state indexed on Component name (like state encryption) can "leak" to another Component. It also makes it impossible to implement the hot reloader reconciler as that controller is level based, and without a single unique name to index on, the reconciler cannot determine which Components have been deleted/editted. It is also confusing to different requirements to Kubernetes whether a Component name _must_ be unique within a namespace.

PR implements a check to ensure that Component names are unique. The component store implements a pending & lock mechanism to ensure that another Component of the same name cannot be added whilst another is initialising.

Adds an integration test to ensure daprd errors when multiple Components with the same name are added.


pkg/runtime/compstore/components.go
```diff
@@ -13,31 +13,73 @@ limitations under the License.
...
-		if comp.Spec.Type == componentType && comp.ObjectMeta.Name == name {
+		if comp.ObjectMeta.Name == name {
 			return c.components[i], true
 		}
 	}
 	return compsv1alpha1.Component{}, false
 }
 
-func (c *ComponentStore) AddComponent(component compsv1alpha1.Component) {
+func (c *ComponentStore) AddPendingComponentForCommit(component compsv1alpha1.Component) error {
```
daixiang0, line 34: Why need a pending queue as we add component one by one?

JoshVanL: I don't think it is necessarily true when we introduce hot reloading. Adding the name as pending means we can reserve name before committing when `Init` is successful.

yaron2: With hot reloading components can be updated in the sidecar in parallel, as opposed to being initialized sequentially when loaded at startup.

berndverst (approved):

Issue #7193, opened 2023-11-15
Placement server stops working completely "failed to invoke target x after 3 retries" requires manual intervention


fabistb:
## In what area(s)?

> /area runtime

> /area operator

 /area placement

> /area docs

> /area test-and-release

## What version of Dapr?

1.11.x
1.12.x
> edge: output of `git describe --dirty`

## Expected Behavior
When running a high amount of operations like deployment and restarting many services at the same time the placement server keeps the ip addresses of the applications up to date.

## Actual Behavior
The placement server runs out of sync and tries to call ip addresses which don't exist anymore.
It also looks like that the service doesn't automatically recover and the placement server needs to be scaled to 0 and scaled up again.

Behavior appears when the placement server is configured using the in memory mode.

**Example:**
Ip of service before deployment / rollout restart:
`random-application  2/2   Running  192.168.26.105`

Error message application insights:
![image](https://github.com/dapr/dapr/assets/63794579/099ffef4-4f66-4567-b3d3-102e82c19ae5)

Ip of services after deployment / rollout restart:
`random-application  2/2  Running  192.168.26.113`

**Placement Server Log:**
There is one log in the placement server which I haven't seen before however a couple of times after the error appeared.

`level=error msg="Stream is disconnected before member is added rpc error: code = Canceled desc = context canceled" instance=dapr-placement-server-1 scope=dapr.placement type=log ver=1.12.0`

## Steps to Reproduce the Problem
I haven't found a reliably way to reproduce the issue. 
Rollout restarting all applications from time to time causes this behavior in a semi forced way.

## Release Note

RELEASE NOTE:

fabistb:
@yaron2 , @artursouza do you spontaneously have any idea here?
I see this sometimes multiple times a day which is really unfortunate.

ItalyPaleAle:
> Running the 1.12.3-rc.1 version unfortunately it hasn't resolved the issue on our side.

Same error? Or something different?

fabistb:
> > Running the 1.12.3-rc.1 version unfortunately it hasn't resolved the issue on our side.
> 
> Same error? Or something different?

Some one unfortunately.

fabistb:
> @fabistb also on the Placement service or just in the sidecars?

@ItalyPaleAle  sorry mainly focused on the application / sidecar.

I also have to alter this statement from myself:
[...]

PR #7197, opened 2023-11-15
Fix dissemination of placement tables

ItalyPaleAle:
Fixes (hopefully) #7193

Fixes two issues in the process of disseminating placement tables.

First, dissemination should always have a background context, and not be tied to the context of a specific stream. Otherwise, if the stream is disconnected before the dissemination, all connected hosts may be left in a broken state.

Second, the dissemination process was meant to follow a 3-step process: acquire a lock, disseminate the tables, then release the lock. The current implementation performed all 3 operations in parallel, effectively not waiting for all nodes to be locked before disseminating the tables. This means that the nodes could potentially have inconsistent tables.


artursouza:
LGTM overall. Waiting on the tests to pass.

artursouza:
I assume this affects 1.12 - if so, can this be merged to 1.12 instead?

mukundansundar:
> I assume this affects 1.12 - if so, can this be merged to 1.12 instead?

@ItalyPaleAle Is this being merged into 1.12 to be released as a hot fix?

ItalyPaleAle:
> @ItalyPaleAle Is this being merged into 1.12 to be released as a hot fix?

This is up to you to decide. Can also be cherry-picked from master as another option.

daixiang0 (approved):
LGTM

Issue #6992, opened 2023-10-03
Delete Actor Support


gverstraete:
## In what area(s)?
/area runtime

## Describe the feature
There idea is to have a "Delete Actor" functionality.

In our use case we have long running actors (multiple years) in a "digital twin" methodology.
At some point time the digital twin is no longer valid and needs to be cleaned up.

I'm sure there are many other use cases as well.

We now go under the asssumption that we remove all the state & reminders/timers on a custom implementation per actor type and by that the actor is removed.
In reality the actor still lives some time in memory, but that is a side effect which is acceptable.

This idea is here to have one generic way to handle Actor deletion with all of it's aspects by just one operation..

## Release Note

RELEASE NOTE:

alicejgibbons:
Note: this is more than just the actor TTL as it would be an API call to explicitly delete the actor state, as well as any reminders and timers associated with it.

papabytes:
This approach looks great for our scenario. Any thoughts @gverstraete ?

papabytes:
Sorry to jump in, but would it be possible to add a flag or a delete endpoint based on the actor type where we could remove (from the placement server only) all the instances of an Actor Type?

This may come in handy for us when we have a cumulative number of communication exceptions between our apps and the sidecar hosting a certain actor and it would be great if we could have a "manual reset button" to get a clear state on the actors and where they are running

gverstraete:
This has been planned, so should not be closed right? We would still like this functionality

alicejgibbons:
To be *partially* completed in https://github.com/dapr/cli/pull/1543

WhitWaldo:
> To be completed in [dapr/cli#1543](https://github.com/dapr/cli/pull/1543)

Would this include removal of actor state as well? The linked issue seems to suggest it just removes data from the scheduler.

PR #7223, opened 2023-11-21
Adding in API to support deleting of actor state

RyanLettieri:
# Description

Added in new API for deleting actor state

## Issue reference

Please reference the issue this PR will close: #_[[6992](https://github.com/dapr/dapr/issues/6992)]_

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing


ItalyPaleAle:
Necessary clarification: this PR deletes actor **state**, not actors per se

dapr/proto/runtime/v1/dapr.proto
```diff
@@ -82,6 +82,9 @@ service Dapr {
   // Gets the state for a specific actor.
   rpc GetActorState(GetActorStateRequest) returns (GetActorStateResponse) {}
 
+  // Gets the state for a specific actor.
+  rpc DeleteActorState(DeleteActorStateRequest) returns (google.protobuf.Empty) {}
```
ItalyPaleAle, line 86: A bit of a nit, but for good hygiene consider creating a response type `DeleteActorStateResponse` which is empty, instead of `google.protobuf.Empty`. In the future there may be the need to add properties, and creating a custom type already makes this easier (incl. on SDKs)

pkg/grpc/universalapi/api_actors.go
```diff
@@ -47,3 +53,51 @@ func (a *UniversalAPI) WaitForActorsReady(ctx context.Context) {
 	case <-a.actorsReadyCh:
 	}
 }
+
+// This function makes sure that the actor subsystem is ready.
+func (a *UniversalAPI) actorReadinessCheck(ctx context.Context) error {
```
ItalyPaleAle, line 58: Looks like this was copy-pasted from pkg/grpc/api.go... Please don't copy/paste code as that makes it harder to maintain it.

Instead, make this method exported, and then change pkg/grpc to use "actorReadinessCheck" from the Universal property

pkg/grpc/universalapi/api_actors.go
```diff
@@ -47,3 +53,51 @@ func (a *UniversalAPI) WaitForActorsReady(ctx context.Context) {
...
+
+	return nil
+}
+
+func (a *UniversalAPI) DeleteActorState(ctx context.Context, in *runtimev1pb.DeleteActorStateRequest) (*emptypb.Empty, error) {
+	if err := a.actorReadinessCheck(ctx); err != nil {
+		return &emptypb.Empty{}, err
+	}
+
+	actorType := in.ActorType
```
ItalyPaleAle, line 74: Nit: No need to copy these in a local variable, since they are not changed. Just use `in.ActorType` in all cases below. (Same for `in.ActorId` and `in.Key`)

pkg/grpc/universalapi/api_actors.go
```diff
@@ -47,3 +53,51 @@ func (a *UniversalAPI) WaitForActorsReady(ctx context.Context) {
...
+	actorID := in.ActorId
+	key := in.Key
+
+	hosted := a.Actors.IsActorHosted(ctx, &actors.ActorHostedRequest{
```
ItalyPaleAle, line 78: This may need further discussion. By doing this check, you're essentially making it so only an actor can delete its own state? It may be what we want, but then it requires an actor to implement the logic to clean up after itself.

I looked at the issue that was linked and couldn't find details on how this API was meant to be used. Deciding on how this API will be used may be key first

pkg/grpc/universalapi/api_actors.go
```diff
@@ -47,3 +53,51 @@ func (a *UniversalAPI) WaitForActorsReady(ctx context.Context) {
...
+		return &emptypb.Empty{}, err
+	}
+
+	req := actors.DeleteStateRequest{
+		ActorType: actorType,
+		ActorID:   actorID,
+		Key:       key,
+	}
+
+	err := a.Actors.DeleteState(ctx, &req)
```
ItalyPaleAle, line 95: Nit :)


```suggestion
	err := a.Actors.DeleteState(ctx, &actors.DeleteStateRequest{
		ActorType: actorType,
		ActorID:   actorID,
		Key:       key,
	})
```

pkg/http/api_actors.go
```diff
@@ -0,0 +1,174 @@
...
+func (a *api) onDeleteActorStateHandler() http.HandlerFunc {
+	return UniversalHTTPHandler(
+		a.universal.DeleteActorState,
+		UniversalHTTPHandlerOpts[*runtimev1pb.DeleteActorStateRequest, *emptypb.Empty]{
+			InModifier: func(r *http.Request, in *runtimev1pb.DeleteActorStateRequest) (*runtimev1pb.DeleteActorStateRequest, error) {
+				in.ActorId = chi.URLParam(r, actorIDParam)
+				in.ActorType = chi.URLParam(r, actorTypeParam)
+				return in, nil
+			},
+			SuccessStatusCode: http.StatusAccepted,
```
ItalyPaleAle, line 172: The "accepted" status code is normally used to indicate that an operation has been accepted and will be executed in background. Here, the operation is already executed synchronously.

Suggest using either 204 (No Content) or 200 (OK)
```suggestion
			SuccessStatusCode: http.StatusNoContent,
```

go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/dapr/dapr
 
-go 1.20
+go 1.21
```
ItalyPaleAle, line 3: Please don't update to Go 1.21 in this PR :) That will need to be done eventually, but let's do that in a separate PR!

ItalyPaleAle (changes requested):

ItalyPaleAle:
I've been thinking a bit more about this PR.

I believe the intent of #6992 is to have a one-stop call to purge all state for an actor, without having to track which keys are in use.

The way this is implemented in this PR, it requires users to pass each key they want to delete. In this sense, this change doesn't really add much value compared to using the existing methods to perform a state store transaction, which already allow deleting keys (and do that more efficiently!). As a consequence, I don't believe this is really the fix for #6992.

Our state store APIs don't have a "list" method (and it can't be implemented on _all_ actor state stores), so we don't have a way to know all keys that are to be deleted. The only way this could be fixed would be to keep track of all keys in the state store that an actor uses, but that is problematic too as it requires maintaining a separate index (which at the very least has a negative perf impact).

ItalyPaleAle:
More thoughts. Here's how we can implement this.

Premises:

- This is possible because actor state always has a prefix that indicates the actor that created it, which is: `app-id||actor-type||actor-id||key`
- This method will not work on all state stores that support actors. It will definitely work on all relational ("SQL") databases. Should work on MongoDB too, possibly a few more. Will not work on Cosmos DB, may not work with Redis.

How this works:

- We add a new API to components called `DeleteAllWithPrefix` which accepts a key prefix (such as `app-id||actor-type||actor-id||`)
- This method is optional and as such has its own interface. It _can_ be implemented by transactional state stores, but it is not required for transactional state stores.
- If the method is not implemented by the state store, invoking the API will return an error.

JoshVanL:
@RyanLettieri Please remember to also include [integration tests](https://github.com/dapr/dapr/blob/master/tests/docs/writing-integration-test.md) with for this new API.

RyanLettieri:
> More thoughts. Here's how we can implement this.
> 
> Premises:
> 
> * This is possible because actor state always has a prefix that indicates the actor that created it, which is: `app-id||actor-type||actor-id||key`
> * This method will not work on all state stores that support actors. It will definitely work on all relational ("SQL") databases. Should work on MongoDB too, possibly a few more. Will not work on Cosmos DB, may not work with Redis.
> 
> How this works:
> 
> * We add a new API to components called `DeleteAllWithPrefix` which accepts a key prefix (such as `app-id||actor-type||actor-id||`)
> * This method is optional and as such has its own interface. It _can_ be implemented by transactional state stores, but it is not required for transactional state stores.
> * If the method is not implemented by the state store, invoking the API will return an error.

@dapr/maintainers-dapr Thoughts on this idea for the implementation?

ItalyPaleAle:
> > We add a new API to components called DeleteAllWithPrefix which accepts a key prefix (such as app-id||actor-type||actor-id||)
> 
> Is this new API meant to be a part of the Actor API or the state API? Personally I think this should go under state API the way it's described in your comment. Alternatively, this can become an API call specifically for actors that takes as input the actor type and id, and then components expose to the runtime a new feature for deleting actors.

Could be either way. The components will just implement a “delete with prefix” which doesn’t care _what_ the prefix is. Whether we expose it to users only for actors, or also for regular state store operations, up to you if you think there’s value.

yaron2:
> > > We add a new API to components called DeleteAllWithPrefix which accepts a key prefix (such as app-id||actor-type||actor-id||)
> > 
> > 
> > Is this new API meant to be a part of the Actor API or the state API? Personally I think this should go under state API the way it's described in your comment. Alternatively, this can become an API call specifically for actors that takes as input the actor type and id, and then components expose to the runtime a new feature for deleting actors.
> 
> Could be either way. The components will just implement a “delete with prefix” which doesn’t care _what_ the prefix is. Whether we expose it to users only for actors, or also for regular state store operations, up to you if you think there’s value.

I recommend we expose this as part of the actors API.

dapr/proto/runtime/v1/dapr.proto
```diff
@@ -82,6 +82,9 @@ service Dapr {
   // Gets the state for a specific actor.
   rpc GetActorState(GetActorStateRequest) returns (GetActorStateResponse) {}
 
+  // Gets the state for a specific actor.
```
ItalyPaleAle, line 85:
```suggestion
  // Deletes all state created by an actor
```

dapr/proto/runtime/v1/dapr.proto
```diff
@@ -546,6 +549,18 @@ message GetActorStateRequest {
   string key = 3;
 }
 
+// DeleteActorStateRequest is the message to get key-value states from specific actor.
+message DeleteActorStateRequest {
+  string actor_type = 1;
+  string actor_id = 2;
+  string key = 3;
```
ItalyPaleAle, line 556: This isn't used as we delete all state

```suggestion
```

dapr/proto/runtime/v1/dapr.proto
```diff
@@ -546,6 +549,18 @@ message GetActorStateRequest {
   string key = 3;
 }
 
+// DeleteActorStateRequest is the message to get key-value states from specific actor.
```
ItalyPaleAle, line 552:
```suggestion
// DeleteActorStateRequest is the request message for DeleteActorState.
```

dapr/proto/runtime/v1/dapr.proto
```diff
@@ -546,6 +549,18 @@ message GetActorStateRequest {
...
+message DeleteActorStateRequest {
+  string actor_type = 1;
+  string actor_id = 2;
+  string key = 3;
+}
+
+// DeleteActorStateResponse is the response for a delete actor state request
```
ItalyPaleAle, line 559:
```suggestion
// DeleteActorStateResponse is the response message for DeleteActorState.
```

pkg/actors/actors.go
```diff
@@ -718,6 +720,32 @@ func (a *actorsRuntime) GetState(ctx context.Context, req *GetStateRequest) (*St
 	}, nil
 }
 
+func (a *actorsRuntime) DeleteState(ctx context.Context, req *DeleteStateRequest) (*runtimev1pb.DeleteActorStateResponse, error) {
```
ItalyPaleAle, line 723: Please don't return a proto here. Create a specific `DeleteStateResponse` object

Also, best to avoid pointers:


```suggestion
func (a *actorsRuntime) DeleteState(ctx context.Context, req DeleteStateRequest) (DeleteStateResponse, error) {
```

pkg/actors/actors.go
```diff
@@ -718,6 +720,32 @@ func (a *actorsRuntime) GetState(ctx context.Context, req *GetStateRequest) (*St
...
+	storeName, store, err := a.stateStore()
+	if err != nil {
+		return &runtimev1pb.DeleteActorStateResponse{}, err
+	}
+
+	key := constructCompositeKey(a.actorsConfig.Config.AppID, req.ActorID, req.ActorType, req.ActorKey())
```
ItalyPaleAle, line 729:
```suggestion
	key := constructCompositeKey(a.actorsConfig.Config.AppID, req.ActorID, req.ActorType)
```

pkg/actors/actors.go
```diff
@@ -718,6 +720,32 @@ func (a *actorsRuntime) GetState(ctx context.Context, req *GetStateRequest) (*St
...
+
+	policyRunner := resiliency.NewRunner[state.DeleteWithPrefixResponse](ctx,
+		a.resiliency.ComponentOutboundPolicy(storeName, resiliency.Statestore),
+	)
+
+	storeReq := state.DeleteWithPrefixRequest{
+		Prefix: key,
+	}
+
+	prefixDelete := store.(state.DeleteWithPrefix)
```
ItalyPaleAle, line 739: Need to check if this type casting is possible. If the state store doesn't support DeleteWithPrefix, this will panic.


```suggestion
	prefixDelete, ok := store.(state.DeleteWithPrefix)
```

Then check if `!ok` return an error that says that the operation isn't possible.

pkg/grpc/universalapi/api_actors.go
```diff
@@ -47,3 +53,46 @@ func (a *UniversalAPI) WaitForActorsReady(ctx context.Context) {
 	case <-a.actorsReadyCh:
 	}
 }
+
+// This function makes sure that the actor subsystem is ready.
+func (a *UniversalAPI) ActorReadinessCheck(ctx context.Context) error {
+	a.WaitForActorsReady(ctx)
+
+	if a.Actors == nil {
+		return messages.ErrActorRuntimeNotFound
```
ItalyPaleAle, line 62: Missing the debug log that was in the old function

```go
apiServerLogger.Debug(messages.ErrActorRuntimeNotFound)
```

pkg/grpc/universalapi/api_actors.go
```diff
@@ -47,3 +53,46 @@ func (a *UniversalAPI) WaitForActorsReady(ctx context.Context) {
...
+	if !hosted {
+		err := status.Errorf(codes.Internal, messages.ErrActorInstanceMissing)
+		a.Logger.Debug(err)
+		return &runtimev1pb.DeleteActorStateResponse{Count: 0}, err
+	}
+
+	res, err := a.Actors.DeleteState(ctx, &actors.DeleteStateRequest{
+		ActorType: in.GetActorType(),
+		ActorID:   actorID,
+		Key:       in.GetKey(),
```
ItalyPaleAle, line 89:
```suggestion
```

pkg/http/api_actors.go
```diff
@@ -0,0 +1,173 @@
...
+			Version:         apiVersionV1,
+			Group:           endpointGroupActorV1State,
+			FastHTTPHandler: a.onGetActorState,
+			Settings: endpoints.EndpointSettings{
+				Name: "GetActorState",
+			},
+		},
+		{
+			Methods: []string{nethttp.MethodDelete},
+			Route:   "actors/{actorType}/{actorId}/state/{key}",
```
ItalyPaleAle, line 99:
```suggestion
			Route:   "actors/{actorType}/{actorId}/state",
```

pkg/http/api_actors.go
```diff
@@ -0,0 +1,173 @@
...
+
+// Route: DELETE "actors/{actorType}/{actorId}/state/{key}"
+func (a *api) onDeleteActorStateHandler() nethttp.HandlerFunc {
+	return UniversalHTTPHandler(
+		a.universal.DeleteActorState,
+		UniversalHTTPHandlerOpts[*runtimev1pb.DeleteActorStateRequest, *runtimev1pb.DeleteActorStateResponse]{
+			InModifier: func(r *nethttp.Request, in *runtimev1pb.DeleteActorStateRequest) (*runtimev1pb.DeleteActorStateRequest, error) {
+				in.ActorId = chi.URLParam(r, actorIDParam)
+				in.ActorType = chi.URLParam(r, actorTypeParam)
+				in.Key = chi.URLParam(r, stateKeyParam)
```
ItalyPaleAle, line 168:
```suggestion
```

ItalyPaleAle (changes requested):
Pretty good first iteration :)

cicoyle:
@RyanLettieri this would be great to have in next release if possible :)

PR #7231, opened 2023-11-22
Internal and remote actor invocation optimization

ItalyPaleAle:
This PR changes the data structures used for actor invocation (direct messaging, reminders, timers), optimizing them for when internal and remote actors are invoked.

1. When invoking internal actor such as workflow actors, with this PR we do not need to create a "virtual app channel" anymore, which added significant overhead due to data being copied and re-encoded various time, and complexity with routing. When an actor is internal (and local), it is invoked directly.
1. Currently, the "standard" data structure used in the actors package is `pkg/messaging/v1.InvokeMethodRequest`, which is optimized for invocations on the app channel. This changes the main data structure to be `pkg/proto/internals/v1.InternalInvokeRequest`, which is the same that is used for local and remote actor invocation. This avoids unnecessary conversions to/from the messaging data structure, which is meant for a different purpose and optimized for streaming.
2. Lastly, internal actors have been improved by allowing the actor runtime itself manage each instance of an active actor. This way, implementation of internal actors can save state locally (in the same object) and don't need to maintain a map of states. It simplifies the development and maintenance of internal actors, as well as improving performance by using more efficient state maps and avoid locking (based on haxmap)

All told these changes should improve performance, especially when invoking internal actors, and reduce allocations.


pkg/actors/actors.go
```diff
@@ -582,64 +602,145 @@ func (a *actorsRuntime) callLocalActor(ctx context.Context, req *invokev1.Invoke
...
+		if md != nil && len(md["Dapr-Reentrancy-Id"]) > 0 {
+			reentrancyID = ptr.Of(md["Dapr-Reentrancy-Id"][0])
+		} else {
+			var uuidObj uuid.UUID
+			uuidObj, err = uuid.NewRandom()
+			if err != nil {
+				return nil, fmt.Errorf("failed to generate UUID: %w", err)
+			}
+			reentrancyID = ptr.Of(uuidObj.String())
+			if md == nil {
```
DeepanshuA, line 687: md is already initialized with length, 1 greater than metadata length - do we need this check here and re-initialize it?

pkg/actors/internal_actor.go
```diff
@@ -21,145 +21,57 @@ import (
...
-	internalTypeName := actorType
-	if !strings.HasPrefix(actorType, InternalActorTypePrefix) {
-		internalTypeName = InternalActorTypePrefix + actorType
-	}
-	if _, exists := c.actors[internalTypeName]; exists {
-		return fmt.Errorf("internal actor named '%s' already exists", actorType)
-	}
-	c.actors[internalTypeName] = actorImpl
-	return nil
+type InternalActorTimer struct {
```
DeepanshuA, line 57: Can we use single struct for `InternalActorReminder` and `InternalActorTimer`?

pkg/actors/internal_actor.go
```diff
@@ -21,145 +21,57 @@ import (
...
+	"github.com/dapr/dapr/pkg/actors/internal"
 )
 
 const InternalActorTypePrefix = "dapr.internal."
 
 // InternalActor represents the interface for invoking an "internal" actor (one which is built into daprd directly).
 type InternalActor interface {
 	SetActorRuntime(actorsRuntime Actors)
-	InvokeMethod(ctx context.Context, actorID string, methodName string, data []byte) (any, error)
+	InvokeMethod(ctx context.Context, actorID string, methodName string, data []byte, metadata map[string][]string) ([]byte, error)
```
DeepanshuA, line 33: Why do we need extra `metadata` here ?

pkg/actors/internal_actor.go
```diff
@@ -21,145 +21,57 @@ import (
...
 	DeactivateActor(ctx context.Context, actorID string) error
-	// InvokeReminder invokes reminder logic for an internal actor.
-	// Note that the DecodeInternalActorReminderData function should be used to decode the [data] parameter.
-	InvokeReminder(ctx context.Context, actorID string, reminderName string, data []byte, dueTime string, period string) error
-	InvokeTimer(ctx context.Context, actorID string, timerName string, params []byte) error
+	InvokeReminder(ctx context.Context, reminder InternalActorReminder, metadata map[string][]string) error
```
DeepanshuA, line 35: Same - is extra `metadata` required here?

pkg/grpc/api.go
```diff
@@ -1188,12 +1177,8 @@ func (a *api) InvokeActor(ctx context.Context, in *runtimev1pb.InvokeActorReques
 	// should technically wait forever on the locking mechanism. If we timeout while
 	// waiting for the lock, we can also create a queue of calls that will try and continue
 	// after the timeout.
-	policyRunner := resiliency.NewRunnerWithOptions(ctx, policyDef,
-		resiliency.RunnerOpts[*invokev1.InvokeMethodResponse]{
-			Disposer: resiliency.DisposerCloser[*invokev1.InvokeMethodResponse],
-		},
-	)
-	resp, err := policyRunner(func(ctx context.Context) (*invokev1.InvokeMethodResponse, error) {
+	policyRunner := resiliency.NewRunner[*internalv1pb.InternalInvokeResponse](ctx, policyDef)
```
DeepanshuA, line 1180: For my understanding: We were using Disposer to take care of failure scenarios. Won't we need it now?

DeepanshuA (commented):
Doing first pass

pkg/actors/actors.go
ItalyPaleAle: We can't be sure md != nil here

pkg/actors/internal_actor.go
ItalyPaleAle: Good question, let me see if I can get away with removing it

pkg/grpc/api.go
ItalyPaleAle: No need now because the data is not held in a stream anymore for actors!

pkg/actors/internal_actor.go
ItalyPaleAle: The metadata includes things like headers and re-entrancy IDs. I think it may be safe to leave it there in case an internal actor needs them. I don't believe it's the case today, but probably not a bad practice

pkg/actors/internal_actor.go
ItalyPaleAle: I think so, it would probably help simplifying. I'll change that.

ItalyPaleAle: Done

DeepanshuA (approved):
LGTM

pkg/actors/actors.go
```diff
@@ -158,20 +162,20 @@ func NewActors(opts ActorsOpts) ActorRuntime {
...
+		grpcConnectionFn:   opts.GRPCConnectionFn,
+		actorsConfig:       opts.Config,
+		timers:             timers.NewTimersProvider(clock),
+		tracingSpec:        opts.TracingSpec,
+		resiliency:         opts.Resiliency,
+		storeName:          opts.StateStoreName,
+		placement:          opts.MockPlacement,
+		actorsTable:        &sync.Map{},
+		clock:              clock,
+		internalActorTypes: haxmap.New[string, InternalActorFactory](4), // Initial capacity should be enough for the current built-in actors
```
daixiang0, line 175: Magic number?

pkg/actors/actors.go
```diff
@@ -158,20 +162,20 @@ func NewActors(opts ActorsOpts) ActorRuntime {
...
+		internalActors:     haxmap.New[string, InternalActor](32),
```
daixiang0, line 176: Magic number?

pkg/actors/actors.go
```diff
@@ -370,40 +374,55 @@ func (a *actorsRuntime) haltAllActors() error {
...
+		// If the actor was loaded in-memory, call DeactivateActor on it
+		if loaded {
+			err = internalAct.DeactivateActor(ctx)
+			if err != nil {
+				diag.DefaultMonitoring.ActorDeactivationFailed(act.actorType, "internal")
+				return fmt.Errorf("failed to deactivate internal actor: %w", err)
+			}
+		}
+	} else if a.appChannel != nil {
+		req := invokev1.NewInvokeMethodRequest("actors/"+act.actorType+"/"+act.actorID).
```
daixiang0, line 403: Could we have a func to generate request URL, if so we can easily update URL and do not change many codes.

daixiang0, line 403: Also use `daprSeparator`

pkg/actors/actors.go
```diff
@@ -547,39 +577,45 @@ func (a *actorsRuntime) getOrCreateActor(act *internalv1pb.Actor) *actor {
...
+	}
+	defer imReq.Close()
 
 	// Reentrancy to determine how we lock.
 	var reentrancyID *string
 	if a.actorsConfig.GetReentrancyForType(act.actorType).Enabled {
-		if headerValue, ok := req.Metadata()["Dapr-Reentrancy-Id"]; ok {
-			reentrancyID = &headerValue.GetValues()[0]
+		if md := imReq.Metadata()["Dapr-Reentrancy-Id"]; md != nil && len(md.GetValues()) > 0 {
+			reentrancyID = ptr.Of(md.GetValues()[0])
```
daixiang0, line 594: Please add an example about whole value.

pkg/actors/actors.go
```diff
@@ -547,39 +577,45 @@ func (a *actorsRuntime) getOrCreateActor(act *internalv1pb.Actor) *actor {
...
 		} else {
-			uuidObj, err := uuid.NewRandom()
+			var uuidObj uuid.UUID
```
daixiang0, line 596: Any reason that use `var`?

pkg/actors/actors.go
ItalyPaleAle: It's just an initial capacity that makes sense. If more capacity is needed, it'll be allocated

pkg/actors/actors.go
ItalyPaleAle: It's explained there in a comment. it's an initial capacity that makes sense given the current number of internal actors. If more capacity is needed it'll be allocated.

pkg/actors/actors.go
ItalyPaleAle: Maybe but at this stage we will not change URLs until a "v2".

ItalyPaleAle: No, daprSeparator is || and here we use /

pkg/actors/actors.go
ItalyPaleAle: I'm not sure I follow your ask. This is not new code, it was there before

pkg/actors/actors.go
ItalyPaleAle: To appease our linter overloard

daixiang0:
@ItalyPaleAle see conflicts.

ItalyPaleAle:
> @ItalyPaleAle see conflicts.

Fixed

pkg/api/grpc/daprinternal.go
```diff
@@ -281,34 +281,23 @@ func (a *api) CallLocalStream(stream internalv1pb.ServiceInvocation_CallLocalStr
...
-	defer req.Close()
-
 	// We don't do resiliency here as it is handled in the API layer. See InvokeActor().
-	resp, err := a.Actors().Call(ctx, req)
+	res, err := a.Actors().Call(ctx, in)
 	if err != nil {
-		// We have to remove the error to keep the body, so callers must re-inspect for the header in the actual response.
 		actorErr, isActorErr := actorerrors.As(err)
-		if resp != nil && isActorErr {
-			defer resp.Close()
```
yaron2, line 296: Don't we want to keep this?

ItalyPaleAle: No, as part of the refactoring this is now not a stream (since the data was never a stream to begin with for actor invocation), so no need to clsoe

daixiang0:
Please fix conflicts.

Codes looks good to me.

ItalyPaleAle:
> Please fix conflicts.
> 
> Codes looks good to me.

Fixed

daixiang0 (approved):
LGTM

yaron2 (approved):

PR #7239, opened 2023-11-26
Hot Reloading: SelfHosted Componenthot reloading

JoshVanL:
Part of #1172

Adds hot reloading functionality to Daprd, which is gated behind the HotReload preview feature.

If enabled, Daprd in Selfhosted mode will watch for resource updates from watching the resource directory(s). If a resource is detected as changed (spec is different to what is currently loaded), then Dapr will close/create/update the resource.

Supports Component _except_ HTTP middlware which requires further changes to the HTTP server handler machinery. A warning is thrown if hot reloading is enabled and a HTTP middleware is updated.

The hot reloader reconciler is made generic to enable other resource types to be added in future.

---

When running in standalone mode, the disk loader watches the directories passed by --resources-path for updates to yaml files containing Components.

When an event occurs, the reconciler and differ are responsible for determining whether any resources have been created, updated, or deleted, by comparing the local store specs with that of the remote. If any have changed, the resource is closed (if it exists), and then re-initialized (if it exists). A resource will only be closed if it has been deleted, only initialized if it has been created, and closed & initialized if it has been updated.

We consider a resource to have changed generally if anything apart from its Kubernetes metadata or type meta object meta has changed, and therefore needs some kind of reloading.

---

Currently, if a reloading component errors and `spec.ignoreErrors=false` then Daprd will gracefully close, like a component loaded at startup today. It is intended that in future the component will be re-inited on a backoff queue in future in the case of errors, even if `spec.ignoreErros=true`.

HTTP middleware component hot reloading is not supported as it requires further changes to the HTTP server handler machinery, and I didn't want to grow this PR further.

The differ, loader, and reconciler have been made generic so that they can be reused when other resource types (HTTPEndpoint, Subscriptions etc.) can be added to hot reloading in future.

To use the HotReloading feature, which is currently only available as an alpha feature in Selfhosted mode, users need to add the following Configuration to the target Daprd;

```yaml
apiVersion: dapr.io/v1alpha1
kind: Configuration
metadata:
  name: hotreloading
spec:
  features:
  - name: HotReload
    enabled: true
```


pkg/runtime/compstore/crypto.go
```diff
@@ -13,7 +13,9 @@ limitations under the License.
 
 package compstore
 
-import "github.com/dapr/components-contrib/crypto"
+import (
```
daixiang0, line 16: Let's keep it as it was.

pkg/apis/components/v1alpha1/register.go
```diff
@@ -24,8 +24,8 @@ import (
 // SchemeGroupVersion is group version used to register these objects.
 var SchemeGroupVersion = schema.GroupVersion{Group: components.GroupName, Version: "v1alpha1"}
 
-// Kind takes an unqualified kind and returns back a Group qualified GroupKind.
-func Kind(kind string) schema.GroupKind {
+// GroupKind takes an unqualified kind and returns back a Group qualified GroupKind.
+func GroupKindFromKind(kind string) schema.GroupKind {
```
daixiang0, line 28:
```suggestion
func GroupKind(kind string) schema.GroupKind {
```

pkg/apis/httpEndpoint/v1alpha1/register.go
```diff
@@ -24,8 +24,9 @@ import (
 // SchemeGroupVersion is group version used to register these objects.
 var SchemeGroupVersion = schema.GroupVersion{Group: httpendpoint.GroupName, Version: "v1alpha1"}
 
-// Kind takes an unqualified kind and returns back a Group qualified GroupKind.
-func Kind(kind string) schema.GroupKind {
+// GroupKindFromKind takes an unqualified kind and returns back a Group
+// qualified GroupKind.
+func GroupKindFromKind(kind string) schema.GroupKind {
```
daixiang0, line 29: Same as above.

tests/integration/framework/process/logline/logline.go
```diff
@@ -39,6 +39,8 @@ type LogLine struct {
 	stderr            io.Reader
 	stderrExp         io.WriteCloser
 	stderrLinContains map[string]bool
+
+	outCheck chan map[string]bool
```
daixiang0, line 43:
```suggestion
	checkOut chan map[string]bool
```

daixiang0:
@JoshVanL thanks for your great job! 

I see you close two PRs before, which make some conversations missing. Could we make changes as much as possible in one place to avoid reviewers starting over every time?

ping @ItalyPaleAle @mukundansundar @cicoyle

JoshVanL:
> @JoshVanL thanks for your great job! 
> 
> 
> 
> I see you close two PRs before, which make some conversations missing. Could we make changes as much as possible in one place to avoid reviewers starting over every time?
> 
> 
> 
> ping @ItalyPaleAle @mukundansundar @cicoyle

Thanks @daixiang0 :)

This was done intentionally as the original hot reloading PR has diverged significantly enough (this PR is only selfhosted, doesn't include middleware, doesn't include HTTPEndpoints, has loads of integration tests) that force pushing to the same PR would cause more confusion than it was worth imo.

pkg/runtime/compstore/crypto.go
JoshVanL: Done=.

pkg/internal/apis/namevaluepair.go
```diff
@@ -42,3 +46,11 @@ func (g GenericNameValueResource) GetSecretStore() string {
 func (g GenericNameValueResource) NameValuePairs() []common.NameValuePair {
 	return g.Pairs
 }
+
+func (g GenericNameValueResource) LogName() string {
+	return g.Name + " (" + g.ResourceKind + ")"
```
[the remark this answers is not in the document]
JoshVanL: Done

pkg/apis/components/v1alpha1/register.go
JoshVanL: Done

pkg/apis/httpEndpoint/v1alpha1/register.go
JoshVanL: Done

pkg/runtime/hotreload/differ/differ.go
```diff
@@ -0,0 +1,130 @@
...
+	// missing are the resources which exist remotely but which don't exist
+	// locally or have changed.
+	missing := detectDiff(resources.Local, resources.Remote, nil)
+
+	// deleted are the resources which exist locally but which don't exist
+	// remotely or have changed.
+	deleted := detectDiff(resources.Remote, resources.Local, func(r T) bool {
+		if comp, ok := any(r).(componentsapi.Component); ok {
+			// Ignore the built-in Kubernetes secret store and workflow engine.
+			if comp.Name == secretstores.BuiltinKubernetesSecretStore &&
```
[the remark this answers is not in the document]
JoshVanL: The in-built components are not stored in the loader (i.e. they do not exist in the referenced `--resources-path` or as Kubernetes objects) therefore they need to be special cased. If they weren't special cased then they would be marked for deletion by the differ.

pkg/runtime/hotreload/differ/differ.go
```diff
@@ -0,0 +1,130 @@
...
+			}
+		}
+
+		return false
+	})
+
+	var result Result[T]
+
+	for i := range deleted {
+		if _, ok := missing[deleted[i].GetName()]; !ok {
```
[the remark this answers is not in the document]
JoshVanL: Yes, there are 3 states a object is in which will take the following actions:

- CREATED: object will be created (init).
- UPDATED: object with be _deleted_ and then _created_ again with the new spec.
- DELETED: object will be deleted (closed).

tests/integration/framework/process/logline/logline.go
JoshVanL: We can't name it this because we already have a func called this https://github.com/dapr/dapr/pull/7239/files#diff-b975baf9f073a4828870a3b544cc351547501ed77e4c82459856babb979c261dR99

pkg/runtime/hotreload/loader/disk/generic.go
```diff
@@ -0,0 +1,128 @@
...
+			log.Errorf("Failed to load resources from disk: %s", err)
+			continue
+		}
+
+		result := differ.Diff(resources)
+		if result == nil {
+			continue
+		}
+
+		for _, group := range []struct {
```
mukundansundar, line 106: please add comments here.

pkg/runtime/hotreload/loader/store/store_test.go
```diff
@@ -0,0 +1,48 @@
...
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	componentsapi "github.com/dapr/dapr/pkg/apis/components/v1alpha1"
+	"github.com/dapr/dapr/pkg/runtime/compstore"
+)
+
+func Test_component(t *testing.T) {
```
mukundansundar, line 27: camelCase for tests also ...

mukundansundar (commented):
first pass review

pkg/internal/apis/namevaluepair.go
```diff
@@ -42,3 +48,11 @@ func (g GenericNameValueResource) GetSecretStore() string {
 func (g GenericNameValueResource) NameValuePairs() []common.NameValuePair {
 	return g.Pairs
 }
+
+func (g GenericNameValueResource) LogName() string {
```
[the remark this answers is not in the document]
JoshVanL: This is consistent with other API types https://github.com/dapr/dapr/blob/master/pkg/apis/components/v1alpha1/types.go#L45

pkg/runtime/hotreload/differ/differ.go
```diff
@@ -0,0 +1,130 @@
...
+	meta.Resource
+}
+
+type Result[T Resource] struct {
+	Deleted []T
+	Updated []T
+	Created []T
+}
+
+type LocalRemoteResources[T Resource] struct {
```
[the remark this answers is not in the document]
JoshVanL: Which one? `[T Resource]` is singular as the type is singular, `LocalRemoteResources` is plural because is contains multiple `Resource`s.

pkg/runtime/hotreload/hotreload.go
```diff
@@ -0,0 +1,85 @@
...
+	Processor      *processor.Processor
+}
+
+type Reloader struct {
+	isEnabled  bool
+	components *reconciler.Reconciler[componentsapi.Component]
+}
+
+func NewDisk(ctx context.Context, opts OptionsDisk) (*Reloader, error) {
+	loader, err := disk.New(ctx, disk.Options{
```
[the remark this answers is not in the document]
JoshVanL: This loader is not limited to components, and will also contain other Resources (HTTPEndpoitns, Subscriptions..) in future.

pkg/runtime/hotreload/loader/store/store_test.go
JoshVanL: Same as above ^

dapr/proto/operator/v1/operator.proto
```diff
@@ -40,6 +40,21 @@ service Operator {
   rpc HTTPEndpointUpdate (HTTPEndpointUpdateRequest) returns (stream HTTPEndpointUpdateEvent) {}
 }
 
+// ResourceEventType is the type of event to a resource.
+enum ResourceEventType {
+  // UNKNOWN indicates that the event type is unknown.
+  UNKNOWN = 0;
```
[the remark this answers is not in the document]
JoshVanL: It's common/best practice to have the zero enum value to be `UNKNOWN` https://stackoverflow.com/questions/10392952/how-to-design-for-a-future-additional-enum-value-in-protocol-buffers

artursouza (dismissed):

berndverst:
This fails if the component `namespace` is defined, but not set to `default`. It even prevents the component from being loaded at all now.

JoshVanL:
@berndverst is this not expected behaviour? daprd in self-hosted mode does still have the concept of namespacing (though this isn't documented well and not clear to the user). The namespace of daprd is set via the environment variable `NAMESPACE` which itself is set by Kubernetes so is transparent to the user but still clear through context. daprd will default the namespace to `default` when the env var is not set.

If it is the case that daprd would previously consume Components through file in seflhosted mode, even if the namespace set did not match the `NAMESPACE` env, then this can be considered a breaking change. Please let me know if this is the case and if it is a breaking change we do not want to accept.

Issue #3002, opened 2021-03-30
Add parallelism to Docker build in E2Es


artursouza:
Speed up E2E tests by parallelizing Docker images build and publish.

mukundansundar:
@artursouza 

I am thinking if make can run parallel builds if so can the `build-e2e-app-all` job should be run in parallel. 
Though for build, i think it should be split into two steps:
1. Get the base image for the apps
2. Build e2e apps in parallel
https://github.com/dapr/dapr/blob/master/tests/dapr_tests.mk#L80 and same for `push-e2e-app-all`. 

For Linux build: https://www.gnu.org/software/make/manual/html_node/Parallel.html
Potential for Windows build? https://stackoverflow.com/questions/1533425/make-parallel-jobs-on-windows

MregXN:
I added `-j4`  when `Build Dapr and its Docker images` and `Build and push E2E test apps`, but it seems not to make any difference.

I made a simple test in my PR：

|     | before add -j4  | after add -j4   | 
|  ----  | ----  |----  |
| Build Dapr and its Docker images for linux  | 24s |29s |
| Build and push E2E test apps for linux  | 2m22s |2m23s | 
| Build Dapr and its Docker images for windows| 1m39s |3m52s|
| Build and push E2E test apps for windows| 4m57s |4m12s | 

action [link](https://github.com/dapr/dapr/actions/runs/7029226580/job/19126497122) of `before add -j4`
action [link](https://github.com/dapr/dapr/actions/runs/7033507164) of `after add -j4`


I don’t know the reason yet. Maybe we also need to find some other ways to speed up the build process.
@artursouza @mukundansundar

PR #7256, opened 2023-11-29
Add parallelism to Docker build in E2Es

MregXN:
# Description

Please refer to the issue #3002

### DIFF
--- a/pkg/runtime/wfengine/activity.go
+++ b/pkg/runtime/wfengine/activity.go
@@ -78,6 +78,7 @@ func (a *activityActor) SetActorRuntime(actorsRuntime actors.Actors) {
 // returns immediately after creating the reminder, enabling the workflow to continue processing other events
 // in parallel.
 func (a *activityActor) InvokeMethod(ctx context.Context, actorID string, methodName string, data []byte) (any, error) {
+	wfLogger.Debugf("Activity actor '%s': invoking method '%s'", actorID, methodName)
 	var ar ActivityRequest
 	if err := actors.DecodeInternalActorData(bytes.NewReader(data), &ar); err != nil {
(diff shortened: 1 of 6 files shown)


.github/workflows/dapr-test.yml
```diff
@@ -334,12 +334,12 @@ jobs:
       - name: Build Dapr and its Docker images
         if: env.TEST_PREFIX != ''
         run: |
-          make build
+          make build -j4
```
ItalyPaleAle, line 337: I believe our Actions runners have 2 cores, so I wouldn't set to more than 2 parallel jobs.

MregXN: it seems that I do not have access to the runner's details. Thank you for the information！

ItalyPaleAle: Also note that in _most_ cases, the images are cached, so they shouldn't be rebuilt... unless something changes :)

MregXN: Thanks！do you have any idea about how to speed up the build process? Feel free to comment here or in the issue.

ItalyPaleAle: I think the images that take the longest are the Java and Python ones, IIRC? When they are not cached

ItalyPaleAle: Let's take a look at a recent E2E test: https://github.com/dapr/dapr/actions/runs/7039695179

The longest thing (aside from test) is actually retrieving the Go dependencies, which is very slow especially on Windows (lots of disk I/O with lots of small files). This is a problem with all our actions since we have A LOT of dependencies. I don't think there's much we can do there to speed that up sadly :(

Issue #6068, opened 2023-03-13
Dapr-specific error codes in APIs


artursouza:
## In what area(s)?

> /area runtime

> /area operator

> /area placement

> /area docs

> /area test-and-release
## Describe the proposal

In addition to the usual status codes for HTTP and gRPC APIs for Dapr, it should also include Dapr-specific error codes. It can quickly be used to identify troubleshooting guides and even (in some cases) trigger recovery code paths in the application.

Examples:
- Error code for eTag mismatch in state store
- Error code for invalid Resiliency policy
- Error code for secret not found

berndverst:
@robertojrojas will you be moving the proposal to a PR in the Dapr org somewhere so we can comment on it there as well? Ideally in the `dapr/proposals` repo?

robertojrojas:
@berndverst I just created a [PR](https://github.com/dapr/proposals/pull/21) with the proposal.

cicoyle:
Pubsub api error standardization PR is open 👍🏻

mukundansundar:
moved overarching tracking issue to 1.14 and added merged PRs into 1.13.

PR #7257, opened 2023-11-29
Richer error codes model

elena-kolevska:
# Description

This PR is based on the proposal for standardising Dapr error codes: https://github.com/dapr/proposals/blob/main/0009-BCIRS-error-handling-codes.md

It has a dependency on https://github.com/dapr/kit/pull/77

## Issue reference

https://github.com/dapr/dapr/issues/6068

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests
* [X] Unit tests passing
* [x] Extend reference documentation https://github.com/dapr/docs/issues/3907
* [X] Extended the documentation for go-sdk: https://github.com/dapr/go-sdk/pull/482
* [X] Created an issue for improving DevEx for error handling in the python-sdk repo: https://github.com/dapr/python-sdk/issues/649


pkg/apis/errors/errors.go
ItalyPaleAle, line 1: Note that pkg/apis is for the K8s APIs (like, CRDs) only (i know, not the best naming)...

elena-kolevska: I'm sorry, this was supposed to be `api`. I'll fix it.

tests/integration/framework/framework.go
```diff
@@ -20,6 +20,8 @@ import (
 	"github.com/dapr/dapr/tests/integration/framework/process"
 )
 
+var Domain = "dapr.io"
```
cicoyle, line 23: I have this exported in kit as `Domain`, you can use kitErrors.Domain

elena-kolevska: For integration tests I kind of prefer having the expected values on the test side, so we can catch if something changes in the code and prevent possible breaking changes to clients. This is especially true for error codes, but I think it applies to ErrorInfo details too.

Makefile
```diff
@@ -373,8 +373,8 @@ test-integration: test-deps
 ################################################################################
 # Target: lint                                                                 #
 ################################################################################
-# Please use golangci-lint version v1.51.2 , otherwise you might encounter errors.
-# You can download version v1.51.2 at https://github.com/golangci/golangci-lint/releases/tag/v1.51.2
+# Please use golangci-lint version v1.55.2 , otherwise you might encounter errors.
+# You can download version v1.55.2 at https://github.com/golangci/golangci-lint/releases/tag/v1.51.2
```
JoshVanL, line 377: Revert

dapr/proto/components/v1/state.proto
```diff
@@ -286,4 +286,24 @@ message BulkSetRequest {
...
+// MultiMaxSizeRequest is the request for MultiMaxSize. It is empty because
+// there are no parameters.
+message MultiMaxSizeRequest {}
+
+// MultiMaxSizeResponse is the response for MultiMaxSize.
+message MultiMaxSizeResponse {
+  // The maximum number of operations that can be performed in a single
+  // transaction.
+  int64 max_size = 1;
+}
```
JoshVanL, line 309: This has been added for support in the gRPC pluggable state store.

docs/development/developing-dapr.md
```diff
@@ -63,7 +63,7 @@ This command will:
 - format, test and lint all the code 
 - check if you forgot to `git commit` something
 
-Note: To run linter locally, please use golangci-lint version v1.51.2, otherwise you might encounter errors. You can download version v1.55.2 [here](https://github.com/golangci/golangci-lint/releases/tag/v1.55.2).
+Note: To run linter locally, please use golangci-lint version v1.55.2, otherwise you might encounter errors. You can download version v1.55.2 [here](https://github.com/golangci/golangci-lint/releases/tag/v1.55.2).
```
JoshVanL, line 66: Revert

pkg/api/errors/state.go
```diff
@@ -0,0 +1,108 @@
...
+package errors
+
+import (
+	"fmt"
+	"net/http"
+	"strconv"
+
+	grpcCodes "google.golang.org/grpc/codes"
+
+	kitErrors "github.com/dapr/kit/errors"
```
JoshVanL, line 23: We only use import aliases if there is a package import name conflict. We also don't use camel casing in package names- if we really needed an alias for `kit/errors`, we would use the alias `kiterrors`.

```suggestion
	"google.golang.org/grpc/codes"

	"github.com/dapr/kit/errors"
```

pkg/api/errors/state.go
```diff
@@ -0,0 +1,108 @@
...
+)
+
+func StateStoreNotConfigured() *kitErrors.Error {
```
JoshVanL: We should return the built in `error` type here.

```suggestion
func StateStoreNotConfigured() error {
```

pkg/grpc/api.go
```diff
@@ -24,6 +24,8 @@ import (
 	"sync/atomic"
 	"time"
 
+	apierrors "github.com/dapr/dapr/pkg/api/errors"
+
```
JoshVanL, line 28: Move import to `dapr` block below.

pkg/api/errors/state.go
```diff
@@ -0,0 +1,108 @@
...
+		http.StatusInternalServerError,
+		"state store does not support querying",
+		"ERR_STATE_STORE_NOT_SUPPORTED",
+	).
+		WithErrorInfo(kitErrors.CodePrefixStateStore+"QUERYING_"+kitErrors.CodeNotSupported, nil).
+		WithResourceInfo("state", storeName, "", "")
+}
+
+func StateStoreQueryFailed(storeName string, detail string) *kitErrors.Error {
+	// TODO: @elena-kolevska #change-in-next-major: The http and grpc codes are wrong, but they're legacy.
```
JoshVanL, line 105: What is wrong about the codes? We also probably will never have a new `major` version of Dapr so it will likely be this way forever :smile:

pkg/components/state/pluggable.go
```diff
@@ -321,6 +321,17 @@ func (ss *grpcStateStore) Multi(ctx context.Context, request *state.Transactiona
 	return err
 }
 
+// MultiMaxSize returns the maximum number of operations allowed in a transactional request.
+func (ss *grpcStateStore) MultiMaxSize() int {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
```
JoshVanL, line 327: Can we please have at least a timeout here? 5 seconds would be more then enough.

pkg/grpc/universalapi/api_state_query.go
```diff
@@ -18,30 +18,31 @@ import (
 	"encoding/json"
 	"time"
 
+	"github.com/dapr/dapr/pkg/api/errors"
```
JoshVanL, line 21: Move to block below.

pkg/http/api.go
```diff
@@ -25,6 +25,8 @@ import (
 	"strings"
 	"time"
 
+	apierrors "github.com/dapr/dapr/pkg/api/errors"
```
JoshVanL, line 28: Move to block below.

pkg/http/responses.go
```diff
@@ -143,6 +145,13 @@ func respondWithError(w http.ResponseWriter, err error) {
 		return
 	}
 
+	// Check if it's a kiterrors.Error object
+	apiKitErr, ok := err.(*kitErrors.Error)
```
JoshVanL: I think we would ideally have `FromError` as a helper function in `kit/errors`

tests/integration/framework/framework.go
JoshVanL: I agree that this doesn't belong here- at a minimum this should be a `const` and not live in `framework.go` as it is test data. Ideally, we would use `"dapr.io"` string literals in test assertions.

pkg/http/responses_fasthttp.go
```diff
@@ -16,13 +16,42 @@ package http
 import (
 	"net/http"
 
+	kitErrors "github.com/dapr/kit/errors"
+
 	"github.com/valyala/fasthttp"
 
 	"github.com/dapr/dapr/pkg/messages"
```
JoshVanL, line 24:
```suggestion
	"github.com/valyala/fasthttp"
	
	"github.com/dapr/kit/errors"
	"github.com/dapr/dapr/pkg/messages"
```

tests/integration/framework/process/statestore/inmemory/inmemory.go
```diff
@@ -0,0 +1,138 @@
...
+	return nil, nil
+}
+
+func (w *WrappedTransactionalMultiMaxSize) MultiMaxSize() int {
+	w.lock.Lock()
+	defer w.lock.Unlock()
+	if w.transactionalStoreMultiMaxSizeFn != nil {
+		return w.transactionalStoreMultiMaxSizeFn()
+	}
+	return 10000000
```
JoshVanL, line 123: Should we make this `-1` instead?

tests/integration/framework/process/statestore/statestore.go
```diff
@@ -54,8 +54,6 @@ func New(t *testing.T, fopts ...Option) *StateStore {
 	socketFile := util.RandomString(t, 8)
 
 	require.NotNil(t, opts.statestore)
-	_, ok := opts.statestore.(state.TransactionalStore)
-	require.True(t, ok, "statestore must implement state.TransactionalStore")
```
JoshVanL, line 57: Should we add require checks here which tests that `opts.statestore` implement the relevant interfaces if `opts.queryFunc` or `opts.transactionalStoreMultiMaxSizeFn` are not nil?

tests/integration/suite/daprd/state/grpc/errors.go
```diff
@@ -0,0 +1,483 @@
...
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials/insecure"
+
+	apierrors "github.com/dapr/dapr/pkg/api/errors"
+
+	"google.golang.org/genproto/googleapis/rpc/errdetails"
+	grpcCodes "google.golang.org/grpc/codes"
+
+	"github.com/dapr/components-contrib/state"
+	commonv1 "github.com/dapr/dapr/pkg/proto/common/v1"
```
JoshVanL, line 37:
```suggestion
	"testing"

	"golang.org/x/net/nettest"
	"google.golang.org/grpc/status"
	"github.com/stretchr/testify/require"
	"google.golang.org/grpc"
	"google.golang.org/grpc/credentials/insecure"
	"google.golang.org/genproto/googleapis/rpc/errdetails"
	grpcCodes "google.golang.org/grpc/codes"
	
	apierrors "github.com/dapr/dapr/pkg/api/errors"
	"github.com/dapr/components-contrib/state"
	commonv1 "github.com/dapr/dapr/pkg/proto/common/v1"
```

tests/integration/suite/daprd/state/grpc/errors.go
```diff
@@ -0,0 +1,483 @@
...
+		require.Nil(t, errInfo.GetMetadata())
+
+		require.NotNil(t, resInfo, "ResourceInfo should be present")
+		require.Equal(t, "state", resInfo.GetResourceType())
+		require.Equal(t, stateStoreName, resInfo.GetResourceName())
+		require.Empty(t, resInfo.GetOwner())
+		require.Empty(t, resInfo.GetDescription())
+	})
+
+	// TODO: test for NewErrStateStoreTransactionsNotSupported
```
JoshVanL, line 360: Stil TODO?

tests/integration/suite/daprd/state/http/errors.go
```diff
@@ -0,0 +1,623 @@
...
+	"path/filepath"
+	"strings"
+	"testing"
+
+	apierrors "github.com/dapr/dapr/pkg/api/errors"
+
+	"github.com/stretchr/testify/require"
+	"golang.org/x/net/nettest"
+
+	"github.com/dapr/components-contrib/state"
```
JoshVanL, line 32:
```suggestion
	"github.com/stretchr/testify/require"
	"golang.org/x/net/nettest"

	apierrors "github.com/dapr/dapr/pkg/api/errors"
	"github.com/dapr/components-contrib/state"
```

tests/integration/suite/daprd/state/http/errors.go
```diff
@@ -0,0 +1,623 @@
...
+
+	return []framework.Option{
+		framework.WithProcesses(storeWithNoTransactional, storeWithQuerier, storeWithMultiMaxSize, e.daprd),
+	}
+}
+
+func (e *errors) Run(t *testing.T, ctx context.Context) {
+	e.daprd.WaitUntilRunning(t, ctx)
+
+	// postURL := fmt.Sprintf("http://localhost:%d/v1.0/state/mystore", b.daprd.HTTPPort())
```
JoshVanL, line 146:
```suggestion
```

JoshVanL (changes requested):

Makefile
elena-kolevska: Should I send a separate PR for this?

pkg/components/state/pluggable.go
```diff
@@ -321,6 +322,18 @@ func (ss *grpcStateStore) Multi(ctx context.Context, request *state.Transactiona
 	return err
 }
 
+// MultiMaxSize returns the maximum number of operations allowed in a transactional request.
+func (ss *grpcStateStore) MultiMaxSize() int {
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
```
ItalyPaleAle, line 327: Value should be cached, as it should be a constant

pkg/http/api.go
```diff
@@ -1920,17 +1917,17 @@ func (a *api) onPostStateTransaction(reqCtx *fasthttp.RequestCtx) {
...
 		universalFastHTTPErrorResponder(reqCtx, err)
 		return
 	}
 
 	transactionalStore, ok := store.(state.TransactionalStore)
-	if !ok {
-		msg := NewErrorResponse("ERR_STATE_STORE_NOT_SUPPORTED", fmt.Sprintf(messages.ErrStateStoreNotSupported, storeName))
-		fasthttpRespond(reqCtx, fasthttpResponseWithError(nethttp.StatusInternalServerError, msg))
-		log.Debug(msg)
+	if !ok || !state.FeatureTransactional.IsPresent(store.Features()) {
```
ItalyPaleAle, line 1927: A conformance test in components-contrib is in place to ensure that if `TransactionalStore` is implemented, the feature is exposed, and vice-versa.

```suggestion
	if !ok {
```

pkg/components/state/pluggable.go
```diff
@@ -321,6 +322,18 @@ func (ss *grpcStateStore) Multi(ctx context.Context, request *state.Transactiona
...
+	defer cancel()
+
+	resp, err := ss.Client.MultiMaxSize(ctx, new(proto.MultiMaxSizeRequest))
+	if err != nil {
```
ItalyPaleAle, line 331: Should handle errors depending on the error.

If the error is `Unimplemented`, then -1 is a valid answer. Otherwise, should probably return an error

pkg/components/state/pluggable.go
```diff
@@ -321,6 +322,18 @@ func (ss *grpcStateStore) Multi(ctx context.Context, request *state.Transactiona
...
+		return -1
+	}
+	return int(resp.GetMaxSize())
```
ItalyPaleAle, line 334: Needs overflow check on 32-bit systems

ItalyPaleAle (changes requested):

pkg/http/api.go
elena-kolevska: We needed this here because when we're using a pluggable component, the interface check will always be satisfied.

pkg/components/state/pluggable.go
elena-kolevska: This would happen if the pluggable component runs on a 64-bit machine and dapr/dapr on a 32-bit one. It's probably never going to happen in reality, or at least I don't know of any databases that have a limitation of a few billion transactions, but it's always nice to be precise. 
So how should we handle this? Should we limit it to `math.MaxInt32`?

pkg/components/state/pluggable.go
elena-kolevska: We need to satisfy the MultiMaxSize interface though, which doesn't return an error.

pkg/components/state/pluggable.go
ItalyPaleAle: Correct, it is unlikely but still _possible_

I think a simple way would be to check if it's higher than math.MaxInt32, then return an error?

pkg/components/state/pluggable.go
ItalyPaleAle: Ok how about:

- If error is `Unimplemented`, return -1
- If there's another error, log it as a warning

elena-kolevska: And what do we return if there is another error? Still a -1?

pkg/components/state/pluggable.go
elena-kolevska: So that's the thing, it's the same issue as above; we're bound by the interface and can't return am error. So maybe just log it, as you suggested above, and return math.MaxInt32?

tests/integration/framework/process/statestore/statestore.go
elena-kolevska: `opts.queryFunc` and `opts.transactionalStoreMultiMaxSizeFn` would only have been set if `opts.statestore` was created through the `NewQuerier` and `NewTransactionalMultiMaxSize` factories. These factories return `WrappedQuerier` and `WrappedTransactionalMultiMaxSize` respectively, which implement default/dummy `Query` and `MultiMaxSize` methods, so a check can be possibly seen as redundant.

pkg/api/errors/state.go
```diff
@@ -0,0 +1,114 @@
...
+package errors
+
+import (
+	"fmt"
+	"net/http"
+	"strconv"
+
+	"google.golang.org/grpc/codes"
+
+	"github.com/dapr/kit/errors"
```
cicoyle, line 23: for consistency, maybe consider making all the kit errors imports be `kitErrors`. I went with this approach in my draft PR. that way there are 2 distinct names that are also standard and easily identifiable: `kitErrors` and `apiErrors`
```suggestion
	kitErrors "github.com/dapr/kit/errors"
```

pkg/components/state/pluggable.go
elena-kolevska: Done

pkg/components/state/pluggable.go
elena-kolevska: Ok, added a log for any error, but we continue to return -1

elena-kolevska: @ItalyPaleAle  Can we consider this resolved? :)

pkg/components/state/pluggable.go
```diff
@@ -321,6 +331,46 @@ func (ss *grpcStateStore) Multi(ctx context.Context, request *state.Transactiona
...
+	if err != nil {
+		log.Error("failed to get multi max size from state store", err)
+		ss.multiMaxSize = ptr.Of(-1)
+		return *ss.multiMaxSize
+	}
+
+	// If the pluggable component is on a 64bit system and the dapr runtime is on a 32bit system,
+	// the response will be larger than the maximum int32 value.
+	// In this case, we set the max size to the maximum possible value for a 32bit system.
+	is32bitSystem := unsafe.Sizeof(int(0)) == 4
```
daixiang0, line 363: Do we need this temp value?

elena-kolevska: Considering this is a local variable that's allocated on the stack, I think the extra readability is worth it, because it's making the code base more accessible to contributors from all levels. Happy to change it if that's not something that's very relevant at this point though.

pkg/http/responses_fasthttp.go
```diff
@@ -78,6 +80,14 @@ func universalFastHTTPErrorResponder(reqCtx *fasthttp.RequestCtx, err error) {
 		return
 	}
 
+	// Check if it's a kitErrors.Error object
+	var kitErr kiterrors.Error
+	ok = errors.As(err, &kitErr)
```
cicoyle, line 85:
```suggestion
	standardizedErr, ok := kitErrors.FromError(err)
```

I think we can use the `FromError()` here and then pass `standardizedErr` to the `fasthttpRespond`

elena-kolevska: Ah, thanks! I knew there were more places where we were doing this, but couldn't find it. Fixed.

elena-kolevska:
@ItalyPaleAle  @JoshVanL Do you feel like this is ready to be merged?

JoshVanL:
Hi @elena-kolevska, I'll take another look later today or tomorrow

pkg/api/errors/README.md
```diff
@@ -0,0 +1,56 @@
...
+- **Dynamically Constructed Errors**: These errors are created on the fly within the code, typically to handle specific situations or errors that are not covered by the predefined set.
+    
+As we move predefined errors to the rich error model and a new location (`pkg/api/errors`), the first step in updating to the new error model is to familiarize yourself with the existing error handling patterns, both predefined and dynamically constructed. This understanding will be crucial when replacing them with the richer error model aligned with gRPC standards.
+
+#### Step 2: Adding New Errors
+
+1. **Check for existing errors**: Before creating a new error, check if a relevant error file exists in `/pkg/api/errors/`.
+2. **Create a new error file**: If no relevant file exists, create a new file following the pattern `<building block>.go`.
```
cicoyle, line 22: nit:
```suggestion
2. **Create a new error file**: If no relevant file exists, create a new file following the pattern `<building-block>.go`.
```

pkg/api/errors/messages.go
```diff
@@ -0,0 +1,19 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package errors
+
+const (
+	// StateTransaction
+	MsgStateTransactionsNotSupported = "state store %s doesn't support transaction"
```
cicoyle, line 18: This might make more sense to live in the `api/errors/state.go`? I doubt this will be reused by other building blocks, so I don't think it should live in it's own generic file.

pkg/api/errors/state.go
```diff
@@ -0,0 +1,114 @@
...
+		Build()
+}
+
+/**** Transactions ****/
+
+func StateStoreTransactionsNotSupported(storeName string) error {
+	return kiterrors.NewBuilder(
+		codes.Unimplemented,
+		http.StatusInternalServerError,
+		fmt.Sprintf(MsgStateTransactionsNotSupported, storeName),
```
cicoyle, line 66: Actually, I think this const should just live here in this func since it's only used once as far as my ctrl-f can find

cicoyle (dismissed):
I left a comment, otherwise looks great!

JoshVanL (changes requested):

JoshVanL (dismissed):

pkg/api/errors/state.go
```diff
@@ -0,0 +1,114 @@
...
+		Build()
+}
+
+/**** Transactions ****/
+
+func StateStoreTransactionsNotSupported(storeName string) error {
+	return kiterrors.NewBuilder(
+		codes.Unimplemented,
+		http.StatusInternalServerError,
+		fmt.Sprintf("state store %s doesn't support transaction", storeName),
```
ItalyPaleAle, line 66: Nit:

```suggestion
		fmt.Sprintf("state store %s doesn't support transactions", storeName),
```

pkg/api/errors/state.go
```diff
@@ -0,0 +1,114 @@
...
+		WithErrorInfo(kiterrors.CodePrefixStateStore+"QUERYING_"+kiterrors.CodeNotSupported, nil).
+		WithResourceInfo("state", storeName, "", "").
+		Build()
+}
+
+func StateStoreQueryFailed(storeName string, detail string) error {
+	return kiterrors.NewBuilder(
+		codes.Internal,
+		http.StatusInternalServerError,
+		fmt.Sprintf("state store %s query failed: %s", storeName, detail),
```
ItalyPaleAle, line 108: Do we need the name of the state store? It should be in the resource info so probably redundant here.
```suggestion
		fmt.Sprintf("query failed: %s", storeName, detail),
```

pkg/components/state/pluggable.go
```diff
@@ -321,6 +330,46 @@ func (ss *grpcStateStore) Multi(ctx context.Context, request *state.Transactiona
...
+
+	if multiMaxSize != nil {
+		return *multiMaxSize
+	}
+
+	// In a worst case scenario, two or more callers may get to this point,
+	// meaning that we unnecessarily call MultiMaxSize multiple ties,
+	// however this is functionally not a problem and the cost is minimal.
+	ss.lock.Lock()
+	defer ss.lock.Unlock()
```
ItalyPaleAle, line 347: The fix to this "problem" is pretty simple, just check if multiMaxSize is non-nil after acquiring the write lock

```suggestion
	ss.lock.Lock()
	defer ss.lock.Unlock()

	// Check the cached value again in case another goroutine set it
    if multiMaxSize != nil {
		return *multiMaxSize
	}
```

pkg/components/state/pluggable.go
```diff
@@ -321,6 +330,46 @@ func (ss *grpcStateStore) Multi(ctx context.Context, request *state.Transactiona
...
+
+	resp, err := ss.Client.MultiMaxSize(ctx, new(proto.MultiMaxSizeRequest))
+	if err != nil {
+		log.Error("failed to get multi max size from state store", err)
+		ss.multiMaxSize = ptr.Of(-1)
+		return *ss.multiMaxSize
+	}
+
+	// If the pluggable component is on a 64bit system and the dapr runtime is on a 32bit system,
+	// the response will be larger than the maximum int32 value.
```
ItalyPaleAle, line 360: Nit:
```suggestion
	// the response could be larger than the maximum int32 value.
```

ItalyPaleAle (changes requested):
A few more small comments but otherwise LGTM

pkg/api/errors/state.go
elena-kolevska: That makes sense. Personally, I prefer the redundancy here though, because existing applications that haven't implemented the new error handling will still get meaningful error messages.

elena-kolevska:
> A few more small comments but otherwise LGTM

Thanks @ItalyPaleAle 🙏  Addressed.

ItalyPaleAle (approved):

artursouza (approved):

PR #7260, opened 2023-11-30
Hot Reloading: Kubernetes Component hot reloading

JoshVanL:
Branched from https://github.com/dapr/dapr/pull/7239

Part of #1172

Adds hot reloading functionality to Daprd, which is gated behind the HotReload preview feature.

If enabled, Daprd in Kubernetes mode will watch for Component resource
updates from the Operator. If a resource is detected as changed (spec is
different to what is currently loaded), then Dapr will
close/create/update the resource.

Supports Component _except_ HTTP middlware which requires further
changes to the HTTP server handler machinery. A warning is thrown if hot
reloading is enabled and a HTTP middleware is updated.

---

When running in Kubernetes mode, the operator loader streams updates from
the Operator API. When an event occurs, the event type will trigger the
resource to be created, updated, or deleted. Updated components will be
closed, and then re-initialized.

We consider a resource to have changed generally if anything apart from
its Kubernetes metadata or type meta object meta has changed, and
therefore needs some kind of reloading.

---

Currently, if a reloading component errors and `spec.ignoreErrors=false`
then Daprd will gracefully close, like a component loaded at startup
today. It is intended that in future the component will be re-inited on
a backoff queue in future in the case of errors, even if
`spec.ignoreErros=true`.

HTTP middleware component hot reloading is not supported as it requires
further changes to the HTTP server handler machinery, and I didn't want
to grow this PR further.

To use the HotReloading feature, which is currently only available as an
alpha feature in Selfhosted mode, users need to add the following
Configuration to the target Daprd;

```yaml
apiVersion: dapr.io/v1alpha1
kind: Configuration
metadata:
  name: hotreloading
spec:
  features:
  - name: HotReload
    enabled: true
```


pkg/config/configuration.go
```diff
@@ -1,5 +1,6 @@
 /*
-Copyright 2021 The Dapr Authors
+Copyright 2GetConfiguration
+021 The Dapr Authors
```
ItalyPaleAle, line 3: Please revert
```suggestion
Copyright 2021 The Dapr Authors
```

pkg/runtime/hotreload/loader/operator/operator.go
```diff
@@ -0,0 +1,57 @@
...
+	}
+}
+
+func (o *operator) Close() error {
+	errs := make([]error, 0, 2)
+	if err := o.component.close(); err != nil {
+		errs = append(errs, err)
+	}
+
+	return errors.Join(errs...)
```
ItalyPaleAle, line 52:
```suggestion
	return o.component.close()
```

pkg/runtime/hotreload/loader/operator/resource.go
```diff
@@ -0,0 +1,167 @@
...
+}
+
+func (r *resource[T]) stream(ctx context.Context, eventCh chan<- *loader.Event[T]) {
+	for {
+		for {
+			event, err := r.streamer.recv()
+			if err != nil {
+				r.streamer.close()
+				// Retry on stream error.
+				log.Errorf("error from operator stream: %s", err)
```
ItalyPaleAle, line 128: Nit: consistency with other logs
```suggestion
				log.Errorf("Error from operator stream: %v", err)
```

pkg/runtime/hotreload/loader/operator/resource.go
```diff
@@ -0,0 +1,167 @@
...
+				return
+			}
+		}
+
+		for {
+			select {
+			case <-ctx.Done():
+				return
+			default:
+			}
```
ItalyPaleAle, line 144:
```suggestion
            if ctx.Err() != nil {
                return
            }
```

pkg/runtime/runtime.go
```diff
@@ -196,7 +196,18 @@ func newDaprRuntime(ctx context.Context,
...
+			Namespace:      namespace,
+			Client:         operatorClient,
+			Config:         globalConfig,
+			ComponentStore: compStore,
+			Authorizer:     authz,
+			Processor:      processor,
+		})
+		if err != nil {
+			return nil, err
+		}
```
ItalyPaleAle, line 210: There's no err in this block
```suggestion
```

ItalyPaleAle (changes requested):

pkg/runtime/hotreload/loader/operator/resource.go
```diff
@@ -0,0 +1,165 @@
...
+
+	return &differ.LocalRemoteResources[T]{
+		Local:  r.store.List(),
+		Remote: remotes,
+	}, nil
+}
+
+func (r *resource[T]) Stream(ctx context.Context) (<-chan *loader.Event[T], error) {
+	if r.closed.Load() {
+		return nil, errors.New("stream is closed")
```
yaron2, line 93: Who is responsible to re-establish the stream if its closed?

JoshVanL: This is a sanity check to return early if we call `Stream` after the resource streamer manager has been closed. Here we retry forever with an exponential backoff until the context is cancelled to reconnect to the operator.

https://github.com/dapr/dapr/pull/7260/files#diff-1e05ef0dac6089bfebe37de39be1346f5722fb7c8c88c78f686d8c77150be98eR139-R155

yaron2 (approved):

PR #7264, opened 2023-12-01
Change target qps for actor reminder test

yaron2:
The move to PostgreSQL for actor tests reduced the actual QPS from +50 to ~38. This PR changes the target QPS accordingly so performance tests can run to completion again.


yaron2:
> Note that #7129 is changing it too, _increasing_ it to 56 (conservatively)

That's great, I'll close this then.

Issue #4313, opened 2022-03-01
allow annotation to specify shutdown pause


trondhindenes:
## In what area(s)?

/area runtime

## Describe the feature
We're struggling with making dapr lifecycle robust, especially during shutdown. We have some pods that need a few seconds (up to 30) to shutdown, and during those 30 seconds, they require the dapr sidecar to be available so that messages can be sent. However, since these pods are in the "draining" state, we don't want them to receive more traffic

I propose that a new annotation is added: `dapr.io/shutdown-wait-secs`. If specified, the sidecar is kept alive for the specified number of seconds, but only for outbound traffic (e.g. it will not pick up new messages etc).
This annotation could be used to setup a `preStop` hook on the injected container and do something like
`command: ["/bin/sh", "-c", "/bin/dapr -c prepShutdown && sleep 30 && /bin/dapr -c doShutdown"]`
The point is that there needs to be a way to instruct dapr to go into "passive" mode where it will still forward messages to the queue etc, but not receive anything inbound. Without this it's very difficult to build a robust messaging system using dapr.

trondhindenes:
I tried to use the dapr _service_ instead of the sidecar address in the meantime, but at least the health endpoint (`curl http://<app-id>-dapr:80/v1.0/healthz`) just gives a http 400 back :-(

artursouza:
Summarizing the approach:

Stops new incoming messages right away for PubSub, Binding and Service Invocation.
Continue allowing outgoing API calls during grace period for all APIs.
Allow this to be configurable via K8s annotation.

CrazyHZM:
I found `dapr.io/graceful-shutdown-seconds` to meet this requirement, can you use `dapr.io/graceful-shutdown-seconds` to try to solve your problem? @trondhindenes

yaron2:
cc @trondhindenes

trondhindenes:
sorry, I was completely out of the loop here. Yes I believe it does.
There is a situation which I believel we aren't fully covering:

Imagine a pubsub scenario where a messagehandler processes messages, and for each processed message, it needs to send a message to another queue (confirmation message). Each message might take between 10 seconds and 10 minutes to complete (this is real-life btw, we used dapr to respond to uploaded audio files that would vary a lot in file size and thus processing them also varied a lot).

[...]

PR #7268, opened 2023-12-05
Adds Daprd option `--dapr-block-shutdown-duration`

JoshVanL:
Closes https://github.com/dapr/dapr/issues/4313
Docs: https://github.com/dapr/docs/pull/3893

PR adds the `--dapr-block-shutdown-duration` CLI flag and corresponding `dapr.io/block-shutdown-duration` Kubernetes annotation which configures Daprd to block the graceful shutdown procedure until _either_, the block shutdown duration has elapsed _or_ the application has become unhealthy, as according to the normal app health status.

Accepts a Go time duration string.

By default, this option is unset, and therefore there is no effect to the current behaviour of graceful shutdown. When set, Daprd will block the interrupt signal cascading into runtime until the above requirements have been met.

The framework process `Cleanup` order has been reversed as a stack to mimic `t.Cleanup` and allow the `logline` process to function correctly for exec's which don't exit before `Cleanup`.


artursouza (approved):

cmd/daprd/options/options.go
```diff
@@ -109,6 +112,7 @@ func New(args []string) *Options {
 	flag.StringVar(&opts.UnixDomainSocket, "unix-domain-socket", "", "Path to a unix domain socket dir mount. If specified, Dapr API servers will use Unix Domain Sockets")
 	flag.IntVar(&opts.DaprHTTPReadBufferSize, "dapr-http-read-buffer-size", runtime.DefaultReadBufferSize, "Increasing max size of read buffer in KB to handle sending multi-KB headers")
 	flag.IntVar(&opts.DaprGracefulShutdownSeconds, "dapr-graceful-shutdown-seconds", int(runtime.DefaultGracefulShutdownDuration/time.Second), "Graceful shutdown time in seconds")
+	flag.IntVar(&blockShutdownSeconds, "dapr-block-shutdown-seconds", 0, "If enabled, will block graceful shutdown after terminate signal is received until either the number of seconds has been reached or the app reports unhealthy. Disabled by default")
```
ItalyPaleAle, line 115: Please don't use "seconds" or other things with complex units. We should standardize on Go durations which allow more flexibility, at least for new flags. See #7028

ItalyPaleAle (changes requested):

cmd/daprd/options/options.go
yaron2: On the other hand, it might be more familiar for K8s users with `terminationGracePeriodSeconds`.

artursouza (dismissed):

artursouza:
@JoshVanL Please, look at the failing unit and integration tests.

yaron2 (approved):

PR #7270, opened 2023-12-05
Fix timeouts in HTTP service invocation when resiliency policies with timeouts are applied

### LINKED ISSUE #7145: DAPR 1.12 cross-service REST call cuts response body
## In what area(s)?

 /area runtime

## What version of Dapr?

> 1.12 has bug. Previous version s 1.11.2 and 1.11.5 are stable

## Expected Behavior

We may reproduce the on docker compose when first service calls second service rest API through dapper.
 Producer service uses ASP.NET minimal REST API (F# language) to provider json response from object serialization
 
image

Consumer service is DAPR actor that calls producer to download JSON.
image

## Actual Behavior

See attached file with interrupted JSON stream in the end where error starts ".....Valid JSON stream and then unexpectedly: "{"errorCode":"ERR_DIRECT_INVOKE","message":"failed to invoke, id: expressport-adoconnector, err: error receiving message: rpc error: code = Canceled desc = context canceled"}"

On 16 core machine with 64 GB ram from actor call DAPR another service by REST and will see that sometimes response is cut with error in the end.

### LINKED ISSUE #7173: 1.12 service invocation failures with resiliency timeout policies and large responses
## In what area(s)?

/area runtime

> /area operator

> /area placement

> /area docs

> /area test-and-release

## What version of Dapr?

- 1.12.0
- edge: v1.11.3-336-g78b7271f-dirty

> 1.1.x
> 1.0.x

## Expected Behavior

Invocation of method with 10-100MB response succeeds.

## Actual Behavior

Invocation of method fails returning error:

`{"errorCode":"ERR_DIRECT_INVOKE","message":"failed to invoke, id: back-end, err: error receiving message: rpc error: code = Canceled desc = context canceled"}`

Notes:

 - Invoking the application HTTP endpoint directly succeeds
 - Invoking the method via the "back-end" application's Dapr sidecar HTTP endpoint succeeds 
 - Remove the resiliency policy and method invocation via the "front-end" Dapr sidecar will then succeed
 - Invocation succeeds on previous version of Dapr runtime (1.11.x)
 - Timing (i.e. hardware) reportedly makes a difference (reproduces consistently on 16+ vCore CPU, 32+ GB RAM)

Some earlier investigation seems to point to the "front-end" Dapr sidecar canceling the request context while the "back-end" Dapr sidecar is still streaming its response.  The original issue was with .NET-based applications using the Dapr .NET SDK, but it also reproduces with a Node.js application not using *any* SDK.

[dapr-bug.zip](https://github.com/dapr/dapr/files/13323169/dapr-bug.zip)

1. Expand attached repro application
2. `npm install`
3. `dapr run -f ./dapr.yaml`
4. Execute HTTP request `POST http://localhost:50050/v1.0/invoke/back-end/method/bugReproduce`

This invokes the `bugReproduce` method of the "back-end" application via the "front-end" Dapr sidecar, and should fail.

Notes:

 - Make the same request on port `55050` (the "back-end" application's Dapr sidecar) and the request should succeed.
 - Make the same request on port `3000` (the "back-end" applications direct endpoint) and the request should succeed.
 - Remove the resiliency policy targeting calls to the "back-end" application and all the requests should succeed.

## Release Note

RELEASE NOTE:

ItalyPaleAle:
Fixes #7173
Fixes #7145

When resiliency policies with a timeout are applied, HTTP service invocation could fail with a truncated response and this error:

```
{"errorCode":"ERR_DIRECT_INVOKE","message":"failed to invoke, id: back-end, err: error receiving message: rpc error: code = Canceled desc = context canceled"}
```

This was reproducible mostly on faster machines, indicating it was a timing issue.

The root cause of the issue was due to the fact that in HTTP service invocation, we pass to the `Invoke` method a context that is tied to the resiliency policy runner's context. This meant that when the policy function returns, the context passed to the `Invoke` method is canceled, even if the request has not timed out. However, with streaming, the context has to remain valid outside of the policy runner function, to be able to respond to the caller.

The fix involves sending the response within the policy function. This way, the context is still valid throughout that time, and the timeout applies to the entire response.

The fix was validated with the repro code in #7173, which at least on my M1 Pro Mac allowed reproducing the error reliably.

It is sadly not possible to write a test for this, because it was a timing bug that is hard to reproduce deterministically (and it mostly appears on fast machines, so not on GH Actions). In fact, our E2E tests already have test cases for service invocation with resiliency, and never detected this issue.

> Additionally, this PR fixes a race condition in `pkg/resiliency/policy.go`, which could have caused a goroutine leak (due to writing on a channel with no reader) if an operation completed _immediately_ after the timeout.

### DIFF
--- a/pkg/http/api_directmessaging.go
+++ b/pkg/http/api_directmessaging.go
@@ -23,7 +23,9 @@ import (
 	"net/url"
 	"path"
 	"strings"
+	"sync/atomic"
 
+	"github.com/cenkalti/backoff/v4"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 
@@ -32,7 +34,6 @@ import (
 	"github.com/dapr/dapr/pkg/messages"
 	invokev1 "github.com/dapr/dapr/pkg/messaging/v1"
 	"github.com/dapr/dapr/pkg/resiliency"
-	"github.com/dapr/dapr/pkg/resiliency/breaker"
 )
 
 // directMessagingSpanData is the data passed by the onDirectMessage endpoint to the tracing middleware
@@ -150,6 +151,7 @@ func (a *api) onDirectMessage(w http.ResponseWriter, r *http.Request) {
 			Disposer: resiliency.DisposerCloser[*invokev1.InvokeMethodResponse],
 		},
 	)
+	success := atomic.Bool{}
 	// Since we don't want to return the actual error, we have to extract several things in order to construct our response.
 	resp, err := policyRunner(func(ctx context.Context) (*invokev1.InvokeMethodResponse, error) {
 		rResp, rErr := a.directMessaging.Invoke(ctx, targetID, req)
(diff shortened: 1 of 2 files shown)


pkg/http/api_directmessaging.go
```diff
@@ -198,63 +200,76 @@ func (a *api) onDirectMessage(w http.ResponseWriter, r *http.Request) {
...
+		if !success.CompareAndSwap(false, true) {
+			// This error will never be returned to a client but it's here to prevent retries
+			return rResp, backoff.Permanent(errors.New("already completed"))
+		}
 
-	var codeErr codeError
-	if errors.As(err, &codeErr) {
-		if len(codeErr.headers) > 0 {
-			invokev1.InternalMetadataToHTTPHeader(r.Context(), codeErr.headers, w.Header().Add)
+		if rResp == nil {
```
philliphoff, line 211: Are multiple instances of the function running in parallel? Otherwise, how does execution get to 211 if the first execution returns an error that prevents retries?  (I'm sure it's just my lack of experience with Go and the policy runner APIs.)

ItalyPaleAle: Yes, multiple instances of the function _can_ be running in parallel if one of them times out. When there's a timeout the policy runner cancels the context, which is a request (not an order) to stop processing; at the same time, it invokes the function again.

The reason why we don't have issues with concurrency is the lines just above:

```go
if !success.CompareAndSwap(false, true) {
	return rResp, backoff.Permanent(errors.New("already completed"))
}
```

This code uses an atomic compare-and-swap:

- if `success` is false (initial value), then its value is changed to true (atomically) and the function returns true
- if `success` is already true, then the function returns false, so we return a permanent error

In all the lines below that, we return all errors as `backoff.Permanent`, which makes the policy runner not retry in case of errors.

If the retry was due to a timeout, instead, it would hit the compare-and-swap and return right away. In this case we can't go past this `if` block because we have already started sending data to the client.

daixiang0 (approved):
LGTM

berndverst (approved):

yaron2:
@ItalyPaleAle before we merge, can you please look into intg tests failing? I re-ran several times

ItalyPaleAle:
> @ItalyPaleAle before we merge, can you please look into intg tests failing? I re-ran several times

It looks like tests that fail are mostly actors-related, the same ones that are failing in master. I've re-run the unit tests, but they seemed unrelated too.

Is there any test in particular I should be concerned with?

PR #7272, opened 2023-12-06
Updates dapr/kit and dapr/components-contrib to HEAD

JoshVanL:
Updates dapr/kit and dapr/components-contrib to HEAD. This primarily includes updating the fswatcher from the old `fswatcher.Watch` to the new `fs, err := fswatcher.New(fswatcher.Options)` & `fs.Run` pattern.

`pkg/security` Provider now includes a `fswatcherInterval` option which is used as the trust bundle fswatcher interval duration. This is the interval to batch file events for the same file in fswatcher. This is always set to 500ms except for tests where it is set to 0 to speed up the test run. We want to keep this at 500ms as the live value to reduce CPU & IO usage on trust bundle file event churn which occurs on Kubernetes volumes or some text editors on "single" write events.

https://ahmet.im/blog/kubernetes-inotify/
https://www.extrema.is/blog/2022/03/04/vim-and-inotify

---

With these changes not in master, other developers wanting to make changes to dapr/dapr with an updated dapr/kit branch have to write these changes themself which is a pain & adds noise to PRs. I think moving forward, we should make an effort to ensure any updates to dapr/kit & dapr/components-contrib, particularly breaking changes, are updated in the dapr/dapr go.mod and merged into master as soon as possible. We can make the go.mod dapr/dapr PR a prereq like we do for docs PRs on new Dapr features.

### DIFF
--- a/cmd/sentry/main.go
+++ b/cmd/sentry/main.go
@@ -128,11 +128,16 @@ func main() {
 	}
 
 	// Watch for changes in the watchDir
-	err = mngr.Add(func(ctx context.Context) error {
-		log.Infof("Starting watch on filesystem directory: %s", watchDir)
-		return fswatcher.Watch(ctx, watchDir, issuerEvent)
+	fs, err := fswatcher.New(fswatcher.Options{
+		Targets: []string{watchDir},
 	})
 	if err != nil {
+		log.Fatal(err)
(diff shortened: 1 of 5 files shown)


daixiang0 (approved):
LGTM

PR #7276, opened 2023-12-06
Integration: adds daprd/resources/namespace test

JoshVanL:
Integration adds a test to ensure that components loaded from disk which are a different namespace to that of Daprd are ignored.

Note that for an empty namespace, Daprd assumes the resource is for it's own namespace rather than 'default'.


tests/integration/suite/daprd/resources/namespace.go
```diff
@@ -0,0 +1,111 @@
...
+  name: ghi
+  namespace: notmynamespace
+spec:
+  type: state.in-memory
+  version: v1
+---
+apiVersion: dapr.io/v1alpha1
+kind: Component
+metadata:
+  name: abc
```
mukundansundar, line 64: nitpick change name to something else.

cicoyle: It might be clearer if you moved up the `name: abc` w/ the `namespace: notmynamespace` right below the first one with `name: abc` but no namespace to prove your point in a more reader friendly manner, otherwise we see `name: ghi` which is next, but not sequentially next in the alphabet and `def` later which is just out of order and then takes more time to look at to figure out what you are trying to show with these examples.

tests/integration/suite/daprd/resources/namespace.go
```diff
@@ -0,0 +1,111 @@
...
+	"github.com/dapr/dapr/tests/integration/framework/process/daprd"
+	"github.com/dapr/dapr/tests/integration/framework/process/exec"
+	"github.com/dapr/dapr/tests/integration/suite"
+)
+
+func init() {
+	suite.Register(new(namespace))
+}
+
+// namespace ensures that the component disk loader does not load components
```
cicoyle, line 36: Is this comment right? maybe you can add the comment about same name, diff namespace is ok here too?

cicoyle (dismissed):
Minor comment about the test yaml, otherwise LGTM

mukundansundar (approved):

tests/integration/suite/daprd/resources/namespace.go
JoshVanL: Yes, this comment is correct- the component disk loader should only load `Component` manifests which are configured for the same namespace as the `daprd`. I think same name different namespace is implicit as the component in the different namespace will not be loaded.

tests/integration/suite/daprd/resources/namespace.go
JoshVanL: Done

mukundansundar (approved):

artursouza (approved):

Issue #7203, opened 2023-11-16
Design the interface that allows multiple versions of Reminders/ Actors implementation


rabollin:
This is a placeholder to add a design proposal for Interface that allows multiple reminders/ actior implementation(versions) in-tree and implement it in 1.13 once the design proposal is approved

DeepanshuA:
Based on the discussion that I had with Alessandro (and I think he had with Yaron), I have put together this:

## Interfaces for Actors and Reminders 

There are two places where APIs or functions required for Actors and Reminders are defined. 

1. Protos in dapr/proto package. 

2. Interfaces in pkg/actors/internal for placement and reminders to be updated 

Out of these 2, protos are not to be updated, as every implementation will have their own RPCs defined accordingly. 

Only interfaces in pkg/actors/internal have to be updated. Though, a major work has already been done for them in release 1.12 already by segregating actors, reminders and timers interfaces (https://github.com/dapr/dapr/pull/6669 ). 

![Screenshot 2023-12-12 at 11 05 47 PM](https://github.com/dapr/dapr/assets/24491296/c7ea777b-9f96-4760-a414-a9678a587ed5)

 

 
![Screenshot 2023-12-12 at 11 06 57 PM](https://github.com/dapr/dapr/assets/24491296/e2ba1c27-78d2-45d3-9c0b-87a72818db60)


 

 

Idea behind this is that interfaces defined in pkg/actors/internal will be used to form a Building Block / Client which will implement all these methods defined in these interfaces. 

Whereas, methods defined for Services in protos will be implemented by respective new/existing Control Plane Service(s). 

And, the methods from these clients will call methods of the servers, wherever required. 

 
![Screenshot 2023-12-12 at 11 07 26 PM](https://github.com/dapr/dapr/assets/24491296/3950816b-4eae-4673-9e47-833fe23ffe37)

 

 

Depending on the implementation, it can be decided that what all RPCs need to be defined for a Service. Example, for a Placement Service which just creates and pushes Actor Types v/s Dapr Hosts table information to All sidecars, only table related `ReportDaprStatus`needs to be as rpc. 

Whereas, for a Service, where Actor information is Pulled by sidecars from Actors Service needs some methods like LookupActor or ReportActorDeactivation, which thus form a part of RPCs. 

 

For some minor changes left in interfaces, https://github.com/dapr/dapr/pull/7281 has been raised. 

 

### How to build and deploy Actors/Scheduler/Placement Service? 

 

[...]

PR #7281, opened 2023-12-07
Standardize initialization actor services: Placement and Reminders

ItalyPaleAle:
Part of #7203

This is some initial work towards #7203 and it involves standardizing how objects that provide actors services are initialized: Placement and Reminders. There are no functional changes.

1. All objects receive the same struct with properties that are used by both placement and reminder providers
2. All objects are initialized in the `newActorsWithClock` method, while before placement was initialized in `Init`


pkg/actors/actors.go
```diff
@@ -1111,16 +1118,16 @@ func ValidateHostEnvironment(mTLSEnabled bool, mode modes.DaprMode, namespace st
 	return nil
 }
 
-func (a *actorsRuntime) stateStore() (internal.TransactionalStateStore, error) {
+func (a *actorsRuntime) stateStore() (string, internal.TransactionalStateStore, error) {
```
daixiang0, line 1121: Is there any reason for this? Get name for runtime looks good.

ItalyPaleAle: The caller of this method is also in a separate package (pkg/actor/reminders), so this avoids passing a new property to the factory method.

daixiang0: OK.

pkg/actors/internal/reminders.go
```diff
@@ -31,12 +29,10 @@ type ExecuteReminderFn func(reminder *Reminder) bool
 // LookupActorFn is the type of a function that returns whether an actor is locally-hosted and the address of its host.
 type LookupActorFn func(ctx context.Context, actorType string, actorID string) (isLocal bool, actorAddress string)
 
-// StateStoreProviderFn is the type of a function that returns the state store provider.
-type StateStoreProviderFn func() (TransactionalStateStore, error)
+// StateStoreProviderFn is the type of a function that returns the state store provider and its name.
+type StateStoreProviderFn func() (string, TransactionalStateStore, error)
```
DeepanshuA, line 33: till now, we were using state store name from ActorStateStoreName - is deriving this data from StateStoreProviderFn necessary?

pkg/actors/actors.go
```diff
@@ -1028,7 +1034,7 @@ func (a *actorsRuntime) RegisterInternalActor(ctx context.Context, actorType str
 		log.Debugf("Registering internal actor type: %s", actorType)
 		actor.SetActorRuntime(a)
 		a.actorsConfig.Config.HostedActorTypes.AddActorType(actorType, actorIdleTimeout)
-		if a.placement != nil {
+		if a.placementEnabled {
```
DeepanshuA, line 1037: if placementEnabled is true but if placement is not initialised properly, can it be an issue here?

pkg/actors/internal/reminders.go
ItalyPaleAle: It allows not having to pass another option to `ActorsProviderOpts` since it is only needed by local providers. See: https://github.com/dapr/dapr/pull/7281#discussion_r1421879007

pkg/actors/internal/opts.go
```diff
@@ -0,0 +1,41 @@
+/*
```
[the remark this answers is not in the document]
ItalyPaleAle: Done

pkg/actors/actors.go
ItalyPaleAle: I don't think it should be. Before, `a.placement` was `nil` if placement was disabled. Now, it could be non-nil, but `a.placementEnabled` will be false. See: https://github.com/dapr/dapr/pull/7281/files#diff-aeb183ea3b81464cb579aa3e309be6bbfa88d6d900cfdbddb8c74d89ee7d4dc6L242-L243

DeepanshuA (approved):

mukundansundar (approved):

daixiang0 (approved):
LGTM

PR #7283, opened 2023-12-08
Configure alternate backend for Dapr workflow

ASHIQUEMD:
# Description

Allow users to configure alternate backend for Dapr Workflow apart from Actor.

Currently supported workflow backends
1. Actor
2. Sqlite

This PR refactors wfengine.go file to be independent of Actor backend.

SQLite backend component example yaml

```yaml
apiVersion: dapr.io/v1alpha1
kind: Component
metadata:
  name: sqlitebackend
spec:
  type: workflowbackend.sqlite
  version: v1
  metadata:
    - name: filePath
      value: "file::memory:"
    - name: orchestrationLockTimeout
      value: "130000000000ms"
    - name: activityLockTimeout
      value: "130000000000ms"
```

Actor backend component example yaml.

```yaml
apiVersion: dapr.io/v1alpha1
kind: Component
metadata:
  name: actorbackend
spec:
  type: workflowbackend.actor
  version: v1
```

Note: If there is no backend component defined, actor backend will be initialized by default, so its not required for user to specify the actor backend component.

## Issue reference

Please reference the issue this PR will close: #7127

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Unit tests passing
* [x] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: [dapr/docs#_[3924]_](https://github.com/dapr/docs/issues/3924)
* [x] Specification has been updated / Created issue in the https://github.com/dapr/docs/ repo: [dapr/docs#_[3924]_](https://github.com/dapr/docs/issues/3924)
* [x] Provided sample for the feature / Created issue in the https://github.com/dapr/docs/ repo: [dapr/docs#_[3924]_](https://github.com/dapr/docs/issues/3924)


pkg/runtime/wfengine/backendFactory.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+
+	if orchestrationLockTimeout, ok := backendComponentInfo.WorkflowBackendMetadata.Properties["orchestrationLockTimeout"]; ok {
+		orchestrationLockTimeout, err := time.ParseDuration(orchestrationLockTimeout)
+		sqliteOptions.OrchestrationLockTimeout = orchestrationLockTimeout
+
+		if err != nil {
+			sqliteOptions.OrchestrationLockTimeout = orchestrationLockTimeout
+		} else {
+			log.Errorf("invalid orchestrationLockTimeout provided in backend workflow component: %v", err)
```
cgillum, line 59: It looks like the error handling logic is reversed here and further below.

pkg/runtime/wfengine/wfengine_test.go
```diff
@@ -340,9 +345,11 @@ func TestRecreateCompletedWorkflow(t *testing.T) {
 func TestInternalActorsSetupForWF(t *testing.T) {
 	ctx := context.Background()
 	_, engine := startEngine(ctx, t, task.NewTaskRegistry())
-	assert.Len(t, engine.GetInternalActorsMap(), 2)
-	assert.Contains(t, engine.GetInternalActorsMap(), workflowActorType)
-	assert.Contains(t, engine.GetInternalActorsMap(), activityActorType)
+	if abe, ok := engine.Backend.(*wfengine.ActorBackend); ok {
```
cgillum, line 348: I think we need to assert/require that the cast succeeded since the test is otherwise invalid.

pkg/runtime/runtime.go
```diff
@@ -582,10 +586,26 @@ func (a *DaprRuntime) appHealthReadyInit(ctx context.Context) (err error) {
...
+
 func (a *DaprRuntime) initWorkflowEngine(ctx context.Context) {
 	wfComponentFactory := wfengine.BuiltinWorkflowFactory(a.workflowEngine)
 
-	a.workflowEngine.SetActorRuntime(a.actor)
+	if a.workflowEngine.BackendType == wfengine.ActorBackendType {
+		err := setActorRuntimeForBackend(ctx, a.workflowEngine.Backend, a.actor)
+
+		if err != nil {
+			log.Errorf("could not set actor runtime for backend: %v", err)
```
cgillum, line 605: It looks like we're just swallowing the error but continuing to run. Is that going to be okay or do we need to stop the startup of the Dapr runtime?

pkg/runtime/runtime.go
```diff
@@ -556,17 +558,19 @@ func (a *DaprRuntime) appHealthReadyInit(ctx context.Context) (err error) {
...
-			a.initWorkflowEngine(ctx)
-
 			a.daprUniversalAPI.SetActorRuntime(a.actor)
 		}
 	} else {
 		// If actors are not enabled, still invoke SetActorRuntime on the workflow engine with `nil` to unblock startup
-		a.workflowEngine.SetActorRuntime(nil)
+		err = setActorRuntimeForBackend(ctx, a.workflowEngine.Backend, nil)
+		if err != nil {
+			log.Errorf("error while setting actor runtime for backend to nil: %v", err)
```
cgillum, line 567: Shouldn't we be returning an error here instead of logging?
```suggestion
			return fmt.Errorf("error while setting actor runtime for backend to nil: %w", err)
```

cgillum (commented):
I haven't had a chance to do a full review yet, but from what I've skimmed so far, the approach you're taking matches my expectations, which is great to see. :)

There are a few minor things I was able to spot while looking quickly at the code.

pkg/runtime/runtime.go
ASHIQUEMD: SetActorRuntime is simplified as it was earlier, where we only initializing the actor and nothing else. So, returning the error doesn't add any value as it always be nil. So, removing this error check.

pkg/runtime/wfengine/backendFactory.go
ASHIQUEMD: Fixed.

cmd/daprd/components/workflowbackend_sqlite.go
```diff
@@ -0,0 +1,25 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package components
+
+import (
+	"github.com/dapr/components-contrib/workflows"
+	backendLoader "github.com/dapr/dapr/pkg/components/workflowBackend"
```
ItalyPaleAle, line 20: We normally don't use uppercase letters in package names.

"workflow-backend" or just "wfbackend"?

pkg/runtime/compstore/workflowBackend.go
```diff
@@ -0,0 +1,41 @@
+/*
```
ItalyPaleAle, line 1: This file too, please don't put uppercase characters in the file name

pkg/runtime/wfengine/backendFactory.go
```diff
@@ -0,0 +1,89 @@
...
+
+func getSqliteBackend(appId string, wfe *WorkflowEngine, backendComponentInfo *workflows.WorkflowBackendComponentInfo, log logger.Logger) backend.Backend {
+	wfe.BackendType = SqliteBackendType
+	sqliteOptions := &sqlite.SqliteOptions{
+		OrchestrationLockTimeout: 2 * time.Minute,
+		ActivityLockTimeout:      2 * time.Minute,
+		FilePath:                 "",
+	}
+
+	if connectionString, ok := backendComponentInfo.WorkflowBackendMetadata.Properties["connectionString"]; ok {
```
ItalyPaleAle, line 51: Please use https://github.com/dapr/kit/tree/main/metadata to parse the metadata like we do in components-contrib. This is more complete and also supports case-insensitive property names

pkg/runtime/wfengine/backendFactory.go
```diff
@@ -0,0 +1,89 @@
...
+)
+
+const (
+	ActorBackendType  = "workflowbackend.actor"
+	SqliteBackendType = "workflowbackend.sqlite"
+)
+
+type BackendFactory func(appId string, wfe *WorkflowEngine, backendComponentInfo *workflows.WorkflowBackendComponentInfo, log logger.Logger) backend.Backend
+
+var backendFactories = map[string]BackendFactory{
```
ItalyPaleAle, line 33: Is there a way to make this allow some "drop-in" things? Such as what we do in cmd/daprd/components where we can drop a file and it register itself. Otherwise, for each new backend that dtf-go adds, it needs to be added here which is more frail.

ASHIQUEMD: @ItalyPaleAle Thank you for your suggestion! I appreciate the thought you've put into making the integration more user-friendly. After considering your idea, I wanted to share some thoughts that came to mind.

I really like the simplicity of our current implementation, particularly how adding a new backend only requires a modification in one place, specifically in backend_factory.go. It keeps things neat and focused.

While the idea of a "drop-in" feature is intriguing, I'm concerned that it might introduce additional complexity, especially considering that the number of backends is likely to be much smaller than the variety of dapr components. Keeping the initialization logic within backend_factory.go seems to maintain a straightforward and easy-to-follow structure.

I'm open to further discussion on this, and I'd love to hear your thoughts. Let me know what you think!

ItalyPaleAle: My concern is the amount of inter-dependency that is being added where there's part of the initialization in this repo, and part in dtf-go.

Ideally all of this should be moved to dtf-go. Dapr will just pass through everything to dtf-go and the initialization will happen in the upstream library. This way, adding a new backend does not require changes to dapr/dapr.

cgillum:
> Ideally, if dtf-go adds a new backend, we should be able to have that exposed in Dapr without having to make code changes in dapr itself.

@ItalyPaleAle I would push back on this _a little_. I agree this would be convenient, but on the other hand I worry about whether this creates a bit of a support problem.

By allowing Dapr to be in control over which workflow backends are used, Dapr is then able to document a set of supported workflow backends and the community is better able to help debug issues with those backends (since they were explicitly onboarded for use in a Dapr context). It also allows Dapr to provide a more Dapr-native experience when it comes to configuration or integration with other features.

If we instead open it up to anything that a separate project supports, then Dapr can't effectively document or provide 1st class integration with those workflow backends. I'm worried about the hypothetical case where durabletask-go introduces some exotic workflow backend option and customers come here complaining why it doesn't work in a Dapr context when Dapr maintainers have maybe never even heard of this exotic workflow backend.

So while I agree that it's more empowering to the user and less burden to Dapr maintainers to add support for new things, I worry that there's a hidden cost that we're not considering. I'm curious to hear your thoughts on this concern.

ItalyPaleAle:
@cgillum that's a fair point.

On one hand, however, consider that we would still need to update the pinned version of dtf-go in this repo to be able to add new backends.

On the other hand, perhaps there could be some middle ground where the initialization happens in dtf-go, and we just keep a list of allowed backends here. This should be done in a "drop-in" way like we do for components in cmd/daprd/components.

WDYT?

cgillum:
> On one hand, however, consider that we would still need to update the pinned version of dtf-go in this repo to be able to add new backends.

Yes, I'm aware of this, but Dapr will need to update this dependency periodically for other reasons too, and I'm worried about cases where new backends get picked up "accidentally".

> On the other hand, perhaps there could be some middle ground where the initialization happens in dtf-go, and we just keep a list of allowed backends here.

This would solve the support problem, which is definitely the biggest one in my mind, so I like this direction.

There's still a question of whether/how Dapr will be able to apply configuration to the selected workflow backend. I see in [this PR](https://github.com/microsoft/durabletask-go/pull/53) that it's relying on a well-known name and well-known key/value pairs. I understand that this is the normal way to configure components in Dapr, but now we're forcing `durabletask-go` to also support that declarative configuration model. As a simple library, `durabletask-go` was never really designed to support unstructured, declarative configuration.

For these reasons, I still feel like it would be wiser and more ergonomic to have backend initialization happen in Dapr rather than in `durabletask-go`.

DeepanshuA:
My thoughts:
I feel DTF-go has to provide the support for backends but Dapr, providing support for multiple components, should provide initialization of the component (database/ actor) and let DTF-go know the backend through NewOrchestrationWorker/NewActivityTaskWorker call.

Here also, user will need to just drop-in the component file. But, yes - to initialize a new component, Dapr will need to support that.
Consider a usecase, where-in, as a user, I want to use the same database cluster (may be a different partition) that I am already using for some different use-case (may be statestore) in Dapr. Then, ideally I would expect it to be initialized only once i.e. in Dapr.

cmd/daprd/main.go
```diff
@@ -103,7 +105,8 @@ func main() {
 		WithBindings(bindingsLoader.DefaultRegistry).
 		WithCryptoProviders(cryptoLoader.DefaultRegistry).
 		WithHTTPMiddlewares(httpMiddlewareLoader.DefaultRegistry).
-		WithWorkflows(workflowsLoader.DefaultRegistry)
+		WithWorkflows(workflowsLoader.DefaultRegistry).
+		WithWorkflowBackends(wfbeLoader.DefaultRegistry)
```
DeepanshuA, line 109: Just a thought:
In this PR, we have used here 3 naming conventions: `workflowbackend`, `wfbackend` and `wfbe`.
Should we make it consistent?

go.mod
```diff
@@ -455,3 +468,7 @@ replace github.com/microcosm-cc/bluemonday => github.com/microcosm-cc/bluemonday
 //
 // Then, run `make modtidy` in this repository.
 // This ensures that go.mod and go.sum are up-to-date.
+
+replace github.com/dapr/components-contrib => ../components-contrib
```
DeepanshuA, line 472: !! I know you are aware, but just leaving this comment here, so that the we don't forget to replace this.

go.mod
```diff
@@ -455,3 +468,7 @@ replace github.com/microcosm-cc/bluemonday => github.com/microcosm-cc/bluemonday
...
+
+replace github.com/microsoft/durabletask-go => ../durabletask-go
```
DeepanshuA, line 474: Same as above

DeepanshuA (commented):
Just Few initial comments.

cmd/daprd/main.go
ASHIQUEMD: I have followed existing naming convention in the codebase, wfbackend is the package name, wfbe is the variable name, where as workflowbackend is a type.

I have fixed the names at few places now to make the variable name consistent.

cmd/daprd/components/wfbackend.go
```diff
@@ -0,0 +1,24 @@
...
+*/
+
+package components
+
+import (
+	wfbeLoader "github.com/dapr/dapr/pkg/components/wfbackend"
+)
+
+func init() {
+	wfbeLoader.DefaultRegistry.RegisterComponent(wfbeLoader.NewWorkflowBackendComp, "actor", "sqlite")
```
ItalyPaleAle, line 23: The goal of having these files is that it makes it so components can be enabled or disabled simply by creating/deleting files.

We need to have a separate file for each workflow backend component, so they can be enabled/disabled at will.

This is related to my earlier point around the fact that we need to have more decoupling so we can do things more "drop-in"...

tests/perf/workflows/workflow_test.go
```diff
@@ -35,8 +35,8 @@ import (
 )
 
 var tr *runner.TestRunner
-
-var testAppNames = []string{"perf-workflowsapp"}
+var backends = []string{"", "sqlite"}
```
ItalyPaleAle, line 38: I don't think we need to run perf tests for each backend, first and foremost because each test already takes almost 2 hours. If you add 2 backends, the test takes 4 hours. A regular workday has 8 hours, so you see where I'm going :)

Second, the SQLite backend is primarily meant for dev/test and its perf isn't key.

Maybe one option is to make the backend configurable via an env var? By default it would use actors.

ItalyPaleAle (changes requested):

pkg/runtime/wfengine/actor_backend.go
```diff
@@ -110,24 +117,41 @@ func getActivityScheduler(activityWorkItemChan chan *backend.ActivityWorkItem) a
...
-	internalActors[be.config.workflowActorType] = be.workflowActor
-	internalActors[be.config.activityActorType] = be.activityActor
+	internalActors[abe.config.workflowActorType] = abe.workflowActor
+	internalActors[abe.config.activityActorType] = abe.activityActor
 	return internalActors
 }
 
-func (be *actorBackend) SetActorRuntime(actors actors.Actors) {
-	be.actors = actors
+func (abe *ActorBackend) SetActorRuntime(actorRuntime actors.ActorRuntime, ctx context.Context) {
```
cgillum, line 127: nit: I believe it's a common Go convention that the `ctx` parameter be the first parameter in a function signature.

pkg/runtime/wfengine/actor_backend.go
```diff
@@ -110,24 +117,41 @@ func getActivityScheduler(activityWorkItemChan chan *backend.ActivityWorkItem) a
...
+	abe.actors = actorRuntime
+	abe.actorRuntime = actorRuntime
+	if abe.actorsReady.CompareAndSwap(false, true) {
```
cgillum, line 130: Are we anticipating that multiple goroutines might call this concurrently? If so, should we also only assign the `actors` and `actorRuntime` fields within this block?

pkg/runtime/wfengine/wfengine.go
```diff
@@ -63,20 +58,23 @@ func IsWorkflowRequest(path string) bool {
...
+	if ok && backendComponentInfo != nil {
+		backendType = backendComponentInfo.WorkflowBackendType
+	} else {
+		backendType = ActorBackendType
+	}
 
-// GetInternalActorsMap returns a map of internal actors that are used to implement workflows
-func (wfe *WorkflowEngine) GetInternalActorsMap() map[string]actors.InternalActor {
-	return wfe.backend.GetInternalActorsMap()
+	be := InitializeWorkflowBackend(appID, backendType, engine, backendComponentInfo, wfBackendLogger)
```
cgillum, line 74: Do we have an info-level log statement anywhere indicating which backend we selected?

pkg/runtime/wfengine/wfengine_backend_test.go
```diff
@@ -0,0 +1,119 @@
...
+
+func TestSqliteBackendSetupWithMetadata(t *testing.T) {
+	mockWorkflowBackendManager := new(daprt.MockSqliteBackendManager)
+	spec := config.WorkflowSpec{MaxConcurrentWorkflowInvocations: 100, MaxConcurrentActivityInvocations: 100}
+
+	engine := wfengine.NewWorkflowEngine(testAppID, spec, mockWorkflowBackendManager)
+
+	_, ok := engine.Backend.(*wfengine.ActorBackend)
+
+	assert.False(t, ok, "actor is not of type ActorBackend")
```
cgillum, line 48: Did you mean "engine.Backend" here (and in the tests below)?
```suggestion
	assert.False(t, ok, "engine.Backend is not of type ActorBackend")
```

pkg/runtime/wfengine/wfengine_backend_test.go
```diff
@@ -0,0 +1,119 @@
...
+
+	assert.True(t, ok, "actor is of type ActorBackend")
+	assert.Equal(t, wfengine.ActorBackendType, engine.BackendType)
+}
+
+func TestInvalidSqliteBackendComponentSetup(t *testing.T) {
+	mockWorkflowBackendManager := new(daprt.MockSqliteBackendComponentInvalidManager)
+	spec := config.WorkflowSpec{MaxConcurrentWorkflowInvocations: 100, MaxConcurrentActivityInvocations: 100}
+
+	engine := wfengine.NewWorkflowEngine(testAppID, spec, mockWorkflowBackendManager)
```
cgillum, line 101: if a component is configured in an invalid way, would it be better to fail the engine startup rather than quietly defaulting to the actor backend?

pkg/runtime/wfengine/wfengine_test.go
```diff
@@ -862,9 +874,15 @@ func startEngineAndGetStore(ctx context.Context, t *testing.T, r *task.TaskRegis
 	return client, engine, store
 }
 
-func getEngine(t *testing.T) *wfengine.WorkflowEngine {
+func getEngine(t *testing.T, ctx context.Context) *wfengine.WorkflowEngine {
```
cgillum, line 877: nit: it's more conventional to have `ctx` be the first parameter in Go functions.

cgillum (commented):
I left some minor comments for the wfengine-related changes. I'll defer to @ItalyPaleAle for the rest.

pkg/runtime/wfengine/actor_backend.go
ItalyPaleAle: I see that this method was copied over from another file. The reason for the CAS was to be able to call `close()` on the channel, which is used to send a signal that the workflow engine has set an actor runtime to everyone that was waiting. See: https://github.com/dapr/dapr/pull/6972

pkg/runtime/wfengine/wfengine_backend_test.go
ItalyPaleAle: Also very much a nit but this should likely be a `require`

pkg/runtime/wfengine/wfengine_backend_test.go
ItalyPaleAle: Agree

pkg/runtime/wfengine/wfengine_test.go
ItalyPaleAle: This is a tricky case because `testing.T` is _also_ supposed to be the first parameter. I believe when we have both, we keep `testing.T` first and then `context.Context`.

cgillum: Ah, good point - I didn't even look carefully to see what the first parameter was. 😅 I agree, we should keep `testing.T` first.

pkg/runtime/wfengine/wfengine.go
ASHIQUEMD: Added info log.

pkg/runtime/wfengine/wfengine_backend_test.go
ASHIQUEMD: If component is misconfigured, we don't fall back to actor, we are [logging the error message](https://github.com/dapr/dapr/pull/7283/files#diff-500c4c9a401c87ceef033c9e69aff0b306bd732e68e521ca8bd9892cb50bb7a4R83-R85) and backend is not initialized.

Are you suggesting to not set the [`a.workflowEngine = wfe`](https://github.com/dapr/dapr/pull/7283/files#diff-e56f9bb947923c6f8bb4c995bf4b9469a0ed117276223d16b177dcb04399a4bcR462-R464) if backend is not configured.

cmd/daprd/components/wfbackend.go
ASHIQUEMD: I initially I had created separate files for each backend component, but I merged them together because there was no other change except the name of the component.

Now also, if we have to add or remove a backend, we only need to remove or add the new component name. In both cases we need to make changes and release it. Do you see other benefits of having these components in their own file?

cmd/daprd/components/wfbackend_actor.go
```diff
@@ -0,0 +1,24 @@
+//go:build allcomponents
```
ItalyPaleAle, line 1: This should be built with stable components too, as it's built-in

```suggestion
//go:build allcomponents || stablecomponents
```

cmd/daprd/components/wfbackend_sqlite.go
```diff
@@ -0,0 +1,24 @@
+//go:build allcomponents
```
ItalyPaleAle, line 1: I wonder if all workflow backends should be built with "stablecomponents" too

pkg/components/wfbackend/metadata.go
```diff
@@ -0,0 +1,57 @@
...
+	defaultOrchestrationLockTimeout = 2 * time.Minute
+	defaultActivityLockTimeout      = 2 * time.Minute
+)
+
+// Metadata represents a set of workflow specific properties.
+type Metadata struct {
+	metadata.Base `json:",inline"`
+}
+
+type sqliteMetadata struct {
```
ItalyPaleAle, line 36: This is specific to a component (SQLite), so should be moved to its own file please

Perhaps it may even make sense to put this business logic into the `cmd/daprd/components/wfbackend_sqlite.go`, so it's all together.

pkg/components/wfbackend/metadata.go
```diff
@@ -0,0 +1,57 @@
...
+	"time"
+
+	"github.com/microsoft/durabletask-go/backend/sqlite"
+
+	"github.com/dapr/components-contrib/metadata"
+	metadataUtil "github.com/dapr/kit/metadata"
+)
+
+const (
+	defaultFilePath                 = ""
```
ItalyPaleAle, line 26: These constants are specific for SQLite, so should have a prefix (unless you move the Metadata struct to a separate file per my other comment)

pkg/components/wfbackend/wfbackend_component.go
```diff
@@ -0,0 +1,34 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package wfbackend
+
+import (
+	"github.com/dapr/kit/logger"
+)
+
+func NewWorkflowBackendComp(logger logger.Logger) WorkflowBackend {
```
ItalyPaleAle, line 21: What is the purpose of this method?

pkg/components/wfbackend/wfbackend_component.go
```diff
@@ -0,0 +1,34 @@
...
+		logger: logger,
+	}
+}
+
+type workflowBackendComponent struct {
+	logger logger.Logger
+}
+
+func (c *workflowBackendComponent) Init(metadata Metadata) error {
+	c.logger.Info("Initializing Dapr workflow backend component ")
```
ItalyPaleAle, line 32: Missing something?

pkg/runtime/compstore/compstore.go
```diff
@@ -35,41 +36,43 @@ import (
...
+	secretsConfigurations     map[string]config.SecretsScope
+	secrets                   map[string]secretstores.SecretStore
+	inputBindings             map[string]bindings.InputBinding
+	inputBindingRoutes        map[string]string
+	outputBindings            map[string]bindings.OutputBinding
+	locks                     map[string]lock.Store
+	pubSubs                   map[string]PubsubItem
+	topicRoutes               map[string]TopicRoutes
+	workflowComponents        map[string]workflows.Workflow
+	workflowBackendComponents map[string]wfbe.WorkflowBackend
```
ItalyPaleAle, line 51: Nit (not required): consider renaming this to `workflowBackend` since other component stores don't say "components"

pkg/runtime/wfengine/sqlite_backend.go
```diff
@@ -0,0 +1,44 @@
...
+func init() {
+	RegisterBackendFactory(SqliteBackendType, getSqliteBackend)
+}
+
+func getSqliteBackend(appID string, backendComponentInfo *wfbe.WorkflowBackendComponentInfo, log logger.Logger) backend.Backend {
+	log.Infof("Initializing sqlite backend for appID: %s", appID)
+	sqliteMetadata := wfbe.NewSqliteMetadata()
+	err := sqliteMetadata.Parse(backendComponentInfo.WorkflowBackendMetadata.Properties)
+	if err != nil {
+		log.Errorf("Failed to parse sqlite backend metadata: %s, sqlite backend is not initialized", err)
```
ItalyPaleAle, line 39:
```suggestion
		log.Errorf("Failed to parse SQLite backend metadata; SQLite backend is not initialized: %v", err)
```

pkg/runtime/wfengine/sqlite_backend.go
```diff
@@ -0,0 +1,44 @@
...
+const (
+	SqliteBackendType = "workflowbackend.sqlite"
+)
+
+func init() {
+	RegisterBackendFactory(SqliteBackendType, getSqliteBackend)
+}
+
+func getSqliteBackend(appID string, backendComponentInfo *wfbe.WorkflowBackendComponentInfo, log logger.Logger) backend.Backend {
+	log.Infof("Initializing sqlite backend for appID: %s", appID)
```
ItalyPaleAle, line 35:
```suggestion
	log.Infof("Initializing SQLite backend for appID: %s", appID)
```

cmd/daprd/components/wfbackend_sqlite.go
ASHIQUEMD: Not all components will be stable by default, but sqlite should stable I believe. I will mark it stable.

pkg/components/wfbackend/wfbackend_component.go
ASHIQUEMD: This function is used when components get initialized for backend component like any other component, for example redis_state which gets initialized from [daprd/components](https://github.com/dapr/dapr/blob/9e2bcdc3d8a64b02466a9274e9259f166eedc95a/cmd/daprd/components/state_redis.go#L24) and implementation is present in [component-contrib](https://github.com/dapr/components-contrib/blob/087a597ec5a06484edd28445a96107fa8811fc02/state/redis/redis.go#L109).

Reason for moving the changes from components-contrib to dapr/dapr was because backend components are not fully fledged components like others and it will be easier to maintain them in dapr/dapr itself.

pkg/components/wfbackend/wfbackend_component.go
ASHIQUEMD: Its basically like [workflow engine components init](https://github.com/dapr/dapr/blob/9e2bcdc3d8a64b02466a9274e9259f166eedc95a/pkg/runtime/wfengine/component.go#L73-L76), which doesn't have any business logic but only to be initialized. In the same way, we are initializing workflow backend component and logging the message.

pkg/runtime/compstore/compstore.go
ASHIQUEMD: Go it, I followed workflow engine components just above this :p

pkg/components/wfbackend/wfbackend_component.go
ItalyPaleAle: I meant to say that this looks like it's missing something as the error message ends with a space! What is this used for btw?

pkg/components/wfbackend/metadata.go
```diff
@@ -0,0 +1,23 @@
...
+limitations under the License.
+*/
+
+package wfbackend
+
+import (
+	"github.com/dapr/components-contrib/metadata"
+)
+
+// Metadata represents a set of workflow specific properties.
```
ItalyPaleAle, line 20: Nit:
```suggestion
// Metadata represents a set of properties specific for workflow backends.
```

pkg/components/wfbackend/registry.go
```diff
@@ -0,0 +1,84 @@
...
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/dapr/dapr/pkg/components"
+	"github.com/dapr/kit/logger"
+)
+
+// Registry is an interface for a component that returns registered state store implementations.
```
ItalyPaleAle, line 24: Nit:
```suggestion
// Registry is an interface for a component that returns registered workflow backend implementations.
```

pkg/components/wfbackend/sqlite_metadata.go
```diff
@@ -0,0 +1,51 @@
+/*
```
ItalyPaleAle, line 1: Can you move this file to `pkg/wfengine/sqlite/metadata.go`? This way it's together with the backend file

pkg/runtime/wfengine/sqlite_backend.go
```diff
@@ -0,0 +1,44 @@
+/*
```
ItalyPaleAle, line 1: Can you move this file to `pkg/wfengine/sqlite/backend.go`? It's nice to have a folder for each component. We can then keep the metadata.go file in the same folder

pkg/runtime/wfengine/sqlite_backend.go
```diff
@@ -0,0 +1,44 @@
...
+	RegisterBackendFactory(SqliteBackendType, getSqliteBackend)
+}
+
+func getSqliteBackend(appID string, backendComponentInfo *wfbe.WorkflowBackendComponentInfo, log logger.Logger) backend.Backend {
+	log.Infof("Initializing SQLite backend for appID: %s", appID)
+	sqliteMetadata := wfbe.NewSqliteMetadata()
+	err := sqliteMetadata.Parse(backendComponentInfo.WorkflowBackendMetadata.Properties)
+	if err != nil {
+		log.Errorf("Failed to parse SQLite backend metadata; SQLite backend is not initialized: %v", err)
+		return nil
```
ItalyPaleAle, line 40: Should this return the error?

pkg/runtime/wfengine/sqlite_backend.go
ASHIQUEMD: I wanted to do the same, but I had to give sqlite as package and then it creates cyclic dependency wfengine, so I had to keep it this way. It doesn't allow to give package as wfengine inside sqlite folder

pkg/components/wfbackend/wfbackend_component.go
ASHIQUEMD: Fixed now.

ItalyPaleAle (dismissed):

pkg/runtime/wfengine/backends/actors/backend.go
```diff
@@ -110,24 +131,41 @@ func getActivityScheduler(activityWorkItemChan chan *backend.ActivityWorkItem) a
...
+	internalActors[abe.config.workflowActorType] = abe.workflowActor
+	internalActors[abe.config.activityActorType] = abe.activityActor
 	return internalActors
 }
 
-func (be *actorBackend) SetActorRuntime(actors actors.Actors) {
-	be.actors = actors
+func (abe *ActorBackend) SetActorRuntime(ctx context.Context, actorRuntime actors.ActorRuntime) {
+	abe.actors = actorRuntime
+	abe.actorRuntime = actorRuntime
```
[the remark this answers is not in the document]
ASHIQUEMD: Good catch! removed it.

pkg/runtime/wfengine/component.go
```diff
@@ -80,8 +80,10 @@ func (c *workflowEngineComponent) Start(ctx context.Context, req *workflows.Star
 		return nil, errors.New("a workflow name is required")
 	}
 
+	// Init with capacity of 3 as worst-case scenario
+	opts := make([]api.NewOrchestrationOptions, 0, 3)
```
DeepanshuA, line 84: What is the logic behind init with capacity 3?

ItalyPaleAle: I added a comment that it's the "worst-case scenario". If you count the number of times something _could_ get appended to opts below, it's 3 at most. So this is just a micro-optimization

ASHIQUEMD: Here we know it will have 3 or fewer elements, so it's better choice to pre-allocate the necessary memory and Go doesn't need to reallocate memory every time we append an element to the slice.

DeepanshuA (approved):
LGTM

ItalyPaleAle:
@ASHIQUEMD I just chatted with @cgillum and there are some concerns with making SQLite available as a backend today. It may be a bit premature.

We don't want to remove the component from the codebase, also because having it available for tests (E2E, perf, IT) is helpful. But we don't think users should be able to use it just yet.

Perhaps we could add a build tag on the component file so it's built only when we run tests? For example, change:

https://github.com/dapr/dapr/pull/7283/files#diff-a5739e0a3a8f8aa325f794a09252631bfead81e7d12ceb4dfc76280313d26b27R1

To:

```go
//go:build e2e || unit || integration
```

And then change the pipelines for tests (including E2E and perf) to build with the e2e or integration tags (just need to set `DAPR_GO_BUILD_TAGS=e2e` or `DAPR_GO_BUILD_TAGS=unit` or `DAPR_GO_BUILD_TAGS=integration` in the pipeline when calling `make build`)

mukundansundar:
waiting for PR https://github.com/dapr/dapr/pull/7373 to be merged first

mukundansundar (approved):
lgtm

JoshVanL:
@mukundansundar @ASHIQUEMD this PR fails integration tests as we now always have a workflow component returned on the component list from the metadata endpoint. Tests need updating to reflect this.

JoshVanL:
@mukundansundar @ASHIQUEMD Please can you also consider hot reloading, and how hot reloading works for this component type? It doesn't look like there are any tests covering this.

ItalyPaleAle:
> @mukundansundar @ASHIQUEMD Please can you also consider hot reloading, and how hot reloading works for this component type? It doesn't look like there are any tests covering this.

Just like for the actor state store, we shouldn't support hot reloading here as it could cause serious consistency issues.

PR #7286, opened 2023-12-11
Hot Reloading: HTTP Middleware Reloading 

JoshVanL:
Part of https://github.com/dapr/dapr/issues/1172
Branched from https://github.com/dapr/dapr/pull/7260

Adds support for HTTP middleware hot reloading.

PR adds support for HTTP middlewares to be dynamically updated. The HTTP
Middleware returns a middleware.HTTP pipleine built from a config spec.
This pipeline implements a single handler which it itself runs the spec
configured HTTP middleware chain. When a middleware Component is added
or removed, the HTTP middleware store is updated and both the HTTP
server and App HTTP channel's pipeline's chains are updated dynamically.

Like today, if a pipeline spec contains handlers whose names, version,
or type does not match that from the store, then the handler is skipped
for that chain. Pipeline handler order is preserved (i.e. reverse order).

The middleware store has been made generic so can be used in future for
planned gRPC middleware support.

Middleware init has been moved from the runtime channel manager to the
runtime processor init procedure to allow for dynamic loading. The
middleware HTTP manager is passed to the runtime processor to expose the
store, and relevant built pipelines passed to the HTTP app channel and
HTTP server.

Middleware integration tests has been added for daprd to check
functionality, ordering, and hotreloading,

Signed-off-by: joshvanl <me@joshvanl.dev>


pkg/middleware/http/http.go
```diff
@@ -0,0 +1,83 @@
+/*
+Copyright 2023 The Dapr Authors
```
daixiang0, line 2:
```suggestion
Copyright 2024 The Dapr Authors
```

pkg/middleware/http/http.go
```diff
@@ -0,0 +1,83 @@
...
+	"github.com/dapr/dapr/pkg/middleware/store"
+	"github.com/dapr/kit/logger"
+)
+
+var log = logger.NewLogger("dapr.middleware.http")
+
+// HTTP returns HTTP middleware pipelines. These pipelines dynamically update
+// the middleware chain when a component is added or removed from the store.
+// Callers need only build a Pipeline once for a given spec.
+type HTTP struct {
```
daixiang0, line 31:
```suggestion
type Chains struct {
```

Something not like HTTP is good to me.

pkg/middleware/http/http.go
yaron2: @JoshVanL Please commit suggestion

pkg/channel/http/http_channel.go
```diff
@@ -231,34 +231,26 @@ func (h *Channel) invokeMethodV1(ctx context.Context, req *invokev1.InvokeMethod
...
-				_, _ = io.Copy(wr, clientResp.Body)
-			}
-			if clientErr != nil {
-				err = clientErr
-			}
-		}))
-		execPipeline.ServeHTTP(rw, channelReq)
-		resp = rw.Result() //nolint:bodyclose
-	} else {
+	// Exec pipeline only if at least one handler is specified
```
yaron2, line 234: With `if len(h.pipeline.Handlers) > 0 {` removed, is this still true?

JoshVanL: Removed

PR #7287, opened 2023-12-11
Hot Reloading: Actor State Store Reloading

JoshVanL:
Part of https://github.com/dapr/dapr/issues/1172
Branched from https://github.com/dapr/dapr/pull/7260

Adds support for actor state store hot reloading.

PR updates the actors package so that the actor state store is dynamically
retrieved from the component store every time the state it invoked. The
passed static actor state store name previously used to retrieve the
state store has been removed.

The component store has been updated to track the current actor state
store. It is no longer tracked by the state processor, and instead the
state processor updates the latest in the component store.

Like before, the actor state store _must_ be configured on Dapr startup,
else the actor subsystem runtime will not be enabled, even if an actor
state store is hot reloaded later.

Adds actor hotreloading integration tests for the actor state store.


ItalyPaleAle (changes requested):
I would like us to be very careful with hot-reloading actor state store please, especially as this interacts with the reminder store. The actor state store is not meant to be hot-reloaded today, and the impact on reminders is unclear.

For example, in the current code I do not see the method that triggers reloading of reminders when the actor state store changes. This could have very major impacts on the actor subsystem.

Another thing to consider is that there could be active actors which loaded state, and assume that they can store it back (with etag matching) later. If the state store changes, that could cause some significant issues.

Issue #6082, opened 2023-03-15
Pluggable components on Windows


roozbehid-ic:
## In what area(s)?

> /area runtime

> /area operator

> /area placement

> /area docs

> /area test-and-release

## Describe the feature
I noticed a log message that pluggable components feature is not available on Windows. And a quick search showed it is using unis sockets and probably because of this it is not supported.
I think after some Windows 10 and newers Windows Server it is supported. Why not just enable it on those supported OSes and fail on those unsupported Windows OSes instead of just disabling it always on Windows?

ItalyPaleAle:
Support for Windows 2022 will land with Dapr 1.11, the next release.

CC: @philliphoff and @shubham1172 can comment on plans to support pluggable components after we have support for Windows 2022

roozbehid-ic:
Thanks. Still doing a research to figure out exact versions that support it, it seems all versions after 2017 should have this enabled which also means Windows 2019 and Windows 11 and most Windows 10 versions.

mukundansundar:
a related issue opened in CLI for unix domain socket support in Windows https://github.com/dapr/cli/issues/1261 
Potentially a separate issue can be opened for unix domain socket support in runtime also and then the pluggable components support can be added.

mukundansundar:
related to USD support in Windows : https://github.com/dapr/dapr/issues/4128

roozbehid:
@philliphoff maybe a label for this...?

philliphoff:
So I thought I'd take another look at this and, surprisingly, I got it to work!  It requires a couple of the runtime:

 - On Windows, UDS paths apparently require the format `unix:<file path>`, whereas, on other platforms the format `unix://<file path>` is used.
  - On Windows, UDS files do not have the `ModeSocket` flag set, so there's no way to differentiate between "normal" and UDS files.
 
    The runtime currently tests for the flag to ensure that only socket files are loaded from the sockets folder.  For Windows, it is probably "good enough" to simply see if the socket file path appears "socket shaped", as done in other repos.  For example, the runtime might simply assume files with the extension `.sock` are socket files.  The Java/.NET pluggable component SDKs already support a `DAPR_COMPONENTS_SOCKETS_EXTENSION` to alter the extension; the runtime might use that same variable to also allow alternative extensions to `.sock`

Supporting Windows would be helpful, in particular, to .NET developers doing Dapr custom components.

@yaron2 Consider reopening?

PR #7289, opened 2023-12-11
Enable pluggable components on Windows

philliphoff:
# Description

Removes the restriction on use of pluggable components on Windows.  Updates the Unix Domain Socket syntax and detection logic to handle Windows' quirks (namely using `unix:` instead of `unix://` and not relying on the `ModeSocket` flag).

In the case of Windows, as the `ModeSocket` flag is not set, the Dapr runtime instead relies on looking for "socket shaped" sockets which, by default, have the `.sock` extension. If the user desires to use another extension, they can set the `DAPR_COMPONENTS_SOCKETS_EXTENSION` environment variable to an arbitrary extension. (This environment variable is already supported in the Java and .NET pluggable component SDKs.)

With this change, the Dapr runtime will also respect temp directories specified via `$TMPDIR`, whereas it previously used `/tmp`.

## Issue reference

Please reference the issue this PR will close: #6082 (closed earlier for lack of comments)

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Unit tests passing


pkg/components/pluggable/grpc.go
```diff
@@ -103,6 +104,11 @@ func socketDialer(socket string, additionalOpts ...grpc.DialOption) GRPCConnecti
 // SocketDial creates a grpc connection using the given socket.
 func SocketDial(ctx context.Context, socket string, additionalOpts ...grpc.DialOption) (*grpc.ClientConn, error) {
 	udsSocket := "unix://" + socket
+
+	if runtime.GOOS == "windows" {
+		udsSocket = "unix:" + socket
```
daixiang0, line 109: Could we add a test to cover it?

pkg/components/pluggable/discovery.go
```diff
@@ -47,13 +47,16 @@ func removeExt(fileName string) string {
 }
 
 const (
-	SocketFolderEnvVar  = "DAPR_COMPONENTS_SOCKETS_FOLDER"
-	defaultSocketFolder = "/tmp/dapr-components-sockets"
+	SocketsFolderEnvVar  = "DAPR_COMPONENTS_SOCKETS_FOLDER"
+	defaultSocketsFolder = "dapr-components-sockets"
```
daixiang0, line 51:
```suggestion
	DefaultSockets\Folder = "dapr-components-sockets"
```

utils/utils.go
```diff
@@ -74,7 +81,15 @@ func GetIntValOrDefault(val int, defaultValue int) int {
 
 // IsSocket returns if the given file is a unix socket.
 func IsSocket(f fs.FileInfo) bool {
```
daixiang0, line 83: Please add tests for it.

daixiang0 (changes requested):
@philliphoff Thank you for great contribution!

Just some nits here.

pkg/components/pluggable/discovery.go
```diff
@@ -47,13 +47,16 @@ func removeExt(fileName string) string {
...
 )
 
-// GetSocketFolderPath returns the shared unix domain socket folder path
-func GetSocketFolderPath() string {
-	return utils.GetEnvOrElse(SocketFolderEnvVar, defaultSocketFolder)
+// GetSocketsFolderPath returns the shared unix domain socket folder path
+func GetSocketsFolderPath() string {
```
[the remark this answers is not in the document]
philliphoff: Happy to change it back, but the existing variable `DAPR_COMPONENTS_SOCKETS_FOLDER`, so thought the change would be more consistent (especially with `DAPR_COMPONENTS_SOCKETS_EXTENSION`).

pkg/components/pluggable/discovery.go
philliphoff: Wasn't sure if there was a reason for the existing naming, so didn't change it originally, but can do so.

pkg/components/pluggable/grpc.go
philliphoff: I'll see what can be added.

utils/utils.go
philliphoff: PR has tests added to `SocketExists()` to cover the Windows cases, which *should* cover this function, being the only place it's called.

JoshVanL:
@philliphoff please can you re-enable the socket integration test for windows in this PR? :) they are also quite limited so consider extending them a bit if you want
https://github.com/dapr/dapr/blob/53ceb4aa59a57a25de9d76b61b298584f5339354/tests/integration/suite/daprd/pluggable/basic.go#L48

philliphoff:
> please can you re-enable the socket integration test for windows in this PR? :) they are also quite limited so consider extending them a bit if you want

Enabled those tests, which seem to pass on Windows and other platforms (though looks like there's some flakyness in other tests).

pkg/components/pluggable/grpc.go
philliphoff: Found some more related tests to unblock on Windows.

philliphoff:
@daixiang0 Anything more you'd like me to add?

m3nax:
Yesterday, version 0.3.0 of the dapr-sandbox/components-dotnet-sdk was released which includes the initialisation of Unix Domain Sockets on windows, can you enable the search in the component folder on windows as well so that people using Visual Studio can also develop on it?

philliphoff:
> Yesterday, version 0.3.0 of the dapr-sandbox/components-dotnet-sdk was released which includes the correct initialisation of Unix Domain Sockets on windows, can you enable the search in the component folder on windows as well so that people using Visual Studio like me can also develop on it?

@m3nax Can you be a bit more specific about what you're trying to do?  I'm not sure what you mean by "search in the component folder" with respect to pluggable components.  (Usually the UDS is placed in a temp folder separate from the Dapr components folder.)

m3nax:
@philliphoff It's quite simple, I'm using Visual Studio Dapr Extension for Visual Studio 2022 whitch starts debugging on my pc, whitch is Windows. The task manager confirm it.

image

 Then, when I try to register the binding to use it in another application I get the current message: 
``` 
level=debug msg="the current OS does not support pluggable components feature, skipping initialization" app_id=myapp
```

philliphoff:
> This makes it impossible to debug pluggable components with Visual Studio 2022. Is there any other way?

@m3nax Until this PR gets committed, likely not.  the VS Dapr extension *does* support limited debugging of applications within WSL, but I've not tried to add a pluggable component into the mix so I can't say for certain whether that would work.

m3nax:
Not stale

philliphoff:
(Just a note that I'll be unlikely to have time in the near term to freshen up this PR, also given that there have been some significant changes for component hot-reload that intersect these changes.  I welcome any interested people to take that up if they have time/interest, else I'll probably just close this PR.)

PR #7294, opened 2023-12-12
Actor State TTL: Remove comment suggesting to remove feature gate in 1.12

JoshVanL:
The comment associated with the ActorStateTTL feature gate suggests to remove the gate in v1.12. This is obviously not the case so let's remove the comment. The project can decide when we do want to remove it in future.


mukundansundar (approved):

JoshVanL:
> Add a small comment that it will be removed in future release?

@mukundansundar Is this not implied for all feature gates?

artursouza (approved):

PR #7299, opened 2023-12-14
Configure middlewares within the component

### LINKED ISSUE #6660: Proposal: Auto-configure middleware pipelines via priority and pipeline type properties
This is a proposal to add a new mechanism for configuring Middlewares, directly from the component YAML **instead of the Configuration CRD**.

Each middleware would support the optional `priority` property via which the ordering can be determined in which a middleware components should be applied to a pipeline.

Another property should be `pipelineType` with values `httpPipeline` and `appHttpPipeline` at this time.

This also will allow scoping of middleware components to individual apps in Kubernetes without having to rely on a global configuration which may be invalid for an individual app.

sadath-12:
# Description

## Issue reference

Please reference the issue this PR will close: #6660 

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing


pkg/runtime/compstore/components.go
```diff
@@ -90,6 +91,18 @@ func (c *ComponentStore) ListComponents() []compsv1alpha1.Component {
 	return comps
 }
 
+func (c *ComponentStore) ListMatchComponents(match string) []compsv1alpha1.Component {
```
ItalyPaleAle, line 94: I would recommend changing this to prefix-matching. Then you can pass as prefix `middleware.`. This is safer than doing a match in the entire string, as imagine one day we had a component called `state.mymiddleware` for example :)

pkg/runtime/channels/channels.go
```diff
@@ -107,7 +109,14 @@ func (c *Channels) Refresh() error {
 	log.Debug("Refreshing channels")
 
 	// Create a HTTP channel for external HTTP endpoint invocation
-	pipeline, err := c.buildHTTPPipelineForSpec(c.appHTTPPipelineSpec, "app channel")
+	var pipeline middlehttp.Pipeline
+	var err error
+	if c.appHTTPPipelineSpec != nil {
```
ItalyPaleAle, line 114: This change applies the middlewares to the appHttpPipeline, but we also have a httpPipeline that is defined in pkg/channel/http. A pipeline needs to be applied there too.

pkg/runtime/channels/channels.go
```diff
@@ -152,6 +161,81 @@ func (c *Channels) Refresh() error {
 	return nil
 }
 
+func (c *Channels) BuildHTTPPipelineFromComponents(comps []compsv1alpha1.Component, expectedPiplineType string) (middlehttp.Pipeline, error) {
```
ItalyPaleAle, line 164: This should be moved to a common package so it can be used for the httpPipeline too

msfussell:
Please create docs PR before code PR is merge.

pkg/runtime/compstore/components.go
sadath-12: Ya good catch tho

pkg/runtime/channels/channels.go
sadath-12: Yeah , your right I have passed ```expectedPiplineType``` arg to support that .

pkg/runtime/channels/channels.go
sadath-12: Yeah , your right I have passed ```expectedPiplineType``` arg to support that .  And also adding making it stay in same package as channels is making more easy to use the utilities . Please let me know if I missed something

sadath-12:
> Please create docs PR before code PR is merge.

Yes , Once this PR is approved ill make sure about that . thanks for reminding

hhunter-ms:
@sadath-12 @msfussell Hey there - I'm working on prioritizing docs issues for v1.14 milestone. Is this still aiming to be ready for v1.14 release?

cicoyle:
@sadath-12 gentle ping on this PR :)

sadath-12:
hey @ItalyPaleAle, this one's been on my mind again. #6660 is still open so guessing the problem is still relevant - is the component-metadata approach here still the direction you'd want, or did thinking move somewhere else since? happy to pick it back up and rebase if it's still worth doing

berndverst:
@sadath-12 neither @ItalyPaleAle nor I (@berndverst ) are active in Dapr anymore

Issue #7101, opened 2023-10-26
[Workflow] Support for workflow ID reuse policies


cgillum:
## In what area(s)?

> /area runtime

## Describe the feature

This is a continuation of a discussion [here](https://github.com/dapr/dapr/issues/6434). The idea is simplify the creation and re-creation of workflows that have a well-known workflow ID.

When scheduling the creation of a new workflow instance, the following options should be available (taken from the proposal by @clintsinger):

- `TerminateIfExists` - Terminates any existing workflows with the same instance ID and then schedules a new instance as one atomic action, similar to on-demand ContinueAsNew.
- `SkipIfExists` - If there is an existing workflow already scheduled, then the scheduler does nothing.
- `ThrowIfExists` - If there is an existing workflow then the scheduler throws an exception (this is the current behavior).

As an optional extension of this, we can consider exposing an "Execution ID" or "Generation ID" property to the workflow which can help distinguish different instantiations of the given workflow instance ID. The Durable Task Framework has this concept already (it's an auto-generated UUID) but we don't currently surface it anywhere.

## Release Note

RELEASE NOTE: **ADD** Workflow creation policy options.

olitomlinson:
While we're on the subject of scheduling workflows, @cgillum I raised some feedback a while back that it would be useful to improve the DevEx for the scenario when trying to start a workflow when that workflow is already running. 

Right now, it throws an Exception, Ideally i'd prefer it not to throw, as it's not necessarily an exceptional circumstance, it's entirely use-case specific.

However, in the spirit of being pragmatic, if an Exception must remain, it would be nice to have it throw a specific exception, rather than relying on matching a vague/ambiguous string prefix, which could change!

this is what I do right now :

```c#
try
{
    result = await workflowClient.ScheduleNewWorkflowAsync(
        name: nameof(SagaWorkflow),
        instanceId: workflowId,
        input: orderInfo);
}
catch(Grpc.Core.RpcException ex) when (ex.StatusCode == Grpc.Core.StatusCode.Unknown && ex.Status.Detail.StartsWith("an active workflow with ID"))
{
    ...
}
```

cgillum:
Based on the proposal above, this would come into play only when using the (default) `ThrowIfExists` policy. I agree that we need to have a better exception handling story than this.

kaibocai:
Working items for this proposal. 
- [X] durabletask-go update https://github.com/microsoft/durabletask-go/pull/46
[...]

PR #7308, opened 2023-12-15
Support orchestration id reuse policy

kaibocai:
# Description

## Issue reference

This PR tries to update the logic to support reuse orchestration ID, more details can be found https://github.com/microsoft/durabletask-go/issues/42, https://github.com/dapr/dapr/issues/7101
Corresponding protobuf updates can be found https://github.com/microsoft/durabletask-protobuf/pull/19

1. This PR requires https://github.com/microsoft/durabletask-go/pull/51 in durabletask-go
2. This PR upgrade the OTEL dependency to be compatible with durabletask-go as a newer version is used in durabletask-go
3. This PR update the purge instance logic a bit so that only instance in `[Completed, Failed, Terminated]` will be purged. 

Please reference the issue this PR will close: https://github.com/dapr/dapr/issues/7101

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests
* [X] Unit tests passing
* [X] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[3909]_
* [X] Specification has been updated / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[3909]_
* [X] Provided sample for the feature / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[3909]_


pkg/runtime/wfengine/wfengine_test.go
```diff
@@ -162,16 +164,133 @@ func TestSingleActivityWorkflow(t *testing.T) {
...
+			client.WaitForOrchestrationStart(ctx, id)
+			pivotTime := time.Now()
+			// schedule again, it should ignore creating the new orchestration
+			id, err = client.ScheduleNewOrchestration(ctx, "SingleActivity", api.WithInput("World"), api.WithInstanceID(id), api.WithOrchestrationIdReusePolicy(reuseIdPolicy))
+			require.NoError(t, err)
+			timeoutCtx, cancelTimeout := context.WithTimeout(ctx, 30*time.Second)
+			defer cancelTimeout()
+			metadata, err := client.WaitForOrchestrationCompletion(timeoutCtx, id)
+			require.NoError(t, err)
+			assert.Equal(t, true, metadata.IsComplete())
```
ItalyPaleAle:
```suggestion
			assert.True(t, metadata.IsComplete())
```

pkg/runtime/wfengine/wfengine_test.go
```diff
@@ -162,16 +164,133 @@ func TestSingleActivityWorkflow(t *testing.T) {
...
+	for _, opt := range GetTestOptions() {
+		suffix := opt(engine)
+		t.Run(opt(engine), func(t *testing.T) {
+			instanceID := api.InstanceID("IGNORE_IF_RUNNING_OR_COMPLETED_" + suffix)
+			id, err := client.ScheduleNewOrchestration(ctx, "SingleActivity", api.WithInput("世界"), api.WithInstanceID(instanceID))
+			require.NoError(t, err)
+			id, err = client.ScheduleNewOrchestration(ctx, "SingleActivity", api.WithInput("World"), api.WithInstanceID(id))
+			if assert.Error(t, err) {
+				assert.Contains(t, err.Error(), "already exists")
+			}
```
ItalyPaleAle:
```suggestion
			require.Error(t, err)
			require.ErrorContains(t, err, "already exists")
```

pkg/runtime/wfengine/workflow.go
```diff
@@ -184,20 +196,54 @@ func (wf *workflowActor) createWorkflowInstance(ctx context.Context, actorID str
...
-	// We block (re)creation of existing workflows unless they are in a completed state
-	// Or if they still have any pending activity result awaited.
-	if !created {
-		runtimeState := getRuntimeState(actorID, state)
-		if !runtimeState.IsCompleted() {
-			return fmt.Errorf("an active workflow with ID '%s' already exists", actorID)
-		}
-		if wf.activityResultAwaited.Load() {
-			return fmt.Errorf("a terminated workflow with ID '%s' is already awaiting an activity result", actorID)
+	// orchestration doesn't exist, create it.
```
ItalyPaleAle, line 199: Nit:
```suggestion
	// orchestration didn't exist and was just created
```

kaibocai:
tag @cgillum , @ItalyPaleAle , @RyanLettieri for review.

go.mod
```diff
@@ -456,3 +455,5 @@ replace github.com/microcosm-cc/bluemonday => github.com/microcosm-cc/bluemonday
 //
 // Then, run `make modtidy` in this repository.
 // This ensures that go.mod and go.sum are up-to-date.
+
+replace github.com/microsoft/durabletask-go v0.3.1 => github.com/microsoft/durabletask-go v0.3.2-0.20231215171910-b4fe18e3027d
```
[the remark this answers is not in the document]
kaibocai: Thanks for the note, will do that once https://github.com/microsoft/durabletask-go/pull/51 is merged to main

pkg/runtime/wfengine/workflow.go
kaibocai: update a bit please check again if it makes sense. Thank you!

pkg/runtime/wfengine/wfengine_test.go
```diff
@@ -162,16 +164,132 @@ func TestSingleActivityWorkflow(t *testing.T) {
 
 	ctx := context.Background()
 	client, engine := startEngine(ctx, t, r)
+
 	for _, opt := range GetTestOptions() {
+		suffix := opt(engine)
 		t.Run(opt(engine), func(t *testing.T) {
```
cgillum, line 170: Should we be using `suffix` here?
```suggestion
		t.Run(suffix, func(t *testing.T) {
```

pkg/runtime/wfengine/workflow.go
```diff
@@ -62,6 +62,11 @@ type recoverableError struct {
 	cause error
 }
 
+type PolicyAndEventData struct {
```
cgillum, line 65: Let's give this a more specific name, like `CreateWorkflowInstanceRequest`

pkg/runtime/wfengine/workflow.go
```diff
@@ -62,6 +62,11 @@ type recoverableError struct {
...
+	Policy    *api.OrchestrationIdReusePolicy `json:"policy"`
+	EventData []byte                          `json:"eventData"`
```
cgillum, line 67: For consistency with existing code, I suggest `StartEventData`.
```suggestion
	StartEventBytes []byte                    `json:"startEventBytes"`
```

pkg/runtime/wfengine/workflow.go
```diff
@@ -184,20 +196,54 @@ func (wf *workflowActor) createWorkflowInstance(ctx context.Context, actorID str
...
-			return fmt.Errorf("a terminated workflow with ID '%s' is already awaiting an activity result", actorID)
+	// orchestration didn't exist and just create it.
+	if created {
+		return wf.createInstance(ctx, actorID, startEvent, state)
+	}
+
+	// orchestration already exists, apply reuse id policy
+	runtimeState := getRuntimeState(actorID, state)
+	runtimeStatus := runtimeState.RuntimeStatus()
+	targetStatusValues := backend.BuildStatusSet(reuseIDPolicy.GetOperationStatus())
```
cgillum, line 207: This goes back to my comment on the durabletask-go repo, but I don't think we need to expose a `BuildStatusSet` utility function for this simple purpose. I think a simple for-loop over `reuseIDPolicy.GetOperationStatus()` and comparing to `runtimeStatus` on each loop iteration is good enough since we only have a small number of possible status values. This might even be more performant than creating a map and doing hashing operations for the key lookup.

pkg/runtime/wfengine/workflow.go
```diff
@@ -184,20 +196,54 @@ func (wf *workflowActor) createWorkflowInstance(ctx context.Context, actorID str
...
 	}
+	if wf.activityResultAwaited.Load() {
+		return fmt.Errorf("a terminated workflow with ID '%s' is already awaiting an activity result", actorID)
+	}
+	wfLogger.Infof("Workflow actor '%s': workflow was previously completed and is being recreated", actorID)
+	state.Reset()
+	return wf.createInstance(ctx, actorID, startEvent, state)
+}
 
+func (wf *workflowActor) createInstance(ctx context.Context, actorID string, startEvent *backend.HistoryEvent, state *workflowState) error {
```
cgillum, line 246: I'm a little confused about the function name here. It's called `createInstance` but the only thing it does is create a reminder? Can we give this a more specific name, like `scheduleWorkflowStart`?

pkg/runtime/wfengine/workflow.go
```diff
@@ -209,6 +255,36 @@ func (wf *workflowActor) createWorkflowInstance(ctx context.Context, actorID str
 	return wf.saveInternalState(ctx, actorID, state)
 }
 
+// This method cleans up a workflow associated with the given actorID
+func (wf *workflowActor) cleanupOrchestrationStateInternal(ctx context.Context, actorID string, state *workflowState, requireCompleted bool) error {
+	runtimeState := getRuntimeState(actorID, state)
```
cgillum, line 260: We've already loaded the runtime state in `createWorkflowInstance`. Should we reuse that and pass it as a parameter rather than getting the runtime state again?

pkg/runtime/wfengine/workflow.go
```diff
@@ -209,6 +255,36 @@ func (wf *workflowActor) createWorkflowInstance(ctx context.Context, actorID str
...
+	runtimeStatus := runtimeState.RuntimeStatus()
+	isCompleted := runtimeStatus == api.COMPLETED || runtimeStatus == api.TERMINATED || runtimeStatus == api.FAILED
```
cgillum, line 262: This requires a change in durabletask-go, but I wonder if we should prefix these constants with `RUNTIME_STATUS_` so that it's clear what these values are for.

cgillum, line 262: Can't we just use the existing `runtimeStatus.IsCompleted()` for this instead of checking specific status values? That would be more maintainable.

kaibocai: The reason I am not using `runtimeStatus.IsCompleted()` is that I am not quite sure if it equals `[COMPLETED , TERMINATED , FAILED]`. From the code, the `completedEvent` is assigned when the orchestration completed at https://github.com/microsoft/durabletask-go/blob/aa335e20a192395649eb9ede97b57c9f647238ea/backend/runtimestate.go#L134, but when orchestration is terminated we don't assign a `completedEvent`, instead we create a `ExecutionTerminatedEvent` at https://github.com/microsoft/durabletask-go/blob/aa335e20a192395649eb9ede97b57c9f647238ea/backend/runtimestate.go#L197, so it seems `runtimeStatus.IsCompleted()` only check if orchestration is completed but not check if it's terminated/failed.

cgillum: According to [this integration test code](https://github.com/microsoft/durabletask-go/blob/aa335e20a192395649eb9ede97b57c9f647238ea/tests/orchestrations_test.go#L749-L757), a terminated orchestration results in `runtimeStatus.IsCompleted()` being `true`.

cgillum: Actually, hang on - I need to double check. The code I pointed to is for orchestration metadata, not the orchestration runtime state object...

cgillum: Okay, so confirming this wasn't as straightforward as I thought. But the short answer is that a terminated orchestration should always be completed. When an orchestration processes an ExecutionTerminated event, the SDK will generate a `completedEvent` with a status of "Terminated". This can be seen in the SDK code [here](https://github.com/microsoft/durabletask-go/blob/aa335e20a192395649eb9ede97b57c9f647238ea/task/orchestrator.go#L577-L579).

kaibocai: Got it. Updated!

pkg/runtime/wfengine/backend.go
```diff
@@ -139,17 +140,33 @@ func (be *actorBackend) CreateOrchestrationInstance(ctx context.Context, e *back
...
+	}
+
 	// Invoke the well-known workflow actor directly, which will be created by this invocation
 	// request. Note that this request goes directly to the actor runtime, bypassing the API layer.
 	req := invokev1.
 		NewInvokeMethodRequest(CreateWorkflowInstanceMethod).
 		WithActor(be.config.workflowActorType, workflowInstanceID).
-		WithRawDataBytes(eventData).
+		WithRawDataBytes(requestBytes).
 		WithContentType(invokev1.OctetStreamContentType)
```
cgillum, line 170: nit: we should change this to specify that the payload is JSON. I don't think it actually matters from a runtime perspective, but it will at least help mitigate possible confusion.

pkg/runtime/wfengine/workflow.go
```diff
@@ -250,29 +331,8 @@ func (wf *workflowActor) purgeWorkflowState(ctx context.Context, actorID string)
...
-		return err
-	}
-	// This will do the purging
-	err = wf.actors.TransactionalStateOperation(ctx, req)
-	if err != nil {
-		return err
-	}
-	wf.states.Delete(actorID)
-	return nil
+	return wf.cleanupOrchestrationStateInternal(ctx, actorID, state, true && !runtimeState.IsCompleted())
```
cgillum, line 335:
```suggestion
	return wf.cleanupOrchestrationStateInternal(ctx, actorID, state, !runtimeState.IsCompleted())
```

pkg/runtime/wfengine/workflow.go
```diff
@@ -184,20 +196,62 @@ func (wf *workflowActor) createWorkflowInstance(ctx context.Context, actorID str
...
+	runtimeStatus := runtimeState.RuntimeStatus()
+	// if target status doesn't match, fall back to original logic, create instance only if previous one is completed
+	if !isStatusMatch(reuseIDPolicy.GetOperationStatus(), runtimeStatus) {
+		return wf.createIfCompleted(ctx, runtimeState, actorID, state, startEvent)
+	}
+
+	switch reuseIDPolicy.GetAction() {
+	case api.REUSE_ID_ACTION_IGNORE:
+		// Log an warning message and ignore creating new instance
+		wfLogger.Warnf("An instance with ID '%s' already exists; dropping duplicate create request", actorID)
```
cgillum, line 215: For consistency, we should follow the logging format that's used elsewhere in this file. For example:
```suggestion
		wfLogger.Warnf("Workflow actor '%s': ignoring request to recreate the current workflow instance", actorID)
```

pkg/runtime/wfengine/workflow.go
```diff
@@ -209,6 +263,33 @@ func (wf *workflowActor) createWorkflowInstance(ctx context.Context, actorID str
 	return wf.saveInternalState(ctx, actorID, state)
 }
 
+// This method cleans up a workflow associated with the given actorID
+func (wf *workflowActor) cleanupOrchestrationStateInternal(ctx context.Context, actorID string, state *workflowState, requiredAndNotCompleted bool) error {
+	// Only purge orchestration in the ['COMPLETED', 'FAILED', 'TERMINATED'] statuses,
```
cgillum, line 268: This comment should probably be updated.

cgillum (commented):
Just some minor things.

cgillum (approved):
LGTM! I'll go ahead and merge the `durabletask-go` changes and create a release so that this PR can be unblocked.

kaibocai:
> LGTM! I'll go ahead and merge the `durabletask-go` changes and create a release so that this PR can be unblocked.

sorry that I didn't see this one and created the release myself.

kaibocai:
Hi @cgillum, @ItalyPaleAle and @RyanLettieri, the duraletask-go is updated to version v0.4.0 and I think this PR should be ready for final review and merge. Please help to take another look. Thanks!

kaibocai:
tag @mukundansundar for review as well.

ItalyPaleAle (approved):

pkg/runtime/wfengine/workflow.go
```diff
@@ -184,20 +196,62 @@ func (wf *workflowActor) createWorkflowInstance(ctx context.Context, actorID str
...
+
+	// orchestration already existed: apply reuse id policy
+	runtimeState := getRuntimeState(actorID, state)
+	runtimeStatus := runtimeState.RuntimeStatus()
+	// if target status doesn't match, fall back to original logic, create instance only if previous one is completed
+	if !isStatusMatch(reuseIDPolicy.GetOperationStatus(), runtimeStatus) {
+		return wf.createIfCompleted(ctx, runtimeState, actorID, state, startEvent)
+	}
+
+	switch reuseIDPolicy.GetAction() {
```
DeepanshuA, line 212: The PR LGTM, I just wanted to understand how are we skipping the creation of a new orchestration here, (i.e. the option `SkipIfExists` from https://github.com/microsoft/durabletask-go/issues/42#issue-1966761413)

kaibocai:
```go
case api.REUSE_ID_ACTION_IGNORE:
		// Log an warning message and ignore creating new instance
		wfLogger.Warnf("Workflow actor '%s': ignoring request to recreate the current workflow instance", actorID)
		return nil
```

So if users choose `REUSE_ID_ACTION_IGNORE` and if there is an existing instance with the same instance ID, then we will enter the above block and just log a message and directly return.

DeepanshuA: my bad, I missed that Ignore is Skip, thanks for clarification

DeepanshuA (approved):

cgillum:
Hi @artursouza, @yaron2, @mukundansundar. This workflow PR has been signed off and is ready to be merged.

Issue #7311, opened 2023-12-16
Placement loses leadership "too frequently" causing "no route to host"


artursouza:
## In what area(s)?

> /area runtime

> /area operator

/area placement

> /area docs

> /area test-and-release

## What version of Dapr?

> 1.1.x
> 1.0.x
> edge: output of `git describe --dirty`
1.12.2
1.12.3-rc.1

## Expected Behavior

Placement service keeps the leadership on cluster startup until the leader node goes down.
Sidecar nodes are not stuck in the placement table after being deleted (pod terminated).

## Actual Behavior

Placement service loses Raft leadership dozens times within an hour, sometimes twice in a minute.
If there are pods being terminated, a host might be stuck in the placement table and actor clients receive "no route to host" forever until placement service is restarted.

```
time="2023-12-15T19:24:47.342396471Z" level=debug msg="heartbeat timeout reached, starting election,last-leader-addr=,last-leader-id=" instance=dapr-placement-server-1 scope=dapr.placement.raft type=log ver=1.12.3-rc.1
```

## Release Note

RELEASE NOTE: **FIXED** race condition between placement leadership and pod termination

PR #7312, opened 2023-12-16
Fix race due to pod removed during leadership change

artursouza:
# Description

Change placement's Raft setting to reduce leadership changes.

## Issue reference

Please reference the issue this PR will close: https://github.com/dapr/dapr/issues/7311

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing

### DIFF
--- a/charts/dapr/charts/dapr_placement/templates/dapr_placement_statefulset.yaml
+++ b/charts/dapr/charts/dapr_placement/templates/dapr_placement_statefulset.yaml
@@ -186,8 +186,6 @@ spec:
             valueFrom:
               fieldRef:
                 fieldPath: metadata.namespace
-          - name: PLACEMENT_RAFT_SETTINGS
-            value: v2
       serviceAccountName: dapr-placement
       volumes:
       - name: dapr-trust-bundle
--- a/charts/dapr/charts/dapr_placement/templates/dapr_placement_statefulset.yaml
+++ b/charts/dapr/charts/dapr_placement/templates/dapr_placement_statefulset.yaml
@@ -186,6 +186,8 @@ spec:
             valueFrom:
               fieldRef:
                 fieldPath: metadata.namespace
+          - name: PLACEMENT_RAFT_SETTINGS
+            value: v2
       serviceAccountName: dapr-placement
       volumes:
       - name: dapr-trust-bundle
--- a/charts/dapr/charts/dapr_sidecar_injector/templates/dapr_sidecar_injector_deployment.yaml
+++ b/charts/dapr/charts/dapr_sidecar_injector/templates/dapr_sidecar_injector_deployment.yaml
@@ -154,6 +154,11 @@ spec:
 {{- if not .Values.global.actors.enabled }}
         - name: SKIP_PLACEMENT
           value: "true"
+{{- else }}
+{{- if .Values.global.ha.enabled }}
+        - name: PLACEMENT_ADDRESSES
+          value: "dapr-placement-server-0.dapr-placement-server.{{ .Release.Namespace }}.svc.cluster.local:50005,dapr-placement-server-1.dapr-placement-server.{{ .Release.Namespace }}.svc.cluster.local:50005,dapr-placement-server-2.dapr-placement-server.{{ .Release.Namespace }}.svc.cluster.local:50005"
+{{- end }}
 {{- end }}
         - name: SIDECAR_RUN_AS_NON_ROOT
           value: "{{ .Values.sidecarRunAsNonRoot }}"
--- a/charts/dapr/charts/dapr_sidecar_injector/templates/dapr_sidecar_injector_deployment.yaml
+++ b/charts/dapr/charts/dapr_sidecar_injector/templates/dapr_sidecar_injector_deployment.yaml
@@ -154,11 +154,6 @@ spec:
 {{- if not .Values.global.actors.enabled }}
         - name: SKIP_PLACEMENT
           value: "true"
-{{- else }}
-{{- if .Values.global.ha.enabled }}
-        - name: PLACEMENT_ADDRESSES
-          value: "dapr-placement-server-0.dapr-placement-server.{{ .Release.Namespace }}.svc.cluster.local:50005,dapr-placement-server-1.dapr-placement-server.{{ .Release.Namespace }}.svc.cluster.local:50005,dapr-placement-server-2.dapr-placement-server.{{ .Release.Namespace }}.svc.cluster.local:50005"
-{{- end }}
 {{- end }}
         - name: SIDECAR_RUN_AS_NON_ROOT
           value: "{{ .Values.sidecarRunAsNonRoot }}"
--- a/pkg/placement/raft/server.go
+++ b/pkg/placement/raft/server.go
@@ -214,15 +214,15 @@ func (s *Server) StartRaft(ctx context.Context, sec security.Handler, config *ra
 		// Set default configuration for raft
 		s.config = &raft.Config{
 			ProtocolVersion:    raft.ProtocolVersionMax,
-			HeartbeatTimeout:   1000 * time.Millisecond,
-			ElectionTimeout:    1000 * time.Millisecond,
-			CommitTimeout:      50 * time.Millisecond,
+			HeartbeatTimeout:   2 * time.Second,
+			ElectionTimeout:    2 * time.Second,
+			CommitTimeout:      100 * time.Millisecond,
 			MaxAppendEntries:   64,
 			ShutdownOnRemove:   true,
 			TrailingLogs:       10240,
 			SnapshotInterval:   120 * time.Second,
 			SnapshotThreshold:  8192,
-			LeaderLeaseTimeout: 500 * time.Millisecond,
+			LeaderLeaseTimeout: 2 * time.Second,
 		}
 	} else {
 		s.config = config
--- a/pkg/placement/raft/server.go
+++ b/pkg/placement/raft/server.go
@@ -18,6 +18,7 @@ import (
 	"errors"
 	"fmt"
 	"net"
+	"os"
 	"path/filepath"
 	"sync"
 	"time"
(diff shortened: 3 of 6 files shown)


pkg/placement/raft/server.go
```diff
@@ -214,15 +214,15 @@ func (s *Server) StartRaft(ctx context.Context, sec security.Handler, config *ra
 		// Set default configuration for raft
 		s.config = &raft.Config{
 			ProtocolVersion:    raft.ProtocolVersionMax,
-			HeartbeatTimeout:   1000 * time.Millisecond,
-			ElectionTimeout:    1000 * time.Millisecond,
-			CommitTimeout:      50 * time.Millisecond,
+			HeartbeatTimeout:   2 * time.Second,
```
DeepanshuA, line 217: Does increasing this time solve issue majorly or it's a temporary solution?

ItalyPaleAle: I would love to understand _why_ this is hitting the timeout under load. Or there's always the risk that even with the timeout increased, it may be hit under heavier load.

Also, perhaps these should be configurable as CLI flags?

artursouza: This will be the new default going forward. I believe this should be configurable but we need to unblock users to deploy 1.12 to production. It is temporary in the sense that the ideal solution is to change how actor placement is discovered - but users cannot wait for that. We are fixing the plane while flying it.

artursouza: I want user to try an RC before we release this. I am working on a new test for Dapr but I need more time. We have a timing problem now.

artursouza: I suggest we make this a 1.12.4-rc.1 release first.

mukundansundar:
@artursouza Please wait to merge this after 1.12.3 is cut, since this is not going in for 1.12.3 ....

daixiang0 (approved):

charts/dapr/charts/dapr_placement/templates/dapr_placement_statefulset.yaml
```diff
@@ -186,6 +186,8 @@ spec:
             valueFrom:
               fieldRef:
                 fieldPath: metadata.namespace
+          - name: PLACEMENT_RAFT_SETTINGS
+            value: v2
```
artursouza, line 190: I am allowing users to revert to v1 in case the new setting shows issues at scale.

charts/dapr/charts/dapr_sidecar_injector/templates/dapr_sidecar_injector_deployment.yaml
```diff
@@ -154,6 +154,11 @@ spec:
 {{- if not .Values.global.actors.enabled }}
         - name: SKIP_PLACEMENT
           value: "true"
+{{- else }}
+{{- if .Values.global.ha.enabled }}
+        - name: PLACEMENT_ADDRESSES
+          value: "dapr-placement-server-0.dapr-placement-server.{{ .Release.Namespace }}.svc.cluster.local:50005,dapr-placement-server-1.dapr-placement-server.{{ .Release.Namespace }}.svc.cluster.local:50005,dapr-placement-server-2.dapr-placement-server.{{ .Release.Namespace }}.svc.cluster.local:50005"
```
artursouza, line 160: This is important for the sidecar to pick the endpoint instead of relying on the k8s service to pick which one to route to.

artursouza:
I am creating a separate PR for a new test I wrote for this but I could not reproduce the issue, it does not happen deterministically and requires a production scale cluster: > 100 pods.

pkg/placement/raft/server.go
artursouza: I added a flag to pick the timeout settings.

charts/dapr/charts/dapr_sidecar_injector/templates/dapr_sidecar_injector_deployment.yaml
```diff
@@ -154,6 +154,11 @@ spec:
 {{- if not .Values.global.actors.enabled }}
         - name: SKIP_PLACEMENT
           value: "true"
+{{- else }}
+{{- if .Values.global.ha.enabled }}
```
ItalyPaleAle, line 158: This needs to take into account the cluster DNS suffix, it may not always be `cluster.local`. Must use ``.Values.global.dnsSuffix``

However you may be able to get away without hardcoding this, by doing a DNS resolution on the service, assuming it has ClusterIP=None?

pkg/placement/raft/server.go
```diff
@@ -212,17 +213,35 @@ func (s *Server) StartRaft(ctx context.Context, sec security.Handler, config *ra
...
-			HeartbeatTimeout:   1000 * time.Millisecond,
-			ElectionTimeout:    1000 * time.Millisecond,
-			CommitTimeout:      50 * time.Millisecond,
-			MaxAppendEntries:   64,
-			ShutdownOnRemove:   true,
-			TrailingLogs:       10240,
-			SnapshotInterval:   120 * time.Second,
-			SnapshotThreshold:  8192,
-			LeaderLeaseTimeout: 500 * time.Millisecond,
+		if os.Getenv("PLACEMENT_RAFT_SETTINGS") == "v2" {
```
ItalyPaleAle, line 216: Do we still need this given that the bug was unrelated?

Also, this makes v1 the default, do we want to default to v2?

charts/dapr/charts/dapr_sidecar_injector/templates/dapr_sidecar_injector_deployment.yaml
JoshVanL: > instead of relying on the k8s service to pick which one to route to.
Could you elaborate on this a bit? Placement is using a [headless](https://github.com/dapr/dapr/blob/b99cf8b5476a503f655d74ce478e681a1f9a37f0/charts/dapr/charts/dapr_placement/templates/dapr_placement_service.yaml#L26) service which will return the same thing but will be more correct, i.e. correct port, removes unhealthy hosts.

mukundansundar: How does this value interact with the Placement host address annotation that can be set in the app?

pkg/placement/raft/server.go
artursouza: The goal for this change is to reduce the change in actor leadership. It is part of the solution. I made v2 as the default via the helm chart. I can make it the default here instead of people prefer.

charts/dapr/charts/dapr_sidecar_injector/templates/dapr_sidecar_injector_deployment.yaml
artursouza: OK. I will try the dnsSuffix.

charts/dapr/charts/dapr_sidecar_injector/templates/dapr_sidecar_injector_deployment.yaml
artursouza: Responding to Josh. Kubernetes has no idea about which one is the leader, so the sidecars have to keep reconnecting to the service until one of them accepts as the leader. Only the leader node can accept the streaming connections and do placement table disseminations. Sidecar already has logic to try multiple addresses until it finds the leader.

artursouza: > How does this value interact with the Placement host address annotation that can be set in the app?

It does not make a difference because I changed the default value, way before the annotation is handled.

JoshVanL: Indeed however daprd currently handles this by DNS querying the single headless service, then finds the leader through round robin on the returned set.

https://github.com/dapr/dapr/blob/b99cf8b5476a503f655d74ce478e681a1f9a37f0/pkg/actors/placement/placement.go#L129
https://github.com/dapr/dapr/blob/b99cf8b5476a503f655d74ce478e681a1f9a37f0/pkg/actors/placement/placement.go#L596C6-L596C26
https://github.com/dapr/dapr/blob/b99cf8b5476a503f655d74ce478e681a1f9a37f0/pkg/actors/placement/placement.go#L137
https://github.com/dapr/dapr/blob/b99cf8b5476a503f655d74ce478e681a1f9a37f0/pkg/actors/placement/grpc_opts.go#L65-L69

pkg/placement/raft/server.go
artursouza: Default to v2 in code now.

charts/dapr/charts/dapr_sidecar_injector/templates/dapr_sidecar_injector_deployment.yaml
artursouza: Using dnsSuffix now.

charts/dapr/charts/dapr_sidecar_injector/templates/dapr_sidecar_injector_deployment.yaml
artursouza: Great. I will remove this from the sidecar injector. Learned something new. Thanks.

charts/dapr/charts/dapr_sidecar_injector/templates/dapr_sidecar_injector_deployment.yaml
artursouza: Removed since Josh showed me that this is handled in the sidecar via dns resolution.

charts/dapr/charts/dapr_sidecar_injector/templates/dapr_sidecar_injector_deployment.yaml
artursouza: Resolving this comment since I deleted this from the solution.

JoshVanL (approved):

cicoyle (approved):
LGTM

mikeee (approved):

artursouza:
This PR was reverted and PR #7349 replaces this.

PR #7313, opened 2023-12-18
Remove daprClient in workflow e2e test

MregXN:
# Description

please refer to issue [#1205](https://github.com/dapr/dotnet-sdk/issues/1205) in dapr/dotnet-sdk.

### DIFF
--- a/tests/apps/workflowsapp/Controller.cs
+++ b/tests/apps/workflowsapp/Controller.cs
@@ -15,6 +15,7 @@ using Microsoft.AspNetCore.Mvc;
 using System.Threading.Tasks;
 using Dapr.Client;
 using System;
+using Dapr.Workflow;
 
 namespace DaprDemoActor
 {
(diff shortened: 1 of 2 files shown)


tests/apps/workflowsapp/Controller.cs
```diff
@@ -15,75 +15,67 @@
...
-
-      return getResponse.InstanceId;
+              instanceId: instanceID);
+      var getResponse = await daprClient.GetWorkflowStateAsync(instanceID);
+      return startResponse;
     }
 
     [HttpPost("PurgeWorkflow/{workflowComponent}/{instanceID}")]
-    public async Task<ActionResult<bool>> PurgeWorkflow([FromRoute] string instanceID, string workflowComponent)
+    public async Task<ActionResult<bool>> PurgeWorkflow([FromServices]DaprWorkflowClient daprClient,[FromRoute] string instanceID, string workflowComponent)
```
cgillum, line 48: A few things we need to fix in all these methods:
1. Remove `workflowComponent` from both the route and the parameter list. You may also need to fix the code which invokes these routes to remove their use of `workflowComponent`.
1. Rename `daprClient` to `workflowClient` to avoid confusion with the `DaprClient` class, which we're not using anymore.
2. Please fix the lack of whitespace characters here

Here's an example of what these method signatures should look like:
```csharp
[HttpPost("PurgeWorkflow/{instanceID}")]
public async Task<ActionResult<bool>> PurgeWorkflow([FromServices] DaprWorkflowClient daprClient, [FromRoute] string instanceID)
```

tests/apps/workflowsapp/Controller.cs
```diff
@@ -15,75 +15,67 @@
 using System.Threading.Tasks;
 using Dapr.Client;
 using System;
+using Dapr.Workflow;
```
cgillum, line 18: nit: please sort the `using` statements alphabetically.

cgillum:
Also tagging @RyanLettieri on this PR since he's the original author of the workflow E2E test.

tests/apps/workflowsapp/Controller.cs
MregXN: Thanks!

tests/apps/workflowsapp/Controller.cs
MregXN: Thanks!

MregXN:
this pr will be ready for review after this [issue](https://github.com/dapr/dotnet-sdk/issues/1215) in dotnet-sdk is fixed.

PR #7314, opened 2023-12-18
Load component for control plane service as well, based on filters

DeepanshuA:
# Description

This PR lets control plane service get required components loaded.
The component should have namespace as control plane namespace AND one scope should be defined with control plane service name.
This component, meant for Control Plane, won't be visible to sidecar and vice-versa.

## Issue reference

Please reference the issue this PR is linked to: #7203 

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing


JoshVanL (changes requested):
@DeepanshuA please can you describe what the use case of this PR is trying to accomplish? If the intention is to limit which components are sent to the client based on scope, this _must_ be done using the X.509 client SPIFFE identity not the provided client string which is _not_ verifiable. We already have a PR open to authorize the requested namespace matches the client identity here https://github.com/dapr/dapr/pull/7111

It is not the case that all services running in the control plane namespace are control plane

DeepanshuA:
> @DeepanshuA please can you describe what the use case of this PR is trying to accomplish? If the intention is to limit which components are sent to the client based on scope, this _must_ be done using the X.509 client SPIFFE identity not the provided client string which is _not_ verifiable. We already have a PR open to authorize the requested namespace matches the client identity here #7111
> 
> It is not the case that all services running in the control plane namespace are control plane

hi @JoshVanL, This PR is for letting dapr-operator being able to send components list to a control plane service as well.

This is part of work for #7203. Here, the control plane service name can vary (as per implementation for actors/reminders) and that's why it uses:
1. "dapr-" prefix in pod name to understand if it is a control plane service.
2. Namespace comparison.

In addition, the component defined would have to be:
1. namespaced (same as control plane namespace)
2. Scope defined with prefix "dapr-"

Screenshot 2023-12-18 at 5 09 29 PM

JoshVanL:
Hi @DeepanshuA 
There needn't be any special casing for control plane components retrieving components in the control plane namespace. As clients, they are treated no differently to daprds. Their spiffe ids (app ids) are the same, e.g ns/dapr-system/dapr-placement

JoshVanL:
The he control plane service id shouldn't change, even if the implementation is different.

ItalyPaleAle (changes requested):
One small change. What we discussed was that the scoping for CP services should be in the format `dapr:<foo>` (while the pod name is `dapr-<foo>`). Because the colon isn't valid in an app ID for sidecars, then there's no risk sidecars could accidentally receive those components.

pkg/operator/api/api.go
```diff
@@ -192,7 +200,23 @@ func (a *apiServer) ListComponents(ctx context.Context, in *operatorv1pb.ListCom
 		Components: [][]byte{},
 	}
 	for i := range components.Items {
+		// By default assume that it is not a component for control plane service.
+		controlPlaneComp := false
 		c := components.Items[i] // Make a copy since we will refer to this as a reference in this loop.
+		// If the component is in the same namespace as the operator AND has a scope defined with prefix dapr-, then it is a control plane component.
```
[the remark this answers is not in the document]
JoshVanL: Control plane components have identity today so there needn't be any special casing for them. I'd definitely be in favor of adding logic on the server side to filter out Components with a scope that don't match the client, however this must come from the gRPC context identity.

It seems like we are making magic scope values here and as a reviewer it is not clear why we are doing this, and why the identity we have today cannot be used to achieve what we want. I'd like to see a proposal so we can make a decision on a significant API change.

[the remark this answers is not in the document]
ItalyPaleAle: My comment isn't about the identity passed to the operator, but rather the value of the scope. To make sure that components scoped to control plane services aren't sent to sidecars, a simple solution is to use the scope `dapr:<foo>` which is invalid for sidecars.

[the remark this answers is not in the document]
JoshVanL: What is the reason that we would like to make a distinction between control plane and sidecar clients, rather than using the control plane identity itself, i.e. `dapr-placement`, `dapr-sentry`, etc.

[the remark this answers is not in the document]
ItalyPaleAle: Correct me if i'm wrong. scopes are not authenticated. an app (sidecar) can choose app-id "dapr-sentry" and it would then be able to receive components scoped for that.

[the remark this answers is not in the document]
JoshVanL: Correct, there is currently _no_ authorization that is happening on the operator side for controlling what component resources are sent to clients. Sidecars are currently being sent all resources in that namespace. The sidecar is currently self electing whether the component should be `init`ed or not. I'm definitely in favor of this authorization happening in the operator, and for all resource types.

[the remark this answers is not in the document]
ItalyPaleAle: Thanks for confirming. So, the idea was that using `dapr:<foo>` as scope was a cheap but effective way to obtain some "access control", since no sidecar can have an app-id with a `:`.

Other options considered included:

- Adding a new field to the Component CRD such as "controlPlane" to indicate it's for control plane services only. This however requires changing the CRD which is not ideal
- Ignoring access control, but that may allow sidecars to get access to components they shouldn't see. Assuming the component type isn't supported by the sidecar the risk shouldn't be significant, but it still felt like it'd be nicer to have some separation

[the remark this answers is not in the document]
JoshVanL: I think we agree that we should have server side authorization on scoped resources and we should implement this as it should exist today already. If the fundamental intention of this PR is to to do this, then we can merge https://github.com/dapr/dapr/pull/7111, and then follow on to include a function to filter non-matching scopes.

I question why we need any concept of "control plane" service as a scope `group` concept when we have a well defined list of control plane services and their identities. Equally, we need not need to separate or special case control plane and side car identities at all- they are both represented as SPIFFE IDs when connecting to peers over gRPC mTLS.

As a reviewer I cannot understand where the motivation is coming from where we need non-known control plan services which require loading of components- this should be raised as a proposal. Remember that control plane service identities _must_ be known before deployment or must be discoverable, else mTLS  fundamentally itself will not work.

ItalyPaleAle (approved):
LGTM!

JoshVanL:
This PR needs some more work before being merged; I think there is still more discussion that needs to be had about why we need a scope group added for control plane components over using the control plane identities themselves. We need to expand the scopes API change to all resource endpoints, specifically `ComponentUpdate`, `ListComponents`, `GetConfiguration`, `ListSubscriptions`, `GetResiliency`, `ListResiliency`, `ListSubscriptionsV2`, `ListHTTPEndpoints`, `HTTPEndpointUpdate`. We should also add integration tests to cover the new code and API changes, as well as documentation describing this scope group.

ItalyPaleAle:
> We need to expand the scopes API change to all resource endpoints, specifically `ComponentUpdate`, `ListComponents`, `GetConfiguration`, `ListSubscriptions`, `GetResiliency`, `ListResiliency`, `ListSubscriptionsV2`, `ListHTTPEndpoints`, `HTTPEndpointUpdate`

The other methods are not necessary, just needs to get components (and not even subscribe)

artursouza (changes requested):
Before I can review this PR in detail I need some clarification.

What problem is this solving? Why does control plane needs to load components? Like was mentioned here, a proposal will be needed. As of now, the distributed scheduler service in POC stage does not plan to load any component. Even if we decide that it is needed there, it is still a proposal in review.

What integration and E2E tests are being included? Unit tests is not enough to guarantee quality.

artursouza:
Marking this as do-not-merge temporarily until there is further clarification.

DeepanshuA:
As a user, if I am facing issues due to placement service and reminders, then, I just want to be able to use Dapr core functionality of Actors/reminders i.e. leverage Actors/reminders interfaces BUT I want to implement these interfaces in a new Control Plane service, which as a user, I should be allowed to design as per the way it solves issues for me.
Infact, that is the basis of issue #7203, for which I believe Maintainers already have been having some discussions. 

Now, this Control Plane Service may require some component, like a database etc. which I should be able to load, using Operator. But, components visible to a Control Plane service should not be visible to Daprd sidecar and vice-versa.
Hence, just a logical grouping is required.
Neither it intends to solve any security issue, nor it introduces any.

For E2E test, Dapr doesn't have any Control Plane service right now which loads components, so I really don't know how to add an E2E test for a CP loading components. Any ideas here?
In E2E, It is already being tested that Daprd sidecar is able to load components. What I can add in a new/existing E2E test is that it is not able to load a component, which is for CP.

I believe same would be apply for Integration test, will check that out.

DeepanshuA:
> Before I can review this PR in detail I need some clarification.
> 
> What problem is this solving? Why does control plane needs to load components? Like was mentioned here, a proposal will be needed. As of now, the distributed scheduler service in POC stage does not plan to load any component. Even if we decide that it is needed there, it is still a proposal in review.
> 
> What integration and E2E tests are being included? Unit tests is not enough to guarantee quality.

As a user, if I am facing issues due to placement service and reminders, then, I just want to be able to use Dapr core functionality of Actors/reminders i.e. leverage Actors/reminders interfaces BUT I want to implement these interfaces in a new Control Plane service, which as a user, I should be allowed to design as per the way it solves issues for me.
Infact, that is the basis of issue https://github.com/dapr/dapr/issues/7203, for which I believe Maintainers already have been having some discussions.

Now, this Control Plane Service may require some component, like a database etc. which I should be able to load, using Operator. But, components visible to a Control Plane service should not be visible to Daprd sidecar and vice-versa.
Hence, just a logical grouping is required.
Neither it intends to solve any security issue, nor it introduces any.

For E2E test, Dapr doesn't have any Control Plane service right now which loads components, so I really don't know how to add an E2E test for a CP loading components. Any ideas here?
In E2E, It is already being tested that Daprd sidecar is able to load components. What I can add in a new/existing E2E test is that it is not able to load a component, which is for CP.

I believe same would be apply for Integration test, will check that out.

DeepanshuA:
Thanks @JoshVanL for letting me borrow your tests/util/pki.go code from #7111 to mock properly context with proper grpccredentials.
I was following https://github.com/spiffe/go-spiffe/blob/16893f62d83d74be95036e818955b9bcb5c7d35e/v2/spiffegrpc/grpccredentials/credentials_test.go but it was taking time to mimic on same lines, using pki.go with updated version of 7111 was lot faster.

cmd/operator/main.go
```diff
@@ -57,6 +57,8 @@ func main() {
 		WatchdogCanPatchPodLabels:           opts.WatchdogCanPatchPodLabels,
 		APIPort:                             opts.APIPort,
 		HealthzPort:                         opts.HealthzPort,
+		ActorsServiceName:                   opts.ActorsServiceName,
```
ItalyPaleAle, line 60: How about we pass a more generic "additional-control-plane-services" flag? Which can use `[]string` (we use pflag now so we can use `[]string` and it supports repeated values or CSV)

This way it's more generic for the future

PS: We should probably add the same flag for Sentry so it can issue certs without hardcoding the control plane services names

ItalyPaleAle (changes requested):

JoshVanL:
I believe it is only necessary for us to need to expose the address of the service which implements an API. I do not think it is necessary and don't think we should have a CLI flag which determines the list of control plane component identities. Control plane identities e.g. `placement`, `scheduler`, `operator` etc. never change- they are well defined services that have an identity and expose some kind of API that the service type implements. When a different implementation of this service is deployed, it is expected that it might be the case the hostname/address of the service changes and thus needs to be reflected in the configuration, however fundamentally the identity of that service has _not_ changed. By making the the service identity configurable, we not longer have a mapping of `API&identity` -> `address`, but instead have a mapping of `API` -> `identity` -> `address` removing the point of having an service identity at all.

Adding the ability to manipulate the list of known control plane services has a couple issues- one being that this increases the risk of bad configuration causing a security incident whereby an application gains (or is prevented from) getting a control plane identity which is highly privileged. The other is configuration sprawl whereby the set of additional control plane components now needs to be plumbed to all Dapr services i.e. Sentry, injector, operator, daprds.. When a peer wants to make an mTLS connection to a placement service as an example, the code will no longer ensure it is connecting the `dapr-placement`, but instead to a variable which poses another security risk by programming error when plumbing the service name.

cmd/operator/main.go
DeepanshuA: > How about we pass a more generic "additional-control-plane-services" flag? Which can use `[]string` (we use pflag now so we can use `[]string` and it supports repeated values or CSV)
> 
> This way it's more generic for the future
> 
> PS: We should probably add the same flag for Sentry so it can issue certs without hardcoding the control plane services names

I have incorporated changes

ItalyPaleAle:
I see the points @JoshVanL .

I do think that having a flag that is more generic such as “additional control plane services” for both Operator and Sentry is the more future-proof option.

You are mentioning having different implementations for placement and scheduler. In some cases those may be the same application, so that would be a problem. Also, I share Deepanshu’s concerns that overloading a name with 2 different implementations may be confusing to end users.

As from the configuration point of view, the Operator is only deployed on K8s and via Helm. Users are not expected to manually run the Operator and set CLI flags. We wouldn’t need to expose additional configuration to users, besides the “global” option “tell me which reminder service you want to use”

JoshVanL:
> I do think that having a flag that is more generic such as “additional control plane services” for both Operator and Sentry is the more future-proof option.

Since the control plane is an atomic versioned deployment, it should never be the case that we need to introduce an "escape hatch" to allow control plane components to learn of new control plane components which are not part of their known set at that Dapr version.

> You are mentioning having different implementations for placement and scheduler. In some cases those may be the same application, so that would be a problem. Also, I share Deepanshu’s concerns that overloading a name with 2 different implementations may be confusing to end users.

I don't think that the identity list needs to be bubbled up to the user at all- simply setting the hostname is enough and is in-fact simpler for the user for writing configuration. With this PRs approach, a user will need to set both the identity _and_ the host name of an API service. To be clear, I'm not suggesting that a user must deploy their custom scheduler service using the exact same Deployment name, Service name, Service Account etc. only that the Deployment _must_ contain the [control plane service identity annotation](https://github.com/dapr/dapr/blob/70e3f699793bb138d927079682261999b010e40e/pkg/sentry/server/validator/kubernetes/kubernetes.go#L242) declaring that it has the identity of that service type. I don't understand what we gain by allowing users to set which ever identity they like here, when it will need to be copied across to the Dapr configuration anyway and replace the default target control plane identity. I think it would make sense to a user that they need to assign the `scheduler` identity to their custom scheduler service deployment.

> As from the configuration point of view, the Operator is only deployed on K8s and via Helm. Users are not expected to manually run the Operator and set CLI flags. We wouldn’t need to expose additional configuration to users, besides the “global” option “tell me which reminder service you want to use”

Sure, but the identity which is being used by a particular service _will_ need to be propagated everywhere. mTLS doesn't work unless the client knows the identity of the peer beforehand. If the client doesn't know what the identity is beforehand, we loose the strong mutual authentication that mTLS provides and allows for impersonation of control plane services.

charts/dapr/charts/dapr_operator/templates/dapr_operator_deployment.yaml
```diff
@@ -171,6 +171,14 @@ spec:
 {{- end }}
 {{- if .Values.global.operator.watchdogCanPatchPodLabels }}
         - "--watchdog-can-patch-pod-labels"
+{{- end }}
+{{- if .Values.global.actors.enabled }}
```
ItalyPaleAle, line 175: Should this be `.serviceName`?

```suggestion
{{- if .Values.global.actors.serviceName }}
```

charts/dapr/charts/dapr_sentry/templates/dapr_sentry_deployment.yaml
```diff
@@ -176,6 +176,14 @@ spec:
 {{- end }}
         - "--trust-domain"
         - {{ .Values.global.mtls.controlPlaneTrustDomain }}
+{{- if .Values.global.actors.enabled }}
```
ItalyPaleAle, line 179: Here too?
```suggestion
{{- if .Values.global.actors.serviceName }}
```

cmd/operator/options/options.go
```diff
@@ -80,6 +81,8 @@ func New() *Options {
 	flag.IntVar(&opts.APIPort, "port", 6500, "The port for the operator API server to listen on")
 	flag.IntVar(&opts.HealthzPort, "healthz-port", 8080, "The port for the healthz server to listen on")
 
+	flag.Var(&opts.AdditionalCPServices, "additional-control-plane-services", "Name of the additional control plane services, if any")
```
ItalyPaleAle, line 84: Nit: should this be `additional-control-plane-service` since each value can contain a single service?

cmd/sentry/options/options.go
```diff
@@ -77,6 +78,7 @@ func New(origArgs []string) *Options {
 	fs.StringVar(&opts.TrustDomain, "trust-domain", "localhost", "The CA trust domain")
 	fs.IntVar(&opts.Port, "port", config.DefaultPort, "The port for the sentry server to listen on")
 	fs.IntVar(&opts.HealthzPort, "healthz-port", 8080, "The port for the healthz server to listen on")
+	fs.StringSliceVar(&opts.AdditionalCPServices, "additional-control-plane-services", []string{"placement"}, "Name of the additional control plane services, if any")
```
ItalyPaleAle, line 81: Here too, nit: should be `-service`

pkg/operator/api/api.go
```diff
@@ -182,6 +187,26 @@ func (a *apiServer) GetConfiguration(ctx context.Context, in *operatorv1pb.GetCo
...
+		log.Debugf("failed to get SPIFFE ID from gRPC connection context: %v", err)
+		return nil, err
+	}
+	if !ok {
+		// Apply the default action
+		log.Debugf("Error while reading spiffe id from client cert. applying default global policy action")
+	}
+
+	if in.GetNamespace() == security.CurrentNamespace() {
+		for _, service := range a.additionalCPServices {
```
ItalyPaleAle, line 203: I think here we need to list all control plane services. The "additional" ones are ...additional. But we should have the built-in ones too, such as sentry & placement (not that it's needed today, but for completness)

pkg/security/x509source.go
```diff
@@ -476,18 +485,29 @@ func renewalTime(notBefore, notAfter time.Time) time.Time {
...
 		"dapr-placement",
 		"dapr-injector",
 		"dapr-sentry":
 		return true
-	default:
-		return false
 	}
+	if additionalCPServices != nil {
+		return contains(id, additionalCPServices)
+	}
```
ItalyPaleAle, line 509: Nit: the stdlib has a built-in method for this:


```suggestion
	if len(additionalCPServices) != 0 {
		return slices.Contains(additionalCPServices, id)
	}
```

pkg/security/x509source.go
```diff
@@ -476,18 +485,29 @@ func renewalTime(notBefore, notAfter time.Time) time.Time {
...
+			return true
+		}
+	}
+	return false
+}
+
 // isControlPlaneService returns true if the app ID corresponds to a Dapr
 // control plane service.
-func isControlPlaneService(id string) bool {
+func isControlPlaneService(id string, additionalCPServices []string) bool {
```
ItalyPaleAle, line 499: Maybe this method can be moved to a shared package so the operator can use it too. This has the hardcoded ones (sentry, placement, injector) + supports the additional ones.

charts/dapr/charts/dapr_operator/templates/dapr_operator_deployment.yaml
DeepanshuA: Tried to check `enabled` actually for actors and thought that if `enabled` is true, then serviceName is provided. But, I am fine checking serviceName directly. Done the change.

pkg/operator/api/api.go
DeepanshuA: Had not added them - but as you said, for the sake of completeness - added them and as per another comment, shifted to utils, to be used by operator and x509source.

pkg/security/x509source.go
DeepanshuA: shifted func to utils

pkg/operator/api/api.go
```diff
@@ -40,21 +40,25 @@ import (
...
 )
 
 const (
-	APIVersionV1alpha1    = "dapr.io/v1alpha1"
-	APIVersionV2alpha1    = "dapr.io/v2alpha1"
-	kubernetesSecretStore = "kubernetes"
+	APIVersionV1alpha1        = "dapr.io/v1alpha1"
+	APIVersionV2alpha1        = "dapr.io/v2alpha1"
+	kubernetesSecretStore     = "kubernetes"
+	controlPlanePodNamePrefix = "dapr-"
```
ItalyPaleAle, line 52: I believe this is now-unused, so can be removed
```suggestion
```

pkg/security/x509source.go
```diff
@@ -46,11 +46,13 @@ import (
...
 	sentryToken "github.com/dapr/dapr/pkg/security/token"
+	"github.com/dapr/dapr/utils"
 )
 
 const (
-	sentrySignTimeout = time.Second * 3
-	sentryMaxRetries  = 5
+	sentrySignTimeout         = time.Second * 3
+	sentryMaxRetries          = 5
+	controlPlanePodNamePrefix = "dapr-"
```
ItalyPaleAle, line 55: I believe this is now-unused, so can be removed
```suggestion
```

utils/utils.go
```diff
@@ -146,3 +148,26 @@ func GetNamespaceOrDefault(defaultNamespace string) string {
...
+	if additionalCPServices != nil {
+		prefixedCPServices = make([]string, len(additionalCPServices))
+		for i, service := range additionalCPServices {
+			prefixedCPServices[i] = controlPlanePodNamePrefix + service
+		}
+	}
+	if len(prefixedCPServices) != 0 {
+		return slices.Contains(prefixedCPServices, id)
+	}
+	return false
```
ItalyPaleAle, line 172: Small improvement that avoids allocating a new slice
```suggestion

	if len(additionalCPServices) == 0 {
		return false
	}

	after, found := strings.CutPrefix(id, controlPlanePodNamePrefix)
	return found && slices.Contains(additionalCPServices, after)
```

ItalyPaleAle (approved):

DeepanshuA:
I have opened #7359 instead of this PR, as the way it used to work is quite different.
Will close this one, once #7359 is merged.

PR #7322, opened 2023-12-19
PubSub API: standardize err msgs

cicoyle:
# Description

Making this a draft since I need to add integration tests.

This is the code to standardize many of the PubSub api errors to use the newly merged kit pkg for standardizing errors around the richer error model based on this proposal: https://github.com/dapr/proposals/blob/main/0009-BCIRS-error-handling-codes.md.

I added todos for a follow up PR to enrich more of the pubsub errors, as there were already so many that were updated in this PR.

# Issue Reference
https://github.com/dapr/dapr/issues/6068

## Checklist
* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Unit tests passing
* [x] Extended the documentation: Elena has a PR for it: https://github.com/dapr/docs/pull/3908


pkg/http/api_test.go
```diff
@@ -302,7 +303,7 @@ func TestPubSubEndpoints(t *testing.T) {
 			// assert
 			assert.Equal(t, 403, resp.StatusCode, "unexpected success publishing with %s", method)
 			assert.Equal(t, "ERR_PUBSUB_FORBIDDEN", resp.ErrorBody["errorCode"])
-			assert.Equal(t, "topic topic is not allowed for app id test", resp.ErrorBody["message"]) //nolint:dupword
+			assert.Equal(t, "topic topic is not allowed for app id fakeAPI", resp.ErrorBody["message"]) //nolint:dupword
```
cicoyle, line 306: this should be `fakeAPI`, as if you look that is the app id that is set for the test. `test` is the pubsub name I believe, so I made it the true app id name.

sadath-12:
@cicoyle  had a thought like instead of keeping under ```pkg/api``` we could keep it in ```pkg/utils``` something like that so we don't make the packages confusing between ```pkg/api``` and ```pkg/apis```

go.mod
```diff
@@ -462,10 +462,11 @@ replace (
 // check for retracted versions: go list -mod=mod -f '{{if .Retracted}}{{.}}{{end}}' -u -m all
 replace github.com/microcosm-cc/bluemonday => github.com/microcosm-cc/bluemonday v1.0.24
 
-// Uncomment for local development for testing with changes in the components-contrib repository.
+// Uncomment for local development for testing with changes in the components-contrib && kit repositories.
```
cicoyle, line 465: From a new contributor perspective I think it would be nice to have kit mentioned here in the off chance that a user is testing a change in kit. It is more explicit and easier on developers to uncomment a line rather than add it in. Not sure how we can disable the failing CI check on me adding this in?

ItalyPaleAle: It's here: https://github.com/dapr/dapr/blob/master/.github/scripts/check_go_mod.mjs

cicoyle: I'll pull this into a separate PR 👍🏻

cicoyle:
> @cicoyle had a thought like instead of keeping under `pkg/api` we could keep it in `pkg/utils` something like that so we don't make the packages confusing between `pkg/api` and `pkg/apis`

We decided to go with this location, as we want to avoid `pkg/utils` becoming a catch all location of flat packages.

pkg/api/errors/pubsub.go
```diff
@@ -0,0 +1,204 @@
...
+	"fmt"
+	"net/http"
+
+	grpcCodes "google.golang.org/grpc/codes"
+
+	kiterrors "github.com/dapr/kit/errors"
+)
+
+const (
+	PostFixNameEmpty = "NAME_EMPTY"
```
daixiang0, line 26: What is the meaning of PostFix?

pkg/api/errors/pubsub.go
```diff
@@ -0,0 +1,204 @@
...
+)
+
+func PubSubNotFound(name string, pubsubType string, metadata map[string]string) error {
```
daixiang0, line 29: Could we remove `PubSub` prefix as it is in pubsub.go?

daixiang0, line 29:
```suggestion
func NotFound(component, name string, type string, metadata map[string]string) error {
```

If we use something like this, we can reduce many duplicate codes.

pkg/runtime/pubsub/errors.go
```diff
@@ -20,7 +20,7 @@ import (
 	"github.com/dapr/dapr/pkg/messages"
 )
 
-var ErrMessageDropped = errors.New("pubsub message dropped")
+var ErrMessageDropped = errors.New("pubsub message dropped") // TODO: remove this and use apierrors.PubSubMsgDropped
```
artursouza, line 23: todo?

cicoyle: yes, this error was not created in my PR, so I just added it as an explicit todo since  ~13 errors were already being updated in this PR

pkg/api/errors/pubsub.go
cicoyle: We could make the errors more generic, but then we would have to have more parameters in the function since different APIs have different legacy tags or error codes in general that users depend on. We can't change those legacy values because we don't want to break users. For example, [StateStoreNotFound](https://github.com/dapr/dapr/pull/7257/files#diff-8d5b272f8b9930484ae10ac077dc7ee4c0656649270f3b1f7b3054b80ea92130R37), and PubSubNotFound return different values to users for legacy tags, http codes, etc. 

@elena-kolevska what are your thoughts? We would have to coordinate this change and try to combine as many as we could by making the functions general and pass in the values we need.

pkg/api/errors/pubsub.go
cicoyle: done

pkg/api/errors/pubsub.go
elena-kolevska: Yes, exactly, the NotFound errors for different components are different because we include the component name and details.
I think merging them into a single, parametrised function will complicate the developer experience, without a big benefit.

yaron2: Agree

yaron2: > Could we remove PubSub prefix as it is in pubsub.go?

Its in `pubsub.go` but the different files containing the error codes are under the same package, so method names would conflict.

cicoyle: I went ahead and made a generic `NotFound` && `NotConfigured` func to be used by both state store and pubsub

cicoyle: generic errors can be found in: `pkg/api/errors/errors.go`

pkg/api/grpc/grpc.go
```diff
@@ -122,29 +126,37 @@ func NewAPI(opts APIOpts) API {
...
 
 	rawPayload, metaErr := contribMetadata.IsRawPayload(reqMeta)
 	if metaErr != nil {
-		return nil, "", "", false, messages.ErrPubSubMetadataDeserialize.WithFormat(metaErr)
+		err = apierrors.PubSubMetadataDeserialize(pubsubName, pubsubType, reqMeta, metaErr)
+		return nil, "", "", false, err
 	}
 
-	return thepubsub.Component, pubsubName, topic, rawPayload, nil
+	return thepubsub.Component, pubsubName, topic, rawPayload, err
```
ItalyPaleAle, line 159:
```suggestion
	return thepubsub.Component, pubsubName, topic, rawPayload, nil
```

pkg/api/grpc/grpc.go
```diff
@@ -19,11 +19,16 @@ import (
...
+	"net/http"
 	"strconv"
 	"sync"
 	"sync/atomic"
 	"time"
 
+	kiterrors "github.com/dapr/kit/errors"
+
+	apierrors "github.com/dapr/dapr/pkg/api/errors"
+
```
ItalyPaleAle, line 31: Nit: please ensure imports are sorted in 3 blocks only:

1. Standard library
2. External packages
3. github.com/dapr/*

pkg/api/grpc/grpc.go
```diff
@@ -426,16 +441,19 @@ func (a *api) BulkPublishEventAlpha1(ctx context.Context, in *runtimev1pb.BulkPu
...
 		if errors.As(err, &runtimePubsub.NotAllowedError{}) {
-			nerr = status.Errorf(codes.PermissionDenied, err.Error())
+			nerr = apierrors.PubSubPublishForbidden(pubsubName, string(contribMetadata.PubSubType), topic, a.AppID(), err)
 		}
 
 		if errors.As(err, &runtimePubsub.NotFoundError{}) {
-			nerr = status.Errorf(codes.NotFound, err.Error())
+			nerr = apierrors.PubSubTestNotFound(pubsubName, string(contribMetadata.PubSubType), topic, err)
 		}
+
```
ItalyPaleAle, line 456: Nit: please use a switch statement for readability with the 3 cases:

1. `errors.As(err, &runtimePubsub.NotAllowedError{})`
2. `errors.As(err, &runtimePubsub.NotFoundError{})`
3. default

Also, shouldn't this be using `errors.Is` rather than `As`, since you're not reading from the "casted" error?

pkg/api/http/http.go
```diff
@@ -25,6 +25,8 @@ import (
 	"strings"
 	"time"
 
+	kiterrors "github.com/dapr/kit/errors"
```
ItalyPaleAle, line 28: Same here regarding blocks

pkg/api/http/http.go
```diff
@@ -1619,22 +1620,19 @@ func (a *api) onPublish(reqCtx *fasthttp.RequestCtx) {
 	diag.DefaultComponentMonitoring.PubsubEgressEvent(context.Background(), pubsubName, topic, err == nil, elapsed)
 
 	if err != nil {
-		status := nethttp.StatusInternalServerError
-		msg := NewErrorResponse("ERR_PUBSUB_PUBLISH_MESSAGE",
-			fmt.Sprintf(messages.ErrPubsubPublishMessage, topic, pubsubName, err.Error()))
+		var nerr error
+		nerr = apierrors.PubSubPublishMessage(pubsubName, string(contribMetadata.PubSubType), topic, err)
```
ItalyPaleAle, line 1624: Same here, please use a switch statement and conside rusing `errors.Is`

pkg/api/http/http.go
```diff
@@ -1794,30 +1789,40 @@ func (a *api) onBulkPublish(reqCtx *fasthttp.RequestCtx) {
...
 		}
-		status := nethttp.StatusInternalServerError
 		bulkRes.ErrorCode = "ERR_PUBSUB_PUBLISH_MESSAGE"
 
-		if errors.As(err, &runtimePubsub.NotAllowedError{}) {
-			msg := NewErrorResponse("ERR_PUBSUB_FORBIDDEN", err.Error())
-			status = nethttp.StatusForbidden
-			fasthttpRespond(reqCtx, fasthttpResponseWithError(status, msg), closeChildSpans)
-			log.Debug(msg)
+		if !errors.As(err, &runtimePubsub.NotAllowedError{}) && !errors.As(err, &runtimePubsub.NotFoundError{}) {
```
ItalyPaleAle, line 1794: Here too, please use a switch statement which makes this code a lot more readable (there are multiple `ifs` and `returns` and it's hard to follow)

pkg/api/http/http.go
```diff
@@ -1827,30 +1832,35 @@ func (a *api) onBulkPublish(reqCtx *fasthttp.RequestCtx) {
...
 	topic := reqCtx.UserValue(wildcardParam).(string)
 	if topic == "" {
-		msg := NewErrorResponse("ERR_TOPIC_EMPTY", fmt.Sprintf(messages.ErrTopicEmpty, pubsubName))
-		return nil, "", "", nethttp.StatusNotFound, &msg
+		err = apierrors.PubSubTopicEmpty(pubsubName, pubsubType, metadata)
+		return nil, "", "", err
 	}
-	return thepubsub.Component, pubsubName, topic, nethttp.StatusOK, nil
+
+	return thepubsub.Component, pubsubName, topic, err
```
ItalyPaleAle, line 1863:
```suggestion
	return thepubsub.Component, pubsubName, topic, nil
```

pkg/api/universal/state_query.go
```diff
@@ -16,8 +16,14 @@ package universal
 import (
 	"context"
 	"encoding/json"
+	httpCodes "net/http"
```
ItalyPaleAle, line 19: Please don't rename `http` to `httpCodes` as this package includes more than the codes, and if someone needs something else from `net/http` in the future, this will be confusing

pkg/api/universal/state_query.go
```diff
@@ -16,8 +16,14 @@ package universal
...
 	"time"
 
+	"google.golang.org/grpc/codes"
+
+	contribMetadata "github.com/dapr/components-contrib/metadata"
+	kiterrors "github.com/dapr/kit/errors"
```
ItalyPaleAle, line 25: Import blocks here too

pkg/api/grpc/grpc.go
cicoyle: I didn't author that code. I can update it as suggested to use `errors.Is()`

pkg/api/http/http.go
cicoyle: if err is nil, it will return nil. I can still change it if desired

pkg/api/universal/state_query.go
cicoyle: there were 2 http imports from what I saw earlier, but that may have just been a local issue. I'll switch this back 👍🏻

pkg/api/grpc/grpc.go
cicoyle: if err is nil, it will return nil here anyways.

pkg/api/grpc/grpc.go
cicoyle: is there some config I need to set in my IDE or go for this? I ran the `gofumpt` && `goimport` and this is what I got

pkg/api/grpc/grpc.go
cicoyle: done

pkg/api/grpc/grpc.go
cicoyle: done

pkg/api/grpc/grpc.go
cicoyle: Actually, switching from errors.As -> errors.Is was causing test failures. I'm switching back and keeping it as `errors.As`

pkg/api/http/http.go
```diff
@@ -1794,30 +1788,35 @@ func (a *api) onBulkPublish(reqCtx *fasthttp.RequestCtx) {
...
 			return
+		default:
+			err = apierrors.PubSubPublishMessage(pubsubName, string(contribMetadata.PubSubType), topic, err)
+			log.Debug(err)
 		}
 
 		// Return the error along with the list of failed entries.
 		resData, _ := json.Marshal(bulkRes)
-		fasthttpRespond(reqCtx, fasthttpResponseWithJSON(status, resData, nil), closeChildSpans)
+		standardizedErr, ok := kiterrors.FromError(err)
```
elena-kolevska, line 1816: Just a nit; maybe this can be more concise as:

```
if standardizedErr, ok := kiterrors.FromError(err); ok {
    fasthttpRespond(reqCtx, fasthttpResponseWithJSON(standardizedErr.HTTPStatusCode(), resData, map[string]string{"responseData": string(resData), "error": standardizedErr.Error()}), closeChildSpans)
}
```

elena-kolevska (approved):

pkg/api/http/http.go
```diff
@@ -696,15 +696,15 @@ func (a *api) getStateStoreWithRequestValidation(reqCtx *fasthttp.RequestCtx) (s
 	storeName := a.getStateStoreName(reqCtx)
 
 	if a.universal.CompStore().StateStoresLen() == 0 {
-		err := apierrors.NotConfigured(storeName, string(contribMetadata.StateStoreType)+" store", map[string]string{"appID": a.universal.AppID()}, codes.FailedPrecondition, nethttp.StatusInternalServerError, "ERR_STATE_STORE_NOT_CONFIGURED", kiterrors.CodePrefixStateStore+kiterrors.CodeNotConfigured)
+		err := apierrors.NotConfigured(storeName, string(contribMetadata.StateStoreType)+" store", nil, codes.FailedPrecondition, nethttp.StatusInternalServerError, "ERR_STATE_STORE_NOT_CONFIGURED", kiterrors.CodePrefixStateStore+kiterrors.CodeNotConfigured)
```
yaron2, line 699: Why is the metadata removed here?

cicoyle: It didn't exist previously until my PR, but when I added it in it was failing a bunch of integration tests, so I removed it to minimize the test failures. Its back now. Still working on the test failures tho.

elena-kolevska:
@cicoyle you can "fix" the failing windows tests by adding this check to your tests:

```
if runtime.GOOS == "windows" {
		t.Skip("skipping unix socket based test on windows")
	}
```

Check out Josh's PR for details: https://github.com/dapr/dapr/pull/7377/files#diff-63a8642d7d0d77c4c1c98a0b9a603522277d8bef886479027e0d7030d9cdfa47R57

artursouza (approved):

artursouza:
Does any of the current reviewers have any concern before I merge this?

yaron2 (approved):

cicoyle:
> @cicoyle you can "fix" the failing windows tests by adding this check to your tests:
> 
> ```
> if runtime.GOOS == "windows" {
> 		t.Skip("skipping unix socket based test on windows")
> 	}
> ```
> 
> Check out Josh's PR for details: https://github.com/dapr/dapr/pull/7377/files#diff-63a8642d7d0d77c4c1c98a0b9a603522277d8bef886479027e0d7030d9cdfa47R57

If you look at the commit right before your comment, that's where I added the 'fix'. Its the actor health that is failing for me on Windows. I assume that is somewhat flaky bc I've seen it before.

PR #7324, opened 2023-12-20
Fix handling errors in HTTP service invocation

ItalyPaleAle:
1. Make sure error is always logged
2. If data has already been sent to users, log the error only and do not append it to the response, which could cause data corruption for the caller

### DIFF
--- a/pkg/http/api_directmessaging.go
+++ b/pkg/http/api_directmessaging.go
@@ -240,8 +240,22 @@ func (a *api) onDirectMessage(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	// Handle errors; successful operations are already complete
+	// If success is true, it means that headers have already been sent, so we can't send the error to the user, because:
+	// headers cannot be re-sent, and adding to response body may cause corrupted data to be sent
+	if success.Load() {
+		// Use Warn log here because it's the only way users are notified of the error
+		log.Warnf("HTTP service invocation failed to complete with error: %v", err)
+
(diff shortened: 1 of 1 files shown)


pkg/http/api_directmessaging.go
```diff
@@ -240,8 +240,22 @@ func (a *api) onDirectMessage(w http.ResponseWriter, r *http.Request) {
...
+	if success.Load() {
+		// Use Warn log here because it's the only way users are notified of the error
+		log.Warnf("HTTP service invocation failed to complete with error: %v", err)
+
+		// Do nothing else, as at least some data was already sent to the client
+		return
+	}
+
+	// Log with debug level and send the error to the client in the body
+	log.Debugf("HTTP service invocation failed to complete with error: %v", err)
```
yaron2, line 255: We need to be able to differentiate between calls that failed due to errors coming from the target sidecar as opposed to errors coming from the target app

ItalyPaleAle: Looks like this is already handled. If error is of type `codeError` (invalid status code) or `invokeError` (other errors from the sidecar), the logs are formatted according to:

https://github.com/ItalyPaleAle/dapr/blob/00cd74047155233276ca66d18f3d8506be47a2bd/pkg/http/api_directmessaging.go#L382-L400

yaron2: Validated it 👍

yaron2 (approved):

PR #7327, opened 2023-12-20
[release-1.12] Fix handling errors in HTTP service invocation

ItalyPaleAle:
Cherry-pick #7324 into release-1.12

> I don't think we need release notes as it could be considered a fixup from the other PR

1. Make sure error is always logged
2. If data has already been sent to users, log the error only and do not append it to the response, which could cause data corruption for the caller

### DIFF
--- a/pkg/http/api_directmessaging.go
+++ b/pkg/http/api_directmessaging.go
@@ -194,8 +194,22 @@ func (a *api) onDirectMessage(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	// Handle errors; successful operations are already complete
+	// If success is true, it means that headers have already been sent, so we can't send the error to the user, because:
+	// headers cannot be re-sent, and adding to response body may cause corrupted data to be sent
+	if success.Load() {
+		// Use Warn log here because it's the only way users are notified of the error
+		log.Warnf("HTTP service invocation failed to complete with error: %v", err)
+
+		// Do nothing else, as at least some data was already sent to the client
+		return
+	}
+
+	// Log with debug level and send the error to the client in the body
+	log.Debugf("HTTP service invocation failed to complete with error: %v", err)
+
 	if resp != nil {
-		defer resp.Close()
+		resp.Close()
 
 		// Set headers if present (if resp is not nil, they haven't been sent already)
 		headers := resp.Headers()
@@ -204,7 +218,6 @@ func (a *api) onDirectMessage(w http.ResponseWriter, r *http.Request) {
 		}
 	}
 
-	// Handle errors; successful operations are already complete
 	var (
 		codeErr   codeError
 		invokeErr invokeError


JoshVanL:
@ItalyPaleAle please can we include tests so we don't have a regression/covers the bug introduced in the previous PR

ItalyPaleAle:
> @ItalyPaleAle please can we include tests so we don't have a regression/covers the bug introduced in the previous PR

Not a regression introduced by the other PR, just something else that was left out: adding some logs and making sure errors aren't appended to the body (something observed with 1.12.0 too).

Not sure we can create a test for this as it requires creating complex situations such as gRPC connections failing while a request is in progress. I've been working with @yaron2 and 2 customers this morning and found that this change helped identifying a bug with the gRPC stream between sidecars. We are still investigating the root cause of that bug.

JoshVanL:
> > @ItalyPaleAle please can we include tests so we don't have a regression/covers the bug introduced in the previous PR
> 
> Not a regression introduced by the other PR, just something else that was left out: adding some logs and making sure errors aren't appended to the body (something observed with 1.12.0 too).
> 
> Not sure we can create a test for this as it requires creating complex situations such as gRPC connections failing while a request is in progress. I've been working with @yaron2 and 2 customers this morning and found that this change helped identifying a bug with the gRPC stream between sidecars. We are still investigating the root cause of that bug.

In the case of code being untestable we should refactor the code to enable introducing mid request errors via a middleware, though it looks like this should be possible with the code today AFAICT.

pkg/http/api_directmessaging.go
```diff
@@ -194,8 +194,22 @@ func (a *api) onDirectMessage(w http.ResponseWriter, r *http.Request) {
...
 	}
 
+	// Handle errors; successful operations are already complete
+	// If success is true, it means that headers have already been sent, so we can't send the error to the user, because:
+	// headers cannot be re-sent, and adding to response body may cause corrupted data to be sent
+	if success.Load() {
+		// Use Warn log here because it's the only way users are notified of the error
+		log.Warnf("HTTP service invocation failed to complete with error: %v", err)
+
+		// Do nothing else, as at least some data was already sent to the client
```
artursouza, line 204: Does it mean the user will get a successful response with incomplete data?

ItalyPaleAle: No, if the response was successful, the data is already sent to the client within the policy function.

If we are here, it means that the request failed. There are 2 possible options here:

1. The request failed without ever sending any data to the user. In this case, down below we send a response to the client with a HTTP status code indicating an error, and the error details in the body
2. The request failed but at least a partial response has already been sent to the client. This is what we've observed when working on #7270 (which was a bug in Dapr now fixed) and now in bugs reported by 2 separate users (which seem to be due to the gRPC connection between sidecars failing - we are investigating that separately). In this case, we were here trying to re-set the status code (pointlessly since the headers have already been flushed), and more importantly we were appending the error message to the response body, which meant we were sending a "corrupted" response to the caller.

The main change here is recognizing that if some data has already been sent to the client, we should abort the request without sending _more_ data.

At the same time, we were never logging errors. Now errors are logged with debug level if they are also sent to the client in the body; if not, they are logged as warnings.

yaron2 (approved):

PR #7328, opened 2023-12-20
Fix: enable explicit keep-alives in the sidecar-to-sidecar gRPC connection

ItalyPaleAle:
Sets explicit keep-alives in the gRPC server and client, to make sure that they are enabled.

Also sets some other keep-alive parameters more explicitly and documents some choices.

> Updates the gRPC library as there are some fixes included that are related to keep-alives


yaron2:
Note for maintainers we're awaiting user feedback on this.

ItalyPaleAle:
Closing this, it doesn't seem to be the right thing to do. Switching to #7329 instead

PR #7329, opened 2023-12-21
Switch gRPC sidecar-to-sidecar to create one connection per each pod

ItalyPaleAle:
> This a second attempt at fixing the issue started in #7328

Dependencies:

- https://github.com/dapr/kit/pull/80
- https://github.com/dapr/components-contrib/pull/3284

Users are reporting errors in service invocation that indicate that the gRPC connection between sidecar has failed. Logs (with #7327 merged) include something like:

> {"app_id":"app-id","instance":"...","level":"warning","msg":"HTTP service invocation failed to complete with error: readfrom tcp 127.0.0.1:3500-\u003e127.0.0.1:47102: error receiving message: rpc error: code = Unavailable desc = closing transport due to: connection error: desc = \"error reading from server: unexpected EOF\", received prior goaway: code: NO_ERROR, debug data: \"max_age\"","scope":"dapr.runtime.http","time":"2023-12-20T23:11:38.819884017Z","type":"log","ver":"edge"}

Upon further investigation over the last 2 days together with @yaron2 , we identified the issue to be related to load balancing of gRPC connections, which was interacting poorly with our use of MaxAge. You can read more details here: https://kubernetes.io/blog/2018/11/07/grpc-load-balancing-on-kubernetes-without-tears/

This PR attempts to fix the issue by following the recommendation in the blog post above, specifically about load balancing pools.

- Dapr already had a load balancing pool, which was implemented a long time ago. However, in our pool we were putting connections created against a DNS name.
- In environments like K8s, a DNS name can resolve with multiple A/AAAA records, so gRPC automatically creates connections to each of them.
- We also set a MaxAge to allow the connection to be re-created after a delay. This was necessary because if the list of pods that resolve with a given DNS name change (e.g. when the app is scaled horizontally), gRPC doesn't automatically refresh them. It is required to close and re-create the connection.
- The MaxAge was causing servers to send GOAWAY requests to clients, causing the bug shown above

The way this PR attempts to solve the issue is to:

- Add a new, optional method to nameresolvers `ResolveIDMulti`, which returns a list of addresses.
   - In the K8s nameresolver, this performs actual DNS resolution and returns a list of IPs
- If the nameresolver implements that method, when creating a gRPC connection to the target sidecar we pick a random address from the list of addresses returned
   - To avoid performing DNS resolution on every invocation, the result is cached. The value in the cache has a TTL of 30s and is automatically deleted after certain failures (such as connection errors).
   - This works because when Dapr creates Service resources, we make them headless, so we can get the list of A/AAAA records via a normal DNS lookup.
   - Note: all other nameresolvers already return a single IP in `ResolveID` (such as Consul, mDNS, SQLite). They already pick a random one from a pool.  By returning a different IP every time, they already cause gRPC to create individual connections, just like this PR does for K8s. Those don't need to be updated to use `ResolveIDMulti`, unless we want to use the cache (which may be useful for performance).
- This way, we create a connection to one specific sidecar, and not a load-balanced IP.
- Remove MaxAge from the gRPC server, since it's not needed anymore, and we can get load balancing (with automatic refresh of the addresses) without forcefully destroying connections. There's still a timeout on idle connections, both on the server and client.

Add a new E2E test to validate that if an app is deployed with > 1 replica, requests are load balanced across the 2 instances.


go.mod
```diff
@@ -11,8 +11,8 @@ require (
 	github.com/argoproj/argo-rollouts v1.4.1
 	github.com/cenkalti/backoff/v4 v4.2.1
 	github.com/cloudevents/sdk-go/v2 v2.14.0
-	github.com/dapr/components-contrib v1.12.1-0.20231219201927-02c6b21ec621
-	github.com/dapr/kit v0.12.2-0.20231219171931-d7281f0c24d0
```
ItalyPaleAle, line 15: Note that these depend on:

- https://github.com/dapr/kit/pull/80
- https://github.com/dapr/components-contrib/pull/3284

yaron2:
@ItalyPaleAle Please resolve conflicts

ItalyPaleAle:
> @ItalyPaleAle Please resolve conflicts

Done

yaron2 (approved):

PR #7333, opened 2023-12-22
[release-1.12] Switch gRPC sidecar-to-sidecar to create one connection per each pod

ItalyPaleAle:
Depends on:

- https://github.com/dapr/kit/pull/80
- https://github.com/dapr/components-contrib/pull/3287

Backport of #7329 into release-1.12 - see that PR for details


pkg/messaging/direct_messaging.go
```diff
@@ -548,37 +576,62 @@ func (d *directMessaging) addForwardedHeadersToMetadata(req *invokev1.InvokeMeth
...
+			ID:        res.id,
+			Namespace: res.namespace,
+			Port:      d.grpcPort,
+		}
+
+		// If the component implements ResolverMulti, we can use caching
+		if d.resolverMulti != nil {
+			// Check if the value is in the cache
+			res.cacheKey = request.CacheKey()
+			addresses, _ := d.resolverCache.Get(res.cacheKey)
```
yaron2, line 608: How is the cache updated with IP addresses of new pods that come up from a scale up event?

JoshVanL: Similarly, how do we handle the case when hosts disappear or the IP is reserved by a completely different app ID during this 30 second window.

ItalyPaleAle: > How is the cache updated with IP addresses of new pods that come up from a scale up event?

It isn't, for as long as the IPs are in the cache (30s). The idea is that if an app scales horizontally, and a caller was invoking apps with that ID, it will take an average of 15s (best case 0s, worst case 30s) before messages are routed to that app too, which is considered acceptable.
We can tweak the duration of the cache if you think. Maybe 30s is too much, and 10s or just 5s is better. Of course, the shorter the cache, the more namresolutions will have to happen, with a negative (but supposedly small) perf impact.

> Similarly, how do we handle the case when hosts disappear or the IP is reserved by a completely different app ID during this 30 second window.

Good question, let me investigate that. I don't know if the target app already has a way to reject calls for the wrong app ID.

JoshVanL: > Good question, let me investigate that. I don't know if the target app already has a way to reject calls for the wrong app ID.

My assumption is that it will fail mTLS and error out. This is actually a security vulnerability whereby a malicious actor can assume an IP address and then take incoming connections meant for a different host.

Given that the cache is not related to the error we are trying to fix here, and there hasn't been any issues (as far as I can see (?)) related to performance hits of DNS querying in service invocation, I think we should remove the cache. The nature of cloud native environments where IP address are completely ephemeral means a cache of the DNS state of the cluster can be assumed to be wrong all of the time and I see a lot of errors being reported because of it.

ItalyPaleAle: Yes, if mTLS is enabled that should fail. Emphasis on _if_ it's enabled.

I could stand behind removing the cache if we have concerns about stale data.

Today the Kubernetes nameresolver is the only one that doesn't perform an actual DNS resolution in the component. Every other NR does do a full lookup, whether that's Consul, mDNS, or SQLite. The lookup in the K8s component is performed by making a regular DNS resolution call, which _should_ be very quick.

yaron2: A perf test can easily point if the lookup is acceptable. I support removing the cache if perf is ok

ItalyPaleAle: I have pushed a commit that disables the cache (but can be re-added quickly). Will run perf tests now

ItalyPaleAle: Looks like removing the cache did have a significant negative impact on the perf tests

- Dapr 1.12: https://github.com/dapr/dapr/actions/runs/7453524347
- This PR without caching: https://github.com/dapr/dapr/actions/runs/7453244709

yaron2: Well, then the cache stays. I'm not worried about this idea architecturally, Envoy and other proxies are doing a similar thing for gRPC load balancing, but we should make sure our error handling is as efficient as possible, maybe have the ability to remove endpoints from the cache if we get an unreachable error and then retry the request until it succeeds on at least one endpoint and if no addresses exist we return

ItalyPaleAle: > Well, then the cache stays.

I'll re-add the cache, then see what the perf impact is.

> maybe have the ability to remove endpoints from the cache if we get an unreachable error and then retry the request until it succeeds on at least one endpoint and if no addresses exist we return

Yes, this was already implemented: https://github.dev/ItalyPaleAle/dapr/blob/a805b2b466ad03ba784a3b17c2627aa835641469/pkg/messaging/direct_messaging.go#L230-L233

Although I should probably add that here too for network errors: https://github.dev/ItalyPaleAle/dapr/blob/a805b2b466ad03ba784a3b17c2627aa835641469/pkg/messaging/direct_messaging.go#L223-L225

yaron2: Perf is very bad even with the cache re-introduced, but it shouldn't be as once an address is in the cache it should behave exactly like it does today, so I think there's a low hanging fruit here that'll fix this

ItalyPaleAle: @yaron2 i got it to work. I had made a mistake so the cache wasn't actually working. I have now fixed it so the cache is actually used. Perf tests are passing and doesn't look like there's a significant difference

pkg/messaging/direct_messaging.go
```diff
@@ -246,7 +241,13 @@ func (d *directMessaging) invokeWithRetry(
 				}
 				return rResp, fmt.Errorf("failed to invoke target %s after %d retries. Error: %w", app.id, attempt-1, rErr)
 			}
+
 			teardown(false)
+			// Remove the resolved name from the cache
+			if app.cacheKey != "" && d.resolverCache != nil {
+				d.resolverCache.Delete(app.cacheKey)
```
yaron2, line 248: Should an address be removed only if the error is for an unavailable endpoint?

yaron2: According to the gRPC docs it can also mean a transient network error

ItalyPaleAle: I've removed it

pkg/messaging/direct_messaging.go
```diff
@@ -208,9 +234,14 @@ func (d *directMessaging) invokeWithRetry(
 			code := status.Code(rErr)
 			if code == codes.Unavailable || code == codes.Unauthenticated {
 				// Destroy the connection and force a re-connection on the next attempt
+				// We also remove the resolved name from the cache
 				teardown(true)
+				if app.cacheKey != "" && d.resolverCache != nil {
+					d.resolverCache.Delete(app.cacheKey)
```
yaron2, line 240: Deletion now occurs when unavailable which is correct, but also applies to unauthenticated. I believe we should only remove if unavailable. For example, an unauthenticated message will happen during a certificate rollover of the target sidecar. In this case, the endpoint is there but the calls should be retried until it succeeds (when the target sidecar's server has a valid cert)

ItalyPaleAle: I think deletion from cache when un-authenticated is fine. This will cover the case of an app that is replacing a different app at the same IP.

Cert rollovers _shouldn't_ cause errors, because we rotate certs before their validity ends. As long as they are signed by the same CA, callers should continue to accept them with 0 downtime.

I would also err on the side of caution personally and over-purging from the cache, than potentially keeping stale values.

yaron2 (approved):

Issue #6666, opened 2023-07-14
Sidecar tries to delete reminder even though app has unregistered it


sujitdmello:
## In what area(s)?

> /area runtime

## What version of Dapr?

> 1.11.1

## Expected Behavior

Java app responding to a **receiveReminder** callback, calls **super.unregisterReminder().block()** which should delete the reminder and not result in any errors. 

## Actual Behavior

The receiveReminder method calls **super.unregisterReminder().block()** and receives no errors, and the reminder finishes processing. Then the sidecar logs an error *Could not find active reminder with key*.

```log
== APP == 2023-07-14 08:13:57.039  INFO 13768 --- [nio-9002-exec-4] com.service.CheckoutActorImpl            : Reminder received for long retry # 1
time="2023-07-14T08:13:57.0567597-04:00" level=debug msg="Found reminder with key: CheckoutActor||1000||constantRetryReminder1. Deleting reminder" app_id=checkout-actor instance=DESKTOP-70K7KP5 scope=dapr.runtime.actor type=log ver=1.11.1
time="2023-07-14T08:13:57.1813354-04:00" level=debug msg="starting to read reminders for actor type CheckoutActor (migrate=false), with metadata id 00000000-0000-0000-0000-000000000000 and 0 partitions" app_id=checkout-actor instance=DESKTOP-70K7KP5 scope=dapr.runtime.actor type=log ver=1.11.1
time="2023-07-14T08:13:57.1976031-04:00" level=debug msg="read reminders from actors||CheckoutActor without partition" app_id=checkout-actor instance=DESKTOP-70K7KP5 scope=dapr.runtime.actor type=log ver=1.11.1
time="2023-07-14T08:13:57.2031242-04:00" level=debug msg="finished reading reminders for actor type CheckoutActor (migrate=false), with metadata id 00000000-0000-0000-0000-000000000000 and no partitions: total of 1 reminders" app_id=checkout-actor instance=DESKTOP-70K7KP5 scope=dapr.runtime.actor type=log ver=1.11.1
== APP == 2023-07-14 08:13:57.321  INFO 13768 --- [nio-9002-exec-4] com.service.CheckoutActorImpl            : Reminder constantRetryReminder1 unregistered.
== APP == 2023-07-14 08:13:57.322  INFO 13768 --- [nio-9002-exec-4] com.service.CheckoutActorImpl            : Reading actor state for actor Id 1000
[...]

PR #7337, opened 2024-01-02
Allow canceling recurring reminders after execution for non-internal actors

ItalyPaleAle:
> Related issue: #6666

Internal actors have the ability to cancel a recurring reminder by responding to an execution with the special error [`ErrReminderCanceled`](https://github.com/dapr/dapr/blob/9e2bcdc3d8a64b02466a9274e9259f166eedc95a/pkg/actors/actors.go#L894-L898) after a reminder is executed.

However, the capability was never exposed to "external" actors.

This PR adds the ability for an actor to stop recurring reminders (and timers) after they are executed also for "external" actors. This is possible by including the header `X-DaprReminderCancel` with a truthy value in the response when a reminder or timer is executed.

This will require support in SDKs, but it allows something that is particularly helpful (and today can be buggy, as in #6666)

Integration test is included


mukundansundar:
> > Related issue: #6666
> 
> Internal actors have the ability to cancel a recurring reminder by responding to an execution with the special error [`ErrReminderCanceled`](https://github.com/dapr/dapr/blob/9e2bcdc3d8a64b02466a9274e9259f166eedc95a/pkg/actors/actors.go#L894-L898) after a reminder is executed.
> 
> However, the capability was never exposed to "external" actors.
> 
> This PR adds the ability for an actor to stop recurring reminders (and timers) after they are executed also for "external" actors. This is possible by including the header `X-DaprReminderCancel` with a truthy value in the response when a reminder or timer is executed.
> 
> This will require support in SDKs, but it allows something that is particularly helpful (and today can be buggy, as in #6666)
> 
> Integration test is included

@ItalyPaleAle So in this scenario, as soon as a reminder is deregistered/unregistered, then the SDK/app sends the the specific header, that returns the response `ErrReminderCancelled`. But won't the reminder still be there in the DB to be executed again? 

Or is this a specific scenario where the app deregisters the reminder first and then sets the header? 

Finally why is an error returned in `callLocalActor`? Wouldn't that trigger the resiliency retry if present which we do not want for deregistered reminders?

ItalyPaleAle:
> > > Related issue: #6666
> > 
> > 
> > Internal actors have the ability to cancel a recurring reminder by responding to an execution with the special error [`ErrReminderCanceled`](https://github.com/dapr/dapr/blob/9e2bcdc3d8a64b02466a9274e9259f166eedc95a/pkg/actors/actors.go#L894-L898) after a reminder is executed.
> > However, the capability was never exposed to "external" actors.
> > This PR adds the ability for an actor to stop recurring reminders (and timers) after they are executed also for "external" actors. This is possible by including the header `X-DaprReminderCancel` with a truthy value in the response when a reminder or timer is executed.
> > This will require support in SDKs, but it allows something that is particularly helpful (and today can be buggy, as in #6666)
> > Integration test is included
> 
> @ItalyPaleAle So in this scenario, as soon as a reminder is deregistered/unregistered, then the SDK/app sends the the specific header, that returns the response `ErrReminderCancelled`. But won't the reminder still be there in the DB to be executed again?
> 
> Or is this a specific scenario where the app deregisters the reminder first and then sets the header?
> 
> Finally why is an error returned in `callLocalActor`? Wouldn't that trigger the resiliency retry if present which we do not want for deregistered reminders?

It's a lot simpler than that :) The feature was already implemented and used by internal actors (and it was wired via `callLocalActor` because reminder/timer invocation still goes through the regular actor invocation paths). It allows an app to tell Dapr to stop the execution of recurring reminders (so the reminder is also removed from the DB).

A reminder is executed by Dapr sending a HTTP request to the app. The app responds and includes a status code.

What's been added now (bringing it on-par with internal actors) is that at the same time as the app responds, it can tell Dapr to stop a recurring reminder.

This can be useful in scenarios such as what #6666 tries to do, but can also be useful to implement reliable reminders which offer at-least-once delivery guarantee

msfussell:
@ItalyPaleAle - I presume this needs a docs issue/PR, yes? If so, please create a docs PR before merging this issue.

mukundansundar:
@ItalyPaleAle can you fix the conflicts in this PR please?

ItalyPaleAle:
Done

yaron2 (approved):

marcduiker:
@holopin-bot @ItalyPaleAle

PR #7340, opened 2024-01-02
Cascade terminate/Purge Workflow Support

shivamkm07:
# Description

Support for cascade terminate/purge was added in DTF-Go: https://github.com/microsoft/durabletask-go/pull/47.

This PR adds the support for the same in Dapr. Also it makes the cascade feature default i.e. when no option is provided, it would assume cascade terminate/purge `true` The related components-contrib PR is https://github.com/dapr/components-contrib/pull/3296
## Issue reference

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing


ItalyPaleAle:
LGTM but all builds are failing

pkg/proto/common/v1/common.pb.go
```diff
@@ -19,9 +19,9 @@
 package common
 
 import (
+	any1 "github.com/golang/protobuf/ptypes/any"
```
ItalyPaleAle, line 22: These changes look odd. Please make sure you're using the same version of the  protobuf compiler we use: https://github.com/dapr/dapr/tree/master/dapr#proto-client-generation

shivamkm07: Fixed.

DeepanshuA (approved):

pkg/http/api_test.go
```diff
@@ -2817,6 +2817,23 @@ func TestV1Beta1Workflow(t *testing.T) {
...
+			Recursive bool `json:"recursive"`
+		}{
+			Recursive: true,
+		}
+		body, _ := json.Marshal(data)
+		resp := fakeServer.DoRequest("POST", apiPath, body, nil)
+		assert.Equal(t, 500, resp.StatusCode)
+		assert.Equal(t, "ERR_TERMINATE_WORKFLOW", resp.ErrorBody["errorCode"])
+		assert.Equal(t, fmt.Sprintf(messages.ErrTerminateWorkflow.Message(), "instanceID", daprt.ErrFakeWorkflowTerminateRecursiveError), resp.ErrorBody["message"])
+	})
```
mukundansundar, line 2835: can we have this as a query string parameter instead?

mukundansundar (changes requested):
please add IT/E2E for this parameter

pkg/runtime/wfengine/workflow.go
```diff
@@ -597,12 +597,21 @@ func (wf *workflowActor) runWorkflow(ctx context.Context, actorID string, remind
 			if errMarshal != nil {
 				return errMarshal
 			}
-
+			requestBytes := eventData
```
shivamkm07, line 600: @kaibocai This was a bug that got left out of #7308 and was causing errors in creating sub-orchestrations. Please take a look if this is the correct behavior while sending the policy to sub-orchestrations

pkg/http/api_test.go
shivamkm07: Done

shivamkm07:
> please add IT/E2E for this parameter

Added ITs

pkg/api/http/workflow.go
```diff
@@ -238,7 +238,19 @@ func (a *api) onTerminateWorkflowHandler() http.HandlerFunc {
...
-			InModifier:        workflowInModifier[*runtimev1pb.TerminateWorkflowRequest],
+			InModifier: func(r *http.Request, in *runtimev1pb.TerminateWorkflowRequest) (*runtimev1pb.TerminateWorkflowRequest, error) {
+				in.SetWorkflowComponent(chi.URLParam(r, workflowComponent))
+				in.SetInstanceId(chi.URLParam(r, instanceID))
+
+				// Default to true if not specified
+				in.Recursive = true
+				// Extract recursive option from query string
+				recursive := r.URL.Query().Get("recursive")
+				if recursive == "false" {
```
mukundansundar, line 249: why is the default `true`?

shivamkm07: That seems to be the intended behavior: https://github.com/microsoft/durabletask-go/pull/47#discussion_r1424793527
right @cgillum ?

mukundansundar: @shivamkm07 If the intended behaviour is to have a recursive terminate/purge, lets break the API ... But have the query string as `non-recursive`. It's is not the right behaviour to have a query string that is implicitly set to true.
cc @cgillum

pkg/runtime/wfengine/workflow.go
shivamkm07: Creating a separate PR for this and removing it from here as it's P0 blocking other works.

shivamkm07:
Integration test will fail in this PR for now. Will be resolved once #7385 is merged.

shivamkm07:
> Integration test will fail in this PR for now. Will be resolved once #7385 is merged.

Should pass now

mukundansundar (approved):
lgtm.
@cgillum Do we need the API to have cascade terminate by default?

pkg/api/http/http.go
```diff
@@ -99,6 +99,7 @@ const (
 	workflowComponent        = "workflowComponent"
 	workflowName             = "workflowName"
 	instanceID               = "instanceID"
+	nonRecursive             = "non_recursive"
```
cgillum, line 102: Can we make this `recursive` instead of `non_recursive`? It's harder for people to understand negatives.

pkg/runtime/wfengine/backends/actors/backend.go
```diff
@@ -343,8 +343,13 @@ func (abe *ActorBackend) GetActivityWorkItem(ctx context.Context) (*backend.Acti
 }
 
 // GetOrchestrationRuntimeState implements backend.Backend
-func (*ActorBackend) GetOrchestrationRuntimeState(context.Context, *backend.OrchestrationWorkItem) (*backend.OrchestrationRuntimeState, error) {
-	return nil, errors.New("not supported")
+func (abe *ActorBackend) GetOrchestrationRuntimeState(ctx context.Context, owi *backend.OrchestrationWorkItem) (*backend.OrchestrationRuntimeState, error) {
+	state, err := LoadWorkflowState(ctx, abe.actorRuntime, string(owi.InstanceID), abe.config)
```
cgillum, line 347: Any particular reason for implementing this?

cgillum (commented):
My only question about this PR is whether we can make the parameter name be `recursive` instead of `non_recursive`. I understand that we like to default Boolean parameter values to `false`, but I wonder if it would actually be better to name the query string parameter to `recursive` and default its value to `true`.

pkg/api/http/http.go
DeepanshuA: I kind of incline towards `recursive` set to `true` by default. Understanding this was quite easier for me.
Is there any reasoning behind keeping all flags to false by default, apart from general practice?

pkg/runtime/wfengine/backends/actors/backend.go
shivamkm07: @cgillum this is required for cascade purge: we need to get list of sub-orchestrations to purge: https://github.com/microsoft/durabletask-go/blob/b58abfd02e48667ae6d8622de1531b322178b8b9/backend/backend.go#L135

pkg/api/http/http.go
ItalyPaleAle: I personally agree with Mukundan that it's more intuitive when the default value of a boolean is false.

If the default value is "true", then you are introducing 3 states: set true, set false, null.

PR #7348, opened 2024-01-04
Chore: update artifact actions

ItalyPaleAle:
The `master` tag was pointing to a very old build that sometimes failed (the "master" branch is now "main"). More appropriately pointing to a version now. Also, changed to avoid compressing artifacts that are already compressed.

In the process, also updated other actions including setup-go, which comes with its own caching for Go deps so we don't need the custom cache tasks


artursouza (approved):

daixiang0 (approved):
LGTM

artursouza:
There are conflicts now.

ItalyPaleAle:
> There are conflicts now.

Fixed

daixiang0 (dismissed):
LGTM

.github/workflows/dapr.yml
```diff
@@ -400,27 +303,30 @@ jobs:
           make release-flavor
         shell: bash
       - name: upload artifacts
-        uses: actions/upload-artifact@master
+        uses: actions/upload-artifact@v4
+        # Avoid publishing duplicate Windows artifacts, which will cause an error
+        if: matrix.windows_version != '1809'
```
ItalyPaleAle, line 308: This is key to fixing the builds right now: https://github.com/dapr/dapr/actions/runs/7534402274/job/20516204786

(The binaries are the same on 1809 and LTSC2022)

yaron2 (approved):

PR #7358, opened 2024-01-10
Update k6 dependency to fix perf test build

yaron2:
Perf tests on master branch fail to build due to an error with a k6 dependency. This is an attempt to upgrade to a version that doesn't have the issue.


yaron2:
Ok this fixed the issue. (The perf test goes past the build stage)

artursouza (approved):

PR #7359, opened 2024-01-10
Component loading for Control Plane Service rework

DeepanshuA:
# Description

It's a rework of #7314. Allows to load component for Control Plane Services.

## Issue reference

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Unit tests passing


charts/dapr/values.yaml
```diff
@@ -57,8 +57,11 @@ global:
   #  placement:
   #  operator:
   #  injector:
+
```
ItalyPaleAle, line 60: Please revert changes to this file since they are just comments - will avoid merge conflicts

DeepanshuA:
Marking WIP - I had merged conflicts but there is some mess created.

ItalyPaleAle (approved):

pkg/operator/api/authz.go
```diff
@@ -24,12 +22,7 @@ import (
 
 // authzRequest ensures that the requesting identity resides in the same
 // namespace as that of the requested namespace.
-func (a *apiServer) authzRequest(ctx context.Context, namespace string) error {
-	spiffeID, ok, err := spiffe.FromGRPCContext(ctx)
-	if err != nil || !ok {
-		return status.New(codes.PermissionDenied, "failed to determine identity").Err()
-	}
```
mukundansundar, line 31: do not move this to utils. have a function here itself to get spiffe ID or use the provided `FromGRPCContext` function from the module directly in `api.go`.

mukundansundar (changes requested):
one small change.

pkg/operator/api/authz.go
DeepanshuA: Though I believe that GetSpiffeIdFromContext can be used at other places as well and better placed in utils. But, I have moved that to this file itself.

ItalyPaleAle: I agree utils may not be the best package, but I do like the idea of it being in _a_ shared package. What about some security package?

pkg/operator/api/api.go
```diff
@@ -224,7 +240,23 @@ func (a *apiServer) ListComponents(ctx context.Context, in *operatorv1pb.ListCom
...
+				scope := c.Scopes[s]
+				if utils.IsControlPlaneService(scope) {
+					controlPlaneComp = true
+					break
+				}
+			}
+		}
+		if (!controlPlaneServiceReq && controlPlaneComp) || (controlPlaneServiceReq && !controlPlaneComp) {
+			continue
+		}
```
JoshVanL, line 258: Why do we want it so that a Component which has a scope that defines any control plane service will not also be sent to a non control plane application, rather than continuing to filter based solely on the service identity?

JoshVanL (changes requested):

pkg/operator/api/api.go
```diff
@@ -449,7 +493,11 @@ func (a *apiServer) ListResiliency(ctx context.Context, in *operatorv1pb.ListRes
 // TODO: @joshvanl: Authorize pod name and namespace matches the SPIFFE ID of
 // the caller.
 func (a *apiServer) ComponentUpdate(in *operatorv1pb.ComponentUpdateRequest, srv operatorv1pb.Operator_ComponentUpdateServer) error { //nolint:nosnakecase
-	if err := a.authzRequest(srv.Context(), in.GetNamespace()); err != nil {
+	spiffeID, err := getSpiffeIDFromContext(srv.Context())
```
JoshVanL, line 496: Filtering must also be done here, else applications will receive the Components which they are not scoped to when they are updated..

JoshVanL (changes requested):
Please include integration tests which extend the existing authz tests.

https://github.com/dapr/dapr/blob/master/tests/integration/suite/operator/api/authz.go

pkg/operator/api/api.go
DeepanshuA: I have updated it to load only components, NOT meant for control-plane.
And, do not want to add ComponentUpdate feature to control plane components as of now.

pkg/operator/api/api.go
DeepanshuA: Yeah, that's the ask. A request is identified whether to be from a CP or non-CP.
And, components are identified whether meant for CP or non-CP.
Only matching components are sent across.

pkg/operator/api/authz.go
mukundansundar: A security package if needed works.

mukundansundar (dismissed):
lgtm

pkg/operator/api/api.go
JoshVanL: We gain no security posture by not extending this check to also `ComponentUpdate` as control plane/non-control plane components will be available to any client which calls it and the component has some informer event.

Can you please describe the purpose of having any distinction at all between control plane components and app IDs in the scope entity list? It is still not clear to me why we would not do a simple filter based on the entities inside the scope, rather than the (in my eyes) surprising behaviour of special casing known control plane services and blocking any app ID from having access to it?

pkg/operator/api/api.go
```diff
@@ -224,7 +240,23 @@ func (a *apiServer) ListComponents(ctx context.Context, in *operatorv1pb.ListCom
 		Components: [][]byte{},
 	}
 	for i := range components.Items {
+		// By default assume that it is not a component for control plane service.
+		controlPlaneComp := false
 		c := components.Items[i] // Make a copy since we will refer to this as a reference in this loop.
+		// If the component is in the same namespace as the operator AND has a scope defined with a dynamic control plane service name, then it is a control plane component.
```
JoshVanL, line 246:
```suggestion
		// If the component is in the same namespace as the operator AND has a scope defined with a control plane service name, then it is a control plane component.
```

pkg/operator/api/api.go
```diff
@@ -224,7 +240,23 @@ func (a *apiServer) ListComponents(ctx context.Context, in *operatorv1pb.ListCom
...
+			for s := range c.Scopes {
+				scope := c.Scopes[s]
+				if utils.IsControlPlaneService(scope) {
+					controlPlaneComp = true
+					break
+				}
+			}
+		}
+		if (!controlPlaneServiceReq && controlPlaneComp) || (controlPlaneServiceReq && !controlPlaneComp) {
+			continue
```
JoshVanL, line 257:
```suggestion
		// By default assume that it is not a component for control plane service.
		inScope := len(c.Scopes) == 0
		c := components.Items[i]
			for s := range c.Scopes {
			   if s == spiffeID.AppID() {
			     inScope = true
			     break
			   }
			}
		if !inScope {
			continue
```

tests/integration/suite/operator/api/listcomponents.go
```diff
@@ -88,6 +89,18 @@ func (l *listcomponents) Setup(t *testing.T) []framework.Option {
...
 
+	l.ctrlPlaneComp = &compapi.Component{
+		TypeMeta:   metav1.TypeMeta{Kind: "Component", APIVersion: "dapr.io/v1alpha1"},
+		ObjectMeta: metav1.ObjectMeta{Name: "ctrl-plane-comp", Namespace: "default"},
+		Spec: compapi.ComponentSpec{
+			Type:    "state.inmemory",
+			Version: "v1",
+		},
+		Scoped: common.Scoped{
+			Scopes: []string{"dapr-placement"},
```
JoshVanL, line 100: What happens when I include app IDs here?

pkg/operator/api/api_test.go
```diff
@@ -271,6 +272,58 @@ func TestComponentUpdate(t *testing.T) {
...
+			for key := range api.allConnUpdateChan {
+				api.allConnUpdateChan[key] <- &ComponentUpdateEvent{
+					Component: &c,
+				}
+				close(api.allConnUpdateChan[key])
+			}
+		}()
+
+		os.Setenv("NAMESPACE", "ns1")
+		defer os.Unsetenv("NAMESPACE")
```
JoshVanL, line 317:
```suggestion
		t.Setenv("NAMESPACE", "ns1")
```

pkg/operator/api/api_test.go
```diff
@@ -470,6 +523,85 @@ func TestListsNamespaced(t *testing.T) {
...
+				Scoped: commonapi.Scoped{
+					Scopes: []string{"dapr-placement"},
+				},
+			}).
+			Build()
+
+		api := NewAPIServer(Options{Client: client}).(*apiServer)
+
+		os.Setenv("NAMESPACE", "dapr-mockns")
+		defer os.Unsetenv("NAMESPACE")
```
JoshVanL, line 562:
```suggestion
		t.Setenv("NAMESPACE", "dapr-mockns")
```

pkg/operator/api/api_test.go
```diff
@@ -470,6 +523,85 @@ func TestListsNamespaced(t *testing.T) {
...
+					Namespace: "namespace-a",
+				},
+			}, &componentsapi.Component{
+				TypeMeta: typeMeta,
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "obj2",
+					Namespace: "dapr-mockns",
+				},
+				Scoped: commonapi.Scoped{
+					Scopes: []string{"dapr-placement"},
```
JoshVanL, line 554: What happens when I include non control plane IDs here?

pkg/operator/api/api.go
```diff
@@ -475,6 +523,14 @@ func (a *apiServer) ComponentUpdate(in *operatorv1pb.ComponentUpdateRequest, srv
...
 			return
 		}
+		if c.ObjectMeta.Namespace == security.CurrentNamespace() {
+			for s := range c.Scopes {
+				scope := c.Scopes[s]
+				if utils.IsControlPlaneService(scope) {
+					return
+				}
+			}
+		}
```
JoshVanL, line 533: Why are we doing this here? Control plane components must support component updates eventually to have feature parity across Dapr. It also is not appropriate to have logic gating this functionality on the operator server side, rather than on the client side.

JoshVanL (changes requested):

tests/integration/suite/operator/api/listcomponents.go
DeepanshuA: They will be ignored - Want to keep components for Apps and CP separate.

pkg/operator/api/api.go
DeepanshuA: But, here I feel it is better to not open up a component meant for Control Plane to other Apps as well.
Let's say, an app having the access to same database which a Control Plane Service uses to Control data plane, seems like a Security issue to me. I know that if the App wants, they can still decide to use another component yaml and provide only their scope but I believe similar can be done for many other security points. It is about reducing the attack surface.
@mukundansundar @ItalyPaleAle @yaron2 WDYT?

pkg/operator/api/api.go
DeepanshuA: I have updated to do ComponentUpdate for both CP and non-CP. I have kept the same logic as is currently for ListComponents.

pkg/operator/api/api.go
DeepanshuA: have updated ComponentUpdate to be same as ListComponents

pkg/operator/api/api_test.go
DeepanshuA: They will be ignored - Want to keep components for Apps and CP separate.

pkg/operator/api/api.go
DeepanshuA: Please note that component for CP also needs to be namespaced same as CP namespace.

PR #7366, opened 2024-01-11
Sentry JWKS validator: add support for passing root CA certs

ItalyPaleAle:
Allows fetching a JWKS from a HTTPS server that uses a custom CA certificate


JoshVanL:
Moving to v1.14 milestone as we are beyond v1.13 feature freeze and imminently going into code freeze.

ItalyPaleAle:
This is a P0 for Microsoft for 1.13. It's only 5 LOCs if you exclude tests.

JoshVanL:
@ItalyPaleAle this has not been discussed with the project more widely and cannot be pushed through without consensus. Please create an issue as to why we need an exception to be included and the bug this is fixing. New features will not be accepted at this stage.

ItalyPaleAle:
> @ItalyPaleAle this has not been discussed with the project more widely and cannot be pushed through without consensus. Please create an issue as to why we need an exception to be included and the bug this is fixing. New features will not be accepted at this stage.

@mukundansundar is aware and agreed.

Reason is pretty self-evident and it's related to security, we need to set the CA cert for when fetching a JWKS from a HTTPS endpoint. It's required for security otherwise the only option is HTTP, which doesn't allow authenticating the origin server, so the JWKS could be spoofed.

ItalyPaleAle:
Thank you @yaron2.

I agree that sooner would have been better. We didn't realize this was needed this till yesterday.

yaron2 (approved):

artursouza:
I agree to accept this change. I also recommend vendors to use internal fork to have their urgent needs met without the need to workaround the OSS release process.

artursouza (approved):

Issue #7354, opened 2024-01-09
Scheduling of workflows


shivamkm07:
## In what area(s)?

 /area runtime

> /area operator

> /area placement

> /area docs

> /area test-and-release

## Describe the feature
Dapr workflows should support scheduling of workflows at a later time.

As per implementation, it seems that users are expected to pass time as metadata while sending the request with startTime set as `dapr.workflow.start_time`: https://github.com/dapr/dapr/blob/328b4b33df71f073b2da0d96bb30940880179955/pkg/runtime/wfengine/component.go#L96

However, this is not respected and carried at all places. 
1. It is not being set as part of the http request: https://github.com/dapr/dapr/blob/328b4b33df71f073b2da0d96bb30940880179955/pkg/http/api_workflow.go#L195
2. DTF-Go(which dapr uses for orchestration management) doesn't forward the scheduled time to backend: https://github.com/microsoft/durabletask-go/issues/57
3. Once the above issue is resolved, required delay needs to be added while creating the workflow reminder to actually schedule the workflow. Currently it schedules immediately even if scheduledTime is passed while creating the workflow: https://github.com/dapr/dapr/blob/328b4b33df71f073b2da0d96bb30940880179955/pkg/runtime/wfengine/workflow.go#L258

## Release Note

RELEASE NOTE:

msfussell:
@shivamkm07 Before merging this PR, please create a Docs PR for the workflow changes

shivamkm07:
> @shivamkm07 Before merging this PR, please create a Docs PR for the workflow changes

This PR mostly won't be going in 1.13. For other PR that has been merged, tracking issue was created: https://github.com/dapr/docs/issues/3949. Will create the PR soon.

PR #7370, opened 2024-01-12
Workflow execution and scheduling latency

ASHIQUEMD:
# Description

We are adding metrics to measure Dapr workflow execution and scheduling latency. The calculation strategy for these metrics is defined [here](https://github.com/dapr/dapr/issues/7109#issuecomment-1888693347).

This PR should be merged after [dtf-go pr](https://github.com/microsoft/durabletask-go/pull/62) is merged.

## Issue reference

Please reference the issue this PR will close: #7109

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Unit tests passing


pkg/runtime/wfengine/backends/actors/workflow_actor.go
```diff
@@ -482,11 +493,14 @@ func (wf *workflowActor) runWorkflow(ctx context.Context, actorID string, remind
 		}
 		return newRecoverableError(fmt.Errorf("failed to schedule a workflow execution: %w", err))
 	}
-	// Record metrics for workflow execution
+
+	wf.recordWorkflowSchedulingLatency(ctx, isWorkflowExecutionStartedEvent, state, scheduledStartTimestamp, workflowName, executionStatus)
```
shivamkm07, line 497: `executionStatus` should not be used here. It is being set only when this method is returning. At this point, it would be `diag.StatusFailed` only.

pkg/runtime/wfengine/backends/actors/workflow_actor.go
```diff
@@ -635,11 +649,24 @@ func (wf *workflowActor) runWorkflow(ctx context.Context, actorID string, remind
 				// Setting executionStatus to failed if workflow has failed/terminated/cancelled
 				executionStatus = diag.StatusFailed
 			}
+			wfExecutionElapsedTime = diag.ElapsedSince(state.workflowStartTime)
```
shivamkm07, line 652: This needs to be set at other places as well e.g. when workflow is returning with `recoverable error` or when workflow is completing successfully.

pkg/runtime/wfengine/backends/actors/workflow_actor.go
```diff
@@ -635,11 +649,24 @@ func (wf *workflowActor) runWorkflow(ctx context.Context, actorID string, remind
...
 		}
 	}
 	return nil
 }
 
+func (wf *workflowActor) recordWorkflowSchedulingLatency(ctx context.Context, isWorkflowExecutionStartedEvent bool, state *workflowState, scheduledStartTimestamp time.Time, workflowName string, executionStatus string) {
+	if isWorkflowExecutionStartedEvent {
+		state.workflowStartTime = time.Now()
+		// If scheduledStartTimestamp is zero, then scheduling latency is zero
```
shivamkm07, line 661: This case needs to be handled separately. In normal scenarios when scheduledStartTime is not given explicitly, workflow triggers immediately but still there would be some latency due to Reminder creation and invocation. Probably need to take the `startTimeStamp` as the time when reminder was created? thoughts @DeepanshuA @mukundansundar ?

pkg/diagnostics/workflow_monitoring.go
```diff
@@ -123,6 +137,26 @@ func (w *workflowMetrics) WorkflowExecutionEvent(ctx context.Context, workflowNa
 	stats.RecordWithTags(ctx, diagUtils.WithTags(w.workflowExecutionCount.Name(), appIDKey, w.appID, namespaceKey, w.namespace, workflowNameKey, workflowName, statusKey, status), w.workflowExecutionCount.M(1))
 }
 
+func (w *workflowMetrics) WorkflowExecutionLatency(ctx context.Context, workflowName, status string, elapsed float64) {
```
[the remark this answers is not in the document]
ASHIQUEMD: It's better to keep them separate as they do different logging, and this makes these methods focused on doing one thing.

pkg/runtime/wfengine/backends/actors/workflow_actor.go
```diff
@@ -417,13 +417,24 @@ func (wf *workflowActor) runWorkflow(ctx context.Context, actorID string, remind
...
 	for _, e := range state.Inbox {
 		var taskID int32
 		if ts := e.GetTaskCompleted(); ts != nil {
 			taskID = ts.GetTaskScheduledId()
 		} else if tf := e.GetTaskFailed(); tf != nil {
 			taskID = tf.GetTaskScheduledId()
 		} else {
+			// If the event is an execution started event, then we need to record the scheduled start timestamp
+			if es := e.GetExecutionStarted(); es != nil {
+				isWorkflowExecutionStartedEvent = true
```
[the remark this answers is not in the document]
ASHIQUEMD: We isWorkflowExecutionStartedEvent is more suitable here, please at the GetExecutionStarted method name.

pkg/runtime/wfengine/backends/actors/workflow_actor.go
ASHIQUEMD: In that case, we should have same value for both timeStamp and scheduledStartTime, and it's better to read value from only one property. I think you made some changes in [this PR](https://github.com/microsoft/durabletask-go/pull/60/files) to fix scheduleStartTime issue, we should fix this value there so that, it always has correct value, its better if caller does not check different properties for one value.

pkg/runtime/wfengine/backends/actors/workflowstate.go
```diff
@@ -127,6 +129,14 @@ func (s *workflowState) GetSaveRequest(actorID string) (*actors.TransactionalReq
 		return nil, err
 	}
 
+	// if s.workflowStartTime is not nil then append req.operations with a new operation to update the workflowStartTime
+	if !s.workflowStartTime.IsZero() {
```
ItalyPaleAle, line 133: This adds one more round-trip to the DB per each request. The performance impact is non-trivial, and I'm not sure it's worth for a metric?

Can we find another way that doesn't require writing and loading a new state key?

ItalyPaleAle (changes requested):

pkg/runtime/wfengine/backends/actors/workflowstate.go
yaron2: Agree, from both a performance and cost perspective this isn't desirable

dapr.sln
ItalyPaleAle, line 1: Please remove this file

pkg/diagnostics/workflow_monitoring.go
```diff
@@ -79,6 +83,14 @@ func newWorkflowMetrics() *workflowMetrics {
 			"runtime/workflow/activity/execution/latency",
 			"The total time taken to run an activity to completion.",
 			stats.UnitMilliseconds),
+		workflowExecutionLatency: stats.Float64(
+			"runtime/workflow/execution/latency",
+			"The total time taken to run workflow to completion.",
+			stats.UnitMilliseconds),
+		workflowSchedulingLatency: stats.Float64(
+			"runtime/workflow/scheduling/latency",
+			"The time taken between workflow execution request and actual workflow execution.",
```
ItalyPaleAle, line 92:
```suggestion
			"Interval between workflow execution request and workflow execution.",
```

pkg/runtime/wfengine/backends/actors/workflow_actor.go
```diff
@@ -417,13 +417,24 @@ func (wf *workflowActor) runWorkflow(ctx context.Context, actorID string, remind
 
 	// The logic/for loop below purges/removes any leftover state from a completed or failed activity
 	transactionalRequests := make(map[string][]actors.TransactionalOperation)
+	isWorkflowExecutionStartedEvent := false
+	scheduledStartTimestamp := time.Time{}
```
ItalyPaleAle, line 421: Nit: `time.Time{}` can be confusing (at least it is to me) as one needs to remember that this is not the current time.


```suggestion
	var (
        isWorkflowExecutionStartedEvent bool
	    scheduledStartTimestamp time.Time
    )
```

pkg/runtime/wfengine/backends/actors/workflow_actor.go
DeepanshuA: I agree that when it is scheduled immediately, we should consider all times involved i.e. Reminder creation, invocation - basically anything that accounts for time taken before actual execution could trigger.

ASHIQUEMD: Sure @DeepanshuA, changes has been made, please have a look.

pkg/runtime/wfengine/backends/actors/workflow_actor.go
ASHIQUEMD: As discussed offline, this is the only place we need to calculate wfExecutionElapsedTime.

shivamkm07:
LGTM. Just one comment - We can possibly add an integration test for execution latency here: https://github.com/dapr/dapr/blob/007ccf55219639ff7df8949ff22f00565463245a/tests/integration/suite/daprd/metrics/metrics.go#L186
e.g. a workflow with a sleep(5 seconds) should have executionLatency > 5sec
(We can actually add for scheduling latency too like for workflow with `ScheduledStartTimeStamp` set to `T`, `SchedulingLatency` > `T - currentTime` but this test won't pass since scheduling doesn't work correctly as of now: https://github.com/dapr/dapr/issues/7354)

artursouza (approved):

PR #7373, opened 2024-01-12
Add DAPR_GO_BUILD_TAGS=wfbackendsqlite for e2e test

ASHIQUEMD:
# Description

This PR adds DAPR_GO_BUILD_TAGS=wfbackendsqlite for e2e tests, where it is required to load sqlite backend component in dapr binaries for e2e tests related to sqlite backend.

This will fix the failing e2e tests [here](https://github.com/dapr/dapr/actions/runs/7498321229/job/20413628280#step:21:1726) for [this PR](https://github.com/dapr/dapr/pull/7283/files) related to configurable backend.

### DIFF
--- a/.github/workflows/dapr-test.yml
+++ b/.github/workflows/dapr-test.yml
@@ -330,6 +330,7 @@ jobs:
           echo "DAPR_TEST_TAG=${TEST_PREFIX}" >> $GITHUB_ENV
           echo "TEST_RESOURCE_GROUP=Dapr-E2E-${TEST_PREFIX}" >> $GITHUB_ENV
           echo "WINDOWS_VERSION=${{ matrix.windows_version }}" >> $GITHUB_ENV
+          echo "DAPR_GO_BUILD_TAGS=wfbackendsqlite" >> $GITHUB_ENV
         shell: bash
       - name: Build Dapr and its Docker images
         if: env.TEST_PREFIX != ''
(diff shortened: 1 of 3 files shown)


.github/workflows/dapr-test.yml
```diff
@@ -330,6 +330,7 @@ jobs:
           echo "DAPR_TEST_TAG=${TEST_PREFIX}" >> $GITHUB_ENV
           echo "TEST_RESOURCE_GROUP=Dapr-E2E-${TEST_PREFIX}" >> $GITHUB_ENV
           echo "WINDOWS_VERSION=${{ matrix.windows_version }}" >> $GITHUB_ENV
+          echo "DAPR_GO_BUILD_TAGS=wfbackendsqlite" >> $GITHUB_ENV
```
ItalyPaleAle, line 333: Let's find a different name for this tag please, such as "e2etest" , so it's not tied to a specific feature

ItalyPaleAle (changes requested):

.github/workflows/dapr-test.yml
mukundansundar: I really want this to be tied to this specific feature, When/If we have more different components like this later, we can generalise it. But for now I want this to be specific to wfbackendsqlite. 
Ideally we should not be having these tag specific components a lot.

Issue #7360, opened 2024-01-10
Remove vnodes from dissemination table


elena-kolevska:
## In what area(s)?

/area placement

## Describe the feature

Currently, the virtual nodes (vnodes) of the consistent hashing ring are disseminated by the placement service to the dapr sidecars as part of the placement table. The number of vnodes is defined by the `replicationFactor` parameter of the placement service and its default value is 100.

This configuration results in 200 additional entries for each actor type in the system: 100 entries for hosts and another 100 in the sorted set of hosts. For illustration purposes, here’s an example of how the table might look like:

```json
{
  "entries": {
    "myactortype-0": {
      "hosts": {
        "10056481384176189962": "192.168.1.89:62362",
        "10100244799470048543": "192.168.1.89:62362",
        "10136976372505843558": "192.168.1.89:62362",
        "10210620769057790472": "192.168.1.89:62362",
        "10278294263954349758": "192.168.1.89:62345",
		    ...
      },
      "sorted_set": [
        3430792224860206,
        10660957680821358,
        69520260237277928,
        96174763468296109,
        110417171835179485,
        ...
      ],
      "load_map": {
        "192.168.1.89:62345": {
          "name": "192.168.1.89:62345",
          "id": "08aa8f86-df9b-45bd-b407-f35c67535073"
        },
        "192.168.1.89:62362": {
          "name": "192.168.1.89:62362",
          "id": "140d66c3-66b4-4642-ade5-0d17b4cd0402"
        },
        "192.168.1.89:62367": {
          "name": "192.168.1.89:62367",
          "id": "6c1f823e-b3b0-43d6-b7ca-ce844ae1ae2c"
        }
      }
    },
    "myactortype-1": {
      "hosts": {
        "10056481384176189962": "192.168.1.89:62362",
        "10100244799470048543": "192.168.1.89:62362",
        "10136976372505843558": "192.168.1.89:62362",
        ...
      },
      "sorted_set": [
        69520260237277928,
        110417171835179485,
        18297525576440347375,
       ...
      ],
      "load_map": {
        "192.168.1.89:62362": {
          "name": "192.168.1.89:62362",
          "id": "140d66c3-66b4-4642-ade5-0d17b4cd0402"
        }
      }
    }
  },
  "version": "3",
  "api_level": 20
}
```

In systems with a large number of actors and sidecars, this approach leads to excessively large placement table sizes,  and causes `ResourceExhausted` errors:

```json
[...]

PR #7382, opened 2024-01-15
Calculate vnodes on daprd instead of the placement service

elena-kolevska:
# Description

(Note: Please refer to the [issue](https://github.com/dapr/dapr/issues/7360) for the reason behind this PR)

This PR removes the vnodes and their sorted set from the placement table that's disseminated from the placement service to the dapr sidecars, and instead, it calculates them on the dapr sidecars.
It accounts for mismatch of versions in daprd/placement services (older daprd version with placement on 1.13+ and vice versa). In the next Dapr version we should be able to remove these checks and clean up the code. 

A few notes that might help during the review:
- `replicationFactor` is an argument received on placement service startup, whose default value is 100. 
- We pass the replication factor as part of the placement table message to daprd, so we can calculate the vnodes there
- We attach the api level of the daprd as metadata to the stream connection. The placement service will look at this data and decide if it needs to send the vnodes based on it (needed for daprd sidecars on older versions). This way we're not relying on the minimum API level of the cluster, but serve data based on what the particular sidecar needs.
- We still need to keep the vnodes in state on the placement side because of backwards compatibility.

## Issue reference
https://github.com/dapr/dapr/issues/7360

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests
* [x] Unit tests passing

N/A


pkg/actors/internal/api_level.go
```diff
@@ -27,10 +27,14 @@ const ActorAPILevel = 20
...
 const (
-	// Enables serializing reminders as protobuf rather than JSON in the pkg/actors/reminders package
+	// APILevelFeatureRemindersProtobuf Enables serializing reminders as protobuf rather than JSON in the pkg/actors/reminders package
 	// When serialized as protobuf, reminders have the "\0pb" prefix
 	// Note this only control serializations; when un-serializing, legacy JSON is always supported as fallback
 	APILevelFeatureRemindersProtobuf apiLevelFeature = 20
+
+	// NoVirtualNodesInPlacementTables - The placement table doesn't include the virtual nodes
+	// Instead, they're calculated on the client (daprd) side
+	NoVirtualNodesInPlacementTables apiLevelFeature = 20
```
artursouza, line 37: Why the same number as above?

artursouza (changes requested):
This PR warms my heart because it addresses an obvious problem that nobody dared to point out AND fix. Thanks a lot.

pkg/actors/internal/api_level.go
ItalyPaleAle: This indicates the API level. For Dapr 1.12, it's 10. In 1.13 it will be 20. #6838

pkg/placement/hashing/consistent_hash.go
```diff
@@ -90,7 +82,33 @@ func NewConsistentHash() *Consistent {
 }
 
 // NewFromExisting creates a new consistent hash from existing values.
-func NewFromExisting(hosts map[uint64]string, sortedSet []uint64, loadMap map[string]*Host) *Consistent {
+func NewFromExisting(loadMap map[string]*Host, replicationFactor int) *Consistent {
```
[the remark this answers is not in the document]
elena-kolevska: So my thinking was that we rename the old behaviour to `NewFromExistingWithVirtNodes` and then we remove it in the next version. And we keep the new behaviour with a name that's gonna read more naturally in this version, but also in the next ones.

elena-kolevska:
I added caching for the virtual hosts. I'll be adding a few more tests today and also I'll fix the `Test_Integration/placement/apilevel/noMax/run` test that seems I've broken.

pkg/placement/membership.go
```diff
@@ -325,15 +329,17 @@ func (p *Service) performTableDissemination(ctx context.Context) error {
 	p.disseminateLock.Lock()
 	defer p.disseminateLock.Unlock()
 
-	state := p.raftNode.FSM().PlacementState()
+	state := p.raftNode.FSM().PlacementState(false)
+	stateWithVirtualNodes := p.raftNode.FSM().PlacementState(true)
```
elena-kolevska, line 333: I wonder if we can use `p.minAPILevel` to check if there are any older daprds connected to the service. If not, we don't need to calculate `stateWithVirtualNodes`. 
I think that's what that property is, but I have to double-check. If anyone knows and confirm that for me in the meantime, it would b great.

ItalyPaleAle: `minAPILevel` is cluster-wide. Although in the state we do keep the API level of each connected host.

pkg/actors/placement/placement.go
```diff
@@ -532,7 +536,14 @@ func (p *actorPlacement) updatePlacements(in *v1pb.PlacementTables) {
 			for lk, lv := range v.GetLoadMap() {
 				loadMap[lk] = hashing.NewHost(lv.GetName(), lv.GetId(), lv.GetLoad(), lv.GetPort())
 			}
-			p.placementTables.Entries[k] = hashing.NewFromExisting(v.GetHosts(), v.GetSortedSet(), loadMap)
+
+			// TODO in v1.14 remove the check for versions < 1.13
```
ItalyPaleAle, line 540: Probably 1.15, we try to keep n-2 compat

pkg/placement/hashing/consistent_hash.go
```diff
@@ -90,7 +82,104 @@ func NewConsistentHash() *Consistent {
...
+		data: make(map[int64]*hashMap),
+	}
+}
+
+// GetHashes retrieves the hashes for the given replication factor and IP address.
+func (hc *VirtualNodesCache) GetHashes(replicationFactor int64, host string) []uint64 {
+	hc.RLock()
+	if hashMap, exists := hc.data[replicationFactor]; exists {
+		if hashes, found := hashMap.hashes[host]; found {
+			return hashes
```
ItalyPaleAle, line 143: Don't we need `hc.RUnlock()` before this?

pkg/placement/hashing/consistent_hash.go
```diff
@@ -90,7 +82,104 @@ func NewConsistentHash() *Consistent {
...
+		}
+	}
+
+	hashMap := newHashMap()
+	hashMap.hashes[host] = []uint64{}
+
+	c := &Consistent{}
+	for i := 0; i < int(replicationFactor); i++ {
+		h := c.hash(fmt.Sprintf("%s%d", host, i))
```
ItalyPaleAle, line 170: Please avoid the use of `fmt.Sprintf` for concatenation in a hot path. Can be replaced with `host + strconv.Itoa(i)`

pkg/placement/hashing/consistent_hash.go
```diff
@@ -116,15 +205,24 @@ func (c *Consistent) Add(host, id string, port int64) bool {
 	}
 
 	c.loadMap[host] = &Host{Name: host, AppID: id, Load: 0, Port: port}
+
+	// TODO in v1.14
```
ItalyPaleAle, line 209: 1.15?

pkg/placement/membership.go
```diff
@@ -366,19 +372,29 @@ func (p *Service) performTablesUpdate(ctx context.Context, hosts []placementGRPC
 		}
 	}
 
+	// Enforce maximum API level
```
ItalyPaleAle, line 375: Perhaps this enforcement can be done just once for both `newTable` and `newTableWithVirtualNodes`, where we compute the "final" API level

pkg/placement/hashing/consistent_hash.go
```diff
@@ -90,7 +82,104 @@ func NewConsistentHash() *Consistent {
...
+	}
+
+	hashMap := newHashMap()
+	hashMap.hashes[host] = []uint64{}
+
+	c := &Consistent{}
+	for i := 0; i < int(replicationFactor); i++ {
+		h := c.hash(fmt.Sprintf("%s%d", host, i))
+		hashMap.hashes[host] = append(hashMap.hashes[host], h)
+	}
```
ItalyPaleAle, line 172: If I understand correctly, this will contain `replicationFactor` items, so we can improve the allocator efficiency:

```suggestion
	hashMap.hashes[host] = make([]uint64, replicationFactor)

	c := &Consistent{}
	for i := 0; i < int(replicationFactor); i++ {
		hashMap.hashes[host][i] = c.hash(host + strconv.Itoa(i))
	}
```

pkg/actors/placement/placement.go
elena-kolevska: Got it. Updating all instances of this note.

pkg/placement/membership.go
elena-kolevska: Great! Thank you! So I'll add a check, if `minApiLevel` is >=20, I won't calculate`stateWithVirtualNodes`. 🙏

ItalyPaleAle: Since we're on topic:

> We attach the api level of the daprd as metadata to the stream connection. The placement service will look at this data and decide if it needs to send the vnodes based on it (needed for daprd sidecars on older versions). This way we're not relying on the minimum API level of the cluster, but serve data based on what the particular sidecar needs.

Just a note that if a sidecar tries to connect to Placement and the reported API level is lower than the cluster's, it's rejected (at least, rejected as actor host).

elena-kolevska: Thanks for pointing that out. I went at looked at https://github.com/dapr/dapr/pull/7134 in the meantime so it's more clear now.
I guess the comment still applies though, right? It would be a case where placement is upgraded before the daprd-s. I can update the comment to reflect this better.

ItalyPaleAle: Yes I think your code is fine. My comment was more of a clarification on how this works.

pkg/placement/hashing/consistent_hash.go
elena-kolevska: What a save! Thank you.

pkg/placement/membership.go
elena-kolevska: We could have any combination of  `newTable` and `newTableWithVirtualNodes` being present, so I preferred leaving them separate like this, so we don't have to "untangle" them in a few months when we have to remove `newTableWithVirtualNodes`.

ItalyPaleAle: Ok

pkg/actors/internal/api_level.go
artursouza: I am new to this one. Thanks.

artursouza (approved):

Issue #7384, opened 2024-01-15
Workflow Sub-Orchestration creation not working


shivamkm07:
## In what area(s)?

 /area runtime

> /area operator

> /area placement

> /area docs

> /area test-and-release

## What version of Dapr?

> edge: output of `git describe --dirty`

## Expected Behavior

We should be able to create Child workflow and run it without any errors.

## Actual Behavior

When trying to create and run a child workflow, workflow execution simply stalls with workflow never completing, and finally erroring out with `Context deadline exceeded`.

In Daprd logs, following log can be found:
`iowriter.go:89: Test_Integration/daprd/workflow/workflow/daprd: time="2024-01-15T22:18:22.228973448+05:30" level=warning msg="Workflow actor '73550139-539f-443d-a5ce-bcdab2d65832': execution failed with a recoverable error and will be retried later: 'method CreateWorkflowInstance on actor '73550139-539f-443d-a5ce-bcdab2d65832:0000' returned an error: failed to unmarshal createWorkflowInstanceRequest: invalid character '\\b' looking for beginning of value'" app_id=myapp instance=MININT-0A33K9K scope=dapr.wfengine.backend.actors type=log ver=unknown`

## Steps to Reproduce the Problem

Try to run a workflow that creates a child workflow. It errors out as described above.

## Release Note

RELEASE NOTE:

PR #7385, opened 2024-01-15
Fix workflow subOrchestration creation

shivamkm07:
# Description

It fixes workflow sub-orchestration creation by passing correctly the `CreateWorkflowInstanceRequest` at the time of Sub-orchestration creation. The struct was added as part of #7308. 

The PR also adds integration test for the same. 

## Issue reference

Please reference the issue this PR will close: #7384 

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing


mukundansundar (approved):

DeepanshuA (approved):
I tested it at my end as well, there was an issue and with this fix, it is working. LGTM

pkg/runtime/wfengine/backends/actors/workflow_actor.go
```diff
@@ -598,12 +598,21 @@ func (wf *workflowActor) runWorkflow(ctx context.Context, actorID string, remind
 			if errMarshal != nil {
 				return errMarshal
 			}
-
+			requestBytes := eventData
```
ItalyPaleAle, line 601: Is this not backwards-incompatible? If a sidecar that is running 1.12 receives this message, it may not know how to parse this?

shivamkm07: I think not. If a sidecar is running 1.12, here we will be sending only `startEvent` as rawData: https://github.com/dapr/dapr/blob/release-1.12/pkg/runtime/wfengine/workflow.go#L485

and then we would be parsing that only in `CreateWorkflowInstanceMethod`: https://github.com/dapr/dapr/blob/986d38b8d0524753bb5331a3349e11f17200d3ca/pkg/runtime/wfengine/workflow.go#L170

The issue here is PR #7308 added the struct here: https://github.com/dapr/dapr/blob/ded0ace346ebb129ac319b961dd05516bc0b098d/pkg/runtime/wfengine/backends/actors/workflow_actor.go#L180 but missed to add here and so the inconsistency causing the issue.

cgillum:
If this was a regression, we should consider adding a unit test in `pkg/runtime/wfengine/wfengine_test.go` to cover the child workflows scenario (a.k.a. sub-orchestrations).

pkg/runtime/wfengine/backends/actors/workflow_actor.go
ItalyPaleAle: So #7308 is not backwards-compatible?

shivamkm07: I meant #7308 is backwards compatible. It's just that it introduced a bug and the current PR fixes it. But it should worknwell even if the sidecar is from 1.12.

mukundansundar (approved):

PR #6818, opened 2023-08-16
House cleaning: Move grpc, http, and universal under `pkg/api`. Make Universal immutable.

JoshVanL:
General house cleaning PR to move the `grpc`, `http` and `universal` packages under `pkg/api`.

Makes the shared universal object immutable by callers.

Based on https://github.com/ItalyPaleAle/dapr/pull/97

PR #7390, opened 2024-01-16
Fix build failures with "subtlecrypto" tag and re-add E2E tests for building block

ItalyPaleAle:
Fixes a regression introduced by #6818 that caused builds to fail when the "subtlecrypto" tag is enabled.

Re-adds the E2E tests for the building block (that were never merged), so errors like these can be caught in the future.


.github/workflows/dapr-test.yml
```diff
@@ -55,8 +55,13 @@ env:
   AZURE_ARM_REGIONS: "eastus"
   # Container registry where to cache e2e test images
   DAPR_CACHE_REGISTRY: "dapre2eacr.azurecr.io"
+  # Name of the Azure Key Vault resource used in tests
+  # The credentials defined in AZURE_CREDENTIALS must have permissions to perform operations in this vault
+  AZURE_KEY_VAULT_NAME: "dapre2ekv"
   # Whether to collect TCP dumps
   TCP_DUMPS: "false"
+  # Additional build tags for Dapr
+  DAPR_GO_BUILD_TAGS: "subtlecrypto,wfbackendsqlite"
```
ItalyPaleAle, line 64: The subtlecrypto E2E tests only run on AKS and not KIND

yaron2 (approved):

yaron2 (approved):

PR #7391, opened 2024-01-16
Updated pinned components-contrib

ItalyPaleAle:
Updates the pinned components-contrib as we get into code freeze

Delete with Prefix added to in-memory state store as part of https://github.com/dapr/components-contrib/pull/3314


daixiang0 (approved):
LGTM

mukundansundar (approved):

PR #7394, opened 2024-01-17
Adds actors healthz integration tests

### LINKED ISSUE #7355: Having healthz endpoint with appHealthCheck false
## In what area(s)?

/area runtime

> /area operator

/area placement

> /area docs

> /area test-and-release

## What version of Dapr?

## Expected Behavior

If app health check is disabled, it shouldn't be required for applications to implement `healthz` endpoint.

## Actual Behavior

Even with app health disabled, If app doesn't implement healthz, actors(and so Dapr workflows) don't work. This happens because placement detects app as unhealthy and disconnects with log `Disconnecting from placement service by the unhealthy app.` 

1. Run `tests/integration/suite/actors/healthz/deactivateOnPlacementFail` test with healthz endpoint commented out: https://github.com/dapr/dapr/blob/328b4b33df71f073b2da0d96bb30940880179955/tests/integration/suite/actors/healthz/deactivate-on-placement-fail.go#L58
2. It would fail even when app health check is false

## Release Note

RELEASE NOTE:

JoshVanL:
Adds integration tests for testing actor healthz behaviour when entities and or app healthz is defined.

Closes #7355

~~Mark actors as always healthy when there is no app channel or their have been~~
~~no actor entities (actor types) provided by the application config endpoint.~~
~~This is important for enabling actor related APIs such as workflows, which~~
~~don't rely on the application implementing actors itself.~~

~~Adds integration tests for actor healthz endpoints Most importantly, ensuring~~
~~that actors are marked as always healthy when the both the application healthz~~
~~endpoint is disabled, and the application reports no actor entities.~~


tests/integration/suite/actors/healthz/endpoint/noappentities.go
```diff
@@ -0,0 +1,142 @@
...
+		prochttp.WithHandlerFunc("/dapr/config", func(w http.ResponseWriter, r *http.Request) {
+			w.Write([]byte(`{}`))
+		}),
+		prochttp.WithHandlerFunc("/healthz", func(w http.ResponseWriter, r *http.Request) {
+			w.WriteHeader(http.StatusOK)
+			onceNoHealthz.Do(func() {
+				close(n.noHealthzCalled)
+			})
+		}),
+	)
```
mukundansundar, line 81: qq when there are no actor types and health check is disabled, why is a healthz endpoint needed and is being called?

mukundansundar (approved):

mukundansundar:
@JoshVanL The IT seem to failing in the specific test that was written in this PR.

JoshVanL:
Thanks @mukundansundar, it's clear this PR is causing some kind of unexpected side effect somewhere. I'm taking a look

pkg/actors/placement/placement.go
```diff
@@ -180,6 +180,7 @@ func (p *actorPlacement) Start(ctx context.Context) error {
 		defer p.shutdownConnLoop.Done()
 		ch := p.appHealthFn(ctx)
 		if ch == nil {
+			p.appHealthy.Store(true)
```
ItalyPaleAle, line 183: This seems a bit off. I don't think we can do this because if the app has a HTTP app channel, `ch` will be non-nil.

Can you check `pkg/actors/actors.go`? Specifically, where `a.checker` is initialized:

https://github.com/JoshVanL/dapr/blob/a447ea75cd47e4c28256d6a5e4374dfadeb156ea/pkg/actors/actors.go#L278

Perhaps the way to fix this issue is to simply not initialize it if `len(hat) == 0`?

https://github.com/JoshVanL/dapr/blob/a447ea75cd47e4c28256d6a5e4374dfadeb156ea/pkg/actors/actors.go#L261

```go
	if len(hat) > 0 {
		a.checker, err = a.getAppHealthChecker()
		if err != nil {
			return fmt.Errorf("actors: couldn't create health check: %w", err)
		}
	}
```

ItalyPaleAle (changes requested):

pkg/actors/placement/placement.go
JoshVanL: We should be covering this scenario in the `noentities` test.

https://github.com/dapr/dapr/pull/7394/files/a447ea75cd47e4c28256d6a5e4374dfadeb156ea#diff-104d2d64415550335dafcda3cb0730e5d866648f66f25ce74a40faffad722b58

ItalyPaleAle: Hold on, then if `ch` is already `nil` when entities is empty.... There's nothing we need to do in this PR? Was the issue already fixed? (And just needed tests?)

ItalyPaleAle: When we start placement, we already set appHealthy to true:

https://github.dev/JoshVanL/dapr/blob/a447ea75cd47e4c28256d6a5e4374dfadeb156ea/pkg/actors/placement/placement.go#L159

And if `ch` is nil, then this can never be set to false

(And if that weren't the case, workflow tests would have failed already sinced they don't implement /healthz)

JoshVanL: Without the code change made in this PR, the new test `noentities` will fail. i.e. actors will never become ready.

ItalyPaleAle: It works on my machine without this line. Please try it!

artursouza:
This is a regression because a sidecar that does not host actors can always use actors that are hosted by other instances. If this is the case, this is a hotfix.

tests/integration/suite/actors/healthz/endpoint/noapp.go
```diff
@@ -0,0 +1,106 @@
...
+	"github.com/dapr/dapr/tests/integration/framework/util"
+	"github.com/dapr/dapr/tests/integration/suite"
+)
+
+func init() {
+	suite.Register(new(noapp))
+}
+
+// noapp ensures that the daprd `/healthz` endpoint is called and actors
+// will respond successfully when invoked, even when the app healthz endpoint is
```
ItalyPaleAle, line 40:
```suggestion
// will respond successfully when invoked, even when the app health checks is
```

tests/integration/suite/actors/healthz/endpoint/noappentities.go
```diff
@@ -0,0 +1,142 @@
...
+	"github.com/dapr/dapr/tests/integration/framework/util"
+	"github.com/dapr/dapr/tests/integration/suite"
+)
+
+func init() {
+	suite.Register(new(noappentities))
+}
+
+// noappentities ensures that actors work even though no entities have been
+// defined and the app healthz endpoint is disabled.
```
ItalyPaleAle, line 40:
```suggestion
// defined and the app healthchecks is disabled.
```

pkg/actors/placement/placement.go
artursouza: @JoshVanL Please, double check this because we already have E2E to validate that sidecars that don't serve actors types can still invoke other actors.

artursouza: BTW, happy to take the integration test part of it.

artursouza (dismissed):

pkg/actors/placement/placement.go
JoshVanL: Done! I have just included the integration tests now.

ItalyPaleAle (approved):

PR #7400, opened 2024-01-17
Update placement API Level gRPC context key to `dapr-placement-api-level`

JoshVanL:
Updates the placement API Level gRPC context key from `ApiLevel` to `dapr-placement-api-level`. Although collisions are highly unlikely, it could be considered more "correct" and best practice to namespace our application specific metadata keys. It could also help in downstream tracers and metric consumers debugging RPCs.

This isn't a breaking change as this key hasn't been part of a release yet.

cc @elena-kolevska


elena-kolevska (approved):
Thanks Josh

cicoyle (approved):

PR #7423, opened 2024-01-22
Moves public daprd metadata server to new authorized port

### LINKED ISSUE #7397: Authenticate daprd public port
Currently, daprd is exposing the public port which exposes the healthz and metadata endpoints. This is exposed publicly on the network. This server has no authentication and is serving requests in plaintext.

The metadata endpoint returns information like actors, subscriptions, components etc.

This represents a massive privacy exposure. This port must be disabled on any multi-tenant deployment of Dapr, however this is not documented anywhere and the daprd logs don't explain that this is happening either. Users are currently exposing this information without knowledge.

We should update this API server to use mTLS based on the dapr identity certificate, and provide an allow list of identities which can connect to this endpoint to authorize it consumption. This is a breaking change, however the current behaviour is not appropriate and should be considered a security vulnerability.

JoshVanL:
Currently, daprd is exposing the public port which exposes the healthz and metadata endpoints. This is exposed publicly on the network. This server has no authentication and is serving requests in plaintext.

The metadata endpoint returns information like actors, subscriptions, components etc.

This represents a massive privacy exposure. This port must be disabled on any multi-tenant deployment of Dapr, however this is not documented anywhere and the daprd logs don't explain that this is happening either. Users are currently exposing this information without knowledge.

This PR moves the public metadata endpoint to a new port, and adds authentication and authorization to the endpoint. If mTLS is enabled, the endpoint will serve using the same TLS as the internal API server. The server will only serve requests to SPIFFE peers that present in the `--dapr-metadata-authorized-ids` CLI flag. If mTLS is disabled, this flag is ignored.

PR extends the mTLS Configuration API with a new `metadataAuthorizedIDs` fields which takes a slice of SPIFFE IDs. This slice is appended to the authorized list when consumed by daprd. When set on the global `dapr-system` configuration, the sidecar-injector sets this value on the sidecar container. This allows users to configure the authorized IDs globally, or on a per-pod basis.

Closes https://github.com/dapr/dapr/issues/7397


ItalyPaleAle (changes requested):
Although I appreciate that this is done the "right way" and requires mTLS, we should consider how the metadata API is likely being used, which is that apps may call it to get information on the sidecar, or possibly even external monitoring/cataloging services.

Adding the requirement of using mTLS, fetching certs from Sentry, in this case is IMHO very burdensome given the kind of consumers.

I would rather add a more simple IP-based allowlisting. Or just trust that users will not expose the port in the K8s service.

artursouza:
> Although I appreciate that this is done the "right way" and requires mTLS, we should consider how the metadata API is likely being used, which is that apps may call it to get information on the sidecar, or possibly even external monitoring/cataloging services.
> 
> Adding the requirement of using mTLS, fetching certs from Sentry, in this case is IMHO very burdensome given the kind of consumers.
> 
> I would rather add a more simple IP-based allowlisting. Or just trust that users will not expose the port in the K8s service.

I agree with @ItalyPaleAle. At least, offer a way for people to keep the existing behavior. We should do that anyway because it is a breaking change.

artursouza (changes requested):
The current behavior must be kept and the new behavior should be done via a config flag. Default must be current behavior, following the deprecation policy.

JoshVanL:
The project should always strive to do things the right way, especially so for security. I don't think we should follow the deprecation policy in this instance as this is an undocumented privacy vulnerability in multi tenant deployments.

Applications that currently consume the metadata API can continue to do so via the localhost endpoints. Communication between Dapr hosts must always be mTLS by default in Kubernetes. Any system which is currently consuming this API should have enough knowledge of Dapr already to be able to request an identity certificate from sentry using the gRPC protos. We must have a mechanism to authorize this endpoint- IP allowlists are not appropriate in cloud native environments as IP addresses are ephemeral.

artursouza:
> The project should always strive to do things the right way, especially so for security. I don't think we should follow the deprecation policy in this instance as this is an undocumented privacy vulnerability in multi tenant deployments.
> 
> Applications that currently consume the metadata API can continue to do so via the localhost endpoints. Communication between Dapr hosts must always be mTLS by default in Kubernetes. Any system which is currently consuming this API should have enough knowledge of Dapr already to be able to request an identity certificate from sentry using the gRPC protos. We must have a mechanism to authorize this endpoint- IP allowlists are not appropriate in cloud native environments as IP addresses are ephemeral.

I understand your rationale. On the other hand, an OSS project have to be careful on forcing a migration on a minor version upgrade. It can lead to the opposite result, where users stick to an older version of Dapr to plan ahead their migration in order to upgrade to a version that requires a significant change of their environment - there can be internal tools depending on this behavior that only the user knows.

ItalyPaleAle:
(Please also note that this metadata API is not enabled by default, so users must enable it explicitly)

Before we make decisions here we should really inquire on how users are using this API. My feeling is that it used by external services for cataloging or other purposes, for whom requesting a certificate from Sentry would not be feasible and would make the API unusable. (Also, we do not have public-facing SDKs for Sentry)

Agree IP allowlisting isn't great, but this leaves us with 2 options:

1. Let users figure it out. We can add a warning when the metadata endpoint is enabled that it can expose sensitive information and users should be careful when exposing that.
2. Shared key authentication, combined with TLS (not mTLS). Not great but better than nothing.

JoshVanL:
@ItalyPaleAle it is enabled by default in Kubernetes.

ItalyPaleAle:
> @ItalyPaleAle it is enabled by default in Kubernetes.

Ok that's a different problem then. I thought the PR that implemented that made it optional

WhitWaldo:
For what it's worth, I'll chime in to repeat what I said on today's release call: the API token is set by developers to proactively opt into security on the Dapr APIs, which to me is ample support that they've consented and acknowledged the risks in restricting all Dapr APIs to that token, metadata included.

This has collected feedback for quite a long time and I'd urge us to close this gap and lock it down.

Issue #7411, opened 2024-01-18
Breaking Change: Can no longer create component with name `dapr`


JoshVanL:
It is now no longer possible to create a Component with the name `dapr`.

```
FATA[0000] Fatal error from runtime: process component dapr error: [INIT_COMPONENT_FAILURE]: initialization error occurred for dapr (workflow.dapr/v1): component dapr already exists  app_id=foo instance=purple scope=dapr.runtime type=log ver=unknown
```

JoshVanL:
We should rename the internal workflow component name to be `_dapr_workflow` or `_dapr` so that it is internally namespaced.

We should change the type of this component to be `workflowbackend.dapr`.

We should remove this always installed Component from the metadata endpoint Comeponent list, and instead report the workflow backend status on the workflow metadata fields.

mukundansundar:
@JoshVanL This seems to be a larger issue than `workflow.dapr` component having a `dapr` name. 
For example in 1.12.4, you can have a component with name `pubsub` of `pubsub.redis` type and a component with name `pubsub` of `statestore.redis`. This seems to work at least in the init scenario (I have not verified the behavior yet(as in which component def gets the precedence), just that init works).

But in master, that is currently not possible. Any component loaded with a particular name, cannot be loaded again. 
This I believe is the main issue, it is not entirely related to the `workflow.dapr` component since that component has existed in the code even before changes for 1.13. 

cc @yaron2 @artursouza

JoshVanL:
@mukundansundar This behaviour was changed in https://github.com/dapr/dapr/pull/7195. Allowing same name different types was at best a bug, and at worst allowing users to put components in corrupted states.

This only effected self hosted mode and was considered on oversight. Kubernetes enforces uniqueness of names on resource objects already.

Without having a unique index on name, hot reloading will not work.

mukundansundar:
@JoshVanL This specific component should not be changed to `workflowbackend.dapr` type since it is specifically the dtf engine that is being initialized here and not the backend. 
@ASHIQUEMD will provide more details on this

ASHIQUEMD:
Hi @mukundansundar,

I appreciate your clarification, and I want to emphasize the distinction between the two types of components in the context of workflow:

**`workflow.dapr` (Workflow Engine component):**

This type represents the workflow engine component.
It is initialized by default without the need for a separate component YAML file.


**`workflowbackend.sqlite` or `workflowbackend.actor` (Workflow Backend component):**

[...]

PR #7427, opened 2024-01-22
Make Workflow `dapr` Component internal

JoshVanL:
The internal workflow engine `dapr` component is currently being registered in the processor component store, making it visible to the user via the metadata registered component list and preventing the user creating a Component of the same name.

PR updates the workflow engine machinery to `Init` the component manually, bypassing the processor and preventing it being committed to the component store.

Adds integration test to ensure a `dapr` named component can be created by the user.

Updates tests that previously expected the workflow engine component to be registered in the component store.

closes https://github.com/dapr/dapr/issues/7411

/cc @mukundansundar @ASHIQUEMD

### DIFF
--- a/pkg/runtime/processor/processor.go
+++ b/pkg/runtime/processor/processor.go
@@ -48,8 +48,7 @@ import (
 	"github.com/dapr/dapr/pkg/runtime/processor/pubsub"
 	"github.com/dapr/dapr/pkg/runtime/processor/secret"
 	"github.com/dapr/dapr/pkg/runtime/processor/state"
-	wfbeProcessor "github.com/dapr/dapr/pkg/runtime/processor/wfbackend"
-	"github.com/dapr/dapr/pkg/runtime/processor/workflow"
+	"github.com/dapr/dapr/pkg/runtime/processor/wfbackend"
 	"github.com/dapr/dapr/pkg/runtime/registry"
 	"github.com/dapr/kit/concurrency"
 	"github.com/dapr/kit/logger"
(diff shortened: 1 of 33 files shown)


pkg/runtime/runtime.go
```diff
@@ -644,12 +645,22 @@ func (a *DaprRuntime) initWorkflowEngine(ctx context.Context) error {
...
+		return fmt.Errorf("failed to initialize Dapr workflow component: %w", err)
 	}
-	return nil
+
+	log.Info("Workflow engine initialized.")
+
+	return a.runnerCloser.AddCloser(func() error {
+		return wfe.Close(wfengine.ComponentDefinition())
+	})
 }
```
JoshVanL, line 664: This is the main crux of the PR.

pkg/runtime/processor/processor.go
```diff
@@ -209,12 +208,7 @@ func New(opts Options) *Processor {
 			components.CategorySecretStore:     secret,
 			components.CategoryStateStore:      state,
 			components.CategoryWorkflowBackend: wfbe,
-			components.CategoryWorkflow: workflow.New(workflow.Options{
-				Registry:       opts.Registry.Workflows(),
-				ComponentStore: opts.ComponentStore,
-				Meta:           opts.Meta,
-			}),
-			components.CategoryMiddleware: middleware.New(),
+			components.CategoryMiddleware:      middleware.New(),
```
JoshVanL, line 211: Ensures the workflow Component type cannot be added by the user.

ASHIQUEMD (approved):
LGTM

artursouza (approved):

PR #7428, opened 2024-01-22
Placement service lock optimisations and race condition fix

elena-kolevska:
# Description

This PR introduces a few placement service optimisations:
- Calculates the vnodes for all hosts outside of the placementTable lock
- Calculates the vnodes for a single host outside of lock when adding a host to the placement table (which is still done for backwards compatibility)

A race condition fix:
- The `resetPlacementTables` method was called without a lock from two places - on actor placement start and on reconnection to the placement service.

This PR It also clean ups the `apiLevel` parameter from the `connectToServer` method, because it's not used anymore.

## Checklist

* [X] Code compiles correctly
* [X] Unit tests passing


pkg/actors/placement/client.go
```diff
@@ -48,7 +48,7 @@ type placementClient struct {
 
 // connectToServer initializes a new connection to the target server and if it succeeds replace the current
 // stream with the connected stream.
-func (c *placementClient) connectToServer(ctx context.Context, serverAddr string, apiLevel uint32) error {
+func (c *placementClient) connectToServer(ctx context.Context, serverAddr string) error {
```
elena-kolevska, line 51: This is just a leftover cleanup from https://github.com/dapr/dapr/pull/7415

JoshVanL (approved):

PR #7429, opened 2024-01-23
Add option to restore high cardinality for metrics for HTTP server

### LINKED ISSUE #7295: Daprd Metrics Breaking Change
With https://github.com/dapr/dapr/pull/6919 merged, Dapr has a breaking change to the metrics which are exposed. This can be seen with the failing version skew metrics test https://github.com/dapr/dapr/actions/runs/7169171492/job/19518959642.

As a project, we need to make the decision as to whether we accept this as a breaking change, or introduce a flag to revert the previous behaviour.

ItalyPaleAle:
This is configured using the Configuration CRD. High cardinality is enabled by default in Dapr 1.13, but shows a warning. The default behavior will be changed in 1.14.

Fixes #7295


msfussell:
Before merging this issue, please create a docs PR to update the following
1) https://docs.dapr.io/operations/configuration/configuration-overview/
2) https://docs.dapr.io/reference/resource-specs/configuration-schema/

charts/dapr/crds/configuration.yaml
```diff
@@ -286,6 +296,16 @@ spec:
                 properties:
                   enabled:
                     type: boolean
+                  http:
+                    description: MetricHTTP defines configuration for metrics for
+                      the HTTP server
+                    properties:
+                      increasedCardinality:
+                        description: 'If false, metrics for the HTTP server are collected
```
mukundansundar, line 304: same nitpick

pkg/apis/configuration/v1alpha1/types.go
```diff
@@ -212,9 +212,20 @@ type ZipkinSpec struct {
...
+	HTTP *MetricHTTP `json:"http,omitempty"`
+	// +optional
 	Rules []MetricsRule `json:"rules,omitempty"`
 }
 
+// MetricHTTP defines configuration for metrics for the HTTP server
+type MetricHTTP struct {
+	// If false, metrics for the HTTP server are collected with increased cardinality.
+	// The default is true in Dapr 1.13, but will be changed to false in 1.14+
+	// TODO @ItalyPaleAle: Metrics cardinality
```
JoshVanL, line 224: Intended to be committed?

pkg/config/configuration.go
```diff
@@ -257,6 +259,25 @@ func (m MetricSpec) GetEnabled() bool {
 	return m.Enabled == nil || *m.Enabled
 }
 
+// GetHTTPIncreasedCardinality returns true if increased cardinality is enabled for HTTP metrics
+func (m MetricSpec) GetHTTPIncreasedCardinality(log logger.Logger) bool {
+	if m.HTTP == nil || m.HTTP.IncreasedCardinality == nil {
+		// The default is true in Dapr 1.13, but will be changed to false in 1.14+
+		// TODO @ItalyPaleAle: Metrics cardinality
+		log.Warn("The default value for 'metric.http.increasedCardinality' will change to 'false' in Dapr 1.14")
```
JoshVanL, line 267:
```suggestion
		log.Warn("The default value for 'spec.metrics.http.increasedCardinality' will change to 'false' in Dapr 1.14")
```

pkg/diagnostics/http_monitoring.go
```diff
@@ -113,19 +116,30 @@ func (h *httpMetrics) IsEnabled() bool {
 	return h != nil && h.enabled
 }
 
-func (h *httpMetrics) ServerRequestCompleted(ctx context.Context, method string, status string, reqContentSize, resContentSize int64, elapsed float64) {
+func (h *httpMetrics) ServerRequestCompleted(ctx context.Context, method, path, status string, reqContentSize, resContentSize int64, elapsed float64) {
```
JoshVanL, line 119: Consider creating a struct for the passed in parameters as we have a few now.

tests/integration/suite/daprd/metrics/helpers.go
```diff
@@ -0,0 +1,149 @@
...
+	m.grpcConn, err = grpc.DialContext(ctx,
+		m.daprd.GRPCAddress(),
+		grpc.WithTransportCredentials(insecure.NewCredentials()),
+		grpc.WithBlock(),
+	)
+	require.NoError(t, err)
+	t.Cleanup(func() { require.NoError(t, m.grpcConn.Close()) })
+	m.grpcClient = runtimev1pb.NewDaprClient(m.grpcConn)
+
+	t.Logf("Metrics URL: http://localhost:%d/metrics", m.daprd.MetricsPort())
```
JoshVanL, line 88: Consider removing as it's just adding to the log size and isn't particularly useful in test output.

```suggestion
```

tests/integration/suite/daprd/metrics/helpers.go
```diff
@@ -0,0 +1,149 @@
...
+	// Body is closed below but the linter isn't seeing that
+	//nolint:bodyclose
+	res, err := m.httpClient.Do(req)
+	require.NoError(t, err)
+	defer closeBody(res.Body)
+	require.True(t, res.StatusCode >= 200 && res.StatusCode <= 299)
+}
+
+// Drain body before closing
+func closeBody(body io.ReadCloser) error {
```
JoshVanL, line 143:
```suggestion
func closeBody(t *testing.T, body io.ReadCloser) {
```

tests/integration/suite/daprd/metrics/helpers.go
```diff
@@ -0,0 +1,149 @@
...
+	"github.com/dapr/dapr/tests/integration/framework/util"
+)
+
+// Base struct for metrics tests
+type base struct {
+	daprd      *procdaprd.Daprd
+	httpClient *http.Client
+	grpcClient runtimev1pb.DaprClient
+	grpcConn   *grpc.ClientConn
+	place      *placement.Placement
```
JoshVanL, line 44: I am not a fan of having a "base" suite test case as it obfuscates what is being setup/run to the reader. We should be over explicit in tests.

Can we move the base code into its consumed test cases?

tests/integration/suite/daprd/metrics/httpserver_highcardinality.go
```diff
@@ -0,0 +1,97 @@
+/*
```
JoshVanL, line 1: Consider moving both http tests into a `http` directory. This improves test organisation.

tests/integration/suite/daprd/metrics/httpserver_lowcardinality.go
```diff
@@ -0,0 +1,97 @@
...
+	configFile := filepath.Join(t.TempDir(), "config.yaml")
+	require.NoError(t, os.WriteFile(configFile, []byte(`
+apiVersion: dapr.io/v1alpha1
+kind: Configuration
+metadata:
+  name: lowcardinality
+spec:
+  metrics:
+    http:
+      increasedCardinality: false
```
JoshVanL, line 53: Please also have a test for when this value is nil.

tests/integration/suite/daprd/metrics/httpserver_lowcardinality.go
```diff
@@ -0,0 +1,97 @@
...
+		require.NoError(t, err)
+		m.doRequest(t, req)
+
+		// Verify metrics
+		metrics := m.getMetrics(t, ctx)
+		assert.Equal(t, 1, int(metrics["dapr_http_server_request_count|app_id:myapp|method:InvokeService/myapp|status:200"]))
+	})
+
+	t.Run("state stores", func(t *testing.T) {
+		reqCtx, reqCancel := context.WithTimeout(ctx, time.Second)
```
JoshVanL, line 77: Same here re timeout

tests/integration/suite/daprd/metrics/workflow.go
```diff
@@ -0,0 +1,106 @@
...
+			return nil, err
+		}
+		return nil, nil
+	})
+	taskhubClient := client.NewTaskHubGrpcClient(m.grpcConn, backend.DefaultLogger())
+	taskhubCtx, cancelTaskhub := context.WithCancel(ctx)
+	taskhubClient.StartWorkItemListener(taskhubCtx, r)
+	defer cancelTaskhub()
+
+	time.Sleep(5 * time.Second)
```
JoshVanL, line 74: Remove
```suggestion
```

JoshVanL (changes requested):
Some comments from me :)

pkg/apis/configuration/v1alpha1/types.go
ItalyPaleAle: Yes

tests/integration/suite/daprd/metrics/helpers.go
```diff
@@ -0,0 +1,149 @@
...
+
+	var err error
+	m.grpcConn, err = grpc.DialContext(ctx,
+		m.daprd.GRPCAddress(),
+		grpc.WithTransportCredentials(insecure.NewCredentials()),
+		grpc.WithBlock(),
+	)
+	require.NoError(t, err)
+	t.Cleanup(func() { require.NoError(t, m.grpcConn.Close()) })
+	m.grpcClient = runtimev1pb.NewDaprClient(m.grpcConn)
```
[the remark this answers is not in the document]
ItalyPaleAle: I also need grpcconn here

tests/integration/suite/daprd/metrics/helpers.go
ItalyPaleAle: We can but then the code would be replicated multiple times and be harder to maintain

pkg/diagnostics/http_monitoring.go
ItalyPaleAle: This is the same number of arguments that this method had in 1.11, and i don't see them changing (as that would mean adding even more cardinality). If needed, can be done separately.

tests/integration/suite/daprd/metrics/helpers.go
```diff
@@ -0,0 +1,149 @@
...
+
+	t.Logf("Metrics URL: http://localhost:%d/metrics", m.daprd.MetricsPort())
+}
+
+// Returns a subset of metrics scraped from the metrics endpoint
+func (m *base) getMetrics(t *testing.T, ctx context.Context) map[string]float64 {
+	t.Helper()
+
+	reqCtx, reqCancel := context.WithTimeout(ctx, time.Second)
+	defer reqCancel()
```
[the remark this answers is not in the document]
ItalyPaleAle: What do you mean? This is a timeout to make it so the request don't hang forever. I should increase it to 5 seconds however.

tests/integration/suite/daprd/metrics/httpserver_lowcardinality.go
ItalyPaleAle: Added test to `pkg/config/configuration_test.go` since the value will change in the future and the test is duplicate.

tests/integration/suite/daprd/metrics/httpserver_lowcardinality.go
ItalyPaleAle: Done

pkg/apis/configuration/v1alpha1/types.go
ItalyPaleAle: I changed the text to be more explicit

yaron2 (approved):

tests/integration/suite/daprd/metrics/httpserver_defaultcardinality.go
```diff
@@ -0,0 +1,83 @@
...
+}
+
+// httpServerDefaultCardinality tests daprd metrics for the HTTP server configured with the default cardinality
+// TODO @ItalyPaleAle [MetricsCardinality] Change default in 1.14 to be based on low cardinality
+type httpServerDefaultCardinality struct {
+	base
+}
+
+func (m *httpServerDefaultCardinality) Setup(t *testing.T) []framework.Option {
+	return m.testSetup(t)
```
JoshVanL, line 42: Please be explicit for the reader and author who will need to update this test when the default changes.

```suggestion
configFile := filepath.Join(t.TempDir(), "config.yaml")
	require.NoError(t, os.WriteFile(configFile, []byte(`
apiVersion: dapr.io/v1alpha1
kind: Configuration
metadata:
  name: lowcardinality
spec:
  metrics:
    http:
      increasedCardinality: null
`), 0o600))

	return m.testSetup(t, procdaprd.WithConfigs(configFile))
```

JoshVanL (dismissed):

tests/integration/suite/daprd/metrics/httpserver_defaultcardinality.go
ItalyPaleAle: This is a different test then. I thought you wanted to validate the default behavior, which is when there's no Configuration.

JoshVanL: It should be the same behaviour when the field is set to null, which is the default. It (should be) the same test, we are just being explicit for the reader.

artursouza: There are multiple scenarios being discussed here and one does not exclude the other:
a. low cardinality by setting `increasedCardinality` to False.
b. low cardinality by setting `increasedCardinality` to nil.
c. low cardinality by not setting a config.

If someone changes the default behavior, then (a) will continue to pass while (b) and (c) will fail. For (b) and (c) to behave differently, it would be an odd bug (not necessarily impossible).

I am fine to keep this test as-is for now and have a new PR into master to add more tests but proceed with the release.

artursouza:
I am OK to merge this PR but need the MacOS integration tests to pass.

artursouza:
This PR is adding flakiness to integration tests or a bug: https://github.com/dapr/dapr/actions/runs/7716239633?pr=7429

ItalyPaleAle:
> This PR is adding flakiness to integration tests or a bug: https://github.com/dapr/dapr/actions/runs/7716239633?pr=7429

Those are the same errors we have in master

artursouza:
Will fix flakiness in master and re-assess this PR. We need to fix both issues to release anyway.

yaron2 (approved):

PR #7432, opened 2024-01-23
Fix operator when healthz when running without webhooks

JoshVanL:
Operator will currently always fail healthz check and exit error when webhooks are disabled.

PR correctly sets healthz and prevents the concurrency manager from exiting early.


famarting (approved):
thank you!

yaron2 (approved):

artursouza (approved):

PR #7446, opened 2024-01-24
Refactor placement config and client out of internal

artursouza:
# Description

Refactor placement config and client out of internal package to allow control plane service to also connect to placement service.

No behavior change.

## Issue reference

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing

### DIFF
--- a/pkg/actors/config/options.go
+++ b/pkg/actors/config/options.go
@@ -11,7 +11,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package internal
+package config
 
 import (
 	"context"
--- a/pkg/actors/placement/placement_interface.go
+++ b/pkg/actors/placement/placement_interface.go
@@ -11,7 +11,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package internal
+package placement
 
 import (
 	"context"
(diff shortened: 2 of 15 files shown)


pkg/actors/placement/placement_interface.go
```diff
@@ -11,7 +11,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package internal
```
ItalyPaleAle, line 14: Please create a new package for the interfaces, don't move them inside `placement` which is a concrete implementation.

pkg/actors/config/options.go
```diff
@@ -11,7 +11,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package internal
```
ItalyPaleAle, line 14: This file should not be in `config`. Maybe more in a package like `interfaces`, together with the placement interface.

ItalyPaleAle (changes requested):

artursouza:
Putting this on-hold for now. I might not need it.

pkg/actors/placement/placement_interface.go
artursouza: No problem. Will move.

pkg/actors/config/options.go
artursouza: No problem. Will move.

cicoyle:
If I remember correctly we were thinking we needed this for the Scheduler service if Scheduler knew about Placement, but we ended up not needing to go this route, so I'm closing this.

Issue #7449, opened 2024-01-25
workflow engine not started if `/TaskHubSidecarService/GetWorkItems` is not called


famarting:
## In what area(s)?

/area runtime

## What version of Dapr?

> edge: v1.11.3-549-gcc9d51e1-dirty

## Expected Behavior

I'm testing the the upcoming workflow go sdk and I was getting an error like `did not find address for actor dapr.internal.myns.myapp.workflow/a7a4168d-3a1c-41da-8a4f-e7f6d9c718d9` when my app tried to start a workflow. Its expected to not have any errors and my workflow to start successfully.

## Actual Behavior

As I said I was getting an error like  `did not find address for actor dapr.internal.myns.myapp.workflow/a7a4168d-3a1c-41da-8a4f-e7f6d9c718d9`

Digging into my dapr sidecar logs I noticed only two requests were being made in my test:
- first a request to `/TaskHubSidecarService/Hello`
- then a request to `/dapr.proto.runtime.v1.Dapr/StartWorkflowBeta1`

The error I was getting means that the internal actors used for workflow were not actually registered, and looking at the workflow engine code, the internal actors are only registered when the workflow engine is started and I wasn't seeing this log entry https://github.com/dapr/dapr/blob/master/pkg/runtime/wfengine/wfengine.go#L160C17-L160C40

So the root cause must have been that the go sdk was not making a request to `/TaskHubSidecarService/GetWorkItems` and because of that this callback was not being executed https://github.com/dapr/dapr/blob/master/pkg/runtime/wfengine/wfengine.go#L90, resulting in the workflow engine not being initialized...

A simple solution I can propose consists of adding a callback to autoStart the workflow engine to https://github.com/dapr/dapr/blob/master/pkg/runtime/wfengine/component.go#L53 and that callback would be called on every function of `workflowEngineComponent` except for the `Init` function, this way whenever a request to the workflow management APIs it will be guaranteed that the workflow engine is started

## Steps to Reproduce the Problem

Try the workflow example included here https://github.com/dapr/go-sdk/pull/487

## Release Note

RELEASE NOTE:

yaron2:
cc @cgillum @RyanLettieri

cgillum:
Indeed, this is related to the work we did to start the workflow engine only after we've detected that an app has connected to it. This is unfortunately one of the side effects of that design decision. I think your proposed solution sounds like a good one that still maintains the spirit of the original delayed-start design.

[...]

PR #7450, opened 2024-01-25
bugfix: lazy load workload engine

famarting:
# Description

Fixes https://github.com/dapr/dapr/issues/7449

## Issue reference

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing


yaron2:
@famarting please fix the linter issues

famarting:
I was wondering if we should close this PR as after this comment https://github.com/dapr/dapr/issues/7449#issuecomment-1910786156 it became clear that the strategy implemented here won't be helpful for the majority of the usecases

artursouza (changes requested):
Can there be an integration test to reproduce the issue?

cgillum (approved):
I think it's worth taking this fix. Appreciate your contribution!

shivamkm07:
@famarting Did this PR fix the issue for you? i.e. if `GetWorkItems` is not called explicitly then is Engine autostarting when calling the Dapr Workflow APIs? 

Looking into the code, seems this solution won't work. When Dapr APIs are called, currently we already wait for engine to be started here: https://github.com/dapr/dapr/blob/c53fc16211e265cf486701bba9c2e94f50ee000f/pkg/api/universal/workflow.go#L37
and so the component implementation won't be called itself. 

If we want to start the Engine at every Dapr Workflow API call, it needs to be done at API level itself.

But IMO we shouldn't do this. The reason being if `GetWorkItems` is not called i.e. there is no stream listening for workItems, then the workflow won't be executed itself. The DTF Grpc executor will be anyway be blocked waiting for an application to connect (and send the work item). So isn't it better and more cleaner to wait for that at API itself, which is what we are doing rn. We can actually add a log line saying `Waiting for engine to be ready`.. something like that. Thoughts @cgillum ?

yaron2:
> We can actually add a log line saying Waiting for engine to be ready

That log message won't mean anything to users and will hardly be discoverable.

cgillum:
> But IMO we shouldn't do this. The reason being if GetWorkItems is not called i.e. there is no stream listening for workItems, then the workflow won't be executed itself. The DTF Grpc executor will be anyway be blocked waiting for an application to connect (and send the work item). So isn't it better and more cleaner to wait for that at API itself, which is what we are doing rn. We can actually add a log line saying Waiting for engine to be ready.. something like that. Thoughts cgillum ?

I think it's a valid point worth considering. The choice is to either block the API caller until a worker is ready or to accept the message and block its _processing_ until the worker is ready. Either way, these APIs are designed to be asynchronous, which is why I didn't see any harm in accepting the message even without a ready worker. It's one less error condition for the developer to worry about. But at the same time, if we know that the worker is supposed to be running, I can see that it might make it easier to diagnose problems if the client call times out because a worker didn't come online within a reasonable amount of time. Without a timeout, the developer is left wondering why nothing is happening after they try to start a workflow.

I could go either way on this.

famarting:
I'm closing this PR, as it has been mentioned this change does not address the problem I described. 

For me my root problem was that the streamed connection to GetWorkItems was failing on my end for other reasons. So the changes in this PR do not bring any improvements

PR #7455, opened 2024-01-26
Fixes MacOS integration tests

JoshVanL:
Fixes MacOS integration test runner.


tests/integration/suite/actors/healthz/endpoint/noappentities.go
```diff
@@ -70,7 +70,7 @@ func (n *noappentities) Setup(t *testing.T) []framework.Option {
 
 	srvNoHealthz := prochttp.New(t,
 		prochttp.WithHandlerFunc("/dapr/config", func(w http.ResponseWriter, r *http.Request) {
-			w.Write([]byte(`{}`))
+			w.Write([]byte(`{"entities": []}`))
```
ItalyPaleAle, line 73: How does this fix the issue? If `entities` is unset and it's causing issues, maybe this is a bug in Dapr?

JoshVanL: I'm going to open up a follow-up test PR to ensure that an empty object can be parsed by `/dapr/config` successfully- this may be a bug in the config client.

artursouza: Why is this a bug on MacOs only?

ItalyPaleAle: > Why is this a bug on MacOs only?

My question exactly

artursouza (approved):
This has done the trick.

artursouza:
Moving back to draft as Josh is still trying this out.

pkg/actors/actors.go
```diff
@@ -405,7 +405,16 @@ func (a *actorsRuntime) haltAllActors() error {
 func (a *actorsRuntime) deactivateActor(act *actor) error {
 	// This uses a background context as it should be unrelated from the caller's context
 	// Once the decision to deactivate an actor has been made, we must go through with it or we could have an inconsistent state
-	ctx := context.Background()
```
JoshVanL, line 408: This was causing graceful shutdown to timeout

yaron2 (approved):

PR #7472, opened 2024-01-31
Bug fixes and test stabilization to unblock 1.13 release

artursouza:
# Description

* Use "127.0.0.1" instead of "localhost" in tests - we have seen instability in resolving "localhost" to an IP address that could not respond before
* Use of DAPR_HOST_IP to force sidecar to advertise itself as "127.0.0.1" in integration-tests
* Fix IsActorLocal() to also compare port in case of localhost
* Fix reminder rebalance bug (by @elena-kolevska)

## Issue reference

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing


tests/integration/suite/actors/healthz/endpoint/noentities.go
```diff
@@ -125,10 +135,17 @@ func (n *noentities) Run(t *testing.T, ctx context.Context) {
 
 	client := util.HTTPClient(t)
 
-	req, err := http.NewRequestWithContext(ctx, http.MethodPost, fooActorURL(n.daprd), nil)
-	require.NoError(t, err)
-	resp, err := client.Do(req)
-	require.NoError(t, err)
-	assert.Equal(t, http.StatusOK, resp.StatusCode)
-	require.NoError(t, resp.Body.Close())
+	assert.EventuallyWithT(t, func(c *assert.CollectT) {
```
JoshVanL, line 138: We shouldn't _have_ to eventually here as a request should always succeed in creating the actor after daprd reports placement ready

JoshVanL (dismissed):

tests/integration/suite/actors/healthz/endpoint/noentities.go
artursouza: Reverted. It was part of the troubleshooting. Thanks.

artursouza:
Even if all tests pass, let me re-run them a few times to make sure the flakiness is over.

pkg/api/grpc/proxy/handler_test.go
```diff
@@ -833,7 +832,7 @@ func (s *proxyTestSuite) SetupSuite() {
 
 	clientConn, err := grpc.DialContext(
 		context.Background(),
-		strings.Replace(s.proxyListener.Addr().String(), "127.0.0.1", "localhost", 1),
+		s.proxyListener.Addr().String(), // DO NOT USE "localhost" as it does not resolve to loopback in some environments.
```
ItalyPaleAle, line 835: I don't think this will work on IPv6-only networks?

elena-kolevska (approved):

pkg/api/grpc/proxy/handler_test.go
artursouza: I remember these fixing flakiness here. I will run another PR without the localhost changes for us to compare now that we have a green baseline. If ipv6 only networks is the concern (although rare for local networks), I can take this value from an env variable.

artursouza: Let's compare: https://github.com/dapr/dapr/pull/7492

artursouza: As I mentioned in Discord, this is only for tests and not in runtime code.

yaron2 (approved):

artursouza:
I created 2 issues to follow up on this and not block the release since the issues on MacOS runner might be an environment issue but requires further investigation.

https://github.com/dapr/dapr/issues/7501
https://github.com/dapr/dapr/issues/7502

PR #7498, opened 2024-02-05
Workflow delete: Remove `non_recursive` with hardcoded always enabled `recursive`

JoshVanL:
The workflow terminate and purge APIs currently take a `non_recursive` parameter to indicate whether to delete the workflow and its children. This is a double negative and is confusing for users, needing a few more brain cycles to understand the field. We should avoid double negatives in APIs.

Removes this option so that `recursive` is always enabled with no API field exposing the option.


dapr/proto/common/v1/common.proto
```diff
@@ -158,3 +158,15 @@ message ConfigurationItem {
   // the metadata which will be passed to/from configuration store component.
   map<string,string> metadata = 3;
 }
+
+// Boolean represents a boolean value which can be defaulted, true, or false.
+enum Boolean {
```
ItalyPaleAle, line 163: Let's not introduce different booleans please. This is a overly complicated.

ItalyPaleAle (changes requested):

dapr/proto/common/v1/common.proto
mukundansundar: +1

shivamkm07: So if we need to set `recursive` as `true` as the default parameter, we possibly would need a separate boolean. This is because, otherwise e.g. if someone uses Dapr Grpc client for 
[terminating a workflow](https://github.com/dapr/dapr/blob/2b6b301afd86ae643d46fe8e1521bf7c187cfb1c/pkg/api/universal/workflow.go#L111), the default value for recursive would be false ,and so in the default case, it won't be recursive termination. Although it won't be a usual case for someone using Dapr client directly instead of workflow SDKs(SDKs would still need to set
true as default), but still, it won't be then consistent between Dapr HTTP client and GRPC client.

This was the positive for using `non_recursive` as query param as it allows for not requiring another `default` type. I would leave this to the maintainers to decide what's better: `non_recursive` as `false` or having a custom `Boolean` type. Btw choice for `non_recursive` was based on this discussion: https://github.com/dapr/dapr/pull/7340#discussion_r1453865073

JoshVanL: I don't think it is complicated, using enum to represent field values is common and standard for gRPC APIs.

The default value of `recursive` is `true`. This is the case for both gRPC and HTTP clients. A default API value occurs, fundamentally, when the field has been omitted by the client.

---

Another way to solve this problem would be to change the API field to be `deletion_strategy` which accepts an enum, being either `recursive` or `non-recursive`. `recursive` being the default. This keeps the door open for adding other deletion strategies in future which we have not considered.

mikeee: The alternative solution proposed by Josh does highlight how it can be even more clunky. 

I do think the enum type implementation is the best with most people being happy with the default strategy anyway.

yaron2: `deletion_strategy` makes the field's purpose much clearer, where the API leads with what the field is. I support changing it to an enum as well.

ItalyPaleAle: We were talking with @cgillum in a Teams chat this morning. Perhaps in the case of Dapr we should remove this option entirely and leave "recursive" as the only option.

FWIW I'd prefer a "deletion_strategy" enum than a boolean-like enum.

artursouza: I vote for `deletion_strategy` enum or remove the option and run on way only, like @ItalyPaleAle is talking to @cgillum about.

cgillum: I'm fine with either approach, leaving out the option or creating a new "strategy" enum.

yaron2: @JoshVanL as per your suggestion, please change the field to an enum type named deletion_strategy

shivamkm07: If adding a strategy-based API field, please rename it to something other than `deletion_strategy` as we don't `delete` in `Terminate` call. In Purge that makes sense.

DeepanshuA: Considering all these scenarios, removing the option entirely sounds like a best choice here, to me personally.
A separate enum, something like`deletion_strategy` also sounds fine, better than new `Boolean` type.

shivamkm07: `deletion_strategy` would only make sense in purge/delete API call. For Terminate, either a separate enum `termination_strategy` would be required or some other common name.

DeepanshuA: or rather put a generic name `strategy` ?

DeepanshuA: With `strategy`, user can perceive it in appropriate context.
And, we also don't remove the option - if any such use case arises, where-in child and parent wf need to remain independent.

artursouza: We are so close to cut the RC. We need to reach a decision here.

Can we remove the config and run with "always recursive" for 1.13 and we can decide on the param for 1.14?

yaron2: If it's possible to remove the configuration entirely then that's best. If not, then a delete_strategy field should be the way to go

mukundansundar: I am OK to either remove the whole parameter or rename it to `strategy`

JoshVanL: I am fine with removing the option entirely and have `recursive` always set to true. We can come back to this in future if and when it is requested.

PR is updated to move the option.

mukundansundar (approved):

ItalyPaleAle (approved):

PR #7503, opened 2024-02-06
Workflow grpc start API: Generate instanceID if not present

shivamkm07:
# Description

Dapr Grpc API to [start workflow](https://github.com/dapr/dapr/blob/6b7c94fa4104f34cb8cb9e6c8d1f85604d6e480d/pkg/api/universal/workflow.go#L71) errors out if no instanceID is given: https://github.com/dapr/dapr/blob/6b7c94fa4104f34cb8cb9e6c8d1f85604d6e480d/pkg/api/universal/workflow_test.go#L74
However, when using HTTP API, instanceID is generated if empty, as expected. This causes inconsistency between HTTP and Grpc clients. This PR updates Grpc API to have the same behavior as http.

## Issue reference

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing


JoshVanL:
Please mark the instanceID as optional in the proto definition. Are other fields optional?

https://github.com/dapr/dapr/blob/6b7c94fa4104f34cb8cb9e6c8d1f85604d6e480d/dapr/proto/runtime/v1/dapr.proto#L1046

JoshVanL (dismissed):
Please mark the instanceID as optional in the proto definition. Are other fields optional?

https://github.com/dapr/dapr/blob/6b7c94fa4104f34cb8cb9e6c8d1f85604d6e480d/dapr/proto/runtime/v1/dapr.proto#L1046

DeepanshuA:
> Please mark the instanceID as optional in the proto definition. Are other fields optional?
> 
> https://github.com/dapr/dapr/blob/6b7c94fa4104f34cb8cb9e6c8d1f85604d6e480d/dapr/proto/runtime/v1/dapr.proto#L1046

Do we require optional here? 
Default value of string is blank
Do we need to differentiate between user-set `blank` (but present) v/s not-present `blank`? I don't think  we need that differentiation.
 
If instanceId is blank, then we just need to assign a random value to it.
I think we can use here a no-presence semantic, instead of explicit-presence.

JoshVanL:
@DeepanshuA Fields which are optional ( `instancedID` in this case ) which are not required by the client in the API request should always be marked as optional in protobuf. It signals to the user what is required of them to make use of the API. If an empty string is not a valid instance ID I strongly recommend also erroring to follow proper schema validation.

ItalyPaleAle (approved):

PR #7508, opened 2024-02-06
Increase Sentry retry to 30s from Sidecar Injector

yaron2:
Follow up to https://github.com/dapr/dapr/pull/7507

Increases retries to `30s` which is the maximum mutation request allowed duration when the K8s API server calls an admission webhook.

Also moves connection init to `New` to encapsulate the logic in the requester.

### DIFF
--- a/pkg/injector/sentry/sentry.go
+++ b/pkg/injector/sentry/sentry.go
@@ -53,18 +53,20 @@ type Requester struct {
 }
 
 // New returns a new instance of the Requester.
-func New(opts Options) *Requester {
+func New(ctx context.Context, opts Options) (*Requester, error) {
 	_, kubeMode := os.LookupEnv("KUBERNETES_SERVICE_HOST")
-	return &Requester{
+	r := &Requester{
 		sentryAddress:  opts.SentryAddress,
 		sentryID:       opts.SentryID,
 		sec:            opts.Security,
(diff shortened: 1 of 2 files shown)


JoshVanL (approved):

pkg/injector/sentry/sentry.go
```diff
@@ -109,7 +111,7 @@ func (r *Requester) RequestCertificateFromSentry(ctx context.Context, namespace
 			Token:          token,
 			Namespace:      namespace,
 			TokenValidator: tokenValidator,
-		}, grpcRetry.WithMax(3), grpcRetry.WithPerRetryTimeout(time.Second*3))
+		}, grpcRetry.WithMax(10), grpcRetry.WithPerRetryTimeout(time.Second*3))
```
ItalyPaleAle, line 114: Then do we need to set a "WithMax"? Could we just abort on when the context was canceled, which happens automatically?

```suggestion
		}, grpcRetry.WithPerRetryTimeout(2*time.Second))
```

yaron2: WithMax is required to enable retries. Only having `WithPerRetryTimeout` will not enable it. Context cancellation will not retry the request, it will serve as a 30s timeout for a single request.

PR #7511, opened 2024-02-07
Injector: add `DAPR_HOST_IP` env var to daprd

ItalyPaleAle:
The `DAPR_HOST_IP` env var is used in various places in Dapr for a sidecar to know its own IP address, for example for service invocation or actor invocation.

When using the Dapr injector to add the daprd container, we should use the downstream APIs to add the `DAPR_HOST_IP` env var based on data from the controller.


yaron2 (approved):

artursouza (dismissed):
LGTM. Please, change it release-1.13 branch. This way we can avoid cherry-picking.

ItalyPaleAle:
As discussed on Discord, this is now optional and can be enabled using Helm

Rebased onto release-1.13

yaron2 (approved):

artursouza (approved):

pkg/injector/consts/consts.go
```diff
@@ -37,6 +37,7 @@ const (
 	UserContainerAppProtocolName   = "APP_PROTOCOL"                         // Name of the variable exposed to the app containing the app protocol.
 	UserContainerDaprHTTPPortName  = "DAPR_HTTP_PORT"                       // Name of the variable exposed to the app containing the Dapr HTTP port.
 	UserContainerDaprGRPCPortName  = "DAPR_GRPC_PORT"                       // Name of the variable exposed to the app containing the Dapr gRPC port.
+	DaprContainerHostIP            = "DAPR_HOST_IP"                         // Name of the variable injected in the daprd container containing the pod's IP
```
yevgen-el8, line 40: I'm very late to the party :) Why it was named `DAPR_HOST_IP` when it's really DAPR_POD_IP? I was looking for a way to get HOST_IP into the daprd sidecar to configure `OTEL_EXPORTER_OTLP_ENDPOINT` with `HOST_IP` and  this name makes it confusing. Even the comment says that it's a pod's IP.

PR #7514, opened 2024-02-08
Test Integration: Hot Reload `spec.ignoreErrors`

JoshVanL:
Adds integration tests that assert that the `spec.ignoreErrors` field is respected when hot reloading a component. Namely that daprd exits error when `spec.ignoreErrors=false`, and continues running when `spec.ignoreErrors=true`.

No implementation changes.


cicoyle (dismissed):
LGTM. There are some errors in CI for `placement vnodes` and `component already exists`, maybe those are flaky?

artursouza (dismissed):

artursouza (approved):

Issue #7489, opened 2024-02-02
Error Standardization: Metadata API


cicoyle:
Richer errors in Dapr for the **Metadata API**.

**Note**: Please clone this issue and assign the clone of this issue to yourself if you would like to contribute to the error standardization efforts.

# Scope of work

- [ ] Choose 5-10 errors in the Metadata API to enrich. Ideally, cover as many errors as possible from the API.
- [ ] Follow the steps outlined in [this README](https://github.com/dapr/dapr/blob/master/pkg/api/errors/README.md) for each error being updated

# Definition of done

- [ ] Update 5-10 errors in the Metadata API
- [ ] Update the unit tests, if necessary
- [ ] Write integration tests to cover each updated errors

# Keep in mind

Unless a major version is being released and there is a solid reason for introducing backwards incompatibility, we can't change the API, meaning the HTTP and gRPC status codes, and the error codes must stay the same. An exception to this is the gRPC `StatusCode.UNKNOWN` which should be converted to a specific status.

See the [kit errors pkg README](https://github.com/dapr/kit/blob/main/errors/README.md) for another usage example. For each error being updated, you can use the following helper functions to add error details, see the kit [errors pkg](https://github.com/dapr/kit/tree/main/errors) for more details:

- `WithHelpLink()`
-  `WithHelp()`
-  `WithErrorInfo()` - **Required**
-  `WithFieldViolation()`
-  `WithDetails()`

# Reference PRs
Existing error standardization for: 
- [State API](https://github.com/dapr/dapr/pull/7257)
- [PubSub API](https://github.com/dapr/dapr/pull/7322)

Bharadwajshivam28:
I would like to work on this issue can you guide me on how to get started with this to solve the issue

cicoyle:
Hey @Bharadwajshivam28  - sure thing! 🙂

First, I suggest copying this issue into a new one, and assigning that issue to yourself. From there, I would choose the Metadata API errors that you might have seen before. You'll want to create a new `metadata.go` file to construct your errors in, similar to how I made a `pubsub.go` [errors file here in my PR](https://github.com/dapr/dapr/pull/7322/files#diff-e43e0a8c79c816cbeddcbc2a00b07650bbadfedaf118dbe4c8f6beda6ea1654c). Once you have updated a few Metadata API errors using the newly constructed ones, run the unit tests, see if you need to tweak anything, then write some integration tests to force those errors you updated.

[...]

PR #7515, opened 2024-02-08
Enhancing errors for the metadata api

Bharadwajshivam28:
# Description

I am working on modifying the error handling for metadata API. Currently i have added 2 of them and `working on writing more rich error handlers`  Need some suggestions till here is it going correct?

## Issue reference
refs #7489

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing


pkg/api/errors/metdata.go
```diff
@@ -0,0 +1,34 @@
+package errors
+
+import (
+	"fmt"
+    "net/http"
```
cicoyle, line 5: looks like you need to run: `make lint` locally

pkg/api/errors/metdata.go
```diff
@@ -0,0 +1,34 @@
...
+	"google.golang.org/grpc/codes"
+	kiterrors "github.com/dapr/kit/errors"
+)
+
+func IncomingContextMetadataNotFound() error {
```
cicoyle, line 10: We have a generic not found error in the errors.go file in this same directory you can use instead of creating one specific to the Metadata API

pkg/api/errors/metdata.go
```diff
@@ -0,0 +1,34 @@
...
+func MetadataFromIncomingContextFailed() error {
+	message := "failed to retrieve metadata from incoming context"
+	return kiterrors.NewBuilder(
+		grpcCodes.InvalidArgument
+        http.StatusBadRequest,
+		message,
+		"ERR_METADATA_RETRIEVAL_FAILED",
+	).
+	WithErrorInfo("METADATA_RETRIEVAL_FAILED", nil).
+	WithResourceInfo("", "", "", message).
```
cicoyle, line 32: Ideally we are passing in more resource info here, otherwise its really not worth adding since you already added the `message` above

pkg/api/grpc/metadata/metadata.go
```diff
@@ -15,7 +15,7 @@ package metadata
 
 import (
 	"context"
-
+    "erros"
```
cicoyle, line 18: typo?

cicoyle: actually, why was this added? I don't see this import being used

pkg/api/grpc/metadata/metadata.go
```diff
@@ -31,7 +31,7 @@ type ctxKey struct{}
 func FromIncomingContext(ctx context.Context) (MD, bool) {
 	md, ok := ctx.Value(ctxKey{}).(MD)
 	if !ok {
-		return nil, false
```
cicoyle, line 34: Did you check where this func is being called? I don't think we should be changing a return of type `bool` to being an `error`. I don't know if the caller will expect that and it might cause other issues. 

The idea behind the error standardization for the Metadata API was to update `err` variables that are already in our code base to be our standardized error with more relevant error details.

pkg/api/errors/metdata.go
Bharadwajshivam28: Oh... thanks

pkg/api/grpc/metadata/metadata.go
Bharadwajshivam28: > typo?

Oh not saw this

pkg/api/grpc/metadata/metadata.go
Bharadwajshivam28: > Did you check where this func is being called? I don't think we should be changing a return of type `bool` to being an `error`. I don't know if the caller will expect that and it might cause other issues. 
> 
> The idea behind the error standardization for the Metadata API was to update `err` variables that are already in our code base to be our standardized error with more relevant error details. 

Thanks for suggesting I will look in these

cicoyle:
Thanks for the start on this, however I think there is a long ways to go still. Specifically, there may be some confusion on the Metadata API, and what that is. Here is the[ Dapr docs for the Metadata API](https://docs.dapr.io/reference/api/metadata_api/#:~:text=Dapr%20has%20a%20metadata%20API,and%20attributes%20with%20information%20attached.). From there, try to force some Metadata API errors, and update those existing returned `err`s to be the standardized errors, which are defined from the kit builder.

Bharadwajshivam28:
> Thanks for the start on this, however I think there is a long ways to go still. Specifically, there may be some confusion on the Metadata API, and what that is. Here is the[ Dapr docs for the Metadata API](https://docs.dapr.io/reference/api/metadata_api/#:~:text=Dapr%20has%20a%20metadata%20API,and%20attributes%20with%20information%20attached.). From there, try to force some Metadata API errors, and update those existing returned `err`s to be the standardized errors, which are defined from the kit builder. 

I looked into the docs..

I understood that these are the things whichever metadata api will return

It would be great if I get a small example... Although I saw the pr for state and pubsub..

A small example would really help me.. a small and clear one  @cicoyle

Bharadwajshivam28:
Just an example like how one error handler is created for metadata and how it is used @cicoyle

cicoyle:
There are examples of using the API in the docs link I provided. I sent a very clear example or 2 with links to the line in our codebase in our discord conversation as well. My suggestion is to read the docs link I provided, and try to force an error, then trace that in the codebase.

I would also suggest making this a draft PR, since the tests need to be written as well and polishing up the changes.

Bharadwajshivam28:
> There are examples of using the API in the docs link I provided. I sent a very clear example or 2 with links to the line in our codebase in our discord conversation as well. My suggestion is to read the docs link I provided, and try to force an error, then trace that in the codebase.
> 
> I would also suggest making this a draft PR, since the tests need to be written as well and polishing up the changes. 

Thanks..
I will implement things by taking help of resources you shared... 

Ya let me make it draft pr

Bharadwajshivam28:
Hello @cicoyle I referred the PR for PubSub API and State API both and the Metadata api docs also [api docs](https://docs.dapr.io/reference/api/metadata_api/#:~:text=Dapr%20has%20a%20metadata%20API,and%20attributes%20with%20information%20attached.)

I have a few questions:-
1) I saw in the f`iles of pubsub and state` and for `pubsub in the line ```WithErrorInfo(kiterrors.CodePrefixPubSub+PostFixNameEmpty, metadata).` we are using CodePrefixPubSub and same in state file we are using `WithErrorInfo(kiterrors.CodePrefixStateStore+kiterrors.CodeIllegalKey, nil)` we are using `CodePrefixStateStore`
My question is that from where we are getting it? When i searched in the Repo I saw we are using it from here ``kiterrors `"github.com/dapr/kit/errors" `so how can i find these for `Metadata` or can i see how it is created for `Pubsub or State`

2) In the PubSub and State files we are declaring functions and in Pubsub it takes three parameters `func PubSubNameEmpty(name string, pubsubType string, metadata map[string]string)` in most of the functions and for state file  we pass params but it differs in functions`func StateStoreInvalidKeyName(storeName string, key string, msg string)`.

My question is that is there any good way to find what our function will return?

Bharadwajshivam28:
I am currently not knowing that what functions will take as Parameters. 
Will they take parameters what they will return? Like in Metadata api docs Subscriptions list of pub/sub subscriptions which includes pub/sub name, topic, routes, dead letter topic, and the metadata associated with the subscription.

So do we need to take Parameters based on these?

cicoyle:
> Hello @cicoyle I referred the PR for PubSub API and State API both and the Metadata api docs also [api docs](https://docs.dapr.io/reference/api/metadata_api/#:~:text=Dapr%20has%20a%20metadata%20API,and%20attributes%20with%20information%20attached.)
> 
> I have a few questions:-
> 
> 1. I saw in the `files of pubsub and state` and for `pubsub in the line` `WithErrorInfo(kiterrors.CodePrefixPubSub+PostFixNameEmpty, metadata).` we are using `CodePrefixPubSub` and same in state file we are using `WithErrorInfo(kiterrors.CodePrefixStateStore+kiterrors.CodeIllegalKey, nil)` we are using `CodePrefixStateStore`.
> 
> My question is that from where we are getting it? When i searched in the Repo I saw we are using it from here `kiterrors github.com/dapr/kit/errors` `so how can i find these for `Metadata`or can i see how it is created for`Pubsub or State`
> 
> 2. In the PubSub and State files we are declaring functions and in `Pubsub` it takes three parameters `func PubSubNameEmpty(name string, pubsubType string, metadata map[string]string)` in most of the functions and for state file we pass params but it differs in functions`func StateStoreInvalidKeyName(storeName string, key string, msg string)`.
> 
> My question is that is there any good way to find what our function will return?
> 
> Thanks.

The `CodePrefixPubSub` comes from the kit pkg, [specifically this file](https://github.com/dapr/kit/blob/main/errors/codes.go). You posted the import above, the `kiterrors.CodePrefixPubSub` is referencing the pkg from the import, which you can find on github. Are you familiar with Go and how to use imports? That might be worth looking into if you are not familiar. You should have seen several links to the kit errors pkg, as I know I linked to it from the issue. Did you by chance review all the links provided?

I recommend reading the code there to understand better how to use the pkg when constructing the error with the builder pattern. If you are using an IDE, you should be able to `ctrl+click` and it should take you to where variables are defined.

The function always returns the error type. If you are asking about the parameters the function accepts, that depends on the API and what makes sense for that specific error you are standardizing.

PR #7516, opened 2024-02-09
[1.12] Pin contrib 1.12.9, add release notes, fix CI

berndverst:
# Description

Release 1.12: Pin contrib 1.12.9

Fixes an issue where Event Hubs Bindings and PubSub subscriptions encounter an error and never recover - causing the sidecar to silently end the subscriptions with no way to recover.

Also cherry-picks ed125cc from @artursouza to fix standalone validation workflow.

### DIFF
--- a/.github/workflows/dapr-standalone-validation.yml
+++ b/.github/workflows/dapr-standalone-validation.yml
@@ -47,26 +47,29 @@ jobs:
       - name: Check out code
         uses: actions/checkout@v4
         with:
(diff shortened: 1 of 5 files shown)


JoshVanL (changes requested):
Thanks @berndverst, please can you add release notes for these changes for 1.12.5?

JoshVanL (approved):

yaron2 (approved):

PR #7518, opened 2024-02-09
Healthz overhaul

JoshVanL:
Currently, the healthz endpoint on Dapr services do not represent the actual health of the service as they are mostly only spinning up a HTTP server and immediately returning a 200 OK. This results in services receiving traffic before they are ready to actually serve requests. This issue manifests in integration tests where, for example, no metrics are available on the metrics endpoint even though the service reports as healthy and therefore fails the test.

This PR overhauls the healthz package to allow registering health targets during the service `New` process chain. Each module marks their target as healthy and the healthz server reports the overall health of the service. Although not perfect, service healthz endpoints should now be more representative of the actual health of the program.


yaron2:
@JoshVanL please resolve conflicts

JoshVanL:
@yaron2 fixed

pkg/placement/tables.go
```diff
@@ -27,7 +29,7 @@ type HostInfo struct {
 }
 
 // GetPlacementTables returns the current placement host infos.
-func (p *Service) GetPlacementTables() (*PlacementTables, error) {
+func (p *Service) GetPlacementTables() ([]byte, error) {
```
yaron2, line 32: What's the purpose of this change in relation to the PR?

pkg/operator/operator.go
```diff
@@ -253,42 +250,28 @@ func (o *operator) Run(ctx context.Context) error {
 			if rErr != nil {
 				return rErr
 			}
-			healthzServer.Ready()
+			o.tshealthz[0].Ready()
```
yaron2, line 253: Ideally we can have a more explicit reference to the server instead of using an index

pkg/placement/tables.go
JoshVanL: This is because the healthz server extra endpoints option has been generalized to now just accept a `([]byte, error)` "`Getter`". I think this is cleaner/easier to consume. This function was only ever used for that anyway, so makes sense to json marshal in-func also.

pkg/operator/operator.go
JoshVanL: I agree- I'll make as named vars.

JoshVanL: Done!

msfussell:
This will need a doc PR.

artursouza (approved):

Issue #6027, opened 2023-03-06
Dapr state store encryption should switch to AES-CBC


ItalyPaleAle:
## In what area(s)?

/area runtime

## Details

Currently, Dapr state store encryption uses AES in GCM mode, with a fixed key (i.e. the same key is used to encrypt all messages), and a randomly-generated nonce.

Per the requirements set by the NIST, in a scenario like ours we should not be encrypting more than 2^32 values with the same key, as the likelihood of a nonce reuse becomes too high after that (a GCM nonce is just 96 bits).

Recommendation from experts we consulted with is to switch to a different cipher, such as AES in CBC mode.

- The cipher is AES-CBC with PKCS#7 padding (support for that is [being added in dapr/kit already](https://github.com/dapr/kit/pull/38)). We are also going to use HMAC-SHA256 for authenticating the message.
- To maintain backwards-compatibility, we will include a "version" field in the ciphertext that also indicates the ciphertext, the ID of the key used, and other information such as the IV and authentication tag
- This is a good opportunity to also include the name of the key of the key-value (the "row name", not the crypto key) as AAD and fix the issue described in #5998

### Alternatives

Using AES-CBC is the only NIST-approved option.

If we want to, we can offer as an alternative ChaCha20-Poly1305 ("ChaPoly"), for example as a config flag (this is being added to dapr/kit too). Although not NIST-approved, it's widely used (it's part of the TLS 1.3 standard) and it can be faster than AES when hardware acceleration is not present (e.g. Raspberry Pi). ChaPoly is a stream cipher so it doesn't require padding.  
When we include the name of the cipher used in the ciphertext (as described above), we can allow Dapr to decrypt everything even though the config flag sets the format Dapr uses to encrypt data.

### Why can't we just have users rotate the keys?

An alternative that doesn't require code changes is to instruct users to rotate their keys at most every 2^32 operations. Although this would address the issue, it's not ideal:

1. Dapr supports a primary and secondary key to allow key rotation. Only the primary key is used to encrypt data, and the secondary key is only used to decrypt data encrypted with the old key. When the key is rotated, data encrypted with the old key is not automatically re-encrypted. Which means that users can only rotate a key once, so **they can only perform 2^33 operations in total** with Dapr to be safe (they can manually re-encrypt data with the new key, but that still counts towards the 2^32 limit of operations per key)
2. Rotating keys requires users to track the number of times a key was used, or at least track how many operations per day are performed to get the approximated number of days.
[...]

PR #7524, opened 2024-02-09
[1.13] Add warning that Dapr state store encryption could lead to catastrophic failures

ItalyPaleAle:
If the same key is used to encrypt more than 2^32 values (ie. more than 2^32 "Save" operations), it can lead to the private keys being exposed, breaking the entire encryption.

See: #6027

Until the issue above is fixed, we should inform our users of the risks

### DIFF
--- a/pkg/runtime/processor/state/state.go
+++ b/pkg/runtime/processor/state/state.go
@@ -86,8 +86,8 @@ func (s *state) Init(ctx context.Context, comp compapi.Component) error {
 	secretStoreName := s.meta.AuthSecretStoreOrDefault(&comp)
 
 	secretStore, _ := s.compStore.GetSecretStore(secretStoreName)
-	encKeys, encErr := encryption.ComponentEncryptionKey(comp, secretStore)
-	if encErr != nil {
+	encKeys, err := encryption.ComponentEncryptionKey(comp, secretStore)
+	if err != nil {
 		diag.DefaultMonitoring.ComponentInitFailed(comp.Spec.Type, "creation", comp.ObjectMeta.Name)
 		return rterrors.NewInit(rterrors.CreateComponentFailure, fName, err)
 	}
@@ -95,7 +95,8 @@ func (s *state) Init(ctx context.Context, comp compapi.Component) error {
 	if encKeys.Primary.Key != "" {
 		ok := encryption.AddEncryptedStateStore(comp.ObjectMeta.Name, encKeys)
 		if ok {
-			log.Infof("automatic encryption enabled for state store %s", comp.ObjectMeta.Name)
+			log.Infof("Automatic encryption enabled for state store %s", comp.ObjectMeta.Name)
+			log.Warnf("WARNING: ⚠️ Automatic state store encryption should never be used to encrypt more than 2^32 values. Encrypting more than that can cause the private key to be exposed.")
 		}
 	}
 
--- a/pkg/runtime/processor/state/state.go
+++ b/pkg/runtime/processor/state/state.go
@@ -96,7 +96,7 @@ func (s *state) Init(ctx context.Context, comp compapi.Component) error {
 		ok := encryption.AddEncryptedStateStore(comp.ObjectMeta.Name, encKeys)
 		if ok {
 			log.Infof("Automatic encryption enabled for state store %s", comp.ObjectMeta.Name)
-			log.Warnf("WARNING: ⚠️ Automatic state store encryption should never be used to encrypt more than 2^32 values. Encrypting more than that can cause the private key to be exposed.")
+			log.Info("WARNING: Automatic state store encryption should never be used to store more than 4 billion items in the state store (including updates). Storing more items than that can cause the private key to be exposed.")
 		}
 	}
 


pkg/runtime/processor/state/state.go
```diff
@@ -86,16 +86,17 @@ func (s *state) Init(ctx context.Context, comp compapi.Component) error {
...
 	}
 
 	if encKeys.Primary.Key != "" {
 		ok := encryption.AddEncryptedStateStore(comp.ObjectMeta.Name, encKeys)
 		if ok {
-			log.Infof("automatic encryption enabled for state store %s", comp.ObjectMeta.Name)
+			log.Infof("Automatic encryption enabled for state store %s", comp.ObjectMeta.Name)
+			log.Warnf("WARNING: ⚠️ Automatic state store encryption should never be used to encrypt more than 2^32 values. Encrypting more than that can cause the private key to be exposed.")
 		}
 	}
```
tmacam, line 101: How about:

```suggestion
	if encKeys.Primary.Key != "" {
		ok := encryption.AddEncryptedStateStore(comp.ObjectMeta.Name, encKeys)
		if ok {
			log.Infof("Automatic encryption enabled for state store %s", comp.ObjectMeta.Name)
			log.Warnf("WARNING: ⚠️ Automatic state store encryption should never be used to encrypt more than 2^32 values. Encrypting more than that can cause the private key to be exposed.")
		}
	} else {
		log.Infof("Automatic encryption disabled for state store %s", comp.ObjectMeta.Name)
	}
```

tmacam (approved):

pkg/runtime/processor/state/state.go
JoshVanL: I don't think it is _super_ necessary to have this as a log warning over putting it in the docs as it could seem a bit spooky for users even when doing the right thing.

If we do want to log a message it should be at the INFO level, and we should use a human number to put 2^32 in context (`over 4 billion values`). We should also state that the key can be rotated.

JoshVanL (changes requested):
Comment from me :) 

Keen for us to address https://github.com/dapr/dapr/issues/6027 soonish. I'd vote for ChaCha as default.

pkg/runtime/processor/state/state.go
yaron2: > I don't think it is _super_ necessary to have this as a log warning over putting it in the docs as it could seem a bit spooky for users even when doing the right thing.
> 
> If we do want to log a message it should be at the INFO level, and we should use a human number to put 2^32 in context (`over 4 billion values`). We should also state that the key can be rotated.

Agree with proposed changes

berndverst: I also agree:
Info level logging, and using the human friendly "4 billion". But in the docs, let's actually make it a **warning** not just some note / info thing.

ItalyPaleAle: >  it could seem a bit spooky for users even when doing the right thing.

Note that 2^32 is a number the NIST sets as a general guideline. The problem is that the nonce is random and 96-bit, and you must never use the same nonce twice with the same key. Given the birthday problem, the NIST stated that after 2^32 the _likelihood_ of nonce reuse (with random nonces) is too high. However, a collision could happen after 2^80 or even just 2^1.

> Info level logging, and using the human friendly "4 billion"

I think warning is important, but agree "4 billion" is better

ItalyPaleAle:
>  I'd vote for ChaCha as default.

We should use FIPS-compliant crypto as default, and ChaPoly is not FIPS-compliant. Sure it'd be nice to have as an option.

pkg/runtime/processor/state/state.go
yaron2: Please, change to Info, change the text to 4 billion and remove the warning sign emoji. Other than that LGTM and we can merge

ItalyPaleAle: Done

JoshVanL:
@ItalyPaleAle from my side I think we should make dapr the most secure by default using the best tech in the biz. Using the best tools for the job and abstracting the burden/work of choosing frameworks/boilerplate/algos away is the core value prop of dapr after all.

Whilst I appreciate the good research FIPS does and the fact that businesses are compelled to use it to adhere to regulation, FIPS specifications are inherently slow to release and therefore prevent users using more secure software sooner. I have previously shipped software that was patched to _remove_ support for larger RSA bit sizes specifically because FIPS didn't technically allow it in its specification.

It's of my opinion that we should default to the best, and give the user the tools to use different algos which adhere to their regulatory requirements as and when needed.

ItalyPaleAle:
A lot of US users have requirements that software must be FIPS-compliant. For companies including banks, manufacturers, government, etc, non-FIPS compliant is a non-starter.

I'm happy to support ChaPoly and other non-FIPS crypto as an opt-in capability.

It is also not true that ChaPoly is safer: it offers 256-bit security like AES-GCM. The advantage of ChaPoly is that when hardware crypto isn't available, it's faster (however when hardware crypto is available, AES is almost always faster).

In any case, neither AES-GCM nor ChaPoly are suitable for encrypting data as-is, as they are both vulnerable to nonce reuse attacks (and both use 96-bit nonces). The right solution will involve some sort of envelope encryption, where the data is encrypted (with a stream cipher like AES-GCM or ChaPoly) and the key is random.

yaron2 (approved):

Issue #7413, opened 2024-01-19
Some pubsub components can emit conflicting `content-length` header


artursouza:
## In what area(s)?

> /area runtime

> /area operator

> /area placement

> /area docs

> /area test-and-release

## What version of Dapr?

> 1.1.x
> 1.0.x
> edge: output of `git describe --dirty`

1.11.1 but it might be a day 0 issue.

## Expected Behavior

PubSub messages can be delivered even when the publisher adds a `content-length` metadata.

Although I found Pulsar to be the one causing this, it can happen with other pubsub components and the fix can be centralized in runtime.

## Actual Behavior

Message cannot be delivered to a gRPC service in Java (does not require Dapr's Java SDK and can impact other gRPC implementations too).

```txt
== APP == WARNING: Stream Error
== APP == io.grpc.netty.shaded.io.netty.handler.codec.http2.Http2Exception$StreamException: Received amount of data 446 does not match content-length header 777
== APP == 	at io.grpc.netty.shaded.io.netty.handler.codec.http2.Http2Exception.streamError(Http2Exception.java:153)
== APP == 	at io.grpc.netty.shaded.io.netty.handler.codec.http2.DefaultHttp2ConnectionDecoder$ContentLength.increaseReceivedBytes(DefaultHttp2ConnectionDecoder.java:806)
== APP == 	at io.grpc.netty.shaded.io.netty.handler.codec.http2.DefaultHttp2ConnectionDecoder.verifyContentLength(DefaultHttp2ConnectionDecoder.java:230)
== APP == 	at io.grpc.netty.shaded.io.netty.handler.codec.http2.DefaultHttp2ConnectionDecoder.access$300(DefaultHttp2ConnectionDecoder.java:53)
== APP == 	at io.grpc.netty.shaded.io.netty.handler.codec.http2.DefaultHttp2ConnectionDecoder$FrameReadListener.onDataRead(DefaultHttp2ConnectionDecoder.java:303)
== APP == 	at io.grpc.netty.shaded.io.netty.handler.codec.http2.Http2InboundFrameLogger$1.onDataRead(Http2InboundFrameLogger.java:48)
== APP == 	at io.grpc.netty.shaded.io.netty.handler.codec.http2.DefaultHttp2FrameReader.readDataFrame(DefaultHttp2FrameReader.java:415)
[...]

PR #7421, opened 2024-01-20
Fix pubsub metadata conflict

artursouza:
# Description

Fix issue where pubsub event metadata can conflict with gRPC metadata and make the message undeliverable.

## Issue reference

Please reference the issue this PR will close: https://github.com/dapr/dapr/issues/7413

## Checklist

Please make sure you've  completed the relevant tasks for this PR, out of the following list:

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing
* [ ] End-to-end tests passing
* [ ] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_
* [ ] Specification has been updated / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_
* [ ] Provided sample for the feature / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_

PR #7525, opened 2024-02-09
[Cherry-pick] Fix pubsub metadata conflict (1.12)

artursouza:
# Description

Cherry picks fix for pubsub metadata conflict (#7421)

## Issue reference

Please reference the issue this PR will close: https://github.com/dapr/dapr/issues/7413

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing

### DIFF
--- a/pkg/messaging/v1/util.go
+++ b/pkg/messaging/v1/util.go
@@ -50,12 +50,12 @@ const (
 
 	// ContentTypeHeader is the header key of content-type.
 	ContentTypeHeader = "content-type"
+	// ContentLengthHeader is the header key of content-length.
+	ContentLengthHeader = "content-length"
 	// DaprHeaderPrefix is the prefix if metadata is defined by non user-defined http headers.
 	DaprHeaderPrefix = "dapr-"
 	// gRPCBinaryMetadata is the suffix of grpc metadata binary value.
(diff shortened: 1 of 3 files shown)


pkg/messaging/v1/util.go
```diff
@@ -516,6 +518,13 @@ func ProtobufToJSON(message protoreflect.ProtoMessage) ([]byte, error) {
 // WithCustomGRPCMetadata applies a metadata map to the outgoing context metadata.
 func WithCustomGRPCMetadata(ctx context.Context, md map[string]string) context.Context {
 	for k, v := range md {
+		if strings.EqualFold(k, ContentTypeHeader) ||
```
ItalyPaleAle, line 521: Nit, but here should have used `==` and not `EqualFold`, and mute the linter. `EqualFold` can be a lot slower as it's unicode-aware, and these strings are never Unicode...

ItalyPaleAle (approved):

pkg/messaging/v1/util.go
artursouza: == will not work because it does not ignorecase. Unless we use Lower() the string. So, is it better to use EqualFold() or == with Lower()

Issue #7523, opened 2024-02-09
Regression: cron and sql components are not loading in quickstarts starting 1.13


paulyuk:
## Expected Behavior

Just like in 1.12.0 release, I see components for cron and sql bindings load

```bash
dapr init --runtime-version=1.12.0
...
Running shell 'bash -c' with command: `dapr run --app-id batch-sdk --app-port 5005 --dapr-http-port 3500 --resources-path ../../../components -- node index.js`
...
                ℹ️  Starting Dapr with id batch-sdk. HTTP Port: 3500. gRPC Port: 59130
                time="2024-02-09T08:53:31.808539-08:00" level=info msg="Starting Dapr Runtime -- version 1.12.0 -- commit 2149fca96cdf11627c387bda26dcc027d1c47354" app_id=batch-sdk instance=Pauls-MacBook-Pro-2.local scope=dapr.runtime type=log ver=1.12.0
                time="2024-02-09T08:53:31.808562-08:00" level=info msg="Log level set to: info" app_id=batch-sdk instance=Pauls-MacBook-Pro-2.local scope=dapr.runtime type=log ver=1.12.0
                time="2024-02-09T08:53:31.809095-08:00" level=warning msg="mTLS is disabled. Skipping certificate request and tls validation" app_id=batch-sdk instance=Pauls-MacBook-Pro-2.local scope=dapr.runtime.security type=log ver=1.12.0
                time="2024-02-09T08:53:31.813961-08:00" level=info msg="standalone mode configured" app_id=batch-sdk instance=Pauls-MacBook-Pro-2.local scope=dapr.runtime type=log ver=1.12.0
                time="2024-02-09T08:53:31.814418-08:00" level=info msg="app id: batch-sdk" app_id=batch-sdk instance=Pauls-MacBook-Pro-2.local scope=dapr.runtime type=log ver=1.12.0
                time="2024-02-09T08:53:31.814273-08:00" level=info msg="metrics server started on :59131/" app_id=batch-sdk instance=Pauls-MacBook-Pro-2.local scope=dapr.runtime type=log ver=1.12.0
                time="2024-02-09T08:53:31.814673-08:00" level=info msg="Dapr trace sampler initialized: DaprTraceSampler(P=1.000000)" app_id=batch-sdk instance=Pauls-MacBook-Pro-2.local scope=dapr.runtime type=log ver=1.12.0
[...]

PR #7527, opened 2024-02-12
Revert selfhosted disk loader to not respect namespace

JoshVanL:
Revert the selfhosted component disk loader to not respect the namespace. This will allow the selfhosted disk loader to load components from any namespace from file, regardless of what the namespace of daprd actually is. This is to maintain previous behavior.

Ensures namespace is ignored in selfhosted mode on both boot and hot reloading.

Fixes https://github.com/dapr/dapr/issues/7523


artursouza (changes requested):
The problem is for the case when NAMESPACE env is not set. If NAMESPACE env is set, it should still be respected.

yaron2:
> Ensures namespace is ignored in selfhosted mode on both boot and hot reloading when NAMESPACE set

Just to be clear based on the Discord discussion - if NAMESPACE is set and a component has a namespace explicitly set, Dapr will not ignore NAMESPACE and will only load the component if the two namespaces match - is that correct?

JoshVanL:
@yaron2 Correct, if will only load components which either have no namespace specified or when the namespace matches. This is covered in https://github.com/dapr/dapr/pull/7527/files#diff-a320eeb99f18d56db46fd7af1778028ccdc5e2e3a721e709f9f5e01a477b49a3R130

yaron2 (approved):

PR #7534, opened 2024-02-13
Test Integration: add basic and app restart actor timer tests

JoshVanL:
Adds integration tests for basic timer functionality, as well as
ensuring a timer will not continue to be triggered after the application
has been marked as unhealthy.

Adds the following environment variables in order to speed up the actor
app being marked as unhealthy. This speeds the test to 4s from 20s
without the changes.

These environment variables will be removed as and when the actor health
check is merged into the app health check.

```
DAPR_EXPERIMENTAL_ACTOR_HEALTH_FAILURE_THRESHOLD
DAPR_EXPERIMENTAL_ACTOR_HEALTH_UNHEALTHY_INTERVAL
DAPR_EXPERIMENTAL_ACTOR_HEALTH_HEALTHY_INTERVAL
DAPR_EXPERIMENTAL_ACTOR_HEALTH_REQUEST_TIMEOUT
```


pkg/actors/actors.go
```diff
@@ -63,6 +64,11 @@ const (
 
 	errStateStoreNotFound      = "actors: state store does not exist or incorrectly configured"
 	errStateStoreNotConfigured = `actors: state store does not exist or incorrectly configured. Have you set the property '{"name": "actorStateStore", "value": "true"}' in your state store component file?`
+
+	envKeyHealthFailureThreshold = "DAPR_EXPERIMENTAL_ACTOR_HEALTH_FAILURE_THRESHOLD"
```
artursouza, line 68: Have we agreed to the "DAPR_EXPERIMENTAL" as the official way to introduce env variables?

I don't have a problem with it. Just checking we had done the vetting on this.

JoshVanL: No we haven't, and I'd definitely like feedback for it. It _seems_ like the right thing to do, but there may already exist a mechanism in Dapr I'm not aware of.

Ideally, we would merge the actor and app health checks into the same routine which would drop the need for this at all.

The main reason for this to exist is for integration tests not to take an unreasonable amount of time, though real users could see benefit from it in theory.

PR #7536, opened 2024-02-13
Reruns Dapr builds a few times, if needed.

artursouza:
# Description

Reruns the build workflow a few times if needed, not against pull requests.

## Issue reference

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing

### DIFF
--- a/.github/workflows/dapr.yml
+++ b/.github/workflows/dapr.yml
@@ -696,5 +696,4 @@ jobs:
       - env:
           GH_REPO: ${{ github.repository }}
           GH_TOKEN: ${{ github.token }}
-          GH_DEBUG: api
         run: gh workflow run rerun.yml -F run_id=${{ github.run_id }}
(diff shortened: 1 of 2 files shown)


.github/workflows/dapr.yml
```diff
@@ -686,3 +686,13 @@ jobs:
...
+  rerun:
+    name: Reruns in case of failure
+    needs: helmpublish
+    if: failure() && fromJSON(github.run_attempt) < 3 && github.event_name != 'pull_request'
+    runs-on: ubuntu-latest
+    steps:
+      - env:
+          GH_REPO: ${{ github.repository }}
+          GH_TOKEN: ${{ github.token }}
+        run: gh workflow run rerun.yml -F run_id=${{ github.run_id }}
```
ItalyPaleAle, line 698: Rather than doing this, which IMHO adds complexity and it's hard to follow, let's just add autmatic retries to tests when not building a PR (just one CLI flag passed to gotestsum). Although I think this should only be done for releases.

ItalyPaleAle (changes requested):

artursouza:
Keeping it as a draft to collect feedback on this change.

artursouza:
Closing the PR. We can always get back to this if we decide it is something we want to proceed with.

PR #7537, opened 2024-02-14
Add content-length to http channel

yaron2:
For 1.13 we removed the content-length metadata from gRPC calls based on grpc-go's recommendation. The content-length header was also removed from HTTP calls between Dapr and the app, and that causes some webservers to abruptly close connections. This PR returns the content-length header to HTTP calls which is essential.

### DIFF
--- a/tests/e2e/pubsub/pubsub_test.go
+++ b/tests/e2e/pubsub/pubsub_test.go
@@ -278,12 +278,7 @@ func testPublish(t *testing.T, publisherExternalURL string, protocol string) rec
 	require.NoError(t, err)
 	offset += numberOfMessagesToPublish + 1
 
-	// Test bug where content-length metadata conflict makes message undeliverable in grpc subscriber.
-	// We set an arbitrarily large number that it is unlikely to match the size of the payload daprd delivers.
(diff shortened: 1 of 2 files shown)


mukundansundar (commented):
Can we add IT confirming that the right value is being set for the content length as we did remove in gRPC because of incorrect value.

ItalyPaleAle (approved):

tests/e2e/pubsub/pubsub_test.go
```diff
@@ -278,12 +278,7 @@ func testPublish(t *testing.T, publisherExternalURL string, protocol string) rec
 	require.NoError(t, err)
 	offset += numberOfMessagesToPublish + 1
 
-	// Test bug where content-length metadata conflict makes message undeliverable in grpc subscriber.
-	// We set an arbitrarily large number that it is unlikely to match the size of the payload daprd delivers.
-	metadataContentLengthConflict := map[string]string{
```
artursouza, line 283: It means that someone cannot set this metadata - which is fine for me.

artursouza (approved):

PR #7541, opened 2024-02-15
Test Integration: change instances of `127.0.0.1` to `localhost`

JoshVanL:


artursouza:
That is a nice try. Although my experience shows that MacOS runs get flakier with "localhost", I never spent the time to dive deep into the why. You are welcome to explore this if you are interested.

JoshVanL:
@artursouza indeed, this was a bit of a scratch pad for me to investigate further what is going on. I haven't found what it is, and haven't had time to delve any deeper.

PR #7546, opened 2024-02-17
Standardize max-body-size and read-buffer-size flags

ItalyPaleAle:
Related to #7028 (fixes part of the issue)

Adds new CLI flags (and corresponding annotations) for daprd:

- `max-body-size` accepts a resource quantity for the max request body, for the HTTP and gRPC servers, and for the app channel responses
   - The mis-named `dapr-http-max-request-size` (mis-named because it doesn't only apply to HTTP, and it doesn't only apply to requests, but also to responses in the app channel) remains available and is used if `max-body-size` isn't present (and interpreted as MBs like it is today)
- `read-buffer-size` accepts a resource quantity for the read buffer size (also used for reading HTTP headers)
   - The old `dapr-http-read-buffer-size` remains available and is used if `read-buffer-size` isn't present

**Resource quantity** uses the same library that K8s uses for parsing resources (such as [memory limits](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#meaning-of-memory)). It can accept unit-less values, which are interpreted as bytes, or values with suffixes such as "Ki" (kilobytes, binary), "Mi" (megabytes, binary), etc

Corresponding annotations have been added for K8s:

- `dapr.io/max-body-size`
- `dapr.io/read-buffer-size`

The injector adds the CLI flags only if the annotations are explicitly present.

These features are covered in existing tests, including E2E ones. Added unit tests for the handling of the CLI flags.


cmd/daprd/options/options_test.go
```diff
@@ -21,112 +21,304 @@ import (
...
+		opts, err := New([]string{
 			"-config=cfg1.yaml", // Single dash
 			"--config", "cfg2.yaml,cfg3.yaml",
 		})
+		require.NoError(t, err)
 		require.Equal(t, []string{"cfg1.yaml", "cfg2.yaml", "cfg3.yaml"}, opts.Config)
 	})
 }
 
+func TestMaxBodySize(t *testing.T) {
```
daixiang0, line 129: Please add upward out of bounds error.

pkg/api/grpc/manager/manager.go
```diff
@@ -49,13 +49,13 @@ type ConnCreatorFn = func() (grpc.ClientConnInterface, error)
...
-	EnableTLS            bool
-	MaxRequestBodySizeMB int
-	ReadBufferSizeKB     int
-	BaseAddress          string
+	Port               int
+	MaxConcurrency     int
+	TracingSpec        config.TracingSpec
+	EnableTLS          bool
+	MaxRequestBodySize int // In bytes
+	ReadBufferSize     int
```
daixiang0, line 57: Please add comment "// In bytes" for all or not, keep same style.

pkg/runtime/channels/channels.go
```diff
@@ -61,10 +61,10 @@ type Options struct {
 	// AppMiddlware is the application middleware.
 	AppMiddleware middleware.HTTP
 
-	// MaxRequestBodySize is the maximum request body size.
+	// MaxRequestBodySize is the maximum request body size, in bytes.
```
daixiang0, line 64:
```suggestion
	// MaxRequestBodySize is the maximum request body size in bytes.
```

pkg/runtime/channels/channels.go
```diff
@@ -61,10 +61,10 @@ type Options struct {
...
 	MaxRequestBodySize int
 
-	// ReadBufferSize is the read buffer size.
+	// ReadBufferSize is the read buffer size, in bytes
```
daixiang0, line 67:
```suggestion
	// ReadBufferSize is the read buffer size in bytes.
```

pkg/runtime/config.go
```diff
@@ -101,7 +103,7 @@ type Config struct {
 	Mode                         string
 	Config                       []string
 	UnixDomainSocket             string
-	DaprHTTPReadBufferSize       int
+	ReadBufferSize               int
```
daixiang0, line 106: Same as above.

pkg/api/grpc/manager/manager.go
ItalyPaleAle: Thanks

cmd/daprd/options/options_test.go
ItalyPaleAle: Since our GH Actions runner are 64-bit, it is not possible to write a test like that, because the constant `buildinfo.PtrSize` cannot be 32

pkg/runtime/config.go
ItalyPaleAle: Done

daixiang0 (approved):
LGTM

PR #7547, opened 2024-02-19
Revert ApiLevel controlling vnodes back to context metadata

elena-kolevska:
# Description

This PR removes ApiLevel as the factor for calculating vnodes on daprd or placement. This si now controlled on a per-connection basis, using context metadata.

## Checklist

* [X] Code compiles correctly
* [X] Unit tests passing
* [X] Integration tests passing


pkg/actors/placement/client.go
```diff
@@ -17,6 +17,10 @@ import (
 	"context"
 	"sync"
 
+	"google.golang.org/grpc/metadata"
+
+	"github.com/dapr/dapr/pkg/placement"
+
 	"google.golang.org/grpc"
```
JoshVanL, line 24: nit: order imports

pkg/actors/placement/client_test.go
```diff
@@ -19,6 +19,10 @@ import (
 	"sync"
 	"testing"
 
+	"google.golang.org/grpc/metadata"
+
+	"github.com/dapr/dapr/pkg/placement"
+
```
JoshVanL, line 25: Imports

pkg/placement/membership.go
```diff
@@ -335,16 +339,37 @@ func (p *Service) performTableDissemination(ctx context.Context) error {
...
 	copy(streamConnPool, p.streamConnPool)
+
+	hasOldDaprds := false
+	hasNewDaprds := false
+	for _, stream := range streamConnPool {
+		if !hostExpectsVNodes(stream) {
+			hasNewDaprds = true
+		} else {
+			hasOldDaprds = true
+		}
```
JoshVanL, line 361: Can we break early if both are true?

pkg/placement/membership.go
```diff
@@ -362,7 +387,7 @@ func (p *Service) performTableDissemination(ctx context.Context) error {
 // It first locks so no further dapr can be taken it. Once placement table is locked
 // in runtime, it proceeds to update new table to Dapr runtimes and then unlock
 // once all runtimes have been updated.
-func (p *Service) performTablesUpdate(ctx context.Context, hosts []placementGRPCStream, newTable *v1pb.PlacementTables) error {
+func (p *Service) performTablesUpdate(ctx context.Context, hosts []placementGRPCStream, newTable *v1pb.PlacementTables, newTableWithVNodes *v1pb.PlacementTables) error {
```
JoshVanL, line 390: Parsing two parameters of the same type can be tricky for the caller to get right. Please can we use and pass a `tablesUpdate` struct

pkg/placement/membership.go
```diff
@@ -33,6 +36,7 @@ const (
 	// raftApplyCommandMaxConcurrency is the max concurrency to apply command log to raft.
 	raftApplyCommandMaxConcurrency = 10
 	barrierWriteTimeout            = 2 * time.Minute
+	GRPCContextKeyExpectsVNodes    = "dapr-expects-vnodes"
```
JoshVanL, line 39: Not a fan of using `expects` as the client verb. Can we have something like `dapr-accept-vnodes`?.

pkg/placement/membership.go
```diff
@@ -335,16 +339,37 @@ func (p *Service) performTableDissemination(ctx context.Context) error {
 	p.disseminateLock.Lock()
 	defer p.disseminateLock.Unlock()
 
-	state := p.raftNode.FSM().PlacementState()
+	// Check if the cluster has daprd hosts that expect vnodes;
+	// older daprd versions (pre 1.13) do expect them, and newer versions (1.13+) do not.
+	state := &v1pb.PlacementTables{}
+	stateWithVNodes := &v1pb.PlacementTables{}
```
JoshVanL, line 345: Should these be nil by default?

pkg/placement/membership.go
```diff
@@ -452,3 +482,23 @@ func (p *Service) disseminateOperation(ctx context.Context, target placementGRPC
 		func() { log.Debug("Dissemination successful after failure") },
 	)
 }
+
+func hostExpectsVNodes(stream placementGRPCStream) bool {
```
JoshVanL, line 486:
```suggestion
func hostAcceptsVNodes(stream placementGRPCStream) bool {
```

pkg/placement/membership.go
```diff
@@ -452,3 +482,23 @@ func (p *Service) disseminateOperation(ctx context.Context, target placementGRPC
...
+	if len(expectsVNodes) == 0 {
+		return true // default to older versions that need vnodes
+	}
+
+	res, err := strconv.ParseBool(expectsVNodes[0])
+	if err != nil {
+		return true
+	}
+
+	return res
```
JoshVanL, line 503:
```suggestion
	// Extract apiLevel from metadata
	vmd := md.Get(GRPCContextKeyExpectsVNodes)
  return !(len(vmd) > 0 && strings.ToLower(vmd[0]) == "false")
```

tests/integration/suite/placement/vnodes/vnodes.go
```diff
@@ -0,0 +1,181 @@
...
+	)
+
+	return []framework.Option{
+		framework.WithProcesses(v.place),
+	}
+}
+
+func (v *vnodes) Run(t *testing.T, ctx context.Context) {
+	t.Run("register host without vnodes metadata (simulating daprd <1.13)", func(t *testing.T) {
+		v.place.WaitUntilRunning(t, ctx)
```
JoshVanL, line 50: Move to line above. This should only be done once

tests/integration/suite/placement/vnodes/vnodes.go
```diff
@@ -0,0 +1,181 @@
...
+		placementMessageCh := v.place.RegisterHostWithMetadata(t, ctx, msg, nil)
+
+		require.EventuallyWithT(t, func(t *assert.CollectT) {
+			select {
+			case <-ctx.Done():
+				return
+			case placementTables := <-placementMessageCh:
+				if ctx.Err() != nil {
+					return
+				}
```
JoshVanL, line 71: Remove

tests/integration/suite/placement/vnodes/vnodes.go
```diff
@@ -0,0 +1,181 @@
...
+			Port:     1234,
+			Entities: []string{"someactor"},
+			Id:       "myapp1",
+			ApiLevel: uint32(10),
+		}
+
+		// The host won't send the "dapr-expects-vnodes" metadata, simulating an older daprd version
+		placementMessageCh := v.place.RegisterHostWithMetadata(t, ctx, msg, nil)
+
+		require.EventuallyWithT(t, func(t *assert.CollectT) {
```
JoshVanL, line 64: Does this need to be in an Eventually?

tests/integration/suite/placement/vnodes/vnodes.go
```diff
@@ -0,0 +1,181 @@
...
+			Port:     2222,
+			Entities: []string{"someactor2"},
+			Id:       "myapp2",
+			ApiLevel: uint32(level2),
+		}
+
+		ctx1, cancel1 := context.WithCancel(ctx)
+		defer cancel1()
+		ctx2, cancel2 := context.WithCancel(ctx)
+		defer cancel2()
```
JoshVanL, line 137: Remove

JoshVanL (changes requested):
Some comments from me 🙂 
I'd like for us to be clear why this change needs to be made, and why we want to do it so close to release.
If it's purely for assumed performance benefits, we should push this back to 1.14

elena-kolevska:
No, it's not performance-related, as far as I understand. It concerns the APILevel usage and the "breaking change" it would introduce when using actor reminders and downgrading from 1.13 to 1.12 (1.13 stores actor reminders as protobuf and 1.12 as json).

pkg/placement/membership.go
elena-kolevska: Oh, I agree, this is ugly, and I'm actually considering a struct too, but I'm torn between having the code cleaner now vs. having these changes easier/obvious to revert when we do them in 1.15. What do you think?

yaron2:
@elena-kolevska please ping once this moves out of draft so we can run version skew tests.

tests/integration/suite/placement/vnodes/vnodes.go
elena-kolevska: Yeah, it does, because we're getting empty tables at first

elena-kolevska:
> @elena-kolevska please ping once this moves out of draft so we can run version skew tests.

Ping @yaron2  :)

artursouza:
> Some comments from me 🙂 I'd like for us to be clear why this change needs to be made, and why we want to do it so close to release. If it's purely for assumed performance benefits, we should push this back to 1.14

This was the original design but it changed to use apiVersion instead. Now, with the decision to force a lower minimum apiVersion (JSON reminders as default), this feature will not be enabled anymore. We want to enable this feature independently of others. ApiVersion bundles multiple features at a given number.

pkg/actors/placement/placement.go
```diff
@@ -551,10 +551,10 @@ func (p *actorPlacement) updatePlacements(in *v1pb.PlacementTables) {
 			// TODO: @elena in v1.15 remove the check for versions < 1.13
 			// only keep `hashing.NewFromExisting`
 
-			if p.apiLevel < raft.NoVirtualNodesInPlacementTablesAPILevel {
-				entries[k] = hashing.NewFromExistingWithVirtNodes(v.GetHosts(), v.GetSortedSet(), loadMap)
-			} else {
+			if in.GetReplicationFactor() > 0 && len(v.GetHosts()) == 0 {
```
JoshVanL, line 554: Couldn’t len hosts be 0 in correct operation? I think it would make sense for placement to return a header which daprd checks for here. `dapr-content-vnodes=false`.

pkg/actors/placement/placement_test.go
```diff
@@ -36,6 +36,7 @@ import (
 	"github.com/dapr/dapr/pkg/placement/hashing"
 	placementv1pb "github.com/dapr/dapr/pkg/proto/placement/v1"
 	"github.com/dapr/dapr/pkg/resiliency"
+
```
JoshVanL, line 39: Remove

pkg/placement/membership.go
```diff
@@ -33,6 +35,7 @@ const (
 	// raftApplyCommandMaxConcurrency is the max concurrency to apply command log to raft.
 	raftApplyCommandMaxConcurrency = 10
 	barrierWriteTimeout            = 2 * time.Minute
+	GRPCContextKeyExpectsVNodes    = "dapr-accept-vnodes"
```
JoshVanL, line 38: Rename const name based on new value

pkg/placement/membership.go
JoshVanL: I think it is better to do the struct now. We’ll keep the struct forever so it would only be a case of removing the field in future. Using a struct will also move the need for `hasNewDapr` and `hasOldDaprds` above. Not a massive fan of `old` and `new` anywho as that doesn’t mean too much.

pkg/placement/membership_test.go
```diff
@@ -22,14 +22,17 @@ import (
...
 	"time"
 
+	"google.golang.org/grpc/metadata"
+
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"google.golang.org/grpc"
 	clocktesting "k8s.io/utils/clock/testing"
 
+	"github.com/dapr/kit/logger"
```
JoshVanL, line 32: Move back

pkg/placement/membership_test.go
```diff
@@ -22,14 +22,17 @@ import (
 	"testing"
 	"time"
 
+	"google.golang.org/grpc/metadata"
+
```
JoshVanL, line 26: Remove

pkg/placement/placement.go
```diff
@@ -34,6 +34,7 @@ import (
 	placementv1pb "github.com/dapr/dapr/pkg/proto/placement/v1"
 	"github.com/dapr/dapr/pkg/security"
 	"github.com/dapr/dapr/pkg/security/spiffe"
+
```
JoshVanL, line 37: Remove

pkg/placement/placement.go
```diff
@@ -250,9 +251,16 @@ func (p *Service) ReportDaprStatus(stream placementv1pb.Placement_ReportDaprStat
 				registeredMemberID = req.GetName()
 				p.addStreamConn(stream)
 				// We need to use a background context here so dissemination isn't tied to the context of this stream
-				placementTable := p.raftNode.FSM().PlacementState()
+				placementTableWithoutNodes := &placementv1pb.PlacementTables{}
+				placementTableWithNodes := &placementv1pb.PlacementTables{}
```
JoshVanL, line 255:
```suggestion
				var placementTableWithoutNodes *placementv1pb.PlacementTables
				var placementTableWithNodes *placementv1pb.PlacementTables
```

JoshVanL (changes requested):

pkg/actors/placement/placement.go
elena-kolevska: Yes, `v.GetHosts()` will always be empty in correct operation in 1.13+. 
For 1.12 and lower, it should never be empty, because the disseminated placement table is dynamically generated based on the host data saved in Raft, meaning - if there is an entry for an actor type (`v`) it will always have a `load_map` and consequentially - `hosts` and `sorted_set`.

JoshVanL: >  if there is an entry for an actor type (v) it will always have a load_map and consequentially - hosts and sorted_set.

Doesn't this mean it will be empty when there is no actor type?}

elena-kolevska: The placement table pre 1.13 looks like this:
```
{
  "entries": {
    "myactortype-0": {
      "hosts": {
        "10056481384176189962": "192.168.1.89:62362",
        "10100244799470048543": "192.168.1.89:62362",
        "10136976372505843558": "192.168.1.89:62362",
        "10210620769057790472": "192.168.1.89:62362",
        "10278294263954349758": "192.168.1.89:62345",
		    ...
        "9959649394905760010": "192.168.1.89:62345",
        "999880486694283580": "192.168.1.89:62362"
      },
      "sorted_set": [
        3430792224860206,
        10660957680821358,
        69520260237277928,
        96174763468296109,
        110417171835179485,
        ...
        18409429381221104921,
        18426863874365556280
      ],
      "load_map": {
        "192.168.1.89:62345": {
          "name": "192.168.1.89:62345",
          "id": "08aa8f86-df9b-45bd-b407-f35c67535073"
        },
        "192.168.1.89:62362": {
          "name": "192.168.1.89:62362",
          "id": "140d66c3-66b4-4642-ade5-0d17b4cd0402"
        },
        "192.168.1.89:62367": {
          "name": "192.168.1.89:62367",
          "id": "6c1f823e-b3b0-43d6-b7ca-ce844ae1ae2c"
        }
      }
    },
    "myactortype-1": {
      "hosts": {
        "10056481384176189962": "192.168.1.89:62362",
        "10100244799470048543": "192.168.1.89:62362",
        "10136976372505843558": "192.168.1.89:62362",
        ...
        "9475149095182223555": "192.168.1.89:62362",
        "999880486694283580": "192.168.1.89:62362"
      },
      "sorted_set": [
        69520260237277928,
        110417171835179485,
       ...
        18297525576440347375,
        18375786333053603218
      ],
      "load_map": {
        "192.168.1.89:62362": {
          "name": "192.168.1.89:62362",
          "id": "140d66c3-66b4-4642-ade5-0d17b4cd0402"
        }
      }
    }
  },
  "version": "3",
  "api_level": 20
}
```

If an actor type is not present, it won't be a part of the placement table in the first place.

JoshVanL: OK. It seems a bit flaky to me to rely on the data structure contents to make this decision, rather than the same API surface which we use to signal whether to return the vnodes. If we are sure that this will never be problem, than fine.

pkg/placement/placement.go
```diff
@@ -77,6 +77,52 @@ type hostMemberChange struct {
 	host    raft.DaprHostMember
 }
 
+type placementTablesUpdate struct {
+	withoutVirtualNodes *placementv1pb.PlacementTables
```
JoshVanL, line 81:
```suggestion
type tablesUpdateRequest struct {
   hosts []placementGRPCStream
	withVNodes *placementv1pb.PlacementTables
	noVNodes *placementv1pb.PlacementTables
```

pkg/placement/membership.go
```diff
@@ -335,21 +338,40 @@ func (p *Service) performTableDissemination(ctx context.Context) error {
...
 
-	state := p.raftNode.FSM().PlacementState()
-
-	log.Infof(
-		"Start disseminating tables. memberUpdateCount: %d, streams: %d, targets: %d, table generation: %s",
-		cnt, nStreamConnPool, nTargetConns, state.GetVersion())
 	p.streamConnPoolLock.RLock()
 	streamConnPool := make([]placementGRPCStream, len(p.streamConnPool))
 	copy(streamConnPool, p.streamConnPool)
+
```
JoshVanL, line 344: This is more what I meant.

```suggestion
	req := &placementTablesUpdate{
		hosts: streamConnPool,
	}

	for _, stream := range streamConnPool {
		if hostAcceptsVNodes(stream) {
			if req.wtihVNodes == nil {
				req.withVNodes = fsm.PlacementState(false)
			}
		} else {
			if req.noVNodes == nil {
				req.noVNodes = fsm.PlacementState(false)
			}
		}

		if req.withVNodes != nil && req.noVNodes != nil {
			break
		}
	}
 
```

JoshVanL (changes requested):

pkg/actors/placement/placement.go
artursouza: I agree with @elena-kolevska. I like that we use the content to decide here.

pkg/actors/placement/placement.go
```diff
@@ -551,10 +551,10 @@ func (p *actorPlacement) updatePlacements(in *v1pb.PlacementTables) {
 			// TODO: @elena in v1.15 remove the check for versions < 1.13
```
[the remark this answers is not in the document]
elena-kolevska: It won't hurt if it's left in, but I think the code will be tidier if we remove it. Of course, this is only the case if we're keeping with the decision to support rolling upgrades/downgrades only one version in each direction.

artursouza (approved):

PR #7548, opened 2024-02-19
Actor Reminders: Default JSON serialization.

JoshVanL:
To support downgrades to 1.12 from 1.13, this PR changes the reminder serialization storage format back to JSON by default. This means a 1.12 actor reminder client can read reminders written by 1.13 actors.

1.13 will continue to understand both JSON and protobuf. Protobuf serialization can be enabled with the `ActorReminderStorageProtobuf` feature gate. The actor "API Level" has been changed back to 10.

Adds test to ensure the default serialization is JSON.

### DIFF
--- a/pkg/actors/actors.go
+++ b/pkg/actors/actors.go
@@ -41,7 +41,7 @@ import (
 	"github.com/dapr/dapr/pkg/actors/internal"
 	"github.com/dapr/dapr/pkg/actors/timers"
 	"github.com/dapr/dapr/pkg/channel"
-	configuration "github.com/dapr/dapr/pkg/config"
+	"github.com/dapr/dapr/pkg/config"
 	diag "github.com/dapr/dapr/pkg/diagnostics"
 	diagUtils "github.com/dapr/dapr/pkg/diagnostics/utils"
 	invokev1 "github.com/dapr/dapr/pkg/messaging/v1"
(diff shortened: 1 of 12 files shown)


ItalyPaleAle (changes requested):
This need not be a configuration option, because no one (eventually) will want to have JSON.

To keep 1.13 to use JSON by default you simply just need to change this line:

https://github.com/dapr/dapr/blob/518753d502cbb1b57b244031b574862944ad86ce/charts/dapr/charts/dapr_placement/values.yaml#L26

Change it to `10` and you're done :)

JoshVanL:
@ItalyPaleAle Helm values currently aren't testable so is safer to do in code imo. It also doesn't cover placement in self-hosted mode.

ItalyPaleAle:
> @ItalyPaleAle Helm values currently aren't testable so is safer to do in code imo. It also doesn't cover placement in self-hosted mode.

This is not what was discussed in the call according to @mukundansundar ?

I really do not want a configuration option in the CRD as it will be there forever. The entire point of using API version in placement was so we could migrate more quickly, but adding a Configuration CRD option is a perennial choice.

For self-hosted we need to update the Docker Compose file examples. I would not be too concerned about self hosted as actor usage there is likely a tiny fraction.

JoshVanL:
The protobuf serialization is opt-in through a feature gate rather than a dedicated config CRD field, and expected to be removed in 1.14 where protobuf is the only supported option.

elena-kolevska (approved):

yaron2 (approved):

PR #7565, opened 2024-02-28
Always tick renewal check every minute

JoshVanL:
It is sometimes the case that Dapr services are suspended for a period of time, for example when developer closes their laptop over night, before being resumed after a significant period of time. Due to the renewal timer being a Go Timer, it will also be suspended during this time and only resume after the process is resumed.

Because X.509 certificates expiration validity follow wall-clock time and the renewal timer follows process up-time time, it is possible for a Dapr identity certificates to expire and not be renewed until enough observed time has passed for the renewal timer to trigger.

This commit introduces a safety check whereby the renewal timer will always trigger either every minute or at renewal time, whichever is sooner. Once ticked, the renew loop will check to ensure whether the renewal time has actually been reached before renewing.

This change helps out developers by not needing to restart the Dapr cluster when they suspend their Dapr clusters for a period of time. While it is not expected that this change be useful in production, it is possible users are suspending and resuming virtual machines in production.

Adds a renewal integration test to ensure Dapr identity certificate is renewed based on expiration.


artursouza (approved):

cicoyle (approved):
Thanks for adding the tests 🎉

PR #7568, opened 2024-02-28
chore: Convert actor HTTP APIs off fasthttp

ItalyPaleAle:
Changed APIs for actors to not be based on fasthttp.

- 2 APIs have been converted to universal
- Remaining APIs have been converted to net/http (not universal) due to differences in how they managed data between gRPC and HTTP
- APIs are now in their own "actors.go" file

There are no changes in behavior.


pkg/messages/predefined.go
```diff
@@ -102,6 +91,15 @@ var (
 	// Actor.
 	ErrActorReminderOpActorNotHosted = APIError{"operations on actor reminders are only possible on hosted actor types", "ERR_ACTOR_REMINDER_NON_HOSTED", http.StatusForbidden, grpcCodes.PermissionDenied}
 	ErrActorRuntimeNotFound          = APIError{`the state store is not configured to use the actor runtime. Have you set the - name: actorStateStore value: "true" in your state store component file?`, "ERR_ACTOR_RUNTIME_NOT_FOUND", http.StatusInternalServerError, grpcCodes.Internal}
+	ErrActorInstanceMissing          = APIError{"actor instance is missing", "ERR_ACTOR_INSTANCE_MISSING", http.StatusBadRequest, grpcCodes.Internal}
```
cicoyle, line 94: Since you are updating this anyways, could you move it to the `api/errors` folder and build the actor errors in the standardized way? Its basically the same thing but a standardized format, [here are the examples we currently have](https://github.com/dapr/dapr/tree/master/pkg/api/errors). Its a small tweak, but I think a nice to have for users. 

Also, is the `ERR_ACTOR_INSTANCE_MISSING` legacy or are you adding this with this PR? Ideally, I'd like to keep the format we have for the standardized errors where [legacy codes should be retained here](https://github.com/dapr/dapr/blob/master/pkg/api/errors/pubsub.go#L44), but we [create a new 'reason' here](https://github.com/dapr/dapr/blob/master/pkg/api/errors/pubsub.go#L46) where it should be starting with the api: `ACTOR_INSTANCE_MISSING` for example, where we dont want `ERR` prefixed anymore.

cicoyle: If the gRPC code is `Internal` I would assume we should use the equivalent http code - unless this was intentional?

```suggestion
	ErrActorInstanceMissing          = APIError{"actor instance is missing", "ERR_ACTOR_INSTANCE_MISSING", http.StatusInternalServerError, grpcCodes.Internal}
```

ItalyPaleAle: The goal of this PR was to maintain the existing behavior and just migrate the handlers. I would rather migrate errors in a separate PR if possible.

As for the different status codes  between HTTP and gRPC... Yes, that's an inconsistency in the current code, which I did not want to touch now. It's not the only inconsistency even: for example, if you try to save an actor state using gRPC and read it using HTTP, it may not work. Even worse, there's an HTTP API to read a reminder, but that equivalent API doesn't exist for gRPC:

JoshVanL (approved):

JoshVanL (approved):

PR #7569, opened 2024-02-28
chore: convert crypto APIs to net/http

ItalyPaleAle:
Similar to #7568, converts the 2 crypto APIs (high-level) from fasthttp to net/http (these 2 cannot be universal)

No changes in behavior


cicoyle (approved):
LGTM. Although, can you elaborate as to why we want to move away from `fasthttp` -> `http`?

ItalyPaleAle:
> LGTM. Although, can you elaborate as to why we want to move away from `fasthttp` -> `http`?

We have already migrated the web server from fasthttp to the standard library. See https://github.com/dapr/dapr/issues/4979

However, a lot of the handlers are still written as fasthttp handlers. We did not convert them all mostly due to resourcing issues, and to do things more gradually. For handlers that are still written as fasthttp handlers, we have a "shim" that converts the request and response in the stdlib format. Understandably, the shim adds a negative performance impact (due to data being copied over multiple times), increases complexity (for example debuggability is harder), and forces us to maintain a dependency on fasthttp.

JoshVanL (approved):

daixiang0 (approved):
LGTM

ItalyPaleAle:
@daixiang0 automerge isn't working, can you merge please?

daixiang0:
@ItalyPaleAle sure.

PR #7573, opened 2024-02-29
Switch to OTel ParentSampler for correct sampling behavior

andrejpk:
# Description

- [x] Remove custom sampler logic and use OTel's ParentSampler instead for correct behavior
- [x] Update tests to correctly test different behavior for no span parent vs parent flag = 0
- [x] ~~Add an additional sampling config to differentiate between SamplingRate=0 and tracing disabled~~ seems like a separate change and maybe a broader look at improving tracing config

Instead of [problematic] custom logic, we can simply use the OpenTelemetry ParentSampler to a more sane default behavior. See the issue for more details. 

This PR fixes the underlying issue without modifying tracing configuration options, but a separate issue and PR would make sense to provide a more complete tracing configuration. For example, there's no way to enable tracing but set the trace level to 0 (which makes total sense if you are relying on upstream calls to set the trace flag). But a change like this should take a broader look at tracing with one config schema change.

## Issue reference

Please reference the issue this PR will close: #7574

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Unit tests passing
* [x] **N/A** ~~Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo:  dapr/docs#_[issue number]_~~
* [x] **N/A** ~~Specification has been updated / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_~~
* [X] **N/A** ~~Provided sample for the feature / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_~~


ItalyPaleAle (commented):
Thank you, this is a good improvement!

Can you please take a look at the linter failures? (Locally, that's `make lint`)

daixiang0 (approved):
LGTM

marcduiker:
@holopin-bot @andrejpk

Issue #7139, opened 2023-11-02
Hot-reloading of Subscriptions


JoshVanL:
## In what area(s)?

/area runtime

/area operator

## Describe the feature

Subscription resources are only loaded once on startup- both in Kubernetes and Standalone mode. This means that they must be installed _before_ daprd has been deployed, or daprd needs to be restarted to pickup new/drop old subscription config.

Daprd should implement hot-reloading of Subscription configuration live, when that config changes.

## Release Note

RELEASE NOTE: **ADD** Hot-reloading for Dapr Subscriptions

olitomlinson:
+1 this ask has appeared a couple of times on Discord from users

PR #7582, opened 2024-03-03
Subscriptions: Declarative Go Generic Loader

JoshVanL:
Following Components and HTTPEndpoints, creates a generic disk &
kubernetes loader for Subscriptions. Removes the adhok loaders from
existing runtime pubsub packages. This is required for adding
Subscriptions to the hot reloader reconciler and making a more
consistent & testable manifest loader package.

The disk loader loads both `v1alpha1` and `v2alpha1` Subscriptions,
converting `v1alpha1` to `v2alpha1` on successful load. Since the
operator returns only `v2alpha1` Subscriptions (Kubernetes API converts
`v1alpha1` to `v2alpha1`), the Kubernetes loader only loads `v2alpha1`.

`APIVersion() string` func has been added to the generic `meta.Resource`
interface type to allow for the generic loader to determine between
resource versions, supporting differentiating `v1alpha1` and `v2alpha1`
Subscription resource versions.

To ensure backwards compatibility of the previous disk Subscription
loader, the generic disk loader now tracks the order in which manifests
are loaded. This ensures that, even though `v1alpha1` and `v2alpha1`
Subscriptions use separate loaders, their file position order is
preserved once `v1alpha1` Subscriptions are converted to `v2alpha1`.
Subscription backwards compatibility of parsing & ingestion priority is
covered by existing extensive integration tests.

Notice that _ZERO_ of the comprehensive subscription integration tests
have been modified, proving no behaviour change has occurred for loading
and actuating Subscriptions.

Explanation of integration test changes:
- daprd/hotreload/selfhosted/crypto: loader now mandates `apiVersion` on
  disk manifests which is more correct. Previous behaviour ignoring this
  field should be considered a bug. Since we use Kubernetes resource
  API schema, `apiVersion` is (and has been) _always_ required.
- framework/process/grpc/operator/server.go: returning non-nil gRPC
  object for mocked operator Subscription RPC call prevents underlying
  gRPC library unmarshal nil errors- more correct than previous
  implementation.

Part of [Subscription hot-reloading](https://github.com/dapr/dapr/issues/7139).


pkg/internal/loader/disk/disk_test.go
```diff
@@ -195,13 +195,136 @@ metadata:
...
+  type: state.couchbase
+`), fs.FileMode(0o600)))
+			}
+		}
+
+		d := NewComponents(tmp1, tmp2, tmp3).(*disk[compapi.Component])
+		set, err := d.loadWithOrder()
+		require.NoError(t, err)
+		assert.Equal(t, []manifestOrder{
+			{dirIndex: 0, fileIndex: 0, manifestIndex: 0},
```
cicoyle, line 307: Can you explain why the `manifestIndex` goes from 0 -> 2 here?

JoshVanL: Indeed- we have 3 directories, writing 3 files each, writing 3 manifests each- so index 0,1,2 on each. Since we are creating a `Component` disk loader, the test is ensuring that the `Subscription` manifest at index `1` is ignored. Hence why `manifestIndex` `1` is skipped and not expected on the result.

pkg/internal/loader/disk/manifest.go
```diff
@@ -0,0 +1,202 @@
...
+			errs = append(errs, err)
+			continue
+		}
+
+		if ti.APIVersion != m.d.apiVersion || ti.Kind != m.d.kind {
+			continue
+		}
+
+		if errS := path.IsValidPathSegmentName(ti.Name); len(errS) > 0 {
+			errs = append(errs, fmt.Errorf("invalid name %q for %q: %s", ti.Name, m.d.kind, strings.Join(errS, "; ")))
```
cicoyle, line 135: Why concatenate errS with a `;`? Should `errS` be `pathErrs` to avoid confusion with `errs`.

pkg/internal/loader/disk/subscriptions.go
```diff
@@ -0,0 +1,63 @@
...
+limitations under the License.
+*/
+
+package disk
+
+import (
+	"context"
+	"sort"
+
+	subv1api "github.com/dapr/dapr/pkg/apis/subscriptions/v1alpha1"
```
cicoyle, line 20:
```suggestion
"github.com/dapr/dapr/pkg/apis/subscriptions/v1alpha1"
```
Just a thought - Since its already in the `subscriptions.go` file I think we can just use `v1alpha1`. `subv1api` could be misinterpreted. or even just `v1api`?

pkg/internal/loader/disk/manifest.go
JoshVanL: Done! :slightly_smiling_face: 

This kube package is older than `errors.Join` hence why it is returning a slice of strings. Name change makes sense to me :+1:

pkg/internal/loader/disk/subscriptions.go
JoshVanL: Yeah, using the bare package name makes sense to me.

pkg/internal/loader/kubernetes/components_test.go
```diff
@@ -27,7 +27,7 @@ import (
 	"google.golang.org/grpc/credentials/insecure"
 
 	"github.com/dapr/dapr/pkg/apis/components/v1alpha1"
-	subscriptions "github.com/dapr/dapr/pkg/apis/subscriptions/v1alpha1"
+	subapi "github.com/dapr/dapr/pkg/apis/subscriptions/v2alpha1"
```
cicoyle, line 30: is this right? You are now using the v2alpha import? should that `v2` be included in the import name: `subapi` -> `subv2api`

JoshVanL: Yes, that's right- In Kubernetes (and internally really), `v2alpha1` is our "target" and stored version where created `v1alpha1` Subscription resources are converted into `v2alpha1` types during admission. Dapr Kube clients will only deal with the `v2alpha1` type. 

This `ListSubscriptionsV2` test func was incorrectly using `v1alpha`, event though the func name itself is `V2` (!)

pkg/runtime/wfengine/component.go
```diff
@@ -22,10 +22,8 @@ import (
 
 	"github.com/microsoft/durabletask-go/api"
 	"github.com/microsoft/durabletask-go/backend"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
-	"github.com/dapr/components-contrib/workflows"
-	componentsV1alpha1 "github.com/dapr/dapr/pkg/apis/components/v1alpha1" // This will be removed
+	"github.com/dapr/components-contrib/workflows" // This will be removed
```
cicoyle, line 26: why will the import be removed?

JoshVanL: I don't know. Seems to be a weird copy/paste error inheriting the code comment from the removed line in the diff. Removed.

pkg/internal/loader/disk/disk_test.go
```diff
@@ -195,13 +195,136 @@ metadata:
...
+kind: Component
+metadata:
+  name: statestore
+spec:
+  type: state.couchbase
+---
+apiVersion: dapr.io/v1alpha1
+kind: Component
+metadata:
+  name: statestore
```
cicoyle, line 250: Can these names be different to avoid confusion since this isn't valid config in Dapr? It confuses the example you are testing since the `metadata.name` && `spec.type` are the same for each but the `manifestIndex` is different each time. Or make a note that its not valid config but its testing the indexing properly?

cicoyle (approved):
Looks great Josh, just had a few minor comments :)

JoshVanL:
Thanks @cicoyle, updated!

yaron2 (approved):

PR #7583, opened 2024-03-04
Hot Reload: Subscription

JoshVanL:
Adds support for Subscription Hot Reloading.

If the Hot Reloading preview feature is enabled, create, update and
delete events to declarative Subscriptions will be reflected in the
Subscriptions runtime without needing to restart the Dapr runtime. Works
in both self-hosted and Kubernetes modes.

Note that, because topic routes and not indexed by Subscription
resources, all Subscriptions will be cancelled and topic routes re-built
and re-subscribed when a Subscription event occurs. A possible
improvement would to only cancel all topic route subscriptions which are
related to the Subscription pubsub or include an index of the
Subscription which a topic is routed for, however this is out of scope
of this change.

Closes https://github.com/dapr/dapr/issues/7139
Docs https://github.com/dapr/docs/pull/4116


pkg/operator/api/components.go
```diff
@@ -0,0 +1,240 @@
...
+			case <-ctx.Done():
+			}
+		}(connUpdateChan)
+	}
+	wg.Wait()
+	a.connLock.Unlock()
+}
+
+// ComponentUpdate updates Dapr sidecars whenever a component in the cluster is modified.
+// TODO: @joshvanl: Authorize pod name and namespace matches the SPIFFE ID of
```
yaron2, line 61: Will this require a K8s API server call?

pkg/runtime/processor/subscriptions.go
```diff
@@ -0,0 +1,89 @@
...
+func (p *Processor) AddPendingSubscription(ctx context.Context, subscriptions ...subapi.Subscription) bool {
+	p.lock.Lock()
+	defer p.lock.Unlock()
+	if p.shutdown.Load() {
+		return false
+	}
+
+	scopedSubs := make([]subapi.Subscription, 0, len(subscriptions))
+	for i, subscription := range subscriptions {
+		if len(subscription.Scopes) > 0 && !utils.Contains[string](subscription.Scopes, p.appID) {
```
yaron2, line 32: Can unit tests be added to this method? It contains authorization logic

yaron2 (changes requested):

pkg/operator/api/components.go
JoshVanL: No. This comment isn't written very well as pod name isn't involved- it is app ID we care about. It is referring to server-side Component scope filtering, implemented [here](https://github.com/dapr/dapr/pull/7615).

Component updates are broadcast to the correct streaming clients based on their identity derived via mTLS, stored in-memory.

pkg/runtime/processor/subscriptions.go
JoshVanL: Sure thing- note that security is a mute point here as this is sidecar code so we already have the data. While fine to check again (and have to for backwards compat), we need to add server side filtering for subscriptions [like components](https://github.com/dapr/dapr/pull/7615).

JoshVanL: Done in https://github.com/dapr/dapr/pull/7583/commits/4c984c4ab68b35cb8fa377bf4d13a568ac031fb7

pkg/runtime/processor/subscriptions.go
```diff
@@ -0,0 +1,93 @@
...
+
+func (p *Processor) AddPendingSubscription(ctx context.Context, subscriptions ...subapi.Subscription) bool {
+	p.lock.Lock()
+	defer p.lock.Unlock()
+	if p.shutdown.Load() {
+		return false
+	}
+
+	scopedSubs := p.scopeFilterSubscriptions(subscriptions)
+	if len(scopedSubs) == 0 {
```
yaron2, line 31: This will not load a subscription tha has no scopes, but its possible today to use a subscription with no scopes.

JoshVanL: The list is of subscription manifests. If after filtering subscriptions we end up with an empty list, we exit early.

yaron2: To be clear, this will return true which signals the subscription was added but it will not in effect be added to the comp store. Is that the case?

JoshVanL: It returns `true` indicating that processing of this resource was successful. We classify that not processing this non-scoped subscription is a happy path result.

yaron2: So returning to my original point - if we allow loading non-scoped subscriptions on init, then shouldn't this method also do so?

JoshVanL: I think I'm mangling my words a bit.

This function removes subscription manifests which this app is not scoped for. A subscription is scoped when the manifest has either an empty scope string slice _or_ the appID is in the scope list. This scoping behaviour is that same as on init today. After filtering the list results in an empty list of subscriptions, there are no subscriptions to process and we can exit early signalling success.

yaron2: Ok understood

yaron2 (approved):

marcduiker:
@holopin-bot @JoshVanL

PR #7588, opened 2024-03-05
Refactor api/errors into builders

JoshVanL:
Refactors api/errors into functional builders. The current API for `api/errors` is heavily string concatenation based with long parameter arguments. This is fragile/error prone to the developer and difficult to understand for the consumer/reader.

This change should make it clearer to the reader/consumer what the intended errors are and how to use them. Moving implementation details of the error protocol codes into the error package also helps to clean up API implementation code. PR also help de-duplicate common kit error params.


pkg/api/errors/pubsub.go
```diff
@@ -17,170 +17,229 @@ import (
...
-	).
-		WithErrorInfo(kiterrors.CodePrefixPubSub+PostFixForbidden, map[string]string{"topic": topic, "error": err.Error()}).
-		WithResourceInfo(pubsubType, name, "", message).
-		Build()
-}
+// UnmarshalEvents only occurs in http/api.go
+func (p *PubSubTopicError) UnmarshalEvents(err error) error {
+	message := fmt.Sprintf(
+		"error when unmarshaling the request for topic %s pubsub %s: %s",
+		p.topic, p.p.p.name, err,
```
cicoyle, line 212: `p.p.p.name` - is that a typo or is it really accessed in this manner now?

JoshVanL: Yes, it is because we are wrapping with more context which this error type needs. We can pick a different variable name if you prefer.

cicoyle:
I see what you are doing, but I wonder if this change makes it more complicated for contributors. I've been seeing quite a few questions from contributors wanting to assist in the error efforts and I almost wonder if adding the pointer receiver in this manner makes it more complicated to understand and add to.

pkg/components/state/state_config.go
```diff
@@ -66,7 +66,7 @@ func SaveStateConfiguration(storeName string, metadata map[string]string) error
 
 func GetModifiedStateKey(key, storeName, appID string) (string, error) {
 	if err := checkKeyIllegal(key); err != nil {
-		return "", errors.StateStoreInvalidKeyName(storeName, key, err.Error())
+		return "", errors.StateStore(storeName).InvalidKeyName(key, err.Error())
```
cicoyle, line 69: I do like this look tho, it's less flat and can be easier to view by api

cicoyle:
Looks like something is up with the secret integration tests. Not sure if that is a result of this change, but its on all the errors so maybe?

JoshVanL:
> Looks like something is up with the secret integration tests. Not sure if that is a result of this change, but its on all the errors so maybe?

@cicoyle yeah, those really need fixing.. The issue is with the fuzz data generator producing invalid data for the test resulting in errors.

pkg/api/errors/state.go
```diff
@@ -20,73 +20,136 @@ import (
...
+			http.StatusBadRequest,
+			msg,
+			"ERR_STATE_STORE_NOT_FOUND",
+		),
+		errors.CodeNotFound,
+		meta,
+	)
+}
+
+func (s *StateStoreError) NotConfigured(appID string) error {
```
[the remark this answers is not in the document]
cicoyle: I wonder if it makes sense to still have a generic error. I can think of a few off the top of my head that apply to more than one of our APIs:

- NotConfigured
- Empty
- NotFound
- etc

tests/integration/suite/daprd/pubsub/grpc/errors.go
```diff
@@ -190,7 +189,7 @@ func (e *standardizedErrors) Run(t *testing.T, ctx context.Context) {
 			}
 		}
 		require.NotNil(t, errInfo, "ErrorInfo should be present")
-		require.Equal(t, kiterrors.CodePrefixPubSub+apierrors.PostFixNameEmpty, errInfo.GetReason())
```
cicoyle, line 193: if you dont want the concatenation of the consts, Id rather a file dedicated to these consts over magic strings.

tests/integration/suite/daprd/pubsub/http/fuzz.go
```diff
@@ -19,7 +19,7 @@ import (
 	"crypto/rand"
 	"encoding/base64"
 	"encoding/json"
-	"errors"
+	stderrors "errors"
```
cicoyle, line 22: is this change so the focus for `errors` tests can be consistent across all apis? Ik its overlapping the names

tests/integration/suite/daprd/pubsub/grpc/errors.go
elena-kolevska: In the context of tests, I would cast my vote for strings. Since this is as an api exposed to users, they rely on that specific string, so that's why comparing to it directly is safer.

tests/integration/suite/daprd/pubsub/http/fuzz.go
elena-kolevska: Yeah, I think the three of us talked about this and had decided that we keep the std packages with their own names and import the kit error under the same alias everywhere. I'm fine with whatever decision, as long as it's consistent.

pkg/api/errors/pubsub.go
```diff
@@ -17,170 +17,229 @@ import (
...
+			"error": err.Error(),
+		},
+	}
 }
 
-func PubSubPublishMessage(name string, pubsubType string, topic string, err error) error {
-	message := fmt.Sprintf("error when publishing to topic %s in pubsub %s: %s", topic, name, err.Error())
-	return kiterrors.NewBuilder(
-		grpcCodes.Internal,
+func (p PubSubError) PublishMessage(topic string, err error) error {
```
elena-kolevska, line 77: I really like the "fluent-like" way this refactor builds errors. I'm a bit divided about the `PubSubMetadataError` in the middle, but I understand why it's there. It has its pros and cons.
It just feels a bit weird to me to have the intermediate methods that return `PubSubMetadataError` for some errors, but not for others. Right now we have final methods (the one that calls `build`) on both `PubSubError` and `PubSubMetadataError` and that might be a bit confusing when you see it as an IDE suggestion.

tests/integration/suite/daprd/pubsub/grpc/errors.go
JoshVanL: Indeed :+1

tests/integration/suite/daprd/pubsub/http/fuzz.go
JoshVanL: :+1:

pkg/api/errors/pubsub.go
JoshVanL: How should be proceed here, remove the the `Metadata` intermediate struct?

pkg/api/errors/state.go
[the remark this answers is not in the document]
cicoyle: As discussed offline, this will be the forward moving implementation that each api will implement these generic errors for their api since the `trade off is totally fine given the benefit of things being typed.` We want to avoid having untyped string like params.

artursouza (changes requested):
@elena-kolevska has an important comment on a design decision, please, do not merge until that is reconciled.

pkg/api/errors/pubsub.go
elena-kolevska: Since the metadata struct is very useful for setting the appid metadata, I think we can just leave it for now. If we see the relationships between the three (PubSubError, PubSubMetadataError and PubSubTopicError) get more complex in the future, we can always refactor.

elena-kolevska (approved):

yaron2 (approved):

PR #7591, opened 2024-03-05
chore: bump go to 1.21.8 and protobuf lib to 1.33.0

mikeee:
# Description

Bumps go to 1.21.8 - https://groups.google.com/g/golang-announce/c/5pwGVUPoMbg
Bumps google.golang.org/protobuf to 1.33.0 - https://groups.google.com/g/golang-announce/c/ArQ6CDgtEjY

## Issue reference

Please reference the issue this PR will close: #N/A

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing

### DIFF
--- a/.github/workflows/dapr-perf-components.yml
+++ b/.github/workflows/dapr-perf-components.yml
@@ -222,7 +222,7 @@ jobs:
         id: setup-go
         uses: actions/setup-go@v5
         with:
-          go-version-file: 'go.mod'
+          go-version: '1.21.8'
       - name: Login to Azure
         if: env.CHECKOUT_REPO != ''
         uses: azure/login@v1
@@ -367,7 +367,7 @@ jobs:
         id: setup-go
         uses: actions/setup-go@v5
         with:
-          go-version-file: 'go.mod'
+          go-version: '1.21.8'
       - uses: azure/setup-kubectl@v3
         with:
           version: ${{ env.KUBECTLVER }}
--- a/.github/workflows/dapr-perf.yml
+++ b/.github/workflows/dapr-perf.yml
@@ -244,7 +244,7 @@ jobs:
         id: setup-go
         uses: actions/setup-go@v5
         with:
-          go-version-file: 'go.mod'
+          go-version: '1.21.8'
       - name: Login to Azure
         if: env.CHECKOUT_REPO != ''
         uses: azure/login@v1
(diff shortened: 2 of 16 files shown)


berndverst (commented):
While this works, I prefer doing it differenlt:

Instead change the Go version in the `.github/workflows` pipelines. We have done that before.

This is because we generally want the `go.mod` itself to use the latest `go1.21` available.

We only need to change the workflow to specifically install Go 1.21.8 temporarily (since that doesn't get installed by Github until next week)

JoshVanL:
> While this works, I prefer doing it differenlt:
> 
> Instead change the Go version in the `.github/workflows` pipelines. We have done that before.
> 
> This is because we generally want the `go.mod` itself to use the latest `go1.21` available.
> 
> We only need to change the workflow to specifically install Go 1.21.8 temporarily (since that doesn't get installed by Github until next week)

I don't have super strong opinions either way but we need to make sure we update _every_ instance of `actions/setup-go@v5` in `.github/workflows` or _every_ version directive in all `go.mod`s in the project to to `1.21.8` to ensure correctness/completeness. Personally, I think changing every `go.mod` is a bit better on the diff/sed -i but that's neither here nor there.

N.B. I wish we used nix :slightly_smiling_face:

ItalyPaleAle:
A few things:

1. We need to update this in the **release-1.13** branch, and only there. The build will be cut from that branch
2. We need to update only here: https://github.com/dapr/dapr/blob/aad43e4337a4ab6c28cad81e1798f315f68ba659/.github/workflows/dapr.yml#L52-L56 we do not need to update it anywhere else, as the other places do not influence the binary that is built

mikeee:
> 2. We need to update only here: https://github.com/dapr/dapr/blob/aad43e4337a4ab6c28cad81e1798f315f68ba659/.github/workflows/dapr.yml#L52-L56
>     we do not need to update it anywhere else, as the other places do not influence the binary that is built

I think all the references updated are actually build steps e.g.
https://github.com/dapr/dapr/blob/a32f9615b50eca296519fbb8490b589397e2e211/.github/workflows/kind-e2e.yaml#L93-L97

ItalyPaleAle:
> > 2. We need to update only here: https://github.com/dapr/dapr/blob/aad43e4337a4ab6c28cad81e1798f315f68ba659/.github/workflows/dapr.yml#L52-L56
> >    
> >    we do not need to update it anywhere else, as the other places do not influence the binary that is built
> 
> I think all the references updated are actually build steps e.g.
> 
> https://github.com/dapr/dapr/blob/a32f9615b50eca296519fbb8490b589397e2e211/.github/workflows/kind-e2e.yaml#L93-L97

Yes there may be other refgerecnes, you're right, but the only file that matters is `.github/workflows/dapr.yml`

Other files don't need changing as they don't impact the build that we ship

PR #7612, opened 2024-03-09
feat: optionally disable increasedCardinality in helm chart

jfreeland:
# Description

This is an attempt to provide users of the helm chart the option to toggle off `increasedCardinality` per https://docs.dapr.io/operations/observability/metrics/metrics-overview/#high-cardinality-metrics.

I've recently run into this being an actual problem and need this feature myself.  I can just use my own local chart for now but would appreciate seeing this added to the main chart.

## Issue reference

Please reference the issue this PR will close: n/a

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing


jfreeland:
I also opened https://github.com/dapr/dapr/issues/7613 as a potential area for improvement.

jfreeland:
And I don't like the idea of using `{{ .Values.global.prometheus.enabled }}` to indicate whether or not metrics should be enabled.  I suppose that could be an independent value, but it seems strange that you could potentially have `metrics.enabled: false` and `prometheus.enabled: true`.

charts/dapr/charts/dapr_config/templates/dapr_default_config.yaml
```diff
@@ -9,6 +9,14 @@ metadata:
     {{ $key }}: {{ tpl $value $ }}
     {{- end }}
 spec:
+  metrics:
+    enabled: {{ .Values.global.prometheus.enabled }}
```
jfreeland, line 13: I don't like the idea of using `{{ .Values.global.prometheus.enabled }}` to indicate whether or not metrics should be enabled. I suppose that could be an independent value, but it seems strange that you could potentially have `metrics.enabled: false` and `prometheus.enabled: true`.

yaron2:
I don't think this change will apply globally, as this configuration setting is done on a per-namespace configuration object that needs to be explicitly referenced from an app. @ItalyPaleAle for more info on whether the Dapr operator takes the dapr-system configuration into account for this field.

ItalyPaleAle:
> I don't think this change will apply globally, as this configuration setting is done on a per-namespace configuration object that needs to be explicitly referenced from an app. @ItalyPaleAle for more info on whether the Dapr operator takes the dapr-system configuration into account for this field.

I think you're right, @yaron2 . This currently comes from a Configuration that is dynamically loaded, and not from the "daprsystem" Configuration.

We _could_ look into an option to set a per-cluster default, but I am not sure if this is worth it given that low cardinality will be the default in 1.14.

jfreeland:
I'm just starting to dig around today and noticed these other oddities as well.

https://github.com/dapr/dapr/blob/master/charts/dapr/crds/configuration.yaml#L244-L305

It's not clear to me what the difference in `metric:` vs. `metrics:` is supposed to be, but the description for `increasedCardinality` differs between the two entries.

ItalyPaleAle:
> I'm just starting to dig around today and noticed these other oddities as well.
> 
> https://github.com/dapr/dapr/blob/master/charts/dapr/crds/configuration.yaml#L244-L305
> 
> It's not clear to me what the difference in `metric:` vs. `metrics:` is supposed to be, but the description for `increasedCardinality` differs between the two entries.

Oh, that I can explain.

`metric` was the old name for `metrics`: #5808

As for why the description is different, probably a copy-paste error on my side: I must have updated it on one and not the other (our CRDs are not exactly auto-generated: the auto-generated ones end up in different files and we need to copy/paste the changes manually - not sure why)

jfreeland:
Confirmed this works at the application level.  I opened and linked a feature request in case it's something the team wants to pursue at a later date.

PR #7616, opened 2024-03-11
Disk Loader: Scope filtering

JoshVanL:
Adds scope filtering to the disk manifest loader.

Though this doesn't have any security benefits, it does mean manifests
are filtered much sooner in the runtime when running in self-hosted mode
and prevents logging/errors downstream in modules like the hot reloading
reconciler.

Branched from https://github.com/dapr/dapr/pull/7579


yaron2:
Please resolve conflicts

JoshVanL:
@yaron2 done :+1:

yaron2 (approved):

PR #7623, opened 2024-03-13
Test Integration: reserve ports in ordered blocks

JoshVanL:
Dapr opens a lot of ports. Integration tests use upwards of a 1000 ports through a run. Tests are run in parallel and ports are reserved on demand through port 0 then being released before being used by a test. There was a high chance that a port would attempted to be used by multiple tests/processes at the same time causing conflict errors and false test failures.

To prevent these conflicts, the ports utils has been moved to a process which will reserve 300 blocks of ports to be used by tests. These ports are reserved by incrementing through port ranges 1024 to 65535 in a ring. This always prevents conflicts on this test runner, and should also decrease port conflicts across multiple simultaneous test runners.


tests/integration/framework/process/placement/placement.go
```diff
@@ -61,15 +62,16 @@ func New(t *testing.T, fopts ...Option) *Placement {
 	uid, err := uuid.NewUUID()
 	require.NoError(t, err)
 
-	fp := util.ReservePorts(t, 4)
+	fp := ports.Reserve(t, 4)
+	intIP := fp.Port(t)
```
cicoyle, line 66:
```suggestion
	port := fp.Port(t)
```
Can this just be `port` for consistency sake, you have port further down for the slowappstart example. intIP isn't as self explanatory and its only used twice

tests/integration/suite/placement/quorum/insecure.go
```diff
@@ -66,9 +67,8 @@ func (i *insecure) Setup(t *testing.T) []framework.Option {
 		placement.New(t, append(opts, placement.WithID("p3"))...),
 	}
 
-	fp.Free(t)
```
cicoyle, line 69: We no longer need this?

JoshVanL: No, as it is now a proper process, this will happen during `Run`.

tests/integration/framework/process/placement/placement.go
JoshVanL: :+1:

cicoyle (approved):

yaron2 (approved):

Issue #7629, opened 2024-03-14
input binding incorrectly sends unimplemented events to apps when direction is defined


Syuparn:
## In what area(s)?

> /area runtime

## What version of Dapr?

> 1.13.0

## Expected Behavior

Input binding sends only events that the app subscribes.
For example, if an app `myapp` only subscribes binding `myevent`, events other than `myevent` is not sent to `myapp`.

## Actual Behavior

The Dapr sidecar of `myapp` unexpectedly receives unsubscribed events.

```
$ kubectl logs myapp-7c789fb77f-kqkvq daprd | tail -n 1
time="2024-03-14T11:17:17.950372149Z" level=debug msg="error from app consumer for binding [otherevent]: error invoking app: rpc error: code = Unknown desc = binding not implemented: otherevent" app_id=myapp instance=myapp-7c789fb77f-kqkvq scope=dapr.runtime.processor.binding type=log ver=1.13.0
```

## Steps to Reproduce the Problem

I used a sample project to reproduce this issue.
https://github.com/Syuparn/dapr-binding-sample/tree/two-bindings

- prepare a Kubernetes cluster
- deploy Dapr and components `myevent`, `otherevent`
    - components define `direction` explicitly

```yaml
  - name: direction 
    value: "input,output"
```

- deploy an app `myapp`, which implements `myevent`
- request an event of `otherevent`
    - `kubectl port-forward svc/myapp 3500:3500`
    - `curl -XPOST -i http://localhost:3500/v1.0/bindings/otherevent -d '{"data": {}, "operation": "create"}'`
- Dapr sidecar of `myapp` receives the event and displays the log above

If the direction is not defined, this does not happen.

## Release Note

RELEASE NOTE: **FIX** input binding subscription when direction is defined

Syuparn:
Currently, binding is subscribed if the app implements it or the binding defines the input direction. I suppose it should be subscribed if the app implements it **and** it defines the input direction.

https://github.com/dapr/dapr/blob/ed56f692122d0afef7c79147f77eb283acb10467/pkg/runtime/processor/binding/send.go#L90-L99

Syuparn:
This is an expected behavior. When directions are explicitly defined, scopes should also be defined.

https://github.com/dapr/dapr/pull/7634#issuecomment-2026588134

PR #7634, opened 2024-03-18
FIX input binding subscription when direction is defined

Syuparn:
# Description

Fix input binding to send only events that the app subscribes when direction is explicitly defined.

## Issue reference

Please reference the issue this PR will close: #7629

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Unit tests passing


JoshVanL:
Thanks @Syuparn! Please can you add integration tests for this change?
https://github.com/dapr/dapr/tree/master/tests/integration

yaron2:
@Syuparn the purpose of the input direction is to stop Dapr from sending an OPTIONS request to the app to check if the path is available, since this is explicitly telling Dapr that the app is expected to have a valid endpoint.

Syuparn:
@yaron2 
> since this is explicitly telling Dapr that the app is expected to have a valid endpoint.

Does it mean that all apps in the cluster should be able to subscribe to the binding component when the input direction is specified in the component?

(If any app does not subscribe to it, the binding event may be sent to the app and an error is raised.)

yaron2:
> @yaron2
> 
> > since this is explicitly telling Dapr that the app is expected to have a valid endpoint.
> 
> Does it mean that all apps in the cluster should be able to subscribe to the binding component when the input direction is specified in the component?
> 
> (If any app does not subscribe to it, the binding event may be sent to the app and an error is raised.)

You can use component scopes to limit the binding only to apps that are subscribing: https://docs.dapr.io/operations/components/component-scopes/

Syuparn:
@yaron2 
> You can use component scopes to limit the binding only to apps that are subscribing: https://docs.dapr.io/operations/components/component-scopes/

Thank you for the information! When scopes are specified, events are sent to the expected app. I understand the input/output direction works as expected.

PR #7643, opened 2024-03-22
Fixed Actor metadata is stored under 'actors||actortype' partitionkey

markbosch:
# Description

Changed that the Actor metadata is stored under correct partitionkey via 'constructCompositionKey' instead of 'calculatePartitionKey' which is used for the reminders partitions.

This results in that the Actor metadata can be found and updated when a reminder gets stored or deleted, otherwise this will create a new metadata record with incorrect metadata configuration like 'partitionCount'.

## Issue reference

Please reference the issue this PR will close: 3380 (components-contrib)

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests
* [x] Unit tests passing


markbosch:
Small comment to elaborate on the PR:

I did some investigations regarding the issue [3380](https://github.com/dapr/components-contrib/issues/3380) which is reported to the compontents-contrib and I compared the behaviour in v1.10.4 (where it was working) and where it got broken (versions above that) and it turns out that when you create a reminder it will do a 'getActorMetadata' request based with a partitionKey 'constructCompositeKey("actors", actorType)', but this key is not created during the initial 'migrate' which creates the actor metadata in the fist place. During the initial creating of the actor metadata the partitionKey is based on the metadata.ID which is a GUID and passed through the transaction as a partitionKey.

This results in that the metadata can't be found when you call 'storeReminder' or 'deleteReminder' and it will create a new actor metadata record with a incorrect configuration e.g. partitionKey 0.

The fix is that it will execute two requests (transactions) instead of one with different metadata on the transactions. I can imaging this is maybe not what you want, but I don't see any way on how to distinguish between reminder partitionKeys and actor metadata partitionKeys in one transaction. Maybe somebody can point me to that?

ItalyPaleAle:
> The fix is that it will execute two requests (transactions) instead of one with different metadata on the transactions.

I don't think we can do this sadly, because at this stage it's not a single transaction, so if Dapr crashed half-way we'd have inconsistent state (or even if more than 1 Dapr sidecar is trying to write the same data)

markbosch:
> > The fix is that it will execute two requests (transactions) instead of one with different metadata on the transactions.
> 
> I don't think we can do this sadly, because at this stage it's not a single transaction, so if Dapr crashed half-way we'd have inconsistent state (or even if more than 1 Dapr sidecar is trying to write the same data)

@ItalyPaleAle thanks for your reply. 

Had the same thought indeed and did some digging into the `state`  implementation of **cosmos** in the components-contrib. 

It's unfortunately not allowed to execute a `transaction` with different `partitionKey`s, so the `actor-metadata` and the `reminders` needs to be created with a 'predictable' `partitionKey` which can always be reconstructed. Otherwise the `getActorTypeMetadata` will always fail, because that is retrieving data based on the following `partitionKey`: `metadataPartitionKey: constructCompositeKey("actors", actorType),`

So, if I change that (use the `constructCompositeKey(...)`) it will work, but then it makes me wonder why this `databasePartitionKey` is calculated?

```golang
// Get the database partition key (needed for CosmosDB)
databasePartitionKey := actorMetadata.calculateDatabasePartitionKey(stateKey)
```

Do you know if there is any reason why there is chosen for a GUID as a partitionKey? Or would be `constructCompositeKey` also be fine? Or any other thoughts on this?

Or,

Is there a way to obtain the new de facto `metadataID`, which is stated in the following comment:

```golang
// Also create a request to save the new metadata, so the new "metadataID" becomes the new de facto referenced list for reminders
stateOperations[len(stateOperations)-1] = r.saveActorTypeMetadataRequest(actorType, actorMetadata, stateMetadata)
```

so that it can be used in the `getActorTypeMetadata`

pkg/actors/reminders/reminders.go
```diff
@@ -969,14 +989,27 @@ func (r *reminders) migrateRemindersForActorType(ctx context.Context, storeName
 	}
 
 	// Also create a request to save the new metadata, so the new "metadataID" becomes the new de facto referenced list for reminders
-	stateOperations[len(stateOperations)-1] = r.saveActorTypeMetadataRequest(actorType, actorMetadata, stateMetadata)
+	//stateOperations[len(stateOperations)-1] = r.saveActorTypeMetadataRequest(actorType, actorMetadata, stateMetadata)
```
cicoyle, line 992:
```suggestion
```

yaron2:
@markbosch please resolve the conflicts and we can look into merging this. Thanks!

markbosch:
> @markbosch please resolve the conflicts and we can look into merging this. Thanks!

@yaron2 merge conflict is resolved

ItalyPaleAle:
I am still not sure we should move forward with this, as it does perform 2 separate transactions, so it could lead to data corruption. I understand the problem this is trying to solve, which is a really valid one, but I don't think there's a solution given the current architecture.

Issue #7641, opened 2024-03-20
[observability]:  allow configurable bucket settings for distributions


jfreeland:
## In what area(s)?

/area runtime

> /area operator

> /area placement

> /area docs

> /area test-and-release

/area observability

## Describe the feature
https://github.com/dapr/dapr/blob/d3df68746dd1ae98ba6eaf7cc9a4b5e1a82af5fa/pkg/diagnostics/http_monitoring.go#L43-L44

Some people might consider this to be an excessive number of buckets, and it introduces a tremendous amount of unnecessary cardinality.

I'm already setting https://docs.dapr.io/operations/observability/metrics/metrics-overview/#high-cardinality-metrics to false, but 34 different buckets, per app, per method, per namespace, per service, per status.  It's a lot.  I care about like
`5, 50, 500, 5000`

Please make this configurable by the user.

## Release Note

RELEASE NOTE:

ItalyPaleAle:
I'm not a maintainer but I do think this is a good idea.

Would you be interested in opening a PR?

jfreeland:
> I'm not a maintainer but I do think this is a good idea.

Doh, my bad. :)

> Would you be interested in opening a PR?

I can probably get around to it this weekend.

ItalyPaleAle:
> I can probably get around to it this weekend.

Awesome :)

Open to other options, but I would probably suggest adding it to the Configuration CRD: https://github.com/dapr/dapr/blob/master/pkg/apis/configuration/v1alpha1/types.go

Then you can run `make code-generate` to have the CRDs generated from that. Note the command puts the YAMLs in a different folder, so you'll have to manually copy-paste the changes to the https://github.com/dapr/dapr/blob/master/charts/dapr/crds/configuration.yaml file

Also, the `types.go` file above is used for code generation, but then the configuration sent to the app is copied to this struct: https://github.com/dapr/dapr/blob/master/pkg/config/configuration.go

jfreeland:
I updated the PR a little bit tonight.  Probably need some feedback from a maintainer before I spend any more time on it.

PR #7646, opened 2024-03-24
feat(observability)(WIP): allow custom latency distribution buckets

jfreeland:
# Description

Work In Progress.

This is intended to address #7641 and add support for users to provide an alternative list of distributions thresholds.

No tests.  Not entirely sure this is the best approach.  Not validated at all yet.

## Issue reference

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing


yaron2:
@jfreeland are you still working on this?

jfreeland:
I think there's still a lot of work to do here but I think this is 'closer'.  I didn't get a chance to dig around the `Makefile` and reverse engineer the quickest path to building my own containers so that I could test this out in my development environment.

I seem to recall a couple months ago that I regenerated the generated files myself but I didn't do that tonight.  I don't recall seeing that I needed to in `CONTRIBUTING.md`, but maybe I wasn't looking in the right place.

It might be cool if there was some simple app that could be used for local testing with https://tilt.dev like https://github.com/hashicorp-demoapp/hashicups-setups.

I also spotted a little typo https://github.com/dapr/dapr/blob/master/pkg/diagnostics/grpc_monitoring.go#L57 `healthProbeRoundripLatency` should be `Roundtrip` along the way that bothered me enough to make a note.  Probably worthy of a separate PR just to ensure consistency and searchability.

jfreeland:
> I seem to recall a couple months ago that I regenerated the generated files myself but I didn't do that tonight. I don't recall seeing that I needed to in `CONTRIBUTING.md`, but maybe I wasn't looking in the right place.

I found @ItalyPaleAle 's [reference to `make code-generate`](https://github.com/dapr/dapr/issues/7641#issuecomment-2010569725) and ran it again but there was no diff.  I updated `configuration.yaml` by hand in this PR.  That was probably not correct.

jfreeland:
this is not ready to rerun tests yet.  please hold.

jfreeland:
Update...

Some of the unit test errors were because I was missing the [new-ish `statusKey`](https://github.com/jfreeland/dapr/blame/8519d226924000dab0793bdecbc8191cea10a0a5/pkg/diagnostics/component_monitoring.go#L170).  Maybe that's also related to integration test failures but I didn't quickly determine that.

I haven't yet tracked down what's causing the `runtime/service_invocation/res_recv_latency_ms: cannot register view "runtime/service_invocation/res_recv_latency_ms"; a different view with the same name is already registered` errors.  I removed this from `handler_test.go` in https://github.com/dapr/dapr/pull/7646/commits/d5c590a1cf51b04ce35dbfc59dbf01d1da26ec88 but that was actually probably the wrong thing to do.  I believe there's something about these tests running in parallel that was working fine before but no longer is the case after the distribution bucket changes, which is not what I expected.  Now that I've addressed the `statusKey` change I've got more `cannot register view` errors in e.g. https://github.com/jfreeland/dapr/blob/8519d226924000dab0793bdecbc8191cea10a0a5/pkg/diagnostics/resiliency_monitoring_test.go#L188.

PR #7654, opened 2024-03-27
Remove ActorScanInterval

ItalyPaleAle:
Remove the "ActorScanInterval" configuration from actors which is harmful. Idle actors are now deactivated automatically shortly after the idle timeout for their actor type.

We have been using this patch in production for many months and saw various improvements and no issues.

Benefits:


pkg/actors/actors_idleprocessor.go
```diff
@@ -0,0 +1,42 @@
+/*
```
JoshVanL, line 1: Nit: rename file to `idleprocessor.go`.

pkg/actors/actors_idleprocessor.go
```diff
@@ -0,0 +1,42 @@
...
+func (a *actorsRuntime) idleProcessorExecuteFn(act *actor) {
+	// This function is outlined for testing
+	if !a.idleActorBusyCheck(act) {
+		return
+	}
+
+	// Proceed with deactivating the actor
+	err := a.deactivateActor(act)
+	if err != nil {
+		log.Errorf("Failed to deactivate actor %s: %v", act.Key(), err)
```
JoshVanL, line 29: Question: what are the real failure modes where this error occurs, and what is result of just logging the error?

JoshVanL (changes requested):
Great change to me! Just a nit and question. Another question- it is correct that the scan interval was only ever configurable via the config app endpoint?

Please can you create a PR for the [docs](https://docs.dapr.io/developing-applications/building-blocks/actors/actors-runtime-config/)? We also need to create issues in the SDKs to surface deprecation for this option in the appropriate way for each language.

pkg/actors/actors_idleprocessor.go
ItalyPaleAle: You can take a look at `deactivateActor` for what the possible failures are, and it depends on the actor backend used.

Since this method is running in a background goroutine, there's not much use in cascading the error, we can just log it to indicate _something_ is wrong with the actor subsystem.

cicoyle:
Given we are facing an actor reminder registration perf regression currently, I'd like to request this PR gets perf tested before being merged to ensure we don't add a regression.

yaron2 (approved):

yaron2:
Great PR, @ItalyPaleAle please update the docs now this is merged

Issue #7653, opened 2024-03-27
blocked-duration-in-seconds does not seem to work


virgilp:
## In what area(s)?

 /area runtime

## What version of Dapr?

> 1.13.0

## Expected Behavior

When I restart the deployment while a message is actively being processed, I am expecting daprd to block shutdown for the specified period & keep on processing the active message. 

## Actual Behavior

While shutdown sequence is indeed blocked (and I can e.g. still send messages using the sidecar), the currently-processed message is imediatelly/unilaterally cancelled.  This is a bug, since it leads to duplicate "full" processing of the same message (if my app, attempting to do graceful shutdown, actually finishes the processing of the message before shutting down). 
This is how the sidecar logs look like:

```
{"app_id":"my-app","instance":"my-app-6b94bcfbdc-bbxdm","level":"info","msg":"Received signal 'terminated'; beginning shutdown","scope":"dapr.signals","time":"2024-03-27T10:57:23.821429774Z","type":"log","ver":"1.13.0"}
{"app_id":"my-app","instance":"my-app-6b94bcfbdc-bbxdm","level":"info","msg":"Blocking graceful shutdown for 1h1m40s or until app reports unhealthy...","scope":"dapr.runtime","time":"2024-03-27T10:57:23.821530477Z","type":"log","ver":"1.13.0"}
{"app_id":"my-app","instance":"my-app-6b94bcfbdc-bbxdm","level":"warning","msg":"encountered a retriable error while publishing a subscribed message to topic my-topic, err: Post \"http://127.0.0.1:7187/internalApi/myTopic/myTopicRequest\": context canceled","scope":"dapr.runtime.processor.pubsub","time":"2024-03-27T10:57:23.821942387Z","type":"log","ver":"1.13.0"}
```

Daprd should not cancel the active message handler, it defeats the purpose of the "graceful shutdown"/ "block shutdown". 

## Steps to Reproduce the Problem
It's described above - just try to restart a deployment while a message is being processed, notice the sidecar immediately giving up on it.  See also discussion in https://discord.com/channels/778680217417809931/901141467840524289/threads/1222501189829263491

My setup is:
- C# / .NET8 app processing messages. ServiceBus Topics configured as pubsub component, maxActiveMessages 1, maxConcurrentHandlers 1. Generous (1day) handlerTimeoutInSec & defaultMessageTimeToLiveInSec

Pod annotations/ configuration:
```
Labels:
                           dapr.io/app-id=my-app
[...]

PR #7657, opened 2024-03-27
Pubsub: Execute in-flight subscribed messages in background context.

JoshVanL:
When block shutdown is enabled, Dapr will block the shutdown sequence for a configurable amount of time or until the app becomes unhealthy. During this time the out-bound Dapr APIs remain available but all in-bound requests are blocked. Currently, in-flight PubSub messages are canceled when Dapr receives a SIGTERM, which can result in lost messages.

Patch updates the pubsub subscribe message handler policy runner to execute sending the message to the app channel in a background context. This ensures that when Dapr cancels the Subscription due to a blocking shutdown, the in-flight message is not effected.

Enables the `dapr_component_pubsub_egress_bulk_count` and `dapr_component_pubsub_egress_bulk_latency` metrics for bulk pubsub egress which were previously not enabled, as they are used to test that an in-flight message is correctly sent after Dapr has begun to shutdown.

fixes https://github.com/dapr/dapr/issues/7653

### DIFF
--- a/pkg/diagnostics/component_monitoring.go
+++ b/pkg/diagnostics/component_monitoring.go
@@ -174,6 +174,8 @@ func (c *componentMetrics) Init(appID, namespace string) error {
 		diagUtils.NewMeasureView(c.bulkPubsubEventIngressCount, []tag.Key{appIDKey, componentKey, namespaceKey, processStatusKey, topicKey}, view.Count()),
 		diagUtils.NewMeasureView(c.pubsubEgressLatency, []tag.Key{appIDKey, componentKey, namespaceKey, successKey, topicKey}, defaultLatencyDistribution),
 		diagUtils.NewMeasureView(c.pubsubEgressCount, []tag.Key{appIDKey, componentKey, namespaceKey, successKey, topicKey}, view.Count()),
+		diagUtils.NewMeasureView(c.bulkPubsubEgressLatency, []tag.Key{appIDKey, componentKey, namespaceKey, successKey, topicKey}, defaultLatencyDistribution),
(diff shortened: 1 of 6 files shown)


mukundansundar (approved):
If as it is [described](https://github.com/dapr/dapr/pull/7657#issue-2212016088) that there is a chance of messages being lost, I believe this should be a patch in 1.13, 1.12 and 1.11 if the code paths are same, since it's in a stable API.
cc @dapr/maintainers-dapr thoughts?

yaron2:
> If as it is [described](https://github.com/dapr/dapr/pull/7657#issue-2212016088) that there is a chance of messages being lost, I believe this should be a patch in 1.13, 1.12 and 1.11 if the code paths are same, since it's in a stable API. cc @dapr/maintainers-dapr thoughts?

Messages aren't lost, they are simply retried at the broker level and this is still guaranteed at least once delivery

virgilp:
@yaron2 there's a small wart: on my tests, the next message (that I'm not sure I posted) was this:
```
{"app_id":"my-app","instance":"my-app-6b94bcfbdc-bbxdm","level":"error","msg":"error sending message to dead letter, origin topic: my-topic dead letter topic deadLetter-my-topic err: %!w(pubsub.NotAllowedError={deadLetter-my-topic my-app})","scope":"dapr.runtime.processor.pubsub","time":"2024-03-27T10:57:23.821993889Z","type":"log","ver":"1.13.0"}
```
It seems that dapr immediately tries to dead-letter the message (even though it's configured to retry 10 times). Now, it fails to do so, probably due to misconfiguration on my side (I don't think I gave it publish rights on the deadletter topic) BUT, had it succeeded in dead-lettering the message, that's a fairly significant contract breach (even if the broker itself retries).

PR #7662, opened 2024-03-28
[1.13] store legacy `dapr_http_server_response_count` HTTP metric

JoshVanL:
Backport of https://github.com/dapr/dapr/pull/7656


tests/integration/suite/daprd/metrics/http/basic.go
```diff
@@ -0,0 +1,78 @@
...
+	})
+
+	t.Run("state stores", func(t *testing.T) {
+		body := `[{"key":"myvalue", "value":"hello world"}]`
+		b.daprd.HTTPPost2xx(t, ctx, "/v1.0/state/mystore", strings.NewReader(body), "content-type", "application/json")
+
+		b.daprd.HTTPGet2xx(t, ctx, "/v1.0/state/mystore/myvalue")
+
+		metrics := b.daprd.Metrics(t, ctx)
+		assert.Equal(t, 1, int(metrics["dapr_http_server_request_count|app_id:myapp|method:POST|path:/v1.0/state/mystore|status:204"]))
```
filintod, line 75: should there be a dapr_http_server_response_count checke here?

JoshVanL: This metric is part of the high cardinality metric group which is disabled by default for 1.13, hence why it isn't present here.

filintod: was wondering how we tested we did not miss some as this one before

PR #7664, opened 2024-03-31
Test Integration: hardcode binary tmp for darwin

JoshVanL:
On Darwin (MacOS), `os.TempDir()` nicely returns a truly ephemeral process directory is `/var/folders/...`, however in order to reap the benefits of Go cache magic when re-building binaries between integration test runs we need a slightly more permanent directory, i.e. `/tmp`. This directory will persist across integration test runs and speed up the test execution boot times from s to ms.

Updates integration test binary builder to use`/tmp` as the root dir when running on darwin.

### DIFF
--- a/tests/integration/framework/binary/binary.go
+++ b/tests/integration/framework/binary/binary.go
@@ -60,7 +60,13 @@ func Build(t *testing.T, name string) {
 
 		// Use a consistent temp dir for the binary so that the binary is cached on
 		// subsequent runs.
-		binPath := filepath.Join(os.TempDir(), "dapr_integration_tests/"+name)
+		var tmpdir string
+		if runtime.GOOS == "darwin" {
+			tmpdir = "/tmp"
+		} else {
+			tmpdir = os.TempDir()
+		}
+		binPath := filepath.Join(tmpdir, "dapr_integration_tests/"+name)
 		if runtime.GOOS == "windows" {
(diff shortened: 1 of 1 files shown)


artursouza (approved):

tests/integration/framework/binary/binary.go
```diff
@@ -60,7 +60,13 @@ func Build(t *testing.T, name string) {
 
 		// Use a consistent temp dir for the binary so that the binary is cached on
 		// subsequent runs.
-		binPath := filepath.Join(os.TempDir(), "dapr_integration_tests/"+name)
+		var tmpdir string
+		if runtime.GOOS == "darwin" {
+			tmpdir = "/tmp"
+		} else {
+			tmpdir = os.TempDir()
```
daixiang0, line 67: Good catch!

daixiang0: Is it a design issue for Go os package? Maybe we can submit an issue.

daixiang0 (approved):
LGTM

tests/integration/framework/binary/binary.go
JoshVanL: I actually think Go is doing the more correct thing here as a sane default, however because we want a slightly more persistent target directory we need to divert from the default.

Issue #7665, opened 2024-04-02
HA spread of pods on single zone deployments not preferring different nodes


filintod:
## In what area(s)?

> /area runtime
Chart deployment HA antiaffinity only if multiple zones available

> /area operator
Chart deployment HA antiaffinity only if multiple zones available

> /area placement
Chart deployment HA antiaffinity only if multiple zones available

## What version of Dapr?

1.13, 1.12

## Expected Behavior

If we have a single zone, pods should be tried to be spread on different nodes for the same deployment/statefulset

## Actual Behavior

There is no anti-affinity that takes into account the node names

## Steps to Reproduce the Problem

Set system with HA enable and 3 replicas, see there is no soft/hard antiaffinity with `topologyKey: "kubernetes.io/hostname"`

## Release Note

RELEASE NOTE:

olitomlinson:
@yaron2 can we re-open this, seems like a reasonable ask

PR #7666, opened 2024-04-02
Try spread pods in HA to different nodes on single zone deployments

filintod:
# Description

Add soft antiaffinity with topology key `kubernetes.io/hostname` with a weight of 50 to still give more priority to zone antiaffinity if multiple zones available

## Issue reference

Please reference the issue this PR will close: https://github.com/dapr/dapr/issues/7665

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing


JoshVanL:
@filintod, the kube-scheduler already scores based on Node locality for a replica set. Setting this value to me seems like we are just adding more processing time to the scheduling process. Same with the exiting zone affinity rule- do we actually need this?

filintod:
@JoshVanL was seeing some pods for the same dapr system to be running on the same nodes in some customers, even though it could also be there was a single node, but seem strange for prod systems.   Where do you see this default node anti-affinity for replicasets, as that was not expected in the past, not sure nowadays.

JoshVanL:
@filintod I think we should do some experimenting with also removing the zone affinity rules. Scheduler also takes this into account by default. By adding custom rules we are causing changes to the default scoring which might be having unintended consequences and fighting sane defaults.

I would have thought having better pod priority to be the thing to focus on to ensure uptime of the control plane.

filintod:
one thing is that you cannot say HA and have things running on the same node

JoshVanL:
I said uptime not HA 🙂 I see these as two separate things where uptime is a subset property of achieving HA, but HA also incorporating the idea of replication.

The Dapr control plane is not sensitive to churn like an application serving business traffic needing network failover would be. Single replica Dapr control plane can handle plenty, it just needs to be up, somewhere, and needs have higher priority of being up over consuming Dapr apps.

filintod:
yes, I meant for `you cannot say` in term of `dapr cannot say` as we have here https://github.com/dapr/dapr/blob/master/charts/dapr/values.yaml#L29 `HA`  and people might expect fault tolerance and spreading on different nodes/az is part of it. 

Uptime is for sure important and we should find ways to make these services increase in priority for when one of them is going to be kick out to not be the one of the last in the list. but probably different to this PR.

filintod:
btw, actually thought I had seen the replica set node locality you mentioned but actually it was different.  https://kubernetes.io/docs/reference/scheduling/config/ . So lmk if you can point to me where you see it. From the default ones, you could get balance if all nodes have somewhat balanced load allocatable, but overtime that is usually not feasible without some sort of rebalancing (that there are tools around). If that is the case, we do need to have antiaffinity to tilt the balance on not having all pods on the same node if possible.

daixiang0 (approved):

PR #7675, opened 2024-04-04
Adds `listen-address` CLI flags for all services servers.

JoshVanL:
Address corresponding `--listen-address` CLI flags for all services servers. This sets the listening address for these TCP servers.

This is added to allow users to both bind the address to localhost to prevent access to a particular server on the network, as well as for integration testing as it is more correct to bind to localhost and prevents triggering MacOS firewall popups.


cmd/daprd/options/options.go
```diff
@@ -117,8 +119,10 @@ func New(origArgs []string) (*Options, error) {
 	fs.StringVar(&opts.DaprHTTPPort, "dapr-http-port", strconv.Itoa(runtime.DefaultDaprHTTPPort), "HTTP port for Dapr API to listen on")
 	fs.StringVar(&opts.DaprAPIListenAddresses, "dapr-listen-addresses", runtime.DefaultAPIListenAddress, "One or more addresses for the Dapr API to listen on, CSV limited")
 	fs.StringVar(&opts.DaprPublicPort, "dapr-public-port", "", "Public port for Dapr Health and Metadata to listen on")
+	fs.StringVar(&opts.DaprPublicListenAddress, "dapr-public-listen-address", "0.0.0.0", "Public listen address for Dapr Health and Metadata")
```
daixiang0, line 122:
```suggestion
	fs.StringVar(&opts.DaprPublicAddress, "dapr-public-address", "0.0.0.0", "Public listen address for Dapr Health and Metadata")
```

cmd/daprd/options/options.go
```diff
@@ -117,8 +119,10 @@ func New(origArgs []string) (*Options, error) {
...
 	fs.StringVar(&opts.DaprAPIGRPCPort, "dapr-grpc-port", strconv.Itoa(runtime.DefaultDaprAPIGRPCPort), "gRPC port for the Dapr API to listen on")
 	fs.StringVar(&opts.DaprInternalGRPCPort, "dapr-internal-grpc-port", "", "gRPC port for the Dapr Internal API to listen on")
+	fs.StringVar(&opts.DaprInternalGRPCListenAddress, "dapr-internal-grpc-listen-address", "0.0.0.0", "gRPC listen address for the Dapr Internal API")
```
daixiang0, line 125:
```suggestion
	fs.StringVar(&opts.DaprInternalGRPCAddress, "dapr-internal-grpc-address", "0.0.0.0", "gRPC listen address for the Dapr Internal API")
```

cmd/injector/app/app.go
```diff
@@ -102,6 +102,7 @@ func Run() {
 
 	inj, err := service.NewInjector(service.Options{
 		Port:                    opts.Port,
+		ListenAddress:           opts.ListenAddress,
```
daixiang0, line 105: Same as above, do not repeat below.

cmd/daprd/options/options.go
cicoyle: I personally like having `Listen` in the name otherwise it's not explicit to developers setting it without looking at the description in this one file or via the cli description.

daixiang0: How about the `DaprPublicPort`, I prefer to keep same.

JoshVanL: I agree- `address` without `listen` gives the impression of a target host. We are already using the `listen` key word in an existing flag so this also keeping consistent. https://github.com/dapr/dapr/blob/8bd7e07a852a4d0727e518563d300bdb5d92a34f/cmd/daprd/options/options.go#L118

pkg/api/http/server.go
```diff
@@ -174,7 +174,7 @@ func (s *server) StartNonBlocking() error {
 		s.setupRoutes(publicR, s.api.PublicEndpoints())
 
 		healthServer := &http.Server{
-			Addr:              fmt.Sprintf(":%d", *s.config.PublicPort),
+			Addr:              fmt.Sprintf("%s:%d", s.config.PublicListenAddress, *s.config.PublicPort),
```
artursouza, line 177: The old code was omitting the listening address while the new code is sending "0.0.0.0" instead. Does it mean that we now will only listen to IPv4 addresses while before it would listen also on IPv6? Can this be validated? I am asking because maybe the default value should probably be empty string to keep backwards compatibility.

artursouza (changes requested):
I have one important question about backwards compatibility.

pkg/api/http/server.go
JoshVanL: Thanks @artursouza, good catch! I've updated the default listen addresses CLI flags to be empty strings which will be the same value as before.

daixiang0 (approved):
LGTM

PR #7676, opened 2024-04-04
refactor(cockroachdb): disallow aws iam auth, as only for pg

sicoyle:
# Description

Needed as a side effect of enabling postgres AWS IAM authentication. We need to explicitly disallow this for cockroach DB as it uses postgres v2 under the hood.

For additional context, see: https://github.com/dapr/components-contrib/pull/3324
Specifically this comment asking me to make these changes: https://github.com/dapr/components-contrib/pull/3324#discussion_r1532362078

I will update the components contrib dependency here as soon as that PR gets merged so this builds and passes fine.

## Issue reference

https://github.com/dapr/components-contrib/issues/3254

## Checklist

* [X] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing

### DIFF
--- a/cmd/daprd/components/state_cockroachdb.go
+++ b/cmd/daprd/components/state_cockroachdb.go
@@ -26,7 +26,7 @@ func init() {
 	stateLoader.DefaultRegistry.RegisterComponentWithVersions("cockroachdb", components.Versioning{
 		Preferred: components.VersionConstructor{
 			// For v2, this component uses the same implementation as the postgres state store
-			Version: "v2", Constructor: pgv2.NewPostgreSQLStateStoreWithAzureAD,
+			Version: "v2", Constructor: pgv2.NewPostgreSQLStateStoreWithTrueOptions,
 		},
 		Others: []components.VersionConstructor{
 			{Version: "v1", Constructor: cockroachdb.New},
--- a/cmd/daprd/components/state_cockroachdb.go
+++ b/cmd/daprd/components/state_cockroachdb.go
@@ -26,7 +26,7 @@ func init() {
 	stateLoader.DefaultRegistry.RegisterComponentWithVersions("cockroachdb", components.Versioning{
 		Preferred: components.VersionConstructor{
 			// For v2, this component uses the same implementation as the postgres state store
-			Version: "v2", Constructor: pgv2.NewPostgreSQLStateStore,
+			Version: "v2", Constructor: pgv2.NewPostgreSQLStateStoreWithAzureAD,
 		},
 		Others: []components.VersionConstructor{
 			{Version: "v1", Constructor: cockroachdb.New},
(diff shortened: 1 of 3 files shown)


cmd/daprd/components/state_cockroachdb.go
```diff
@@ -26,7 +26,7 @@ func init() {
 	stateLoader.DefaultRegistry.RegisterComponentWithVersions("cockroachdb", components.Versioning{
 		Preferred: components.VersionConstructor{
 			// For v2, this component uses the same implementation as the postgres state store
-			Version: "v2", Constructor: pgv2.NewPostgreSQLStateStore,
+			Version: "v2", Constructor: pgv2.NewPostgreSQLStateStoreWithAzureAD,
```
ItalyPaleAle, line 29: `NewPostgreSQLStateStoreWithOptions` and pass option `NoAzureAd` and `NoAWS` (or whatever they were).

CockroachDB doesn't support AAD either.

sicoyle: I made a function called `NewPostgreSQLStateStoreWithTrueOptions` instead for now to address your feedback. I cannot use the `NewPostgreSQLStateStoreWithOptions`. This is an issue in dapr/dapr that reaches beyond the extent of what this PR is intended for, so that should be addressed in a different PR and initiative.

The problem with using `NewPostgreSQLStateStoreWithOptions` as you are recommending is that no other component is using the ...WithOptions function. The dapr/dapr registry is setup to only allow constructors of a single logger parameter. Me using the `NewPostgreSQLStateStoreWithOptions` as you are recommending leads to this panic:
```
panic: constructor for v2 is not a state store

```

Basically the component registry is not set up to support v2 components such as postgres with options support. This would be a high touch effort that I do not want to add to my AWS IAM PR. 

For better context:
```


// Need more changes because of: signature does not allow for options
panic: constructor for v2 is not a state store

goroutine 1 [running]:
github.com/dapr/dapr/pkg/components/state.toConstructor({{0x108c88bf0?, 0x108c92a84?}, {0x10ae82700?, 0x140014daa20?}})
        /Users/samcoyle/go/src/github.com/forks/dapr/dapr/pkg/components/state/registry.go:78 +0x8c
github.com/dapr/dapr/pkg/components/state.(*Registry).RegisterComponentWithVersions(0x10e3ea940, {0x108ca6732, 0xb}, {{{0x108c88bf0, 0x2}, {0x10ae82700, 0x140014daa20}}, {0x0, 0x0, 0x0}, ...})
        /Users/samcoyle/go/src/github.com/forks/dapr/dapr/pkg/components/state/registry.go:63 +0xb8
github.com/dapr/dapr/cmd/daprd/components.init.104()
        /Users/samcoyle/go/src/github.com/forks/dapr/dapr/cmd/daprd/components/state_cockroachdb.go:29 +0x1ac
```

sicoyle: Thoughts @ItalyPaleAle ?

PR #7691, opened 2024-04-11
Secrets: Error standardization - Secrets API

pglushko:
# Description

## Issue reference
#7690 
#7484 

Please reference the issue this PR will close: #7690 

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing


pkg/api/errors/secrets.go
```diff
@@ -0,0 +1,104 @@
...
+	)
+}
+
+func (s *SecretsStoreError) PermissionDenied(key string) error {
+	return s.build(
+		errors.NewBuilder(
+			codes.PermissionDenied,
+			http.StatusForbidden,
+			fmt.Sprintf("access denied by policy to get %q from %q", key, s.name),
+			"ERR_SECRETS_PERMISSION_DENIED",
```
cicoyle, line 55:
```suggestion
			"ERR_PERMISSION_DENIED",
```
We have to keep this as the legacy tag for backwards compatibility

pkg/api/errors/secrets.go
```diff
@@ -0,0 +1,104 @@
...
+
+func (s *SecretsStoreError) GetSecret(key string, error string) error {
+	return s.build(
+		errors.NewBuilder(
+			codes.Internal,
+			http.StatusInternalServerError,
+			fmt.Sprintf("failed getting secret with key %s from secret store %s: %s", key, s.name, error),
+			"ERR_SECRET_GET",
+		),
+		"SECRET_GET",
```
cicoyle, line 81:
```suggestion
		"GET_SECRET",
```
Since we prefix the `reason` with the `errors.CodePrefixSecretStore` in the `build()`, it might be worth changing  it to `SECRET_GET_SECRET` - then its roughly `api_operation_thing`.

pkg/api/errors/secrets.go
```diff
@@ -0,0 +1,104 @@
...
+
+func (s *SecretsStoreError) BulkSecretGet(error string) error {
+	return s.build(
+		errors.NewBuilder(
+			codes.Internal,
+			http.StatusInternalServerError,
+			fmt.Sprintf("failed getting secrets from secret store %s: %v", s.name, error),
+			"ERR_SECRET_GET",
+		),
+		"SECRET_GET",
```
cicoyle, line 93:
```suggestion
		"GET_BULK_SECRET",
```

cicoyle (commented):
Great start @pglushko - thanks for keeping the error codes the same for backwards compatibility 🙂 Next will be writing the integration tests. I don't see a secrets folder in the `integration/framework` folder, so that will be the next step to add in order to write the tests for the errors. Lmk if you need any assistance 🙂

pkg/messages/predefined.go
```diff
@@ -75,13 +75,6 @@ var (
 	ErrOutboundHealthNotReady = APIError{"dapr outbound is not ready", "ERR_OUTBOUND_HEALTH_NOT_READY", http.StatusInternalServerError, grpcCodes.Internal}
 	ErrHealthAppIDNotMatch    = APIError{"dapr app-id does not match", "ERR_HEALTH_APPID_NOT_MATCH", http.StatusInternalServerError, grpcCodes.Internal}
 
-	// Secrets.
-	ErrSecretStoreNotConfigured = APIError{"secret store is not configured", "ERR_SECRET_STORES_NOT_CONFIGURED", http.StatusInternalServerError, grpcCodes.FailedPrecondition}
```
pglushko, line 79: @cicoyle looks like typo here btw. ERR_SECRET_STORES_NOT_CONFIGURED => ERR_SECRET_STORE_NOT_CONFIGURED. Fix it?

pkg/api/errors/secrets.go
pglushko: Makes sense
Should tags be updated accordingly? ERR_SECRET_GET => ERR_GET_SECRET (for other scenarios where applicable also)

pglushko:
> Great start @pglushko - thanks for keeping the error codes the same for backwards compatibility 🙂 Next will be writing the integration tests. I don't see a secrets folder in the `integration/framework` folder, so that will be the next step to add in order to write the tests for the errors. Lmk if you need any assistance 🙂

I took a glance at integration tests and  I would definitively need some guidance there. How should I tackle it?

pkg/messages/predefined.go
cicoyle: We will have to keep the legacy tag as is: `ERR_SECRET_STORES_NOT_CONFIGURED`, the new tag that gets the api prefixed in the `build()` can be whatever makes the most sense.

pkg/api/errors/secrets.go
cicoyle: `"ERR_SECRET_GET"` will have to remain as is since its the legacy tag. The new tag (or errCode) can be updated accordingly, as you have correctly already 🙂

cicoyle:
> > Great start @pglushko - thanks for keeping the error codes the same for backwards compatibility 🙂 Next will be writing the integration tests. I don't see a secrets folder in the `integration/framework` folder, so that will be the next step to add in order to write the tests for the errors. Lmk if you need any assistance 🙂
> 
> I took a glance at integration tests and I would definitively need some guidance there. How should I tackle it?

The integration tests will involve writing a new secret `integration/framework/process` folder and the relevant code inside that such that you can force the errors you are updating in the codebase. If the errors aren't easily forced as is with the integration process, then you'll need to write a pluggable component. Please view the [pubsub](https://github.com/dapr/dapr/pull/7322) & [statestore](https://github.com/dapr/dapr/pull/7257) PRs as examples of how to write those files. You will likely have the following structure:
`integration/framework/process/secrets/` with the following files `component.go`, `options.go`, `secrets.go` with the relevant code in each. Then you'll use the `secrets.go` `New()` to create the secrets process in the `errors.go` and spin up the daprd sidecar with that component, [similar to this](https://github.com/dapr/dapr/pull/7322/files#diff-fcef716f53ca56a7e277239de2b38437cfec31e2a639c2de54894e62c2ce78d4R73-R85).

yaron2:
ping @pglushko

PR #7713, opened 2024-05-01
Bump to Go 1.22 - Pin latest contrib with Go 1.22

berndverst:
# Description

Pin the latest contrib which now uses Go 1.22, as a result use Go 1.22 in Runtime also.

PR #7720, opened 2024-05-07
Makefile: export `go/bin` to PATH for darwin

JoshVanL:
Exports the `go/bin/darwin_$(arch)` directory to the PATH environment
variable when running on Darwin. This ensures that binaries installed by
`go install` are available in the Makefile exec path.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -84,6 +84,7 @@ ifeq ($(LOCAL_OS),Linux)
    TARGET_OS_LOCAL = linux
 else ifeq ($(LOCAL_OS),Darwin)
    TARGET_OS_LOCAL = darwin
+   PATH := $(PATH):$(HOME)/go/bin/darwin_$(GOARCH)
 else
    TARGET_OS_LOCAL = windows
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -84,6 +84,7 @@ ifeq ($(LOCAL_OS),Linux)
    TARGET_OS_LOCAL = linux
 else ifeq ($(LOCAL_OS),Darwin)
    TARGET_OS_LOCAL = darwin
+   PATH := $(PATH):$(HOME)/go/bin/darwin_$(GOARCH)
```
artursouza, line 87: Is this true for everyone? I don't have this folder.

artursouza (changes requested):

JoshVanL:
Close in favor of https://github.com/dapr/dapr/pull/7713

Makefile
JoshVanL: Same for me, but this is the case for the Github CI runner. We are only appending the path so there is no harm.

PR #7729, opened 2024-05-14
Update controller-gen

yaron2:
title


yaron2:
Ref: https://github.com/kubernetes-sigs/kubebuilder/issues/3792#issuecomment-2110937515

Issue #7719, opened 2024-05-07
Low cardinality metrics issues


antontroshin:
## Intro
With the low cardinality metrics enabled, a lot of metrics are losing their value and in some cases even become completely unusable as they do not have any useful information provided except the application ID.
Monitoring systems and dashboards previously using high cardinality settings, are broken and there's seems no suitable replacement can be found with low cardinality enabled.
The value of having a low cardinality setting is understandable and the use-cases differ between the users, however having the option to have a middle ground would be great.
Please consider some scenarios where low cardinality might render metrics unusable (see below).

## Test method
I've done an investigation by building a few simple applications using the code samples mentioned below.
Using simple applications to send/receive HTTP requests using 2 different methods, based on the examples provided in:
https://docs.dapr.io/developing-applications/building-blocks/service-invocation/howto-invoke-discover-services/#invoke-the-service (plain HTTP calls)
https://github.com/dapr/go-sdk/tree/main/examples/service/serving/http (SDK server)
https://github.com/dapr/go-sdk/tree/main/examples/service/client (SDK client)

## Metrics checked

Application: `service-invocation-http-client`

Metrics: `dapr_http_server_latency_` (`_sum`, `_count`, `_bucket`)

High cardinality:
```
dapr_http_server_latency_bucket{app_id="service-invocation-http-client",method="POST",path="/post-service-invocation-plain",status="200",le="100"} 1160
dapr_http_server_latency_bucket{app_id="service-invocation-http-client",method="GET",path="/get-service-invocation-plain",status="500",le="100"} 321
dapr_http_server_latency_bucket{app_id="service-invocation-http-client",method="DELETE",path="/delete-service-invocation-plain",status="500",le="100"} 299
dapr_http_server_latency_bucket{app_id="service-invocation-http-client",method="PUT",path="/put-service-invocation-plain",status="200",le="100"} 1149
```

Low cardinality:
```
dapr_http_server_latency_bucket{app_id="service-invocation-http-client",method="InvokeService/service-invocation-http-server",status="200",le="100"} 3584
```

Metrics: `dapr_http_server_request_count`

High cardinality:
```
dapr_http_server_request_count{app_id="service-invocation-http-client",method="GET",path="/get-service-invocation-plain",status="200"} 1742
dapr_http_server_request_count{app_id="service-invocation-http-client",method="PUT",path="/put-service-invocation-plain",status="200"} 1693
```

Low cardinality:
```
[...]

Issue #6681, opened 2023-07-16
Reduce cost of allocations/processing of adding tags to methods


filintod:
## In what area(s)?

/area runtime

Following discussion on https://github.com/dapr/dapr/pull/6674#issuecomment-1637080556 in regards to frequents allocations and processing to add tags to metrics on hot path, we should find ways to pre-create/allocate these tags and use stats.Record instead of stats.RecordWithOptions.

As an example on a call to actors we see:

![image](https://github.com/dapr/dapr/assets/1373693/0b1b80cc-a91f-4aab-9fdc-e4119ec8260f)

Where a fair amount allocations are dedicated only to the addition of these tags. We also have these inside a set of two nested resiliencies, meaning that when issues occur, we start allocating memory frequently that is pretty much the same all the time and could have been reused.

## Describe the feature

Some tags can also be pre-allocated for longer duration like appID as it is used all the time. Some could be created just before resiliencies and added to the context via tag.New (like in https://github.com/census-instrumentation/opencensus-go/blob/master/stats/record.go#L155) as done in stats.RecordWithOptions.

## Release Note

RELEASE NOTE:

filintod:
using the bug one

PR #7746, opened 2024-05-20
HTTP Metrics Path Matching

nelson-parente:
# Description

This PR implements the proposal https://github.com/dapr/proposals/pull/58. 

```yaml
spec:
  metric:
    enabled: true
    http:
      increasedCardinality: true
      pathMatching:
        ingress:
        - /orders/{orderID}/items/{itemID}
        - /users/{userID}
        egress:
        - /orders
        - /orders/{orderID}/items/{itemID}
```

PathMatching configuration is opt-in and ensures consistent representation of HTTP request paths in server metrics. 

## Issue reference

HTTP Metrics Path Matching Proposal: https://github.com/dapr/proposals/pull/58
Dapr Low Cardinality Issue with initial motivation for the proposal: https://github.com/dapr/dapr/issues/7719

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing


pkg/diagnostics/http_monitoring.go
```diff
@@ -307,3 +365,75 @@ func (h *httpMetrics) HTTPMiddleware(next http.Handler) http.Handler {
 		h.ServerRequestCompleted(r.Context(), method, path, status, reqContentSize, respSize, elapsed)
 	})
 }
+
+func (h *httpMetrics) initPathNormalization(config *config.PathNormalization) pathNormalizationConfig {
+	if config == nil {
```
filintod, line 370: should we also check `config.enabled == false` to return here as well?

filintod: ok. I see you check later.

pkg/diagnostics/utils/path_normalization_utils.go
```diff
@@ -0,0 +1,28 @@
...
+limitations under the License.
+*/
+
+package utils
+
+import "net/http"
+
+type PathNormalizationRW struct {
+	http.ResponseWriter
+	NormalizedPath *string
```
filintod, line 20: why we need this to be a pointer? I did not see a check against nil or something like it that would warrant it, but maybe I missed it. Asking as you already use pointers against the struct.

nelson-parente: I used the pointer to pass by reference instead of passing by value and with that, we can perform the path transformation by modifying the var.

nelson-parente: thanks for the tip! as you pointed out the issue was on the return value. Fixed in this commit: https://github.com/dapr/dapr/pull/7746/commits/8dc24c680ddb1b98d850a101cd258802a06101b4

pkg/apis/configuration/v1alpha1/types.go
```diff
@@ -222,6 +222,17 @@ type MetricHTTP struct {
 	// If false (the default), metrics for the HTTP server are collected with increased cardinality.
 	// +optional
 	IncreasedCardinality *bool `json:"increasedCardinality,omitempty"`
+	// +optional
+	PathNormalization *PathNormalization `json:"pathNormalization,omitempty"`
+}
+
+type PathNormalization struct {
+	// If enabled, the path normalization will be applied to the HTTP server metrics.
+	Enabled *bool `json:"enabled"`
```
ItalyPaleAle, line 231: No need for `*bool` since the default is false, just `bool`

Please don't forget `,omitempty` here and below

pkg/config/configuration.go
```diff
@@ -261,21 +260,41 @@ func (m MetricSpec) GetEnabled() bool {
...
+// GetHTTPPathNormalization returns the path normalization configuration for HTTP metrics
+func (m MetricSpec) GetHTTPPathNormalization() *PathNormalization {
+	if m.HTTP == nil || m.HTTP.PathNormalization == nil {
+		return nil
+	}
+	return &PathNormalization{
+		Enabled:      m.HTTP.PathNormalization.Enabled,
+		IngressPaths: m.HTTP.PathNormalization.IngressPaths,
+		EgressPaths:  m.HTTP.PathNormalization.EgressPaths,
+	}
```
ItalyPaleAle, line 280: Couldn't this be just `return m.HTTP.PathNormalization`? Why creating a new object?

```suggestion
	if m.HTTP == nil {
		return nil
	}
	return m.HTTP.PathNormalization
```

pkg/diagnostics/http_monitoring.go
```diff
@@ -190,13 +235,21 @@ func (h *httpMetrics) ClientRequestCompleted(ctx context.Context, method, path,
 			diagUtils.WithTags(h.clientRoundtripLatency.Name(), appIDKey, h.appID, httpPathKey, h.convertPathToMetricLabel(path), httpMethodKey, method, httpStatusCodeKey, status),
 			h.clientRoundtripLatency.M(elapsed))
 	} else {
+		completedCountTags := diagUtils.WithTags(h.clientCompletedCount.Name(), appIDKey, h.appID, httpStatusCodeKey, status)
+		roundTripLatencyTags := diagUtils.WithTags(h.clientRoundtripLatency.Name(), appIDKey, h.appID, httpStatusCodeKey, status)
+
+		if h.pathNormalization.enabled {
+			completedCountTags = append(completedCountTags, diagUtils.WithTags(h.clientCompletedCount.Name(), httpPathKey, path, httpMethodKey, method)...)
+			roundTripLatencyTags = append(roundTripLatencyTags, diagUtils.WithTags(h.clientRoundtripLatency.Name(), httpPathKey, path, httpMethodKey, method)...)
+		}
```
ItalyPaleAle, line 244: Using `append` causes re-allocations in a hoth path. Best to create 2 separate slices for the enabled and disabled cases

pkg/diagnostics/http_monitoring.go
```diff
@@ -162,15 +191,26 @@ func (h *httpMetrics) ClientRequestStarted(ctx context.Context, method, path str
...
 	if h.legacy {
 		stats.RecordWithTags(
 			ctx,
 			diagUtils.WithTags(h.clientSentBytes.Name(), appIDKey, h.appID, httpPathKey, h.convertPathToMetricLabel(path), httpMethodKey, method),
 			h.clientSentBytes.M(contentSize))
 	} else {
+		tags := diagUtils.WithTags(h.clientSentBytes.Name(), appIDKey, h.appID)
+
+		if h.pathNormalization.enabled {
+			tags = append(tags, diagUtils.WithTags(h.clientSentBytes.Name(), httpPathKey, path, httpMethodKey, method)...)
```
ItalyPaleAle, line 208: Here too, avoid using `append` and just have 2 else's:

- `else if h.pathNormalization.enabled` and `else`

pkg/diagnostics/http_monitoring.go
```diff
@@ -307,3 +365,93 @@ func (h *httpMetrics) HTTPMiddleware(next http.Handler) http.Handler {
...
+
+	if path == "" {
+		return "", false
+	}
+
+	if !strings.HasPrefix(path, "/") {
+		path = "/" + path
+	}
+
+	req, err := http.NewRequest(http.MethodGet, path, nil)
```
ItalyPaleAle, line 443: The proposal said

> [...] normalizing metrics without relying on regex's, which are known to be CPU-intensive.

However here you are:

1. Creating `http.Request` objects, which allocates a lot of resources
2. Invoking a HTTP server
3. Have the HTTP server do routing, which internally most likely relies on some regular expressions

If you do want to do this in a lightweight fashion, you should look into other options that don't involve all this work. This is a very hot path and it should be optimized.

Some ideas could include using a router only, possibly finding one that uses optimized trees and doesn't involve regexs.

ItalyPaleAle (changes requested):
Normally the mainainters would tell you it's a bit premature to do the work before the proposal is even reviewed...!

artursouza:
This proposal was reviewed among maintainers and people got invited to participate in the discussion too. While there is space for discussion, multiple maintainers thought this proposal to be reasonable and low risk to start implementation.

artursouza:
@nelson-parente Please, address the optimization suggestions as this is a hot path for Dapr and can cause a performance regression.

Thanks, @ItalyPaleAle

ItalyPaleAle:
> This proposal was reviewed among some runtime maintainers and people got invited to participate in the discussion too. While there is space for discussion, multiple maintainers thought this proposal to be reasonable and low risk to start implementation.

This is interesting as every time I was in @nelson-parente's shoes, the response I got from maintainers was very different. I have not seen voting happening (to my understanding, you will need the majority of 3 votes - 1 person for each company), and only 1 person commented.

yaron2:
The PR is still pending proposal approval

pkg/diagnostics/http_monitoring.go
JoshVanL: Yep- we should be using `http.Request` to instantiate the object :+1: 

We are not invoking any HTTP server here- we are invoking the mux router to hit a resulting handler. This mux doesn't use `regexp` which the reason why it was chosen.

pkg/diagnostics/http_monitoring.go
nelson-parente: Since the tags are the same as in legacy mode instead of having 2 slices I added `|| h.pathnormalization.enabled` to the if.  Do you agree with this change?

pkg/apis/configuration/v1alpha1/types.go
nelson-parente: I think `*bool` would help us in case we want to change the default value, this way we always fallback to false.

Anyway, I removed it and added a test for the default case when pathNormalization is not in the spec.

pkg/diagnostics/http_monitoring.go
nelson-parente: Changed the instantiation to use http.Request{...}.

As @filintod mentions we are not relying on regex evaluation here.

Given that we are registering the handlers once at startup and not invoking the HTTP server I think the solution using the mux routing primitives does not add much overhead other than the allocation of the request object.

pkg/config/configuration.go
```diff
@@ -273,11 +273,7 @@ func (m MetricSpec) GetHTTPPathNormalization() *PathNormalization {
 	if m.HTTP == nil || m.HTTP.PathNormalization == nil {
```
ItalyPaleAle, line 273: This last nil-check for `m.HTTP.PathNormalization` is redundant

pkg/diagnostics/http_monitoring.go
ItalyPaleAle: As long as it makes sense works for me!

pkg/diagnostics/http_monitoring.go
ItalyPaleAle: Could you run some benchmarks please?

nelson-parente: ofc! I will work on that!

pkg/diagnostics/http_monitoring.go
```diff
@@ -307,3 +346,94 @@ func (h *httpMetrics) HTTPMiddleware(next http.Handler) http.Handler {
...
+
+	if path == "" {
+		return "", false
+	}
+
+	if !strings.HasPrefix(path, "/") {
+		path = "/" + path
+	}
+
+	req := &http.Request{
```
filintod, line 424: you might want to benchmark with `sync.pool` to create this as well, to compare with/without it, as sync.pool might make sense in this use case

pkg/diagnostics/http_monitoring.go
nelson-parente: **Benchmark Results**

All runs were done over this commit https://github.com/dapr/dapr/pull/7746/commits/1fccd567e0cd39e68b64267c14f7f7e49600cd52 (this will be reverted before merge)

_BenchmarkHTTPMiddlewareWithoutPathNormalization_

```
$  go test -benchtime=100x -benchmem -run=^$ -bench ^BenchmarkHTTPMiddlewareWithoutPathNormalization$ github.com/dapr/dapr/pkg/diagnostics
goos: darwin
goarch: arm64
pkg: github.com/dapr/dapr/pkg/diagnostics
BenchmarkHTTPMiddlewareWithoutPathNormalization-14    	     100	   5669265 ns/op	    4865 B/op	      85 allocs/op
PASS
ok  	github.com/dapr/dapr/pkg/diagnostics	0.951s

$ go test -benchtime=500x -benchmem -run=^$ -bench ^BenchmarkHTTPMiddlewareWithoutPathNormalization$ github.com/dapr/dapr/pkg/diagnostics
goos: darwin
goarch: arm64
pkg: github.com/dapr/dapr/pkg/diagnostics
BenchmarkHTTPMiddlewareWithoutPathNormalization-14    	     500	   5676510 ns/op	    4811 B/op	      84 allocs/op
PASS
ok  	github.com/dapr/dapr/pkg/diagnostics	3.162s

$ go test -benchtime=1000x -benchmem -run=^$ -bench ^BenchmarkHTTPMiddlewareWithoutPathNormalization$ github.com/dapr/dapr/pkg/diagnostics
goos: darwin
goarch: arm64
pkg: github.com/dapr/dapr/pkg/diagnostics
BenchmarkHTTPMiddlewareWithoutPathNormalization-14    	    1000	   5653330 ns/op	    4806 B/op	      84 allocs/op
PASS
ok  	github.com/dapr/dapr/pkg/diagnostics	5.963s
```

_BenchmarkHTTPMiddlewareWithPathNormalization_

```
$ go test -benchtime=100x -benchmem -run=^$ -bench ^BenchmarkHTTPMiddlewareWithPathNormalization$ github.com/dapr/dapr/pkg/diagnostics
goos: darwin
goarch: arm64
pkg: github.com/dapr/dapr/pkg/diagnostics
BenchmarkHTTPMiddlewareWithPathNormalization-14    	     100	   5673615 ns/op	    6440 B/op	     114 allocs/op
PASS
ok  	github.com/dapr/dapr/pkg/diagnostics	0.896s

$ go test -benchtime=500x -benchmem -run=^$ -bench ^BenchmarkHTTPMiddlewareWithPathNormalization$ github.com/dapr/dapr/pkg/diagnostics
goos: darwin
goarch: arm64
pkg: github.com/dapr/dapr/pkg/diagnostics
BenchmarkHTTPMiddlewareWithPathNormalization-14    	     500	   5717264 ns/op	    6342 B/op	     112 allocs/op
PASS
ok  	github.com/dapr/dapr/pkg/diagnostics	3.180s

$ go test -benchtime=1000x -benchmem -run=^$ -bench ^BenchmarkHTTPMiddlewareWithPathNormalization$ github.com/dapr/dapr/pkg/diagnostics
goos: darwin
goarch: arm64
pkg: github.com/dapr/dapr/pkg/diagnostics
BenchmarkHTTPMiddlewareWithPathNormalization-14    	    1000	   5652908 ns/op	    6331 B/op	     112 allocs/op
PASS
ok  	github.com/dapr/dapr/pkg/diagnostics	5.976s
```

_BenchmarkHTTPMiddlewareWithPathNormalizationSyncPool_

```
$ go test -benchtime=100x -benchmem -run=^$ -bench ^BenchmarkHTTPMiddlewareWithPathNormalizationSyncPool$ github.com/dapr/dapr/pkg/diagnostics
goos: darwin
goarch: arm64
pkg: github.com/dapr/dapr/pkg/diagnostics
BenchmarkHTTPMiddlewareWithPathNormalizationSyncPool-14    	     100	   5755573 ns/op	    6179 B/op	     112 allocs/op
PASS
ok  	github.com/dapr/dapr/pkg/diagnostics	0.902s

$ go test -benchtime=500x -benchmem -run=^$ -bench ^BenchmarkHTTPMiddlewareWithPathNormalizationSyncPool$ github.com/dapr/dapr/pkg/diagnostics
goos: darwin
goarch: arm64
pkg: github.com/dapr/dapr/pkg/diagnostics
BenchmarkHTTPMiddlewareWithPathNormalizationSyncPool-14    	     500	   5721990 ns/op	    6049 B/op	     111 allocs/op
PASS
ok  	github.com/dapr/dapr/pkg/diagnostics	3.171s

$ go test -benchtime=1000x -benchmem -run=^$ -bench ^BenchmarkHTTPMiddlewareWithPathNormalizationSyncPool$ github.com/dapr/dapr/pkg/diagnostics
goos: darwin
goarch: arm64
pkg: github.com/dapr/dapr/pkg/diagnostics
BenchmarkHTTPMiddlewareWithPathNormalizationSyncPool-14    	    1000	   5691965 ns/op	    6024 B/op	     111 allocs/op
PASS
ok  	github.com/dapr/dapr/pkg/diagnostics	6.022s
```

ItalyPaleAle: The number of allocations went up by 1/3, which is quite a bit of added pressure to the GC.
Looks like the sync.Pool didn't help much as there was just 1 allocation saved.

ItalyPaleAle: I am un-convinced that using the full HTTP request stack is particularly efficient. I don't have an exact number to give you as to what's an acceptable number of added allocations, but it'd be great to limit it.

Maybe consider using a third-party router to integrate? And just the router? https://github.com/julienschmidt/httprouter is supposed to be very fast with very low allocations, if the caveats aren't a concern.

nelson-parente: @filintod noticed something in the test setup that is interesting which is the fact that since all tests were running in low cardinality the allocation increase is not related to the HTTP routing but to the fact that more tags are added in the requests with path normalization since these include path and method keys/values.

An issue was already opened about this: https://github.com/dapr/dapr/issues/6681

Given this, I ran the benchmarks again with high cardinality and the results with and without path normalization are very similar. This validates that the memory allocation increase observed in the first benchmark run was related to the fact that more tags are added and not to the path-matching implementation. 

_BenchmarkHTTPMiddlewareWithoutPathNormalization_

```
$ go test -benchtime=100x -benchmem -run=^$ -bench ^BenchmarkHTTPMiddlewareWithoutPathNormalization$ github.com/dapr/dapr/pkg/diagnostics

goos: darwin
goarch: arm64
pkg: github.com/dapr/dapr/pkg/diagnostics
BenchmarkHTTPMiddlewareWithoutPathNormalization-14    	     100	   5620625 ns/op	    7681 B/op	     126 allocs/op
PASS
ok  	github.com/dapr/dapr/pkg/diagnostics	0.938s

$ go test -benchtime=500x -benchmem -run=^$ -bench ^BenchmarkHTTPMiddlewareWithoutPathNormalization$ github.com/dapr/dapr/pkg/diagnostics
goos: darwin
goarch: arm64
pkg: github.com/dapr/dapr/pkg/diagnostics
BenchmarkHTTPMiddlewareWithoutPathNormalization-14    	     500	   5627427 ns/op	    7802 B/op	     125 allocs/op
PASS
ok  	github.com/dapr/dapr/pkg/diagnostics	3.136s

$ go test -benchtime=1000x -benchmem -run=^$ -bench ^BenchmarkHTTPMiddlewareWithoutPathNormalization$ github.com/dapr/dapr/pkg/diagnostics
goos: darwin
goarch: arm64
pkg: github.com/dapr/dapr/pkg/diagnostics
BenchmarkHTTPMiddlewareWithoutPathNormalization-14    	    1000	   5639435 ns/op	    7804 B/op	     126 allocs/op
PASS
ok  	github.com/dapr/dapr/pkg/diagnostics	5.949s
```

_BenchmarkHTTPMiddlewareWithPathNormalization_

```
$ go test -benchtime=100x -benchmem -run=^$ -bench ^BenchmarkHTTPMiddlewareWithPathNormalization$ github.com/dapr/dapr/pkg/diagnostics
goos: darwin
goarch: arm64
pkg: github.com/dapr/dapr/pkg/diagnostics
BenchmarkHTTPMiddlewareWithPathNormalization-14    	     100	   5641299 ns/op	    7121 B/op	     126 allocs/op
PASS
ok  	github.com/dapr/dapr/pkg/diagnostics	0.876s

$ go test -benchtime=500x -benchmem -run=^$ -bench ^BenchmarkHTTPMiddlewareWithPathNormalization$ github.com/dapr/dapr/pkg/diagnostics
goos: darwin
goarch: arm64
pkg: github.com/dapr/dapr/pkg/diagnostics
BenchmarkHTTPMiddlewareWithPathNormalization-14    	     500	   5667083 ns/op	    7018 B/op	     124 allocs/op
PASS
ok  	github.com/dapr/dapr/pkg/diagnostics	3.156s

$ go test -benchtime=1000x -benchmem -run=^$ -bench ^BenchmarkHTTPMiddlewareWithPathNormalization$ github.com/dapr/dapr/pkg/diagnostics
goos: darwin
goarch: arm64
pkg: github.com/dapr/dapr/pkg/diagnostics
BenchmarkHTTPMiddlewareWithPathNormalization-14    	    1000	   5670418 ns/op	    7007 B/op	     125 allocs/op
PASS
ok  	github.com/dapr/dapr/pkg/diagnostics	5.980s
```

_BenchmarkHTTPMiddlewareWithPathNormalizationSyncPool_

```
$ go test -benchtime=100x -benchmem -run=^$ -bench ^BenchmarkHTTPMiddlewareWithPathNormalizationSyncPool$ github.com/dapr/dapr/pkg/diagnostics

goos: darwin
goarch: arm64
pkg: github.com/dapr/dapr/pkg/diagnostics
BenchmarkHTTPMiddlewareWithPathNormalizationSyncPool-14    	     100	   5673832 ns/op	    6857 B/op	     124 allocs/op
PASS
ok  	github.com/dapr/dapr/pkg/diagnostics	0.878s

$ go test -benchtime=500x -benchmem -run=^$ -bench ^BenchmarkHTTPMiddlewareWithPathNormalizationSyncPool$ github.com/dapr/dapr/pkg/diagnostics

goos: darwin
goarch: arm64
pkg: github.com/dapr/dapr/pkg/diagnostics
BenchmarkHTTPMiddlewareWithPathNormalizationSyncPool-14    	     500           5687737 ns/op	    6714 B/op	     122 allocs/op
PASS
ok  	github.com/dapr/dapr/pkg/diagnostics	3.160s

$ go test -benchtime=1000x -benchmem -run=^$ -bench ^BenchmarkHTTPMiddlewareWithPathNormalizationSyncPool$ github.com/dapr/dapr/pkg/diagnostics

goos: darwin
goarch: arm64
pkg: github.com/dapr/dapr/pkg/diagnostics
BenchmarkHTTPMiddlewareWithPathNormalizationSyncPool-14    	    1000	   5663766 ns/op	    6712 B/op	     123 allocs/op
PASS
ok  	github.com/dapr/dapr/pkg/diagnostics	5.983s
```

pkg/apis/configuration/v1alpha1/types.go
```diff
@@ -222,6 +222,15 @@ type MetricHTTP struct {
 	// If false (the default), metrics for the HTTP server are collected with increased cardinality.
 	// +optional
 	IncreasedCardinality *bool `json:"increasedCardinality,omitempty"`
+	// +optional
+	PathMatching *PathMatching `json:"pathMatching,omitempty"`
+}
+
+type PathMatching struct {
```
JoshVanL, line 229: Add API comments.

pkg/config/configuration_test.go
```diff
@@ -627,6 +627,54 @@ func TestMetricsGetHTTPIncreasedCardinality(t *testing.T) {
...
 			},
 		}
-		assert.False(t, m.GetHTTPIncreasedCardinality(log))
+		assert.False(t, m.GetHTTPIncreasedCardinality())
+	})
+}
+
+func TestMetricsGetHTTPPathMatching(t *testing.T) {
+	log := logger.NewLogger("test")
+	log.SetOutput(io.Discard)
```
JoshVanL, line 636:
```suggestion
```

pkg/diagnostics/http_monitoring_benchmark_test.go
```diff
@@ -0,0 +1,115 @@
+package diagnostics
```
JoshVanL, line 1: Add copyright headers.

pkg/diagnostics/http_monitoring.go
```diff
@@ -36,6 +40,13 @@ var (
 	httpStatusCodeKey = tag.MustNewKey("status")
 	httpPathKey       = tag.MustNewKey("path")
 	httpMethodKey     = tag.MustNewKey("method")
+
+	log = logger.NewLogger("dapr.runtime.diagnostics")
+)
+
+const (
+	Egress = iota
+	Ingress
```
JoshVanL, line 49:
```suggestion
```

pkg/diagnostics/http_monitoring.go
```diff
@@ -56,14 +67,22 @@ type httpMetrics struct {
...
 	// Enable legacy metrics, which includes the full path
 	legacy bool
+
+	pathMatching pathMatchingConfig
+}
+
+type pathMatchingConfig struct {
+	enabled           bool
+	virtualIngressMux *http.ServeMux
+	virtualEgressMux  *http.ServeMux
```
JoshVanL, line 85: Create a `patchMaching` struct containing only a `mux *http.ServeMux`. `httpMetrics` can then have 2 `patchMatching`s, ingress and egress.

pkg/diagnostics/http_monitoring.go
```diff
@@ -162,7 +186,12 @@ func (h *httpMetrics) ClientRequestStarted(ctx context.Context, method, path str
 		return
 	}
 
-	if h.legacy {
+	matchedPath, ok := h.matchPath(path, Ingress)
+	if ok {
+		path = matchedPath
+	}
+
+	if h.legacy || h.pathMatching.enabled {
```
JoshVanL, line 194: Why does having path matching rules enabled put us in the legacy code path?

pkg/diagnostics/http_monitoring.go
```diff
@@ -307,3 +346,94 @@ func (h *httpMetrics) HTTPMiddleware(next http.Handler) http.Handler {
 		h.ServerRequestCompleted(r.Context(), method, path, status, reqContentSize, respSize, elapsed)
 	})
 }
+
+func (h *httpMetrics) initPathMatching(config *config.PathMatching) pathMatchingConfig {
```
JoshVanL, line 350:
```suggestion
func  newPathMatching(paths []string) *pathMatching {
```

pkg/diagnostics/http_monitoring.go
```diff
@@ -307,3 +346,94 @@ func (h *httpMetrics) HTTPMiddleware(next http.Handler) http.Handler {
...
+	for _, pattern := range config.IngressPaths {
+		pm.virtualIngressMux.Handle(pattern, http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			rw, ok := w.(*diagUtils.PathMatchingRW)
+			if !ok {
+				log.Errorf("Failed to cast to PathMatchingRW")
+				return
+			}
+			rw.MatchedPath = pattern
+		}))
+	}
```
JoshVanL, line 385: Rather than passing around strings and doing any/interface conversions- can we put the `stats.RecordWithTags` inside these request handlers?

JoshVanL (changes requested):

pkg/diagnostics/http_monitoring.go
nelson-parente: I changed that after this comment https://github.com/dapr/dapr/pull/7746#discussion_r1608424871

With path matching the keys used are the same as legacy code path the goal here was to simplify the code since these keys are not used in low cardinality.

nelson-parente: path matching is orthogonal to low/high cardinality so we need to have this in the legacy code path to allow path matching in legacy (high cardinality)

pkg/diagnostics/http_monitoring.go
nelson-parente: I don't think we can avoid the interface conversions here since we always need to cast the appropriate response writer 🙁 

Moving the metric producers inside the handlers would require us to have one rw per metric type. Personally, I'd prefer to keep things as they are now and not introduce that change in this PR if you are OK.

pkg/diagnostics/http_monitoring_path_matching.go
```diff
@@ -0,0 +1,113 @@
...
+	"strings"
+)
+
+type pathMatching struct {
+	mux *http.ServeMux
+}
+
+func newPathMatching(paths []string, legacy bool) *pathMatching {
+	if paths == nil {
+		return &pathMatching{}
```
JoshVanL, line 28: Should we return `nil` here?

pkg/diagnostics/http_monitoring_path_matching.go
```diff
@@ -0,0 +1,113 @@
...
+		},
+	}
+
+	crw := &pathMatchingRW{MatchedPath: path}
+	pm.mux.ServeHTTP(crw, req)
+
+	return crw.MatchedPath, true
+}
+
+const unmatchedPathPlaceholder = "_"
```
JoshVanL, line 101: Please put `const`s at the top of the file.

pkg/diagnostics/http_monitoring_path_matching.go
```diff
@@ -0,0 +1,113 @@
...
+			if !legacy {
+				rw, ok := w.(*pathMatchingRW)
+				if !ok {
+					log.Errorf("Failed to cast to PathMatchingRW")
+					return
+				}
+				rw.MatchedPath = unmatchedPathPlaceholder
+			}
+		}))
+	}
```
JoshVanL, line 64: Rather than have 2 `HandlerFunc` registrations, we can optionally append `"/"` if in non-legacy mode and remove dupes in the paths slice.


```go
if !legacy {
  paths = append(paths, "/")
  slices.Sort(paths)
  paths = slices.Compact(paths)
}
```


```suggestion
```

pkg/diagnostics/http_monitoring_path_matching.go
```diff
@@ -0,0 +1,113 @@
...
+func newPathMatching(paths []string, legacy bool) *pathMatching {
+	if paths == nil {
+		return &pathMatching{}
+	}
+
+	if len(paths) == 0 {
+		return &pathMatching{}
+	}
+
+	catchAllRegistered := false
```
JoshVanL, line 35:
```suggestion
```

pkg/diagnostics/http_monitoring_path_matching.go
```diff
@@ -0,0 +1,113 @@
...
+	if !pm.enabled() {
+		return "", false
+	}
+
+	if path == "" {
+		return "", false
+	}
+
+	if !strings.HasPrefix(path, "/") {
+		path = "/" + path
```
JoshVanL, line 85: Why is this required?

pkg/diagnostics/http_monitoring_path_matching.go
```diff
@@ -0,0 +1,113 @@
...
+	pm.mux.ServeHTTP(crw, req)
+
+	return crw.MatchedPath, true
+}
+
+const unmatchedPathPlaceholder = "_"
+
+type pathMatchingRW struct {
+	http.ResponseWriter
+	MatchedPath string
```
JoshVanL, line 105:
```suggestion
	matchedPath string
```

pkg/diagnostics/http_monitoring_path_matching.go
```diff
@@ -0,0 +1,113 @@
...
+}
+
+func (w *pathMatchingRW) WriteHeader(statusCode int) {
+}
+
+func (w *pathMatchingRW) Write(b []byte) (int, error) {
+	return len(b), nil
+}
```
JoshVanL, line 113: `http.ResponseWriter` is a member of the struct so these should not be necessary. 

```suggestion
```

pkg/diagnostics/http_monitoring.go
```diff
@@ -242,6 +273,10 @@ func (h *httpMetrics) Init(appID string, legacy bool) error {
 	} else {
 		serverTags = []tag.Key{appIDKey, httpMethodKey, httpStatusCodeKey}
 		clientTags = []tag.Key{appIDKey, httpStatusCodeKey}
+		if h.pathMatchingEnabled() {
+			serverTags = append(serverTags, httpPathKey)
+			clientTags = append(clientTags, httpPathKey, httpMethodKey)
```
JoshVanL, line 278: Should a non-empty slice in one direction also enable the other? Is that what is happening?

JoshVanL (changes requested):
Thanks @nelson-parente, couple more from me :slightly_smiling_face:

pkg/diagnostics/http_monitoring_path_matching.go
nelson-parente: The goal for these is that they serve as no-ops and don't do any actual work since we just want to have use routing and not the default rw mechanism. wdyt?

pkg/diagnostics/http_monitoring_path_matching.go
nelson-parente: Paths not starting `/` are invalid and the request will return an error on the handle function. 

I noticed this when testing and we had errors with `dapr/config`path.

pkg/diagnostics/http_monitoring.go
nelson-parente: yeah, we can split each case 👍

pkg/diagnostics/http_monitoring_path_matching.go
nelson-parente: I think we need the 2 handlers registration because we want to distinguish when the user explicitly sets the "/" path or not.

When the user sets the "/" path we match it and use it otherwise if the user does not set a "/" we fallback to the unmatched placeholder "_" so we need to have different handler funcs for those cases.

pkg/diagnostics/http_monitoring_path_matching.go
nelson-parente: Update: all invoke method paths are coming without the initial slash. I fixed the dapr/config but all these others are using the `*invokev1.InvokeMethodRequest` and I don't really know the why these don't have the initial slash.

pkg/diagnostics/http_monitoring_path_matching.go
nelson-parente: done! we always need to register the catch all otherwise serveHTTP will throw an error when it doesn't match the path.

nelson-parente: done!

JoshVanL (approved):

artursouza (approved):

marcduiker:
@holopin-bot @nelson-parente Enjoy your Dapr badge!

PR #7747, opened 2024-05-20
Subscriptions: bi-directional subscription streaming.

JoshVanL:
Adds SubscribeTopicEvents proto API which dynamically subscribes to pubsub topics based on https://github.com/dapr/proposals/pull/52.

This is a basic gRPC implementation of the API whereby, like Subscription hot-reloading today, subscribing to a topic will reload _every_ active subscription for the current daprd. In a future PR, reloading of Subscriptions will be granular to the specific pubsub topic.

Stream subscriptions are also only active once daprd declares the application as both present and ready. Dynamic stream subscriptions should be active both whether a app is running or not, as well as whether it is ready or not. This will be addressed in a future PR.


dapr/proto/runtime/v1/dapr.proto
```diff
@@ -58,6 +59,10 @@ service Dapr {
   // Bulk Publishes multiple events to the specified topic.
   rpc BulkPublishEventAlpha1(BulkPublishRequest) returns (BulkPublishResponse) {}
 
+  // SubscribeTopicEvents subscribes to a PubSub topic and receives topic events
+  // from it.
+  rpc SubscribeTopicEvents(stream SubscribeTopicEventsRequest) returns (stream TopicEventRequest) {}
```
yaron2, line 64: This should have Alpha1

pkg/api/grpc/endpoints.go
```diff
@@ -47,6 +47,9 @@ var endpoints = map[string][]string{
 	"publish.v1alpha1": {
 		daprRuntimePrefix + "v1.Dapr/BulkPublishEventAlpha1",
 	},
+	"subscribe.v1": {
+		daprRuntimePrefix + "v1.Dapr/SubscribeTopicEvents",
```
yaron2, line 51: Add Alpha1

pkg/proto/runtime/v1/dapr_tracing.go
```diff
@@ -148,6 +148,15 @@ func (*StartWorkflowRequest) AppendSpanAttributes(rpcMethod string, m map[string
 func (*SubscribeConfigurationRequest) AppendSpanAttributes(rpcMethod string, m map[string]string) {
 	// TODO
 }
+func (*SubscribeTopicEventsInitialRequest) AppendSpanAttributes(rpcMethod string, m map[string]string) {
+	// TODO
```
yaron2, line 152: Is this to be added in a future PR?

yaron2 (changes requested):

pkg/proto/runtime/v1/dapr_tracing.go
JoshVanL: I  copied this from the other funcs in the [file](https://github.com/dapr/dapr/blob/a6bafa6babec9856259c6a1bb2a73beab0d245e8/pkg/proto/runtime/v1/dapr_tracing.go). Not entirely sure why they are `TODO`, but are needed to satisfy an interface.

pkg/runtime/processor/pubsub/streamerconn.go
```diff
@@ -0,0 +1,56 @@
...
+	stream           rtv1pb.Dapr_SubscribeTopicEventsAlpha1Server
+}
+
+func (s *streamconn) notifyPublishResponse(ctx context.Context, resp *rtv1pb.SubscribeTopicEventsResponseAlpha1) {
+	s.lock.RLock()
+	ch, ok := s.publishResponses[resp.GetId()]
+	s.lock.RUnlock()
+
+	if !ok {
+		log.Errorf("no client stream expecting publish response for id %q", resp.GetId())
```
yaron2, line 36: Is there a scenario where this is a valid path and thus warrants a Warn instead of Error?

JoshVanL: No, this likely means that the client is reporting the processing result of a message with an ID which has already been reported or was never published to the app in the first place.

yaron2 (approved):

daixiang0 (approved):
LGTM

Issue #7736, opened 2024-05-16
hight cardinality metrics are shown for actors but increasedCardinality is set to false


fabistb:
## In what area(s)?
> /area runtime

## What version of Dapr?

> 1.13.2

## Expected Behavior
When setting in the appconfig _increasedCardinality_ to false I don't expect any high cardinality metrics in the logging server.

**App Config:**

```yaml
kind: Configuration
spec:
  metric:
    http:
      increasedCardinality: false
```

## Actual Behavior
There are now new high cardinality `dapr_http_server_latency_bucket` metcis related to the _InvokeActor_ method.

```
dapr_http_server_latency_bucket{app_id="supplement",method="InvokeActor/SupplementActor.7fba4d69-bd63-4782-83aa-f523223c26df",status="200",le="50000"} 3
dapr_http_server_latency_bucket{app_id="supplement",method="InvokeActor/SupplementActor.7fba4d69-bd63-4782-83aa-f523223c26df",status="200",le="100000"} 3
dapr_http_server_latency_bucket{app_id="supplement",method="InvokeActor/SupplementActor.7fba4d69-bd63-4782-83aa-f523223c26df",status="200",le="+Inf"} 3
dapr_http_server_latency_sum{app_id="supplement",method="InvokeActor/SupplementActor.7fba4d69-bd63-4782-83aa-f523223c26df",status="200"} 357
dapr_http_server_latency_count{app_id="supplement",method="InvokeActor/SupplementActor.7fba4d69-bd63-4782-83aa-f523223c26df",status="200"} 3
dapr_http_server_latency_bucket{app_id="supplement",method="InvokeActor/SupplementActor.8032a1a1-ec2e-4518-965c-a9ad9adbff31",status="200",le="1"} 0
dapr_http_server_latency_bucket{app_id="supplement",method="InvokeActor/SupplementActor.8032a1a1-ec2e-4518-965c-a9ad9adbff31",status="200",le="2"} 0
dapr_http_server_latency_bucket{app_id="supplement",method="InvokeActor/SupplementActor.8032a1a1-ec2e-4518-965c-a9ad9adbff31",status="200",le="3"} 0
dapr_http_server_latency_bucket{app_id="supplement",method="InvokeActor/SupplementActor.8032a1a1-ec2e-4518-965c-a9ad9adbff31",status="200",le="4"} 0
[...]

PR #7756, opened 2024-05-23
Update http monitoring method parsing in low cardinality cases

jmprusi:
# Description

Modified the http monitoring logic to reduce Method cardinality in low cardinality scenarios where the actorID is appended to the endpoint.

## Issue reference

closes: [#7736](https://github.com/dapr/dapr/issues/7736)

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Unit tests passing

### DIFF
--- a/pkg/api/http/actors.go
+++ b/pkg/api/http/actors.go
@@ -76,7 +76,7 @@ func actorInvocationMethodNameFn(r *http.Request) string {
 
 func (a *api) constructActorEndpoints() []endpoints.Endpoint {
 	methodNameFn := actorInvocationMethodNameWithIDFn
-	if !a.metricSpec.GetHTTPIncreasedCardinality() {
+	if a.metricSpec != nil && !a.metricSpec.GetHTTPIncreasedCardinality() {
 		methodNameFn = actorInvocationMethodNameFn
 	}
 	return []endpoints.Endpoint{
--- a/pkg/api/http/actors.go
+++ b/pkg/api/http/actors.go
@@ -66,11 +66,19 @@ func appendActorInvocationSpanAttributesFn(r *http.Request, m map[string]string)
 	m[diagConsts.DaprAPISpanNameInternal] = "CallActor/" + actorType + "/" + chi.URLParam(r, "method")
 }
 
-func actorInvocationMethodNameFn(r *http.Request) string {
+func actorInvocationMethodNameWithIDFn(r *http.Request) string {
 	return "InvokeActor/" + chi.URLParam(r, actorTypeParam) + "." + chi.URLParam(r, actorIDParam)
 }
 
+func actorInvocationMethodNameFn(r *http.Request) string {
+	return "InvokeActor/" + chi.URLParam(r, actorTypeParam)
+}
+
 func (a *api) constructActorEndpoints() []endpoints.Endpoint {
+	methodNameFn := actorInvocationMethodNameWithIDFn
+	if !a.metricSpec.GetHTTPIncreasedCardinality() {
+		methodNameFn = actorInvocationMethodNameFn
+	}
 	return []endpoints.Endpoint{
 		{
 			Methods: []string{http.MethodPost, http.MethodPut},
@@ -90,7 +98,7 @@ func (a *api) constructActorEndpoints() []endpoints.Endpoint {
 				Name:                 endpoints.EndpointGroupActors,
 				Version:              endpoints.EndpointGroupVersion1,
 				AppendSpanAttributes: appendActorInvocationSpanAttributesFn,
-				MethodName:           actorInvocationMethodNameFn,
+				MethodName:           methodNameFn,
 			},
 			Handler: a.onDirectActorMessage,
 			Settings: endpoints.EndpointSettings{
--- a/pkg/api/http/http.go
+++ b/pkg/api/http/http.go
@@ -73,6 +73,7 @@ type api struct {
 	sendToOutputBindingFn func(ctx context.Context, name string, req *bindings.InvokeRequest) (*bindings.InvokeResponse, error)
 	readyStatus           bool
 	outboundReadyStatus   bool
+	metricSpec            config.MetricSpec
 	tracingSpec           config.TracingSpec
 	maxRequestBodySize    int64 // In bytes
 }
@@ -114,6 +115,7 @@ type APIOpts struct {
 	PubsubAdapter         runtimePubsub.Adapter
 	SendToOutputBindingFn func(ctx context.Context, name string, req *bindings.InvokeRequest) (*bindings.InvokeResponse, error)
 	TracingSpec           config.TracingSpec
+	MetricSpec            config.MetricSpec
 	MaxRequestBodySize    int64 // In bytes
 }
 
@@ -126,6 +128,7 @@ func NewAPI(opts APIOpts) API {
 		pubsubAdapter:         opts.PubsubAdapter,
 		sendToOutputBindingFn: opts.SendToOutputBindingFn,
 		tracingSpec:           opts.TracingSpec,
+		metricSpec:            opts.MetricSpec,
 		maxRequestBodySize:    opts.MaxRequestBodySize,
 	}
 
--- a/pkg/api/http/http.go
+++ b/pkg/api/http/http.go
@@ -73,7 +73,7 @@ type api struct {
 	sendToOutputBindingFn func(ctx context.Context, name string, req *bindings.InvokeRequest) (*bindings.InvokeResponse, error)
 	readyStatus           bool
 	outboundReadyStatus   bool
-	metricSpec            config.MetricSpec
+	metricSpec            *config.MetricSpec
 	tracingSpec           config.TracingSpec
 	maxRequestBodySize    int64 // In bytes
 }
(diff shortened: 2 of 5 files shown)


JoshVanL (changes requested):
Thanks @jmprusi, please also extend the [HTTP metrics cardinality integration tests](https://github.com/dapr/dapr/tree/master/tests/integration/suite/daprd/metrics/http/cardinality) for coverage :slightly_smiling_face:

pkg/diagnostics/http_monitoring.go
```diff
@@ -299,7 +300,16 @@ func (h *httpMetrics) HTTPMiddleware(next http.Handler) http.Handler {
 			endpointData, _ := r.Context().Value(endpoints.EndpointCtxKey{}).(*endpoints.EndpointCtxData)
 			method = endpointData.GetEndpointName()
 			if endpointData != nil && endpointData.Group != nil && endpointData.Group.MethodName != nil {
-				method = endpointData.Group.MethodName(r)
+				MethodName := endpointData.Group.MethodName(r)
+				parsedMethodName := strings.SplitN(MethodName, "/", 2)
+				switch parsedMethodName[0] {
```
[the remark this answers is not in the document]
jmprusi: Updated the PR, PTAL

pkg/diagnostics/http_monitoring.go
```diff
@@ -299,7 +300,15 @@ func (h *httpMetrics) HTTPMiddleware(next http.Handler) http.Handler {
 			endpointData, _ := r.Context().Value(endpoints.EndpointCtxKey{}).(*endpoints.EndpointCtxData)
 			method = endpointData.GetEndpointName()
 			if endpointData != nil && endpointData.Group != nil && endpointData.Group.MethodName != nil {
-				method = endpointData.Group.MethodName(r)
+				methodName := endpointData.Group.MethodName(r)
```
ItalyPaleAle, line 303: This kind of "ad-hoc" things have bit us before in the past, where before #6919 the business logic for monitoring (logs, metrics, tracing) was a bit scattered all over the place, and it caused bugs and inconsistencies.

Instead of hardcoding method names in the `diagnostics` package, it'd be preferrable to add a new rule to EndpointGroup (https://github.com/dapr/dapr/blob/b9843e84a86b924255f9ab115c4d163cca9593b7/pkg/api/http/endpoints/endpointgroup.go#L21) which is a method that can be invoked (alternatively, to "Endpoint" if it's specific for an endpoint and not a group)

Then when the EndpointGroup/Endpoint [is defined](https://github.com/dapr/dapr/blob/b9843e84a86b924255f9ab115c4d163cca9593b7/pkg/api/http/actors.go#L73), the rule can be added there.

We do something similar for logs and other metrics today, for example for PubSub to append span attributes: https://github.com/dapr/dapr/blob/b9843e84a86b924255f9ab115c4d163cca9593b7/pkg/api/http/http.go#L264

ItalyPaleAle (changes requested):

pkg/diagnostics/http_monitoring.go
jmprusi: @ItalyPaleAle, I've updated the PR. Let me know if it makes more sense now. Thank you.

artursouza (dismissed):

tests/integration/suite/daprd/metrics/http/cardinality/high.go
```diff
@@ -22,6 +22,8 @@ import (
 
 	"github.com/stretchr/testify/assert"
 
+	"github.com/dapr/dapr/tests/integration/framework/process/placement"
+
```
JoshVanL, line 26:
```suggestion
```

pkg/api/http/actors.go
```diff
@@ -70,7 +70,15 @@ func actorInvocationMethodNameFn(r *http.Request) string {
 	return "InvokeActor/" + chi.URLParam(r, actorTypeParam) + "." + chi.URLParam(r, actorIDParam)
 }
 
+func actorInvocationLowCardinalityMethodNameFn(r *http.Request) string {
```
JoshVanL, line 73: I would like us to get away from using "low cardinality" as a concept in Dapr. Let's rename this func to `actorInvocationMethodNameFn` and the above func to `actorInvocationMethodNameWithIDFn`.

pkg/api/http/actors.go
```diff
@@ -90,7 +98,7 @@ func (a *api) constructActorEndpoints() []endpoints.Endpoint {
 				Name:                 endpoints.EndpointGroupActors,
 				Version:              endpoints.EndpointGroupVersion1,
 				AppendSpanAttributes: appendActorInvocationSpanAttributesFn,
-				MethodName:           actorInvocationMethodNameFn,
+				MethodName:           methodNameFn,
```
JoshVanL, line 101: Should we be using a new field on `EndpointGroup` for this?

pkg/api/http/http.go
```diff
@@ -73,6 +73,7 @@ type api struct {
 	sendToOutputBindingFn func(ctx context.Context, name string, req *bindings.InvokeRequest) (*bindings.InvokeResponse, error)
 	readyStatus           bool
 	outboundReadyStatus   bool
+	metricSpec            config.MetricSpec
```
JoshVanL, line 76: Where is this set?

tests/integration/suite/daprd/metrics/http/cardinality/low.go
```diff
@@ -20,6 +20,8 @@ import (
 	"strings"
 	"testing"
 
+	"github.com/dapr/dapr/tests/integration/framework/process/placement"
+
```
JoshVanL, line 24:
```suggestion
```

tests/integration/suite/daprd/metrics/http/cardinality/high.go
```diff
@@ -35,39 +37,48 @@ func init() {
...
+		daprd.WithPlacementAddresses(h.place.Address()),
+		daprd.WithInMemoryActorStateStore("mystore"),
 		daprd.WithConfigManifests(t, `
 apiVersion: dapr.io/v1alpha1
 kind: Configuration
 metadata:
   name: lowcardinality
 spec:
-  metrics:
+ metrics:
```
JoshVanL, line 69:
```suggestion
  metrics:
```

JoshVanL (changes requested):

pkg/api/http/actors.go
jmprusi: fixed

pkg/api/http/http.go
jmprusi: fixed

pkg/api/http/actors.go
jmprusi: I don't have a strong opinion here. Having a single attribute that handles the logic of low/high cardinality seems to be ok to me.

pkg/api/http/http.go
JoshVanL: Concerned that we didn't have failing tests, even though we weren't setting this field- do we have enough test coverage?

JoshVanL (changes requested):

pkg/api/http/http.go
jmprusi: I agree! Double-checked tests; this should be better and properly fails when new code is. PTAL

JoshVanL: Maybe I am missing something but with the current HEAD, the tests still pass for me if I remove setting this field.

jmprusi: As we are only using that field this method:

```
// GetHTTPIncreasedCardinality returns true if increased cardinality is enabled for HTTP metrics
func (m MetricSpec) GetHTTPIncreasedCardinality(log logger.Logger) bool {
	if m.HTTP == nil || m.HTTP.IncreasedCardinality == nil {
		// The default is false
		return false
	}
}
```

The method returns false ("low cardinality") as m.HTTP is nil. That stops and returns early. If instead of an "or"(||), we had an "And" (&&), we would have a nice panic there. That's why I've missed it :)  (I played with a similar example here https://goplay.tools/snippet/EMgw3OgHLPH) 

If you remove the negation on the comparison here: https://github.com/dapr/dapr/pull/7756/files#diff-a6c70a4296446b8b8cbe6bebb3d298107381d89c750b3bb47177629cd57e6a0bR79 You should get the tests failing.

pkg/diagnostics/http_monitoring.go
artursouza: @ItalyPaleAle ping?

pkg/api/http/http.go
artursouza: @JoshVanL ping ?

artursouza (dismissed):

pkg/api/http/http.go
JoshVanL: Please add a test case which hits this code path, i.e. would panic if this field was omitted. Tests continue to pass for me when this field has been removed.

FYI there looks to be a lint failure https://github.com/dapr/dapr/actions/runs/9508989782/job/26211228529?pr=7756

jmprusi: Fixed the linting and moved the field to a pointer --> this should allow us to diff between unset/false.

JoshVanL (approved):

marcduiker:
@holopin-bot @jmprusi Thank you Joaquim!

PR #7757, opened 2024-05-23
Subscriptions: bi-directional subscription streaming- subscribe on no healthz.

JoshVanL:
Refactors pubsub machinery to allow for gRPC bi-directional subcription streaming when there is no application, or the application in unhealhty.

dapr/proposals#52


pkg/runtime/compstore/subscriptions.go
```diff
@@ -14,95 +14,195 @@ limitations under the License.
...
+	defer c.lock.RUnlock()
 
-	c.declarativeSubscriptions = append(c.declarativeSubscriptions, subscriptions...)
+	var subs []rtpubsub.Subscription
+	taken := make(map[string]int)
 
-	return nil
+	for _, name := range c.subscriptions.declarativesList {
+		sub := c.subscriptions.declaratives[name].Subscription
+		key := sub.PubsubName + "||" + sub.Topic
```
yaron2, line 105: We have the concept of namespaced pub/sub - does this require a change here? https://docs.dapr.io/developing-applications/building-blocks/pubsub/howto-namespace/

pkg/runtime/processor/subscriber/subscriber.go
```diff
@@ -0,0 +1,372 @@
...
+		return nil
+	}
+
+	appChannel := s.channels.AppChannel()
+	if appChannel == nil {
+		log.Warn("app channel not initialized, make sure -app-port is specified if pubsub subscription is required")
+		return nil
+	}
+
+	s.hasInitProg = true
```
yaron2, line 338: This means the method `initProgramaticSubscriptions` cannot be retried - this might be ok, just wanted to call out that's the case.

pkg/api/grpc/subscribe.go
```diff
@@ -24,19 +29,71 @@ func (a *api) SubscribeTopicEventsAlpha1(stream runtimev1pb.Dapr_SubscribeTopicE
...
+
+	if len(req.GetPubsubName()) == 0 {
+		return errors.New("pubsubName is required")
+	}
+
+	if len(req.GetTopic()) == 0 {
+		return errors.New("topic is required")
+	}
+
+	// TODO: @joshvanl: handle duplicate key names.
```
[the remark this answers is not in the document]
JoshVanL: In the next PR we error out if the client tries to subscribe to an existing topic/pubsub pair https://github.com/dapr/dapr/pull/7770/files#diff-97540f3f58ecaa882c9bdc06116ce1f371a0b73407ca2123859c46650bc2d8f8R74

pkg/runtime/compstore/subscriptions.go
JoshVanL: To me- it looks like this namespacing is done on the PubSub component spec level. The _name_ of the pubsub still needs to be unique for that namespace, so there would be no possible for conflict here.

pkg/runtime/processor/subscriber/subscriber.go
JoshVanL: As today, this function will continue to only be called once.

yaron2 (approved):

PR #7767, opened 2024-05-28
[WIP] feat: add method and empty path for low cardinality

nelson-parente:
# Description

## Issue reference

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing

### DIFF
--- a/pkg/diagnostics/http_monitoring.go
+++ b/pkg/diagnostics/http_monitoring.go
@@ -23,7 +23,6 @@ import (
 	"go.opencensus.io/stats/view"
 	"go.opencensus.io/tag"
 
-	"github.com/dapr/dapr/pkg/api/http/endpoints"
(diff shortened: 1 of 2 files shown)


tests/e2e/metrics/metrics_test.go
```diff
@@ -231,10 +231,10 @@ func findHTTPMetricFromPrometheus(t *testing.T, app string, res *http.Response)
 					case "method":
 						if count.GetValue() > 0 {
 							switch val {
-							case "Healthz":
```
artursouza, line 234: Is there a way to have the information on the healthz calls too? Maybe add it to the path since it can be interpreted as an allowed path.

nelson-parente: I was holding that for another PR, adding well-known paths like `/dapr/config`, `/healthz`, etc

When we make that change it makes sense to add a new switch case testing the healthz path

Issue #7605, opened 2024-03-07
v1.14 Release Planning


JoshVanL:
[Release Tracking Board](https://github.com/orgs/dapr/projects/80)

Release lead: @mikeee 
Release lead shadow(s): -
Perf test lead: @cicoyle 
Longhaul test: @elena-kolevska 
Build manager: @JoshVanL 

Read about the [Dapr release lifecycle](https://github.com/dapr/community/blob/master/release-process.md)

## Milestone dates

- [x] Feature definition starts: 12th March 2024 (12/03/2024)
- [x] Feature definition ends: 26th March 2024 (26/03/2024)
- [x] Feature freeze: ~~14th May 2024~~ -> 28th May 2024
- [x] Code freeze (Endgame starts ~2 weeks stabilization):  ~~28th May 2024~~ -> ~~11th June 2024~~ -> 25th June 2024
- [x] RC Release day: ~~28th May 2024~~ -> ~~11th June 2024~~ -> ~~25th June 2024~~ -> 03rd July 2024
- [x] Release day: ~~11th June 2024~~ -> ~~9th July 2024~~ -> 13th August 2024
- [ ] Announcement day: ~~10th July 2024~~ -> 14th August 2024 < - Docs, Samples and Runtime, CLI, Release notes

## Priorities for this release

- [x] [P0] Longhauls/tests: azurecr non-existent - https://github.com/dapr/test-infra/issues/225
- [x] [P0] Distributed Scheduler Implementation #7246 @cicoyle 
  - [x] cli changes https://github.com/dapr/cli/pull/1405
  - [x] go-sdk implementation https://github.com/dapr/go-sdk/issues/561 @mikeee 
- [x] [P0] Pubsub bidirectional streams [Proposal #52](https://github.com/dapr/proposals/pull/52) @JoshVanL 
  - [x] API -> health/z checks -> split subscriptions into new structures 
  - [x] Go SDK implementation https://github.com/dapr/go-sdk/pull/578
  - [x] #814 @JoshVanL
- [x] [P0] Opencensus / Open telemetry [dapr#7703](https://github.com/dapr/dapr/issues/7703)
- [x] [P1] Finish standardization of env vars java-sdk
  - [x] java-sdk [java-sdk#1028](https://github.com/dapr/java-sdk/issues/1028) @artursouza 
- [x] [P0] Namespaced actors #3786 @elena-kolevska  - https://github.com/dapr/dapr/pull/7745
[...]

Issue #7246, opened 2023-11-27
POC: Distributed Scheduler


cicoyle:
The intent of this issue is to break down the work into bite size pieces. This may not be a fully inclusive list, ie things may be added/tweaked.

The work will be done in [my branch](https://github.com/cicoyle/dapr/tree/feat-dist-scheduler) in my dapr fork. PRs should be made into my branch.

- [ ] Building Block
    - [x] protos @cicoyle
    - [x] service layer funcs @cicoyle 
    - [ ] [P2] tests @cicoyle
    - [ ] [P2] update error standardization format @cicoyle 
- [ ] Service
    - [x] binary builds & scheduler service runs @DeepanshuA @cicoyle @JoshVanL 
    - [x] write charts @cicoyle
    - [ ] [P0] scale (perf) tests @artursouza @JoshVanL 
      - [ ] (optimize) scale loading jobs from etcd into cron scheduler @JoshVanL (wip)
      - [ ] [P2] (optimize) producer/consumer for triggering jobs: scheduling too many jobs=too many go routines
    - [x] protos @cicoyle 
      - [ ] [P1] add internal job to scheduler.proto @JoshVanL 
    - [x] service layer funcs @cicoyle
    - [ ] [P0] authorization of service layer funcs @JoshVanL 
    - [x] [P0] namespace, appID, scope first class fields (outside of metadata) where scope=ENUM(app-jobs, namespace-actors), actor fields still in metadata @cicoyle @JoshVanL  -> Should just need scope added and rm appID/ns from metadata once lib accts for sending it back (add the scheduleJobMetadata for the typing of job vs actor reminder for entry in etcd prefix)
    - [x] etcd cron library @artursouza @JoshVanL 
      - [x] ensure persistence
      - [x] job TTL
      - [x] job repeat count
      - [x] virtual partitioning of jobs, such that a scheduler service owns a subset of all jobs that exist
      - [x] trigger back the proper amt of times with repeats set @artursouza @JoshVanL 
      - [x] [P0] optimize cron lib @JoshVanL wip
    - [ ] embedded etcd
        - [ ] [P2] mtls etcd @JoshVanL 
        - [x] [P0] persistent volume @JoshVanL 
        - [ ] [P1] ensure if pods are added or removed & rejoined everything works still @elena-kolevska
          - [ ] [P1] should fix this err: `error running scheduler: member <id> has already been bootstrapped`
[...]

PR #7768, opened 2024-05-28
Jobs Building Block & Scheduler Control Plane Service

cicoyle:
This PR adds the relevant code for:
1. a new Jobs Building Block 
2. a new Scheduler Control Plane Service

This PR follows [this official proposal,](https://github.com/dapr/proposals/pull/44). 

This is an end to end solution where an app is able to schedule a job to the daprd sidecar and perform basic crud operations with the job via the `Jobs api`. The daprd sidecar sends the job to the `Scheduler` control plane service, which then stores the job in its embedded etcd. At trigger time, the scheduler will send the job back to the daprd sidecar where the sidecar will send the triggered job along to the app. 

This includes work to replace the actor reminder subsystem, by scheduling the actor reminders in the Scheduler control plane service embedded etcd via the `SchedulerReminders` feature flag.

Perf numbers can be found in the [proposal comments here](https://github.com/dapr/proposals/pull/44).

There were several individuals involved in making this work happen - thank you to all 🚀 

## Issue reference

[1.14 release P0](https://github.com/dapr/dapr/issues/7605)
[Here is the issue tracking the work](https://github.com/dapr/dapr/issues/7246). There are a few remaining items that will be completed before the release date.

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [ ] Unit tests passing


dapr/proto/runtime/v1/dapr.proto
```diff
@@ -193,6 +193,15 @@ service Dapr {
...
   rpc Shutdown (ShutdownRequest) returns (google.protobuf.Empty) {}
+
+  // Create and schedule a job
+  rpc ScheduleJob(ScheduleJobRequest) returns (google.protobuf.Empty) {}
+
+  // Gets a scheduled job
+  rpc GetJob(GetJobRequest) returns (GetJobResponse) {}
+
+  // Delete a job
+  rpc DeleteJob(DeleteJobRequest) returns (google.protobuf.Empty) {}
```
mikeee, line 204: Do these definitions need to be appended with `Alpha1`  as a suffix?

tests/perf/actor_reminder/actor_reminder_test.go
```diff
@@ -152,3 +191,248 @@ func TestActorReminderRegistrationPerformance(t *testing.T) {
...
+		sidecarUsage,
+		appUsage)
+
+	report.SetTotalRestartCount(restarts)
+	err = utils.UploadAzureBlob(report)
+
+	if err != nil {
+		t.Error(err)
+	}
+
```
mikeee, line 270:
```suggestion
```

This blob upload step should be removed

charts/dapr/charts/dapr_scheduler/templates/dapr_scheduler_poddisruptionbudget.yaml
```diff
@@ -0,0 +1,32 @@
...
+  maxUnavailable: {{ .Values.global.ha.disruption.maximumUnavailable }}
+{{- end }}
+  selector:
+    matchLabels:
+      app: dapr-scheduler-server
+      {{- range $key, $value := .Values.global.k8sLabels }}
+      {{ $key }}: {{ tpl $value $ }}
+      {{- end }}
+      {{- with .Values.global.labels }}
+        {{- toYaml . | nindent 6 }}
```
ItalyPaleAle, line 30: Shouldn't this be 8?

cmd/scheduler/options/options.go
```diff
@@ -0,0 +1,132 @@
...
+	EtcdCompactionRetention string
+
+	Logger  logger.Options
+	Metrics *metrics.Options
+
+	taFile string
+}
+
+func New(origArgs []string) (*Options, error) {
+	// We are using pflag to parse the CLI flags
```
ItalyPaleAle, line 61: This is probably not needed for a new service, since it was only for backwards-compat for daprd.

cmd/daprd/options/options.go
```diff
@@ -248,6 +251,13 @@ func New(origArgs []string) (*Options, error) {
 		opts.DaprBlockShutdownDuration = nil
 	}
 
+	if !fs.Changed("scheduler-host-address") {
```
ItalyPaleAle, line 254: Since this uses StringSliceVar, the parsing of CSV values is automatic. StringSliceVar supports both CSV and repeating the flag

cmd/scheduler/options/options.go
```diff
@@ -0,0 +1,132 @@
...
+
+	// Create a flag set
+	fs := pflag.NewFlagSet("scheduler", pflag.ExitOnError)
+	fs.SortFlags = true
+
+	fs.IntVar(&opts.Port, "port", 50006, "The port for the scheduler server to listen on")
+	fs.IntVar(&opts.HealthzPort, "healthz-port", 8080, "The port for the healthz server to listen on")
+	fs.StringVar(&opts.HealthzListenAddress, "healthz-listen-address", "", "The listening address for the healthz server")
+
+	fs.StringVar(&opts.ListenAddress, "listen-address", "127.0.0.1", "The address for the Scheduler to listen on")
```
ItalyPaleAle, line 86: `0.0.0.0`? 127.0.0.1 is not good in containers (aside from sidecars)

dapr/proto/scheduler/v1/scheduler.proto
```diff
@@ -0,0 +1,158 @@
...
+  optional uint32 repeats = 2;
+
+  // Optional: sets time at which or time interval before the callback is invoked for the first time.
+  optional string due_time = 3;
+
+  // Optional: Time To Live to allow for auto deletes (accounting for Actor Reminders).
+  optional string ttl = 4;
+
+  // Job data.
+  google.protobuf.Any data = 5;
```
ItalyPaleAle, line 35: Why this instead of `bytes`?

dapr/proto/scheduler/v1/scheduler.proto
```diff
@@ -0,0 +1,158 @@
...
+
+message Job {
+  // The schedule for the job.
+  optional string schedule = 1;
+
+  // Optional: jobs with fixed repeat counts (accounting for Actor Reminders).
+  optional uint32 repeats = 2;
+
+  // Optional: sets time at which or time interval before the callback is invoked for the first time.
+  optional string due_time = 3;
```
ItalyPaleAle, line 29: Should probably be a google.protobuf.Timestamp?

dapr/proto/scheduler/v1/scheduler.proto
```diff
@@ -0,0 +1,158 @@
...
+
+  // actorTypes is the optional list of actor types to watch for.
+  repeated string actorTypes = 3;
+}
+
+// WatchJobsRequestResult is the result of a job execution to allow the job to
+// be marked as processed.
+message WatchJobsRequestResult {
+  // uuid is the uuid of the job that has finished processing.
+  uint64 uuid = 1;
```
ItalyPaleAle, line 99: uuid is 128 bits so can't fit in uint64?

dapr/proto/scheduler/v1/scheduler.proto
```diff
@@ -0,0 +1,158 @@
...
+// WatchJobsRequestInitial is the initial request to start watching for jobs.
+message WatchJobsRequestInitial {
+  // app_id is the App ID of the requester.
+  string app_id = 1;
+
+  // namespace is the namespace of the requester.
+  string namespace = 2;
+
+  // actorTypes is the optional list of actor types to watch for.
+  repeated string actorTypes = 3;
```
ItalyPaleAle, line 92: `actor_types` because in protobuf the convention is to use snake_case

pkg/actors/internal/reminders.go
```diff
@@ -33,6 +33,7 @@ type LookupActorFn func(ctx context.Context, actorType string, actorID string) (
 type StateStoreProviderFn func() (string, TransactionalStateStore, error)
 
 // RemindersProvider is the interface for the object that provides reminders services.
+// Deprecated: this will be removed after we have migrated to Scheduler Service.
```
ItalyPaleAle, line 36: There are dependencies on this, should not be deprecated

pkg/actors/actors.go
```diff
@@ -1143,11 +1182,84 @@ func (a *actorsRuntime) doExecuteReminderOrTimer(ctx context.Context, reminder *
...
+
+	return ptr.Of("@every " + duration.String()), repeats, nil
+}
+
 func (a *actorsRuntime) CreateReminder(ctx context.Context, req *CreateReminderRequest) error {
 	if !a.actorsConfig.Config.HostedActorTypes.IsActorTypeHosted(req.ActorType) {
 		return ErrReminderOpActorNotHosted
 	}
 
+	if a.schedulerClients != nil && a.schedulerReminders {
```
ItalyPaleAle, line 1212: Please move to a separate function for readability

pkg/actors/actors.go
```diff
@@ -230,6 +235,25 @@ func newActorsWithClock(opts ActorsOpts, clock clock.WithTicker) (ActorRuntime,
 
 	a.timers.SetExecuteTimerFn(a.executeTimer)
 
+	if opts.SchedulerReminders {
+		if opts.Config.SchedulerClients == nil {
+			return nil, fmt.Errorf("scheduler reminders are enabled, but no Scheduler clients are available")
+		}
+		log.Debug("Using Scheduler service for reminders.")
+		// We want to delete "a.actorsReminders" once we move to Scheduler service.
+		a.actorsReminders = reminders.NoOpReminders() // disable old reminder system if using Scheduler for reminders
```
ItalyPaleAle, line 244: Why wasn't this implemented using the `Reminders` interface? Deepanshu and I created that specifically so it could be used by other implementations. We have one and the idea was that the scheduler would use it too (and that's why it can't be deprecated). That would help keeping the code maintainable.

pkg/actors/actors.go
```diff
@@ -1175,6 +1287,27 @@ func (a *actorsRuntime) DeleteReminder(ctx context.Context, req *DeleteReminderR
 		return ErrReminderOpActorNotHosted
 	}
 
+	if a.schedulerClients != nil && a.schedulerReminders {
+		internalDeleteJobReq := &schedulerv1pb.DeleteJobRequest{
```
ItalyPaleAle, line 1291: Ditto, please migrate to a separate function.

But--this should have been implemented in a struct that implements the `Reminder` interface

pkg/actors/actors.go
```diff
@@ -1183,6 +1316,55 @@ func (a *actorsRuntime) GetReminder(ctx context.Context, req *GetReminderRequest
 		return nil, ErrReminderOpActorNotHosted
 	}
 
+	if a.schedulerClients != nil && a.schedulerReminders {
+		internalGetJobReq := &schedulerv1pb.GetJobRequest{
```
ItalyPaleAle, line 1320: Ditto

pkg/api/errors/errors.go
```diff
@@ -0,0 +1,80 @@
...
+		httpCode,
+		message,
+		legacyTag,
+	).
+		WithErrorInfo(reason, metadata).
+		Build()
+}
+
+func NotConfigured(name string, componentType string, metadata map[string]string, grpcCode codes.Code, httpCode int, legacyTag string, reason string) error {
+	message := fmt.Sprintf("%s %s is not configured", componentType, name)
```
ItalyPaleAle, line 46: fmt.Sprintf should be avoided for simle string concatenation

```suggestion
	message := componentType + " " + name + " is not configured"
```

pkg/scheduler/server/internal/pool.go
```diff
@@ -0,0 +1,212 @@
...
+			conns:     make(map[uint64]*conn),
+		}
+
+		p.nsPool[req.GetNamespace()] = nsPool
+	}
+
+	ok = true
+	var uuid uint64
+	for ok {
+		uuid = rand.Uint64() //nolint:gosec
```
ItalyPaleAle, line 94: Should listen to the linter :) math/rand doesn't produce good random numbers. Use crypto/rand instead.
(But this should probably use a true UUID which has ~128 bits of entropy instead of 64...)

dapr/proto/runtime/v1/dapr.proto
```diff
@@ -193,6 +193,15 @@ service Dapr {
   rpc RaiseEventWorkflowBeta1 (RaiseEventWorkflowRequest) returns (google.protobuf.Empty) {}
   // Shutdown the sidecar
   rpc Shutdown (ShutdownRequest) returns (google.protobuf.Empty) {}
+
+  // Create and schedule a job
+  rpc ScheduleJobAlpha1(ScheduleJobRequest) returns (google.protobuf.Empty) {}
```
ItalyPaleAle, line 198: 1. Dont' return Emtpy. Instead, create a new message `ScheduleJobAlpha1Response` and make it empty. Same below.
2. Don't forget `Alpha` in the Request type

dapr/proto/internals/v1/reminders.proto
```diff
@@ -35,3 +35,10 @@ message Reminder {
 message Reminders {
   repeated Reminder reminders = 1;
 }
+
+// TimerFiredEvent is the event that is sent to the actor when a timer fires.
+message TimerFiredEvent {
+  google.protobuf.Timestamp fireAt = 1;
```
ItalyPaleAle, line 41: `fire_at`

ItalyPaleAle (changes requested):

charts/dapr/charts/dapr_scheduler/templates/dapr_scheduler_poddisruptionbudget.yaml
cicoyle: based on sentry, looks like 6 is fine: https://github.com/dapr/dapr/blob/master/charts/dapr/charts/dapr_sentry/templates/dapr_sentry_poddisruptionbudget.yaml#L30

cmd/daprd/options/options.go
cicoyle: This is needed for the e2e tests to not break on the n-1 and n-2 versions since `scheduler-host-address` doesn't exist in those versions.

dapr/proto/internals/v1/reminders.proto
cicoyle: I will update the protos to ensure the additions use `_`

dapr/proto/runtime/v1/dapr.proto
cicoyle: I don't see this being done for other protos, I'd prefer consistency. For example:

```
  rpc SubtleGetKeyAlpha1(SubtleGetKeyRequest) returns (SubtleGetKeyResponse);
  // SubtleEncryptAlpha1 encrypts a small message using a key stored in the vault.
  rpc SubtleEncryptAlpha1(SubtleEncryptRequest) returns (SubtleEncryptResponse);
```

charts/dapr/charts/dapr_scheduler/templates/dapr_scheduler_poddisruptionbudget.yaml
ItalyPaleAle: Ok I misunderstood the code. The issue is that there are 8 spaces at the start of the line and it caused confusion.
```suggestion
      {{- toYaml . | nindent 6 }}
```

cmd/daprd/options/options.go
ItalyPaleAle: Not following sorry. That should be irrelevant for the n-1/n-2 tests? If it's passed or not...

charts/dapr/charts/dapr_scheduler/templates/dapr_scheduler_poddisruptionbudget.yaml
cicoyle: It works as is, are you saying there is an issue with indentation? It is consistent with the other pod disruption budgets in dapr.

cmd/daprd/options/options.go
cicoyle: It needs to include this line due to the n-1/n-2 e2e tests not knowing the scheduler flag.

charts/dapr/charts/dapr_scheduler/templates/dapr_scheduler_poddisruptionbudget.yaml
ItalyPaleAle: It works because there's the `-` at the start of `{{`, but it is _technically_ confusing (since the line has 8 spaces at the start instead of 6). So more like a nit now.

dapr/proto/runtime/v1/dapr.proto
ItalyPaleAle: Ok. At least don't use `google.protobuf.Empty` please! This also helps with the tracing code as it expects a type for each message.

dapr/proto/scheduler/v1/scheduler.proto
cicoyle: The type is for consistency with actor reminders due time, which is string.

pkg/actors/internal/reminders.go
yaron2: Indeed

dapr/proto/runtime/v1/dapr.proto
yaron2: Agree on not using Empty

pkg/actors/actors.go
cicoyle: Of all funcs in the reminders interface, only 3 relate to scheduler: create, get, delete. It didnt make sense to implement the interface. The abstraction assumes the sidecars are responsible to keeping reminders and triggering them, but distributed scheduler elevates the abstraction.

pkg/actors/actors.go
cicoyle: sure thing

pkg/actors/actors.go
cicoyle: sure thing I can mv to a separate func, but plz see my response above as to why we didn't implement the reminder interface

pkg/actors/actors.go
cicoyle: same reply as above

pkg/api/errors/errors.go
cicoyle: sure

cmd/scheduler/options/options.go
cicoyle: The comment or the `OrigArgs`?

ItalyPaleAle: The entire block where the args are transformed. That was only needed in daprd, not for control plane services

pkg/actors/actors.go
ItalyPaleAle: Please still implement the interface. Methods that do nothing can be no-ops. But using the interface allows the code to be much more readable and especially maintainable, since there are different implementations.

dapr/proto/runtime/v1/dapr.proto
ItalyPaleAle: Once a type is defined, it then needs to be added here for tracing: https://github.com/dapr/dapr/blob/master/pkg/proto/runtime/v1/dapr_tracing.go (it's ok to leave implementation as TODO for now)

daixiang0:
@cicoyle maybe we can make all UT pass then do perf test.

cicoyle:
> @cicoyle maybe we can make all UT pass then do perf test.

The unit tests are failing due to non Scheduler specific things. Specifically, its failing due to the dependency update. [See the fix in this PR](https://github.com/dapr/dapr/pull/7806). Also, there is some general flakiness in master with unit tests with the `informer` that is not related to Scheduler changes. [Workflows perf tests passed here](https://github.com/dapr/dapr/pull/7768#issuecomment-2158702313), we are doing some fine tuning gathering data points still.

charts/dapr/charts/dapr_scheduler/templates/dapr_scheduler_poddisruptionbudget.yaml
cicoyle: I appreciate the nit. For consistency sake, I will leave this as is because all the other pod disruption budgets in dapr (specifically the `operator`, `injector`, `sentry`, and `placement`) all have this exact config. To avoid confusion, maybe all of them should be updated with your suggestion. However, I would like to avoid updating all of them in the scheduler PR

cmd/scheduler/options/options.go
cicoyle: From what I can find, it looks like the other control plane services include this diff as well. [See the PR where you added this to all the other control plane services.](https://github.com/dapr/dapr/commit/ace5b645edc32c1f95f6e689d1f65a939c47153e#diff-b0e58d8b097ed95fbd37b293e3e629e39829972fc09efce2360a15348ea03914) I understand that since Scheduler is a new service its not necessary to maintain backwards compatability, but maybe for consistency sake across the services it should remain- thoughts?

cmd/scheduler/options/options.go
cicoyle: I'm hesitant to put `0.0.0.0` as a default value due to the security implications.

dapr/proto/runtime/v1/dapr.proto
cicoyle: I have the TODOs in the `dapr_tracing` file in this PR currently. I can update it to not use Empty.

cmd/scheduler/options/options.go
ItalyPaleAle: The default listen address for the other CP services is `0.0.0.0`. It won't work in Docker otherwise.

cmd/scheduler/options/options.go
ItalyPaleAle: I'd remove it. That block is only useful for legacy reasons

dapr/proto/runtime/v1/dapr.proto
ItalyPaleAle: Yes do not use Empty, because it makes things easier in the future. Ok to leave TODOs in the tracing file for now

dapr/proto/internals/v1/reminders.proto
cicoyle: https://github.com/cicoyle/dapr/pull/71

dapr/proto/runtime/v1/dapr.proto
cicoyle: https://github.com/cicoyle/dapr/pull/71

pkg/actors/internal/reminders.go
cicoyle: https://github.com/cicoyle/dapr/pull/71

cmd/scheduler/options/options.go
cicoyle: [please see this PR with that change](https://github.com/cicoyle/dapr/pull/71/files#diff-0de3b642b59f384f86664dea2875d7602cdc27323f3220286b52ce0126c758f8)

cmd/scheduler/options/options.go
yaron2: @cicoyle was this change request addressed?

cmd/scheduler/options/options.go
cicoyle: After further consideration, I left this as is since as a user I would not want anything listening on `0.0.0.0` by default bc if I forgot to change the default value then that is a security concern. Does anyone else feel strongly about this and could weigh in? cc/ @daixiang0 @yaron2 @artursouza @JoshVanL @mukundansundar

dapr/proto/scheduler/v1/scheduler.proto
cicoyle: This is used as an incremental counter, not really a `uuid`. I can change the name to be job `id` and put an explicit note in the comment description for it, or do you have a suggestion for a different name?

ItalyPaleAle: Yes please, rename it. `id` or `seq` (sequence number) could work. The name `uuid` implies it's a UUID which is not a number

pkg/scheduler/server/internal/pool.go
cicoyle: This is used as indexing rather than a uuid. I can update the name to just be id to avoid confusion. Since there are no security implication here and it is used for indexing purposes, I can change this to id++.

ItalyPaleAle: Ok `id++` is fine.

It's not just about security, it's also that `math/rand` doesn't guarantee good randomness, so conflicts can arise. If you can use an increment, that's better

cmd/scheduler/options/options.go
artursouza: Alessandro is correct that we need to listen to all addresses for this control plane to run from the container without having to change CLI to pass additional params (it would be a breaking change too for CLI to work with older Dapr versions or old CLI to work with newer Dapr versions).

All other control plane services (right now) listen on "" for this reason. I suggest we keep scheduler consistent (in this PR) and also listen on "". Then we have a separate PR to change all control planes (and the shared Dockerfile) to have them to listen to "" automatically from inside the docker image, without CLI having to pass additional params. We can have a new env var named DAPR_CONTROL_PLANE_LISTEN_ADDR or something that only control plane services will understand and it is set in Dockerfile globally.

JoshVanL: Software generally should not be made to listen on all interfaces and therefore be publicly available on the network by default. This is a clear security issue. Dapr services which are supported to be run from the CLI should default `localhost` for listening addresses. In the case of both kube and container target deployments, we have control over what arguments are given to the process via Helm and `dapr` CLI.

We should change `sentry`, `placement`, `daprd` and `scheduler` listen address defaults to `localhost`. I don't think there is a need for more config surface to be added because CLI flags already exist to change these values.

cicoyle: I've updated the default for Scheduler to be `""` for consistency. Although, we should revisit changing the defaults separate of the Scheduler PR

tests/integration/integration.go
```diff
@@ -74,7 +74,7 @@ func RunIntegrationTests(t *testing.T) {
 			options := tcase.Setup(t)
 
 			t.Logf("setting up test case")
-			ctx, cancel := context.WithTimeout(context.Background(), 45*time.Second)
+			ctx, cancel := context.WithTimeout(context.Background(), 2*time.Minute)
```
JoshVanL, line 77: Please revert this change. No integration test should be getting close to 45s, and is already a generous upper bound. This is here so that buggy code/tests don't make runs go on forever.

JoshVanL (changes requested):

cmd/scheduler/options/options.go
ItalyPaleAle: > Software generally should not be made to listen on all interfaces and therefore be publicly available on the network by default. 

While this is true for "regular" binaries, the difference is that Dapr control plane components are only meant to run inside Docker containers, even in self-hosted mode. We do not support running outside of a Docker container. Inside Docker, binding to 127.0.0.1 is not the desired behavior in almost all cases; if users want to make sure the container isn't exposed to the world, that should be done on Docker and not the binary.

artursouza: > Software generally should not be made to listen on all interfaces and therefore be publicly available on the network by default. This is a clear security issue. Dapr services which are supported to be run from the CLI should default `localhost` for listening addresses. In the case of both kube and container target deployments, we have control over what arguments are given to the process via Helm and `dapr` CLI.
> 
> We should change `sentry`, `placement`, `daprd` and `scheduler` listen address defaults to `localhost`. I don't think there is a need for more config surface to be added because CLI flags already exist to change these values.

Like I mentioned, requiring new params from CLI would be a breaking change. The CLI is meant to work with Dapr runtime version +-1, the container image should be done self-sufficient to run all binaries without extra params. I would not be in favor of breaking this compatibility if the cost is just a new env that users won't even need to set and come pre-set in the container image.

Anyway, for the purpose of this change, consistency is better than fixing for only one service and having to write on-off logic for scheduler.

artursouza: > > Software generally should not be made to listen on all interfaces and therefore be publicly available on the network by default.
> 
> While this is true for "regular" binaries, the difference is that Dapr control plane components are only meant to run inside Docker containers, even in self-hosted mode. We do not support running outside of a Docker container. Inside Docker, binding to 127.0.0.1 is not the desired behavior in almost all cases; if users want to make sure the container isn't exposed to the world, that should be done on Docker and not the binary.

Users can download the binaries and run it themselves in standalone mode without using Docker at all, but I agree that this would be rare and have not heard about it yet. Just a possible scenario.

artursouza: Moving the discussion on "if" and "how" to change control planes to listen to localhost to this issue: https://github.com/dapr/dapr/issues/7834

For the sake of this PR, can we all agree to keep consistency? 👍 or 👎

yaron2: I agree that it's a much better security posture to change to localhost across all control plane services, and that it should be done, outside the scope of this PR.

dapr/proto/runtime/v1/appcallback.proto
```diff
@@ -45,6 +46,9 @@ service AppCallback {
   // User application can save the states or send the events to the output
   // bindings optionally by returning BindingEventResponse.
   rpc OnBindingEvent(BindingEventRequest) returns (BindingEventResponse) {}
+
+  // Sends job back to the app's endpoint at trigger time.
+  rpc OnJobEvent (JobEventRequest) returns (JobEventResponse);
```
artursouza, line 51: This needs to be moved to AppCallbackAlpha since it is as Alpha API

artursouza (commented):
Only this comment. Let me try to make this change and merge it now.

dapr/proto/runtime/v1/appcallback.proto
artursouza: Just made the changes. Observing the build.

cicoyle:
Looks like the integration tests that are failing are unrelated: 
`Test_Integration/placement/vnodes/vnodes` && 
`Test_Integration/daprd/secret/http/componentName` && `Test_Integration/daprd/hotreload/operator/informer/scopes/components`

The e2e test `TestPubSubGRPC` failure is unrelated as well I believe

yaron2:
@cicoyle see minor conflict needs resolving

cicoyle:
looks like that int test is flaky and unrelated: `tests/integration.Test_Integration/placement/dissemination/notls`

PR #7789, opened 2024-06-03
feat(observability): allow custom latency distribution buckets

jfreeland:
# Description

This is intended to address https://github.com/dapr/dapr/issues/7641 and add support for users to provide an alternative list of distributions thresholds.

This was a bit tricky because when we use a static value for defautlLatencyDistribution we are consuming an identical instance of that variable to register our global views with OpenCensus.  In unit testing, we register the same views over and over again without resetting them, which is generally fine because we're using an identical value.  When we use a dynamic value, even though the names and distribution buckets are the same, the instance is different in the view of OpenCensus and we cannot re-register the same instance name.  I had created an alternative version based on a previous approach to demonstrate what we otherwise could have done, https://github.com/jfreeland/dapr/pull/1.  I think the solution here is a reasonable approach.

## Issue reference

Please reference the issue this PR will close: https://github.com/dapr/dapr/issues/7641

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Unit tests passing


jfreeland:
I was able to get unit tests to pass in https://github.com/jfreeland/dapr/pull/1, but this looks pretty strange.

I haven't yet fundamentally understood WHY it is that we did NOT have to unregister ANY latency views BEFORE my changes but we now have to unregister ALL latency views at all points during unit testing, because they are apparently NOW registered as global metrics but they were not before??

The primary thing that I've changed is that now to `InitMetrics` and subsequently `Init` each of the individual groupings' metrics we now have to explicitly pass in the latencyDistribution `*view.Aggregation`, which is either specified in the configuration or uses the default.

But somehow that has fundamentally changed how the metrics are getting treated in units tests when I run `make test`.

pkg/diagnostics/resiliency_monitoring_test.go
```diff
@@ -29,6 +29,19 @@ const (
...
+// type customMetricsExporter struct{}
+//
+// func (ce *customMetricsExporter) ExportView(vd *view.Data) {
+// 	log.Printf("vd.View: %+v\n%v\n", vd.View, vd.Rows)
+// 	for i, row := range vd.Rows {
+// 		log.Printf("\tRow: %d: %v\n", i, row)
+// 	}
+// 	log.Printf("StartTime: %s EndTime: %s\n\n", vd.Start.Round(0), vd.End.Round(0))
+// }
+
```
jfreeland, line 44: So I added this block of code and the `TODO` a couple lines below it in both this PR and in `master` and what I'm seeing is that latency views are simply not getting registered on `master`, while they are attempting to get registered repeatedly in this branch.  I have to clean them up in tests, https://github.com/jfreeland/dapr/pull/1, in order to not stomp on the namespace.

jfreeland:
I must be doing something really silly.  Maybe I'll get this solved tomorrow.

If I do something as trivial as https://github.com/dapr/dapr/pull/7794, I run into the same problem with `go test -v ./pkg/diagnostics/...  -tags=unit` failing.  As best I can tell, I'm doing the same thing in that example as what was happening before with https://github.com/dapr/dapr/blob/master/pkg/diagnostics/http_monitoring.go#L44.  I haven't yet figured out how or what is causing the `latency` view to not err on master but err on that example.

pkg/diagnostics/metrics.go
```diff
@@ -54,7 +101,7 @@ func InitMetrics(appID, namespace string, rules []config.MetricsRule, legacyMetr
 		return err
 	}
 
-	if err := DefaultHTTPMonitoring.Init(appID, legacyMetricsHTTPMetrics); err != nil {
+	if err := DefaultHTTPMonitoring.Init(appID, metricSpec.GetEnabled()); err != nil {
```
jfreeland, line 104: I think this is the correct replacement?  I just figured it'd be easier to pass the entire `metricsSpec`.

pkg/diagnostics/component_monitoring.go
```diff
@@ -166,27 +166,41 @@ func (c *componentMetrics) Init(appID, namespace string) error {
 	c.enabled = true
 	c.namespace = namespace
 
+	// This is a gross hack that facilitates unit testing.  Once it's set here
+	// the rest of the unit tests in the diagnostics package consume the
+	// identical global view that was first set here because alphabetically this
+	// file was encountered first.
+	//
+	// NOTE: It might be more practical to explicitly run this check in each of
+	// the diagnostic package _test.go files where we call x.Init.
```
jfreeland, line 175: This should actually say "it might make sense to run this check in each of the x.Init functions."

In normal runtime this would be set by InitMetrics.  In unit tests, this would be set by `component_monitoring_test.go` first.

pkg/diagnostics/metrics.go
jfreeland: Nope.  I should have gone back and checked.

yaron2:
@jfreeland please fix the conflicts and we can move into merging.

cc @filintod for review

pkg/config/configuration.go
```diff
@@ -269,10 +270,29 @@ func (m MetricSpec) GetHTTPIncreasedCardinality(log logger.Logger) bool {
 	return *m.HTTP.IncreasedCardinality
 }
 
+// GetLatencyDistribution returns a *view.Aggregration to be used for latency histograms
+func (m MetricSpec) GetLatencyDistribution(log logger.Logger) *view.Aggregation {
+	defaultLatencyDistribution := []float64{1, 2, 3, 4, 5, 6, 8, 10, 13, 16, 20, 25, 30, 40, 50, 65, 80, 100, 130, 160, 200, 250, 300, 400, 500, 650, 800, 1_000, 2_000, 5_000, 10_000, 20_000, 50_000, 100_000}
```
filintod, line 275: maybe this could be a variable and that way reuse in tests?

pkg/diagnostics/component_monitoring.go
filintod: you could add a initGlobals (or something like it) function (with a sync.once probably) that all x.Init will call. That way we don't rely on import ordering.  I also imagine adding custom Size distribution config in the future.

tests/integration/suite/daprd/metrics/http/distributionbuckets/low.go
```diff
@@ -0,0 +1,89 @@
...
+}
+
+func (l *low) Run(t *testing.T, ctx context.Context) {
+	l.daprd.WaitUntilRunning(t, ctx)
+
+	t.Run("service invocation", func(t *testing.T) {
+		l.daprd.HTTPGet2xx(t, ctx, "/v1.0/invoke/myapp/method/hi")
+		metrics := l.daprd.Metrics(t, ctx)
+		httpServerLatencyBuckets := 0
+		for k := range metrics {
```
filintod, line 82: could we create a slice with the values and compare expected slice with found one? that way we make sure the values are there and not just empty for example. that will also cover count implicitly

jfreeland:
Need revisiting after merge.

yaron2:
@jfreeland please see lint and test failures

pkg/diagnostics/global_monitoring.go
```diff
@@ -0,0 +1,74 @@
+/*
```
jfreeland, line 1: @filintod  - I don't particularly like this or think this is 'the way'.  I just moved this out to get it separate for now.  Might be best for you our someone more close to the code to take over or otherwise make this look how you prefer.

tests/integration/suite/daprd/metrics/http/distributionbuckets/low.go
jfreeland: I added a print statement here so we can see the values.  I don't know what the entry for infinity looks like in the list.

Because we're simply parsing a string, I think we're seeing on entry per `bucket=` label, so we should see a separate line for each `bucket=5, bucket=50, bucket=500, bucket=5000`, and a `>5000`.  I could add those to a slice as we see them and validate the values we see for these labels match `[]{5, 50, 500, 5000}`.

pkg/diagnostics/grpc_monitoring.go
```diff
@@ -53,8 +54,8 @@ type grpcMetrics struct {
 	clientRoundtripLatency *stats.Float64Measure
 	clientCompletedRpcs    *stats.Int64Measure
 
-	healthProbeCompletedCount  *stats.Int64Measure
-	healthProbeRoundripLatency *stats.Float64Measure
+	healthProbeCompletedCount   *stats.Int64Measure
+	healthProbeRoundtripLatency *stats.Float64Measure
```
jfreeland, line 58: The latest PR changed it to `RoundTrip`.  This should be `Roundtrip` to be consistent with the other metric names.

tests/integration/suite/daprd/metrics/http/distributionbuckets/low.go
jfreeland: Wrong.  `golangci-lint` failed.

pkg/diagnostics/global_monitoring.go
filintod: understood, but relying on import order is not robust enough. another idea could be to have a single Init for the package that calls all the individual xInits, no need to delay this PR more about it though. could be a future refactor PR. thanks for the feedback.

filintod: you know actually we already have that setup this way in https://github.com/dapr/dapr/blob/a04348b0099df1890a88627333d0b1ecb3534e70/pkg/diagnostics/metrics.go#L48. We could just call this globa init in this InitMetrics. 

this is the better place to call instead of each individual xxx.Init(), sorry got confused when it was mentioned about import order

filintod: I can see that each unit tests would need to call InitGlobals if moved out of each metric.Init, but that makes more sense that having it in each main path or in one that depends on certain order.  Another option, is to add an `init` method in one of the _test files to run this InitGlobals method in all the unit tests.

filintod (approved):
lgtm. there is a minor style refactor that can be done in a future PR.

jfreeland:
> lgtm. there is a minor style refactor that can be done in a future PR.

do you want to just take over or otherwise borrow the ideas here for a separate PR?  i don't think this is in a place that i'd be comfortable merging but i don't have time to look any further right now.

filintod:
@jfreeland I can take over with the minor refactor here

filintod:
couldn't push to your branch, created this one https://github.com/dapr/dapr/pull/7843

PR #7804, opened 2024-06-11
add global config override for increasedCardinality

filintod:
# Description

As part of the work defined in https://github.com/dapr/dapr/issues/7719#issuecomment-2115859125. Adding a global configuration override to help users to set this value in a central place as it is a cross cutting concern, instead of having to change all configurations

## Issue reference

Please reference the issue this PR will close: #7719 

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing

### DIFF
--- a/charts/dapr/charts/dapr_config/templates/dapr_default_config.yaml
+++ b/charts/dapr/charts/dapr_config/templates/dapr_default_config.yaml
@@ -18,6 +18,6 @@ spec:
   {{ if .Values.global.metrics.http }}
   metrics:
     http:
-      increasedCardinality: {{ default .Values.global.metrics.http.increasedCardinality false }}
+      increasedCardinality: {{ default false .Values.global.metrics.http.increasedCardinality }}
   {{- end }}
 {{- end }}
--- a/charts/dapr/charts/dapr_config/templates/dapr_default_config.yaml
+++ b/charts/dapr/charts/dapr_config/templates/dapr_default_config.yaml
@@ -15,4 +15,9 @@ spec:
     allowedClockSkew: {{ .Values.global.mtls.allowedClockSkew }}
     controlPlaneTrustDomain: {{ .Values.global.mtls.controlPlaneTrustDomain }}
     sentryAddress: {{ if .Values.global.mtls.sentryAddress }}{{ .Values.global.mtls.sentryAddress }}{{ else }}dapr-sentry.{{ .Release.Namespace }}.svc.cluster.local:443{{ end }}
+  {{ if .Values.global.metrics.http }}
+  metrics:
+    http:
+      increasedCardinality: {{ default .Values.global.metrics.http.increasedCardinality false }}
+  {{- end }}
 {{- end }}
--- a/charts/dapr/charts/dapr_config/templates/dapr_default_config.yaml
+++ b/charts/dapr/charts/dapr_config/templates/dapr_default_config.yaml
@@ -15,7 +15,7 @@ spec:
     allowedClockSkew: {{ .Values.global.mtls.allowedClockSkew }}
     controlPlaneTrustDomain: {{ .Values.global.mtls.controlPlaneTrustDomain }}
     sentryAddress: {{ if .Values.global.mtls.sentryAddress }}{{ .Values.global.mtls.sentryAddress }}{{ else }}dapr-sentry.{{ .Release.Namespace }}.svc.cluster.local:443{{ end }}
-  {{ if .Values.global.metrics.http }}
+  {{- if .Values.global.metrics.http }}
   metrics:
     http:
       increasedCardinality: {{ default false .Values.global.metrics.http.increasedCardinality }}
--- a/charts/dapr/values.yaml
+++ b/charts/dapr/values.yaml
@@ -35,6 +35,9 @@ global:
   prometheus:
     enabled: true
     port: 9090
+  metrics: {}
+  # http:
+  #   increasedCardinality: false # set to true to revert to old behavior by default (dapr < 1.14)
   mtls:
     enabled: true
     workloadCertTTL: 24h
--- a/pkg/operator/api/configurations.go
+++ b/pkg/operator/api/configurations.go
@@ -17,6 +17,9 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"os"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"k8s.io/apimachinery/pkg/types"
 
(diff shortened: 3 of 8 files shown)


charts/dapr/charts/dapr_config/templates/dapr_default_config.yaml
```diff
@@ -15,4 +15,9 @@ spec:
     allowedClockSkew: {{ .Values.global.mtls.allowedClockSkew }}
     controlPlaneTrustDomain: {{ .Values.global.mtls.controlPlaneTrustDomain }}
     sentryAddress: {{ if .Values.global.mtls.sentryAddress }}{{ .Values.global.mtls.sentryAddress }}{{ else }}dapr-sentry.{{ .Release.Namespace }}.svc.cluster.local:443{{ end }}
+  {{ if .Values.global.metrics.http }}
```
ItalyPaleAle, line 18:
```suggestion
  {{- if .Values.global.metrics.http }}
```

charts/dapr/values.yaml
```diff
@@ -35,6 +35,9 @@ global:
   prometheus:
     enabled: true
     port: 9090
+  metrics: {}
+  # http:
```
ItalyPaleAle, line 39: Please update the README.md too

charts/dapr/values.yaml
```diff
@@ -35,6 +35,9 @@ global:
   prometheus:
     enabled: true
     port: 9090
+  metrics: {}
```
ItalyPaleAle, line 38: This needs the `{}` removed

pkg/operator/api/configurations.go
```diff
@@ -25,6 +30,69 @@ import (
...
+
+	if c.Spec.MetricsSpec.HTTP != nil {
+		if c.Spec.MetricSpec == nil {
+			c.Spec.MetricSpec = &configurationapi.MetricSpec{}
+		}
+		c.Spec.MetricSpec.HTTP = c.Spec.MetricsSpec.HTTP
+	}
+}
+
+func (a *apiServer) overrideConfigDefaults(ctx context.Context, cfg *configurationapi.Configuration) error {
```
ItalyPaleAle, line 76: I'm not understanding what this code does. Why do we apply the daprSystem config to sidecars?

Also, what about non-K8s environments?

ItalyPaleAle (changes requested):

pkg/operator/api/configurations.go
filintod: @ItalyPaleAle The code mainly overrides, if not present on the regular config, config attributes (in this case only one but can be expanded in the future) present on the daprsystem config. Mostly useful for cross cutting concerns where you can define a common config to all.  Now saying this, I'm probably missing the use case of an app without a defined config but for which an operator still want to apply common config setting in the cluster different to dapr internal defaults.

is there something similar to daprsystem config in non-k8s environment? I could look to add to that if there is

charts/dapr/values.yaml
filintod: was following same pattern as later in this file like
```
  extraVolumes: {}
  #  sentry:
  #  placement:
  #  operator:
  #  injector:
```

ItalyPaleAle: I don't believe that's valid YAML syntax

pkg/operator/api/configurations.go
ItalyPaleAle: The daprSystem config is meant to configure the control plane, not individual sidecars.

In non-K8s there's no operator so this code path won't be executed. See the fact that it's not meant to configure individual sidecars.

charts/dapr/values.yaml
filintod: it is, an empty map in yaml, I can remove it but will have to change the others for consistency

ItalyPaleAle:
```
metrics: {}
```

is valid

... but if you uncomment the line below:

```
metrics: {}
   http:
```

is not

filintod: this is what I mean of consistency:
```
  metrics: {}
  # http:
  #   increasedCardinality: false # set to true to revert to old behavior by default (dapr < 1.14)
  mtls:
    enabled: true
    workloadCertTTL: 24h
    allowedClockSkew: 15m
    controlPlaneTrustDomain: "cluster.local"
    # If set to true, a bound service account token will be mounted and used to
    # authenticate to Sentry.
    mountSentryVolume: true
    # Used to override `dapr-sentry.{{ .Release.Namespace }}.svc.cluster.local:443`
    #sentryAddress:
  # extraVolumes and extraVolumeMounts are used to mount additional volumes to
  # the Dapr control plane pods. Useful for using alternative authentication
  # credentials to sentry.
  extraVolumes: {}
  #  sentry:
  #  placement:
  #  operator:
  #  injector:
  extraVolumeMounts: {}
  #  sentry:
  #  placement:
  #  operator:
  #  injector:
```

pkg/operator/api/configurations.go
JoshVanL: `mtls.enabled` is an example of an existing configuration today which configures sidecars.

ItalyPaleAle: Sort-of, as that is still something that also configures the control plane and the entire cluster

We don't have an existing case of using daprSystem for each individual sidecar, also because it doesn't work in standalone mode.

JoshVanL: `mTLS` is always enabled for control plane components in Kubernetes, so I would argue this setting is only for `daprd`s.

charts/dapr/values.yaml
filintod: @ItalyPaleAle actually this does not work and will lead to npe :
```
helm template dapr
Error: template: dapr/charts/dapr_config/templates/dapr_default_config.yaml:18:15: executing "dapr/charts/dapr_config/templates/dapr_default_config.yaml" at <.Values.global.metrics.http>: nil pointer evaluating interface {}.http
```

charts/dapr/charts/dapr_config/templates/dapr_default_config.yaml
filintod: done

charts/dapr/values.yaml
filintod: done

charts/dapr/values.yaml
ItalyPaleAle: What doesn't work?

filintod: removing the `{}`

ItalyPaleAle: Yes please do that.

Every other option is invalid

```yaml
# This is invalid
metrics:

# This is invalid too
metrics: {}
  http: ...
```

artursouza (commented):
LGTM

I think that there is a legit use case for global configuration of sidecars. I do have a concern of this being just a partial implementation and not have all the other settings to also be applied. I am OK to do that as a follow up in 1.15 since the cardinality option is the change with highest priority now.

pkg/operator/api/configurations.go
artursouza: I like the idea to have a central place to change sidecar config without having to manually override each Deployment with annotation.

Now, there is a different approach to this where we could simply allow a Configuration in the "dapr-system" namespace (or whatever the install NS is) to be tagged as the "default" configuration that the sidecar injector will use if Deployment does snot pick one. The pro is that it will be a much simpler change and easy to understand. The disadvantage is that users can only override their app's configuration with an entire new config instead of a merge.

In summary, my question is only if we want to have a "merge" of the default and app's configuration or to have a complete overwrite.

ItalyPaleAle: This still doesn't solve the problem of Dapr self-hosted. My understanding is that we the project's goal is to maintain feature parity between the two

PR #7806, opened 2024-06-11
Up grpc version

cicoyle:
When updating the grpc version, I noticed the test failing due to the return value changing

There is a note in the [v1.62.0 dependencies section](https://github.com/grpc/grpc-go/releases/tag/v1.62.0) about:
image

Therefore, I've updated the unit test accordingly since:
- previously, when a nil message was passed to the Marshal function, it resulted in an error.
This means if the gRPC server method returned nil, nil, it would cause the server to return an error due to the marshaling failure. 
- now, when a nil message is passed to the Marshal function, it outputs zero bytes without an error.
As a result, if the gRPC server method returns nil, nil, the server now returns an empty message without an error.


JoshVanL (approved):

cicoyle:
last commit was to fix this panic: https://github.com/dapr/dapr/actions/runs/9470168103/job/26090820065?pr=7806

go.mod
```diff
@@ -63,9 +63,9 @@ require (
 	golang.org/x/exp v0.0.0-20240119083558-1b970713d09a
 	golang.org/x/net v0.24.0
 	golang.org/x/sync v0.6.0
-	google.golang.org/genproto/googleapis/api v0.0.0-20231120223509-83a465c0220f
-	google.golang.org/genproto/googleapis/rpc v0.0.0-20231212172506-995d672761c0
-	google.golang.org/grpc v1.60.1
+	google.golang.org/genproto/googleapis/api v0.0.0-20240304212257-790db918fca8
+	google.golang.org/genproto/googleapis/rpc v0.0.0-20240304212257-790db918fca8
+	google.golang.org/grpc v1.62.1
```
mikeee, line 68: The upgraded version is still vulnerable to a recently fixed DOS exploit.
https://github.com/grpc/grpc-go/releases/tag/v1.62.2

mikeee (changes requested):
Should we go to 1.64.0 (we would need to change the way we initialize connections)

go.mod
cicoyle: I see. I'll change it to 1.62.2 then, but don't want to address the deprecations with upgrading to 1.64 quite yet. This PR was just to address a unit test failure with the Scheduler branch in isolation.

mikeee (approved):
lgtm

artursouza (approved):

marcduiker:
@holopin-bot @cicoyle Enjoy your Dapr badge!

PR #7817, opened 2024-06-17
Fix Membership Unit Test Flakiness

cicoyle:
Update membership test to fix unit test flakiness. Should be no diff other than indentation and wrapping the `membercount` call in the `require.EventuallyWithT`

PR #7818, opened 2024-06-17
Fix Unit Test General Flakiness

cicoyle:
Working to fix the flakiness in unit tests in CI. They pass locally, so need to test a few things in CI since the behavior is different.


cicoyle:
Looks like the last unit test failure is due to the membership test, [which is fixed here](https://github.com/dapr/dapr/pull/7817)

pkg/operator/api/informer/informer_test.go
```diff
@@ -75,6 +75,11 @@ func Test_WatchUpdates(t *testing.T) {
 			},
 			operator.ResourceEventType_UPDATED,
 		)
+
+		require.Eventually(t, func() bool {
+			return i.batchID == 1
+		}, 5*time.Second, 100*time.Millisecond)
```
JoshVanL, line 81: We want to print out the values if it does fail. This also has a data race so we need to change `i.batchID` to be a `atomic.Uint32.

```suggestion
		require.EventuallyWithT(t, func(c *assert.CollectT) {
			assert.Equal(t, 1, i.batchID)
		}, 5*time.Second, 100*time.Millisecond)
```

pkg/operator/api/informer/informer_test.go
```diff
@@ -86,6 +91,11 @@ func Test_WatchUpdates(t *testing.T) {
 			},
 			operator.ResourceEventType_UPDATED,
 		)
+
+		require.Eventually(t, func() bool {
+			return i.batchID == 2
```
JoshVanL, line 96: Same here.

pkg/operator/api/informer/informer_test.go
```diff
@@ -94,6 +104,10 @@ func Test_WatchUpdates(t *testing.T) {
 			operator.ResourceEventType_CREATED,
 		)
 
+		require.Eventually(t, func() bool {
+			return i.batchID == 3
```
JoshVanL, line 108: Same here.

JoshVanL (changes requested):

cicoyle:
it looks like the memory utilization actually decreased, but idk why the go routines upped by 1 for this diff:
`75th percentile for number of go routines (new): 37.00`

pkg/diagnostics/resiliency_monitoring_test.go
```diff
@@ -407,7 +407,7 @@ func TestResiliencyActivationsCountMonitoring(t *testing.T) {
 					})
 				}
 				// let the circuit breaker to go to half open state (5x cb timeout) and then return success to close it
-				time.Sleep(500 * time.Millisecond)
+				time.Sleep(1000 * time.Millisecond)
```
cicoyle, line 410: I think its just a timing issue for this: https://github.com/dapr/dapr/actions/runs/9570009389/job/26384075690?pr=7818

JoshVanL (approved):

PR #7820, opened 2024-06-18
HTTP Metrics: Add verbs back and preserve static paths

nelson-parente:
# Description

This PR addresses a part of the discussion initiated in issue [#7719](https://github.com/dapr/dapr/issues/7719).

Currently, in low cardinality, Dapr metrics perform a gRPC-like conversion on the method label. Since Dapr is moving away from this conversion (as discussed [here](https://github.com/dapr/dapr/issues/7719#issuecomment-2130730650)), we propose adding back the HTTP method to the metrics (GET, POST, PUT, etc.).

The new config option `excludeVerbs` allows users to configure whether HTTP verbs are added or not to the metrics. The default for this option is false, by default these should be included, that's why I decided to call the config ExcludeVerbs instead of IncludeVerbs, this way it's more explicit the act of excluding verbs from the metrics.

```
apiVersion: dapr.io/v1alpha1
kind: Configuration
metadata:
  name: excludeverbs
spec:
  metrics:
    enabled: true
    http:
      excludeVerbs: true
```

This change addresses one of the current issues with Dapr HTTP metrics: all HTTP calls are grouped into the same bucket. For instance, if my app has GET, POST, and PUT methods on the same resource, these are all exposed as InvokeService/myapp-id, making it difficult to differentiate between them.

```
OLD - 3 HTTP calls to GET/POST/PUT on the same resource would end all in the same bucket: 
dapr_http_server_request_count{app_id="my-service",method="InvokeService/my-service",status="200",le="100"} 3

NEW - 3 HTTP calls to GET/POST/PUT on the same resource: 
dapr_http_server_request_count{app_id="my-service",method="GET", path="", status="200",le="100"} 1
dapr_http_server_request_count{app_id="my-service",method="PUT", path="", status="200",le="100"} 1
dapr_http_server_request_count{app_id="my-service",method="POST", path="", status="200",le="100"} 1

NEW - 3 HTTP calls to GET/POST/PUT on the same resource with ExcludeVerbs == true: 
dapr_http_server_request_count{app_id="my-service",method="", path="", status="200",le="100"} 3
```
With this update, we will reintroduce the method labels, allowing us to distinguish between different HTTP methods in the metrics while maintaining constrained cardinality in paths. The path keys are also reintroduced as empty strings. Although this solution is not perfect (as it does not distinguish between different paths), it represents a more natural iteration towards HTTP protocol metrics. Users can further refine this with path matching, as proposed [here](https://github.com/dapr/proposals/pull/58).

Additionally, this PR preserves well-known dapr paths, I identified this initial list of static dapr paths:
```
"/dapr/config"
"/dapr/metrics"
"/dapr/subscribe"
"/healthz":    
```

## Issue reference

https://github.com/dapr/dapr/issues/7719

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing

### DIFF
--- a/charts/dapr/crds/configuration.yaml
+++ b/charts/dapr/crds/configuration.yaml
@@ -262,6 +262,9 @@ spec:
                         type: array
                         items:
                           type: string
+                      excludeVerbs:
+                        description: If true (default is false) HTTP verbs (e.g., GET, POST) are excluded from the metrics.
+                        type: boolean
                     type: object
                   rules:
                     items:
@@ -315,6 +318,9 @@ spec:
                         type: array
                         items:
                           type: string
+                      excludeVerbs:
+                          description: If true (default is false) HTTP verbs (e.g., GET, POST) are excluded from the metrics.
+                          type: boolean
                     type: object
                   rules:
                     items:
--- a/pkg/apis/configuration/v1alpha1/types.go
+++ b/pkg/apis/configuration/v1alpha1/types.go
@@ -224,6 +224,9 @@ type MetricHTTP struct {
 	// TODO: [MetricsCardinality] Change default in 1.15+
 	// +optional
 	IncreasedCardinality *bool `json:"increasedCardinality,omitempty"`
+	// If true (default is false) HTTP verbs (e.g., GET, POST) are excluded from the metrics.
+	// +optional
+	ExcludeVerbs *bool `json:"excludeVerbs,omitempty"`
 	// +optional
 	PathMatching []string `json:"pathMatching,omitempty"`
 }
--- a/pkg/apis/configuration/v1alpha1/types.go
+++ b/pkg/apis/configuration/v1alpha1/types.go
@@ -224,11 +224,11 @@ type MetricHTTP struct {
 	// TODO: [MetricsCardinality] Change default in 1.15+
 	// +optional
 	IncreasedCardinality *bool `json:"increasedCardinality,omitempty"`
+	// +optional
+	PathMatching []string `json:"pathMatching,omitempty"`
 	// If true (default is false) HTTP verbs (e.g., GET, POST) are excluded from the metrics.
 	// +optional
 	ExcludeVerbs *bool `json:"excludeVerbs,omitempty"`
-	// +optional
-	PathMatching []string `json:"pathMatching,omitempty"`
 }
 
 // MetricsRule defines configuration options for a metric.
--- a/pkg/config/configuration.go
+++ b/pkg/config/configuration.go
@@ -296,12 +296,6 @@ type MetricHTTP struct {
 	ExcludeVerbs         *bool    `json:"excludeVerbs,omitempty" yaml:"excludeVerbs,omitempty"`
 }
 
-// PathMatching defines configuration options for path matching.
-type PathMatching struct {
-	IngressPaths []string `json:"ingress,omitempty" yaml:"ingress,omitempty"`
-	EgressPaths  []string `json:"egress,omitempty" yaml:"egress,omitempty"`
-}
-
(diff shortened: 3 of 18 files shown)


ItalyPaleAle:
Can you make this optional please, or offer an option to turn it off? For some users the increased memory usage isn't worth knowing the difference between GET/POST/PUT etc

artursouza:
> Can you make this optional please, or offer an option to turn it off? For some users the increased memory usage isn't worth knowing the difference between GET/POST/PUT etc

For HTTP, having this information is super important. Maybe having a flag to disable HTTP metrics entirely if memory is the only factor a specific user cares about.

artursouza (dismissed):

artursouza:
@nelson-parente lint is failing

ItalyPaleAle:
> For HTTP, having this information is super important.

it's fine if it's enabled by default, but adding an option to disable it shouldn't be too hard and is helpful to some users

nelson-parente:
I understand the concern about memory usage, but adding a configuration option to disable HTTP metrics might not be necessary. This change doesn't significantly increase memory usage because we have bounded HTTP methods and static paths, so memory shouldn't be an issue here.

For users concerned about memory, we already have the metrics.enable option available. It seems inconsistent to have only HTTP metrics be individually disabled while others remain initialized.

If we decide we really want this option, I'd suggest moving it to a different PR and applying it to all metrics we have (gRPC, HTTP, component, resiliency, and workflow).

JoshVanL:
I think it makes sense to have an option to toggle whether verbs are attached to metrics.

We _must_ only accepts verbs from a well-known set, with any unrecognized verb being put in an `UNKNOWN` bucket. As it stands it looks like we currently have unbound cardinality from the `method` label.

nelson-parente:
> I think it makes sense to have an option to toggle whether verbs are attached to metrics.
> 
> We _must_ only accepts verbs from a well-known set, with any unrecognized verb being put in an `UNKNOWN` bucket. As it stands it looks like we currently have unbound cardinality from the `method` label.

Thank you for the feedback. I will add the check for valid HTTP verbs, as I expected requests with invalid verbs to be rejected early on, but testing locally shows they are currently accepted.

Regarding the configuration option to toggle verbs, I still believe that including the verb always makes sense. Without it, metrics like `dapr_http_server_request_count{app_id="my-service", path="", status="200",le="100"}` provide limited value to the user, as they lack essential HTTP context.

ItalyPaleAle:
> Regarding the configuration option to toggle verbs, I still believe that including the verb always makes sense. 

For many users, for sure. When the low-cardinality metrics feature was implemented, we had user requirements to not have pivots for each HTTP verb. If you could add a config flag (even if off by default) that would be very important

yaron2:
> > Regarding the configuration option to toggle verbs, I still believe that including the verb always makes sense.
> 
> For many users, for sure. When the low-cardinality metrics feature was implemented, we had user requirements to not have pivots for each HTTP verb. If you could add a config flag (even if off by default) that would be very important

@nelson-parente Please add this config option. I accept this is valid user feedback

nelson-parente:
I will add the config 🙂 

Should this config be specific to low cardinality, or will it also affect high cardinality if enabled?

nelson-parente:
Added the new config, once I have positive feedback on this I will update my other docs PR :)

artursouza (approved):

nelson-parente:
Tests were failing after [this](https://github.com/dapr/dapr/pull/7826) was merged, fix 👍

artursouza (approved):

marcduiker:
@holopin-bot @nelson-parente Thank you Nelson!

PR #7825, opened 2024-06-19
[Temporary Test] Scheduler workflow perf test app + original workflow perf test app

cicoyle:
Updating the workflow perf test to have 2 apps, one for the original workflow test, and one for the workflows using scheduler under the hood. Going to test this in CI rq


cicoyle:
Just needed to test my perf test changes. Closing since they passed :)

PR #7826, opened 2024-06-19
Revert "Change default cardinality to low for metrics (#7522)"

artursouza:
# Description

This reverts commit c3baea50a19e8e00793f5d79929edce45fc4c5e2.

Proposal to postpone decision to make low cardinality default, done in https://github.com/dapr/dapr/pull/7522

## Issue reference

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing

### DIFF
--- a/charts/dapr/crds/configuration.yaml
+++ b/charts/dapr/crds/configuration.yaml
@@ -253,8 +253,9 @@ spec:
                       the HTTP server
                     properties:
                       increasedCardinality:
-                        description: |-
-                          If false (the default), metrics for the HTTP server are collected with increased cardinality.
+                        description: 'If true, metrics for the HTTP server are collected
+                          with increased cardinality. The default is true in Dapr 1.13,
+                          but will be changed to false in 1.15+'
                         type: boolean
                       pathMatching:
                         description: PathMatching defines the path matching configuration for HTTP server metrics.
--- a/pkg/apis/configuration/v1alpha1/types.go
+++ b/pkg/apis/configuration/v1alpha1/types.go
@@ -219,7 +219,9 @@ type MetricSpec struct {
 
 // MetricHTTP defines configuration for metrics for the HTTP server
 type MetricHTTP struct {
-	// If false (the default), metrics for the HTTP server are collected with increased cardinality.
+	// If false, metrics for the HTTP server are collected with increased cardinality.
+	// The default is true in Dapr 1.13, but will be changed to false in 1.15+
+	// TODO: [MetricsCardinality] Change default in 1.15+
 	// +optional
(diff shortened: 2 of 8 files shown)


charts/dapr/crds/configuration.yaml
```diff
@@ -253,8 +253,9 @@ spec:
                       the HTTP server
                     properties:
                       increasedCardinality:
-                        description: |-
-                          If false (the default), metrics for the HTTP server are collected with increased cardinality.
+                        description: 'If true, metrics for the HTTP server are collected
+                          with increased cardinality. The default is true in Dapr 1.13,
+                          but will be changed to false in 1.14+'
```
artursouza, line 258: I can update this message too, just waiting for this run to finish.

ItalyPaleAle, line 258: You may need to update this

pkg/apis/configuration/v1alpha1/types.go
```diff
@@ -219,7 +219,9 @@ type MetricSpec struct {
 
 // MetricHTTP defines configuration for metrics for the HTTP server
 type MetricHTTP struct {
-	// If false (the default), metrics for the HTTP server are collected with increased cardinality.
+	// If false, metrics for the HTTP server are collected with increased cardinality.
+	// The default is true in Dapr 1.13, but will be changed to false in 1.14+
+	// TODO @ItalyPaleAle [MetricsCardinality] Change default in 1.14
```
ItalyPaleAle, line 224: Update this

ItalyPaleAle (changes requested):

yaron2:
Context:

After testing the low cardinality mode as default for a while and having received a satisfactory amount of user feedback, it becomes clear that while this enabling or disabling by default cannot be summarized into a right/wrong decision and both carry pros and cons that will better optimize for a certain class of users. The overall feedback received is that users want the ability to easily turn off high cardinality metrics if they need to, and a subset of users would very much benefit more granular control over the metrics configuration, such as path matching and regex, to allow power users to tailor metrics to their needs. With the implications of making metrics less usable out of the box for new users and potentially disrupting dashboards, monitoring and observability platform ingesting Dapr metrics of existing users, it would be preferable to follow how way sister projects like Istio and Linkerd handle the situation, which is to keep high cardinality as default and provide users with the best experience of opting out or arriving at a desired custom configuration.

cicoyle (approved):
Changes look fine, but there are some failing int tests that I believe should be passing before this gets merged

artursouza:
> Changes look fine, but there are some failing int tests that I believe should be passing before this gets merged

I think I fixed now. Let's see.

yaron2 (approved):

PR #7830, opened 2024-06-20
Add libc to docker image

yaron2:
Changes the base image to `base-nossl` which contains all the dependencies in `static` plus libc. This enabled Dapr to work with advanced monitoring tools such as Dynatrace and helps users uncover issues faster.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,6 +1,6 @@
 # current directory must be ./dist
 
-FROM gcr.io/distroless/static:nonroot
+FROM gcr.io/distroless/base-nossl:nonroot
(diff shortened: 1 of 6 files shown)


artursouza (approved):

ItalyPaleAle:
This is not an easy change actually and wish there was some discussion. Adding libc to the container means that now we have an additional binary in the container for which there can be CVEs. We are now responsible to re-publish container images just because the base image has changed due to security vulnerability, or users' security scanners will flag this.

Couldn't this have been made optional? A different variant?

yaron2:
Yeah I can add this as a variant

Issue #7490, opened 2024-02-02
Error Standardization: Placement API


cicoyle:
Richer errors in Dapr for the **Placement API**.

**Note**: Please clone this issue and assign the clone of this issue to yourself if you would like to contribute to the error standardization efforts.

# Scope of work

- [ ] Choose 5-10 errors in the Placement API to enrich. Ideally, cover as many errors as possible from the API.
- [ ] Follow the steps outlined in [this README](https://github.com/dapr/dapr/blob/master/pkg/api/errors/README.md) for each error being updated

# Definition of done

- [ ] Update 5-10 errors in the Placement API
- [ ] Update the unit tests, if necessary
- [ ] Write integration tests to cover each updated errors

# Keep in mind

Unless a major version is being released and there is a solid reason for introducing backwards incompatibility, we can't change the API, meaning the HTTP and gRPC status codes, and the error codes must stay the same. An exception to this is the gRPC `StatusCode.UNKNOWN` which should be converted to a specific status.

See the [kit errors pkg README](https://github.com/dapr/kit/blob/main/errors/README.md) for another usage example. For each error being updated, you can use the following helper functions to add error details, see the kit [errors pkg](https://github.com/dapr/kit/tree/main/errors) for more details:

- `WithHelpLink()`
-  `WithHelp()`
-  `WithErrorInfo()` - **Required**
-  `WithFieldViolation()`
-  `WithDetails()`

# Reference PRs
Existing error standardization for: 
- [State API](https://github.com/dapr/dapr/pull/7257)
- [PubSub API](https://github.com/dapr/dapr/pull/7322)

fowlerlee:
Hi @cicoyle , I would like to work on this ticket, since no one has commented, may I continue with this work?

fowlerlee:
Hi @cicoyle, I made this [WIP] PR to get early feedback, to know if I am on the right path. I added the test based on what I saw in the integration tests in tests/integration/suite/placement/apilevel/... as well as the commit on #7257, where the author replaced several errors of the form `ErrBadXX = APIError{"invalid request: %v", "ERR_BAD_XX", http.StatusBadRequest, grpcCodes.InvalidArgument}`. What I don't see however is typical errors from pkg/messages/predefined.go, related to the Placement API, which can be converted to the Google API guide format for Errors. Could you guide me where I can find typical errors that should be handled or converted to the mentioned format? Thank you.

by-nelson:
[...]

PR #7835, opened 2024-06-22
added rich error handling for placement api

fowlerlee:
# Description

This PR addresses the rich error handling for the placement api issue https://github.com/dapr/dapr/issues/7490.
The error handling follows the form specified **[here](https://github.com/dapr/dapr/blob/master/pkg/api/errors/README.md)**   

## Issue reference

Please reference the issue this PR will close:  **[7490]( https://github.com/dapr/dapr/issues/7490)** 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [ ] Unit tests passing


cicoyle (commented):
Thanks @fowlerlee for this valuable addition. I think this can really help users gain more visibility into these errors. Mind adding some integration tests forcing these errors? I want to ensure we don't change any of the http/grpc error codes.

elena-kolevska:
@fowlerlee for reference, here's an integration test that checks one of the errors you updated (although it's in a different context): https://github.com/dapr/dapr/blob/master/tests/integration/suite/placement/authz/mtls.go#L110

Since the placement tests are heavier to run (they spin up more processes) I'm ok with checking the correctness of the errors inside of existing tests. So, instead of setting up different scenarios just to test the errors, use the existing scenarios to also check the details of the returned error.

pkg/placement/placement.go
```diff
@@ -234,12 +236,12 @@ func (p *Service) ReportDaprStatus(stream placementv1pb.Placement_ReportDaprStat
...
 	clientID, err := p.validateClient(stream)
 	if err != nil {
-		return err
+		return apiErrors.PlacementServiceInternalError(fmt.Sprintf("Internal Error: %s", codes.Internal))
 	}
 
 	firstMessage, err := p.receiveAndValidateFirstMessage(stream, clientID)
 	if err != nil {
-		return err
+		return apiErrors.PlacementServiceUnAuthenticated(fmt.Sprintf("Failed to get client ID from context with code: %s", codes.Unauthenticated))
```
elena-kolevska, line 244: `p.receiveAndValidateFirstMessage` can return `codes.PermissionDenied` or `codes.Internal`  status codes with a specific error message, but here you're overriding them and replacing it with `codes.Unauthenticated`.

pkg/placement/placement.go
```diff
@@ -269,7 +271,7 @@ func (p *Service) ReportDaprStatus(stream placementv1pb.Placement_ReportDaprStat
 		case nil:
 
 			if clientID != nil && req.GetId() != clientID.AppID() {
-				return status.Errorf(codes.PermissionDenied, "client ID %s is not allowed", req.GetId())
+				return apiErrors.PlacementServicePermissionDenied(fmt.Sprintf("Failed wih code %s. Client ID %s is not allowed", codes.PermissionDenied, req.GetId()))
```
elena-kolevska, line 274: Same comment as above, let's follow the go styling guide and start all error messages with lowercase letters. Also, no need to wrap the error message with "Failed with code"

pkg/placement/placement.go
```diff
@@ -182,16 +184,16 @@ func NewPlacementService(opts PlacementServiceOpts) *Service {
 // Blocks until the service is closed and all connections are drained.
 func (p *Service) Run(ctx context.Context, listenAddress, port string) error {
 	if p.closed.Load() {
-		return errors.New("placement service is closed")
+		return apiErrors.PlacementServiceIsClosedOnRun("placement service is closed")
```
elena-kolevska, line 187: The errors in the Start function are not part of the API, and we don't need to transform them into richer grpc errors. @JoshVanL would you agree?

pkg/placement/placement.go
```diff
@@ -234,12 +236,12 @@ func (p *Service) ReportDaprStatus(stream placementv1pb.Placement_ReportDaprStat
 
 	clientID, err := p.validateClient(stream)
 	if err != nil {
-		return err
+		return apiErrors.PlacementServiceInternalError(fmt.Sprintf("Internal Error: %s", codes.Internal))
```
elena-kolevska, line 239: Same comment as below, please propagate the original errors returned in `p.validateClient`

elena-kolevska (changes requested):
Thanks for your PR @fowlerlee! A few comments from me.
There's also some errors that are returned but have not been converted to the new format.

pkg/placement/placement.go
JoshVanL: Yes, that's correct. We don't and should have a typed API error here.

pkg/api/errors/placement.go
```diff
@@ -0,0 +1,110 @@
...
+	PlacementServiceAlreadyRunning = "ALREADY_RUNNING"
+
+	CodePrefixPlacement = "DAPR_PLACEMENT_"
+)
+
+func PlacementServiceIsClosedOnRun(msg string) error {
+	return kiterrors.NewBuilder(
+		codes.Unavailable,
+		http.StatusExpectationFailed,
+		fmt.Sprintln(msg),
```
elena-kolevska, line 36:
```suggestion
		msg,
```

pkg/api/errors/placement.go
```diff
@@ -0,0 +1,110 @@
...
+		CodePrefixPlacement+kiterrors.CodePostfixQueryFailed,
+	).WithErrorInfo(CodePrefixPlacement+kiterrors.CodePostfixQueryFailed, nil).
+		Build()
+}
+
+func PlacementServiceIsAlreadyRunning(msg string) error {
+	return kiterrors.NewBuilder(
+		codes.AlreadyExists,
+		http.StatusExpectationFailed,
+		fmt.Sprintln(msg),
```
elena-kolevska, line 46:
```suggestion
		msg,
```

pkg/api/errors/placement.go
elena-kolevska: ...and similar for the other instances.

pkg/api/errors/placement.go
fowlerlee: Thanks for the review and tips @elena-kolevska: regarding the integration tests, i've been trying to figure out what to do from mtls.go and follower.go, but I'm not able to figure how to force the errors for the Service struct in the integration tests. I'm probably just missing some understanding, but I'm keen to get it right, if I can get some tips to bridge my understanding.

fowlerlee: I am still keen to fix these integration tests, but am stuck with how to force the errors, as stated previously. If I could get some advice I would jump on it asap.

elena-kolevska: Hey @fowlerlee , did you see my previous comment where I linked to an example I prepared for the follower error?
https://github.com/elena-kolevska/dapr/commit/81a00b3c1eb53f59ccbdbc59050080070ffa383d
Testing some of these errors is indeed quite complex, I can take a look at the PR again and see if I can find a way to force that Sentry error, for example.

elena-kolevska: @fowlerlee can we address the outstanding comments about the code first before proceeding to the tests? Otherwise, we might end up having to change the tests again after we change the code.
For example, the one about the different errors in `validateClient` is really important.

fowlerlee: @elena-kolevska, I have pushed fixes for outstanding comments now. It's on my branch and ready for review. Thank you for your time.

pkg/placement/placement.go
```diff
@@ -266,12 +268,12 @@ func (p *Service) ReportDaprStatus(stream placementv1pb.Placement_ReportDaprStat
 
 	clientID, err := p.validateClient(stream)
 	if err != nil {
-		return err
+		return apiErrors.PlacementServiceInternalError(fmt.Sprintf("internal or unauthenticated error: %s", err))
```
elena-kolevska, line 271: We can't replace the error `validateClient` returns here, because it could be different error types. Please take a closer look at the function and update the errors in there directly.

pkg/placement/placement.go
```diff
@@ -266,12 +268,12 @@ func (p *Service) ReportDaprStatus(stream placementv1pb.Placement_ReportDaprStat
...
 	}
 
 	firstMessage, err := p.receiveAndValidateFirstMessage(stream, clientID)
 	if err != nil {
-		return err
+		return apiErrors.PlacementServicePermissionDenied(fmt.Sprintf("internal of permission denied error: %s", err))
```
elena-kolevska, line 276: Same here. We should respect the error `receiveAndValidateFirstMessage` returns.

pkg/placement/placement.go
```diff
@@ -320,13 +322,13 @@ func (p *Service) ReportDaprStatus(stream placementv1pb.Placement_ReportDaprStat
...
 			if clientID != nil && req.GetId() != clientID.AppID() {
-				return status.Errorf(codes.PermissionDenied, "client ID %s is not allowed", req.GetId())
+				return apiErrors.PlacementServicePermissionDenied(fmt.Sprintf("client ID %s is not allowed", req.GetId()))
 			}
 
 			if registeredMemberID == "" {
 				registeredMemberID, err = p.handleNewConnection(req, daprStream, namespace)
 				if err != nil {
-					return err
+					return apiErrors.PlacementServiceFailedPrecondition(fmt.Sprintln("handle connection error: ", err))
```
elena-kolevska, line 331: Same here.

elena-kolevska (changes requested):

elena-kolevska:
Also, please fix the linter issues so the tests can run. I'm expecting to see integration tests failing because some of the error codes changed. And then we iterate and improve until they pass again :)

elena-kolevska (changes requested):
Here are a few more comments. Don't forget to run `make test` for unit tests and `make test-integration` for all integration tests while developing. If they're failing, you'll know you still have a few more tweaks to do.
For quicker iteration, you can run only placement related tests:
```
cd tests/integration
go test --race -v --tags integration --focus="placement"
```

And for unit tests, run everything in the `pkg/actors/ and `pkg/placement` directories.
And finally, don't forget the linter. Otherwise, CI won't run.

PR #7843, opened 2024-06-25
custom latency distribution

filintod:
# Description

Follow up of https://github.com/dapr/dapr/pull/7789

## Issue reference

Please reference the issue this PR will close: #7641 

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing


pkg/config/configuration.go
```diff
@@ -271,6 +272,23 @@ func (m MetricSpec) GetHTTPIncreasedCardinality(log logger.Logger) bool {
 	return *m.HTTP.IncreasedCardinality
 }
 
+// GetLatencyDistribution returns a *view.Aggregration to be used for latency histograms
+func (m MetricSpec) GetLatencyDistribution(log logger.Logger) *view.Aggregation {
+	defaultLatencyDistribution := []float64{1, 2, 3, 4, 5, 6, 8, 10, 13, 16, 20, 25, 30, 40, 50, 65, 80, 100, 130, 160, 200, 250, 300, 400, 500, 650, 800, 1_000, 2_000, 5_000, 10_000, 20_000, 50_000, 100_000}
```
artursouza, line 277: Is this the default today?

filintod: from https://github.com/dapr/dapr/blob/master/pkg/diagnostics/http_monitoring.go#L48

defaultLatencyDistribution = view.Distribution(1, 2, 3, 4, 5, 6, 8, 10, 13, 16, 20, 25, 30, 40, 50, 65, 80, 100, 130, 160, 200, 250, 300, 400, 500, 650, 800, 1_000, 2_000, 5_000, 10_000, 20_000, 50_000, 100_000)

filintod:
parallel integration tests seems flaky locally and probably in CI as well

pkg/api/grpc/proxy/handler_test.go
```diff
@@ -74,7 +74,10 @@ func metricsCleanup() {
 		serviceInvocationResponseRecvName)
 }
 
-var testLogger = logger.NewLogger("proxy-test")
+var (
+	testLogger = logger.NewLogger("proxy-test")
+	_          = diag.SetLatencyDistribution(nil)
```
JoshVanL, line 79: Is there a strong reason this should be a global variable rather than done on a per test basis?
Global state in tests is very burdensome/error prone long term, especially since `diag` itself is globally stateful (which we should address eventually).

pkg/apis/configuration/v1alpha1/types.go
```diff
@@ -215,6 +215,9 @@ type MetricSpec struct {
 	HTTP *MetricHTTP `json:"http,omitempty"`
 	// +optional
 	Rules []MetricsRule `json:"rules,omitempty"`
+	// Latency distribution buckets. If not set, the default buckets are used.
+	// +optional
+	LatencyDistributionBuckets []int `json:"latencyDistributionBuckets,omitempty"`
```
JoshVanL, line 220: q: is there a semantic meaning between "unset" and "empty set", i.e. should this be a pointer to slice to differentiate `nil` pointer and empty slice?
I ask because it has been labeled as optional.

artursouza (dismissed):
This is an important change to help reduce cardinality problem for users. We have far too many buckets by default.

pkg/apis/configuration/v1alpha1/types.go
JoshVanL, line 220: Please can we add a more detailed specific comment about what this tunes? I can probably _guess_ what this does, but it would be good to include what latency in this context means, what metric types this effects (with some examples), and some example values a user would typically include.

Without reading further, latency of daprd to app is going to be vastly different to daprd to an external flaky HTTP IoT device so a blanket bucket set could be confusing.

pkg/config/configuration.go
```diff
@@ -253,9 +254,11 @@ func (o OtelSpec) GetIsSecure() bool {
 // MetricSpec configuration for metrics.
 type MetricSpec struct {
 	// Defaults to true
-	Enabled *bool         `json:"enabled,omitempty" yaml:"enabled,omitempty"`
-	HTTP    *MetricHTTP   `json:"http,omitempty" yaml:"http,omitempty"`
-	Rules   []MetricsRule `json:"rules,omitempty" yaml:"rules,omitempty"`
+	Enabled *bool       `json:"enabled,omitempty" yaml:"enabled,omitempty"`
+	HTTP    *MetricHTTP `json:"http,omitempty" yaml:"http,omitempty"`
+	// Latency distribution buckets. If not set, the default buckets are used.
+	LatencyDistributionBuckets *[]int        `json:"latencyDistributionBuckets,omitempty" yaml:"latencyDistributionBuckets,omitempty"`
```
JoshVanL, line 260: We use a pointer here 🙂

pkg/config/configuration.go
```diff
@@ -275,6 +278,24 @@ func (m MetricSpec) GetHTTPIncreasedCardinality(log logger.Logger) bool {
 	return *m.HTTP.IncreasedCardinality
 }
 
+// GetLatencyDistribution returns a *view.Aggregration to be used for latency histograms
+func (m MetricSpec) GetLatencyDistribution(log logger.Logger) *view.Aggregation {
+	defaultLatencyDistribution := []float64{1, 2, 3, 4, 5, 6, 8, 10, 13, 16, 20, 25, 30, 40, 50, 65, 80, 100, 130, 160, 200, 250, 300, 400, 500, 650, 800, 1_000, 2_000, 5_000, 10_000, 20_000, 50_000, 100_000}
```
JoshVanL, line 283: q: is this particular step a common standard?

JoshVanL (changes requested):

pkg/config/configuration.go
filintod: this is the current default, not really a standard, it is too many

filintod: https://github.com/dapr/dapr/pull/7843#discussion_r1651993783

pkg/apis/configuration/v1alpha1/types.go
filintod: sure, this is per config, so people can adjust that to the respective app. will provide some more context

pkg/api/grpc/proxy/handler_test.go
filintod: don't think so, let me revisit

pkg/apis/configuration/v1alpha1/types.go
filintod: should we call this `CustomLatencyDistributionBuckets` to differentiate from default?

JoshVanL: Tbh, I actually prefer LatencyDistributionBuckets. Although the user doesn't necessarily see it, the field itself (at least internally) gets set to the default value when unset, rather than acting as some kind of override. A good visualisation would be the API docs reference table whereby one of the columns is the default. If it was an override kind of field then the default would be nil rather than the actual end result value.
A lot of Dapr Configuration APIs would end up with Custom prefixes under the logic they customise a behaviour imo, which I don't think we want.

tests/integration/framework/process/daprd/daprd.go
```diff
@@ -327,17 +328,41 @@ func (d *Daprd) ProfilePort() int {
 	return d.profilePort
 }
 
+type MetricBucket struct {
+	UpperBound      float64
+	CumulativeCount uint64
+}
```
JoshVanL, line 334:
```suggestion
```

tests/integration/framework/process/daprd/daprd.go
```diff
@@ -327,17 +328,41 @@ func (d *Daprd) ProfilePort() int {
...
+
+type metricsOptions struct {
+	includeBuckets bool
+}
```
JoshVanL, line 338: Please move `includeBuckets` as a field to `options` in `options.go` with an appropriate name.

tests/integration/framework/process/daprd/daprd.go
```diff
@@ -327,17 +328,41 @@ func (d *Daprd) ProfilePort() int {
...
+
+func WithIncludeBuckets() func(metricsOptions *metricsOptions) {
+	return func(m *metricsOptions) {
+		m.includeBuckets = true
+	}
+}
```
JoshVanL, line 344: Move to `options.go`.

tests/integration/framework/process/daprd/daprd.go
```diff
@@ -327,17 +328,41 @@ func (d *Daprd) ProfilePort() int {
...
+
 // Returns a subset of metrics scraped from the metrics endpoint
-func (d *Daprd) Metrics(t *testing.T, ctx context.Context) map[string]float64 {
+func (d *Daprd) Metrics(t *testing.T, ctx context.Context, opts ...func(*metricsOptions)) map[string]float64 {
 	t.Helper()
 
+	curOpts := metricsOptions{}
+	for _, o := range opts {
+		o(&curOpts)
+	}
```
JoshVanL, line 353: Is there a reason we should ever _not_ include the bucket metrics to the returned test func? Is this option really needed?

tests/integration/framework/process/daprd/daprd.go
```diff
@@ -327,17 +328,41 @@ func (d *Daprd) ProfilePort() int {
...
+
 	req, err := http.NewRequestWithContext(ctx, http.MethodGet, fmt.Sprintf("http://%s/metrics", d.MetricsAddress()), nil)
 	require.NoError(t, err)
 
 	resp, err := d.httpClient.Do(req)
 	require.NoError(t, err)
 	require.Equal(t, http.StatusOK, resp.StatusCode)
 
+	// read resp.Body to variable
+	body, _ := io.ReadAll(resp.Body)
```
JoshVanL, line 363: Always check error.
```suggestion
	body, err := io.ReadAll(resp.Body)
	require.NoError(t, err)
```

tests/integration/framework/process/daprd/daprd.go
```diff
@@ -327,17 +328,41 @@ func (d *Daprd) ProfilePort() int {
...
+	resp.Body = io.NopCloser(bytes.NewReader(body))
```
JoshVanL, line 364: Why are we doing this? Seems superfluous :raised_eyebrow:

tests/integration/suite/daprd/metrics/http/cardinality/high.go
```diff
@@ -63,7 +63,7 @@ func (h *high) Setup(t *testing.T) []framework.Option {
 apiVersion: dapr.io/v1alpha1
 kind: Configuration
 metadata:
-  name: lowcardinality
+  name: highcardinality
```
JoshVanL, line 66: :+1:

tests/integration/suite/daprd/metrics/http/distributionbuckets/util.go
```diff
@@ -0,0 +1,22 @@
+package distributionbuckets
```
JoshVanL, line 1: Add copyright headers.

tests/integration/suite/daprd/metrics/http/distributionbuckets/util.go
```diff
@@ -0,0 +1,22 @@
...
+
+import (
+	"fmt"
+	"strconv"
+	"strings"
+)
+
+func getBucketFromKey(k string) (float64, error) {
```
JoshVanL, line 9: Consider passing in a `testing.T` so we can `require` inside the func, cleaning up the code a bit. 

```suggestion
func getBucketFromKey(t *testing.T, k string) float64 {
```

JoshVanL (changes requested):

tests/integration/framework/process/daprd/daprd.go
filintod: not a big reason. the function description said a subset and why I added this as an optional one. 

I guess not including will save some memory and a bit of speed, even though for tests that might not be the most important thing.

I can have it always return it.

filintod: @JoshVanL let me know your recommendation to change one way or another

tests/integration/framework/process/daprd/daprd.go
filintod: yeah, that was me debugging, let me remove

tests/integration/framework/process/daprd/daprd.go
JoshVanL: Let's always include all metrics in the output. Int tests are bound by `time.Sleep`s in Dapr runtime and test `eventually`s so we're not saving anything excluding them.

filintod: done

tests/integration/framework/process/daprd/daprd.go
filintod: removed

pkg/config/configuration.go
```diff
@@ -275,6 +278,24 @@ func (m MetricSpec) GetHTTPIncreasedCardinality(log logger.Logger) bool {
...
+	log.Infof("metric spec: %v", m)
+	if m.LatencyDistributionBuckets == nil || len(*m.LatencyDistributionBuckets) == 0 {
```
JoshVanL, line 285: This means that a user can't disable adding latency buckets. That's fine, but want to be clear that's what we want to do.

pkg/diagnostics/component_monitoring_test.go
```diff
@@ -5,6 +5,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/dapr/dapr/pkg/config"
```
JoshVanL, line 8: Nit: move dapr imports as 3rd block after 3rd party imports.

JoshVanL (changes requested):
Couple last things from me.

pkg/config/configuration.go
filintod: correct, it is just strange to have a histogram without buckets.  Maybe we could decide on a more sane default that the current very specific and long one.

JoshVanL (approved):

artursouza (approved):

PR #7873, opened 2024-07-01
default scheduler addr for sidecar automatically

cicoyle:
default scheduler addr for sidecar automatically

### DIFF
--- a/cmd/daprd/options/options.go
+++ b/cmd/daprd/options/options.go
@@ -168,7 +168,7 @@ func New(origArgs []string) (*Options, error) {
 	// --placement-host-address is a legacy (but not deprecated) flag that is translated to the actors-service flag
 	var placementServiceHostAddr string
 	fs.StringVar(&placementServiceHostAddr, "placement-host-address", "", "Addresses for Dapr Actor Placement servers (overrides actors-service)")
-	fs.StringSliceVar(&opts.SchedulerAddress, "scheduler-host-address", nil, "Addresses of the Scheduler service instance(s), as comma separated host:port pairs")
+	fs.StringSliceVar(&opts.SchedulerAddress, "scheduler-host-address", []string{"127.0.0.1:50006"}, "Addresses of the Scheduler service instance(s), as comma separated host:port pairs")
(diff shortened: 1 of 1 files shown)


yaron2 (approved):

cmd/daprd/options/options.go
```diff
@@ -168,7 +168,7 @@ func New(origArgs []string) (*Options, error) {
 	// --placement-host-address is a legacy (but not deprecated) flag that is translated to the actors-service flag
 	var placementServiceHostAddr string
 	fs.StringVar(&placementServiceHostAddr, "placement-host-address", "", "Addresses for Dapr Actor Placement servers (overrides actors-service)")
-	fs.StringSliceVar(&opts.SchedulerAddress, "scheduler-host-address", nil, "Addresses of the Scheduler service instance(s), as comma separated host:port pairs")
+	fs.StringSliceVar(&opts.SchedulerAddress, "scheduler-host-address", []string{"127.0.0.1:50006"}, "Addresses of the Scheduler service instance(s), as comma separated host:port pairs")
```
JoshVanL, line 171: Does this not mean a bare daprd run without any other control plane service would error? It's currently always been the case that no flag was needed to be set to empty or nil on a bare daprd to run successfully. Strictly only --app-id iirc

JoshVanL (changes requested):

cmd/daprd/options/options.go
JoshVanL: Btw, running a bare daprd binary (or container) with no other services is something we definitely want to (continue to) support

yaron2: Yes, this code should not make Dapr exit. Currently running a bare `daprd` with this line is entering a retry loop on the scheduler connection which isn't ideal. This behavior should be modified

artursouza (changes requested):
daprd should run by default without requiring placement or scheduler.

cmd/daprd/options/options.go
artursouza: If the sidecar does not connect to the scheduler, it should not continue. It is unhealthy condition.

yaron2:
Closing as we need to change what happens when scheduler isn't reachable to prevent a process freeze. That should be done for 1.15

PR #7875, opened 2024-07-02
Don't use dns by default for placement service in k8s. Get placement addresses from env variable.

elena-kolevska:
# Description

This PR changes the default placement service address in k8s deployments from the headless service to the addresses of the pods.
Placement addresses can also be specified through an env variable.

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Unit tests passing


charts/dapr/charts/dapr_placement/templates/dapr_placement_statefulset.yaml
```diff
@@ -97,7 +97,7 @@ spec:
   {{- end }}
 {{- end }}
         ports:
-          - containerPort: {{ .Values.ports.apiPort }}
```
artursouza, line 100: Why is this needed?

artursouza (commented):
minor nit

charts/dapr/charts/dapr_placement/templates/dapr_placement_statefulset.yaml
elena-kolevska: This was a variable under the placement charts which defines the port of the placement service.
In order to be able to access it from the injector chart, I moved it to global.

JoshVanL:
I don't fully understand the need for this PR- the benefit of using the headless service is that endpoints are added/removed from the DNS response according to each Pods health. Sure this is behaviour we want to keep?

elena-kolevska:
Closing because the problem that this PR was trying to address was solved in https://github.com/dapr/dapr/pull/7919

PR #7876, opened 2024-07-02
dont spam users if the job can't be sent to dapr

cicoyle:
tweaking to not spam users if the job can't be sent back to daprd. log once


pkg/scheduler/server/internal/pool.go
```diff
@@ -26,7 +26,10 @@ import (
 	"github.com/dapr/kit/logger"
 )
 
-var log = logger.NewLogger("dapr.runtime.scheduler")
+var (
+	issueSendingJobToDapr = atomic.Bool{}
```
artursouza, line 30: Why global?

cicoyle: fair, let me change that

cicoyle: oh wait, we want it for the entire process lifetime to only log once

pkg/scheduler/server/internal/pool.go
```diff
@@ -182,7 +185,13 @@ func (p *Pool) getConn(meta *schedulerv1pb.JobMetadata) (*conn, error) {
 
 	nsPool, ok := p.nsPool[meta.GetNamespace()]
 	if !ok {
-		return nil, fmt.Errorf("no connections available for namespace: %s", meta.GetNamespace())
+		// no connections available for namespace
+		// TODO: add a dead-letter go routine and log once, but
+		// don't err and spam users if the job can't be sent back
+		if issueSendingJobToDapr.CompareAndSwap(false, true) {
+			log.Warn("No daprd connections available to send triggered job back to")
```
JoshVanL, line 192: This should always be logged at the debug level at least, and should probably include information as to why.

I don't see the need to log this once as a warning as it's entirely client/job specific.

pkg/scheduler/server/internal/pool.go
JoshVanL: I still don't think this should be a global variable.

JoshVanL (changes requested):

yaron2 (approved):

pkg/scheduler/server/internal/pool.go
```diff
@@ -182,7 +182,12 @@ func (p *Pool) getConn(meta *schedulerv1pb.JobMetadata) (*conn, error) {
 
 	nsPool, ok := p.nsPool[meta.GetNamespace()]
 	if !ok {
-		return nil, fmt.Errorf("no connections available for namespace: %s", meta.GetNamespace())
+		// no connections available for namespace
+		// TODO: add a dead-letter go routine and log once, but
+		// don't err and spam users if the job can't be sent back
+		log.Debugf("No daprd connections available to send triggered job back to for namespace: %s, %+v", meta.GetNamespace(), meta)
```
ItalyPaleAle, line 188: Couldn't `meta` contain PII that is now being logged?

```suggestion
		log.Debugf("No daprd connections available to send triggered job back to for namespace: %s", meta.GetNamespace())
```

JoshVanL (approved):

pkg/scheduler/server/api.go
```diff
@@ -179,7 +179,7 @@ func (s *Server) triggerJob(ctx context.Context, req *api.TriggerRequest) bool {
 	}); err != nil {
 		// TODO: add job to a queue or something to try later this should be
 		// another long running go routine that accepts this job on a channel
-		log.Errorf("Error sending job to connection stream: %s", err)
+		log.Debugf("Error sending job %s to connection stream: %s", req.GetName()[idx+2:], err)
```
mukundansundar, line 182: qq: If there are errors sending jobs to dapr, is it retried in place or is currently dropped in the alpha implementation?

cicoyle: Please see the [failure policy proposal PR lines for the answer here](https://github.com/dapr/proposals/pull/66/files#diff-eabc9eb37f9fcfc26ba06ecae4e6391a5a4ebda446a2bc8f85178f042609ef6bR14-R15) :)

cicoyle:
this isn't needed once we have the staging queue, so closing it

PR #7877, opened 2024-07-02
Makes raft server healthy earlier

elena-kolevska:
Trying to fix the e2e actor reminder tests by reporting the raft server healthy earlier.

When running on Kubernetes, we use the headless placement service as the "placement-address" for sidecars. The gRPC client library, configured to use round-robin, resolves the A records provided by the headless service. However, gRPC checks for pod liveness and caches the results. This behaviour can lead to issues if the leader's address is not cached, as the client will only return follower addresses until the cache is refreshed on the next DNS lookup. Consequently, most actor operations may fail during this period.

To fix this, we're moving the code line where the raft server reports its health status earlier. This adjustment should help the gRPC client library detect the leader's address more reliably, improving overall stability.


artursouza (dismissed):

elena-kolevska:
The problem this PR is trying to solve was solved in https://github.com/dapr/dapr/pull/7962 and https://github.com/dapr/dapr/pull/7919, so the change here is no longer necessary, although it's still a nice to have.

artursouza (approved):

PR #7891, opened 2024-07-03
Scheduler: expose securityContext.fsGroup in helm chart

JoshVanL:
Expose the `securityContext.fsGroup` field in the helm chart for scheduler, which is useful when mounting a custom volume.


cicoyle:
We do want this change for Scheduler, but the e2e tests are failing due to an `unknown field "fsGroup" in io.k8s.api.core.v1.SecurityContext`.

yaron2:
> We do want this change for Scheduler, but the e2e tests are failing due to an `unknown field "fsGroup" in io.k8s.api.core.v1.SecurityContext`.

Might be helpful: https://github.com/zammad/zammad-helm/issues/40#issuecomment-609715022.

yaron2 (approved):

PR #7895, opened 2024-07-05
Extend Actor Placement Heartbeat 1s -> ~5m~ 5s

JoshVanL:
Extends the actor placement host report heartbeat from 1s to 5m. Sending a host report every second from every sidecar to placement is very unnecessary, as it consumes resources and saturated the network.

We now only report host status to placement when first connecting, and when the app health changes.

Placement correctly shuts down the client stream when it is closing, as we can no longer rely on receiving a messagee to error as placement will no longer receive messages from every sidecar every second.

We still have a "heartbeat" status report every 5 minutes however it is not needed. Kept purely to reduce code churn late in release.

EDIT: to account for version skew tests passing, the value has been changed to 5s as v1.13 placement will not close server streams during shutdown. The heartbeat should be completely removed in v1.15.


artursouza (approved):

artursouza:
Thanks. Will merge once all validations pass.

antontroshin:
@JoshVanL `5m` seems like a big interval to report actor "liveness", some systems or dashboards using this metric might consider the actor type "dead" after such a long period of not reporting.
Is it possible to reduce this to `30s` or at least `1m`, to be somewhat consistent with the default Prometheus scrape intervals of 1m?

JoshVanL:
> @JoshVanL `5m` seems like a big interval to report actor "liveness", some systems or dashboards using this metric might consider the actor type "dead" after such a long period of not reporting.
> 
> Is it possible to reduce this to `30s` or at least `1m`, to be somewhat consistent with the default Prometheus scrape interval of `1m`?

Hey @antontroshin 🙂 
The 5m is intentionally made to be a "never" interval, and it would equally be valid to make this value max(int64) instead. Currently, daprds are reporting the actor types they implement _every second_. This is both superfluous, and more importantly, very resource wasteful on both the runtime and network. Daprds should only report actor types in the event of the app being healthy, unhealthy, daprd shutting down or, soon in future, (de)registering internal workflow actor types when a workflow stream is (in)active.
The presence of the gRPC connection stream to placement determines that the daprd is still healthy and thus implements the currently registered actor type. Normal operation 99.9% of the time should see this connection gracefully disconnect when either service exits. Ungracefull shutdown is handled by the underlying gRPC keep alive. These keep alive values might want to be changed.
The concept/codepath of the heartbeat has been kept in this PR to reduce LoC churn late in release. 1.15 should see the code removed entirely.

elena-kolevska:
The actor heartbeat is currently important for two things:

- Metric reporting: Actor liveness is a metric Dapr provides (`placement/actor_heartbeat_timestamp`), so it's important to announce any changes to it clearly and possibly update the metric so it only shows an active/inactive status

- Placement failover: In case of a placement service failover we rely on the actor liveness timestamps to know if a sidecar that had been connected to the previous leader hasn't connected to the newly elected one. If so - we trigger a dissemination. Otherwise, if all sidecars reconnect to the new leader in the required interval (`faultyHostDetectDuration`) we skip the dissemination, avoiding the locking it incurs. This is actually the primary need for replicating the raft data. If we choose to remove the heartbeat, and disseminate the tables from scratch on failover, we can simplify our raft implementation and only use the algorithm for leader election.

elena-kolevska (changes requested):
Currently, `statusReportHeartbeatInterval` is also used to send quick updates to the placement service when there is a change in the hosted actor types. This is especially important for internal (workflow) actors, since they get registered and unregistered dynamically. 
Admittedly, we don't need a heartbeat for this, we can trigger it when needed, but that needs to be added together with the heartbeat change.
I'm happy to help with this, if you want.

Issue #7897, opened 2024-07-05
Exception thrown during Job schedule API call


rochabr:
While running this [HTTP quickstart ](https://github.com/rochabr/quickstarts/tree/distr-scheduler/job/go/http)(not released, yet). Dapr throws an exception and the job is not scheduled.

## In what area(s)?

/area runtime

> /area operator

> /area placement

> /area docs

> /area test-and-release

## What version of Dapr?

1.14 rc.1
> 1.1.x
> 1.0.x
> edge: output of `git describe --dirty`

## Actual Behavior

The API returns an empty response and this exception is thrown in the sidecar:

```
2024/07/05 10:27:23 http: panic serving [::1]:63000: runtime error: invalid memory address or nil pointer dereference
goroutine 94 [running]:
net/http.(*conn).serve.func1()
        /Users/runner/hostedtoolcache/go/1.22.4/arm64/src/net/http/server.go:1898 +0xb0
panic({0x10a4275e0?, 0x10e3d16e0?})
        /Users/runner/hostedtoolcache/go/1.22.4/arm64/src/runtime/panic.go:770 +0x124
github.com/dapr/dapr/pkg/runtime/scheduler/clients.(*Clients).Next(...)
        /Users/runner/work/dapr/dapr/pkg/runtime/scheduler/clients/clients.go:65
github.com/dapr/dapr/pkg/api/universal.(*Universal).ScheduleJobAlpha1(0x1400179c480, {0x10b061120, 0x1400134c660}, 0x1400134c690)
        /Users/runner/work/dapr/dapr/pkg/api/universal/jobs.go:66 +0x464
github.com/dapr/dapr/pkg/api/http.UniversalHTTPHandler[...].func1(0x140016ba240)
        /Users/runner/work/dapr/dapr/pkg/api/http/universal.go:105 +0x10c
net/http.HandlerFunc.ServeHTTP(0x10b061120?, {0x115c4b108?, 0x14000789ae0?}, 0x1049ac7e0?)
        /Users/runner/hostedtoolcache/go/1.22.4/arm64/src/net/http/server.go:2166 +0x38
github.com/dapr/dapr/pkg/api/http.(*server).handle.(*server).unescapeRequestParametersHandler.func1({0x115c4b108, 0x14000789ae0}, 0x140016ba240)
        /Users/runner/work/dapr/dapr/pkg/api/http/server.go:400 +0x78
[...]

PR #7898, opened 2024-07-05
Fix panic when scheduler is not initialized and job API is used.

artursouza:
# Description

Fixes https://github.com/dapr/dapr/issues/7897

New integration test reproduces the bug gracefully. In this case, the check for `c == nil` was required. It means that the `.Next()` function is called even when the instance is nil.

## Issue reference

Please reference the issue this PR will close: https://github.com/dapr/dapr/issues/7897

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing


pkg/runtime/scheduler/clients/clients.go
```diff
@@ -62,6 +62,12 @@ func New(ctx context.Context, opts Options) (*Clients, error) {
 
 // Next returns the next client in a round-robin manner.
 func (c *Clients) Next() schedulerv1pb.SchedulerClient {
+	// This is called even when no scheduler is initialized.
+	// Keeping these nil checks here as a central location instead of spreading "ifs".
+	if c == nil || c.clients == nil || len(c.clients) == 0 {
+		return &schedulerClientNotInitializedInstance
```
JoshVanL, line 68: Does it make more sense to not return an error here instead? Or better, also pass a context whereby it waits for client to become available. We return the context error when it has been cancelled.

pkg/runtime/scheduler/clients/schedulerclientnotinit.go
```diff
@@ -0,0 +1,43 @@
+/*
```
JoshVanL, line 1: Rename file to `noinit` as prefix is superfluous through package import name.

JoshVanL (changes requested):

pkg/runtime/scheduler/clients/clients.go
artursouza: I am keeping this change minimal for the sake of this release.

pkg/runtime/scheduler/clients/schedulerclientnotinit.go
artursouza: Sure.

yaron2 (approved):

pkg/runtime/scheduler/clients/clients.go
artursouza: If you want to make another PR in master improving this, I am all in. For 1.14, I recommend us being minimalistic.

PR #7908, opened 2024-07-15
[TEST] Fix Actor Reminder Triggering Perf Test Regression

cicoyle:
Fix the [failing perf test](https://github.com/dapr/dapr/actions/runs/9670616762/job/26679809967) starts to pass. If it passes, that means we should up the test time since we will be triggering more snapshots in code which might take more time.

```
tests/perf/actor_reminder TestActorReminderTriggerPerformance
...
    actor_reminder_test.go:332: Created 5000 reminders in 2m28.613813864s (33.6qps)
    actor_reminder_test.go:334: 
        	Error Trace:	/home/runner/work/dapr/dapr/tests/perf/actor_reminder/actor_reminder_test.go:334
        	Error:      	"2m20s" is not greater than or equal to "2m28.613813864s"
        	Test:       	TestActorReminderTriggerPerformance
        	Messages:   	expected to create reminders 5000 in less than 140s
```

It looks like [this PR never got the dueTime just right, see the failing perf tests](https://github.com/dapr/dapr/pull/7885), so I'm upping the value a little more.


cicoyle:
oops, didn't mean to close

PR #7911, opened 2024-07-17
Do placement dns check

JoshVanL:
# Description

## Issue reference

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing


elena-kolevska:
I found this in the [grpc docs](https://github.com/grpc/grpc/blob/master/doc/environment_variables.md), it might be helpful:
```
GRPC_DNS_RESOLVER Declares which DNS resolver to use. The default is ares if gRPC is built with c-ares support. Otherwise, the value of this environment variable is ignored. Available DNS resolver include:

    - ares (default on most platforms except iOS, Android or Node)- a DNS resolver based around the c-ares library
    - native - a DNS resolver based around getaddrinfo(), creates a new thread to perform name resolution

NetBIOS and DNS: If your network relies on NetBIOS name resolution or a mixture of DNS and NetBIOS name resolution (e.g. in some Windows networks) then you should use the 'native' DNS resolver or make sure all NetBIOS names are also configured in DNS. The 'ares' DNS resolver only supports DNS name resolution.
```

ItalyPaleAle:
@JoshVanL what's the reason for this PR?

Normally we don't want to resolve the names in Dapr, and let K8s do that, since DNS names can change if the pods are restarted?

JoshVanL:
@ItalyPaleAle we’re trying to investigate why the placement client DNS round robin is misbehaving in Windows clusters. Just a dummy draft PR 🙂

PR #7915, opened 2024-07-18
Update file mounts to be consistent with FHS

JoshVanL:
Changes scheduler data dir in Helm chart to be under `/var/lib/dapr/scheduler` as this is strictly more correct under FHS.

Updates paths `/var/run/..` to `/run/...`/

Updates placement raft log path from `/var/run/..` to `/var/lib/...`.

https://refspecs.linuxfoundation.org/FHS_3.0/fhs-3.0.html#varlibVariableStateInformation


cicoyle:
This change looks fine, but there are e2e failures with scheduler, looks to be from a prior change tho bc I see: `Failed to initialize scheduler clients: scheduler client not initialized for address dapr-scheduler-server-0.dapr-scheduler-server.dapr-tests.svc.bq0sc14cqklujfo00hvyg2cvkh.bx.internal.cloudapp.net:50006: context deadline exceeded:`

cicoyle (dismissed):

artursouza (changes requested):
What is the motivation for this change? It would make it inconsistent with placement's raft logs location.

JoshVanL:
> What is the motivation for this change? It would make it inconsistent with placement's raft logs location.

`/var/lib` is the more appropriate location to store database data under FHS.

artursouza:
> Changes scheduler data dir in Helm chart to be under `/var/lib/dapr/scheduler` as this is strictly more correct under FHS.
> 
> https://refspecs.linuxfoundation.org/FHS_3.0/fhs-3.0.html#varlibVariableStateInformation

What is the motivato

artursouza:
> > What is the motivation for this change? It would make it inconsistent with placement's raft logs location.
> 
> `/var/lib` is the more appropriate location to store database data under FHS.

Sure. I don't have any strong preference towards complying with FHS or not. For me, consistency of using Dapr is more important. I am OK with this change if we make it consistent for other volumes too, like in Placement.

JoshVanL:
@artursouza makes sense! I went ahead and made all paths valid under _modern_ FHS as I understand it. Let me know what you think.

PR #7919, opened 2024-07-19
Fixing gRPC round robin for placement service

elena-kolevska:
Related to actors e2e tests failing

References:
- https://github.com/grpc/grpc/blob/master/doc/environment_variables.md
- https://github.com/grpc/grpc/blob/master/doc/load-balancing.md


artursouza (commented):
No concerns from my side. Waiting on Cassie's approval to merge.

pkg/actors/placement/client.go
```diff
@@ -104,20 +117,24 @@ var noop = func() {}
...
 	c.streamConnectedCond.L.Lock()
 	if !c.streamConnAlive {
 		c.streamConnectedCond.Broadcast()
 		c.streamConnectedCond.L.Unlock()
 		return
 	}
 
-	c.drain(c.clientStream, c.clientConn)
+	if closeConnection {
+		c.drain(c.clientStream, c.clientConn)
```
cicoyle, line 134: I think we can get rid of the bool here and conditional, and do the nil check in the drain func like you already have

cicoyle: Elena mentioned this is intentional to reuse the same client conn for grpc load balancing. this is fine then 👍🏻

cicoyle (approved):
LGTM assuming tests are passing consistently now :)

artursouza (approved):

PR #7925, opened 2024-07-21
Helm: Fix Service annotation indentation

JoshVanL:
Fixes annotation indentation for Service manifests.

Adds integration tests to cover combinations of custom annotations and
prometheus enabled/disabled.

Adds int test framework `tee.buffer` to implement a buffered tee for
multiple writers. This allows for doing io reads on a io.Reader returned
from running processes, called at any time.

Fix placement failover int test.

/cc @sicoyle

### DIFF
--- a/charts/dapr/charts/dapr_operator/templates/dapr_operator_service.yaml
+++ b/charts/dapr/charts/dapr_operator/templates/dapr_operator_service.yaml
@@ -15,9 +15,7 @@ metadata:
     prometheus.io/path: "/"
     {{- end }}
   {{- if .Values.apiService.annotations }}
-    {{- range $key, $value := .Values.apiService.annotations }}
-    {{ $key }}: "{{ $value }}"
-    {{- end }}
+  {{- .Values.apiService.annotations | toYaml | nindent 4 }}
   {{- end }}
   {{- end }}
 spec:
--- a/charts/dapr/charts/dapr_operator/templates/dapr_operator_service.yaml
+++ b/charts/dapr/charts/dapr_operator/templates/dapr_operator_service.yaml
@@ -16,7 +16,7 @@ metadata:
     {{- end }}
   {{- if .Values.apiService.annotations }}
     {{- range $key, $value := .Values.apiService.annotations }}
-      {{ $key }}: "{{ $value }}"
+    {{ $key }}: "{{ $value }}"
     {{- end }}
   {{- end }}
   {{- end }}
@@ -49,6 +49,7 @@ metadata:
     {{ $key }}: {{ tpl $value $ }}
     {{- end }}
 {{- if .Values.webhookService.annotations }}
+  annotations:
 {{ toYaml .Values.webhookService.annotations | indent 4}}
 {{- end }}
 spec:
--- a/charts/dapr/charts/dapr_scheduler/templates/dapr_scheduler_service.yaml
+++ b/charts/dapr/charts/dapr_scheduler/templates/dapr_scheduler_service.yaml
@@ -18,7 +18,7 @@ metadata:
     {{- end }}
   {{- if .Values.service.annotations }}
     {{- range $key, $value := .Values.service.annotations }}
-      {{ $key }}: "{{ $value }}"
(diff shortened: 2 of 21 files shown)


charts/dapr/charts/dapr_operator/templates/dapr_operator_service.yaml
```diff
@@ -49,6 +49,7 @@ metadata:
     {{ $key }}: {{ tpl $value $ }}
     {{- end }}
 {{- if .Values.webhookService.annotations }}
+  annotations:
```
cicoyle, line 52: I think this is still under the annotations from above on line 11
```suggestion
```

JoshVanL: Sorry. Not sure what you mean

cicoyle: I mean this `annotations` line you added can be removed since its already on line 11 above I believe, so the webhook service annotations are already under that field so we don't need to add `annotations` here. Hopefully that makes sense?

JoshVanL: Sorry, still not sure. Line 11 is a different resource, we want annotations on both, no?

sicoyle: I think this is fine. I do think line 53 on the `toYaml` needs a dash.
```
{{ - toYaml .Values.webhookService.annotations | indent 4}}
```

Pretty sure Ale commented on this to be fixed on all of these.

sicoyle: also, if we have the toYaml line, then we don't need the range loop like the service above has?

tests/integration/suite/placement/ha/failover.go
```diff
@@ -72,11 +72,15 @@ func (n *failover) Setup(t *testing.T) []framework.Option {
 	n.srv = prochttp.New(t, prochttp.WithHandler(handler))
 
 	return []framework.Option{
-		framework.WithProcesses(n.fp, n.srv),
+		framework.WithProcesses(n.fp, n.srv, n.placements[0], n.placements[1], n.placements[2]),
```
sicoyle, line 75: can we always guarantee these indices will be present and not result in an out of range issue?

charts/dapr/charts/dapr_operator/templates/dapr_operator_service.yaml
cicoyle: oh, I misread it being a diff resource. yeah I agree now

charts/dapr/charts/dapr_scheduler/templates/dapr_scheduler_service.yaml
```diff
@@ -18,7 +18,7 @@ metadata:
     {{- end }}
   {{- if .Values.service.annotations }}
     {{- range $key, $value := .Values.service.annotations }}
-      {{ $key }}: "{{ $value }}"
+    {{ $key }}: "{{ $value }}"
```
ItalyPaleAle, line 21: Why not using `{{- .Values.service.annotations | toYaml | nindent 4 }}` like in the other palces? Similar comment in other places

tests/integration/suite/placement/ha/failover.go
JoshVanL: Yes, this is defined here https://github.com/dapr/dapr/pull/7925/files/d74972b73e88cd0f8d4c5e15870b8e462decd62a#diff-d95aa018c0eb8cb641069009547a188d45010c2a2c3fcbea585f9e390a029773R54

charts/dapr/charts/dapr_scheduler/templates/dapr_scheduler_service.yaml
JoshVanL: Thanks, I'll update with that :slightly_smiling_face:

PR #7935, opened 2024-07-23
Fix crud integration scheduler test

JoshVanL:
Fix crud integration scheduler test

Fix integration CRUD test. Splits test into 2 for faster runtime.


tests/integration/suite/actors/reminders/scheduler/upsert.go
```diff
@@ -0,0 +1,128 @@
...
+
+	_, err = gclient.RegisterActorReminder(ctx, &rtv1.RegisterActorReminderRequest{
+		ActorType: "myactortype",
+		ActorId:   "myactorid",
+		Name:      "xyz",
+		DueTime:   "1s",
+	})
+	require.NoError(t, err)
+
+	assert.Eventually(t, func() bool {
```
elena-kolevska, line 109: I think this check has the potential to be flaky under some conditions, but given the reminder has a whole second until it triggers, it should be ok

elena-kolevska (approved):

yaron2 (approved):

PR #7954, opened 2024-07-29
Fixes reminder deletion return

elena-kolevska:


cicoyle (approved):
This doesn't per say fix a race condition, but rather fixes the return from muddying output. Change lgtm

yaron2:
Linter

pkg/actors/actors.go
```diff
@@ -1153,19 +1153,38 @@ func (a *actorsRuntime) ExecuteLocalOrRemoteActorReminder(ctx context.Context, r
...
+
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case errC := <-errCh:
+			if errC != nil {
+				fmt.Println("--- Error deleting reminder", errC)
+				return errC
+			} else {
+				fmt.Println("--- Deleted reminder", errC)
```
mikeee, line 1184: I don't think the nil-error `errC` should be printed here.
Are these prints going to be updated to a errorf/debugf?

yaron2: Likely here for debug purposes only

elena-kolevska:
Closing this, as @cicoyle is topuching on it as a part of a bigger PR

PR #7957, opened 2024-07-30
Cleanup Reminders, Dont spam scheduler logs with failure watching jobs

cicoyle:
Call to delete the reminder when the reminder is canceled. Also dont spam users with `failureWatchingJobs` so I added an atomic bool there so it should log once during the process lifetime

### DIFF
--- a/pkg/actors/actors.go
+++ b/pkg/actors/actors.go
@@ -28,6 +28,7 @@ import (
 
 	"github.com/alphadose/haxmap"
 	"github.com/cenkalti/backoff/v4"
+	"github.com/dapr/dapr/pkg/messages"
 	"github.com/google/uuid"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
@@ -546,6 +547,27 @@ func (a *actorsRuntime) Call(ctx context.Context, req *internalv1pb.InternalInvo
 	}
 
 	if err != nil {
+		if errors.Is(err, ErrReminderCanceled) {
+			go func() {
+				select {
+				case <-ctx.Done():
+					log.Debugf("Context canceled, stopping reminder deletion: %s", req.Actor.GetActorKey())
+					return
+				default:
+					reqCtx := context.Background()
+					derr := a.DeleteReminder(reqCtx, &DeleteReminderRequest{
+						Name:      req.Actor.GetActorKey(),
+						ActorType: req.Actor.GetActorType(),
+						ActorID:   req.Actor.GetActorId(),
+					})
+					if derr != nil {
+						log.Errorf("Failed to delete reminder. %s", derr.Error())
+					}
+					return
+				}
+			}()
+		}
+
 		if res != nil && actorerrors.Is(err) {
 			return res, err
 		}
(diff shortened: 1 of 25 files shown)


yaron2:
Linter

pkg/actors/actors.go
```diff
@@ -546,6 +547,27 @@ func (a *actorsRuntime) Call(ctx context.Context, req *internalv1pb.InternalInvo
 	}
 
 	if err != nil {
+		if errors.Is(err, ErrReminderCanceled) {
```
cgillum, line 550: Reminder cancellation is not intended to be valid in a normal actor call context. It seems a little strange that we'd include this logic here.

pkg/actors/actors.go
```diff
@@ -1055,12 +1077,32 @@ func (a *actorsRuntime) executeTimer(reminder *internal.Reminder) bool {
...
 			log.Debug("Reminder " + reminder.ActorKey() + " was canceled by the actor")
+
+			go func() {
+				select {
+				case <-ctx.Done():
+					log.Debugf("Context canceled, stopping reminder deletion: %s", reminder.Key())
+					return
+				default:
+					reqCtx := context.Background()
+					derr := a.DeleteReminder(reqCtx, &DeleteReminderRequest{
```
cgillum, line 1095: The whole point of reminder cancellation was to have an optimized mechanism for deleting these reminders at the end of their execution. But if we're going to just send a delete command in response to ErrReminderCanceled anyways, should we instead just update the implementations of workflow_actor.go and activity_actor.go to make this API call directly, and remove the reminder cancellation concept?

pkg/actors/actors.go
```diff
@@ -1087,7 +1129,28 @@ func (a *actorsRuntime) doExecuteReminderOrTimerOnInternalActor(ctx context.Cont
 
 		err = internalAct.InvokeTimer(ctx, reminder, md)
 		if err != nil {
-			if !errors.Is(err, ErrReminderCanceled) {
+			if errors.Is(err, ErrReminderCanceled) {
```
cgillum, line 1132: This code path is for timers, but we don't have timer cancellation, only reminder cancellation.

pkg/api/grpc/daprinternal.go
```diff
@@ -312,6 +312,25 @@ func (a *api) CallActorReminder(ctx context.Context, in *internalv1pb.Reminder)
 		Period:    in.GetPeriod(),
 		TTL:       in.GetExpirationTime().String(),
 	})
+	if errors.Is(err, actors.ErrReminderCanceled) {
```
cgillum, line 315: I haven't looked at the code flow closely, but shouldn't this error be handled internally in the actor runtime instead of here?

pkg/actors/actors.go
cicoyle: I think that would be cleaner to remove the cancel concept. Its a bit confusing without the context and especially since I don't think we were properly cleaning up when they were set to cancel.

pkg/actors/actors.go
cicoyle: This function is called from the `workflow_actor.go` file. This func makes a call to `a.isActorLocal` which then does a check if its an internal actor. I thought the reminder cancel error was relevant to internal actors. I can add to the conditional to check if its an internal actor only then go into this logic. Or are you saying this isn't related to reminders in general?

pkg/api/grpc/daprinternal.go
cicoyle: oh yes, I added it to the `ExecuteLocalOrRemoteActorReminder` so this check can be removed.

pkg/actors/actors.go
cicoyle: I'll rm this. I think you are saying this is not needed here

pkg/actors/actors.go
```diff
@@ -1234,7 +1231,18 @@ func (a *actorsRuntime) doExecuteReminderOrTimer(ctx context.Context, reminder *
 		return a.callLocalActor(ctx, req)
 	})
 	if err != nil {
-		if !errors.Is(err, ErrReminderCanceled) {
+		if errors.Is(err, ErrReminderCanceled) {
```
artursouza, line 1234: This is outside the resiliency policy now.

pkg/runtime/wfengine/backends/actors/activity_actor.go
```diff
@@ -364,6 +393,13 @@ func (a *activityActor) createReliableReminder(ctx context.Context, data any) er
 	if err != nil {
 		return fmt.Errorf("failed to encode data as JSON: %w", err)
 	}
+
+	if a.actorRuntime.UsingSchedulerReminders() == true {
```
artursouza, line 397: This is an abstraction violation. The WF actors should work regardless of the scheduler service. Does it mean that actors that implemented in the app will have the problem with reminders not being deleted too if they also don't have this abstraction violation?

cicoyle:
Closing for now, will pull out what I need

PR #7962, opened 2024-08-02
Placement leadership fix

elena-kolevska:
Fixes a situation where the placement cluster would be stuck in a leaderless state, even though a raft leader has been elected.
The placement server gets in this state when a raft node has been elected leader, then follower, and then leader again. Raft re-election like this can be triggered due to:
- faulty network
- lost majority (two non-leader pods going down at around the same time)

### DIFF
--- a/pkg/placement/raft/server.go
+++ b/pkg/placement/raft/server.go
@@ -390,3 +390,11 @@ func (s *Server) ApplyCommand(cmdType CommandType, data DaprHostMember) (bool, e
 	resp := future.Response()
 	return resp.(bool), nil
 }
+
+func (s *Server) GetID() string {
+	return s.id
+}
+
+func (s *Server) GetRaftBind() string {
+	return s.raftBind
+}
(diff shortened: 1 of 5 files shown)


pkg/placement/raft/server.go
```diff
@@ -390,3 +390,11 @@ func (s *Server) ApplyCommand(cmdType CommandType, data DaprHostMember) (bool, e
 	resp := future.Response()
 	return resp.(bool), nil
 }
+
+func (s *Server) GetID() string {
```
artursouza, line 394: Although this is used just for tests. I am OK with it.

pkg/placement/raft/server.go
```diff
@@ -390,3 +390,11 @@ func (s *Server) ApplyCommand(cmdType CommandType, data DaprHostMember) (bool, e
...
+	return s.id
+}
+
+func (s *Server) GetRaftBind() string {
```
artursouza, line 398: Same.

artursouza (approved):
Pending test runs completing. Flakiness seem unrelated to this change.

PR #7967, opened 2024-08-05
Revert docker images to static:nonroot images

yaron2:
Accepted suggestion by @ItalyPaleAle to make `static:nonroot` the default images again due to potential future CVEs associated with libc. We will look into adding additional variants in the future.


ItalyPaleAle (approved):
Thanks!

PR #7968, opened 2024-08-05
Fix scheduler reminders for workflows

cicoyle:
For now setting repeats to 2 which will trigger once, then again but its a no op. Setting repeats to 1 currently causes issues that need investigation. updating cron library to latest with the fix for cleaning up the counters upon deleting a job


artursouza (dismissed):

artursouza:
Pending tests to pass.

artursouza (approved):

artursouza (dismissed):

cicoyle:
It looks like dapr is not gracefully shutting down in all cases. That issue is unrelated to this PR, as I can see this run had an issue shutting daprd down gracefully as well. I also removed all my changes and got the same result of the test failure.
[see CI failure from this run](https://github.com/dapr/dapr/actions/runs/10288408130/job/28474160735)

yaron2 (approved):

PR #7980, opened 2024-08-12
Remove 70s timeout for integration test to default 45s

JoshVanL:
Most integration tests should take between 1-2s max, with some outliers being closer to 10s. 45s is a safe max to allow for slow CI runners. 70s is too large and not appropriate for integration tests- tests which take this long should be moved to e2e.


artursouza:
I am happy to take this PR in master AFTER integration tests are stable.

yaron2:
> I am happy to take this PR in master AFTER integration tests are stable.

Integration tests looks stable

yaron2 (approved):

PR #7982, opened 2024-08-12
Fix some flaky int tests

cicoyle:
Trying to fix some flakiness with:

- [reminders/scheduler/unregister](https://github.com/dapr/dapr/actions/runs/10353200043/job/28656085181)
- binding appready - cant find the build link again, but the binding call was starting as 1, so Im starting it explicitly as 0 for the test at the start
- [pubsub/grpc/appready](https://github.com/dapr/dapr/actions/runs/10355699565/job/28664197983)
- [slow unit test for windows: healthtest for pkg/actors/health](https://github.com/dapr/dapr/actions/runs/10393654023/job/28782088664?pr=7982)


tests/integration/suite/actors/reminders/scheduler/unregister.go
```diff
@@ -124,7 +124,7 @@ func (u *unregister) Run(t *testing.T, ctx context.Context) {
 		prevLast := last
 		last = called
 		assert.Equal(ct, int(prevLast), int(called))
-	}, time.Second*15, time.Second*2)
+	}, time.Second*15, time.Second*1)
```
JoshVanL, line 127: We must wait 2 seconds. The reminder fires every second so we are now racing

tests/integration/suite/daprd/binding/input/appready.go
```diff
@@ -99,6 +99,7 @@ spec:
 func (a *appready) Run(t *testing.T, ctx context.Context) {
 	a.daprd.WaitUntilRunning(t, ctx)
 
+	a.bindingCalled.Store(0)
```
JoshVanL, line 102: This is superfluous as its zeroth value is 0

tests/integration/suite/daprd/binding/input/appready.go
```diff
@@ -108,21 +109,19 @@ func (a *appready) Run(t *testing.T, ctx context.Context) {
...
 	require.Eventually(t, func() bool { return a.healthCalled.Load() > called }, time.Second*5, time.Millisecond*10)
-
 	assert.Eventually(t, func() bool {
 		resp, err := httpClient.Do(req)
 		require.NoError(t, err)
 		defer resp.Body.Close()
 		return resp.StatusCode == http.StatusInternalServerError
 	}, time.Second*5, 10*time.Millisecond)
 
-	time.Sleep(time.Second * 2)
```
JoshVanL, line 125: This is important to ensure that value doesn't change.

tests/integration/suite/daprd/pubsub/grpc/appready.go
```diff
@@ -143,12 +143,14 @@ func (a *appready) Run(t *testing.T, ctx context.Context) {
 		return resp.StatusCode == http.StatusOK
 	}, time.Second*5, 10*time.Millisecond)
 
-	_, err = client.PublishEvent(ctx, &rtv1.PublishEventRequest{
-		PubsubName: "mypubsub",
-		Topic:      "mytopic",
-		Data:       []byte(`{"status": "completed"}`),
-	})
-	require.NoError(t, err)
```
JoshVanL, line 151: This should never fail- what error are we seeing?

tests/integration/suite/daprd/binding/input/appready.go
cicoyle: the test was erring due to that being 1 before it gets called later on in the test

tests/integration/suite/daprd/binding/input/appready.go
cicoyle: I can add it back no problem

tests/integration/suite/daprd/pubsub/grpc/appready.go
cicoyle: I forgot to grab the build failure link, but was seeing an err here

cicoyle: https://github.com/dapr/dapr/actions/runs/10355704565/job/28664159218?pr=7973

cicoyle: actually, maybe that wasn't the exact same err, just the same test I happened to see in an open PR currently

tests/integration/suite/daprd/binding/input/appready.go
JoshVanL: It should never be 1 because the app isn't healthy yet so the binding should not be called. If it is, there is a bug.

tests/integration/suite/daprd/pubsub/grpc/appready.go
JoshVanL: The link you shared doesn't have anything to do with this block.

cicoyle:
I can't repro that unit test failure ^ This PR addresses all others I've seen thus far

tests/integration/suite/daprd/binding/input/appready.go
cicoyle: maybe this is a flaky bug then. I can remove this change, but will keep it on the radar

artursouza (approved):

artursouza (approved):

JoshVanL (approved):

PR #7983, opened 2024-08-12
Fix Duplicate workflow reminder execution in Scheduler mode

JoshVanL:
Workflow reminders must have a period set of repeating forever. This is what fundamentally ensures the durability of Workflows.

PR updates all workflow reminders to have an infinite period.

Workflow reminders repeat execution every 1 minute. In some cases, such as during extremely high load, a workflow reminder may execute whilst a previous execution is still in progress. This leads to situations such as double activity invocation. This patch adds a duplicate internal reminder detector, which silently returns if a duplicate reminder is detected.

Removes the background contexts to Scheduler APIs. The Scheduler API server must respect inbound client contexts.

Removes unused `--placement` flag from the Scheduler.

### DIFF
--- a/cmd/scheduler/options/options.go
+++ b/cmd/scheduler/options/options.go
@@ -73,7 +73,6 @@ func New(origArgs []string) (*Options, error) {
 	fs.StringVar(&opts.TrustDomain, "trust-domain", "localhost", "Trust domain for the Dapr control plane")
 	fs.StringVar(&opts.taFile, "trust-anchors-file", securityConsts.ControlPlaneDefaultTrustAnchorsPath, "Filepath to the trust anchors for the Dapr control plane")
 	fs.StringVar(&opts.SentryAddress, "sentry-address", fmt.Sprintf("dapr-sentry.%s.svc:443", security.CurrentNamespace()), "Address of the Sentry service")
-	fs.StringVar(&opts.PlacementAddress, "placement-address", "", "Addresses for Dapr Actor Placement service")
 	fs.StringVar(&opts.Mode, "mode", string(modes.StandaloneMode), "Runtime mode for Dapr Scheduler")
 
 	fs.StringVar(&opts.ID, "id", "dapr-scheduler-server-0", "Scheduler server ID")
--- a/pkg/actors/actors.go
+++ b/pkg/actors/actors.go
@@ -1126,11 +1126,13 @@ func (a *actorsRuntime) doExecuteReminderOrTimerOnInternalActor(ctx context.Cont
 			delete(a.internalReminderInProgress, key)
 			a.lock.Unlock()
 		}
-		if err != nil && !errors.Is(err, ErrReminderCanceled) {
-			log.Errorf("Error executing reminder for internal actor '%s': %v", reminder.Key(), err)
-		}
 
-		return err
+		if err != nil {
+			if !errors.Is(err, ErrReminderCanceled) {
+				log.Errorf("Error executing reminder for internal actor '%s': %v", reminder.Key(), err)
+			}
+			return err
+		}
 	}
 
 	return nil
--- a/pkg/actors/actors.go
+++ b/pkg/actors/actors.go
@@ -93,8 +93,6 @@ type ActorRuntime interface {
 }
 
 // Actors allow calling into virtual actors as well as actor state management.
-//
-//nolint:interfacebloat
 type Actors interface {
 	// Call an actor.
 	Call(ctx context.Context, req *internalv1pb.InternalInvokeRequest) (*internalv1pb.InternalInvokeResponse, error)
@@ -116,8 +114,6 @@ type Actors interface {
 	DeleteTimer(ctx context.Context, req *DeleteTimerRequest) error
 	// ExecuteLocalOrRemoteActorReminder executes a reminder on a local or remote actor.
 	ExecuteLocalOrRemoteActorReminder(ctx context.Context, reminder *CreateReminderRequest) error
-	// UsingScheduledReminders returns a bool indicating if the actor implementation is using the scheduler service
-	UsingSchedulerReminders() bool
 }
 
 // GRPCConnectionFn is the type of the function that returns a gRPC connection
@@ -129,7 +125,6 @@ type actorsRuntime struct {
 	appChannel         channel.AppChannel
 	placement          internal.PlacementService
 	placementEnabled   bool
-	schedulerReminders bool
 	grpcConnectionFn   GRPCConnectionFn
 	actorsConfig       Config
 	timers             internal.TimersProvider
@@ -149,6 +144,9 @@ type actorsRuntime struct {
 	closed             atomic.Bool
 	closeCh            chan struct{}
 	apiLevel           atomic.Uint32
+
+	lock                       sync.Mutex
+	internalReminderInProgress map[string]struct{}
 
 	// TODO: @joshvanl Remove in Dapr 1.12 when ActorStateTTL is finalized.
 	stateTTLEnabled bool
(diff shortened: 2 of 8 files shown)


cicoyle:
I've confirmed the db is clean after several thousand workflow iterations and the reminders are not triggered indefinitely even while removing the Repeat from 2, should be good with this diff 👍🏻

cicoyle (dismissed):

pkg/scheduler/server/api.go
```diff
@@ -57,9 +57,7 @@ func (s *Server) ScheduleJob(ctx context.Context, req *schedulerv1pb.ScheduleJob
 		Payload:  req.GetJob().GetData(),
 	}
 
-	reqCtx, cancel := context.WithCancel(context.Background())
-	defer cancel()
-	err = s.cron.Add(reqCtx, jobName, job)
+	err = s.cron.Add(ctx, jobName, job)
```
artursouza, line 60: This makes sense to me.

pkg/runtime/scheduler/streamer.go
```diff
@@ -110,23 +109,6 @@ func (s *streamer) handleJob(ctx context.Context, job *schedulerv1pb.WatchJobsRe
...
-			// this err is expected if the reminder was triggered once already, the next time it goes to get
-			// triggered by scheduler it sees that it's been canceled and does not invoke the 2nd time. This
-			// more relevant for workflows which currently has a repeat set to 2 since setting it to 1 caused
-			// issues. This will be updated in the future releases, but for now we will see this err. To not
-			// spam users only log if the error is not reminder canceled because this is expected for now.
-			if errors.Is(err, actors.ErrReminderCanceled) {
-				return
-			}
-
-			// If the actor was hosted on another instance, the error will be a gRPC status error,
```
artursouza, line 122: This was the change that Elena did this morning. Have you validated that the spam of error logs is not present after removing this?

pkg/actors/actors.go
```diff
@@ -1000,6 +995,10 @@ func (a *actorsRuntime) drainRebalancedActors() {
 	// visit all currently active actors.
 	var wg sync.WaitGroup
 
+	a.lock.Lock()
+	a.internalReminderInProgress = make(map[string]struct{})
```
artursouza, line 999: Why is this needed? Don't we have single thread guarantees anymore when using scheduler for reminders?

artursouza (changes requested):

cmd/scheduler/options/options.go
```diff
@@ -73,7 +73,6 @@ func New(origArgs []string) (*Options, error) {
 	fs.StringVar(&opts.TrustDomain, "trust-domain", "localhost", "Trust domain for the Dapr control plane")
 	fs.StringVar(&opts.taFile, "trust-anchors-file", securityConsts.ControlPlaneDefaultTrustAnchorsPath, "Filepath to the trust anchors for the Dapr control plane")
 	fs.StringVar(&opts.SentryAddress, "sentry-address", fmt.Sprintf("dapr-sentry.%s.svc:443", security.CurrentNamespace()), "Address of the Sentry service")
-	fs.StringVar(&opts.PlacementAddress, "placement-address", "", "Addresses for Dapr Actor Placement service")
```
artursouza, line 76: How we missed that? Thanks.

pkg/actors/actors.go
```diff
@@ -116,8 +114,6 @@ type Actors interface {
 	DeleteTimer(ctx context.Context, req *DeleteTimerRequest) error
 	// ExecuteLocalOrRemoteActorReminder executes a reminder on a local or remote actor.
 	ExecuteLocalOrRemoteActorReminder(ctx context.Context, reminder *CreateReminderRequest) error
-	// UsingScheduledReminders returns a bool indicating if the actor implementation is using the scheduler service
-	UsingSchedulerReminders() bool
```
artursouza, line 120: I am happy that we don't need this hack anymore.

pkg/actors/actors.go
```diff
@@ -1103,15 +1102,35 @@ func (a *actorsRuntime) doExecuteReminderOrTimerOnInternalActor(ctx context.Cont
...
+		key := reminder.Key()
+
+		log.Debugf("Executing reminder for internal actor '%s'", key)
+
+		a.lock.Lock()
+		if _, ok := a.internalReminderInProgress[key]; ok {
+			a.lock.Unlock()
+			// We don't need to return cancel here as the first invocation will
+			// delete the reminder.
+			log.Debugf("Duplicate concurrent reminder invocation detected for '%s', likely due to long processing time. Ignoring in favour of the active invocation", key)
```
artursouza, line 1114: This scares me.

pkg/runtime/scheduler/streamer.go
elena-kolevska: I think we're not going to get this errors anymore, because we're deleting the reminder.

elena-kolevska: I think it's because we're now deleting the reminders when completed and returning nil instead of the original ErrReminderCancelled error

pkg/actors/actors.go
JoshVanL: If an internal (workflow) actor reminder execution takes longer than 1m to process, then a duplicate reminder will be executed. To prevent duplication, we need to track what internal actor reminders are in progress to check against.

Existing application reminder guarantees should remain the same

pkg/actors/actors.go
JoshVanL: We could gate this map check behind the scheduler actor reminder feature gate?

> Is there a bug in how scheduler integrates into reminders that causes this abstraction leak? Firing reminders always had single thread guarantees in Dapr.

> Is there a bug in how scheduler integrates into reminders that causes this abstraction leak? Firing reminders always had single thread guarantees in Dapr.

We are only dealing with internal actor reminders here, which as generally implemented today, are a bit of a hack anyway w.r.t. how actors works generally in application code.

pkg/actors/actors.go
artursouza: As per our call, this code will "ensure" that there is not a memory leak. This code is executed for all actors (internal and external) but is not populating the map. I am fine with this for 1.14 since we plan to revisit internal actors' implementation in 1.15.

pkg/runtime/scheduler/streamer.go
artursouza: We are bringing it back as there was no manual verification that this is not needed anymore. Ideally, we should have automation but, at this stage, we need to be conservative.

artursouza (dismissed):

artursouza (approved):

cicoyle:
LGTM. Just ran my local test and saw the duplicated triggering since the run activity took longer than a minute and saw the log about ignoring it since one is still active, then deleting it. Post workflow run of 4k workflow iterations my db was clean still 👍🏻 

```
scheduler-0-1      | time="2024-08-13T13:49:41.788503301Z" level=debug msg="Triggering job: actorreminder||default||dapr.internal.default.workflow-a.activity||2953-1::0::1||run-activity" instance=bcbc20ef8d3b scope=dapr.scheduler.server type=log ver=edge
scheduler-0-1      | time="2024-08-13T13:50:41.004162967Z" level=debug msg="Triggering job: actorreminder||default||dapr.internal.default.workflow-a.activity||2953-1::0::1||run-activity" instance=bcbc20ef8d3b scope=dapr.scheduler.server type=log ver=edge

workflow-dapr-a-1  | time="2024-08-13T13:50:51.394679014Z" level=debug msg="Executing reminder for internal actor 'dapr.internal.default.workflow-a.activity||2953-1::0::1||run-activity'" app_id=workflow-a instance=e455363315f3 scope=dapr.runtime.actor type=log ver=edge
workflow-dapr-a-1  | time="2024-08-13T13:50:51.394693722Z" level=debug msg="Duplicate concurrent reminder invocation detected for 'dapr.internal.default.workflow-a.activity||2953-1::0::1||run-activity', likely due to long processing time. Ignoring in favour of the active invocation" app_id=workflow-a instance=e455363315f3 scope=dapr.runtime.actor type=log ver=edge
workflow-dapr-a-1  | time="2024-08-13T13:50:51.394715305Z" level=debug msg="Deleting reminder which was cancelled: dapr.internal.default.workflow-a.activity||2953-1::0::1||run-activity" app_id=workflow-a instance=e455363315f3 scope=dapr.runtime.actor type=log ver=edge
```

cicoyle (approved):

mikeee (approved):

PR #7998, opened 2024-08-14
Fix: Don't track duplicate in-flight reminders for non-scheduler reminders

JoshVanL:
Don't track in-flight internal (workflow) reminders when not using scheduler reminders. This is because they only get un-tracked when they are deleted in the scheduler code-path. This caused workflow activities to not fire when triggering a workflow multiple times with the same instance ID - since the had the same reminder key.

Adds int tests for both scheduler and non-scheduler reminders which trigger a workflow with the same instance ID multiple times.

### DIFF
--- a/pkg/actors/actors.go
+++ b/pkg/actors/actors.go
@@ -145,8 +145,9 @@ type actorsRuntime struct {
 	closeCh            chan struct{}
 	apiLevel           atomic.Uint32
 
-	lock                       sync.Mutex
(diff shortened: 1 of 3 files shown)


yaron2 (approved):

cicoyle:
[I have a fix for the unrelated failing int test in this PR](https://github.com/dapr/dapr/pull/7982)

cicoyle (approved):

Issue #7955, opened 2024-07-26
Actor service failure detection by Placement Service Needs to be configurable


rahulpoddar-fyndna:
The frequency at which placement service detects failure of the pod hosting actor services needs to be minimum so that failures of http request are minimized. 
If an actor service receives multiple concurrent requests and if the actor service fails during this time, the placement service detects this failure and instantiates the actors on one of the existing pods. As of now the placement services polls/detects in every 2 seconds. If we have this frequency of detecting actor pod configurable, then we can tune minimise the http requests failures. The retry resiliency can be configured to keep it close to polling interval of placement service.

cicoyle:
docs are needed for this too plz

elena-kolevska:
I think we're not documenting the placement cli arguments in docs, only daprd. We do document the values in both the helm values readme and the cli itself.

elena-kolevska:
@elena-kolevska to write a documentation section about running placement in production ([on this page](https://docs.dapr.io/operations/hosting/kubernetes/kubernetes-production/)).

elena-kolevska:
Here's the docs PR: https://github.com/dapr/docs/pull/4365

PR #8015, opened 2024-08-27
Configurable placement keep alive time, keep alive timeout and dissemination timeout

elena-kolevska:
# Description
Since the grpc keep alive time and timeout depend largely on the environment (network stability), we add the possibility to adjust them through environment variables.

## Issue reference

https://github.com/dapr/dapr/issues/7955

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests
* [X] Unit tests passing
* [X] Extended the documentation / Created PR in the docs repo: https://github.com/dapr/docs/pull/4365


cmd/placement/options/options.go
```diff
@@ -148,3 +177,22 @@ func parsePeersFromFlag(val []string) []raft.PeerInfo {
 
 	return peers[:i]
 }
+
+// getEnvIntWithRange returns the integer value of the environment variable with the given name.
+func getEnvIntWithRange(envVar string, defaultValue int, min int, max int) (int, error) {
```
artursouza, line 182: Cut a PR in dapr/kit too. I will not block this, but it can be an easy follow up to remove this utility method from here.

artursouza (dismissed):

cmd/placement/options/options.go
elena-kolevska: I expected this feedback :) 
Ok, on it.

cmd/placement/options/options.go
```diff
@@ -36,6 +37,14 @@ const (
...
 	defaultReplicationFactor = 100
 	envMetadataEnabled       = "DAPR_PLACEMENT_METADATA_ENABLED"
+
+	envKeepAliveTime      = "DAPR_PLACEMENT_KEEPALIVE_TIME"
+	envKeepAliveTimeout   = "DAPR_PLACEMENT_KEEPALIVE_TIMEOUT"
+	envDisseminateTimeout = "DAPR_PLACEMENT_DISSEMINATE_TIMEOUT"
+
+	defaultKeepAliveTime      = 2 // in seconds
+	defaultKeepAliveTimeout   = 3 // in seconds
+	defaultDisseminateTimeout = 2 // in seconds
```
JoshVanL, line 47: Can we please make these Go duration strings to be consistent?

cmd/placement/options/options.go
```diff
@@ -63,12 +72,16 @@ type Options struct {
 
 	ReplicationFactor int
 
+	KeepAliveTime      int
+	KeepAliveTimeout   int
+	DisseminateTimeout int
```
JoshVanL, line 77:
```suggestion
	KeepAliveTime      time.Duration
	KeepAliveTimeout   time.Duration
	DisseminateTimeout time.Duration
```

cmd/placement/options/options.go
```diff
@@ -36,6 +37,14 @@ const (
 	defaultPlacementPort     = 50005
 	defaultReplicationFactor = 100
 	envMetadataEnabled       = "DAPR_PLACEMENT_METADATA_ENABLED"
+
+	envKeepAliveTime      = "DAPR_PLACEMENT_KEEPALIVE_TIME"
+	envKeepAliveTimeout   = "DAPR_PLACEMENT_KEEPALIVE_TIMEOUT"
+	envDisseminateTimeout = "DAPR_PLACEMENT_DISSEMINATE_TIMEOUT"
```
JoshVanL, line 43: Why are these environment variables, rather than CLI flags?

JoshVanL (changes requested):

cmd/placement/options/options.go
elena-kolevska: Here's the kit PR: https://github.com/dapr/kit/pull/102

cmd/placement/options/options.go
elena-kolevska: I thought about it too, but defaulted to integers because values outside of the specified ranges (~1-10s) don't make sense and will be rejected, and also to follow what we already have in daprd (`app-health-probe-timeout`, `app-health-probe-interval`, `dapr-graceful-shutdown-seconds`).
I'm not opposed to changing it if you feel strongly about it, the benefit would be that people could use fractional values like 1.5 seconds.

cmd/placement/options/options.go
elena-kolevska: That was my first approach, too if you look at the commit history, but the initial feedback was that it's better not to expose these values too much because wrong configuration values can substantially degrade the service.

cmd/placement/options/options.go
JoshVanL: We can continue to have validation on ranges, regardless of the format.

As a project we are trying to converge on flags and options using unitised values- it's otherwise very confusing for consumers to attempt to guess or other wise have to think _more_ about the what unit the field is in. see https://github.com/dapr/dapr/commit/c920e26671cc75805fc7ca1b7bb50720538e4241 https://github.com/dapr/dapr/pull/7268/files

cmd/placement/options/options.go
JoshVanL: We should never silently drop configuration values- this results in users having no idea about whether configuration has actually been applied to the system or not and basically having a miss-trust in the software completely. We have this issue today already in our CRDs like Components and Resiliency.

> wrong configuration values can substantially degrade the service

If the configuration is wrong, we should error exit daprd. We shouldn't try to hide configuration- if the configuration itself has large performance impact, this should be clearly documented as such.

Using CLI flags is the _main_ method for configuring runtime options on daprd. Environment variables should be considered an escape hatch.

elena-kolevska: I agree completely on the failing silently comment, and we actually do that ([example](https://github.com/dapr/dapr/pull/8015/files#diff-ef615d6fc2542a98f506d5b7aae40e875a21770cc30427e8a20f2f942152f453R102)).
When I said "wrong configuration values" I meant values that aren't optimal for one network but would be completely valid for another (a better term would have been "suboptimal configuration values"). Values outside the range of what we consider acceptable will be rejected on init.

I also agree with the documentation part and tried to document it as clearly as possible in this [PR](https://github.com/dapr/docs/pull/4327/files).

Re: CLI flags vs env variables: the feedback was that we want to avoid users seeing the flags exposed and feel enticed to change their values without doing a proper benchmark of their network latency and stability. The counter-argument to that is that users in production systems usually spend a lot of time researching and tweaking every argument, but both of these are just assumptions, of course. I'm open to reverting to flags if the team agrees, that way we keep everything nice and organised.

@artursouza @alicejgibbons  what do you think?

cmd/placement/options/options.go
elena-kolevska: Great, I'm on it!
I love that we're making a move to standardise this.

cmd/placement/options/options.go
alicejgibbons: The initial reason for using env vars was that this was a very niche option for specific use cases and didn't necessarily need to be "as discoverable" as CLI flags. If we are moving away from env vars to standardize on cli flags then that approach is fine.

There were also concerns that these cli flags might not be backwards compatible, so if they are provided in version 1.15 daprd then downgraded to this could cause errors with older versions.

elena-kolevska: I think the downgrading problem is not specific to these flags; it's a more generic thing we should address on a different level.

Maybe we should just do this the "dapr way" and revert to flags. As @JoshVanL was saying above, having the parameters clearly documented should prevent misconfigurations by users.

cicoyle:
@elena-kolevska can you fix the lint issues?

elena-kolevska:
Docs PR: https://github.com/dapr/docs/pull/4365

elena-kolevska:
Thanks @JoshVanL, this was because of the new linter version and rules that were merged in the meantime. Fixed now.

elena-kolevska:
@JoshVanL , as agreed, I updated the `disseminateTimeoutMax` value to 3 seconds and sent a docs PR: https://github.com/dapr/docs/pull/4372.
As a next step, I'll update this code to use an exponential rate limiting algorithm, for better performance.

JoshVanL (approved):

Issue #7958, opened 2024-07-31
Add `securityContext/runAsGroup` and `securityContext/runAsUser` as  sidecar injector options.


rochabr:
## In what area(s)?

> /area runtime

/area operator

> /area placement

> /area docs

> /area test-and-release

## Describe the feature
The [Gatekeeper Library](https://github.com/open-policy-agent/gatekeeper-library) is throwing two warnings for sidecar containers:

```
[allowed-user-ranges] Container <daprd> is attempting to run without a required securityContext/runAsGroup. Allowed runAsGroup: {"ranges": [{"max": 65535, "min": 1}], "rule": "MustRunAs"}

[allowed-user-ranges] Container <daprd> is attempting to run without a required securityContext/runAsUser. Allowed runAsUser: {"ranges": [{"max": 65535, "min": 1}], "rule": "MustRunAs"}
```

To mitigate that, we need to explicitly define `securityContext/runAsGroup` and `securityContext/runAsUser` as [sidecar injector options](https://github.com/dapr/dapr/blob/master/charts/dapr/README.md#dapr-sidecar-injector-options), which are not available today. They need to be added.

artursouza:
@siri-varma how is your progress on this issue?

siri-varma:
@artursouza I did not make much progress here and I will not be able for a week. I will unassign myself.

PR #8019, opened 2024-08-28
Add support for runAsUser and runAsGroup for injected sidecar

antontroshin:
# Description

Add support for `securityContext` fields `runAsUser` and `runAsGroup` on the injected `daprd` sidecar.
Some security scanners require the presence of these when `securityContext.runAsNonRoot` is set to `true`

## Issue reference

Please reference the issue this PR will close: #7958

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Unit tests passing


daixiang0 (dismissed):
LGTM

pkg/injector/service/config.go
```diff
@@ -204,3 +227,15 @@ func isTruthyDefaultTrue(val string) bool {
 	}
 	return kitutils.IsTruthy(val)
 }
+
+func parseStringToInt64Pointer(intStr string) (*int64, error) {
```
ytimocin, line 231: Should this be in **dapr/kit**?

antontroshin: Thanks for the suggestion. The conversion here is very specific to this use case. I've searched through the `dapr/dapr` code for a similar conversion of `string` to `int64` pointer, and all I've found is regular `string` to `int` (Atoi) conversion. 
At this point, I don't see a good reason to add this to `dapr/kit`, but we might in the future if more use cases will appear.

artursouza (approved):

PR #8023, opened 2024-08-29
Updates k8s version for e2e tests on AKS

elena-kolevska:
# Description

e2e tests have been failing for a while because of old k8s version that is not supported any more in the region we're using. The currently supported versions in the westus3 are 1.27 through 1.30. This PR updates the k8s version we're using to 1.30.


JoshVanL (approved):
👍 1.27 has been out of support for a while
https://kubernetes.io/releases/

mikeee (approved):
Lgtm, I think 1.27 has been dropped in all aks regions following the 1.31 GA release 👍

artursouza (approved):

PR #8047, opened 2024-09-04
Workflows: Fix daprd memory leak

### LINKED ISSUE #7777: Potential workflow -sidecar memory leak
## In what area(s)?

/area runtime

> /area operator

> /area placement

> /area docs

/area test-and-release

## What version of Dapr?

edge/master

> 1.1.x
> 1.0.x

## Expected Behavior

The sidecar resource usage should remain consistent.

## Actual Behavior

There is potentially a memory leak, the sidecar memory profile is increasing exponentially with the emitter application resource usage remaining consistent.
image
image

Note: This is currently only observed in the **weekly** longhaul cluster.

Investigate what the backend is doing and whether this can be attributed to use of an in-memory backend.

## Release Note

RELEASE NOTE:

JoshVanL:
Fixes a memory leak in Workflows whereby the internal actor state held by daprd would never be released- meaning daprd would grow in memory indefinitely after each workflow execution. This memory leak occurs for both scheduler and non-scheduler reminders.

Specifically, the actor table and internal actor workflow state maps where never deleted after workflow orchestration and activity completion. PR updates the workflow and activity actors to track whether it has been completed. This is signalled to the actor runtime after a workflow reminder execution, and the actor runtime will then delete the actor state.

Adds integration tests to ensure daprd memory usage does not grow for both scheduler and non-scheduler workflows.

Will close #7777

### DIFF
--- a/pkg/runtime/wfengine/backends/actors/activity_actor.go
+++ b/pkg/runtime/wfengine/backends/actors/activity_actor.go
@@ -22,6 +22,7 @@ import (
 	"errors"
 	"fmt"
 	"strings"
+	"sync/atomic"
 	"time"
 
 	"github.com/microsoft/durabletask-go/api"
@@ -46,6 +47,7 @@ type activityActor struct {
 	defaultTimeout   time.Duration
 	reminderInterval time.Duration
 	config           actorsBackendConfig
+	completed        atomic.Bool
 }
 
 // ActivityRequest represents a request by a worklow to invoke an activity.
@@ -138,7 +140,10 @@ func (a *activityActor) InvokeReminder(ctx context.Context, reminder actors.Inte
 	timeoutCtx, cancelTimeout := context.WithTimeout(ctx, a.defaultTimeout)
 	defer cancelTimeout()
 
-	err := a.executeActivity(timeoutCtx, reminder.Name, state.EventPayload)
+	completed, err := a.executeActivity(timeoutCtx, reminder.Name, state.EventPayload)
+	if completed == runCompletedTrue {
+		a.completed.Store(true)
+	}
 
 	var recoverableErr *recoverableError
 	// Returning nil signals that we want the execution to be retried in the next period interval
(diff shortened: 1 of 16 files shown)


cicoyle (dismissed):
Great stuff

pkg/runtime/wfengine/backends/actors/activity_actor.go
```diff
@@ -138,7 +140,10 @@ func (a *activityActor) InvokeReminder(ctx context.Context, reminder actors.Inte
 	timeoutCtx, cancelTimeout := context.WithTimeout(ctx, a.defaultTimeout)
 	defer cancelTimeout()
 
-	err := a.executeActivity(timeoutCtx, reminder.Name, state.EventPayload)
+	completed, err := a.executeActivity(timeoutCtx, reminder.Name, state.EventPayload)
+	if completed == runCompletedTrue {
+		a.completed.Store(true)
```
famarting, line 145: completed will be forever true once there has been one reminder execution that set this to true.... dont you need some logic to make it go to false again?

or what are the semantics of this new `Completed() bool` function that you added?

JoshVanL: Correct- because workflows (and activities) reach a terminal state, they will be set as `Completed` 'forever'. When checked by the actor runtime, this being true if what makes this struct be deleted from the internal actors map and garbage collected. Without this, we have the situation today where internal actor state (workflow history etc.) is never garbage collected and the memory leak occurs.

pkg/runtime/wfengine/backends/actors/workflow_actor.go
```diff
@@ -672,7 +682,7 @@ func (wf *workflowActor) runWorkflow(ctx context.Context, reminder actors.Intern
 
 	err = wf.saveInternalState(ctx, state)
 	if err != nil {
-		return err
+		return runCompletedTrue, err
```
yaron2, line 685: Shouldn't this be `runCompletedFalse`? saving state is retriable?

JoshVanL: Yes, I think it probably should to but the _current_ implementation treats this as a terminal error and didn't want to change behaviour in this PR.

https://github.com/dapr/dapr/blob/ca411ccd824f60a2d7286ff39c9c7f4d3bd64102/pkg/runtime/wfengine/backends/actors/workflow_actor.go#L166

yaron2: Ok. lets leave this for now, but this is a question for @cgillum also.

yaron2:
One of the tests seems to be failing on Windows now consistently?

JoshVanL:
@yaron2 this is due to the sqlite limitations on windows... I've skipped this test for now on Windows. This is a known issue which hasn't been resolved yet. https://github.com/dapr/dapr/blob/09f6a0f8f6264a2534adf9a4577c13baed751514/tests/integration/suite/actors/reminders/serialization/protobuf.go#L54

yaron2 (dismissed):

mikeee (dismissed):

yaron2 (approved):

yaron2:
@JoshVanL please backport to 1.14

pkg/runtime/wfengine/backends/actors/workflow_actor.go
cgillum: I unfortunately don't remember this clearly. It seems like it should be a recoverable error. Perhaps I was worried about the cases where it's not recoverable, such as when you run into certain state store limitations around payload sizes. Happy to revisit this later.

PR #8067, opened 2024-09-08
fix: add apiVersion and kind to volumeClaimTemplates

aqeelat:
# Description

Kubernetes adds apiVersion and kind to volumeClaimTemplates, which makes the StatefulSets always appear out of sync when using any manifest diffing tool such as helm diff, argocd, or flux.

This PR adds them to the helm template to avoid this issue.

## Issue reference

Please reference the issue this PR will close: none

## Checklist

- [ ] Code compiles correctly
- [ ] Created/updated tests
- [ ] Unit tests passing

### DIFF
--- a/charts/dapr/charts/dapr_placement/templates/dapr_placement_statefulset.yaml
+++ b/charts/dapr/charts/dapr_placement/templates/dapr_placement_statefulset.yaml
@@ -256,7 +256,9 @@ spec:
 {{- if or (eq .Values.global.ha.enabled true) (eq .Values.ha true) }}
   {{- if eq .Values.cluster.forceInMemoryLog false }}
   volumeClaimTemplates:
-  - metadata:
+  - apiVersion: v1
+    kind: PersistentVolumeClaim
+    metadata:
       name: raft-log
     spec:
       accessModes:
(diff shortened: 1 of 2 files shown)


JoshVanL:
Thanks for the contribution @aqeelat, lgtm!

For those coming across this problem, a general resolution to this is enabling the **"server side apply diff" (or equivalent)** in whatever tool you are using. This may be a config option, CLI flag, or resource annotation depending on the tool your using.

Kube CD tools can, and often do by default, determine resource difference/drift by doing simple string diffs of fields between desired and current states of the world. The Kuberneres API server set a default value on some fields which are committed by the client/applier (like the fields on this PR). Although the resource manifests are _semantically_ identical, the yaml manifest strings are different. Setting this option will set the API server defaults (and execute mutation webhooks) on the resource before doing the diff, resolving the discrepancy on strings of the two resources and stopping the tool falsely complaining.

It is generally good practice to always have this option enabled globally.

You can read more discussion on this [arogo-cd issue](https://github.com/argoproj/argo-cd/issues/11143).

JoshVanL (approved):

PR #8075, opened 2024-09-10
Automatically update the dapr version in long haul tests on new releases

elena-kolevska:
Automatically update the dapr version in long haul tests whenever we have a new release. This will require a GitHub application and updating two Github Secrets.

Related to https://github.com/dapr/test-infra/pull/235

Note: We will do the same change in the cli repo.

Issue:
https://github.com/dapr/dapr/issues/8089


.github/workflows/dapr.yml
```diff
@@ -687,3 +687,38 @@ jobs:
...
+    runs-on: ubuntu-latest
+    needs: helmpublish
+    if: startswith(github.ref, 'refs/tags/v') && github.repository_owner == 'dapr'
+    steps:
+      - name: Get Token
+        id: get_workflow_token
+        uses: peter-murray/workflow-application-token-action@v2
+        with:
+          application_id: ${{ secrets.LONGHAUL_APPLICATION_ID }}
+          application_private_key: ${{ secrets.LONGHAUL_APPLICATION_PRIVATE_KEY }}
```
mikeee, line 702: https://github.com/actions/create-github-app-token
I think it'd be worth throwing this official action into the mix for consideration

mikeee (changes requested):
Lgtm - just a concern about the action used to generate a short-lived token

.github/workflows/dapr.yml
elena-kolevska: Yep, done! I had automatically gone with the one I knew from another project, but official is better, for sure!

elena-kolevska:
Also added a check if the version is newer than the one previously deployed in the longhaul tests (ex. we don't want to update them with an old patch version on 1.13)

.github/scripts/compare_versions.py
```diff
@@ -0,0 +1,48 @@
...
+    if new_ver > existing_ver:
+        print("New version is greater than existing version.")
+        with open(os.getenv("GITHUB_ENV"), "a") as github_env:
+            github_env.write("UPDATE_REQUIRED=true\n")
+
+    else:
+        print("New version is not greater. Skipping update.")
+        with open(os.getenv("GITHUB_ENV"), "a") as github_env:
+            github_env.write("UPDATE_REQUIRED=false\n")
+    sys.exit(0)
```
mikeee, line 39: minor nit: I think a boolean should be returned rather than exiting within this function and duplicating the write steps

mikeee (changes requested):
Just one thing is needed to prevent the job from starting immediately as soon as the workflow is triggered. I've suggested the change to trigger the job after the final publish step.
(feel free to ignore the nit)

mikeee (approved):
Lgtm

.github/scripts/compare_versions.py
```diff
@@ -0,0 +1,55 @@
...
+
+    update_required = new_ver > existing_ver
+    status_message = "New version is greater than existing version." if update_required else "New version is not greater. Skipping update."
+    print(status_message)
+
+    # Write the update requirement status to GITHUB_ENV
+    github_env_path = os.getenv("GITHUB_ENV")
+    if github_env_path:
+        with open(github_env_path, "a") as github_env:
+            github_env.write(f"UPDATE_REQUIRED={str(update_required).lower()}\n")
```
cicoyle, line 41: Should this be `VERSION_UPDATE_REQUIRED` to be explicit given its an env var?

cicoyle (commented):
Has this been tested - do you by chance have a link to this working in your fork? Just want to spot check this in case there are issues ahead of a merge :)

.github/scripts/compare_versions.py
elena-kolevska: Good point, I'll change.

elena-kolevska:
@cicoyle  Here's a successful run of the action, adapted for another repo: https://github.com/elena-kolevska/repoA/actions/runs/10960439399

.github/workflows/dapr.yml
```diff
@@ -687,3 +687,57 @@ jobs:
...
+        with:
+          app-id: ${{ secrets.APPLICATION_ID }}
+          private-key: ${{ secrets.APPLICATION_PRIVATE_KEY }}
+          repositories: ${{ env.LONG_HAUL_REPO }}
+      - name: Check out code into the Go module directory
+        uses: actions/checkout@v4
+        with:
+          path: dapr
+      - name: Checkout the longhaul tests repo
+        uses: actions/checkout@v3
```
mikeee, line 711:
```suggestion
        uses: actions/checkout@v4
```

mikeee (changes requested):

mikeee (approved):

yaron2 (approved):

Issue #8077, opened 2024-09-10
Fix Jobs HTTP API to be native to application/json


artursouza:
## In what area(s)?

> /area runtime

> /area operator

> /area placement

> /area docs

> /area test-and-release

## What version of Dapr?

> 1.1.x
> 1.0.x
> edge: output of `git describe --dirty`

1.14.2

## Expected Behavior

Applications written with expressJS (or similar frameworks) that expect trigger payloads to be native JSON should work directly with the Jobs API over HTTP.

Instead, to schedule an API over HTTP, it should be native JSON without any wrapping or metadata for protobuf (serialization to etcd is agnostic of the HTTP API and can use protobuf):

```
curl --location 'http://localhost:3500/v1.0-alpha1/jobs/myjob' \
--header 'Content-Type: application/json' \
--data-raw '{
  "data": "someData",
  "dueTime": "1s"
}'
```

It also means that when triggering a job for an app over HTTP, it should always include `application/json` content-type header and the body:

```
POST /job/myjob HTTP/1.1
Host: 127.0.0.1:3000
User-Agent: Go-http-client/1.1
Content-Length: 999999
Accept-Encoding: gzip
Content-Type: application/json
"someData"
```

## Actual Behavior

See: https://github.com/dapr/dapr/issues/8076

## Steps to Reproduce the Problem

See https://github.com/dapr/dapr/issues/8076

## Release Note

RELEASE NOTE: **FIXED** Jobs API's data over HTTP to be JSON

Issue #8076, opened 2024-09-10
Jobs API is base64 encoding the data on trigger


artursouza:
## In what area(s)?

 /area runtime

> /area operator

> /area placement

> /area docs

> /area test-and-release

## What version of Dapr?

> 1.1.x
> 1.0.x
> edge: output of `git describe --dirty`
1.14.2

## Expected Behavior

Trigger should send the data as-is when it was scheduled at the Jobs API, all over HTTP.

## Actual Behavior

Data is base64 encoded.

## Steps to Reproduce the Problem

- Install Dapr 1.14.2 locally: `dapr init --runtime-version=1.14.2`
- Run it locally: `npm install -g http-echo-server`; `dapr run --dapr-http-port 3500 --app-port 3000 -- http-echo-server 3000`
- On a new terminal, schedule a job:
```
curl --location 'http://localhost:3500/v1.0-alpha1/jobs/myjob' \
--header 'Content-Type: application/json' \
--data-raw '{
  "data": {
	  "@type": "type.googleapis.com/google.protobuf.StringValue",
	  "value": "\"someData\""
    },
    "dueTime": "1s"
}'
```
See bug:
```
== APP == --> POST /job/myjob HTTP/1.1
== APP == --> Host: 127.0.0.1:3000
== APP == --> User-Agent: Go-http-client/1.1
== APP == --> Content-Length: 89
== APP == --> Accept-Encoding: gzip
== APP == --> 
== APP == --> {"type_url":"type.googleapis.com/google.protobuf.StringValue","value":"Cgoic29tZURhdGEi"}
```

## Release Note

RELEASE NOTE: **FIXED** Jobs data serialization/deserialization over HTTP API

mikeee:
@artursouza Did you appraise the gRPC implementation as well? iirc this is also something that needs to be addressed, the example I implemented required me to retrieve the value and then base64 decode the payload.
ref (implementation in the go-sdk):
https://github.com/dapr/go-sdk/pull/562/files#diff-4975dda9114c3801d072d31bde655b3482bb5ae38e745aa99546ef22f56c05a9R98-R114

artursouza:
[...]

PR #8083, opened 2024-09-11
[1.14] Jobs: change HTTP Create data request body to JSON

JoshVanL:
Previously, the HTTP Create data request body accepted only a fully qualified proto Any message- a object of type URL and Value. The API now accepts a JSON object string, i.e. number, string, boolean, object, etc. On trigger, the HTTP app will receive a JSON string of the Job data. In the cases where the data is not of type `google.protobuf.Value` or some other primitive encoded type, the data may contain binary data.

An internal `JobHTTPRequest` proto message has been added for the HTTP endpoint to allow of validation and ingestion of JSON string data in the request body.

The HTTP and gRPC channels now construct trigger job requests separately to allow for the different data serializations.

Closes https://github.com/dapr/dapr/issues/8077 https://github.com/dapr/dapr/issues/8076

Docs: https://github.com/dapr/docs/pull/4349


yaron2:
Linter

JoshVanL:
@yaron2 updated

cicoyle (dismissed):
I tested this locally with my test-apps and it works like magic 🪄

tests/integration/suite/daprd/jobs/http/basic.go
```diff
@@ -84,7 +84,8 @@ func (b *basic) Run(t *testing.T, ctx context.Context) {
 		for _, body := range []string{
 			`{"schedule": "@daily"}`,
 			`{"schedule": "@daily", "repeats": 3, "due_time": "10s", "ttl": "11s"}`,
-			`{"schedule": "@daily", "repeats": 3, "due_time": "10s", "ttl": "11s", "data": {"@type": "type.googleapis.com/google.protobuf.StringValue", "value": "Hello, World!"}}`,
+			`{"schedule": "@daily", "repeats": 3, "due_time": "10s", "ttl": "11s", "data": "{\"@type\": \"type.googleapis.com/google.protobuf.StringValue\", \"value\": \"Hello, World!\"}"}`,
```
artursouza, line 87: I thought we agreed not to have this format anymore. I am OK for now, but I vote for this to be removed for HTTP in 1.15.

JoshVanL: Yes. We don't ingest data payloads as proto objects and instead as JSON strings. as agreed.

It's still valid JSON though, and will be ingested as such. Please see the addition of quotations, and the next test case.

artursouza (approved):

Issue #8078, opened 2024-09-10
Workflow API: No Activity State Found


cicoyle:
## In what area(s)?

> /area runtime

> /area operator

> /area placement

> /area docs

> /area test-and-release

## What version of Dapr?

> 1.1.x
> 1.0.x
> edge: output of `git describe --dirty`

## Expected Behavior

Workflow to execute successfully both with and without scheduler and not throw errors about `no activity state found`.

## Actual Behavior

Currently user is seeing the following w/ v 1.14.2-rc.3 w/o Scheduler:
```
Error executing reminder for internal actor <run-activity>: no activity state found, scope: dapr.runtime.actor
error invoking reminder on actor <activity actor>: no activity state found, scope: dapr.runtime.actor
```
With Scheduler enabled it appears to be a similar err:
```
failed to invoke scheduled actor reminder named run-activity due to: no activity state found
Error executing reminder for internal actor <run-activity>: no activity state found, scope: dapr.runtime.actor
failed to invoke scheduled actor reminder named: run-activity due to: no activity state found, scope: dapr.runtime.scheduler
```

## Steps to Reproduce the Problem

Use Workflows without Scheduler, also seems to impact if using Scheduler so once root issue is fixed it _should_ auto work with Scheduler.

## Release Note

RELEASE NOTE: FIXED Workflow API no activity state found

arturotrenard:
Hello I remain attentive thank you very much

PR #7949, opened 2024-07-26
activity actor: prevent nil pointer

famarting:
# Description

If a workflow activity somehow is being retried after a successfull execution the activity actor can incurr into a nil pointer error on line 141

## Issue reference

Please reference the issue this PR will close: #_[issue number]_

## Checklist

Please make sure you've  completed the relevant tasks for this PR, out of the following list:

* [ ] Code compiles correctly
* [ ] Created/updated tests
* [ ] Unit tests passing
* [ ] End-to-end tests passing
* [ ] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_
* [ ] Specification has been updated / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_
* [ ] Provided sample for the feature / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_

PR #8084, opened 2024-09-11
Workflow: Prevent nil ptr & terminating early

cicoyle:
[reverting this change](https://github.com/dapr/dapr/pull/7949) while still preventing the nil ptr. This is done by returning an empty struct and logging the err instead of terminating the logic which is causing workflows to not finish and err out bc that PR changed the execution logic to terminate invoking the internal actor. Reverting that change and logging instead, such that users shouldn't see the err and it wont terminate the logic.

See err here:
Screenshot 2024-09-12 at 9 49 10 AM

I am unable to see this err with our existing tests and having a hard time reproing it, so this PR is simply reverting the logical change from [this PR](https://github.com/dapr/dapr/pull/7949) but still logging it.

This PR once complete will close [this issue.](https://github.com/dapr/dapr/issues/8078)

### DIFF
--- a/pkg/runtime/wfengine/backends/actors/activity_actor.go
+++ b/pkg/runtime/wfengine/backends/actors/activity_actor.go
@@ -134,7 +134,7 @@ func (a *activityActor) InvokeReminder(ctx context.Context, reminder actors.Inte
 	state, err := a.loadActivityState(ctx)
 	// TODO: On error, reply with a failure - this requires support from durabletask-go to produce TaskFailure results
 	if err != nil || len(state.EventPayload) == 0 {
-		wfLogger.Errorf("no activity state found for reminder: %s, err: %s", reminder.Name, err)
+		wfLogger.Errorf("no activity state found for internal wf activity actor: %s, err: %s", reminder.Name, err)
 	}
 
 	timeoutCtx, cancelTimeout := context.WithTimeout(ctx, a.defaultTimeout)
--- a/pkg/runtime/wfengine/backends/actors/activity_actor.go
+++ b/pkg/runtime/wfengine/backends/actors/activity_actor.go
@@ -131,10 +131,10 @@ func (a *activityActor) InvokeMethod(ctx context.Context, methodName string, dat
 func (a *activityActor) InvokeReminder(ctx context.Context, reminder actors.InternalActorReminder, metadata map[string][]string) error {
 	wfLogger.Debugf("Activity actor '%s': invoking reminder '%s'", a.actorID, reminder.Name)
 
-	state, _ := a.loadActivityState(ctx)
+	state, err := a.loadActivityState(ctx)
 	// TODO: On error, reply with a failure - this requires support from durabletask-go to produce TaskFailure results
-	if state == nil {
-		return errors.New("no activity state found")
+	if err != nil || len(state.EventPayload) == 0 {
+		wfLogger.Errorf("no activity state found for reminder: %s, err: %s", reminder.Name, err)
 	}
 
 	timeoutCtx, cancelTimeout := context.WithTimeout(ctx, a.defaultTimeout)
@@ -312,7 +312,7 @@ func (a *activityActor) loadActivityState(ctx context.Context) (*activityState, 
 	}
 	res, err := a.actorRuntime.GetState(ctx, &req)
 	if err != nil {
-		return nil, fmt.Errorf("failed to load activity state: %w", err)
+		return &activityState{}, fmt.Errorf("failed to load activity state: %w", err)
 	}
 
 	if len(res.Data) == 0 {
@@ -323,7 +323,7 @@ func (a *activityActor) loadActivityState(ctx context.Context) (*activityState, 
 	state := &activityState{}
 	err = json.Unmarshal(res.Data, state)
 	if err != nil {
-		return nil, fmt.Errorf("failed to unmarshal activity state: %w", err)
+		return &activityState{}, fmt.Errorf("failed to unmarshal activity state: %w", err)
 	}
 	return state, nil
 }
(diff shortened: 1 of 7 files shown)


pkg/runtime/wfengine/backends/actors/backend.go
```diff
@@ -412,7 +412,7 @@ func (abe *ActorBackend) String() string {
 
 func (abe *ActorBackend) validateConfiguration() error {
 	if abe.actorRuntime == nil {
-		return errors.New("actor runtime has not been configured")
+		return errors.New("actor runtime has not been configured. Ensure Placement is running")
```
cicoyle, line 415: added this for for clarity, see discord user facing this and it [not being clear placement was not running](https://discord.com/channels/778680217417809931/1075836407156850698/1283744676884512871)

pkg/runtime/wfengine/backends/actors/activity_actor.go
```diff
@@ -131,10 +131,10 @@ func (a *activityActor) InvokeMethod(ctx context.Context, methodName string, dat
...
-	state, _ := a.loadActivityState(ctx)
+	state, err := a.loadActivityState(ctx)
 	// TODO: On error, reply with a failure - this requires support from durabletask-go to produce TaskFailure results
-	if state == nil {
-		return errors.New("no activity state found")
+	if err != nil || len(state.EventPayload) == 0 {
+		wfLogger.Errorf("no activity state found for reminder: %s, err: %s", reminder.Name, err)
 	}
 
 	timeoutCtx, cancelTimeout := context.WithTimeout(ctx, a.defaultTimeout)
```
mukundansundar, line 140: does the following logic need to be executed? 
Can't we move directly to the error switch case below or do we still support executing the workflow activity without state data?

pkg/runtime/wfengine/backends/actors/backend.go
mukundansundar: nitpick: "... Placement service ..."
but is this clarified in terms of the scheduler backend?

pkg/runtime/wfengine/backends/actors/activity_actor.go
cicoyle: Based on the user seeing this err and having issues with his workflows, where their only way to recover was to delete the state table in Postgres, I believe we do want to support workflow activity without state data or at least not return here. The [prior PR](https://github.com/dapr/dapr/pull/7949) Im reverting changed the logic to return here, whereas before we weren't returning for this case.

cc @cgillum for more info here please, I see the activity state is used for the history details (event payload) saved to a statestore in case there is need to recover from a failure. I see we save state upon invoking the actor method and we purge it once the methodname is the purgeWorkflowState. I see the state gets filled with the activity request history event `ar.HistoryEvent` so it _could_ be empty after [decoding here](https://github.com/dapr/dapr/blob/master/pkg/runtime/wfengine/backends/actors/activity_actor.go#L103-L120) if the actor req data was empty to begin with. Im not familiar if this could occur and why...

For the case where it might be possible to have nil activity state, should we terminate or continue along (as we did before [this PR](https://github.com/dapr/dapr/pull/7949))?

cicoyle: actually, I think `GetOrchestrationRuntimeState` in backend.go might cause the data to be empty bc I dont see it populating the data field in the req along the way and the data is used for the activity actor state. So I think it is valid and we should not return. I changed the return to a log, but will change it to debug now.

pkg/runtime/wfengine/backends/actors/backend.go
cicoyle: fair point, I've changed it to this: `actor runtime has not been configured. Ensure actor backend is running, either Placement service or Scheduler service`. I dont want to bubble up which is being used bc then that changes the code to know which system is being used, but its clear now for users to check them. Rather then having users spend time to figure out they need to check placement or scheduler.

pkg/runtime/wfengine/backends/actors/activity_actor.go
mukundansundar: Sounds good.

mukundansundar (commented):
Please add UT/IT for this scenario.

famarting:
After looking at the activity actor code to refresh my memory on how the activity state works...

An activity is created by invoking the "Execute" on the activity actor, that method's logic first persists the activity state and then creates a reminder... so in theory the activity reminder should require the activity state to exist

in a normal execution the activity state is deleted by the workflow actor when an activity is completed or failed, so having into account all the above the workflow actor is assuming the activity actor already executed the reminder and it wont execute it again.

So IMO if the reminder for an activity is executed after its state has been deleted that is a bug, not the fact that the state is missing. The bug is the fact that the reminder got executed.

@cgillum do you know about this part of the codebase? wdyt?

cgillum:
@famarting yes, the fact that the reminder exists without any activity data *might* be a bug. Another possibility is that a workflow was terminated and purged while the activity was running, the activity failed to complete (for whatever reason - maybe because the workflow actor is no longer valid) and now the timer is firing again to "retry", but the state has already been purged.

Regardless of how we got into this state, I think the right thing to do is exit gracefully and make sure that the reminder is properly deleted.

tests/integration/suite/daprd/workflow/backend/singular/activityactorstate.go
```diff
@@ -0,0 +1,343 @@
...
+				queryCtx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+				defer cancel()
+				s.dbMutex.Lock()
+				_, err := s.db.GetConnection(t).QueryContext(queryCtx, fmt.Sprintf("DELETE FROM %s", s.db.GetTableName(t)))
+				s.dbMutex.Unlock()
+				require.NoError(t, err)
+				fmt.Printf("cassie deelted from table")
+				break
+
+				// tried deleting all records
```
cicoyle, line 190: I tried deleting everything from the table both from inside the activity func here and below as well, but still dont see the log about no activity state found, meaning it was always able to get the state in my test... Can this PR be accepted without a test since this is reverting a change in behavior that also had no test? 

Not sure if @JoshVanL you wanna take another stab at reproducing this issue? I thought my issue was due to the state being cached bc I see that is a field that can be set, but even when I manually changed it to not cache for the test runs I still was not seeing the no activity state found error.

cicoyle:
Going to close this to open a PR to the release branch since there were a ton of commits showing up once I changed the target branch

PR #8101, opened 2024-09-13
fix(wf): update resp code on wf start

sicoyle:
# Description

In cases like this where users are trying to start a workflow that already exists, they should see a 400 error code to indicate that this is a "bad" request that they are trying to perform. A 500 response code is extreme here in the case that this is a client side error.

```
{"errorCode":"ERR_START_WORKFLOW","message":"error starting workflow 'SimpleWorkflow': unable to start workflow: failed to start orchestration: error from internal actor: an active workflow with ID 'SimpleWorkflow' already exists"}
```

where would be the best place to add tests for this change? @cicoyle @JoshVanL

## Issue reference

## Checklist

- [ ] Code compiles correctly
- [ ] Created/updated tests
- [ ] Unit tests passing


cicoyle:
You could probably add to this test: https://github.com/dapr/dapr/blob/master/tests/integration/suite/daprd/workflow/basic.go#L269

JoshVanL:
> You could probably add to this test: https://github.com/dapr/dapr/blob/master/tests/integration/suite/daprd/workflow/basic.go#L269

Please can we add a new `tests/integration/suite/daprd/workflow/errors.go` file?

PR #8105, opened 2024-09-15
Placement conn fix

elena-kolevska:
This  PR addresses daprd not being able to connect to placement on k8s after placement leader failover.


artursouza (approved):

JoshVanL:
Is there an integration test we can add for this?

elena-kolevska:
> Is there an integration test we can add for this?
Unfortunately, the issue is only reproducible on k8s so an integration test wouldn't have helped here. 
As @artursouza suggested, we should probably start thinking about investing in chaos testing.

PR #8110, opened 2024-09-16
Updates the 1.14.3 release notes with a warning

elena-kolevska:


docs/release_notes/v1.14.3.md
```diff
@@ -1,5 +1,15 @@
 # Dapr 1.14.3
 
+⚠️ When upgrading from Dapr versions `v1.14.0`, `v1.14.1`, or `v1.14.2` to `v1.14.3`, the upgrade will fail with the following error:
```
msfussell, line 3:
```suggestion
⚠️ When upgrading from Dapr versions `v1.14.0`, `v1.14.1`, or `v1.14.2` to `v1.14.3`, the upgrade fails with the following error:
```

docs/release_notes/v1.14.3.md
```diff
@@ -1,5 +1,15 @@
...
+
+```
+Error: UPGRADE FAILED: cannot patch "dapr-scheduler-server" with kind StatefulSet: StatefulSet.apps "dapr-scheduler-server" is invalid: spec: Forbidden: updates to statefulset spec for fields other than 'replicas', 'ordinals', 'template', 'updateStrategy', 'persistentVolumeClaimRetentionPolicy' and 'minReadySeconds' are forbidden
+```
+
+The workaround for the user is to Delete the Scheduler StatefulSet before re-running the upgrade with the new storage size, but we recommend skipping this version and upgrading to `v1.14.4` instead.
```
msfussell, line 9:
```suggestion
The workaround for the user is to delete the Scheduler StatefulSet before re-running the upgrade with the new storage size. This issue will be addressed in the v1.14 releases and it is recommend to wait for this release.
```
This sentence needs to tell you what the storage size should be. Put a link to doc below.

JoshVanL (approved):

msfussell (commented):
Also this sentence needs fixing
"Creating a reasonable number of Jobs or Actor Reminders (also through workflows) with the Scheduler, would case the embedded Scheduler to run out of memory."

Should be
"Creating a significant number of jobs or Actor reminders (including with workflows) with the Scheduler, causes the embedded Scheduler to run out of memory."

docs/release_notes/v1.14.3.md
elena-kolevska: There already is a fix for the issue in 1.14.4-rc.1 (we'll be releasing the final version asap), so maybe we can just say `but we recommend skipping this version and upgrading to `v1.14.4` or later instead.`?
The [1.14.4 release notes](https://github.com/dapr/dapr/blob/release-1.14/docs/release_notes/v1.14.4.md) cover the issue in more detail, and the docs were updated as well, we can probably just link to them.

PR #8123, opened 2024-09-18
Scheduler Client: Make client connect a runner

JoshVanL:
There is currently a race condition whereby daprd and the jobs APIs are available, even though the scheduler clients are not ready yet.

Moves the runtime Scheduler Clients into a runner which is executed in the background, along with all other background runnables. This allows daprd to correctly signal ready on the readiness endpoint once all Scheduler clients have connected (or always if no addresses were provided).

Removes the `NoInit` client, as it is no longer needed as it is handled by the main Client.

Client getters now give a context to `Next` and `All` to wait for the clients to connect.


yaron2:
Linter

pkg/actors/reminders/scheduler.go
```diff
@@ -124,10 +122,12 @@ func (s *scheduler) CreateReminder(ctx context.Context, reminder *internal.Creat
...
-	_, err = s.clients.Next().ScheduleJob(ctx, internalScheduleJobReq,
-		retry.WithMax(3),
-		retry.WithPerRetryTimeout(time.Second/2),
-	)
+	client, err := s.clients.Next(ctx)
+	if err != nil {
+		return fmt.Errorf("error getting scheduler client: %w", err)
+	}
+
+	_, err = client.ScheduleJob(ctx, internalScheduleJobReq)
```
cicoyle, line 130: do we want the `grpc.WaitForReady(true)` here as well?

JoshVanL: No it isn’t necessary

cicoyle (approved):
lgtm

yaron2:
the go-routine validation fails, need to look into that

JoshVanL:
@yaron2 the number of go routines has increased from 36 to 38- this is acceptable.

yaron2:
> @yaron2 the number of go routines has increased from 36 to 38- this is acceptable.

Can you change the threshold in this PR? This validation is a required step in releases

yaron2:
> > @yaron2 the number of go routines has increased from 36 to 38- this is acceptable.
> 
> Can you change the threshold in this PR? This validation is a required step in releases

Nevermind, the baseline is based on latest branch sidecar utilization

JoshVanL:
@yaron2 It's based on p-value so there isn't anything to tune for go routines. It compares the master branch with the PR branch.
`dapr-standalone-validation` isn't a required step in our release pipeline.

PR #8134, opened 2024-09-23
Feat Scheduler Metrics

cicoyle:
Added Metrics for the Scheduler Service along with integration tests for each:

- dapr_scheduler_sidecars_connected
- dapr_scheduler_jobs_created_total
- dapr_scheduler_jobs_triggered_total
- dapr_scheduler_trigger_latency


tests/integration/suite/actors/reminders/scheduler/metrics.go
```diff
@@ -0,0 +1,210 @@
...
+	m.etcdPort = port2
+	clientPorts := []string{
+		"scheduler-0=" + strconv.Itoa(m.etcdPort),
+	}
+	m.scheduler = scheduler.New(t,
+		scheduler.WithID("scheduler-0"),
+		scheduler.WithInitialCluster(fmt.Sprintf("scheduler-0=http://localhost:%d", port1)),
+		scheduler.WithInitialClusterPorts(port1),
+		scheduler.WithEtcdClientPorts(clientPorts),
+	)
```
JoshVanL, line 83:
```suggestion
m.scheduler = scheduler.New(t)
```

tests/integration/suite/actors/reminders/scheduler/metrics.go
```diff
@@ -0,0 +1,210 @@
...
+	configFile := filepath.Join(t.TempDir(), "config.yaml")
+	require.NoError(t, os.WriteFile(configFile, []byte(`
+apiVersion: dapr.io/v1alpha1
+kind: Configuration
+metadata:
+  name: schedulerreminders
+spec:
+  features:
+  - name: SchedulerReminders
+    enabled: true`), 0o600))
```
JoshVanL, line 69:
```suggestion
```

tests/integration/suite/actors/reminders/scheduler/metrics.go
```diff
@@ -0,0 +1,210 @@
...
+
+	m.daprd = daprd.New(t,
+		daprd.WithConfigs(configFile),
+		daprd.WithInMemoryActorStateStore("mystore"),
+		daprd.WithPlacementAddresses(m.place.Address()),
+		daprd.WithSchedulerAddresses(m.scheduler.Address()),
+		daprd.WithAppPort(app.Port()),
+	)
+
+	fp.Free(t)
```
JoshVanL, line 103:
```suggestion
```

tests/integration/suite/actors/reminders/scheduler/metrics.go
```diff
@@ -0,0 +1,210 @@
...
+		framework.WithProcesses(app, m.scheduler, m.place, m.daprd),
+	}
+}
+
+func (m *metrics) Run(t *testing.T, ctx context.Context) {
+	m.scheduler.WaitUntilRunning(t, ctx)
+	m.place.WaitUntilRunning(t, ctx)
+	m.daprd.WaitUntilRunning(t, ctx)
+
+	frameworkClient := frameworkclient.HTTP(t)
```
JoshVanL, line 114:
```suggestion
	httpclient := client.HTTP(t)
```

tests/integration/suite/actors/reminders/scheduler/metrics.go
```diff
@@ -0,0 +1,210 @@
...
+
+	client := m.daprd.GRPCClient(t, ctx)
+
+	etcdClient := clients.Etcd(t, clientv3.Config{
+		Endpoints:   []string{fmt.Sprintf("localhost:%d", m.etcdPort)},
+		DialTimeout: 5 * time.Second,
+	})
```
JoshVanL, line 121:
```suggestion
	etcdClient :=  m.scheduler.ETCDClient(t, ctx)
```

Add the following to `/framework/process/scheduler/scheduler.go`

```go
func (s *Scheduler) ETCDClient(t *testing.T) *clientv3.Client {
       t.Helper()

       client, err := clientv3.New(clientv3.Config{
               Endpoints:   []string{fmt.Sprintf("127.0.0.1:%s", s.EtcdClientPort())},
               DialTimeout: 40 * time.Second,
       })
       require.NoError(t, err)

       t.Cleanup(func() {
               require.NoError(t, client.Close())
       })

       return client
}
```

tests/integration/suite/actors/reminders/scheduler/metrics.go
```diff
@@ -0,0 +1,210 @@
...
+
+	// Use "path/filepath" import, it is using OS specific path separator unlike "path"
+	etcdKeysPrefix := filepath.Join("dapr", "jobs")
```
JoshVanL, line 124: @cicoyle I think we should consider this a bug in the etcd-cron library implementation- `v1.15` is likely the last opportunity we have the change this for windows. Should be do this?

The path should always be `dapr/jobs` imo.

tests/integration/suite/actors/reminders/scheduler/metrics.go
```diff
@@ -0,0 +1,210 @@
...
+
+	assert.EventuallyWithT(t, func(c *assert.CollectT) {
+		keys, rerr := etcdClient.ListAllKeys(ctx, etcdKeysPrefix)
+		require.NoError(c, rerr)
+		assert.Empty(c, keys)
+	}, time.Second*10, 10*time.Millisecond)
```
JoshVanL, line 130:
```suggestion
		resp, err := etcdClient.Get(ctx, "dapr/jobs", clientv3.WithPrefix())
		require.NoError(t, err)
		assert.Empty(t, resp.KVs)
```

tests/integration/suite/actors/reminders/scheduler/metrics.go
```diff
@@ -0,0 +1,210 @@
...
+			foundMetric = true
+			break
+		}
+	}
+	if expectedCount > 0 {
+		assert.True(t, foundMetric, "Expected metric %s not found", expectedMetric)
+	} else {
+		assert.False(t, foundMetric)
+	}
+}
```
JoshVanL, line 210: We should move the metrics call to the Scheduler process `func (s *Scheduler) Metrics(t *testing.T, ctx context.Context) map[string]float64`.

This is what we are doing for daprd https://github.com/dapr/dapr/blob/b26a1951c482b6754d7486c0a1ec0b9999eb99a0/tests/integration/framework/process/daprd/daprd.go#L328

tests/integration/suite/actors/reminders/scheduler/metrics.go
```diff
@@ -0,0 +1,210 @@
...
+	etcdKeysPrefix := filepath.Join("dapr", "jobs")
+
+	assert.EventuallyWithT(t, func(c *assert.CollectT) {
+		keys, rerr := etcdClient.ListAllKeys(ctx, etcdKeysPrefix)
+		require.NoError(c, rerr)
+		assert.Empty(c, keys)
+	}, time.Second*10, 10*time.Millisecond)
+
+	// assert false, since count is 0 here
+	m.assertMetricExists(t, ctx, frameworkClient, "dapr_scheduler_jobs_created_total", 0)
```
JoshVanL, line 133:
```suggestion
	assert.Equal(t, 0, m.scheduler.Metrics(t, ctx)["dapr_scheduler_jobs_created_total"])
```

tests/integration/suite/actors/reminders/scheduler/metrics.go
```diff
@@ -0,0 +1,210 @@
...
+		DueTime:   "0s",
+		Period:    "1s",
+	})
+	require.NoError(t, err)
+
+	assert.EventuallyWithT(t, func(c *assert.CollectT) {
+		keys, rerr := etcdClient.ListAllKeys(ctx, etcdKeysPrefix)
+		require.NoError(c, rerr)
+		assert.Len(c, keys, 1)
+	}, time.Second*10, 10*time.Millisecond)
```
JoshVanL, line 155:
```suggestion
		resp, err := etcdClient.Get(ctx, "dapr/jobs", clientv3.WithPrefix())
		require.NoError(t, err)
		assert.Len(t, resp.KVs, 1)
```

tests/integration/suite/actors/reminders/scheduler/metrics.go
```diff
@@ -0,0 +1,210 @@
...
+
+	// true, since count is 1
+	m.assertMetricExists(t, ctx, frameworkClient, "dapr_scheduler_jobs_created_total", 1)
```
JoshVanL, line 158:
```suggestion
	assert.Equal(t, 1, m.scheduler.Metrics(t, ctx)["dapr_scheduler_jobs_created_total"])
```

tests/integration/suite/actors/reminders/scheduler/metrics.go
```diff
@@ -0,0 +1,210 @@
...
+
+	assert.EventuallyWithT(t, func(c *assert.CollectT) {
+		assert.GreaterOrEqual(c, m.triggered.Load(), int64(1))
+	}, 30*time.Second, 10*time.Millisecond, fmt.Sprintf("failed to wait for 'triggered' to be greatrer or equal 1, actual value %d", m.triggered.Load()))
```
JoshVanL, line 162:
```suggestion
```

tests/integration/suite/scheduler/metrics/basic.go
```diff
@@ -0,0 +1,187 @@
...
+					Target: &schedulerv1.JobTargetMetadata{
+						Type: new(schedulerv1.JobTargetMetadata_Job),
+					},
+				},
+			}
+
+			_, err := client.ScheduleJob(ctx, req)
+			require.NoError(t, err)
+
+			assert.True(t, b.etcdHasJob(t, ctx, name))
```
JoshVanL, line 123: Either use full path on the ETCD client Get for the job, or use the Scheduler gRPC client to get the Job.

```suggestion
```

tests/integration/suite/scheduler/metrics/basic.go
```diff
@@ -0,0 +1,187 @@
+/*
+Copyright 2024 The Dapr Authors
```
JoshVanL, line 2: Update with suggestions from other file.

tests/integration/suite/scheduler/metrics/daprconnections.go
```diff
@@ -0,0 +1,129 @@
+/*
```
JoshVanL, line 1: Same here.

tests/integration/suite/scheduler/metrics/daprconnections.go
```diff
@@ -0,0 +1,129 @@
...
+
+func init() {
+	suite.Register(new(daprconnections))
+}
+
+type daprconnections struct {
+	scheduler *scheduler.Scheduler
+	daprd     *daprd.Daprd
+
+	connectionlogline *logline.LogLine
```
JoshVanL, line 47: Not a fan of using the log line here for testing as it is quite fragile. I think we can improve the test by running 3 daprds in `Run`, and doing a `Cleanup` of each. We should test that we observe the `dapr_scheduler_sidecars_connected_total` metric return `0, 1, 2, 3, 2, 1, 0`

tests/integration/suite/scheduler/metrics/jobstriggered.go
```diff
@@ -0,0 +1,277 @@
+/*
```
JoshVanL, line 1: Same here.

JoshVanL (changes requested):

tests/integration/suite/actors/reminders/scheduler/metrics.go
```diff
@@ -0,0 +1,210 @@
...
+			m.triggered.Add(1)
+		}),
+		app.WithHandlerFunc("/actors/myactortype/myactorid/method/foo", func(http.ResponseWriter, *http.Request) {}),
+		app.WithConfig(`{"entities": ["myactortype"]}`),
+	)
+
+	m.place = placement.New(t)
+
+	m.daprd = daprd.New(t,
+		daprd.WithConfigs(configFile),
```
JoshVanL, line 96:
```suggestion
		daprd.WithConfigManifests(t, `
apiVersion: dapr.io/v1alpha1
kind: Configuration
metadata:
  name: schedulerreminders
spec:
  features:
  - name: SchedulerReminders
    enabled: false
`),
```

tests/integration/suite/actors/reminders/scheduler/metrics.go
cicoyle: agreed. Ill add it to the list to fix 👍🏻

tests/integration/suite/scheduler/metrics/basic.go
cicoyle: I think this is done?

tests/integration/suite/scheduler/metrics/basic.go
cicoyle: I think this is done

tests/integration/suite/scheduler/metrics/daprconnections.go
cicoyle: Not sure what this is referring to

tests/integration/suite/scheduler/metrics/jobstriggered.go
cicoyle: Not sure what is meant here

tests/integration/suite/scheduler/metrics/daprconnections.go
cicoyle: ok I wouldn't say its fragile, but I see what you are saying. I can work to make this better per your suggestion.

Can you re-send/share the comment you made about using a metric map. I vaguely remember seeing a comment from you about it, but am having a hard time finding it now.

tests/integration/suite/actors/reminders/scheduler/metrics.go
cicoyle: that really cleans things up. Wish I saw this earlier!

tests/integration/suite/actors/reminders/scheduler/metrics.go
cicoyle: I added this, but then realized the test was failing and forgot I added this, bc this makes the schedulerreminders false, it needed to be true. that was fun to realize I changed the value... the suggestion was added :)

tests/integration/suite/scheduler/metrics/daprconnections.go
cicoyle: I noticed with the `Cleanup()` they werent shutting down in time, so what I did was test the metric by adding the 3 daprs and then just clean them up but not test the metric count going down.

tests/integration/framework/client/etcd.go
```diff
@@ -64,3 +64,17 @@ func (c *EtcdClient) ListAllKeys(ctx context.Context, prefix string) ([]string,
...
+		return nil, err
+	}
+
+	keys := make([]string, len(resp.Kvs))
+	for i, kv := range resp.Kvs {
+		keys[i] = string(kv.Key)
+	}
+
+	return keys, nil
+}
```
JoshVanL, line 80:
```suggestion
func (c *EtcdClient) Get(t *testing.T, ctx context.Context, prefix string, opts ...clientv3.OpOption) error {
	t.Helper()
	resp, err := c.client.Get(ctx, prefix, opts...)
	require.NoError(t, err)

	keys := make([]string, len(resp.Kvs))
	for i, kv := range resp.Kvs {
		keys[i] = string(kv.Key)
	}

	return keys
}
```

tests/integration/framework/process/scheduler/scheduler.go
```diff
@@ -215,6 +219,17 @@ func (s *Scheduler) DataDir() string {
 	return s.dataDir
 }
 
+func (s *Scheduler) ETCDClient(t *testing.T) *client.EtcdClient {
+	t.Helper()
+
+	client := client.Etcd(t, clientv3.Config{
+		Endpoints:   []string{"127.0.0.1:" + s.EtcdClientPort()},
+		DialTimeout: 40 * time.Second, //nolint:mnd
```
JoshVanL, line 227: Linter is being removed

```suggestion
		DialTimeout: 40 * time.Second,
```

tests/integration/framework/process/scheduler/scheduler.go
```diff
@@ -215,6 +219,17 @@ func (s *Scheduler) DataDir() string {
 	return s.dataDir
 }
 
+func (s *Scheduler) ETCDClient(t *testing.T) *client.EtcdClient {
+	t.Helper()
+
+	client := client.Etcd(t, clientv3.Config{
```
JoshVanL, line 225:
```suggestion
	return client.Etcd(t, clientv3.Config{
```

tests/integration/suite/actors/reminders/scheduler/remove.go
```diff
@@ -140,14 +133,14 @@ func (r *remove) Run(t *testing.T, ctx context.Context) {
...
-		keys, rerr := etcdClient.ListAllKeys(ctx, etcdKeysPrefix)
+		keys, rerr := r.scheduler.ETCDClient(t).ListAllKeys(ctx, etcdKeysPrefix)
 		require.NoError(c, rerr)
 		assert.Len(c, keys, 1)
 	}, time.Second*10, 10*time.Millisecond)
 
 	assert.EventuallyWithT(t, func(c *assert.CollectT) {
 		assert.GreaterOrEqual(c, r.triggered.Load(), int64(1))
-	}, 30*time.Second, 10*time.Millisecond, fmt.Sprintf("failed to wait for 'triggered' to be greatrer or equal 1, actual value %d", r.triggered.Load()))
+	}, 30*time.Second, 10*time.Millisecond, fmt.Sprintf("failed to wait for 'triggered' to be greater or equal 1, actual value %d", r.triggered.Load())) //nolint:mnd
```
JoshVanL, line 143:
```suggestion
	}, 30*time.Second, 10*time.Millisecond, fmt.Sprintf("failed to wait for 'triggered' to be greater or equal 1, actual value %d", r.triggered.Load()))
```

tests/integration/suite/daprd/hotreload/operator/informer/scopes/components.go
```diff
@@ -139,7 +139,7 @@ func (c *components) Run(t *testing.T, ctx context.Context) {
 	c.kubeapi.Informer().Modify(t, &comp)
 	require.EventuallyWithT(t, func(ct *assert.CollectT) {
 		assert.Len(ct, c.daprd.GetMetaRegisteredComponents(ct, ctx), 1)
-	}, time.Second*10, time.Millisecond*10)
+	}, time.Second*15, time.Millisecond*10) //nolint:mnd
```
JoshVanL, line 142:
```suggestion
	}, time.Second*15, time.Millisecond*10)
```

tests/integration/suite/scheduler/metrics/actors.go
```diff
@@ -0,0 +1,129 @@
...
+
+	metrics := a.scheduler.Metrics(t, ctx)
+	assert.Equal(t, 0, int(metrics["dapr_scheduler_jobs_created_total"]))
+
+	_, err := grpcClient.InvokeActor(ctx, &runtimev1pb.InvokeActorRequest{
+		ActorType: "myactortype",
+		ActorId:   "myactorid",
+		Method:    "foo",
+	})
+	require.NoError(t, err)
```
JoshVanL, line 106:
```suggestion
```

tests/integration/suite/scheduler/metrics/daprconnections.go
```diff
@@ -0,0 +1,106 @@
...
+
+func (c *daprconnections) Setup(t *testing.T) []framework.Option {
+	c.scheduler = scheduler.New(t)
+	srv := app.New(t)
+
+	c.daprdA = daprd.New(t,
+		daprd.WithSchedulerAddresses(c.scheduler.Address()),
+		daprd.WithAppProtocol("grpc"),
+		daprd.WithAppPort(srv.Port(t)),
+		daprd.WithAppID("A"),
```
JoshVanL, line 49: Please try not to use constant app IDs because it causes issues with cache mDNS entries. Use `daprd.AppID()` to get the UUID which is generated in the framework.

JoshVanL (changes requested):

tests/integration/framework/process/scheduler/scheduler.go
cicoyle: this fixes a linter issue. anywhere there is a number it throws a lint err

tests/integration/suite/actors/reminders/scheduler/remove.go
cicoyle: fixes linter

tests/integration/suite/scheduler/metrics/daprconnections.go
cicoyle: since Im not using the logline htis can be removed, I just forgot to rm it earlier. Its removed now

elena-kolevska (commented):
The `sidecarsConnectedTotal` metric seems to be cumulative here, it represents the total number of sidecars that have ever connected to the scheduler server, correct? I think a more useful metric type would be a gauge that we can increase when a sidecar connects and decrease when it disconnects. That way we'll have an accurate representation of currently connected sidecars.

cicoyle:
> The `sidecarsConnectedTotal` metric seems to be cumulative here, it represents the total number of sidecars that have ever connected to the scheduler server, correct? I think a more useful metric type would be a gauge that we can increase when a sidecar connects and decrease when it disconnects. That way we'll have an accurate representation of currently connected sidecars.

@elena-kolevska I mentioned above, but when I did have that code to manually shutdown and check the metric it was not shutting down in time for the test. Int tests have a ctx timeout of 45s. https://github.com/dapr/dapr/pull/8134#discussion_r1781700161

elena-kolevska:
> @elena-kolevska I mentioned above, but when I did have that code to manually shutdown and check the metric it was not shutting down in time for the test. Int tests have a ctx timeout of 45s. [#8134 (comment)](https://github.com/dapr/dapr/pull/8134#discussion_r1781700161)

This is expected, because we're using the Counter instrument which only counts up. It will never go down to three.
I'm currently working on moving our metrics from OpenCensus to OpenTelemetry and with it, we can use [UpDownCounters](https://opentelemetry.io/docs/specs/otel/metrics/api/#updowncounter). Unfortunately, OpenCensus doesn't provide that option, and we'd have to manually track the number of connected instances and report it both when a sidecar is connected and disconnected.
I know this is a tad bit more code, but I think the metric is not useful otherwise.
I'll put a suggestion in code to show what I mean.

pkg/scheduler/monitoring/metrics.go
```diff
@@ -0,0 +1,98 @@
...
+	triggerLatency = stats.Float64(
+		"scheduler/trigger_latency",
+		"The total time it takes to trigger a job from the scheduler service.",
+		stats.UnitMilliseconds)
+)
+
+// RecordSidecarsConnectedCount records the number of dapr sidecars connected to the scheduler service
+func RecordSidecarsConnectedCount() {
+	stats.RecordWithTags(context.Background(), utils.WithTags(sidecarsConnectedTotal.Name()), sidecarsConnectedTotal.M(1))
+}
```
elena-kolevska, line 50:
```suggestion
func RecordSidecarsConnectedCount(cnt int64) {
	stats.RecordWithTags(context.Background(), utils.WithTags(sidecarsConnectedTotal.Name()), sidecarsConnectedTotal.M(cnt))
}
```
Instead of always setting the value to one and checking how many times it was called later (`utils.NewMeasureView(sidecarsConnectedTotal, []tag.Key{}, view.Count())`), we'll record the actual value we're observing and check the last value (`utils.NewMeasureView(sidecarsConnectedTotal, []tag.Key{}, view.LastValue())`).

pkg/scheduler/monitoring/metrics.go
```diff
@@ -0,0 +1,98 @@
...
+// RecordTriggerDuration records the time it takes to send the job to dapr from the scheduler service
+func RecordTriggerDuration(ns string, appID string, start time.Time) {
+	elapsed := time.Since(start).Milliseconds()
+	stats.RecordWithTags(context.Background(), utils.WithTags(triggerLatency.Name(), ns, appID), triggerLatency.M(float64(elapsed)))
+}
+
+// InitMetrics initialize the scheduler service metrics.
+func InitMetrics() error {
+	err := view.Register(
+		utils.NewMeasureView(sidecarsConnectedTotal, []tag.Key{}, view.Count()),
```
elena-kolevska, line 91:
```suggestion
		utils.NewMeasureView(sidecarsConnectedTotal, []tag.Key{}, view.LastValue()),
```

pkg/scheduler/server/api.go
```diff
@@ -145,6 +146,7 @@ func (s *Server) WatchJobs(stream schedulerv1pb.Scheduler_WatchJobsServer) error
 	}
 
 	s.connectionPool.Add(req.GetInitial(), stream)
+	monitoring.RecordSidecarsConnectedCount()
```
elena-kolevska, line 149:
```suggestion
	s.cnt.Add(1)
	monitoring.RecordSidecarsConnectedCount(s.cnt.Load())

	defer func() {
		s.cnt.Add(-1)
		monitoring.RecordSidecarsConnectedCount(s.cnt.Load())
	}()
```
Assuming `cnt atomic.Int64`

elena-kolevska (changes requested):
These changes should get us to the current instead of cumulative number of connected sidecars.

tests/integration/suite/scheduler/metrics/daprconnections.go
cicoyle: I got this working

cicoyle:
> > @elena-kolevska I mentioned above, but when I did have that code to manually shutdown and check the metric it was not shutting down in time for the test. Int tests have a ctx timeout of 45s. [#8134 (comment)](https://github.com/dapr/dapr/pull/8134#discussion_r1781700161)
> 
> This is expected, because we're using the Counter instrument which only counts up. It will never go down to three. I'm currently working on moving our metrics from OpenCensus to OpenTelemetry and with it, we can use [UpDownCounters](https://opentelemetry.io/docs/specs/otel/metrics/api/#updowncounter). Unfortunately, OpenCensus doesn't provide that option, and we'd have to manually track the number of connected instances and report it both when a sidecar is connected and disconnected. I know this is a tad bit more code, but I think the metric is not useful otherwise. I'll put a suggestion in code to show what I mean.

I figured it out, then came back to tell you I got it to see all your comments. Please see latest

elena-kolevska:
> I figured it out, then came back to tell you I got it, then saw all your comments. Please see latest

Ha! I know the feeling :)
I think it looks great now, I'd only use `sidecarConnectionCount atomic.Int64` to make it a bit more readable, but other than that - it LGTM!

elena-kolevska (approved):

yaron2 (approved):

docs/development/dapr-metrics.md
```diff
@@ -54,6 +55,15 @@ Dapr uses prometheus process and go collectors by default.
 * dapr_sentry_issuercert_changed_total: The number of issuer cert updates, when issuer cert or key is changed
 * dapr_sentry_issuercert_expiry_timestamp: The unix timestamp, in seconds, when issuer/root cert will expire.
 
+## Dapr Scheduler metrics
+
+[monitoring metrics](../../pkg/scheduler/monitoring/metrics.go)
+
+* dapr_scheduler_sidecars_connected_total: The total number of dapr sidecars connected to the scheduler service.
+* dapr_scheduler_jobs_created_total: The total number of jobs scheduled.
+* dapr_scheduler_jobs_triggered_total: The total number of successfully triggered jobs.
```
JoshVanL, line 64: How will this behave once we have added failure policy- I guess we will need to add labels for `SUCCESS`, `FAILED`, and `UNDELIVERED`.

`UNDELIVERED` will be tricky as we might need to track when they do get success/failure triggered so this gauge can decrease.

pkg/scheduler/server/api.go
```diff
@@ -181,7 +185,9 @@ func (s *Server) triggerJob(ctx context.Context, req *api.TriggerRequest) bool {
 		// another long running go routine that accepts this job on a channel
 		log.Errorf("Error sending job to connection stream: %s", err)
 	}
+	monitoring.RecordTriggerDuration(meta.GetNamespace(), meta.GetAppId(), now)
```
JoshVanL, line 188: We shouldn't have app ID as a label as this produces high cardinality. Once could also say the same for namespace, though that it probably fine.

Users interested in app ID specific metrics should be getting that from daprd.

tests/integration/framework/client/etcd.go
```diff
@@ -64,3 +64,16 @@ func (c *EtcdClient) ListAllKeys(ctx context.Context, prefix string) ([]string,
 
 	return r, nil
 }
+
+func (c *EtcdClient) Get(t *testing.T, ctx context.Context, prefix string, opts ...clientv3.OpOption) []string {
+	t.Helper()
+	resp, err := c.client.Get(ctx, prefix, opts...)
```
JoshVanL, line 70: `clientv3.WithPrefix` ? Or should the parameter not be `prefix` ?

tests/integration/framework/process/scheduler/scheduler.go
JoshVanL: `mnd` has definitely been disabled though so we shouldn't need this.

tests/integration/suite/daprd/hotreload/operator/informer/scopes/components.go
JoshVanL: Same as above.

tests/integration/suite/scheduler/metrics/basic.go
JoshVanL: Please can we not just use the scheduler client to get the job so everything is fully qualified? I don't like the fact we are doing suffix checks with `||` to check for existence.

tests/integration/suite/scheduler/metrics/daprconnections.go
```diff
@@ -0,0 +1,121 @@
...
+		// 3 sidecars connected
+		c.daprdC.Run(t, ctx)
+		c.daprdC.WaitUntilRunning(t, ctx)
+		assert.EventuallyWithT(t, func(ct *assert.CollectT) {
+			metrics = c.scheduler.Metrics(t, ctx)
+			assert.Equal(ct, 3, int(metrics["dapr_scheduler_sidecars_connected"])) //nolint:mnd
+		}, 15*time.Second, 10*time.Millisecond, "daprdC sidecar didn't connect to Scheduler in time") //nolint:mnd
+
+		// 2 sidecars connected
+		c.daprdA.Cleanup(t)
```
JoshVanL, line 101: Please duplicate all daprd `Cleanup` function calls to just after the daprd `.Run` with a `t.Cleanup`. This ensures that daprds are _always_ cleandup, regardless of test failure. You'll need to add a `sync.Once` wrapper over the `exec.Cleanup` in the daprd framework process to ensure exec is attempted to kill twice.

```go
t.Cleanup(func() {c.daprdA.Cleanup(t)}())
```

JoshVanL (changes requested):

docs/development/dapr-metrics.md
cicoyle: I think adding labels will be the easiest. Since that change is not in, Id like to not wait or add a pre-req to this PR since its been open a while already.

tests/integration/framework/client/etcd.go
cicoyle: I added it as a default option

pkg/scheduler/server/api.go
cicoyle: I can rm the appID

pkg/scheduler/monitoring/metrics.go
```diff
@@ -0,0 +1,102 @@
...
+		jobType = "unknown"
+	}
+
+	stats.RecordWithTags(context.Background(), utils.WithTags(jobsTriggeredTotal.Name(), jobType), jobsTriggeredTotal.M(1))
+}
+
+// RecordTriggerDuration records the time it takes to send the job to dapr from the scheduler service
+func RecordTriggerDuration(ns string, start time.Time) {
+	elapsed := time.Since(start).Milliseconds()
+	stats.RecordWithTags(context.Background(), utils.WithTags(triggerLatency.Name(), ns), triggerLatency.M(float64(elapsed)))
```
JoshVanL, line 89: I'm still not sure about having the namespace label here, as it means cardinality will grow indefinitely, unless we remove the metric once the ns has been deleted. Wdyt?

JoshVanL (changes requested):

pkg/scheduler/monitoring/metrics.go
cicoyle: fair, I must have missed this. Its rmed now.

JoshVanL (approved):

PR #8138, opened 2024-09-24
Fix Perf Infra: up k8s version

cicoyle:
Fix Perf Infra: up k8s version

### DIFF
--- a/.github/workflows/dapr-perf-components.yml
+++ b/.github/workflows/dapr-perf-components.yml
@@ -34,7 +34,7 @@ env:
   # Configure proxy for Go modules
   GOPROXY: https://proxy.golang.org
   # Version of kubectl
-  KUBECTLVER: "v1.27.6"
+  KUBECTLVER: "v1.30.0"
   # Version of Helm
   HELMVER: "v3.10.0"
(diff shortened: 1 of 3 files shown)


elena-kolevska (approved):
These changes seem to only be updating the kubectl version. Based on this command in the workflow file
```
az deployment group create \
      --resource-group "Dapr-Perf-daprprfd9b5c79d49" \
      --template-file ./tests/test-infra/azure-aks.bicep
```
we should have the k8s version in the `./tests/test-infra/azure-aks.bicep`, and we already updated that to 1.30 a while ago.

I don't know what were the failed runs, but based on the latest ones - the build stage passes.
I'm still ok updating the kubectl version though.

elena-kolevska:
Ah, I just ran into an instance of a failed deploy now: https://github.com/dapr/dapr/actions/runs/10901844797/job/30252456951. This was on the 1.14 though, so we can cherrypick this commit into 1.14 to fix it: https://github.com/dapr/dapr/commit/3ce38aff38fcf99402cfe621a784b05cb8f41320. I can send a PR.

JoshVanL (approved):

PR #8146, opened 2024-09-26
Linter: Update implementation code

JoshVanL:
Update Dapr implementation code to adhere to new linter rules. All test files have been purposefully omitted to ensure that the implementation code is still functionally correct. A follow-up commit will contain the update to all test code, and removal of the "only new changes linted" from the github linter workflow.


.golangci.yml
```diff
@@ -16,28 +16,29 @@ run:
   build-tags:
     - unit
```
mikeee, line 17:
```suggestion
    - unit
    - allcomponents
    - subtlecrypto
```

Missing build tags - as per workflow

.golangci.yml
```diff
@@ -86,9 +84,8 @@ linters-settings:
     # enable or disable analyzers by name
     enable:
       - atomicalign
+      - shaddow
```
mikeee, line 87:
```suggestion
      - shadow
```
sp.

pkg/api/errors/pubsub.go
```diff
@@ -150,7 +150,7 @@ func (p *PubSubMetadataError) TopicEmpty() error {
 	return p.build(
 		codes.InvalidArgument,
 		http.StatusNotFound,
-		fmt.Sprintf("topic is empty in pubsub %s", p.p.name),
+		"topic is empty in pubsub "+p.p.name,
```
mikeee, line 153:
```suggestion
		"topic is empty in pubsub " + p.p.name,
```
I'm surprised gofmt didn't complain about this

mikeee (changes requested):

elena-kolevska (approved):
LGTM, provided that Mike's suggestions are accepted

.golangci.yml
```diff
@@ -16,28 +16,29 @@ run:
...
-  # autogenerated files. If it's not please let us know.
-  # skip-files:
-  #  - ".*\\.my\\.go$"
-  #  - lib/bad.go
-  
 # output configuration options
 output:
   # colored-line-number|line-number|json|tab|checkstyle, default is "colored-line-number"
-  format: tab
+  formats: tab
```
elena-kolevska, line 41:
```suggestion
  formats:
    - format: tab
```

.golangci.yml
```diff
@@ -60,7 +61,7 @@ linters-settings:
     # [deprecated] comma-separated list of pairs of the form pkg:regex
     # the regex is used to ignore names within pkg. (default "fmt:.*").
     # see https://github.com/kisielk/errcheck#the-deprecated-method for details
-    ignore: fmt:.*,io/ioutil:^Read.*
+    exclude-functions: fmt:.*,io/ioutil:^Read.*
```
elena-kolevska, line 64:
```suggestion
    exclude-functions:
      - fmt:.*
      - io/ioutil:^Read.*
```

.golangci.yml
```diff
@@ -86,9 +84,8 @@ linters-settings:
     # enable or disable analyzers by name
     enable:
       - atomicalign
+      - shaddow
     enable-all: false
-    disable:
-      - shadow
     disable-all: false
   golint:
```
elena-kolevska, line 90: The `golint` linter was removed, so it should be removed from settings. Same with `maligned`

elena-kolevska (changes requested):
The typo `shaddow` made me think there could be other errors on which the tool failed silently, so I did a quick search and found out about the `golangci-lint config verify` command. It brought a few additional errors we need to fix. I put a few of them as suggestions, and here are two more I can't add as comments in code:

Under `linter-settings`:
```
# `check-exported: false` is not  valid anymore, I suggest: 
unused:
    # Mark all exported fields as used.
    exported-fields-are-used: false
```    
```
# We used to have `allow-case-traling-whitespace: true` which is not allowed anymore. We can either delete the config completely, or force a new line as so:
wsl:
  ...
      # If the number of lines in a case block is equal to or lager than this number,
    # the case *must* end white a newline.
    # https://github.com/bombsimon/wsl/blob/master/doc/configuration.md#force-case-trailing-whitespace
    # Default: 0
    force-case-trailing-whitespace: 1
```

JoshVanL:
Thanks @elena-kolevska @mikeee, please take another look!

elena-kolevska (approved):

mikeee (approved):
Lgtm

pkg/api/errors/pubsub.go
elena-kolevska: TIL: gofmt doesn't force spaces around operators if the expression is an argument of a function with more than one argument 🤷‍♀️ 
Ref: https://github.com/golang/go/issues/34426, https://github.com/golang/go/issues/12105

cicoyle (approved):

PR #8157, opened 2024-10-01
Scheduler: ListJobs API

JoshVanL:
Adds the `ListJobs` rpc API to the Scheduler gRPC service. This API returns the either
- the Jobs for the given app ID and namespace, or
- the ActorReminders for a given namespace & actor type, with the option to also scope the list to an actor ID.

```proto
rpc ListJobs(ListJobsRequest) returns (ListJobsResponse) {}
```

This API has been added to allow for migrating ActorReminder from the
"legacy" state store implementation, into Scheduler via daprd. daprd
needs to be able to list ActorReminders in Scheduler to reconcile the
difference and then apply the missing ActorReminders to Scheduler.

Adds `NamedJob` message to `scheduler.proto` so the Job response also
includes the Jobs name.

The `ListResponse` includes the field `has_more` to indicate that there
are more jobs that match the list query, however the reponse size
exceeds Etcd's maximum response size. In the cron library requests for
the maximum number of jobs keys possible, using a [limit of
`0`](https://github.com/diagridio/go-etcd-cron/pull/43/files#diff-5b9b550fff634f1fd5596893d63ceeb6afa6a355427db919a644a21f097b7834R190).

Since Jobs contain payload data, the message for a Jobs List response
can easily exceed the dedault 4MB gRPC message receive size limit. To
manage this, the scheduler Etcd client and daprd scheduler client has
increased the gRPC receive message size limit to `math.MaxInt32`.


cicoyle (dismissed):
Mind fixing the conflicts? Then lgtm

cicoyle (approved):

PR #8161, opened 2024-10-02
proto: updates protoc to 25.4

JoshVanL:


cicoyle (approved):

mikeee (approved):
Lgtm, not a whole load of breaking changes that could bite us in the SDKs I don't think.

mikeee (commented):
I do think the dockerfiles at `docker/` should also be updated alongside the dapr tools install shell script

JoshVanL:
Thanks @mikeee, updated!

yaron2 (approved):

mikeee (approved):
lgtm

dapr/proto/common/v1/README.md
```diff
@@ -5,7 +5,7 @@ This folder is intended for the Common protos amongst the packages in the `dapr/
 ## Proto client generation
 
 Pre-requisites:
-1. Install protoc version: [v4.24.4](https://github.com/protocolbuffers/protobuf/releases/tag/v4.24.4)
+1. Install protoc version: [v4.25.4](https://github.com/protocolbuffers/protobuf/releases/tag/v4.25.4)
```
elena-kolevska, line 8:
```suggestion
1. Install protoc version: [v25.4](https://github.com/protocolbuffers/protobuf/releases/tag/v25.4)
```
Shouldn't this be `v25.4`? From what I understand, the 4.X.X would be the language runtime-specific version, but since Go is not included in the compiler and instead used as a plugin, it probably doesn't make sense for us to refer to 4.X.X here.
Please correct me if my understanding is wrong here, I find protobuf versioning to be quite confusing.

elena-kolevska, line 8:
```suggestion
1. Install protoc version: [v25.4](https://github.com/protocolbuffers/protobuf/releases/tag/v25.4)
```
Shouldn't this be `v25.4`? From what I understand, the 4.X.X would be the language runtime-specific version, but since Go is not included in the compiler and instead used as a plugin, it probably doesn't make sense for us to refer to 4.X.X here.
Please correct me if my understanding is wrong here, I find protobuf versioning to be quite confusing.

.github/workflows/dapr.yml
```diff
@@ -42,7 +42,7 @@ jobs:
         target_arch: ["amd64"]
     env:
       GOLANGCILINT_VER: "v1.61.0"
-      PROTOC_VERSION: "24.4"
+      PROTOC_VERSION: "25.4"
```
elena-kolevska, line 45: Just a question: How do we decide which version we're updating to? The latest stable version seems to be 28.2, with an RC1 for 29 already out there.

dapr/proto/common/v1/README.md
JoshVanL: To be quite honest, I'm not sure.

.github/workflows/dapr.yml
JoshVanL: I assumed because we want to keep the same major version (`4`).

dapr/proto/common/v1/README.md
mikeee: I agree with this assessment although the tags point and align at the same commit references thus making a minimal difference, we should move towards a clearer version/tag consistent with what we already have.

mikeee:
Out of curiousity, was there a particular reason for choosing 25.4 over the latest 25.5?

https://github.com/protocolbuffers/protobuf/releases/tag/v25.5

JoshVanL:
> Out of curiousity, was there a particular reason for choosing 25.4 over the latest 25.5?
> 
> https://github.com/protocolbuffers/protobuf/releases/tag/v25.5

It's not in latest nixpkgs yet

cicoyle (approved):

elena-kolevska (approved):

PR #8186, opened 2024-10-10
Add role support to conversational api

yaron2:
Updates to latest contrib and adds role support to a conversational input message.


dapr/proto/runtime/v1/dapr.proto
```diff
@@ -1297,6 +1297,14 @@ message ConversationAlpha1Request {
   map<string, string> metadata = 5;
 }
 
+message ConversationInput {
+  // The message to send to the llm
+  string message = 1;
+
+  // The role to set for the message
+  optional string role = 2;
```
JoshVanL, line 1305: Im not super familiar in the LLM space when it comes to shared API surface between providers. Is `role` common, or is this a particular implementation detail of one provider, and should be moved to a generic metadata map?

JoshVanL (changes requested):

dapr/proto/runtime/v1/dapr.proto
yaron2: It's common across most LLMs and there's a very common definition too between the major providers (OpenAI, AWS Bedrock, Anthropic) so it's very easy for us to abstract.

JoshVanL: Gotcha, thanks 👍

JoshVanL (approved):

Issue #8189, opened 2024-10-10
Enable PII scrubbing for LLM inputs/outputs


yaron2:
A major challenge that GenAI developers and security teams face is the unintentional sending of PII (personal identifiable information) into an LLM or getting it in response.

This becomes a bigger problem with RAG based operations where data that can be both structured and unstructured is not guaranteed to be sanitized, and also in smaller fit-for-purpose models that undergo fine tuning with custom data.

With the recently added Conversational API, Dapr is in a prime position to serve as secure entrypoint into LLM invocation and can apply PII scrubbing across every underlying model it interacts with.

PR #8191, opened 2024-10-10
Add PII scrubbing to conversational api

yaron2:
Closes https://github.com/dapr/dapr/issues/8189.


dapr/proto/runtime/v1/dapr.proto
```diff
@@ -1295,6 +1295,9 @@ message ConversationAlpha1Request {
 
   // The metadata passing to conversation components.
   map<string, string> metadata = 5;
+
+  // Scrub PII data that comes back from the LLM
+  optional bool piiScrubOutput = 6;
```
JoshVanL, line 1300: Nit: 1. Is `Output` seems superfluous here (?), 2. make first word the verb, 3. Capitalize acronym  

```suggestion
  optional bool scrubPII = 6;
```

dapr/proto/runtime/v1/dapr.proto
```diff
@@ -1295,6 +1295,9 @@ message ConversationAlpha1Request {
 
   // The metadata passing to conversation components.
   map<string, string> metadata = 5;
+
```
JoshVanL, line 1298: Nit: do we want to rename `conversationContext` to `context` since `conversation` is superfluous as it's derived through context (?)

The description is also confusing, and suggests we should use a new field name altogether- such as `session`, `sessionID`,  `chatID` etc.

```
the Id of an existing chat room (like in ChatGPT)
```

dapr/proto/runtime/v1/dapr.proto
```diff
@@ -1303,6 +1306,9 @@ message ConversationInput {
 
   // The role to set for the message
   optional string role = 2;
+
+  // Scrub PII data that goes into the LLM
+  optional bool piiScrubInput = 3;
```
JoshVanL, line 1311:
```suggestion
  optional bool ScrubPII = 3;
```

pkg/api/universal/conversation.go
```diff
@@ -51,9 +53,29 @@ func (a *Universal) ConverseAlpha1(ctx context.Context, req *runtimev1pb.Convers
 		return nil, err
 	}
 
+	scrubber, err := piiscrubber.NewDefaultScrubber()
```
JoshVanL, line 56: Do we want to be caching `scrubber`, or creating it on `universal.New()`. ?

pkg/api/universal/conversation.go
```diff
@@ -51,9 +53,29 @@ func (a *Universal) ConverseAlpha1(ctx context.Context, req *runtimev1pb.Convers
...
+		if i.GetPiiScrubInput() {
+			scrubbed, sErr := scrubber.ScrubTexts([]string{i.GetMessage()})
+			if sErr != nil {
+				sErr = messages.ErrConversationInvoke.WithFormat(req.GetName(), sErr.Error())
+				a.logger.Debug(sErr)
+				return &runtimev1pb.ConversationAlpha1Response{}, sErr
+			}
+
+			msg = scrubbed[0]
+		}
```
JoshVanL, line 75: We are returning on the first error here rather than doing an `errors.Join` of collected errors on scrubed inputs. That's fine, but is a design decision. Would a user of the API expect all errors of the input be returned, or just the first?

pkg/api/universal/conversation.go
```diff
@@ -86,10 +108,23 @@ func (a *Universal) ConverseAlpha1(ctx context.Context, req *runtimev1pb.Convers
...
+			if req.GetPiiScrubOutput() {
+				scrubbed, sErr := scrubber.ScrubTexts([]string{o.Result})
+				if sErr != nil {
+					sErr = messages.ErrConversationInvoke.WithFormat(req.GetName(), sErr.Error())
+					a.logger.Debug(sErr)
+					return &runtimev1pb.ConversationAlpha1Response{}, sErr
+				}
+
+				res = scrubbed[0]
+			}
```
JoshVanL, line 123: Similarly here re: first error vs appended list.

tests/integration/suite/daprd/conversation/grpc/scrubpii.go
```diff
@@ -0,0 +1,151 @@
...
+package http
+
+import (
+	"context"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	rtv1 "github.com/dapr/dapr/pkg/proto/runtime/v1"
+
```
JoshVanL, line 23:
```suggestion

```

pkg/api/universal/conversation.go
```diff
@@ -16,6 +16,8 @@ package universal
 import (
 	"context"
 
+	piiscrubber "github.com/aavaz-ai/pii-scrubber"
```
JoshVanL, line 19: Maybe this is a fine pii scrubber default, but what is the long term plan for users to pick/choose their own scrubber?

tests/integration/suite/daprd/conversation/grpc/scrubpii.go
```diff
@@ -0,0 +1,151 @@
...
+				{
+					Message:       "well hello there, my phone number is +2222222222",
+					PiiScrubInput: &scrubInput,
+				},
+			},
+		})
+		require.NoError(t, err)
+		require.Len(t, resp.GetOutputs(), 1)
+		require.Equal(t, "well hello there, my phone number is <PHONE_NUMBER>", resp.GetOutputs()[0].GetResult())
+	})
```
JoshVanL, line 76:
```suggestion
	})
	
	t.Run("scrub input great phone number", func(t *testing.T) {
		scrubInput := true

		resp, err := client.ConverseAlpha1(ctx, &rtv1.ConversationAlpha1Request{
			Name: "echo",
			Inputs: []*rtv1.ConversationInput{
				{
					Message:       "well hello there, my phone number is +4422222222",
					PiiScrubInput: &scrubInput,
				},
			},
		})
		require.NoError(t, err)
		require.Len(t, resp.GetOutputs(), 1)
		require.Equal(t, "well hello there, my phone number is <PHONE_NUMBER>", resp.GetOutputs()[0].GetResult())
	})
```

JoshVanL (changes requested):

pkg/api/universal/conversation.go
yaron2: It's an extremely cheap pass-through method that can't ever return an error, so I'm fine keeping it as is

pkg/api/universal/conversation.go
yaron2: Long term we should consider making PII scrubbers first class components. There are many good integrations such as [project Presidio](https://github.com/microsoft/presidio), AWS Comprehend and others.

dapr/proto/runtime/v1/dapr.proto
yaron2: Agree, field name changed

dapr/proto/runtime/v1/dapr.proto
yaron2: Changed, including description

tests/integration/suite/daprd/conversation/grpc/scrubpii.go
yaron2: Changed

tests/integration/suite/daprd/conversation/grpc/scrubpii.go
yaron2: Added :D

pkg/api/universal/conversation.go
yaron2: I'm a fan of returning early and and not using joined errors as they're harder to parse through. I believe I've seen in batched inferencing (which is what we're mimicking with accepting an input array of messages) early return of errors

JoshVanL (approved):

PR #8192, opened 2024-10-11
Error Codes Const Consolidation

jake-engelberg:
# Description

Screenshot 2024-10-15 at 1 36 19 PM

Error codes that were defined and used literally at the point of use in the debug log statements has been moved to a singular `errorcodes` package to be defined there.

The intention is to simplify/clarify error codes in assistance with adding metrics for these error codes.

## Issue reference

Proposal in progress:
https://github.com/dapr/proposals/pull/67

## Checklist

- [x] Code compiles correctly
- [x] Created/updated tests
- [ ] Unit tests passing


artursouza (approved):

pkg/messages/errorcodes/errorcodes.go
```diff
@@ -0,0 +1,103 @@
...
+	PubsubEmpty                     = "ERR_PUBSUB_EMPTY"
+	PubsubEventsSer                 = "ERR_PUBSUB_EVENTS_SER"
+	PubsubForbidden                 = "ERR_PUBSUB_FORBIDDEN"
+	PubsubNotConfigured             = "ERR_PUBSUB_NOT_CONFIGURED"
+	PubsubNotFound                  = "ERR_PUBSUB_NOT_FOUND"
+	PubsubPublishMessage            = "ERR_PUBSUB_PUBLISH_MESSAGE"
+	PubsubRequestMetadata           = "ERR_PUBSUB_REQUEST_METADATA"
+	PurgeWorkflow                   = "ERR_PURGE_WORKFLOW"
+	RaiseEventWorkflow              = "ERR_RAISE_EVENT_WORKFLOW"
+	ResumeWorkflow                  = "ERR_RESUME_WORKFLOW"
```
cicoyle, line 80: This is a nit, but can you make all the API specific error codes start with the respective API? So for ex: 
```suggestion
	WorkflowResume                  = "ERR_RESUME_WORKFLOW"
```
This will help developers later on if we standardize this a bit more by putting the API first.  I see several in this list: `PauseWorkflow`, `GetWorkflow`, `PurgeWorkflow`, `RaiseEventWorkflow`, etc

Then can you also group the errorcodes in this file by API? I see a group of workflow API error codes below, so this should get moved down there

elena-kolevska: +1 on this

pkg/components/state/pluggable.go
```diff
@@ -45,7 +45,7 @@ var (
 	ErrTransactOperationNotSupported = errors.New("transact operation not supported")
 )
 
-// errors code
+// errorcodes code
```
elena-kolevska, line 48: Was this change intended or a copy/paste error?

tests/integration/suite/daprd/pubsub/http/errors.go
```diff
@@ -112,7 +113,7 @@ spec:
 	}
 }
 
-func (e *errorcodes) Run(t *testing.T, ctx context.Context) {
+func (e *interrorcodes) Run(t *testing.T, ctx context.Context) {
```
elena-kolevska, line 116: Nit: can we please name the receiver `i`, for consistency?

elena-kolevska (changes requested):
I think this PR is a great improvements to our error codes, but I believe since we're here we can make them even nicer. For example it's not very clean that we have error messages in `predefined.go` and the corresponding error codes in `errorcodes.go` and they are disconnected, when in fact they form pairs.
In addition the `predefined.go` file also contains the APIError objects. Could we maybe do something like

```
type DaprErr struct {
	code    string
	message string
}

func (e DaprErr) Code() string {
	return e.code
}

func (e DaprErr) Msg(args ...interface{}) string {
	return fmt.Sprintf(e.message, args...)
}

var (
	NotSupportedStateOperation = DaprErr{
		code:    "ERR_STATE_OPERATION_NOT_SUPPORTED",
		message: "The state operation '%s' is not supported",
	}
)
```

And then use the errors like this:
```
msg := NewErrorResponse(
				NotSupportedStateOperation.Code(),
				NotSupportedStateOperation.Msg("my arg"))
```

This is just a quick draft without thinking about it too much.
If we don't like the new struct idea, we can at least align the const names, so it's `errorcodes.NotSupportedStateOperation` and `messages.NotSupportedStateOperation` instead of `messages.ErrNotSupportedStateOperation`.

PR #8212, opened 2024-10-18
[PoC] Support isolated Activities

olitomlinson:
# Description

### Background

Typically in a Dapr Workflows App, the `Workflows` and `Activity` code are co-located and **registered** in the same App instance.

After conversations with yaron, there is good reason to think that it should be possible to break this co-location, such that `Activities` can be registered on a different App instances to the `Workflow`, yet still have them interoperate. 

### Why?

This would allow the isolation/ring fencing of compute & memory to the `Activities` which are separated from the compute & memory of the `Workflow`. In a k8s environment, this would mean a `deployment` for the main App which contains the registered `Workflow` and then another `deployment` which contains the registered `Activities`. `Activities` could then scale independently of the rest of the main Workflow App.

### Challenge

As long as both apps share the same dapr `app-Id`, wfengine doesn't care that the workflows and activities are are not co-located. It's weird, but it works...

However, the problem lies in that the internal workflow Actors and internal Activity Actors are **both loaded by default** on each sidecar. Therefore, during the running of a Workflow, eventually the placement service will try and schedule an Activity onto the App which only has the Workflows registered. The Workflow fails at this point with the message

>  `dapr.workflow.failure.error_message": "Task '<ACTIVITY_NAME>' (#0) failed with an unhandled exception: No activity task named '<ACTIVITY_NAME>' was found.",`

### PoC

This PoC attempts to change this **default**, such that a sidecar can either care about _both_ Workflows & Activities, or _only_ Workflows, or _only_ Activities. 

### How?

In a complete solution, there would be a handshake between the Workflow Client SDK and the Workflow Engine to establish what is registered (Activities and/or Workflows), and then the sidecar would apply the desired settings to match. 

In the absence of a complete solution, for this PoC, I've simply added two new arguments to the `dapr run` in order to manually generate the desired effect.

- `--register-workflow-backend` (default `true`)
- `--register-activity-backend` (default `true`)

These flags will respectively swap out the *real* Durable Task Workers with a new [No-Op Task worker](https://github.com/dapr/dapr/blob/28097042c63800f1d4ebd83811decf867957ae81/pkg/runtime/wfengine/no_op_task_worker.go)

## Issue reference

## Checklist

- [ ] Code compiles correctly
- [ ] Created/updated tests
- [ ] Unit tests passing


yaron2:
Looks like this is solved by https://github.com/dapr/dapr/pull/8830

PR #8217, opened 2024-10-21
Actors: Reminder migration state->scheduler

JoshVanL:
Adds a backwards compatible migration mechanism to migrate Actor State Reminders to the new Scheduler based reminder system. Is activated on every Placement dissemination event, where all stored Actor reminders for the locally implemented Actor types are migrated to Scheduler Actor Reminer Jobs.

Migration of reminders are implemented on a per daprd basis, whereby upon dissemination, each daprd will list the reminders in the state store and Scheduler for actor types that it implements. Reminders for actor IDs which are not hosted for that daprd are discarded preventing multiple daprd's from migrating the same reminder. Daprd will determine which reminders to migrate based on missing items or differing contents from state store and Scheduler.


pkg/actors/reminders/scheduler.go
```diff
@@ -24,48 +24,73 @@ import (
...
 func (s *scheduler) SetStateStoreProviderFn(fn internal.StateStoreProviderFn) {
+	s.stateReminder.SetStateStoreProviderFn(fn)
 }
 
 func (s *scheduler) SetLookupActorFn(fn internal.LookupActorFn) {
+	s.lookUpActorFn = fn
+	s.stateReminder.SetLookupActorFn(fn)
 }
 
 func (s *scheduler) SetMetricsCollectorFn(fn remindersMetricsCollectorFn) {
```
cicoyle, line 79: should we fill in this func too while you are in this file? also fine to do as a follow up

tests/integration/suite/actors/reminders/migration/scheduler/distributed.go
```diff
@@ -0,0 +1,164 @@
...
+	assert.Len(t, d.scheduler.EtcdJobs(t, ctx), 100)
+
+	daprd6.Run(t, ctx)
+	daprd6.WaitUntilRunning(t, ctx)
+
+	assert.Len(t, d.db.ActorReminders(t, ctx, "myactortype").Reminders, 100)
+	assert.Len(t, d.db.ActorReminders(t, ctx, "myactortype2").Reminders, 100)
+	assert.Len(t, d.scheduler.EtcdJobs(t, ctx), 200)
+
+	daprd2.Cleanup(t)
```
cicoyle, line 159: should these be deferred in case the test runs into an err?

cicoyle (approved):
The changes lgtm.

The 1 e2e test failure, just needs the failure policy in as a pre-req since that test is shutting down/restarting dapr which from schedulers perspective will drop the jobs/reminders at trigger time until failure policy is in. 

I agree with Josh's suggestion for it, we can disable that test for this PR then re-add it with the failure policy PR

tests/integration/suite/actors/reminders/migration/scheduler/distributed.go
JoshVanL: Yeah I agree, thanks @cicoyle

JoshVanL:
Thanks @cicoyle, please take another look!

cicoyle (approved):
lgtm

PR #8220, opened 2024-10-22
Scheduler: Updates cron library to include staging queue

JoshVanL:
Updates the go-etcd-cron library to add support for the staging queue.
When a daprd Job or Actor Reminder is triggered but that job type is not
supported, the job is added to the staging queue. Once a daprd connects
to scheduler which implements that type, the job will be immediately
triggered to that host.

The `WatchJobsRequestResultStatus` proto has been added for the watch
jobs response to signal whether a job has succeeded or failed. This has
not been wired up with daprd yet.

Tests have been added for staging, and the default failure policy which
is built in to the cron library.


pkg/actors/actors.go
```diff
@@ -1220,7 +1217,7 @@ func (a *actorsRuntime) ExecuteLocalOrRemoteActorReminder(ctx context.Context, r
 		}
 		a.lock.Unlock()
 		go func() {
-			log.Debugf("Deleting reminder which was cancelled: %s", reminder.Key())
+			log.Errorf("Deleting reminder which was cancelled: %s", reminder.Key())
```
yaron2, line 1220: Could this not cause log explosion?

JoshVanL: Yes, thanks. Good catch.

Was using for testing.

cicoyle (approved):
lgtm

PR #8222, opened 2024-10-22
Scheduler: Expose FailurePolicy on internal API

JoshVanL:
Exposes the cron FailurePolicy on the internal Scheduler CRUD API.
Currently unused by any implementation- first expected is ActorReminders
with Workflows in following commits.

Adds the FailurePolicy proto to the scheduler.v1 package which is a
clone of the proto in the go-etcd-cron library.


tests/integration/suite/scheduler/failurepolicy/constant/interval.go
```diff
@@ -0,0 +1,78 @@
...
+	triggered := i.scheduler.WatchJobsFailed(t, ctx, &schedulerv1.WatchJobsRequestInitial{
+		AppId: "appid1", Namespace: "namespace",
+	})
+
+	assert.EventuallyWithT(t, func(c *assert.CollectT) {
+		assert.ElementsMatch(c, []string{"test"}, triggered.Slice())
+	}, time.Second*10, time.Millisecond*10)
+
+	time.Sleep(time.Second * 2)
+	assert.ElementsMatch(t, []string{"test"}, triggered.Slice())
```
elena-kolevska, line 73: I feel like this part might be flakey, but I don't know what would be a better way to test it. Lowering the sleep to 1 and the tick to 5ms should probably help.

elena-kolevska (approved):

tests/integration/suite/scheduler/failurepolicy/constant/interval.go
JoshVanL: Thanks, have updated.

cicoyle (approved):

elena-kolevska (approved):

PR #8225, opened 2024-10-23
[Actors API] update err desc to be more explicit

cicoyle:
Add to an err description to be more explicit to actor api newbies. I mixed up the actor id/type in my client code and couldnt figure out easily what I messed up or if I was seeing a bug. Made the err a bit more explicit for quicker user debugging experience

### DIFF
--- a/pkg/actors/actors.go
+++ b/pkg/actors/actors.go
@@ -76,7 +76,7 @@ var (
 	log = logger.NewLogger("dapr.runtime.actor")
 
 	ErrIncompatibleStateStore        = errors.New("actor state store does not exist, or does not support transactions which are required to save state - please see https://docs.dapr.io/operations/components/setup-state-store/supported-state-stores/")
-	ErrReminderOpActorNotHosted      = errors.New("operations on actor reminders are only possible on hosted actor types")
+	ErrReminderOpActorNotHosted      = errors.New("operations on actor reminders are only possible on hosted actor types. Check that the actor id/type is correct and hosted by the expected instance")
 	ErrTransactionsTooManyOperations = errors.New("the transaction contains more operations than supported by the state store")
 	ErrReminderCanceled              = internal.ErrReminderCanceled
 )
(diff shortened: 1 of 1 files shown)


pkg/actors/actors.go
```diff
@@ -76,7 +76,7 @@ var (
 	log = logger.NewLogger("dapr.runtime.actor")
 
 	ErrIncompatibleStateStore        = errors.New("actor state store does not exist, or does not support transactions which are required to save state - please see https://docs.dapr.io/operations/components/setup-state-store/supported-state-stores/")
-	ErrReminderOpActorNotHosted      = errors.New("operations on actor reminders are only possible on hosted actor types")
+	ErrReminderOpActorNotHosted      = errors.New("operations on actor reminders are only possible on hosted actor types. Check that the actor id/type is correct and hosted by the expected instance")
```
JoshVanL, line 79: Only type is relevant in this context afaik

JoshVanL (changes requested):

pkg/actors/actors.go
cicoyle: In my client code I swapped the actor id/type accidentally like this:
```
...
	err = client.RegisterActorReminder(ctx, &dapr.RegisterActorReminderRequest{
		ActorType: "playerActorType",
		ActorID:   actorID,
		Name:      "healthReminder",
		DueTime:   "1s",
		Period:    "2s",
		Data:      []byte(`"Reminder triggered"`),
	})
...
	err = client.UnregisterActorReminder(ctx, &dapr.UnregisterActorReminderRequest{
		ActorType: actorID, <------------------HERE
		ActorID:  "playerActorType", <------------------HERE
		Name:      "healthReminder",
	})
...
```

From a user perspective, it would have sped up my debugging bc I assumed it was a bug and missed that I swapped the id/type

elena-kolevska: I agree that only the type is relevant for this error though.

elena-kolevska (dismissed):

nelson-parente (approved):

Issue #8227, opened 2024-10-23
Allow configuring dapr grpc ports to values outsdie of dynamic range in kubernetes


passuied:
## In what area(s)?

/area runtime

## Describe the feature
We are experiencing intermittent errors when the dapr sidecar starts in all our environments (including production) with following error:
```
Failed to listen for gRPC server on TCP address :50002 with error: listen tcp :50002: bind: address already in use
```
After investigation, we understand this is due to the default GRPC port is part of the dynamic port range and the linkerd sidecar is choosing this port before it was claimed by dapr.
We would like to be able to pick a port outside of the dynamic range to avoid this issue and potential similar others.
After talking to @yaron2, we agreed that this feature could be helpful and easy to introduce.

## Release Note
RELEASE NOTE: **ADD** Provide ability to configure dapr grpc ports through dapr pod annotations.

PR #8228, opened 2024-10-23
Provide ability to configure dapr grpc ports through dapr pod annotations

passuied:
# Description

Provide ability to configure dapr grpc ports through dapr pod annotations

## Issue reference

#8227

## Checklist

- [x] Code compiles correctly
- [x] Created/updated tests
- [x] Unit tests passing

### DIFF
--- a/pkg/injector/patcher/sidecar.go
+++ b/pkg/injector/patcher/sidecar.go
@@ -56,8 +56,6 @@ type SidecarConfig struct {
 	RemindersService            string
 	SentrySPIFFEID              string
 	SidecarHTTPPort             int32 `default:"3500"`
-	SidecarAPIGRPCPort          int32 `default:"50001"`
-	SidecarInternalGRPCPort     int32 `default:"50002"`
 	SidecarPublicPort           int32 `default:"3501"`
 
 	Enabled                             bool    `annotation:"dapr.io/enabled"`
@@ -77,6 +75,8 @@ type SidecarConfig struct {
 	EnableDebug                         bool    `annotation:"dapr.io/enable-debug" default:"false"`
 	SidecarDebugPort                    int32   `annotation:"dapr.io/debug-port" default:"40000"`
 	Env                                 string  `annotation:"dapr.io/env"`
+	SidecarAPIGRPCPort                  int32   `annotation:"dapr.io/grpc-port" default:"50001"`
+	SidecarInternalGRPCPort             int32   `annotation:"dapr.io/internal-grpc-port" default:"50002"`
 	SidecarCPURequest                   string  `annotation:"dapr.io/sidecar-cpu-request"`
 	SidecarCPULimit                     string  `annotation:"dapr.io/sidecar-cpu-limit"`
 	SidecarMemoryRequest                string  `annotation:"dapr.io/sidecar-memory-request"`
--- a/pkg/injector/patcher/sidecar_container_test.go
+++ b/pkg/injector/patcher/sidecar_container_test.go
@@ -367,6 +367,69 @@ func TestGetSidecarContainer(t *testing.T) {
 		assert.Equal(t, corev1.PullAlways, container.ImagePullPolicy)
 	})
 
+	t.Run("get sidecar container with custom grpc ports", func(t *testing.T) {
+		c := NewSidecarConfig(&corev1.Pod{
+			ObjectMeta: metav1.ObjectMeta{
+				Annotations: map[string]string{
+					annotations.KeyAppID:            "app_id",
+					annotations.KeyConfig:           "config",
+					annotations.KeyAppPort:          "5000",
+					annotations.KeyLogAsJSON:        "true",
+					annotations.KeyAPITokenSecret:   "secret",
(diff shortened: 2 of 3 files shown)


yaron2:
Thanks for the PR!

You also need to make sure the sidecar container is configured with the `--dapr-internal-grpc-port` flag if the annotation exists. This is done here: https://github.com/dapr/dapr/blob/master/pkg/injector/patcher/sidecar_container.go#L39

In addition, the Dapr Operator needs to examine the annotations of a Dapr annotated resource and configure the Service that it creates with the right port. This is done here: https://github.com/dapr/dapr/blob/master/pkg/operator/handlers/dapr_handler.go#L265

passuied:
> You also need to make sure the sidecar container is configured with the `--dapr-internal-grpc-port` flag if the annotation exists. This is done here: https://github.com/dapr/dapr/blob/master/pkg/injector/patcher/sidecar_container.go#L39
From what I can tell this is already taken care of by pulling the port values mapped to the annotations... Am I reading this wrong?

yaron2:
> > You also need to make sure the sidecar container is configured with the `--dapr-internal-grpc-port` flag if the annotation exists. This is done here: https://github.com/dapr/dapr/blob/master/pkg/injector/patcher/sidecar_container.go#L39
> 
> From what I can tell this is already taken care of by pulling the port values mapped to the annotations... Am I reading this wrong?

Yes, that's right! So the only place left to modify is the Service creation logic in the operator

pkg/injector/patcher/sidecar.go
```diff
@@ -56,8 +56,6 @@ type SidecarConfig struct {
 	RemindersService            string
 	SentrySPIFFEID              string
 	SidecarHTTPPort             int32 `default:"3500"`
```
yevgen-el8, line 58: for consistency, do we want to make this one configurable as well?

yaron2: We could, however I'd suggest doing it in a separate PR and this is less pressing than changing 50002 as 3500 is within the non-dynamic range.

evhen14:
I've posted a PR to the dapr/docs repo to reflect the changes: https://github.com/dapr/docs/pull/4401

passuied:
> Yes, that's right! So the only place left to modify is the Service creation logic in the operator
PR updated with Service creation logic and tests

yaron2:
@passuied see the linter errors in `dapr/static checks`

yaron2 (dismissed):

pkg/injector/patcher/sidecar_container_test.go
```diff
@@ -30,6 +30,8 @@ import (
 	"github.com/dapr/kit/ptr"
 )
 
+const daprImage = "daprio/dapr"
```
JoshVanL, line 33: Nit: refrain from using constants in tests as they can potentially hide data values long term, and is always best to be literal in expression assertions.

```suggestion
```

pkg/operator/meta/meta.go
```diff
@@ -11,6 +13,16 @@ func IsAnnotatedForDapr(a map[string]string) bool {
 	return utils.IsTruthy(a[annotations.KeyEnabled])
 }
 
+func GetAnnotationIntValueOrDefault(a map[string]string, annotationKey string, defaultValue int) int {
```
JoshVanL, line 16: Should we be using a int32 flavour of this for ports?

pkg/operator/handlers/dapr_handler.go
```diff
@@ -349,6 +353,14 @@ func (h *DaprHandler) getMetricsPort(wrapper ObjectWrapper) int {
 	return metricsPort
 }
 
+func (h *DaprHandler) getGRPCPort(wrapper ObjectWrapper) int {
+	return meta.GetAnnotationIntValueOrDefault(wrapper.GetTemplateAnnotations(), annotations.KeyAPIGRPCPort, daprSidecarDefaultAPIGRPCPort)
+}
+
+func (h *DaprHandler) getInternalGRPCPort(wrapper ObjectWrapper) int {
```
JoshVanL, line 360: ?

```suggestion
func (h *DaprHandler) getInternalGRPCPort(wrapper ObjectWrapper) int32 {
```

JoshVanL (changes requested):
Some nits from me, otherwise lgtm

pkg/injector/patcher/sidecar_container_test.go
passuied: These were due to static check complaints. Should I just suppress them?

JoshVanL: Yes please. Best is to put the nolint tag above `package` at the top.

JoshVanL (dismissed):

JoshVanL:
Thanks @passuied

yaron2 (approved):

yaron2:
Small lint issues

marcduiker:
@holopin-bot @passuied Thank you PAtrick!

PR #8238, opened 2024-10-24
Verify Scheduler Enabled

cicoyle:
Verify Scheduler is enabled before connecting.


artursouza (approved):

pkg/runtime/runtime.go
```diff
@@ -334,6 +332,16 @@ func newDaprRuntime(ctx context.Context,
 		},
 	)
 
+	if runtimeConfig.SchedulerEnabled() {
+		err := rt.runnerCloser.Add(
+			rt.schedulerClients.Run,
+			rt.jobsManager.Run,
+		)
+		if err != nil {
+			log.Errorf("Scheduler enabled, but dapr failed to connect. Unable to use the Jobs API.")
```
JoshVanL, line 341: This error must be returned.

pkg/runtime/runtime.go
```diff
@@ -334,6 +332,16 @@ func newDaprRuntime(ctx context.Context,
 		},
 	)
 
+	if runtimeConfig.SchedulerEnabled() {
```
JoshVanL, line 335: Both of these runners are designed to be able to run, regardless of whether there are any addresses/clients. Scheduler is also signaled as enabled if there is a non zero number of addresses. It's not clear to me why we need this change at all.

JoshVanL (changes requested):

PR #8256, opened 2024-10-30
Metric Error Codes Recording + Error Code Prefixing

jake-engelberg:
# Description

Error codes that were defined and used on a per-developer basis for debug logging has been temporarily moved to a singular `errorcodes` package to be defined and prefixed for cohesion. Eventually, they will be converted to structures similar to `predefined.go` for all errors and associated messages.

Additionally, error codes are now retrieved using a function which also records a count metric for tracking errors.

No renaming has been done to assure backwards compatability.

The metric can be filtered by its keys for app-id, error code, and error category.

The error metric recording is opt-in via the flag `metrics.recordErrorCodes`.

Metric fields/schema:
```
dapr_error_code_total
{
    app_id = "errorgen-app",
    category = "pubsub",
    dapr_io_enabled = "true",
    error_code = "ERR_PUBSUB_NOT_FOUND",
    instance = "10.244.0.35:9090",
    job = "kubernetes-service-endpoints",
    namespace = "crud-app",
    node = "dapr-master-control-plane",
    service = "errorgen-app-dapr"
}
```

Example of query `dapr_error_code_total` in Grafana:
Screenshot 2024-11-14 at 10 00 41 PM

Example of query `rate(dapr_error_code_total[2m])` in Grafana:
Screenshot 2024-11-14 at 10 00 12 PM

Another img:
Screenshot 2024-11-27 at 11 44 37 AM

## Issue reference

Please reference the issue this PR will close: https://github.com/dapr/proposals/pull/67

## Checklist

- [x] Code compiles correctly
- [x] Created/updated tests
- [x] Unit tests passing
- [x] Extended the documentation / Created issue in the <https://github.com/dapr/docs/> repo: https://github.com/dapr/docs/pull/4422
- [x] Specification has been updated / Created issue in the <https://github.com/dapr/docs/> repo: https://github.com/dapr/docs/pull/4422
- [x] Provided sample for the feature / Created issue in the <https://github.com/dapr/docs/> repo: https://github.com/dapr/docs/pull/4422


tests/e2e/healthcheck/healthcheck_test.go
```diff
@@ -266,7 +266,7 @@ func testAppHealthCheckProtocol(t *testing.T, protocol string) {
 			// Service invocation should fail
 			go func() {
 				res, status := invokeFoo(t)
-				require.Contains(t, string(res), "ERR_DIRECT_INVOKE")
+				require.Contains(t, string(res), "ERR_CONVERSATION_DIRECT_INVOKE")
```
filintod, line 269: let's use the constant value from `errorcodes` package. the same in all tests

elena-kolevska: I'm in the camp of always using hardcoded error codes in tests, so that it's very obvious if we break compatibility in a later version.

elena-kolevska (changes requested):
I'll be doing a more detailed review tomorrow, but sending my comments about breaking changes early so you can work on it while we're looking at the rest.

pkg/api/http/directmessaging_test.go
```diff
@@ -411,7 +411,7 @@ func TestV1DirectMessagingEndpoints(t *testing.T) {
 		// assert
 		mockDirectMessaging.AssertNumberOfCalls(t, "Invoke", 1)
 		assert.Equal(t, 500, resp.StatusCode)
-		assert.Truef(t, strings.HasPrefix(string(resp.RawBody), "{\"errorCode\":\"ERR_MALFORMED_RESPONSE\",\"message\":\""), "code not found in response: %v", string(resp.RawBody))
+		assert.Truef(t, strings.HasPrefix(string(resp.RawBody), "{\"errorCode\":\"ERR_COMMON_MALFORMED_RESPONSE\",\"message\":\""), "code not found in response: %v", string(resp.RawBody))
```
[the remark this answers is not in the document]
jake-engelberg: I'm aware of this, and I don't exactly have much of a preference of whether or not to change the errors; It was described to me as breaking but logs are not in the contract of what we have to be super weary of; the old error codes were inconsistent in naming, like:
ERR_RAISE_EVENT_WORKFLOW
ERR_WORKFLOW_NAME_MISSING
...so I envision someday we'd want to rename these errors, unless you see errors like ERR_WORKFLOW_NAME_MISSING not as being prefixed but just a phrase that happens to start with the component.

In the case of the latter, as well as generally just wanting to prevent breaking changes, I can revert to the original error codes.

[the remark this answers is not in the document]
elena-kolevska: Aren't these the same codes that are returned in http responses too? I haven't looked deeper into the PR yet, this is just something that caught my eye at first glance.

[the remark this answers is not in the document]
jake-engelberg: Sometimes, yes; in that case, it would be more a potential breaking change if users rely on consistent responses, so it adds another reason to revert to original error codes which I will do so.

pkg/config/testdata/override_gen.yaml
```diff
@@ -25,6 +25,7 @@ spec:
         allowedClockSkew: 1h
     metric:
         enabled: true
+        recorderrorcodes: null
```
JoshVanL, line 28:
```suggestion
        recordErrorCodes: null
```

pkg/config/testdata/override_spec_gen.yaml
```diff
@@ -6,6 +6,7 @@ mtls:
     allowedClockSkew: 1h
 metric:
     enabled: true
+    recorderrorcodes: null
```
JoshVanL, line 9:
```suggestion
    recordErrorCodes: null
```

pkg/diagnostics/errorcode_monitoring.go
```diff
@@ -0,0 +1,52 @@
+package diagnostics
```
JoshVanL, line 1: Add copyright headers.

pkg/diagnostics/errorcode_monitoring.go
```diff
@@ -0,0 +1,52 @@
...
+
+	diagUtils "github.com/dapr/dapr/pkg/diagnostics/utils"
+	"github.com/dapr/dapr/pkg/messages/errorcodes"
+)
+
+type errorCodeMetrics struct {
+	errorCodeCount *stats.Int64Measure
+
+	appID   string
+	ctx     context.Context
```
JoshVanL, line 18:
```suggestion
```

pkg/diagnostics/errorcode_monitoring.go
```diff
@@ -0,0 +1,52 @@
...
+
+func newErrorCodeMetrics() *errorCodeMetrics {
+	return &errorCodeMetrics{ //nolint:exhaustruct
+		errorCodeCount: stats.Int64(
+			"error_code/count",
+			"Number of times an error with a specific errorcode was encountered.",
+			stats.UnitDimensionless),
+
+		ctx:     context.Background(),
+		enabled: false,
```
JoshVanL, line 30: AppID?

pkg/diagnostics/errorcode_monitoring_test.go
```diff
@@ -0,0 +1,55 @@
+package diagnostics
```
JoshVanL, line 1: Copyright headers.

pkg/diagnostics/errorcode_monitoring.go
```diff
@@ -0,0 +1,52 @@
...
+
+// Init registers the errorcode metrics view.
+func (m *errorCodeMetrics) Init(id string) error {
+	m.enabled = true
+	m.appID = id
+
+	return view.Register(
+		diagUtils.NewMeasureView(m.errorCodeCount, []tag.Key{appIDKey, errorCodeKey, categoryKey}, view.Count()),
+	)
+}
```
JoshVanL, line 42: Remove, and move to `New`.

pkg/messages/errorcodes/errorcodes.go
```diff
@@ -0,0 +1,138 @@
+/*
+Copyright 2022 The Dapr Authors
```
JoshVanL, line 2:
```suggestion
Copyright 2024 The Dapr Authors
```

JoshVanL (changes requested):

pkg/diagnostics/service_monitoring.go
```diff
@@ -23,13 +23,15 @@ var (
 	namespaceKey        = tag.MustNewKey("namespace")
 	resiliencyNameKey   = tag.MustNewKey("name")
 	policyKey           = tag.MustNewKey("policy")
+	errorCodeKey        = tag.MustNewKey("errorcode")
```
nelson-parente, line 26: nit: follow the same pattern we have on other keys error_code or errorCode (we have both, choose one I guess)

pkg/config/configuration.go
```diff
@@ -254,8 +254,9 @@ func (o OtelSpec) GetIsSecure() bool {
 // MetricSpec configuration for metrics.
 type MetricSpec struct {
 	// Defaults to true
-	Enabled *bool       `json:"enabled,omitempty" yaml:"enabled,omitempty"`
-	HTTP    *MetricHTTP `json:"http,omitempty" yaml:"http,omitempty"`
+	Enabled          *bool       `json:"enabled,omitempty" yaml:"enabled,omitempty"`
+	RecordErrorCodes *bool       `json:"recordErrorCodes,omitempty"`
```
nelson-parente, line 258: add yaml tag

pkg/api/http/metadata.go
```diff
@@ -22,6 +22,7 @@ import (
 
 	"github.com/dapr/dapr/pkg/api/http/endpoints"
 	"github.com/dapr/dapr/pkg/messages"
+
```
nelson-parente, line 25: nit: remove this new line

pkg/diagnostics/errorcode_monitoring.go
jake-engelberg: app-id is supplied with the `Init()` function, following the other metrics' structures
https://github.com/jake-engelberg/dapr//blob/5c3b42bf03519008739c1fc2ec8030dba5f1fd00/pkg/diagnostics/errorcode_monitoring.go#L48

pkg/diagnostics/errorcode_monitoring.go
jake-engelberg: the metric is optional and so the Init() is a secondary function to call when the flag is checked

https://github.com/jake-engelberg/dapr//blob/5c3b42bf03519008739c1fc2ec8030dba5f1fd00/pkg/diagnostics/metrics.go#L53

elena-kolevska (changes requested):
Thank you @jake-engelberg I'm happy we're moving even further in making Dapr errors easier to work with. 🚀 

I left a few comments, but I also wanted to point something that's missing: 
The current error metric recording relies on the error code we used to use for http mostly, but we don't have that for new code like the scheduler and the Jobs API where we use the Rich grpc errors exclusively and we use the `ErrorInfo` string to derive the http error code (you can see how that's done [here](https://github.com/dapr/kit/blob/main/errors/errors.go#L214)).
It's a simple change though, you can look through the [scheduler errors file](https://github.com/dapr/dapr/blob/master/pkg/api/errors/scheduler.go) and look at the `WithErrorInfo` calls. The first argument is the tag you need to record.

We'll also need to do this for the control plane services, but we can leave that for a separate PR.

jake-engelberg:
> Thank you @jake-engelberg I'm happy we're moving even further in making Dapr errors easier to work with. 🚀
> 
> I left a few comments, but I also wanted to point out that we seem not to be recording scheduler errors; The current error metric recording relies on the error code we used to use for http mostly, but we don't have that for new code like the scheduler and the Jobs API where we use the Rich grpc errors exclusively and we use the `ErrorInfo` string to derive the http error code (you can see how that's done [here](https://github.com/dapr/kit/blob/main/errors/errors.go#L214)). It's a simple change though, you can look through the [scheduler errors file](https://github.com/dapr/dapr/blob/master/pkg/api/errors/scheduler.go) and look at the `WithErrorInfo` calls. The first argument is the tag you need to record.
> 
> We'll also need to do this for the control plane services, but we can leave that for a separate PR.

Thanks for this point brought up; I skipped over the Jobs API errors because of those composite codes, but I went back and made some specific functions to allow recording of those various codes even though they may not be defined in the new `errorcodes` package.

While I would have liked to either modify WithErrorInfo, or wrap it, to minimize repetition, I decided against it for these two reasons:
1. I wanted to keep dependencies within `dapr/dapr`, so I decided to not make changes to WithErrorInfo in `dapr/kit`

2. Within `dapr/dapr`, I can't make a wrapper to record and call `WithErrorInfo()` as well as `RecordAndGet()` as the calling struct is defined in `dapr/kit`, not local to anywhere in `dapr/dapr`, so I can't do:
```
func (b *kiterrors.ErrorBuilder) WithErrorInfoAndRecord(){
    //Performs metric recording as well as has original WithErrorInfo() function call
}
```

LMK if you see some other way to reduce repetition, for my solution meant adding a `RecordAndGetComp()` call with each `WithErrorInfo()`.

elena-kolevska:
@jake-engelberg  Fix linter please, so the tests can run 🙏

filintod:
could we change the suffix to be `_total` instead of `_count` to follow convention https://prometheus.io/docs/practices/naming/. We don't follow a single convention in dapr but probably to move to common ones as we go

pkg/api/http/actors.go
```diff
@@ -193,13 +193,13 @@ func (a *api) onCreateActorReminder(w http.ResponseWriter, r *http.Request) {
 	err = a.universal.Actors().CreateReminder(ctx, &req)
 	if err != nil {
 		if errors.Is(err, actors.ErrReminderOpActorNotHosted) {
-			msg := messages.ErrActorReminderOpActorNotHosted
+			msg := messages.ErrActorReminderOpActorNotHosted.RecordAndGet()
```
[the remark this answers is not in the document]
jake-engelberg: Changes related to above: https://github.com/dapr/dapr/pull/8256/commits/7c6a69175d553c82f11b99f847ca5108000e3a9b

I've went through and pushed the metric recording up in the call stack as far as I could understand, and ended up with a few functions that take different structures depending on how the end of the chain's parameters were / how the error's structure ended up looking like:

1. `RecordAPIErrorCode()`
    * Used at the end of most HTTP responses, which are provided an APIError struct
2. `RecordErrorCode()`
    * Used at the end of various chains, which are provided the new ErrorCode struct, like some other HTTP responses, the HTTP error response helper `NewErrorResponse()`, some crypto gRPC errors, Pubsub error builder
3. `RecordCompAndGet()`
    * Used at the end of various chains, which are provided various composited errors not following the `"ERR_X_X_X" format, some general error builders, scheduler errors, state error builder


In my opinion, it might be somewhat less understandable than having the recording added to each usage of the error codes, but it did lower the lines changed drastically and seems like a workable solution for the time being.

pkg/diagnostics/errorcode_monitoring_test.go
```diff
@@ -0,0 +1,55 @@
...
+
+	"github.com/dapr/dapr/pkg/messages/errorcodes"
+)
+
+func TestRecordErrorCode(t *testing.T) {
+	t.Run("record single error code", func(t *testing.T) {
+		m := newErrorCodeMetrics()
+		_ = m.Init("app-id")
+
+		m.RecordErrorCode(errorcodes.ActorInstanceMissing)
```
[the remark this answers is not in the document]
jake-engelberg: There is an additional test for 2 more error codes at the same time, can you expand on this suggestion?
Screenshot 2024-11-13 at 12 36 32 PM

elena-kolevska (changes requested):
We're getting there @jake-engelberg , thanks for being patient with us. It's so important we get this right :)
I have to say personally I don't love the record and get thing. A method with two verbs is usually a code smell and it violates single responsabillity principle, but I'm not going to hold on approval based on that (as soon as we clean everything up).

We also need to wait for [this PR](https://github.com/dapr/dapr/pull/8237) to get merged, because we're gonna have to adjust some code.

pkg/api/universal/subtlecrypto_subtlecrypto.go
```diff
@@ -158,8 +158,9 @@ func (a *Universal) SubtleEncryptAlpha1(ctx context.Context, in *runtimev1pb.Sub
 	if err != nil {
 		// We are not going to return the exact error from the component to the user, because an error that is too specific could allow for various side channel attacks (e.g. AES-CBC and padding oracle attacks)
 		// We will log the full error as a debug log, but only return a generic one to the user
-		a.logger.Debug(messages.ErrCryptoOperation.WithFormat(err))
-		err = messages.ErrCryptoOperation.WithFormat("failed to encrypt")
+		apiError := messages.ErrCryptoOperation
```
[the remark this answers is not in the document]
jake-engelberg: My understanding of the call hiearchy is that the functions in this file are handled by `UniversalHTTPHandler()`
Screenshot 2024-11-14 at 3 07 19 PM

Which then handles lots of APIErrors already but also the one's you commented about here.
Screenshot 2024-11-14 at 3 08 39 PM

pkg/api/errors/errors.go
```diff
@@ -38,7 +40,7 @@ func NotFound(name string, componentType string, metadata map[string]string, grp
 		message,
 		legacyTag,
 	).
-		WithErrorInfo(reason, metadata).
+		WithErrorInfo(messages.RecordCompAndGet(reason, errorcodes.CategoryCommon), metadata).
```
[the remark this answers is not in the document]
jake-engelberg: I think it actually is because it ends up at `UniversalHTTPHandler()` like [this comment as well](https://github.com/dapr/dapr/pull/8256#discussion_r1842830509), I'll remove.

pkg/api/universal/actors.go
```diff
@@ -55,11 +55,12 @@ func (a *Universal) ActorReadinessCheck(ctx context.Context) error {
 	a.WaitForActorsReady(ctx)
 
 	if a.Actors() == nil {
+		err := messages.ErrActorRuntimeNotFound
```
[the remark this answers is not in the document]
jake-engelberg: `respondWithError()`
Screenshot 2024-11-14 at 3 23 48 PM

pkg/api/grpc/crypto.go
```diff
@@ -42,6 +43,7 @@ func (a *api) EncryptAlpha1(stream runtimev1pb.Dapr_EncryptAlpha1Server) (err er
 	if err != nil {
 		// This is already an APIError object.
 		a.logger.Debug(err)
+		messages.RecordErrorCode(errorcodes.CommonBadRequest)
```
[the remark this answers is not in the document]
jake-engelberg: You are correct, I made a wrapper to record any of the returned APIErrors for both, lmk if this is bad practice:
Screenshot 2024-11-14 at 3 25 10 PM

pkg/api/universal/actors.go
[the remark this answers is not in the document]
elena-kolevska: That's only on the http side though, no? If the universal API is accessed through the gRPC side, it won't record it.
Please feel free to correct me if I'm missing something, you've been looking into this way more than me.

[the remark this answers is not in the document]
jake-engelberg: This is correct, I ended up adding to the gRPC metric recording which is a middleware that will intercept the errors to do the same thing as the universal HTTP recorder.

elena-kolevska:
Hey @jake-engelberg  the PR I referred to before is merged, please update your branch 🙏 

i saw the changes you made and i think we're going in the right direction, by centralising the recording as much as possible.

jake-engelberg:
https://github.com/dapr/kit/pull/109

This PR makes some changes to dapr/kit that are necessary for the latest commit I made.

pkg/api/http/responses.go
```diff
@@ -122,7 +124,7 @@ func respondWithProto(w http.ResponseWriter, m protoreflect.ProtoMessage, status
 		EmitUnpopulated: emitUnpopulated,
 	}.Marshal(m)
 	if err != nil {
-		msg := NewErrorResponse("ERR_INTERNAL", "failed to encode response as JSON: "+err.Error())
+		msg := NewErrorResponse(errorcodes.CommonInternal, "failed to encode response as JSON: "+err.Error())
```
elena-kolevska, line 127: Seems like we're not recording this error

pkg/components/pluggable/errors.go
```diff
@@ -113,7 +113,7 @@ func KitErrorFromStatus(s *status.Status) (error, bool) {
 		return s.Err(), false
 	}
 
-	return kiterrors.NewBuilder(s.Code(), httpCode, s.Message(), info.GetMetadata()[tagField]).
+	return kiterrors.NewBuilder(s.Code(), httpCode, s.Message(), info.GetMetadata()[tagField], "unknown").
```
elena-kolevska, line 116: We need to look a bit deeper into this, if I remember correctly how pluggable components work, we won't be able to extract the original category, but we could maybe create a new generic one, for example, `pluggable-state`.  @JoshVanL what do you think?
For context (so you don't have to look through the whole PR), we added a `category` field needed for recording metrics to kit and APIErrors, but we never return it as part of the final error the client sees. Does that mean a pluggable component won't have access to it?

elena-kolevska: I'm also ok marking this as category `pluggable-component` for now, and split it in the future if necessary.

pkg/diagnostics/grpc_monitoring.go
```diff
@@ -227,6 +227,7 @@ func (g *grpcMetrics) UnaryServerInterceptor() func(ctx context.Context, req int
 			size = g.getPayloadSize(resp)
 		}
 		g.ServerRequestSent(ctx, info.FullMethod, status.Code(err).String(), int64(g.getPayloadSize(req)), int64(size), start)
+		RecordErrorCode(err)
```
elena-kolevska, line 230: Let's check if err is nil before calling the record function

tests/integration/framework/process/daprd/daprd.go
```diff
@@ -350,6 +350,7 @@ func (d *Daprd) Metrics(t *testing.T, ctx context.Context) map[string]float64 {
 		for _, m := range mf.GetMetric() {
 			metricName := mf.GetName()
 			labels := ""
+			// t.Log(metricName)
```
elena-kolevska, line 353:
```suggestion

```

elena-kolevska (changes requested):
Looks like we're almost done! Just a few small nits.

pkg/diagnostics/grpc_monitoring.go
filintod: wdyt on adding a nil check inside `RecordErrorCode`?

like:
```
func RecordErrorCode(err error) bool {
  if err == nil { return false}
  ...
}
```

filintod: but no need to change more in any case

elena-kolevska: Yep, even better!

elena-kolevska:
Looking great! Let's wait for the tests the pass and I'll approve. Good job Jake!

elena-kolevska (dismissed):

elena-kolevska (approved):

artursouza (approved):

marcduiker:
@holopin-bot @jake-engelberg Thanks Jake!

PR #8264, opened 2024-11-05
Unregister actor in placement

elena-kolevska:
# Description

This PR makes it possible for a dapr host to unregister actor types. It is used for dynamic actor registration in workflows.

Previously, Dapr treated any instance that didn’t host actors as a non-actor host, skipping the Raft state update and dissemination. Now, Dapr also considers that such an instance could have previously hosted actor types that have since been unregistered.

## Checklist

- [X] Code compiles correctly
- [X] Created/updated tests
- [X] Unit tests passing


pkg/placement/membership_test.go
```diff
@@ -91,30 +91,20 @@ func TestMembershipChangeWorker(t *testing.T) {
...
-
-				if assert.Len(c, operations1, 3) {
-					require.Equal(c, lockOperation, operations1[0])
-					require.Equal(c, updateOperation, operations1[1])
-					require.Equal(c, unlockOperation, operations1[2])
+					if assert.Len(c, tables.GetEntries(), 2) &&
+						assert.Contains(c, tables.GetEntries(), "actor1") &&
+						assert.Contains(c, tables.GetEntries(), "actor2") {
+						return
+					}
```
elena-kolevska, line 105: Updated the test so that it doesn't depend on order of receiving the stream messages. Removes flakiness.

pkg/placement/raft/server.go
```diff
@@ -381,6 +381,11 @@ func (s *Server) ApplyCommand(cmdType CommandType, data DaprHostMember) (bool, e
 	s.lock.RLock()
 	defer s.lock.RUnlock()
 
+	// If the command is to upsert a member, but the member has no entities, treat it as a remove command.
+	if cmdType == MemberUpsert && len(data.Entities) == 0 {
+		cmdType = MemberRemove
+	}
```
elena-kolevska, line 387: Since our raft store is specialised for storing actor hosts, we don't need to do upsert here, instead a command that says "this host has no actor" is equivalent to removing that host from the table.

yaron2 (approved):

PR #8266, opened 2024-11-05
hotreload: add component callback support

lrascao:
# Description

Component callback is a useful dapr integration functionality for applications to react on when components are loaded. This patch bridges the gap left between this and hot reload providing the same experience on both scenarios.

## Checklist

- [X] Code compiles correctly
- [ ] Created/updated tests
- [ ] Unit tests passing


pkg/runtime/hotreload/reconciler/components.go
```diff
@@ -68,6 +70,14 @@ func (c *components) update(ctx context.Context, comp compapi.Component) {
 	if c.proc.AddPendingComponent(ctx, comp) {
 		log.Infof("Component updated: %s", comp.LogName())
 		c.proc.WaitForEmptyComponentQueue()
+
+		if c.cb != nil {
+			if err := c.cb(registry.ComponentRegistry{
```
JoshVanL, line 75: The callbacks here are at the wrong level- they need to be inside the component processor in `Init` and `Close`. Please return errors.

https://github.com/dapr/dapr/blob/334ae9eea43d487a7b29a0e4aef904e3eba57a10/pkg/runtime/processor/components.go#L30
https://github.com/dapr/dapr/blob/334ae9eea43d487a7b29a0e4aef904e3eba57a10/pkg/runtime/processor/components.go#L51

JoshVanL (changes requested):

pkg/runtime/hotreload/reconciler/components.go
lrascao: updated

pkg/runtime/runtime.go
```diff
@@ -215,6 +215,7 @@ func newDaprRuntime(ctx context.Context,
 		ActorsEnabled:   len(runtimeConfig.actorsService) > 0,
 		Registry:        runtimeConfig.registry,
 		ComponentStore:  compStore,
+		ComponentCb:     runtimeConfig.registry.ComponentsCallback(),
```
famarting, line 218: is it true that at some point in the runtime initialization components callback is set to nil? I think I recall something like that

famarting:
love this btw

pkg/runtime/runtime.go
JoshVanL: It’s a pain to have a nil checker for this. If we can, let’s make sure this is always at least a noop func

pkg/runtime/processor/components.go
```diff
@@ -44,7 +45,19 @@ func (p *Processor) Init(ctx context.Context, comp componentsapi.Component) erro
 		return errors.Join(err, p.compStore.DropPendingComponent())
 	}
 
-	return p.compStore.CommitPendingComponent()
+	if err := p.compStore.CommitPendingComponent(); err != nil {
+		return fmt.Errorf("error committing component: %w", err)
+	}
+
+	if p.componentCb != nil {
+		if err := p.componentCb(registry.ComponentRegistry{
```
JoshVanL, line 53: This is not capturing init errors. I’m not sure if that’s intended.. we likely need to update the signature and wrap this init and close to pass that on.

JoshVanL (changes requested):

pkg/runtime/runtime.go
famarting: ah no, ignore me, its `a.appHealthReady` what its set to nil

pkg/runtime/processor/components.go
JoshVanL: It also follows that we should close the component of callback fails.

lrascao:
closing in favour of https://github.com/dapr/dapr/pull/8268

PR #8268, opened 2024-11-07
Add a resource reporter

lrascao:
# Description

Allows applications that embed daprd to be notified of relevant events on the resource lifecycle (eg. init, close).

## Issue reference

N/A

## Checklist

- [X] Code compiles correctly
- [X] Created/updated tests
- [X] Unit tests passing
- [X] End-to-end tests passing


dapr/proto/operator/v1/component.proto
```diff
@@ -0,0 +1,87 @@
+/*
```
JoshVanL, line 1: Let's rename the file to `resource.proto`.

dapr/proto/operator/v1/component.proto
```diff
@@ -0,0 +1,87 @@
...
+  FAILURE = 2;
+}
+
+// ResourceType is the type of the resource.
+enum ResourceType {
+  // UNKNOWN_RESOURCE indicates that the resource type is unknown.
+  RESOURCE_UNKNOWN = 0;
+
+  // COMPONENT indicates that the resource type is a component.
+  COMPONENT = 1;
```
JoshVanL, line 40:
```suggestion
  // RESOURCE_COMPONENT indicates that the resource type is a component.
  RESOURCE_COMPONENT = 1;
```

dapr/proto/operator/v1/component.proto
```diff
@@ -0,0 +1,87 @@
...
+// EventType is the type of the event.
+enum EventType {
+  // UNKNOWN indicates that the event type is unknown.
+  EVENT_UNKNOWN = 0;
+
+  // INIT indicates that the event type is init.
+  INIT = 1;
+
+  // CLOSE indicates that the event type is close.
+  CLOSE = 2;
```
JoshVanL, line 52:
```suggestion
  // EVENT_INIT indicates that the event type is init.
  EVENT_INIT = 1;

  // EVENT_CLOSE indicates that the event type is close.
  EVENT_CLOSE = 2;
```

pkg/runtime/processor/components.go
```diff
@@ -23,11 +23,44 @@ import (
...
+
+	if err := p.reporter(ctx, comp,
+		operatorv1.ResourceResult{
+			ResourceType:        operatorv1.ResourceType_COMPONENT,
+			EventType:           operatorv1.EventType_INIT,
+			Name:                comp.Name,
+			Condition:           condition,
+			Reason:              reason,
+			Message:             message,
+			ObservedGeneration:  comp.Generation,
```
JoshVanL, line 53:
```suggestion
			ObservedGeneration:  comp.GetGeneration(),
```

pkg/runtime/processor/components.go
```diff
@@ -23,11 +23,44 @@ import (
...
+		condition = operatorv1.ResourceConditionStatus_FAILURE
+		reason = ptr.Of("ERROR")
+		message = ptr.Of(initerr.Error())
+	}
+
+	if err := p.reporter(ctx, comp,
+		operatorv1.ResourceResult{
+			ResourceType:        operatorv1.ResourceType_COMPONENT,
+			EventType:           operatorv1.EventType_INIT,
+			Name:                comp.Name,
```
JoshVanL, line 49:
```suggestion
			Name:                comp.GetName(),
```

pkg/runtime/processor/components.go
```diff
@@ -23,11 +23,44 @@ import (
...
+			Condition:           condition,
+			Reason:              reason,
+			Message:             message,
+			ObservedGeneration:  comp.Generation,
+			LastTransactionTime: timestamppb.New(time.Now()),
+		}); err != nil {
+		return errors.Join(initerr, fmt.Errorf("error reporting component init result: %w", err))
```
JoshVanL, line 56:
```suggestion
		return errors.Join(initerr, fmt.Errorf("error reporting component init result: %w", err), p.Close(comp))
```

pkg/runtime/processor/components.go
```diff
@@ -44,11 +77,45 @@ func (p *Processor) Init(ctx context.Context, comp componentsapi.Component) erro
...
+
+	if err := p.reporter(context.TODO(), comp,
+		operatorv1.ResourceResult{
+			ResourceType:        operatorv1.ResourceType_COMPONENT,
+			EventType:           operatorv1.EventType_CLOSE,
+			Name:                comp.Name,
+			Condition:           condition,
+			Reason:              reason,
+			Message:             message,
+			ObservedGeneration:  comp.Generation,
```
JoshVanL, line 108:
```suggestion
			ObservedGeneration:  comp.GetGeneration(),
```

pkg/runtime/processor/components.go
```diff
@@ -44,11 +77,45 @@ func (p *Processor) Init(ctx context.Context, comp componentsapi.Component) erro
...
+		condition = operatorv1.ResourceConditionStatus_FAILURE
+		reason = ptr.Of("ERROR")
+		message = ptr.Of(closeErr.Error())
+	}
+
+	if err := p.reporter(context.TODO(), comp,
+		operatorv1.ResourceResult{
+			ResourceType:        operatorv1.ResourceType_COMPONENT,
+			EventType:           operatorv1.EventType_CLOSE,
+			Name:                comp.Name,
```
JoshVanL, line 104:
```suggestion
			Name:                comp.GetName(),
```

pkg/runtime/processor/components.go
```diff
@@ -44,11 +77,45 @@ func (p *Processor) Init(ctx context.Context, comp componentsapi.Component) erro
...
+	// after performing the initialization, report the result
+	condition := operatorv1.ResourceConditionStatus_SUCCESS
+	var reason, message *string
+	if closeErr != nil {
+		condition = operatorv1.ResourceConditionStatus_FAILURE
+		reason = ptr.Of("ERROR")
+		message = ptr.Of(closeErr.Error())
+	}
+
+	if err := p.reporter(context.TODO(), comp,
```
JoshVanL, line 100: How much work is it to update `Close` to accept a context? Ideally we should.

pkg/runtime/processor/processor.go
```diff
@@ -179,6 +183,11 @@ func New(opts Options) *Processor {
 		Meta:           opts.Meta,
 	})
 
+	// ensure a default no-op reporter
+	if opts.Reporter == nil {
+		opts.Reporter = DefaultReporter
+	}
```
JoshVanL, line 189: Please don't write to options.

```suggestion
	reporter := DefaultReporter
	if opts.Reporter != nil {
	  reporter = opts.Reporter
	}
```

pkg/runtime/processor/processor_test.go
```diff
@@ -462,3 +487,170 @@ func TestProcessNoWorkflow(t *testing.T) {
...
+			err := proc.processComponentAndDependents(context.Background(), pubsubComponent)
+			require.NoError(t, err)
+
+			select {
+			case result := <-resultChan:
+				assert.Equal(t, operatorv1.ResourceType_COMPONENT, result.ResourceType)
+				assert.Equal(t, operatorv1.EventType_INIT, result.EventType)
+				assert.Equal(t, operatorv1.ResourceConditionStatus_SUCCESS, result.Condition)
+				assert.Equal(t, pubsubComponent.Name, result.Name)
+			case <-time.After(20 * time.Second):
```
JoshVanL, line 533: Hopefully computers are quicker than this.
```suggestion
			case <-time.After(5 * time.Second):
```

pkg/runtime/processor/processor_test.go
```diff
@@ -462,3 +487,170 @@ func TestProcessNoWorkflow(t *testing.T) {
...
+			err := proc.processComponentAndDependents(context.Background(), pubsubComponent)
+			require.Error(t, err)
+
+			select {
+			case result := <-resultChan:
+				assert.Equal(t, operatorv1.ResourceType_COMPONENT, result.ResourceType)
+				assert.Equal(t, operatorv1.EventType_INIT, result.EventType)
+				assert.Equal(t, operatorv1.ResourceConditionStatus_FAILURE, result.Condition)
+				assert.Equal(t, pubsubComponent.Name, result.Name)
+			case <-time.After(20 * time.Second):
```
JoshVanL, line 569:
```suggestion
			case <-time.After(5 * time.Second):
```

pkg/runtime/processor/processor_test.go
```diff
@@ -462,3 +487,170 @@ func TestProcessNoWorkflow(t *testing.T) {
...
+			err = proc.Close(pubsubComponent)
+			require.NoError(t, err)
+
+			select {
+			case result := <-resultChan:
+				assert.Equal(t, operatorv1.ResourceType_COMPONENT, result.ResourceType)
+				assert.Equal(t, operatorv1.EventType_CLOSE, result.EventType)
+				assert.Equal(t, operatorv1.ResourceConditionStatus_SUCCESS, result.Condition)
+				assert.Equal(t, pubsubComponent.Name, result.Name)
+			case <-time.After(20 * time.Second):
```
JoshVanL, line 612:
```suggestion
			case <-time.After(5 * time.Second):
```

pkg/runtime/processor/processor_test.go
```diff
@@ -462,3 +487,170 @@ func TestProcessNoWorkflow(t *testing.T) {
...
+			err = proc.Close(pubsubComponent)
+			require.Error(t, err)
+
+			select {
+			case result := <-resultChan:
+				assert.Equal(t, operatorv1.ResourceType_COMPONENT, result.ResourceType)
+				assert.Equal(t, operatorv1.EventType_CLOSE, result.EventType)
+				assert.Equal(t, operatorv1.ResourceConditionStatus_FAILURE, result.Condition)
+				assert.Equal(t, pubsubComponent.Name, result.Name)
+			case <-time.After(20 * time.Second):
```
JoshVanL, line 652:
```suggestion
			case <-time.After(5 * time.Second):
```

JoshVanL (changes requested):
Fine for now, but we should expect the reporter to be moved to the operator at some point in the future. It tracks that the `reporter` doesn't really belong to the `registry` at all, but again I think this is fine for now though consumers should keep this in mind.

pkg/runtime/processor/components.go
lrascao: checking

lrascao: seemed easy and added it, tagged you on a bit that i'm not totally sure

pkg/runtime/runtime.go
```diff
@@ -342,7 +343,7 @@ func newDaprRuntime(ctx context.Context,
 			for _, comp := range comps {
 				go func(comp compapi.Component) {
 					log.Infof("Shutting down component %s", comp.LogName())
-					errCh <- rt.processor.Close(comp)
+					errCh <- rt.processor.Close(ctx, comp)
```
lrascao, line 346: @JoshVanL here, easier to use `ctx` passed in to `newDaprRuntime` but perhaps it should be receiving it from the `AddCloser` func instead

JoshVanL: We can't use this context because it will have been cancelled when shutting down, meaning it won't actually send the close  report. I think instead we actually want to use a `context.Background()` with a `5s` timeout. Sorry!

JoshVanL: And to be clear, Close() should revert to not accept a context like before, and the Background be made just before the report call.

famarting:
> Fine for now, but we should expect the reporter to be moved to the operator at some point in the future. It tracks that the `reporter` doesn't really belong to the `registry` at all, but again I think this is fine for now though consumers should keep this in mind.

not to the registry, but I think it makes a lot of sense to be an abstraction within the sidecar runtime. Whether its implementation relates to the operator or its a custom function implementation is the key IMO

JoshVanL:
>be an abstraction within the sidecar runtime.
@famarting Why?

JoshVanL (approved):

famarting:
> > be an abstraction within the sidecar runtime.
> 
> @famarting Why?

because then its easy to provide a custom implementation for users that use dapr as a library

Issue #8284, opened 2024-11-17
Failed publishing to DLQ ignores errors


evhen14:
## In what area(s)?

/area runtime

## What version of Dapr?

1.14.x

## Expected Behavior

when pub/sub fails to send a message to DLQ, I expect it to get stuck and not proceed forward.

## Actual Behavior

if publishing to DLQ fails, the error is ignored and processing moves on to the next message.
I observed the error `error sending message to dead letter, origin topic` but wasn't able to trace down the message that caused it, however, looking at the Dapr runtime logic I concluded that the message will be lost.

The fix will potentially break some users' expectations, especially those that treat the DLQ as a black hole. For them, it might be more important to move to the next message, whereas in our case, it's more important to preserve all messages.

## Steps to Reproduce the Problem

1. Configure DLQ in dapr, but not in your message broker 
2. Publish a broken message
3. observe no message in DLQ and processing moved on to the next message

Related code: https://github.com/dapr/dapr/blob/master/pkg/runtime/subscription/subscription.go#L271-L290
In the first case, the warning is logged
in the second, the error is completely ignored.

## Release Note

RELEASE NOTE: **FIX** Bug in runtime: In some cases, if delivery to the dead letter queue fails, messages can be lost. The fix ensures no messages are lost.

PR #8291, opened 2024-11-19
do not ingore dlq errors

evhen14:
# Description

All failed DLQ attempts must be reported as an error. Before publishing to subscribers it does, after it ignores those. It leads to lost messages

## Issue reference

Please reference the issue this PR will close: #8284

## Checklist

- [x] Code compiles correctly
- [ ] Created/updated tests
- [x] Unit tests passing

### DIFF
--- a/pkg/runtime/subscription/subscription.go
+++ b/pkg/runtime/subscription/subscription.go
@@ -281,7 +281,8 @@ func New(opts Options) (*Subscription, error) {
 			}
 			return nil, pErr
 		})
-		if err != nil {
+		// when runtime shutting down, don't send to DLQ
+		if err != nil && err != context.Canceled {
 			// Sending msg to dead letter queue.
 			// If no DLQ is configured, return error for backwards compatibility (component-level retry).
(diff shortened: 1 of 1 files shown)


elena-kolevska (commented):
Thanks @evhen14!
If the change in behaviour is approved, we're going to need a test for this.
From my point of view, I prefer strictness, but we have to consider that this will cause a bad upgrade experience for some users, so it should be properly documented.

evhen14:
> Thanks @evhen14! If the change in behaviour is approved, we're going to need a test for this. From my point of view, I prefer strictness, but we have to consider that this will cause a bad upgrade experience for some users, so it should be properly documented.

What would be the best place to document it? I see that messages are normally not dropped when failing to send to DLQ (unmarshal, checking `IsRawPayload`, etc.); they only happen after sending to the subscribers.

passuied (approved):

yaron2 (approved):

marcduiker:
@holopin-bot @evhen14 Thank you!

PR #8304, opened 2024-11-21
use dapr_scheduler.ha to allow for scaling independently of the global.ha

cicoyle:
allow the use of `dapr_scheduler.ha` to scale independely of the global ha flag. This allows for global ha to be false and to scale the schedulers to 3 instances.

I confirmed the following to be true by testing all cases locally.

High level (flag breakdown): 

- scheduler enabled is required for schedulers to exist (default is true)
- global ha being true is fully respected
   - this mean, you cannot override the flag by setting the local ha flag to false. for example, if I have `--set global.ha.enabled=true` neither `--set dapr_scheduler.ha=false --set dapr_placement.ha=false` are respected meaning you are unable to be globally ha and have either of the placement or scheduler as a single instance. global being true overrides local being false (bc it defaults to false)
 - if global ha is false, you can have schedulers scaled to 3 instances by setting `--set dapr_scheduler.ha=true`


elena-kolevska (changes requested):
I think we should also support the case when global HA is true and scheduler ha is false.
Otherwise, lgtm.

cicoyle:
> I think we should also support the case when global HA is true and scheduler HA is false. For comparison, placement will allow HA when its own local HA is true, and the global is false. Otherwise, lgtm.

Hey @elena-kolevska! Thanks for the review! 

re: your comment: `I think we should also support the case when global HA is true and scheduler HA is false.`

Are you saying you want the local scheduler flag to override the global ha flag specifically for the case where `--set global.ha.enabled=true` & allow someone to have 1 scheduler by putting `--set dapr_scheduler.ha=false`? I intentionally didn't acct for this for Scheduler bc this is not how it works for placement. In my local testing, when I set `--set global.ha.enabled=true`, I was unable to have 1 instance of either scheduler/placement. Meaning, I could not override `--set global.ha.enabled=true` by setting `--set dapr_scheduler.ha=false --set dapr_placement.ha=false`

re: your comment: `For comparison, placement will allow HA when its own local HA is true, and the global is false. `. This works the same for scheduler with this PR change, meaning I did account for using the local scheduler ha flag to scale to 3 scheduler instances when global ha is false.

elena-kolevska:
Yeah, nevermind, I had commented relying just on the comment in the helm file itself without reading the PR comment, the behaviour is fully consistent with placement!

elena-kolevska (approved):

yaron2 (approved):

Issue #8245, opened 2024-10-28
Expose raft and placement service leadership status as a metric


jmprusi:
## In what area(s)?

> /area runtime

> /area operator

/area placement

> /area docs

> /area test-and-release

## Describe the feature

During a load-testing scenario, the placement server of DAPR ended in a non-consistent state. Workflows failed. To fix the issue, we had to remove the local disk state. (We couldn't reproduce this issue later in a similar environment, so take this with a grain of salt).

Unfortunately, the issue is that we had no indication that something was broken from outside the system, so this feature request is for a way to expose the raft consensus state as a metric (Prometheus) so we can debug/alert based on it.

## Release Note

RELEASE NOTE: **ADD** Exposed raft and placement service leadership status as metrics (`dapr_placement_leader_status` and `dapr_placement_raft_leader_status`).

Issue #7031, opened 2023-10-12
Placement server fails to disseminate placement tables.


MattCosturos:
## What version of Dapr?

1.11.2, running in AKS (aks version 1.25.5)

## Expected Behavior

When I update my application deployment, and new pods are started with new IPs, I expect the Placement Server to distribute the updated tables to all pods
We use a Helm chart to manage our application. Our Release pipeline does a `helm upgrade --namespace myNamespace --install --wait --debug --create-namespace  /path/to/myChart` to install the application.

## Actual Behavior

The Placement Server Starts the dissemination, but never completes, and never generates a timeout nor error.

Tail of logs shown here. We can see the last line is a `Start disseminating tables` but we never get a `Completed dissemination` nor do we get any error logs.

```
time="2023-10-10T19:38:39.299921228Z" level=info msg="Completed dissemination. memberUpdateCount: 2, streams: 37, targets: 8, table generation: 770" instance=dapr-placement-server-0 scope=dapr.placement type=log ver=1.11.2
time="2023-10-10T19:39:00.121337481Z" level=info msg="Start disseminating tables. memberUpdateCount: 2, streams: 30, targets: 6, table generation: 772" instance=dapr-placement-server-0 scope=dapr.placement type=log ver=1.11.2
time="2023-10-10T19:39:00.197528468Z" level=info msg="Completed dissemination. memberUpdateCount: 2, streams: 30, targets: 6, table generation: 772" instance=dapr-placement-server-0 scope=dapr.placement type=log ver=1.11.2
time="2023-10-11T20:53:09.620082431Z" level=info msg="Start disseminating tables. memberUpdateCount: 2, streams: 39, targets: 8, table generation: 774" instance=dapr-placement-server-0 scope=dapr.placement type=log ver=1.11.2
time="2023-10-11T20:53:09.805892521Z" level=info msg="Completed dissemination. memberUpdateCount: 2, streams: 39, targets: 8, table generation: 774" instance=dapr-placement-server-0 scope=dapr.placement type=log ver=1.11.2
time="2023-10-11T20:53:30.619910487Z" level=info msg="Start disseminating tables. memberUpdateCount: 2, streams: 30, targets: 6, table generation: 776" instance=dapr-placement-server-0 scope=dapr.placement type=log ver=1.11.2
[...]

PR #8326, opened 2024-12-02
Fix placement dissemination hang

elena-kolevska:
This PR addresses stuck dissemination when a stream hangs without returning an error. This can happen when an app becomes unresponsive and stops consuming messages from the stream, overflowing its buffer and resulting in a stuck. stream send operation on the placement service side.

Additional changes:
- Refactors the leadership change code for clarity, adding two new metrics: `dapr_placement_leader_status` and `dapr_placement_raft_leader_status`.
- It introduces a small refactor in the init code, bringing it closer to what we have in the scheduler service.

## Issue reference

- https://github.com/dapr/dapr/issues/7031
- https://github.com/dapr/dapr/issues/8245

## Checklist

- [X] Code compiles correctly
- [X] Created/updated tests
- [X] Unit tests passing
- [x] Extended the documentation / Created issue in the <https://github.com/dapr/docs/> repo: dapr/docs#_[issue number]_


pkg/placement/placement.go
```diff
@@ -467,16 +503,36 @@ func (p *Service) checkAPILevel(req *placementv1pb.Host) error {
 	return nil
 }
 
-func requiresUpdateInPlacementTables(req *placementv1pb.Host, isActorHost *bool) bool {
+func (p *Service) handleErrorOnStream(err error, hostName string, isActorHost *atomic.Bool, namespace string) error {
+	if errors.Is(err, io.EOF) || errors.Is(err, context.Canceled) {
+		log.Infof("Stream connection for app %s is disconnected gracefully: %s", hostName, err)
+	} else {
+		log.Infof("Stream connection for app %s is disconnected with the error: %v.", hostName, err)
```
antontroshin, line 510: Should this be `log.Errorf`? It would be easier to spot in logs and logging apps like Kibana, Loki, etc., when filtering by level.

pkg/placement/placement.go
```diff
@@ -176,14 +177,32 @@ type ServiceOpts struct {
...
+		o.MaxAPILevel = ptr.Of(uint32(apiLevel))
+	}
+}
+
+// New returns a new placement service.
+func New(opts ServiceOpts) *Service {
+	raftServer := raft.New(opts.Raft)
+	if raftServer == nil {
+		log.Fatal("Failed to create raft server.")
+	}
```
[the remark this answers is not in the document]
elena-kolevska: I updated the code so it fails with a proper error. 
It can't be retried because it's a user error in startup arguments.

pkg/placement/monitoring/metrics.go
```diff
@@ -39,6 +41,16 @@ var (
...
+	leaderStatus = stats.Int64(
+		"placement/leader_status",
+		"Placement server leadership status (1 for leader, 0 for not leader).",
+		stats.UnitDimensionless)
+
+	raftLeaderStatus = stats.Int64(
+		"placement/raft_leader_status",
+		"Raft server leadership status (1 for leader, 0 for not leader).",
+		stats.UnitDimensionless)
+
```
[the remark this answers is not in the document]
elena-kolevska: I had done this last week, per Alice's suggestion, but hadn't pushed. Thanks for catching it.

pkg/placement/placement.go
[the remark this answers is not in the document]
JoshVanL: Please don’t panic in the middle of runtime.

If the error is not recoverable, we should error exit the process. If we _really_ have to exit the process (?)

pkg/placement/leadership.go
```diff
@@ -31,85 +32,98 @@ const barrierWriteTimeout = 2 * time.Minute
...
-	)
-	close(loopNotRunning)
-
-	var leaderLoopCancel context.CancelFunc
-	var leaderCtx context.Context
+	raftLeaderCh := raft.LeaderCh()
+
+	startLeaderLoop := func() {
+		if leaderCtx != nil {
+			log.Error("attempted to start the leader loop while running", p.raftNode.GetID())
```
JoshVanL, line 48: Why don't we catch this error?

pkg/placement/leadership.go
```diff
@@ -31,85 +32,98 @@ const barrierWriteTimeout = 2 * time.Minute
...
+
+		leaderCtx, leaderCancelFn = context.WithCancel(ctx)
+
+		leaderLoopWg.Add(1)
+
+		go func(ctx context.Context) {
+			defer leaderLoopWg.Done()
+			err := p.leaderLoop(ctx)
+			if err != nil {
+				log.Error("placement leader loop failed", "error", err)
```
JoshVanL, line 60: Same here

pkg/placement/leadership.go
```diff
@@ -31,85 +32,98 @@ const barrierWriteTimeout = 2 * time.Minute
...
+		}(leaderCtx)
+		log.Info("raft cluster leadership acquired", p.raftNode.GetID())
+		return
+	}
+
+	stopLeaderLoop := func() {
+		if leaderCtx == nil || leaderCancelFn == nil {
+			log.Error("attempted to stop the leader loop while not running", p.raftNode.GetID())
+			return
+		}
```
JoshVanL, line 71: Why would this ever happen?

pkg/placement/membership.go
```diff
@@ -333,18 +332,46 @@ func (p *Service) disseminateOperation(ctx context.Context, target daprdStream,
 			return nil
 		}
 
-		if err := target.stream.Send(o); err != nil {
-			remoteAddr := "n/a"
-			if p, ok := peer.FromContext(target.stream.Context()); ok {
-				remoteAddr = p.Addr.String()
+		sendCtx, sendCancelFn := context.WithTimeout(ctx, 2*time.Second) // We should make this configurable
```
JoshVanL, line 335: This isn't a very long time, why 2 seconds?

JoshVanL (changes requested):

pkg/placement/membership.go
elena-kolevska: Yes, I wanted to discuss this value. I believe it should be configurable because it's a function of the underlying network as well as the number of dissemination messages we're expecting and how quickly can the sidecars process them; it can't be too long because it would imply we're locking all daprds for a long time (during dissemination). What would you suggest as a good default?

pkg/placement/leadership.go
elena-kolevska: Logically, we should be able to ignore this error. We would only get it if `p.raftNode.Raft(ctx)` returns an error, which only happens if the underlying raft server is not ready in time. But we would have never gotten the isLeader signal in the first palce if the raft server wasn't ready.

JoshVanL (approved):

JoshVanL:
@elena-kolevska tests failing

elena-kolevska:
> @elena-kolevska tests failing

Yeah, that's cause I reverted the error code fixes and put them in a separate PR. All the placement tests have been passing consistently for a few runs though. I just merged latest release, let's see if they pass.

JoshVanL (approved):

PR #8333, opened 2024-12-05
Integration: Test for default actor reminder store

JoshVanL:


tests/integration/suite/actors/reminders/store.go
```diff
@@ -0,0 +1,80 @@
...
+	_, err := s.daprd.GRPCClient(t, ctx).RegisterActorReminder(ctx, &rtv1.RegisterActorReminderRequest{
+		ActorType: "myactortype",
+		ActorId:   "myactorid",
+		Name:      "remindermethod",
+		DueTime:   "100s",
+		Data:      []byte("reminderdata"),
+	})
+	require.NoError(t, err)
+
+	assert.Len(t, s.scheduler.EtcdJobs(t, ctx), 1)
```
cicoyle, line 79: Does that need to be wrapped in an assert eventually- it's failing on windows/mac... Or was this the test proving the default to true wasn't working?

JoshVanL: It was to prove this wasn't working. This is now fixed in https://github.com/dapr/dapr/pull/8334 so this can be closed.

PR #8346, opened 2024-12-11
Centralises JobsAPI errors and gRPC error codes

elena-kolevska:
We're in the process of centralising all errors and error codes as much as possible for easier metrics, documentation and general code simplicity.
This PR centralises all the error codes for the Jobs API in a single file where all the errors are defined for easy parsing and docs generation.
In addition, we're adding a gRPC code field to the ErrorCode struct, so we can track all the possible gRPC errors as well.

Task list:
- [x] Add error codes for the Jobs API
- [x] Centralise all gRPC error codes
- [x] Add the description of every error as an inline comment
- [x] Create a parser script that will convert the errors file into a markdown table for docs. (Will sent as PR in the docs repo, probably)
- [x] Updates Jobs API to follow [Dapr error codes convention](https://github.com/dapr/proposals/blob/main/20230511-BCIRS-error-handling-codes.md#errorinfo-required). This breaks compatibility with 1.14 because it changes API error codes, but the API is still in Alpha, so this is the time to do it. I checked the SDKs that support this feature and none of them seem to be relying on these codes.

This is how an autogenerated table would look like. No value in the gRPC code field means we don't have a rich gRPC error for that code yet.
Screenshot 2024-12-12 at 15 18 13


jake-engelberg (commented):
I don't see any non-testing usage of `NotFound()`, `NotConfigured()`, or `IncorrectNegative()`.

Should we remove them to reduce confusion upon looking through this file?

elena-kolevska:
> I don't see any non-testing usage of NotFound(), NotConfigured(), or IncorrectNegative().
Should we remove them to reduce confusion upon looking through this file?

Yeah, they're not used. Let's delete them, since we're here. 👍

pkg/api/errors/errors.go
```diff
@@ -23,13 +23,6 @@ import (
 	kiterrors "github.com/dapr/kit/errors"
 )
 
-const (
-	InFixName     ReasonSegment = "NAME_"
-	InFixNegative ReasonSegment = "NEGATIVE_"
-	PostFixName   ReasonSegment = "NAME"
-	PostFixEmpty  ReasonSegment = "EMPTY"
-)
-
```
elena-kolevska, line 32: Deleting because they're not used anywhere.

pkg/api/errors/errors.go
```diff
@@ -44,42 +37,15 @@ func NotFound(name string, componentType string, metadata map[string]string, grp
...
-		httpCode,
-		message,
-		legacyTag,
-		string(category),
-	).
-		WithErrorInfo(reason, metadata).
-		Build()
-}
-
-func Empty(name string, metadata map[string]string, reason string, category errorcodes.Category) error {
```
elena-kolevska, line 61: Deleting because they're not used anywhere.

pkg/api/errors/errors.go
```diff
@@ -44,42 +37,15 @@ func NotFound(name string, componentType string, metadata map[string]string, grp
 		Build()
 }
 
-func NotConfigured(name string, componentType string, metadata map[string]string, grpcCode codes.Code, httpCode int, legacyTag string, reason string, category errorcodes.Category) error {
```
elena-kolevska, line 47: Deleting because not used anywhere.

pkg/api/errors/errors.go
```diff
@@ -44,42 +37,15 @@ func NotFound(name string, componentType string, metadata map[string]string, grp
...
 		message,
 		"",
-		string(category),
+		string(errorCode.Category),
 	).
-		WithErrorInfo(reason, metadata).
-		Build()
-}
-
-func IncorrectNegative(name string, metadata map[string]string, reason string, category errorcodes.Category) error {
```
elena-kolevska, line 74: Deleting because not used anywhere.

pkg/api/errors/scheduler.go
```diff
@@ -22,86 +22,63 @@ import (
...
-	InFixSchedule       ReasonSegment = "SCHEDULE_"
-	PostFixRepeats      ReasonSegment = "REPEATS"
-	PostFixJob          ReasonSegment = "JOB"
-	PostFixJobs         ReasonSegment = "JOBS"
-
-	MsgScheduleJob = "failed to schedule job"
-	MsgGetJob      = "failed to get job"
-	MsgListJobs    = "failed to list jobs"
-	MsgDeleteJob   = "failed to delete job"
-)
```
elena-kolevska, line 41: Moved to `pkg/messages/errorcodes/errorcodes.go` to align with other error codes

elena-kolevska:
FYI @msfussell

elena-kolevska:
@jake-engelberg  We'll have to update this once [your PR](https://github.com/dapr/dapr/pull/8345) gets merged.

pkg/messages/errorcodes/errorcodes.go
```diff
@@ -42,108 +43,124 @@ func (e *ErrorCode) Error() string {
...
+	CommonAppChannelNil        = ErrorCode{"ERR_APP_CHANNEL_NIL", "", CategoryCommon}        // App channel is nil
+	CommonBadRequest           = ErrorCode{"ERR_BAD_REQUEST", "", CategoryCommon}            // Bad request
+	CommonBodyRead             = ErrorCode{"ERR_BODY_READ", "", CategoryCommon}              // Error reading request body
+	CommonInternal             = ErrorCode{"ERR_INTERNAL", "", CategoryCommon}               // Internal error
+	CommonMalformedRequest     = ErrorCode{"ERR_MALFORMED_REQUEST", "", CategoryCommon}      // Malformed request
+	CommonMalformedRequestData = ErrorCode{"ERR_MALFORMED_REQUEST_DATA", "", CategoryCommon} // Malformed request data
+	CommonMalformedResponse    = ErrorCode{"ERR_MALFORMED_RESPONSE", "", CategoryCommon}     // Malformed response
+
+	// ### Scheduler/Jobs API
+	SchedulerScheduleJob   = ErrorCode{"SCHEDULER_SCHEDULE_JOB", "SCHEDULER_SCHEDULE_JOB", CategoryJob}     // Error scheduling job
```
cicoyle, line 155: I think this is what you are wanting to prefix `DAPR_` with - correct?

```suggestion
	SchedulerScheduleJob   = ErrorCode{"SCHEDULER_SCHEDULE_JOB", "DAPR_SCHEDULER_SCHEDULE_JOB", CategoryJob}     // Error scheduling job
```

cicoyle: We should probably add a string check to kit to check if `DAPR_` is prefixed and if not to add it to avoid this going forward. This was missed bc the const lives in runtime atm, but all the others live in kit and are prefixed in kit. I think I assumed that prefix was added to the const when I added it to runtime

cicoyle: Since scheduler isn't stable 'yet', my vote would be update this to include the dapr prefix. Elena, could you go ahead with this since you have this PR open?

elena-kolevska: Yes, definitely! Since we're still in Alpha now is the right time to change this.

jake-engelberg (approved):

JoshVanL (approved):

PR #8347, opened 2024-12-13
Remove vnodes metadata from Placement

elena-kolevska:
In 1.13 we removed the passing of generated virtual nodes between the placement and daprd and started calculating them on the daprd side. We kept the feature behind a context metadata flag for backwards compatibility and marked it to be completely removed in 1.15.
This PR completely removes the calculation and passing of vnodes on the placement side because 1.13+ sidecars are capable of calculating them on their side.


pkg/placement/membership.go
```diff
@@ -226,23 +225,10 @@ func (p *Service) performTableDissemination(ctx context.Context, ns string) erro
 		streams = append(streams, *stream)
 	})
 
-	// Check if the cluster has daprd hosts that expect vnodes;
-	// older daprd versions (pre 1.13) do expect them, and newer versions (1.13+) do not.
 	req := &tablesUpdateRequest{
 		hosts: streams,
 	}
-	// Loop through all streams and check what kind of tables to disseminate (with/without vnodes)
```
elena-kolevska, line 234: We don't need to disseminate tables with vnodes anymore, so we're removing this.

JoshVanL (approved):
Thanks @elena-kolevska 

we seem to be hitting a catastrophic panic in macos int tests, but I don’t believe it is related to this PR.

PR #8351, opened 2024-12-17
[1.14] update contrib to 1.14.5

yaron2:
Update Contrib to 1.14.5 to include a fix for a Kafka edge case during shutdowns.


artursouza (dismissed):

elena-kolevska (commented):
There's some lint errors

artursouza (approved):

PR #8355, opened 2024-12-20
sentry: allow for self-hosted CA on Kubernetes

lrascao:
# Description

Currently, any user running on kubernetes interesting on running a standalone Sentry is forced to provide both a `dapr-trust-bundle` Secret and ConfigMap. This is not ideal in the cases where you're generating the private key pair from eg. cert-manager and distributing trust in the cluster using trust-manager. On these use-cases it's simply enough to mount these resources and supply the correct CLI args.

## Checklist

- [X] Code compiles correctly
- [X] Created/updated tests
- [X] Unit tests passing
- [X] End-to-end tests passing

### DIFF
--- a/cmd/sentry/options/options.go
+++ b/cmd/sentry/options/options.go
@@ -14,12 +14,14 @@ limitations under the License.
 package options
 
 import (
+	"fmt"
 	"path/filepath"
 
 	"github.com/spf13/pflag"
 	"k8s.io/client-go/util/homedir"
 
 	"github.com/dapr/dapr/pkg/metrics"
+	"github.com/dapr/dapr/pkg/modes"
 	"github.com/dapr/dapr/pkg/sentry/config"
 	"github.com/dapr/kit/logger"
 )
@@ -43,6 +45,7 @@ type Options struct {
 	Kubeconfig            string
 	Logger                logger.Options
 	Metrics               *metrics.FlagOptions
+	Mode                  string
 
 	RootCAFilename     string
 	IssuerCertFilename string
@@ -81,6 +84,7 @@ func New(origArgs []string) *Options {
 	fs.StringVar(&opts.ListenAddress, "listen-address", "", "The listen address for the sentry server")
 	fs.IntVar(&opts.HealthzPort, "healthz-port", 8080, "The port for the healthz server to listen on")
 	fs.StringVar(&opts.HealthzListenAddress, "healthz-listen-address", "", "The listening address for the healthz server")
+	fs.StringVar(&opts.Mode, "mode", string(modes.StandaloneMode), "Runtime mode for Dapr Sentry")
 
 	if home := homedir.HomeDir(); home != "" {
 		fs.StringVar(&opts.Kubeconfig, "kubeconfig", filepath.Join(home, ".kube", "config"), "(optional) absolute path to the kubeconfig file")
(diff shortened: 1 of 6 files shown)


cmd/sentry/options/options.go
```diff
@@ -81,6 +82,7 @@ func New(origArgs []string) *Options {
 	fs.StringVar(&opts.ListenAddress, "listen-address", "", "The listen address for the sentry server")
 	fs.IntVar(&opts.HealthzPort, "healthz-port", 8080, "The port for the healthz server to listen on")
 	fs.StringVar(&opts.HealthzListenAddress, "healthz-listen-address", "", "The listening address for the healthz server")
+	fs.BoolVar(&opts.SelfHostedCA, "self-hosted-ca", false, "Runs the CA in self-hosted mode, ie. only arguments provided through command line are considered")
```
JoshVanL, line 85: Please can we use the same “mode” option used for other services?

JoshVanL (changes requested):

JoshVanL:
@lrascao please rebase against release-1.15. You likely need to do a cherry pick as “change base” is goofy in GH UI.

cmd/sentry/options/options.go
JoshVanL: We should favour consistency over the confusing here (though I don’t think it’s super confusing- it’s self hosting your own PKI vs using kubernetes resources as the backend).

lrascao: updated, now takes `--mode`, defaults to `KuberrnetesMode` to keep existing behaviour.

cmd/sentry/options/options.go
```diff
@@ -81,6 +83,7 @@ func New(origArgs []string) *Options {
 	fs.StringVar(&opts.ListenAddress, "listen-address", "", "The listen address for the sentry server")
 	fs.IntVar(&opts.HealthzPort, "healthz-port", 8080, "The port for the healthz server to listen on")
 	fs.StringVar(&opts.HealthzListenAddress, "healthz-listen-address", "", "The listening address for the healthz server")
+	fs.StringVar(&opts.Mode, "mode", string(modes.KubernetesMode), "Runtime mode for Dapr Sentry")
```
JoshVanL, line 86: Default mode should be selfhosted, and set to kubernetes in the helm chart. We should check mode is either self hosted or kube else error.

JoshVanL (changes requested):

cmd/sentry/options/options.go
JoshVanL: Default mode should be self hosted else this breaks container environments

cmd/sentry/options/options.go
lrascao: right, the helm chart is the one that sets this, that makes sense

JoshVanL: @lrascao please add this check ^

JoshVanL: I’m also assuming some int tests will fail as they were relying on the old “IsKubernstesHosted” method of checking, and so now instead we need to supply this flag. Could be wrong tho

lrascao: > @lrascao please add this check ^

check is being done in `opts.Validate`, unless you had some other in mind?

JoshVanL: Missed it, was expecting it in the options.New func

JoshVanL:
@lrascao lint is failing and need DCO signoff

lrascao:
> @lrascao lint is failing and need DCO signoff

yeah, noticed the lint failure and pushed a66ce94f9c8f3567d23e84de559f5a2ead99a47d, seems CI didn't re-run
To fix the DCO i'll need to force push, is that ok?

JoshVanL:
@lrascao indeed, you need to re-write history so that all commits have DCO

lrascao:
> @lrascao indeed, you need to re-write history so that all commits have DCO

squashed and force-pushed

PR #8361, opened 2025-01-06
Adds WatchHosts mechanism for daprd -> scheduler

JoshVanL:
Currently, for daprd to connect to every scheduler istance it must have the full set of addresses on start. This can be problematic since all daprds need to be restarted when changing daprd to ha mode or visa versa. When the address list is misaligned, daprds will either _not_ receive jobs which are managed by an unconnected instance, or, daprd will spam logs trying to connect to scheduler instance which does not exist.

By using the dynamic leadership feature add to go-etcd-cron, we now store the host:port of scheduler instances at their leadership keys, which are then shared to all scheduler instances. This means all schedulers know the addresses of all _live_ active scheduler instances. Using a new `WatchHosts` gRPC stream RPC, scheduler broadcasts all scheduler hosts to connected daprd instances.

daprds will receive updated lists of active scheduler addresses, meaning that we handle full scale up/scale down scenarios with this service discovery. Note that daprd need only connect to _1_ scheduler instance on this RPC. The scheduler server API Kubernetes service has therefore changed to a simple Cluster IP so that DNS returns the IPs of active scheduler pods.

We no longer have a replica-count CLI flag for scheduler, since go-etcd-cron now self-manages partition total and ID. The cron ID is now the ETCD ID which is the pod name.

We use a new environment variable to set the single scheduler Cluster IP service address `DAPR_SCHEDULER_HOST_ADDRESS_DNS_A`. We must use a new env var rather than using the existing for backwards compat- older versions of daprd would no longer connect to all hosts(!).

The call to `WatchHosts` is effectively wrapping the existing logic which handles connecting & reconnecting to each scheduler instance.


pkg/runtime/scheduler/scheduler.go
```diff
@@ -36,183 +42,222 @@ type Options struct {
...
+		}
 	}
+}
 
-	req.GetInitial().AcceptJobTypes = acceptJobTypes
+func (s *Scheduler) connSchedulerHosts(ctx context.Context) (schedulerv1pb.Scheduler_WatchHostsClient, []string, error) {
+	log.Infof("Attempting to connect to scheduler: %s", s.addresses[0])
 
-	clients, err := m.clients.All(ctx)
+	cl, err := client.New(ctx, s.addresses[0], s.security)
```
cicoyle, line 223: Can you add a comment for future reference that mentions this LB under the hood due to the clusterIP?

pkg/scheduler/server/internal/cron/cron.go
```diff
@@ -50,29 +53,38 @@ type Interface interface {
 	// framework and database. Blocks until Etcd and the Cron library are ready.
 	Client(ctx context.Context) (api.Interface, error)
 
-	// AddWatch adds a watch for jobs to the connection pool.
-	AddWatch(*schedulerv1pb.WatchJobsRequestInitial, schedulerv1pb.Scheduler_WatchJobsServer) error
+	// AddJobsWatch adds a watch for jobs to the connection pool.
+	AddJobsWatch(*schedulerv1pb.WatchJobsRequestInitial, schedulerv1pb.Scheduler_WatchJobsServer) error
```
cicoyle, line 57: For consistency, can it be?
```suggestion
	JobsWatch(*schedulerv1pb.WatchJobsRequestInitial, schedulerv1pb.Scheduler_WatchJobsServer) error
```

pkg/scheduler/server/internal/cron/cron.go
```diff
@@ -108,28 +120,70 @@ func (c *cron) Run(ctx context.Context) error {
 		return err
 	}
 
+	ch := make(chan []*anypb.Any)
```
cicoyle, line 123: nit for readability:
```suggestion
	WatchLeadershipCh := make(chan []*anypb.Any)
```

cicoyle (commented):
few minor comments, overall lgtm

yaron2:
E2E is failing on the Build/Push stage

pkg/scheduler/server/internal/cron/cron.go
JoshVanL: :+1:

go.mod
```diff
@@ -506,3 +506,5 @@ replace github.com/stretchr/testify => github.com/stretchr/testify v1.8.4
 //
 // Then, run `make modtidy-all` in this repository.
 // This ensures that go.mod and go.sum are up-to-date for each go.mod file.
+
+replace github.com/dapr/kit => ../kit
```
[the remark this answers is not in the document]
JoshVanL: Still waiting for https://github.com/dapr/kit/pull/111

JoshVanL:
@yaron2 We need to merge https://github.com/dapr/kit/pull/111

pkg/runtime/scheduler/scheduler.go
```diff
@@ -36,183 +42,224 @@ type Options struct {
...
+				addresses = make([]string, 0, len(resp.GetHosts()))
+				for _, host := range resp.GetHosts() {
+					addresses = append(addresses, host.GetAddress())
+				}
+
+				log.Infof("Received updated scheduler hosts addresses: %v", addresses)
+
+				return nil
+			},
+		).Run(ctx)
```
elena-kolevska, line 123: I don't think I understand this block, how do we ensure `s.connectClients` will run after we've received new addresses? Don't we have a potential race condition on accessing `addresses`?

elena-kolevska: Ok, nevermind, I see it now, `s.connectClients` won't return until there's an error on the stream.
Would be nice to trigger the reconnect automatically without reconnecting, but I see you already have a TODO for that.

elena-kolevska (approved):

PR #8364, opened 2025-01-08
Workflow: gracefully recover from legacy state read errors

JoshVanL:
Currently, after upgrading from Dapr v1.14 to v1.15 and executing a workflow which has state from v1.14, the workflow state will fail to parse and cause the workflow to error. This is due to the storage API changing moving to v1 GA.

To resolve this, we fall back to parsing the workflow metadata key as JSON, then in the case of event history, ignoring the existing state if it fails to parse. Practically, this means a mid-running workflow will be abandoned when upgrading to v1.15 which seems acceptable. The other approach to this would be for use intervention to manually clean up the actor state store which is unreasonable.

cc @olitomlinson

### DIFF
--- a/pkg/runtime/wfengine/state/state.go
+++ b/pkg/runtime/wfengine/state/state.go
@@ -15,6 +15,7 @@ package state
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"strings"
 	"time"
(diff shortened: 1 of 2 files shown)


olitomlinson:
@JoshVanL 

> Practically, this means a mid-running workflow will be abandoned

Does this mean the workflow instance would be given a first-class terminal status of `FAILED` ?

olitomlinson:
> @olitomlinson it will be as if the state doesn't exist at all

So querying for the workflow status would give a 404? (as though the workflow never existed)

elena-kolevska (approved):

yaron2 (approved):

PR #8369, opened 2025-01-10
Workflow: Stream runtime status to durabletask, rather than poll

JoshVanL:
Instead of constituently polling for the runtime status to change state to observe a different phase of a workflow, durabletask now streams the state, whereby updates are only sent when they occur. Updates dapr to support this.

A new "generic" `CallActorStream` internal Dapr proto has been added which implements a internal invoke request stream that receives invoke responses. This RPC is used to stream the runtime status of a workflow which is a hosted on another dapr host, i.e. a remote running workflow.

Creates `CallStream` will child local and remote calls to the target actor to implement the new remote RPC, as well as the entry point for the workflow actor backend implementation. Only the "workflow" actor target implements this RPC to watch for runtime state changes.

Upon invoke stream invocation, the workflow actor target will first lock the actor to ensure we get the current orchestration metadata state, then adds a subscription to metadata changes, sends the current state, then unlocks so that the workflow can continue to execute. Upon changes, the client will receive changes.

Relies on kit and durabletask PR changes.

# Description

## Issue reference

## Checklist

- [ ] Code compiles correctly
- [ ] Created/updated tests
- [ ] Unit tests passing


yaron2:
https://github.com/dapr/durabletask-go/pull/8 is merged, the dependency can be updated

famarting (approved):

famarting:
go-sdk will need to be updated with this same durabletask-go update

yaron2 (approved):

PR #8380, opened 2025-01-14
Integration: Fix

JoshVanL:


antontroshin (dismissed):

elena-kolevska (approved):

elena-kolevska (dismissed):

cicoyle (approved):

antontroshin (approved):

tests/integration/suite/actors/reminders/scheduler/remote.go
```diff
@@ -74,7 +74,7 @@ spec:
   - name: SchedulerReminders
     enabled: true`), 0o600))
 
-	r.actorIDsNum = 500
+	r.actorIDsNum = 400
```
elena-kolevska, line 77: I think this was related to the app closing before daprd deactivated all actors. Lowering the number here works, but we might have to update the framework to account for this in the future.
I started a draft PR for it here: https://github.com/dapr/dapr/pull/8362/files that controls the app closing through the test

antontroshin: @elena-kolevska I've also tried something similar, but in my case, it still did not work when trying a larger number of actors, for example, 3000.
What did work, is to add the retry on actor deactivation. See this PR https://github.com/dapr/dapr/pull/8377

PR #8387, opened 2025-01-17
Workflow: fixup

JoshVanL:
Make GetActorState not dependent hosted so workflow doesn't need to be executing to get terminal status.

Create placement lock which doesn't deadlock on shutdown or workflow execution.

Ensure scheduler shutdown doesn't cause deadlock.

Update durabletask to use runtime status proto


famarting (approved):

yaron2:
Conflict after latest merge

cicoyle (approved):

yaron2 (approved):

PR #8395, opened 2025-01-21
Scheduler: ETCD mTLS

JoshVanL:
Updates scheduler for ETCD to run with mTLS between hosts. Updates sentry to make dapr-scheduler identity certificates to be signed with `dapr-scheduler-server.${ns}.svc` DNS. Updates ETCD embeded config to use mTLS on this DNS when mTLS is enabled.

Since etcd requires the identity certs to be a file reference, we write them to the temporary file dir.

Removes Etcd Client HTTP port.

Uses https://github.com/dapr/kit/pull/112


charts/dapr/charts/dapr_scheduler/templates/_helpers.tpl
```diff
@@ -33,7 +33,7 @@ Create initial cluster peer list dynamically based on replicaCount.
 {{- range $i, $e := until $replicaCount -}}
 {{- $instanceName := printf "dapr-scheduler-server-%d" $i -}}
 {{- $svcName := printf "%s.dapr-scheduler-server.%s.svc.cluster.local" $instanceName $namespace -}}
-{{- $peer := printf "%s=http://%s:%d" $instanceName $svcName (int $.Values.ports.etcdGRPCPeerPort) -}}
+{{- $peer := printf "%s=https://%s:%d" $instanceName $svcName (int $.Values.ports.etcdGRPCPeerPort) -}}
```
cicoyle, line 36: Doesn't this need to be wrapped around a conditional on if mtls is enabled?

JoshVanL: @cicoyle no because we always do mtls on the control plane in Kubernetes environments

pkg/sentry/server/server.go
```diff
@@ -184,6 +184,8 @@ func (s *Server) signCertificate(ctx context.Context, req *sentryv1pb.SignCertif
 		dns = []string{fmt.Sprintf("dapr-sidecar-injector.%s.svc", req.GetNamespace())}
 	case req.GetNamespace() == security.CurrentNamespace() && req.GetId() == "dapr-operator":
 		dns = []string{fmt.Sprintf("dapr-webhook.%s.svc", req.GetNamespace())}
+	case req.GetNamespace() == security.CurrentNamespace() && req.GetId() == "dapr-scheduler":
+		dns = []string{fmt.Sprintf("dapr-scheduler-server.%s.svc", req.GetNamespace())}
```
elena-kolevska, line 188: Should we make this `dapr-scheduler-server-a` to match what we did in [this pr](https://github.com/dapr/dapr/pull/8401/files#diff-4d6aae0e9704c4ff709a3ff0afaa2d61e740bf745d2fafdece7bc859381aac40)?

JoshVanL: The ETCD DNS lookup fails with `-a` so we need to use this service. This means we will always need both Services now- this is fine.

elena-kolevska (approved):

cicoyle (approved):

cicoyle:
actually - @JoshVanL I saw there is a random `foo` file that can get deleted I think :)

JoshVanL:
Thanks @cicoyle! Fixed

cicoyle (approved):

PR #8399, opened 2025-01-21
Improve test stability

cicoyle:
e2e test tweaks & cleanup and fix some flaky ints


tests/e2e/actor_features/actor_features_test.go
```diff
@@ -382,7 +382,7 @@ func TestActorFeatures(t *testing.T) {
 		req := actorReminderOrTimer{
 			Data:    "reminderdata",
 			DueTime: "1s",
-			Period:  "10s",
+			Period:  "20s",
```
JoshVanL, line 385:
```suggestion
			Period:  "10s",
```

This changes what the test is testing. This is fixed in https://github.com/dapr/dapr/pull/8398

pkg/runtime/wfengine/state/state.go
```diff
@@ -326,22 +329,28 @@ func LoadWorkflowState(ctx context.Context, state state.Interface, actorID strin
 	for i := range metadata.GetInboxLength() {
 		key = getMultiEntryKeyName(inboxKeyPrefix, i)
 		if bulkRes[key] == nil {
-			return nil, fmt.Errorf("failed to load inbox state key '%s': not found", key)
+			// If an inbox entry is missing, it means we caught the state in between saves.
+			// Return the state we have so far. This likely happened due to the reminder firing
+			// before all the inbox entries from the previous transactions went thru in the state store
+			wfLogger.Warnf("Inbox entry '%s' not found, state may be partially saved", key)
+			break
```
JoshVanL, line 336: Why are we doing this? We have to fail here as the workflow is corrupt??

pkg/runtime/wfengine/state/state.go
```diff
@@ -326,22 +329,28 @@ func LoadWorkflowState(ctx context.Context, state state.Interface, actorID strin
...
 		if bulkRes[key] == nil {
 			return nil, fmt.Errorf("failed to load history state key '%s': not found", key)
 		}
-		wState.History[i] = &protos.HistoryEvent{}
-		if err = proto.Unmarshal(bulkRes[key], wState.History[i]); err != nil {
+		event := &protos.HistoryEvent{}
+		if err = proto.Unmarshal(bulkRes[key], event); err != nil {
 			return nil, fmt.Errorf("failed to unmarshal history event from history state key '%s': %w", key, err)
 		}
+		wState.History = append(wState.History, event)
```
JoshVanL, line 353: What's the purpose of this change?

JoshVanL (changes requested):

tests/e2e/actor_features/actor_features_test.go
cicoyle: thats a diff test, this diff is from the 3xx lines, the diff from the linked PR impacts the 1xx lines. I dont think its the same issue

JoshVanL: I think you really are changing the test behaviour though, and we shouldn't. Please revert.

pkg/runtime/wfengine/state/state.go
cicoyle: I was investigating a race condition with the inbox/metadata not being found and was trying this for a fix. If this is intentional meaning the wf is corrupt I can remove the diff and look elsewhere

tests/e2e/actor_features/actor_features_test.go
cicoyle: I increased the period bc there was flakiness in CI, but not locally. I believe the reminder was firing back due to the 10s and CI being slower, I put 20s just to give CI more time. The period being upped should not impact the intended behavior of the test, but accounts for a slower env in CI. I can revert it since you seem adamant about it, but if it is flaky then I suggest upping the time a little. I understand why you don't want it upped to 20s since we 'reset' it to the 20s

cicoyle:
I still cant repro this issue locally. I've tried a few diff changes, but am having a hard time fixing it bc I can't debug it locally. This test passes locally

artursouza (dismissed):

elena-kolevska (commented):
lgtm, but had one question in the comments

pkg/actors/api/reminder.go
```diff
@@ -76,14 +76,15 @@ func (r Reminder) RepeatsLeft() int {
 func (r *Reminder) TickExecuted() (done bool) {
 	if r.Period.repeats > 0 {
 		r.Period.repeats--
-	}
```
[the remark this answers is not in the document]
cicoyle: There is a check if `reminder.RepeatsLeft() == 0` in the `statestore.go` code before the `CallReminder` [call here](https://github.com/dapr/dapr/blob/release-1.15/pkg/actors/internal/reminders/storage/statestore/statestore.go#L1061-L1067) that ensures we never call a reminder with no repeats left. With that and this change we are handling all cases properly due to the finite repeats getting stopped once they decrement to 0 and the infinite repeats (noted with a -1) will continue as expected.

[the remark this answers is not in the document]
cicoyle: Let me investigate the unit tests

cicoyle:
Looks like the actor reminder state store triggering the wrong number of times is resolved as of the last e2e run, but others fail. Something appears to be off in the actor system in general

JoshVanL (approved):

elena-kolevska (approved):

PR #8401, opened 2025-01-22
Two scheduler services for backwards compatibility

elena-kolevska:
We saw errors upgrading from 1.14 to 1.15 where the scheduler service, which used to be headless is now clusterIP.
The errors were due to the field `spec.clusterIP` being immutable, so the existing service wasn't properly updated.

This PR creates the new ClusterIP scheduler service under a new name, and keeps the old service for backwards compatibility.


charts/dapr/charts/dapr_scheduler/templates/dapr_scheduler_service.yaml
```diff
@@ -44,3 +44,49 @@ spec:
   {{- end}}
   type: ClusterIP
 {{- end }}
+---
+{{- if (eq .Values.global.scheduler.enabled true) }}
```
elena-kolevska, line 48: This is the old service, exactly as we used to have it.

JoshVanL, line 48: Please can you add a TODO to remove this Service in the next version?

JoshVanL (changes requested):
The integration tests are failing on helm tests.

yaron2 (approved):

PR #8403, opened 2025-01-23
Fix Scheduler address in standalone mode

antontroshin:
# Description

This PR attempts to fix the issue in Dapr runtime when used in standalone mode and it tries to connect to the scheduler but receives an error on the internal `WatchHosts` rpc call.
It returns internal Docker addresses that are inaccessible from the host machine.
In this case, will use supplied addresses via `--scheduler-host-address` argument, or if it is not supplied, use default local address `:50006`.

## Issue reference

## Checklist

- [ ] Code compiles correctly
- [ ] Created/updated tests
- [ ] Unit tests passing


pkg/runtime/config.go
```diff
@@ -247,6 +243,11 @@ func FromConfig(ctx context.Context, cfg *Config) (*DaprRuntime, error) {
 		}
 	}
 
+	// Add local scheduler address if not provided and reminders are enabled (which they are by default)
+	if intc.mode == modes.StandaloneMode && globalConfig.IsFeatureEnabled(config.SchedulerReminders) && len(intc.schedulerAddress) == 0 {
```
yaron2, line 247: For 1.15 reminders are enabled by default so shouldn't need the check

yaron2 (changes requested):

pkg/runtime/config.go
elena-kolevska: Would this disrespect explicitly turned off reminders, further down the stack?

pkg/runtime/config.go
```diff
@@ -247,6 +243,11 @@ func FromConfig(ctx context.Context, cfg *Config) (*DaprRuntime, error) {
 		}
 	}
 
+	// Add local scheduler address if not provided in standalone mode
+	if intc.mode == modes.StandaloneMode && len(intc.schedulerAddress) == 0 {
+		intc.schedulerAddress = append(intc.schedulerAddress, ":50006")
```
JoshVanL, line 248: Why are we appending the address, rather than just setting it to this?

paulyuk:
Hey @antontroshin will this fix https://github.com/dapr/dapr/issues/8412 or is that likely a different issue?

antontroshin:
The latest CLI RC2 has this issue fixed for quickstarts https://github.com/dapr/cli/pull/1475
The issue may be experienced on local machine when the version does not pass the "edge"/version check and parameter is being removed.
Added this PR (https://github.com/dapr/cli/pull/1479), to add "dev" version tag to this check.

pkg/runtime/scheduler/scheduler.go
```diff
@@ -254,6 +258,13 @@ func (s *Scheduler) connSchedulerHosts(ctx context.Context) (schedulerv1pb.Sched
 		addresses = append(addresses, host.GetAddress())
 	}
 
+	// Let the code above run to this point to validate the connection and WatchHosts RPC call
+	// If we are in standalone mode and only one address is provided, we need to return the address as provided in config
+	// Otherwise the resolved address from response will have the address of the docker network
+	if len(s.addresses) > 0 && s.mode == modes.StandaloneMode {
+		addresses = s.addresses
+	}
```
JoshVanL, line 266: We can't do this- this is going to fail in standalone mode on an existing deployment when the scheduler replicas are scaled up (or down). This is one of the main reasons for this API existing in the first place. Not least, daprd side hacks like these are super fragile and can back us into a corner long term.

If scheduler is listening on localhost and this is inaccessible to the host machine, how is scheduler reachable at all? If it's listening on another interface, then that IP too should be reachable from the host machine and the existing logic in Scheduler to use the right address to broadcast should also work.

If we *must must* do something here, I think it best to add a `--override-broadcast-address` CLI flag on the Scheduler side.

JoshVanL (changes requested):

antontroshin:
Superseded by https://github.com/dapr/dapr/pull/8416

PR #8404, opened 2025-01-23
Workflow: event channel size 1

JoshVanL:
Make workflow internal event channel a size of 1 to ensure we don't block execution on a busy workflow queue.


elena-kolevska:
I don't know the new workflow code well, so I'll just ask here: why are we going with a size of 1? Who are the contenders for sending? If it's more than two, would a bigger buffer make sense?

JoshVanL:
@elena-kolevska this ensures there is no circular dependency of different workers waiting for each other I think, and ensure things can always return

cicoyle (approved):

cicoyle:
actually, could this PR cause this new e2e failure? Ive not seen it before this PR: `/home/runner/work/dapr/dapr/tests/e2e/workflows/workflow_test.go:254`

JoshVanL:
@cicoyle no we have seen this error for a small while. It _should_ be resolved by the PR https://github.com/dapr/dapr/pull/8408

PR #8406, opened 2025-01-23
Scheduler: Remove space quota

JoshVanL:
Previously, we set a pseudo random space quota where Etcd would go into maintenance mode when hit. Having any kind of limit here doesn't make sense as it will artificially limit scheduler at some point in time.

Allow the user to set a sufficient disk size for Scheduler.


cmd/scheduler/options/options.go
```diff
@@ -92,7 +87,6 @@ func New(origArgs []string) (*Options, error) {
 	fs.StringVar(&opts.EtcdDataDir, "etcd-data-dir", "./data", "Directory to store scheduler etcd data")
 	fs.StringSliceVar(&opts.EtcdClientPorts, "etcd-client-ports", []string{"dapr-scheduler-server-0=2379"}, "Ports for etcd client communication")
 	fs.StringSliceVar(&opts.EtcdClientHTTPPorts, "etcd-client-http-ports", nil, "Ports for etcd client http communication")
-	fs.StringVar(&opts.etcdSpaceQuota, "etcd-space-quota", resource.NewQuantity(defaultEtcdStorageQuota, resource.BinarySI).String(), "Space quota for etcd")
```
elena-kolevska, line 95: I think we should keep the option so users can set it if they want it. It's zero-cost to keep it there, as long as we add a note in the docs explaining that changing the value once set will require manual process.
Or do you think that might add confusion?

JoshVanL: Sounds good, I have added the flag back with 9.2E default.

elena-kolevska (approved):

cicoyle (approved):

PR #8410, opened 2025-01-23
Workflow: Enable remote state transaction & fix status stream

JoshVanL:
Updates actor state transaction to allow workflows to change state of a remote actor, resolving "no instance" errors.

Fixes runtime status watch stream to no longer cancel context early when watching remotely, and properly close stream when daprd shutting down.

Enables workflow reconnect integration test now that it passes.

Enables workflow fanoutthree integration test now that it passes.


famarting (approved):
love it

cicoyle (approved):

PR #8416, opened 2025-01-25
Scheduler: adds --override-broadcast-host-port

JoshVanL:
Override the address (host:port) which is broadcast by this scheduler host that daprd instances will use to connect to this scheduler. This option should only be set by the CLI when in standalone mode, or in exotic environments whereby the routable scheduler address (host:port) is different from its own understood routable address, i.e. in a layered or natted network.

Adds int tests to ensure value it propagated correctly.


antontroshin (approved):

elena-kolevska (approved):
lgtm

PR #8437, opened 2025-01-29
Actors: State store actor reminder drain wait group

JoshVanL:
Add wait group which waits for state store reminder loops to exit before continuing from a placement dissemination reminder drain.

Ensure ticks are always recorded to ensure we don't miss app reminder executions, and result in double invocations during disemination events.

Fixed e2e actor reminder e2e test.


cicoyle (approved):
I was close, but missing the wg. nice

PR #8442, opened 2025-01-30
Placement: increase client heartbeat to 120s

JoshVanL:
Currently, placement clients will send their host information every second. This is completely unnecessary, floods the network and wakes routines that don't need CPU time. Ups the heartbeat to 120s.


famarting:
should be reopened against main branch

PR #8446, opened 2025-01-31
 Scheduler: Use random Etcd ID suffix

JoshVanL:
To prevent existing member ID errors, use a random Etcd ID suffix for
the member ID and force a new cluster on boot. Ensure that the host
names have the suffix removed when advertising.

Suffix only used in Kubernetes deployments.

Branched from https://github.com/dapr/dapr/pull/8395


cmd/scheduler/app/app.go
```diff
@@ -109,7 +110,6 @@ func Run() {
 				EtcdSpaceQuota:          opts.EtcdSpaceQuota,
 				EtcdCompactionMode:      opts.EtcdCompactionMode,
 				EtcdCompactionRetention: opts.EtcdCompactionRetention,
-				EtcdClientHTTPPorts:     opts.EtcdClientHTTPPorts,
```
daixiang0, line 112: Do we need to set it? If removed, use default one?

JoshVanL: We only need a gRPC client port for cron.

PR #8447, opened 2025-01-31
Workflow: Cache workflow state for 10s after Termination

JoshVanL:
Delay deleting the workflow orchestration actor from being deleted from the actor table for 10s in order to hold the cache for new streaming clients to observe the state.

Don't error the workflow when the inbox state cannot be found, which happens when the workflow has been terminated.

10s is enough, as the client is Terminating the workflow anyway.

### DIFF
--- a/pkg/runtime/wfengine/state/state.go
+++ b/pkg/runtime/wfengine/state/state.go
@@ -326,7 +326,8 @@ func LoadWorkflowState(ctx context.Context, state state.Interface, actorID strin
 	for i := range metadata.GetInboxLength() {
 		key = getMultiEntryKeyName(inboxKeyPrefix, i)
 		if bulkRes[key] == nil {
-			return nil, fmt.Errorf("failed to load inbox state key '%s': not found", key)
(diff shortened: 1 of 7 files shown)


pkg/runtime/wfengine/state/state.go
```diff
@@ -326,7 +326,8 @@ func LoadWorkflowState(ctx context.Context, state state.Interface, actorID strin
 	for i := range metadata.GetInboxLength() {
 		key = getMultiEntryKeyName(inboxKeyPrefix, i)
 		if bulkRes[key] == nil {
-			return nil, fmt.Errorf("failed to load inbox state key '%s': not found", key)
+			wfLogger.Warnf("Failed to load inbox state key '%s': not found", key)
```
yaron2, line 329: Is this an actionable item the user can fix? Maybe can this be debug level? If this is expected to not show up frequently (even in a situation where it triggers) we can keep as Warn. Same applies for line 340 below

yaron2 (approved):

PR #8456, opened 2025-02-04
Actors: Regression: return 200 on reminder not found

JoshVanL:
Update Get Actor Reminder to return a 200 rather than 500 when Scheduler reminders are enabled. This matches the same behavior as the state store reminder get request.

Adds integration test.

### DIFF
--- a/pkg/actors/internal/reminders/storage/scheduler/scheduler.go
+++ b/pkg/actors/internal/reminders/storage/scheduler/scheduler.go
@@ -18,6 +18,8 @@ import (
 	"fmt"
 	"time"
 
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 	"google.golang.org/protobuf/types/known/durationpb"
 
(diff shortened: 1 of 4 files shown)


passuied (approved):

pkg/actors/internal/reminders/storage/scheduler/scheduler.go
```diff
@@ -203,6 +205,11 @@ func (s *scheduler) Get(ctx context.Context, req *api.GetReminderRequest) (*api.
 			"jobType":   "reminder",
 		}
 		log.Errorf("Error getting reminder job %s due to: %s", req.Name, err)
```
passuied, line 207: Should it still log an error? Or it good be moved to a debug?

yaron2: Debug makes sense IMO

yaron2 (approved):
LGTM

PR #8457, opened 2025-02-04
Make error code metrics check in integration tests async

elena-kolevska:
Make error code metrics check in integration tests async

### DIFF
--- a/tests/integration/suite/actors/state/http.go
+++ b/tests/integration/suite/actors/state/http.go
@@ -20,6 +20,7 @@ import (
 	nethttp "net/http"
 	"strings"
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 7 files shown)


tests/integration/suite/actors/state/http.go
```diff
@@ -77,8 +78,9 @@ func (h *http) Run(t *testing.T, ctx context.Context) {
 	require.NoError(t, err)
 	require.NoError(t, resp.Body.Close())
 	assert.Equal(t, `{"errorCode":"ERR_ACTOR_INSTANCE_MISSING","message":"actor instance is missing"}`, string(b))
-
-	h.app.Daprd().Metrics(t, ctx).MatchMetricAndSum(2, "dapr_error_code_total", "category:actor", "error_code:ERR_ACTOR_INSTANCE_MISSING")
+	assert.Eventually(t, func() bool {
+		return h.app.Daprd().Metrics(t, ctx).MatchMetricAndSum(1, "dapr_error_code_total", "category:actor", "error_code:ERR_ACTOR_INSTANCE_MISSING")
```
elena-kolevska, line 82: The value change from 2 to 1 is intentional.
This line wasn't actually evaluated before, that's why we never caught the error.

JoshVanL (approved):

PR #8458, opened 2025-02-04
Workflow: remove reminder policy on remote actor call

JoshVanL:


pkg/actors/engine/engine.go
```diff
@@ -117,7 +117,9 @@ func (e *engine) Call(ctx context.Context, req *internalv1pb.InternalInvokeReque
 }
 
 func (e *engine) CallReminder(ctx context.Context, req *api.Reminder) error {
-	if !req.SkipLock {
+	if req.SkipLock {
+		return e.callReminder(ctx, req)
+	} else {
```
antontroshin, line 122: Is it intentional that we're skipping the whole "resiliency" wrapped call below?

JoshVanL: yes

JoshVanL: For Activities we don't want that. Only Workflow activities will have this skip option set.

antontroshin (approved):

daixiang0 (approved):
LGTM

PR #8463, opened 2025-02-04
Integration: Increase timeout for http state error metric

JoshVanL:


elena-kolevska (approved):
Lgtm, but if we're seeing this failure here, we'll probably see it in all the other metrics checks. Once it becomes a problem I can go ahead and bump the timeout in all of them.

PR #8464, opened 2025-02-05
Scheduler: Enforce 3 replicas on Kubernetes

JoshVanL:
Always enforce 3 replicas on Kubernetes to prevent member ID issues, and ensure data integrity. 

Update the client listen address to be only over localhost, removing the need for mTLS.

Refactor scheduler server code to moving config and etcd server setup into `etcd` package.

Updates dapr/kit to v0.15.0


elena-kolevska (changes requested):
Thanks for the PR! I've been testing this approach and noticed a few potential concerns:

First, the node removal in defer doesn't seem to work, it returns `{"instance":"dapr-scheduler-server-2","level":"error","msg":"Error removing member from etcd: context canceled","scope":"dapr.scheduler.server.cron","time":"2025-02-05T16:04:28.27225771Z","type":"log","ver":"edge"}`. I suspect this has to do with the node removing itself, but have to look into it deeper.

Then, if we solve the error above, after a member has been removed and the scheduler restarts, we can't start anymore with the default `--initial-cluster-state=new` flag or the `--initial-cluster-state=existing`, we would have to do a `member add` because the cluster has already been bootstrapped and in addition - delete the old data dir, otherwise we get `"error":"the member has been permanently removed from the cluster"`.

We should also check how does the member removal affect quorum, and if it's even possible to remove more than half of the members.

And finally, and probably not that important, but worth mentioning anyway, this approach only works for graceful shutdown when we hit the defer, which excludes situations like network partitions, OOM errors, or any hard crashes.

cicoyle (commented):
I tested this locally and still see the same error when going from a single instance to ha:
```
{"level":"warn","ts":"2025-02-05T16:53:53.251658Z","caller":"rafthttp/http.go:500","msg":"request cluster ID mismatch","local-member-id":"6deb7341ee1ec37","local-member-cluster-id":"68f3781e2b1eea2d","local-member-server-version":"3.5.16","local-member-server-minimum-cluster-version":"3.0.0","remote-peer-server-name":"f6c3419d2b0c5c1e","remote-peer-server-version":"3.5.16","remote-peer-server-minimum-cluster-version":"3.0.0","remote-peer-cluster-id":"c07e8d523c24b9ab"}
```

yaron2 (approved):

PR #8435, opened 2025-01-29
Issue 8254: Added support for setting allowPrivilegeEscalation and readOnlyRootFilesystem as applicable

mustaFAB53:
Issue 8254: Added support for setting allowPrivilegeEscalation and readOnlyRootFilesystem as applicable

PR raised in continuation to discussion on https://github.com/dapr/dapr/pull/8409

# Description

I have added a support to set allowPrivilegeEscalation and readOnlyRootFilesystem kubernetes securityContext attributes for relevant dapr control components as suggested and finalized as part of #8254 

## Issue reference

#8254 

## Checklist

Please make sure you've completed the relevant tasks for this PR, out of the following list:

- [ ] Code compiles correctly
- [ ] Created/updated tests
- [ ] Unit tests passing
- [ ] End-to-end tests passing
- [ ] Extended the documentation / Created issue in the <https://github.com/dapr/docs/> repo: dapr/docs#_[issue number]_
- [ ] Specification has been updated / Created issue in the <https://github.com/dapr/docs/> repo: dapr/docs#_[issue number]_
- [ ] Provided sample for the feature / Created issue in the <https://github.com/dapr/docs/> repo: dapr/docs#_[issue number]_

PR #8465, opened 2025-02-05
Fix Scheduler readOnlyRootFilesystem: true

cicoyle:
Add writable identity directory for Scheduler's etcd TLS certificates to support readOnlyRootFilesystem.

### DIFF
--- a/charts/dapr/charts/dapr_scheduler/templates/dapr_scheduler_statefulset.yaml
+++ b/charts/dapr/charts/dapr_scheduler/templates/dapr_scheduler_statefulset.yaml
@@ -172,7 +172,7 @@ spec:
           capabilities:
             add: ["SYS_PTRACE"]
   {{- else }}
-          readOnlyRootFilesystem: {{ .Values.securityContext.readOnlyRootFilesystem }}
+          readOnlyRootFilesystem: true
           capabilities:
             drop: ["ALL"]
   {{- end }}
--- a/charts/dapr/charts/dapr_scheduler/templates/dapr_scheduler_statefulset.yaml
+++ b/charts/dapr/charts/dapr_scheduler/templates/dapr_scheduler_statefulset.yaml
@@ -95,6 +95,8 @@ spec:
             mountPath: {{ .Values.cluster.etcdDataDirPath }}/
           {{- end }}
             readOnly: false
+          - name: dapr-scheduler-writeable-identity
+            mountPath: /tmp
           - name: dapr-trust-bundle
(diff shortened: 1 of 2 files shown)


cicoyle:
See it passing in CI for `readOnlyRootFilesystem` set to both `false` && for `true` now. See [this CI run from the first commit for true](https://github.com/dapr/dapr/actions/runs/13166272027/job/36747051634). I tested locally and downloaded the container logs and grepped for `read-only` and dont see it so we should be good on the `read-only` err. 

On another note, @JoshVanL looks like those flaky actor tests are still around. I dont think the diff in this PR would cause those unless Im mistaken:
```
TestActorReminder/actorreminderscheduler/Actor_reminder_with_repetition_should_run_correct_number_of_times (20.03s)
    actor_reminder_test.go:455: Sleeping for 20 seconds ...
    actor_reminder_test.go:458: Getting logs from 172.18.0.2:31804/test/logs to see if reminders did trigger ...
    actor_reminder_test.go:462: Checking if all reminders did trigger for app actorreminderscheduler...
    actor_reminder_test.go:464: 
        	Error Trace:	/home/runner/work/dapr/dapr/tests/e2e/actor_reminder/actor_reminder_test.go:464
        	Error:      	Not equal: 
        	            	expected: 5
        	            	actual  : 0
        	Test:       	TestActorReminder/actorreminderscheduler/Actor_reminder_with_repetition_should_run_correct_number_of_times

```

Please note the `{"level":"warn","ts":"2025-02-05T20:53:59.593936Z","caller":"embed/config_logging.go:170","msg":"rejected connection on client endpoint","remote-addr":"127.0.0.1:56882","server-name":"dapr-scheduler-server.dapr-tests.svc","error":"remote error: tls: bad certificate"}` from the Scheduler log I believe is fixed in [this PR](https://github.com/dapr/dapr/pull/8464)

elena-kolevska (commented):
Do we see a situation where a user have to change the default value of `readOnlyRootFilesystem`? Would it make sense to keep it safer and hardcode it to false?

cicoyle:
> Do we see a situation where users have to change the default value of `readOnlyRootFilesystem`? Would it make sense to keep it safer and hardcode it to true? IIRC we were getting security tools complaining about it being set to false.

I thought people wanted to toggle it based on [this open PR](https://github.com/dapr/dapr/pull/8435), but Im happy to hard code it to true for security reasons and if users request a param we can add it back in later on. I'll switch it back to true. Was just letting CI run the tests.

JoshVanL (approved):

Issue #8100, opened 2024-09-12
Allow user to completely remove/disable scheduler in 1.14+


artursouza:
## In what area(s)?

> /area runtime

> /area operator

> /area placement

> /area docs

> /area test-and-release

## What version of Dapr?

> 1.1.x
> 1.0.x
> edge: output of `git describe --dirty`
1.14.x

## Expected Behavior

If user does not want to use scheduler, via the Helm chart setting `global.scheduler.enabled`, should allow scheduler's k8s resources not to be provisioned at all (no statefulset, service, nothing). The sidecar should also be smart enough (via env var in the sidecar injector, maybe) to not try to connect to scheduler on start up. Job's API should error proactively saying that scheduler is not available instead of trying to perform any operation.

This should work equally if user does not install scheduler's container on Docker for standalone mode.

## Actual Behavior

```
failed to watch scheduler jobs, retrying: rpc error: code = Unavailable desc = connection error: desc = \"transport: Error while dialing: dial tcp: lookup dapr-scheduler-server-1.dapr-scheduler-server.dapr-system.svc.cluster.local on 172.20.0.10:53: no such host\""
```

## Steps to Reproduce the Problem

Disable scheduler (via `global.scheduler.enabled` or setting replica to 0 for scheduler's statefulset) and try to run any app on K8s.

## Release Note

RELEASE NOTE: **UPDATED** control plane and sidecar to work without Scheduler service

artursouza:
As a workaround, set the following:

- For K8s mode: set annotation on app's Deployment: `dapr.io/scheduler-host-address: ""
- For standalone mode, set env: `export DAPR_SCHEDULER_HOST_ADDRESS=""`

cicoyle:
To fix part of this issue, we should have dynamic host resolution via the injector, rn that code isnt dynamic. We can also be more graceful with our errors in the jobs api, bc it assumes scheduler is there if the jobs api is being used.

rgitpriv:
> As a workaround, set the following:
> 
> * For K8s mode: set annotation on app's Deployment: `dapr.io/scheduler-host-address: ""`
> * For standalone mode, set env: `export DAPR_SCHEDULER_HOST_ADDRESS=""`

Hi, that workaround is not working for us on 1.14.4, can you please confirm? I've tried with our deployments on EKS but here's a repro using DAPR resources, on a clean minikube install with k8s version 1.29.7:

[...]

PR #8473, opened 2025-02-07
Respect empty scheduler/placement annotations & handle actorStateStore without placement

cicoyle:
This PR fixes two issues related to disabling connections to the scheduler & placement service via annotations.

1. Respect Explicitly Empty Annotations
Users can disable connecting to scheduler and placement per [our docs](https://docs.dapr.io/reference/arguments-annotations-overview/) by setting the annotations field to `""` or `" "`.
Currently, the sidecar injector overwrites these values with the k8s address instead of respecting them.
Now, the sidecar injector properly respects `""` or `" "` for both placement & scheduler addresses.

2. Handle `actorStateStore=true` with Placement Disabled
If `actorStateStore: true` is set but placement is explicitly disabled, the previous behavior resulted in errors like:
```
time="2025-02-07T14:55:50.519322722Z" level=debug msg="api error: code = Internal desc = dapr is not ready"
time="2025-02-07T14:55:50.892991847Z" level=error msg="Failed to connect to placement: rpc error: code = Unavailable desc = connection error: desc = \"transport: Error while dialing: dial tcp: address \\\"\\\": missing port in address\""
```

```
time="2025-02-07T15:35:07.952844299Z" level=info msg="Actor runtime disabled: api error: code = Internal desc = the state store is not configured to use the actor runtime."
```
Now, if `actorStateStore=true` but placement is disabled, Dapr will simply log that placement is disabled.

I also explicitly log about if placement is disabled that actors + workflow api cannot be used along with the scheduler not being connected when configured.

[Should fix this issue](https://github.com/dapr/dapr/issues/8100)


pkg/runtime/config.go
```diff
@@ -317,6 +318,17 @@ func (c *Config) toInternal() (*internalConfig, error) {
...
 
+	placementAddrs := strings.Split(strings.TrimPrefix(c.ActorsService, "placement:"), ",")
+	hasValidPlacementAddr := false
+	for _, addr := range placementAddrs {
+		if strings.TrimSpace(addr) != "" {
+			hasValidPlacementAddr = true
+			break
+		}
+	}
+
```
JoshVanL, line 329: Why wouldn’t a non valid placement address not error?

JoshVanL (changes requested):

pkg/runtime/config.go
cicoyle: The double negative is confusing me, but I did double check tho that this is no longer needed with the injector respecting the user annotations now for `""` && `" "`. removed the check

pkg/runtime/scheduler/scheduler.go
```diff
@@ -82,8 +83,10 @@ func New(opts Options) *Scheduler {
 }
 
 func (s *Scheduler) Run(ctx context.Context) error {
-	if len(s.addresses) == 0 {
+	if len(s.addresses) == 0 ||
+		len(s.addresses) == 1 && strings.TrimSpace(s.addresses[0]) == "" {
```
JoshVanL, line 87:
```suggestion
		(len(s.addresses) == 1 && strings.TrimSpace(s.addresses[0]) == "") {
```

pkg/runtime/wfengine/wfengine.go
```diff
@@ -142,6 +143,13 @@ func (wfe *engine) RegisterGrpcServer(server *grpc.Server) {
 }
 
 func (wfe *engine) Run(ctx context.Context) error {
+	if wfe.actors.RuntimeStatus().GetRuntimeStatus() == runtimev1pb.ActorRuntime_DISABLED {
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		}
```
JoshVanL, line 150:
```suggestion
  <-ctx.Done()
  return ctx.Error()
```

tests/integration/suite/daprd/actors/actors.go
```diff
@@ -0,0 +1,71 @@
...
+
+	"github.com/dapr/dapr/tests/integration/framework"
+	"github.com/dapr/dapr/tests/integration/framework/process/daprd"
+	"github.com/dapr/dapr/tests/integration/framework/process/exec"
+	"github.com/dapr/dapr/tests/integration/framework/process/logline"
+	"github.com/dapr/dapr/tests/integration/suite"
+)
+
+func init() {
+	suite.Register(new(actors))
```
JoshVanL, line 28: This should live under `/suite/actors`. That package itself is probably in the wrong place and should live under `/suite/daprd/actors` but we shouldn't change that in this PR.

Rename this test (struct) to `emptyaddress` or similar.

JoshVanL (changes requested):

tests/integration/suite/daprd/actors/actors.go
```diff
@@ -0,0 +1,71 @@
+/*
```
JoshVanL, line 1: 👍

tests/integration/suite/daprd/actors/actors.go
cicoyle: yeah I agree that the `/suite/actors` should move to `/suite/daprd/actors`. For now, I've changed the struct to be `emptyaddress` per your suggestion.

JoshVanL (changes requested):
@cicoyle after we've move the int test file under `/suite/actors` we can merge.

elena-kolevska (dismissed):

cicoyle:
> @cicoyle after we've move the int test file under `/suite/actors` we can merge.

I thought it was fine for the new test I added to be in the proper path and a follow up PR would move the `suite/actors` tests to the proper dir that I added a test to which is `suite/daprd/actors`, but that would be in a separate PR and not in my PR here

cicoyle:
cc @JoshVanL I opened[ this PR](https://github.com/dapr/dapr/pull/8481) to move those tests to the proper path. I would appreciate if this current PR did not have to wait since I opened the other PR to move the tests :)

JoshVanL (approved):
Let’s make sure to rename the added int test file to `emptyaddress`.

JoshVanL (approved):

PR #8474, opened 2025-02-07
Scheduler: Attempt to remove member during shutdown

JoshVanL:
During Scheduler shutdown, attempt to remove ones self as a member to prevent member ID collisions.

### DIFF
--- a/pkg/scheduler/server/internal/etcd/etcd.go
+++ b/pkg/scheduler/server/internal/etcd/etcd.go
@@ -17,6 +17,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"time"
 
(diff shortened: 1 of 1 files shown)


pkg/scheduler/server/internal/etcd/etcd.go
```diff
@@ -126,6 +127,11 @@ func (e *etcd) Close() error {
 
 	var err error
 	if e.client != nil {
+		ctx, cancel := context.WithTimeout(context.Background(), time.Second*3)
+		defer cancel()
+		if _, err = e.client.MemberRemove(ctx, uint64(e.etcd.Server.ID())); err != nil {
```
artursouza, line 132: Won't this compromise the durability of the data on etcd? If all instances are shutdown at once?

PR #8480, opened 2025-02-11
Scheduler: Cleanup data dir on first boot

JoshVanL:
To prevent member and cluster ID missmatch/errors on 1.15 upgrade, we now will delete the target data directory on first boot. This ensures that we have a fresh Etcd on 1.15 upgrade. Users will have data loss in Scheduler upgrading from 1.14.

We use a file in the data dir to signal that we have already had the first boot of that replica in 1.15.


pkg/scheduler/server/internal/etcd/etcd.go
```diff
@@ -135,3 +144,48 @@ func (e *etcd) Close() error {
...
+	if err == nil {
+		log.Infof("Found existing cluster data flag, preserving data dir: %s", e.config.Dir)
+		return nil
+	}
+
+	if !os.IsNotExist(err) {
+		return err
+	}
+
+	log.Infof("No existing cluster data flag found, deleting data dir contents: %s", e.config.Dir)
```
cicoyle, line 159: I dont think that'll mean anything to a user so I think we can just rm that key work.
```suggestion
	log.Infof("No existing cluster data found, deleting data dir contents: %s", e.config.Dir)
```

pkg/scheduler/server/internal/etcd/config.go
```diff
@@ -39,20 +42,38 @@ func config(opts Options) (*embed.Config, error) {
...
+		if err != nil {
+			return nil, err
+		}
+
+		certs, err := pem.DecodePEMCertificatesChain(b)
+		if err != nil {
+			return nil, err
+		}
+
+		if !slices.Contains(certs[0].DNSNames, fmt.Sprintf("*.dapr-scheduler-server.%s.svc", opts.Security.ControlPlaneNamespace())) {
```
cicoyle, line 73: I tested locally and am seeing this logged from etcd:
```
{"level":"warn","ts":"2025-02-11T19:28:42.785510Z","caller":"embed/config_logging.go:161","msg":"rejected connection on peer endpoint","remote-addr":"10.244.1.12:44178","server-name":"dapr-scheduler-server-1.dapr-scheduler-server.dapr-tests.svc","ip-addresses":[],"dns-names":["*.dapr-scheduler-server.dapr-tests.svc"],"error":"tls: \"10.244.1.12\" does not match any of DNSNames [\"*.dapr-scheduler-server.dapr-tests.svc\"]"}
```
I dont think etcd likes the wildcard

pkg/scheduler/server/internal/etcd/config.go
```diff
@@ -78,12 +99,14 @@ func config(opts Options) (*embed.Config, error) {
 	case modes.KubernetesMode:
 		config.Dir = opts.DataDir
 		etcdURL.Host = "0.0.0.0:" + etcdURL.Port()
+		config.AdvertiseClientUrls[0].Scheme = "https"
```
yaron2, line 102: Shouldn't this be set in SelfHosted mode too if TLS is enabled?

JoshVanL: Yes! Thanks
Updated

yaron2 (approved):

PR #8483, opened 2025-02-12
Adds workflow tests with durable task client

elena-kolevska:
Needs durabletask-go PR to be merged first: https://github.com/dapr/durabletask-go/pull/16

### DIFF
--- a/tests/integration/suite/daprd/workflow/scheduler/basicdtclient.go
+++ b/tests/integration/suite/daprd/workflow/scheduler/basicdtclient.go
@@ -1,0 +1,382 @@
+/*
+Copyright 2024 The Dapr Authors
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://wwb.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package scheduler
+
+import (
+	"context"
+	"database/sql"
+	"errors"
+	"fmt"
+	"net/http"
+	"os"
+	"path/filepath"
+	"strconv"
+	"sync/atomic"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"google.golang.org/grpc/status"
+
+	runtimev1pb "github.com/dapr/dapr/pkg/proto/runtime/v1"
+	"github.com/dapr/dapr/tests/integration/framework"
+	frameworkclient "github.com/dapr/dapr/tests/integration/framework/client"
+	"github.com/dapr/dapr/tests/integration/framework/process/daprd"
+	prochttp "github.com/dapr/dapr/tests/integration/framework/process/http"
+	"github.com/dapr/dapr/tests/integration/framework/process/placement"
+	procscheduler "github.com/dapr/dapr/tests/integration/framework/process/scheduler"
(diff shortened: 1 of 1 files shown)


tests/integration/suite/daprd/workflow/scheduler/basicdtclient.go
```diff
@@ -0,0 +1,320 @@
+/*
```
JoshVanL, line 1: We don't need to make this test so involved to check that the purge API is doing the right thing. We should try to keep integration tests specific to one scenario, both for readability, search-ability, and being mindful about test run time.

I was expecting something more like the following.


```go
package workflow

import (
	"context"
	"testing"

	"github.com/dapr/dapr/tests/integration/framework"
	"github.com/dapr/dapr/tests/integration/framework/process/workflow"
	"github.com/dapr/dapr/tests/integration/suite"
	"github.com/dapr/durabletask-go/task"
)

func init() {
	suite.Register(new(purge))
}

type purge struct {
	workflow *workflow.Workflow
}

func (p *purge) Setup(t *testing.T) []framework.Option {
	p.workflow = workflow.New(t)

	return []framework.Option{
		framework.WithProcesses(p.workflow),
	}
}

func (p *purge) Run(t *testing.T, ctx context.Context) {
	p.workflow.WaitUntilRunning(t, ctx)

	p.workflow.Registry().AddOrchestratorN("purge", func(ctx *task.OrchestrationContext) (any, error) {
		require.NoError(t, ctx.CallActivity("abc").Await(nil))
		return nil, nil
	})
	p.workflow.Registry().AddActivityN("abc", func(ctx task.ActivityContext) (any, error) {
		return nil, nil
	})

	client := p.workflow.BackendClient(t, ctx)

	id, err := client.ScheduleNewOrchestration(ctx, "purge")
	require.NoError(t, err)

	db := p.workflow.DB().GetConnection(t)
	tableName := p.workflow.DB().TableName()

	var count int
	require.NoError(t, db.QueryRowContext(ctx, "SELECT COUNT(*) FROM %s"+tableName).Scan(&count))
	assert.Equal(t, 1, count)

	_, err = client.WaitForOrchestrationCompletion(ctx, id)
	require.NoError(t, err)

	require.NoError(t, client.PurgeOrchestrationState(ctx, id))

	require.NoError(t, db.QueryRowContext(ctx, "SELECT COUNT(*) FROM %s"+tableName).Scan(&count))
	assert.Equal(t, 1, count)
}
```

elena-kolevska: Thanks, I pushed a simpler test.

tests/integration/suite/daprd/workflow/purge.go
```diff
@@ -0,0 +1,75 @@
...
+
+	id, err := client.ScheduleNewOrchestration(ctx, "purge")
+	require.NoError(t, err)
+
+	db := p.workflow.DB().GetConnection(t)
+	tableName := p.workflow.DB().TableName()
+
+	var count int
+	require.NoError(t, db.QueryRowContext(ctx, "SELECT COUNT(*) FROM "+tableName).Scan(&count))
+	assert.GreaterOrEqual(t, count, 0)
```
JoshVanL, line 66: Shouldn’t we test for >= 1?

JoshVanL (changes requested):

tests/integration/suite/daprd/workflow/purge.go
elena-kolevska: Oh yeah, that was supposed to be >=1. I had it pinned to ==5 (that's three history records, one metadata and one custom status), but I don't know if that's deterministic or there would be more in case of retries...
Should I go with >=5, to be more strict, or is >=1 ok?

JoshVanL: Let’s go with >= 5 🙂

elena-kolevska: Done! Thanks

JoshVanL (approved):

PR #8491, opened 2025-02-17
Scheduler: version the Scheduler data dir

JoshVanL:
Version the data dir used by Scheduler Etcd using a new directory (`dapr-0.1`) so that downgrading to 1.14 does not use the same existing metadata. Without this, 1.14 which doesn't use TLS will never achieve quorum and fail to start.


yaron2:
Unit tests

JoshVanL:
@yaron2 updated

cicoyle (dismissed):

cicoyle (commented):
IT tests seem flaky

yaron2 (approved):

yaron2:
> IT tests seem flaky

Yep but since there wasn't any material change from the last RC I'm giving it the benefit of the doubt that it's a runner issue. We'll see later in scheduled tests / other PRs if it repeats

PR #8492, opened 2025-02-17
Fix workflow engine starting when actor runtime is disabled

cicoyle:
The workflow engine was starting even when the actor runtime was disabled due to empty placement addresses. This happened because we are checking the actor runtime status in [wfengine here](https://github.com/dapr/dapr/blob/release-1.15/pkg/runtime/wfengine/wfengine.go#L146) bc the `a.disabled.Load()` in the actors pkg is not externally available to the wf engine pkg. This additional check ensures the workflow engine correctly detects when actors are disabled. 

I was running the `emtpyaddress` IT test locally and noticed it running the wf engine by this log: `workflow engine started` which does not occur anymore with this additional check.

-------

Apparently for backwards compatibility for the metadata API I could not proceed with the prior approach. Therefore, I switched to checking the actor engine which is essentially the same check thru a diff approach. This prevents wf from running if placement is disabled.


elena-kolevska (commented):
lgtm, but seems like there's a related test failing

cicoyle:
The IT test failures from last run were unrelated. Will verify with this next run again

cicoyle:
The failing IT test is unrelated

pkg/runtime/wfengine/wfengine.go
```diff
@@ -146,7 +145,8 @@ func (wfe *engine) RegisterGrpcServer(server *grpc.Server) {
 }
 
 func (wfe *engine) Run(ctx context.Context) error {
-	if wfe.actors.RuntimeStatus().GetRuntimeStatus() == runtimev1pb.ActorRuntime_DISABLED {
+	_, err := wfe.actors.Engine(ctx)
+	if err != nil {
 		<-ctx.Done()
 		return ctx.Err()
```
JoshVanL, line 151:
```suggestion
	_, err := wfe.actors.Engine(ctx)
	if err != nil {
		return err
```

JoshVanL (changes requested):

cicoyle:
That e2e test has been flaky and is unrelated to the diff in this PR I believe

JoshVanL (approved):

PR #8503, opened 2025-02-20
security: force to use kubernetes validator when DAPR_SENTRY_TOKEN_FILE is provided

famarting:
# Description

I have come across this issue when trying to write integration tests using the kubernetes mode.

Dapr already looks into `/var/run/secrets/dapr.io/sentrytoken/token` for the sentry token, but I cannot write into the `/var/run/secrets` directory unless we run the integration tests as root... which doesnt seem like a good idea.

If the change in this PR is not welcome or seems risky, another option I can think of is to make the path `/var/run/secrets/dapr.io/sentrytoken/token` configurable via an environment variable

## Issue reference

## Checklist

- [ ] Code compiles correctly
- [ ] Created/updated tests
- [ ] Unit tests passing


pkg/security/token/token.go
```diff
@@ -54,7 +54,12 @@ func GetSentryToken(allowKubernetes bool) (token string, validator sentryv1pb.Si
 		if path == "" {
 			return "", sentryv1pb.SignCertificateRequest_UNKNOWN, errors.New("environmental variable DAPR_SENTRY_TOKEN_FILE is set with an empty value")
 		}
-		return GetSentryTokenFromFile(path)
+		token, validator, err = GetSentryTokenFromFile(path)
```
JoshVanL, line 57: Can you update the const comment to say that this env var enforces the kube validator when running in kube, else uses JWKs in self hosted 

https://github.com/dapr/dapr/blob/8cefe6a312867d8517a750d2dc300e10821bdf3a/pkg/security/consts/consts.go#L38

JoshVanL (changes requested):

JoshVanL (approved):

PR #8518, opened 2025-02-24
Fix Actor & Scheduler deadlocks and Integration Test fixes

JoshVanL:
This changes includes a number of changes to prevent dead locks in the actor runtime & Scheduler connectivity/job triggering sequences. There a number of disparate scenarios covered below. This change ensures that no deadlocks occur, particularly when running workflows, during scale (up or down) events, loosing connection or destruction of the placement and scheduler control plane services. Makes integration tests more stable, particularly in CI/slow machine runners.

Scenarios and fixes:

Adds a new `scheduler` fields to the gRPC and HTTP metadata output, which includes an object containing the list of addresses which the daprd is currently connected to. This address list can change over time, as daprd dynamically connects/looses connection to scheduler hosts. Int tests added for this field, as well as being used to check the address list in some integration tests where the number of hosts connected is a dependency of that test scenario (e.g. job load balancing).

Moves the placement lock to `dapr/kit/concurrency/lock` as a new shared `OuterCancel` lib. This struct was also fixed so that a deadlock could not occur on `RLock` whereby the given context was cancelled, but `RLock` did not return with error- now it does. This caused deadlocks in workflow actors whereby dissemination (deactivation) of actors would not complete as their was a circular lock which did not break.

Removes a deadlock in the Scheduler server whereby a new daprd stream was stuck adding to the connection pool. This was due to adding the new connections as deliverable prefixes, locks the underlying go-etcd-cron queue for those prefixes. If a job trigger was occurring at the same time, there was a circular lock with no break since the same Scheduler lock adding the new connection was RLocked to get the connection for that job trigger to send the job event to daprd. The use of the new `Context` lock from `dapr/concurrency/lock` ensures that getting the connection for a triggered job will fail the RLock if it does not gain the lock after 250ms. This breaks the circular lock. When this RLock fails, the job will be reported back to go-etcd-cron that the job is not deliverable and unlock the queue lock- since the job _is_ actually deliverable, go-etcd-cron will correctly re-enqueue the job immediately whilst releasing that queue lock in the mean time. See: https://github.com/diagridio/go-etcd-cron/blob/c9126ef694ffed83011125c5acbf6127f33bf784/internal/queue/queue.go#L209-L217

Fixes another deadlock in the Scheduler server whereby a connection to daprd would not correctly be identified as a closed or cancelled connection. This would happen when the connection is closed or cancelled, whilst there is a send & job handle in progress for a particular connection. This has been resolved by adding another go routine to the connection handler to detect when either Recv or Send has returned, the pool total is closed, or the stream context has closed, then closing and removing that connection entirely. This correctly ensures that the connection is removed from the pool, as well as returning from the stream at the top gRPC RPC level.

Permanently backoff when calling a remote reminder when the remote peer is not available, i.e. during a scale down event. This prevents future retries, which slows down the entire placement dissemination for that namespace.

Mark the placement client as ready later- after the placement client has sent the first host request, so to better reflect the real ready state and prevent unnecessary timing errors which cause retries downstream.

Move `HaltAll` and `Drain` implementation to the same func as they were the same implementation, only with one using a conditional for the drain.

Pre-fetch the actor sub Interfaces on the Factory initialization of workflow and activity actors. This prevents a circular deadlock during cases of dissemination where a workflow/activity actor would be deactivated whilst attempting to retrieve one of these interfaces (like state).

Move the workflow actor internal lock to only be used in the _actual_ critical shared memory locations. Previous use of locking around actor resource usage like state would cause the same circular deadlock as described above.

The Scheudler client has been changed to be more aggressive for scheduler reconnects, particularly on `WatchHosts`. This prevents reconnect errors which were observed in scenarios whereby the entire scheduler cluster was to go down temporarily. Also removes the retry loop for the Scheduler client connector since a reconnect at the top level is preferred.

Mark the Scheduler client readiness to true only when all Scheduler host clients have actually sent the initial message to their respective host. Sets readiness to false during shutdown and reconnect events.

JobsWatch passes a context up which will be closed when either Send/Recv or shutdown is in progress to ensure the gRPC API is returned on disconnect events.

Set Scheduler log level to `debug` in e2e CI runs.

We need to register our custom proxy codec in the global registrar, but only after all gRPC internal codecs have been registered. This is because we are squatting the conflicted base name "proto" which we use to inject our custom marshal code. We do this to optionally passthrough proxy data if the gRPC frame encoding is a message we don't recognise- i.e. a user is doing a direct message using their own message type. Since 'd' comes before 'g' in the alphabet and go `init` func execution order is now sane, we have to register our conflicting base codec as runtime. It is also the case that gRPC uses a global variable codec registrar so we can't build a custom one that we propagate. The solution is to keep this as is, or find a way to bypass the codec stack further down the transport layer in a wrapper. I assume we can use a custom message type to wrap user messages which does not have a conflicting name with the base codec.

Don't register actor types to Scheduler if the SchedulerActorReminders preview feature is disabled.

Don't lock the table on calls to Actor Type list.

Ensure that shutdown or fatal placement connection errors do not gracefully drain active actors on the host, and instead deactivate them immediately. Refactor actors table to de-dupe logic and reduce string concat.

Add backoff retry to legacy reminder state store `getRemindersForActorType` to account for etag missmatch.

Integration Tests:

- Set all exec processes to use `debug` log level to make debugging easier when Integration tests fail.
- Fixes shutdown timeout of HTTP app Cleanup.
- Fix log line process Cleanup logic to ensure all exec process logs are flushed before checking for a log line existence.
- Fixes timing errors in scheduler failure policy tests.
-  Adds actor type handler not implementing the deactivation endpoints.
-  Reduce the number of components in fuzz component name tests to prevent shutdown timeouts on slow CI runnners.
- Fix broken pipe network errors which occur on slow CI runners.
- Adds tests for the new scheduler metadata field, both for gRPC and HTTP.
- Fix state metrics tests by using the correct Enventually tester on asserts.
- Correctly use Workflow `WaitForSingleEvent` for workflow coordination, rather than cross scope channels for singling.
- Use eventuallys to prevent EOF errors during placement dissemination tests.
- Prevent cleanup timeout on reminder remote tests using non-closing channels for signaling.
- Prevent exec panic on pipe output close timeouts.
- Prevent exec error panicing whole suite & include timeout before closing exec std pipes.


pkg/actors/internal/placement/placement.go
```diff
@@ -98,7 +99,7 @@ type placement struct {
 }
 
 func New(opts Options) (Interface, error) {
-	lock := lock.New()
+	lock := lock.NewOuterCancel(errors.New("placement is disseminating"), time.Second*2)
```
yaron2, line 102: What's the effect of upping or reducing `time.Second*2`?

JoshVanL: This value determines the upper bound for which placement will wait for in-flight actor requests to finish before cancelling the context to them, effectively force closing them. Making this number 0s would mean that all in-flight requests like actor invocations which will always be cancelled, resulting in more retry-able errors in the system and double calls etc. Making this number larger will mean more long running requests will have the opportunity to complete without being cancelled, but will result in longer dissemination times. 2 seconds is a random number I picked that seemed like long enough for simple calls to finish, while not going to far to cover all long running requests.

Note that it will still always be the case that all requests will be waited on to be finished (either gracefully or by cancel) before the lock is achieved to do dissemination. Even if that time is longer than this 2s. This ensures the single host property of Dapr actors. This property is how the deadlocks fixed in this PR come about. It is also the case that the lock will be achieved sooner if the in-flight requests complete sooner. No in-flight requests can start during this graceful termination period.

yaron2:
Conflict after latest merge

pkg/actors/table/table.go
```diff
@@ -153,26 +151,36 @@ func (t *table) Len() map[string]int {
...
-// Drain drains actors who return true on the given func.
-func (t *table) Drain(fn func(actorType, actorID string) bool) {
+// HaltAll halts all actors currently in the table.
+func (t *table) HaltAll() error {
+	return t.Drain(func(actorType, actorID string) bool {
+		return true
+	})
+}
+
+func (t *table) Drain(fn func(actorType, actorID string) bool) error {
```
cicoyle, line 161: This is super nit, but can you add back & improve the comment above this func? I think future developers could benefit from a simple comment on the diff btw `drain` vs `haltall`

tests/e2e/scheduler/scheduler_test.go
```diff
@@ -168,11 +168,13 @@ func TestJobs(t *testing.T) {
 			require.NoError(t, err)
 
 			// Check if the length of triggeredJobs matches the expected length of scheduled jobs
-			assert.Equal(c, numIterations*numJobsPerGoRoutine, len(triggeredJobs))
+			assert.Len(c, triggeredJobs, numIterations*numJobsPerGoRoutine)
 		}, 10*time.Second, 100*time.Millisecond)
 		t.Log("Done.")
 	})
 
+	time.Sleep(time.Second * 5)
```
cicoyle, line 176: Why add a `time.sleep`? If the tests are interfering with one another can we instead add a `ttl` to the individual jobs or add a`delete` call at the end of the first `t.run`?

tests/integration/framework/process/daprd/daprd.go
```diff
@@ -80,7 +80,7 @@ func New(t *testing.T, fopts ...Option) *Daprd {
 		publicPort:       fp.Port(t),
 		metricsPort:      fp.Port(t),
 		profilePort:      fp.Port(t),
-		logLevel:         "info",
+		logLevel:         "debug",
```
cicoyle, line 83: Will this get put back to `info` after further flaky test squashing?
```suggestion
		logLevel:         "info",
```

tests/e2e/scheduler/scheduler_test.go
JoshVanL: Thanks, was a left over from my debugging. I agree, I've removed it

tests/integration/framework/process/injector/injector.go
```diff
@@ -54,7 +54,7 @@ func New(t *testing.T, fopts ...Option) *Injector {
 
 	fp := ports.Reserve(t, 3)
 	opts := options{
-		logLevel:      "info",
+		logLevel:      "debug",
```
cicoyle, line 57: same here, is this getting reverted? Calling it out so we don't lose it in the large diff
```suggestion
		logLevel:      "info",
```

tests/integration/framework/process/daprd/daprd.go
JoshVanL: I think it's correct for us to keep it debug actually. When a test fails a dev will wish that debug was set, so it makes sense to have that default.

The only thing we lose having this as debug is the extra bit of RAM holding the extra debug log lines that will get de-referenced and quickly garbage collected after that test case is successful.

tests/integration/framework/process/injector/injector.go
JoshVanL: Replied above :slightly_smiling_face:

tests/integration/suite/placement/dissemination/streamhang.go
```diff
@@ -68,10 +68,17 @@ func (n *streamHang) Run(t *testing.T, ctx context.Context) {
...
 	t.Cleanup(cancel1)
-	stream1, streamCancel1 := n.getStream(t, ctx1)
-	t.Cleanup(streamCancel1)
-	err := stream1.Send(host1)
-	require.NoError(t, err, "Failed to send host1")
+	var err error
+	var stream1 v1pb.Placement_ReportDaprStatusClient
+	var streamCancel1 context.CancelFunc
+	assert.EventuallyWithT(t, func(c *assert.CollectT) {
+		stream1, streamCancel1 = n.getStream(t, ctx1)
```
cicoyle, line 75: Shall we update `getStream` to accept `c` (here and further down in this file) for consistency and to ensure failures are properly tracked?
```suggestion
		stream1, streamCancel1 = n.getStream(c, ctx1)
```

tests/integration/framework/process/daprd/daprd.go
cicoyle: makes sense to me, and will be helpful for us later 👍🏻

tests/integration/suite/placement/dissemination/streamhang.go
JoshVanL: Done!

JoshVanL: Done!

cicoyle (approved):
lgtm pending CI

yaron2 (approved):

yaron2:
Hallelujah

PR #8527, opened 2025-02-26
Fix unmarshalling legacy custom state

elena-kolevska:
If proto unmarshall errors out, the second attempt of unmarshalling should be to json


pkg/runtime/wfengine/state/state.go
```diff
@@ -350,11 +350,13 @@ func LoadWorkflowState(ctx context.Context, state state.Interface, actorID strin
 		wState.CustomStatus = &wrapperspb.StringValue{}
 		err = proto.Unmarshal(bulkRes[customStatusKey], wState.CustomStatus)
 		if err != nil {
+			// Fallback to JSON unmarshaling
+			var customStatusValue string
 			// TODO: @famarting: remove in v1.16
```
cicoyle, line 355:
```suggestion
```
Can the TODO comment be removed with this PR going into 1.15?

elena-kolevska: Since the TODO is for 1.16 I think we should keep it as a reminder, just so it doesn't slip. We have a similar comment on line 317.

JoshVanL (approved):

cicoyle (approved):

PR #8529, opened 2025-02-26
Actor: transcode period string getting Scheduler reminders

JoshVanL:
Transcode Scheduler reminder period strings on get and lists to transcode `@every` strings to the previous ISO8601 duration format to ensure backwards compatibility.


JoshVanL:
We are going to keep this string being returned for reminders, and update the SDKs to understand the new format.

PR #8530, opened 2025-02-26
Feat scheduler actor reminder metrics

cicoyle:
Add the `dapr_runtime_actor_reminders_fired_total` metric when using scheduler reminders. This is needed for supporting the metric when using the scheduler, as it was previously being supported with statestore reminders.

I removed the `actorerrors.ErrReminderCanceled` check bc that was prior to one shot wf reminders, so I think we should be able to remove it now - will confirm with IT tests.

There is no way to support the prior actor reminder metrics:
-`ActorTimerFired`
-`ActorReminders`
-`ActorTimers`
As the timer logic was from the `inmemory` code land previously and we don't really have a distinction btw timer and reminder anymore. The `ActorReminders` one isn't really feasible since we don't keep a global inmemory map anymore with scheduler since that logic is managed by scheduler + etcd and not in dapr runtime.

As a follow up, I think it would be easy to add CRUD actor reminder metrics, but that is out of scope for this PR.


tests/integration/suite/daprd/metrics/actors/actors.go
```diff
@@ -0,0 +1,106 @@
...
+	})
+	require.NoError(t, err)
+
+	assert.EventuallyWithT(t, func(c *assert.CollectT) {
+		assert.GreaterOrEqual(c, a.triggered.Load(), int64(1))
+	}, 30*time.Second, 10*time.Millisecond, "failed to wait for 'triggered' to be greater or equal 1, actual value %d", a.triggered.Load())
+
+	assert.EventuallyWithT(t, func(c *assert.CollectT) {
+		metrics := a.daprd.Metrics(c, ctx).All()
+		assert.GreaterOrEqual(c, a.triggered.Load(), int64(metrics["dapr_runtime_actor_reminders_fired_total"]))
```
JoshVanL, line 104: Shouldn't this be Equal?

pkg/runtime/scheduler/internal/cluster/streamer.go
```diff
@@ -118,15 +119,6 @@ func (s *streamer) handleJob(ctx context.Context, job *schedulerv1pb.WatchJobsRe
 
 	case *schedulerv1pb.JobTargetMetadata_Actor:
 		if err := s.invokeActorReminder(ctx, job); err != nil {
-			// this err is expected if the reminder was triggered once already, the next time it goes to get
```
JoshVanL, line 121: It's not worth changing this now I would say.

JoshVanL (changes requested):

tests/integration/suite/daprd/metrics/actors/actors.go
cicoyle: I did not set a repeat for the reminder, so I cannot guarantee the exact count, especially with CI timing variations, hence why I put the `GreaterOrEqual`. I can set a Repeat if desired, but we have other tests already ensuring correct trigger count at the same time.

cicoyle: We do the same for [this other test with no repeats set](https://github.com/dapr/dapr/blob/release-1.15/tests/integration/suite/scheduler/metrics/actors.go#L118)

JoshVanL: Yes, however surely the metric number should be the same as the number of triggers?

cicoyle: sure, I can change it for an additional check for count

cicoyle:
We should go thru the IT suite and see what other metrics are in the wrong place and move them here `tests/integration/suite/daprd/metrics`, also, we should really have IT tests for metrics for all the APIs... Maybe a good opportunity for community support with good first issues. I can take note of which ones do/don't and open issues accordingly.

JoshVanL (approved):

cicoyle:
that IT test failure is unrelated to my test in this PR

JoshVanL (changes requested):

cicoyle:
I ran the commands given in CI to fix DCO bc I forgot to sign the last commit and it looks like it signed other commits outside of this PR. Going to close this out and open a new one.

PR #8531, opened 2025-02-26
[1.15] Run Perf Tests

cicoyle:
This PR is simply used to run perf tests against the release branch


cicoyle:
I canceled that first run bc it should be `actor_reminder`

PR #8539, opened 2025-02-28
fix: workflow error string format

Eileen-Yu:
small fix for workflow error string formatting, current:
```
unable to find workflow with the provided instance ID: instanceID%!(EXTRA *fmt.wrapError=no such instance exists: failed to fetch orchestration metadata: error from internal actor: no such instance exists
```

after fix:
```
unable to find workflow with the provided instance ID: instanceID: no such instance exists: failed to fetch orchestration metadata: error from internal actor: no such instance exists
```

Also added a unit testcase for it.


cicoyle (approved):
Im fine to target the release branch if we are doing another rc, but then we will need to cherry-pick to master too please

cicoyle:
weird, tests show failing after merging in latest. They were passing when I initially approved

Issue #8537, opened 2025-02-27
Sidecar GRPC response size is limited and cannot be set


jbalmer89:
## In what area(s)?

 /area runtime
## What version of Dapr?

1.15.0-rc18

## Expected Behavior

We want to send responses to the dapr sidecar > 4Mb
## Actual Behavior

When receiving a message > 4 Mb we get an error as follows: 

Error: An unexpected error occurred invoking 'CreateMethod' on the server. DaprApiException: error invoke actor method: rpc error: code = ResourceExhausted desc = grpc: received message larger than max (9517075 vs. 4194304)

## Steps to Reproduce the Problem

We are experiencing issues where the dapr sidecar receives grpc messages greater than 4Mb. Normally this can be set on the grpc channel by changing MaxReceiveMessageSize, but it looks like there is no option in the CLI to set the size. There is max-body-size but this doesn't seem to have any effect.

## Release Note
RELEASE NOTE: **ADD** Set GRPC max recieve message size via CLI

RELEASE NOTE:

jbalmer89:
We also tried it with versions 1.13 and 1.14, we never managed to set the max size for the grpc communication

yaron2:
@jbalmer89 Can you ping me on Discord (handle is `yaron2`) so we can debug this step by step?

yaron2:
Closed via #8541

PR #8541, opened 2025-02-28
[1.15] Add missing grpc call options when invoking actors

yaron2:
Fixes https://github.com/dapr/dapr/issues/8537.

### DIFF
--- a/pkg/actors/actors.go
+++ b/pkg/actors/actors.go
@@ -71,7 +71,8 @@ type Options struct {
 	Healthz            healthz.Healthz
 	CompStore          *compstore.ComponentStore
 	// TODO: @joshvanl Remove in Dapr 1.12 when ActorStateTTL is finalized.
-	StateTTLEnabled bool
+	StateTTLEnabled    bool
+	MaxRequestBodySize int
 }
 
 type InitOptions struct {
@@ -110,7 +111,8 @@ type actors struct {
 	healthz            healthz.Healthz
 	compStore          *compstore.ComponentStore
 	// TODO: @joshvanl Remove in Dapr 1.12 when ActorStateTTL is finalized.
-	stateTTLEnabled bool
+	stateTTLEnabled    bool
+	maxRequestBodySize int
 
 	reminders      reminders.Interface
 	table          table.Interface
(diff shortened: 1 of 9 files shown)


pkg/actors/engine/engine.go
```diff
@@ -74,6 +76,12 @@ type engine struct {
 
 	lock  *fifo.Mutex
 	clock clock.Clock
+
+	actorCallOpts actorCallOptions
+}
+
+type actorCallOptions struct {
+	maxRequestBodySize int
```
JoshVanL, line 84: I don't it's necessary for us to have a new struct here for this

pkg/actors/engine/engine.go
```diff
@@ -74,6 +76,12 @@ type engine struct {
 
 	lock  *fifo.Mutex
 	clock clock.Clock
+
+	actorCallOpts actorCallOptions
```
JoshVanL, line 80: We can pre build the gRPC options on `New`.

```suggestion
	callOptions []grpc.CallOption
```

pkg/actors/engine/engine.go
```diff
@@ -89,6 +97,9 @@ func New(opts Options) Interface {
 		locker:             opts.Locker,
 		lock:               fifo.New(),
 		clock:              clock.RealClock{},
+		actorCallOpts: actorCallOptions{
+			maxRequestBodySize: opts.MaxRequestBodySize,
+		},
```
JoshVanL, line 102:
```suggestion
  callOptions: []grpc.CallOption{
		grpc.MaxCallRecvMsgSize(opts.MaxRequestBodySize),
		grpc.MaxCallSendMsgSize(opts.MaxRequestBodySize),
	},
```

JoshVanL (changes requested):

pkg/actors/actors.go
```diff
@@ -71,7 +71,8 @@ type Options struct {
 	Healthz            healthz.Healthz
```
JoshVanL, line 71: Please add a new `docs/release_notes/v1.15.1.md` for this change.

JoshVanL (changes requested):

pkg/actors/engine/engine.go
yaron2: Changed

pkg/actors/engine/engine.go
yaron2: Changed

pkg/actors/engine/engine.go
yaron2: Changed

pkg/actors/actors.go
yaron2: Added

JoshVanL (approved):

Issue #8538, opened 2025-02-22
Actor `TryGetStateAsync()`/ `GetStateAsync()`` is returning stale / previous state


olitomlinson:
Runtime 1.15.0-rc15
SDK 1.15.0-rc.04

## Expected Behavior

**Update** : I _think_ this is related to state operations that are performed within the context of a Reminder.

When a call is made to `this.StateManager.TryGetStateAsync()` then I expect the newest/current data to be retrieved.

## Actual Behavior

When a call is made to `this.StateManager.TryGetStateAsync()` then it is retrieving a cached/previous/stale version of the state.

![Image](https://github.com/user-attachments/assets/9dbebc27-fda5-4062-a815-fd221c5e5359)

However, if I use the Actor HTTP API to retrieve the same state key, I can see the new/current state returned as expected

![Image](https://github.com/user-attachments/assets/ae99f6e9-730f-47c6-8a63-c205be5928a2)

If I then kill the dapr/app process, and restart it, the Actor will now log out the correct state

![Image](https://github.com/user-attachments/assets/6b072a25-7494-4b58-82db-5f6212c9bd43)

## Release Note

RELEASE NOTE:

JoshVanL:
Hey all!

Thanks for the great write up of the issue 🙂 
I've added a fix for this here: https://github.com/dapr/dapr/pull/8542

WhitWaldo:
@JoshVanL Can this be closed as completed?

PR #8542, opened 2025-02-28
[v1.15] Actors: Don't sent Reentrancy ID header when disabled

JoshVanL:
Fix sending Actor Reentrancy ID HTTP header when disabled

Problem
Calling `this.StateManager.TryGetStateAsync()` in the DotNet Actor SDK would return stale data during some invocation scenarios.

Impact
The latest Actor state data was not being correctly returned during some Actor invocation scenarios using the DotNet Actor SDK.

Root cause
When Reentrancy was disabled, the Actor Reentrancy ID HTTP header was still being sent to Actor HTTP servers. The DotNet SDK uses the existence of this HTTP header in logic to determine what state should be returned to the Actor.

Solution
Don't send the Actor Reentrancy ID HTTP header (`"Dapr-Reentrancy-Id"`) when Reentrancy is disabled.

Fixes https://github.com/dapr/dapr/issues/8538#issuecomment-2689485765

### DIFF
--- a/pkg/actors/internal/locker/lock.go
+++ b/pkg/actors/internal/locker/lock.go
@@ -31,6 +31,8 @@ import (
 
 var ErrLockClosed = errors.New("actor lock is closed")
 
+const headerReentrancyID = "Dapr-Reentrancy-Id"
+
 type lockOptions struct {
 	actorType         string
 	reentrancyEnabled bool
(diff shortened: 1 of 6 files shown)


pkg/actors/internal/locker/lock.go
```diff
@@ -229,19 +231,19 @@ func newInflight(id string) *inflight {
 }
 
 func (l *lock) idFromRequest(req *internalv1pb.InternalInvokeRequest) (string, bool) {
-	if req == nil {
+	if !l.reentrancyEnabled || req == nil {
```
JoshVanL, line 234: This is the fix.

yaron2 (dismissed):

cicoyle (dismissed):

yaron2 (approved):

PR #8544, opened 2025-03-01
WIP: allow actor invocation without state store requirement

cicoyle:
Currently, Dapr instances without an actor state store configured are prevented from connecting to the placement service and marked as not-ready/unhealthy. This incorrectly prevents these instances from acting as actor clients, even when they only need to invoke actors on other hosts.

I still need to fix: `dapr is not becoming ready` and `dapr did not find address for actor`


cicoyle:
closing since [this PR](https://github.com/dapr/dapr/pull/8546) got merged

PR #8555, opened 2025-03-05
[1.15] Fix memory leak in Dapr scheduler

JoshVanL:
Fix Daprd continuously growing in memory

Problem

Fixes an issue where Daprd would continually grow in memory when using Workflows.

Impact

Daprd would eventually use all available memory on the node or cgroup, causing an OOM crash.

Root cause

An internal Actor lock object was not being released from Workflow Activities.

Solution

Release lock memory during Workflow Activity completion.

Fix Scheduler continuously growing in memory

Problem

Scheduler would continuously grow in memory when under heavy usage, for example using Workflows.

Impact

Scheduler would eventually use all available memory on the node or cgroup, causing an OOM crash.

Root cause

Etcd does not automatically Defragment after compaction, causing unused memory to not be released.

Solution

Every 10 minutes, each Scheduler host will check whether the total memory is twice the size of used memory and if so, will defragment that host's Etcd database.

# Description

## Issue reference

## Checklist

- [ ] Code compiles correctly
- [ ] Created/updated tests
- [ ] Unit tests passing

### DIFF
--- a/pkg/scheduler/server/internal/etcd/etcd.go
+++ b/pkg/scheduler/server/internal/etcd/etcd.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"time"
 
 	clientv3 "go.etcd.io/etcd/client/v3"
 	"go.etcd.io/etcd/server/v3/embed"
@@ -26,6 +27,7 @@ import (
 	"github.com/dapr/dapr/pkg/healthz"
 	"github.com/dapr/dapr/pkg/modes"
 	"github.com/dapr/dapr/pkg/security"
+	"github.com/dapr/kit/concurrency"
 	"github.com/dapr/kit/logger"
 )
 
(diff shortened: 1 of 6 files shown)


pkg/scheduler/server/internal/etcd/etcd.go
```diff
@@ -135,6 +142,38 @@ func (e *etcd) Client(ctx context.Context) (*clientv3.Client, error) {
...
+func (e *etcd) runDefrag(ctx context.Context) error {
+	for {
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case <-time.After(10 * time.Minute):
+			log.Debug("Checking if Etcd needs Defragmentation")
+			resp, err := e.client.Maintenance.Status(ctx, e.config.ListenClientUrls[0].Host)
+			if err != nil {
+				return err
```
yaron2, line 154: Will returning an error here and anywhere else in this method not cause the `runDefrag` logic to stop forever and ot repeat anymore when the cause of the error might be transient?

yaron2: Yeah, just trying to prevent a situation where transient errors cause a destructive event like scheduler shutdown

JoshVanL: :+1: done!

yaron2 (approved):

PR #8571, opened 2025-03-10
feat: enable richer details on app health status

sicoyle:
# Description

Right now, we use a uint8 to depict the app health connectivity status of healthy or unhealthy. This PR refactors this to be a struct to associate the time and reason for a given app health state change. This will better bubble up health status info to Dapr users and provide more meaning when debugging as to why this might have occurred. It also cleans up some of the conditional logic, making it easier to read with well defined helpers on app health status. We can then iterate on this refactor later to better leverage standardized errors even for app health checks :) 

## Issue reference

## Checklist

- [ ] Code compiles correctly
- [ ] Created/updated tests
- [ ] Unit tests passing


pkg/apphealth/status.go
```diff
@@ -0,0 +1,31 @@
+package apphealth
+
+import "time"
+
+const (
+	AppStatusUnhealthy uint8 = 0
+	AppStatusHealthy   uint8 = 1
```
JoshVanL, line 7: Can this just be a bool?

pkg/apphealth/status.go
```diff
@@ -0,0 +1,31 @@
+package apphealth
```
JoshVanL, line 1: Missing copyright headers.

pkg/apphealth/status.go
```diff
@@ -0,0 +1,31 @@
...
+
+func (s *Status) IsHealthy() bool {
+	return s.State == AppStatusHealthy
+}
+
+// DefaultStatus returns a default status for the app.
+// This defaults to healthy so subscriptions can start right away.
+func DefaultStatus() *Status {
+	return &Status{
+		State:    AppStatusHealthy,
```
JoshVanL, line 27: Should this not be false be default?

pkg/apphealth/status.go
```diff
@@ -0,0 +1,31 @@
...
+	Reason   *string `json:"reason,omitempty"`
+}
+
+func (s *Status) IsHealthy() bool {
+	return s.State == AppStatusHealthy
+}
+
+// DefaultStatus returns a default status for the app.
+// This defaults to healthy so subscriptions can start right away.
+func DefaultStatus() *Status {
```
JoshVanL, line 25:
```suggestion
// NewStatus returns a default status for the app.
// This defaults to healthy so subscriptions can start right away.
func NewStatus() *Status {
```

sicoyle:
looking into tests now

JoshVanL (changes requested):
More widely, it would be great for us in future to also name the `pkg/healthz Target` so that we can display better error messages describe what specifically is causing daprd to be not healthy.

pkg/apphealth/status.go
sicoyle: this was existing code that I only moved to a more logical place, so I left as is to minimize changes, but I suppose I could update

pkg/apphealth/status.go
sicoyle: no, there is a note somewhere in the code that it should default to healthy so subscriptions start up right away, so left as is to not bring in random changes, and just keep it a clean refactor :)

sicoyle: actually yeah i'll leave defaults alone and make sure the part accounting for this sets it to healthy

pkg/apphealth/status.go
sicoyle: went ahead and made bool instead bc yeah no value in uint8 type and it wasn't a big diff

sicoyle:
I'm seeing lots of weird build issues totally unrelated 😢 
```
Error: pkg/runtime/pubsub/publisher/publisher.go:95:35: undefined: rtpubsub.NewDefaultBulkPublisher

```
somehow the unit test tag was spelled wrong in a build workflow file but I'm working through other issues here that again idk how are coming up bc i didn't touch the stuff failing 🤷‍♀️

JoshVanL (changes requested):
Can we also add an integration test for this?

sicoyle:
ready cc @JoshVanL @cicoyle

pkg/apphealth/status.go
```diff
@@ -0,0 +1,39 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package apphealth
+
+import "time"
+
+const (
+	// reportMinInterval is the minimum interval between health reports.
+	reportMinInterval = time.Second
```
JoshVanL, line 21: q: why are we reporting the health at second interval? Seems wasteful to be waking this go routine when there is no change. Why do we not report only where there is a change?

JoshVanL (changes requested):

pkg/apphealth/status.go
sicoyle: this is existing code that I just moved to a more logic file. I did not write this nor change it to minimize changes that my PR introduced. This is a great follow up to be done in a separate issue/pr

JoshVanL: Can we change this now since we are already changing the data structure?

sicoyle: done

sicoyle:
I'm assuming the unit tests are flaky given the other ones ran fine in the build... can someone pls rerun to see?

JoshVanL (approved):

cicoyle (approved):

pkg/apphealth/status.go
coip: Bonus points imo for just `type Status uint8` in the right pkg or unexported for the plumbing/porcelain approach if I’m understanding the point of the repo

JoshVanL: Hello! :slightly_smiling_face: 

It is the case that the app channel health is a boolean Healthy or Unhealthy- either the channel is open for communication or it is not. The reason why it would be `Unhealthy`, i.e. with the reasons you have above, would go into `Reason`. Those could be typed/well known, or we create a new `Message string ` which has the "error" message and make `Reason` well typed- following kube convention.

coip: Howdy, hope alls well across the pond! 🤠

could def see that being a handy way to enable deeper integration, good thinkin w the k8s interface there. hope the ref/granularity is handy as y'all build more things and see that uint8 stuff 🤘 😎 there's cases for string/interfaces, but def out of context on this thread at this point 😄 josh, lmk if you open an issue or anything on the k8s status wrapper Id be curious to see it & might chime in. sounds interesting. the iota things something I contributed on a golang proposal thread for, so its something I try to keep... idk familiarity with I suppose.

cheers 🍻

coip: The nature of enums is where it’s interesting to incorporate the message (have done a number of error libs w this), since enums sort of implicitly lack a msg. Interesting to get into if it were not only a comment on this thread lol

PR #8582, opened 2025-03-12
Update go version to v1.24.1 & golangci-lint to v1.64.6

JoshVanL:
Update code for new linter checks

Upgrade `golang.org/x/net` `v0.34.0` -> `v0.36.0` to fix CVE

Fix golangci-lint version check.


JoshVanL:
`Standalone validations` is going to fail because it is being `daprd` from master, which has an incompatible dependency because of the Go upgrade to v1.24

https://github.com/dapr/dapr/actions/runs/13823311036/job/38673395212?pr=8582

antontroshin (approved):

cicoyle (approved):

Issue #8365, opened 2025-01-08
Set Placement service to runAsNonRoot when forceInMemoryLog is false


rochabr:
When looking at the Dapr Placement service, it looks to only honor the `runAsNonRoot` setting if `forceInMemoryLog` is true.

The request is to allow the service to `runAsNonRoot` if `forceInMemoryLog` is false.

alicejgibbons:
> We should fix this completely (for both Scheduler & Placement) for all deployment environments by creating a data directory in the container image itself with the correct file permissions. This involves doing a mkdir & chown for the container building process (docker build). I think the magic user group we need to set is 65532. The data directory will be in /run.. somewhere.

@elena-kolevska you still good to make this change requested by Josh to close this issue?

elena-kolevska:
I can't officially commit to this, but I can try my best to look into it in the next two weeks.

mikeee:
Docs needed

PR #8591, opened 2025-03-15
Don’t run Placement service as root by default (sets fsGroup)

elena-kolevska:
# Description

When `forceInMemoryLog` was false the Placement service would run as root (`runAsUser: 0`). This change removes that setting and ensures the service runs with non-root privileges regardless of storage configuration.

This matches the approach used in the Scheduler service.

## Issue reference

https://github.com/dapr/dapr/issues/8365


daixiang0:
Could you update title something like `set fsgroup as default`?

LGTM for change.

antontroshin (approved):

cicoyle (approved):

mikeee (approved):

Issue #8599, opened 2025-03-18
Slowness when a workflow is scheduled


AlbertoVPersonal:
## In what area(s)?

/area runtime

## What version of Dapr?

```powershell
CLI version: 1.15.0
Runtime version: 1.15.3
NuGet package: 1.15.2
````

## Expected Behavior

The service is executed on less than 50 miliseconds.

## Actual Behavior

The service is executed between 10-30 seconds.
After each iteration, the time is worse.

```powershell
> .\ConsoleGrpc.exe
Start: 17:28::19.253
End: 17:28::29.139

> .\ConsoleGrpc.exe
Start: 17:28::31.685
End: 17:28::46.055

> .\ConsoleGrpc.exe
Start: 17:28::48.927
End: 17:29::04.049

> .\ConsoleGrpc.exe
Start: 17:29::06.533
End: 17:29::21.038

> .\ConsoleGrpc.exe
Start: 17:31::17.594
End: 17:31::26.086
```

*UPD 18:09*: I'm repeating the same tests, and the time is better (about 5 seconds) but it does not achieve our goals.

## Steps to Reproduce the Problem

1. Run the project from the main repo.

My command line:

```powershell
cls | dapr run --app-id test --app-port 9099 --resources-path .\components\ --placement-host-address localhost:50002 --scheduler-host-address localhost:56003 -- .\DaprTestIssueWithSlowness.exe
```

2. Run a gRPC call using the following code (sub-folder in the repo). And the command line to call to the gRPC service:

```powershell
.\ConsoleGrpc.exe
```

## Workaround

I think that if we changed this line

```csharp
await daprWorkflowClient.ScheduleNewWorkflowAsync(nameof(MyWorkflow), s_instanceId, 0);
```

by this one, the time improves a bit.

```csharp
await daprWorkflowClient.ScheduleNewWorkflowAsync(nameof(MyWorkflow), s_instanceId, 0, DateTime.Now.AddMilliseconds(10));
```

---

## Release Note

RELEASE NOTE: **FIX** Solved a bug with the workflows execution due to an issue slowness

---

### Repository

**Main**: https://github.com/AlbertoVPersonal/DaprIssueWithWorkflows

**Sub-folder with ConsoleGrpc**: https://github.com/AlbertoVPersonal/DaprIssueWithWorkflows/tree/main/ConsoleGrpcForDapr

yaron2:
cc @cicoyle @JoshVanL

JoshVanL:
[...]

PR #8606, opened 2025-03-21
Scheduler: Fix client connection prune

JoshVanL:
Fixes an issue whereby Scheduler client (daprd) connections where not properly pruned from the connection pool for a given Namespace's appID/actorTypes set. This would lead to jobs/actor reminders being sent to stale client connections that were no longer active. This caused Jobs to fail, and enter failure policy retry loops.

This was particularly noticeable when running Workflow applications which were of single run/low number of iterations. Typically, every `dapr run` of these Workflow applications would leave a stale connection in the Scheduler pool. The end result would be Workflows which would take an artificial amount of time to complete, with longer run times the more times the application would be run.

This change refactors the Scheduler connection pool logic to properly abstract the connection instance handling, and moves logic into the connection pool store. Also fixes a bug whereby a cancelled actor reminder would be treated as a Failed trigger, rather than a Success.

This change should see more performant Workflow execution times in environments where there is any churn whatsoever.

This change should be backported to release-1.15.

Thanks to @AlbertoVPersonal for their in-depth issue report.

Fixes https://github.com/dapr/dapr/issues/8599

### DIFF
--- a/pkg/runtime/scheduler/internal/cluster/streamer.go
+++ b/pkg/runtime/scheduler/internal/cluster/streamer.go
@@ -118,29 +118,25 @@ func (s *streamer) handleJob(ctx context.Context, job *schedulerv1pb.WatchJobsRe
 		return schedulerv1pb.WatchJobsRequestResultStatus_SUCCESS
 
 	case *schedulerv1pb.JobTargetMetadata_Actor:
-		if err := s.invokeActorReminder(ctx, job); err != nil {
-			// this err is expected if the reminder was triggered once already, the next time it goes to get
-			// triggered by scheduler it sees that it's been canceled and does not invoke the 2nd time. This
-			// more relevant for workflows which currently has a repeat set to 2 since setting it to 1 caused
-			// issues. This will be updated in the future releases, but for now we will see this err. To not
-			// spam users only log if the error is not reminder canceled because this is expected for now.
-			if errors.Is(err, actorerrors.ErrReminderCanceled) {
-				return schedulerv1pb.WatchJobsRequestResultStatus_SUCCESS
-			}
-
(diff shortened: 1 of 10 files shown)


pkg/runtime/scheduler/internal/cluster/streamer.go
```diff
@@ -118,29 +118,25 @@ func (s *streamer) handleJob(ctx context.Context, job *schedulerv1pb.WatchJobsRe
 		return schedulerv1pb.WatchJobsRequestResultStatus_SUCCESS
 
 	case *schedulerv1pb.JobTargetMetadata_Actor:
-		if err := s.invokeActorReminder(ctx, job); err != nil {
```
JoshVanL, line 121: Needs release notes

pkg/scheduler/server/internal/cron/cron.go
```diff
@@ -222,9 +223,19 @@ func (c *cron) HostsWatch(stream schedulerv1pb.Scheduler_WatchHostsServer) error
 	}
 }
 
-func (c *cron) triggerJob(ctx context.Context, req *api.TriggerRequest) *api.TriggerResponse {
+func (c *cron) triggerHandler(ctx context.Context, req *api.TriggerRequest) *api.TriggerResponse {
 	log.Debugf("Triggering job: %s", req.GetName())
```
cicoyle, line 227: Should this log line be deleted since you added the one below? Otherwise we log twice for triggering the job - unless that was intentional?
```suggestion
```

JoshVanL: Yes it's intentional, so we get a _debug_ log, before and after we have completed execution.

yaron2 (approved):

Issue #8595, opened 2025-03-17
User thrown actor exceptions triggering built in retries


MrMint:
## In what area(s)?

> /area runtime

> /area operator

> /area placement

> /area docs

> /area test-and-release

## What version of Dapr?

1.15.3

## Expected Behavior

When an actor throws an exception from within a method, the calling service should receive the error without any retries.

## Actual Behavior

When the actor throws the exception, we are seeing 5 retries in the daprd logs before returning to the calling service.

```"message": "{\"error\":\"Internal Server Error\",\"error_msg\":\"{\\\"errorCode\\\":\\\"ERR_ACTOR_INVOKE_METHOD\\\",\\\"message\\\":\\\"error invoke actor method: failed to invoke target 192.168.20.59:59961 after 5 retries. Error: rpc error: code = Internal desc = error invoke actor method: error from actor service: (500) {}\\\"}\",\"status\":500}",```

<img width="1458" alt="Image" src="https://github.com/user-attachments/assets/29f56414-f92a-4213-b540-55ea59608a80" />

## Steps to Reproduce the Problem

Minimal Repro: https://github.com/MrMint/dapr-actor-bug

1. Create an actor with a method that will throw an exception/error.
2. Call that method from a separate client service.
3. Inspect daprd logs for retries.

## Release Note

RELEASE NOTE:

PR #8624, opened 2025-03-25
Actors: Don't retry remote calls returning 500

JoshVanL:
Fixes https://github.com/dapr/dapr/issues/8595

Should be backported to release-1.15

Fix remote Actor invocation 500 retry

Problem

An actor invocation across hosts which result in a 500 HTTP header response code would result in the request being retried 5 times.

Impact

Services which return a 500 HTTP header response code would result in requests under normal operation to return slowly, and request the service on the same request multiple times.

Root cause

The Actor engine considered a 500 HTTP header response code to be a retriable error, rather than a successful request which returned a non-200 status code.

Solution

Remove the 500 HTTP header response code from the list of retriable errors.

### DIFF
--- a/tests/integration/suite/actors/call/non200.go
+++ b/tests/integration/suite/actors/call/non200.go
@@ -1,0 +1,109 @@
+/*
+Copyright 2025 The Dapr Authors
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implieh.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package call
+
+import (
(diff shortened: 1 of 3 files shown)


JoshVanL:
cc @MrMint

tests/integration/suite/actors/call/non200.go
```diff
@@ -0,0 +1,109 @@
...
+		})
+		require.Error(t, err)
+		status, ok := status.FromError(err)
+		require.True(t, ok)
+		assert.Equal(t, "error invoke actor method: error from actor service: (500) custom error", status.Message())
+		assert.Equal(t, codes.Internal, status.Code())
+		assert.Equal(t, int64(i+1), n.called.Load())
+	}
+
+	time.Sleep(time.Second * 3)
```
cicoyle, line 83: Rather than putting a sleep here (and below), can we wrap this in an assert eventually?

JoshVanL: No because we are asserting that it doesn't change

cicoyle: ah ok

cicoyle (dismissed):

antontroshin (dismissed):

cicoyle (approved):

PR #8631, opened 2025-03-27
Update Maintainers List

cicoyle:
Adding myself and removing `artursouza` & `mukundansundar` from the maintainers list. I have been adding the `automerge` label, but was not in this list so it was not doing anything - this PR should fix that. Its ordered alphabetically now.


WhitWaldo:
You might also consider giving the list at the top of https://github.com/dapr/dapr/blob/master/.github/scripts/dapr_bot.js a pass to remove any absent code owners.

cicoyle:
> You might also consider giving the list at the top of https://github.com/dapr/dapr/blob/master/.github/scripts/dapr_bot.js a pass to remove any absent code owners.

Good catch. I forgot about that. Let me update it too

.github/scripts/dapr_bot.js
```diff
@@ -2,34 +2,16 @@
 // IMPORTANT: Make sure usernames are lower-cased
```
cicoyle, line 2: If more people should be removed lmk

yaron2 (changes requested):
Italypaleale should be removed too

JoshVanL (approved):

PR #8640, opened 2025-03-31
[1.15] Fix Scheduler Client connection pruning

JoshVanL:
Fix Scheduler Client connection pruning

Problem

Daprd would attempt to connect to stale Scheduler addresses.

Impact

Network resource usage and error reporting from service mesh sidecars.

Root cause

Daprd would not close Scheduler gRPC connections to hosts which no longer exist.

Solution

Daprd now closes connections to Scheduler hosts when they are no longer in the list of active hosts.

### DIFF
--- a/pkg/runtime/scheduler/internal/clients/clients.go
+++ b/pkg/runtime/scheduler/internal/clients/clients.go
@@ -17,6 +17,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"sync"
 	"sync/atomic"
 
 	schedulerv1pb "github.com/dapr/dapr/pkg/proto/scheduler/v1"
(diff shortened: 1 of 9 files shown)


pkg/runtime/scheduler/internal/cluster/connector.go
```diff
@@ -64,7 +64,7 @@ func (c *connector) run(ctx context.Context) error {
 
 	err = (&streamer{
 		stream:   stream,
-		resultCh: make(chan *schedulerv1pb.WatchJobsRequest),
+		resultCh: make(chan *schedulerv1pb.WatchJobsRequest, 10),
```
antontroshin, line 67: What's the reason behind `10` specifically?

JoshVanL: Was doing some testing, have removed.

pkg/runtime/scheduler/internal/clients/clients.go
```diff
@@ -77,3 +81,9 @@ func (c *Clients) Next() (schedulerv1pb.SchedulerClient, error) {
 func (c *Clients) All() []schedulerv1pb.SchedulerClient {
 	return c.clients
 }
+
+func (c *Clients) Close() {
+	for _, closeFn := range c.closeFns {
+		closeFn()
+	}
```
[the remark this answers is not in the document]
JoshVanL: Yeah makes sense, happy to make this concurrent!

antontroshin (approved):

PR #8658, opened 2025-04-09
Actors: Vanity rename engine -> router

JoshVanL:
Vanity package and struct rename of actor `engine` to `router`. The name `router` far better fits the purpose of this package since its main/sole focus is routing messages to correct targets. The whole `/pkg/actor` package hierarchy should be considered the engine.


pkg/runtime/wfengine/state/state.go
```diff
@@ -116,6 +116,11 @@ func (s *State) AddToInbox(e *backend.HistoryEvent) {
 	s.inboxAddedCount++
 }
 
+func (s *State) AddToHistory(e *backend.HistoryEvent) {
```
cicoyle, line 119: Is this an intentional add? I dont see any tests for it or references in this PR other than the diff here.

JoshVanL: Thanks, good catch. Removed.

cicoyle (approved):

PR #8666, opened 2025-04-15
ci: bring back dapr alphas

nelson-parente:
# Description

Previous work:
- https://github.com/dapr/dapr/pull/8267 : Added alpha versions
- https://github.com/dapr/dapr/pull/8293 : Stopped publishing due to the CLI installing them as the latest

The motivation is the same as before: publishing earlier non-RC versions for earlier testing.

This shouldn't bring any of the issues we saw in the past with the CLI, since this is considered a pre-release version:
https://github.com/dapr/cli/blob/master/pkg/version/version.go#L122-L136

```
> sh .github/scripts/tag.sh
Latest tag: 1.15.4
Next alpha: 1.16.0-alpha.20250415

```

My main concern here is that we will be creating a new tag per day. I don't mind adding a step that clears all alpha tags except the last X ones.

## Issue reference

There is an issue for this in the 1.16 board but is in draft so I can't reference the number 🤷 

https://github.com/orgs/dapr/projects/89/views/1?pane=issue&itemId=105561623

## Checklist

- [ ] Code compiles correctly
- [ ] Created/updated tests
- [ ] Unit tests passing

### DIFF
--- a/.github/workflows/dapr.yml
+++ b/.github/workflows/dapr.yml
@@ -517,7 +517,7 @@ jobs:
       - name: lists artifacts
         run: ls -l ${{ env.ARTIFACT_DIR }}
       - name: publish binaries to github
-        if: startswith(github.ref, 'refs/tags/v') && ${{ env.ALPHA_RELEASE }} != "true"
+        if: startswith(github.ref, 'refs/tags/v')
         run: |
           echo "installing github-release-cli..."
           sudo npm install --silent --no-progress -g github-release-cli@2.1.0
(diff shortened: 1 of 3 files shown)


.github/workflows/dapr.yml
mikeee, line 1: Looking good, I would have expected some significant changes to the main dapr workflow (or refactoring the create release workflow) to ensure that we are not overriding the `:latest` tag. Is this not necessary?

nelson-parente: Let me double-check that!

nelson-parente:
Reopened this PR and implemented the changes I mentioned in my last comment on the issue. Let me know if anything else is needed!

alicejgibbons:
@mikeee can you take a peek at this one again 🙏

mikeee (commented):
Thinking about the implementation, I think it'd be pertinent to follow best practices - having a clear separation of concerns with a separate release script that deals solely with the alpha releases for a clear script path.

Just thinking ahead to de-risk this further.

Wdyt?

Issue #8402, opened 2025-01-22
Don't overwrite jobs using same name


WhitWaldo:
/area runtime

## Describe the proposal
I would like to propose a slight change to the Jobs API. Today, if a job is scheduled with the name "MyJob" and another inadvertently scheduled tomorrow using the same name, Dapr will overwrite it without hesitation, removing the previous job and its payload irrevocably. Today's approach means that intentionally scheduled jobs that are documented only by their payloads can be inadvertently deleted by something as accidental as having an insufficiently unique name on a service.

I would like to request a change to this behavior while the Jobs API is still immature.

I propose that once a job is created, it can only be removed via the delete method. This reflects a sense of intentionality on the part of the developer to remove an existing job and replace with another.  An attempt to replace an existing job should fail with a status code such as Conflict that gives information to the SDKs to properly share such feedback with developers.

### Example
Here's an example scenario demonstrating this issue based on how this is handled today:

Say a service wants to trigger a backup that its data on a regular basis, so it registers a job called "backup". The payload includes the name of the service, the name of the external state store to persist the backup to and a directory to write it into.

Another developer working on another service in the same app has a similar idea and also registers a job called "backup", but there's no payload necessary as the backup details are included in the app configuration.

Both services are deployed to production and upon launch, only one ever performs any backups because the other had its job overwritten. Something happens and they relaunch in a different order and the winner flips, but still, only one ever receives backups and the never shows anything in the logs indicating anything was amiss - the job was successfully scheduled and only one of them ever sees trigger invocations.

### Proposed Example
Take the example above with my proposed change to the API - one service will register its backup and the other will show a log indicating that the registration failed because the name is already in use. The developers will have a clear understanding from the logs as to what happened and what the path forward is to resolving the issue.

olitomlinson:
+1

ajaysingh77:
+1

WhitWaldo:
@javier-aliaga Do note that I've already got a PR to address this but which needs to instead use the new AddOrUpdate method on the etcd fork instead of the approach I took. But you're welcome to take over if you'd like.

javier-aliaga:
thanks @WhitWaldo, I have already followed that approach. The PR is still in draft, but you can take a look

javier-aliaga:
[...]

PR #8672, opened 2025-04-21
Prevent overwriting existing job when creating another

WhitWaldo:
## Description

When scheduling a job, if it already exists, throw an error and require that if the user wishes to replace it, they must explicitly delete it first.

## Issue reference

Please reference the issue this PR will close: #8402

## Checklist

- [X] Code compiles correctly
- [ ] Created/updated tests
- [ ] Unit tests passing


pkg/scheduler/server/api.go
```diff
@@ -39,6 +39,13 @@ func (s *Server) ScheduleJob(ctx context.Context, req *schedulerv1pb.ScheduleJob
 		return nil, err
 	}
 
+	_, err = cron.Get(ctx, serialized.Name())
```
cicoyle, line 42: Hey @WhitWaldo thanks for the PR and initiative here 🎉 

Did you by chance see [this PR](https://github.com/diagridio/go-etcd-cron/pull/74/files)? My understanding was that we will want to expose to users the option to simply `Add` or `AddIfNotExists` following the function names, basically overwrite the job or not. They should be able to toggle it, then we will want to call the functions exposed from that PR instead of doing a `Get` call.

Issue #8667, opened 2025-04-16
Event metadata/headers are not available via streaming subscription


akotsarelation:
## In what area(s)?

> /area runtime

## What version of Dapr?

> 1.15.4

## Expected Behavior

When using streaming subscriptions with a `pubsub.kafka` component, I expect to be able to receive the headers associated with the message, similar to how they are available as HTTP headers (e.g. `__key` or `value_format`) when using an HTTP-based subscription.

I expect the metadata/headers to be available via `TopicEventRequest.Extensions` and the payload to be part of the `TopicEventRequest.Data`.

## Actual Behavior

None of the headers make it through to the client (via `TopicEventRequest`). The message payload is included in the `Extensions` field. The `Data` field is always empty.

I was able to trace the issue to the [GRPCEnvelopeFromSubscriptionMessage method](https://github.com/dapr/dapr/blob/master/pkg/runtime/pubsub/subscriptions.go#L315), which does NOT use any of the information in `msg.Metadata`, where the headers information is located. It is only using `msg.Metadata[MetadataKeyPubSub]` to extract the pubsub name and all the remaining information is ignored.

The [publishMessageGRPC method](https://github.com/dapr/dapr/blob/master/pkg/runtime/subscription/publish.go#L136) is also using `GRPCEnvelopeFromSubscriptionMessage`. However, it then adds the metadata via `invokev1.WithCustomGRPCMetadata(ctx, msg.Metadata)` after invoking `GRPCEnvelopeFromSubscriptionMessage`. This does not happen for streaming subscriptions.

Also, unlike [publishMessageHTTP](https://github.com/dapr/dapr/blob/master/pkg/runtime/subscription/publish.go#L40), which is using `SubscribedMessage.Data`, `GRPCEnvelopeFromSubscriptionMessage` is using `cloudEvent` data to extract the payload. However, [due to the way cloudEvent data is prepared](https://github.com/dapr/dapr/blob/e4b2cd69cd4bf73496af4755b1c9e89dc26366e6/pkg/runtime/subscription/subscription.go#L197), full event payload becomes the cloudEvent object instead of it being under `cloudEvent["data"]`, as `GRPCEnvelopeFromSubscriptionMessage` expects it to be. As a result, the produced `TopicEventRequest` object does not have any `Data` associated with it and all the payload is interpreted as `Extensions`, leaving no room for metadata/headers even if it was used.

### Potential fix / Local experimentation (to give more context)

[...]

PR #8676, opened 2025-04-23
fix: missing pubsub streaming metadata & incorrect placement of event payload in gRPC messages

akotsarelation:
# Description

This resolves the issue of missing pubsub event metadata when using Subscription Streaming and the event payload passed via `TopicEventRequest.Extensions` instead of being in `TopicEventRequest.Data` (see https://github.com/dapr/dapr/issues/8667).

Below is an overview of all the changes and how they affect different subscription types.

- HTTP subscription
    - Regular payload (Avro or Json)
        - ✅ No change
    - Raw payload
        - ✅ No change
- Regular gRPC subscription
    - Regular payload (Avro or Json)
        - ✅ No change
    - Raw payload
        - ✅ No change
- gRPC Streaming
    - Regular payload (Avro or Json)
        - `message.data()`:
            - ✅ No change (`None`)
        - ⚠️ `message.extensions()`:
            - Before: *deserialized event payload*
                
                ```python
                {
                	'payload1': 'data1',
                	'payload2': 123,
                }
                ```
                
            - After: *deserialized object with*:
                - event payload
                - event metadata prefixed with `_metadata_`
                
                ```python
                {
                	'payload1': 'data1',
                	'payload2': 123,
                	'_metadata_pubsubname': 'kafka-pubsub',
                	'_metadata___key': 'local:131074',
                	'_metadata___offset': '0',
                	'_metadata_value_format': 'avro'
                	// other metadata fields
                }
                ```
                
    - Raw payload
        - `message._data_content_type`:
            - ✅ No change (`application/octet-stream`)
        - `message.data()`:
            - ✅ No change (`None`)
        - `message._raw_data`:
            - ✅ No change (*json serialized event payload*)
            
            ```python
            b'{"payload1":"data1","payload2":123}'
            ```
            
        - ⚠️ `message.extensions()`:
            - Before: *deserialized object with only `pubsubname` and `topic` fields*
                
                ```python
                {
                	'pubsubname': 'kafka-pubsub',
                	'topic': 'el8.snapshot.practice.physician'
                }
                ```
                
            - After: *deserialized object with*:
                - `pubsubname` and `topic` fields
                - event metadata prefixed with `_metadata_`
                
                ```python
                {
                	'pubsubname': 'kafka-pubsub',
                	'topic': 'el8.snapshot.practice.physician'
                	'_metadata_value_format': 'avro',
                	'_metadata___key': 'local:131074',
                	'_metadata___offset': '9',
                	// other metadata fields
                }
                ```

## Issue reference

https://github.com/dapr/dapr/issues/8667

## Checklist

- [X] Code compiles correctly
- [X] Created/updated tests
- [X] Unit tests passing

(_The docs issues will be created once the future of this PR is clear_)


passuied:
@JoshVanL any chance you could look at this one before 1.16?

yaron2:
See conflicts. Also, does this cause any breaking change?

passuied:
> See conflicts. Also, does this cause any breaking change?

Yes but the initial implementation was wrong... This is why I think the only time we could introduce this change is before the API gets stable...

PS: Andrey will resolve conflicts Today

akotsarelation:
> See conflicts. Also, does this cause any breaking change?

The conflicts are resolved now. 

Regarding breaking changes - it does contain changes that _may_ be breaking for some consumers. Primarily around gRPC streaming. I did my best to mark all the potential changes in the PR description, they're marked with ⚠️. Previously, when using with gRPC (regular or streaming) in non-raw mode, the `event.Data` field was always empty. All the payload data was in the `event.Extensions` field mixed together with metadata. In case of event streaming, the metadata was missing entirely. I.e, the `Data` field would be empty and `Metadata` would contain the actual data.

Note, the primary PubSub use case of using HTTP endpoints is completely unchanged.

akotsarelation:
Re-based and force-pushed to include commit signatures.

yaron2:
For the route of gRPC regular, I would keep the extension field as it was before to prevent breaking changes for existing users.

pkg/runtime/subscription/subscription.go
```diff
@@ -206,6 +207,30 @@ func New(opts Options) (*Subscription, error) {
...
+					allKeysPresent = false
+					break
+				}
+			}
+			if allKeysPresent {
+				cloudEvent = cloudEventData
+			} else {
+				cloudEvent = make(map[string]interface{})
+				cloudEvent[contribpubsub.DataField] = cloudEventData
+				cloudEvent[contribpubsub.DataContentTypeField] = "application/json"
```
yaron2: Is this overriding any user provided value for this field?

akotsarelation: No. In this code branch `cloudEvent` is an empty map and we're setting only the data and content type fields of the cloud event. We also know for sure that by that time the data is JSON because we just deserialized it.

akotsarelation:
> For the route of gRPC regular, I would keep the extension field as it was before to prevent breaking changes for existing users.

I changed the approach. It is now only affecting Streaming and should be fully backward compatible. The main downside is that all 3 mechanisms (HTTP, gRPC, and Streaming) are using different ways of passing the metadata.

pkg/runtime/subscription/subscription.go
passuied: Thanks @akotsarelation. @yaron2 do you mind taking another look?

PR #8702, opened 2025-05-07
Workflow: Rerun workflow instance from timers

JoshVanL:
Branched from https://github.com/dapr/dapr/pull/8699

    Updates the Workflow rerun rpc to enable rerunning a workflow from an
    event ID which targets a timer (or raise event timer). Also enables
    tracking active timers in a fanout scenario and creates the accordingly.

    Does not add support for targeting suborchestations.

    Refactors the workflow and activity actor targets into a new `workflow`
    target package structure.


pkg/actors/targets/workflow/orchestrator/activity.go
```diff
@@ -11,7 +11,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package workflow
+package orchestrator
```
cicoyle, line 14: Are you moving everything into the `orchestrator` folder and adding that folder to align with durabletask naming conventions a bit more? Just curious, I personally liked it under the `workflow` folder. imo orchestrator is basically the same thing as workflow... I'm working on the cross app bits, and I have to change things in these files, so moving them around is just an extra step for me later on conflict resolution wise.

JoshVanL: Yes the idea was to move it to have naming which is closer to durabletask- workflow is the overall feature, orchestrator is the top level concept driving an instance of a workflow.

I have moved both actor packages under `/workflow` so that there is better namespacing in the event other actor target implementations are added :crossed_fingers:

pkg/actors/targets/workflow/orchestrator/run.go
```diff
@@ -132,122 +127,56 @@ func (w *workflow) runWorkflow(ctx context.Context, reminder *actorapi.Reminder)
...
+		default:
+			return todo.RunCompletedTrue, fmt.Errorf("workflow actor '%s': don't know how to process outbound message '%v'", o.actorID, msg)
 		}
 	}
 
-	wg.Wait()
-	if len(errs) > 0 {
-		return compl, errors.Join(errs...)
+	if err = o.callAddEventStateMessage(ctx, addWorkflows); err != nil {
+		return todo.RunCompletedTrue, err
```
cicoyle: Shouldn't this return false? I think before it returned false
```suggestion
		return todo.RunCompletedFalse, err
```

pkg/actors/targets/workflow/orchestrator/run.go
```diff
@@ -132,122 +127,56 @@ func (w *workflow) runWorkflow(ctx context.Context, reminder *actorapi.Reminder)
...
+	}
+
+	if err = o.callCreateWorkflowStateMessage(ctx, createWorkflows); err != nil {
+		return todo.RunCompletedTrue, err
```
cicoyle:
```suggestion
		return todo. RunCompletedFalse, err
```

JoshVanL:
Thanks @cicoyle!

PR #8721, opened 2025-05-20
chore: Use custom round robin implementation for placement raft servers

javier-aliaga:
# Description
- Avoid logging errors when the client is trying to find the leader
- Use custom round robin placement connector as grpc client may not see the leader due to cache

## Proof of Work
- Log is printed with debug level
imagen

- Round robin across all 3 pods
imagen

## Issue reference

- https://github.com/dapr/dapr/issues/8724
- 

## Checklist

- [x] Code compiles correctly
- [x] Created/updated tests
- [x] Unit tests passing


Makefile
```diff
@@ -272,6 +272,15 @@ ADDITIONAL_HELM_SET ?= ""
 ifneq ($(ADDITIONAL_HELM_SET),)
 	ADDITIONAL_HELM_SET := --set $(ADDITIONAL_HELM_SET)
 endif
+ifneq ($(RUN_AS_NON_ROOT),)
```
javier-aliaga: This is useful if you want to ssh the containers. You can set DEBUG=1 and the images will be built with golang base image instead of distroless images

pkg/actors/internal/placement/client/client.go
```diff
@@ -160,10 +178,17 @@ func (c *Client) Run(ctx context.Context) error {
 
 		cancel()
 		c.ready.Store(false)
+		c.retryCount++
 		// Re-enable once healthz of daprd is not tired to liveness.
 		// c.htarget.NotReady()
 
-		log.Errorf("Error communicating with placement: %s", err)
+		if status.Code(err) == codes.FailedPrecondition && c.retryCount%100 == 0 {
```
javier-aliaga: We will log this as an error every 100 failures, just in case there is no leader for some reason

pkg/actors/internal/placement/client/client.go
```diff
@@ -160,10 +178,17 @@ func (c *Client) Run(ctx context.Context) error {
 
 		cancel()
 		c.ready.Store(false)
+		c.retryCount++
 		// Re-enable once healthz of daprd is not tired to liveness.
 		// c.htarget.NotReady()
 
-		log.Errorf("Error communicating with placement: %s", err)
+		if status.Code(err) == codes.FailedPrecondition && c.retryCount%100 != 0 {
```
javier-aliaga: Log as an error every 100 retries

pkg/actors/internal/placement/client/client.go
```diff
@@ -91,20 +100,29 @@ func New(opts Options) (*Client, error) {
 		)
 	}
 
+	var rr connector.Interface
 	if len(addresses) == 1 && strings.HasPrefix(addresses[0], "dns:///") {
 		// In Kubernetes environment, dapr-placement headless service resolves multiple IP addresses.
 		// With round robin load balancer, Dapr can find the leader automatically.
 		gopts = append(gopts, grpc.WithDefaultServiceConfig(grpcServiceConfig))
+		rr, err = roundrobin.New(roundrobin.Options{
```
JoshVanL: If we are using a custom dns resolver, then we should be removing the gRPC service config option and const?

JoshVanL: Using round robin connector to be predicated on the fact that we are in Kubernetes (I.e. a passed down bool, rather than magic string match?)

pkg/actors/internal/placement/client/client.go
JoshVanL: 100 seems high to me. 10?

pkg/actors/internal/placement/client/client.go
```diff
@@ -239,24 +264,31 @@ func (c *Client) connect(ctx context.Context) error {
...
-	// If we're trying to connect to the same address, we reuse the existing
-	// connection.
-	// This is not just an optimisation, but a necessary feature for the
-	// round-robin load balancer to work correctly when connecting to the
-	// placement headless service in k8s
+
+	// In Kubernetes, we use our custom round-robin implementation.
+	// gRPC library does not do the placement round-robin correctly.
+	// It seems it will not invalidate its local DNS cache and will never see a leader potentially.
+	if len(c.addresses) == 1 && strings.HasPrefix(c.addresses[0], "dns:///") {
```
JoshVanL: Please can we pass a bool down and not use magic string checks?

pkg/actors/internal/placement/client/client.go
```diff
@@ -239,24 +264,31 @@ func (c *Client) connect(ctx context.Context) error {
...
 		}
 	}
 
 	client, err := v1pb.NewPlacementClient(c.conn).ReportDaprStatus(ctx)
 	if err != nil {
-		err = fmt.Errorf("failed to create placement client: %s", err)
+		err = fmt.Errorf("failed to create placement client: %w", err)
 		if strings.Contains(err.Error(), "connect: connection refused") {
-			// reset gRPC connenxt to reset the round robin load balancer state to
+			// reset gRPC context to reset the round robin load balancer state to
```
JoshVanL: This doesn't seem true anymore

pkg/actors/internal/placement/client/roundrobin/roundrobin.go
```diff
@@ -0,0 +1,91 @@
+/*
```
JoshVanL: Move package under /connector

pkg/actors/internal/placement/client/client.go
```diff
@@ -22,6 +22,13 @@ import (
 	"sync/atomic"
 	"time"
 
+	"github.com/dapr/dapr/pkg/actors/internal/placement/client/connector"
+
+	"github.com/dapr/dapr/pkg/actors/internal/placement/client/roundrobin"
+
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+
```
JoshVanL: Fix import ordering

pkg/actors/internal/placement/client/roundrobin/roundrobin.go
```diff
@@ -0,0 +1,91 @@
...
+type roundrobin struct {
+	dnsEntries []string
+	host       string
+	port       string
+	gOpts      []grpc.DialOption
+	resolver   *net.Resolver
+}
+
+func New(opts Options) (connector.Interface, error) {
+	addr := strings.TrimPrefix(opts.Address, "dns:///")
```
JoshVanL: Seems clear to me that we can remove the dns string prefix now?

pkg/actors/internal/placement/client/roundrobin/roundrobin.go
```diff
@@ -0,0 +1,91 @@
...
+
+	return conn, nil
+}
+
+func (r *roundrobin) refreshEntries(ctx context.Context) error {
+	addrs, err := r.resolver.LookupHost(ctx, r.host)
+	if err != nil {
+		return fmt.Errorf("failed to lookup addresses: %w", err)
+	}
+
```
JoshVanL: We must also error if it returns 0 entries. We are currently panicking in this case.

pkg/actors/internal/placement/client/roundrobin/roundrobin_test.go
```diff
@@ -0,0 +1,38 @@
+/*
+Copyright 2024 The Dapr Authors
```
JoshVanL:
```suggestion
Copyright 2025 The Dapr Authors
```

tests/dapr_tests.mk
```diff
@@ -279,6 +279,10 @@ setup-pubsub-subs-perf-test-components: setup-test-env-rabbitmq setup-test-env-p
 
 e2e-build-deploy-run: create-test-namespace setup-3rd-party build docker-push docker-deploy-k8s setup-test-components build-e2e-app-all push-e2e-app-all test-e2e-all
 
+e2e-build-deploy: create-test-namespace setup-3rd-party build docker-push docker-deploy-k8s setup-test-components build-e2e-app-all push-e2e-app-all
```
JoshVanL: Why are these different to other existing targets?

Makefile
```diff
@@ -272,6 +272,15 @@ ADDITIONAL_HELM_SET ?= ""
...
+	ADDITIONAL_HELM_SET := --set dapr_operator.runAsNonRoot=$(RUN_AS_NON_ROOT) $(ADDITIONAL_HELM_SET)
```
JoshVanL: I don't fully understand the point of these additions, and how that relates to the base image

pkg/actors/internal/placement/client/client.go
```diff
@@ -239,24 +264,31 @@ func (c *Client) connect(ctx context.Context) error {
...
-	// connection.
-	// This is not just an optimisation, but a necessary feature for the
-	// round-robin load balancer to work correctly when connecting to the
-	// placement headless service in k8s
+
+	// In Kubernetes, we use our custom round-robin implementation.
+	// gRPC library does not do the placement round-robin correctly.
+	// It seems it will not invalidate its local DNS cache and will never see a leader potentially.
+	if len(c.addresses) == 1 && strings.HasPrefix(c.addresses[0], "dns:///") {
+		c.conn, err = c.roundRobinConn.Connect(ctx)
```
JoshVanL: Please can we create a non dns host lookup round robin client connector, so we don't need these if check special paths.

pkg/actors/internal/placement/client/client.go
```diff
@@ -57,12 +64,14 @@ type Client struct {
 	table table.Interface
 	lock  *lock.OuterCancel
 
-	conn      *grpc.ClientConn
-	client    v1pb.Placement_ReportDaprStatusClient
-	sendQueue chan *v1pb.Host
-	recvQueue chan *v1pb.PlacementOrder
+	conn           *grpc.ClientConn
```
JoshVanL: We shouldn't cache a connection because the connector should be returning one.

Makefile
javier-aliaga: if you run `make build` with `export DEBUG=1` images are built using `https://github.com/dapr/dapr/blob/master/docker/Dockerfile-debug`. If you try to deploy that images it it will complain as they are using nonRootUser so to be able to deploy to the cluster you also need to set the runAsNonRoot as true. I can demo it on friday afternoon meeting.

JoshVanL: In this case, please can we split out this work to another PR as this is unrelated to tje placement client change.

It should be the case that all of our images are non root and deployed as such.

tests/dapr_tests.mk
javier-aliaga: `e2e-build-deploy` I use this one to initialize the cluster but do not run e2e
`build-deploy` When the cluster is initialize and you only want to build and deploy dapr components

javier-aliaga: I will also move this out of this commit

pkg/actors/internal/placement/client/client.go
javier-aliaga: @JoshVanL I found of this a bit confusing. We are adding dns to all addresses that are not IPs, but only checking this if there is only one with DNS prefix. What happens with multiple dns addresses? How it should behave? Should we error in that case?

JoshVanL: There are two modes which we want to cater for-
- In Kubernetes, we use a headless service whereby the remote host addresses are discovered via `LookupHost`. We round robin over the _discovered_ hosts. We cache the DNS host response, but only attempt connection once. When we're out of addresses, we re-lookup to populate our local host address cache again. For Kubernetes mode, we must have at least one address given, and we use the first one in the list (practically, we will always have a single host address set.
- In non-Kubernetes mode, we will be given a static list of addresses to connect to. In this mode we the same round robin, however in this case there is no custom host lookup & cache functionality.

Technically, these host addresses supplied can be named DNS hosts, or simply IPs.
The `dns://` protocol suffix is used by the gRPC library that it should use the DNS resolver. This was needed to enable the gRPC round robin service config, which we are now no longer using.

javier-aliaga: yes, but how we know we are running in kubernetes mode. If there is only one address and it is not an IP we assume it is a Kubernetes mode?  What happens if there are multiple addresses, then it is a non Kubernetes mode?

javier-aliaga: something like this

```
func isKubernetesMode(addr []string) bool {
	if len(addr) != 1 {
		return false
	}

	host, _, err := net.SplitHostPort(addr[0])
	if err == nil && net.ParseIP(host) == nil {
		return true
	}

	return false
}
```

JoshVanL: It is passed in as a CLI flag https://github.com/dapr/dapr/blob/c0279fbe579ec76ed7f884ccec2fcd99eaa6684b/cmd/placement/options/options.go#L132

javier-aliaga: but that is not in the sidecar. Does the sidecar know which mode is the placement server running?

javier-aliaga:
@JoshVanL I did a refactor with your comments, take a look when you can

pkg/actors/internal/placement/client/client.go
JoshVanL: We do! it's here :slightly_smiling_face: 

https://github.com/dapr/dapr/blob/939ef414b6ddf2297b5b21b9bed0109910a7aac9/cmd/daprd/options/options.go#L121

pkg/actors/internal/placement/client/client.go
```diff
@@ -18,10 +18,16 @@ import (
...
 	"sync/atomic"
 	"time"
 
+	"github.com/dapr/dapr/pkg/actors/internal/placement/client/connector/roundrobin"
+
+	"github.com/dapr/dapr/pkg/actors/internal/placement/client/connector"
+
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+
```
JoshVanL: Move imports into blocks- stdlib, external, internal

pkg/actors/internal/placement/client/connector/roundrobin/dnsroundrobin.go
```diff
@@ -0,0 +1,84 @@
+/*
```
JoshVanL: Please can we split out `dnsroundrobin` and `staticroundrobin` into 2 separate sub packages of `connector` - `dsnlookup` and `static` respectively?

This means we can use `New` and `Options` type names in the constructors, leaving context to the package path/name.

pkg/actors/internal/placement/client/connector/roundrobin/staticroundrobin.go
```diff
@@ -0,0 +1,58 @@
...
+type StaticOptions struct {
+	GRPCOptions []grpc.DialOption
+	Addresses   []string
+}
+
+func (r *staticRoundRobin) Connect(ctx context.Context) (*grpc.ClientConn, error) {
+	r.addressIndex++
+	if r.addressIndex >= len(r.addresses) {
+		r.addressIndex = 0
+	}
```
JoshVanL:
```suggestion
	r.addressIndex = (r.addressIndex+1)%len(r.addresses)
```

pkg/actors/internal/placement/client/connector/roundrobin/dnsroundrobin.go
```diff
@@ -0,0 +1,84 @@
...
+	"context"
+	"fmt"
+	"net"
+
+	"google.golang.org/grpc"
+)
+
+type LookUpResolver interface {
+	LookupHost(ctx context.Context, host string) (addrs []string, err error)
+}
```
JoshVanL: nit: don't export, and don't need an interface for a single func.

```suggestion
type lookupFunc (ctx context.Context, host string) (addrs []string, err error)
```

JoshVanL (changes requested):
Looks good, just a couple of nits, and piping the `mode`.

pkg/actors/internal/placement/client/connector/roundrobin/dnsroundrobin.go
javier-aliaga: https://github.com/dapr/dapr/pull/8721/commits/fcf75fea9f6fb6383fd01f10a64bd11d7c0983c9

pkg/actors/internal/placement/client/client.go
```diff
@@ -17,26 +17,32 @@ import (
...
 	"github.com/spiffe/go-spiffe/v2/spiffeid"
 	"google.golang.org/grpc"
 
+	"github.com/dapr/kit/concurrency"
+	"github.com/dapr/kit/concurrency/lock"
+	"github.com/dapr/kit/logger"
+
```
JoshVanL: nit: code blocks

pkg/actors/internal/placement/client/connector/dnslookup/dnslookup.go
```diff
@@ -0,0 +1,107 @@
...
+import (
+	"context"
+	"fmt"
+	"net"
+
+	"github.com/dapr/kit/logger"
+
+	"github.com/dapr/dapr/pkg/actors/internal/placement/client/connector"
+
+	"google.golang.org/grpc"
```
JoshVanL: nit

pkg/actors/internal/placement/client/connector/dnslookup/dnslookup_test.go
```diff
@@ -0,0 +1,57 @@
...
+*/
+package dnslookup
+
+import (
+	"context"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/stretchr/testify/assert"
```
JoshVanL: nit

pkg/actors/internal/placement/placement.go
```diff
@@ -22,6 +22,11 @@ import (
 	"sync/atomic"
 	"time"
 
+	"github.com/dapr/kit/concurrency"
+	"github.com/dapr/kit/concurrency/fifo"
```
JoshVanL: nit

javier-aliaga:
> Just the nix, otherwise lgtm!

Done, I hope... otherwise I will destroy goland :-)

JoshVanL (approved):
Would help if I clicked the correct button :)

marcduiker:
@holopin-bot @javier-aliaga Thank you! Here's a digital badge as a small token of appreciation.

PR #8726, opened 2025-05-21
Scheduler server control loop

JoshVanL:
Updates the Scheduler server connection pool handler to be made of
control loops;

`connections{} ->> stream[]`

Although use of control loops here instead increases complexity, it
is important for the latter addition of broadcast job management and
sending to clients as to prevent deadlocks.


JoshVanL:
Part of https://github.com/dapr/dapr/issues/501

pkg/scheduler/server/internal/pool/loops/connections/connections.go
```diff
@@ -0,0 +1,198 @@
...
+		IDx:      streamIDx,
+		Channel:  add.Channel,
+		Request:  add.Request,
+		ConnLoop: c.loop,
+	})
+
+	c.wg.Add(1)
+	go func() {
+		defer c.wg.Done()
+		err := streamLoop.Run(ctx)
```
cicoyle, line 131: Shall we use the runner manager here?

cicoyle (approved):
One suggestion to use the runner manager, but otherwise lgtm

pkg/scheduler/server/internal/pool/loops/connections/connections.go
JoshVanL: Unfortunately, we can't here because we are both dynamically creating stream loops over time, and this function _must_ not block- because the connection loop itself is single threaded.

cicoyle: ah ok fine by me 👍🏻

cicoyle:
Can you verify those IT test failures are not from this PR? There are a few new ones I haven't seen before

cicoyle (approved):

PR #8731, opened 2025-05-24
Workflow: Prevent panic when no actor state store is configured

JoshVanL:
Before, attempting to schedule a workflow on a daprd where no actor state store would result in a panic. This is because the workflow actor would attempt to use the actor state store, which is nil if no actor state store is configured.

Updates the workflow actor to return a typed error if no actor state store is configured.

Adds int tests.

cc @javier-aliaga

### DIFF
--- a/pkg/actors/targets/workflow/invoke.go
+++ b/pkg/actors/targets/workflow/invoke.go
@@ -25,6 +25,7 @@ import (
 
 	actorapi "github.com/dapr/dapr/pkg/actors/api"
 	actorerrors "github.com/dapr/dapr/pkg/actors/errors"
+	"github.com/dapr/dapr/pkg/messages"
 	invokev1 "github.com/dapr/dapr/pkg/messaging/v1"
 	commonv1pb "github.com/dapr/dapr/pkg/proto/common/v1"
 	internalsv1pb "github.com/dapr/dapr/pkg/proto/internals/v1"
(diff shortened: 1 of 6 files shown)


tests/e2e/service_invocation/service_invocation_test.go
```diff
@@ -1558,7 +1558,6 @@ func TestNegativeCases(t *testing.T) {
 				// - `rpc error: code = Internal desc = failed to invoke, id: serviceinvocation-callee-0, err: rpc error: code = Internal desc = error invoking app channel: Post \"http://127.0.0.1:3000/timeouterror\": context deadline exceeded``
 				// - `rpc error: code = DeadlineExceeded desc = context deadline exceeded`
 				assert.Contains(t, testResults.RawError, "rpc error:")
-				assert.Contains(t, testResults.RawError, "RST_STREAM with error code: CANCEL")
```
nelson-parente, line 1561: why this change and the one in mtls.go?

pkg/actors/targets/workflow/workflow.go
```diff
@@ -208,6 +209,10 @@ func (w *workflow) InvokeMethod(ctx context.Context, req *internalsv1pb.Internal
 func (w *workflow) executeMethod(ctx context.Context, methodName string, request []byte) ([]byte, error) {
 	log.Debugf("Workflow actor '%s': invoking method '%s'", w.actorID, methodName)
 
+	if w.actorState == nil {
+		return nil, messages.ErrActorRuntimeNotFound
```
nelson-parente, line 213: nit: should this err be more verbose to the user clearly stating that this needs to be enabled and point to docs?

nelson-parente (approved):
lgtm

pkg/actors/targets/workflow/workflow.go
JoshVanL: I don't know how we can be any more verbose- this error is piped and returned to the client API call as a failure.

tests/e2e/service_invocation/service_invocation_test.go
JoshVanL: I believe this string check is no longer correct because of a lib dep upgrade, I have updated the string check above to check for status code check via string match.

nelson-parente (approved):

cicoyle (approved):

cicoyle (approved):

PR #8734, opened 2025-05-27
feat: Add framework for Redis-to-Kafka state replicator

alvidofaisal:
This PR introduces a framework for state replication within Dapr, along with an initial implementation for a Redis-to-Kafka state replicator.

The main goals of this PR are:
- To establish the foundational components and registration mechanism for state replicators.
- To provide a concrete example of a replicator (Redis to Kafka).
- To include initial performance testing infrastructure for this new capability.

### DIFF
--- a/cmd/daprd/components/replicator_redis_kafka.go
+++ b/cmd/daprd/components/replicator_redis_kafka.go
@@ -1,0 +1,53 @@
+//go:build allcomponents
+
+package components
+
+import (
+	// This is a hypothetical import. It assumes that a component named
+	// 'redis_kafka_replicator' exists in 'components-contrib' and
+	// provides a 'NewRedisKafkaReplicator' constructor.
+	// Since we can't add that here, we'll use a placeholder.
+	// "github.com/dapr/components-contrib/state/redis_kafka_replicator"
+
+	replicatorLoader "github.com/dapr/dapr/pkg/components/replicators"
+	"github.com/dapr/kit/logger"
+	// "fmt" // For placeholder - not actually used in the placeholder code
+)
+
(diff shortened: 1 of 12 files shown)


daixiang0:
Could you share some context for this PR? Like why need to implement it, which issue can it fix?

alvidofaisal:
Hi, thanks for the question. 🙂

Right now Dapr gives us a solid Redis state store and a Kafka pub-sub component, yet there’s no built-in way to turn a write that lands in Redis into a Kafka event. That forces developers to double-write from service code (cache plus topic) or attach an external CDC pipeline, which undermines Dapr’s “let the sidecar do the heavy lifting” promise.

**What the replicator adds**

1. **Better developer experience.** Your service still calls `POST /state` once; the sidecar quietly pushes a CloudEvent to Kafka. No duplicate error handling or boilerplate.
2. **Common real-world patterns.** CQRS, audit trails, multi-region cache fills, and real-time analytics all need those state changes downstream. Today every team hand-rolls its own bridge.
3. **Operational safety.** We listen to Redis key notifications (or Streams) and rely on Kafka for ordering and retention, so retries are safe. If Kafka is down, writes continue in Redis until the producer recovers.

The first cut keeps scope small: subscribe to key events, map them to CloudEvents, produce to Kafka with at-least-once delivery. Once that is stable we can discuss exactly-once semantics, schema registry support, or a simple `replicateTo` annotation so any state store can opt in.

Hope this clarifies the motivation. Happy to adjust anything or answer more questions!

yaron2:
@alvidofaisal have you looked at https://docs.dapr.io/developing-applications/building-blocks/state-management/howto-outbox/

PR #8740, opened 2025-05-28
Integration: fix scheduler get job race

JoshVanL:
Fix get job race by ensuring the job is created before every test.

### DIFF
--- a/tests/integration/suite/daprd/jobs/get.go
+++ b/tests/integration/suite/daprd/jobs/get.go
@@ -1,0 +1,79 @@
+/*
+Copyright 2024 The Dapr Authors
+Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


tests/integration/suite/daprd/jobs/get.go
```diff
@@ -0,0 +1,79 @@
+/*
+Copyright 2024 The Dapr Authors
```
nelson-parente: nit: 2025

nelson-parente (approved):

tests/integration/suite/daprd/jobs/get.go
cicoyle:
```suggestion
Copyright 2025 The Dapr Authors
```

cicoyle (approved):

PR #8741, opened 2025-05-28
Jobs: Move `overwrite` field outside of job proto

JoshVanL:
Moves the `overwrite` job field to be outside of the job object itself, since its a API level field signal, rather than part of the stored job object.


tests/integration/suite/actors/reminders/overwrite.go
```diff
@@ -0,0 +1,65 @@
+/*
+Copyright 202d The Dapr Authors
```
nelson-parente: nit 2025

nelson-parente (dismissed):

tests/integration/suite/actors/reminders/overwrite.go
cicoyle:
```suggestion
Copyright 2025 The Dapr Authors
```

cicoyle (approved):

nelson-parente (approved):

nelson-parente (approved):

cicoyle (approved):

nelson-parente (approved):

cicoyle (approved):

PR #8757, opened 2025-06-04
[1.15] Actors: Fix memory leak in lock goroutines

JoshVanL:
Running Actor or Workflow workloads would see the daprd process consume more and more memory over time.

Running Actor or Workflow for long enough periods of time would see the daprd process consume all available memory, leading to an OOM crash.

The Actor locking mechanism would not release object memory after use, in the case where the daprd it calling a remote Actor.

Defer Actor message locking to only the daprd which is hosting that Actor ID to ensure the lock object memory is always released after use.


tests/integration/suite/actors/lock/reminders/self/grpc.go
```diff
@@ -11,7 +11,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package reminders
+package self
```
daixiang0: How about rename `self` to `local` is better, let us keep `self` away from package name.

cicoyle: +1, and can you bring in all the changes I suggested from the other PR for this diff to the release branch?

JoshVanL: Done https://github.com/dapr/dapr/pull/8757/commits/73f697b70d56398eddfe5fe34c46ae1c119879e0

yaron2 (dismissed):

cicoyle (dismissed):

cicoyle (approved):

PR #8771, opened 2025-06-07
[1.15] Workflow: Allow non-stream Workflow Schedule

JoshVanL:
Remove the registered actor type gate on the Scheduler workflow internal server to allow Scheduling workflows on a daprd which does not have a client listener stream.

This makes the Schedule Workflow RPC in line with the other Workflow RPCs.

Problem

Attempting to schedule a Workflow on a daprd with no Workflow listener stream would hang.

Impact

Attempting to schedule a Workflow on a daprd with no Workflow listener stream would hang indefinitely.

Root cause

The daprd would wait until a Workflow listener stream was established before processing the schedule request.

Solution

Allow scheduling of Workflows without requiring a Workflow listener stream to be established first. This allows for immediate


cicoyle (dismissed):
lgtm

cicoyle (dismissed):

cicoyle (approved):

PR #8782, opened 2025-06-10
Update Scheduler connection after dissemination

JoshVanL:
Problem

Connecting a Workflow stream when there is worked queued in the system would cause a large number errors.

Impact

Connecting a Workflow stream when there is work queued in the system would cause a large number of errors, leading to degraded performance and potential timeouts on existing Workflows.

Root cause

Daprd would connect to Scheduler to receive Workflow work before Placement had disseminated.

Solution

Ensure that the daprd connects to the Scheduler only after Placement has disseminated, preventing errors and ensuring that the daprd is ready to process Workflow work without contention.

### DIFF
--- a/tests/integration/framework/process/daprd/daprd.go
+++ b/tests/integration/framework/process/daprd/daprd.go
@@ -206,7 +206,7 @@ func (d *Daprd) WaitUntilTCPReady(t *testing.T, ctx context.Context) {
 		}
 		net.Close()
 		return true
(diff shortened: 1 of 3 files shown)


cicoyle:
@JoshVanL can you fix the conflicts for this one?

nelson-parente (approved):

cicoyle (approved):

PR #8790, opened 2025-06-12
Filinto/streaming conversation api

filintod:
# Description

## Issue reference

## Checklist

- [ ] Code compiles correctly
- [ ] Created/updated tests
- [ ] Unit tests passing


dapr/proto/runtime/v1/dapr.proto
```diff
@@ -1344,4 +1347,51 @@ message ConversationResponse {
...
   // An array of results.
   repeated ConversationResult outputs = 2;
+}
+
+// ConversationStreamResponse is the streaming response for Conversation.
+message ConversationStreamResponse {
+  oneof response_type {
+    ConversationStreamChunk chunk = 1;
+    ConversationStreamComplete complete = 2;
+    ConversationStreamError error = 3;
```
JoshVanL: Why is this part of the message rather than an actual error?

dapr/proto/runtime/v1/dapr.proto
```diff
@@ -1344,4 +1347,51 @@ message ConversationResponse {
 
   // An array of results.
   repeated ConversationResult outputs = 2;
+}
+
+// ConversationStreamResponse is the streaming response for Conversation.
+message ConversationStreamResponse {
+  oneof response_type {
+    ConversationStreamChunk chunk = 1;
+    ConversationStreamComplete complete = 2;
```
JoshVanL, line 1417: Why is closing the stream by the server not enough to indicate that that stream has finished- why is this a message?

dapr/proto/runtime/v1/dapr.proto
```diff
@@ -1344,4 +1347,51 @@ message ConversationResponse {
...
+    ConversationStreamError error = 3;
+  }
+}
+
+// ConversationStreamChunk represents a streaming content chunk.
+message ConversationStreamChunk {
+  // Streaming content chunk
+  string content = 1;
+  // Whether this chunk ends a token/word boundary
+  optional bool token_boundary = 2 [json_name = "tokenBoundary"];
```
JoshVanL: Why is this important?

dapr/proto/runtime/v1/dapr.proto
```diff
@@ -1344,4 +1347,51 @@ message ConversationResponse {
...
+  // Delta parameters for this chunk
+  map<string, google.protobuf.Any> delta_parameters = 3 [json_name = "deltaParameters"];
```
JoshVanL: What is a delta parameter, and why can they not be typed?

pkg/api/universal/conversation_middleware.go
```diff
@@ -0,0 +1,137 @@
...
+
+// IsTokenBoundary determines if a chunk ends at a token boundary (heuristic)
+func (s *StreamingPIIScrubber) IsTokenBoundary(chunk []byte) *bool {
+	if len(chunk) == 0 {
+		result := false
+		return &result
+	}
+
+	lastChar := chunk[len(chunk)-1]
+	result := lastChar == ' ' || lastChar == '.' || lastChar == ',' ||
```
JoshVanL: Strings.Contiains?

dapr/proto/runtime/v1/dapr.proto
filintod: this is very specific to AI vendors, like a type of chunk

dapr/proto/runtime/v1/dapr.proto
filintod: I'll remove this, not seen it much actually, more like nice but not needed

dapr/proto/runtime/v1/dapr.proto
JoshVanL: Please link to some literature or documentation for how this would be used.

pkg/api/universal/conversation_middleware.go
filintod: not sure, it is a single byte at this point, bytes.contains maybe, I'll check for options to simplify

dapr/proto/runtime/v1/dapr.proto
filintod: the complete has usage information that we are not sending with each chunk, we could send usage with each chunk and avoid this and let the user keep the counter, let me think on it.

dapr/proto/runtime/v1/dapr.proto
filintod: Let me change to an error instead and pass the ai provider error downstream with it

pkg/api/universal/conversation_stream.go
```diff
@@ -0,0 +1,296 @@
...
+		err := messages.ErrConversationMissingInputs.WithFormat(req.GetName())
+		a.logger.Debug(err)
+		return err
+	}
+
+	// Process inputs
+	for _, i := range req.GetInputs() {
+		msg := i.GetMessage()
+
+		// TODO: Add PII scrubbing for input if needed
```
filintod: we're missing this as in https://github.com/dapr/dapr/blob/36d47e5aeb78c4eeae213a657a6e6d8a579c0cdc/pkg/api/universal/conversation.go#L66

dapr/proto/runtime/v1/dapr.proto
filintod: we need to keep as the interface in langchaingo does not allow for usage per chunk as the sdk direct

go.mod
```diff
@@ -511,7 +509,8 @@ replace (
 // Uncomment for local development for testing with changes in the components-contrib && kit repositories.
 // Don't commit with this uncommented!
 //
-// replace github.com/dapr/components-contrib => ../components-contrib
+replace github.com/dapr/components-contrib => ../components-contrib
```
filintod, line 508: Temporary while developing and while component-contrib dependency is needed. I'll need to update the mod sum after that is merged

dapr/proto/runtime/v1/dapr.proto
```diff
@@ -220,6 +220,9 @@ service Dapr {
 
   // Converse with a LLM service
   rpc ConverseAlpha1(ConversationRequest) returns (ConversationResponse) {}
+
+  // Converse with a LLM service using streaming
+  rpc ConverseStreamAlpha1(ConversationRequest) returns (stream ConversationStreamResponse) {}
```
JoshVanL, line 225: Why is this API not bi-directional- with a client being able to send more prompts during the same "conversation"?

dapr/proto/runtime/v1/dapr.proto
```diff
@@ -1357,4 +1360,41 @@ message ConversationResponse {
...
+
+// ConversationUsage represents token usage statistics.
+message ConversationUsage {
+  // Number of tokens in the prompt
+  optional int32 prompt_tokens = 1 [json_name = "promptTokens"];
+  // Number of tokens in the completion
+  optional int32 completion_tokens = 2 [json_name = "completionTokens"];
+  // Total number of tokens used
+  optional int32 total_tokens = 3 [json_name = "totalTokens"];
+}
```
JoshVanL: nit: eof newline

dapr/proto/runtime/v1/dapr.proto
```diff
@@ -1357,4 +1360,41 @@ message ConversationResponse {
...
+
+
+// ConversationUsage represents token usage statistics.
+message ConversationUsage {
+  // Number of tokens in the prompt
+  optional int32 prompt_tokens = 1 [json_name = "promptTokens"];
+  // Number of tokens in the completion
+  optional int32 completion_tokens = 2 [json_name = "completionTokens"];
+  // Total number of tokens used
+  optional int32 total_tokens = 3 [json_name = "totalTokens"];
```
JoshVanL: Any reason signed & 32 bit?

```suggestion
  optional uint64 prompt_tokens = 1 [json_name = "promptTokens"];
  // Number of tokens in the completion
  optional uint64 completion_tokens = 2 [json_name = "completionTokens"];
  // Total number of tokens used
  optional uint64 total_tokens = 3 [json_name = "totalTokens"];
```

dapr/proto/runtime/v1/dapr.proto
```diff
@@ -1357,4 +1360,41 @@ message ConversationResponse {
...
   // An array of results.
   repeated ConversationResult outputs = 2;
+
+  // Usage statistics if available
+  optional ConversationUsage usage = 3;
+}
+
+// ConversationStreamResponse is the streaming response for Conversation.
+message ConversationStreamResponse {
+  oneof response_type {
```
JoshVanL, line 1415: Will this message _ever_ contain other fields outside of `oneof`?

Generally it's good practice to move the `oneof` definition to a separate message to preserve field number sequence.

pkg/api/universal/conversation.go
```diff
@@ -24,9 +24,58 @@ import (
...
+	"github.com/dapr/kit/logger"
 	kmeta "github.com/dapr/kit/metadata"
 )
 
+// needsInputScrubber checks if PII scrubbing is needed for any input
+func needsInputScrubber(req *runtimev1pb.ConversationRequest) bool {
+	for _, input := range req.GetInputs() {
+		if input.GetScrubPII() {
+			return true
+		}
```
JoshVanL: Why is the scrubber enabled for _all_ when only some inputs have it enabled?

pkg/api/universal/conversation.go
```diff
@@ -24,9 +24,58 @@ import (
 	"github.com/dapr/dapr/pkg/messages"
 	runtimev1pb "github.com/dapr/dapr/pkg/proto/runtime/v1"
 	"github.com/dapr/dapr/pkg/resiliency"
+	"github.com/dapr/kit/logger"
 	kmeta "github.com/dapr/kit/metadata"
 )
 
+// needsInputScrubber checks if PII scrubbing is needed for any input
+func needsInputScrubber(req *runtimev1pb.ConversationRequest) bool {
```
JoshVanL: This helper func is strange to have in the same package namespace as other universal API handler. Please move to a dedicated `conversation` package which as the context to this func.

pkg/api/universal/conversation.go
```diff
@@ -24,9 +24,58 @@ import (
...
+			return true
+		}
+	}
+	return false
+}
+
+// needsOutputScrubber checks if PII scrubbing is needed for outputs
+func needsOutputScrubber(req *runtimev1pb.ConversationRequest) bool {
+	return req.GetScrubPII()
+}
```
JoshVanL: Seem redundant.

pkg/api/universal/conversation.go
```diff
@@ -24,9 +24,58 @@ import (
...
+
+// getInputsFromRequest gets the inputs from the request and scrubs them if PII scrubbing is enabled on the input
+func getInputsFromRequest(req *runtimev1pb.ConversationRequest, scrubber piiscrubber.Scrubber, logger logger.Logger) ([]conversation.ConversationInput, error) {
```
JoshVanL: This function name is not appropriate in a package for the entire universal API set.

Why are we passing a logger?

pkg/api/universal/conversation.go
```diff
@@ -24,9 +24,58 @@ import (
...
+	if reqInputs == nil {
+		return nil, messages.ErrConversationMissingInputs.WithFormat(req.GetName())
+	}
+
+	inputs := make([]conversation.ConversationInput, 0, len(reqInputs))
+	for _, i := range reqInputs {
+		msg := i.GetContent()
+
+		if i.GetScrubPII() && scrubber != nil {
+			scrubbed, sErr := scrubber.ScrubTexts([]string{i.GetContent()})
```
JoshVanL: Why is content only a single element?

JoshVanL: Please keep the `error` variable name to `err` to reduce cognitive load to the reader.

Use `var scrubbed ...` in the proceeding line in the cases of shadow variable complaints.

```suggestion
			scrubbed, err := scrubber.ScrubTexts([]string{i.GetContent()})
```

pkg/api/universal/conversation.go
```diff
@@ -24,9 +24,58 @@ import (
...
+
+			msg = scrubbed[0]
+		}
+
+		c := conversation.ConversationInput{
+			Message: msg,
+			Role:    conversation.Role(i.GetRole()),
+		}
+
+		inputs = append(inputs, c)
```
JoshVanL:
```suggestion
		inputs = append(inputs, 	c := conversation.ConversationInput{
			Message: msg,
			Role:    conversation.Role(i.GetRole()),
		})
```

pkg/api/universal/conversation.go
```diff
@@ -24,9 +24,58 @@ import (
...
+				logger.Debug(sErr)
+				return nil, sErr
+			}
+
+			msg = scrubbed[0]
+		}
+
+		c := conversation.ConversationInput{
+			Message: msg,
+			Role:    conversation.Role(i.GetRole()),
```
JoshVanL: What is `Role`?

pkg/api/universal/conversation.go
```diff
@@ -55,35 +104,29 @@ func (a *Universal) ConverseAlpha1(ctx context.Context, req *runtimev1pb.Convers
...
-		c := conversation.ConversationInput{
-			Message: msg,
-			Role:    conversation.Role(i.GetRole()),
+	if needsInputPIIScrubbing || needsOutputPIIScrubbing {
+		scrubber, err = piiscrubber.NewDefaultScrubber()
+		if err != nil {
+			err := messages.ErrConversationMissingInputs.WithFormat(req.GetName())
+			a.logger.Debug(err)
+			return &runtimev1pb.ConversationResponse{}, err
 		}
```
JoshVanL: We have lost the error contents here. I assume the original error was meant to be passed to the `Debug` ?

Please use Error log level, not debug.

pkg/api/universal/conversation.go
```diff
@@ -24,9 +24,58 @@ import (
...
+
+	inputs := make([]conversation.ConversationInput, 0, len(reqInputs))
+	for _, i := range reqInputs {
+		msg := i.GetContent()
+
+		if i.GetScrubPII() && scrubber != nil {
+			scrubbed, sErr := scrubber.ScrubTexts([]string{i.GetContent()})
+			if sErr != nil {
+				sErr = messages.ErrConversationInvoke.WithFormat(req.GetName(), sErr.Error())
+				logger.Debug(sErr)
```
JoshVanL: This is an error log.

pkg/api/universal/conversation.go
```diff
@@ -55,35 +104,29 @@ func (a *Universal) ConverseAlpha1(ctx context.Context, req *runtimev1pb.Convers
...
 
-		c := conversation.ConversationInput{
-			Message: msg,
-			Role:    conversation.Role(i.GetRole()),
+	if needsInputPIIScrubbing || needsOutputPIIScrubbing {
+		scrubber, err = piiscrubber.NewDefaultScrubber()
+		if err != nil {
+			err := messages.ErrConversationMissingInputs.WithFormat(req.GetName())
+			a.logger.Debug(err)
+			return &runtimev1pb.ConversationResponse{}, err
```
JoshVanL:
```suggestion
			return nil, err
```

pkg/api/universal/conversation.go
```diff
@@ -55,35 +104,29 @@ func (a *Universal) ConverseAlpha1(ctx context.Context, req *runtimev1pb.Convers
...
 		}
+	}
 
-		request.Inputs = append(request.Inputs, c)
+	inputs, err := getInputsFromRequest(req, scrubber, a.logger)
+	if err != nil {
+		return &runtimev1pb.ConversationResponse{}, err
```
JoshVanL:
```suggestion
		return nil}, err
```

pkg/api/universal/conversation_middleware.go
```diff
@@ -0,0 +1,115 @@
+/*
```
JoshVanL: This file doesn't belong in `universal`.

JoshVanL: Please refrain from creating file names with underscore deliminator between words - needing to do so signals that we should organize into different packages.

pkg/api/universal/conversation_middleware.go
```diff
@@ -0,0 +1,115 @@
...
+
+// StreamingPIIScrubber handles real-time PII scrubbing for streaming content
+type StreamingPIIScrubber struct {
+	buffer     []byte
+	windowSize int // Look-ahead window for PII detection
+	scrubber   piiscrubber.Scrubber
+}
+
+// PIIPattern represents a PII detection pattern
+type PIIPattern struct {
```
JoshVanL: Used?

pkg/api/universal/conversation_middleware.go
```diff
@@ -0,0 +1,115 @@
...
+	Pattern     *regexp.Regexp
+	Replacement string
+}
+
+// NewStreamingPIIScrubber creates a new streaming PII scrubber with a look-ahead window
+func NewStreamingPIIScrubber(windowSize int) (*StreamingPIIScrubber, error) {
+	scrubber := &StreamingPIIScrubber{
+		buffer:     make([]byte, 0, windowSize*2),
```
JoshVanL: Why by 2? Why do we need a buffer, and why are we not using `bytes.Buffer`?

pkg/api/universal/conversation_middleware.go
```diff
@@ -0,0 +1,115 @@
...
+		return chunk
+	}
+
+	// Append to buffer
+	s.buffer = append(s.buffer, chunk...)
+
+	// Process what we can safely process (leaving window for PII detection)
+	safeLength := len(s.buffer) - s.windowSize
+	if safeLength <= 0 {
+		return nil // Need more data
```
JoshVanL:
```suggestion
		return nil
```

pkg/api/universal/conversation_middleware.go
```diff
@@ -0,0 +1,115 @@
...
+}
+
+// PIIPattern represents a PII detection pattern
+type PIIPattern struct {
+	Pattern     *regexp.Regexp
+	Replacement string
+}
+
+// NewStreamingPIIScrubber creates a new streaming PII scrubber with a look-ahead window
+func NewStreamingPIIScrubber(windowSize int) (*StreamingPIIScrubber, error) {
```
JoshVanL: Why do we need a window size?

pkg/api/universal/conversation_middleware.go
```diff
@@ -0,0 +1,115 @@
...
+	if len(chunk) == 0 {
+		return chunk
+	}
+
+	// Append to buffer
+	s.buffer = append(s.buffer, chunk...)
+
+	// Process what we can safely process (leaving window for PII detection)
+	safeLength := len(s.buffer) - s.windowSize
+	if safeLength <= 0 {
```
JoshVanL: Why is it not this? The comments suggest have a buffer size the _same_ size as the window size is "safe".
```suggestion
	if safeLength < 0 {
```

pkg/api/universal/conversation_middleware.go
```diff
@@ -0,0 +1,115 @@
...
+		return nil // Need more data
+	}
+
+	// Extract safe portion
+	safeContent := s.buffer[:safeLength]
```
JoshVanL: What do we mean by "safe"?

pkg/api/universal/conversation_middleware.go
```diff
@@ -0,0 +1,115 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to create default PII scrubber: %w", err)
+	}
+	scrubber.scrubber = defaultScrubber
+
+	return scrubber, nil
+}
+
+// ProcessChunk processes a chunk of streaming content with PII scrubbing
+func (s *StreamingPIIScrubber) ProcessChunk(chunk []byte) []byte {
```
JoshVanL: This scrubber looks like it should be processed via an `io.Reader`.

pkg/api/universal/conversation_middleware.go
```diff
@@ -0,0 +1,115 @@
...
+
+// Flush processes and returns any remaining content in the buffer
+func (s *StreamingPIIScrubber) Flush() []byte {
+	if len(s.buffer) == 0 {
+		return nil
+	}
+
+	// Process remaining buffer content
+	scrubbed := s.scrubPII(s.buffer)
+	s.buffer = s.buffer[:0] // Clear buffer
```
JoshVanL: Why is this necessary?

pkg/api/universal/conversation_middleware.go
```diff
@@ -0,0 +1,115 @@
...
+	// Convert to string for processing
+	text := string(content)
+
+	// Use the default scrubber to process the text
+	scrubbed, err := s.scrubber.ScrubTexts([]string{text})
+	if err != nil {
+		// If scrubbing fails, return original content
+		// TODO: Add logging here
+		return content
+	}
```
JoshVanL: Why are we swallowing the error and failing silently?

pkg/api/universal/conversation_middleware.go
```diff
@@ -0,0 +1,115 @@
...
+
+	if len(scrubbed) > 0 {
+		return []byte(scrubbed[0])
+	}
```
JoshVanL: What is this doing?

pkg/api/universal/conversation_stream.go
```diff
@@ -0,0 +1,312 @@
...
+	"time"
+
+	piiscrubber "github.com/aavaz-ai/pii-scrubber"
+
+	"github.com/dapr/components-contrib/conversation"
+	diag "github.com/dapr/dapr/pkg/diagnostics"
+	"github.com/dapr/dapr/pkg/messages"
+	runtimev1pb "github.com/dapr/dapr/pkg/proto/runtime/v1"
+	"github.com/dapr/kit/logger"
+	kmeta "github.com/dapr/kit/metadata"
```
JoshVanL:
```suggestion
	"github.com/dapr/kit/metadata"
```

pkg/api/universal/conversation_stream.go
```diff
@@ -0,0 +1,312 @@
...
+)
+
+// defaultPIIScrubberWindowSize is the default window size (bytes) for PII scrubbing
+// this allows for a look-ahead window of stream data to be scrubbed
+// TODO: make this configurable
+const defaultPIIScrubberWindowSize = 50
```
JoshVanL: Why was `50` chosen? Please use a typical data chunk size like `1024`.

pkg/api/universal/conversation_stream.go
```diff
@@ -0,0 +1,312 @@
...
+
+// ConverseStreamAlpha1 implements the streaming conversation API
+func (a *Universal) ConverseStreamAlpha1(req *runtimev1pb.ConversationRequest, stream runtimev1pb.Dapr_ConverseStreamAlpha1Server) (retErr error) {
```
JoshVanL: Please do not use return values for least surprising. It makes it more difficult to read the function contents.

Name should at least be `error`.

```suggestion
func (a *Universal) ConverseStreamAlpha1(req *runtimev1pb.ConversationRequest, stream runtimev1pb.Dapr_ConverseStreamAlpha1Server) error {
```

JoshVanL: It is not being used till the end :upside_down_face:

pkg/api/universal/conversation_stream.go
```diff
@@ -0,0 +1,312 @@
...
+	// Validate component exists
+	if a.compStore.ConversationsLen() == 0 {
+		err := messages.ErrConversationNotFound
+		a.logger.Debug(err)
+		return err
+	}
```
JoshVanL:
```suggestion
```

pkg/api/universal/conversation_stream.go
```diff
@@ -0,0 +1,312 @@
...
+	component, ok := a.compStore.GetConversation(req.GetName())
+	if !ok {
+		err := messages.ErrConversationNotFound.WithFormat(req.GetName())
+		a.logger.Debug(err)
+		return err
+	}
+
+	// Prepare component conversation request (similar to non-streaming version)
+	request := &conversation.ConversationRequest{}
+	err := kmeta.DecodeMetadata(req.GetMetadata(), request)
```
JoshVanL: What is in metadata?

Why do we not have named typed field instead?

pkg/api/universal/conversation_stream.go
```diff
@@ -0,0 +1,312 @@
...
+	}
+
+	if len(req.GetInputs()) == 0 {
+		err := messages.ErrConversationMissingInputs.WithFormat(req.GetName())
+		a.logger.Debug(err)
+		return err
+	}
+
+	// Create streaming pipeline
+	pipelineImpl := NewStreamingPipelineImpl(a.logger)
```
JoshVanL: Why are we passing a logger?

pkg/api/universal/conversation_stream.go
```diff
@@ -0,0 +1,312 @@
...
+	needsInputPIIScrubbing := needsInputScrubber(req)
+	needsOutputPIIScrubbing := needsOutputScrubber(req)
+
+	if needsInputPIIScrubbing || needsOutputPIIScrubbing {
+		scrubber, err = piiscrubber.NewDefaultScrubber()
+		if err != nil {
+			err := messages.ErrConversationMissingInputs.WithFormat(req.GetName())
+			a.logger.Debug(err)
+			return err
+		}
```
JoshVanL: Swallowing error again.

pkg/api/universal/conversation_stream.go
```diff
@@ -0,0 +1,312 @@
...
+
+		// Add PII scrubbing middleware for streaming output scrubbing
+		if needsOutputPIIScrubbing {
+			scrubberMiddleware, scrubberErr := NewStreamingPIIScrubber(defaultPIIScrubberWindowSize)
+			if scrubberErr != nil {
+				return fmt.Errorf("failed to create streaming PII scrubber middleware: %w", scrubberErr)
+			}
```
JoshVanL: Use same err variabnle name.

pkg/api/universal/conversation_stream.go
```diff
@@ -0,0 +1,312 @@
...
+			pipelineImpl.AddMiddleware(scrubberMiddleware)
```
JoshVanL: Why is this being added as a middleware in this way if it is always going to be added?

pkg/api/universal/conversation_stream.go
```diff
@@ -0,0 +1,312 @@
...
+		}
+	}
+
+	// get inputs from request and scrub them if PII scrubbing is enabled
+	request.Inputs, err = getInputsFromRequest(req, scrubber, a.logger)
```
JoshVanL: Why pass logger?

pkg/api/universal/conversation_stream.go
```diff
@@ -0,0 +1,312 @@
...
+
+	// Process streaming request
+	ctx := stream.Context()
+	start := time.Now()
+	var usage *runtimev1pb.ConversationUsage
+	defer func() {
+		elapsed := diag.ElapsedSince(start)
+		diag.DefaultComponentMonitoring.ConversationInvoked(ctx, req.GetName(), retErr == nil, elapsed, diag.StreamingConversation, int64(usage.GetPromptTokens()), int64(usage.GetCompletionTokens()))
+	}()
+	usage, retErr = pipelineImpl.ProcessStream(ctx, stream, component, request)
```
JoshVanL: Context shouldn't be passed as a parameter. It in in the stream.

```suggestion
	usage, retErr = pipelineImpl.ProcessStream(stream, component, request)
```

pkg/api/universal/conversation_stream.go
```diff
@@ -0,0 +1,312 @@
...
+	request.Temperature = req.GetTemperature()
+
+	// Process streaming request
+	ctx := stream.Context()
+	start := time.Now()
+	var usage *runtimev1pb.ConversationUsage
+	defer func() {
+		elapsed := diag.ElapsedSince(start)
+		diag.DefaultComponentMonitoring.ConversationInvoked(ctx, req.GetName(), retErr == nil, elapsed, diag.StreamingConversation, int64(usage.GetPromptTokens()), int64(usage.GetCompletionTokens()))
+	}()
```
JoshVanL: This all looks like it should be moved to `ProcessStream` - usage returned `ProcessStream` by is only being used by metrics in the defer.

pkg/api/universal/conversation_stream.go
```diff
@@ -0,0 +1,312 @@
...
+}
+
+// StreamingMiddleware is a simple interface for processing streaming chunks
+type StreamingMiddleware interface {
+	ProcessChunk(chunk []byte) []byte
+	Flush() []byte
+}
+
+// NewStreamingPipelineImpl creates a new streaming pipeline
+func NewStreamingPipelineImpl(logger logger.Logger) *StreamingPipelineImpl {
```
JoshVanL: This don't look like they should be exported.

pkg/api/universal/conversation_stream.go
```diff
@@ -0,0 +1,312 @@
...
+		elapsed := diag.ElapsedSince(start)
+		diag.DefaultComponentMonitoring.ConversationInvoked(ctx, req.GetName(), retErr == nil, elapsed, diag.StreamingConversation, int64(usage.GetPromptTokens()), int64(usage.GetCompletionTokens()))
+	}()
+	usage, retErr = pipelineImpl.ProcessStream(ctx, stream, component, request)
+	return retErr
+}
+
+// StreamingPipelineImpl handles the streaming conversation pipeline
+type StreamingPipelineImpl struct {
+	middleware []StreamingMiddleware
```
JoshVanL: Why do we have a slice of middlwares?

pkg/api/universal/conversation_stream.go
```diff
@@ -0,0 +1,312 @@
...
+	logger     logger.Logger
+}
+
+// StreamingMiddleware is a simple interface for processing streaming chunks
+type StreamingMiddleware interface {
+	ProcessChunk(chunk []byte) []byte
+	Flush() []byte
+}
```
JoshVanL: `io.Reader`?

pkg/api/universal/conversation_stream.go
```diff
@@ -0,0 +1,312 @@
...
+func (p *StreamingPipelineImpl) AddMiddleware(middleware StreamingMiddleware) {
+	if middleware == nil {
+		return
+	}
+
+	p.middleware = append(p.middleware, middleware)
+}
+
+// ProcessStream handles the streaming pipeline from LangChain Go to gRPC stream
+func (p *StreamingPipelineImpl) ProcessStream(
```
JoshVanL: Exported?

pkg/api/universal/conversation_stream.go
```diff
@@ -0,0 +1,312 @@
...
+		// Send processed chunk immediately (if any ready)
+		if len(processedChunk) > 0 {
+			if err := p.sendChunk(stream, processedChunk); err != nil {
+				p.logger.Errorf("Failed to send streaming chunk: %v", err)
+				return err
+			}
+		}
+
+		return nil
+	}
```
JoshVanL: Why is this function a variable?

pkg/api/universal/conversation_stream.go
```diff
@@ -0,0 +1,312 @@
...
+	ctx context.Context,
+	stream runtimev1pb.Dapr_ConverseStreamAlpha1Server,
+	streamer conversation.StreamingConversation,
+	request *conversation.ConversationRequest,
+) (*runtimev1pb.ConversationUsage, error) {
+	var finalResp *conversation.ConversationResponse
+	var streamErr error
+
+	// Create the streaming function that LangChain Go will call for each chunk
+	streamFunc := func(chunkCtx context.Context, chunk []byte) error {
```
JoshVanL:
```suggestion
	streamFunc := func(ctx context.Context, chunk []byte) error {
```

pkg/api/universal/conversation_stream.go
```diff
@@ -0,0 +1,312 @@
...
+				p.logger.Errorf("Failed to send streaming chunk: %v", err)
+				return err
+			}
+		}
+
+		return nil
+	}
+
+	// Call the component's streaming method
+	finalResp, streamErr = streamer.ConverseStream(ctx, request, streamFunc)
```
JoshVanL: Is it necessary to pass a function as a parameter here?

JoshVanL:
```suggestion
	resp, err = streamer.ConverseStream(ctx, request, streamFunc)
```

pkg/api/universal/conversation_stream.go
```diff
@@ -0,0 +1,312 @@
...
+	if remaining := p.flushMiddleware(); len(remaining) > 0 {
+		if sendErr := p.sendChunk(stream, remaining); sendErr != nil {
+			p.logger.Errorf("Failed to send final streaming chunk: %v", sendErr)
+			// Return the component error if it exists, otherwise the send error
+			if streamErr != nil {
+				return nil, streamErr
+			}
+			return nil, sendErr
+		}
+	}
```
JoshVanL: It seems somewhat clumsy to need to be flushing here, _after_ we have returned from `ConverseStream`.

pkg/api/universal/conversation_stream.go
```diff
@@ -0,0 +1,312 @@
...
+
+	// Flush any remaining content in middleware buffers
+	if remaining := p.flushMiddleware(); len(remaining) > 0 {
+		if sendErr := p.sendChunk(stream, remaining); sendErr != nil {
+			p.logger.Errorf("Failed to send final streaming chunk: %v", sendErr)
+			// Return the component error if it exists, otherwise the send error
+			if streamErr != nil {
+				return nil, streamErr
+			}
+			return nil, sendErr
```
JoshVanL:
```suggestion
		if err = p.sendChunk(stream, remaining); err != nil {
			p.logger.Errorf("Failed to send final streaming chunk: %v", err)
			// Return the component error if it exists, otherwise the send error
			if err != nil {
				return nil, err
			}
			return nil, err
```

pkg/api/universal/conversation_stream.go
```diff
@@ -0,0 +1,312 @@
...
+		}
+	}
+
+	if streamErr != nil {
+		return nil, streamErr
+	}
```
JoshVanL: We should be checking the error straight after `ConverseStream`. This error is going to be ignored because the block above is going to error.

pkg/api/universal/conversation_stream.go
```diff
@@ -0,0 +1,312 @@
...
+
+	// Send completion with context and usage stats
+	contextID := ""
+	var usage *runtimev1pb.ConversationUsage
+	if finalResp != nil {
+		contextID = finalResp.ConversationContext
+		usage = convertUsageToProto(finalResp)
+	}
+	return usage, p.sendComplete(stream, contextID, usage)
```
JoshVanL:
```suggestion
	return usage, p.sendComplete(stream, resp.GetConversationContext(), usage)
```

pkg/api/universal/conversation_stream.go
```diff
@@ -0,0 +1,312 @@
...
+	if streamErr != nil {
+		return nil, streamErr
+	}
+
+	// Send completion with context and usage stats
+	contextID := ""
+	var usage *runtimev1pb.ConversationUsage
+	if finalResp != nil {
+		contextID = finalResp.ConversationContext
+		usage = convertUsageToProto(finalResp)
```
JoshVanL: `convertUsageToProto` can just return `nil` if `resp` is `nil`.

Would `resp` ever be `nil`?

pkg/api/universal/conversation_stream.go
```diff
@@ -0,0 +1,312 @@
...
+	// Call the regular Converse method
+	resp, err := component.Converse(ctx, request)
+	if err != nil {
+		return nil, err
+	}
+
+	var contextID string
+
+	// Simulate streaming by sending the complete response as chunks
+	if resp != nil {
```
JoshVanL: `resp` should _not_be `nil` if `err` is `nil`.

pkg/api/universal/conversation_stream.go
```diff
@@ -0,0 +1,312 @@
...
+		contextID = resp.ConversationContext
+		if len(resp.Outputs) > 0 {
```
JoshVanL: Redundant.

```suggestion
```

pkg/api/universal/conversation_stream.go
```diff
@@ -0,0 +1,312 @@
...
+			for _, output := range resp.Outputs {
+				// Break the result into chunks to simulate streaming
+				content := output.Result
+				chunkSize := 50 // Send in 50-character chunks
```
JoshVanL: Why `50`?

gRPC by default has a maximum message size of 4MB (~4,000,000 bytes)- It's incredibly inefficient to be sending small payloads like this.

I don't think we need to be chunking at all. But if we must..  we should be fetching what the max message size configured is, and chunking on a slightly smaller number to account for headers and other fields in the message.

tests/integration/suite/daprd/conversation/grpc/streaming.go
```diff
@@ -0,0 +1,451 @@
...
+  version: v1
+  metadata:
+  - name: key
+    value: testkey`
+
+	// Define AI provider components
+
+	// Add AI provider components if their API keys are available
+	for _, provider := range liveConversationAIProviders {
+		if apiKey := os.Getenv(provider.envVar); apiKey != "" {
```
JoshVanL: Move to an end to end test.

Integration tests are always self contained and offline. Use the `echo` conversation component type, or use a mock server.

tests/integration/suite/daprd/conversation/grpc/basic.go
```diff
@@ -34,6 +34,14 @@ type basic struct {
 	daprd *daprd.Daprd
 }
 
+func getEchoEstimatedTokens(msg ...string) int {
```
JoshVanL, line 38: Please do not put helper functions at the top of the file.

tests/integration/suite/daprd/conversation/grpc/streaming.go
```diff
@@ -0,0 +1,451 @@
...
+	return []framework.Option{
+		framework.WithProcesses(s.daprd),
+	}
+}
+
+func (s *streaming) Run(t *testing.T, ctx context.Context) {
+	s.daprd.WaitUntilRunning(t, ctx)
+
+	client := s.daprd.GRPCClient(t, ctx)
+
```
JoshVanL, line 65: Is the `oneof` order (completion last) being tested?

dapr/proto/runtime/v1/dapr.proto
filintod: wouldn't that be the same on the new message? don't follow you here. or that is if we ever need new fields, from your previous question? I don't know if we'll ever need another field here.

"complete" is only expected once at the end, so I guess the other option is have it optional? it most cases just nil.

dapr/proto/runtime/v1/dapr.proto
filintod: most providers do not support that, or the few that do have complete separate apis for "live/realtime".  The streaming is usually just the response been sent on SSE json lines. 

It would be good to have dapr implement a context accumulator that is required more multi-turn conversation if that is what you mean, even though most agent frameworks would implement themselves as "context engineering" can get complicated.

pkg/api/universal/conversation.go
filintod: true,

pkg/api/universal/conversation.go
filintod: true but I was trying to separate input/output conditions, as GetScrubPII is not very informative of direction of data flow that needs scrubbing.

pkg/api/universal/conversation.go
filintod: user/assistant/system (but usually user or assistant in a conversation)

pkg/api/universal/conversation_middleware.go
filintod: true, not really needed here that we are just finishing it

tests/integration/suite/daprd/conversation/grpc/streaming.go
filintod: don't think I have a specific test for it to be last

pkg/api/universal/conversation.go
filintod: actually, this is just to say if we need a scrubber instance, not that it is true for all inputs, we do check on each input if that input in particular needs to be scrubbed

yaron2:
conflicts

pkg/api/universal/conversation.go
filintod: there is only a single response on this more simple (no tool or parts) implemetnation. we do a more comprehensive on the tool calling pr

tests/integration/suite/daprd/conversation/grpc/streaming.go
filintod: e2e tests are slow and a pain to run tbh, and these only run if you have the API keys setup and there is no infra to setup, so they would actually not run in CI but people can run locally to check quickly, in an ideal world we would have api keys in CI for all providers though.

pkg/api/universal/conversation_middleware.go
filintod: not really, thought I would but no, I'll remove

dapr/proto/runtime/v1/dapr.proto
filintod: I see we do use the one-of in some of our stream responses already so I guess this is ok for now. As mentioned, I don't foresee adding new fields different to the one of one

go.mod
filintod: @sicoyle fyi to comment back when [comp](https://github.com/dapr/components-contrib/pull/3857) is merged

dapr/proto/runtime/v1/dapr.proto
filintod: I actually used uint32, that is 4G and more than enough for token count.   there were also some issues with protojson converting 64 bit number to string

JoshVanL:
In favor of https://github.com/dapr/dapr/pull/8851

PR #8795, opened 2025-06-16
Add target names to the healthz checks

acroca:
# Description

Updated the health check target registration across various components to include specific names for better identification.

Also modified the health check implementation to log the unhealthy targets

## Checklist

- [ ] Code compiles correctly
- [ ] Created/updated tests
- [ ] Unit tests passing


pkg/healthz/server/server.go
```diff
@@ -50,6 +50,9 @@ func New(opts Options) *Server {
 	mux := http.NewServeMux()
 	mux.HandleFunc("/healthz", func(w http.ResponseWriter, r *http.Request) {
 		if !opts.Healthz.IsReady() {
+			for _, name := range opts.Healthz.GetUnhealthyTargets() {
+				opts.Log.WithFields(map[string]any{"target": name}).Debug("target is unhealthy")
```
JoshVanL: I think we should only print a single log line for all unhealthys as a list

pkg/healthz/healthz.go
```diff
@@ -14,43 +14,67 @@ limitations under the License.
...
-	AddTarget() Target
-	AddTargetSet(int) []Target
+	GetUnhealthyTargets() []string
+	AddTarget(name string) Target
 }
 
 type healthz struct {
-	targetsReady atomic.Int64
-	targets      atomic.Int64
+	mu        sync.RWMutex
```
JoshVanL: Nit 
```suggestion
	lock        sync.RWMutex
```

pkg/healthz/server/server.go
acroca: I thought about it, but I find it more convenient to have one message per target so messages can be handled independently for alerting and also to figure out if a target is failing you can just do simple matching and not something like 'contains'. But I'm fine either way, I see pros/cons of either approach.

JoshVanL: Yes, I say because this will be incredibly chatty on daprd, particularly on start up

pkg/actors/internal/placement/placement.go
```diff
@@ -129,7 +129,7 @@ func New(opts Options) (Interface, error) {
 		hostname:      opts.Hostname,
 		operationLock: fifo.New(),
 		apiLevel:      opts.APILevel,
-		htarget:       opts.Healthz.AddTarget(),
+		htarget:       opts.Healthz.AddTarget("placement-service"),
```
cicoyle: I see `placement-service` twice - is that intentional? Or should this be called:
```suggestion
		htarget:       opts.Healthz.AddTarget("internal-placement-service"),
```

PR #8799, opened 2025-06-17
[1.15] Fix Scheduler Deadlock under high load

JoshVanL:
Problem

Under high load, like with high Workflow through put, a Scheduler would deadlock.

Impact

Workflows and the Jobs API would fail to trigger.

Root cause

The Scheduler would not retry or handle transient Etcd errors related to high contention.

Solution

Add retry logic to the Scheduler to handle transient Etcd errors, preventing deadlocks and ensuring that Workflows and Jobs can be triggered successfully under high load conditions.


docs/release_notes/v1.15.6.md
```diff
@@ -117,3 +118,21 @@ Daprd would connect to Scheduler to receive Workflow work before Placement had d
 ### Solution
 
 Ensure that the daprd connects to the Scheduler only after Placement has disseminated, preventing errors and ensuring that the daprd is ready to process Workflow work without contention.
+
+## Fix Scheduler Deadlock under high load
+
+### Problem
+
+Under high load, like with high Workflow through put, a Scheduler would deadlock.
```
nelson-parente, line 126: nit: should we define high load? what are the figures that we consider high load today?

nelson-parente (approved):

docs/release_notes/v1.15.6.md
JoshVanL: It is kind of arbitrary. It happens when the Etcd database server is under resource contention that this happens, which is sort of dependent of the size of the box.

yaron2 (approved):

PR #8803, opened 2025-06-17
Integration: add test for terminated a completed workflow

JoshVanL:


cicoyle (commented):
lgtm but linter was failing

PR #8804, opened 2025-06-17
Fix Scheduler Deadlock under high load

JoshVanL:
Fix Scheduler Deadlock under high load

Problem

Under high load, like with high Workflow through put, a Scheduler would deadlock.

Impact

Workflows and the Jobs API would fail to trigger.

Root cause

The Scheduler would not retry or handle transient Etcd errors related to high contention.

Solution

Add retry logic to the Scheduler to handle transient Etcd errors, preventing deadlocks and ensuring that Workflows and Jobs can be triggered successfully under high load conditions.


cicoyle (approved):

yaron2:
Conflicts

nelson-parente (approved):

PR #8814, opened 2025-06-20
Add support for slash containing keys in statestore get operation

### LINKED ISSUE #8746: GET statestore API fails for keys with slashes despite successful save
> /area runtime

## What version of Dapr?

> 1.15.5

## Expected Behavior

GET operations to the Dapr State API should support keys that contain any valid character, including slashes (`/`). This flexibility is critical for many use cases where structured or hierarchical keys are used (e.g., `user/123/settings`).

Saving such keys (with slashes) works fine using the POST/PUT methods, indicating that the underlying state store does not impose this limitation. Therefore, the GET operation should not impose restrictions that are inconsistent with the save operation or the underlying store capabilities.

## Actual Behavior

When attempting to retrieve a key with a slash (/) in its value via the GET endpoint:

```http 
GET http://localhost:<daprPort>/v1.0/state/<storename>/<key-with-slash>
```
The following error is returned:
```json
{
  "error": "Not Found",
  "error_msg": "{\"errorCode\":\"ERR_DIRECT_INVOKE\",\"message\":\"failed getting app id either from the URL path or the header dapr-app-id\"}",
  "status": 404
}
```

This error is misleading—it refers to a failure to get the app id, which is unrelated to the actual issue (a malformed URL due to the unescaped slash).

1. Save a state value with a key that includes a slash:

```bash
curl -X POST http://localhost:<daprPort>/v1.0/state/<storename> \
-H "Content-Type: application/json" \
-d '[{"key": "user/123/settings", "value": "dark_mode"}]'
````

2. Try to retrieve the state:

```bash
curl http://localhost:<daprPort>/v1.0/state/<storename>/user/123/settings
```

3. Observe the 404 response and unrelated error message.

### Technical Proposal

**Suggested Fix:**
Allow the GET endpoint to accept keys that contain special characters, especially slashes, by:

```http
GET /v1.0/state/<storename>/user%2F123%2Fsettings
```

Alternatively, allow the key to be passed as a query parameter to bypass path-related encoding issues:

```http
GET /v1.0/state/<storename>?key=user/123/settings
```

This approach ensures full support for hierarchical and structured keys without breaking existing storage patterns or introducing unnecessary constraints.

#### Release Note

RELEASE NOTE: FIX Allow GET state API to support keys with special characters (e.g., slashes) by enabling proper decoding or supporting query parameter access.

cr3ativ3cod3r:
# Description

To ensure backward compatibility, i added a new endpoint that will take key as a request parameter. This endpoint also uses the same handler. In the handler if the key is not found in the url, it will check for the key in the parameter. I also added tests for all test cases existing for the current endpoint.

## Issue reference

Please reference the issue this PR will close: #8746 

## Checklist

- [x] Code compiles correctly
- [x] Created/updated tests
- [x] Unit tests passing
- [x] Extended the documentation / Created issue in the <https://github.com/dapr/docs/> repo: dapr/docs#4681
- [x] Specification has been updated / Created issue in the <https://github.com/dapr/docs/> repo: dapr/docs#4681
- [x] Provided sample for the feature / Created issue in the <https://github.com/dapr/docs/> repo: dapr/docs#4681


knotseaborg:
@cr3ativ3cod3r 

Since malformed URL due to special characters in `<key>` is the issue, this endpoint will also require attention.

```http
DELETE http://localhost:<daprPort>/v1.0/state/<storename>/<key>
```

I noticed that optional parameters like `metadata` are passed as URL query parameters to the state management endpoints, as documented here: https://docs.dapr.io/reference/api/state_api/. This seems to be the established pattern.

This fix proposed in #8746 should be reconsidered as it neither breaks the pattern nor fragments existing endpoints.

> Suggested Fix:
> Allow the GET endpoint to accept keys that contain special characters, especially slashes, by:
> 
> GET /v1.0/state/\<storename\>/user%2F123%2Fsettings

cicoyle (changes requested):
Hey @cr3ativ3cod3r thanks for the PR here 🎉 mind adding an integration for this? You can see [state integration tests in this directory.](https://github.com/dapr/dapr/tree/master/tests/integration/suite/daprd/state)

Also, mind removing the additional changes in format in all the other files shown, as it makes the diff a lot bigger

cr3ativ3cod3r:
@cicoyle Thank you for taking the time to review my pr. I have removed additional changes in format in other files and also wrote integration tests as requested. Looking forward to your feedback.

cr3ativ3cod3r:
Hi @cicoyle, just checking in to see if there's anything else I should address from my side.

daixiang0:
@cr3ativ3cod3r please fix CI

PR #8829, opened 2025-06-23
adds converstation API tool calling

filintod:
# Description

Depends on https://github.com/dapr/components-contrib/pull/3857
Stacked over https://github.com/dapr/dapr/pull/8790

It adds tool calling capability to dapr for 1.16

## Issue reference

## Checklist

- [ ] Code compiles correctly
- [ ] Created/updated tests
- [ ] Unit tests passing

### DIFF
--- a/pkg/conversation/streaming.go
+++ b/pkg/conversation/streaming.go
@@ -15,6 +15,7 @@ package conversation
 
 import (
 	"context"
+	"errors"
 
 	piiscrubber "github.com/aavaz-ai/pii-scrubber"
 
@@ -91,7 +92,13 @@ func (p *StreamingPipelineImpl) ProcessStream(
 ) (*runtimev1pb.ConversationUsage, error) {
 	// Try to get the underlying LangChain Go model for streaming support
 	if streamer, ok := component.(contribConverse.StreamingConversation); ok {
-		return p.processRealStreaming(ctx, stream, streamer, request, componentName, scrubber)
+		usage, err := p.processRealStreaming(ctx, stream, streamer, request, componentName, scrubber)
+		if err == nil {
+			return usage, nil
+		}
+		if !(errors.Is(err, contribConverse.ErrStreamingNotSupported) || errors.Is(err, contribConverse.ErrToolCallStreamingNotSupported)) {
+			return nil, err
+		}
 	}
 
 	// Fallback: simulate streaming with existing Converse method
--- a/tests/integration/suite/daprd/conversation/grpc/streaming.go
+++ b/tests/integration/suite/daprd/conversation/grpc/streaming.go
@@ -435,4 +435,97 @@ func (s *streaming) Run(t *testing.T, ctx context.Context) {
 			}
 		})
 	}
+
+	// Test streaming fallback mechanism - this specifically tests the automatic fallback from
+	// real streaming to simulated streaming when Anthropic component supports streaming
+	// but returns ErrToolCallStreamingNotSupported when doing streaming + tool calling
+	t.Run("streaming fallback mechanism - anthropic tool calling", func(t *testing.T) {
+		apiKey := os.Getenv("ANTHROPIC_API_KEY")
+		if apiKey == "" {
+			t.Skip("ANTHROPIC_API_KEY not set, skipping Anthropic streaming fallback test")
+		}
+
+		// Use Anthropic with tool calling - this should trigger the fallback mechanism
+		// because Anthropic supports streaming but not streaming + tool calling
+		stream, err := client.ConverseStreamAlpha1(ctx, &rtv1.ConversationRequest{
+			Name: "anthropic",
+			Inputs: []*rtv1.ConversationInput{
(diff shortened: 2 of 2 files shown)


dapr/proto/runtime/v1/dapr.proto
```diff
@@ -1340,6 +1343,16 @@ message ConversationInput {
 
   // Scrub PII data that goes into the LLM
   optional bool scrubPII = 3;
+  
+  // Tool calling support
```
nelson-parente: should these be wrapped in their own object?

dapr/proto/runtime/v1/dapr.proto
```diff
@@ -1357,4 +1377,82 @@ message ConversationResponse {
...
+}
+
+// ConversationStreamChunk represents a streaming content chunk.
+message ConversationStreamChunk {
+  // Streaming content chunk
+  string content = 1;
+  
+  // Tool calling support for streaming
+  // Tool calls in this chunk (partial or complete)
+  repeated ToolCall tool_calls = 2;
```
nelson-parente: how is this handled when sent partially?

dapr/proto/runtime/v1/dapr.proto
```diff
@@ -1357,4 +1377,82 @@ message ConversationResponse {
...
+  // Number of tokens in the completion
+  optional int32 completion_tokens = 2 [json_name = "completionTokens"];
+  // Total number of tokens used
+  optional int32 total_tokens = 3 [json_name = "totalTokens"];
+}
+
+// Tool represents a function that can be called by the LLM (used on the request to the LLM)
+message Tool {
+  // The type of tool (always "function" for now)
+  string type = 1;
```
nelson-parente, line 1457: should this be typed?

pkg/diagnostics/component_monitoring.go
```diff
@@ -360,18 +380,29 @@ func (c *componentMetrics) ConfigurationInvoked(ctx context.Context, component,
 	}
 }
 
-// SecretInvoked records the metrics for a secret event.
-func (c *componentMetrics) ConversationInvoked(ctx context.Context, component string, success bool, elapsed float64) {
+// ConversationInvoked records the metrics for a secret event.
```
nelson-parente, line 383: nit: for a ... conversation event.

pkg/diagnostics/component_monitoring_test.go
```diff
@@ -203,29 +205,114 @@ func TestSecrets(t *testing.T) {
 	})
 }
 
+// removeDuplicatedDataRows when dealing with counters for the same view, we need to remove the duplicated data rows that could appear in back to back calls
+func removeDuplicatedDataRows(prevData, currData []*view.Row) []*view.Row {
+	// If there are no current data, return nil
```
nelson-parente, line 210: nit: if there is no data, return nil

pkg/api/universal/conversation_stream_test.go
```diff
@@ -0,0 +1,697 @@
...
+}
+
+func (m *mockConversationComponent) Close() error {
+	return nil
+}
+
+// Mock tool calling conversation component that simulates OpenAI-style tool calling
+
+// Mock streaming-capable conversation component
+type mockStreamingConversationComponent struct {
```
nelson-parente, line 69: nit: split mock definition from test

pkg/api/universal/conversation_stream.go
```diff
@@ -0,0 +1,358 @@
...
+	"github.com/dapr/dapr/pkg/messages"
+	runtimev1pb "github.com/dapr/dapr/pkg/proto/runtime/v1"
+	"github.com/dapr/kit/logger"
+	kmeta "github.com/dapr/kit/metadata"
+)
+
+// defaultPIIScrubberWindowSize is the default window size (bytes) for PII scrubbing
+// this allows for a look-ahead window of stream data to be scrubbed
+// TODO: make this configurable
+const defaultPIIScrubberWindowSize = 50
```
nelson-parente: why 50?

yaron2:
conflicts

pkg/diagnostics/component_monitoring.go
filintod: do not follow your comment?

dapr/proto/runtime/v1/dapr.proto
filintod: we could use an enum for it

dapr/proto/runtime/v1/dapr.proto
filintod: you mean tools, probably actually here we should not return tools as that is the definition, more than the request/response, but I need to test as I've had issues with some providers when the response is not in the right sequence

filintod:
copying to streaming PR to consolidate there

pkg/diagnostics/component_monitoring.go
filintod: I see now on the comment, fix

PR #8830, opened 2025-06-23
Feat cross app workflow activities

cicoyle:
Enable cross app workflows such that one app can host the orchestration while another may simply host and run activities. Previously, workflows were limited to single app execution whereby both orchestration & activities had to run within the same application. This PR enables cross app workflow activity calls enabling the following scenarios:
- Cross-language workflow development (e.g., Python orchestrator calling Go activities)
- Resource isolation between orchestration and activity execution
- Distributed workload execution across multiple machines/services
- Specialized deployment patterns for AI/ML workloads

TODO: cross app orchestration calls


go.mod
```diff
@@ -513,6 +513,8 @@ replace (
 //
 // replace github.com/dapr/components-contrib => ../components-contrib
 // replace github.com/dapr/kit => ../kit
+replace github.com/dapr/durabletask-go => github.com/cicoyle/durabletask-go v0.0.0-20250618220139-02008f613541
```
cicoyle: [I'll rm this once the durabletask-go PR is merged](https://github.com/dapr/durabletask-go/pull/26)

cicoyle: Ill fix conflicts and update this commit sha tomorrow

pkg/actors/targets/workflow/activity/execute.go
```diff
@@ -109,9 +109,19 @@ func (a *activity) executeActivity(ctx context.Context, name string, taskEvent *
 		return todo.RunCompletedTrue, err
 	}
 
+	// send completed event to orchestrator wf actor
+	var wfActorType string
+	wfActorType = a.workflowActorType
+	if router := taskEvent.GetRouter(); router != nil {
+		sourceAppID := router.GetSource()
+		if sourceAppID != "" && sourceAppID != a.appID { // indicates cross app activity
+			wfActorType = fmt.Sprintf("dapr.internal.%s.%s.workflow", a.namespace, sourceAppID)
```
JoshVanL: Please make this a helper function in `pkg/actors/targets/workflow`.

pkg/actors/targets/workflow/activity/execute.go
```diff
@@ -109,9 +109,19 @@ func (a *activity) executeActivity(ctx context.Context, name string, taskEvent *
 		return todo.RunCompletedTrue, err
 	}
 
+	// send completed event to orchestrator wf actor
+	var wfActorType string
+	wfActorType = a.workflowActorType
+	if router := taskEvent.GetRouter(); router != nil {
+		sourceAppID := router.GetSource()
+		if sourceAppID != "" && sourceAppID != a.appID { // indicates cross app activity
```
JoshVanL: We should be checking for `nil`s for existence, not empty strings.

JoshVanL: Why do this check at all?

```go
sourceAppID != a.appID 
```

pkg/actors/targets/workflow/activity/execute.go
```diff
@@ -109,9 +109,19 @@ func (a *activity) executeActivity(ctx context.Context, name string, taskEvent *
 		return todo.RunCompletedTrue, err
 	}
 
+	// send completed event to orchestrator wf actor
+	var wfActorType string
+	wfActorType = a.workflowActorType
```
JoshVanL:
```suggestion
	wfActorType := a.workflowActorType
```

pkg/actors/targets/workflow/activity/reminder.go
```diff
@@ -39,6 +39,9 @@ func (a *activity) createReminder(ctx context.Context, his *backend.HistoryEvent
 		return err
 	}
 
+	// The activity actor should always create reminders for its own actor type and ID
+	log.Debugf("Creating activity reminder for actorType %s and actorID %s", a.actorType, a.actorID)
```
JoshVanL: This is already being logged above.

```suggestion
```

pkg/actors/targets/workflow/orchestrator/activity.go
```diff
@@ -67,15 +67,29 @@ func (o *orchestrator) callActivity(ctx context.Context, e *backend.HistoryEvent
...
 	}
 
-	targetActorID := buildActivityActorID(o.actorID, e.GetEventId(), generation)
+	activityActorType := o.activityActorType
+	var targetAppID, sourceAppID string
+
+	if router := e.GetRouter(); router != nil {
+		targetAppID = router.GetTarget()
+		sourceAppID = router.GetSource()
+		log.Debugf("Cross-app activity call: target appID=%s, source appID=%s", targetAppID, sourceAppID)
```
JoshVanL: We don't need to be logging this- workflow is already incredibly chatty, and workflow instances are logged with their types already.
```suggestion
```

pkg/actors/targets/workflow/orchestrator/activity.go
```diff
@@ -67,15 +67,29 @@ func (o *orchestrator) callActivity(ctx context.Context, e *backend.HistoryEvent
...
+	}
+
+	// If target app is specified and different from current app, use cross-app actor type
+	if targetAppID != "" && targetAppID != o.appID {
+		activityActorType = fmt.Sprintf("dapr.internal.%s.%s.activity", o.namespace, targetAppID)
+	}
```
JoshVanL: Shouldn't check empty string for existence.

Use helper func for string construction.

pkg/actors/targets/workflow/orchestrator/reminder.go
```diff
@@ -53,8 +52,14 @@ func (o *orchestrator) createReminder(ctx context.Context, namePrefix string, da
 		}
 	}
 
+	actorType := o.actorType
+	if targetAppID != "" && targetAppID != o.appID {
+		actorType = fmt.Sprintf("dapr.internal.%s.%s.workflow", o.namespace, targetAppID)
+	}
+	log.Debugf("Workflow actor '%s||%s': creating '%s' reminder with DueTime = '%s'", actorType, o.actorID, reminderName, delay)
+
```
JoshVanL: Same here.

tests/integration/suite/daprd/workflow/crossapp/appdown.go
```diff
@@ -0,0 +1,165 @@
...
+func init() {
+	suite.Register(new(appdown))
+}
+
+// appdown tests error handling when a target app goes down during execution
+type appdown struct {
+	daprd1 *daprd.Daprd
+	daprd2 *daprd.Daprd
+	place  *placement.Placement
+	sched  *scheduler.Scheduler
```
JoshVanL, line 47:
```suggestion
	workflow  *workflow.Workflow
```

JoshVanL: Use `workflow.WithDaprds` and move app ID configuration into framework .

tests/integration/suite/daprd/workflow/crossapp/appdown.go
```diff
@@ -0,0 +1,165 @@
...
+
+	// App2: Activity that will be called before the app goes down
+	a.registry2.AddActivityN("ProcessData", func(ctx task.ActivityContext) (any, error) {
+		var input string
+		if err := ctx.GetInput(&input); err != nil {
+			return nil, fmt.Errorf("failed to get input in app2: %w", err)
+		}
+
+		// Sleep for a while to ensure the workflow is actively executing when app2 is stopped
+		time.Sleep(10 * time.Second)
```
JoshVanL: Please do not use sleep. Use channels to signal processes are at different checkpoints.

tests/integration/suite/daprd/workflow/crossapp/appdown.go
```diff
@@ -0,0 +1,165 @@
...
+
+	id, err := client1.ScheduleNewOrchestration(ctx, "AppDownWorkflow", api.WithInput("Hello from app1"))
+	assert.NoError(t, err)
+
+	// Wait a bit for the workflow to start executing & Stop app2 to simulate app going down
+	time.Sleep(1 * time.Second)
+	cancel()
+	a.daprd2.Cleanup(t)
+
+	// Expect completion to hang, so timeout
```
JoshVanL: Why do we expect completion to hang?

tests/integration/suite/daprd/workflow/crossapp/appdown.go
```diff
@@ -0,0 +1,165 @@
...
+	err := client1.StartWorkItemListener(ctx, a.registry1)
+	assert.NoError(t, err)
+	err = client2.StartWorkItemListener(wctx, a.registry2)
+	assert.NoError(t, err)
+
+	id, err := client1.ScheduleNewOrchestration(ctx, "AppDownWorkflow", api.WithInput("Hello from app1"))
+	assert.NoError(t, err)
+
+	// Wait a bit for the workflow to start executing & Stop app2 to simulate app going down
+	time.Sleep(1 * time.Second)
```
JoshVanL: Don't use sleep, use channels as signals of check points.

tests/integration/suite/daprd/workflow/crossapp/basiclogline.go
```diff
@@ -0,0 +1,171 @@
+/*
```
JoshVanL: Using log matching to check for cross app functionality doesn't seem quite right.
We should test functionality by the fact of activities etc. being executed.

tests/integration/suite/daprd/workflow/crossapp/callactivity.go
```diff
@@ -0,0 +1,142 @@
...
+func init() {
+	suite.Register(new(callactivity))
+}
+
+// callactivity demonstrates calling activities across different Dapr applications.
+type callactivity struct {
+	daprd1 *daprd.Daprd
+	daprd2 *daprd.Daprd
+	place  *placement.Placement
+	sched  *scheduler.Scheduler
```
JoshVanL: Worklflows here and other places.

tests/integration/suite/daprd/workflow/crossapp/callactivity.go
```diff
@@ -0,0 +1,142 @@
...
+	app1 := app.New(t)
+	app2 := app.New(t)
+
+	// Create registries for each app and register orchestrator/activity
+	c.registry1 = task.NewTaskRegistry()
+	c.registry2 = task.NewTaskRegistry()
+
+	appID1 := uuid.New().String()
+	appID2 := uuid.New().String()
+
```
JoshVanL: Not pre generate app IDs. Use `daprd.AppID()`

tests/integration/suite/daprd/workflow/crossapp/callactivity.go
```diff
@@ -0,0 +1,142 @@
...
+			return nil, fmt.Errorf("failed to execute activity in app2: %w", err)
+		}
+		return output, nil
+	})
+
+	c.daprd1 = daprd.New(t,
+		daprd.WithInMemoryActorStateStore("mystore"),
+		daprd.WithPlacementAddresses(c.place.Address()),
+		daprd.WithScheduler(c.sched),
+		daprd.WithAppID(appID1),
```
JoshVanL:
```suggestion
```

tests/integration/suite/daprd/workflow/crossapp/invalidappid.go
```diff
@@ -0,0 +1,132 @@
...
+		if err == nil {
+			return nil, fmt.Errorf("expected error when calling non-existent app, but got none")
+		}
+		return fmt.Sprintf("Error handled: %v", err), nil
+	})
+
+	i.actorNotFoundLogLine = logline.New(t,
+		logline.WithStdoutLineContains(
+			"did not find address for actor",
+		),
```
JoshVanL: Don't use log lines to check for failure in APIs.

tests/integration/suite/daprd/workflow/crossapp/invalidappid.go
```diff
@@ -0,0 +1,132 @@
...
+	i.daprd1.WaitUntilRunning(t, ctx)
+
+	// Start workflow listener for app1
+	client1 := client.NewTaskHubGrpcClient(i.daprd1.GRPCConn(t, ctx), backend.DefaultLogger())
+
+	err := client1.StartWorkItemListener(ctx, i.registry1)
+	assert.NoError(t, err)
+
+	// ctx cancel bc it will hang
+	wCtx, wcancel := context.WithTimeout(ctx, 20*time.Second)
```
JoshVanL: This is too long. Use channels for checkpointing.

tests/integration/suite/daprd/workflow/crossapp/invalidappid.go
```diff
@@ -0,0 +1,132 @@
...
+		var result string
+		err := ctx.CallActivity("ProcessData",
+			task.WithActivityInput(input),
+			task.WithAppID("nonexistent-app")).
+			Await(&result)
+
+		// This should fail, so we expect an error
+		if err == nil {
+			return nil, fmt.Errorf("expected error when calling non-existent app, but got none")
+		}
```
JoshVanL:
```suggestion
```

tests/integration/suite/daprd/workflow/crossapp/manyapps.go
```diff
@@ -0,0 +1,259 @@
...
+	m.registry1 = task.NewTaskRegistry()
+	m.registry2 = task.NewTaskRegistry()
+	m.registry3 = task.NewTaskRegistry()
+	m.registry4 = task.NewTaskRegistry()
+	m.registry5 = task.NewTaskRegistry()
+
+	appID1 := uuid.New().String()
+	appID2 := uuid.New().String()
+	appID3 := uuid.New().String()
+	appID4 := uuid.New().String()
```
JoshVanL: Don't set custom uuids

tests/integration/suite/daprd/workflow/crossapp/manyapps.go
```diff
@@ -0,0 +1,259 @@
...
+// manyapps demonstrates complex cross-app workflows with many apps
+type manyapps struct {
+	daprd1 *daprd.Daprd
+	daprd2 *daprd.Daprd
+	daprd3 *daprd.Daprd
+	daprd4 *daprd.Daprd
+	daprd5 *daprd.Daprd
+	place  *placement.Placement
+	sched  *scheduler.Scheduler
+
```
JoshVanL: wokrflow.Workflow

tests/integration/suite/daprd/workflow/crossapp/parallel.go
```diff
@@ -0,0 +1,272 @@
...
+			return nil, fmt.Errorf("failed to get input in app3: %w", err)
+		}
+
+		p.activityStarted.Add(1)
+		select {
+		case <-p.releaseCh:
+		case <-time.After(10 * time.Second):
+			return nil, fmt.Errorf("timeout waiting for release")
+		}
+
```
JoshVanL: Remove all instances of these

tests/integration/suite/daprd/workflow/crossapp/parallel.go
```diff
@@ -0,0 +1,272 @@
...
+	p.registry2.AddActivityN("ProcessData2", func(ctx task.ActivityContext) (any, error) {
+		var input string
+		if err := ctx.GetInput(&input); err != nil {
+			return nil, fmt.Errorf("failed to get input in app2: %w", err)
+		}
+
+		p.activityStarted.Add(1)
+		select {
+		case <-p.releaseCh:
+		case <-time.After(10 * time.Second):
```
JoshVanL: Do we have a test for parallel activities?

tests/integration/suite/daprd/workflow/crossapp/restart.go
```diff
@@ -0,0 +1,132 @@
...
+		daprd1.Cleanup(t)
+		daprd2.Cleanup(t)
+	}
+
+	// Ensure all workflows take similar amounts of time.
+	for _, d1 := range timeTaken {
+		for _, d2 := range timeTaken {
+			assert.InDelta(t, d1.Seconds(), d2.Seconds(), float64(time.Second))
+		}
+	}
```
JoshVanL: Why?

tests/integration/suite/daprd/workflow/crossapp/security.go
```diff
@@ -0,0 +1,150 @@
...
+
+	id, err := client1.ScheduleNewOrchestration(ctx, "CrossAppWorkflow", api.WithInput("Hello from app1"))
+	require.NoError(t, err)
+
+	_, err = client1.WaitForOrchestrationStart(ctx, id)
+	require.NoError(t, err)
+
+	// this is key, and should fail
+	_, err = client2.FetchOrchestrationMetadata(ctx, id, api.WithFetchPayloads(true))
+	require.Error(t, err, "no such instance exists", "security breach, this should err out")
```
JoshVanL: I guess thie is validating `FetchOrchestrationMetadata` is scoped to a single app ID? This doesn't belong in the `crossapps` package.

Please add a test showing that calling an activity of an app ID in another _namespace_ doesn't work.

tests/integration/suite/daprd/workflow/security.go
```diff
@@ -0,0 +1,149 @@
...
+}
+
+func (s *security) Run(t *testing.T, ctx context.Context) {
+	s.sched.WaitUntilRunning(t, ctx)
+	s.place.WaitUntilRunning(t, ctx)
+	s.daprd1.WaitUntilRunning(t, ctx)
+	s.daprd2.WaitUntilRunning(t, ctx)
+
+	// Start workflow listeners for each app
+	client1 := client.NewTaskHubGrpcClient(s.daprd1.GRPCConn(t, ctx), backend.DefaultLogger())
```
JoshVanL: Duplicate of above?

JoshVanL (changes requested):

pkg/actors/targets/workflow/activity/execute.go
cicoyle: I made a new `common` pkg there otherwise was seeing cyclical dep issues.

pkg/actors/targets/workflow/activity/reminder.go
cicoyle: oops, yes, I think I logged this additionally and was using this locally for debugging and forgot to remove it. good catch

tests/integration/suite/daprd/workflow/crossapp/appdown.go
cicoyle: The workflow engine is designed to retry failed activities indefinitely until they succeed. In a real-world scenario, if an app goes down and never comes back up, the workflow would hang forever. However, since we can't let tests run indefinitely, I use a context timeout to verify that the workflow does indeed hang (as expected) rather than completing successfully. I've tested the transient app going down and back up again, via the restart test.

tests/integration/suite/daprd/workflow/crossapp/basiclogline.go
cicoyle: I added various tests ensuring the functionality already, this was more of a nice to have added test verifying logs that I was personally looking at while debugging. I can remove it.

tests/integration/suite/daprd/workflow/crossapp/parallel.go
cicoyle: I was aligning the cross app parallel test with [this existing parallel test written by you](https://github.com/dapr/dapr/blob/master/tests/integration/suite/daprd/workflow/parallel.go#L99-L102)... You still want it removed?

tests/integration/suite/daprd/workflow/crossapp/parallel.go
cicoyle: yes, [please see this existing test I was aligning with logic wise](https://github.com/dapr/dapr/blob/master/tests/integration/suite/daprd/workflow/parallel.go#L99-L102), just the crossapp folder test does a cross app call.

tests/integration/suite/daprd/workflow/crossapp/restart.go
cicoyle: [alignment with this non cross app test written by you](https://github.com/dapr/dapr/blob/master/tests/integration/suite/daprd/workflow/restart.go). happy to remove this last part

tests/integration/suite/daprd/workflow/crossapp/security.go
cicoyle: added a new `tests/integration/suite/daprd/workflow/security/appscoping.go` and `tests/integration/suite/daprd/workflow/crossapp/crossnamespace.go`

tests/integration/suite/daprd/workflow/security.go
cicoyle: rmed this file instead, and am using the appscoping and crossnamespace tests

tests/integration/suite/daprd/workflow/crossapp/appdown.go
cicoyle: Per our discussion, this will stay as is - similar to how the reconnect scheduler test requires individual processes rather than the test framework

pkg/actors/targets/workflow/activity/execute.go
```diff
@@ -109,9 +110,19 @@ func (a *activity) executeActivity(ctx context.Context, name string, taskEvent *
...
+	// send completed event to orchestrator wf actor
+	wfActorType := a.workflowActorType
+	if router := taskEvent.GetRouter(); router != nil {
+		sourceAppID := router.GetSource()
+		wfActorType, err = common.GetActorType(a.namespace, sourceAppID, common.WfActor)
+		if err != nil {
+			return todo.RunCompletedFalse, err
+		}
+	}
+
```
acroca, line 117: Would it be useful here to add a `log.Debug` like in the orchestrator `callActivity`?

pkg/actors/targets/workflow/common/common.go
```diff
@@ -0,0 +1,36 @@
...
+func GetActorType(ns, appID string, wfActorType WfActorTypes) (string, error) {
+	switch wfActorType {
+	case WfActor:
+		return "dapr.internal." + ns + "." + appID + ".workflow", nil
+	case WfActivityActor:
+		return "dapr.internal." + ns + "." + appID + ".activity", nil
+	default:
+		return "", errors.New("unknown workflow actor type")
+	}
+}
```
acroca: What do you think if we create a function for each type? this way we don't need to deal with errors in the caller side and keeps the code simpler too.
```suggestion
func GetWorkflowActorType(ns, appID string) string {
	return "dapr.internal." + ns + "." + appID + ".workflow"
}

func GetActivityActorType(ns, appID string) string {
	return "dapr.internal." + ns + "." + appID + ".activity"
}
```

pkg/runtime/wfengine/backends/actors/actors.go
```diff
@@ -387,7 +391,7 @@ func (*Actors) CompleteActivityWorkItem(ctx context.Context, wi *backend.Activit
 }
 
 // CompleteOrchestrationWorkItem implements backend.Backend
-func (*Actors) CompleteOrchestrationWorkItem(ctx context.Context, wi *backend.OrchestrationWorkItem) error {
+func (abe *Actors) CompleteOrchestrationWorkItem(ctx context.Context, wi *backend.OrchestrationWorkItem) error {
```
acroca: Do we really need a name here? Is it to make the linter happy?
```suggestion
func (a *Actors) CompleteOrchestrationWorkItem(ctx context.Context, wi *backend.OrchestrationWorkItem) error {
```

tests/integration/framework/process/workflow/workflow.go
```diff
@@ -31,11 +32,11 @@ import (
 )
 
 type Workflow struct {
-	registry *task.TaskRegistry
-	db       *sqlite.SQLite
-	place    *placement.Placement
-	sched    *scheduler.Scheduler
-	daprds   []*daprd.Daprd
+	taskregistry atomic.Value // map[int]*task.TaskRegistry
```
acroca: Q: Do we really need this to be atomic? It's only set during the `New`.

tests/integration/framework/process/workflow/workflow.go
```diff
@@ -134,13 +159,30 @@ func (w *Workflow) WaitUntilRunning(t *testing.T, ctx context.Context) {
 }
 
 func (w *Workflow) Registry() *task.TaskRegistry {
-	return w.registry
+	registries := w.taskregistry.Load().(map[int]*task.TaskRegistry)
+	return registries[0]
```
acroca:
```suggestion
	return w.RegistryN(0)
```

tests/integration/framework/process/workflow/workflow.go
```diff
@@ -134,13 +159,30 @@ func (w *Workflow) WaitUntilRunning(t *testing.T, ctx context.Context) {
...
+	registries := w.taskregistry.Load().(map[int]*task.TaskRegistry)
+	return registries[index]
 }
 
 func (w *Workflow) BackendClient(t *testing.T, ctx context.Context) *client.TaskHubGrpcClient {
 	t.Helper()
 	backendClient := client.NewTaskHubGrpcClient(w.daprds[0].GRPCConn(t, ctx), logger.New(t))
-	require.NoError(t, backendClient.StartWorkItemListener(ctx, w.registry))
+	require.NoError(t, backendClient.StartWorkItemListener(ctx, w.Registry()))
+	return backendClient
```
acroca, line 179:
```suggestion
	return w.BackendClientN(0)
```

tests/integration/suite/daprd/workflow/crossapp/callactivity.go
```diff
@@ -0,0 +1,90 @@
...
+	return []framework.Option{
+		framework.WithProcesses(c.workflow),
+	}
+}
+
+func (c *callactivity) Run(t *testing.T, ctx context.Context) {
+	c.workflow.WaitUntilRunning(t, ctx)
+
+	// Add orchestrator to app0's registry
+	c.workflow.Registry().AddOrchestratorN("CrossAppWorkflow", func(ctx *task.OrchestrationContext) (any, error) {
```
acroca, line 53: Since we're adding more later on, I think it's clearer to be explicit about the index
```suggestion
	c.workflow.RegistryN(0).AddOrchestratorN("CrossAppWorkflow", func(ctx *task.OrchestrationContext) (any, error) {
```

acroca: If you agree with this change, we'd need to replace all the `Registry()` and `BackendClient()` to the indexed one in the crossapp tests. But feel free to ignore if you have a different preference :)

tests/integration/suite/daprd/workflow/crossapp/crossnamespace.go
```diff
@@ -0,0 +1,99 @@
...
+
+func (c *crossnamespace) Run(t *testing.T, ctx context.Context) {
+	c.workflow.WaitUntilRunning(t, ctx)
+
+	// Start workflow listener for apps
+	client0 := c.workflow.BackendClient(t, ctx)
+	c.workflow.BackendClientN(t, ctx, 1)
+
+	// Expect completion to hang, so timeout
+	// dapr will log about 'did not find address for actor'
```
acroca, line 102: Should we add the `logline.WithStdoutLineContains("did not find address for actor")` like in other tests then?

tests/integration/suite/daprd/workflow/crossapp/multihop.go
```diff
@@ -0,0 +1,119 @@
...
+	"github.com/dapr/dapr/tests/integration/suite"
+	"github.com/dapr/durabletask-go/api"
+	"github.com/dapr/durabletask-go/task"
+)
+
+func init() {
+	suite.Register(new(multihop))
+}
+
+// multihop demonstrates multi-hop cross-app workflow: app0 → app1 → app2
```
acroca: I think this scenario is already covered in the `manyapps` test, isn't it?

tests/integration/suite/daprd/workflow/crossapp/orderingpreserved.go
```diff
@@ -0,0 +1,239 @@
...
+
+	// Helper func to record execution order atomically
+	recordExecution := func(activityName, appID string) {
+		o.activityCount.Add(1)
+		order := o.executionOrder.Load()
+
+		orderSlice := order.([]string)
+		orderSlice = append(orderSlice, activityName)
+		o.executionOrder.Store(orderSlice)
+	}
```
acroca: Don't you need a mutex here? If I'm not mistaken, If multiple parallel calls enter this block, it can happen that two goroutines execute the `Load()` at the same time, then each will append the `activityName` to the list they got and the `Store()` will override each other. Also with a mutex the `executionOrder` doesn't need to be an atomic value.

acroca: Well, this is going to execute sequentially because it calls `Await` on every activity call, neither the mutex or the atomic value is really required.

tests/integration/suite/daprd/workflow/crossapp/orderingpreserved.go
```diff
@@ -0,0 +1,239 @@
...
+	"github.com/dapr/durabletask-go/api"
+	"github.com/dapr/durabletask-go/task"
+)
+
+func init() {
+	suite.Register(new(orderingpreserved))
+}
+
+// orderingpreserved demonstrates that activity execution order is preserved
+// across local and cross-app calls using an atomic slice to track order
```
acroca: I'm not sure this test is covering much in terms of crossapp execution. The ordering is happening because of how the workflow code is structured, regardless of where the activities run. Or in other words, this is just how the workflows work, the `Await` is kind of enforcing the order.

tests/integration/suite/daprd/workflow/crossapp/parallel.go
```diff
@@ -0,0 +1,196 @@
...
+	"github.com/dapr/dapr/tests/integration/suite"
+	"github.com/dapr/durabletask-go/api"
+	"github.com/dapr/durabletask-go/task"
+)
+
+func init() {
+	suite.Register(new(parallel))
+}
+
+// parallel demonstrates parallel cross-app activity calls: app0 → app1, app0 → app2
```
acroca: Same here, I don't see a lot cross-app specific functionality tested here. This test is just making sure an existing feature (parallel activities) also works with cross-app, but I think we've covered cross-app execution and parallel execution, I don't know if testing both things together is so beneficial. 
Is parallel execution any different in the cross-app world?

tests/integration/suite/daprd/workflow/crossapp/restart.go
```diff
@@ -0,0 +1,89 @@
...
+	// Add activity to app1's registry
+	r.workflow.RegistryN(1).AddActivityN("bar", func(c task.ActivityContext) (any, error) {
+		return "processed by app1", nil
+	})
+
+	timeTaken := make([]time.Duration, 0, 5)
+	for range 5 {
+		// Start workflow listeners for each app with their respective registries
+		client0 := r.workflow.BackendClient(t, ctx) // app0 (orchestrator)
+		r.workflow.BackendClientN(t, ctx, 1)        // app1 (activity)
```
acroca: If I understand correctly, all the 5 calls to this function will create a client that can process the activity, so by the time we reach the 5th iteration there will be a pool of 5 'workers' that can process the activity, right? 
If that's correct, we might want to add something in this client so we can `Close()` them somehow so we make sure each iteration starts clean. What do you think?

tests/integration/suite/daprd/workflow/crossapp/restart.go
```diff
@@ -0,0 +1,89 @@
...
+	"github.com/dapr/dapr/tests/integration/suite"
+	"github.com/dapr/durabletask-go/api"
+	"github.com/dapr/durabletask-go/task"
+)
+
+func init() {
+	suite.Register(new(restart))
+}
+
+type restart struct {
```
acroca, line 45: it's missing the little comment that explains what this test covers. Also I'm not sure what this test really tests, it's just creating N clients to both apps and trigger the workflow, and make sure it all run in a similar amount of time (delta 1s). I don't see anything to do with restarts, but I might be missing something.

pkg/actors/targets/workflow/activity/execute.go
cicoyle: There is a log line above this section of code on line 102 `log.Debugf("Activity actor '%s': activity completed for workflow with instanceId '%s' activityName '%s'", a.actorID, wi.InstanceID, name)`

pkg/actors/targets/workflow/common/common.go
cicoyle: sure

pkg/runtime/wfengine/backends/actors/actors.go
cicoyle: this was a linter tweak bc I had a comment for a todo in here but removed the comment and _think_ I can remove this diff assuming linter remains happy

tests/integration/framework/process/workflow/workflow.go
cicoyle: think I'll tweak it to `w.Registry()` which is the same as the above 2 options, (what it currently is, and what you are suggesting to change it to) but this way I dont need to pass in the index

tests/integration/framework/process/workflow/workflow.go
cicoyle: same as the above comment, can tweak to `w.BackendClient()` which should be the 0th index, which is the same as your suggestion but cleaner to read

tests/integration/suite/daprd/workflow/crossapp/callactivity.go
cicoyle: I kept the `Registry()` as what it was previously and added the `<func>N` for anything that needs to be indexed - beyond 0. I originally took in an index for this func, but received feedback to keep the original intent and add the `N` funcs for indexes. I would prefer to use `Registry()` instead of `RegistryN(0)` even tho they are the exact same bc that is how it is for every test atm due to feedback I received.

tests/integration/suite/daprd/workflow/crossapp/crossnamespace.go
cicoyle: sure I can add it

pkg/actors/targets/workflow/activity/execute.go
```diff
@@ -109,9 +110,19 @@ func (a *activity) executeActivity(ctx context.Context, name string, taskEvent *
 		return todo.RunCompletedTrue, err
 	}
 
+	// send completed event to orchestrator wf actor
+	wfActorType := a.workflowActorType
+	if router := taskEvent.GetRouter(); router != nil {
+		sourceAppID := router.GetSource()
+		wfActorType, err = common.GetActorType(a.namespace, sourceAppID, common.WfActor)
+		if err != nil {
+			return todo.RunCompletedFalse, err
```
JoshVanL:
```suggestion
			return todo.RunCompletedTrue, err
```

pkg/actors/targets/workflow/activity/execute.go
```diff
@@ -109,9 +110,19 @@ func (a *activity) executeActivity(ctx context.Context, name string, taskEvent *
 		return todo.RunCompletedTrue, err
 	}
 
+	// send completed event to orchestrator wf actor
+	wfActorType := a.workflowActorType
+	if router := taskEvent.GetRouter(); router != nil {
+		sourceAppID := router.GetSource()
+		wfActorType, err = common.GetActorType(a.namespace, sourceAppID, common.WfActor)
```
JoshVanL:
```suggestion
		wfActorType, err = common.GetActorType(a.namespace, router.GetSource(), common.WfActor)
```

pkg/actors/targets/workflow/common/common.go
JoshVanL: :+1:

pkg/actors/targets/workflow/orchestrator/activity.go
```diff
@@ -67,15 +68,25 @@ func (o *orchestrator) callActivity(ctx context.Context, e *backend.HistoryEvent
 		return err
 	}
 
-	targetActorID := buildActivityActorID(o.actorID, e.GetEventId(), generation)
+	activityActorType := o.activityActorType
+	if router := e.GetRouter(); router != nil {
+		if router.Target != nil {
```
JoshVanL:
```suggestion
	if router := e.GetRouter(); router != nil && router.Target != nil {
```

pkg/actors/targets/workflow/orchestrator/add.go
```diff
@@ -47,7 +47,14 @@ func (o *orchestrator) addWorkflowEvent(ctx context.Context, historyEventBytes [
...
 
-	if _, err := o.createReminder(ctx, "new-event", nil, nil); err != nil {
+	// For activity completion events, we want to create the reminder on the same app where this workflow actor is
+	// hosted, so use the source app from the router
+	targetApp := o.appID
+	if e.GetRouter() != nil {
+		targetApp = e.GetRouter().GetSource()
+	}
+
+	if _, err := o.createReminder(ctx, "new-event", nil, nil, targetApp); err != nil {
```
JoshVanL: This is being sent to a different orchestrator actor ID, why?

pkg/actors/targets/workflow/orchestrator/run.go
```diff
@@ -68,6 +69,10 @@ func (o *orchestrator) runWorkflow(ctx context.Context, reminder *actorapi.Remin
 	for _, e := range state.Inbox {
 		if es := e.GetExecutionStarted(); es != nil {
 			esHistoryEvent = e
+			// Set the source app ID for cross-app routing in durabletask-go
+			esHistoryEvent.Router = &protos.TaskRouter{
+				Source: o.appID,
+			}
```
JoshVanL, line 75: This doesn't look like it is needed (?)

tests/integration/framework/process/workflow/options.go
```diff
@@ -16,18 +16,30 @@ package workflow
...
+	}
+	activities []struct {
+		index int
+		name  string
+		fn    func(task.ActivityContext) (any, error)
+	}
+	daprdOptions []struct {
+		index int
+		opts  []daprd.Option
+	}
```
JoshVanL: Please move to types defined outside of this struct - every time we write to these fields, we have to redefine these struct types again.

tests/integration/framework/process/workflow/workflow.go
JoshVanL: We do not need to make this atomic. It should be a slice.

tests/integration/framework/process/workflow/workflow.go
```diff
@@ -85,20 +85,45 @@ spec:
...
+
+	for i := range daprds {
+		for _, daprdOpt := range opts.daprdOptions {
+			if daprdOpt.index == i {
+				baseDopts = append(baseDopts, daprdOpt.opts...)
+				break
+			}
+		}
+
+		daprds[i] = daprd.New(t, baseDopts...)
```
JoshVanL: What is this doing?

tests/integration/framework/process/workflow/workflow.go
```diff
@@ -85,20 +85,45 @@ spec:
...
-	return &Workflow{
-		registry: opts.registry,
-		db:       db,
-		place:    place,
-		sched:    sched,
-		daprds:   daprds,
+	registries := make(map[int]*task.TaskRegistry)
+	for i := range daprds {
+		registries[i] = task.NewTaskRegistry()
+	}
```
JoshVanL, line 105: Registries should be a slice.

tests/integration/suite/daprd/workflow/crossapp/invalidappid.go
```diff
@@ -0,0 +1,98 @@
...
+		return fmt.Sprintf("Error handled: %v", err), nil
+	})
+
+	// Start workflow listener for app0
+	client0 := i.workflow.BackendClient(t, ctx)
+
+	// ctx cancel bc it will hang
+	wCtx, wcancel := context.WithTimeout(ctx, 15*time.Second)
+	defer wcancel()
```
JoshVanL: 15 seconds is too long to be waiting.

tests/integration/suite/daprd/workflow/crossapp/orderingpreserved.go
```diff
@@ -0,0 +1,239 @@
...
+	// App1: Remote activities
+	o.workflow.RegistryN(1).AddActivityN("RemoteActivity2", func(ctx task.ActivityContext) (any, error) {
+		recordExecution("RemoteActivity2", o.workflow.DaprN(1).AppID())
+		time.Sleep(10 * time.Millisecond)
+		return "remote2", nil
+	})
+
+	o.workflow.RegistryN(1).AddActivityN("RemoteActivity4", func(ctx task.ActivityContext) (any, error) {
+		recordExecution("RemoteActivity4", o.workflow.DaprN(1).AppID())
+		time.Sleep(10 * time.Millisecond)
```
JoshVanL: Why are we sleeping?

JoshVanL (changes requested):

tests/integration/framework/process/workflow/workflow.go
cicoyle: actually I cant bc that is the func name here. going to leave it as is to avoid another func hop

tests/integration/suite/daprd/workflow/crossapp/orderingpreserved.go
cicoyle: I can remove this test bc I have other tests ensuring input/output matches which is essentially ordering

tests/integration/suite/daprd/workflow/crossapp/parallel.go
cicoyle: its testing parallel activities that are cross app. I thought it was a nice to have, but not required test. I can remove it bc we separately test parallel activities elsewhere

pkg/actors/targets/workflow/orchestrator/add.go
cicoyle: this is just a naming issue. `targetapp` is the `sourceapp`, let me just use sourceapp instead for the var name

tests/integration/framework/process/workflow/workflow.go
cicoyle: To enable using the wf framework with diff dapr options per dapr instance like where I test crossnamespace for example, I need to be able to pass in diff opts per dapr which I accomplish via: 
```
workflow.WithDaprdOptions(0, daprd.WithNamespace("default"))
workflow.WithDaprdOptions(1, daprd.WithNamespace("other"))
```

or this:
```
workflow.WithDaprdOptions(1, 
    daprd.WithNamespace("other"),
    daprd.WithAppID("app2"), 
    daprd.WithPort(3001)
)
```
Keeps the base dapr options needed to make wf run like the statestore but also takes in N other options for that specific dapr instance

tests/integration/suite/daprd/workflow/crossapp/orderingpreserved.go
cicoyle: removed

pkg/actors/targets/workflow/orchestrator/run.go
cicoyle: It is necessary, upon starting the workflow, to pass along the source appID. This informs durabletask-go about the orchestration's originating appID, ensuring that the appID is propagated properly throughout the durabletask-go code. This is essential for correct routing and context in cross-app workflow scenarios.

tests/integration/suite/daprd/workflow/crossapp/restart.go
cicoyle: I rewrote this, so resolving

tests/integration/suite/daprd/workflow/crossapp/parallel.go
cicoyle: removed

tests/integration/suite/daprd/workflow/crossapp/invalidappid.go
cicoyle: I can remove this if preferred. Felt like it enhanced the test tho

tests/integration/suite/daprd/workflow/crossapp/orderingpreserved.go
cicoyle: removed

pkg/actors/targets/workflow/activity/execute.go
cicoyle: removed this

pkg/actors/targets/workflow/orchestrator/orchestrator.go
```diff
@@ -80,6 +81,7 @@ type orchestrator struct {
 
 type Options struct {
 	AppID             string
+	Namespace         string
```
JoshVanL: Remove

pkg/actors/targets/workflow/activity/activity.go
```diff
@@ -62,6 +63,7 @@ type activity struct {
 
 type Options struct {
 	AppID              string
+	Namespace          string
```
JoshVanL: Remove

pkg/runtime/wfengine/backends/actors/actors.go
```diff
@@ -72,6 +72,7 @@ type Options struct {
 
 type Actors struct {
 	appID             string
+	namespace         string
```
JoshVanL, line 76: Remove

cicoyle: I cannot. This is 100% needed so that I know of the namespace when building out the `wfActorType = common.GetWorkflowActorType(a.namespace, router.GetSource())` inside `pkg/actors/targets/workflow/activity/execute.go` in the `executeActivity` func under the commented line for `// send completed event to orchestrator wf actor` bc namespace is needed when constructing the `wfActorType`

pkg/actors/targets/workflow/activity/activity.go
cicoyle: same reason as below

pkg/actors/targets/workflow/orchestrator/orchestrator.go
cicoyle: I also cannot remove this. This is 100% needed so that I know of the namespace when building out the `activityActorType = common.GetActivityActorType(o.namespace, router.GetTarget())` inside `pkg/actors/targets/workflow/orchestrator/activity.go` & `pkg/actors/targets/workflow/orchestrator/reminder.go` in the `callActivity` & `createReminder` func bc namespace is needed when constructing the ActivityActorType

pkg/actors/targets/workflow/common/common.go
```diff
@@ -0,0 +1,22 @@
...
+
+package common
+
+func GetWorkflowActorType(ns, appID string) string {
+	return "dapr.internal." + ns + "." + appID + ".workflow"
+}
+
+func GetActivityActorType(ns, appID string) string {
+	return "dapr.internal." + ns + "." + appID + ".activity"
+}
```
JoshVanL: Use a struct to hold the namespace, as this doesn't change.

```suggestion

type ActorTypeBuilder struct {
  ns string
}

func NewActorTypeBuilder(namespace string) *ActorTypeBuilder{
  return &ActorTypeBuilder{
    ns: namespace,
  },
}

func (a *ActorTypeBuilder) Workflow(appID string) string {
	return "dapr.internal." + a.ns + "." + appID + ".workflow"
}

func (a *ActorTypeBuilder) Activity (appID string) string {
	return "dapr.internal." + a.ns + "." + appID + ".activity"
}
```

cicoyle: Sure - However, I would like to note, this was the commented suggestion from you/Albert. I wish we caught this new suggestion when the change you are requesting to change again was being suggested originally to save a review loop + dev cycle loop.

pkg/actors/targets/workflow/activity/execute.go
```diff
@@ -109,9 +110,16 @@ func (a *activity) executeActivity(ctx context.Context, name string, taskEvent *
 		return todo.RunCompletedTrue, err
 	}
 
+	// send completed event to orchestrator wf actor
+	wfActorType := a.workflowActorType
+	if router := taskEvent.GetRouter(); router != nil {
+		builder := common.NewActorTypeBuilder(a.namespace)
```
JoshVanL: Builder should be passed in to `Factory` in `Options`, rather than being created on every execute.

cicoyle: Fair point - thanks. Now, I create it once now in the `actors.go` and then assign to both orchestration and activity options in the `RegisterActors` func in the `actors.go` file where we create the options from the actor backend struct which has the `namespace` on it already. 

With the above, I don't need to update the factory function and opted to do it in the `RegisterActors` func bc we have the actors backend `abe` var where `namespace` is and is used to build the options with. Let me know if you would still rather it be in the `Factory` func, but I opted to do it here bc the `abe` struct with namespace is here and that way the orchestration doesnt need both `namespace` and `builder` in its options.

PR #8838, opened 2025-06-27
Reduce contention on placement by removing locks

lrascao:
# Description

Reducing placement lock contention by removing (seemingly) unnecessary locks.


JoshVanL:
These locks are entirely necessary for turn based actors to be correct.

PR #8842, opened 2025-06-30
Actors: reduce contention of mutex locks

JoshVanL:
Updates actor router, table and targets to reduce lock contention on hight throughput actor scenarios.

Move the reentrancy locker to only be used inside the app actor target. Do no use a long running go routine managing a queue, and instead use the same thread for managing waiting for the lock.

Use simple fifo channels for workflow orchestrator and activity locks, which also respect context.

Change the actor table implementation to use `sync.Map` with optimistic atomic stores and atomic loads in order to remove outer locks on access to the table.

This change should significantly reduce contention on the actor router, table and targets, allowing for higher throughput.

cc @lrascao @jmprusi


cicoyle (approved):
lgtm

PR #8846, opened 2025-07-02
Scheduler: reduce stream handle contention

JoshVanL:
Reduce lock contention in the scheduler loop stream trigger handler by using a sync.Map.

cc @jmprusi


cicoyle:
I believe the unit tests are related to this change. Can you take a look?

PR #8848, opened 2025-07-03
Scheduler: Add multiple configurable watch job streams per Scheduler

JoshVanL:
Adds a new CLI flag `scheduler-job-streams`, defaulting to 3, which configures the number of watch job streams daprd will use per scheduler service. Increasing this number can increase the throughput of job processing in high load scenarios.


yaron2:
What's the downside of defaulting to a higher number?

JoshVanL:
@yaron2 higher memory usage and number of open sockets on Scheduler.

In more dramatic typologies where there are 1000s of daprd to a scheduler cluster, one would want to reduce this value to 1. On the other scale, if there are limited daprds but huge throughput, one may consider increasing this value.

cicoyle (dismissed):

cicoyle (approved):

PR #8850, opened 2025-07-07
Actors: reduce lock contention

JoshVanL:
Update actors to remove all locking to the shared table. Instead, each
actor type is locked by the corresponding actor target factory. This is
advantageous, as different actor targets have different lock
requirements given the amount of state the cache/hold.

It is now the case that getting/creating actors are no longer done
inside the table, removing the need for locking at the global level.

Each actor type uses a sync.Map to hold their own actor table, using
optimistic loads and stores.

Only the app target is deactivated via an Idle timeout.


cicoyle:
@JoshVanL can you fix the conflicts?

cicoyle (dismissed):

cicoyle (approved):

PR #8851, opened 2025-07-07
feat(conversation api): add streaming + tool call support

sicoyle:
# Description

this pr is a duplicate of https://github.com/dapr/dapr/pull/8790 I have to open another pr as i don't have perms in filinto's fork and he is on vacation so i have to work in a diff pr unfortunately

To make things easier picking this task up I had to go through and understand things. Posting in case this is helpful to others:
image
Key:
- Blue is what i need to look at next
- pink is tool related
- red is a rename i want to make
- yellow is memo to self
- black/grey is the flow

## Issue reference

## Checklist

- [ ] Code compiles correctly
- [ ] Created/updated tests
- [ ] Unit tests passing


sicoyle:
fixing build now

Makefile
```diff
@@ -541,6 +541,30 @@ check-proto-diff:
...
+	buf lint .
+
+################################################################################
+# Target: buf-format                                                           #
+################################################################################
+.PHONY: buf-format
+buf-format: ## Format protobuf files using buf
+	@command -v buf >/dev/null 2>&1 || { echo "buf is not installed. Install it from https://docs.buf.build/installation"; exit 1; }
+	buf format -w .
+
```
JoshVanL: Remove. Not relevant to PR.

buf.yaml
```diff
@@ -0,0 +1,8 @@
+version: v1
```
JoshVanL: Remove file

dapr/proto/runtime/v1/dapr.proto
```diff
@@ -1329,32 +1332,145 @@ message ConversationRequest {
...
+
+  // Tools available for the LLM to call
+  repeated Tool tools = 8;
 }
 
 message ConversationInput {
-  // The content to send to the llm
-  string content = 1;
+  // DEPRECATED: Use parts instead for new implementations
+  string content = 1 [deprecated = true];
```
JoshVanL: Please create another version of this message. It is wrong for us to have deprecated fields in RPC which are not yet v1.

dapr/proto/runtime/v1/dapr.proto
```diff
@@ -1329,32 +1332,145 @@ message ConversationRequest {
...
 
   // The role to set for the message
   optional string role = 2;
 
   // Scrub PII data that goes into the LLM
   optional bool scrubPII = 3;
+
+  // Content parts for rich content within each agent's input.
+  // In multi-turn conversations the output becomes part of the next request input
+  repeated ContentPart parts = 4;
```
JoshVanL:
```suggestion
  // Content parts for rich content within each agent's input.
  // In multi-turn conversations the output becomes part of the next request input
  repeated Content content = 4;
```

dapr/proto/runtime/v1/dapr.proto
```diff
@@ -1329,32 +1332,145 @@ message ConversationRequest {
...
+}
+
+// Content part supporting text and tool calling (rich media out of scope)
+message ContentPart {
```
JoshVanL: Namespace message names with the associated API.
```suggestion
message ConversationContent {
```

dapr/proto/runtime/v1/dapr.proto
```diff
@@ -1329,32 +1332,145 @@ message ConversationRequest {
...
+  oneof content_type {
+    TextContent text = 1;
+    ToolCallContent tool_call = 2;
+    ToolResultContent tool_result = 3;
```
JoshVanL: Namespace message names.

dapr/proto/runtime/v1/dapr.proto
```diff
@@ -1329,32 +1332,145 @@ message ConversationRequest {
 
   // Temperature for the LLM to optimize for creativity or predictability
   optional double temperature = 7;
+
+  // Tools available for the LLM to call
+  repeated Tool tools = 8;
```
JoshVanL: Namespace message names with API.

dapr/proto/runtime/v1/dapr.proto
JoshVanL: Why `ToolResultContent` (_result_) a type on the request?

dapr/proto/runtime/v1/dapr.proto
```diff
@@ -1329,32 +1332,145 @@ message ConversationRequest {
...
+  string arguments = 4;               // Function arguments as JSON string
+}
+
+// Tool result as content part
+message ToolResultContent {
+  string tool_call_id = 1;
+  string name = 2;
+  string content = 3;                 // Tool result as text
+  optional bool is_error = 4;         // Indicates tool execution error
 }
```
JoshVanL: Namespace message names with API.

dapr/proto/runtime/v1/dapr.proto
```diff
@@ -1329,32 +1332,145 @@ message ConversationRequest {
...
+message TextContent {
+  string text = 1;
+}
+
+// Tool call as content part
+message ToolCallContent {
+  string id = 1;
+  string type = 2;                    // "function"
+  string name = 3;
+  string arguments = 4;               // Function arguments as JSON string
```
JoshVanL: Are non of these fields optional?

JoshVanL: Align comments.

dapr/proto/runtime/v1/dapr.proto
```diff
@@ -1329,32 +1332,145 @@ message ConversationRequest {
...
+
+// Simple text content
+message TextContent {
+  string text = 1;
+}
+
+// Tool call as content part
+message ToolCallContent {
+  string id = 1;
+  string type = 2;                    // "function"
```
JoshVanL: Why not call this field `function`?

dapr/proto/runtime/v1/dapr.proto
JoshVanL: JSON strings should have type `google.protobuf.StringValue`.

Why is this a JSON string, instead of a types message?

dapr/proto/runtime/v1/dapr.proto
```diff
@@ -1329,32 +1332,145 @@ message ConversationRequest {
...
+  string name = 3;
+  string arguments = 4;               // Function arguments as JSON string
+}
+
+// Tool result as content part
+message ToolResultContent {
+  string tool_call_id = 1;
+  string name = 2;
+  string content = 3;                 // Tool result as text
+  optional bool is_error = 4;         // Indicates tool execution error
```
JoshVanL: Why is this not part of the RPC error?

Should this message not be a `oneof` with one type being the error type? We should have error context (string).

dapr/proto/runtime/v1/dapr.proto
```diff
@@ -1329,32 +1332,145 @@ message ConversationRequest {
...
+  string type = 2;                    // "function"
+  string name = 3;
+  string arguments = 4;               // Function arguments as JSON string
+}
+
+// Tool result as content part
+message ToolResultContent {
+  string tool_call_id = 1;
+  string name = 2;
+  string content = 3;                 // Tool result as text
```
JoshVanL: All tools return strings (?).

dapr/proto/runtime/v1/dapr.proto
```diff
@@ -1329,32 +1332,145 @@ message ConversationRequest {
...
+message ToolCallContent {
+  string id = 1;
+  string type = 2;                    // "function"
+  string name = 3;
+  string arguments = 4;               // Function arguments as JSON string
+}
+
+// Tool result as content part
+message ToolResultContent {
+  string tool_call_id = 1;
```
JoshVanL: At least remove the string `tool`, this is given through context via the message type.

```suggestion
  string id = 1;
```

dapr/proto/runtime/v1/dapr.proto
```diff
@@ -1329,32 +1332,145 @@ message ConversationRequest {
...
+}
+
+// Simple text content
+message TextContent {
+  string text = 1;
+}
+
+// Tool call as content part
+message ToolCallContent {
+  string id = 1;
```
JoshVanL, line 1371: What is the different between `id` and `name`?

dapr/proto/runtime/v1/dapr.proto
```diff
@@ -1329,32 +1332,145 @@ message ConversationRequest {
...
+  optional bool is_error = 4;         // Indicates tool execution error
 }
 
-// ConversationResult is the result for one input.
+// ConversationResult represents a single output from the assistant or tool in response to a conversation input.
 message ConversationResult {
-  // Result for the one conversation input.
-  string result = 1;
+  // DEPRECATED: Use parts instead for new implementations
+  string result = 1 [deprecated = true];
```
JoshVanL: Again, we shouldn't be deprecating message types on pre-v1 RPCs.

dapr/proto/runtime/v1/dapr.proto
```diff
@@ -1329,32 +1332,145 @@ message ConversationRequest {
...
+
+  // Usage statistics if available
+  optional ConversationUsage usage = 3;
+}
+
+// ConversationStreamResponse is the streaming response for Conversation.
+message ConversationStreamResponse {
+  oneof response_type {
+    ConversationStreamChunk chunk = 1;
+    ConversationStreamComplete complete = 2;
```
JoshVanL:
```suggestion
    ConversationStreamEOF complete = 2;
```

dapr/proto/runtime/v1/dapr.proto
```diff
@@ -1329,32 +1332,145 @@ message ConversationRequest {
...
+  }
+}
+
+// ConversationStreamChunk represents a streaming content chunk.
+message ConversationStreamChunk {
+  // Reason why streaming stopped for this chunk (e.g., "stop", "tool_calls")
+  optional string finish_reason = 1;
```
JoshVanL, line 1422: Why is this not part of the `EOF` type? Why not have another `oneof` type?

dapr/proto/runtime/v1/dapr.proto
```diff
@@ -1329,32 +1332,145 @@ message ConversationRequest {
...
+
+  // Content parts in streaming
+  repeated ContentPart parts = 2;
+
+  // Chunk metadata
+  optional int32 chunk_index = 3;
+  optional bool is_delta = 4;          // True if incremental
```
JoshVanL: Align comments.

JoshVanL: This doesn't mean anything reading the API spec.

dapr/proto/runtime/v1/dapr.proto
```diff
@@ -1329,32 +1332,145 @@ message ConversationRequest {
...
+// ConversationStreamChunk represents a streaming content chunk.
+message ConversationStreamChunk {
+  // Reason why streaming stopped for this chunk (e.g., "stop", "tool_calls")
+  optional string finish_reason = 1;
+
+  // Content parts in streaming
+  repeated ContentPart parts = 2;
+
+  // Chunk metadata
+  optional int32 chunk_index = 3;
```
JoshVanL, line 1428: Redundant.
```suggestion
```

dapr/proto/runtime/v1/dapr.proto
```diff
@@ -1329,32 +1332,145 @@ message ConversationRequest {
...
+  optional bool is_delta = 4;          // True if incremental
+}
+
+// ConversationStreamComplete indicates the streaming conversation has completed.
+message ConversationStreamComplete {
+  // Final context ID
+  optional string contextID = 1;
```
JoshVanL: Don't mix camel and snake case.

What is a "context ID", and why is it optional?

```suggestion
  optional string context_id = 1;
````

File: dapr/proto/runtime/v1/dapr.proto [hunk 21]
```diff
@@ -1329,32 +1332,145 @@ message ConversationRequest {
...
+  // Usage statistics if available
+  optional ConversationUsage usage = 2;
+  // Outputs accumulated outputs/tool calls from the streaming
+  repeated ConversationResult outputs = 3;
```
JoshVanL: Result vs output - I think we should be using the same word here.

dapr/proto/runtime/v1/dapr.proto
```diff
@@ -1329,32 +1332,145 @@ message ConversationRequest {
...
+
+// ConversationUsage represents token usage statistics.
+message ConversationUsage {
+  // Number of tokens in the prompt
+  optional uint32 prompt_tokens = 1 [json_name = "promptTokens"];
+  // Number of tokens in the completion
+  optional uint32 completion_tokens = 2 [json_name = "completionTokens"];
+  // Total number of tokens used
+  optional uint32 total_tokens = 3 [json_name = "totalTokens"];
+}
```
JoshVanL: What's the purpose of this?

dapr/proto/runtime/v1/dapr.proto
```diff
@@ -1329,32 +1332,145 @@ message ConversationRequest {
...
+
+// Tool represents a function that can be called by the LLM (used on the request to the LLM)
+message Tool {
```
JoshVanL: Namespace message name.

dapr/proto/runtime/v1/dapr.proto
```diff
@@ -1329,32 +1332,145 @@ message ConversationRequest {
...
+  // The type of tool (e.g., "function", "web_search", etc.)
+  string type = 1;
+  // The name of the function
+  string name = 2;
+  // Description of what the function does
+  string description = 3;
+  // JSON schema for the function parameters as a string
+  string parameters = 4;
```
JoshVanL: Use `google.protobuf.StringValue` type for JSON strings.

dapr/proto/runtime/v1/dapr.proto
```diff
@@ -1329,32 +1332,145 @@ message ConversationRequest {
...
+}
+
+// ToolCall represents a function call requested by the LLM (used on the response from the LLM)
+message ToolCall {
```
JoshVanL: Namespace message names.

dapr/proto/runtime/v1/dapr.proto
```diff
@@ -1329,32 +1332,145 @@ message ConversationRequest {
...
+  // Unique identifier for this tool call
+  string id = 1;
+  // The type of tool call (e.g., "function")
+  string type = 2;
+  // Name of the function to call
+  string name = 3;
```
JoshVanL: This should be `function_name`?

dapr/proto/runtime/v1/dapr.proto
```diff
@@ -1329,32 +1332,145 @@ message ConversationRequest {
...
+  // Function arguments as a JSON string
+  string arguments = 4;
```
JoshVanL: Use `google.protobuf.StringValue` type for JSON strings.

pkg/api/universal/conversation.go
```diff
@@ -42,69 +43,63 @@ func (a *Universal) ConverseAlpha1(ctx context.Context, req *runtimev1pb.Convers
...
-	diag.DefaultComponentMonitoring.ConversationInvoked(ctx, req.GetName(), err == nil, elapsed)
+	usage := conversation.ConvertUsageToProto(resp)
+
+	// Safe usage extraction for monitoring
+	var promptTokens, completionTokens int64
+	if usage != nil {
+		promptTokens = int64(usage.GetPromptTokens())         //nolint:gosec // Intentional use of int64 for backward compatibility
+		completionTokens = int64(usage.GetCompletionTokens()) //nolint:gosec // Intentional use of int64 for backward compatibility
+	}
+	diag.DefaultComponentMonitoring.ConversationInvoked(ctx, req.GetName(), err == nil, elapsed, diag.NonStreamingConversation, promptTokens, completionTokens)
```
JoshVanL: Why define vars? Move `iag.DefaultComponentMonitoring` into `if` block above.

pkg/api/universal/conversation.go
```diff
@@ -115,26 +110,89 @@ func (a *Universal) ConverseAlpha1(ctx context.Context, req *runtimev1pb.Convers
...
-				res = scrubbed[0]
-			}
+// ConverseStreamAlpha1 implements the streaming conversation API
+func (a *Universal) ConverseStreamAlpha1(req *runtimev1pb.ConversationRequest, stream runtimev1pb.Dapr_ConverseStreamAlpha1Server) error {
+	// Validate component exists
+	if a.compStore.ConversationsLen() == 0 {
+		err := messages.ErrConversationNotFound
+		a.logger.Debug(err)
+		return err
+	}
```
JoshVanL: Redundant

```suggestion
```

pkg/api/universal/conversation.go
```diff
@@ -115,26 +110,89 @@ func (a *Universal) ConverseAlpha1(ctx context.Context, req *runtimev1pb.Convers
...
+			promptTokens = int64(usage.GetPromptTokens())         //nolint:gosec // Intentional use of int64 for backward compatibility
+			completionTokens = int64(usage.GetCompletionTokens()) //nolint:gosec // Intentional use of int64 for backward compatibility
+		}
+		diag.DefaultComponentMonitoring.ConversationInvoked(ctx, req.GetName(), retErr == nil, elapsed, diag.StreamingConversation, promptTokens, completionTokens)
+	}()
+
+	// we use non-streaming scrubber on the non-streamingoutput
+	// we get usage to use in monitoring
+	usage, retErr = pipelineImpl.ProcessStream(ctx, stream, component, componentRequest, req.GetName(), outputScrubber)
+	return retErr
```
JoshVanL: Please move the sending process into a new func. Assigning vars (`usage`) which are re-referenced in a `defer` mid function is confusing to read.

pkg/conversation/consts.go
```diff
@@ -0,0 +1,25 @@
...
+limitations under the License.
+*/
+
+package conversation
+
+const (
+	// SimulatedStreamingChunkSize is the default chunk size (bytes) for simulated streaming conversation data where
+	// the entire response is sent in chunks to simulate a streaming response and allow clients to process data as it arrives
+	// even if the model does not support streaming natively.
+	SimulatedStreamingChunkSize = 1 << 12 // 4 KiB
```
JoshVanL, line 20: Why are we choosing this arbitrary number, and why are we calling this "simulated"?

pkg/conversation/consts.go
```diff
@@ -0,0 +1,25 @@
...
+
+	// PIIScrubberStreamingWindowSize is the default window size (bytes) for PII scrubbing while streaming conversation data.
+	// this allows for a look-ahead window of stream data to be scrubbed
+	PIIScrubberStreamingWindowSize = 128 // 128 bytes compromise between performance, memory usage and latency.
```
JoshVanL, line 24: `// 128 bytes compromise between performance, memory usage and latency.`

How was this determined?

pkg/conversation/conversation.go
```diff
@@ -0,0 +1,356 @@
...
+*/
+
+package conversation
+
+import (
+	piiscrubber "github.com/aavaz-ai/pii-scrubber"
+	contribConverse "github.com/dapr/components-contrib/conversation"
+	"github.com/dapr/dapr/pkg/messages"
+	runtimev1pb "github.com/dapr/dapr/pkg/proto/runtime/v1"
+	"github.com/dapr/kit/ptr"
```
JoshVanL: Organize imports.

pkg/conversation/conversation.go
```diff
@@ -0,0 +1,356 @@
...
+	if resp == nil || resp.Usage == nil {
+		return nil
+	}
+
+	// Safely convert uint64 to uint32 for token counts
+	// Token counts realistically never exceed uint32 range (~4.3 billion)
+	return &runtimev1pb.ConversationUsage{
+		PromptTokens:     ptr.Of(uint32(resp.Usage.PromptTokens)),
+		CompletionTokens: ptr.Of(uint32(resp.Usage.CompletionTokens)),
+		TotalTokens:      ptr.Of(uint32(resp.Usage.TotalTokens)),
```
JoshVanL: Why do we have a discrepancy between typing?

pkg/conversation/conversation.go
```diff
@@ -0,0 +1,356 @@
...
+
+// NeedsInputScrubber checks if PII scrubbing is needed for any input
+func NeedsInputScrubber(req *runtimev1pb.ConversationRequest) bool {
+	for _, input := range req.GetInputs() {
+		if input.GetScrubPII() {
+			return true
+		}
+	}
+	return false
+}
```
JoshVanL, line 68: Surely we should only be doing the scrubbing on inputs where it is enabled? 
This is returning true for all (?).

pkg/conversation/conversation.go
```diff
@@ -0,0 +1,356 @@
...
+
+// NeedsOutputScrubber checks if PII scrubbing is needed for outputs
+func NeedsOutputScrubber(req *runtimev1pb.ConversationRequest) bool {
+	return req.GetScrubPII()
+}
```
JoshVanL, line 73: Redundant
```suggestion
```

pkg/conversation/conversation.go
```diff
@@ -0,0 +1,356 @@
...
+
+// extractToolNameFromParts extracts tool name from tool result parts
+func extractToolNameFromParts(parts []*runtimev1pb.ContentPart) string {
+	for _, part := range parts {
+		if toolResult := part.GetToolResult(); toolResult != nil {
+			return toolResult.GetName()
+		}
+	}
+	return ""
+}
```
JoshVanL: Why do we not error if it is not found? Why is only one "`part`" have a tool name?

pkg/conversation/conversation.go
```diff
@@ -0,0 +1,356 @@
...
+
+// extractToolCallIDFromParts extracts tool call ID from tool result parts
+func extractToolCallIDFromParts(parts []*runtimev1pb.ContentPart) string {
+	for _, part := range parts {
+		if toolResult := part.GetToolResult(); toolResult != nil {
+			return toolResult.GetToolCallId()
+		}
+	}
+	return ""
+}
```
JoshVanL: Same here- why no error, and why do we only return _one_ tool id?

pkg/conversation/conversation.go
```diff
@@ -0,0 +1,356 @@
...
+					Text: &runtimev1pb.TextContent{
+						Text: result,
+					},
+				},
+			},
+		}
+	}
+
+	return nil
+}
```
JoshVanL: What is this doing?

pkg/conversation/conversation.go
```diff
@@ -0,0 +1,356 @@
...
+			parts = append(parts, contribConverse.TextContentPart{Text: msg})
+		}
+
+		// Create conversation input with content parts
+		inputs = append(
+			inputs,
+			contribConverse.ConversationInput{
+				Message: msg, // Keep for backward compatibility. TODO: remove this field
+				Role:    role,
+				Parts:   parts,
```
JoshVanL: Please create new message types.

pkg/conversation/conversation.go
```diff
@@ -0,0 +1,356 @@
...
+			for _, part := range i.GetParts() {
+				// Extract text content from parts
+				msg = part.GetText().GetText()
+
+				// Extract tool call ID and name from parts for role detection
+				toolCallID := extractToolCallIDFromParts([]*runtimev1pb.ContentPart{part})
+				toolName := extractToolNameFromParts([]*runtimev1pb.ContentPart{part})
+
+				// Handle tool result messages - determine role based on tool call ID and name
+				if toolCallID != "" && toolName != "" && (role == contribConverse.RoleUser || role == "") {
```
JoshVanL: These fields should be marked as optional (?).

pkg/conversation/middlewares.go
```diff
@@ -0,0 +1,211 @@
...
+## SCRUBBER LIBRARY BUG DISCOVERED
+
+During testing, we discovered a significant bug in the github.com/aavaz-ai/pii-scrubber library:
+
+**BUG**: When IP addresses and credit card numbers appear in the same text, the library
+fails to detect credit card numbers. This affects ALL usage of the library, not just streaming.
+
+**EVIDENCE**:
+- ✅ "Card number 4263982640269299" → "Card number <CREDIT_CARD>" (WORKS)
+- ❌ "IP is 192.168.1.100 and card 4263982640269299" → "IP is <IP> and card 4263982640269299" (FAILS)
```
JoshVanL, line 46: Please do not use emoji art in code comments.

pkg/conversation/middlewares.go
```diff
@@ -0,0 +1,211 @@
...
+
+**WORKAROUNDS IMPLEMENTED**:
```
JoshVanL, line 48: Why is this a "middleware"?

pkg/conversation/middlewares.go
```diff
@@ -0,0 +1,211 @@
...
+
+			// If no content after delimiter, it's a good word boundary
+			if !hasContentAfter {
+				return i + 1
+			}
+		}
+	}
+
+	return -1
+}
```
JoshVanL, line 115: What is the intention of this doing?

pkg/conversation/scrub_streaming_test.go
```diff
@@ -0,0 +1,1153 @@
+/*
```
JoshVanL, line 1: Do not use underscores in file names (except with `_test.go`).

pkg/conversation/scrub_streaming_test.go
```diff
@@ -0,0 +1,1153 @@
...
+
+During testing, we discovered a critical bug in the github.com/aavaz-ai/pii-scrubber library:
+
+**BUG**: When IP addresses and credit card numbers appear in the same text, the library
+fails to detect credit card numbers. This affects ALL usage of the library.
+
+**EVIDENCE**:
+- ✅ "Card number 4263982640269299" → "Card number <CREDIT_CARD>" (WORKS)
+- ❌ "IP is 192.168.1.100 and card 4263982640269299" → "IP is <IP> and card 4263982640269299" (FAILS)
+
```
JoshVanL, line 30: Please do not use emoji art in code comments- it's distracting.

pkg/conversation/middlewares.go
```diff
@@ -0,0 +1,211 @@
...
+		return nil, nil
+	}
+
+	// Find the last complete word within the window
+	windowContent := s.buffer[:s.windowSize]
+	lastCompletePos := findLastCompleteWordPosition(windowContent)
+
+	if lastCompletePos == -1 {
+		// No complete words - prevent infinite buffering
+		if len(s.buffer) > s.windowSize*2 {
```
JoshVanL, line 136: Why are we using flushers, manual buffers etc. over actual `io` streams?

pkg/conversation/scrub_streaming_test.go
```diff
@@ -0,0 +1,1153 @@
...
+					} else {
+						t.Logf("   Credit card: ✅ SCRUBBED")
+					}
+
+					if ipMissed {
+						t.Logf("   IP address:  ❌ MISSED (192.168.1.100 still visible)")
+					} else {
+						t.Logf("   IP address:  ✅ SCRUBBED")
+					}
+
```
JoshVanL, line 328: Remove all `t.Log` from tests.

pkg/conversation/scrub_streaming_test.go
```diff
@@ -0,0 +1,1153 @@
...
+					if cardMissed && !ipMissed {
+						t.Logf("🔍 PATTERN: IP scrubbed, card missed (same as IP-first case)")
+					} else if !cardMissed && ipMissed {
+						t.Logf("🔍 PATTERN: Card scrubbed, IP missed (opposite of IP-first case)")
+					} else if cardMissed && ipMissed {
+						t.Logf("🔍 PATTERN: Both missed (mutual interference)")
+					} else {
+						t.Logf("🔍 PATTERN: Both scrubbed (order matters - this works!)")
+					}
```
JoshVanL, line 337: Is this unit test actually checking anything?

It seems clear that this file has been written with AI- but it doesn't look like the AI is actually asserting any test cases???

pkg/conversation/scrub_streaming_test.go
```diff
@@ -0,0 +1,1153 @@
...
+
+		// Test flush
+		remaining, err := pipeline.flushMiddleware()
+		require.NoError(t, err)
+
+		t.Logf("Flushed successfully: %v", remaining != nil)
+
+		t.Logf("✅ SUCCESS: Pipeline properly handles error returns from middleware")
+	})
+}
```
JoshVanL, line 1153: It's impossible to review this file.

pkg/conversation/streaming.go
```diff
@@ -0,0 +1,352 @@
...
+// StreamingPipelineImpl handles the streaming conversation pipeline
+type StreamingPipelineImpl struct {
+	Middleware         []StreamingMiddleware
+	logger             logger.Logger
+	simulatedChunkSize int
+}
+
+// StreamingMiddleware interface for processing streaming chunks with error handling
+type StreamingMiddleware interface {
+	ProcessChunk(chunk []byte) ([]byte, error)
```
JoshVanL, line 36: Use proper Go `io` streams when processing streams of data.

pkg/proto/runtime/v1/dapr_tracing.go
```diff
@@ -88,123 +88,163 @@ func (*DeleteJobRequest) AppendSpanAttributes(rpcMethod string, m map[string]str
 func (*DecryptRequest) AppendSpanAttributes(rpcMethod string, m map[string]string) {
```
JoshVanL, line 88: Please revert this file.

pkg/diagnostics/component_monitoring.go
```diff
@@ -31,6 +31,13 @@ const (
 	CryptoOp                 = "crypto_op"
 )
 
+type ConversationType string
+
+const (
+	NonStreamingConversation ConversationType = "non_streaming"
+	StreamingConversation    ConversationType = "streaming"
```
JoshVanL, line 38: Please use `unary` or some other verbiage to describe this mode, rather than the negation of the other mode.

pkg/scheduler/server/internal/etcd/etcd.go
```diff
@@ -72,25 +79,53 @@ type etcd struct {
 }
 
 func New(opts Options) (Interface, error) {
-	config, err := config(opts)
+	if opts.Embed {
```
JoshVanL: Changes not related to PR.

tests/integration/framework/process/scheduler/etcd/etcd.go
```diff
@@ -0,0 +1,131 @@
+/*
```
JoshVanL: Stopping review now as there seems to be some broken history or cherry pick causing the diff to become wrong.

JoshVanL (changes requested):
First pass from me

tests/integration/framework/process/scheduler/etcd/etcd.go
sicoyle: i rebased onto main and fixed this

pkg/scheduler/server/internal/etcd/etcd.go
sicoyle: i rebased onto main and fixed this

dapr/proto/runtime/v1/dapr.proto
sicoyle: This is required for mult-turn conversations that involve tool calls. The client sends a ConversationRequest with potentially a tool call. The ConversationResponse would then include a tool that must be called. After making the tool call the client sends a ConversationRequest that now includes the tool_results. 

I could move this out of the oneof to give us a better separation of conversational content from tool execution results. Thoughts?

dapr/proto/runtime/v1/dapr.proto
sicoyle: tools could even be an agent. The most basic tool call type could be a function. I've updated this to denote it's just an example.

dapr/proto/runtime/v1/dapr.proto
sicoyle: good call out. If a tool call is a small part to a larger conversation, and the overall RPC is successful then it is fine to encode tool errors within the message, not as an RPC err. RPC errs should be reserved for transport/protocol level failures across the entire conversation request failing not just a single tool call that is part of the greater conversation. I'll update this to be more like this: (with some minor field name tweaks)
```
message ConversationToolResult {
  string tool_call_id = 1;
  string name = 2;
  string content = 3;
  oneof result {
    string output_text = 3;
    ToolError error = 4;
  }
}

message ToolError {
  string message = 1;
  int32 code = 2;
}
```

JoshVanL: Please can you tell me why an error from a tool is not an RPC error, and is instead part of the message?

dapr/proto/runtime/v1/dapr.proto
sicoyle: If i look at the contrib PR i see an ex payload that has
```
...
{
      "role": "tool",
      "parts": [
        {
          "type": "tool_result",
          "tool_call_id": "toolu_01A5b3c4D5e6F7g8H9i0J1k2",
          "name": "get_weather",
          "content": "Sunny, 72°F in Boston"
        }
      ]
    },
...
```

so my understanding is that each tool call gets a unique instance id and then per tool they will have a name. I can think how to make this clearer

JoshVanL: Why is the ID part of the protos?

dapr/proto/runtime/v1/dapr.proto
sicoyle: Tools don't have to just be a function, so leaving this. But I can think on how to make this cleaner 🤷‍♀️

sicoyle:
The comments above on this PR are from me cherry picking changes in from the existing Conversation API efforts.

I pushed where I'm at so far with addressing feedback and doing cleanup, renaming, etc. Check out my PR description image for a better understanding if you need - it helped me a ton doing that with following the names and whatnot. Also pushed on contrib corresponding changes. That PR still needs quite a bit of love too tbh. I put comments in code here with `TODO(<at my github tag>)` that could use some input @JoshVanL so far, and I will work on addressing the rest of your existing comments (the ones I left unresolved) tomorrow.

dapr/proto/runtime/v1/dapr.proto
sicoyle: hopefully my tldraw I added in the PR desc helps clear this one up :)

dapr/proto/runtime/v1/dapr.proto
sicoyle: This was an add in to bubble up token usage with the llm provider api 🤷‍♀️

pkg/conversation/consts.go
sicoyle: no idea tbh

pkg/conversation/conversation.go
sicoyle: i just deleted this

pkg/conversation/conversation.go
sicoyle: no idea

pkg/conversation/scrub_streaming_test.go
sicoyle: and `_mock` based on the rest of the repo :)

pkg/conversation/middlewares.go
sicoyle: after looking into the existing code on this, it appears that the sdk chosen does not support streaming I/O (so that's why we have manual buffering) and also honestly we need to get rid of this pkg imo bc it has a bug according to the docs in this file where it fails to detect credit card numbers when IP addresses are present in the same text 💀

sicoyle:
closing to do tool calling only for now

Issue #8855, opened 2025-07-08
Workflow start time is ignored


frankbuckley:
Using Dapr 1.15.6, running locally on Windows, starting a workflow with a future time results in the workflow running immediately.

## Expected Behavior

Workflow starts at requested future time.

## Actual Behavior

Workflow starts immediately.

## Steps to Reproduce the Problem

I first observed the problem using the .NET SDK, but it also replicates using the Go SDK.

Using [Go SDK example](https://github.com/dapr/go-sdk/tree/main/examples/workflow), amend `ScheduleNewWorkflow` request as follows:

```go
start := time.Now().Add(24 * time.Hour)
log.Printf("********* Starting workflow at %s", start.Format(time.RFC3339))

// Start workflow test
instanceID, err := wfClient.ScheduleNewWorkflow(ctx, "TestWorkflow", workflow.WithInstanceID(uuid.NewString()), workflow.WithInput(1), workflow.WithStartTime(start))
```

Run sample.

Logs:

```
You're up and running! Both Dapr and your app logs will appear here.

== APP == dapr client initializing for: 127.0.0.1:50001
== APP == client uses API token
== APP == Worker initialized
== APP == TestWorkflow registered
== APP == TestActivity registered
== APP == FailActivity registered
== APP == INFO: 2025/07/08 14:52:39 connecting work item listener stream
== APP == 2025/07/08 14:52:39 work item listener started
== APP == INFO: 2025/07/08 14:52:39 starting background processor
== APP == 2025/07/08 14:52:39 ********* Starting workflow at 2025-07-09T14:52:39+01:00
== APP == runner started
time="2025-07-08T14:52:39.89005+01:00" level=info msg="work item stream established by user-agent: [dapr-sdk-go/v1.12.0 grpc-go/1.70.0]" app_id=workflow instance=UNDWKS02 scope=dapr.runtime.wfengine type=log ver=1.15.6
time="2025-07-08T14:52:39.890563+01:00" level=debug msg="Registering workflow actors" app_id=workflow instance=UNDWKS02 scope=dapr.runtime.wfengine type=log ver=1.15.6
time="2025-07-08T14:52:39.890563+01:00" level=info msg="Updating actor types: [dapr.internal.default.workflow.workflow dapr.internal.default.workflow.activity]" app_id=workflow instance=UNDWKS02 scope=dapr.runtime.actors.placement type=log ver=1.15.6
[...]

PR #8856, opened 2025-07-08
Fix Workflow Start Time

JoshVanL:
Fix Workflow Start Time

Problem

Using the Workflow "Start Time" feature would result in the Workflow starting immediately.

Impact

A workflow could not be scheduled to start at some point in the future.

Root cause

The "Start Time" was not being propagated to the Actor reminder responsible for executing a Workflow instance.

Solution

Correctly propagate the Start Time of a Workflow to the Actor reminder responsible for executing a Workflow instance.

Fixes https://github.com/dapr/dapr/issues/8855


docs/release_notes/v1.15.7.md
```diff
@@ -0,0 +1,23 @@
+# Dapr 1.15.6
```
frankbuckley:
```suggestion
# Dapr 1.15.7
```

yaron2 (approved):

PR #8858, opened 2025-07-08
[1.15] Fix use of DynamoDB as Workflow state store

JoshVanL:
Problem

Using DynamoDB as the Workflow state store would result in an error when trying to run a Workflow.

Impact

It would not be possible to run Workflows using DynamoDB as the state store.

Root cause

The state store implementation used String (S) attributes rather than the Binary (B) attributes resulting in the state not being readable.

Solution

Updated the DynamoDB state store implementation to use Binary (B) attributes for storing Workflow state.

### DIFF
--- a/docs/release_notes/v1.15.7.md
+++ b/docs/release_notes/v1.15.7.md
@@ -3,6 +3,7 @@
 This update includes bug fixes:
 
 - [Fix Workflow Start Time](#fix-workflow-start-time)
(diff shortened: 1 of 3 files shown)


passuied (approved):

cicoyle (approved):

JoshVanL:
@cicoyle done!

cicoyle (approved):

yaron2 (approved):

PR #8861, opened 2025-07-08
Set base Dockerfile USER to 65532:65532

JoshVanL:
Update the base Dockerfile image to set the user and group of all images to run as the non-root user group `65532:65532`.

Adds a security func check to ensure that every service is running as the user group `65532:65532` when running in Kubernetes mode.

Fixes https://github.com/dapr/dapr/issues/8365

### DIFF
--- a/pkg/actors/targets/workflow/activity/activity.go
+++ b/pkg/actors/targets/workflow/activity/activity.go
@@ -26,6 +26,7 @@ import (
 	"github.com/dapr/dapr/pkg/actors/state"
 	"github.com/dapr/dapr/pkg/actors/table"
 	"github.com/dapr/dapr/pkg/actors/targets"
+	"github.com/dapr/dapr/pkg/actors/targets/workflow/common"
 	internalsv1pb "github.com/dapr/dapr/pkg/proto/internals/v1"
 	"github.com/dapr/dapr/pkg/runtime/wfengine/todo"
 	"github.com/dapr/kit/logger"
@@ -47,6 +48,7 @@ type activity struct {
 	actorID           string
 	actorType         string
 	workflowActorType string
+	actorTypeBuilder  *common.ActorTypeBuilder
 
 	table     table.Interface
 	router    router.Interface
@@ -68,6 +70,7 @@ type Options struct {
 	Scheduler          todo.ActivityScheduler
 	Actors             actors.Interface
 	SchedulerReminders bool
+	ActorTypeBuilder   *common.ActorTypeBuilder
 }
 
 func Factory(ctx context.Context, opts Options) (targets.Factory, error) {
@@ -112,6 +115,7 @@ func Factory(ctx context.Context, opts Options) (targets.Factory, error) {
 				reminders:          reminders,
 				scheduler:          opts.Scheduler,
 				schedulerReminders: opts.SchedulerReminders,
+				actorTypeBuilder:   opts.ActorTypeBuilder,
 				lock:               make(chan struct{}, 1),
 			}
 		} else {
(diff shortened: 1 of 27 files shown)


pkg/security/usergroup.go
```diff
@@ -0,0 +1,49 @@
...
+
+import (
+	"fmt"
+	"os"
+	"syscall"
+
+	"github.com/dapr/dapr/pkg/modes"
+)
+
+func checkUIDGID(mode modes.DaprMode) error {
```
nelson-parente: nit: ugly name but I don't think checkUserIDGroupID is any better

pkg/security/usergroup.go
```diff
@@ -0,0 +1,49 @@
...
+	if mode != modes.KubernetesMode {
+		return nil
+	}
+
+	if os.Getenv("DAPR_UNSAFE_SKIP_CONTAINER_UID_CHECK") == "true" {
```
nelson-parente: nit: `DAPR_UNSAFE_SKIP_CONTAINER_UID_GID_CHECK`?

pkg/security/usergroup.go
```diff
@@ -0,0 +1,49 @@
...
+		log.Warn("Skipping container UID/GID check due to env override.")
+		return nil
+	}
+
+	uid := syscall.Geteuid()
+	gid := syscall.Getegid()
+
+	const expuid = 65532
```
nelson-parente: nit: expid since its used for both user id and group id?

tests/integration/suite/actors/reminders/scheduler/remote.go
```diff
@@ -74,7 +74,7 @@ spec:
   - name: SchedulerReminders
     enabled: true`), 0o600))
 
-	r.actorIDsNum = 200
+	r.actorIDsNum = 100
```
nelson-parente, line 77: Just checking if all these changes under tests/integration/... are intentional?

nelson-parente (approved):
LGTM (my comments are all nits)

tests/integration/suite/actors/reminders/scheduler/remote.go
JoshVanL: Stabilizing int tests

tests/integration/suite/sentry/usergroup/disable.go
```diff
@@ -0,0 +1,90 @@
...
+		sentry.WithTrustDomain("integration.test.dapr.io"),
+	)
+
+	return []framework.Option{
+		framework.WithProcesses(d.logline, d.sentry, kubeAPI),
+	}
+}
+
+func (d *disable) Run(t *testing.T, ctx context.Context) {
+	d.logline.EventuallyFoundAll(t)
```
cicoyle, line 89: Do we need to wait for sentry to be running before checking the logs?
```suggestion
	d.sentry.WaitUntilRunning(t, ctx)
	d.logline.EventuallyFoundAll(t)
```

tests/integration/suite/daprd/jobs/streaming/reconnect.go
```diff
@@ -72,7 +72,7 @@ func (r *reconnect) Setup(t *testing.T) []framework.Option {
 	)
 
 	return []framework.Option{
-		framework.WithProcesses(srv, r.daprd, r.scheduler1),
```
cicoyle, line 75: Seems to still be something off with this test as I see its taking longer than the allotted 45s from the IT suite

tests/integration/suite/sentry/usergroup/disable.go
JoshVanL: No it is not needed, this waits until we find the logs.

tests/integration/suite/daprd/jobs/streaming/reconnect.go
JoshVanL: Indeed

PR #8863, opened 2025-07-09
Workflow: don't wait for start when starttime

JoshVanL:
If start time of the workflow is defined, do not wait for the workflow to start.


tests/integration/suite/daprd/workflow/starttime/nodelay.go
```diff
@@ -0,0 +1,71 @@
...
+	n.workflow.Registry().AddOrchestratorN("delay", func(ctx *task.OrchestrationContext) (any, error) {
+		if !ctx.IsReplaying {
+			executed = time.Now()
+		}
+		return nil, nil
+	})
+
+	client := n.workflow.BackendClient(t, ctx)
+
+	id, err := client.ScheduleNewOrchestration(ctx, "delay")
```
nelson-parente: why do we need 2 blocks? shouldn't lines 65..70 be enough? 

same comment for nowait.go

cicoyle (dismissed):

PR #8874, opened 2025-07-14
Retry starting subscriptions

acroca:
# Description

Added a retry mechanism to the subscribers. It will retry to start the subscription until it succeeds, showing an error on every failed attempt.

## Issue reference

In https://github.com/dapr/components-contrib/issues/3715 we get issues where the sidecar doesn't have permissions temporarily for a few seconds until the role assignment happens, which take a few seconds after the pod is created.

## Checklist

- [ ] Code compiles correctly
- [ ] Created/updated tests
- [ ] Unit tests passing


pkg/runtime/processor/subscriber/subscriber.go
```diff
@@ -318,7 +320,13 @@ func (s *Subscriber) StartAppSubscriptions() error {
 	var errs []error
 	for name, ps := range s.compStore.ListPubSubs() {
 		for _, sub := range s.compStore.ListSubscriptionsAppByPubSub(name) {
-			ss, err := s.startSubscription(ps, sub, false)
+			var ss *subscription.Subscription
+			var err error
+			backoff.Retry(func() error {
+				ss, err = s.startSubscription(ps, sub, false)
+				return err
+			}, backoff.WithMaxRetries(backoff.NewConstantBackOff(1*time.Second), 10))
```
JoshVanL: We must also always respect context.

JoshVanL (changes requested):

acroca:
Closing because 10s isn't enough (based on [this comment](https://github.com/dapr/components-contrib/issues/3715#issuecomment-3070529960)). We need a different solution for the problem

pkg/runtime/processor/subscriber/subscriber.go
```diff
@@ -251,6 +260,9 @@ func (s *Subscriber) ReloadDeclaredAppSubscription(name, pubsubName string) erro
 
 	ss, err := s.startSubscription(ps, sub.NamedSubscription, false)
 	if err != nil {
+		log.Errorf("Failed to start declared subscription %s for pubsub %s, topic %s: %s", name, pubsubName, sub.Topic, err)
+		go s.retrySubscription(pubsubName, ps, sub.NamedSubscription)
+
```
JoshVanL, line 264: Is is necessary for this to be in a go routine?

At least let's have a `sync.WaitGroup` to ensure all go routines have returned before continuing after closing those in retry.

pkg/runtime/processor/subscriber/subscriber.go
```diff
@@ -334,6 +350,35 @@ func (s *Subscriber) StartAppSubscriptions() error {
...
+				Subscription: ss,
+			})
+			log.Infof("Successfully started subscription after retry for pubsub %s, topic %s", pubsubName, sub.Topic)
+		} else {
+			// Subscriber was closed while we were retrying, stop the subscription
+			ss.Stop()
+		}
+		s.lock.Unlock()
+		return nil
+	}, backoff.WithContext(backoff.NewExponentialBackOff(backoff.WithMaxElapsedTime(0)), s.retryCtx))
```
JoshVanL: This looks like it will never retry (?)

pkg/runtime/processor/subscriber/subscriber.go
```diff
@@ -425,10 +472,14 @@ func (s *Subscriber) reloadPubSubApp(name string, pubsub *rtpubsub.PubsubItem) e
 
 	var errs []error
 	subs := make([]*namedSubscription, 0, len(s.compStore.ListSubscriptionsAppByPubSub(name)))
+
```
JoshVanL, line 487: Should we not cancel first (?).

pkg/runtime/processor/subscriber/subscriber.go
acroca: it's intentionally sending it to the background. The `retrySubscription` runs forever until it manages to reconnect. Only first attempt blocks this main thread, and if the first attempt fails we return the error but keep trying to reconnect in the background.

pkg/runtime/processor/subscriber/subscriber.go
acroca: with MaxElapsedTime=0 it retries forever: https://github.com/cenkalti/backoff/blob/v4/exponential.go#L60

pkg/runtime/processor/subscriber/subscriber.go
acroca: Correct. Just changed to handle cancellations per pubsub name instead of a global one.

cicoyle (approved):

olitomlinson:
@yaron2  I'm somewhat concerned that this is going to have an unexpected operational impact.

For example, in my org we rely on the sidecar crashing to indicate the system is not healthy at start-up triggering a crashloop, as is a normal approach to take in kubernetes.

My understanding here is that the sidecar will continue to run indefinitely (successfully passing its liveness probe) even though there may be a persistent issue.

IMO there should be at least a max timeout or max retry configuration here to put an upper limit on retries.

acroca:
@olitomlinson The problem is that errors on subscriptions were not crashing the process. They were simply logged.
Failing subscriptions are not a reason to crash if I understand correctly, they are not tied to the liveness check.
In your org do you run into this scenario and find a crash when subscriptions fail? Would you be able to send some logs? Maybe I'm missing something here.

olitomlinson:
> Failing subscriptions are not a reason to crash if I understand correctly

IMO they should crash, eventually. 

To be clear, I fully support having retries around this subscription start-up process, so long as there is a retry limit and when that retry limit exceeds the container is allowed to crash.

> In your org do you run into this scenario and find a crash when subscriptions fail?

It's not something I can say we've exactly had for this scenario, however we do rely generally on dapr crashing when things go wrong for too long. IMO The start-up process (which includes Subscriptions becoming established) shouldn't be allowed to just run indefinitely, in the face of an error that is blocking progress.

@yaron2 your thoughts?

PR #8847, opened 2025-07-03
update durabletask, new pending tasks interface

famarting:
# Description

## Issue reference

Please reference the issue this PR will close: #_[issue number]_

## Checklist

Please make sure you've completed the relevant tasks for this PR, out of the following list:

- [ ] Code compiles correctly
- [ ] Created/updated tests
- [ ] Unit tests passing
- [ ] End-to-end tests passing
- [ ] Extended the documentation / Created issue in the <https://github.com/dapr/docs/> repo: dapr/docs#_[issue number]_
- [ ] Specification has been updated / Created issue in the <https://github.com/dapr/docs/> repo: dapr/docs#_[issue number]_
- [ ] Provided sample for the feature / Created issue in the <https://github.com/dapr/docs/> repo: dapr/docs#_[issue number]_

PR #8883, opened 2025-07-16
workflows: support clustered deployments, provide alternative implementation to pending tasks backend

famarting:
# Description

Provides an experimental implementation after https://github.com/dapr/dapr/pull/8847
The new durabletask-go abstraction allows to provide an alternative implementation that enables to run workflows in scenarios where a dapr sidecar is deployed behind a load balancer, i.e the case of [dapr shared](https://docs.dapr.io/operations/hosting/kubernetes/kubernetes-dapr-shared/)

adds a new flag `workflows-enable-clustered-deployment` to enable this behavior and adds integration tests to validate it

## Issue reference

## Checklist

- [ ] Code compiles correctly
- [ ] Created/updated tests
- [ ] Unit tests passing


cmd/daprd/options/options.go
```diff
@@ -165,6 +166,7 @@ func New(origArgs []string) (*Options, error) {
 	fs.IntVar(&opts.AppHealthProbeTimeout, "app-health-probe-timeout", int(config.AppHealthConfigDefaultProbeTimeout/time.Millisecond), "Timeout for app health probes in milliseconds")
 	fs.IntVar(&opts.AppHealthThreshold, "app-health-threshold", int(config.AppHealthConfigDefaultThreshold), "Number of consecutive failures for the app to be considered unhealthy")
 	fs.StringVar(&opts.AppChannelAddress, "app-channel-address", runtime.DefaultChannelAddress, "The network address the application listens on")
+	fs.BoolVar(&opts.WorkflowsEnableClusteredDeployment, "workflows-enable-clustered-deployment", false, "Enable feature to support workflows in a clustered deployment")
```
cicoyle: Rather than this being a new daprd flag, can you make this a [preview feature](https://docs.dapr.io/operations/support/support-preview-features/) via [config](https://github.com/dapr/dapr/blob/master/pkg/config/configuration.go#L49-L59)?

famarting: okey, thanks, I'll create it with the name `WorkflowsClusteredDeployment`

well, I'll try to fix the integration tests first

famarting:
the integration test failures are weird... I need to troubleshoot them more, but it kind of looks like they are flagging an issue around placement... because the issue is that one of the two daprd instances does not get notified of the actor types registered in the other daprd

famarting:
@yaron2 @cicoyle can you retry tests?

pkg/runtime/wfengine/backends/actors/clustertasks.go
```diff
@@ -0,0 +1,219 @@
+package actors
```
cicoyle, line 15: plz add the header

pkg/runtime/wfengine/backends/actors/clustertasks.go
```diff
@@ -0,0 +1,219 @@
...
+		resp.GetInstanceId(),
+		resp.GetTaskId(),
+	)
+
+	data, err := proto.Marshal(resp)
+	if err != nil {
+		return err
+	}
+
+	req := internalsv1pb.
```
cicoyle, line 67: Since we introduce a new executor actor type for clustered workflow deployments. Should we consider adding metrics for the clustertask functionality for monitoring- Can you open an issue for adding metrics in a follow up PR?

We will also need docs for this new preview feature, please [see here for where the preview features are noted. ](https://v1-11.docs.dapr.io/operations/support/support-preview-features/)

tests/integration/suite/daprd/workflow/loadbalance/config.go
```diff
@@ -0,0 +1,43 @@
...
+kind: Configuration
+metadata:
+    name: workflowsclustereddeployment
+spec:
+    features:
+    - name: WorkflowsClusteredDeployment
+    enabled: true
+`
+	for i := range daprds {
+		wopts = append(wopts, workflow.WithDaprdOptions(i, daprd.WithAppID("app"), daprd.WithConfigManifests(t, config)))
```
cicoyle: Why do you want the daprd's to have `app` as the appID? Can you let the framework use an autogenerated appID so we dont have DNS misses due to caching if the same name is used?

tests/integration/suite/daprd/workflow/loadbalance/config.go
```diff
@@ -0,0 +1,43 @@
...
+package loadbalance
+
+import (
+	"testing"
+
+	"github.com/dapr/dapr/tests/integration/framework/process/daprd"
+	"github.com/dapr/dapr/tests/integration/framework/process/workflow"
+)
+
+func newClusteredDeployment(t *testing.T, daprds int) *workflow.Workflow {
```
cicoyle, line 28: Can you write a test that involves a case where there is a dapr or 2 not using the clustered deployment and one or 2 using the clustered deployment and ensure we dont have any collisions with state, wf names, etc? Just want to ensure there are no issues with a mixed deployment model. I would like to ensure we dont have any collisions/issues. This might need to be something documented explicitly.

tests/integration/suite/daprd/workflow/loadbalance/config.go
```diff
@@ -0,0 +1,43 @@
...
+    name: workflowsclustereddeployment
+spec:
+    features:
+    - name: WorkflowsClusteredDeployment
+    enabled: true
+`
+	for i := range daprds {
+		wopts = append(wopts, workflow.WithDaprdOptions(i, daprd.WithAppID("app"), daprd.WithConfigManifests(t, config)))
+	}
+	return workflow.New(t, wopts...)
```
cicoyle, line 50: Also, do we need a new statestore for clustered wf or are we reusing the actor state store?

cicoyle (changes requested):
Mostly lgtm - thanks for the preview feature add! Just a few comments/questions. Will do another review pass later in the day👍🏻

tests/integration/suite/daprd/workflow/loadbalance/config.go
famarting: I want all the daprds to be configured with the same appid, so they represent a "cluster" of daprds serving the same app, I can refactor to use an autogenerated appID

tests/integration/suite/daprd/workflow/loadbalance/config.go
famarting: reusing the same state store, there is no reason to need a new one

tests/integration/suite/daprd/workflow/loadbalance/config.go
famarting: that would be the same as running workflows on two different appids

famarting: added a conflict test

famarting:
tests failures seem flakes unrelated to this PR

PR #8888, opened 2025-07-17
[1.16] Prevent Jobs from not triggering upserting in high throughput scenarios

JoshVanL:
Problem

Upserting a Job (or Actor reminder) in high job trigger throughput scenarios could result in the Job not being triggered.

Impact

Jobs (or Actor reminders) could not be triggered in high throughput scenarios.

Root cause

Scheduler would find a conflict in jobs being triggered between the old and new job state, and would result in no more jobs to be triggered on that host.

Solution

Ignore the conflict in the scheduler when upserting a job (or Actor reminder) in high throughput scenarios.

### DIFF
--- a/docs/release_notes/v1.15.7.md
+++ b/docs/release_notes/v1.15.7.md
@@ -3,6 +3,7 @@
 This update includes bug fixes:
 
 - [Fix Workflow Start Time](#fix-workflow-start-time)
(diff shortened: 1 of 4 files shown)


cicoyle:
Did you mean to target the 1.16 branch? I see the title says 1.16, but is targeting master.

JoshVanL:
@cicoyle fixed

cicoyle (approved):

Issue #8651, opened 2025-04-07
Dapr 1.15 Workflow - Is it expected that workflows will break after updating?


joshuadmatthews:
## In what area(s)?

> /area runtime

> /area operator

> /area placement

> /area docs

> /area test-and-release

## What version of Dapr?

> 1.15

## Expected Behavior

Upgrade to Dapr 1.15 without breaking workflows

## Actual Behavior

failed to delete activity state for activity actor '1974bc3e-87ff-5b09-964c-a9ecd127f51c::0::1' with error: transaction failed due to operation 0 which failed with status code 404

## Steps to Reproduce the Problem

We were using CosmosDB as our actor state store. We never enabled SchedulerReminders feature flag, I was under the impression that with the 1.15 upgrade all of the existing workflows would be migrated from the actor state store. But instead, we started seeing errors with existing and new workflows. Nothing works after the upgrade. Currently in the process of rolling back to fix our env, shame on us for having auto upgrade enabled in production I know, we were not aware that was enabled.

Is it expected that simply upgrading to 1.15 would break workflows though?

## Release Note

RELEASE NOTE:

abigailrmiller:
I'm seeing the same errors after upgrading to 1.15.0 with a CosmosDB actor state store. Also tried upgrading with SchedulerReminders = false and same result. This seems specific to CosmosDB, if I swap to in-memory or Azure SQL the workflows complete successfully.

dthk-cogmatix:
Experiencing the same issue as above and utilizing CosmosDB as actor state store.

acroca:
The fix should be available in the `1.15.9`, you can try it now using `v1.15.9-rc.1`

PR #8891, opened 2025-07-17
[FIX 1.15] Prevent workflow cleanup process to delete activities that are not in the state

acroca:
# Description

This error surfaced with the Azure Cosmos DB component, reported [here](https://github.com/dapr/dapr/issues/8651).
The problem is that dapr sends a delete operation on the activity, but this actor has never been created in the database and Cosmos returns 404. Out of curiosity I compared it with Redis and MySQL and they both can handle deleting a key that doesn't exist, that's why they don't fail in the same situation.

I've added this check to not do the delete operation for missing keys.

## Issue reference

Please reference the issue this PR will close: https://github.com/dapr/dapr/issues/8651

## Checklist

- [ ] Code compiles correctly
- [ ] Created/updated tests
- [ ] Unit tests passing


acroca:
After a discussion with the team we've decided to fix it in components-contrib, aligning the cosmos component to the others, as the other components don't fail when deleting a key that doesn't exist.

PR #8894, opened 2025-07-18
Workflow: always stream state every 3 seconds

JoshVanL:
Update Workflow state streamer to always stream state every 3 seconds.


cicoyle (approved):
makes sense. hopefully we see some greener runs in CI

cicoyle (approved):

PR #8969, opened 2025-08-07
feat: ensure endgame task comp registration + complete metadata bundle

sicoyle:
# Description

NOTE: build will fail until my contrib PR is merged: https://github.com/dapr/components-contrib/pull/3949

For each dapr release there is an endgame task to ensure that components are properly registered across contrib and dapr/dapr. I took a bit of time to create a new CLI in my contrib PR linked blow to automate this for myself (assigned this task), as well as for others. I noticed several components not registered properly in runtime, or in the docs This PR is a first stab at fully automating verifying component registration across contrib and runtime, and does not hit every check box on my issue I created below, but again, is a great starting point and can be iterated on for future releases. This corrects issues I saw for component registration, as well as ensured that naming conventions are upheld. This means things like:
- runtime binding_* -> bindings_* to conform to the convention of componenttype_blah
- runtime crypto -> cryptography since it is a cryptography component type
- added missing rocketmq component
- correct 4 registration errors on all files with a string addition to include the missing vendor prefix, or say for servicebusqueues add the alias that contrib expects https://github.com/dapr/components-contrib/blob/d29c295f44359339b59f99b7d83e910faddad5ec/tests/config/bindings/azure/servicebusqueues/bindings.yml#L7
- update all registry.go files to have the same setup for a default registry singleton that I look for in my contrib cli so that is a vanity / consistency change

It's also nice bc i print out the counts, so we have accurate metrics in future and can also cross ref in the docs later on which more are missing than I found in my manual cross referencing

My contrib CLI PR that this needs: https://github.com/dapr/components-contrib/pull/3949

## Issue reference

https://github.com/dapr/components-contrib/issues/3948

## Checklist

- [ ] Code compiles correctly
- [ ] Created/updated tests
- [ ] Unit tests passing


cmd/daprd/components/crypto_jwks.go
```diff
@@ -21,5 +21,6 @@ import (
 )
 
 func init() {
-	cryptoLoader.DefaultRegistry.RegisterComponent(jwks.NewJWKSCrypto, "dapr.jwks")
+	// TODO: in future rm the dapr prefix
+	cryptoLoader.DefaultRegistry.RegisterComponent(jwks.NewJWKSCrypto, "jwks", "dapr.jwks")
```
JoshVanL: Please revert all instances of adding non-`dapr` prefixes. These groups are intentional, and describe components whose provider is Dapr itself, rather than some other external provider (Azure, AWS etc.)

JoshVanL: It also seems like this needs reflection on contrib also?
https://github.com/dapr/components-contrib/pull/3949/files

Issue #8975, opened 2025-08-11
Sidecar panics when calling Non-Dapr endpoint with redirect behavior


fabistb:
## In what area(s)?

 /area runtime

## What version of Dapr?
1.15.6
1.15.9

## Expected Behavior
When utilizing the _Call Non-Dapr Endpoint_ functionality there  shouldn't be a difference if the called domain is called directly of if there is some kind or redirect behavior. E.G. adding some localization for the language.

## Actual Behavior
It appears that the Dap sidecar panics when a URL is called which performs some redirects.

Logs:
```
2025/08/11 09:25:52 http: panic serving 127.0.0.1:54852: runtime error: invalid memory address or nil pointer dereference
goroutine 1297 [running]:
net/http.(*conn).serve.func1()
	/opt/hostedtoolcache/go/1.24.4/x64/src/net/http/server.go:1947 +0xb0
panic({0x531d120?, 0xbac6130?})
	/opt/hostedtoolcache/go/1.24.4/x64/src/runtime/panic.go:792 +0x124
github.com/dapr/dapr/pkg/diagnostics.(*pathMatchingRW).Header(0x4000e01840?)
	<autogenerated>:1 +0x28
net/http.Redirect({0x6fd5cb0, 0x400136cd20}, 0x4000a44f00, {0x4000e01840, 0x2d}, 0x12d)
	/opt/hostedtoolcache/go/1.24.4/x64/src/net/http/server.go:2408 +0xe0
net/http.(*redirectHandler).ServeHTTP(0x4000a77800?, {0x6fd5cb0?, 0x400136cd20?}, 0x423cb60?)
	/opt/hostedtoolcache/go/1.24.4/x64/src/net/http/server.go:2449 +0x40
net/http.(*ServeMux).ServeHTTP(0x5e724bd4dd4c49bc?, {0x6fd5cb0, 0x400136cd20}, 0x4000a44f00)
	/opt/hostedtoolcache/go/1.24.4/x64/src/net/http/server.go:2822 +0x1b4
github.com/dapr/dapr/pkg/diagnostics.(*pathMatching).match(0x400084b0f0, {0x4000e005c4?, 0x4000e005c4?})
	/home/runner/work/dapr/dapr/pkg/diagnostics/http_monitoring_path_matching.go:121 +0x140
github.com/dapr/dapr/pkg/diagnostics.(*httpMetrics).getMetricsPath(0x4000855d00, {0x4000e005c4, 0x2f})
[...]

PR #8977, opened 2025-08-11
FIX Call Non-Dapr endpoint with redirect behavior

MyMirelHub:
**Description**

Fixes a `nil pointer dereference panic` in the HTTP metrics middleware that occurs when invoking a non-Dapr endpoint that returns an HTTP redirect. Panic was caused by an uninitialized `http.ResponseWriter` field in the `pathMatchingRW` struct. 

- Added a `Header()` implementation to `pathMatchingRW` that always returns a valid `http.Header` map.
 
- Implemented no-op `Write` and `WriteHeader` methods to satisfy the `http.ResponseWriter` interface when no underlying writer is present.

- Introduced a new unit test `TestHTTPMetricsPathMatchingWithRedirect` to simulate the scenario.

**Issue reference** - #8975

- [x] Code compiles correctly
- [x] Created/updated tests
- [x] Unit tests passing

### DIFF
--- a/pkg/diagnostics/http_monitoring_path_matching.go
+++ b/pkg/diagnostics/http_monitoring_path_matching.go
@@ -124,6 +124,22 @@ func (pm *pathMatching) match(path string) (string, bool) {
 }
 
 type pathMatchingRW struct {
-	http.ResponseWriter
 	matchedPath string
+	header      http.Header
+}
+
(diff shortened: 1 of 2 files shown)


MyMirelHub:
closing for targeted 1.15 and 1.16 release

PR #8979, opened 2025-08-12
Workflows: reduce go routine overhead

JoshVanL:
Update the workflow orchestrator target to reduce the number of go routines which are spawned, in favour of doing work sequentially. This reduces the overhead of go rountine context thrashing.

Uses a main deactivation channel in the orchestration factory to handle cleanup of orchestration state.

Updates the workflow locker to use a `Close` to signal close of the workflow entry points.

### DIFF
--- a/pkg/actors/targets/workflow/orchestrator/orchestrator.go
+++ b/pkg/actors/targets/workflow/orchestrator/orchestrator.go
@@ -95,6 +95,10 @@ func (o *orchestrator) InvokeStream(ctx context.Context, req *internalsv1pb.Inte
 
 // DeactivateActor implements actors.InternalActor
 func (o *orchestrator) Deactivate(ctx context.Context) error {
+	if !o.closed.CompareAndSwap(false, true) {
+		return nil
+	}
+
 	o.wg.Add(1)
 
 	unlock, err := o.lock.ContextLock(ctx)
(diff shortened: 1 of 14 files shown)


pkg/actors/targets/workflow/orchestrator/orchestrator.go
```diff
@@ -103,17 +95,29 @@ func (o *orchestrator) InvokeStream(ctx context.Context, req *internalsv1pb.Inte
...
+	o.state = nil
+	o.rstate = nil
+	o.ometa = nil
+	o.ometaBroadcaster.Close()
+	o.lock.Close()
+	o.wg.Wait()
+	orchestratorCache.Put(o)
 
-	o.cleanup()
-	log.Debugf("Workflow actor '%s': deactivated", o.actorID)
```
cicoyle, line 116: Shall we keep this log? Or did I miss you moving it somewhere else?

cicoyle (changes requested):
One comment, and I see a dep issue with CI -- could you please review that.

Also, could we run performance tests on this PR to check for any impact from changing to sequential execution

pkg/actors/targets/workflow/orchestrator/orchestrator.go
JoshVanL: Just went ahead and removed it- there isn't any useful information a developer/someone debugging can get from this information imo, and is simply taking up cycles.

JoshVanL:
@cicoyle yeah I see the dep error- tis the same of release-1.16 atm. Let's fix separately. 


```
$ ~/go/bin/govulncheck ./...

=== Symbol Results ===

Vulnerability #1: GO-2025-3849
    Incorrect results returned from Rows.Scan in database/sql
  More info: https://pkg.go.dev/vuln/GO-2025-3849
  Standard library
    Found in: database/sql@go1.24.4
    Fixed in: database/sql@go1.24.6
    Example traces found:
      #1: tests/integration/suite/daprd/workflow/purge.go:65:84: workflow.purge.Run calls sql.Row.Scan
      #2: tests/integration/framework/process/sqlite/actors.go:69:30: sqlite.SQLite.ActorReminders calls sql.Rows.Scan

Your code is affected by 1 vulnerability from the Go standard library.
This scan found no other vulnerabilities in packages you import or modules you
require.
Use '-show verbose' for more details.
```

yaron2 (approved):

PR #8984, opened 2025-08-13
[1.16] fix(diagnostics): redirect

MyMirelHub:
**Description**

Fixes a `nil pointer dereference panic` in the HTTP metrics middleware that occurs when invoking a non-Dapr endpoint that returns an HTTP redirect. Panic was caused by an uninitialized `http.ResponseWriter` field in the `pathMatchingRW` struct. 

- Added a `Header()` implementation to `pathMatchingRW` that always returns a valid `http.Header` map.
 
- Implemented no-op `Write` and `WriteHeader` methods to satisfy the `http.ResponseWriter` interface when no underlying writer is present.

- Introduced a new unit test `TestHTTPMetricsPathMatchingWithRedirect` to simulate the scenario.

**Issue reference** - #8975

- [x] Code compiles correctly
- [x] Created/updated tests
- [x] Unit tests passing

### DIFF
--- a/pkg/diagnostics/http_monitoring_path_matching.go
+++ b/pkg/diagnostics/http_monitoring_path_matching.go
@@ -124,6 +124,22 @@ func (pm *pathMatching) match(path string) (string, bool) {
 }
 
 type pathMatchingRW struct {
-	http.ResponseWriter
 	matchedPath string
+	header      http.Header
+}
+
+// Header returns a non-nil header map
+func (rw *pathMatchingRW) Header() http.Header {
+	if rw.header == nil {
+		rw.header = make(http.Header)
+	}
+	return rw.header
+}
+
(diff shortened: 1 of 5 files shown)


pkg/diagnostics/http_monitoring_path_matching.go
```diff
@@ -124,6 +124,22 @@ func (pm *pathMatching) match(path string) (string, bool) {
...
+// Header returns a non-nil header map
+func (rw *pathMatchingRW) Header() http.Header {
+	if rw.header == nil {
+		rw.header = make(http.Header)
+	}
+	return rw.header
+}
+
+func (rw *pathMatchingRW) Write(b []byte) (int, error) {
+	return 0, nil
```
JoshVanL:
```suggestion
	return len(b), nil
```

PR #9012, opened 2025-08-26
sentry: enforce different root/intermediate certificates

lrascao:
# Description

sentry: enforce different root/intermediate certificates


pkg/sentry/server/ca/ca.go
```diff
@@ -128,42 +128,61 @@ func New(ctx context.Context, conf config.Config) (Signer, error) {
...
 
-		generatedBundle, err := ca_bundle.Generate(ca_bundle.GenerateOptions{
-			X509RootKey:        x509RootKey,
-			JWTRootKey:         jwtRootKey,
-			TrustDomain:        conf.TrustDomain,
-			AllowedClockSkew:   conf.AllowedClockSkew,
-			OverrideCATTL:      nil,
-			MissingCredentials: missingCreds,
-		})
+		generatedBundle, err := ca_bundle.Generate(
```
JoshVanL: @lrascao Does this second `ca_bundle.Generate` generate multiple root keys unnecessarily ?

JoshVanL (changes requested):

pkg/sentry/server/ca/ca.go
JoshVanL: Nvm, I see you are setting MissingCredentials: ca_bundle.MissingCredentials{JWT: true},

jjcollinge:
was this change actually needed in the end - did this bit not effectively do the same originally?

```
if missingCreds.X509 {
	bundle.X509 = generatedBundle.X509
}
if missingCreds.JWT {
	bundle.JWT = generatedBundle.JWT
}
```
I guess maybe the logs were confusing because it may have printed `Self-signed certs generated and persisted successfully` if it wrote either the JWT or X509 root. So at least this cleanly separates them.

lrascao:
> was this change actually needed in the end - did this bit not effectively do the same originally?
> 
> ```
> if missingCreds.X509 {
> 	bundle.X509 = generatedBundle.X509
> }
> if missingCreds.JWT {
> 	bundle.JWT = generatedBundle.JWT
> }
> ```
> 
> I guess maybe the logs were confusing because it may have printed `Self-signed certs generated and persisted successfully` if it wrote either the JWT or X509 root. So at least this cleanly separates them.

don't think so, original issue i ran into is one where i'm supplying only my own keys, the previous `MissingRootKeys` would see that JWT keys were missing, generate a new key pair and ignore my provided ones

jjcollinge:
Hmm it'd generate them yes (which I guess is not ideal) but it should not assign them based on the code snippet I shared?

lrascao:
´´

> Hmm it'd generate them yes (which I guess is not ideal) but it should not assign them based on the code snippet I shared?

```
TL;DR
Due to sentryca trying to generate new certs while ignoring the provided ones we're falling into a previously existing behaviour of `selfhosted.store` writing an empty slice over the provided IssuerCertPath file on disk
```

true, there's another possible bug at play here, here's what i found:

* in my test use case i'm running in self-hosted mode
* i'm supplying a single root keypair, that means `RootCertPath` and `IssuerCertPath` point to the same public key, `IssuerKeyPath` when creating a `sentryca` (in `github.com/dapr/dapr/pkg/sentry/server/ca`):

```
	ca, err := sentryca.New(t.Context(),
		sentrycfg.Config{
			Mode:             modes.StandaloneMode,
			RootCertPath:     caPath,
			IssuerCertPath:   certPath,
			IssuerKeyPath:    keyPath,
			WorkloadCertTTL:  time.Hour,
			AllowedClockSkew: time.Second * 5,
		})
	require.NoError(t, err)
```

* `conf.Mode` is not Kubernetes so `castore` is implemented by `selfhosted`
```
		log.Info("Using local file system for trust bundle storage")
		castore = &selfhosted{config: conf}
```

* bundle is retrieved from `castore`
```
	bundle, missingCreds, err := castore.get(ctx)
	if err != nil {
		return nil, fmt.Errorf("failed to get CA bundle: %w", err)
	}
```

* `castore.get` implemented by `selfhosted` loads the PEM files from disk and invokes `verifyBundle`. As part of it's verification process it decodes the PEM chain from `IssuerCertPath`:
```
	issChain, err := pem.DecodePEMCertificatesChain(issChainPEM)
	if err != nil {
		return bundle.Bundle{}, err
	}
```
It then performs some checks on it before re-encoding it back into PEM:
```
	// Re-encode the issuer chain to ensure it contains only the issuer chain,
	// and strip out PEM comments since we don't want to send them to the client.
	issChainPEM, err = pem.EncodeX509Chain(issChain)
	if err != nil {
		return bundle.Bundle{}, err
	}
```
A check for an empty slice on `issChainPEM` would fail at this point, the reason behind that is detailed in https://github.com/dapr/kit/pull/134

* JWT credentials are missing so a new X509 key pair will be generated, it won't be assigned to `bundle` however. What will happen though is that `castore.store` will be invoked:
```
		if err := castore.store(ctx, bundle); err != nil {
			return nil, fmt.Errorf("failed to store CA bundle: %w", err)
		}
```

`selfhosted.store` will now write the previously empty slice back into `IssuerCertPath` and clobber it.

pkg/sentry/server/ca/ca_test.go
```diff
@@ -297,16 +298,13 @@ func TestNew(t *testing.T) {
...
-			int2PEM, int2Crt, int2PKPEM, int2PK := genCrt(t, "int2", int1Crt, int1PK)
+			rootPEM, _, rootPEMKey, _ := genCrt(t, "root", nil, nil)
 
 			//nolint:gocritic
-			rootFileContents := append(rootPEM, rootPEM2...)
+			rootFileContents := rootPEM
 			//nolint:gocritic
-			issuerFileContents := append(int2PEM, int1PEM...)
-			issuerKeyFileContents := int2PKPEM
+			issuerFileContents := rootPEM
```
lrascao, line 305: notice how both root and issuer are the same

pkg/sentry/server/ca/ca_test.go
```diff
@@ -319,6 +317,7 @@ func TestNew(t *testing.T) {
 			require.NoError(t, err)
 			issuerCert, err := os.ReadFile(issuerCertPath)
 			require.NoError(t, err)
+			require.NotEmpty(t, issuerCert)
```
lrascao: this fails as `issuerCertPath` file contents have been clobbered

lrascao:
red test failing as explained above:

```
=== FAIL: pkg/sentry/server/ca TestNew/if_existing_pool_exists,_new_should_load_the_existing_pool_and_generate_a_jwt_bundle (0.33s)
time="2025-08-27T10:57:18.265979293Z" level=info msg="Using local file system for trust bundle storage" instance=pkrvmqc4gcfdwos scope=dapr.sentry.ca type=log ver=unknown
time="2025-08-27T10:57:18.266298723Z" level=info msg="Root and issuer certs found: using credentials from store" instance=pkrvmqc4gcfdwos scope=dapr.sentry.ca type=log ver=unknown
time="2025-08-27T10:57:18.266312799Z" level=info msg="JWT not found: generating self signed" instance=pkrvmqc4gcfdwos scope=dapr.sentry.ca type=log ver=unknown
time="2025-08-27T10:57:18.595339389Z" level=info msg="JWT generated and persisted successfully" instance=pkrvmqc4gcfdwos scope=dapr.sentry.ca type=log ver=unknown
time="2025-08-27T10:57:18.595375016Z" level=info msg="JWT issuing enabled" instance=pkrvmqc4gcfdwos scope=dapr.sentry.ca type=log ver=unknown
time="2025-08-27T10:57:18.595396276Z" level=info msg="Using JWT kid from thumbprint: 1YF+04gQse2fklkGUwHmByJZ/hHHW+Zdy2ct+l/uvHQ=, please ensure this aligns with your JWKS" instance=pkrvmqc4gcfdwos scope=dapr.sentry.ca type=log ver=unknown
    ca_test.go:320: 
        	Error Trace:	/home/runner/work/dapr/dapr/pkg/sentry/server/ca/ca_test.go:320
        	Error:      	Should NOT be empty, but was []
        	Test:       	TestNew/if_existing_pool_exists,_new_should_load_the_existing_pool_and_generate_a_jwt_bundle

=== FAIL: pkg/sentry/server/ca TestNew (0.84s)
```

pkg/sentry/server/ca/ca_test.go
JoshVanL: The root and issuer certs should not point to the same document- Sentry signed from an intermediate, signed by the root

pkg/sentry/server/ca/bundle/bundle.go
```diff
@@ -91,13 +92,13 @@ type GenerateOptions struct {
...
 
-	if opts.MissingCredentials.X509 {
+	if opts.MissingCredentials.X509 && !reflect.ValueOf(opts.X509RootKey).IsNil() {
 		if err := generateX509Bundle(opts, &bundle); err != nil {
 			return Bundle{}, fmt.Errorf("failed to generate X.509 bundle: %w", err)
 		}
 	}
 
-	if opts.MissingCredentials.JWT {
+	if opts.MissingCredentials.JWT && !reflect.ValueOf(opts.JWTRootKey).IsNil() {
```
lrascao: not crazy about this but it's the only way to check that the interface's concrete value is not nil which will happen when JWT is not enabled

JoshVanL:
I have extended this PR to make the checking and generation of X.509 & JWT separated  https://github.com/dapr/dapr/pull/9016/files

lrascao:
rebased, scope changed into just enforcing that root/intermediate certificates are different when supplying a x509 bundle

JoshVanL (approved):

PR #9016, opened 2025-08-27
[1.16] Simplify X.509 & JWT logic

JoshVanL:
Simplify the Sentry X.509 and JWT generation logic to compare and generate separately. Only writes when necessary.

Extends from https://github.com/dapr/dapr/pull/9012

### DIFF
--- a/pkg/sentry/server/ca/ca.go
+++ b/pkg/sentry/server/ca/ca.go
@@ -35,7 +35,7 @@ import (
 	"github.com/dapr/dapr/pkg/security/spiffe"
 	"github.com/dapr/dapr/pkg/sentry/config"
 	"github.com/dapr/dapr/pkg/sentry/monitoring"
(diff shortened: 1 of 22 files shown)


pkg/sentry/server/ca/ca.go
```diff
@@ -114,12 +114,15 @@ func New(ctx context.Context, conf config.Config) (Signer, error) {
...
 
-	bundle, missingCreds, err := castore.get(ctx)
+	bndle, err := castore.get(ctx)
 	if err != nil {
 		return nil, fmt.Errorf("failed to get CA bundle: %w", err)
 	}
 
-	if missingCreds.MissingRootKeys() {
+	var needsWrite bool
+	if bndle.X509 == nil {
```
jjcollinge, line 123: typo?

JoshVanL: `bundle` string is taken up by the `bundle` package name.

cicoyle (approved):

PR #9037, opened 2025-09-03
[1.16] Workflows: Fix & improve streaming

JoshVanL:
Updates orchestrator workflow actor streaming so that it is done synchronously, both for metadata condition reporting & event sinking. This prevents a race in reporting, and fixes a bug on shutdown when a workflow is in a `FAILED` state.

Improves workflow performance by reducing the number of go routines spqawned per workflow, both when waiting for conditions, and when using an event sink. Data is piped via a series of middleware callbacks.

Updates github.com/dapr/durabletask-go to v0.9.0 for implementation.


pkg/actors/targets/workflow/orchestrator/orchestrator.go
```diff
@@ -38,16 +38,23 @@ type orchestrator struct {
...
+	rstate *backend.OrchestrationRuntimeState
+	ometa  *backend.OrchestrationMetadata
 
 	activityResultAwaited atomic.Bool
-	completed             atomic.Bool
 	lock                  *lock.Lock
 	closed                atomic.Bool
 	wg                    sync.WaitGroup
+
+	streamFns map[int64]*streamFn
```
famarting, line 50: is all access to this map protected from concurrent reads and writes?

pkg/actors/targets/workflow/orchestrator/orchestrator.go
```diff
@@ -83,38 +90,40 @@ func (o *orchestrator) InvokeTimer(ctx context.Context, reminder *actorapi.Remin
...
 	o.wg.Add(1)
 	defer o.wg.Done()
-	return o.handleStream(ctx, req, stream)
-}
 
-// DeactivateActor implements actors.InternalActor
-func (o *orchestrator) Deactivate(ctx context.Context) error {
-	if !o.closed.CompareAndSwap(false, true) {
-		return nil
+	unlock, err := o.lock.ContextLock(ctx)
```
famarting, line 97: this is new, no? there wasn't a requirement before that did not allow to InvokeStream on an actor multiple times concurrently

JoshVanL: Yes, this is because of the stream fn map we use to stream the state changes

pkg/actors/targets/workflow/orchestrator/orchestrator.go
JoshVanL: Yes, by virtue of actor in turn lock guarantees

pkg/runtime/wfengine/backends/actors/clustertasks.go
```diff
@@ -194,40 +175,22 @@ func (be *ClusterTasksBackend) WaitForOrchestratorCompletion(ctx context.Context
...
-			break
+	var resp protos.OrchestratorResponse
+	err = router.CallStream(ctx, sreq, func(res *internalsv1pb.InternalInvokeResponse) (bool, error) {
+		if res == nil {
+			return false, errors.New("received nil response from activity completion")
 		}
-		if ctx.Err() != nil {
-			return nil, err
+		if res.GetStatus().GetCode() == int32(codes.Aborted) {
+			return false, api.ErrTaskCancelled
```
famarting, line 184: its crucial that non of the `router.CallStream` machinery incorrectly wraps this error

pkg/actors/targets/workflow/orchestrator/orchestrator.go
famarting: I'm not understanding something, you have a map of streamFn, but it is not possible to cann `InvokeStream` multiple times concurrently... it doesn't make sense, no?

JoshVanL: Invoke stream will lock until it has fetched the current state and sent it, & added its steam to the map of steams which are watching state.

After it has sent the current state and added its reference, it will release the lock.

pkg/runtime/wfengine/backends/actors/clustertasks.go
JoshVanL: Can you elaborate?

famarting (approved):
LGTM ty

PR #9045, opened 2025-09-05
[1.16] Workflows: (runtime) retry completing/cancelling

JoshVanL:
To account for the race between and activity/orchestration being tracked as pending in the durabletask backend, and the client reporting it as being completed/cancelled, add a backoff retry. This prevents activities and orchestrators from being stuck in a pending state forever, which can occur in high throughput scenarios.

Also changes the runtime wf runtime state to change an info to debug log for a nebulous log line.

### DIFF
--- a/pkg/runtime/wfengine/backends/actors/actors.go
+++ b/pkg/runtime/wfengine/backends/actors/actors.go
@@ -585,50 +585,35 @@ func (abe *Actors) ActivityActorType() string {
 
 // CancelActivityTask implements backend.Backend.
 func (abe *Actors) CancelActivityTask(ctx context.Context, instanceID api.InstanceID, taskID int32) error {
-	return backoff.Retry(func() error {
-		err := abe.pendingTasksBackend.CancelActivityTask(ctx, instanceID, taskID)
-		if err != nil && ctx.Err() == nil {
-			log.Warnf("error completing activity task: %v, retrying...", err)
-		}
-		if abe.stopped.Load() {
-			return backoff.Permanent(err)
-		}
-		return err
-	}, backoff.WithContext(backoff.NewConstantBackOff(time.Second), ctx))
+	return abe.callWithBackoff(ctx, func() error {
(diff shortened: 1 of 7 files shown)


pkg/runtime/wfengine/backends/actors/actors.go
```diff
@@ -580,22 +585,58 @@ func (abe *Actors) ActivityActorType() string {
...
+	return backoff.Retry(func() error {
+		err := abe.pendingTasksBackend.CancelActivityTask(ctx, instanceID, taskID)
+		if err != nil && ctx.Err() == nil {
+			log.Warnf("error completing activity task: %v, retrying...", err)
+		}
+		if abe.stopped.Load() {
+			return backoff.Permanent(err)
+		}
+		return err
+	}, backoff.WithContext(backoff.NewConstantBackOff(time.Second), ctx))
```
filintod: should we add a random +/- 100-200ms jitter to distribute retries if they happen in bursts at high throughput? if the errors are not bursty it might not matter

JoshVanL: Done

tests/integration/suite/daprd/workflow/noawait.go
```diff
@@ -50,7 +50,7 @@ func (n *noawait) Run(t *testing.T, ctx context.Context) {
 
 	var barCalled atomic.Int64
 	n.workflow.Registry().AddOrchestratorN("noawait", func(ctx *task.OrchestrationContext) (any, error) {
-		for range 50 {
+		for range 30 {
```
famarting, line 53: What? We have a test like this? Calling activities and not awaiting them? This shouldnt event be supported

JoshVanL: Indeed, this is somewhat undefined behaviour... this test will at least ensure behaviour doesn't change until we have some kind of mechanism to do something different

jjcollinge:
Just for clarity, under what circumstances could the back off expire and then what is the recovery process?

PR #9061, opened 2025-09-15
scheduler: configure embedded etcd with same log level

lrascao:
# Description

Configure embedded etcd with same log level as scheduler


JoshVanL:
@lrascao please resolve conflicts

PR #9071, opened 2025-09-18
[1.16] Optionally configure scheduler client port listen address

JoshVanL:
Fix Scheduler Etcd client port networking in standalone mode

Problem

The Scheduler Etcd client port is not available when running in Dapr CLI standalone mode.

Impact

Cannot perform Scheduler Etcd admin operations in Dapr CLI standalone mode.

Root cause

The Scheduler Etcd client port is only listened on localhost.

Solution

The Scheduler Etcd client listen address is now configurable via the `--scheduler-etcd-client-listen-address` CLI flag, meaning port can be exposed when running in standalone mode.

---

The listen address will continue to be configured as `127.0.0.1` when running in Kubernetes.

### DIFF
--- a/docs/release_notes/v1.16.1.md
+++ b/docs/release_notes/v1.16.1.md
@@ -3,6 +3,8 @@
 This update includes bug fixes:
 
 - [Sidecar Injector Crash with Disabled Scheduler](#sidecar-injector-crash-with-disabled-scheduler)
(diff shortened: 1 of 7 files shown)


acroca (approved):

docs/release_notes/v1.16.1.md
```diff
@@ -0,0 +1,23 @@
+# Dapr 1.15.8
```
javier-aliaga: 1.16.1?

PR #9078, opened 2025-09-25
Workflows bugfix: app health check transitions stops reminders

famarting:
# Description

We have found that app health checks were making the sidecar to stop watching for reminder jobs from the scheduler, resulting in workflows that suddenly stopped being executed

Adds an integration test that requires improvements, we either lack machinery, or I am not aware of it, to check the internal details that this test tries to control in order to exactly reproduce the issue

This test also showed a daprd side-effect between workflows and actors, the metadata API doesn't report the workflow actors if the app health check reports unhealthy

kudos to @lrascao for the help troubleshooting this

## Issue reference

## Checklist

- [ ] Code compiles correctly
- [ ] Created/updated tests
- [ ] Unit tests passing

### DIFF
--- a/tests/integration/suite/daprd/workflow/apphealth/apphealth.go
+++ b/tests/integration/suite/daprd/workflow/apphealth/apphealth.go
@@ -1,0 +1,124 @@
+/*
+Copyright 2025 The Dapr Authors
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://wwb.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package apphealth
+
+import (
+	"context"
+	"errors"
+	"sync/atomic"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"google.golang.org/protobuf/types/known/emptypb"
+
+	rtv1 "github.com/dapr/dapr/pkg/proto/runtime/v1"
+	"github.com/dapr/dapr/tests/integration/framework"
+	"github.com/dapr/dapr/tests/integration/framework/iowriter/logger"
+	"github.com/dapr/dapr/tests/integration/framework/process/daprd"
+	"github.com/dapr/dapr/tests/integration/framework/process/grpc/app"
+	"github.com/dapr/dapr/tests/integration/framework/process/workflow"
+	"github.com/dapr/dapr/tests/integration/suite"
+	"github.com/dapr/durabletask-go/api"
+	durabletask_client "github.com/dapr/durabletask-go/client"
+	"github.com/dapr/durabletask-go/task"
+)
+
+// luis case: app is unhealthy, worker is connected, app transitions to healthy
+
+func init() {
+	suite.Register(new(apphealth))
+}
+
+type apphealth struct {
+	healthy  atomic.Bool
+	app      *app.App
+	workflow *workflow.Workflow
+}
+
+func (a *apphealth) Setup(t *testing.T) []framework.Option {
+	a.healthy.Store(true)
+	a.app = app.New(t,
+		app.WithHealthCheckFn(func(context.Context, *emptypb.Empty) (*rtv1.HealthCheckResponse, error) {
+			if a.healthy.Load() {
+				return &rtv1.HealthCheckResponse{}, nil
+			}
+			return nil, errors.New("app not healthy")
+		}),
+	)
+
+	a.workflow = workflow.New(t,
+		workflow.WithDaprdOptions(0,
+			daprd.WithAppPort(a.app.Port(t)),
+			daprd.WithAppProtocol("grpc"),
+			daprd.WithAppHealthCheck(true),
+			daprd.WithAppHealthProbeInterval(1),
+			daprd.WithAppHealthProbeThreshold(1),
+		),
+	)
+
+	return []framework.Option{
+		framework.WithProcesses(a.app, a.workflow),
+	}
+}
+
+func (a *apphealth) Run(t *testing.T, ctx context.Context) {
+	a.workflow.WaitUntilRunning(t, ctx)
+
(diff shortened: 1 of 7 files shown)


tests/integration/suite/daprd/workflow/apphealth/apphealth.go
```diff
@@ -0,0 +1,124 @@
...
+	"github.com/dapr/dapr/tests/integration/framework/process/daprd"
+	"github.com/dapr/dapr/tests/integration/framework/process/grpc/app"
+	"github.com/dapr/dapr/tests/integration/framework/process/workflow"
+	"github.com/dapr/dapr/tests/integration/suite"
+	"github.com/dapr/durabletask-go/api"
+	durabletask_client "github.com/dapr/durabletask-go/client"
+	"github.com/dapr/durabletask-go/task"
+)
+
+// luis case: app is unhealthy, worker is connected, app transitions to healthy
```
JoshVanL: is this test not saying the app _starts_ as healthy, then is unhealthy, then becomes healthy?

tests/integration/suite/daprd/workflow/apphealth/apphealth.go
```diff
@@ -0,0 +1,124 @@
...
+		return nil, nil
+	})
+	a.workflow.Registry().AddActivityN("bar", func(ctx task.ActivityContext) (any, error) {
+		return nil, nil
+	})
+
+	// app starts unhealthy
+	a.healthy.Store(false)
+	// we have no way of knowing when the sidecar detected if the app is unhealthy
+	time.Sleep(time.Second * 3)
```
JoshVanL, line 95: This should be reported by daprd.

```go
	client := client.HTTP(t)
	req, err := http.NewRequestWithContext(ctx, http.MethodGet,
		fmt.Sprintf("http://%s/v1.0/healthz", a.workflow.Dapr().HTTPAddress()), nil)
	require.NoError(t, err)

		resp, err := client.Do(req)
		require.NoError(t, err)
		assert.NotEqual(c, http.StatusOK, resp.StatusCode)
	}, time.Second*10, time.Millisecond*10)
```

tests/integration/suite/daprd/workflow/apphealth/apphealth.go
```diff
@@ -0,0 +1,124 @@
...
+	require.NoError(t, client.StartWorkItemListener(ctx, a.workflow.Registry()))
+	// this sleep is the key
+	// it makes sure you get the placement tables update before the application becomes healthy
+	// the root cause of the bug we were following was that the app transitioning to healthy was wiping out the actor types
+	time.Sleep(time.Second * 10)
+
+	// app transitions to healthy
+	a.healthy.Store(true)
+	// active actors is dependant on the app health checks
+	// hence we cannot check the active actors while the app is unhealthy
```
JoshVanL: This shouldn't be true.

tests/integration/suite/daprd/workflow/apphealth/apphealth.go
```diff
@@ -0,0 +1,124 @@
...
+	assert.EventuallyWithT(t, func(c *assert.CollectT) {
+		assert.GreaterOrEqual(c,
+			len(a.workflow.Dapr().GetMetadata(t, ctx).ActorRuntime.ActiveActors), 2)
+	}, time.Second*10, time.Millisecond*10)
```
JoshVanL:
```suggestion
			require.GreaterOrEqual(t,
			len(a.workflow.Dapr().GetMetadata(t, ctx).ActorRuntime.ActiveActors), 2)
```

JoshVanL (changes requested):
Please add release notes for `v1.16.1`

tests/integration/suite/daprd/workflow/apphealth/apphealth.go
cicoyle: +1, we should not have to use sleep here.

tests/integration/suite/daprd/workflow/apphealth/apphealth.go
lrascao: yeah, that is correct, app won't start off as unhealthy

tests/integration/suite/daprd/workflow/apphealth/apphealth.go
lrascao: yep, this is the critical bit that ensures this is testing the fix. We need to ensure that placement dissemination happens before setting the app back to healthy, the ideal way is to block until placement dissemination but since we didn't know how to do this went with the dirty sleep instead

tests/integration/suite/daprd/workflow/apphealth/apphealth.go
famarting: luis is correct, the comment is just outdated, I will remove it

tests/integration/suite/daprd/workflow/apphealth/apphealth.go
famarting: in fact the sidecar healthz endpoint doesn't change its response if the app health checks start failing, sidecar only depends on app connectivity on startup, but after that the app health checks feature doesn't seem to be impacting the sidecar healthz endpoint

tests/integration/suite/daprd/workflow/apphealth/apphealth.go
famarting: if I start the tests with the app as unhealthy, then dapr fails to succeed in its health probe and logs this 
```
"api error: code = Internal desc = dapr is not ready: [internal-placement-service actors-reminders-scheduler placement-client]"
```

this makes sense I guess, because if an app port is configured we start to assume that the actors system is not getting fully initialized until the app health checks succeed

tests/integration/suite/daprd/workflow/apphealth/apphealth.go
famarting: just change for require? ok, just say it

tests/integration/suite/daprd/workflow/apphealth/apphealth.go
famarting: found the reason for this, easy fix, will include here

tests/integration/suite/daprd/workflow/apphealth/apphealth.go
famarting: maybe we can get rid of this sleep checking the job types watched in the scheduler, instead of checking if the sidecar is aware of the app health, what we were really chasing here is making sure that the scheduler clients were configured as expected

JoshVanL (approved):

Issue #8761, opened 2025-06-05
Support streaming in service to service invocation HTTP api


tomflenner:
## In what area(s)?

/area runtime

> /area operator

> /area placement

> /area docs

> /area test-and-release

## What version of Dapr?

1.15.5 (not sure when it was introduced by I reproduce in this one)

## Expected Behavior

When using Dapr’s service-to-service invocation API over HTTP/1.1, we expect streamed responses (using Transfer-Encoding: chunked) to be delivered progressively to the client, just as they are when the services communicate directly outside of Dapr.

## Actual Behavior

When invoking a service through Dapr’s sidecar, the streamed response is buffered and only delivered after the stream ends.
This breaks expected streaming behavior and renders chunked transfer encoding unusable.

This issue does not occur when invoking the services directly (bypassing Dapr) — in that case, chunks are received progressively.

## Steps to Reproduce the Problem

A minimal repro is provided here, follow the README everything is explained :
🔗 https://github.com/tomflenner/dapr-service-to-service-invocation-http-stream-issue/tree/main

## Additional Notes
Streaming should have been introduced in Dapr 1.12 : https://blog.dapr.io/posts/2023/10/11/dapr-v1.12-is-now-available/#service-invocation-http-streaming-stable-and-enabled-by-default

This might be a regression according to these :
- https://github.com/dapr/dapr/issues/3830#issuecomment-1339841520
- https://github.com/dapr/dapr/issues/6987#issuecomment-1743252031
- https://github.com/dapr/dapr/issues/6571

Thanks @olitomlinson and @WhitWaldo for sharing stuff around this topic !
## Release Note

RELEASE NOTE: FIX Dapr service invocation now correctly handles chunked HTTP streaming without buffering the response.

tomflenner:
With advice from @olitomlinson I've just tested in 1.12.0 and 1.12.5 and I have the same issue, so this use case seems to not be covered by the actual implementation

danadesrosiers:
I'm experiencing this issue as well. I have endpoints that serve image and audio files. When requesting directly, the response streams. When requesting through the sidecar, the response is not delivered until the response is fully loaded into the sidecar.

danadesrosiers:
UPDATE: Responses are streaming in my production environment (Azure Container Apps), but not in my development environment (self-hosted on localhost, v1.14).

olitomlinson:
@danadesrosiers can you tell us what runtime version of dapr is being used in ACA?

[...]

PR #9081, opened 2025-09-26
Add e2e SSE support

javier-aliaga:
# Description

This PR will enable e2e support for SSE.

## Issue reference

Please reference the issue this PR will close: https://github.com/dapr/dapr/issues/8761

## Checklist

- [ ] Code compiles correctly
- [ ] Created/updated tests
- [ ] Unit tests passing


yaron2:
What does this PR intend to fix?

javier-aliaga:
@yaron2 Enable SSE between sidecars. AppA -> daprA -> daprB - AppB

https://github.com/dapr/dapr/issues/8761

yaron2:
To be clear, SSE between sidecars never worked and was never a supported scenario. The chunking/streaming feature listed [here](https://blog.dapr.io/posts/2023/10/11/dapr-v1.12-is-now-available/#service-invocation-http-streaming-stable-and-enabled-by-default) does not include SSE support.

olitomlinson:
I really hope this PR makes it! Full SSE support is desperately needed in Dapr! 🙏🏻

yaron2:
> I really hope this PR makes it! Full SSE support is desperately needed in Dapr! 🙏🏻

This PR's title is misleading. This won't enable full SSE

yaron2:
> > I really hope this PR makes it! Full SSE support is desperately needed in Dapr! 🙏🏻
> 
> This PR's title is misleading. This won't enable full SSE

Actually I take that back, it looks like the scope of some of the changes here in addition to what was added for 1.16 might do it. I'll work with @javier-aliaga to understand fully

olitomlinson:
> Enable SSE between sidecars. AppA -> daprA -> daprB - AppB

@javier-aliaga 

would this extend to the following : 

AppA -> daprA -> daprB -> AppB -> daprB -> daprC -> AppC 

I expect it would, but just wanted to be clear about it :)

javier-aliaga:
@olitomlinson good question, probably yes

javier-aliaga:
@olitomlinson something like this https://github.com/dapr/dapr/pull/9081/commits/abe7bee471779bdecbd243ca6e9ffeb43fbf3013 ?

pkg/api/grpc/daprinternal.go
```diff
@@ -186,12 +188,46 @@ func (a *api) CallLocalStream(stream internalv1pb.ServiceInvocation_CallLocalStr
...
+
+	if res != nil && isSSE {
+		statusOK := res.Status().GetCode() >= 200 && res.Status().GetCode() < 300
+		msg := "no response received from stream"
+		if statusOK {
+			msg = "no expected response from stream"
+		}
+		statusCode = int32(codes.Internal)
+		return status.Errorf(codes.Internal, messages.ErrChannelInvoke, errors.New(msg))
+	}
```
acroca: I think we need to call `res.Close` in this case. Maybe realigning the ifs so the `defer res.Close()` is called?
Something like:
```go
if res == nil
	if isSSE {
		return nil
	}
	statusCode = int32(codes.Internal)
	return status.Errorf(codes.Internal, messages.ErrChannelInvoke, errors.New("no response received from stream"))
}

defer res.Close()

if isSSE {
	statusOK := res.Status().GetCode() >= 200 && res.Status().GetCode() < 300
	msg := "no response received from stream"
	if statusOK {
		msg = "no expected response from stream"
	}
	statusCode = int32(codes.Internal)
	return status.Errorf(codes.Internal, messages.ErrChannelInvoke, errors.New(msg))
}
```

pkg/api/http/directmessaging.go
```diff
@@ -231,14 +231,65 @@ func (a *api) onDirectMessage(w http.ResponseWriter, r *http.Request) {
...
 
-		_, rErr = io.Copy(w, rResp.RawData())
-		if rErr != nil {
-			// Do not return rResp here, we already have a deferred `Close` call on it
-			return nil, backoff.Permanent(rErr)
+		reader := rResp.RawData()
+		flusher, canFlush := w.(http.Flusher)
+
+		accept := r.Header.Get("Accept")
+		isSSE := strings.HasPrefix(accept, "text/event-stream")
```
acroca: Does it have to be a prefix?
```suggestion
		isSSE := strings.Contains(accept, "text/event-stream")
```

pkg/api/http/directmessaging.go
```diff
@@ -231,14 +231,65 @@ func (a *api) onDirectMessage(w http.ResponseWriter, r *http.Request) {
...
+
+		if isSSE && canFlush {
```
acroca: Should it return some sort of error if `isSSE` is true, but `canFlush` is false? This would be captured by the tests though as we are setting a flushable writer for sse, but just in case that code changes in the future.

pkg/channel/http/http_channel.go
```diff
@@ -381,14 +382,19 @@ func (h *Channel) invokeMethodV1(ctx context.Context, req *invokev1.InvokeMethod
...
 		}
 
 		// Send request to user application
 		// (Body is closed below, but linter isn't detecting that)
 		//nolint:bodyclose
 		clientResp, clientErr := h.client.Do(r)
 		if clientResp != nil {
-			if sse && req.HTTPResponseWriter() != nil {
+			statusOK := clientResp.StatusCode >= 200 && clientResp.StatusCode < 300
+			if sse && req.HTTPResponseWriter() != nil && statusOK {
```
acroca: Is the `else` a good code path when sse is active, but status is not 2xx?

tests/integration/suite/daprd/serviceinvocation/http/sserelay.go
```diff
@@ -0,0 +1,320 @@
...
+				now := time.Now().Format(time.RFC3339Nano)
+				payload := fmt.Sprintf("data: {\"n\":%d,\"t\":%q}\n", count, now)
+				s.eventsA.Store(count, payload)
+				_, err := w.Write([]byte(payload))
+				if err != nil {
+					t.Error(err)
+					return
+				}
+
+				log.Printf(">>SRVA>> sent %d bytes: %q\n", len(payload), payload)
```
acroca:
```suggestion
				t.Logf(">>SRVA>> sent %d bytes: %q\n", len(payload), payload)
```

tests/integration/suite/daprd/serviceinvocation/http/sserelay.go
```diff
@@ -0,0 +1,320 @@
...
+
+			for count := range 5 {
+				now := time.Now().Format(time.RFC3339Nano)
+				payload := fmt.Sprintf("data: {\"n\":%d,\"t\":%q}\n", count, now)
+				s.eventsA.Store(count, payload)
+				_, err := w.Write([]byte(payload))
+				if err != nil {
+					t.Error(err)
+					return
+				}
```
acroca:
```suggestion
				require.NoError(t, err)
```

tests/integration/suite/daprd/serviceinvocation/http/sserelay.go
acroca: Also for the rest of the log lines.

tests/integration/suite/daprd/serviceinvocation/http/sserelay.go
```diff
@@ -0,0 +1,320 @@
...
+	s.daprdC = daprd.New(t, daprd.WithAppID("daprdC"), daprd.WithAppPort(s.srvC.Port()))
+
+	return []framework.Option{
+		framework.WithProcesses(s.srvA, s.srvB, s.srvC, s.daprdA, s.daprdB, s.daprdC),
+	}
+}
+
+func (s *sserelay) Run(t *testing.T, ctx context.Context) {
+	s.daprdA.WaitUntilRunning(t, ctx)
+	s.daprdB.WaitUntilRunning(t, ctx)
```
acroca:
```suggestion
	s.daprdA.WaitUntilRunning(t, ctx)
	s.daprdB.WaitUntilRunning(t, ctx)
	s.daprdC.WaitUntilRunning(t, ctx)
```

tests/integration/suite/daprd/serviceinvocation/http/sserelay.go
```diff
@@ -0,0 +1,320 @@
...
+		}
+	}()
+
+	select {
+	case <-done:
+		assert.EqualValues(t, 5, eventCount.Load())
+		log.Printf("done; received %d events", eventCount.Load())
+	case err = <-errChan:
+		require.NoError(t, err)
+	case <-timeoutCtx.Done():
```
acroca:
```suggestion
	case <-time.After(10 * time.Second)
```
This way you don't need to create a context for this

tests/integration/suite/daprd/serviceinvocation/http/sserelay.go
javier-aliaga: We have a lint rule to not use require inside handlers, do you prefer to nolint it?

pkg/channel/http/http_channel.go
javier-aliaga: I think so, in that case we do not want to stream as there is nothing to stream. This way we can propagate there is no error. I will add a test for this

tests/integration/suite/daprd/serviceinvocation/http/sserelay.go
acroca: I see, it's fine then

pkg/api/grpc/daprinternal.go
```diff
@@ -186,18 +188,57 @@ func (a *api) CallLocalStream(stream internalv1pb.ServiceInvocation_CallLocalStr
 		pw.Close()
 	}()
 
+	header := http.Header{}
+	for k, v := range chunk.GetRequest().GetMetadata() {
+		for _, vv := range v.GetValues() {
+			header.Add(k, vv)
+		}
+	}
```
JoshVanL: Move header creation to inside line 200 block.

JoshVanL: Is there no quicker way to build the header than needing to write every key value?

pkg/api/grpc/daprinternal.go
```diff
@@ -186,18 +188,57 @@ func (a *api) CallLocalStream(stream internalv1pb.ServiceInvocation_CallLocalStr
...
+
+	isSSE := false
+	// Or check the Accept header
+	if accept := header.Get("Accept"); strings.Contains(accept, "text/event-stream") {
```
JoshVanL: Why is this  not an exact match?

pkg/api/grpc/daprinternal.go
```diff
@@ -186,18 +188,57 @@ func (a *api) CallLocalStream(stream internalv1pb.ServiceInvocation_CallLocalStr
...
 	res, err := appChannel.InvokeMethod(ctx, req, "")
 	if err != nil {
 		statusCode = int32(codes.Internal)
 		return status.Errorf(codes.Internal, messages.ErrChannelInvoke, err)
 	}
+
 	if res == nil {
+		if isSSE {
+			return nil
+		}
```
JoshVanL: Please can we create a new SSE handler- we have multiple `isSSE` if statements here so makes comprehending the func tricky.

pkg/api/grpc/daprinternal.go
```diff
@@ -186,18 +188,57 @@ func (a *api) CallLocalStream(stream internalv1pb.ServiceInvocation_CallLocalStr
...
+
+	if isSSE {
+		statusOK := res.Status().GetCode() >= 200 && res.Status().GetCode() < 300
+		msg := "no response received from stream"
+		if statusOK {
+			msg = "no expected response from stream"
+		}
+		statusCode = int32(codes.Internal)
+		return status.Errorf(codes.Internal, messages.ErrChannelInvoke, errors.New(msg))
+	}
```
JoshVanL: Why are we always sending an error here?

pkg/api/grpc/daprinternal.go
```diff
@@ -275,6 +316,56 @@ func (a *api) CallLocalStream(stream internalv1pb.ServiceInvocation_CallLocalStr
 	return nil
 }
 
+type loggingResponseWriter struct {
```
JoshVanL: This doesn't seem like the right name.

Please can we move this response writer implementation out of the file. Ideally, a new package.

pkg/api/grpc/daprinternal.go
```diff
@@ -275,6 +316,56 @@ func (a *api) CallLocalStream(stream internalv1pb.ServiceInvocation_CallLocalStr
...
+	stream internalv1pb.ServiceInvocation_CallLocalStreamServer
+	header http.Header
+	buf    []byte
+	seq    uint64
```
JoshVanL: What is the purpose of writing a sequence?

pkg/api/grpc/daprinternal.go
```diff
@@ -275,6 +316,56 @@ func (a *api) CallLocalStream(stream internalv1pb.ServiceInvocation_CallLocalStr
...
+	r := &internalv1pb.InternalInvokeResponse{}
+
+	if lrw.seq == 0 {
+		r = &internalv1pb.InternalInvokeResponse{
+			Headers: internalv1pb.HTTPHeadersToInternalMetadata(lrw.header),
+			Status: &internalv1pb.Status{
+				Code: int32(codes.OK),
+			},
+		}
+	}
```
JoshVanL: Please don't create a new struct, we already have one on 338. Write to the fields instead.

pkg/api/grpc/daprinternal.go
```diff
@@ -275,6 +316,56 @@ func (a *api) CallLocalStream(stream internalv1pb.ServiceInvocation_CallLocalStr
...
+}
+
+func (lrw *loggingResponseWriter) Flush() {
+	r := &internalv1pb.InternalInvokeResponse{}
+
+	if lrw.seq == 0 {
+		r = &internalv1pb.InternalInvokeResponse{
+			Headers: internalv1pb.HTTPHeadersToInternalMetadata(lrw.header),
+			Status: &internalv1pb.Status{
+				Code: int32(codes.OK),
```
JoshVanL: Why is this not `lrw.status`?

pkg/api/grpc/daprinternal.go
```diff
@@ -275,6 +316,56 @@ func (a *api) CallLocalStream(stream internalv1pb.ServiceInvocation_CallLocalStr
...
+
+	err := lrw.stream.Send(&internalv1pb.InternalInvokeResponseStream{
+		Response: r,
+		Payload: &commonv1pb.StreamPayload{
+			Data: lrw.buf,
+			Seq:  lrw.seq,
+		},
+	})
+	if err != nil {
+		panic(err)
```
JoshVanL: Never panic.

pkg/api/grpc/daprinternal.go
```diff
@@ -275,6 +316,56 @@ func (a *api) CallLocalStream(stream internalv1pb.ServiceInvocation_CallLocalStr
...
+	}
+
+	lrw.buf = []byte{}
```
JoshVanL: Don't reallocate memory unnecessarily. 

```suggestion
	lrw.buf = lrw.buf[:0]
```

pkg/api/http/directmessaging.go
```diff
@@ -231,14 +231,65 @@ func (a *api) onDirectMessage(w http.ResponseWriter, r *http.Request) {
...
-		if rErr != nil {
-			// Do not return rResp here, we already have a deferred `Close` call on it
-			return nil, backoff.Permanent(rErr)
+		reader := rResp.RawData()
+		flusher, canFlush := w.(http.Flusher)
+
+		accept := r.Header.Get("Accept")
+		isSSE := strings.Contains(accept, "text/event-stream")
+
+		if isSSE && canFlush {
```
JoshVanL: Move if out of func

pkg/api/http/directmessaging.go
```diff
@@ -231,14 +231,65 @@ func (a *api) onDirectMessage(w http.ResponseWriter, r *http.Request) {
...
+					err := closer.Close()
+					if err != nil {
+						log.Warnf("Failed to close reader: %v", err)
+					}
+				}
+			}()
+
+			for {
+				// Stream SSE data in real-time
+				buf := make([]byte, 1024)
```
JoshVanL: Can this buffer be reused?

pkg/api/http/directmessaging.go
```diff
@@ -231,14 +231,65 @@ func (a *api) onDirectMessage(w http.ResponseWriter, r *http.Request) {
...
+
+				// Check if context is cancelled
+				select {
+				case <-r.Context().Done():
+					rErr = r.Context().Err()
+					return nil, rErr
+				default:
+				}
```
JoshVanL:
```suggestion
 if err = r.Context().Error(); err != nil {
 return err
 }
```

pkg/api/http/directmessaging.go
```diff
@@ -231,14 +231,65 @@ func (a *api) onDirectMessage(w http.ResponseWriter, r *http.Request) {
...
+					// Flush immediately for SSE
+					flusher.Flush()
+				}
+				if err != nil {
+					if err != io.EOF {
+						rErr = err
+					}
+					break
+				}
+			}
```
JoshVanL: We have 3 error variables here 😅 please move to a new func to clean up error catch and return.

pkg/api/http/directmessaging.go
```diff
@@ -231,14 +231,65 @@ func (a *api) onDirectMessage(w http.ResponseWriter, r *http.Request) {
...
+					return nil, rErr
+				default:
+				}
+
+				n, err := reader.Read(buf)
+				if n > 0 {
+					_, writeErr := w.Write(buf[:n])
+					if writeErr != nil {
+						// Client disconnected - break immediately
+						log.Debugf("Client disconnected during streaming: %v", writeErr)
```
JoshVanL: Why is this debug?

pkg/api/http/directmessaging.go
```diff
@@ -231,14 +231,65 @@ func (a *api) onDirectMessage(w http.ResponseWriter, r *http.Request) {
...
+						rErr = writeErr
+						break
+					}
+					// Flush immediately for SSE
```
JoshVanL: Why? Why not when 1024 or EOF?

pkg/api/http/directmessaging.go
```diff
@@ -231,14 +231,65 @@ func (a *api) onDirectMessage(w http.ResponseWriter, r *http.Request) {
...
+					}
+					break
+				}
+			}
+		} else {
+			// Use regular io.Copy for non-streaming responses
+			_, rErr = io.Copy(w, reader)
+			if rErr != nil {
+				// Do not return rResp here, we already have a deferred `Close` call on it
+				return nil, backoff.Permanent(rErr)
```
JoshVanL: We don't have `backoff.Permanent` elsewhere here? 🤔

pkg/channel/http/http_channel.go
```diff
@@ -381,14 +382,19 @@ func (h *Channel) invokeMethodV1(ctx context.Context, req *invokev1.InvokeMethod
 		sse = isSSE(r)
 		if sse {
 			r.Header.Set(headerAccept, mimeEventStream)
+			// Disable buffering for SSE requests
+			r.Header.Set("Cache-Control", "no-cache")
+			// Disable compression
```
JoshVanL: Why??

pkg/channel/http/http_channel.go
```diff
@@ -381,14 +382,19 @@ func (h *Channel) invokeMethodV1(ctx context.Context, req *invokev1.InvokeMethod
 		sse = isSSE(r)
 		if sse {
 			r.Header.Set(headerAccept, mimeEventStream)
+			// Disable buffering for SSE requests
```
JoshVanL: Why??

tests/integration/suite/daprd/serviceinvocation/http/sserelay.go
```diff
@@ -0,0 +1,315 @@
...
+	eventsA sync.Map
+	eventsB sync.Map
+	eventsC sync.Map
+}
+
+func (s *sserelay) Setup(t *testing.T) []framework.Option {
+	s.srvA = app.New(t,
+		app.WithHandlerFunc("/healthz", func(http.ResponseWriter, *http.Request) {}),
+		app.WithHandlerFunc("/TOA", func(w http.ResponseWriter, r *http.Request) {
+			t.Logf(">>SRVA>>%s\n", r.URL.Path)
```
JoshVanL: Remove all test logging.

tests/integration/suite/daprd/serviceinvocation/http/sserelay.go
```diff
@@ -0,0 +1,315 @@
...
+				select {
+				case <-time.After(600 * time.Millisecond):
+				case <-r.Context().Done():
+					if errors.Is(r.Context().Err(), context.Canceled) {
+						t.Logf(">>SRVA>> client closed connection\n")
+					} else {
+						t.Logf(">>SRVA>> client closed connection: %v\n", r.Context().Err())
+					}
+					return
+				}
```
JoshVanL: Let's remove all arbitrary sleeps, they are not necessary.

tests/integration/suite/daprd/serviceinvocation/http/sserelay.go
```diff
@@ -0,0 +1,315 @@
...
+
+	s.srvB = app.New(t,
+		app.WithHandlerFunc("/healthz", func(http.ResponseWriter, *http.Request) {}),
+		app.WithHandlerFunc("/TOB", func(w http.ResponseWriter, r *http.Request) {
+			url := fmt.Sprintf("http://%s/v1.0/invoke/%s/method/TOA", s.daprdB.HTTPAddress(), s.daprdA.AppID())
+			req, err := http.NewRequest(http.MethodGet, url, nil)
+			if err != nil {
+				http.Error(w, err.Error(), http.StatusInternalServerError)
+				return
+			}
```
JoshVanL: Please can we assert these errors instead.

pkg/api/grpc/daprinternal.go
javier-aliaga: Accept header is a list, you can have multiple values.

Accept: text/html, application/xhtml+xml

JoshVanL: `strings.Contains` matches on a sub string. `header.Get` returns the first element from the header value list. Use `header.Values` & `slices.Contains`.

https://pkg.go.dev/strings#Contains
https://pkg.go.dev/net/http#Header.Get
https://pkg.go.dev/slices#Contains

pkg/api/grpc/daprinternal.go
javier-aliaga: in that block `res != nil` so if 
`SSE && res != nil && statusOk` -> in case sse request is handled correctly we do not expect a response as it is being flushed, so this is not expected
`SSE && res != nil && !statusOk` -> this is a normal error and the request could not be handled properly

pkg/api/grpc/daprinternal.go
javier-aliaga: I was thinking about this. If we want to be strict on the Accept header, we should only mark this as an SSE request if the `text/event-stream` is the first on the list.  

`Accept: text/html, application/xhtml+xml, text/event-stream`

So probably this one should not be considered an SSE request. What do you think?

JoshVanL: Fine, but it still needs to be an exact match.

pkg/api/grpc/daprinternal.go
javier-aliaga: I will rename it to `streamResponseWriter`, but I will keep in the same package

javier-aliaga: Also moved to a new file

pkg/api/grpc/daprinternal.go
javier-aliaga: when we flush we build a `commonv1pb.StreamPayload` it requires a sequence to be incremented

pkg/channel/http/http_channel.go
javier-aliaga: I added them because the test were not working, and one suggestion was to disable both caching and compression and it is one of the common causes that breaks SSE. I will remove it and rerun the test again to check we do not need them

javier-aliaga: Removed

pkg/channel/http/http_channel.go
javier-aliaga: https://github.com/dapr/dapr/pull/9081/files#r2409834125

pkg/api/http/directmessaging.go
```diff
@@ -231,14 +231,65 @@ func (a *api) onDirectMessage(w http.ResponseWriter, r *http.Request) {
 
 		w.WriteHeader(int(rResp.Status().GetCode()))
 
-		_, rErr = io.Copy(w, rResp.RawData())
-		if rErr != nil {
-			// Do not return rResp here, we already have a deferred `Close` call on it
-			return nil, backoff.Permanent(rErr)
```
javier-aliaga, line 237: @JoshVanL fyi https://github.com/dapr/dapr/pull/9081/files#r2409832263

pkg/api/http/directmessaging.go
javier-aliaga: this was already there https://github.com/dapr/dapr/pull/9081/files#r2413999803

JoshVanL: Then why are the other errors not permanent?

pkg/api/http/directmessaging.go
javier-aliaga: We are streaming the response here, as we are not consuming the response just piping it, why we care about reading 1024bytes?  we just want to pipe as fast as possible, then the client that uses this stream should read by line.

EOF -> we may not see EOF here as the stream may not be closed

olitomlinson:
Hey all - whats the latest on this PR? Is there a target to get this in for 1.17 ?

pkg/api/grpc/daprinternal.go
javier-aliaga: We cannot move it inside the bolck as we are using it in the if condition in line 198

pkg/api/http/directmessaging.go
javier-aliaga: We have more permanent errors, in line 219. I am not confident to remove it. I would prefer keep it there and check it later

yaron2:
> Hey all - whats the latest on this PR? Is there a target to get this in for 1.17 ?

That's the goal

tests/integration/suite/daprd/serviceinvocation/http/sserelay.go
javier-aliaga: I will reduce it but if the test wants to simulate sending events randomly we cannot remove it, as the events will be sent as fast as possible

tests/integration/suite/daprd/serviceinvocation/http/sserelay.go
javier-aliaga: Do you mean `require.NoError(t, err)`. Is that the case linter complains

pkg/api/http/directmessaging.go
javier-aliaga: refactored to use only one error var

pkg/api/grpc/stream_response_writer.go
```diff
@@ -0,0 +1,60 @@
+package grpc
```
cicoyle: please add the header

pkg/sse/sse.go
```diff
@@ -0,0 +1,46 @@
+package sse
```
cicoyle: please add the header

tests/integration/suite/daprd/serviceinvocation/http/sserelay.go
```diff
@@ -0,0 +1,295 @@
...
+			assert.Equal(t, val, line)
+
+			val, ok = s.eventsC.Load(int(eventCount.Load()))
+			assert.True(t, ok)
+			assert.Equal(t, val, line)
+
+			eventCount.Add(1)
+			// Stop after receiving a certain number of events for testing
+			if eventCount.Load() > 5 {
+				errChan <- errors.New("received too many events")
```
cicoyle: should this return here?
```suggestion
				errChan <- errors.New("received too many events")
				return
```

pkg/api/http/directmessaging.go
```diff
@@ -231,10 +232,23 @@ func (a *api) onDirectMessage(w http.ResponseWriter, r *http.Request) {
...
+				// Do not return rResp here, we already have a deferred `Close` call on it
+				return nil, backoff.Permanent(rErr)
+			}
+
+			return nil, nil
+		}
+
+		err := sse.FlushSSEResponse(r.Context(), w, reader)
+		if err != nil {
+			return nil, err
```
JoshVanL: Why is this error not `backoff.Permanent`?

JoshVanL (changes requested):

pkg/api/http/directmessaging.go
javier-aliaga: Probably it should

javier-aliaga: Changed

javier-aliaga:
Moving this PR to target master instead of 1.16 branch

https://github.com/dapr/dapr/pull/9149

PR #9084, opened 2025-09-26
refresh dns entries everytime we connect

cicoyle:
Fix issue where if there are 3 placement instances, and placement-server-1 is leader but goes down for whatever reason, dapr was only trying to connect to placement servers 0 & 2 leading to issues where it never connected to the leader.


viswapriya22:
Hi @cicoyle! Does this fix need to be brought into 1.16 as well?

viswapriya22:
@cicoyle following up on the above comment in case it was missed among recent PRs

javier-aliaga:
@cicoyle can we close this one? I do not think we need it

cicoyle:
This was fixed by another PR, so closing

PR #9109, opened 2025-10-08
Fix actor not found after continue as new

filintod:
# Description

I was having issues with workflow when exeucting a simple workflow with an activity and after continue as new it would start failing

## Issue reference

## Checklist

- [ ] Code compiles correctly
- [ ] Created/updated tests
- [ ] Unit tests passing

### DIFF
--- a/pkg/actors/targets/workflow/activity/execute.go
+++ b/pkg/actors/targets/workflow/activity/execute.go
@@ -112,7 +112,10 @@ func (a *activity) executeActivity(ctx context.Context, name string, taskEvent *
 	// send completed event to orchestrator wf actor
 	wfActorType := a.workflowActorType
 	if router := taskEvent.GetRouter(); router != nil {
-		wfActorType = a.actorTypeBuilder.Workflow(router.GetSourceAppID())
+		src := router.GetSourceAppID()
+		if src != "" {
+			wfActorType = a.actorTypeBuilder.Workflow(src)
+		}
 	}
 
 	req := internalsv1pb.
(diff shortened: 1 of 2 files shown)


yaron2 (approved):

JoshVanL:
@filintod please as an integration test for this change. Also we need release notes.

pkg/actors/targets/workflow/activity/execute.go
```diff
@@ -112,7 +112,10 @@ func (a *activity) executeActivity(ctx context.Context, name string, taskEvent *
 	// send completed event to orchestrator wf actor
 	wfActorType := a.workflowActorType
 	if router := taskEvent.GetRouter(); router != nil {
-		wfActorType = a.actorTypeBuilder.Workflow(router.GetSourceAppID())
```
JoshVanL, line 115: It should _not_ be the case that `SourceAppID` is an empty string. Clearly there is something wrong inside of durabletask, which is where this fix should be. It cannot be the case that both `router` is not nil, and this value is empty.

filintod: yes, I'm trying to see where the root cause is at, like was it changed in python durable task, or go durable task, but when testing from python directly against go durable task (no placement/scheduler) it works fine, but when testing against dapr proper (with placement/scheduler) it fails, and why I try to find issues in dapr instead of durable task.

JoshVanL: I think this is the problem

```
        router = pb.TaskRouter()
        router.sourceAppID = self._app_id
        if app_id is not None:
            router.targetAppID = app_id
```

filintod: actually found the bug, it was not the above, we were not adding router when executing continue as new in the set_complete action.  

testing against durabletask-go, seems to be completely independent of router information or not caring, and why it was passing against it

will close this pr

filintod:
root cause was a python durabletask bug missing passing router information on continue as new

PR #9117, opened 2025-10-10
Fix placement dissemination during high churn events

JoshVanL:
Blocked Placement dissemination during high churn

Problem

Placement would fail to ever, or very slowly, disseminate the actor table in high daprd churn scenarios.

Impact

Actors or workflows would fail to be activated, and existing actors or workflows would fail.

Root Cause

Placement used a "small" (100) queue size which when exhausted would cause a deadlock. Placement would also wait for a fully consumed channel queue before disseminating slowing down the dissemination process.

Solution

Increase the queue size to 10000 and change the dissemination logic to not wait for a fully consumed queue before disseminating.

### DIFF
--- a/pkg/placement/membership.go
+++ b/pkg/placement/membership.go
@@ -34,9 +34,6 @@ func (p *Service) membershipChangeWorker(ctx context.Context) {
 	baselineHeartbeatTimestamp := p.clock.Now().UnixNano()
 	log.Infof("Baseline membership heartbeat timestamp: %v", baselineHeartbeatTimestamp)
 
-	disseminateTimer := p.clock.NewTicker(disseminateTimerInterval)
-	defer disseminateTimer.Stop()
-
 	// Reset memberUpdateCount to zero for every namespace when leadership is acquired.
 	p.streamConnPool.forEachNamespace(func(ns string, _ map[uint32]*daprdStream) {
 		val, _ := p.memberUpdateCount.GetOrSet(ns, &atomic.Uint32{})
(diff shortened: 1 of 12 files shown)


pkg/placement/membership.go
```diff
@@ -170,13 +138,20 @@ func (p *Service) processMembershipCommands(ctx context.Context) {
...
+							return
+						}
+						cc := cnt.Load()
+						if cc == 0 {
+							return
+						}
+						log.Debugf("Add raft.TableDisseminate to membershipCh. memberUpdateCountTotal count for namespace %s: %d", op.host.Namespace, cc)
+						log.Debugf("dissemination running for namespace %s", op.host.Namespace)
+						select {
+						case p.membershipCh <- hostMemberChange{cmdType: raft.TableDisseminate, host: raft.DaprHostMember{Namespace: op.host.Namespace}}:
```
famarting: this push should be done in a different gouroutine or sent through a different queue or something, I'm thinking it is not safe to produce and consume to the same channel in the same loop

tests/integration/suite/placement/dissemination/race.go
```diff
@@ -0,0 +1,245 @@
...
+func (r *race) Run(t *testing.T, ctx context.Context) {
+	r.place.WaitUntilRunning(t, ctx)
+
+	require.EventuallyWithT(t, func(c *assert.CollectT) {
+		metrics := r.place.Metrics(c, ctx)
+		assert.Len(c, metrics.MatchMetric("dapr_placement_leader_status"), 1)
+		assert.Len(c, metrics.MatchMetric("dapr_placement_raft_leader_status"), 1)
+	}, time.Second*15, time.Millisecond*10)
+
+	for i := range 110 {
```
famarting: you can make this 1000, I tried locally and it should be fine

tests/integration/suite/placement/dissemination/race.go
```diff
@@ -0,0 +1,245 @@
...
+			Name:      fmt.Sprintf("inithost%d", i),
+			Namespace: fmt.Sprintf("initns%d", i),
+			Port:      1231,
+			Entities:  []string{fmt.Sprintf("initactor%d", i)},
+			Id:        fmt.Sprintf("initapp%d", i),
+			ApiLevel:  uint32(20),
+		}
+		r.simulateHost(t, ctx, host)
+	}
+	time.Sleep(1 * time.Second)
```
famarting: if you increase the previous loop to 1000, I would put here something like 3seconds, just to give more time for the load to build up

tests/integration/suite/placement/dissemination/race.go
```diff
@@ -0,0 +1,245 @@
...
+
+	t.Run("register and unregister in loop", func(t *testing.T) {
```
famarting: rename this, sorry, this test was different originally

PR #9121, opened 2025-10-10
Delete state store reminders

JoshVanL:
The Dapr Scheduler service was introduced in v1.14 as a new control
plane service, with one of its primary goals to specifically replace the
state store actor reminders with a more performant and scalable
implementation. The actor reminder implementation by
Scheduler became the defauilt in v1.15, and the actor reminder store
implementation being deprecated since. Now going into v1.17, it is time
to completely remove the state store implementation- it is slow,
requires some not so nice code to stay around, and shouldn't be used
today.

Remove all actor reminder state store code, references, feature gate,
tests.


cmd/daprd/options/options.go
```diff
@@ -174,8 +174,12 @@ func New(origArgs []string) (*Options, error) {
 	fs.StringVar(&placementServiceHostAddr, "placement-host-address", "", "Addresses for Dapr Actor Placement servers (overrides actors-service)")
 	fs.StringSliceVar(&opts.SchedulerAddress, "scheduler-host-address", nil, "Addresses of the Scheduler service instance(s), as comma separated host:port pairs")
 	fs.UintVar(&opts.SchedulerJobStreams, "scheduler-job-streams", 3, "The number of active job streams to connect to the Scheduler service")
-	fs.StringVar(&opts.ActorsService, "actors-service", "", "Type and address of the actors service, in the format 'type:address'")
+
+	// DEPRECATED.
 	fs.StringVar(&opts.RemindersService, "reminders-service", "", "Type and address of the reminders service, in the format 'type:address'")
+	fs.StringVar(&opts.ActorsService, "actors-service", "", "Type and address of the actors service, in the format 'type:address'")
+	fs.MarkHidden("reminders-service")
+	fs.MarkHidden("actors-service")
```
acroca, line 182: Should we show a deprecation warning if the user configures those fields?

pkg/actors/api/period.go
```diff
@@ -24,6 +24,8 @@ import (
 
 // ReminderPeriod contains the parsed period for a reminder.
 //
+// TODO: @joshvanl: remove this.
+//
```
acroca, line 28: Why isn't this removed in this PR?

pkg/actors/targets/workflow/activity/reminder.go
```diff
@@ -45,8 +37,7 @@ func (a *activity) createReminder(ctx context.Context, his *backend.HistoryEvent
 		ActorID:   a.actorID,
 		DueTime:   "0s",
 		Name:      reminderName,
-		Period:    period,
-		IsOneShot: oneshot,
+		IsOneShot: true,
```
acroca, line 40: This seems to be `true` in both cases it's used. Can we delete this property from `CreateReminderRequest`?

pkg/config/app_configuration.go
```diff
@@ -19,10 +19,12 @@ type ApplicationConfig struct {
...
-	DrainOngoingCallTimeout    string           `json:"drainOngoingCallTimeout"`
-	DrainRebalancedActors      bool             `json:"drainRebalancedActors"`
-	Reentrancy                 ReentrancyConfig `json:"reentrancy,omitempty"`
-	RemindersStoragePartitions int              `json:"remindersStoragePartitions"`
+	DrainOngoingCallTimeout string           `json:"drainOngoingCallTimeout"`
+	DrainRebalancedActors   bool             `json:"drainRebalancedActors"`
+	Reentrancy              ReentrancyConfig `json:"reentrancy,omitempty"`
+
+	// DEPRECATED.
+	RemindersStoragePartitions int `json:"remindersStoragePartitions"`
```
acroca, line 27: Should this show a warning message if used?

tests/e2e/actor_metadata/actor_metadata_test.go
```diff
@@ -1,286 +0,0 @@
...
-				"TEST_APP_ACTOR_TYPE":                 actorName,
-			},
-		},
-	}
-
-	tr = runner.NewTestRunner(testRunnerName, testApps, nil, nil)
-	os.Exit(tr.Start(m))
-}
-
-func TestActorMetadataEtagRace(t *testing.T) {
```
acroca, line 155: Why is this deleted? Sounds unrelated to the change of this PR

tests/e2e/actor_reminder_partition/actor_reminder_partition_test.go
```diff
@@ -1,319 +0,0 @@
...
-			count++
-		}
-	}
-
-	return count
-}
-
-var tr *runner.TestRunner
-
-func TestMain(m *testing.M) {
```
acroca, line 93: Just to be sure, reminder partitioning is not a thing anymore correct?

pkg/actors/targets/workflow/orchestrator/invoke.go
```diff
@@ -107,30 +106,18 @@ func (o *orchestrator) handleReminder(ctx context.Context, reminder *actorapi.Re
...
-			return err
-		}
-		return nil
+		return err
 	default: // Other error
 		log.Errorf("Workflow actor '%s': execution failed with an error: %v", o.actorID, err)
-		if o.schedulerReminders {
-			return err
-		}
-		return actorerrors.ErrReminderCanceled
```
cicoyle, line 134: Can you delete the following as well from this file: `pkg/actors/errors/errors.go` - I dont see it removed in this PR, but I dont think we need it anymore now that we are deleting state store reminders.

```
// ErrReminderCanceled is returned when the reminder has been canceled.
var ErrReminderCanceled = errors.New("reminder has been canceled")
```

pkg/config/configuration.go
```diff
@@ -73,9 +69,7 @@ const (
 	ActionPolicyGlobal  = "global"
 )
 
-var defaultFeatures = map[Feature]bool{
```
cicoyle, line 76: We will also need to update the docs to remove this preview feature.

pkg/runtime/scheduler/internal/cluster/streamer.go
```diff
@@ -138,6 +138,8 @@ func (s *streamer) handleJob(ctx context.Context, job *schedulerv1pb.WatchJobsRe
 		}
 
 		if errors.Is(err, actorerrors.ErrReminderCanceled) {
```
cicoyle, line 140: I dont think we return `actorerrors.ErrReminderCanceled` anymore with the changes from this PR, so I think this conditional check can be removed.

cmd/daprd/options/options.go
JoshVanL: :+1: done

pkg/actors/api/period.go
JoshVanL: To no balloon the PR size- would require a larger refactor of the actor router, timer, etc.

pkg/actors/targets/workflow/activity/reminder.go
JoshVanL: It is not true when being created from the actor HTTP & gRPC APIs.

tests/e2e/actor_metadata/actor_metadata_test.go
JoshVanL: These tests rely on state store reminder.

tests/e2e/actor_reminder_partition/actor_reminder_partition_test.go
JoshVanL: Correct.

pkg/actors/targets/workflow/orchestrator/invoke.go
JoshVanL:
```
		if errors.Is(err, actorerrors.ErrReminderCanceled) {
			// TODO: @joshvanl add support for cancelling a job with a new
			// WatchJobsRequestResultStatus_CANCELED.
			return schedulerv1pb.WatchJobsRequestResultStatus_SUCCESS
		}
```

pkg/config/configuration.go
JoshVanL: :+1:

pkg/runtime/scheduler/internal/cluster/streamer.go
JoshVanL: We do from the app

```
	// Allow stopping a recurring reminder or timer
	if v := res.GetHeaders()["X-Daprremindercancel"]; v != nil && len(v.GetValues()) > 0 && strings.IsTruthy(v.GetValues()[0]) {
		return res, actorerrors.ErrReminderCanceled
	}
```

PR #9126, opened 2025-10-13
Blocked Placement dissemination with high Scheduler dataset

JoshVanL:
Problem

Disseminations would hang for long periods of time when the Scheduler dataset was large.

Impact

Dissemination could take up to hours to complete, causing reminders to not be delivered for a long period of time.

Root Cause

The reminder migration of state store to scheduler reminders does a full decoded scan of the Scheduler database, which would take a long time if there were many entries. During this time the dissemination would be blocked.

Solution

Limit the maximum time spent doing the migration to 3 seconds.


JoshVanL:
> Should target 1.16?

One merged, we will cherry pick

PR #9133, opened 2025-10-20
Fix panic during actor deactivation

JoshVanL:
Problem

Daprd could panic during actor deactivation.

Impact

Daprd sidecar would crash, resulting in downtime for the application.

Root Cause

A race in the actor lock cached memory release and claiming logic meant a stale lock could be used during deactivation, double closing it, and causing a panic.

Solution

Tie the lock's lifecycle to the actor's lifecycle, ensuring the lock is only released when the actor is fully deactivated, and claimed with the actor itself.

### DIFF
--- a/tests/integration/framework/process/scheduler/etcd/etcd.go
+++ b/tests/integration/framework/process/scheduler/etcd/etcd.go
@@ -30,7 +30,7 @@ import (
 )
 
 type Etcd struct {
-	fp        *ports.Ports
+	fp        []*ports.Ports
 	configs   []*embed.Config
 	etcds     []*embed.Etcd
 	endpoints []string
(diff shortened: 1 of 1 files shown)


pkg/actors/targets/app/app.go
```diff
@@ -54,9 +55,12 @@ type app struct {
 
 	lock  *lock.Lock
 	clock clock.Clock
+	wg    sync.WaitGroup
```
acroca: I'm not sure I understand the logic for this wg. The `GetOrCreate` increments the wg, and the invokes call the Done. What happens if we call Invoke more than once? that would call the `done` multiple times, but only was incremented once.

pkg/actors/targets/app/factory_test.go
```diff
@@ -143,21 +169,28 @@ func Test_Idle(t *testing.T) {
...
+	act.InvokeMethod(t.Context(), new(internalv1pb.InternalInvokeRequest))
 	assert.Equal(t, 7, mapLen(ff))
 
 	assert.Eventually(t, clock.HasWaiters, time.Second*10, time.Millisecond*10)
 
 	clock.Step(time.Second * 5)
 	assert.True(t, clock.HasWaiters())
 	assert.Equal(t, 7, mapLen(ff))
-	clock.Step(time.Second * 5)
+	clock.Step(time.Second*5 + 1)
```
acroca: Why?

JoshVanL: Not needed, removed

pkg/actors/targets/app/app.go
JoshVanL: Removed, please take another look.

pkg/actors/targets/app/factory.go
```diff
@@ -99,30 +93,32 @@ func New(opts Options) targets.Factory {
...
 	a, ok := f.table.Load(actorID)
 	if !ok {
-		newApp := f.initApp(appCache.Get(), actorID)
-		var loaded bool
-		a, loaded = f.table.LoadOrStore(actorID, newApp)
-		if loaded {
-			appCache.Put(newApp)
-		}
+		newApp := f.initApp(actorID)
+		a, _ = f.table.LoadOrStore(actorID, newApp)
```
acroca, line 102: The mutex is locked for reading, two readers might get here at the same time. Is that fine?

pkg/actors/targets/app/app.go
```diff
@@ -223,9 +225,12 @@ func (a *app) InvokeTimer(ctx context.Context, reminder *api.Reminder) error {
 }
 
 func (a *app) Deactivate(ctx context.Context) error {
-	defer a.table.Delete(a.actorID)
+	if !a.closed.CompareAndSwap(false, true) {
+		return nil
+	}
```
acroca, line 230: Any reason to use this over `sync.Once`?

pkg/actors/targets/app/factory.go
JoshVanL: `table` is a `sync.Map` so safe on multi thread read/writes by design.

pkg/actors/targets/app/app.go
JoshVanL: I find `sync.Once` ugly, & it would mean we would need to wrap the entire contents of this function in it, or use a wrapper func .

PR #9142, opened 2025-10-23
Propagate APP_API_TOKEN in gRPC metadata for subscriptions, bindings, and jobs

cicoyle:
Fixes gRPC pub/sub subscriptions, input bindings, and scheduler job callbacks to include the `APP_API_TOKEN` in metadata when invoking the app.

### DIFF
--- a/pkg/messaging/grpc_proxy.go
+++ b/pkg/messaging/grpc_proxy.go
@@ -31,6 +31,7 @@ import (
 	invokev1 "github.com/dapr/dapr/pkg/messaging/v1"
 	"github.com/dapr/dapr/pkg/proto/common/v1"
 	"github.com/dapr/dapr/pkg/resiliency"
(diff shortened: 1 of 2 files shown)


pkg/channel/grpc/grpc_channel.go
```diff
@@ -236,3 +237,12 @@ func (g *Channel) HealthProbe(ctx context.Context) (*apphealth.Status, error) {
 func (g *Channel) SetAppHealth(ah *apphealth.AppHealth) {
 	g.appHealth = ah
 }
+
+// AddAppTokenToContext adds the APP_API_TOKEN to the outgoing gRPC ctx
+func AddAppTokenToContext(ctx context.Context) context.Context {
+	appMetadataToken := security.GetAppToken()
```
JoshVanL: This is doing a Getenv lookup on every request. We should cache this value instead.

Issue #9018, opened 2025-08-28
Scheduler doesn't explicitly handle scenarios when workflow state is deleted


atrauzzi:
## In what area(s)?

/area scheduler

## What version of Dapr?

Current `1.15.x` **and** latest `1.16.x` RCs.

## Expected Behavior

Graceful handling around the plausible (for one reason or another) scenario where systems lose the state behind a workflow.

This probably just comes in the form of correctly logging and then cleaning up any corresponding state in scheduler.

## Actual Behavior

Scheduler retries any workflows with lost state...forever. 🧟‍♂️ 

## Steps to Reproduce the Problem

A comprehensive repro can be found here: https://github.com/landonnewberry/dapr-workflow-not-found-repro

## Release Note

RELEASE NOTE:

olitomlinson:
Closely related https://github.com/dapr/dapr/issues/8987

nelson-parente:
Fixed on 1.16.2

Pending https://github.com/dapr/dapr/pull/9157 to master

PR #9157, opened 2025-10-30
Prevent infinite loop when workflow state is corrupted or destroyed

JoshVanL:
Fixes https://github.com/dapr/dapr/issues/9018

Problem

Dapr workflows could enter an infinite reminder loop when the workflow state in the actor state store is corrupted or destroyed.

Impact

Dapr workflows would enter an infinite loop of reminder calls.

Root Cause

When a workflow reminder is triggered, the workflow state is loaded from the actor state store. If the state is corrupted or destroyed, the workflow would not be able to progress and would keep re-triggering the same reminder indefinitely.

Solution

Do not retry the reminder if the workflow state cannot be loaded, and instead log an error and exit the workflow execution.

### DIFF
--- a/pkg/runtime/subscription/subscription.go
+++ b/pkg/runtime/subscription/subscription.go
@@ -23,6 +23,7 @@ import (
 	"sync/atomic"
 	"time"
 
+	contribContenttype "github.com/dapr/components-contrib/contenttype"
 	"github.com/dapr/components-contrib/metadata"
 	contribpubsub "github.com/dapr/components-contrib/pubsub"
 	"github.com/dapr/dapr/pkg/api/grpc/manager"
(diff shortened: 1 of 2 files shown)


pkg/actors/targets/workflow/activity/execute.go
```diff
@@ -124,6 +124,12 @@ func (a *activity) executeActivity(ctx context.Context, name string, taskEvent *
 	_, err = a.router.Call(ctx, req)
 	switch {
 	case err != nil:
+		if strings.HasSuffix(err.Error(), api.ErrInstanceNotFound.Error()) {
```
acroca, line 127: Can't we use `errors.Is`?
```suggestion
		if errors.Is(err, api.ErrInstanceNotFound) {
```

cicoyle (commented):
Code lgtm, but CI looks to be a bit red. mind taking a look?

pkg/actors/targets/workflow/activity/execute.go
JoshVanL: No because the error may be coming from a remote gRPC call which clobbers the type.

cicoyle (approved):

PR #9158, opened 2025-10-30
[1.16] Prevent infinite loop when workflow state is corrupted or destroyed

JoshVanL:
Problem

Dapr workflows could enter an infinite reminder loop when the workflow state in the actor state store is corrupted or destroyed.

Impact

Dapr workflows would enter an infinite loop of reminder calls.

Root Cause

When a workflow reminder is triggered, the workflow state is loaded from the actor state store. If the state is corrupted or destroyed, the workflow would not be able to progress and would keep re-triggering the same reminder indefinitely.

Solution

Do not retry the reminder if the workflow state cannot be loaded, and instead log an error and exit the workflow execution.

### DIFF
--- a/tests/integration/suite/daprd/workflow/delstate.go
+++ b/tests/integration/suite/daprd/workflow/delstate.go
@@ -126,15 +126,15 @@ func (d *delstate) Run(t *testing.T, ctx context.Context) {
 			"dapr_runtime_actor_reminders_fired_total|actor_type:dapr.internal.default.%s.activity|app_id:%s|success:true",
 			d.daprd2.AppID(), d.daprd2.AppID())]
 		assert.InDelta(c, 1.0, v, 0)
+
+		v = d.daprd2.Metrics(t, ctx).All()[fmt.Sprintf(
+			"dapr_runtime_actor_reminders_fired_total|actor_type:dapr.internal.default.%s.activity|app_id:%s|success:false",
+			d.daprd2.AppID(), d.daprd2.AppID())]
+		assert.InDelta(c, 0.0, v, 0)
+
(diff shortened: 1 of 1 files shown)


cicoyle (changes requested):
changes lgtm, just linter and dep review are red and need fixing

pkg/actors/targets/workflow/activity/execute.go
```diff
@@ -124,6 +124,12 @@ func (a *activity) executeActivity(ctx context.Context, name string, taskEvent *
 	_, err = a.router.Call(ctx, req)
 	switch {
 	case err != nil:
+		if strings.HasSuffix(err.Error(), api.ErrInstanceNotFound.Error()) {
```
filintod, line 127: would `errors.Is` not work here? this comparison against a suffix seems a bit strange and potentially brittle.  It seemed to have been used before like in https://github.com/dapr/dapr/blob/ac944977c6ce31e343b6b6ff5249d2fb36b18cc6/pkg/runtime/wfengine/client.go#L103 (2y old so not sure how good it is though)

JoshVanL: https://github.com/dapr/dapr/pull/9157#discussion_r2478365375 
:slightly_smiling_face:

filintod: should it be `contains`? just thinking that we are trusting this to only be found at the end, maybe that is the case. I guess we don't use NOT_FOUND status code in durabletask.

But I guess it is fine and hopefully the tests capture it if we diverge later on

JoshVanL: Yes I hear you- if we want to do that it would be part of a larger change.

We do indeed have test coverage for this

PR #9170, opened 2025-11-05
Implements durabletask `ListInstanceIDs` & `GetInstanceHistory`

JoshVanL:
Today, there is no ways of discovering the list of workflow instances that are currently running or have completed in the past without using external storage queries. The Dapr CLI [introduced list and workflow history commands](https://github.com/dapr/cli/pull/1560) to get information about running and completed workflows, however these commands rely on direct queries to the underlying storage provider. By introducing this functionality into the durabletask framework itself, these commands need only talk to Daprd, removing the requirement for direct access to the storage provider as well as authentication. Daprd can make these queries itself, and use the Actor State Store component to access the underlying storage.

Implements the new durabletask APIs according to https://github.com/dapr/proposals/pull/93

### DIFF
--- a/pkg/runtime/wfengine/backends/actors/actors.go
+++ b/pkg/runtime/wfengine/backends/actors/actors.go
@@ -43,7 +43,9 @@ import (
 	invokev1 "github.com/dapr/dapr/pkg/messaging/v1"
 	internalsv1pb "github.com/dapr/dapr/pkg/proto/internals/v1"
 	"github.com/dapr/dapr/pkg/resiliency"
+	"github.com/dapr/dapr/pkg/runtime/compstore"
 	"github.com/dapr/dapr/pkg/runtime/wfengine/state"
+	"github.com/dapr/dapr/pkg/runtime/wfengine/state/list"
 	"github.com/dapr/dapr/pkg/runtime/wfengine/todo"
 	"github.com/dapr/dapr/utils"
 	"github.com/dapr/durabletask-go/api"
@@ -62,15 +64,15 @@ const (
 	WorkflowNameLabelKey = "workflow"
 	ActivityNameLabelKey = "activity"
 	ExecutorNameLabelKey = "executor"
-	ActorTypePrefix      = "dapr.internal."
 )
 
 type Options struct {
-	AppID      string
-	Namespace  string
-	Actors     actors.Interface
-	Resiliency resiliency.Provider
-	EventSink  orchestrator.EventSink
+	AppID          string
+	Namespace      string
+	Actors         actors.Interface
+	Resiliency     resiliency.Provider
+	EventSink      orchestrator.EventSink
+	ComponentStore *compstore.ComponentStore
 	// experimental feature
 	// enabling this will use the cluster tasks backend for pending tasks, instead of the default local implementation
 	// the cluster tasks backend uses actors to share the state of pending tasks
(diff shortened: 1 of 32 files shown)


pkg/runtime/wfengine/backends/actors/actors.go
```diff
@@ -479,7 +484,7 @@ func (abe *Actors) WatchOrchestrationRuntimeStatus(ctx context.Context, id api.I
 }
 
 // PurgeOrchestrationState deletes all saved state for the specific orchestration instance.
-func (abe *Actors) PurgeOrchestrationState(ctx context.Context, id api.InstanceID) error {
+func (abe *Actors) PurgeOrchestrationState(ctx context.Context, id api.InstanceID, force bool) error {
```
javier-aliaga, line 488: Are we using `force` param?

JoshVanL: We are in [this PR](https://github.com/dapr/dapr/pull/9176) but the durabletask-go commit we are pointing to has this in the signature.

pkg/runtime/wfengine/state/list/list.go
```diff
@@ -0,0 +1,76 @@
...
+	"errors"
+	"fmt"
+	"strings"
+
+	"github.com/dapr/components-contrib/state"
+	"github.com/dapr/dapr/pkg/runtime/compstore"
+)
+
+type ListOptions struct {
+	ComponentStore *compstore.ComponentStore
```
javier-aliaga, line 28: Could we pass here only the StateStore? Why we need all componentStore?

JoshVanL: It is subject to change because of hot reloading. `ComponentStore` is the canonical struct to reference ingested components in the system.

pkg/runtime/wfengine/state/list/list.go
```diff
@@ -0,0 +1,76 @@
...
+	if !ok {
+		return nil, errors.New("no state store with actor support found")
+	}
+
+	ks, ok := store.(state.KeysLiker)
+	if !ok {
+		return nil, fmt.Errorf("state store %T does not support listing keys", store)
+	}
+
+	like := opts.AppID + "||dapr.internal." + opts.Namespace + "." + opts.AppID + ".workflow||%||metadata"
```
cicoyle: Can we switch to use the `ActorTypePrefix` const in `backend/actors/actors`?

tests/integration/suite/daprd/workflow/list/multiple.go
```diff
@@ -0,0 +1,84 @@
+/*
```
cicoyle, line 1: typo nit: can you rename the file `multiple`

tests/integration/suite/daprd/workflow/list/multipule.go
```diff
@@ -0,0 +1,84 @@
...
+
+	reg := dworkflow.NewRegistry()
+	reg.AddWorkflowN("foo", func(ctx *dworkflow.WorkflowContext) (any, error) {
+		return nil, nil
+	})
+
+	const numWf = 5
+
+	wfs := make([]*dworkflow.Client, 0, numWf)
+	for i := range 5 {
```
cicoyle:
```suggestion
	for i := range numWf {
```

cicoyle (changes requested):
lgtm, just a few comments from me

nelson-parente:

JoshVanL:
Thanks @cicoyle, PTAL!

PR #9171, opened 2025-11-05
Workflow: Delete associated reminders on workflow purge

JoshVanL:
Upates the Workflow Purge API to also delete all associated reminders when a workflow is purged. This ensures that no orphaned reminders remain after a workflow is deleted.

To implement this, a new scheduler RPC method `DeleteByMetadata` has been added, which behaves as a bulk delete on a particular job target type. Upon purge, the workflow actor will delete all workflow & activity reminders by invoking this new scheduler RPC.

`DeleteByMetadata` optionally allows deleting ids as a prefix, enabling deleting all workflow activity reminders which have unbounded suffixes as the actor ID.

### DIFF
--- a/tests/integration/suite/daprd/workflow/purge/reminders.go
+++ b/tests/integration/suite/daprd/workflow/purge/reminders.go
@@ -47,7 +47,7 @@ func (r *reminders) Setup(t *testing.T) []framework.Option {
 func (r *reminders) Run(t *testing.T, ctx context.Context) {
 	r.workflow.WaitUntilRunning(t, ctx)
 
-	var here atomic.Bool
+	var inActivity atomic.Bool
 	releaseCh := make(chan struct{})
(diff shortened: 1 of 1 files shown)


msfussell:
@JoshVanL Need to ensure that this behavior is described in the docs.

tests/integration/suite/daprd/workflow/purge/reminders.go
```diff
@@ -0,0 +1,78 @@
...
+
+	return []framework.Option{
+		framework.WithProcesses(r.workflow),
+	}
+}
+
+func (r *reminders) Run(t *testing.T, ctx context.Context) {
+	r.workflow.WaitUntilRunning(t, ctx)
+
+	var here atomic.Bool
```
cicoyle: nit, can this be named `activityStarted`?
```suggestion
	var activityStarted atomic.Bool
```

cicoyle (changes requested):
mostly lgtm, just a minor comment then we can merge

nelson-parente:

PR #9176, opened 2025-11-07
Workflow: Purge Force

JoshVanL:
Implements force purging as from https://github.com/dapr/durabletask-protobuf/pull/20

Force will force a purge of a workflow, regardless of its current runtime state, or whether an active worker can process it, the backend will attempt to delete it anyway. This neccessarily means the purging is executed out side of the workflow state machine, and therefore, can lead to corrupt state or broken workflow execution. Usage of this should only be used when the client knows the workflow is not being currently processed. It is highly recommended to avoid using this flag unless absolutely necessary.

### DIFF
--- a/tests/integration/suite/daprd/workflow/purge/force/reminders.go
+++ b/tests/integration/suite/daprd/workflow/purge/force/reminders.go
@@ -70,7 +70,7 @@ func (r *reminders) Run(t *testing.T, ctx context.Context) {
 	assert.GreaterOrEqual(t, count, 5)
 
 	assert.EventuallyWithT(t, func(c *assert.CollectT) {
-		assert.Len(t, r.workflow.Scheduler().ListAllKeys(t, ctx, "dapr/jobs"), 1)
+		assert.Len(c, r.workflow.Scheduler().ListAllKeys(t, ctx, "dapr/jobs"), 1)
 	}, time.Second*10, time.Millisecond*10)
 
 	require.NoError(t, client.PurgeWorkflowState(ctx, id, dworkflow.WithForcePurge(true)))


pkg/runtime/wfengine/backends/actors/actors.go
```diff
@@ -629,3 +627,62 @@ func (abe *Actors) WaitForActivityCompletion(ctx context.Context, request *proto
...
+		return err
+	}
+
+	err = astate.TransactionalStateOperation(ctx, true, req, false)
+	if err != nil {
+		return err
+	}
+
+	// TODO: @joshvanl: Delete all associated reminders	using once merged
+	// https://github.com/dapr/dapr/pull/9171
```
cicoyle: PR is merged 👍🏻

tests/integration/suite/daprd/workflow/purge/force/midexecution.go
```diff
@@ -0,0 +1,84 @@
...
+
+	return []framework.Option{
+		framework.WithProcesses(m.workflow),
+	}
+}
+
+func (m *midexecution) Run(t *testing.T, ctx context.Context) {
+	m.workflow.WaitUntilRunning(t, ctx)
+
+	var here atomic.Bool
```
cicoyle: Can this be `activityStarted` to be very explicit about where 'here' is?

```suggestion
	var activityStarted atomic.Bool
```

tests/integration/suite/daprd/workflow/purge/force/remote.go
```diff
@@ -0,0 +1,104 @@
...
+	workflow *workflow.Workflow
+}
+
+func (r *remote) Setup(t *testing.T) []framework.Option {
+	uuid, err := uuid.NewRandom()
+	require.NoError(t, err)
+
+	r.workflow = workflow.New(t,
+		workflow.WithDaprds(3),
+		workflow.WithDaprdOptions(0, daprd.WithAppID(uuid.String())),
```
cicoyle, line 46: Why do we need to pass in a uuid for the appID here? I thought the framework already did this for us?

nelson-parente:

tests/integration/suite/daprd/workflow/purge/force/remote.go
JoshVanL: The workflow process does not make all the daprds the same app ID

JoshVanL:
Thanks @cicoyle, PTAL!

tests/integration/suite/daprd/workflow/purge/force/remote.go
cicoyle: oh duh, yes. ok resolving this.

PR #9179, opened 2025-11-10
Workflow: schedule reminders to first history

JoshVanL:
Currently, all reminders work task reminders are sheduled with a due time of `Now()`. The consequence of this is that processing a large number of workflows concurrently leads to workflow execution favouring breadth over depth.

PR updates the reminder scheduling logic to schedule reminders based on the first history event of the workflow. This ensures that reminders are scheduled in the order that workflows were created, leading to a more depth first workflow execution pattern.

PR #9185, opened 2025-11-12
Workflow: State Retention Policy

JoshVanL:
Branched from https://github.com/dapr/dapr/pull/9179

Implementation as described in https://github.com/dapr/proposals/pull/95

PR adds a new workflow configuration spec `stateRetentionPolicy` that allows
users to define a retention policy for workflow state. This policy
determines how long workflows in a terminal state (completed, failed, or
terminated) should retain their state before being purged from the actor
state store.

Users can define a base `anyTerminal` retention duration that applies to
all terminal states, as well as specific retention durations for each
terminal state (`completed`, `failed`, `terminated`). The most specific
policy takes precedence, allowing for granular control over state
retention. The retention fields take a Go duration string, which
represents the state TTL (time-to-live) from the moment the workflow
enters the terminal state.

Once a workflow reaches a terminal state, the configuration policy is
checked, and if so, a corresponding retention actor reminder is created
with a due time based on the defined retention duration. The reminder is
created against a new `retentioner` actor type. This type takes the
typical form of `dapr.internal.<namespace>.<appid>.retentioner`. Like
the orchestrator actor, the retentioner actor ID is the instance ID of
the workflow.

Upon the reminder firing, the retentioner actor calls the `purge` method
on the workflow orchestrator actor with the same actor ID (workflow
instance ID). The orchestrator then purges its state from the actor
state store.

The retentioner reminder is purged from the scheduler in the event that
the workflow is manually purged before the retention period elapses.


pkg/runtime/wfengine/todo/todo.go
```diff
@@ -30,6 +30,9 @@ const (
 	WaitForRuntimeStatus         = "WaitForRuntimeStatus"
 	ForkWorkflowHistory          = "ForkWorkflowHistory"
 	RerunWorkflowInstance        = "RerunWorkflowInstance"
+
+	MetadataActivityRemidnerDueTime = "dueTime"
```
acroca:
```suggestion
	MetadataActivityReminderDueTime = "dueTime"
```

pkg/actors/targets/workflow/activity/invoke.go
```diff
@@ -17,13 +17,16 @@ import (
...
+	"strconv"
+	"time"
 
 	"google.golang.org/protobuf/proto"
 
 	actorapi "github.com/dapr/dapr/pkg/actors/api"
 	invokev1 "github.com/dapr/dapr/pkg/messaging/v1"
 	internalsv1pb "github.com/dapr/dapr/pkg/proto/internals/v1"
 	wferrors "github.com/dapr/dapr/pkg/runtime/wfengine/errors"
+	"github.com/dapr/dapr/pkg/runtime/wfengine/todo"
```
acroca: What does `todo` mean here?

pkg/actors/targets/workflow/activity/invoke.go
```diff
@@ -32,7 +35,18 @@ import (
...
+	dueTime := time.Now()
+	if s, ok := req.GetMetadata()[todo.MetadataActivityRemidnerDueTime]; ok && len(s.GetValues()) > 0 {
+		unix, err := strconv.ParseInt(s.GetValues()[0], 10, 64)
+		if err != nil {
+			return nil, err
+		}
+		dueTime = time.UnixMilli(unix)
+	}
+
+	log.Debugf("Activity actor '%s': invoking method '%s'", a.actorID, method)
```
acroca, line 49: Should we add the `dueTime` here? It sounds like it's invoking it now.

pkg/actors/targets/workflow/orchestrator/activity.go
```diff
@@ -70,6 +79,9 @@ func (o *orchestrator) callActivity(ctx context.Context, e *backend.HistoryEvent
 	_, err = o.router.Call(ctx, internalsv1pb.
 		NewInternalInvokeRequest("Execute").
 		WithActor(activityActorType, targetActorID).
+		WithMetadata(map[string][]string{
+			todo.MetadataActivityRemidnerDueTime: {strconv.FormatInt(dueTime.UnixMilli(), 10)},
+		}).
```
acroca, line 84: Why are we using metadata instead of adding a new attribute?

pkg/actors/targets/workflow/orchestrator/run.go
```diff
@@ -210,8 +211,8 @@ func (o *orchestrator) runWorkflow(ctx context.Context, reminder *actorapi.Remin
 	}
 
 	if runtimestate.IsCompleted(rs) {
-		log.Infof("Workflow Actor '%s': workflow completed with status '%s' workflowName '%s'", o.actorID, runtimestate.RuntimeStatus(rs).String(), workflowName)
-		return todo.RunCompletedTrue, nil
+		log.Infof("Workflow Actor '%s': workflow completed with status '%s' workflowName '%s'", o.actorID, rstatus, workflowName)
+		return todo.RunCompletedTrue, o.handleRetention(ctx, rstatus)
```
acroca: I find this might be confusing for the caller, as it might get an `err` due to the `handleRetention` failing, but still get the run completed. I'm wondering if an error in the creation of the retention reminder should be a reason to not make the workflow completed and make it try again...

pkg/actors/targets/workflow/orchestrator/state.go
```diff
@@ -111,22 +113,53 @@ func (o *orchestrator) saveInternalState(ctx context.Context, state *wfenginesta
...
+				ActorID:         o.actorID,
+				MatchIDAsPrefix: false,
+			})
+		}()
+	}
+
+	errs := make([]error, 0, n)
+	for range n {
+		errs = append(errs, <-errCh)
+	}
```
acroca: This way of running many things in parallel and join all the errors looks like something we could move to `dapr/kit` and make it similar to how `errgroup` works, but returning all errors joined instead of just the first one, wdyt?

tests/integration/suite/daprd/workflow/purge/retention/config/match/completed.go
```diff
@@ -0,0 +1,126 @@
...
+		assert.EventuallyWithT(t, func(c *assert.CollectT) {
+			require.NoError(t, db.QueryRowContext(ctx, "SELECT COUNT(*) FROM "+tableName).Scan(&count))
+			assert.Equal(c, 8, count)
+		}, time.Second*10, time.Millisecond*10)
+
+		require.NoError(t, client.TerminateWorkflow(ctx, id))
+		time.Sleep(time.Second * 3)
+
+		require.NoError(t, db.QueryRowContext(ctx, "SELECT COUNT(*) FROM "+tableName).Scan(&count))
+		assert.Equal(t, 11, count)
```
acroca: What about putting this in a `assert.EventuallyWithT(t, func(c *assert.CollectT) {` with a max time of 3 seconds?

tests/integration/suite/daprd/workflow/purge/retention/config/match/terminated.go
```diff
@@ -0,0 +1,126 @@
...
+      terminated: "0s"
+`)),
+	)
+
+	return []framework.Option{
+		framework.WithProcesses(e.workflow),
+	}
+}
+
+func (e *terminated) Run(t *testing.T, ctx context.Context) {
```
acroca, line 61: I think it's worth adding a note of what this test is about, it took me some time to figure out it's about the configuration having both `anyTerminal` and `terminated`.

tests/integration/suite/daprd/workflow/purge/retention/config/match/terminated.go
```diff
@@ -0,0 +1,126 @@
...
+			assert.Equal(c, 8, count)
+		}, time.Second*10, time.Millisecond*10)
+
+		require.NoError(t, client.TerminateWorkflow(ctx, id))
+
+		assert.EventuallyWithT(t, func(c *assert.CollectT) {
+			require.NoError(t, db.QueryRowContext(ctx, "SELECT COUNT(*) FROM "+tableName).Scan(&count))
+			assert.Equal(c, 0, count)
+			assert.Empty(c, e.workflow.Scheduler().ListAllKeys(t, ctx, "dapr/jobs"))
+		}, time.Second*5, time.Millisecond*10)
```
acroca: I'd be optimistinc and give it 4 seconds to make sure it doesn't hit the 5s from `anyTerminal`...

tests/integration/suite/daprd/workflow/purge/retention/config/completed.go
```diff
@@ -0,0 +1,92 @@
...
+      completed: "1s"
+`)),
+	)
+
+	return []framework.Option{
+		framework.WithProcesses(e.workflow),
+	}
+}
+
+func (e *completed) Run(t *testing.T, ctx context.Context) {
```
acroca, line 57: I think it'd be good to also test what would happen with workflows terminating in a different state than `completed`, for example one that ends up in `terminated` or `failed`. Same with the other 2 states.

pkg/actors/targets/workflow/activity/invoke.go
JoshVanL: This is an invoke here

pkg/actors/targets/workflow/orchestrator/state.go
JoshVanL: Done! The problem with using `errgroup` is that it returns the first error which is not good enough. https://pkg.go.dev/golang.org/x/sync/errgroup#Group.Wait 

https://github.com/dapr/kit/pull/139

pkg/actors/targets/workflow/orchestrator/activity.go
JoshVanL: Not sure what you mean by `attribute` in this case?

pkg/actors/targets/workflow/activity/invoke.go
JoshVanL: This is a invovation not a reminder

JoshVanL:
Thanks @acroca, PTAL!

go.mod
```diff
@@ -525,3 +525,5 @@ replace (
 //
 // Then, run `make modtidy-all` in this repository.
 // This ensures that go.mod and go.sum are up-to-date for each go.mod file.
+
+replace github.com/dapr/kit => github.com/joshvanl/kit v0.0.0-20251114131157-e2e1932e525d
```
acroca, line 529: Needs a release in dapr/kit

pkg/actors/targets/workflow/orchestrator/activity.go
acroca: I mean adding a new property to `InternalInvokeRequest` instead of using metadata.

go.mod
cicoyle: I believe the PR is merged now, so we can just ref the new merge commit instead of your fork

pkg/actors/targets/workflow/orchestrator/state.go
```diff
@@ -111,22 +113,43 @@ func (o *orchestrator) saveInternalState(ctx context.Context, state *wfenginesta
...
+			return o.reminders.DeleteByActorID(ctx, &actorsapi.DeleteRemindersByActorIDRequest{
+				ActorType:       o.activityActorType,
+				ActorID:         o.actorID + "::",
+				MatchIDAsPrefix: true,
+			})
+		},
+	}
+
+	runners = append(runners, func(ctx context.Context) error {
+		return o.reminders.DeleteByActorID(ctx, &actorsapi.DeleteRemindersByActorIDRequest{
```
cicoyle: Why not add this `retentionactortype` deletion func in the above slice for runners?

pkg/actors/targets/workflow/retentioner/retentioner.go
```diff
@@ -0,0 +1,100 @@
...
+
+	return nil
+}
+
+func (r *retentioner) InvokeTimer(ctx context.Context, reminder *actorapi.Reminder) error {
+	return errors.New("timers are not implemented")
+}
+
+func (r *retentioner) Deactivate(context.Context) error {
+	retentionerCache.Put(r)
```
cicoyle, line 79: Should this be a delete?
```suggestion
	retentionerCache.Delete(r)
```

pkg/actors/targets/workflow/workflow.go
```diff
@@ -16,21 +16,66 @@ package workflow
...
+		{
+			Factory: orchFactory,
+			Type:    opts.WorkflowActorType,
+		},
+		{
+			Factory: activityFactory,
+			Type:    opts.ActivityActorType,
+		},
+		{
+			Factory: retentionerFactory,
```
cicoyle, line 64: Should the `retentionerFactory` be added every time or only if the config is set?

pkg/runtime/wfengine/backends/actors/actors.go
```diff
@@ -209,7 +216,11 @@ func (abe *Actors) UnRegisterActors(ctx context.Context) error {
 		return err
 	}
 
-	actorTypes := []string{abe.workflowActorType, abe.activityActorType}
+	actorTypes := []string{
+		abe.workflowActorType,
+		abe.activityActorType,
+		abe.retentionerActorType,
```
cicoyle, line 229: same question here as above - do we want this every time or only if config is set?

nelson-parente:
Related to #9190

pkg/actors/targets/workflow/orchestrator/state.go
JoshVanL: Missed an `if` statement. fixed.

pkg/actors/targets/workflow/retentioner/retentioner.go
JoshVanL: `Put` is correct- we are putting this struct onto the retentioner cache.

pkg/actors/targets/workflow/workflow.go
JoshVanL: Yes because the config can change over time- but we still want to service any retention reminders which may have been created in the past.

pkg/runtime/wfengine/backends/actors/actors.go
JoshVanL: See above

JoshVanL:
Thanks @cicoyle, PTAL!

PR #9200, opened 2025-11-20
proto: refactor runtime `dapr.proto` into multiple files

JoshVanL:
The dapr runtime `dapr.proto` file has grown to be unruly and difficult to navigate. Change breaks up the single file into multiple files based on API group.

### DIFF
--- a/docs/development/setup-dapr-development-env.md
+++ b/docs/development/setup-dapr-development-env.md
@@ -27,7 +27,7 @@ This document helps you get started developing Dapr. If you find any problems wh
 
 2. Install [Delve](https://github.com/go-delve/delve/tree/master/Documentation/installation) for Go debugging, if desired.
 
-3. Install [golangci-lint](https://golangci-lint.run/usage/install) version 1.64.6.
+3. Install [golangci-lint](https://golangci-lint.run/docs/welcome/install/) version 1.64.6.
 
 ## Setup a Kubernetes development environment
 


JoshVanL:
@cicoyle I'm not recalling that being a problem.. that would be surprising to me!

dapr/proto/runtime/v1/metadata.proto
```diff
@@ -0,0 +1,146 @@
...
+  string path = 2;
+}
+
+message SetMetadataRequest {
+  string key = 1;
+  string value = 2;
+}
+
+// ShutdownRequest is the request for Shutdown.
+message ShutdownRequest {
```
cicoyle: should this live in `dapr.proto`?

dapr/proto/runtime/v1/README.md
```diff
@@ -1,10 +1,5 @@
 # Public Dapr Runtime APIs
 
-This folder is intended for user-facing APIs.
```
cicoyle, line 3: Can you tweak this to say something like the following:

```
The `dapr/proto/runtime/v1` folder contains a proto file per the user-facing Dapr APIs.
```

pkg/proto/scheduler/v1/scheduler.pb.go
```diff
@@ -1,7 +1,7 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
 // 	protoc-gen-go v1.32.0
-// 	protoc        v6.30.2
+// 	protoc        v4.25.4
```
cicoyle: why lower the version here?

pkg/proto/scheduler/v1/scheduler_grpc.pb.go
```diff
@@ -1,7 +1,7 @@
 // Code generated by protoc-gen-go-grpc. DO NOT EDIT.
 // versions:
 // - protoc-gen-go-grpc v1.3.0
-// - protoc             v6.30.2
+// - protoc             v4.25.4
```
cicoyle: same here - why lower the version?

mikeee:
> I appreciate the refactor 🎉
> 
> I do recall us not wanting to proceed with this previously due to the impact involving changes to **all** SDKs. Was this considered with the refactor?

It was more the fact that the PR that included this change in the last cycle would have left _very_ little time to make the required changes to the SDKs.

PR #9202, opened 2025-11-20
Actors: Reminder Failure Policy

JoshVanL:
Branched from https://github.com/dapr/dapr/pull/9200

Expose the exiting Jobs Failure Poicy to Actor Reminders, allowing users
to configure how failed reminders are handled. The default failure
policy still stands as a constant retry 3 times with a 1 second
interval.

The workflow orchestration and activity reminders remain unchanged,
being "one shot", constant retry forever.

Exposes failure policy on gRPC API.

### DIFF
--- a/tests/apps/perf/workflowsapp/Dockerfile
+++ b/tests/apps/perf/workflowsapp/Dockerfile
@@ -1,10 +1,21 @@
 
-FROM python:3.8-slim-buster
+FROM python:3.9-slim-buster
 
 WORKDIR /app
 
 COPY . .
 
-RUN pip3 install -r requirements.txt
+RUN pip3 install --no-cache-dir -r requirements.txt gunicorn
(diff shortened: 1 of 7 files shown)


cicoyle (approved):

alicejgibbons:
@JoshVanL can we add a docs pr for this? 

IE. I think [here](https://docs.dapr.io/developing-applications/building-blocks/actors/actors-timers-reminders/#actor-reminders) we could add that the default failure policy is constant retry 3 times with a 1 second interval. 

Also [here](https://docs.dapr.io/operations/resiliency/policies/retries/override-default-retries/) we could either fix or update this table. Assuming this is no longer true? or is it that this is the resiliency to retry the saving the reminder to the state store and what you updated is the retry to actually fire the reminder

image

PR #9203, opened 2025-11-20
Workflow: Adds Rerun parent instance ID

JoshVanL:
Updates the workflow orchestrator to populate the history with the parent instance ID when a workflow is rerun. This enhancement allows for better tracking and correlation of workflow executions, especially in complex scenarios involving multiple reruns.

The two events which are now populated with the parent instance ID are `TimerCreatedEvent` and `TaskScheduledEvent`.

Depends on https://github.com/dapr/durabletask-go/pull/54

### DIFF
--- a/tests/integration/framework/process/otel/collector.go
+++ b/tests/integration/framework/process/otel/collector.go
@@ -15,19 +15,29 @@ package otel
 
 import (
 	"context"
+	"encoding/hex"
+	"fmt"
 	"net"
+	"sort"
 	"sync"
 	"testing"
 	"time"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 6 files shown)


go.mod
```diff
@@ -526,3 +526,5 @@ replace (
 //
 // Then, run `make modtidy-all` in this repository.
 // This ensures that go.mod and go.sum are up-to-date for each go.mod file.
+
+replace github.com/dapr/durabletask-go => github.com/joshvanl/durabletask-go v0.0.0-20251120190650-49029f414b40
```
cicoyle: we should be able to update this :)

JoshVanL: done!

PR #9205, opened 2025-11-21
Scheduler: Namespaces loop

JoshVanL:
Creates a new control loop 'namespaces' which is used to managing the
connection pool for each namespace. This loop enables further
implementation of broadcast jobs which require fanning out triggering
across a single namespace set. The namespace scoping has been removed
from the connections loop stream store.

Uses the kit & go-etcd-cron for the kit update to improve memory
allocations and lock contention.
https://github.com/diagridio/go-etcd-cron/pull/101
https://github.com/dapr/kit/pull/140

Branched from https://github.com/dapr/dapr/pull/9200

Does not change behaviour.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -14,7 +14,7 @@ require (
 	github.com/cloudevents/sdk-go/v2 v2.15.2
 	github.com/coreos/go-oidc/v3 v3.14.1
 	github.com/dapr/components-contrib v1.16.2-0.20251125171221-9b3a3905c4dc
-	github.com/dapr/durabletask-go v0.10.2-0.20251124230813-f3afee228ca1
+	github.com/dapr/durabletask-go v0.10.2-0.20251203182905-2c611fb434fd
 	github.com/dapr/kit v0.16.2-0.20251124175541-3ac186dff64d
 	github.com/diagridio/go-etcd-cron v0.10.1-0.20251125140549-44f1c29e6505
 	github.com/evanphx/json-patch/v5 v5.9.0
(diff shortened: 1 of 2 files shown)


pkg/scheduler/server/internal/pool/loops/stream/add.go
```diff
@@ -0,0 +1,60 @@
...
+	if len(ts) == 0 || slices.Contains(ts, schedulerv1pb.JobTargetType_JOB_TARGET_TYPE_ACTOR_REMINDER) {
+		for _, actorType := range add.GetActorTypes() {
+			log.Infof("Adding a Sidecar connection to Scheduler for actor type: %s/%s.", reqNamespace, actorType)
+			prefixes = append(prefixes, "actorreminder||"+reqNamespace+"||"+actorType+"||")
+		}
+
+		actorTypes = add.GetActorTypes()
+	}
+
+	log.Infof("Adding a Sidecar connection to Scheduler for app: %s/%s (actorTypes=%v).", reqNamespace, reqAppID, actorTypes)
```
cicoyle: There are quite a few log lines in here now. I think we can consolidate into 1 or 2. One like:

```
log.Infof("Sidecar connected: namespace=%s appID=%s actorTypes=%v prefixes=%v",
        reqNamespace, reqAppID, actorTypes, prefixes)
```

Then after the `DeliverablePrefixes()` we could have:
```
log.Debugf("Sidecar prefixes registered for %s/%s", reqNamespace, reqAppID)
```

pkg/scheduler/server/internal/pool/loops/stream/stream.go
```diff
@@ -97,7 +105,7 @@ func (s *stream) Handle(ctx context.Context, event loops.Event) error {
 	case *loops.StreamShutdown:
 		s.handleShutdown()
 	default:
-		return fmt.Errorf("unknown stream event type: %T", e)
+		panic(fmt.Sprintf("unknown stream event type: %T", e))
```
cicoyle, line 108: was the panic intentional instead of the return err?

cicoyle (commented):
2 things from me, appreciate the refactor 🎉

pkg/scheduler/server/internal/pool/loops/stream/stream.go
JoshVanL: Yes, as this will only happen due to programming error, rather than runtime.

PR #9220, opened 2025-12-02
[1.16] Workflow logging loop when no pending task completed

JoshVanL:
Problem

When Daprd is made unhealthy during a workflow execution, pending activity tasks which are made completed will result in looping logs from daprd.

Impact

Daprd will continually print log messages indicating that activity task result has been completed for no pending tasks.

Root Cause

Daprd holds a streaming connection to Schedulers which handles job execution for the Jobs API, Actor Reminders, and workflow execution. Each stream established has a single set of types which the client supports. When the app reports as unhealthy, the stream to Schedulers need to be re-established as daprd no longer supports the Jobs API and Actor Reminders while the app is unhealthy. This restarts the workflow runtime, which clears all pending activity tasks. Resulting task completed from the previous execution are then received with no pending tasks, causing an internal error. This error is intentionally retried indefinitely, resulting in a logging loop.

Solution

The error occurring from no pending tasks is now typed as a non-retryable error, preventing the logging loop.

### DIFF
--- a/.github/workflows/dapr-perf-components.yml
+++ b/.github/workflows/dapr-perf-components.yml
@@ -37,7 +37,7 @@ env:
   # Configure proxy for Go modules
   GOPROXY: https://proxy.golang.org
   # Version of kubectl
-  KUBECTLVER: "v1.30.0"
+  KUBECTLVER: "v1.34.0"
   # Version of Helm
   HELMVER: "v3.10.0"
   # Kubernetes namespace to use
--- a/.github/workflows/dapr-perf.yml
+++ b/.github/workflows/dapr-perf.yml
@@ -43,7 +43,7 @@ env:
   # Configure proxy for Go modules
   GOPROXY: https://proxy.golang.org
   # Version of kubectl
-  KUBECTLVER: "v1.30.0"
+  KUBECTLVER: "v1.34.0"
   # Version of Helm
   HELMVER: "v3.10.0"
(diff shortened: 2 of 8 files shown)


tests/integration/suite/daprd/workflow/unhealthy.go
```diff
@@ -0,0 +1,139 @@
...
+		return nil, nil
+	})
+
+	client := u.workflow.BackendClient(t, ctx)
+
+	id, err := client.ScheduleNewOrchestration(ctx, "bar", api.WithInstanceID("unhealthy-test"))
+	require.NoError(t, err)
+
+	assert.EventuallyWithT(t, func(c *assert.CollectT) {
+		assert.Equal(t, int64(n+1), inActivity.Load())
```
cicoyle: Should this be c?
```suggestion
		assert.Equal(c, int64(n+1), inActivity.Load())
```

go.mod
```diff
@@ -526,3 +526,5 @@ replace (
 //
 // Then, run `make modtidy-all` in this repository.
 // This ensures that go.mod and go.sum are up-to-date for each go.mod file.
+
+replace github.com/dapr/durabletask-go => github.com/joshvanl/durabletask-go v0.0.0-20251202174419-ae2ec1e44e41
```
cicoyle: just merged

cicoyle: just released 👍🏻

PR #9221, opened 2025-12-02
Scheduler: worker pool

JoshVanL:
Workers is the number of workers that handle job events. The higher the number
the more go routines will be spawned, each working over a partition of the
total job space. The higher the number, the higher the number of jobs which
can be concurrently delivered to runtimes. Increasing this number increases
the number of go routines for this instance. This number should be tuned to
the bottleneck of job execution, i.e., CPU, I/O, memory, etc.

Defaults to 128 workers.

`--workers=N` flag for the scheduler binary.

`dapr_scheduler.workers=N` in the Helm chart.

Branched from https://github.com/dapr/dapr/pull/9205

Uses https://github.com/diagridio/go-etcd-cron/pull/103


go.mod
```diff
@@ -527,3 +527,5 @@ replace (
 //
 // Then, run `make modtidy-all` in this repository.
 // This ensures that go.mod and go.sum are up-to-date for each go.mod file.
+
+replace github.com/diagridio/go-etcd-cron => github.com/joshvanl/go-etcd-cron v0.0.0-20251125173513-7e3740c9e201
```
cicoyle: twas merged, so can be updated

cicoyle:
I merged in the other PR and now see conflicts. Will do a final pass once conflicts are resolved 👍🏻 thx!

JoshVanL:
Thanks @cicoyle, PTAL!

cicoyle (approved):

PR #9240, opened 2025-12-12
Workflow: Rerun from Child Workflow Creation

JoshVanL:
Uses fork from https://github.com/dapr/durabletask-go/pull/60

Adds support for rerunning workflows starting from the creation of a child workflow. The new instance will be created, along with the child workflow targeted with a new instance ID.

Workflows can now only be rerun from a root (no parent) workflow.

Child workflorw reruns will generate an instance ID based on the `<parent-id>:<eventID>` pattern to ensure uniqueness. All in-flight child workflows at the time of rerun, are also rerun with this generated instance ID.

Are able to specify new input for the child workflow being rerun. Able to specify the instance ID to use for the new child workflow created.


go.mod
```diff
@@ -522,3 +522,5 @@ replace (
 //
 // Then, run `make modtidy-all` in this repository.
 // This ensures that go.mod and go.sum are up-to-date for each go.mod file.
+
+replace github.com/dapr/durabletask-go => github.com/joshvanl/durabletask-go v0.0.0-20251212130010-ea043c7dece1
```
cicoyle: we should be able to rm this now.

PR #9243, opened 2025-12-15
Integration: Adds IO intensive test option

JoshVanL:
Adds an optional integration test "IOIntensive" that can be enabled via the framework options. When enabled, this test will no run on Github Actions. This option is reserved for tests which complete quickly on local machines but may time out on Github Actions due to IO constraints, i.e. the test which performs 1030 workflows, which takes milliseconds on a decent laptop, but fails to 45s timeout on Github Actions.


nelson-parente (approved):

cicoyle (commented):
Can you check this PR again @JoshVanL - it seems to be missing the `WithIOIntensive` bits - I only see a few IT test fixes in this PR.

JoshVanL:
@cicoyle that commit was added in a separate PR. Will close this PR for now.

PR #8196, opened 2024-10-14
wip: add daprd job metrics

titanventura:
# Description

Adds job metrics to daprd metrics endpoint
* success jobs count
* failure jobs count
* job latency

## Issue reference

Please reference the issue this PR will close: #_[issue number]_

## Checklist

Please make sure you've completed the relevant tasks for this PR, out of the following list:

- [x] Code compiles correctly
- [ ] Created/updated tests
- [ ] Unit tests passing
- [ ] End-to-end tests passing
- [ ] Extended the documentation / Created issue in the <https://github.com/dapr/docs/> repo: dapr/docs#_[issue number]_
- [ ] Specification has been updated / Created issue in the <https://github.com/dapr/docs/> repo: dapr/docs#_[issue number]_
- [ ] Provided sample for the feature / Created issue in the <https://github.com/dapr/docs/> repo: dapr/docs#_[issue number]_

PR #9256, opened 2026-01-02
Add daprd job metrics

### LINKED ISSUE #8160: Jobs API Metrics
## In what area(s)?

/area runtime

> /area operator

> /area placement

> /area docs

> /area test-and-release

## Describe the feature

From Daprd metrics endpoint we should add metrics for the Jobs API including the following:
- Total job latency (start to finish)
- Count failed jobs
- Count successful jobs
- Next time job is scheduled to trigger
- Last time job was triggered

## Release Note

RELEASE NOTE: **ADD** Jobs API Metrics.

JeffreyJPZ:
# Description
Exposes the following metrics for the Jobs API via the daprd metrics endpoint:
- ```dapr_component_job_success_count``` The number of successful job triggers
- ```dapr_component_job_failure_count``` The number of failed job triggers
- ```dapr_component_job_latencies``` The latency of the response from the app that processed the triggered job

## Issue reference

Closes #8160, closes #8196 (which this PR is based on)

## Checklist

- [x] Code compiles correctly
- [x] Created/updated tests
- [x] Unit tests passing


pkg/diagnostics/component_monitoring.go
```diff
@@ -203,6 +223,9 @@ func (c *componentMetrics) Init(meter view.Meter, appID, namespace string, laten
 		diagUtils.NewMeasureView(c.conversationCount, []tag.Key{appIDKey, componentKey, namespaceKey, successKey}, view.Count()),
 		diagUtils.NewMeasureView(c.cryptoLatency, []tag.Key{appIDKey, componentKey, namespaceKey, operationKey, successKey}, latencyDistribution),
 		diagUtils.NewMeasureView(c.cryptoCount, []tag.Key{appIDKey, componentKey, namespaceKey, operationKey, successKey}, view.Count()),
+		diagUtils.NewMeasureView(c.jobLatency, []tag.Key{appIDKey, namespaceKey, operationKey, successKey}, latencyDistribution),
+		diagUtils.NewMeasureView(c.jobSuccessCount, []tag.Key{appIDKey, namespaceKey, operationKey}, view.Count()),
+		diagUtils.NewMeasureView(c.jobFailureCount, []tag.Key{appIDKey, namespaceKey, operationKey}, view.Count()),
```
JeffreyJPZ: Removed component as a label here since it seems to have been overlooked (from [this discussion](https://github.com/dapr/dapr/pull/8196#discussion_r1801860543))

pkg/runtime/scheduler/internal/cluster/streamer.go
```diff
@@ -186,6 +190,7 @@ func (s *streamer) invokeApp(ctx context.Context, job *schedulerv1pb.WatchJobsRe
 	switch codes.Code(statusCode) {
 	case codes.OK:
 		log.Debugf("Sent job %s to app", job.GetName())
+		diag.DefaultComponentMonitoring.JobTriggeredSuccess(ctx, diag.JobTriggerOp, elapsed)
 		return nil
 	case codes.NotFound:
 		log.Errorf("non-retriable error returned from app while processing triggered job %s. status code returned: %v", job.GetName(), statusCode)
```
JeffreyJPZ, line 197: git blame tells us that a non-retriable app error should be considered a success from the scheduler's view, but would this be considered a successful or failed trigger (or neither)?

pkg/diagnostics/component_monitoring.go
```diff
@@ -437,6 +460,34 @@ func (c *componentMetrics) CryptoInvoked(ctx context.Context, component, operati
...
+
+		if elapsed > 0 {
+			stats.RecordWithOptions(
+				ctx,
+				stats.WithRecorder(c.meter),
+				stats.WithTags(diagUtils.WithTags(c.jobLatency.Name(), appIDKey, c.appID, namespaceKey, c.namespace, operationKey, operation, successKey, strconv.FormatBool(success))...),
+				stats.WithMeasurements(c.jobLatency.M(elapsed)))
+		}
+	}
+}
```
JeffreyJPZ, line 489: Changed the implementation of ```jobTriggered``` to use ```stats.RecordWithOptions```, otherwise it uses the default global meter (which causes integration tests to fail).

pkg/diagnostics/component_monitoring.go
JeffreyJPZ, line 228: Removed ```componentKey``` here since it seems to have been overlooked? (from https://github.com/dapr/dapr/pull/8196#discussion_r1801860543). Also removed ```successKey``` for ```jobSuccessCount``` and ```jobFailureCount``` as the metric name makes it redundant (mentioned here https://github.com/dapr/dapr/pull/8196#discussion_r1805329582).

pkg/diagnostics/component_monitoring.go
```diff
@@ -169,6 +174,21 @@ func newComponentMetrics() *componentMetrics {
...
+		jobFailureCount: stats.Int64(
+			"component/job/failure_count",
+			"The number of failed job triggers.",
+			stats.UnitDimensionless,
+		),
+		jobLatency: stats.Float64(
+			"component/job/latencies",
+			"The latency of the response from the app that processed the triggered job.",
+			stats.UnitMilliseconds,
+		),
```
JeffreyJPZ, line 191: Reworded the descriptions so they match other component metrics (new to Dapr so correct me if I'm wrong).

pkg/runtime/scheduler/internal/cluster/streamer.go
cicoyle: yes, the `NotFound` is considered a success from scheduler's perspective bc it is non-retriable. Treating it as a failure would result in incorrect retriggers.

cicoyle (dismissed):
lgtm - thank you! 🎉

pkg/diagnostics/component_monitoring.go
JoshVanL, line 489:
```suggestion
func (c *componentMetrics) JobTriggeredSuccess(ctx context.Context, operation string, elapsed *float64) {
	c.jobTriggered(ctx, operation, true, elapsed, c.jobSuccessCount)
}

// JobTriggeredFailure records the metrics for a job execution failure event.
func (c *componentMetrics) JobTriggeredFailure(ctx context.Context, operation string, elapsed *float64) {
	c.jobTriggered(ctx, operation, false, elapsed, c.jobFailureCount)
}

func (c *componentMetrics) jobTriggered(ctx context.Context, operation string, success bool, elapsed *float64, countMeasure *stats.Int64Measure) {
	if c.enabled {
		stats.RecordWithOptions(
			ctx,
			stats.WithRecorder(c.meter),
			stats.WithTags(diagUtils.WithTags(countMeasure.Name(), appIDKey, c.appID, namespaceKey, c.namespace, operationKey, operation)...),
			stats.WithMeasurements(countMeasure.M(1)))

		if elapsed != nil {
			stats.RecordWithOptions(
				ctx,
				stats.WithRecorder(c.meter),
				stats.WithTags(diagUtils.WithTags(c.jobLatency.Name(), appIDKey, c.appID, namespaceKey, c.namespace, operationKey, operation, successKey, strconv.FormatBool(success))...),
				stats.WithMeasurements(c.jobLatency.M(*elapsed)))
		}
	}
}
```

pkg/runtime/scheduler/internal/cluster/streamer.go
```diff
@@ -178,6 +181,7 @@ func (s *streamer) invokeApp(ctx context.Context, job *schedulerv1pb.WatchJobsRe
 		defer response.Close()
 	}
 
+	elapsed := diag.ElapsedSince(start)
```
JoshVanL: When would `elapsed` ever be 0?

pkg/runtime/scheduler/internal/cluster/streamer.go
JoshVanL, line 199: Should we be monitoring 404s?

tests/integration/suite/daprd/jobs/grpc/metrics.go
```diff
@@ -0,0 +1,110 @@
+/*
+Copyright 2024 The Dapr Authors
```
JoshVanL:
```suggestion
Copyright 2026 The Dapr Authors
```

tests/integration/suite/daprd/jobs/http/metrics.go
```diff
@@ -0,0 +1,90 @@
+/*
+Copyright 2024 The Dapr Authors
```
JoshVanL:
```suggestion
Copyright 2026 The Dapr Authors
```

JoshVanL (changes requested):
Thanks @JeffreyJPZ! Couple things from me

pkg/runtime/scheduler/internal/cluster/streamer.go
JeffreyJPZ: ~~It shouldn't ever be 0 considering the time needed to execute the job and receive the response over the channel.~~ It looks like ```elapsed``` can be 0 if the total time is <1ms due to integer division, but this realistically shouldn't occur due to network latency ([this](https://github.com/dapr/dapr/blob/5e71a8b326f4312534a6066e740f65dabf7a453e/pkg/diagnostics/http_monitoring.go#L355) uses the same calculation). Also, it was carried over from the original PR, and I assumed it was trying to maintain consistency with the surrounding component monitoring code. Happy to refactor it if needed (maybe in a separate PR if not in scope).

Speaking of consistency, I'd probably remove the ```ElapsedSince``` "helper" unless there's any objections?

pkg/runtime/scheduler/internal/cluster/streamer.go
JeffreyJPZ: > Returns the job execution result to scheduler (SUCCESS/FAILURE) to correctly trigger failure policy if required. Adds daprd jobs/actor reminder failure policy integration tests.
> 
> Adds test to job execution to ensure NotFound HTTP status is treated as SUCCESS as it is "not retry-able".

From @cicoyle's comments above and from #8240, I believe client side errors (ex. 404s) are treated as job trigger "successes" to avoid unnecessary retries?

Interestingly, this conflicts with the [scheduler docs](https://docs.dapr.io/concepts/dapr-services/scheduler/) if I'm not misunderstanding something.
> When the Scheduler service triggers a job and it has a client side error, the job is retried by default with a 1s interval and 3 maximum retries.

Personally, it feels like 404s should be monitored (as job trigger failures). Not monitoring them at all might make it difficult to detect misconfigured endpoints, ports, etc. especially for new/naive users. Again, new to Dapr myself so I'm open to suggestions.

JoshVanL (approved):

pkg/runtime/scheduler/internal/cluster/streamer.go
JoshVanL: :+1: this is probably fine

pkg/runtime/scheduler/internal/cluster/streamer.go
JoshVanL: For legacy reasons, 404s are special cased to be "ignored" as job triggers, so are marked as success to the _scheduler_, but we should monitor them as a failure.

pkg/runtime/scheduler/internal/cluster/streamer.go
```diff
@@ -194,6 +199,7 @@ func (s *streamer) invokeApp(ctx context.Context, job *schedulerv1pb.WatchJobsRe
 	default:
 		err := fmt.Errorf("unexpected status code returned from app while processing triggered job %s. status code returned: %v", job.GetName(), statusCode)
 		log.Error(err.Error())
+		diag.DefaultComponentMonitoring.JobTriggeredFailure(ctx, diag.JobTriggerOp, elapsed)
```
JoshVanL: Please can we mark as failure for 404.

JoshVanL (changes requested):
@JeffreyJPZ please fix DCO! 🙂 Will need to force push with sign off on the commits.

pkg/diagnostics/component_monitoring.go
JeffreyJPZ, line 228: Should there be some sort of status code label on the job trigger failure metric? Right now it's missing it and it might be useful to have?

tests/integration/suite/daprd/jobs/grpc/grpc.go
JeffreyJPZ, line 1: Running int tests in parallel sometimes fails due to race conditions (tests share the same sidecar and therefore metrics), splitting these up into their own test cases should fix this.

JoshVanL (approved):

JoshVanL (approved):

marcduiker:
@holopin-bot @JeffreyJPZ runtime-badge Thank you! Here's a digital badge as a small token of appreciation.

PR #9241, opened 2025-12-15
feat: add resp format, model, prompt cache, usage metrics to conversation api

sicoyle:
# Description

This PR reviews the fields in the Conversation API from a Dapr Agents perspective, evaluates which ones we actually need, and aligns our API more closely with OpenAI’s where it makes sense (ie adding fields/commenting on ones we can add in future). Specifically, I’ve:

- Updated the meaning of the metadata field. This is not component metadata (as some quickstarts/docs currently imply), but rather additional metadata passed through to the LLM provider. I do need feedback on this please.
- Identified two fields we don’t use at all (context_id and parameters) and marked them with TODO comments for removal once the API is stable for in future.
- Added a response_format field, which is required for Dapr Agents when using the Conversation API to request responses in a specific JSON schema.
- Bumped the LangChain package version.

Still TODO:
- ~~Update integration tests to reflect these changes.~~
- ~~Open contrib PR bubbling these fields over to the LLM providers and fully test out these fields for the different conversation components~~
- ~~I will likely open a separate PR or just add in usage metrics fields too on the response~~

Opening to get early feedback on the changes.

## Issue reference

Please reference the issue this PR will close: #_[issue number]_

## Checklist

Please make sure you've completed the relevant tasks for this PR, out of the following list:

- [ ] Code compiles correctly
- [ ] Created/updated tests
- [ ] Unit tests passing
- [ ] End-to-end tests passing
- [ ] Extended the documentation / Created issue in the <https://github.com/dapr/docs/> repo: dapr/docs#_[issue number]_
- [ ] Specification has been updated / Created issue in the <https://github.com/dapr/docs/> repo: dapr/docs#_[issue number]_
- [ ] Provided sample for the feature / Created issue in the <https://github.com/dapr/docs/> repo: dapr/docs#_[issue number]_

PR #9265, opened 2026-01-06
feat(conversation): add prompt cache + usage metrics to test cache

sicoyle:
# Description

This PR breaks out work from a larger contrib PR into a smaller, more focused change: https://github.com/dapr/dapr/pull/9241

It reviews the Conversation API from a Dapr Agents perspective, evaluates which fields are required, and aligns the API more closely with OpenAI’s where appropriate (including adding fields now and documenting candidates for future support).

Specifically, this PR:

- Clarifies the semantics of the metadata field to represent request-level metadata passed through to the LLM provider, rather than component metadata as some documentation currently implies.

- Adds a field to enable prompt caching. Dapr currently supports response caching only; this change mirrors OpenAI’s API, allows per-request configuration, and raises the question of whether prompt caching should also be supported as a global, component-level setting.
- Adds token usage metrics as a result of testing that prompt caching indeed works.

- Updates the LangChain dependency.

- Updates integration tests to reflect the API changes.

Note: This PR depends on the corresponding components-contrib change being merged first for the build to pass: https://github.com/dapr/components-contrib/pull/4154

## Issue reference

## Checklist

- [ ] Code compiles correctly
- [ ] Created/updated tests
- [ ] Unit tests passing


dapr/proto/runtime/v1/ai.proto
```diff
@@ -93,6 +100,14 @@ message ConversationRequestAlpha2 {
   // ref: https://python.langchain.com/docs/how_to/tool_choice/
   optional string tool_choice = 9;
 
+  // Retention policy for the prompt cache.
+  // If using OpenAI with this value set to `24h` it enables extended prompt caching,
+  // which keeps cached prefixes active for longer, up to a maximum of 24 hours.
+	// [Learn more](https://platform.openai.com/docs/guides/prompt-caching#prompt-cache-retention).
+  // inspired by openai.ChatCompletionMessageParamUnion.PromptCacheRetention
+  // ref: https://github.com/openai/openai-go/blob/main/chatcompletion.go#L3030
+  optional string prompt_cache_retention = 10;
```
JoshVanL:
```suggestion
  optional google.protobuf.Duration prompt_cache_retention = 10;
```

JoshVanL (changes requested):

dapr/proto/runtime/v1/ai.proto
```diff
@@ -60,14 +65,17 @@ message ConversationRequestAlpha2 {
...
 
   // Parameters for all custom fields.
   map<string, google.protobuf.Any> parameters = 4;
 
-  // The metadata passing to conversation components.
+  // Set of 16 key-value pairs that can be attached to the conversation. 
+  // This can be useful for storing additional information about the object in a structured format, 
+  // and querying for objects via API or the dashboard.
+  // Keys are strings with a maximum length of 64 characters.
+  // Values are strings with a maximum length of 512 characters.
```
JoshVanL, line 78: Rather than this being a field on the request message, should the information which _will_ be set here, go into the HTTP or gRPC request header itself? This is with the assumption that _users_ of Dapr aren't expected to add anything here, and the presence of this field is confusing.

dapr/proto/runtime/v1/ai.proto
```diff
@@ -256,20 +272,71 @@ message ConversationResult {
...
   Created (timestamp): https://github.com/openai/openai-go/blob/main/chatcompletion.go#L173
   Model: https://github.com/openai/openai-go/blob/main/chatcompletion.go#L175
   */
 }
 
+// ref: https://github.com/openai/openai-go/blob/main/completion.go#L162
+// inspired by openai.ChatCompletion.Usage of type CompletionUsage
+message ConversationResultAlpha2CompletionUsage {
+  // Number of tokens in the generated completion.
+  int64 completion_tokens = 1;
```
JoshVanL: Should all these count fields be unsigned rather than signed?

pkg/api/universal/conversation.go
```diff
@@ -195,10 +195,7 @@ func (a *Universal) ConverseAlpha2(ctx context.Context, req *runtimev1pb.Convers
 	// prepare request
 	request := &conversation.Request{}
 	var err error
-	err = kmeta.DecodeMetadata(req.GetMetadata(), request)
-	if err != nil {
-		return nil, err
-	}
+	request.Metadata = req.GetMetadata()
```
JoshVanL, line 198: Here, we would take the metadata from the `ctx`, rather than from the request field.

dapr/proto/runtime/v1/ai.proto
sicoyle: I agree headers would be better. However, this field is already implemented in the SDKs, used in docs/quickstarts, and matches OpenAI's ChatCompletion API metadata field (https://github.com/openai/openai-go/blob/main/chatcompletion.go#L3010), which is passed through to LLM providers. To avoid breaking Alpha2 clients, I'd prefer to keep it for now with a comment noting it will move to headers in a future version. This preserves backward compatibility while aligning with OpenAI's API structure.

dapr/proto/runtime/v1/ai.proto
sicoyle: I matched openai directly here but will update.

PR #9274, opened 2026-01-08
BulkPublishEvent Stable

cicoyle:
BulkPublishEvent stable instead of alpha


dapr/proto/runtime/v1/dapr.proto
```diff
@@ -68,7 +68,7 @@ service Dapr {
   rpc PublishEvent(PublishEventRequest) returns (google.protobuf.Empty) {}
 
   // Bulk Publishes multiple events to the specified topic.
-  rpc BulkPublishEventAlpha1(BulkPublishRequest) returns (BulkPublishResponse) {}
+  rpc BulkPublishEvent(BulkPublishRequest) returns (BulkPublishResponse) {}
```
JoshVanL, line 76: We can't remote the RPC completely- instead we need to make it an deprecated (` option deprecated`), and create a separate BulkPublishEvent RPC.

JoshVanL (changes requested):

dapr/proto/runtime/v1/dapr.proto
cicoyle: oh, I see that now. whoops, lemme fix that

JoshVanL:
@cicoyle looks like everything is unhappy, please take a look

cicoyle:
[I fixed that IT flake in this PR](https://github.com/dapr/dapr/pull/9293)

nelson-parente (approved):

cicoyle:
Once merged, I will cherrypick to the release branch

JoshVanL (approved):

Issue #9266, opened 2026-01-07
Add Support for custom annotations on operator-managed Dapr services


MyMirelHub:
## In what area(s)?

/area operator
/area runtime

## Describe the feature

Currently, the Dapr operator automatically creates services with the suffix `-dapr` for deployments/statefulsets annotated with `dapr.io/enabled: "true"`. These services only support Dapr-specific annotations (like `dapr.io/app-id`) and Prometheus annotations (when metrics are enabled).

**Problem:**
Some Enterprise Kubernetes clusters require mandatory annotations on all resources for policy compliance, security tools, admission controllers, or other operational requirements. Since the operator-generated services don't support custom annotations, they may be rejected by cluster policies.

**Proposed Solution:**
Introduce a new annotation on the Deployment/StatefulSet (e.g., `dapr.io/service-annotations`) that allows users to specify a JSON or comma-separated list of key-value pairs. The Dapr Operator should read this annotation and merge these values into the metadata of the generated -dapr Service.

**Workaround:**
Disable `dapr_operator.serviceReconciler.enabled` and manually create services, but this also requires the overhead of maintaining the service lifecycle, This feature would eliminate the need for this workaround

## Release Note

RELEASE NOTE: **ADD** Support for custom annotations on operator-generated Dapr services via `dapr.io/service-annotation-*` prefix

olitomlinson:
Sounds like a useful addition!

PR #9276, opened 2026-01-09
Add support for custom Dapr service annotations

MyMirelHub:
# Description

Support for custom annotations on Dapr operator-generated services. Previously, the operator-created `-dapr` services could not include custom annotations, which caused issues in clusters with mandatory annotation policies 

- Added new annotation `dapr.io/sidecar-svc-annotations` that accepts comma-separated `key=value` pairs
- Updated operator service reconciler to parse and apply these annotations to the generated `-dapr` services

## Usage

Add the annotation to your Deployment/StatefulSet pod template as a csv of KVs. 

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: test-app
  namespace: default
  labels:
    app: test-app
spec:
  replicas: 1
  selector:
    matchLabels:
      app: test-app
  template:
    metadata:
      labels:
        app: test-app
      annotations:
        dapr.io/enabled: "true"
        dapr.io/app-id: "test-app"
        dapr.io/sidecar-svc-annotations: "com.example.policy.app=test-app,com.example.policy.env=test,com.example.policy.team=platform"
```

And the result will be 

```yaml
apiVersion: v1
kind: Service
metadata:
  annotations:
    com.example.policy.app: test-app
    com.example.policy.env: test
    com.example.policy.team: platform
    dapr.io/app-id: test-app
    prometheus.io/path: /
    prometheus.io/port: "9090"
    prometheus.io/probe: "true"
    prometheus.io/scrape: "true"
  creationTimestamp: "2026-01-09T12:00:59Z"
  labels:
    dapr.io/enabled: "true"
  name: test-app-dapr
```

## Issue reference

Please reference the issue this PR will close: #9266

## Checklist

- [x] Code compiles correctly
- [x] Created/updated tests
- [x] Unit tests passing
- [x] Extended the documentation https://github.com/dapr/docs/pull/5005


olitomlinson:
Thinking defensively, should there be a list of reserved annotations that the end-user is **not** allowed to set via this method? 

I'm thinking that dapr might need exclusive ability to set 1 or more annotations that are vital for daprs operation, and having them conflict with a user specified annotation could cause problems, which would require diagnosis?

Maybe someone from @dapr/maintainers-dapr could advise if this is the case, or if I'm worrying over nothing.

alicejgibbons:
What about adding this as a Helm annotation option instead of as a Daprd annotation? this is the way other annotations on Dapr services are done today: https://github.com/dapr/dapr/blob/master/charts/dapr/README.md#dapr-operator-options

olitomlinson:
> What about adding this as a Helm annotation option instead of as a Daprd annotation? this is the way other annotations on Dapr services are done today: https://github.com/dapr/dapr/blob/master/charts/dapr/README.md#dapr-operator-options

@alicejgibbons I'm very likely being dumb, but I don't follow? 

Can you provide an example pls :)

MyMirelHub:
>  What about adding this as a Helm annotation option instead of as a Daprd annotation? this is the way other annotations on Dapr services are done today: https://github.com/dapr/dapr/blob/master/charts/dapr/README.md#dapr-operator-options

Hey @olitomlinson, just came out of discussing the feature with @alicejgibbons - we can discard this comment ^ (the idea was defining annonations as a list on the operator helm values) - but they would then be global and you would not be able to change annotations per service - so the current approach remains 

> I'm thinking that dapr might need exclusive ability to set 1 or more annotations that are vital for daprs operation, and having them conflict with a user specified annotation could cause problems, which would require diagnosis?

I get the sentiment on this one, but Implementing a blocklist could lead to over-validation and trying to predict edge cases - but keen to also hear from @dapr/maintainers-dapr on this one

olitomlinson:
> but Implementing a blocklist could lead to over-validation and trying to predict edge cases

@MyMirelHub of course I understand the need to be pragmatic, but I don't see where the over-validation & prediction element comes into this? 

The dapr operator must have a known list of annotations that it sets on a service as part of normal dapr operations, right? All im asking is that we blacklist these annotations, so that we don't allow an end-user to create a mess that puts the service into a unknown/unsupported state. That mess would inevitably lead to support cases like "why is dapr not working as I expected" which we obviously want to prevent as much as possible, in the enterprise dapr support space, and the OS community space.

alicejgibbons:
> The dapr operator must have a known list of annotations that it sets on a service as part of normal dapr operations, right? The only annotations that the operator sets on the service is the app id. Thus if disallow that one to Oli's point we should be g2g.

image

MyMirelHub:
> That mess would inevitably lead to support cases like "why is dapr not working as I expected" which we obviously want to prevent as much as possible

You make a good point @olitomlinson. The following now gets dropped silently with a  warning log

```
    dapr.io/app-id: test-app
    prometheus.io/path: /
    prometheus.io/port: "9090"
    prometheus.io/probe: "true"
    prometheus.io/scrape: "true"
```

pkg/operator/handlers/dapr_handler.go
```diff
@@ -43,6 +44,23 @@ const (
...
 
+// Reserved annotation keys that the operator manages and must not be overridden
+var reservedAnnotationKeys = map[string]struct{}{
+	annotations.KeyAppID:       {},
+	annotationPrometheusProbe:  {},
+	annotationPrometheusScrape: {},
+	annotationPrometheusPort:   {},
+	annotationPrometheusPath:   {},
+}
+
```
JoshVanL: Rather than having to have a known set of annotations which are dropped, instead when applying the custom annotations, if the annotation already exists to does this logging.

pkg/operator/handlers/dapr_handler.go
```diff
@@ -278,6 +296,26 @@ func (h *DaprHandler) createDaprServiceValues(ctx context.Context, expectedServi
...
 	}
 
+	if val, ok := wrapper.GetTemplateAnnotations()[annotations.KeySidecarSvcAnnotations]; ok && val != "" {
+		pairs := stdstrings.Split(val, ",")
+		for _, pair := range pairs {
+			pair = stdstrings.TrimSpace(pair)
+			if pair == "" {
+				continue
+			}
+			key, value, found := stdstrings.Cut(pair, "=")
```
JoshVanL, line 286: Log error if malformed.

pkg/operator/handlers/dapr_handler.go
```diff
@@ -278,6 +296,26 @@ func (h *DaprHandler) createDaprServiceValues(ctx context.Context, expectedServi
 		annotationsMap[annotationPrometheusPath] = "/"
 	}
 
+	if val, ok := wrapper.GetTemplateAnnotations()[annotations.KeySidecarSvcAnnotations]; ok && val != "" {
+		pairs := stdstrings.Split(val, ",")
+		for _, pair := range pairs {
+			pair = stdstrings.TrimSpace(pair)
+			if pair == "" {
+				continue
+			}
```
JoshVanL:
```suggestion
```

pkg/operator/handlers/dapr_handler.go
```diff
@@ -278,6 +296,26 @@ func (h *DaprHandler) createDaprServiceValues(ctx context.Context, expectedServi
...
+			key, value, found := stdstrings.Cut(pair, "=")
+			if found && key != "" {
+				key = stdstrings.TrimSpace(key)
+
+				if isReservedAnnotation(key) {
+					log.Warnf("Ignoring reserved annotation %q in %s: reserved keys (%v) are operator-managed", key, annotations.KeySidecarSvcAnnotations, reservedAnnotationKeys)
+					continue
+				}
+				annotationsMap[key] = stdstrings.TrimSpace(value)
```
JoshVanL:
```suggestion
				if _, ok := annotationsMap[key]; ok  {
					log.Warnf("Ignoring reserved annotation %q as it operator-managed", key)
					continue
				}
				annotationsMap[key] = stdstrings.TrimSpace(value)
```

JoshVanL (changes requested):

pkg/operator/handlers/dapr_handler.go
MyMirelHub: makes sense

pkg/operator/handlers/dapr_handler.go
```diff
@@ -43,6 +44,23 @@ const (
...
+}
+
+// isReservedAnnotation checks if an annotation key is reserved (exact key)
+func isReservedAnnotation(key string) bool {
+	if _, ok := reservedAnnotationKeys[key]; ok {
+		return true
+	}
+	return false
+}
+
```
JoshVanL:
```suggestion
```

pkg/operator/handlers/dapr_handler.go
```diff
@@ -278,6 +296,23 @@ func (h *DaprHandler) createDaprServiceValues(ctx context.Context, expectedServi
 		annotationsMap[annotationPrometheusPath] = "/"
 	}
 
+	if val, ok := wrapper.GetTemplateAnnotations()[annotations.KeySidecarSvcAnnotations]; ok && val != "" {
```
JoshVanL:
```suggestion
	if val, ok := wrapper.GetTemplateAnnotations()[annotations.KeySidecarSvcAnnotations]; ok {
```

PR #9283, opened 2026-01-13
fix(grpc): remove singleton connection limit for local app

### LINKED ISSUE #8671: gRPC proxy does not create new HTTP2 connections when concurrent connection cap is reached
## In what area(s)?

 /area runtime

## What version of Dapr?

1.15.4

## Expected Behavior

When the concurrent connection limit for HTTP2 (default 100) is reached, a new HTTP2 connection should be created between dapr sidecars.  This does not appear to be happening.

## Actual Behavior

Concurrent connections are capped at 100 and no new HTTP2 connections are formed.

Full repro (with ASP.NET server/client) is here:  https://github.com/Flern/dapr-streaming-repro

## Release Note

RELEASE NOTE: **FIX** Creates new HTTP2 connections when concurrent connection limit reached in gRPC proxy

Flern:
This change implements gRPC connection pooling for the local app channel, allowing it to scale beyond the 100 concurrent streams limit.

Fixes #8671

# Description

Removed the singleton app connection from the gRPC manager and replaced with access to the local connection pool.  This effectively frees the pool to allocate new connections when the concurrent stream limit is reached.  A teardown method is also returned to allow unused connections to be removed.  Test infrastructure was also updated.  First, the direct setting of the app client connection was replaced with an addition of the test connection to the pool.  Second, tests were added that verify the creation of new client connections when the 100 stream limit is reached.

## Issue reference

Please reference the issue this PR will close: #8671

## Checklist

- [x] Code compiles correctly
- [x] Created/updated tests
- [x] Unit tests passing

### DIFF
--- a/tests/integration/framework/binary/helpers/helmtemplate/go.mod
+++ b/tests/integration/framework/binary/helpers/helmtemplate/go.mod
@@ -28,7 +28,7 @@ require (
 	github.com/distribution/reference v0.5.0 // indirect
 	github.com/docker/cli v25.0.1+incompatible // indirect
 	github.com/docker/distribution v2.8.3+incompatible // indirect
-	github.com/docker/docker v25.0.5+incompatible // indirect
+	github.com/docker/docker v25.0.13+incompatible // indirect
 	github.com/docker/docker-credential-helpers v0.7.0 // indirect
 	github.com/docker/go-connections v0.5.0 // indirect
 	github.com/docker/go-metrics v0.0.1 // indirect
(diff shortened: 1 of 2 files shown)


JoshVanL:
@Flern please can you create an integration test for this? I have not been able to reproduce.

Flern:
> @Flern please can you create an integration test for this? I have not been able to reproduce.

I'll look at the integration tests and see what I can add.  It may not be until Sunday/Monday that I can add more testing.

The current tests artificially add multiple connections to connection pools to test the pooling, but the implementation of the app client connection always returns the existing connection after initialization with the default creation method.  So there's no way you're ever getting more than one gRPC connection per destination host at runtime.

@yaron2 asked for more integration activity, and I did verify that a) more than 100 parallel streams are now supported and b) round-robin distribution to multiple replicas of app-id services is preserved.  Are there integration tests that evaluate these behaviors today (if you know off the top of your head)?

Flern:
Looks like there are issues in existing integration tests.  Trying to resolve those now....

Flern:
I had to clean up my forked repository.  I'll open a new PR from a branch instead.  Test issues should be resolved, but there were wider-reaching updates (e.g., runtime.go) that I had to put in place.  I'll link the new PR when it's available.

PR #9298, opened 2026-01-16
[1.17] Workflow: Clear cache when stalled

JoshVanL:
Since a workflow can be stalled indefinitely, clear the memory cache of the actor to free up resources.


pkg/actors/targets/workflow/orchestrator/factory.go
```diff
@@ -145,9 +145,6 @@ func (f *factory) initOrchestrator(o any, actorID string) *orchestrator {
 	or.closed.Store(false)
 	or.lock.Init()
 
-	or.state = nil
-	or.rstate = nil
-	or.ometa = nil
```
JoshVanL, line 150: Duplicate code- see line 156

cicoyle (approved):

PR #9299, opened 2026-01-16
Placement authorize Dapr actor types

JoshVanL:
Ensure that dapr actor types are properly authorized in placement service. Ensures the client can only register with Dapr actor types with the same app ID and namespace of the client.


pkg/placement/placement.go
```diff
@@ -457,6 +462,23 @@ func (p *Service) validateFirstMessage(firstMessage *placementv1pb.Host, clientI
 		return status.Errorf(codes.PermissionDenied, "client ID %s is not allowed", firstMessage.GetId())
 	}
 
+	if err := p.authorizeMessage(firstMessage, clientID); err != nil {
```
nelson-parente: Can this be considered a breaking change? If so, we need to add it to the 1.17 release notes. Let’s also create a tracking item so we don’t forget.

nelson-parente (dismissed):

pkg/placement/placement.go
JoshVanL: No- these actor types are always only meant to be registered by it's own app. This is considered enforcing a policy which already existed.

pkg/placement/placement.go
```diff
@@ -457,6 +462,23 @@ func (p *Service) validateFirstMessage(firstMessage *placementv1pb.Host, clientI
...
+		return err
+	}
+
+	return nil
+}
+
+func (p *Service) authorizeMessage(msg *placementv1pb.Host, clientID *spiffe.Parsed) error {
```
cicoyle: Should we be using clientID here? Currently, I don't see it being used.
```suggestion
func (p *Service) authorizeMessage(msg *placementv1pb.Host) error {
```

pkg/placement/placement.go
```diff
@@ -457,6 +462,23 @@ func (p *Service) validateFirstMessage(firstMessage *placementv1pb.Host, clientI
...
+	for _, entity := range msg.GetEntities() {
+		split := strings.Split(entity, ".")
+		if len(split) >= 2 && split[0] == "dapr" && split[1] == "internal" {
```
cicoyle: Can you make consts for `dapr` & `internal`?

cicoyle (changes requested):
Mostly lgtm, just the same comments from [this PR](https://github.com/dapr/dapr/pull/9300#pullrequestreview-3678368700)

cicoyle (approved):

PR #9300, opened 2026-01-16
[1.17] Placement authorize internal types

JoshVanL:
Ensure that dapr actor types are properly authorized in placement service. Ensures the client can only register with Dapr actor types with the same app ID and namespace of the client.

### DIFF
--- a/pkg/injector/annotations/annotations.go
+++ b/pkg/injector/annotations/annotations.go
@@ -77,4 +77,5 @@ const (
 	KeyPluggableComponentsInjection     = "dapr.io/inject-pluggable-components"
 	KeyAppChannel                       = "dapr.io/app-channel-address"
 	KeySentryRequestJwtAudiences        = "dapr.io/sentry-request-jwt-audiences"
+	KeySidecarSvcAnnotations            = "dapr.io/sidecar-svc-annotations"
 )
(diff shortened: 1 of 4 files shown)


nelson-parente (approved):
ditto: https://github.com/dapr/dapr/pull/9299#discussion_r2702800214

pkg/placement/placement.go
```diff
@@ -457,6 +462,23 @@ func (p *Service) validateFirstMessage(firstMessage *placementv1pb.Host, clientI
...
 	}
 
+	if err := p.authorizeMessage(firstMessage, clientID); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (p *Service) authorizeMessage(msg *placementv1pb.Host, clientID *spiffe.Parsed) error {
```
cicoyle: Should we be using `clientID` here? Currently, I don't see it being used.
```suggestion
func (p *Service) authorizeMessage(msg *placementv1pb.Host) error {
```

pkg/placement/placement.go
```diff
@@ -457,6 +462,23 @@ func (p *Service) validateFirstMessage(firstMessage *placementv1pb.Host, clientI
...
+	for _, entity := range msg.GetEntities() {
+		split := strings.Split(entity, ".")
+		if len(split) >= 2 && split[0] == "dapr" && split[1] == "internal" {
```
cicoyle: Can you make consts for `dapr` & `internal`?

PR #9301, opened 2026-01-19
[WIP] Implement AES-CBC-AEAD encryption scheme

### LINKED ISSUE #5998: Dapr state store encryption should use associated data for improved security
## In what area(s)?

/area runtime

## Describe the proposal

The built-in [state store encryption feature](https://docs.dapr.io/developing-applications/building-blocks/state-management/howto-encrypt-state/) of Dapr uses AES in GCM mode, which is an AEAD (Authenticated Encryption with Associated Data) mode. However, we are currently not passing any associated data to the cipher.

I recommend we change the way we encrypt the data, to add the key of the item as additional authenticated data (AAD), which would prevent the ciphertext from being decrypted if it's moved to a different key. 

> This is **not** a disclosure of an actual vulnerability in Dapr and it requires users to have access to the underlying data store. However, Dapr can take some steps to increase the security protections we have in place.

Example:

- Imagine an application that uses Dapr to store healthcare data in a state store.
- Alice is a user and stores her data in the application. She is confident the data is encrypted.
- Mallory is an attacker that figured out a way to access the data in the state store (imagine a [database accidentally opened to the Internet](https://thehackernews.com/2015/02/mongodb-database-hacking.html), or a [cloud storage bucket with public access enabled](https://blog.lightspin.io/risks-of-misconfigured-s3-buckets) - this happens more frequently than we think it does; alternatively, Mallory could be a rogue database administrator). Thanks to being able to access the data store, Mallory has access to the data, but the data is encrypted.
- Mallory does not have the key and cannot decrypt the data. However, Mallory can create an account on the same application that Alice uses.
- After Mallory creates an account, there are now 2 records in the database that Dapr manages: Alice's and Mallory's.
- Mallory, then, goes into the data store and swaps the payloads (which are encrypted) of the two records. Alice's data is now under the key for Mallory's.
- Mallory then signs into the application, which retrieves the data from the data store and decrypts it successfully. Mallory now has access to Alice's data in plain-text.

Although this issue isn't a vulnerability in Dapr itself and needs some other (non-trivial) conditions to be successfully exploited, there's something relatively simple that Dapr can do to prevent these attacks: adding the key name as part of the additional authenticated data (AAD) when encrypting/decrypting the value with AES-GCM.

Going back to the example above, if Alice's data is stored in the key `health-data||alice`, if that string is used as AAD when encrypting the message, the same string needs to be passed to decrypt it, or the decryption will fail. This means that when Mallory swaps Alice records into their own, the key is `health-data||mallory`, and when Dapr tries to decrypt the data, the authentication tag check fails and Dapr does not return the decrypted data to Mallory.

This change is not backwards-compatible, but the way we can handle this is adding a version to encrypted data stored in the state store. This will be "v2", so when we see that prefix (or any other way the data is passed), we know we need to use the key name as AAD (bonus: we can also put as prefix the ID of the key used to encrypt the message, so when the key is rotated we don't need to try using both keys to find the right one).

### LINKED ISSUE #6027: Dapr state store encryption should switch to AES-CBC

JeffreyJPZ:
# Description

This PR implements an AES-CBC-AEAD encryption scheme for client-side state encryption, enabled by a `StateV2Encryption` config flag. This strengthens security for state stores when using Dapr, as having a repeated nonce won't run the risk of attackers being able to recover plaintexts.

## Issue reference

Closes #5998, closes #6027

## Checklist

- [x] Code compiles correctly
- [x] Created/updated tests
- [x] Unit tests passing

## TODO
- [ ] Add/update integration tests
- [ ] Update docs
- [ ] Handle parsing of old encryption format when decrypting (try to parse json first, then fallback if needed?)
- [ ] Handle key size non-overlap, "master" key needs to be twice the size so it can be split into MAC and encryption keys, i.e. 128-bit keys will not be supported for v2 encryption while 384 and 512-bit keys are

### DIFF
--- a/pkg/encryption/encryption.go
+++ b/pkg/encryption/encryption.go
@@ -18,7 +18,6 @@ import (
 	"crypto/aes"
 	"crypto/cipher"
 	"crypto/rand"
-	b64 "encoding/base64"
 	"encoding/hex"
 	"errors"
 	"fmt"
@@ -33,11 +32,12 @@ import (
 type Algorithm string
 
 const (
-	primaryEncryptionKey   = "primaryEncryptionKey"
-	secondaryEncryptionKey = "secondaryEncryptionKey"
-	errPrefix              = "failed to extract encryption key"
-	AESCBCAEADAlgorithm    = "AES-CBC-AEAD"
-	AESGCMAlgorithm        = "AES-GCM"
+	primaryEncryptionKey       = "primaryEncryptionKey"
+	secondaryEncryptionKey     = "secondaryEncryptionKey"
+	errPrefix                  = "failed to extract encryption key"
+	AESGCMAlgorithm            = "AES-GCM"
+	AESCBCAEADAlgorithm        = "AES-CBC-AEAD"
+	AESCBCAEADAlgorithmVersion = 2
 )
 
 // ComponentEncryptionKeys holds the encryption keys set for a component.
@@ -55,16 +55,6 @@ type Key struct {
 }
 
 type ComponentEncryptionKeyOptions struct {
-	// TODO: remove when feature flag is removed
-	StateV2EncryptionEnabled bool
-}
-
-type EncryptOptions struct {
-	// TODO: remove when feature flag is removed
-	StateV2EncryptionEnabled bool
-}
-
-type DecryptOptions struct {
 	// TODO: remove when feature flag is removed
 	StateV2EncryptionEnabled bool
 }
@@ -195,51 +185,23 @@ func tryGetEncryptionKeyFromMetadataItem(namespace string, item commonapi.NameVa
 	return Key{}, nil
 }
 
-// Encrypt takes a byte array and encrypts it using a supplied encryption key.
-func encrypt(value []byte, key Key, additionalData []byte, opts EncryptOptions) ([]byte, error) {
-	if opts.StateV2EncryptionEnabled {
-		// TODO: when feature flag is removed, replace old encryption logic with this
-		nsize := make([]byte, key.cipherObjV2.NonceSize())
-		if _, err := io.ReadFull(rand.Reader, nsize); err != nil {
-			return value, err
-		}
-
-		return key.cipherObjV2.Seal(nsize, nsize, value, additionalData), nil
-	}
-
-	nsize := make([]byte, key.cipherObj.NonceSize())
+// Encrypt takes a byte array and encrypts it using a supplied cipher.
+func encrypt(value []byte, additionalData []byte, cipher cipher.AEAD) ([]byte, error) {
+	nsize := make([]byte, cipher.NonceSize())
 	if _, err := io.ReadFull(rand.Reader, nsize); err != nil {
 		return value, err
 	}
 
-	return key.cipherObj.Seal(nsize, nsize, value, nil), nil
+	return cipher.Seal(nsize, nsize, value, additionalData), nil
 }
 
-// Decrypt takes a byte array and decrypts it using a supplied encryption key.
-func decrypt(value []byte, key Key, additionalData []byte, opts DecryptOptions) ([]byte, error) {
-	enc, err := b64.StdEncoding.DecodeString(string(value))
-	if err != nil {
-		return value, err
-	}
+// Decrypt takes a byte array and decrypts it using a supplied cipher.
+// The byte array should be Base64-encoded.
+func decrypt(value []byte, additionalData []byte, cipher cipher.AEAD) ([]byte, error) {
+	nsize := cipher.NonceSize()
+	nonce, ciphertext := value[:nsize], value[nsize:]
 
-	if additionalData == nil {
-		// fallback to old encryption scheme
-		nsize := key.cipherObj.NonceSize()
-		nonce, ciphertext := enc[:nsize], enc[nsize:]
-
-		return key.cipherObj.Open(nil, nonce, ciphertext, nil)
-	}
-
-	if opts.StateV2EncryptionEnabled {
-		// TODO: move to outer scope when feature flag is removed
-		nsize := key.cipherObjV2.NonceSize()
-		nonce, ciphertext := enc[:nsize], enc[nsize:]
-
-		return key.cipherObjV2.Open(nil, nonce, ciphertext, additionalData)
-	}
-
-	// should never get here
-	return nil, nil
+	return cipher.Open(nil, nonce, ciphertext, additionalData)
 }
 
 func createCipher(key Key, algorithm Algorithm) (cipher.AEAD, error) {
--- a/pkg/encryption/encryption.go
+++ b/pkg/encryption/encryption.go
@@ -54,12 +54,12 @@ type Key struct {
 	cipherObjV2 cipher.AEAD // v2 AES-CBC-AEAD encryption scheme
 }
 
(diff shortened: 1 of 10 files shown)


JeffreyJPZ:
Not an expert in cryptography by any means, any pointers would be appreciated.

pkg/encryption/state.go
```diff
@@ -25,6 +29,33 @@ const (
 	separator = "||"
 )
 
+type EncryptedValue struct {
+	// Version of the encryption scheme
+	Version int `json:"v"`
+	// ID of the encryption key as the Base64-encoded SHA-224 hash of the key's bytes
+	KeyID string `json:"kid"`
```
JeffreyJPZ, line 36: The proposal in #6027 wasn't explicit on the encoding for key ID but Base64 is pretty efficient, also used in the old encryption scheme for ciphertext.

pkg/encryption/encryption.go
```diff
@@ -159,27 +185,22 @@ func tryGetEncryptionKeyFromMetadataItem(namespace string, item commonapi.NameVa
...
 		return value, err
 	}
 
-	return key.cipherObj.Seal(nsize, nsize, value, nil), nil
+	return cipher.Seal(nsize, nsize, value, additionalData), nil
 }
 
-// Decrypt takes a byte array and decrypts it using a supplied encryption key.
-func decrypt(value []byte, key Key) ([]byte, error) {
-	enc, err := b64.StdEncoding.DecodeString(string(value))
```
JeffreyJPZ, line 174: Moved Base64 decoding to `TryDecryptValue` since encoding is done in `TryEncryptValue`, better separation of concerns I think (was stuck on this for a while trying to debug the encoding/decoding).

pkg/encryption/encryption.go
```diff
@@ -159,27 +185,22 @@ func tryGetEncryptionKeyFromMetadataItem(namespace string, item commonapi.NameVa
 	return Key{}, nil
 }
 
-// Encrypt takes a byte array and encrypts it using a supplied encryption key.
-func encrypt(value []byte, key Key) ([]byte, error) {
-	nsize := make([]byte, key.cipherObj.NonceSize())
+// Encrypt takes a byte array and encrypts it using a supplied cipher.
+func encrypt(value []byte, additionalData []byte, cipher cipher.AEAD) ([]byte, error) {
```
JeffreyJPZ, line 189: Decided to pass the cipher instead since it maps to a key anyway, avoids the issue of deciding which cipher to use and makes it easier to support other algorithms (e.g. ChaPoly) as long as they implement `cipher.AEAD`.

pkg/encryption/encryption.go
```diff
@@ -46,13 +48,19 @@ type ComponentEncryptionKeys struct {
 
 // Key holds the key to encrypt an arbitrary object.
 type Key struct {
-	Key       string
-	Name      string
-	cipherObj cipher.AEAD
+	Key         string
+	Name        string
+	cipherObj   cipher.AEAD // v1 AES-GCM encryption scheme
+	cipherObjV2 cipher.AEAD // v2 AES-CBC-AEAD encryption scheme
```
JeffreyJPZ, line 54: Need to keep multiple ciphers for backward compatibility, this is the best way I can think of.

pkg/encryption/state.go
```diff
@@ -25,6 +29,33 @@ const (
 	separator = "||"
 )
 
+type EncryptedValue struct {
+	// Version of the encryption scheme
+	Version int `json:"v"`
```
JeffreyJPZ, line 34: Would AES-CBC be version 2? 1? The proposal #6027 was a little ambiguous.

Should opt-in algorithms also require different versions? This would make it easier to select the right scheme to decrypt with down the line. The algorithm name is another possibility but encryption schemes (with different additional data) and algorithms might not be one-to-one.

pkg/encryption/encryption.go
```diff
@@ -111,6 +119,15 @@ func ComponentEncryptionKey(component v1alpha1.Component, secretStore secretstor
...
 
 		cek.Primary.cipherObj = cipherObj
+
+		if opts.StateV2EncryptionEnabled {
+			cipherObjV2, err := createCipher(cek.Primary, AESCBCAEADAlgorithm)
+			if err != nil {
+				return ComponentEncryptionKeys{}, err
+			}
+
+			cek.Primary.cipherObjV2 = cipherObjV2
```
JeffreyJPZ, line 129: Pretty sure keys aren't persisted, so if the flag is enabled, then disabled at some point before the sidecar restarts (or similar), users won't be able to decrypt their data?

pkg/encryption/state.go
```diff
@@ -25,6 +29,33 @@ const (
 	separator = "||"
 )
 
+type EncryptedValue struct {
```
JeffreyJPZ, line 32: Note that the JSON object is serialized before being Base64 encoded.

PR #9305, opened 2026-01-19
Integration: Fix tests

JoshVanL:
- Fix cross-application activity int test.
- Remove superfluous debug log line.
- Fix rerun child async int test.
- Fix workflow app health tests


cicoyle (approved):

cicoyle (changes requested):
changes lgtm but linter is failing

cicoyle:
We might need to allocate etcd ports in the `Run()` to minimize/avoid collision and have fresh ports right before starting etcd.

cicoyle (approved):

PR #9317, opened 2026-01-19
OnBulkTopicEventAlpha1 -> stable

cicoyle:
We bumped the user facing proto to be stable: `BulkPublishEventAlpha1` -> `BulkPublishEvent`

I believe we should also bump the callback proto to be stable, so that users may both publish and subscribe to the stable APIs, and not publish via a stable API and receive back on the Alpha api. Therefore, I am also promoting `OnBulkTopicEventAlpha1` to be stable as `OnBulkTopicEvent` on the AppCallBack service

I did fall back to the alpha callback if the stable one is unimplemented.


cicoyle:
Will fix up CI rq

pkg/runtime/subscription/postman/grpc/grpc.go
```diff
@@ -204,11 +205,30 @@ func (g *grpc) DeliverBulk(ctx context.Context, req *postman.DeliverBulkRequest)
...
 
-	start := time.Now()
-	res, err := clientV1.OnBulkTopicEventAlpha1(ctx, envelope)
-	elapsed := diag.ElapsedSince(start)
+	res, err = call(clientV1.OnBulkTopicEvent)
+	if err != nil {
+		if errStatus, hasErrStatus := status.FromError(err); hasErrStatus && errStatus.Code() == codes.Unimplemented {
+			// fallback: to alpha if unimplemented
+			log.Warnf("falling back to OnBulkTopicEventAlpha1 due to unimplemented error: %s", err)
+			res, err = call(clientAlpha.OnBulkTopicEventAlpha1) //nolint:staticcheck
```
cicoyle, line 229: Im open to rm-ing this if that is preferred.

JoshVanL (approved):

PR #9322, opened 2026-01-20
Use Kube cluster domain for scheduler DNS

JoshVanL:
The Scheduler DNS for its X.509 identity certificate should contain the Kubernetes cluster DNS name, rather than the control plane trust domain.


JoshVanL:
@cicoyle added comment

cicoyle (approved):

PR #9323, opened 2026-01-20
[1.17] Use Kube cluster domain for scheduler DNS

JoshVanL:
The Scheduler DNS for its X.509 identity certificate should contain the Kubernetes cluster DNS name, rather than the control plane trust domain.

### DIFF
--- a/tests/integration/suite/scheduler/kubernetes/prefix.go
+++ b/tests/integration/suite/scheduler/kubernetes/prefix.go
@@ -46,7 +46,7 @@ type prefix struct {
 }
 
 func (p *prefix) Setup(t *testing.T) []framework.Option {
-	/ Skip windows as test requires a resolvconf lookup.
+	// Skip windows as test requires a resolvconf lookup.
 	os.SkipWindows(t)
 
 	tld, err := utils.GetKubeClusterDomain()


tests/integration/suite/scheduler/kubernetes/prefix.go
```diff
@@ -44,7 +46,14 @@ type prefix struct {
 }
 
 func (p *prefix) Setup(t *testing.T) []framework.Option {
-	p.sentry = sentry.New(t)
+	os.SkipWindows(t)
```
cicoyle, line 50: Can you add a comment as to why it gets skipped for windows, so its clearer in the future when we look at it again?

cicoyle (changes requested):
mostly lgtm just one comment

tests/integration/suite/scheduler/kubernetes/prefix.go
cicoyle: on second thought-I see it calls to a `t.skip`. I would still like a comment that is more specific to this test, but wont block the PR

cicoyle (approved):

tests/integration/suite/scheduler/kubernetes/prefix.go
JoshVanL: @cicoyle done

cicoyle (approved):

Issue #9188, opened 2025-11-17
pathMatching breaks increasedCardinality for actor metrics


darraghjones:
## In what area(s)?

/area runtime

> /area operator

> /area placement

> /area docs

> /area test-and-release

> /area injector

> /area scheduler

> /area sentry

## What version of Dapr?

dapr version
CLI version: 1.15.0
Runtime version: 1.16.2

## Expected Behavior

when increasedCardinality is false and pathMatching is used, the path labels should be either an empty string, or one of the matched patterns.

## Actual Behavior

increasedCardinality: false works fine....until I add pathMathcing...at which point "unmatched" paths are included as label values (specfically for actor method calls)

## Steps to Reproduce the Problem

I created a simple aspire-starter app. call WithDaprSidecar on each project....and then convered the weather API service to an actor.
invoke the weather actor like this:
```
public class WeatherApiClient(IActorProxyFactory actorProxyFactory) : IWeatherApiClient
{
    public async Task<WeatherForecast[]> GetWeatherAsync()
    {
        return await actorProxyFactory.CreateActorProxy<IWeatherApiClient>(ActorId.CreateRandom(), "WeatherActor").GetWeatherAsync();
    }
}
```

check the logs exposed by the actor sidecar.

when the config looks like this:
```
  metrics:
    enabled: true
    recordErrorCodes: true
    http:
      increasedCardinality: false
```
the path labels are all emtpy strings.

when the config looks something like this:
```
  metrics:
    enabled: true
    recordErrorCodes: true
    http:
      increasedCardinality: false
      pathMatching:
       - /
```
the full path label for the actor requests are used....e.g.
`dapr_http_server_latency_bucket{app_id="webfrontend",method="PUT",path="//v1.0/actors/WeatherActor/1fb6b63e-7ce9-4f69-acd9-428aba481525/method/GetWeatherAsync",status="200",le="40"} 1`

## Release Note

RELEASE NOTE:

darraghjones:
possibly related https://github.com/dapr/dapr/issues/7736

alicejgibbons:
Hey @darraghjones we don't support path matches for wildcard strings like `/`. I see your point for actors, but if you add all the paths in do you still get this problem or just for `/`?

```
  metrics:
    enabled: true
    recordErrorCodes: true
    http:
      increasedCardinality: false
      pathMatching:
       - /v1.0/actors/WeatherActor/{ID}/method/GetWeatherAsync
```

[...]

Issue #8745, opened 2025-05-29
HTTP Path matching metrics: match substring of path not entire path


alicejgibbons:
## In what area(s)?

> /area runtime

## What version of Dapr?

1.15+

## Expected Behavior

Today when users have an API routes that contain a subset of characters that are in the Dapr API route delimiter, path matching does not work unless the entire route is specified. Ie. the last characters will not be path matched but only the full path.

For example if you are using service invocation to invoke an app `order-app` with method: `/api/v1.0/order/route/{ID}`, the only way to path match this is to add the full route including dapr invoke keywords. Otherwise only the metric on the invokee app will be matched. For example, to ensure matching works correctly the following needs to be added for both invoker and invokee apps:

``` yaml
pathMatching:
- /v1.0/invoke/order-app/method/api/v1.0/order/route/{ID} #invoker app metric
- /api/v1.0/order/route/{ID} #invokee app metric
```

## Actual Behavior

Only one path should have to be added since its the back half of the metric, as follows.

``` yaml
pathMatching:
- /api/v1.0/order/route/{ID} #invokee app metric
```

## Steps to Reproduce the Problem

Do service invocation between two apps and add the invoker app path into the path-matching logic.
Check metrics and see that only the invokee metrics are grouped not invoker app.

## Release Note

RELEASE NOTE: **FIX** HTTP Patch matching metrics bug.

ananyapahwa:
can I work on this issue?

alicejgibbons:
Yes - if you comment /assign it should work @ananyapahwa

aabhas-sao:
hey @alicejgibbons ,

I was going through all docs to re-create this issue locally.

I have two app running using multi app run.
dapr version: 
CLI version: 1.15.1
Runtime version: 1.15.5

macos arm64

```
apiVersion: dapr.io/v1alpha1
kind: Configuration
metadata:
  name: daprConfig
spec:
  tracing:
    samplingRate: "1"
    zipkin:
      endpointAddress: http://localhost:9411/api/v2/spans
  metrics:
    enabled: true
    http:
      increasedCardinality: true
      pathMatching:
      - /orders/route/{ID}
```

I have the config as above. I'm observing below results


![Image](https://github.com/user-attachments/assets/c2fe47ff-92bf-43cd-8486-c112596f9e97)
cc: order-processor app

![Image](https://github.com/user-attachments/assets/be02b794-ca99-4b60-b6e4-38989e86bc29)
cc: checkout app


[...]

PR #9336, opened 2026-01-21
Fix HTTP path matching cardinality leak and support invocation auto-registration

MyMirelHub:
# Description

- `refactor` Removed using a fake response writer to capture matched paths. Now uses the `mux.Handler(req)` method to resolve patterns directly.
- `fix` Implemented a `NormalizeHTTPPath` helper that merges double slashes (`// --> /`) before matching. 
- Preserve legacy vs strict behavior: legacy returns raw path on fallback; strict returns empty for unmatched.
- `fix` Automatically register the Dapr Service Invocation prefix ( `/v1.0/invoke/{app_id}/method/...` ) for all user-configured paths.

Note* Went against implementing auto-registration for Actor paths (e.g., /v1.0/actors/... or internal actor callbacks). Given the high amount of actor path variations (method invocations, reminders, timers, state transactions). I believe actor paths should remain explicitly user-configured. 

## Issue reference

Please reference the issue this PR will close: 

- https://github.com/dapr/dapr/issues/9188
- https://github.com/dapr/dapr/issues/8745

## Checklist

- [x] Code compiles correctly
- [x] Created/updated tests
- [x] Unit tests passing


JoshVanL (approved):
@MyMirelHub please also PR into release-1.17

PR #9357, opened 2026-01-26
Placement server refactor

JoshVanL:
This PR introduces a significant refactor of the placement server
codebase to improve maintainability, readability, and correctness.

Placement now ensures full 3 stage lock dissemination of placement
updates to all members in a namespace. This ensures that all members
have a consistent view of the placement table, and prevents issues where
members have stale placement information.

Placement now reacts immediately to client disconnects, and maintains a
correct state machine of the placement table.

Raft-
Refactor reduces the usage of raft to only use the leader election
functionality. The rest of the storage replication was not actually
being used, so this simplifies the code significantly.

Loop-based architecture -
Breaks up the single monolithic placement loop into multiple control
loops, using the same loop framework used elsewhere. These loops are:

namespaces -> [disseminator] -> [[stream]]

`stream` handles each individual stream, feeding up each report host
back to the disseminator. Handles closing the stream with error, and
authentication/authorization of secondary messages.

`disseminator` handles the dissemination of a single namespace. It
controls sending the dissimination operations to each of the streams. It
tracks the "LOCK", "UNLOCK", and "UPDATE" three stage lock operations.
If a new report comes in while an operation is in progress, it reverts
back to the "LOCK" stage with the new report.
It all handles the timeout of the dissemination of a namespace. If the
full 3 stage lock dissemintion takes longer than the configured timeout
(default 8s), then the disseminator will close the stream(s) responsible
for the timeout and starts a new dissemination.
Store keeps track of the placement table for that namespace.

`namespaces` loop handles all namespaces. It handles routing of reports
and operations to the correct disseminator. It also handles creation
and deletion of disseminators as members from namespaces are added or
removed.

---

Due to the fact that placement will now ensure full 3 stage locks on
dissemination, and the existing placement client will only report every
second, dissemination now takes a full 3 seconds to complete in the
normal case. In the next PR, the placement client will be updated to
report immediately after receiving an operation, so will see
dissemination drop down to the ms range again. This will make testing
slow in the mean time.

The client has not been updated yet to ensure backwards compatibility
and to keep the PR size manageable.

### DIFF
--- a/pkg/placement/internal/loops/disseminator/host.go
+++ b/pkg/placement/internal/loops/disseminator/host.go
@@ -37,7 +37,7 @@ func (d *disseminator) handleReportedHost(ctx context.Context, report *loops.Rep
 	}
 
 	log.Debugf("Received report from stream (idx:%d) (ns=%s) (appID=%s) (op=%s) (ver=%d)",
-		report.StreamIDx, d.namespace, report.Host.Id, op.String(), d.currentVersion)
+		report.StreamIDx, d.namespace, report.Host.GetId(), op.String(), d.currentVersion)
 
 	switch op {
 	case v1pb.HostOperation_REPORT:


charts/dapr/README.md
```diff
@@ -166,7 +166,7 @@ The Helm chart has the follow configuration options that can be supplied:
...
-| `dapr_placement.disseminateTimeout` | Sets the timeout period for dissemination to be delayed after actor membership change (usually related to pod restarts) so as to avoid excessive dissemination during multiple pod restarts. Higher values will reduce the frequency of dissemination, but delay the table dissemination. Accepts values between `1s` and `3s` | `2s` |
+| `dapr_placement.disseminateTimeout` | cccsets the period of time in which a dissemination is considered failed if not completed. Any daprds which have not responded within this time will be considered non-responsive and kicked. Higher values will increase the grace period to complete dissemination in case of network issues, but reduce the chance of skipping daprds which are slow to respond. | `8s` |
```
cicoyle:
```suggestion
| `dapr_placement.disseminateTimeout` | Sets the period of time in which a dissemination is considered failed if not completed. Any daprds which have not responded within this time will be considered non-responsive and kicked. Higher values will increase the grace period to complete dissemination in case of network issues, but reduce the chance of skipping daprds which are slow to respond. | `8s` |
```

dapr/proto/placement/v1/placement.proto
```diff
@@ -26,31 +26,66 @@ service Placement {
...
+  map<uint64, string> hosts = 1 [deprecated = true];
+
+  // Deprecated: sorted_set is no longer used.
+  repeated uint64 sorted_set = 2 [deprecated = true];
+
+  // TODO: @joshvanl: rename to hosts_map.
   map<string, Host> load_map = 3;
-  int64 total_load = 4;
+
+  // total_load is deprecated and no longer used.
```
cicoyle:
```suggestion
  // Deprecated: total_load is no longer used.
```

pkg/placement/internal/leadership/fsm/fsm.go
```diff
@@ -11,28 +11,38 @@ See the License for the specific language governing permissions and
...
-// that may modify the live state.
-type snapshot struct {
-	state *DaprHostMemberState
+type fsm struct{}
+
+func New() raft.FSM {
+	return new(fsm)
+}
+
+func (f *fsm) Apply(log *raft.Log) interface{} {
```
cicoyle: why return `interface{}` here?

pkg/placement/internal/loops/disseminator/timeout/dissemination.go
```diff
@@ -0,0 +1,31 @@
...
+import (
+	"time"
+)
+
+type dissemination struct {
+	version uint64
+	dueTime time.Time
+}
+
+func (d *dissemination) Key() uint64 {
```
cicoyle, line 25: Can this func be named version?

pkg/placement/internal/loops/disseminator/timeout/dissemination.go
```diff
@@ -0,0 +1,31 @@
...
+	return d.version
+}
+
+func (d *dissemination) ScheduledTime() time.Time {
```
cicoyle, line 29: I don't see this used - can you confirm its used? Might be the GH view on my end tho bc its a bigger PR.

pkg/placement/internal/loops/disseminator/disseminator.go
```diff
@@ -0,0 +1,236 @@
...
+		IDx:           streamIDx,
+		Add:           add,
+		NamespaceLoop: d.nsLoop,
+		Authorizer:    d.authorizer,
+	})
+
+	d.wg.Add(1)
+	go func() {
+		defer d.wg.Done()
+		_ = streamLoop.Run(ctx)
```
cicoyle, line 163: should this use a runner manager?

dapr/proto/placement/v1/placement.proto
```diff
@@ -26,31 +26,66 @@ service Placement {
...
+  int64 load = 3 [deprecated = true];
+
   repeated string entities = 4;
+
   string id = 5;
-  string pod = 6;
-  // Version of the Actor APIs supported by the Dapr runtime
-  uint32 api_level = 7;
+
+  // Deprecated, not used
```
acroca:
```suggestion
  // Deprecated: not used
```

dapr/proto/runtime/v1/state.proto
```diff
@@ -16,6 +16,7 @@ syntax = "proto3";
 package dapr.proto.runtime.v1;
 
 import "dapr/proto/common/v1/common.proto";
+import "dapr/proto/placement/v1/placement.proto";
```
acroca, line 19: Is this file actually using any of the types from `placement.proto`?

pkg/actors/internal/placement/placement.go
```diff
@@ -304,12 +304,14 @@ func (p *placement) handleLockOperation(ctx context.Context) {
 }
 
 func (p *placement) handleUpdateOperation(ctx context.Context, in *v1pb.PlacementTables) {
+	//nolint:staticcheck
```
acroca, line 307: The proto states the `api_level` is not used. Shouldn't we delete everything that uses it? Same applies to all calls to deprecated fields.

pkg/placement/internal/authorizer/authorizer.go
```diff
@@ -0,0 +1,141 @@
...
+	if (msg.Operation == nil) != (msg.Version == nil) {
+		return status.Errorf(codes.InvalidArgument, "both operation and version must be set or both must be unset")
+	}
+
+	if len(msg.GetId()) == 0 || len(msg.GetNamespace()) == 0 {
+		return status.Errorf(codes.InvalidArgument, "host ID and namespace must be provided")
+	}
+
+	// Ensure the provided host ID and namespace match those in the client SPIFFE
+	// ID.
```
acroca, line 63: Other lines in this file are longer, why limit the comment only?
```suggestion
	// Ensure the provided host ID and namespace match those in the client SPIFFE ID.
```

pkg/placement/internal/authorizer/authorizer.go
```diff
@@ -0,0 +1,141 @@
...
+	}
+
+	if host.GetId() != clientID.AppID() {
+		log.Errorf("Client ID mismatch: %s != %s", host.GetId(), clientID.AppID())
+		return status.Errorf(
+			codes.PermissionDenied,
+			"client ID %s is not allowed",
+			host.GetId(),
+		)
+	}
```
acroca: This is already checked

pkg/placement/internal/authorizer/authorizer.go
```diff
@@ -0,0 +1,141 @@
...
+
+	return nil
+}
+
+func (a *Authorizer) types(msg *v1pb.Host) error {
```
acroca: Can you find a better name for this function?

pkg/placement/internal/leadership/fsm/fsm.go
acroca: It's the `raft.FSM` interface: https://github.com/hashicorp/raft/blob/9e3b7155dd6f45b128d0679f22df32229319e81f/fsm.go#L16-L49
We could use `any` though

pkg/placement/internal/leadership/fsm/fsm.go
```diff
@@ -11,28 +11,38 @@ See the License for the specific language governing permissions and
...
 )
 
-// snapshot is used to provide a snapshot of the current
-// state in a way that can be accessed concurrently with operations
-// that may modify the live state.
-type snapshot struct {
-	state *DaprHostMemberState
+type fsm struct{}
+
+func New() raft.FSM {
```
acroca, line 24: If I'm not mistaken, this `FSM` is a noop right? Maybe we can name it `NewNoop` to make it clearer?

pkg/placement/internal/leadership/peer.go
```diff
@@ -0,0 +1,45 @@
...
+	"net"
+	"time"
+)
+
+const (
+	nameResolveRetryInterval = time.Second / 2
+	nameResolveMaxRetry      = 240
+)
+
+func (s *Leadership) tryResolveRaftAdvertiseAddr(ctx context.Context, bindAddr string) (*net.TCPAddr, error) {
```
acroca:
```suggestion
func (s *Leadership) resolveRaftAdvertiseAddr(ctx context.Context, bindAddr string) (*net.TCPAddr, error) {
```

pkg/placement/internal/leadership/leadership.go
```diff
@@ -0,0 +1,190 @@
...
+			HeartbeatTimeout:   2 * time.Second,
+			ElectionTimeout:    2 * time.Second,
+			CommitTimeout:      100 * time.Millisecond,
+			MaxAppendEntries:   64,
+			ShutdownOnRemove:   true,
+			TrailingLogs:       1000,
+			SnapshotInterval:   30 * time.Second,
+			SnapshotThreshold:  2000,
+			LeaderLeaseTimeout: 2 * time.Second,
+		}
```
acroca, line 128: I think the timeouts are too optimistic. Some common network topologies will have a latency across nodes of about 4-5 milliseconds in average. Also some cloud providers have frequent spikes in latency getting to 10ms average for short moments. Do you think it's something we can increase to something like 20ms? Or maybe make it configurable?

pkg/placement/internal/leadership/leadership.go
```diff
@@ -0,0 +1,190 @@
...
+			if ra.State() == raft.Leader {
+				select {
+				case <-s.leaderCh:
+				default:
+					close(s.leaderCh)
+				}
+			} else {
+				select {
+				case <-s.leaderCh:
+					return errors.New("lost leadership")
```
acroca, line 175: What will happen with the leader when it loses leadership? will it shut down? If I understand the code correctly, the `Run` will terminate with this error.

pkg/placement/internal/leadership/fsm/fsm.go
JoshVanL: FSM is noop as we don't use raft as a storage layer

pkg/placement/internal/loops/disseminator/timeout/dissemination.go
JoshVanL: No, it is part of the processor quable implementation

https://github.com/dapr/kit/blob/3ac186dff64dc94a98fe04d21180e61130511301/events/queue/queue.go#L22

pkg/placement/internal/loops/disseminator/timeout/dissemination.go
JoshVanL: It's used as the timeout queue. It queues up disseminations for timeouts.

pkg/placement/internal/loops/disseminator/disseminator.go
JoshVanL: No, this needs to go in a go routine to unblock the disseminator loop

dapr/proto/runtime/v1/state.proto
JoshVanL: Yes, `  repeated Host hosts = 1;`

pkg/actors/internal/placement/placement.go
JoshVanL: Yep, but this PR is too large already.

We should also remove everything to do with "loads".

The next PR is tackling the client.

pkg/placement/internal/authorizer/authorizer.go
JoshVanL: My comments wrap at 80.

pkg/placement/internal/authorizer/authorizer.go
JoshVanL: 👍

pkg/placement/internal/leadership/fsm/fsm.go
JoshVanL: Sure 👍

pkg/placement/internal/leadership/leadership.go
JoshVanL: I don't dissagree at all and we can change, however these are the values we were using already.

Let's tune in a follow up in a focused PR.

pkg/placement/internal/leadership/leadership.go
JoshVanL: Yep, placement shuts down

pkg/placement/internal/loops/disseminator/disseminator.go
```diff
@@ -0,0 +1,236 @@
...
+		})
+	}
+}
+
+// handleShutdown handles the shutdown of the streams.
+func (d *disseminator) handleShutdown() {
+	defer d.wg.Wait()
+
+	for _, stream := range d.streams {
+		stream.loop.Close(new(loops.StreamShutdown))
```
acroca: Unrelated. If loops had a separate `Close` and `Wait`, we could signal all streams to close and then wait for all of them to be closed instead of going one by one.

pkg/placement/internal/server/server.go
```diff
@@ -0,0 +1,236 @@
...
+
+	ctx, cancel := context.WithCancelCause(ctx)
+	s.loop = namespaces.New(namespaces.Options{
+		CancelPool:           cancel,
+		ReplicationFactor:    s.replicationFactor,
+		Authorizer:           s.authz,
+		DisseminationTimeout: s.disseminateTimeout,
+	})
+
+	close(s.readyCh)
```
acroca: Can this lead to a race condition? We're closing the channel but nothing is really running yet... Like for example, further down we check for the `isLeader`, which is set in one of the runners so it will never read `true` if the call to `StatePlacementTables` is waiting for this channel to close...

tests/integration/suite/daprd/placement/multiple/actors.go
```diff
@@ -0,0 +1,127 @@
...
+	assert.EventuallyWithT(t, func(c *assert.CollectT) {
+		tables := a.actors[0].Placement().PlacementTables(t, ctx)
+		if !assert.Contains(c, tables.Tables, "default") {
+			return
+		}
+		assert.Equal(c, expHosts, tables.Tables["default"].Hosts)
+		assert.Equal(c, uint64(4), tables.Tables["default"].Version)
+	}, time.Second*20, time.Millisecond*10)
+
+	cancel()
```
acroca, line 114: it's already in the `Cleanup`

tests/integration/suite/daprd/workflow/get/crossactivity.go
```diff
@@ -89,10 +89,10 @@ func (c *crossactivity) Run(t *testing.T, ctx context.Context) {
 
 	assert.NotNil(t, evs[3].GetOrchestratorStarted())
 
-	assert.NotNil(t, evs[4].GetTaskCompleted())
-	assert.Equal(t, c.workflow.DaprN(0).AppID(), evs[4].GetRouter().GetSourceAppID())
-	assert.Equal(t, c.workflow.DaprN(1).AppID(), evs[4].GetRouter().GetTargetAppID())
+	assert.NotNil(t, evs[5].GetTaskCompleted())
```
acroca: Should we check `evs[4]`?

tests/integration/suite/daprd/workflow/get/crossworkflow.go
```diff
@@ -90,10 +90,10 @@ func (c *crossworkflow) Run(t *testing.T, ctx context.Context) {
 
 	assert.NotNil(t, evs[3].GetOrchestratorStarted())
 
-	assert.NotNil(t, evs[4].GetSubOrchestrationInstanceCompleted())
-	assert.Equal(t, c.workflow.DaprN(1).AppID(), evs[4].GetRouter().GetSourceAppID())
-	assert.Equal(t, c.workflow.DaprN(0).AppID(), evs[4].GetRouter().GetTargetAppID())
+	assert.NotNil(t, evs[5].GetSubOrchestrationInstanceCompleted())
```
acroca: Same

tests/integration/suite/placement/quorum/insecure.go
```diff
@@ -139,16 +139,18 @@ func (i *insecure) Run(t *testing.T, ctx context.Context) {
...
-		Name:      "app-1",
-		Namespace: "default",
-		Port:      1234,
-		Load:      1,
-		Entities:  []string{"entity-1", "entity-2"},
-		Id:        "app-1",
-		Pod:       "pod-1",
-	})
-	require.NoError(t, err)
+	for range 4 {
```
acroca: why 4?

tests/integration/suite/placement/quorum/jwks.go
```diff
@@ -150,52 +150,41 @@ func (j *jwks) Run(t *testing.T, ctx context.Context) {
...
+			if p.IsLeader(t, ctx) {
+				leader = p
+				break
+			}
 		}
-		t.Cleanup(func() { require.NoError(t, conn.Close()) })
-		client := v1pb.NewPlacementClient(conn)
+		assert.NotNil(c, leader, "no leader found")
+	}, time.Second*20, time.Millisecond*10)
+	require.NotNil(t, leader, "no leader found")
```
acroca: This is not necessary if we convert the 'eventually' to be `require`

tests/integration/suite/placement/quorum/notls.go
```diff
@@ -63,52 +61,36 @@ func (n *notls) Run(t *testing.T, ctx context.Context) {
...
-		if err != nil {
-			return false
-		}
-		_, err = stream.Recv()
-		if err != nil {
-			return false
-		}
-		return true
-	}, time.Second*10, time.Millisecond*10)
+	require.NotNil(t, leader, "no leader found")
```
acroca: Same

pkg/placement/internal/loops/disseminator/disseminator.go
JoshVanL: Yeah makes sense :+1:

pkg/placement/internal/server/server.go
JoshVanL: I have removed the `readyCh` channel  completely since we don't need it- `isLeader` can be used directly immediately, being an atomic bool.

tests/integration/suite/daprd/placement/multiple/actors.go
JoshVanL: We actually want to `cancel` here as we want the `StartWorker` client to close so that we can test the workflow actors become unregistered.

JoshVanL:
Closing in favour of https://github.com/dapr/dapr/pull/9405

PR #9377, opened 2026-02-03
Placement: Client Refactor

JoshVanL:
PR implements a placement client refactor to resolve for dissemination
events. Client is written in the control loop style for better
maintainability, ready to use against the placement server refactor/v2.

protos have been updated, in preparation for the placement server
refactor.

Locker is now implemented in control loops for better memory management
and contention. Actor not found errors should now be further contained
inside the runtime, and recover faster.

### DIFF
--- a/pkg/actors/internal/placement/loops/disseminator/inflight/inflight.go
+++ b/pkg/actors/internal/placement/loops/disseminator/inflight/inflight.go
@@ -75,8 +75,9 @@ func (i *Inflight) Lock(err error) {
 		Error: err,
 	})
 	i.wg.Wait()
-	lock.LoopFactory.CacheLoop(i.lock)
+	lo := i.lock
 	i.lock = nil
+	lock.LoopFactory.CacheLoop(lo)
 }
 
 func (i *Inflight) Set(in *v1pb.PlacementTables) {


pkg/actors/internal/placement/loops/disseminator/disseminator.go
```diff
@@ -0,0 +1,163 @@
...
+	diss.streamLoop = stream.New(ctx, stream.Options{
+		Channel:       opts.Channel,
+		PlacementLoop: opts.PlacementLoop,
+		IDx:           opts.IDx,
+	})
+
+	diss.wg.Add(1)
+	go func() {
+		_ = diss.streamLoop.Run(ctx)
+		diss.wg.Done()
```
cicoyle: Shall we defer this? And, should we log whatever errors if there is an error thrown to more easily debug?

pkg/actors/internal/placement/loops/placement/placement.go
```diff
@@ -0,0 +1,273 @@
...
+		Scheduler:            p.scheduler,
+		HTarget:              p.htarget,
+		DisseminationTimeout: p.dissTimeout,
+		Cancel:               p.cancel,
+	})
+
+	p.wg.Add(1)
+	go func() {
+		defer p.wg.Done()
+		_ = p.dissLoop.Run(ctx)
```
cicoyle: Should we log whatever errors if there is an error thrown to more easily debug?

pkg/actors/internal/placement/loops/disseminator/inflight/inflight.go
```diff
@@ -0,0 +1,193 @@
...
+}
+
+func (i *Inflight) Unlock(ctx context.Context) {
+	// Recreate the lock to allow queued requests to proceed.
+	lock := lock.New()
+	i.lock = lock
+	i.wg.Add(1)
+	go func() {
+		_ = lock.Run(ctx)
+		i.wg.Done()
```
cicoyle: Should this be deferred? And, should we log whatever errors if there is an error thrown to more easily debug?

pkg/actors/internal/placement/loops/stream/stream.go
```diff
@@ -0,0 +1,104 @@
...
+	stream.loop = LoopFactory.NewLoop(stream)
+
+	stream.wg.Add(1)
+	go func() {
+		err := stream.recvLoop()
+		stream.placeLoop.Enqueue(&loops.ConnCloseStream{
+			Error: err,
+			IDx:   stream.idx,
+		})
+		stream.wg.Done()
```
cicoyle: can this be deferred?

pkg/actors/internal/placement/loops/stream/stream.go
```diff
@@ -0,0 +1,104 @@
...
+	stream := streamCache.Get().(*stream)
+	stream.channel = opts.Channel
+	stream.placeLoop = opts.PlacementLoop
+	stream.idx = opts.IDx
+
+	stream.loop = LoopFactory.NewLoop(stream)
+
+	stream.wg.Add(1)
+	go func() {
+		err := stream.recvLoop()
```
cicoyle, line 63: I think this is an infinite loop and doesnt respect the ctx. I think we need to respect the parent ctx otherwise this go routine will go on forever - can you verify its respecting the ctx?. We likely need a `select` in the `recvLoop` that has a case for the `stream.ctx.Done()` and add ctx to the stream struct

cicoyle (changes requested):
Im still doing a review pass, but thought I'd batch it to speed things up and not all at the end of my review. Here are my first few comments from going thru it.

pkg/actors/internal/placement/loops/stream/stream.go
JoshVanL: It respects the parent context because if that is cancelled, then the disseminator will close the stream look which will close the stream- this in turn causes an error to be returned from the `stream.Recv()`. We cannot select on the `Recv()` because it is a blocking func.

cicoyle: oh yes yes, I see that now. ok, resolving this since its fine as-is

cicoyle (approved):
lgtm, great refactor 🎉 🙌🏻

PR #9399, opened 2026-02-05
[1.17] feat: secretRef for configuration otel headers

Eileen-Yu:
# Description

Adds secretRef support to configuration's `OtelSpec` so users can reference k8s secrets instead of putting API keys in plaintext. The operator fetches and resolves these secrets before returning the configuration to the runtime.

## Issue reference

## Checklist

- [x] Code compiles correctly
- [x] Created/updated tests
- [x] Unit tests passing


JoshVanL (changes requested):
Please generate CRDs (make generate-manifest, then copy out of /config) + also add integration tests

jjcollinge:
@JoshVanL how do you feel about this approach to secretRef... this is a very specific implementation - do we need a more general approach or happy with this for this specific use case?

JoshVanL:
> @JoshVanL how do you feel about this approach to secretRef... this is a very specific implementation - do we need a more general approach or happy with this for this specific use case?

Approach seems fine to me 👍

Eileen-Yu:
> Please generate CRDs (make generate-manifest, then copy out of /config) + also add integration tests

Thx for the reminder! updated, lmk if I missed anything

pkg/apis/configuration/v1alpha1/types.go
```diff
@@ -236,6 +238,12 @@ type OtelSpec struct {
 	Protocol        string `json:"protocol" yaml:"protocol"`
 	EndpointAddress string `json:"endpointAddress" yaml:"endpointAddress"`
 	IsSecure        *bool  `json:"isSecure" yaml:"isSecure"`
+	// +optional
+	Headers string `json:"headers,omitempty" yaml:"headers,omitempty"`
+	// +optional
+	SecretRef *common.SecretKeyRef `json:"secretRef,omitempty" yaml:"secretRef,omitempty"`
+	// +optional
+	Timeout int `json:"timeout,omitempty" yaml:"timeout,omitempty"`
 }
```
JoshVanL, line 248: Optional fields should be pointers to signal presence.
Duration fields should use that type.
Please add description comments on what these fields are/do.
Headers shouldn't just be a list (i think (?)).

```suggestion
	// +optional
	Headers Headers []common.NameValuePair `json:"headers,omitempty" yaml:"headers,omitempty"`
	// +optional
	SecretRef *common.SecretKeyRef `json:"secretRef,omitempty" yaml:"secretRef,omitempty"`
	// +optional
	Timeout *time.Duration `json:"timeout,omitempty" yaml:"timeout,omitempty"`
}
```

pkg/operator/api/configurations.go
```diff
@@ -44,3 +52,34 @@ func (a *apiServer) GetConfiguration(ctx context.Context, in *operatorv1pb.GetCo
 		Configuration: b,
 	}, nil
 }
+
+func processConfigurationSecrets(ctx context.Context, config *configurationapi.Configuration, namespace string, kubeClient client.Client) error {
+	if config.Spec.TracingSpec != nil && config.Spec.TracingSpec.Otel != nil && config.Spec.TracingSpec.Otel.SecretRef != nil {
```
JoshVanL:
```suggestion
	if config.Spec.TracingSpec == nil || config.Spec.TracingSpec.Otel == nil ||config.Spec.TracingSpec.Otel.SecretRef == nil {
	return nil
}
```

pkg/operator/api/configurations.go
```diff
@@ -44,3 +52,34 @@ func (a *apiServer) GetConfiguration(ctx context.Context, in *operatorv1pb.GetCo
...
+		secretRef := config.Spec.TracingSpec.Otel.SecretRef
+
+		if secretRef.Name != "" {
```
JoshVanL: Remove this, this field is present as this point.

```suggestion
```

pkg/operator/api/configurations.go
```diff
@@ -44,3 +52,34 @@ func (a *apiServer) GetConfiguration(ctx context.Context, in *operatorv1pb.GetCo
...
+				Namespace: namespace,
+			}, &secret)
+			if err != nil {
+				return fmt.Errorf("failed to get secret %s for OTel headers: %w", secretRef.Name, err)
+			}
+
+			key := secretRef.Key
+			if key == "" {
+				key = secretRef.Name
+			}
```
JoshVanL: Let's not do this.
```suggestion
```

pkg/apis/configuration/v1alpha1/types.go
Eileen-Yu: shouldn't these be consistent with the internal config https://github.com/dapr/dapr/blob/release-1.17/pkg/config/configuration.go#L274 🤔 

guess these fields already work in the runtime, and this PR just exposing them in the CRD?

pkg/operator/api/configurations.go
Eileen-Yu: have it here bc component uses the same pattern https://github.com/dapr/dapr/blob/release-1.17/pkg/operator/api/components.go#L147

Just concerned if this would be inconsistent and break user expectations, if it's fine then I can definitely remove it!

pkg/apis/configuration/v1alpha1/types.go
JoshVanL: We can change those internal config types

Eileen-Yu: updated, both the CRD and internal config now use the same types for consistency:
-  `Headers []common.NameValuePair`
- `Timeout *metav1.Duration`

pkg/operator/api/configurations.go
Eileen-Yu: updated to return error if key is left as empty

pkg/config/configuration.go
```diff
@@ -276,10 +278,10 @@ type OtelSpec struct {
...
 	// Defaults to true
 	IsSecure *bool `json:"isSecure,omitempty" yaml:"isSecure,omitempty"`
-	// Headers to add to the request
-	Headers string `json:"headers,omitempty" yaml:"headers,omitempty"`
-	// Timeout for the request in milliseconds
-	Timeout int `json:"timeout,omitempty" yaml:"timeout,omitempty"` // Defaults to 10000
+	// Headers to add to the OTLP trace exporter request
+	Headers []commonapi.NameValuePair `json:"headers,omitempty" yaml:"headers,omitempty"`
+	// Timeout for the OTLP trace exporter request
+	Timeout *metav1.Duration `json:"timeout,omitempty" yaml:"timeout,omitempty"`
```
JoshVanL: Internal API does not need kube types or values by reference.

```suggestion
	// Headers to add to the OTLP trace exporter request
	Headers []string`json:"headers,omitempty" yaml:"headers,omitempty"`
	// Timeout for the OTLP trace exporter request
	Timeout *time.Duration `json:"timeout,omitempty" yaml:"timeout,omitempty"`
```

Eileen-Yu:
thx @JoshVanL ! I've updated base on your review, also added integration tests. See CI failed with all integration tc, is this expected? As they passed on my local:
```
# for k8s
❯ go test -v -tags integration -count=1 -timeout 20m ./tests/integration/ -run "Test_Integration" -args -focus="operator/api/getconfiguration/secretref"
...
--- PASS: Test_Integration (14.20s)
    --- PASS: Test_Integration/build_binaries (14.20s)
    --- PASS: Test_Integration/operator/api/getconfiguration/secretref (2.21s)
        --- PASS: Test_Integration/operator/api/getconfiguration/secretref/run (1.69s)
            --- PASS: Test_Integration/operator/api/getconfiguration/secretref/run/configuration_secrets_resolved_and_types_converted (0.21s)
PASS
ok      github.com/dapr/dapr/tests/integration  17.164s
```

```
# for standalone
❯ go test -v -tags integration -count=1 -timeout 20m ./tests/integration/ -run "Test_Integration" -args -focus="daprd/tracing/otel/headers"
...
--- PASS: Test_Integration (1.53s)
    --- PASS: Test_Integration/build_binaries (1.53s)
    --- PASS: Test_Integration/daprd/tracing/otel/headers (7.58s)
        --- PASS: Test_Integration/daprd/tracing/otel/headers/run (7.56s)
            --- PASS: Test_Integration/daprd/tracing/otel/headers/run/standalone_config_with_headers_and_timeout_exports_traces (5.01s)
PASS
ok      github.com/dapr/dapr/tests/integration  9.767s
```

pkg/operator/api/configurations.go
```diff
@@ -36,11 +38,114 @@ func (a *apiServer) GetConfiguration(ctx context.Context, in *operatorv1pb.GetCo
...
+
+	tracing, ok := spec["tracing"].(map[string]interface{})
+	if !ok {
+		return b, nil
+	}
+
+	otel, ok := tracing["otel"].(map[string]interface{})
+	if !ok {
+		return b, nil
+	}
```
JoshVanL: We know the spec, why is any of the `any` json stuff necessary?

tests/integration/suite/daprd/tracing/otel/headers.go
```diff
@@ -0,0 +1,106 @@
...
+    samplingRate: "1.0"
+    otel:
+      endpointAddress: %s
+      protocol: grpc
+      isSecure: false
+      headers:
+        - "x-api-key=test-api-key-123"
+        - "x-custom-header=custom-value"
+      timeout: 30s
+`, h.collector.OTLPGRPCAddress())
```
JoshVanL, line 70: What happens when I set both the config and env vars?

tests/integration/suite/daprd/tracing/otel/headers.go
```diff
@@ -0,0 +1,106 @@
...
+		defer resp.Body.Close()
+		assert.Equal(t, http.StatusOK, resp.StatusCode)
+
+		assert.Eventually(t, func() bool {
+			spans := h.collector.GetSpans()
+			return len(spans) > 0
+		}, time.Second*20, time.Millisecond*100, "should receive spans with custom headers configured")
+
+		spans := h.collector.GetSpans()
+		require.NotEmpty(t, spans, "Should have received spans")
```
JoshVanL: Are we checking the headers are being sent?

tests/integration/suite/operator/api/getconfiguration/secretref.go
```diff
@@ -0,0 +1,152 @@
...
+		framework.WithProcesses(s.kubeapi, s.sentry, s.op),
+	}
+}
+
+func (s *secretref) Run(t *testing.T, ctx context.Context) {
+	s.sentry.WaitUntilRunning(t, ctx)
+	s.op.WaitUntilRunning(t, ctx)
+
+	client := s.op.Dial(t, ctx, s.sentry, "myapp")
+
```
JoshVanL, line 156: We should also be running with daprd to show it is being sent correctly, (and checking headers).

pkg/operator/api/configurations.go
Eileen-Yu: Good point! The `map[string]interface{}` was there because the CRD uses `[]NameValuePair + *metav1.Duration` while the internal config uses `[]string + *time.Duration`, and I couldn't marshal one directly into the other. 

Refactored to deep-copy the CRD, clear those two fields, unmarshal into the typed `config.Configuration`, then set the converted fields directly on the struct.  Let me know if you'd prefer a different approach!

tests/integration/suite/daprd/tracing/otel/headers.go
Eileen-Yu: They are mutually exclusive - if set in the config, it returns early and env vars are not applied at all. The env vars only take effect when no endpoint is configured in yaml, so they don't conflict.

tests/integration/suite/daprd/tracing/otel/headers.go
Eileen-Yu: updated test

tests/integration/suite/operator/api/getconfiguration/secretref.go
Eileen-Yu: updated test

tests/integration/suite/daprd/tracing/otel/headers.go
Eileen-Yu: added an additional tc for this

pkg/operator/api/configurations.go
```diff
@@ -36,11 +39,106 @@ func (a *apiServer) GetConfiguration(ctx context.Context, in *operatorv1pb.GetCo
...
+		ic.Spec.TracingSpec.Otel = &config.OtelSpec{}
+	}
+	ic.Spec.TracingSpec.Otel.Headers = resolvedHeaders
+	if crdConfig.Spec.TracingSpec.Otel.Timeout != nil {
+		timeout := crdConfig.Spec.TracingSpec.Otel.Timeout.Duration
+		ic.Spec.TracingSpec.Otel.Timeout = &timeout
+	}
+
+	return json.Marshal(ic)
+}
```
JoshVanL: We don't want to send the internal type - we should be sending the kube API Configuration

tests/integration/suite/daprd/tracing/otel/headers.go
```diff
@@ -0,0 +1,111 @@
...
```
JoshVanL:
```suggestion
		require.EventuallyWithT(t, func(c *assert.CollectT) {
		  assert.NotEmpty(c, h.collector.GetSpans())
		}, time.Second*20, time.Millisecond*10, "should receive spans with custom headers configured")
```

tests/integration/suite/operator/api/getconfiguration/secretref.go
```diff
@@ -0,0 +1,180 @@
...
+		resp, err := client.Do(req)
+		require.NoError(t, err)
+		defer resp.Body.Close()
+		assert.Equal(t, http.StatusOK, resp.StatusCode)
+
+		assert.Eventually(t, func() bool {
+			return len(s.collector.GetSpans()) > 0
+		}, time.Second*20, time.Millisecond*100, "should receive spans")
+
+		require.NotEmpty(t, s.collector.GetSpans(), "Should have received spans")
```
JoshVanL:
```suggestion
		require.EventuallyWithT(t, func(c *assert.CollectT) {
		  assert.NotEmpty(c, s.collector.GetSpans())
		}, time.Second*20, time.Millisecond*10, "should receive spans with custom headers configured")
```

tests/integration/suite/operator/api/getconfiguration/secretref.go
```diff
@@ -0,0 +1,180 @@
...
+		// Verify the custom headers were sent in the gRPC metadata
+		md := s.collector.GetHeaders()
+		assert.Equal(t, []string{"plain-value"}, md.Get("x-plain-header"))
+		assert.Equal(t, []string{"my-secret-api-key"}, md.Get("x-api-key"))
+	})
+}
+
+func boolPtr(b bool) *bool {
+	return &b
+}
```
JoshVanL:
```suggestion
```

tests/integration/suite/operator/api/getconfiguration/secretref.go
```diff
@@ -0,0 +1,180 @@
...
+			MTLSSpec: &configapi.MTLSSpec{
+				ControlPlaneTrustDomain: "integration.test.dapr.io",
+				SentryAddress:           sen.Address(),
+			},
+			TracingSpec: &configapi.TracingSpec{
+				SamplingRate: "1",
+				Otel: &configapi.OtelSpec{
+					EndpointAddress: s.collector.OTLPGRPCAddress(),
+					Protocol:        "grpc",
+					IsSecure:        boolPtr(false),
```
JoshVanL:
```suggestion
					IsSecure:        ptr.Of(false),
```

tests/integration/suite/daprd/tracing/otel/headers.go
```diff
@@ -0,0 +1,111 @@
...
+	"github.com/dapr/dapr/tests/integration/suite"
+)
+
+func init() {
+	suite.Register(new(headers))
+}
+
+// headers tests that OTel headers and timeout are correctly passed to the
+// exporter in standalone mode.
+type headers struct {
```
JoshVanL, line 40: Please can we have a test where we have both env vars and config defined, demonstrating how the override works?

pkg/operator/api/configurations.go
Eileen-Yu: so guess we want the operator just resolves secrets in-place, and runtime unmarshals the CRD type and does the conversion?  update.

tests/integration/suite/daprd/tracing/otel/headers.go
Eileen-Yu: added!

pkg/config/configuration.go
```diff
@@ -55,12 +56,6 @@ const (
 	// Enables feature to support workflows in a clustered deployment.
 	WorkflowsClusteredDeployment Feature = "WorkflowsClusteredDeployment"
 
-	// Enables a feature to make activities send their results to workflow when
-	// the workflow is running on a different application. Useful when using
-	// cross app workflows. Ensures that activities are not retried forever if
-	// the workflow app is not available, and instead queues the result for when
-	// the workflow app is back online. Strongly recommended to always be enabled
-	// if using the same Dapr version on all daprds.
```
JoshVanL, line 63: Why is this removed?

pkg/config/configuration.go
```diff
@@ -588,6 +583,66 @@ func LoadKubernetesConfiguration(config string, namespace string, podName string
...
+	var crdConf configurationapi.Configuration
+	if err := json.Unmarshal(b, &crdConf); err != nil {
+		return nil, err
+	}
+
+	if crdConf.Spec.TracingSpec == nil || crdConf.Spec.TracingSpec.Otel == nil ||
+		(len(crdConf.Spec.TracingSpec.Otel.Headers) == 0 && crdConf.Spec.TracingSpec.Otel.Timeout == nil) {
+		conf := LoadDefaultConfiguration()
+		if err := json.Unmarshal(b, conf); err != nil {
+			return nil, err
```
JoshVanL: Surely `LoadDefaultConfiguration` does nothing here?

pkg/config/configuration.go
```diff
@@ -588,6 +583,66 @@ func LoadKubernetesConfiguration(config string, namespace string, podName string
...
+	}
+
+	crdCopy := crdConf.DeepCopy()
+	crdCopy.Spec.TracingSpec.Otel.Headers = nil
+	crdCopy.Spec.TracingSpec.Otel.Timeout = nil
+
+	cleaned, err := json.Marshal(crdCopy)
+	if err != nil {
+		return nil, err
+	}
```
JoshVanL: Why are we doing a JSON marshal loop?

pkg/config/configuration.go
```diff
@@ -588,6 +583,66 @@ func LoadKubernetesConfiguration(config string, namespace string, podName string
...
+
+	conf := LoadDefaultConfiguration()
```
JoshVanL: Same here

tests/integration/suite/daprd/tracing/otel/headers.go
```diff
@@ -0,0 +1,185 @@
...
+}
+
+func (h *headersoverride) Setup(t *testing.T) []framework.Option {
+	h.collector = otel.New(t)
+
+	handler := http.NewServeMux()
+	handler.HandleFunc("/test", func(w http.ResponseWriter, r *http.Request) {
+		w.Write([]byte(`OK`))
+	})
+	h.httpapp = prochttp.New(t, prochttp.WithHandler(handler))
```
JoshVanL: Please keep all tests to their own files.

pkg/config/configuration.go
Eileen-Yu: thx for catching, updated

JoshVanL:
@Eileen-Yu looks like tests are failing due to a panic

Eileen-Yu:
@JoshVanL previously panicking tests in CI are now green. The remaining failures seem flaky..?

pkg/config/configuration.go
```diff
@@ -284,18 +284,76 @@ type OtelSpec struct {
...
 
+// UnmarshalJSON handles both the internal config format
+// and the Kubernetes CRD format sent by the operator.
+func (o *OtelSpec) UnmarshalJSON(data []byte) error {
+	var raw struct {
+		Protocol        string          `json:"protocol,omitempty"`
+		EndpointAddress string          `json:"endpointAddress,omitempty"`
+		IsSecure        *bool           `json:"isSecure,omitempty"`
+		Headers         json.RawMessage `json:"headers,omitempty"`
+		Timeout         json.RawMessage `json:"timeout,omitempty"`
```
JoshVanL:
```suggestion
		Timeout         *json.RawMessage `json:"timeout,omitempty"`
```

JoshVanL: Why are these `json.RawMessage` and not concrete types?

pkg/config/configuration.go
```diff
@@ -284,18 +284,76 @@ type OtelSpec struct {
...
+	if err := json.Unmarshal(data, &raw); err != nil {
+		return err
+	}
+
+	o.Protocol = raw.Protocol
+	o.EndpointAddress = raw.EndpointAddress
+	o.IsSecure = raw.IsSecure
+
+	if len(raw.Headers) > 0 {
+		if err := json.Unmarshal(raw.Headers, &o.Headers); err != nil {
```
JoshVanL: Why do we need to double unmarshal again?

pkg/config/configuration.go
```diff
@@ -284,18 +284,76 @@ type OtelSpec struct {
...
+				if len(p.Value) > 0 {
+					//nolint:errcheck
+					json.Unmarshal(p.Value, &val)
+				}
+				o.Headers = append(o.Headers, p.Name+"="+val)
+			}
+		}
+	}
+
+	if len(raw.Timeout) > 0 {
```
JoshVanL:
```suggestion
	if raw.Timeout != nil {
```

pkg/config/configuration.go
```diff
@@ -284,18 +284,76 @@ type OtelSpec struct {
...
+			var s string
+			if err := json.Unmarshal(raw.Timeout, &s); err != nil {
+				return fmt.Errorf("failed to parse otel timeout: %w", err)
+			}
+			parsed, err := time.ParseDuration(s)
+			if err != nil {
+				return fmt.Errorf("invalid otel timeout duration %q: %w", s, err)
+			}
+			d = parsed
+		}
```
JoshVanL: I don't believe we need to unmarshal again here- we should be able to have a concrete type.

pkg/config/configuration.go
Eileen-Yu: thought it needs to handle both formats but it is only called from the k8s path actually.  Thanks for catching this!

PR #9405, opened 2026-02-05
Placement server refactor

JoshVanL:
This PR introduces a significant refactor of the placement server
codebase to improve maintainability, readability, and correctness.

Placement now ensures full 3 stage lock dissemination of placement
updates to all members in a namespace. This ensures that all members
have a consistent view of the placement table, and prevents issues where
members have stale placement information.

Placement now reacts immediately to client disconnects, and maintains a
correct state machine of the placement table.

Raft-
Refactor reduces the usage of raft to only use the leader election
functionality. The rest of the storage replication was not actually
being used, so this simplifies the code significantly.

Loop-based architecture -
Breaks up the single monolithic placement loop into multiple control
loops, using the same loop framework used elsewhere. These loops are:

namespaces -> [disseminator] -> [[stream]]

`stream` handles each individual stream, feeding up each report host
back to the disseminator. Handles closing the stream with error, and
authentication/authorization of secondary messages.

`disseminator` handles the dissemination of a single namespace. It
controls sending the dissimination operations to each of the streams. It
tracks the "LOCK", "UNLOCK", and "UPDATE" three stage lock operations.
If a new report comes in while an operation is in progress, it reverts
back to the "LOCK" stage with the new report.
It all handles the timeout of the dissemination of a namespace. If the
full 3 stage lock dissemintion takes longer than the configured timeout
(default 8s), then the disseminator will close the stream(s) responsible
for the timeout and starts a new dissemination.
Store keeps track of the placement table for that namespace.

`namespaces` loop handles all namespaces. It handles routing of reports
and operations to the correct disseminator. It also handles creation
and deletion of disseminators as members from namespaces are added or
removed.

### DIFF
--- a/pkg/operator/api/informer/informer.go
+++ b/pkg/operator/api/informer/informer.go
@@ -129,14 +129,12 @@ func (i *informer[T]) WatchUpdates(ctx context.Context, ns string) (<-chan *Even
 	i.lock.Unlock()
 
 	return ch, func() {
-		go func() {
-			i.lock.Lock()
-			defer i.lock.Unlock()
-			if _, ok := i.watchers[idx]; ok {
-				delete(i.watchers, idx)
-				close(ch)
-			}
-		}()
+		i.lock.Lock()
+		defer i.lock.Unlock()
+		if _, ok := i.watchers[idx]; ok {
+			delete(i.watchers, idx)
+			close(ch)
+		}
 	}, nil
 }
 
@@ -167,6 +165,8 @@ func (i *informer[T]) handleEvent(ctx context.Context, oldObj, newObj any, event
 	for _, w := range i.watchers {
 		if ev, ok := appEventFromEvent[T](w.id, event); ok {
 			select {
+			case <-ctx.Done():
+				return
 			case w.ch <- ev:
 			case <-w.ctx.Done():
 				// Watcher has disconnected, skip sending the event.


pkg/placement/internal/authorizer/authorizer.go
```diff
@@ -0,0 +1,132 @@
...
+		return status.Errorf(codes.InvalidArgument, "both operation and version must be set or both must be unset")
+	}
+
+	if len(msg.GetId()) == 0 || len(msg.GetNamespace()) == 0 {
+		return status.Errorf(codes.InvalidArgument, "host ID and namespace must be provided")
+	}
+
+	// Ensure the provided host ID and namespace match those in the client SPIFFE
+	// ID.
+	if err = a.matchID(msg, clientID); err != nil {
```
cicoyle: Why not just wrap this in the `a.security.MTLSEnabled()` check instead of checking in the func on clientID?

cicoyle: or add a nil check on `clientID` after the `a.getClientID(stream)` since at this point it is nil and does not need to be checked

pkg/placement/internal/leadership/leadership.go
```diff
@@ -0,0 +1,190 @@
...
+			ShutdownOnRemove:   true,
+			TrailingLogs:       500,
+			SnapshotInterval:   60 * time.Second,
+			SnapshotThreshold:  1000,
+			LeaderLeaseTimeout: 5 * time.Millisecond,
+		}
+	} else {
+		config = &raft.Config{
+			ProtocolVersion:    raft.ProtocolVersionMax,
+			HeartbeatTimeout:   2 * time.Second,
```
cicoyle, line 119: Were these the defaults from before?

pkg/placement/internal/loops/disseminator/disseminator.go
```diff
@@ -0,0 +1,249 @@
...
+		IDx:           streamIDx,
+		Add:           add,
+		NamespaceLoop: d.nsLoop,
+		Authorizer:    d.authorizer,
+	})
+
+	d.wg.Add(1)
+	go func() {
+		defer d.wg.Done()
+		_ = streamLoop.Run(ctx)
```
cicoyle:
```suggestion
		derr := diss.streamLoop.Run(ctx)
		if derr != nil {
			log.Errorf("Placement stream loop ended with error: %s", derr)
		}
```

pkg/placement/internal/loops/disseminator/host.go
```diff
@@ -0,0 +1,171 @@
...
+		// them now in a new dissemination cycle.
+		if len(d.waitingToDisseminate) > 0 {
+			needs := d.waitingToDisseminate[0]
+			d.waitingToDisseminate = d.waitingToDisseminate[1:]
+			d.handleAdd(ctx, needs)
+		}
+	}
+}
+
+func (d *disseminator) allStreamsHaveState(state v1pb.HostOperation) bool {
```
cicoyle: Would it make sense to track atomic counters for each state (like `reportCount`, `lockCount`, `unlockCount`, `updateCount`) and compare against `totalStreams` or `len(d.streams)`? That would turn this func from O(n) to O(1), which could matter when ppl run at scale with thousands of streams.

cicoyle (changes requested):
mostly lgtm - just a few comments -- great job 🎉

pkg/placement/internal/leadership/leadership.go
JoshVanL: Yep!

JoshVanL:
Thank you @cicoyle! Just pushed, PTAL

cicoyle (dismissed):
amazing, thank you! 

Approving, but lets ensure CI is green before merging

JoshVanL:
Thank you @cicoyle ! :pray: 

I push one more thing- resolves a deadlock during shutdown for daprd <-> operator when using hot reloading.

cicoyle (approved):

PR #9426, opened 2026-02-06
Go: upgrade to 1.25.7

JoshVanL:


mikeee:
I think golangci-lint will need to be bumped for 1.25.x, not sure how many linter fixes/config updates will be required.

PR #9456, opened 2026-02-10
Feature Gate: WorkflowsRemoteActivityReminder

JoshVanL:
When a workflow activity completes, it will send the result of the execution to the orchestrator actor via a actor invocation. A problem occurs with cross application workflows whereby, if the orchestrator is no longer available, i.e. the workflow app hosting the workflow is down, the activity actor will restart the activity execution indefinitely attempting to then resend the result.

To prevent this, a new `WorkflowsRemoteActivityReminder` feature gate is added so that the result of an activity is sent to the orchestrator via a reminder instead of an actor invocation. This ensures that if the orchestrator is unavailable, the activity will not be restarted indefinitely and instead the reminder will retry until the orchestrator is available again.

This feature gate is encouraged to always be enabled as it provides a more robust way of sending activity results to the orchestrator, but it is gated behind a feature flag for backward compatibility since older orchestrator will not be able to handle the routing on the new reminder.


JoshVanL:
cc @olitomlinson @WhitWaldo

pkg/actors/targets/workflow/activity/reminder.go
```diff
@@ -52,3 +57,37 @@ func (a *activity) createReminder(ctx context.Context, his *backend.HistoryEvent
 		Data: anydata,
 	})
 }
+
+func (a *activity) createWorkflowResultReminder(ctx context.Context, wfActorType, wfActorID string, result *backend.HistoryEvent) error {
+	const namePrefix = "activity-result-"
```
acroca: shouldn't this use `reminderPrefixActivityResult`?

pkg/actors/targets/workflow/orchestrator/reminder.go
```diff
@@ -29,6 +29,13 @@ import (
 	commonv1pb "github.com/dapr/dapr/pkg/proto/common/v1"
 )
 
+const (
+	reminderPrefixStart          = "start"
+	reminderPrefixNewEvent       = "new-event"
+	reminderPrefixTimer          = "timer-"
+	reminderPrefixActivityResult = "activity-result-"
+)
```
acroca: Maybe for a future change, but wouldn't be better to have something like a `type` in the reminders so we don't have to check prefixes?

pkg/actors/targets/workflow/orchestrator/reminder.go
```diff
@@ -29,6 +29,13 @@ import (
 	commonv1pb "github.com/dapr/dapr/pkg/proto/common/v1"
 )
 
+const (
+	reminderPrefixStart          = "start"
+	reminderPrefixNewEvent       = "new-event"
```
acroca:
```suggestion
	reminderPrefixNewEvent       = "new-event-"
```

pkg/actors/targets/workflow/orchestrator/reminder.go
JoshVanL: Yep :+1:

pkg/actors/targets/workflow/orchestrator/reminder.go
JoshVanL: This is what is was before, so I'm inclined not to change it

pkg/actors/targets/workflow/activity/execute.go
```diff
@@ -101,27 +101,38 @@ func (a *activity) executeActivity(ctx context.Context, name string, taskEvent *
...
+			return err
+		}
+
+		req := internalsv1pb.
+			NewInternalInvokeRequest(todo.AddWorkflowEventMethod).
+			WithActor(wfActorType, workflowID).
+			WithData(resultData).
+			WithContentType(invokev1.ProtobufContentType)
+		_, err = a.router.Call(ctx, req)
+	}
```
javier-aliaga, line 132: @JoshVanL have you consider creating the reminder only if it is remote and it the router call returns an error? 
`_, err = a.router.Call(ctx, req)`

Would it be possible?

JoshVanL: Indeed, perhaps we do this if the remote app ID is not in the actor table :+1:

pkg/actors/targets/workflow/activity/reminder.go
acroca: I meant to use the variables you're adding as constants. But I just noticed they live in a different package, maybe would be good to put them somewhere shared and reference from both places?

JoshVanL:
Thanks for the idea @javier-aliaga :pray:

javier-aliaga:
LGTM

pkg/actors/targets/workflow/orchestrator/add.go
```diff
@@ -22,6 +22,12 @@ import (
 	"github.com/dapr/durabletask-go/backend"
 )
 
+const (
+	reminderPrefixStart    = "start"
+	reminderPrefixNewEvent = "new-event"
+	reminderPrefixTimer    = "timer-"
+)
+
```
acroca, line 30: Should we move this to `common` for consistency?

JoshVanL: @acroca I would rather not because it isn't shared

cicoyle (approved):

PR #9492, opened 2026-02-16
fix(grpc): remove singleton connection limit for local app

### LINKED ISSUE #8671: gRPC proxy does not create new HTTP2 connections when concurrent connection cap is reached

Flern:
This change implements gRPC connection pooling for the local app channel, allowing it to scale beyond the 100 concurrent streams limit.

Fixes #8671

# Description

Removed the singleton `appClientConn` field from the gRPC manager and replaced it with access to the local `ConnectionPool`.  This frees the pool to allocate new connections when the concurrent stream limit (100) is reached. `GetAppClient()` now returns a teardown function that callers use to release connections back to the pool when no longer in use.

All callers of `GetAppClient()` — subscription delivery (`Deliver`, `DeliverBulk`), binding event dispatch (`sendBindingEventToApp`, `getSubscribedBindingsGRPC`), programmatic subscription init (`initProgrammaticSubscriptions`), and the gRPC proxy (`intercept`) — are updated to accept and `defer` the teardown function.

**Graceful shutdown fix**: Moving to pooled connections with proper acquire/release semantics exposed a latent race condition during graceful shutdown.  Previously, `Manager.Close()` (which calls `DestroyAll()` on all gRPC connections) ran as a parallel closer alongside the component closer (which drains in-flight pubsub/binding deliveries).  With the singleton approach, the permanently-leaked reference count masked this race.  With proper teardown, `DestroyAll()` could close connections while `OnTopicEvent` calls were still in-flight.  The fix moves `rt.grpc.Close()` inside the component closer, called sequentially after all components have shut down.

### Changes

- **`pkg/api/grpc/manager/manager.go`**: Replace `appClientConn` singleton with `ConnectionPool`-backed `GetAppClient()` returning a teardown function
- **`pkg/api/grpc/manager/manager_test.go`**: Add unit tests validating connection pool scaling (101 → 2 conns, 250 → 3 conns)
- **`pkg/runtime/subscription/postman/grpc/grpc.go`**: Use teardown in `Deliver` and `DeliverBulk`
- **`pkg/runtime/processor/binding/send.go`**: Use teardown in `sendBindingEventToApp` and `getSubscribedBindingsGRPC`
- **`pkg/runtime/processor/subscriber/subscriber.go`**: Use teardown in `initProgrammaticSubscriptions`
- **`pkg/messaging/grpc_proxy.go`**: Propagate teardown through proxy intercept
- **`pkg/runtime/runtime.go`**: Fix shutdown ordering to close gRPC manager after components drain
- **`tests/integration/.../grpc/connpool.go`**: Integration test sending 150 concurrent gRPC calls through daprd, validating that the connection pool scales beyond 100 streams

## Issue reference

Please reference the issue this PR will close: #8671

## Checklist

- [x] Code compiles correctly
- [x] Created/updated tests
- [x] Unit tests passing
- [x] End-to-end tests passing (2251 integration tests, 0 failures)

### Previous PR (closed / not merged)
https://github.com/dapr/dapr/pull/9283


Flern:
@JoshVanL  or @yaron2  - I could use some help on the remaining test failures.  I don't see how the changes I've got here would drive these problems.

JoshVanL:
Hi @Flern, we still need integration tests to accept this change

Flern:
Hmm... the conpool.go integration test file is suspiciously absent from the commit.  I'll track that down.  Thank you, @JoshVanL!

Flern:
@JoshVanL  - I found and committed that missing test file.  Let me know if you want more than that!

JoshVanL:
Hi @Flern this integration test still passes on master so either the changes are not making an effect, or the test doesn't fully cover the scenario.

JoshVanL:
My suspicion is that this is actually an SDK issue

Flern:
> My suspicion is that this is actually an SDK issue

No, there were two issues with the test, both mine.

**Issue 1: Wrong code path.** I put the integration test under service invocation, but the test was exercising the gRPC proxy path (`dapr-app-id` metadata). On master, `InvokeService` goes through a singleton app channel holding a single `*grpc.ClientConn`, which never hits the connection pool. The proxy path was already updated to use pooled connections, so the test passed on both branches for the wrong reason.

**Issue 2: No stream limit enforced.** The test app's gRPC server (`tests/integration/framework/process/grpc/grpc.go:84`) calls `grpc.NewServer(opts...)` with no `MaxConcurrentStreams` option. The gRPC-Go default is `math.MaxUint32` (unlimited). So even a single connection could handle all 150 test streams -- the pool was never actually needed. The `grpcMaxConcurrentStreams = 100` constant in `pool.go` is a client-side assumption used by the pool's reference counter, not a server-enforced limit.

**What I changed:**

- Fixed the test to use `InvokeService` (matching its location under `serviceinvocation/grpc/`) and added `grpc.MaxConcurrentStreams(100)` on the app server to enforce a realistic per-connection stream limit.
- Updated the gRPC app channel itself to acquire/release pooled connections per-request via `GetAppClient()`, instead of holding a singleton `*grpc.ClientConn`. This brings the `InvokeService` path in line with the proxy, binding, and pubsub callers that were already pooling.

The test now passes on the fix branch (0.15s) and fails on master with `DeadlineExceeded` (the 50 excess requests block on the saturated single connection until timeout).

Flern:
@JoshVanL - this fell off my radar.  Looks like this didn't make it into the last release.  Let me know if there's any action needed from me to proceed with getting this into the baseline code.  Seems like these integration test failures are flaky - I can't reproduce locally.

JoshVanL (approved):

marcduiker:
@holopin-bot @Flern runtime-badge Thank you! Here's a digital badge as a small token of appreciation.

PR #9505, opened 2026-02-19
e2e: Fix flakes

JoshVanL:
Fix timing flakes in e2e tests.
Fix concurrent map read/write on metadata GET.
Retry transient etcd bind already in use errors on scheduler cluster.
Revert not retrying scheduler `Cancelled` errors on mutable calls. 
Ensure all wrappers are finished with scheduler client before reloading.
Fix flake in placement test by using `ElementsMatch` not `Equal`.
Fix panic in placement test checking for nil.
Fix flake in grpc API unit test.
Fix subscription hot reload out-of-order race.
Stablize sub-orchestrator int app down test.

### DIFF
--- a/pkg/runtime/compstore/binding.go
+++ b/pkg/runtime/compstore/binding.go
@@ -13,7 +13,11 @@ limitations under the License.
 
 package compstore
 
-import "github.com/dapr/components-contrib/bindings"
+import (
+	"maps"
+
+	"github.com/dapr/components-contrib/bindings"
+)
 
 func (c *ComponentStore) AddInputBinding(name string, binding bindings.InputBinding) {
 	c.lock.Lock()
@@ -31,11 +35,7 @@ func (c *ComponentStore) GetInputBinding(name string) (bindings.InputBinding, bo
 func (c *ComponentStore) ListInputBindings() map[string]bindings.InputBinding {
 	c.lock.RLock()
 	defer c.lock.RUnlock()
-	mcopy := make(map[string]bindings.InputBinding, len(c.inputBindings))
-	for k, v := range c.inputBindings {
-		mcopy[k] = v
-	}
-	return mcopy
+	return maps.Clone(c.inputBindings)
 }
 
 func (c *ComponentStore) DeleteInputBinding(name string) {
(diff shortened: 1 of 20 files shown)


pkg/runtime/scheduler/internal/clients/clients.go
```diff
@@ -65,6 +66,8 @@ func (c *Clients) Reload(ctx context.Context, addresses []string) error {
 	c.lock.Lock()
 	defer c.lock.Unlock()
 
+	c.wg.Wait()
```
acroca: Would it be safer to do the wait in the `.close()`?

pkg/runtime/compstore/binding.go
```diff
@@ -31,7 +31,11 @@ func (c *ComponentStore) GetInputBinding(name string) (bindings.InputBinding, bo
 func (c *ComponentStore) ListInputBindings() map[string]bindings.InputBinding {
 	c.lock.RLock()
 	defer c.lock.RUnlock()
-	return c.inputBindings
+	mcopy := make(map[string]bindings.InputBinding, len(c.inputBindings))
+	for k, v := range c.inputBindings {
+		mcopy[k] = v
+	}
+	return mcopy
```
acroca: What if we had a map clone function in `kit`? Something like:
```go

func mapClone[K comparable, V any](m map[K]V) map[K]V {
	clone := make(map[K]V, len(m))
	for k, v := range m {
		clone[k] = v
	}
	return clone
}

```

pkg/runtime/scheduler/internal/clients/wrapper/wrapper.go
```diff
@@ -128,21 +128,14 @@ func (w *wrapper) DeleteByNamePrefix(ctx context.Context, req *v1pb.DeleteByName
...
 
 func (w *wrapper) call(ctx context.Context, fn apiFn) error {
-	client, err := w.clients.Next(ctx)
-	if err != nil {
-		return err
-	}
-	return fn(client)
-}
-
-func (w *wrapper) callWithRetry(ctx context.Context, fn apiFn) error {
```
acroca, line 138: I'm not sure this will fix the issue with retrying in operations like create job. We might receive a Cancelled error, and retry, which then will return a 409. [PR](https://github.com/dapr/dapr/pull/9502)

pkg/runtime/compstore/binding.go
JoshVanL: Sure

cicoyle (dismissed):

pkg/runtime/scheduler/internal/clients/wrapper/wrapper.go
JoshVanL: We should no longer get can cancelled unless it comes from the server-
The issue was getting a canclled from a local connection cancel (after the request response) which causes the unnecessary retry.

pkg/runtime/compstore/binding.go
JoshVanL: Used `maps.Clone`.

cicoyle (dismissed):

cicoyle (dismissed):

cicoyle (dismissed):

cicoyle (approved):

PR #9544, opened 2026-02-23
Integration: Breakup resiliency app tests

JoshVanL:
Break up resiliency app tests into separate files for better organization and ensure that we don't hit timeouts in slow CI runners.


cicoyle (commented):
conflicts, but otherwise lgtm

cicoyle (approved):
changes lgtm just CI seems a bit 🔴

PR #9556, opened 2026-02-25
Actors: Honor DrainOngoingCallTimeout during actor dissemination

JoshVanL:
During placement table dissemination, in-flight actor calls are now given the configured DrainOngoingCallTimeout to complete before being forcefully cancelled. Previously this was hardcoded to 2 seconds.

Changes:
- Pass DrainOngoingCallTimeout through placement loop chain to the lock
- Use pointer types for nil-check presence detection
- Default to 2 seconds when not configured
- Respect DrainRebalancedActors setting (when false, skip waiting)
- Use DrainRebalancedActors from hostconfig instead of hardcoding true

Also adds integration tests for:
- Default 2 second timeout behavior
- Custom timeout being respected
- Timeout expiration forcing cancellation
- DrainRebalancedActors=false immediate cancellation


javier-aliaga (approved):

nelson-parente (dismissed):

cicoyle (approved):
changes lgtm assuming no flakes in CI 👍🏻 thx @JoshVanL

PR #9562, opened 2026-02-26
fix: Reset bulk subscribe timer after maxMessagesCount flush

### LINKED ISSUE #9211: Issue with Dapr Bulk Subscription Timer Behavior (Runtime v1.15.4, SDK v1.14.0)
Hi Team,

I’d like to report an issue observed in Dapr’s bulk subscription mechanism related to timer behaviour when using the messagecount and timer configurations.

**Issue Summary**
Our bulk message batching is configured as follows:
maxMessagesCount = 20 → The batch should be dispatched immediately once 20 messages are received.
awaitDuration = 10s → The batch should be dispatched after 10 seconds if the message count threshold isn’t met.

**What version of Dapr?**

Dapr Runtime v1.15.4, 
Dapr SDK v1.14.0

**Expected Behavior**

In our use case, we need the timer to reset once the expected message-count batch is sent. This will ensure that we have sufficient time for the next batch to be processed, allowing us to send messages in batches according to the configured parameters.

**Actual Behavior**

After an early batch is sent upon reaching maxMessagesCount, Dapr continues to honour the original timer. Consequently, the subscriber endpoint is triggered again when the initial 10-second duration elapses, even if the next batch has not reached the message count threshold.

**Impact:**
This results in unnecessary batch triggers and the processing of partially filled batches, reducing overall efficiency

nelson-parente:
## Summary

- Fixes the bulk subscription timer not being reset after a batch is dispatched due to `maxMessagesCount` being reached
- After an early dispatch, the `awaitDuration` ticker is now reset so the next batch gets the full duration window to accumulate messages
- Adds a test verifying the timer reset behavior

## Problem

When using bulk subscriptions with `maxMessagesCount` and `awaitDuration`, after a batch is sent early (because `maxMessagesCount` was reached), Dapr continued to honor the original timer. The subscriber endpoint was triggered again when the initial `awaitDuration` elapsed, even if the next batch had not accumulated enough messages. This resulted in unnecessary batch triggers with partially filled batches.

## Fix

Added `ticker.Reset()` in `processBulkMessages` (`pkg/runtime/pubsub/default_bulksub.go`) after a count-based flush, ensuring the next batch gets the full `awaitDuration` window.

Fixes #9211

## Test plan

- [x] Added `TestProcessBulkMessagesTimerReset` test that verifies the timer is properly reset after a count-based flush
- [x] All existing tests in `pkg/runtime/pubsub/` continue to pass


JoshVanL (changes requested):
Please can we have an integration test for this.

.golangci.yml
```diff
@@ -124,12 +118,6 @@ linters:
               desc: must use github.com/dapr/kit/ptr
     dogsled:
       max-blank-identifiers: 2
```
cicoyle, line 126:
```suggestion
      max-blank-identifiers: 2
    gosec:
      excludes:
        - G703 # Path traversal via taint analysis (false positives)
        - G704 # SSRF via taint analysis (false positives)
        - G705 # XSS via taint analysis (false positives)
        - G706 # Log injection via taint analysis (false positives)
```

.golangci.yml
```diff
@@ -63,12 +63,6 @@ linters:
     - wastedassign
     - wrapcheck
     - wsl
```
cicoyle, line 65:
```suggestion
    - wsl
    # TODO Enforce the below new v2 linters later
    - embeddedstructfieldcheck
    - funcorder
    - godoclint
    - noinlineerr
    - wsl_v5
```

cicoyle (changes requested):
using the ui to revert the change in .golangci.yml

tests/integration/suite/daprd/pubsub/bulksubscribe/timerreset.go
```diff
@@ -0,0 +1,219 @@
...
+
+	reqURL := "http://" + tr.daprd.HTTPAddress() + "/v1.0/publish/mypub/timer-test"
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodPost, reqURL,
+		strings.NewReader(data))
+	require.NoError(t, err)
+
+	req.Header.Set("Content-Type", "application/json")
+
+	resp, err := tr.client.Do(req) //nolint:gosec
```
cicoyle:
```suggestion
	resp, err := tr.client.Do(req)
```

cicoyle (changes requested):

yaron2 (approved):

PR #9423, opened 2026-02-06
Constrain loop event types with scoped sealed interfaces

JoshVanL:
Replace generic `Event`/`any`with per-loop event interfaces (placement + scheduler) so loop.Interface[T] is strongly typed.

Tag events via embedded base types to restrict which loops accept them at compile time.

PR #9570, opened 2026-02-27
Operator: refactor update streams to loops

JoshVanL:
Branched from https://github.com/dapr/dapr/pull/9423

Replace shared channel maps + locks for Subscription/HTTPEndpoint
updates with per-client informer WatchUpdates loops.

Introduce generic loops client/stream + sender abstraction to marshal
resources, process secrets, and send typed events over gRPC.

Wire new subscription/http endpoint informers into API server runner;
remove operator-side cache event handlers.

Update tests to use fake informers instead of timing-based channel
assertions.


pkg/operator/api/informer/informer.go
```diff
@@ -80,6 +80,9 @@ func (i *informer[T]) Run(ctx context.Context) error {
 	var zero T
 	informer, err := i.cache.GetInformer(ctx, zero.ClientObject())
 	if err != nil {
+		if ctx.Err() != nil {
+			return nil
+		}
```
acroca, line 87: I don't see why this is needed, care to explain?

javier-aliaga (approved):

pkg/operator/api/loops/loops.go
```diff
@@ -0,0 +1,54 @@
...
+	operatorv1pb "github.com/dapr/dapr/pkg/proto/operator/v1"
+	"github.com/dapr/dapr/pkg/runtime/meta"
+)
+
+type clientbase struct{}
+
+func (*clientbase) isEventClient() {}
+
+// EventClient is the marker interface for events handled by the client loop.
+type EventClient interface{ isEventClient() }
```
acroca: Move this before the `clientbase` 🙏

pkg/operator/api/loops/client/client.go
```diff
@@ -0,0 +1,153 @@
...
+	"github.com/dapr/dapr/pkg/operator/api/loops/stream"
+	"github.com/dapr/dapr/pkg/runtime/meta"
+	"github.com/dapr/kit/concurrency"
+	"github.com/dapr/kit/events/loop"
+	"github.com/dapr/kit/logger"
+)
+
+var (
+	log              = logger.NewLogger("dapr.operator.api.loops.client")
+	LoopFactoryCache = loop.New[loops.EventClient](16)
```
acroca, line 35: Don't you think this factory cache should be instantiable and passed to the client options? Having this sort of singletons tends to make testing complicated as different test scenarios share the same cache and might interfere with each other, leading to flaky tests where order of execution matters.

pkg/operator/api/httpendpoints.go
```diff
@@ -136,74 +127,47 @@ func (a *apiServer) ListHTTPEndpoints(ctx context.Context, in *operatorv1pb.List
...
-			if !ok {
-				return nil
-			}
-			wg.Go(func() {
-				updateHTTPEndpointFunc(srv.Context(), c)
-			})
-		}
-	}
+	// Cache the client loop for reuse
+	client.CacheLoop()
```
acroca: just future-proofing, maybe move this to a `defer` after the client is created? Just in case we add early returns.

pkg/operator/api/httpendpoints.go
```diff
@@ -136,74 +127,47 @@ func (a *apiServer) ListHTTPEndpoints(ctx context.Context, in *operatorv1pb.List
...
+		Namespace:      in.GetNamespace(),
+		PodName:        in.GetPodName(),
+		KubeClient:     a.Client,
+		ProcessSecrets: processHTTPEndpointSecrets,
+	})
 
-		log.Infof("updated sidecar with http endpoint %s from pod %s/%s", e.GetName(), in.GetNamespace(), in.GetPodName())
+	// Run the client - this will block until context is done or event channel closes
+	if err := client.Run(ctx); err != nil {
+		log.Warnf("http endpoint client loop ended with error: %s", err)
```
acroca, line 167: Is it ok to swallow this error and just send it to the logs?

pkg/operator/api/informer/informer.go
JoshVanL: To prevent fatal errors in the event operator shuts down early.

pkg/operator/api/loops/client/client.go
JoshVanL: Maybe, although I think there is an argument that keeping this way is better since this is the same style which we are using for other loops implementations across the codebase.

pkg/operator/api/httpendpoints.go
JoshVanL: Yes, we don't want to return errors to the client here as they are operator API server concerns- we don't want to leak those errors to the client.

acroca (approved):

cicoyle (approved):

PR #9579, opened 2026-03-03
daprds: Adds SIGHUP handler reloader

JoshVanL:
Adds support to daprd to handle SIGHUP posix signals which will restart
    the daprd runtime in process. This is useful for reloading Configuration
    etc.

    Linking this functionality with the HotReloading feature will be added
    in a follow-up PR.

PR #9582, opened 2026-03-03
hotreload: Configuration, HTTPEndpoint, and Resiliency

JoshVanL:
Branched from https://github.com/dapr/dapr/pull/9579

Adds ConfigurationUpdate and ResiliencyUpdate gRPC streaming RPCs to the
operator API (matching the existing ComponentUpdate/HTTPEndpointUpdate
pattern), and introduces a SIGHUPReconciler that watches these streams
and triggers a runtime restart only when a resource has actually changed.

The reconciler uses the compstore to look up existing resources rather
than maintaining a separate known map, so unchanged update events are
detected via `differ.AreSame` and silently ignored.

- Add ConfigurationUpdate and ResiliencyUpdate RPCs to operator proto
- Add Configuration and Resiliency types to the differ.Resource interface
  (Kind, APIVersion, EmptyMetaDeepCopy, etc.)
- Add compstore storage for raw configapi.Configuration and
  resiliencyapi.Resiliency API objects
- Store initial config and resiliency resources in compstore at startup
- Return raw configapi.Configuration from LoadKubernetesConfiguration
- Add operator loader and store wrappers for all three resource types
- Add integration tests for all three operator update streams
- Add SIGHUP integration tests for change and no-change scenarios


cicoyle (changes requested):
Theres a ton of conflicts - mind resolving them?

cicoyle (dismissed):
lgtm just need conflicts to be resolved 👍🏻

cicoyle (dismissed):
lgtm, lets make sure version skew is 🟢 before merging

JoshVanL:
@cicoyle needed to force push :)

cicoyle (approved):

PR #9609, opened 2026-03-09
Fix Broken Perf Test Suite

cicoyle:
[Perf tests are not running.](https://github.com/dapr/dapr/actions/runs/22832715999/job/66223758666)


cicoyle:
This change fixed it bc before it was unable to even run the tests. Now its running

JoshVanL (approved):

PR #9610, opened 2026-03-09
fix: Replace deep proto.Clone with shallow copy in invoke method request/response

JoshVanL:
proto.Clone performs an expensive deep copy of the entire protobuf message tree. Since ProtoWithData only needs a new Message.Data field and the remaining fields (headers, metadata, status, trailers) are shared read-only references, a shallow struct copy avoids the overhead.


pkg/messaging/v1/invoke_method_request.go
```diff
@@ -236,8 +235,19 @@ func (imr *InvokeMethodRequest) ProtoWithData() (*internalv1pb.InternalInvokeReq
...
+		Metadata: imr.r.GetMetadata(),
+		Message: &commonv1pb.InvokeRequest{
+			Method:        imr.r.GetMessage().GetMethod(),
+			ContentType:   imr.r.GetMessage().GetContentType(),
+			HttpExtension: imr.r.GetMessage().GetHttpExtension(),
+		},
+	}
 
 	// Read the data and store it in the object
 	data, err := imr.RawDataFull()
```
acroca: We could now do the `RawDataFull` before creating the clone, and create the clone with the data directly, no?

pkg/messaging/v1/invoke_method_response.go
```diff
@@ -165,12 +164,21 @@ func (imr *InvokeMethodResponse) ProtoWithData() (*internalv1pb.InternalInvokeRe
...
+		Status:   imr.r.GetStatus(),
+		Headers:  imr.r.GetHeaders(),
+		Trailers: imr.r.GetTrailers(),
+		Message: &commonv1pb.InvokeResponse{
+			ContentType: imr.r.GetMessage().GetContentType(),
+		},
+	}
 
 	// Read the data and store it in the object
 	data, err := imr.RawDataFull()
```
acroca: same

pkg/messaging/v1/invoke_method_request.go
JoshVanL: Agree done

acroca (approved):

cicoyle (approved):

PR #9612, opened 2026-03-10
Fix workflow retention policy CRD fields

JoshVanL:
The workflow stateRetentionPolicy CRD fields (`anyTerminal`, `completed`, `failed`, `terminated`) were incorrectly defined as `type: integer, format: int64` but the Go API types use `metav1.Duration` which serializes as strings (e.g. "1s", "168h"). This caused daprd to crash in Kubernetes mode when the operator sent string-encoded durations that couldn't be unmarshaled into time.Duration (which expects int64 nanoseconds).

Changes:
- Fix CRD field types from integer/int64 to string
- Change API types from *time.Duration to *metav1.Duration
- Add custom UnmarshalJSON on internal WorkflowStateRetentionPolicy to handle metav1.Duration string format from the operator
- Add Kubernetes-mode integration test verifying retention with CRD string durations
- Add e2e test for workflow retention auto-purge

---

The Configuration CRD defined the `stateRetentionPolicy` fields (`anyTerminal`, `completed`, `failed`, `terminated`) as `type: integer, format: int64`, but the Go API types use `metav1.Duration` which serializes as strings (e.g. `"1s"`, `"168h"`). This mismatch caused Kubernetes to reject valid duration string values for these fields, and prevented the workflow state retention policy from being configured correctly via the Kubernetes Configuration CRD.

Users running Dapr in Kubernetes mode could not configure the workflow state retention policy using the Configuration CRD with human-readable duration strings like `"1s"` or `"168h"`. Kubernetes validation rejected these values because the CRD schema expected integers. Additionally, even if integer nanosecond values were used to bypass the CRD schema validation, the internal configuration deserializer could not correctly unmarshal the `metav1.Duration` string format sent by the operator, causing daprd to fail with:

```
Fatal error from runtime: error loading configuration: json: cannot unmarshal string into Go struct field WorkflowStateRetentionPolicy.spec.workflow.stateRetentionPolicy.anyTerminal of type time.Duration
```

The Configuration CRD YAML (`charts/dapr/crds/configuration.yaml`) was not regenerated after the Go API type `WorkflowStateRetentionPolicy` was updated to use `*metav1.Duration` fields.

Updated the CRD schema to use `type: string` for all `stateRetentionPolicy` fields, matching the `metav1.Duration` serialization format. Added a custom `UnmarshalJSON` method on the internal `config.WorkflowStateRetentionPolicy` struct that deserializes via the `configapi.WorkflowStateRetentionPolicy` type (which uses `*metav1.Duration`), correctly handling both the Kubernetes CRD string format and the standalone YAML format.

**This is a change that requires a CRD update.** Kubernetes does not automatically update CRDs when upgrading Dapr via Helm. You must manually update the CRDs before upgrading. See the [Kubernetes upgrade guide](https://docs.dapr.io/operations/hosting/kubernetes/kubernetes-upgrade/) for detailed instructions on how to force update CRDs.

To update CRDs manually:

```bash
kubectl apply -f https://raw.githubusercontent.com/dapr/dapr/v1.17.2/charts/dapr/crds/configuration.yaml
```


tests/e2e/workflow_retention/workflow_retention_test.go
```diff
@@ -0,0 +1,127 @@
...
+
+var tr *runner.TestRunner
+
+func TestMain(m *testing.M) {
+	utils.SetupLogs("workflow_retention")
+	utils.InitHTTPClient(true)
+
+	testApps := []kube.AppDescription{
+		{
+			AppName:             "workflowsapp-retention",
```
cicoyle: shall we make this a const since its also used below in the `externalURL := tr.Platform.AcquireAppExternalURL("workflowsapp-retention")` line?

tests/e2e/workflow_retention/workflow_retention_test.go
```diff
@@ -0,0 +1,127 @@
...
+			DaprEnabled:         true,
+			ImageName:           "e2e-workflowsapp",
+			Replicas:            1,
+			IngressEnabled:      true,
+			IngressPort:         3000,
+			DaprMemoryLimit:     "200Mi",
+			DaprMemoryRequest:   "100Mi",
+			AppMemoryLimit:      "200Mi",
+			AppMemoryRequest:    "100Mi",
+			AppPort:             -1,
```
cicoyle: why `-1` here?

cicoyle (changes requested):
mostly lgtm

tests/e2e/workflow_retention/workflow_retention_test.go
JoshVanL: Semi-random number that can be ignored

tests/e2e/workflow_retention/workflow_retention_test.go
JoshVanL: Done

cicoyle (dismissed):

cicoyle (approved):

PR #9613, opened 2026-03-10
fix: Strip trailing dot from DNS CNAME cluster domain lookup

JoshVanL:
Go's net.Resolver.LookupCNAME returns FQDNs with a trailing dot per DNS convention (e.g. "kubernetes.default.svc.cluster.local."). The cluster domain parser used strings.TrimLeft which only stripped the leading dot separator, leaving a trailing dot in the result. This caused the scheduler's etcd TLS ServerName to include the trailing dot, failing certificate SAN validation even though the expected DNS name was present in the certificate.

Changed strings.TrimLeft to strings.Trim and extracted the parsing logic into a testable helper function with unit tests.

---

Scheduler fails to start due to trailing dot in cluster domain DNS lookup

Problem

The Dapr Scheduler service fails to start in Kubernetes with a fatal error:

```
Fatal error running scheduler: failed to create etcd config: peer certificate does not contain the expected DNS name dapr-scheduler-server-1.dapr-scheduler-server.dapr-system.svc.cluster.local. got [dapr-scheduler-server-0.dapr-scheduler-server.dapr-system.svc.cluster.local dapr-scheduler-server-1.dapr-scheduler-server.dapr-system.svc.cluster.local dapr-scheduler-server-2.dapr-scheduler-server.dapr-system.svc.cluster.local]
```

Impact

The Scheduler service cannot start in any Kubernetes cluster where the DNS CNAME lookup for the cluster domain returns a fully-qualified domain name with a trailing dot (standard DNS behavior). This prevents all scheduler-based functionality including job scheduling.

Root Cause

The scheduler resolves the Kubernetes cluster domain via a DNS CNAME lookup. Per DNS convention, CNAME responses include a trailing dot (e.g. `cluster.local.`). The code only stripped leading dots from the result, leaving the trailing dot intact. This caused the etcd peer TLS server name to end with an extra dot, which did not match the certificate SANs and failed validation.

Solution

Changed `strings.TrimLeft` to `strings.Trim` to strip dots from both ends of the parsed cluster domain, ensuring the trailing dot from DNS CNAME responses is removed.


javier-aliaga (approved):

JoshVanL:
ok to test failures not related

cicoyle (dismissed):
lgtm, plz verify copilots comments then we are good to merge

cicoyle (dismissed):

cicoyle (approved):

PR #9617, opened 2026-03-10
fix: Race condition causes scheduler jobs to fire multiple times during host reconnection

JoshVanL:
The hosts loop closed gRPC client connections immediately upon receiving new scheduler host addresses, but the connector loop had not yet processed the new Connect event. This left active streams running on closed connections, causing in-flight triggers to be marked UNDELIVERABLE and re-staged. When new streams connected and registered DeliverablePrefixes, the staged jobs fired again.

Move connection lifecycle management from the hosts loop to the connector loop. The hosts loop now passes close functions via the Connect event, and the connector closes old connections only after gracefully stopping the previous cluster.


cicoyle (dismissed):

cicoyle (commented):
lgtm, plz resolve copilots comment then we can merge 👍🏻

cicoyle (approved):

Issue #9625, opened 2026-03-11
[Service Invoke] built-in resiliency retries causes OOM kill even when large files are streamed using chunked-transfer


olitomlinson:
Likely affects all versions of Dapr Runtime

**Full repro & analysis here**

https://github.com/olitomlinson/dapr-large-file-streaming

---

**Actual**

When a Service Invoke request is made between 2 daprised apps, using `Transfer-Encoding: Chunked` , the built-in retry policy forces a the **calling sidecar** to buffer the entire request payload into memory.

In large file transfer use-cases this is particularly damaging, easily leading to OOM kills as we have experienced in production.

**Expected**

Dapr recognises that a `Transfer-Encoding: Chunked` request is being made and **disables** the built in-retry policy for that request, **preventing** the calling sidecar from buffering the entire request payload into memory.

**Work around**

Create a resiliency policy that forces `maxRetries: 0`

```yaml
apiVersion: dapr.io/v1alpha1
kind: Resiliency
metadata:
  name: chunk-transfer-no-retry
spec:
  policies:
    retries:
      noRetries:
        policy: constant
        maxRetries: 0
        duration: 0s

  targets:
    apps:
      chunk-receiver:
        retry: noRetries

scopes:
- chunk-sender

```
---

We also need to consider what happens if a retry policy is applied which is specifying a positive number of `maxRetries` -- should the runtime honour it as instructed (potentially causing this same issue as outlined above), or should it be overiden to 0 as well)

olitomlinson:
As discussed with @JoshVanL

olitomlinson:
@JoshVanL can confirm this is fixed in 1.17.2-rc.1 

https://github.com/olitomlinson/dapr-large-file-streaming/blob/main/DAPR_1.17.2_RC1_STREAMING_FIX.md

PR #9627, opened 2026-03-11
Fix ServiceInvocation stream buffering

### LINKED ISSUE #9626: [Service Invoke] HTTP Responses which are streamed via chunked transfer are buffered in sidecar
Probably affects all versions of dapr

---

Full repro instructions here 

https://github.com/olitomlinson/dapr-large-file-streaming/blob/ui/STREAMING_RESPONSE_ISSUE.md

**MAKE SURE TO SWAP TO THE `UI` BRANCH AS `MAIN` IS A DIFFERENT ISSUE**

---

**Actual**

When a request is made which returns a streamed response, the **entire response** is buffered in the sidecar memory.

**Expected**

When a request is made which returns a streamed response, the entire response is NOT buffered in the sidecar memory.

JoshVanL:
Service invocation buffers entire streaming request body in memory

Problem

When sending a request with a streaming body (chunked transfer encoding) through Dapr HTTP service invocation, the sidecar buffered the entire request body in memory before forwarding it. For large payloads—such as file uploads or long-running data streams—this caused excessive memory usage and potential out-of-memory crashes.

Impact

Any HTTP service invocation request without a known `Content-Length` (e.g. chunked uploads, streamed data, piped bodies) had its entire body buffered in memory by the sending sidecar. This made Dapr unsuitable for streaming large payloads between services and could cause sidecar OOM kills in production.

Root Cause

The sidecar's retry mechanism unconditionally buffered the request body into memory so it could replay the body on retry. For streaming requests, the body cannot be replayed because it is consumed as it is read, making the buffering both unnecessary and harmful.

Solution

The sidecar now detects streaming requests (those with no known content length) and skips request body buffering entirely. Both the built-in retry logic and any user-configured resiliency retry policies are automatically bypassed for streaming requests, since retrying would require re-reading a body that has already been consumed. Non-streaming requests with a known `Content-Length` continue to support retries as before.

---

Service invocation buffers entire streaming response body in memory

Problem

When proxying HTTP responses through service invocation, the sidecar buffered the entire response body in memory before forwarding it to the caller.
For large or unbounded streaming responses, this caused excessive memory usage and potential out-of-memory crashes.

Impact

Any service invocation response with a large or streaming body could cause sidecar OOM kills, regardless of HTTP status code.
This made Dapr unsuitable for proxying streaming responses such as server-sent events, file downloads, or long-running data streams between services.

Root Cause

The sidecar's resiliency mechanism read the full response body into memory so it could evaluate whether to retry the request.
When the request itself is a stream that has already been consumed, retries are impossible regardless of the response, making the buffering unnecessary.

Solution

For streaming requests, the sidecar now forwards response bodies directly to the caller without buffering them in memory.
Resiliency features like circuit breakers continue to track failures normally.
Non-streaming requests continue to support retries and buffered error handling as before.

---

Fixes https://github.com/dapr/dapr/issues/9626 https://github.com/dapr/dapr/issues/9625

### DIFF
--- a/docs/release_notes/v1.17.2.md
+++ b/docs/release_notes/v1.17.2.md
@@ -6,6 +6,8 @@ This update includes a breaking change and bug fixes:
 - [Pub/sub messages incorrectly routed to dead-letter queue during graceful shutdown](#pubsub-messages-incorrectly-routed-to-dead-letter-queue-during-graceful-shutdown)
 - [Scheduler jobs with Drop failure policy may fire more than once during host reconnection](#scheduler-jobs-with-drop-failure-policy-may-fire-more-than-once-during-host-reconnection)
 - [Scheduler fails to start due to trailing dot in cluster domain DNS lookup](#scheduler-fails-to-start-due-to-trailing-dot-in-cluster-domain-dns-lookup)
+- [Service invocation buffers entire streaming request body in memory](#service-invocation-buffers-entire-streaming-request-body-in-memory)
+- [Service invocation buffers entire streaming error response in memory](#service-invocation-buffers-entire-streaming-error-response-in-memory)
 
 ## Workflow state retention policy CRD fields use incorrect type (Breaking Change)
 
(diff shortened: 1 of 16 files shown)


pkg/api/http/directmessaging.go
acroca: Shouldn't this `defer` be higher so it's closed even when we do early returns?

pkg/api/http/directmessaging.go
```diff
@@ -202,14 +216,24 @@ func (a *api) onDirectMessage(w http.ResponseWriter, r *http.Request) {
...
 		} else if resStatus.GetCode() < 200 || resStatus.GetCode() > 399 {
-			msg, _ := rResp.RawDataFull()
-			// Returning a `codeError` here will cause Resiliency to retry the request (if retries are enabled), but if the request continues to fail, the response is sent to the user with whatever status code the app returned.
-			return rResp, resiliency.NewCodeError(resStatus.GetCode(), codeError{
-				headers:     rResp.Headers(),
-				statusCode:  int(resStatus.GetCode()),
-				msg:         msg,
-				contentType: rResp.ContentType(),
-			})
+			if !req.IsStreamingRequest() {
```
acroca: Can't we do early return if it's a streaming request?

pkg/messaging/direct_messaging.go
```diff
@@ -212,7 +212,16 @@ func (d *directMessaging) invokeWithRetry(
 	req *invokev1.InvokeMethodRequest,
 ) (*invokev1.InvokeMethodResponse, error) {
 	if !d.resiliency.PolicyDefined(app.id, resiliency.EndpointPolicy{}) {
-		// This policy has built-in retries so enable replay in the request
+		// Streaming requests cannot be retried since the body has already
+		// been consumed, so invoke once and return immediately.
+		if req.IsStreamingRequest() {
```
acroca: Can we make this check as part of the previous `if`? The code here is doing the same as what's after that `if`, except for the warning

pkg/api/http/directmessaging.go
JoshVanL: Thanks

pkg/messaging/direct_messaging.go
JoshVanL: Thanks

pkg/api/http/directmessaging.go
JoshVanL: :+1:

pkg/api/http/directmessaging.go
acroca: That's not what I meant, I meant to do an early return if the response is streaming and not 2xx/3xx, return right away... Do we really need to go through the rest of that function?

JoshVanL: We still need to do headers, no?

docs/release_notes/v1.17.2.md
```diff
@@ -6,6 +6,8 @@ This update includes a breaking change and bug fixes:
 - [Pub/sub messages incorrectly routed to dead-letter queue during graceful shutdown](#pubsub-messages-incorrectly-routed-to-dead-letter-queue-during-graceful-shutdown)
 - [Scheduler jobs with Drop failure policy may fire more than once during host reconnection](#scheduler-jobs-with-drop-failure-policy-may-fire-more-than-once-during-host-reconnection)
 - [Scheduler fails to start due to trailing dot in cluster domain DNS lookup](#scheduler-fails-to-start-due-to-trailing-dot-in-cluster-domain-dns-lookup)
+- [Service invocation buffers entire streaming request body in memory](#service-invocation-buffers-entire-streaming-request-body-in-memory)
+- [Service invocation buffers entire streaming response body in memory](#service-invocation-buffers-entire-streaming-response-body-in-memory)
```
cicoyle, line 10:
```suggestion
```

is this dupped?

cicoyle: I see 2 sections below, can they be combined or at least titled differently?

cicoyle (changes requested):

docs/release_notes/v1.17.2.md
olitomlinson: One is request one is response 😀

cicoyle: oh I see that now :)

cicoyle (approved):

PR #9632, opened 2026-03-11
Updates dapr/components-contrib to v1.17.1

JoshVanL:
Oracle Database state store BulkGet returns HTTP 500 instead of per-key errors

Problem

When using the Oracle Database state store component, a `BulkGet` request that encountered an error for one or more keys returned an HTTP 500 error for the entire request instead of returning per-key errors alongside successful results.

Impact

Applications using `BulkGet` with the Oracle Database state store could not retrieve any results if even a single key encountered an error. Instead of receiving successful results for valid keys with per-key errors for failed keys, the entire operation failed with an HTTP 500 response.

Root Cause

The `BulkGet` implementation in the Oracle Database state store component returned a top-level error when any individual key retrieval failed, rather than collecting the error and associating it with the specific key in the response.

Solution

Updated the `BulkGet` implementation to return per-key errors in the `BulkGetResponse` items instead of returning a top-level error. Successful key retrievals are now returned alongside any per-key errors, matching the expected state store `BulkGet` contract.

Pulsar pub/sub publishes invalid JSON messages when Avro schema is configured

Problem

When the Pulsar pub/sub component was configured with an Avro schema, JSON messages were published without being validated against the schema. Invalid messages that did not conform to the Avro schema were accepted and published to the topic.

### Impact

Applications relying on Avro schema enforcement at the Pulsar pub/sub layer could publish malformed messages that did not conform to the expected schema. Downstream consumers expecting schema-compliant messages could encounter deserialization failures or data integrity issues.

Root Cause

The Pulsar pub/sub component did not validate JSON message payloads against the configured Avro schema before publishing. The schema was used only for consumer-side deserialization, not for producer-side validation.

Solution

Added JSON-to-Avro schema validation in the publish path. Before publishing, the component now validates JSON message payloads against the configured Avro schema and returns an error if the message does not conform, preventing invalid messages from being published to the topic.

Nil pointer dereference in conversation LangChain Go Kit LLM logger

Problem

The conversation component using the LangChain Go Kit could panic with a nil pointer dereference when the LLM logger was invoked.

Impact

Applications using the conversation API with the LangChain Go Kit-based component could experience unexpected crashes due to a nil pointer dereference, causing the Dapr sidecar to restart.

Root Cause

The LLM logger callback in the LangChain Go Kit conversation component was called with a nil pointer, and the logger did not perform a nil check before accessing the pointer.

Solution

Added a nil pointer check in the LangChain Go Kit LLM logger to prevent the dereference, ensuring the conversation component handles the case gracefully without panicking.

### DIFF
--- a/docs/release_notes/v1.17.2.md
+++ b/docs/release_notes/v1.17.2.md
@@ -8,6 +8,9 @@ This update includes a breaking change and bug fixes:
 - [Scheduler fails to start due to trailing dot in cluster domain DNS lookup](#scheduler-fails-to-start-due-to-trailing-dot-in-cluster-domain-dns-lookup)
 - [Service invocation buffers entire streaming request body in memory](#service-invocation-buffers-entire-streaming-request-body-in-memory)
 - [Service invocation buffers entire streaming response body in memory](#service-invocation-buffers-entire-streaming-response-body-in-memory)
+- [Oracle Database state store BulkGet returns HTTP 500 instead of per-key errors](#oracle-database-state-store-bulkget-returns-http-500-instead-of-per-key-errors)
+- [Pulsar pub/sub publishes invalid JSON messages when Avro schema is configured](#pulsar-pubsub-publishes-invalid-json-messages-when-avro-schema-is-configured)
+- [Nil pointer dereference in conversation LangChain Go Kit LLM logger](#nil-pointer-dereference-in-conversation-langchain-go-kit-llm-logger)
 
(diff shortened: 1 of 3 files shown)


cicoyle (dismissed):

cicoyle:
lgtm just a conflict

JoshVanL:
@cicoyle done!

yaron2 (approved):

PR #9638, opened 2026-03-12
Fix ServiceInvocation response stream buffering

### LINKED ISSUE #9626: [Service Invoke] HTTP Responses which are streamed via chunked transfer are buffered in sidecar

JoshVanL:
The HTTP channel's response recorder (RWRecorder) buffered the entire response body in a bytes.Buffer before returning headers to the caller. For large streaming responses, this caused the receiver sidecar's memory to spike proportionally to the response size (e.g. ~750MB for a 256MB response).

Replace RWRecorder with a streaming rwRecorder backed by io.Pipe. The middleware pipeline now runs in a goroutine, writing response data to the pipe. The main goroutine waits for headers via a channel, then returns the pipe reader as the response body — data flows through without buffering.

Also flush chunked response data on the sender sidecar by wrapping the HTTP response writer with a flushWriter that calls Flush() after each Write, ensuring data reaches the client immediately instead of sitting in Go's 4KB bufio.Writer buffer.

Needs backporting but no release notes as this fix is related to the release notes already added.

Fixes https://github.com/dapr/dapr/issues/9626


pkg/api/http/directmessaging.go
```diff
@@ -256,7 +256,15 @@ func (a *api) onDirectMessage(w http.ResponseWriter, r *http.Request) {
...
 		if !isSSE {
 			w.WriteHeader(statusCode)
-			_, rErr = io.Copy(w, reader)
+			// Use a flushing writer to ensure each chunk is sent to the
+			// client immediately. Without this, Go's HTTP server buffers
+			// the response in a 4KB bufio.Writer, preventing true
+			// streaming for chunked responses.
+			dst := io.Writer(w)
+			if f, ok := w.(http.Flusher); ok {
+				dst = &flushWriter{w: w, f: f}
```
acroca, line 265: Can't we just do `defer f.Flush()`? or move this `if` after the copy and flush directly. Won't that be easier?

pkg/channel/http/http_channel.go
```diff
@@ -245,53 +245,83 @@ func (h *Channel) sendJob(ctx context.Context, name string, data *anypb.Any) (*i
...
+		readyCh: make(chan struct{}),
+	}
 
-	execPipeline.ServeHTTP(rw, channelReq)
-	resp := rw.Result() //nolint:bodyclose
+	var handlerErr error
+
+	go func() {
+		defer pw.Close()
+		defer rw.signalReady()
```
acroca: Don't we want to close first?

```suggestion
		defer rw.signalReady()
		defer pw.Close()
```

pkg/channel/http/rwrecorder.go
```diff
@@ -1,111 +1,50 @@
...
 )
 
-type RWRecorder struct {
+// rwRecorder writes response body data to an io.PipeWriter and signals when
+// headers and status code are available.
+type rwRecorder struct {
 	statusCode int
 	h          http.Header
-	W          io.ReadWriter
+	pw         *io.PipeWriter
```
acroca: We could use `io.Writer` instead to make it less implementation specific no?

pkg/api/http/directmessaging.go
JoshVanL: No because this means we would only flush once, rather than on every `Write` which matters for chunked calls. Means we would continue to buffer the whole thing in memory which we are trying to avoid.

pkg/channel/http/http_channel.go
JoshVanL: Addressed

pkg/channel/http/rwrecorder.go
JoshVanL: Addressed

acroca (approved):

PR #9639, opened 2026-03-14
Fix actor placement version check

JoshVanL:
The sidecar disseminator's `UNLOCK` handler set `currentVersion` before comparing it, making the stale-version guard a no-op. Move the assignment after the check so stale UNLOCKs are correctly ignored.

Batch waiting connections on the server side into a single dissemination round instead of processing them one at a time, avoiding O(n) sequential rounds when many replicas connect at once.

---

Marked for backport to v1.17.2

---

Actor placement dissemination failures with many replicas

Problem

After upgrading to Dapr 1.17.x, deployments with many replicas (e.g. 50+) experience frequent "dissemination timeout after 8s" errors, and `/placement/state` showing only a fraction of expected hosts.

Impact

Actor invocations fail intermittently because most sidecars never receive a complete placement table. Rolling restarts and scaling events amplify the problem, making large actor deployments unstable.

Root Cause

Three issues combined to cause a cascading failure during dissemination:

1. **Stale UNLOCK version accepted**: The sidecar disseminator assigned the incoming version before comparing it against the current version, so the guard `currentVersion > version` always evaluated to false. Stale UNLOCK messages were incorrectly applied.
2. **Errors killed the disseminator permanently**: When the sidecar detected a version mismatch on UPDATE or received an unknown operation, it returned a fatal error that terminated the disseminator loop entirely. The sidecar never reconnected to the placement service and remained stuck.
3. **Sequential dissemination rounds for concurrent connections**: When many replicas connected to the placement service simultaneously while a dissemination round was in progress, each waiting connection triggered its own sequential dissemination round on completion. With N waiting replicas, this created N rounds instead of 1, causing timeouts that disconnected other sidecars and produced the cascading failure.

Solution

1. Fixed the UNLOCK version guard to compare before assignment, so stale versions are correctly rejected.
2. Changed version mismatch and unknown operation handling to cancel the stream and trigger a clean reconnection instead of killing the disseminator.
3. Batched all connections that arrive during an active dissemination round into a single round, reducing N sequential rounds to 1.


pkg/actors/internal/placement/loops/disseminator/order.go
```diff
@@ -107,14 +108,14 @@ func (d *disseminator) handleOrder(ctx context.Context, order *loops.StreamOrder
...
 
-		d.currentVersion = version
 		if d.currentVersion > version {
 			log.Errorf("Version mismatch: expected %d, got %d, ignoring unlock",
 				d.currentVersion,
 				version,
 			)
 			return nil
 		}
+		d.currentVersion = version
```
JoshVanL, line 118: This is the main fix.

pkg/actors/internal/placement/loops/disseminator/order.go
```diff
@@ -77,10 +77,11 @@ func (d *disseminator) handleOrder(ctx context.Context, order *loops.StreamOrder
 
 	case operationUpdate:
 		if d.currentVersion > version {
-			return fmt.Errorf("version mismatch: expected %d, got %d",
+			d.cancel(fmt.Errorf("version mismatch: expected %d, got %d",
 				d.currentVersion,
 				version,
-			)
+			))
+			return nil
```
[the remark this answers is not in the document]
JoshVanL: nope.

pkg/actors/table/fake/fake_test.go
```diff
@@ -0,0 +1,25 @@
...
+import (
+	"testing"
+
+	"github.com/dapr/dapr/pkg/actors/table"
+	"github.com/dapr/dapr/pkg/actors/table/fake"
+)
+
+func TestFake(t *testing.T) {
+	var _ table.Interface = fake.New()
+}
```
acroca, line 25: This is already covered in the compile-time interface check.

pkg/healthz/fake/fake_test.go
```diff
@@ -0,0 +1,25 @@
...
+import (
+	"testing"
+
+	"github.com/dapr/dapr/pkg/healthz"
+	"github.com/dapr/dapr/pkg/healthz/fake"
+)
+
+func TestFake(t *testing.T) {
+	var _ healthz.Target = fake.New()
+}
```
acroca, line 25: Same

pkg/runtime/scheduler/client/fake/fake_test.go
```diff
@@ -0,0 +1,25 @@
...
+import (
+	"testing"
+
+	schedclient "github.com/dapr/dapr/pkg/runtime/scheduler/client"
+	"github.com/dapr/dapr/pkg/runtime/scheduler/client/fake"
+)
+
+func TestFake(t *testing.T) {
+	var _ schedclient.Reloader = fake.New()
+}
```
acroca, line 25: Same

pkg/actors/table/fake/fake_test.go
JoshVanL: Removed compile time check

pkg/healthz/fake/fake_test.go
JoshVanL: Same

pkg/runtime/scheduler/client/fake/fake_test.go
JoshVanL: Same

JoshVanL:
Thanks @acroca PTAL

acroca (dismissed):

JoshVanL:
@acroca pushed an integration test fix

acroca (approved):

PR #9644, opened 2026-03-16
Fix workflow max body size

JoshVanL:
Workflow activities with large results fail with gRPC ResourceExhausted error

Problem

Workflow activities that return results larger than ~2MB fail with a `ResourceExhausted` gRPC error when scheduling the activity result reminder via the scheduler:

```
Error scheduling reminder job activity-result-XXXX due to: rpc error: code = ResourceExhausted desc = trying to send message larger than max (37950104 vs. 2097152)
```

Impact

Any workflow activity returning a result larger than the default gRPC send message size limit (~2MB) fails to deliver its result back to the parent orchestration. The orchestration hangs indefinitely waiting for the activity result, eventually timing out or stalling.

Root Cause

The scheduler gRPC client configured `MaxCallRecvMsgSize` to allow receiving large messages, but did not configure `MaxCallSendMsgSize`. This left the send-side limit at the gRPC default (~2MB). When an activity completes, its result is serialized into a reminder job request sent to the scheduler. If the activity result exceeds the default limit, the gRPC client rejects the outgoing message before it reaches the server.

Solution

Added `MaxCallSendMsgSize` to the scheduler gRPC client dial options, matching the existing `MaxCallRecvMsgSize` configuration.

### DIFF
--- a/pkg/scheduler/client/client.go
+++ b/pkg/scheduler/client/client.go
@@ -46,7 +46,10 @@ func New(ctx context.Context, address string, sec security.Handler) (schedulerv1
 	}
 
 	opts := []grpc.DialOption{
-		grpc.WithDefaultCallOptions(grpc.MaxCallRecvMsgSize(math.MaxInt32)),
+		grpc.WithDefaultCallOptions(
+			grpc.MaxCallRecvMsgSize(math.MaxInt32),
+			grpc.MaxCallSendMsgSize(math.MaxInt32),
+		),
 		grpc.WithUnaryInterceptor(unaryClientInterceptor),
 		grpc.WithKeepaliveParams(keepalive.ClientParameters{
 			Time:    time.Second * 3,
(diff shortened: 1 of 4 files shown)


pkg/scheduler/client/client.go
```diff
@@ -46,7 +46,10 @@ func New(ctx context.Context, address string, sec security.Handler) (schedulerv1
 	}
 
 	opts := []grpc.DialOption{
-		grpc.WithDefaultCallOptions(grpc.MaxCallRecvMsgSize(math.MaxInt32)),
+		grpc.WithDefaultCallOptions(
+			grpc.MaxCallRecvMsgSize(math.MaxInt32),
+			grpc.MaxCallSendMsgSize(math.MaxInt32),
+		),
```
JoshVanL, line 52: This is the fix.

yaron2 (approved):
LGTM

acroca (approved):

PR #9651, opened 2026-03-16
Workflow: Delete orphan timer reminders

JoshVanL:
`WaitForSingleEvent` with a timeout creates a timer reminder in the scheduler. When the external event arrives before the timer fires, the timer reminder was never cleaned up, leaving orphan reminders that would eventually fire unnecessarily.

Add two cleanup mechanisms:

1. Mid-execution: `deleteCancelledEventTimers` scans history to match TimerCreated events (with a Name field) against new EventRaised events, and deletes the corresponding timer reminders. Matching is case-insensitive and NotFound errors are ignored for idempotency.

2. On completion: deleteAllReminders bulk-deletes all workflow and activity reminders via DeleteByActorID when hasUnfiredTimers detects pending timers. This handles plain CreateTimer timers that have no associated event name.

---

Should be backported to 1.17

---

Workflow timer reminders not deleted when external event is received before timeout

Problem

When a workflow used `WaitForSingleEvent` with a timeout, a timer reminder was created in the scheduler. If the external event was raised before the timer fired, the timer reminder was never deleted and remained as an orphan in the scheduler until it eventually fired unnecessarily. Additionally, when a workflow completed while timers were still pending (e.g. a `CreateTimer` that had not yet fired), those timer reminders were also left behind.

Impact

Workflows using `WaitForSingleEvent` with timeouts accumulated orphan timer reminders in the scheduler. These timers would eventually fire and trigger unnecessary workflow actor invocations that were silently ignored, wasting scheduler and actor resources. For long-running workflows with many `WaitForSingleEvent` calls or long timeouts, the number of orphan reminders could grow significantly.

Root Cause

The durable task SDK completes the event task when an external event is received, but does not signal the Dapr runtime to delete the associated timer reminder. The runtime had no mechanism to detect that a timer was no longer needed because its associated event had already been received. Similarly, when a workflow completed, there was no cleanup of pending timer reminders that had not yet fired.

Solution

Added two timer cleanup mechanisms to the workflow orchestrator:

1. **Mid-execution cleanup** (`deleteCancelledEventTimers`): After each workflow execution step, the runtime scans the history for `TimerCreated` events associated with `WaitForSingleEvent` calls (identified by the `Name` field on `TimerCreated`). When a matching `EventRaised` event is found in the new events, the corresponding timer reminder is deleted from the scheduler. Event name matching is case-insensitive, and already-deleted timers (e.g. from a crash recovery) are handled gracefully by ignoring `NotFound` errors.

2. **Completion cleanup** (`deleteAllReminders`): When a workflow completes and has unfired timers (detected by comparing `TimerCreated` vs `TimerFired` event counts), all reminders for the workflow and its activities are bulk-deleted via `DeleteByActorID`. This handles timers without a `Name` field (e.g. `CreateTimer`) that cannot be matched to specific events.

### DIFF
--- a/pkg/actors/targets/workflow/orchestrator/run.go
+++ b/pkg/actors/targets/workflow/orchestrator/run.go
@@ -158,6 +158,13 @@ func (o *orchestrator) runWorkflow(ctx context.Context, reminder *actorapi.Remin
 	}
 
 	if !runtimestate.IsCompleted(rs) {
+		// Delete timer reminders for WaitForSingleEvent timers where the event has
+		// been received before the timer fired.
+		if err = o.deleteCancelledEventTimers(ctx, rs); err != nil {
+			executionStatus = diag.StatusRecoverable
+			return todo.RunCompletedFalse, wferrors.NewRecoverable(err)
+		}
+
 		if err = o.createTimers(ctx, rs.GetPendingTimers(), state.Generation); err != nil {
 			executionStatus = diag.StatusRecoverable
 			return todo.RunCompletedFalse, wferrors.NewRecoverable(err)
@@ -220,6 +227,11 @@ func (o *orchestrator) runWorkflow(ctx context.Context, reminder *actorapi.Remin
 
 	if runtimestate.IsCompleted(rs) {
 		log.Infof("Workflow Actor '%s': workflow completed with status '%s' workflowName '%s'", o.actorID, rstatus, workflowName)
+		if hasUnfiredTimers(rs) {
+			if err = o.deleteAllReminders(ctx); err != nil {
+				return todo.RunCompletedFalse, err
+			}
+		}
 		if err = o.handleRetention(ctx, rstatus); err != nil {
 			return todo.RunCompletedFalse, err
 		}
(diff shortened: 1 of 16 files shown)


pkg/actors/targets/workflow/orchestrator/run.go
```diff
@@ -158,6 +158,13 @@ func (o *orchestrator) runWorkflow(ctx context.Context, reminder *actorapi.Remin
 	}
 
 	if !runtimestate.IsCompleted(rs) {
+		// Delete timer reminders for WaitForSingleEvent timers where the event has
+		// been received before the timer fired.
+		if err = o.deleteCancelledEventTimers(ctx, rs); err != nil {
```
cicoyle, line 163: Should this be moved up or outside of the `runtimestate.IsCompleted(rs)` check? I imagine we want to delete canceled event timers regardless of whether the wf completes or not.

cicoyle (changes requested):

pkg/actors/targets/workflow/orchestrator/run.go
JoshVanL: This is covered in this block: https://github.com/dapr/dapr/pull/9651/changes/BASE..626ea30c0975d78d9bea676c69b51e6efb950542#diff-efc41f5661ab19d642030658291592cb0892422d9b9952ba988713cddf9b001dR230-R233

```go

	if runtimestate.IsCompleted(rs) {
		log.Infof("Workflow Actor '%s': workflow completed with status '%s' workflowName '%s'", o.actorID, rstatus, workflowName)
		if hasUnfiredTimers(rs) {
			if err = o.deleteAllReminders(ctx); err != nil {
			}
```

cicoyle (approved):

PR #9660, opened 2026-03-18
Integration: Fix workflow test race

JoshVanL:
The gRPC and HTTP workflow terminate tests had a race condition where close(holdCh) was called immediately after the terminate request, allowing the activity to complete normally before the termination took effect. This caused the orchestration to finish with `COMPLETED` status instead of `TERMINATED`, resulting in intermittent CI failures.

### DIFF
--- a/tests/integration/suite/daprd/workflow/terminate/grpc.go
+++ b/tests/integration/suite/daprd/workflow/terminate/grpc.go
@@ -49,6 +49,7 @@ func (g *grpcclient) Run(t *testing.T, ctx context.Context) {
 	g.workflow.WaitUntilRunning(t, ctx)
 
 	holdCh := make(chan struct{})
+	t.Cleanup(func() { close(holdCh) })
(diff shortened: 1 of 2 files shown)


tests/integration/suite/daprd/workflow/terminate/grpc.go
```diff
@@ -49,6 +49,7 @@ func (g *grpcclient) Run(t *testing.T, ctx context.Context) {
 	g.workflow.WaitUntilRunning(t, ctx)
 
 	holdCh := make(chan struct{})
+	t.Cleanup(func() { close(holdCh) })
```
acroca, line 52: Does it matter for the test if this channel is closed or not? We could just let the GC handle it no?

JoshVanL: The activity is stuck

acroca (approved):

PR #9665, opened 2026-03-20
Added support for timer origins

acroca:
Proposal: https://github.com/dapr/proposals/pull/104
Required durabletask change: https://github.com/dapr/durabletask-go/pull/78

This PR makes use of the new field `origin` from timers when deleting timers associated to external events, keeping a check on the timer name for backwards compatibitily.


pkg/actors/targets/workflow/orchestrator/timer.go
```diff
@@ -155,8 +155,15 @@ func (o *orchestrator) deleteCancelledEventTimers(ctx context.Context, rs *proto
 		newEvents,
 	} {
 		for _, e := range events {
-			if tc := e.GetTimerCreated(); tc != nil && tc.Name != nil {
-				key := strings.ToUpper(tc.GetName())
+			// TODO: We're doing `Name` matching for backwards compatibility. Make it only match with `origin.external_event` in the future and remove the fallback logic.
```
JoshVanL: Please add the Dapr version we can remove this- I guess 1.19?

JoshVanL (changes requested):

tests/integration/suite/daprd/workflow/raise/deletetimer/notimeout.go
```diff
@@ -31,9 +31,8 @@ func init() {
 	suite.Register(new(notimeout))
 }
 
-// notimeout tests that WaitForSingleEvent without a timeout (no timer created)
-// still completes correctly when the event is raised. This verifies the timer
-// deletion logic does not interfere when there is no timer to delete.
+// notimeout tests that WaitForSingleEvent without a timeout creates a
```
JoshVanL, line 34: Please can we create an int test that actually checks the history, and ensure the external event origin is set

JoshVanL (changes requested):

JoshVanL (approved):

cicoyle (approved):

PR #9666, opened 2026-03-20
Add workflow history signing for tamper detection

JoshVanL:
When mTLS is enabled, cryptographically sign workflow history events using the sidecar's SPIFFE X.509 identity. Each execution step creates a chained signature covering the new events, forming a verifiable hash chain persisted alongside the history in the actor state store.

On every state load, the full signature chain is verified — checking event digests, chain linkage, certificate validity, and chain-of-trust against the Sentry CA. If verification fails, reminders are deleted to stop retries and the workflow is reported as FAILED via the metadata API with error type SignatureVerificationFailed. History is never modified.

Catch-up signing covers previously unsigned events when a workflow transitions from a non-signing host to a signing host. Certificate rotation (new SVID, same CA) is handled transparently via a certificate table.

Signing is controlled by the WorkflowSignState feature flag (enabled by default) and requires mTLS.

```
  Event 0  Event 1    Event 2  Event 3    Event 4  Event 5
    │        │           │        │           │        │
    └───┬────┘           └───┬────┘           └───┬────┘
        ▼                    ▼                    ▼
  ┌───────────┐       ┌───────────┐       ┌───────────┐
  │  Sig 0    │──────▶│  Sig 1    │──────▶│  Sig 2    │
  │ [0, 2)    │ prev  │ [2, 4)    │ prev  │ [4, 6)    │
  │ Cert: 0   │digest │ Cert: 0   │digest │ Cert: 1   │
  └───────────┘       └───────────┘       └───────────┘
        │                                       │
        ▼                                       ▼
  ┌───────────┐                           ┌───────────┐
  │ Cert 0    │                           │ Cert 1    │
  │ SVID +    │                           │ SVID +    │
  │ issuer    │                           │ issuer    │
  └─────┬─────┘                           └─────┬─────┘
        └──────────────┬──────────────────────────┘
                       ▼
                ┌─────────────┐
                │  Sentry CA  │
                │ Trust Root  │
                └─────────────┘
```

### DIFF
--- a/pkg/config/configuration.go
+++ b/pkg/config/configuration.go
@@ -74,12 +74,6 @@ const (
 	// When enabled, daprd loads MCPServer manifests at startup,
 	// and makes them available for tool execution via workflow orchestrations.
 	MCPServerResource Feature = "MCPServerResource"
-
-	// WorkflowSignState enables cryptographic signing of workflow history
-	// state. When enabled and mTLS is active, each workflow execution's
-	// history events are signed using the app's X.509 SVID identity,
-	// creating a verifiable chain of signatures. Disabled by default.
-	WorkflowSignState Feature = "WorkflowSignState"
 )
 
 // end feature flags section


JoshVanL:
Closing in favor of https://github.com/dapr/dapr/pull/9778

PR #9669, opened 2026-03-23
fix: Injector ignores non-Dapr pods before checking service account authorization

### LINKED ISSUE #9379: false positive injection failure metrics for non-Dapr pods.
## In what area(s)?

> /area injector

## What version of Dapr?

> 1.1.x 

## Expected Behavior

Pods that do not have Dapr enabled (`dapr.io/enabled`) annotation should be ignored by the sidecar injector. They should **not** trigger error logs or increment the `sidecar_injection_failed_total` metric, regardless of which Service Account created them.

## Actual Behavior

When a pod is created without Dapr annotations, (e.g., infrastructure pods like Vault), the injector still validates the creator's Service Account against the allowed list. If the Service Account is not in the allowed list, the injector:
1. Logs an error: `service account '...' not on the list of allowed controller accounts`.
2. Increments the `dapr_injector_sidecar_injection_failed_total` metric with `reason="pod_patch"`.

This causes false-positive alerts for infrastructure components that were never meant to be Dapr-enabled.

Deploy a standard infrastructure pod (e.g., a simple Nginx pod) using a Service Account that is **not** in the Dapr Injector's `dapr-sidecar-injector` ConfigMap `allowedServiceAccounts` list. Ensure the pod has **no** `dapr.io/enabled: "true"` annotation.

You will see an error: `service account 'system:serviceaccount:default:my-sa' not on the list of allowed controller accounts`.

Check the metrics:
   ```bash
   curl http://localhost:9090/metrics | grep dapr_injector_sidecar_injection_failed_total
   ```
   You will see the counter incremented.

## Release Note

RELEASE NOTE: **FIX** Prevent false positive injection failure metrics for non-Dapr pods.

acroca:
- Fixes #9379: The sidecar injector now checks the `dapr.io/enabled` annotation **before** checking service account authorization, so non-Dapr pods are silently allowed regardless of which service account deploys them.
- Previously, deploying any pod with a non-whitelisted service account (e.g. `kubectl apply --as=...`) would log an error and increment the `sidecar_injection/failed_total` metric — even if the pod had nothing to do with Dapr.
- Refactors `handleRequest` from nested if-else chains into flat guard clauses with early returns


JoshVanL (changes requested):
Please can we have an integration test for this change?

acroca:
> Please can we have an integration test for this change?

We had no integration tests, I just added some to cover some scenarios.

tests/integration/suite/injector/already_injected.go
```diff
@@ -0,0 +1,91 @@
...
+
+	assert.EventuallyWithT(t, func(c *assert.CollectT) {
+		ar := a.injector.SendAdmission(t, ctx, review)
+		if !assert.NotNil(c, ar.Response) {
+			return
+		}
+		assert.True(c, ar.Response.Allowed)
+		assert.Empty(c, ar.Response.Patch, "should not double-inject when daprd is present")
+	}, time.Second*10, 100*time.Millisecond)
+}
```
JoshVanL: Why is this in an eventually?

tests/integration/suite/injector/dapr_disabled.go
```diff
@@ -0,0 +1,88 @@
...
+			Namespace: "default",
+			Operation: "CREATE",
+			UserInfo: authenticationv1.UserInfo{
+				Username: "system:serviceaccount:some-ns:some-sa",
+			},
+			Object: runtime.RawExtension{Raw: podBytes},
+		},
+	}
+
+	assert.EventuallyWithT(t, func(c *assert.CollectT) {
```
JoshVanL: Same for all the others. This `SendAdmission` should work first time.

tests/integration/suite/injector/dapr_disabled.go
```diff
@@ -0,0 +1,88 @@
...
+		ar := d.injector.SendAdmission(t, ctx, review)
+		if !assert.NotNil(c, ar.Response) {
+			return
+		}
+		assert.True(c, ar.Response.Allowed, "dapr-disabled pod should be allowed")
+		assert.Empty(c, ar.Response.Patch, "dapr-disabled pod should not be patched")
+	}, time.Second*10, 100*time.Millisecond)
```
JoshVanL: At least use 10 milliseconds

```suggestion
	}, time.Second*10, 10*time.Millisecond)
```

JoshVanL (changes requested):

JoshVanL (approved):

PR #9670, opened 2026-03-23
Fix stale Content-Length header forwarded

### LINKED ISSUE #9668: Actor HTTP invocation returns an empty body with stale `Content-Length` when the actor method returns `null`
## In what area(s)?

/area runtime

## What version of Dapr?

1.17.2

## Expected Behavior

When an actor method `null` (e.g. returning `None` with the Python SDK), caller HTTP actor invocation should return body `null` with `Content-Length: 4`, same as the actor app returns to daprd and callee side.

## Actual Behavior

When the actor app returns for method invocation:

- status 200
- body `null`
- `Content-Length: 4`

But the caller receives from its local daprd sidecar:

- `Content-Length: 4`
- **empty body**

This causes hard client errors, for example with Python `aiohttp`:

`aiohttp.client_exceptions.ClientPayloadError: Response payload is not completed: <ContentLengthError: 400, message='Not enough data for satisfy content length header.'>`

This only happens when the actor method returns `null`. It does not happen when the actor returns a non-empty value such as `"ok"`.

Switching daprd->app between HTTP/1.1 and HTTP/2 does not change the behavior.

This looks like daprd preserves upstream `Content-Length` but loses the `null` body while rebuilding the actor HTTP response.

1. Run an actor app behind daprd.
2. Implement an actor method that returns `null` (e.g. `None` with the Python SDK).
3. Invoke that actor method over the HTTP actor API.
4. Observe on the callee side that the actor app returns `null` with `Content-Length: 4`.
5. Observe on the caller side that local daprd returns `Content-Length: 4` with an empty body.

Control case:
6. Change the same actor method to return `"ok"` instead of `null`.
7. Observe that the response body is forwarded correctly.

*Following is unverified analysis by ChatGPT*:

> Relevant code paths in 1.17.2:
> 
> - `pkg/channel/http/http_channel.go`
> - `pkg/actors/targets/app/app.go`
> - `pkg/api/http/actors.go`
> - `pkg/messaging/v1/invoke_method_response.go`
> 
> Suspected issue:
> - upstream headers, including `Content-Length`, are preserved
> - actor response is materialized/rebuilt inside daprd
> - in the `None` / `null` case, the body appears to be dropped while `Content-Length` is kept
> 

## Release Note

RELEASE NOTE: **FIX** Actor HTTP invocation could return an empty body with stale `Content-Length` when the actor method returns `null`.

JoshVanL:
When daprd rebuilds HTTP responses from internal protobuf messages, the upstream application's Content-Length header was forwarded verbatim to the caller. If the upstream sent an incorrect Content-Length, callers received unexpected EOF errors or truncated bodies.

Skip Content-Length in InternalMetadataToHTTPHeader so Go's ResponseWriter computes the correct value from the actual body written. Update http_channel's constructRequest to read Content-Length directly from internal metadata for outgoing requests.

Fixes: https://github.com/dapr/dapr/issues/9668

### DIFF
--- a/docs/release_notes/v1.17.3.md
+++ b/docs/release_notes/v1.17.3.md
@@ -1,12 +1,58 @@
 # Dapr 1.17.3
 
-This update contains bug fixes and security fixes:
-- [Actor method invocation returns 200 with empty body over h2c](#actor-method-invocation-returns-200-with-empty-body-over-h2c)
+This update contains security and bug fixes:
 - [Security: Fixes gRPC authorization bypass - CVE-2026-33186](#security-grpc-authorization-bypass)
 - [Security: Fixes TIFF image OOM denial of service - CVE-2026-33809](#security-tiff-image-oom-denial-of-service)
+- [Service invocation and actor responses forward stale Content-Length header](#service-invocation-and-actor-responses-forward-stale-content-length-header)
+- [Actor method invocation returns 200 with empty body over h2c](#actor-method-invocation-returns-200-with-empty-body-over-h2c)
 - [False positive injection failure metrics for non-Dapr pods](#false-positive-injection-failure-metrics-for-non-dapr-pods)
 - [Placement dissemination timeout cascades across all replicas](#placement-dissemination-timeout-cascades-across-all-replicas)
 - [Daprd placement reconnect hangs for 20 seconds on stale DNS](#daprd-placement-reconnect-hangs-for-20-seconds-on-stale-dns)
+
+## Security: gRPC authorization bypass
+
+### Problem
+
+An upstream dependency (google.golang.org/grpc) used by Dapr introduced a vulnerability that could allow gRPC authorization bypass under certain conditions (CVE-2026-33186).
+
+### Impact
+
+Users running affected versions could be exposed to unauthorized gRPC requests.
+
+### Root cause
+
+The issue originated in an upstream library.
+
+### Solution
+
+This release upgrades the affected dependency to a version that resolves CVE-2026-33186. 
+
+Users are strongly encouraged to upgrade to this release.
+
+## Service invocation and actor responses forward stale Content-Length header
+
+### Problem
+
+When a Dapr sidecar forwarded HTTP responses from service invocation or actor method calls, the `Content-Length` header from the upstream application was passed through verbatim to the caller, even though the sidecar rebuilds the response body from an internal protobuf representation.
+If the upstream application sent an incorrect `Content-Length` (or the body size changed during serialization), the caller received a `Content-Length` header that did not match the actual response body.
+
+### Impact
+
+Callers reading the response body with standard HTTP clients (such as Go's `io.ReadAll` or Python's `aiohttp`) received `unexpected EOF` errors or truncated bodies when the forwarded `Content-Length` exceeded the actual body size.
+When the forwarded `Content-Length` was smaller than the actual body, clients silently truncated the response.
+This affected both service invocation (direct messaging) and actor method invocation via HTTP.
+
+### Root Cause
+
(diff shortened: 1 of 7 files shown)


cicoyle (dismissed):
lgtm just merge conflicts

JoshVanL:
@cicoyle thanks, done!

docs/release_notes/v1.17.3.md
```diff
@@ -1,56 +1,83 @@
 # Dapr 1.17.3
 
-This update contains bug fixes and security fixes:
-- [Actor method invocation returns 200 with empty body over h2c](#actor-method-invocation-returns-200-with-empty-body-over-h2c)
```
acroca, line 4: I don't think this is intentional, is it?

JoshVanL: Fixed :+1:

docs/release_notes/v1.17.3.md
```diff
@@ -4,6 +4,7 @@ This update contains bug fixes and security fixes:
 - [Actor method invocation returns 200 with empty body over h2c](#actor-method-invocation-returns-200-with-empty-body-over-h2c)
 - [Security: Fixes gRPC authorization bypass - CVE-2026-33186](#security-grpc-authorization-bypass)
 - [Security: Fixes TIFF image OOM denial of service - CVE-2026-33809](#security-tiff-image-oom-denial-of-service)
+- [Service invocation and actor responses forward stale Content-Length header](#service-invocation-and-actor-responses-forward-stale-content-length-header)
```
acroca, line 6: Just being extra picky, should this index match the order of the sections below? If so, this should move up 1 line :D

pkg/api/http/actors_contentlength_test.go
```diff
@@ -0,0 +1,121 @@
...
+		upstreamContentLength string
+		bodyToWrite           []byte
+		expectedBody          string
+	}{
+		{
+			name:                  "null body - upstream CL matches",
+			upstreamContentLength: "4",
+			bodyToWrite:           []byte("null"),
+			expectedBody:          "null",
+		},
```
acroca: I don't think this has anything to do with `null`, right?
```suggestion
			name:                  "upstream CL matches actual body",
			upstreamContentLength: "5",
			bodyToWrite:           []byte("Hello"),
			expectedBody:          "Hello",
		},
```

pkg/channel/http/http_channel.go
```diff
@@ -480,8 +483,25 @@ func (h *Channel) invokeMethodV1(ctx context.Context, req *invokev1.InvokeMethod
...
+					// Capture Content-Length for metrics before removing
+					// it from forwarded headers, so that the upstream
+					// app's stale or incorrect value is not forwarded to
+					// the caller.
+					if cl := w.Header().Get("Content-Length"); cl != "" {
+						if parsed, parseErr := strconv.ParseInt(cl, 10, 64); parseErr == nil {
+							contentLength = parsed
+						}
+					}
+					w.Header().Del("Content-Length")
```
acroca, line 495: I don't see why this is necessary, the headers are already in the `clientResp` right? `w.Header` doesn't seem to be used anymore. Same with the other places where we delete the header.

JoshVanL: w here is the rwRecorder, so w.Header() returns rw.h. That header map is used at line 541 to build the  http.Response (Header: rw.h), which then flows through parseChannelResponse. If we don't delete Content-Length from w.Header(), the stale upstream value ends up in the response metadata and gets forwarded to the caller. This is the bug this PR fixes. The same pattern applies to the other location at line 291, That rw.h is similarly used to construct the response at ~line 341.

cicoyle (approved):

PR #9676, opened 2026-03-23
Actors: h2c: actor invocation 200 with empty body

### LINKED ISSUE #9672: Actor method invocation can return 200 with `Content-Length` set but empty body in v1.17.2
## In what area(s)?

/area runtime

## What version of Dapr?

1.17.2 (**does not reproduce in 1.17.1**)

## Expected Behavior

When calling an actor method through daprd (`/v1.0/actors/<actor type>/<actor ID>/method/<method>`), the HTTP response body returned by the actor host app should be forwarded intact to the caller.

## Actual Behavior

On v1.17.2, actor method invocation can return empty body (status 200, correct `Content-Length` and `Content-Type`). For example:

```bash
curl "http://127.0.0.1:3500/v1.0/actors/SomeActor/some_actor_id/method/some_method" -d'{}' -v
```

returns:

```text
< HTTP/1.1 200 OK
< Content-Length: 50
< Content-Type: application/json; utf-8
...
* transfer closed with 50 bytes remaining to read
curl: (18) transfer closed with 50 bytes remaining to read
```

Important observations from testing:

* Reproduces on v1.17.2,  **does not reproduce on v1.17.1**
* Does not reproduce if using HTTP/1.1 as app protocol and actor method implementation return long response and `Content-Length` (not chunked). Reproduces otherwise - `h2c` proto, `null` response, chunked response.
* Reproduces most of the time, but rarely (about 1 in 5 in my tests) request does return body.

I reproduced this with two daprd instances and a remote actor invocation flow:

1. Deploy two apps with Dapr sidecars on v1.17.2, using `h2c` app protocol:

   * caller app + caller daprd
   * actor host app + actor host daprd

2. Invoke the actor method through the caller sidecar:

   ```bash
   curl "http://127.0.0.1:3500/v1.0/actors/<actor type>/<actor id>/method/<method>" -d'{}' -v
   ```

3. Observe that the response may contain:

   * HTTP 200
   * non-zero `Content-Length`
   * empty body

## Additional context

I let Claude Code investigate, _unverified_ inference from that report is:

* suspected commit: `4f430036e` (`Fix ServiceInvocation response stream buffering (#9638)`)
* suspected cause: actor invocation path may materialize the response body after the request context is already canceled, so headers are preserved but body bytes are lost

## Release Note

RELEASE NOTE: **FIX** Bug where actor method invocation could return a response with headers and Content-Length but an empty body.

JoshVanL:
With pipe-based response streaming (v1.17.2), HTTP/2 context cancellation — from resiliency timeout cleanup or placement dissemination — resets the stream before io.Copy finishes, causing the pipe to close with 0 bytes. Callers receive 200 OK with correct headers but an empty body.

Two fixes in the HTTP app channel:

1. Propagate io.Copy errors through the pipe via pw.CloseWithError instead of silently closing with EOF, so callers receive an error rather than empty body.

2. For h2c transports only, detach the HTTP request context from the caller's context (context.WithoutCancel). This prevents resiliency timeout and dissemination cancellation from resetting the HTTP/2 stream while body data is in flight. The detached context is cancelled when the pipe reader is closed to prevent goroutine leaks. HTTP/1.1 behavior is unchanged.

Fixes: https://github.com/dapr/dapr/issues/9672

---

Actor method invocation returns 200 with empty body over h2c

Problem

When using the h2c (HTTP/2 cleartext) app protocol, actor method invocations could return HTTP 200 with correct headers (including `Content-Length`) but an empty body.

Impact

Applications using `--app-protocol h2c` with actors could receive empty response bodies from actor method calls, despite the actor handler returning data. This caused silent data loss that was difficult to diagnose because the HTTP status code and headers appeared correct.

Root Cause

Dapr v1.17.2 introduced pipe-based streaming for service invocation response bodies to avoid buffering large payloads in memory. The response headers (including `Content-Length`) are captured when the pipe is ready, and the body streams lazily through an `io.Pipe` via `io.Copy`.

With HTTP/2, response body reads are tied to the request context. When the context is cancelled — either by the resiliency policy runner's `defer cancel()` after `InvokeMethod` returns, or by placement dissemination cancelling actor claims — the HTTP/2 stream is reset (`RST_STREAM`). The goroutine performing `io.Copy` from the HTTP/2 response body then fails, writing 0 bytes to the pipe. The pipe closes normally (EOF), and `ProtoWithData()` reads an empty body. The caller receives 200 OK with the original `Content-Length` header but no data.

HTTP/1.1 is unaffected because TCP buffer reads do not check the request context.

Solution

Two changes in the HTTP app channel:

1. **Pipe error propagation**: `io.Copy` errors are now captured and propagated through the pipe via `pw.CloseWithError(err)` instead of silently closing with EOF. If context cancellation causes the HTTP/2 body read to fail, callers receive an error rather than an empty body.

2. **h2c context detachment**: For HTTP/2 (h2c) transports only, the HTTP request to the app now uses a context detached from the caller's context (`context.WithoutCancel`). This prevents resiliency timeout cancellation and placement dissemination from resetting the HTTP/2 stream while body data is in flight. The detached context is cancelled when the pipe reader is closed, preventing goroutine leaks. HTTP/1.1 behavior is unchanged.

### DIFF
--- a/pkg/channel/http/reader.go
+++ b/pkg/channel/http/reader.go
@@ -36,9 +36,9 @@ func (r *cancelOnCloseReader) Close() error {
 
 // wrapPipeReaderWithCancel wraps a pipe reader so closing it also calls cancel.
 // If cancel is nil, the reader is returned as-is.
-func wrapPipeReaderWithCancel(pr *io.PipeReader, cancel context.CancelFunc) io.ReadCloser {
+func wrapReadCloserWithCancel(rc io.ReadCloser, cancel context.CancelFunc) io.ReadCloser {
 	if cancel == nil {
-		return pr
+		return rc
 	}
-	return &cancelOnCloseReader{ReadCloser: pr, cancel: cancel}
+	return &cancelOnCloseReader{ReadCloser: rc, cancel: cancel}
 }
--- a/pkg/channel/http/reader.go
+++ b/pkg/channel/http/reader.go
@@ -1,0 +1,44 @@
+/*
+Copyright 2026 The Dapr Authors
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package http
+
+import (
+	"context"
+	"io"
+	"sync"
+)
+
+// cancelOnCloseReader wraps an io.ReadCloser and calls a cancel function
+// when Close is called. This is used to cancel the detached h2c request
+// context when the pipe reader is closed, ensuring the HTTP/2 goroutine
+// does not leak.
+type cancelOnCloseReader struct {
+	io.ReadCloser
+	cancel context.CancelFunc
+	once   sync.Once
+}
+
+func (r *cancelOnCloseReader) Close() error {
+	r.once.Do(func() { r.cancel() })
+	return r.ReadCloser.Close()
+}
+
+// wrapPipeReaderWithCancel wraps a pipe reader so closing it also calls cancel.
+// If cancel is nil, the reader is returned as-is.
+func wrapPipeReaderWithCancel(pr *io.PipeReader, cancel context.CancelFunc) io.ReadCloser {
+	if cancel == nil {
+		return pr
+	}
+	return &cancelOnCloseReader{ReadCloser: pr, cancel: cancel}
+}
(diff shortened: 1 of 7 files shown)


pkg/channel/http/reader.go
```diff
@@ -0,0 +1,44 @@
...
+}
+
+// wrapPipeReaderWithCancel wraps a pipe reader so closing it also calls cancel.
+// If cancel is nil, the reader is returned as-is.
+func wrapPipeReaderWithCancel(pr *io.PipeReader, cancel context.CancelFunc) io.ReadCloser {
+	if cancel == nil {
+		return pr
+	}
+	return &cancelOnCloseReader{ReadCloser: pr, cancel: cancel}
+}
```
acroca: What about making it more generic?
```suggestion
func wrapReadCloserWithCancel(rc io.ReadCloser, cancel context.CancelFunc) io.ReadCloser {
	if cancel == nil {
		return rc
	}
	return &cancelOnCloseReader{ReadCloser: rc, cancel: cancel}
}
```

acroca (dismissed):
LGTM, just a small suggestion

acroca (approved):

PR #9686, opened 2026-03-25
[1.16] Scheduler crash fix under high load

JoshVanL:
The Scheduler process crashes with a fatal error during leadership quorum changes when a stale CloseJob event arrives at the router for a counter that no longer exists in the map. This happens because leadership changes (triggered by pod evictions, OOM kills, rolling updates, etc.) cause the cron engine to restart with a fresh router. Stale CloseJob events from the old engine's counter loops can arrive at the new router's empty counter map. Previously this returned a fatal "catastrophic state machine error: lost inner loop reference" which propagated through the event loop, cancelled the router's parent context, and killed the entire Scheduler process.

The remaining instances then see a partition count change (e.g. 3->2), triggering their own engine restarts under the same conditions, causing a cascading failure across the cluster. All workflows, scheduled jobs, and actor reminders stop firing.

The fix in go-etcd-cron (diagridio/go-etcd-cron#127) drops stale CloseJob events for missing counters with a debug log instead of returning a fatal error.

---

This fix is not needed in v1.17, which is not effected.

---

Scheduler crashes with "catastrophic state machine error" during leadership changes

Problem

The Dapr Scheduler process crashes with a fatal error during leadership quorum changes under high job throughput:

```
level=fatal msg="Fatal error running scheduler: catastrophic state machine error: lost inner loop reference"
```

Once a Scheduler instance crashes, the remaining instances experience further quorum instability, which can trigger the same crash on those instances as well, leading to a cascading failure across the entire Scheduler cluster. All workflow executions, scheduled jobs, and actor reminders stop firing until the Scheduler cluster recovers.

Impact

Any Dapr deployment running Scheduler in a multi-instance (HA) configuration is affected. The crash is triggered when a Scheduler instance restarts,  whether due to a Kubernetes pod eviction, OOM kill, node maintenance, rolling update, or transient etcd connectivity issue, while the cluster is under high job throughput. This is especially severe for deployments using Dapr Workflows, as the Scheduler is responsible for dispatching all workflow activity and orchestration timers.

Affected versions: v1.16.0 through v1.16.10.

Root Cause

The Scheduler's internal cron engine uses a router that manages per-job counter loops. When a job completes or is deleted, the counter loop emits a `CloseJob` event to the router, which removes the counter from its internal map.

During a leadership quorum change (e.g., partition count changes from 3 to 2 when a Scheduler instance goes down), the cron engine restarts. In the narrow window between the old engine shutting down and the new engine starting up, stale `CloseJob` events from the old engine's counter loops can arrive at the new engine's router. Because the new router starts with an empty counter map, these stale events reference counters that do not exist.

Previously, this condition returned a fatal error (`"catastrophic state machine error: lost inner loop reference"`), which propagated through the router's event loop, cancelled the router's parent context, and terminated the entire Scheduler process.

Solution

The router now treats a `CloseJob` event for a missing counter as a benign no-op: the stale event is dropped with a debug-level log message and the router continues processing normally. This is safe because:
1. The counter was already cleaned up — there is nothing to close.
2. The event is from a previous engine lifecycle and is no longer relevant.
3. All other `CloseJob` code paths (counter exists, counter reused with new index) are unaffected.

This fix is in the `go-etcd-cron` dependency ([diagridio/go-etcd-cron#127](https://github.com/diagridio/go-etcd-cron/pull/127)).

### DIFF
--- a/tests/integration/suite/scheduler/kubernetes/leadershipchurn.go
+++ b/tests/integration/suite/scheduler/kubernetes/leadershipchurn.go
@@ -72,35 +72,23 @@ func (n *leadershipchurn) Setup(t *testing.T) []framework.Option {
 	)...)
 
 	return []framework.Option{
-		framework.WithProcesses(fp),
+		framework.WithProcesses(fp, n.scheduler1, n.scheduler2, n.scheduler3),
 	}
 }
 
 func (n *leadershipchurn) Run(t *testing.T, ctx context.Context) {
-	n.scheduler1.Run(t, ctx)
-	n.scheduler2.Run(t, ctx)
-	n.scheduler3.Run(t, ctx)
-	t.Cleanup(func() {
-		n.scheduler1.Cleanup(t)
-		n.scheduler2.Cleanup(t)
-		n.scheduler3.Cleanup(t)
-	})
 	n.scheduler1.WaitUntilRunning(t, ctx)
 	n.scheduler2.WaitUntilRunning(t, ctx)
 	n.scheduler3.WaitUntilRunning(t, ctx)
 
 	assert.EventuallyWithT(t, func(c *assert.CollectT) {
 		stream, err := n.scheduler1.Client(t, ctx).WatchHosts(ctx, new(schedulerv1pb.WatchHostsRequest))
-		if !assert.NoError(c, err) {
-			return
 		require.NoError(c, err)
 		defer func() {
 			require.NoError(t, stream.CloseSend())
 		}()
 		resp, err := stream.Recv()
-		if !assert.NoError(c, err) {
-			return
-		}
+		require.NoError(c, err)
 		assert.Len(c, resp.GetHosts(), 3)
 	}, 20*time.Second, 10*time.Millisecond)
 


tests/integration/suite/scheduler/kubernetes/leadershipchurn.go
```diff
@@ -0,0 +1,185 @@
+/*
+Copyright 2024 The Dapr Authors
```
acroca:
```suggestion
Copyright 2026 The Dapr Authors
```

tests/integration/suite/scheduler/kubernetes/leadershipchurn.go
```diff
@@ -0,0 +1,185 @@
...
+		scheduler.WithID("scheduler-1"),
+		scheduler.WithEtcdClientPort(port5),
+	)...)
+	n.scheduler3 = scheduler.New(t, append(clusterOpts,
+		scheduler.WithID("scheduler-2"),
+		scheduler.WithEtcdClientPort(port6),
+	)...)
+
+	return []framework.Option{
+		framework.WithProcesses(fp),
```
acroca: Can't we add the schedulers here?

tests/integration/suite/scheduler/kubernetes/leadershipchurn.go
```diff
@@ -0,0 +1,185 @@
...
+	})
+	n.scheduler1.WaitUntilRunning(t, ctx)
+	n.scheduler2.WaitUntilRunning(t, ctx)
+	n.scheduler3.WaitUntilRunning(t, ctx)
+
+	assert.EventuallyWithT(t, func(c *assert.CollectT) {
+		stream, err := n.scheduler1.Client(t, ctx).WatchHosts(ctx, new(schedulerv1pb.WatchHostsRequest))
+		if !assert.NoError(c, err) {
+			return
+		}
```
acroca:
```suggestion
		require.NoError(c, err)
```

tests/integration/suite/scheduler/kubernetes/leadershipchurn.go
```diff
@@ -0,0 +1,185 @@
...
+		defer func() {
+			require.NoError(t, stream.CloseSend())
+		}()
+		resp, err := stream.Recv()
+		if !assert.NoError(c, err) {
+			return
+		}
```
acroca: Can we use `require`? Or will it not run the `defer`?

tests/integration/suite/scheduler/kubernetes/leadershipchurn.go
```diff
@@ -0,0 +1,173 @@
...
+func (n *leadershipchurn) Run(t *testing.T, ctx context.Context) {
+	n.scheduler1.WaitUntilRunning(t, ctx)
+	n.scheduler2.WaitUntilRunning(t, ctx)
+	n.scheduler3.WaitUntilRunning(t, ctx)
+
+	assert.EventuallyWithT(t, func(c *assert.CollectT) {
+		stream, err := n.scheduler1.Client(t, ctx).WatchHosts(ctx, new(schedulerv1pb.WatchHostsRequest))
+		require.NoError(c, err)
+		defer func() {
+			require.NoError(t, stream.CloseSend())
```
acroca, line 88: I'm not sure this should use `t` or `c`. If the `CloseSend()` fails, should we try again the 'eventually' or fail the test?

PR #9688, opened 2026-03-25
Scheduler fix fatal crash on server restart due to duplicate controller registration

cicoyle:
Fix Perf Tests by fixing the scheduler bug

Fix fatal scheduler crash when the server restart loop iterates more than once -triggered by etcd leadership rebalance during cluster formation.

relevant bug log line: `controller with name namespaces already exists`


pkg/scheduler/server/internal/controller/controller.go
```diff
@@ -37,11 +38,23 @@ var log = logger.NewLogger("dapr.scheduler.server.controller")
...
-	Cron       cron.Interface
 	Healthz    healthz.Healthz
 }
 
-func New(opts Options) (concurrency.Runner, error) {
+// Controller wraps a long-lived controller-runtime manager. It must only be
+// created once per process because controller-runtime registers controller
+// names globally in the metrics registry.
+type Controller struct {
+	cron atomic.Pointer[cron.Interface]
```
acroca: If I'm not mistaken, `atomic.Value` would be better in this case

pkg/scheduler/server/internal/controller/namespace_test.go
```diff
@@ -28,9 +28,50 @@ import (
...
+
+	// Before SetCron is called, reconcile should return an error.
+	_, err := nsctrl.Reconcile(t.Context(), req)
+	require.ErrorContains(t, err, "controller cron not yet initialized")
+
+	// Simulate three server restarts, each with a different cron instance.
+	// Each cron records which "generation" was called so we can verify the
+	// reconciler always uses the latest one.
+	for i := range 3 {
+		gen := i
```
acroca: I think this is not necessary, loop variables are already per-iteration now if I'm not mistaken.

JoshVanL (changes requested):
Possible to add an integration test for this?

JoshVanL (approved):

PR #9697, opened 2026-03-25
[1.16] Scheduler: fix silent cron exit on quorum change under load

JoshVanL:
Marked to backport to 1.17.3 and 1.16.11.

---

When a Scheduler instance joins or rejoins the cluster, the leadership quorum changes (e.g., 2->3). The cron module's runEngine sends updated host addresses to an unbuffered WatchLeadership channel before starting the engine. If the consumer is busy broadcasting to many WatchHosts subscribers, the send blocks. A second quorum change cancels the elected context while the send is blocked, causing runEngine to return context.Canceled. The cron loop treats this as terminal and exits silently. No error is logged, no re-election is attempted, and the leadership key remains stuck at the old partition total. The other instances cannot converge, leaving the affected instance's jobs permanently undeliverable.

Replace the direct channel consumer with a non-blocking events/loop. Enqueue never blocks, so the WatchLeadership channel send always completes before the context cancellation can race with it.

---

Scheduler instance silently stops participating after cluster scale-up

Problem

After a Scheduler pod restart, rolling update, or cluster scale-up event, one or more Scheduler instances can silently stop participating in the cluster. The affected instance's pod remains running and passes health checks, but its cron engine exits and never restarts. The instance stops publishing its address to the WatchHosts API, so daprd sidecars never discover it. Jobs, Actor Reminders or Workflows assigned to the affected instance's partitions are never triggered.

From a user's perspective, workflows, scheduled jobs, and actor reminders randomly stop firing after a Scheduler pod restart. The issue is intermittent and depends on the exact timing of the restart relative to other cluster activity.

Impact

Any Dapr deployment running the Scheduler in a multi-instance (HA) configuration is affected. The issue is triggered when a Scheduler instance joins or rejoins the cluster, causing a leadership quorum change (e.g., partition count changes from 2 to 3). The likelihood increases with the number of connected daprd sidecars, as more WatchHosts subscribers means the host broadcast takes longer, widening the race window.

When the bug is triggered:

- The affected Scheduler instance owns partitions but cannot deliver jobs on them. All workflow activity timers, scheduled jobs, and actor reminders assigned to those partitions stop firing and are logged as `UNDELIVERABLE`.
- Daprd sidecars that call `WatchHosts` may receive an incomplete host list (missing the affected instance), or may never receive a response at all — leaving the sidecar stuck on the `scheduler-watch-hosts` readiness gate indefinitely, preventing the application from becoming ready.
- The remaining healthy instances cannot take over the affected instance's partitions because its leadership key remains in etcd (the lease keep-alive continues independently). The cluster is stuck in a state where all instances are running but quorum can never converge on the correct partition count.

The only recovery is to restart all Scheduler pods simultaneously, which forces a fresh leadership election.

Root Cause

When the Scheduler's internal cron module reaches leadership quorum after a partition change, it calls `runEngine` to start the cron engine. Before starting the engine, `runEngine` sends the updated cluster host addresses to an internal unbuffered Go channel (`WatchLeadership`) so that the WatchHosts API can broadcast them to connected sidecars.

If the channel consumer is busy (for example, broadcasting a previous host update to many WatchHosts subscribers), the send blocks. Meanwhile, if another Scheduler instance joins and causes a second quorum change, the elected context is cancelled while the send is still blocked.

Because the cron module has exited, it never calls `Reelect` to update its leadership key with the new partition total. The other Scheduler instances see this stale key and cannot reach quorum agreement, preventing the entire cluster from converging.

Solution

The internal channel consumer in the Scheduler's cron wrapper has been replaced with a non-blocking event loop (`events/loop`). The loop's `Enqueue` method never blocks.
If the current segment is full, it allocates a new one. This means the channel send from the cron library always completes immediately, regardless of how busy the consumer is with broadcasting.

Since the send no longer blocks, the context cancellation race that caused the silent exit can no longer occur. The cron loop continues to call `Reelect` after each quorum change, leadership keys are updated with the correct partition total, and all instances converge normally.

### DIFF
--- a/pkg/scheduler/server/internal/cron/leadership.go
+++ b/pkg/scheduler/server/internal/cron/leadership.go
@@ -58,9 +58,9 @@ func (h *leadership) Handle(ctx context.Context, anyhosts []*anypb.Any) error {
 		close(h.readyCh)
 		log.Info("Cron is ready")
 	}
-	h.lock.Unlock()
 
 	h.hostBroadcaster.Broadcast(hosts)
+	h.lock.Unlock()
 
 	return nil
 }


ryanisnan:
Thank you @JoshVanL !

PR #9704, opened 2026-03-26
Support glob service accounts

### LINKED ISSUE #9596: Allow sidecar injector to accept full wildcards for service account names
## In what area(s)?

/area injector

## Describe the feature
Allow sidecar injector to accept full wildcards for service account names.  Either `*` or `*:*`.

Currently the field only supports `prefix-*:prefix*`, `*:prefix-*` or `prefix-*:*`.

Note: Currently the setting is named `dapr_sidecar_injector.allowedServiceAccountsPrefixNames`. For clarity, it might be worthwhile to rename this field (perhaps with a backwards compat mapping of the old name?).

Suggested name off the top of my head would be `dapr_sidecar_injector.allowedServiceAccountsNames`

## Release Note
RELEASE NOTE: **ADD** Allow wildcards for service account names

acroca:
Close: https://github.com/dapr/dapr/issues/9596

Docs: https://github.com/dapr/docs/pull/5105

## Summary

Adds glob pattern support to the existing `ALLOWED_SERVICE_ACCOUNTS` env var for the sidecar injector. Service account entries now support full glob syntax (`*`, `?`, `[...]` character classes) via Go's `path.Match`, in addition to the existing exact `namespace:name` format.

This replaces the need for the separate `ALLOWED_SERVICE_ACCOUNTS_PREFIX_NAMES` setting, trailing-`*` prefixes like `ns*:sa*` are valid glob patterns with identical behavior. `ALLOWED_SERVICE_ACCOUNTS_PREFIX_NAMES` is now deprecated (still functional, logs a warning) and users should migrate entries to `ALLOWED_SERVICE_ACCOUNTS`.

### Examples

```
# Exact match (unchanged)
ALLOWED_SERVICE_ACCOUNTS=my-ns:my-sa

# Wildcard — all service accounts in a namespace
ALLOWED_SERVICE_ACCOUNTS=my-ns:*

# Prefix match (replaces ALLOWED_SERVICE_ACCOUNTS_PREFIX_NAMES)
ALLOWED_SERVICE_ACCOUNTS=team-*:deploy-*

# Single-char wildcard
ALLOWED_SERVICE_ACCOUNTS=staging-?:*

# Character class
ALLOWED_SERVICE_ACCOUNTS=proj-*:sa-[abc]*
```

### What changed

- **`ALLOWED_SERVICE_ACCOUNTS`** now accepts glob patterns — backwards compatible since plain `namespace:name` values are valid globs.
- **`ALLOWED_SERVICE_ACCOUNTS_PREFIX_NAMES`** is deprecated. Still works, but logs a warning directing users to use `ALLOWED_SERVICE_ACCOUNTS`.
- The default allowed service accounts (`kube-system` controllers, etc.) are now also matched by name via the same glob matcher, in addition to the existing UID-based path (kept as defense-in-depth).
- Removed the `namespacednamematcher` sub-package. The matcher is now a single function (`NewServiceAccountMatcher`) in the `service` package.

### Breaking changes

None. All existing configurations continue to work without modification.


pkg/injector/namespacednamematcher/equalmatcher.go
```diff
@@ -0,0 +1,78 @@
+/*
+Copyright 2025 The Dapr Authors
```
javier-aliaga: 2026

javier-aliaga: all the other new file had 2025 too

javier-aliaga:
Does this change need docs?

acroca:
> Does this change need docs?

Yes, I was planning to add them after getting an approval, just in case there were concerns about this approach or suggestions

pkg/injector/namespacednamematcher/globmatcher.go
javier-aliaga, line 1: this and the equalmatcher are quite similar, is it worthit to refactor and extract common logic?

pkg/injector/service/serviceaccountmatcher_test.go
```diff
@@ -0,0 +1,146 @@
...
+package service
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestNewServiceAccountMatcherEmpty(t *testing.T) {
```
javier-aliaga, line 23: This test is failing

JoshVanL (changes requested):
Docs & integration tests please

acroca:
@JoshVanL Added integration tests, and docs PR here: https://github.com/dapr/docs/pull/5105

JoshVanL (approved):

PR #9706, opened 2026-03-26
Jobs api stable

cicoyle:
jobs api stable


pkg/api/http/jobs.go
```diff
@@ -64,6 +71,37 @@ func (a *api) constructJobsEndpoints() []endpoints.Endpoint {
 				Name: "GetJob",
 			},
 		},
+		// Stable v1.0 endpoints
```
acroca, line 74: Why are there no routes for `List` and `DeleteJobsByPrefix`? They were not even defined in the alpha, I'm wondering what's the reason.

acroca (dismissed):
LGTM, just a question

pkg/api/http/jobs.go
cicoyle: I dont think there was a technical reason they werent added. When promoting the Jobs API to stable I was just bumping the existing routes and ensuring we meet the criteria for being stable. I think we can add these later on if there is a request for it, but as of now, I dont think there is.

JoshVanL:
@cicoyle need to fix lint

acroca (approved):

PR #9733, opened 2026-03-31
fix: clean up dapr services for non-enabled workloads

### LINKED ISSUE #7933: Dapr sidecar-injector doesn't clean up Kubernetes Service resources
## In what area(s)?

/area runtime

## What version of Dapr?

> 1.13.x

## Expected Behavior

When a Dapr-annotated Kubernetes resource like a `Deployment` get the Dapr sidecar injected, a Kubernetes Service is created for Service to Service invocation to work. 
If the Dapr annotations are removed from the Deployment resource, the pod restart and the Dapr sidecar is not injected, but the Kubernetes Service that was created is never deleted. 

The expected behavior, from a user perspective, is that if the Deployment is not Dapr-enabled, a Service shouldn't be created (or the service should be cleaned up), if there is no Dapr sidecar for that Deployment. 

## Actual Behavior

The Kubernetes Service for Dapr-enabled applications is never cleaned up. If you try to delete the Service manually, the sidecar-injector will recreate the resource. Unless (thanks @alicejgibbons ) you restart the sidecar-injector pod that lives in the Dapr control plane. 

1) Create a Deployment and add Dapr annotations, so the sidecar gets injected
2) Check that a Kubernetes Service was created by Dapr
3) Remove the Dapr annotations from the Deployment and check if the Service is still there.. (it will be there)
4) Try removing the Service manually with `kubectl delete svc ...`
5) Try deleting the sidecar-injector pod inside the `dapr-system` namespace. Now you can delete the service. 

## Release Note

RELEASE NOTE:

abishekgiri:
## Description
This change deletes Dapr-managed Kubernetes Service resources when the owning workload is no longer Dapr-enabled. Without this cleanup, Services created for Dapr-enabled workloads can remain behind after the Dapr annotations are removed, and they may continue to be recreated unexpectedly.

## Issue reference
Fixes #7933

## Checklist
- [x] Code compiles correctly
- [x] Created/updated tests
- [x] Unit tests passing
- [ ] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#7933
- [ ] Specification has been updated / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#7933
- [ ] Provided sample for the feature / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#7933

### DIFF
--- a/pkg/operator/handlers/dapr_handler.go
+++ b/pkg/operator/handlers/dapr_handler.go
@@ -189,6 +189,12 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 			log.Errorf("failed to ensure dapr service present, err: %v", err)
 			return ctrl.Result{Requeue: true}, err
 		}
+	} else {
+		err := r.cleanupOwnedDaprServices(ctx, req.Namespace, req.Name, wrapper.GetObject())
+		if err != nil {
+			log.Errorf("failed to cleanup dapr services, err: %v", err)
+			return ctrl.Result{Requeue: true}, err
+		}
 	}
 
 	return ctrl.Result{}, nil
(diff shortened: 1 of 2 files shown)


abishekgiri:
This PR deletes Dapr-managed Service resources when the owning workload is no longer Dapr-enabled or has been removed. I also added tests to cover both cleanup paths.

abishekgiri:
This PR ensures Dapr services are cleaned up when workloads do not have Dapr enabled, preventing unnecessary resource usage and improving consistency.

PR #9757, opened 2026-04-06
fix flaky crossactivity workflow integration test

JoshVanL:
OrchestratorStarted events can appear at varying positions in the workflow history depending on timing. The test assumed these events only appeared at fixed indices, but on slow CI runners, an extra OrchestratorStarted could be interleaved before TaskScheduled (at index 2), shifting all subsequent events and causing assertion failures. Scan past OrchestratorStarted events to find each meaningful event by type rather than by hardcoded index.vided sample for the feature / Created issue in the <https://github.com/dapr/docs/> repo: dapr/docs#_[issue number]_

### DIFF
--- a/tests/integration/suite/daprd/workflow/get/crossactivity.go
+++ b/tests/integration/suite/daprd/workflow/get/crossactivity.go
@@ -74,34 +74,52 @@ func (c *crossactivity) Run(t *testing.T, ctx context.Context) {
 
 	evs := resp.Events
 
-	// Can have 1 or 2 `GetOrchestratorStarted` events depending on timing.
-	require.True(t, len(evs) == 7 || len(evs) == 6)
+	// OrchestratorStarted events can appear at varying positions depending on
+	// timing, so find the non-OrchestratorStarted events by scanning rather
+	// than assuming fixed indices.
+	require.True(t, len(evs) >= 6 && len(evs) <= 8,
+		"expected 6-8 events, got %d", len(evs))
 
 	assert.NotNil(t, evs[0].GetOrchestratorStarted())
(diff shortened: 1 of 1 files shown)


tests/integration/suite/daprd/workflow/get/crossactivity.go
```diff
@@ -74,34 +74,52 @@ func (c *crossactivity) Run(t *testing.T, ctx context.Context) {
...
-	// The index of the next events depends on whether there are 6 or 7 events
-	// total.
-	i := 4
-	if len(evs) == 7 {
-		i = 5
-		assert.NotNil(t, evs[4].GetOrchestratorStarted())
+
+	// Find ExecutionStarted (first non-OrchestratorStarted event).
+	i := 1
+	for i < len(evs) && evs[i].GetOrchestratorStarted() != nil {
```
acroca: Why doesn't this find the first ExecutionStarted directly instead of the first non-OrchStarted?

tests/integration/suite/daprd/workflow/get/crossactivity.go
```diff
@@ -74,34 +74,52 @@ func (c *crossactivity) Run(t *testing.T, ctx context.Context) {
...
+		i++
+	}
+	require.Less(t, i, len(evs), "ExecutionStarted event not found")
+	assert.NotNil(t, evs[i].GetExecutionStarted())
```
acroca: `es := evs[i].GetExecutionStarted()`

JoshVanL:
@acroca thanks, PTAL

acroca (approved):

PR #9760, opened 2026-04-07
Fix workflow event loss during ContinueAsNew under high event volume

### LINKED ISSUE #8277: [Workflows] Events are disappearing during `ContinueAsNew`
## In what area(s)?

/area runtime

## What version of Dapr?

> 1.14.x

## Expected Behavior

When using the `ContinueAsNew` pattern, When raising many _Events_ (...with the same Name) via `RaiseExternalEvent`, the Events are all buffered, and handled _eventually_ by the Workflow. Never dropped / No data loss.

## Actual Behavior

Some events are never processed by the Workflow, giving the sense that they have been dropped or were never persisted. From my testing it appears that the harder I drive this (more events raised in parrallel) the more likely that one or more of the events will get dropped/disappear.

Download the repo in the link, then do `docker compose build` / `docker compose up` and follow instructions laid out in the issue.

1. https://github.com/olitomlinson/dapr-semaphore-workflow/issues/2

## Release Note

RELEASE NOTE:

JoshVanL:
When the ContinueAsNew tight-loop exceeded `MaxContinueAsNewCount` (20 iterations), the engine's applier had already mutated `wi.State` in place via `*s = *newState`. Because `wi.State` and `o.rstate` were the same pointer, the actor's cached runtime state was left corrupted; the workflow input counter jumped ahead to reflect iterations that were never persisted.

On retry, the workflow resumed from the corrupted counter, silently skipping events that had been counted but never processed. Under high concurrency this caused permanent event loss.

The fix clones `o.rstate` before passing it to the engine, so failed executions cannot corrupt the actor's cached state.

Fixes: https://github.com/dapr/dapr/issues/8277

---

Workflow events lost during ContinueAsNew under high event volume

Problem

When many external events are raised concurrently against a workflow that uses `ContinueAsNew` with `preserveUnprocessedEvents`, some events can be silently lost. The workflow completes with fewer events than were actually sent, or its counter jumps ahead, skipping events that were delivered to the runtime but never processed by the workflow.

Impact

Any workflow that uses the `ContinueAsNew` pattern with `preserveUnprocessedEvents` (or `WithKeepUnprocessedEvents` in the Go SDK) is affected when receiving a burst of external events. The harder the workflow is driven (more events raised in parallel), the more likely events are to disappear. This is a data loss scenario; the events are accepted by the Dapr API but never reach the workflow function.

Root Cause

When the workflow engine processes a batch of events through `ContinueAsNew`, each iteration of the internal tight loop mutates the workflow's runtime state in place. If the loop exceeds the engine's iteration limit (20) and fails, the in-memory cached state is left in a corrupted state; the workflow's input counter has jumped ahead to reflect iterations that were never persisted.

On retry, the workflow resumes from the corrupted counter value instead of the last persisted value, causing it to skip events that were already counted but never actually processed.

Solution

The workflow runtime state is now cloned before being passed to the engine for execution. The engine operates on its own copy, so if execution fails for any reason, the actor's cached state remains consistent with what was last persisted to the state store. Retries start from the correct state and all events are processed.

### DIFF
--- a/pkg/actors/targets/workflow/orchestrator/run.go
+++ b/pkg/actors/targets/workflow/orchestrator/run.go
@@ -115,7 +115,10 @@ func (o *orchestrator) runWorkflow(ctx context.Context, reminder *actorapi.Remin
 	// with o.rstate (same pointer) and may overwrite it during ContinueAsNew
 	// (*s = *newState in the applier). If the engine fails, we restore the
 	// snapshot so the cached state remains consistent with the store.
-	rstateSnapshot := proto.Clone(o.rstate).(*backend.OrchestrationRuntimeState)
+	var rstateSnapshot *backend.OrchestrationRuntimeState
+	if o.rstate != nil {
+		rstateSnapshot = proto.Clone(o.rstate).(*backend.OrchestrationRuntimeState)
+	}
 
 	// TODO: @joshvanl remove.
 	err = o.scheduler(ctx, wi)
@@ -138,7 +141,10 @@ func (o *orchestrator) runWorkflow(ctx context.Context, reminder *actorapi.Remin
 
 	select {
 	case <-ctx.Done(): // caller is responsible for timeout management
-		// Workflow execution failed with recoverable error
+		// The engine may have partially mutated o.rstate via the shared
+		// wi.State pointer before the context was cancelled. Restore the
+		// snapshot so the cached state stays consistent with the store.
+		o.rstate = rstateSnapshot
 		executionStatus = diag.StatusRecoverable
 		return todo.RunCompletedFalse, ctx.Err()
 	case completed := <-callback:
--- a/pkg/actors/targets/workflow/orchestrator/run.go
+++ b/pkg/actors/targets/workflow/orchestrator/run.go
@@ -164,6 +164,7 @@ func (o *orchestrator) runWorkflow(ctx context.Context, reminder *actorapi.Remin
 				state.ApplyRuntimeStateChanges(wi.State)
 				state.Generation++
 				if err = o.saveInternalState(ctx, state); err != nil {
+					o.rstate = rstateSnapshot
 					return todo.RunCompletedFalse, err
 				}
 			} else {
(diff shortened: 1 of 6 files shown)


tests/integration/suite/daprd/workflow/continueasnew/raisebatch.go
```diff
@@ -0,0 +1,203 @@
...
+
+		// In drain mode, process one event and complete. The CAN progress
+		// was saved when MaxContinueAsNewCount was hit, so inc reflects
+		// the saved progress (e.g. 20 after 20 successful CAN iterations).
+		if drainMode.Load() {
+			ctx.WaitForSingleEvent("incr", time.Minute).Await(nil)
+			return inc + 1, nil
+		}
+
+		// Use negative timeout (wait indefinitely) so no durable timer is created.
```
acroca: Timers are always created now

JoshVanL:
@acroca fixed, PTAL!

pkg/actors/targets/workflow/orchestrator/run.go
```diff
@@ -131,12 +141,34 @@ func (o *orchestrator) runWorkflow(ctx context.Context, reminder *actorapi.Remin
...
+			// the original events. On retry, the engine replays from the
+			// saved CAN progress and re-processes the inbox, skipping
+			// already-consumed events via deduplication.
+			// If no CAN progress was made (non-CAN failure), restore the
+			// pre-engine snapshot so the cached state stays consistent.
+			if wi.State.GetContinuedAsNew() {
+				state.ApplyRuntimeStateChanges(wi.State)
+				state.Generation++
+				if err = o.saveInternalState(ctx, state); err != nil {
+					return todo.RunCompletedFalse, err
```
acroca, line 168: Should we do `o.rstate = rstateSnapshot` in case of an error??

JoshVanL: Agreed :+1:

acroca (approved):

PR #9766, opened 2026-04-08
CI: retry transient docker push and crane.Copy failures

JoshVanL:
The `build-and-push` pipeline had no retry logic for transient network errors during `crane.Copy` and docker push operations. Add 3-attempt retries with backoff in the build-tools Go code and in the workflow steps for make docker-push across version-skew, kind-e2e, and dapr-test workflows.


.github/workflows/dapr-test.yml
```diff
@@ -356,10 +356,20 @@ jobs:
         run: |
           if [ "${{ env.TARGET_OS }}" = "linux" ] && [ "${{ env.TARGET_ARCH }}" != "arm" ]; then
             # For Linux, we use images based on Mariner
-            DOCKERFILE=Dockerfile-mariner make docker-push
-          else
-            make docker-push
+            export DOCKERFILE=Dockerfile-mariner
           fi
+          # Retry docker-push up to 3 times for transient registry errors
+          for i in 1 2 3; do
```
acroca: Can't we make the `make docker-push` target do the retry instead of the github workflow? or create a `docker-push-retry` target if you want to keep the docker-push as a single-attempt task...

JoshVanL: :+1:

acroca (approved):

PR #9769, opened 2026-04-08
Workflow: Fix carry over events for ContiueAsNew

JoshVanL:
When the ContinueAsNew tight-loop exceeds MaxContinueAsNewCount (20) and partial progress is saved, the inbox was preserved unchanged. On retry, all original events were re-delivered as NewEvents alongside the carryover events already in OldEvents (History), causing the engine to buffer both sets and the workflow to process duplicate events.

For coordination workflows (e.g. the semaphore pattern), this caused the same request to be dispatched multiple times, leading to runaway workflow state.

The fix separates carryover EventRaised events from the CAN execution events (WorkflowStarted, ExecutionStarted) and moves them to the Inbox, replacing the stale original inbox. On retry, only unprocessed carryover events are delivered as NewEvents, eliminating duplicates.

---

Workflow events lost during ContinueAsNew under high event volume

Problem

When many external events are raised concurrently against a workflow that uses `ContinueAsNew` with `preserveUnprocessedEvents`, events can be silently lost or processed multiple times. The workflow completes with fewer events than were actually sent, its counter jumps ahead skipping events, or the same event is delivered to the workflow function more than once.

Impact

Any workflow that uses the `ContinueAsNew` pattern with `preserveUnprocessedEvents` (or `WithKeepUnprocessedEvents` in the Go SDK) is affected when receiving a burst of external events. The harder the workflow is driven (more events raised in parallel), the more likely events are to be lost or duplicated.

For simple counter workflows, this manifests as event loss (the counter skips ahead). For coordination workflows such as a semaphore pattern, duplicate event delivery causes the same request to be dispatched multiple times, leading to runaway workflow state where the workflow's bookkeeping becomes inconsistent and the workflow cannot recover.

Root Cause

Two related issues cause event loss and duplicate processing when the workflow engine's `ContinueAsNew` tight-loop exceeds the iteration limit (20):

1. **State corruption via shared pointer**: The engine mutates the workflow's runtime state in place during `ContinueAsNew` iterations. If the loop exceeds the iteration limit and fails, the in-memory cached state is left in a corrupted state; the workflow's input counter has jumped ahead to reflect iterations that were never persisted. On retry, the workflow resumes from the corrupted counter value instead of the last persisted value, causing it to skip events.

2. **Duplicate event delivery from stale inbox**: When partial `ContinueAsNew` progress is saved after hitting the iteration limit, the workflow's inbox (containing all original events) was preserved unchanged. On retry, all original events were re-delivered as new events alongside the carryover events already saved in history. This caused the workflow to buffer both sets, processing the same events multiple times. For workflows implementing coordination patterns (e.g. a semaphore), this resulted in duplicate dispatches and runaway workflow state.

Solution

1. **State cloning**: The workflow runtime state is now cloned before being passed to the engine for execution. The engine operates on its own copy, so if execution fails for any reason, the actor's cached state remains consistent with what was last persisted to the state store.

2. **Inbox replacement with carryover events**: When partial `ContinueAsNew` progress is saved, unprocessed carryover events (buffered `EventRaised` events from the engine's `ContinueAsNew` state) are moved from history to the inbox, and the stale original inbox is discarded. On retry, only the unprocessed carryover events are delivered as new events, preventing duplicate processing.


docs/release_notes/v1.17.4.md
```diff
@@ -5,7 +5,6 @@ This update contains bug fixes and security fixes:
 - [Cross-app workflow stuck in PENDING when the first action is a remote activity or child workflow call to an offline app](#cross-app-workflow-stuck-in-pending-when-the-first-action-is-a-remote-activity-or-child-workflow-call-to-an-offline-app)
 - [Dapr forwards hop-by-hop HTTP headers when proxying service invocation requests](#dapr-forwards-hop-by-hop-http-headers-when-proxying-service-invocation-requests)
 - [Workflow events lost during ContinueAsNew under high event volume](#workflow-events-lost-during-continueasnew-under-high-event-volume)
-- [Security: Update Go to 1.26.2 for security fixes](#security-update-go-to-1262-for-security-fixes)
```
acroca, line 8: Is this intentional?

pkg/ports/ports_test.go
```diff
@@ -48,13 +49,14 @@ func TestGetStablePort(t *testing.T) {
...
-		port2 := getPort(t, "myapp2")
-		assert.True(t, (port2 >= 10233 && port2 <= 12280) || (port2 >= 22444 && port2 <= 24491))
-		assert.NotEqual(t, port1, port2)
+	t.Run("Invoking with different appIDs returns different ports", func(t *testing.T) {
+		ports := make(map[int]bool)
+		for i := range 10 {
+			port := getPort(t, fmt.Sprintf("different-app-%d", i))
+			ports[port] = true
+		}
+		assert.Greater(t, len(ports), 1,
```
acroca: Shouldn't the `getPort` return an available port? If so, shouldn't we always get 10 ports here?

tests/integration/suite/daprd/workflow/continueasnew/carryover.go
```diff
@@ -0,0 +1,215 @@
...
+		require.NoError(t, err)
+		_, err = client.WaitForWorkflowStart(ctx, id)
+		require.NoError(t, err)
+
+		for range 3 {
+			require.NoError(t, client.RaiseEvent(ctx, id, "next", api.WithEventPayload("go")))
+			time.Sleep(100 * time.Millisecond)
+		}
+
+		meta, err := client.WaitForWorkflowCompletion(ctx, id)
```
acroca, line 167: I don't understand this test. The workflow only returns `inc` when `val == ""`, but all events have payload `"go"`. Is it due to the timeout?

tests/integration/suite/daprd/workflow/continueasnew/semaphore.go
```diff
@@ -0,0 +1,132 @@
...
+			return st.Dispatched, nil
+		}
+
+		var reqID string
+		ctx.WaitForSingleEvent("request", 5*time.Second).Await(&reqID)
+		if reqID != "" {
+			st.Pending = append(st.Pending, reqID)
+		}
+
+		ctx.ContinueAsNew(st, task.WithKeepUnprocessedEvents())
```
acroca, line 100: If I read this correctly, every iteration of the `ContinueAsNew` will only carry one single pending request ID right? By the time we call the continue as new we've already processed all the pending in `st`...

docs/release_notes/v1.17.4.md
JoshVanL: Yes- 1.17 isn't using 1.26

JoshVanL:
@acroca PTAL!

cicoyle (approved):

PR #9770, opened 2026-04-08
Fix daprd sidecar placement dissemination timeout handling

JoshVanL:
When a slow sidecar delays placement table dissemination, the daprd sidecar's 5-second timeout fired and sent a fatal error through the internal error channel, killing the entire placement subsystem. The sidecar never reconnected. All actor and workflow operations hung because the inflight lock was abandoned in a nil state.

The dissemination timeout now closes the placement stream, triggering the existing handleCloseStream -> handleReconnect path. This matches the recovery behavior for network disconnections (EOF, connection reset), which already reconnected successfully. The timeout was the only error path that did not reconnect.

The same fix is applied to the version mismatch and unknown operation error paths in handleOrder, which had the same fatal-error-instead-of- reconnect behavior.

---

Workflow and actor operations freeze when a slow sidecar delays placement dissemination

Problem

When one daprd sidecar in a namespace is slow to respond during placement table dissemination (for example, due to GC pressure, high actor load, or network latency), all other sidecars in the same namespace experience frozen workflow and actor operations. Scheduling new workflows, invoking actors, and running `dapr workflow terminate` or `dapr workflow purge` all hang indefinitely. After restarting the app, workflows may appear as RUNNING but no activities execute, and `get_workflow_state` returns nothing.

Impact

Any deployment with multiple replicas using actors or workflows is affected. A single slow sidecar can freeze all actor and workflow operations across every other sidecar in the namespace for as long as the placement server takes to time out the slow peer (8 seconds by default), plus the sidecar's own 5-second timeout. In practice, this created a 10-15 second window where:

- All actor method invocations hang
- All workflow scheduling, termination, and purge operations hang
- Actor reminders stop firing
- Applications stop responding to health probes and are declared unhealthy by Kubernetes
- With a workflow SDK client connected, the daprd process enters a zombie state where it can neither serve requests nor exit

The issue is particularly severe during rolling updates, where pod cycling causes repeated dissemination rounds that can trigger the timeout multiple times in succession.

Root Cause

When the daprd sidecar receives a LOCK order from the placement server during actor table dissemination, it blocks all in-flight actor operations (the "inflight lock") until it receives the corresponding UNLOCK. The placement server only sends UNLOCK after every sidecar in the namespace has responded to each phase (LOCK, UPDATE, UNLOCK). If one sidecar is slow, no other sidecar receives UPDATE or UNLOCK.

The daprd sidecar has a 5-second timeout for this wait. When the timeout fired, it killed the entire placement subsystem by sending a fatal error through the internal error channel, which terminated the placement connection permanently. The sidecar never reconnected. Actor operations that were queued during the LOCK phase were silently abandoned, and any new actor operations hung indefinitely because the placement client was dead.

Solution

The dissemination timeout now closes the placement stream and triggers an automatic reconnection, instead of killing the placement subsystem. After the timeout:

1. The sidecar closes its stream to the placement server
2. All actors are halted and the routing table is cleared
3. The sidecar reconnects to placement and re-registers
4. A new dissemination round completes and actor operations resume

This matches the existing recovery behavior for network disconnections (EOF, connection reset), which already reconnected successfully. The timeout was the only error path that did not reconnect.

Additionally, this release fixes two related issues in the dissemination subsystem:

- The placement readiness flag was set immediately after reconnecting, before the new dissemination round completed. This caused the daprd to report healthy to metadata queries while actor operations were still blocked. The readiness flag is now only set after the first successful UNLOCK.
- Recycled disseminator objects from the internal pool could carry a stale timeout version counter from their previous use, potentially causing a valid timeout to be incorrectly ignored. The counter is now reset on reuse.

### DIFF
--- a/docs/release_notes/v1.17.4.md
+++ b/docs/release_notes/v1.17.4.md
@@ -5,6 +5,7 @@ This update contains bug fixes:
 - [Cross-app workflow stuck in PENDING when the first action is a remote activity or child workflow call to an offline app](#cross-app-workflow-stuck-in-pending-when-the-first-action-is-a-remote-activity-or-child-workflow-call-to-an-offline-app)
 - [Dapr forwards hop-by-hop HTTP headers when proxying service invocation requests](#dapr-forwards-hop-by-hop-http-headers-when-proxying-service-invocation-requests)
 - [Workflow events lost during ContinueAsNew under high event volume](#workflow-events-lost-during-continueasnew-under-high-event-volume)
+- [Workflow and actor operations freeze when a slow sidecar delays placement dissemination](#workflow-and-actor-operations-freeze-when-a-slow-sidecar-delays-placement-dissemination)
 
 ## Pulsar pub/sub ignores processMode from component metadata and lacks async backpressure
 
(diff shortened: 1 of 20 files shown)


cicoyle (approved):
lgtm

PR #9775, opened 2026-04-09
Workflow: Adds global and per-name concurrency limits

JoshVanL:
Adds scheduler-enforced concurrency limits for workflows and activities that apply across all replicas, complementing the existing per-sidecar limits. Users can set a global maximum for all workflows or activities, as well as per-name limits for specific workflow or activity names.

The scheduler divides global limits among its instances using floor division and holds triggers in a pending queue when limits are reached, dispatching them as slots free up via the existing ACK protocol.

Per-name limits are enabled by propagating the workflow/activity name as a concurrency_key on JobMetadata when reminders are created. The scheduler maintains independent gates keyed by group (type-level) and group:name (per-name), and a trigger must pass all applicable gates before dispatch.

Configuration:

```yaml
spec:
  workflow:
    globalMaxConcurrentWorkflowInvocations: 50
    globalMaxConcurrentActivityInvocations: 200
    activityConcurrencyLimits:
      - name: SendEmail
        maxConcurrent: 5
    workflowConcurrencyLimits:
      - name: OrderProcess
        maxConcurrent: 20
```


pkg/scheduler/server/internal/pool/loops/connections/connections.go
```diff
@@ -116,18 +126,201 @@ func (c *connections) handleAdd(ctx context.Context, add *loops.ConnAdd) error {
...
+	}
+
+	for key, gate := range c.concurrencyGates {
+		if _, ok := referenced[key]; !ok {
+			for req := gate.dequeue(); req != nil; req = gate.dequeue() {
+				req.ResultFn(api.TriggerResponseResult_UNDELIVERABLE)
+			}
+			delete(c.concurrencyGates, key)
+		}
+	}
```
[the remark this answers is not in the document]
JoshVanL: Intentional it is the safest to mark as undeliverable and let the job go though the full cron queue cycle. This is only a problem in high contention anyway.

pkg/scheduler/server/internal/pool/loops/connections/concurrency.go
```diff
@@ -0,0 +1,84 @@
...
+}
+
+func localLimitFromGlobal(globalLimit, schedulerCount uint32) uint32 {
+	if schedulerCount <= 1 {
+		return globalLimit
+	}
+	local := globalLimit / schedulerCount
+	if local < 1 {
+		return 1
+	}
```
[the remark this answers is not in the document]
JoshVanL: Intentional.

tests/integration/suite/daprd/workflow/globalmaxconcurrent/activity.go
```diff
@@ -0,0 +1,118 @@
...
+	time.Sleep(time.Second * 2)
+	assert.Equal(t, int64(3), inside.Load())
+
+	// Release another, verify last activity starts.
+	doneCh <- struct{}{}
+	assert.EventuallyWithT(t, func(c *assert.CollectT) {
+		assert.Equal(c, int64(4), inside.Load())
+	}, time.Second*10, time.Millisecond*10)
+
+	close(doneCh)
```
acroca: Maybe we can move this to a defer next to the declaration of `doneCh`, just to be cleaner?
Also, it wouldn't hurt to check the workflow is completed... not necessary for this specific check though.

tests/integration/suite/daprd/workflow/globalmaxconcurrent/combined.go
```diff
@@ -0,0 +1,125 @@
...
+	_, err = client1.ScheduleNewWorkflow(ctx, "combined", api.WithStartTime(time.Now()))
+	require.NoError(t, err)
+
+	// With global limit 3 and "limited" per-name limit 1:
+	// - Only 1 "limited" activity should run (per-name gate)
+	// - Up to 2 "unlimited" activities fill the remaining global slots
+	// - Total concurrent = 3 (global gate)
+	assert.EventuallyWithT(t, func(c *assert.CollectT) {
+		total := limitedInside.Load() + unlimitedInside.Load()
+		assert.Equal(c, int64(3), total)
```
acroca: I think it'd be cleaner to check them individually
```suggestion
		assert.Equal(c, int64(1), limitedInside.Load())
		assert.Equal(c, int64(2), unlimitedInside.Load())
```

tests/integration/suite/daprd/workflow/globalmaxconcurrent/instance.go
```diff
@@ -0,0 +1,105 @@
...
+kind: Configuration
+metadata:
+  name: globalmax
+spec:
+  workflow:
+    globalMaxConcurrentWorkflowInvocations: 2
+`
+	const appID = "globalmax-instance"
+	i.workflow = workflow.New(t,
+		workflow.WithDaprds(2),
```
acroca: Should we have 3 instead? Just to make sure it's not limited to 2 because there're 2 daprds....

tests/integration/suite/daprd/workflow/globalmaxconcurrent/namedactivity.go
```diff
@@ -0,0 +1,133 @@
...
+	// "fast" has no per-name limit, so all 4 fast activities should start.
+	// "slow" is limited to 1 concurrent, so only 1 should be running.
+	assert.EventuallyWithT(t, func(c *assert.CollectT) {
+		assert.Equal(c, int64(4), fastInside.Load())
+	}, time.Second*10, time.Millisecond*10)
+
+	// Only 1 slow activity should be running.
+	assert.EventuallyWithT(t, func(c *assert.CollectT) {
+		assert.Equal(c, int64(1), slowInside.Load())
+	}, time.Second*10, time.Millisecond*10)
```
acroca, line 113: I think we can combine it
```suggestion
	assert.EventuallyWithT(t, func(c *assert.CollectT) {
		assert.Equal(c, int64(4), fastInside.Load())
		assert.Equal(c, int64(1), slowInside.Load())
	}, time.Second*10, time.Millisecond*10)
```

tests/integration/suite/scheduler/concurrency/cluster.go
```diff
@@ -0,0 +1,119 @@
...
+				},
+			},
+		})
+		require.NoError(t, err)
+	}
+
+	// Each scheduler has localLimit=2, so total dispatched should be at most 6.
+	// Wait for triggers to arrive.
+	assert.EventuallyWithT(t, func(c *assert.CollectT) {
+		assert.GreaterOrEqual(c, totalDispatched.Load(), int64(3))
```
acroca: Why can't we wait until it's 6? Just in case a scheduler gets only one? If so, can it happen that a scheduler gets no jobs?

acroca (commented):
Only reviewed tests so far

JoshVanL:
Thanks @acroca please take another look

pkg/scheduler/server/internal/pool/loops/connections/concurrency_test.go
```diff
@@ -0,0 +1,216 @@
...
+			expected:       100,
+		},
+		{
+			name:           "even division across schedulers",
+			globalLimit:    90,
+			schedulerCount: 3,
+			expected:       30,
+		},
+		{
+			name:           "floor division ensures max is not exceeded",
```
acroca: Does this mean we'll never get full utilization of the global limit if the division is not perfect? If so, I think we should take into account the index of the scheduler to calculate its local limit, so when having a global limit of 5, one would get a 1, but two of them would get a 2 in a deterministic way.

pkg/scheduler/server/internal/pool/loops/connections/connections.go
```diff
@@ -116,18 +126,201 @@ func (c *connections) handleAdd(ctx context.Context, add *loops.ConnAdd) error {
...
+	if len(gateKeys) == 0 {
+		streamLoop.Enqueue(req)
+		return
+	}
+
+	var acquired []string
+	for _, key := range gateKeys {
+		gate := c.concurrencyGates[key]
+		if gate.tryAcquire(c.getSchedulerCount()) {
+			acquired = append(acquired, key)
```
acroca: I find it hard to track when those acquired gate get unreleased.
I think it'd be good to add some `defer` here to release them in case they have not been released (although we don't have a way to know with the current implementation, maybe the acquire should return a release function?), and log an error message when that happens. wdyt?

pkg/scheduler/server/internal/cron/cron.go
```diff
@@ -160,6 +161,18 @@ func (c *cron) Run(ctx context.Context) error {
 	).Run(ctx)
 }
 
+// schedulerCount returns the number of active scheduler hosts in the cluster.
+func (c *cron) schedulerCount() int32 {
+	c.lock.RLock()
+	defer c.lock.RUnlock()
+	n := len(c.currHosts)
```
acroca: Would it be worth to keep the count up to date in a new field as we update the list of hosts, so we don't have to use the `lock` to read this value?

pkg/scheduler/server/internal/pool/loops/connections/concurrency_test.go
JoshVanL: Thanks, I have updated it so we get a deterministic floor/ceiling division so we get the full limit :+1:

pkg/scheduler/server/internal/cron/cron.go
JoshVanL: Yep- I have moved the scheduler count info to be inside the loops which get pushed out to the loops so that we have no locks here at all now :+1:

pkg/scheduler/server/internal/pool/loops/connections/connections.go
JoshVanL: Done :+1:

JoshVanL:
Thanks @acroca PTAL!

pkg/actors/api/reminders.go
```diff
@@ -45,6 +45,11 @@ type CreateReminderRequest struct {
 	Overwrite *bool      `json:"overwrite"`
 
 	FailurePolicy *commonv1.JobFailurePolicy `json:"failure_policy,omitempty"`
+
+	// ConcurrencyKey is an optional key used for per-name concurrency limiting
+	// at the scheduler level. For activities this is the activity name, for
+	// workflows this is the workflow name.
+	ConcurrencyKey *string `json:"-"`
```
javier-aliaga, line 52: nit: I prefer naming this as TypeName or similar, it feels ConcurrencyKey is an implementation detail more than an attribute for the reminder. Then in the scheduler.go it is clear you are using the typeName as the concurrencyKey.

pkg/scheduler/server/internal/pool/loops/connections/concurrency.go
```diff
@@ -0,0 +1,93 @@
...
+		return nil
+	}
+	req := g.pending[0]
+	g.pending[0] = nil
+	g.pending = g.pending[1:]
+
+	if len(g.pending) == 0 {
+		g.pending = nil
+	} else if cap(g.pending) > 64 && len(g.pending)*4 <= cap(g.pending) {
+		pending := make([]*loops.TriggerRequest, len(g.pending))
```
javier-aliaga, line 83: nit: Should we add the capacity here? after this compact action the slice is full so next queue will trigger a new allocation, if you initialize with a cap of double of the pending it will not trigger it

pkg/actors/api/reminders.go
JoshVanL: This is purposefully made generic and the reminder type is actually the implementation detail. This concurrency key can be used in future for other systems other than reminders.

pkg/scheduler/server/internal/pool/loops/connections/concurrency.go
JoshVanL: The point of the compaction is to reduce allocated memory, adding a larger cap negates the point.

javier-aliaga: But for the same reason, if you receive a queue in the next action it would trigger a new allocation. Maybe doubling the cap is to much as you said, but maybe you can give some extra space to avoid reallocations in the next action

pkg/actors/api/reminders.go
javier-aliaga: I did mean to change it only in the reminder request, not in the scheduler

JoshVanL: Sure, but even then it's not quite right- workflows set this field to the workflow or activity name which aren't actually the reminder type at all.

pkg/scheduler/server/internal/pool/loops/connections/concurrency.go
JoshVanL: The compaction only triggers when we've dropped past 75% which is the signal that the burst is over. If another item does come in right after, it's just one cheap Go grow. Adding slack here would basically undo what the shrink is for.

pkg/actors/api/reminders.go
JoshVanL: It is actually quite nice to be able to set any concurrency key here for a group

JoshVanL:
@acroca @javier-aliaga I changed the scheduler proto definition as I thought it was a bit awkward with the `group` field name being obtuse what that means. I have changed the concurrency limit to take in a `oneof` target which is current only actors, which defines the actor `type`. WDYT? https://github.com/dapr/dapr/pull/9775/commits/12bcd84fd6f752ef61b6fba1496b179a4a714dec

javier-aliaga:
@JoshVanL I think it is much better with the typed field 👍

acroca (dismissed):

acroca (approved):

PR #9778, opened 2026-04-10
Add workflow history signing for tamper detection

JoshVanL:
Adds cryptographic signing and verification of workflow history events using the sidecar's mTLS X.509 SVID identity, creating an auditable chain of signatures that is verified each time workflow state is loaded.

How it works

Each workflow execution step produces a batch of history events that are signed and chained to the previous signature:

```
  ┌─────────────────────────────────────────────────────────────┐
  │                    Workflow History                         │
  │  Event 0 ─ Event 1 ─ Event 2 ─ Event 3 ─ Event 4 ─ Event 5  │
  └─────┬──────────┬─────────┬──────────┬─────────┬──────────┬──┘
        │          │         │          │         │          │
        ▼          ▼         ▼          ▼         ▼          ▼
  ┌──────────────────┐  ┌──────────────────┐  ┌──────────────────┐
  │    Signature 0   │  │    Signature 1   │  │    Signature 2   │
  │  Events [0,2)    │  │  Events [2,4)    │  │  Events [4,6)    │
  │  prevDigest: nil ├─►│ prevDigest: d0   ├─►│  prevDigest: d1  │
  │  cert: 0         │  │  cert: 0         │  │  cert: 1         │
  └────────┬─────────┘  └──────────────────┘  └───────┬──────────┘
           │                                          │
           ▼                                          ▼
  ┌───────────────────┐                    ┌───────────────────┐
  │ Cert 0 (Boot 1)   │                    │ Cert 1 (Boot 2)   │
  │ SVID: /ns/default │                    │ SVID: /ns/default │
  │       /myapp      │                    │       /myapp      │
  └───────────────────┘                    └───────────────────┘
```

On every load, the full chain is verified: linkage, contiguity, digests, cryptographic signatures, cert validity, CA trust, and SPIFFE app identity.

Signing

- Each batch of new history events is deterministically marshaled, digested (SHA-256), and signed with the sidecar's SPIFFE private key.
- Signatures chain to the previous via previousSignatureDigest, forming a hash chain from genesis.
- Raw signature bytes are preserved exactly as stored for deterministic chain linking — signatures are never re-marshaled.
- A single signAndSaveState entry point ensures every state persistence path signs before saving; no code path can bypass signing.

Verification and enforcement

- On every state load, the full signature chain is verified: chain linkage, event contiguity, digest recomputation, cryptographic signature, certificate validity at event time, and CA chain-of-trust.
- Signing is a one-way commitment: enabling signing for a workflow is permanent. Disabling signing on a signed workflow, or enabling signing on an unsigned workflow, is a hard verification error. There is no catch-up signing — unsigned history has no integrity proof and is rejected.
- SPIFFE app identity binding: each signing certificate's SPIFFE ID is validated against the workflow's owning app ID, preventing cross-app signature forgery within the same trust domain.
- Inbox event validation: TaskCompleted, TaskFailed, ChildWorkflowInstanceCompleted, and ChildWorkflowInstanceFailed events are validated against scheduled operations in signed history. Injected events referencing non-existent operations are purged from the inbox.
- Timer event type validation: timer reminders must contain TimerFired events; other event types are rejected.
- Metadata bounds checking: all metadata count fields are capped at 1M entries to prevent OOM from inflated state store values.

Configuration

- Controlled by WorkflowSignState feature flag (disabled by default).
- Requires mTLS (Sentry) to be active; without mTLS, the signer is nil and signing does not occur.

### DIFF
--- a/pkg/actors/targets/workflow/orchestrator/run.go
+++ b/pkg/actors/targets/workflow/orchestrator/run.go
@@ -441,41 +441,43 @@ func filterValidInboxEvents(state *wfenginestate.State) []*backend.HistoryEvent 
 
 	valid := make([]*backend.HistoryEvent, 0, len(state.Inbox))
 	for _, e := range state.Inbox {
-		switch {
-		case e.GetTaskCompleted() != nil:
-			taskID := e.GetTaskCompleted().GetTaskScheduledId()
+		// exhaustive linter will error here on missing types not implemented on
+		// the switch.
+		switch et := e.GetEventType().(type) {
+		case *protos.HistoryEvent_TaskCompleted:
+			taskID := et.TaskCompleted.GetTaskScheduledId()
 			if _, ok := scheduledTaskIDs[taskID]; !ok {
 				log.Warnf("Dropping injected inbox event: task result for task %d not scheduled in signed history", taskID)
 				continue
 			}
-		case e.GetTaskFailed() != nil:
-			taskID := e.GetTaskFailed().GetTaskScheduledId()
+		case *protos.HistoryEvent_TaskFailed:
+			taskID := et.TaskFailed.GetTaskScheduledId()
 			if _, ok := scheduledTaskIDs[taskID]; !ok {
 				log.Warnf("Dropping injected inbox event: task failure for task %d not scheduled in signed history", taskID)
 				continue
 			}
-		case e.GetChildWorkflowInstanceCompleted() != nil:
-			taskID := e.GetChildWorkflowInstanceCompleted().GetTaskScheduledId()
+		case *protos.HistoryEvent_ChildWorkflowInstanceCompleted:
+			taskID := et.ChildWorkflowInstanceCompleted.GetTaskScheduledId()
 			if _, ok := createdChildIDs[taskID]; !ok {
 				log.Warnf("Dropping injected inbox event: child workflow result for task %d not created in signed history", taskID)
 				continue
 			}
-		case e.GetChildWorkflowInstanceFailed() != nil:
-			taskID := e.GetChildWorkflowInstanceFailed().GetTaskScheduledId()
+		case *protos.HistoryEvent_ChildWorkflowInstanceFailed:
+			taskID := et.ChildWorkflowInstanceFailed.GetTaskScheduledId()
 			if _, ok := createdChildIDs[taskID]; !ok {
 				log.Warnf("Dropping injected inbox event: child workflow failure for task %d not created in signed history", taskID)
 				continue
 			}
-		case e.GetEventRaised() != nil,
-			e.GetTimerFired() != nil,
-			e.GetExecutionStarted() != nil,
-			e.GetExecutionTerminated() != nil,
-			e.GetExecutionResumed() != nil,
-			e.GetExecutionSuspended() != nil:
+		case *protos.HistoryEvent_EventRaised,
+			*protos.HistoryEvent_TimerFired,
+			*protos.HistoryEvent_ExecutionStarted,
+			*protos.HistoryEvent_ExecutionTerminated,
+			*protos.HistoryEvent_ExecutionResumed,
+			*protos.HistoryEvent_ExecutionSuspended:
 			// Legitimate inbox event types that do not correspond to a previously
 			// scheduled operation.
 		default:
-			log.Warnf("Dropping injected inbox event: unknown event type")
+			log.Warnf("Dropping injected inbox event: unknown event type %T", et)
 			continue
 		}
 		valid = append(valid, e)
--- a/pkg/actors/targets/workflow/orchestrator/run.go
+++ b/pkg/actors/targets/workflow/orchestrator/run.go
@@ -60,6 +60,11 @@ func (o *orchestrator) runWorkflow(ctx context.Context, reminder *actorapi.Remin
 		}
 
 		timerEvent := durableTimer.GetTimerEvent()
+		// Validate the timer event is actually a TimerFired event. A crafted
+		// reminder could contain arbitrary event types to inject into the inbox.
+		if timerEvent.GetTimerFired() == nil {
+			return todo.RunCompletedTrue, fmt.Errorf("workflow actor '%s': timer reminder contains non-TimerFired event type %T", o.actorID, timerEvent.GetEventType())
+		}
 		// timer fired event is precreated at the moment of creating the timer
 		// set the timestamp to now so it is accurately recorded in the history
 		timerEvent.Timestamp = timestamppb.Now()
@@ -71,6 +76,29 @@ func (o *orchestrator) runWorkflow(ctx context.Context, reminder *actorapi.Remin
 		// for some of those already processed events.
 		log.Debugf("Workflow actor '%s': ignoring run request for reminder '%s' because the workflow inbox is empty", o.actorID, reminder.Name)
 		return todo.RunCompletedTrue, nil
+	}
+
+	// Validate inbox events: result events (TaskCompleted/TaskFailed,
+	// ChildWorkflowInstanceCompleted/Failed) must match operations that
+	// were scheduled in signed history. Invalid events are purged from
+	// the inbox and the state is saved to prevent the same invalid events
+	// from blocking execution on retry.
+	if o.signer != nil {
+		filtered := filterValidInboxEvents(state)
+		if len(filtered) != len(state.Inbox) {
+			log.Warnf("Workflow actor '%s': purged %d injected inbox events that did not match signed history",
+				o.actorID, len(state.Inbox)-len(filtered))
+			// Clear and re-add valid events so state change tracking
+			// generates the correct delete operations for the state store.
+			state.ClearInbox()
+			for _, e := range filtered {
+				state.AddToInbox(e)
+			}
+			if err = o.signAndSaveState(ctx, state); err != nil {
+				return todo.RunCompletedFalse, fmt.Errorf("failed to save state after purging injected events: %w", err)
+			}
+			return todo.RunCompletedFalse, fmt.Errorf("workflow actor '%s': inbox contained injected events that were purged; retrying", o.actorID)
+		}
 	}
 
 	var esHistoryEvent *backend.HistoryEvent
@@ -197,7 +225,7 @@ func (o *orchestrator) runWorkflow(ctx context.Context, reminder *actorapi.Remin
 
 				state.Generation++
 
-				if err = o.saveInternalState(ctx, state); err != nil {
+				if err = o.signAndSaveState(ctx, state); err != nil {
 					o.rstate = rstateSnapshot
 					return todo.RunCompletedFalse, err
 				}
@@ -299,7 +327,7 @@ func (o *orchestrator) runWorkflow(ctx context.Context, reminder *actorapi.Remin
 			rs.NewEvents = filtered
 			state.ApplyRuntimeStateChanges(rs)
 			rs.NewEvents = origNewEvents
-			if saveErr := o.saveInternalState(ctx, state); saveErr != nil {
+			if saveErr := o.signAndSaveState(ctx, state); saveErr != nil {
 				return todo.RunCompletedFalse, saveErr
 			}
 			executionStatus = diag.StatusRecoverable
@@ -313,7 +341,7 @@ func (o *orchestrator) runWorkflow(ctx context.Context, reminder *actorapi.Remin
 	state.ApplyRuntimeStateChanges(rs)
 	state.ClearInbox()
 
-	err = o.saveInternalState(ctx, state)
+	err = o.signAndSaveState(ctx, state)
 	if err != nil {
 		return todo.RunCompletedFalse, err
 	}
@@ -415,4 +443,55 @@ func (o *orchestrator) handleRetention(ctx context.Context, status protos.Orches
 	}
 
 	return nil
+}
+
+// filterValidInboxEvents returns inbox events that pass validation. Result
+// events (TaskCompleted/TaskFailed, ChildWorkflowInstanceCompleted/Failed)
+// must match operations that were scheduled in signed history. Invalid events
+// are dropped and logged.
+func filterValidInboxEvents(state *wfenginestate.State) []*backend.HistoryEvent {
+	// Build sets of scheduled operation event IDs from history.
+	scheduledTaskIDs := make(map[int32]struct{})
+	createdChildIDs := make(map[int32]struct{})
+	for _, e := range state.History {
+		switch {
+		case e.GetTaskScheduled() != nil:
+			scheduledTaskIDs[e.GetEventId()] = struct{}{}
+		case e.GetChildWorkflowInstanceCreated() != nil:
+			createdChildIDs[e.GetEventId()] = struct{}{}
+		}
+	}
+
+	valid := make([]*backend.HistoryEvent, 0, len(state.Inbox))
+	for _, e := range state.Inbox {
+		switch {
+		case e.GetTaskCompleted() != nil:
+			taskID := e.GetTaskCompleted().GetTaskScheduledId()
+			if _, ok := scheduledTaskIDs[taskID]; !ok {
+				log.Warnf("Dropping injected inbox event: task result for task %d not scheduled in signed history", taskID)
+				continue
+			}
+		case e.GetTaskFailed() != nil:
+			taskID := e.GetTaskFailed().GetTaskScheduledId()
+			if _, ok := scheduledTaskIDs[taskID]; !ok {
+				log.Warnf("Dropping injected inbox event: task failure for task %d not scheduled in signed history", taskID)
+				continue
+			}
+		case e.GetChildWorkflowInstanceCompleted() != nil:
+			taskID := e.GetChildWorkflowInstanceCompleted().GetTaskScheduledId()
+			if _, ok := createdChildIDs[taskID]; !ok {
+				log.Warnf("Dropping injected inbox event: child workflow result for task %d not created in signed history", taskID)
+				continue
+			}
+		case e.GetChildWorkflowInstanceFailed() != nil:
+			taskID := e.GetChildWorkflowInstanceFailed().GetTaskScheduledId()
+			if _, ok := createdChildIDs[taskID]; !ok {
+				log.Warnf("Dropping injected inbox event: child workflow failure for task %d not created in signed history", taskID)
+				continue
+			}
+		}
+		valid = append(valid, e)
+	}
+
+	return valid
 }
--- a/pkg/actors/targets/workflow/orchestrator/run.go
+++ b/pkg/actors/targets/workflow/orchestrator/run.go
@@ -78,29 +78,6 @@ func (o *orchestrator) runWorkflow(ctx context.Context, reminder *actorapi.Remin
 		return todo.RunCompletedTrue, nil
 	}
 
-	// Validate inbox events: result events (TaskCompleted/TaskFailed,
-	// ChildWorkflowInstanceCompleted/Failed) must match operations that
-	// were scheduled in signed history. Invalid events are purged from
-	// the inbox and the state is saved to prevent the same invalid events
-	// from blocking execution on retry.
-	if o.signer != nil {
-		filtered := filterValidInboxEvents(state)
-		if len(filtered) != len(state.Inbox) {
-			log.Warnf("Workflow actor '%s': purged %d injected inbox events that did not match signed history",
-				o.actorID, len(state.Inbox)-len(filtered))
-			// Clear and re-add valid events so state change tracking
-			// generates the correct delete operations for the state store.
-			state.ClearInbox()
-			for _, e := range filtered {
-				state.AddToInbox(e)
-			}
-			if err = o.signAndSaveState(ctx, state); err != nil {
-				return todo.RunCompletedFalse, fmt.Errorf("failed to save state after purging injected events: %w", err)
-			}
-			return todo.RunCompletedFalse, fmt.Errorf("workflow actor '%s': inbox contained injected events that were purged; retrying", o.actorID)
-		}
-	}
-
 	var esHistoryEvent *backend.HistoryEvent
 	for _, e := range state.Inbox {
 		if es := e.GetExecutionStarted(); es != nil {
@@ -489,6 +466,17 @@ func filterValidInboxEvents(state *wfenginestate.State) []*backend.HistoryEvent 
 				log.Warnf("Dropping injected inbox event: child workflow failure for task %d not created in signed history", taskID)
 				continue
 			}
+		case e.GetEventRaised() != nil,
+			e.GetTimerFired() != nil,
+			e.GetExecutionStarted() != nil,
+			e.GetExecutionTerminated() != nil,
+			e.GetExecutionResumed() != nil,
+			e.GetExecutionSuspended() != nil:
+			// Legitimate inbox event types that do not correspond to a previously
+			// scheduled operation.
+		default:
+			log.Warnf("Dropping injected inbox event: unknown event type")
+			continue
 		}
 		valid = append(valid, e)
 	}
--- a/pkg/actors/targets/workflow/orchestrator/signing.go
+++ b/pkg/actors/targets/workflow/orchestrator/signing.go
@@ -38,6 +38,15 @@ func (o *orchestrator) signNewEvents(state *wfenginestate.State) error {
 		return nil
 	}
 
+	// Defensive: the added count cannot exceed the history length. Without
+	// this guard, the int subtraction below would underflow and the uint64
+	// cast would produce a massive startIndex, panicking in the marshal loop
+	// with a misleading out-of-bounds error instead of a clear message here.
+	if newEventCount > len(state.History) {
+		return fmt.Errorf("signNewEvents called with newEventCount=%d but history has only %d events",
+			newEventCount, len(state.History))
+	}
+
 	//nolint:gosec
 	startIndex := uint64(len(state.History) - newEventCount)
 
--- a/pkg/actors/targets/workflow/orchestrator/signing.go
+++ b/pkg/actors/targets/workflow/orchestrator/signing.go
@@ -1,0 +1,99 @@
+/*
+Copyright 2026 The Dapr Authors
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package orchestrator
+
+import (
+	"context"
+	"fmt"
+
+	actorsapi "github.com/dapr/dapr/pkg/actors/api"
+	wfenginestate "github.com/dapr/dapr/pkg/runtime/wfengine/state"
+	"github.com/dapr/durabletask-go/backend/historysigning"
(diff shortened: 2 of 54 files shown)


0xbrainkid:
The SPIFFE SVID-anchored workflow history signing is the right architecture — using the sidecar's existing mTLS identity as the signing key means no additional credential management, and the certificate rotation handling (cert field in each signature, chain verification on load) handles the key lifecycle problem cleanly.

A few observations on how this extends to cross-org trust scenarios:

**1. SVID as cross-org identity anchor.** The SPIFFE ID (`/ns/default/myapp`) is namespace-scoped, which works within a trust domain. For workflows that delegate steps across trust domains (Organization A's Dapr sidecar calling Organization B's workflow), the signature chain can verify tamper-evidence within each domain, but cross-domain verification requires both organizations to share a trust anchor. Adding the full X.509 certificate (not just a cert index) to each signature batch makes the chain verifiable by external parties who have access to the intermediate CA but not the full SPIFFE infrastructure.

**2. Behavioral trust derivation from history chains.** A workflow history with a complete, verified signature chain is a behavioral track record — the chain proves what steps executed, in what order, with what identity, with no gaps. This is the raw data for behavioral trust scoring. An agent whose Dapr workflow histories are consistently complete and signed earns higher trust than one with gaps or verification failures. SATP can consume verified history chains as attestation inputs.

**3. Signing granularity.** The current design signs batches of events `[0,2)`, `[2,4)`. For long-running workflows where individual steps involve high-stakes actions, per-step signatures (batch size 1) may be appropriate as an opt-in mode — the overhead is higher but the tamper-evidence is stronger for forensic purposes.

The CA trust + SPIFFE app identity verification on load is essential — chain integrity without identity verification only proves the events are consistent, not that they came from the expected workload.

tests/integration/framework/workflow/signing.go
```diff
@@ -0,0 +1,142 @@
...
+	"github.com/dapr/durabletask-go/backend/historysigning"
+	"github.com/dapr/kit/crypto/spiffe/signer"
+	"github.com/dapr/kit/crypto/spiffe/trustanchors/fake"
+)
+
+// UnmarshalSigningData reads and unmarshals signatures, certificates, and raw
+// history events from the SQLite state store for the given workflow instance.
+// Returns raw signature bytes alongside parsed protos, as raw bytes are needed
+// for digest computation in chain verification.
+func UnmarshalSigningData(t *testing.T, ctx context.Context, db *sqlite.SQLite, instanceID string) (rawSigs [][]byte, sigs []*protos.HistorySignature, certs []*protos.SigningCertificate, rawEvents [][]byte) {
```
acroca: nit: I think returning a struct would look better, otherwise we need to remember constantly which returned value is which.

tests/integration/suite/daprd/workflow/signing/childinjection.go
```diff
@@ -0,0 +1,138 @@
...
+
+	// Wait until the parent is waiting for the external event (child has
+	// completed and the parent has progressed past CallChildWorkflow).
+	require.EventuallyWithT(tt, func(c *assert.CollectT) {
+		meta, err = client.FetchWorkflowMetadata(ctx, id)
+		if !assert.NoError(c, err) {
+			return
+		}
+		assert.Equal(c, dworkflow.StatusRunning, meta.RuntimeStatus)
+	}, time.Second*10, time.Millisecond*100)
```
acroca: Isn't this doing the same thing?

tests/integration/suite/daprd/workflow/signing/childinjection.go
```diff
@@ -0,0 +1,138 @@
...
+	"github.com/dapr/dapr/tests/integration/suite"
+	dworkflow "github.com/dapr/durabletask-go/workflow"
+)
+
+func init() {
+	suite.Register(new(childInjection))
+}
+
+// childInjection verifies that injecting a fake ChildWorkflowInstanceCompleted
+// event into the inbox is rejected when signing is enabled. A
```
acroca, line 46: I fail to see where the fake is injected 😕

tests/integration/suite/daprd/workflow/signing/continueasnew.go
```diff
@@ -0,0 +1,111 @@
...
+	_, err = client.WaitForWorkflowCompletion(ctx, id)
+	require.NoError(t, err)
+
+	fworkflow.VerifySignatureChain(t, ctx, c.db, id, c.sentry.CABundle().X509.TrustAnchors)
+	fworkflow.VerifyCertAppID(t, ctx, c.db, id, c.daprd.AppID())
+
+	// ContinueAsNew resets history, so the history record count should be small.
+	//nolint:dogsled
+	_, _, _, rawEvents := fworkflow.UnmarshalSigningData(t, ctx, c.db, id)
+	assert.Less(t, len(rawEvents), 10, "ContinueAsNew should reset history, keeping it small")
```
acroca: I find this `10` a bit random. I think it'd be better to have the fixed number we expect, even if that means tweaking the test every time we change the add extra events in this kind of workflows.

tests/integration/suite/daprd/workflow/signing/disabled.go
```diff
@@ -0,0 +1,104 @@
...
+	client := dworkflow.NewClient(d.daprd.GRPCConn(tt, ctx))
+	require.NoError(tt, client.StartWorker(ctx, reg))
+
+	id, err := client.ScheduleWorkflow(ctx, "sign-disabled")
+	require.NoError(tt, err)
+
+	_, err = client.WaitForWorkflowCompletion(ctx, id)
+	require.NoError(tt, err)
+
+	assert.Equal(tt, 0, d.db.CountStateKeys(tt, ctx, "signature"))
```
acroca: What do you think if we move the `d.db.CountStateKeys(tt, ctx, "signature")` to a function in the `signing.go` in the workflow framework? This way we ensure all tests go through the same code path. If this code were checking an invalid key it would be passing by mistake, but if it calls a shared function there's no way it'd fail by mistake.

tests/integration/suite/daprd/workflow/signing/enablesigning.go
```diff
@@ -0,0 +1,127 @@
...
+	t.Cleanup(func() { daprd2.Cleanup(t) })
+	daprd2.WaitUntilRunning(t, ctx)
+
+	client2 := dworkflow.NewClient(daprd2.GRPCConn(t, ctx))
+	require.NoError(t, client2.StartWorker(ctx, reg))
+
+	// The workflow should fail to load because the unsigned history
+	// from the non-signing host has no integrity proof.
+	_, err = client2.FetchWorkflowMetadata(ctx, id)
+	require.Error(t, err)
```
acroca, line 126: Should we expose a way to fetch ignoring signatures? I think just make it fail could be problematic for some existing users if they want to start using this feature. 
Also, what will happen with in-flight workflows?

tests/integration/suite/daprd/workflow/signing/inboxinjection.go
```diff
@@ -0,0 +1,193 @@
...
+		Timestamp: timestamppb.Now(),
+		EventType: &protos.HistoryEvent_TaskCompleted{
+			TaskCompleted: &protos.TaskCompletedEvent{
+				TaskScheduledId: int32(9999),
+				Result:          wrapperspb.String(`"injected"`),
+			},
+		},
+	}
+	raw, err := proto.Marshal(fakeEvt)
+	require.NoError(tt, err)
```
acroca, line 140: The workflow is stuck in a `waitForExternalEvent` and doesn't have any activities, wouldn't it fail also because the history doesn't match the code? Shouldn't this test create a fake `continue` event instead?

acroca (commented):
Didn't finish the review, but I've got this so far.

tests/integration/suite/daprd/workflow/signing/enablesigning.go
JoshVanL: yeah we need to make a decision on what we want to do here- currently as it stands its all or nothing, and enabling this feature with in flight workflows will cause them to fail.
We _could_ implement some "catch up" signing logic, however this can be problamtic as you could sign basically anything toggling the feature.

What do you think?

pkg/runtime/wfengine/state/state.go
```diff
@@ -396,29 +650,125 @@ func LoadWorkflowState(ctx context.Context, state state.Interface, actorID strin
...
-		ActorType: s.workflowActorType,
-		ActorID:   actorID,
-		// Initial capacity should be enough to contain the entire inbox, history, and custom status + metadata
-		Operations: make([]api.TransactionalOperation, 0, len(s.Inbox)+len(s.History)+2),
+func verifySignatureChain(s *State, sgn *signer.Signer) error {
+	rawEvents := s.RawHistory
+	if len(s.Signatures) > 0 {
+		last := s.Signatures[len(s.Signatures)-1]
+		coveredEvents := last.GetStartEventIndex() + last.GetEventCount()
+		if coveredEvents < uint64(len(rawEvents)) {
```
cicoyle:
```suggestion
		if coveredEvents != uint64(len(rawEvents)) {
```
Shouldnt it be `!=` bc we should care if its `>` && `<`

tests/integration/suite/daprd/workflow/signing/continueasnew.go
```diff
@@ -0,0 +1,111 @@
...
+	c.daprd.WaitUntilRunning(t, ctx)
+
+	reg := dworkflow.NewRegistry()
+	reg.AddWorkflowN("sign-can", func(ctx *dworkflow.WorkflowContext) (any, error) {
+		var iteration int
+		if err := ctx.GetInput(&iteration); err != nil {
+			return nil, err
+		}
+		if iteration < 3 {
+			ctx.ContinueAsNew(iteration + 1)
```
cicoyle, line 89: Should we check that `GetPreviousSignatureDigest` is empty in this ContinueAsNew IT test?

pkg/actors/targets/workflow/orchestrator/run.go
```diff
@@ -416,3 +444,54 @@ func (o *orchestrator) handleRetention(ctx context.Context, status protos.Orches
...
+		switch {
+		case e.GetTaskScheduled() != nil:
+			scheduledTaskIDs[e.GetEventId()] = struct{}{}
+		case e.GetChildWorkflowInstanceCreated() != nil:
+			createdChildIDs[e.GetEventId()] = struct{}{}
+		}
+	}
+
+	valid := make([]*backend.HistoryEvent, 0, len(state.Inbox))
+	for _, e := range state.Inbox {
```
cicoyle, line 443: The switch handles 4 event types and falls thru to append. Shall we add a default case to reject an unknown type - otherwise we append all types - or flip the switch to a fail close instead of fail open.

pkg/actors/targets/workflow/orchestrator/run.go
```diff
@@ -73,6 +78,29 @@ func (o *orchestrator) runWorkflow(ctx context.Context, reminder *actorapi.Remin
 		return todo.RunCompletedTrue, nil
 	}
 
+	// Validate inbox events: result events (TaskCompleted/TaskFailed,
+	// ChildWorkflowInstanceCompleted/Failed) must match operations that
+	// were scheduled in signed history. Invalid events are purged from
+	// the inbox and the state is saved to prevent the same invalid events
+	// from blocking execution on retry.
+	if o.signer != nil {
+		filtered := filterValidInboxEvents(state)
```
cicoyle: Shouldn't we treat this the same as signature tampering? An inbox event that doesn't match signed history can only get there via state store tampering, which is the same threat model as a forged signature. As-is, the code silently purges the injected events and saves state, destroying the tamper evidence. Can we call `failSignatureVerification` (or an equivalent that can apply here too) and leave the injected events in place for users to investigate and fail the wf?

cicoyle (changes requested):

pkg/runtime/wfengine/state/state.go
```diff
@@ -15,26 +15,40 @@ package state
...
+	historyKeyPrefix   = "history"
+	sigcertKeyPrefix   = "sigcert"
+	signatureKeyPrefix = "signature"
+	customStatusKey    = "customStatus"
+	metadataKey        = "metadata"
+
+	// maxStateEntries is the upper bound for any metadata count field
+	// (inbox, history, signatures, certificates). This prevents OOM from
+	// an attacker inflating metadata values in the state store.
+	maxStateEntries = 1_000_000
```
javier-aliaga, line 51: Is this configurable? If not, should it be? I am not sure this is very unlikely to be hit or something the users will want to reduce.

pkg/runtime/wfengine/state/state.go
```diff
@@ -396,29 +650,125 @@ func LoadWorkflowState(ctx context.Context, state state.Interface, actorID strin
 
 	wfLogger.Debugf("%s: loaded %d state records in %v", actorID, 1+len(bulkRes), time.Since(loadStartTime))
 
+	// Signing enforcement. Once signing is enabled for a cluster, it is a
+	// one-way commitment: workflows created with signing must always run on
+	// signing-enabled hosts, and workflows created without signing cannot be
+	// migrated to signing-enabled hosts. Workflows that violate these rules are
+	// terminated. Operators must ensure all unsigned workflows complete or are
+	// purged before enabling signing cluster-wide.
```
javier-aliaga, line 658: do you have to enable it for a cluster or can it be for an app? Even for an app, it may be a problem for users that want to enable this and have long running workflows. 

Are workflows completed with failed status?

tests/integration/suite/daprd/workflow/signing/enablesigning.go
acroca: All options I can think of to support progressive migration would be subject to tampering.
- Setting a cut-over time where all events from that time need signature. An attacker could just change history to move events before the cut-over and the tampered events will just be accepted
- Signing the existing events the first time the workflow runs. The attacker could just remove all signatures and next time the workflow runs it'd sign it again.
- Sign only new processed events, and leave old ones unsigned, the signature chain would start half-way. Same attack vector, removing signatures breaks it.

The only approach I can think of is to enforce signatures, but introduce a one-off script, maybe in the CLI, that signs all in-flight workflows. An attacker that can only access the state store won't be able to sign anything, and removing signatures will break existing workflows.

tests/integration/suite/daprd/workflow/signing/disabled.go
```diff
@@ -0,0 +1,105 @@
...
+	require.NoError(tt, err)
+
+	_, err = client.WaitForWorkflowCompletion(ctx, id)
+	require.NoError(tt, err)
+
+	assert.Equal(tt, 0, fworkflow.SignatureCount(tt, ctx, d.db, id))
+	assert.Equal(tt, 0, fworkflow.CertificateCount(tt, ctx, d.db, id))
+
+	// Verify that history events were still produced.
+	assert.Positive(tt, d.db.CountStateKeys(tt, ctx, "history"))
```
acroca: Same as https://github.com/dapr/dapr/pull/9778/changes/f444c522ec2f250ff8835b65eed3deda279a4882?new_files_changed=true#r2084838975, I think we should move this to `fworflow` as the signature and certificate counts.

tests/integration/suite/daprd/workflow/signing/multiactivity.go
```diff
@@ -0,0 +1,112 @@
...
+	client := dworkflow.NewClient(m.daprd.GRPCConn(t, ctx))
+	require.NoError(t, client.StartWorker(ctx, reg))
+
+	id, err := client.ScheduleWorkflow(ctx, "sign-multi")
+	require.NoError(t, err)
+
+	_, err = client.WaitForWorkflowCompletion(ctx, id)
+	require.NoError(t, err)
+
+	assert.GreaterOrEqual(t, fworkflow.SignatureCount(t, ctx, m.db, id), 5)
```
acroca: Can we set the specific number here? If I understand correctly this workflow will always generate the same amount of signatures, right?

tests/integration/suite/daprd/workflow/signing/multiapp.go
```diff
@@ -0,0 +1,128 @@
...
+
+	// Schedule on the orchestrator instance.
+	id, err := client1.ScheduleWorkflow(ctx, "sign-xapp")
+	require.NoError(t, err)
+
+	_, err = client1.WaitForWorkflowCompletion(ctx, id)
+	require.NoError(t, err)
+
+	fworkflow.VerifySignatureChain(t, ctx, m.db, id, m.sentry.CABundle().X509.TrustAnchors)
+	fworkflow.VerifyCertAppID(t, ctx, m.db, id, m.daprd1.AppID())
```
acroca, line 132: I think it'd be good to have the negative check using `daprd2`

tests/integration/suite/daprd/workflow/signing/multiappchild.go
```diff
@@ -0,0 +1,134 @@
...
+}
+
+func (m *multiappchild) Run(t *testing.T, ctx context.Context) {
+	m.daprd1.WaitUntilRunning(t, ctx)
+	m.daprd2.WaitUntilRunning(t, ctx)
+
+	reg := dworkflow.NewRegistry()
+	reg.AddWorkflowN("sign-xapp-parent", func(ctx *dworkflow.WorkflowContext) (any, error) {
+		var result string
+		if err := ctx.CallActivity("remote-act").Await(&result); err != nil {
```
acroca: I don't think this is calling the activity on the other app as stated in the scenario description. Or am I missing something?

tests/integration/suite/daprd/workflow/signing/multireplicachild.go
```diff
@@ -0,0 +1,154 @@
...
+			daprd.WithScheduler(m.sched),
+			daprd.WithResourceFiles(m.db.GetComponent(t)),
+			daprd.WithConfigManifests(t, fmt.Sprintf(`apiVersion: dapr.io/v1alpha1
+kind: Configuration
+metadata:
+  name: sign-on-%d
+spec:
+  features:
+  - name: WorkflowSignState
+    enabled: true
```
acroca, line 79: What would happen if one of the replicas doesn't have this feature enabled? Should we add this scenario and make sure it errors?

tests/integration/suite/daprd/workflow/signing/nomtlserror.go
```diff
@@ -0,0 +1,82 @@
...
+	return []framework.Option{
+		framework.WithProcesses(n.logline, db, n.place, n.sched, daprd),
+	}
+}
+
+func (n *nomtlserror) Run(t *testing.T, ctx context.Context) {
+	n.sched.WaitUntilRunning(t, ctx)
+	n.place.WaitUntilRunning(t, ctx)
+
+	n.logline.EventuallyFoundAll(t)
```
acroca, line 85: I don't think this checks daprd actually stopped, right?

tests/integration/suite/daprd/workflow/signing/signaturestripped.go
```diff
@@ -0,0 +1,148 @@
...
+	assert.Positive(tt, fworkflow.SignatureCount(tt, ctx, s.db, id))
+	assert.Positive(tt, fworkflow.CertificateCount(tt, ctx, s.db, id))
+
+	db := s.db.GetConnection(tt)
+	tableName := s.db.TableName()
+
+	// Delete all signature-* and sigcert-* keys from the state store.
+	//nolint:gosec
+	_, err = db.ExecContext(ctx,
+		"DELETE FROM "+tableName+" WHERE key LIKE '%||signature-%' OR key LIKE '%||sigcert-%'",
```
acroca: Do you think it'd be worth to have different test for each of the fields you're removing here? I think the 4 attributes need to exist, so just removing one should make it fail right?

acroca (commented):
I think it'd be good to also cover multi-app as in calling a child workflow or activity in a different app ID

tests/integration/suite/daprd/workflow/signing/enablesigning.go
JoshVanL: Yep- this is the same conclusion I came to as well

I like the script idea which we can put inside the CLI, I think we can investigate this approach. We will want to supplement the CLI anyhow with some signing result output so users can see they workflows being signed as well.

pkg/runtime/wfengine/state/state.go
JoshVanL: This is currently just a cheap OOM gate. We should make this configurable I think in a followup.

pkg/runtime/wfengine/state/state.go
JoshVanL: This is enabled per app. If this is enabled and any signature verification fails, then the workflow fails.

tests/integration/suite/daprd/workflow/signing/multireplicachild.go
JoshVanL: This should be covered in `signinggap.go`. This results in verification failing and the workflow failing.

If the question is around mult-application workflows then it  currently doesn't matter as the workflow signs it's own work and the signatures do not cross boundaries. We will follow up with an attestation for input/output of cross-app things- https://github.com/dapr/durabletask-protobuf/pull/41

tests/integration/suite/daprd/workflow/signing/continueasnew.go
JoshVanL: done in continueasnew.go

pkg/actors/targets/workflow/orchestrator/run.go
JoshVanL: Done

tests/integration/suite/daprd/workflow/signing/signaturestripped.go
JoshVanL: done

pkg/actors/targets/workflow/orchestrator/run.go
```diff
@@ -416,3 +421,65 @@ func (o *orchestrator) handleRetention(ctx context.Context, status protos.Orches
...
+			}
+		case e.GetEventRaised() != nil,
+			e.GetTimerFired() != nil,
+			e.GetExecutionStarted() != nil,
+			e.GetExecutionTerminated() != nil,
+			e.GetExecutionResumed() != nil,
+			e.GetExecutionSuspended() != nil:
+			// Legitimate inbox event types that do not correspond to a previously
+			// scheduled operation.
+		default:
```
acroca, line 479: It's very possible we'll forget to extend this switch whenever we add new event types. I think we have this kind of switch in many other places and I'm wondering if we should have a different way to ensure at compile time that all types are checked. What do you think?

pkg/actors/targets/workflow/orchestrator/signing.go
```diff
@@ -0,0 +1,99 @@
...
+	"github.com/dapr/durabletask-go/backend/historysigning"
+)
+
+// signNewEvents deterministically marshals the newly added history events,
+// creates a HistorySignature covering them, and appends it (along with any
+// new signing certificate) to the state.
+// The marshaled bytes are stored on the state so that GetSaveRequest can
+// persist the exact bytes that were signed.
+// If signer is nil (mTLS disabled or feature flag off), this is a no-op.
+func (o *orchestrator) signNewEvents(state *wfenginestate.State, newEventCount int) error {
```
acroca:
```suggestion
func (o *orchestrator) signNewEvents(state *wfenginestate.State) error {
  newEventCount = state.HistoryAddedCount()
```

pkg/actors/targets/workflow/orchestrator/signing.go
```diff
@@ -0,0 +1,99 @@
...
+		}
+		rawNewEvents[i] = data
+	}
+
+	state.SetMarshaledNewHistory(rawNewEvents)
+
+	var prevSigRaw []byte
+	if len(state.RawSignatures) > 0 {
+		prevSigRaw = state.RawSignatures[len(state.RawSignatures)-1]
+	}
```
acroca, line 79: I think we could be a bit more defensive here and make sure we're doing the very first signature, so the count of events should be same length as the whole history. Just to protect this to situations where the `RawSignatures` are not properly populated and we generate signatures that don't take into consideration the previous ones. I guess those signatures won't be valid though, as the validator requires them...

acroca (commented):
LGTM, just some comments

pkg/actors/targets/workflow/orchestrator/signing.go
JoshVanL: Added this: 

```
+       // Defensive: when there is no previous signature, this must be the very
+       // first signature for the workflow, which requires the batch to cover the
+       // entire history. Otherwise we'd produce a chain-less signature over a
+       // suffix of the history which the verifier would reject, but the defect
+       // would only surface on reload. Fail fast instead.
+       if len(state.RawSignatures) == 0 && startIndex != 0 {
+               return fmt.Errorf("cannot sign partial history (%d events, starting at index %d) without a previous signature",
+                       newEventCount, startIndex)
+       }
```

pkg/actors/targets/workflow/orchestrator/signing.go
```diff
@@ -0,0 +1,110 @@
...
+		return nil
+	}
+
+	newEventCount := state.HistoryAddedCount()
+	if newEventCount == 0 {
+		return nil
+	}
+
+	//nolint:gosec
+	startIndex := uint64(len(state.History) - newEventCount)
```
cicoyle, line 51: Should we add a guard for:
```
if newEventCount > len(state.History) {
    return fmt.Errorf("signNewEvents called with newEventCount=%d but history has only %d events",
        newEventCount, len(state.History))
}
```
Otherwise, `len(state.History) - newEventCount` underflows when cast to uint64 and `startIndex` becomes a huge positive number. Which then the logged err is a bit misleading. I think it could be clearer to just add a new conditional as a check.

pkg/actors/targets/workflow/orchestrator/state.go
```diff
@@ -43,14 +46,34 @@ func (o *orchestrator) loadInternalState(ctx context.Context) (*wfenginestate.St
...
 	if state == nil {
 		// No such state exists in the state store
 		return nil, nil, nil
 	}
+
+	// When signing is enabled, any inbox event that does not match signed
+	// history can only have been written via state store tampering. Treat
+	// the state as unrecoverable: fail the workflow terminally so no
+	// further progress is made on forged input.
+	if o.signer != nil {
```
cicoyle: Should we add a check here to only run this if there is something in the inbox?
```suggestion
	if o.signer != nil && len(state.Inbox) > 0 {
```

pkg/config/configuration.go
```diff
@@ -68,6 +68,12 @@ const (
 	// When enabled, daprd loads MCPServer manifests at startup,
 	// and makes them available for tool execution via workflow orchestrations.
 	MCPServerResource Feature = "MCPServerResource"
+
+	// WorkflowSignState enables cryptographic signing of workflow history
+	// state. When enabled and mTLS is active, each workflow execution's
+	// history events are signed using the app's X.509 SVID identity,
+	// creating a verifiable chain of signatures. Disabled by default.
+	WorkflowSignState Feature = "WorkflowSignState"
```
cicoyle: Should the feature flag be called `WorkflowHistorySigning`?

cicoyle (changes requested):
few more things from me

pkg/actors/targets/workflow/orchestrator/signing.go
JoshVanL: Addressed

pkg/actors/targets/workflow/orchestrator/state.go
JoshVanL: Addressed

pkg/config/configuration.go
JoshVanL: Addressed

acroca (approved):

cicoyle (approved):

PR #9788, opened 2026-04-13
Runtime: adds jitter to scheduler & placement re-conns

JoshVanL:
Adds a jitter to the runtime's scheduler and placement connection loops to prevent thundering herd when many instances reconnect simultaneously.


pkg/retry/retry.go
```diff
@@ -13,9 +13,19 @@ limitations under the License.
...
 	DefaultLinearBackoffInterval = time.Second
 	DefaultLinearRetryCount      = 3
 )
+
+// Jitter returns a random duration in the range [base/2, base*3/2).
+//
+//nolint:gosec
+func Jitter(base time.Duration) time.Duration {
+	return base/2 + time.Duration(rand.Int63n(int64(base)))
+}
```
javier-aliaga: Dont you think this produces values too spread when the time is in the order of minutes? For example for 1 min it could return from 30s to 90s.

I checked the current call and it is 3s and 500ms, what should be fine, I am more worried we use this func in the future with higher values.

Would there be any problem for the Jitter returning a value before the original?

acroca: What if instead we pass another argument with the jitter? so a call like `Jitter(time.Minute, time.Second)` would return a number between 59 and 61 seconds. Too noisy maybe?

JoshVanL: I have updated the `Jitter` func to now take a range which it works on.

javier-aliaga (approved):

acroca (approved):

PR #9789, opened 2026-04-13
Workflow: default WorkflowsRemoteActivityReminder true

JoshVanL:
Mark this feature gate as enabled by default since 1.17 ships with this functionality. This feature is a QoL for sending activity results via reminders from cross app workflows, queuing up results if the target workflow is currently not available.


acroca (approved):

pkg/config/configuration.go
```diff
@@ -83,7 +83,9 @@ const (
 	ActionPolicyGlobal  = "global"
 )
 
-var defaultFeatures = make(map[Feature]bool)
+var defaultFeatures = map[Feature]bool{
+	WorkflowsRemoteActivityReminder: true,
```
acroca, line 87: Users can only enable features right? If they can't disable features, why don't we just delete it as a feature?

JoshVanL: Users can disable features by setting them to false

javier-aliaga (approved):

cicoyle (approved):

cicoyle:
Is there a docs PR for this?

JoshVanL:
@cicoyle good shout, I have opened that here https://github.com/dapr/docs/pull/5128

PR #9790, opened 2026-04-13
Workflows; Adds WorkflowAccessPolicy

JoshVanL:
Introduces a new Kubernetes CRD (WorkflowAccessPolicy) that controls which Dapr app IDs are permitted to schedule specific workflows and activities on a target application. This is a preview feature gated behind the WorkflowAccessPolicy feature flag.

Enforcement:
- Remote calls: callee-side via CallActor/CallActorStream/CallActorReminder gRPC handlers using SPIFFE identity from mTLS peer certificates
- Local calls: router-level WorkflowACLChecker for same-sidecar enforcement
- Non-subject methods (AddWorkflowEvent, PurgeWorkflowState): caller must appear in at least one allow rule (IsCallerKnown)
- PermissionDenied errors propagated back to parent workflows via reminder-based failure events (avoids ClearInbox race)
- Error messages returned to callers are opaque ("access denied by workflow access policy") to avoid leaking internal state

Policy resolution:
- Specificity-based: longest literal prefix wins, exact beats glob, deny beats allow at same specificity
- Glob patterns via Go path.Match (*, ?, [abc])
- Multiple policies merged with aggregate default (deny wins)


0xbrainkid:
WorkflowAccessPolicy with SPIFFE identity enforcement is a strong design — using the mTLS peer certificate as the identity source means no additional credentials are needed, and callee-side enforcement via gRPC handlers closes the server-side gap that caller-side enforcement leaves open (a malicious caller could bypass caller-side checks by calling the gRPC endpoint directly).

A few observations from a cross-org trust perspective:

**1. Opaque error messages are the right call.** Returning "access denied by workflow access policy" rather than the specific violated rule prevents information leakage about the policy structure — an attacker who knows which rules they are hitting can probe the policy boundary. Worth documenting this explicitly as a security-intentional design choice so operators know not to "improve" error verbosity.

**2. Behavioral trust complement to access policy.** WorkflowAccessPolicy answers "is this app ID permitted to call this workflow?" — a static, configuration-driven question. A complementary question is "has this app ID been reliably well-behaved when calling this workflow?" — a dynamic, behavioral question. An app ID that is permitted by policy but has been exhibiting unusual patterns (high failure rate, unexpected activity sequences) is a soft signal worth surfacing alongside the hard policy enforcement.

If the SPIFFE SVID from the workflow history signing feature (#9778) is available, the same identity can anchor behavioral trust lookups — SATP or AgentFolio can maintain per-SPIFFE-ID behavioral profiles that operators can use to adjust policy dynamically in response to behavioral degradation.

**3. IsCallerKnown for non-subject methods.** The "caller must appear in at least one allow rule" check for AddWorkflowEvent/PurgeWorkflowState is a good minimal enforcement point. Worth considering whether these methods should also support behavioral trust gating — a purge from an unknown-but-known-ID caller is different from a purge from an ID with recent anomalous activity.

tests/integration/suite/daprd/workflow/accesspolicy/allow.go
```diff
@@ -0,0 +1,205 @@
...
+	require.NoError(t, client0.StartWorkItemListener(ctx, registry0))
+	client1 := client.NewTaskHubGrpcClient(a.daprd1.GRPCConn(t, ctx), logger.New(t))
+	require.NoError(t, client1.StartWorkItemListener(ctx, registry1))
+
+	assert.EventuallyWithT(t, func(c *assert.CollectT) {
+		assert.GreaterOrEqual(c, len(a.daprd0.GetMetadata(t, ctx).ActorRuntime.ActiveActors), 1)
+		assert.GreaterOrEqual(c, len(a.daprd1.GetMetadata(t, ctx).ActorRuntime.ActiveActors), 1)
+	}, time.Second*20, time.Millisecond*10)
+
+	t.Run("allowed cross-app workflow with activity succeeds", func(t *testing.T) {
```
acroca: do we need this `t.Run`?

tests/integration/suite/daprd/workflow/accesspolicy/crossnamespace.go
```diff
@@ -0,0 +1,100 @@
...
+)
+
+func init() {
+	suite.Register(new(crossnamespace))
+}
+
+// crossnamespace tests that cross-namespace workflow calls are blocked.
+// This uses standalone mode (no mTLS) to test placement-level namespace
+// isolation. The workflow_acl.go checkNamespace function provides additional
+// defense-in-depth when mTLS is enabled (tested via unit tests).
```
acroca: I don't see how this is related with accesspolicy

tests/integration/suite/daprd/workflow/accesspolicy/deny.go
```diff
@@ -0,0 +1,225 @@
...
+	policyStore.Add(&wfaclapi.WorkflowAccessPolicy{
+		TypeMeta:   metav1.TypeMeta{APIVersion: "dapr.io/v1alpha1", Kind: "WorkflowAccessPolicy"},
+		ObjectMeta: metav1.ObjectMeta{Name: "deny-test", Namespace: "default"},
+		Scoped:     common.Scoped{Scopes: []string{"wfacl-target"}},
+		Spec: wfaclapi.WorkflowAccessPolicySpec{
+			DefaultAction: wfaclapi.PolicyActionDeny,
+			Rules: []wfaclapi.WorkflowAccessPolicyRule{{
+				Callers: []wfaclapi.WorkflowCaller{{AppID: "wfacl-caller"}},
+				Operations: []wfaclapi.WorkflowOperationRule{
+					{Type: wfaclapi.WorkflowOperationTypeWorkflow, Name: "AllowedWF", Action: wfaclapi.PolicyActionAllow},
```
acroca: Maybe for simplicity, since this test is about denying, should we change the default to allow and have only the deny rule?

tests/integration/suite/daprd/workflow/accesspolicy/featureflagoff.go
```diff
@@ -0,0 +1,182 @@
...
+	policyStore.Add(&wfaclapi.WorkflowAccessPolicy{
+		TypeMeta:   metav1.TypeMeta{APIVersion: "dapr.io/v1alpha1", Kind: "WorkflowAccessPolicy"},
+		ObjectMeta: metav1.ObjectMeta{Name: "flag-off-test", Namespace: "default"},
+		Scoped:     common.Scoped{},
+		Spec: wfaclapi.WorkflowAccessPolicySpec{
+			DefaultAction: wfaclapi.PolicyActionDeny,
+			Rules: []wfaclapi.WorkflowAccessPolicyRule{{
+				Callers: []wfaclapi.WorkflowCaller{{AppID: "legit-caller"}},
+				Operations: []wfaclapi.WorkflowOperationRule{
+					{Type: wfaclapi.WorkflowOperationTypeWorkflow, Name: "*", Action: wfaclapi.PolicyActionAllow},
```
acroca: Since the `legit-caller` is allowed to do any invocation, I think this test would also pass with the feature enabled, right?

tests/integration/suite/daprd/workflow/accesspolicy/nopolicy.go
```diff
@@ -0,0 +1,104 @@
...
+			daprd.WithConfigManifests(t, configWithFeatureFlag()),
+		),
+		workflow.WithDaprdOptions(1,
+			daprd.WithAppID("nopol-target"),
+			daprd.WithConfigManifests(t, configWithFeatureFlag()),
+		),
+	)
+
+	return []framework.Option{
+		framework.WithProcesses(n.wf),
```
acroca: I think this test should use a similar setup as all the other tests, with the only difference of not having the policy resource, just to make sure it's not something else that's making the tests pass.

acroca (commented):
Reviewed tests for now.

tests/integration/suite/daprd/workflow/accesspolicy/nopolicy.go
JoshVanL: 👍

pkg/acl/workflow/extract.go
```diff
@@ -0,0 +1,111 @@
...
+	default:
+		return "", false
+	}
+}
+
+// ExtractOperationName extracts the workflow or activity name from the request
+// method and payload. Returns the name and true if extraction succeeded, or
+// empty string and false if the method is not subject to access control
+// (e.g. AddWorkflowEvent, PurgeWorkflowState).
+func ExtractOperationName(opType OperationType, method string, data []byte) (string, bool, error) {
```
acroca, line 65: I find we could just return `(string, error)`, making sure all the 'false' cases return an error...

pkg/actors/targets/workflow/orchestrator/activity.go
```diff
@@ -91,15 +94,48 @@ func (o *orchestrator) callActivity(ctx context.Context, e *backend.HistoryEvent
...
 		return fmt.Errorf("failed to invoke activity actor '%s' to execute '%s': %w", targetActorID, ts.GetName(), err)
 	}
 
 	return nil
 }
 
+// failActivity creates a TaskFailed event on the parent orchestrator when
+// the activity call is permanently rejected (e.g. by a WorkflowAccessPolicy).
+// Uses a reminder to deliver the event in a fresh execution cycle.
+func (o *orchestrator) failActivity(ctx context.Context, e *backend.HistoryEvent, callErr error) error {
```
acroca: I'd rename this function to be more specific about failing due to access policy, like `failActivityACL`?
Also, `callErr` is not used.

pkg/actors/targets/workflow/orchestrator/childwf.go
```diff
@@ -75,9 +79,74 @@ func (o *orchestrator) callChildWorkflows(ctx context.Context, startEventName st
...
+}
+
+// failChildWorkflow creates a ChildWorkflowInstanceFailed event on the parent
+// orchestrator when the child workflow call is permanently rejected (e.g. by a
+// WorkflowAccessPolicy). It uses a reminder-based approach to deliver the
+// failure event in a fresh execution cycle, avoiding conflicts with the current
+// run loop's ClearInbox/saveInternalState calls.
+// taskScheduledID is the correlation ID that the parent orchestrator engine
+// uses to match this failure with the original sub-orchestration request.
+func (o *orchestrator) failChildWorkflow(ctx context.Context, taskScheduledID int32, callErr error) error {
```
acroca: Same as before, the function name could be more specific, and `callErr` in this case is only used for logging.

pkg/runtime/runtime.go
```diff
@@ -1385,6 +1424,186 @@ func (a *DaprRuntime) loadHTTPEndpoints(ctx context.Context) error {
...
+}
+
+// warnIfPoliciesExistWithoutFeatureFlag checks whether WorkflowAccessPolicy
+// resources exist even though the feature flag is disabled. This helps catch
+// misconfigurations where an operator creates policies but forgets to enable
+// the WorkflowAccessPolicy feature flag.
+func (a *DaprRuntime) warnIfPoliciesExistWithoutFeatureFlag(ctx context.Context) {
+	switch a.runtimeConfig.mode {
+	case modes.KubernetesMode:
+		resp, err := a.operatorClient.ListWorkflowAccessPolicy(ctx, &operatorv1pb.ListWorkflowAccessPolicyRequest{
```
javier-aliaga: Should we also warn if there is an error? In standalone I can see we are also ignoring the error

JoshVanL: Done :+1:

pkg/runtime/runtime.go
```diff
@@ -1396,6 +1435,194 @@ func (a *DaprRuntime) loadHTTPEndpoints(ctx context.Context) error {
 	return nil
 }
 
+func (a *DaprRuntime) loadWorkflowAccessPolicies(ctx context.Context) error {
```
acroca: This file is already huge. Should we move this new stuff to a different file to not make the problem worse? :)

pkg/acl/workflow/extract.go
JoshVanL: The bool is used to signal whether the operation was listed or not. If we collapsed this into the error, we would have to add error type checks.

pkg/acl/workflow/policy.go
```diff
@@ -0,0 +1,269 @@
...
+			if len(cr.operations) > 0 {
+				rules = append(rules, cr)
+			}
+		}
+	}
+
+	// Build the set of callers that have at least one allow action.
+	// IsCallerKnown uses this to avoid granting access to callers that
+	// only appear in deny rules.
+	allowed := make(map[string]struct{})
```
cicoyle: Hey, nice work on this PR 🎉 

Plz double check my thinking here - I think there is still an exploit path we still need to cover. `IsCallerKnown` allows partial auth callers to tamper with denied wfs by directly invoking actor methods. 

In my tracing I see:
`callActorValidateWorkflowACL` calls `EnforceRequest` in `workflow_acl.go`. `EnforceRequest` calls `ParseActorType`, which returns true for both wf and activity actors in the var called `isWorkflowActor` which, we might want to rename -> maybe to `isWFSystemActor` or simply `wfActors` to be clearer its for both the wf and activity actors. Then `EnforceRequest` calls `ExtractOperationName` which returns true for the `CreateWorkflowInstance` and `Execute` methods bc we have the `name` there. But then the other methods fall thru and call `IsCallerKnown` which is described in this comment:

```
// IsCallerKnown checks whether the given caller appears in at least one rule
// that contains an allow action. Used for methods where we can't extract a
// specific operation name (e.g., AddWorkflowEvent, PurgeWorkflowState).
// Callers that only appear in deny rules are not considered "known" to prevent
// a deny-only entry from granting broader access than being absent entirely.
```

So, with a policy like the following:
```
- callers:
  - appID: wfacl-caller
  operations:
  - type: workflow
    name: AllowedWorkflow
    action: allow
  - type: workflow
    name: DeniedWorkflow
    action: deny

```

`wfacl-caller` is correctly denied from scheduling `DeniedWorkflow` via the `Evaluate` path. But `IsCallerKnown("wfacl-caller")` returns true bc of the allow wf rule. So, `wfacl-caller` can send a direct `CallActor` with the proper wf methods and target a running denied wf instances actor ID.

Can we add a test for this case?

cicoyle (changes requested):

pkg/acl/workflow/policy.go
JoshVanL: Thanks @cicoyle - this was a real gap.

I have tightened `IsCallerKnown` to require name="*" allow with no overriding deny for the same op type (per-opType now), since non-subject methods have no instance name to evaluate against and any deny rule means the caller's trust is conditional. The legitimate cross-app activity-callback case (caller has allow activity=*, no denies) keeps working- the partial-auth case here is now denied.

https://github.com/dapr/dapr/pull/9790/commits/d90d6cf74bfb7113c716a3b95f172f6c39929161

cicoyle (approved):

PR #9795, opened 2026-04-14
[1.16] use contrib patch 1.16.12

cicoyle:
use latest 1.16 patch from contrib


JoshVanL:
@cicoyle we also need to update deps

JoshVanL (approved):

PR #9799, opened 2026-04-15
Tests: flakes

JoshVanL:
Updates more tests to ensure less flakes. Updates port so that we have a _much_ less chance of port collision.d-to-end tests passing.


pkg/ports/ports_test.go
```diff
@@ -43,7 +43,7 @@ func TestGetStablePort(t *testing.T) {
 
 	t.Run("invoking twice should return the same port", func(t *testing.T) {
 		port1 := getPort(t, "myapp")
-		assert.True(t, (port1 >= 10233 && port1 <= 12280) || (port1 >= 22444 && port1 <= 24491))
+		assert.True(t, port1 > 0 && port1 <= 65535)
```
acroca: We should never use ports <1024

JoshVanL: done :+1:

acroca (approved):

cicoyle (approved):

PR #9810, opened 2026-04-21
wf history ctx propagation

cicoyle:
wf history ctx propagation


cicoyle:
This is reviewable - just need to add the tests and look at re-run logic and what I need to change there.

pkg/actors/targets/workflow/activity/invoke.go
```diff
@@ -56,31 +57,68 @@ func (a *activity) handleInvoke(ctx context.Context, req *internalsv1pb.Internal
...
+
+	var legacy backend.HistoryEvent
+	if err := proto.Unmarshal(data, &legacy); err != nil {
+		return nil, nil, err
 	}
 
-	var activityName *string
-	if ts := his.GetTaskScheduled(); ts != nil {
+	return &protos.ActivityInvocation{HistoryEvent: &legacy}, taskScheduledName(&legacy), nil
+}
```
JoshVanL, line 93: Been thinking whether it would be possible to not need to do this optimistic decoding on a new proto type, but I think we have to...
Can we add a TODO to remove the legacy block in a future version (1.19).

tests/integration/suite/daprd/workflow/propagation/allfields.go
```diff
@@ -0,0 +1,249 @@
...
+	"github.com/dapr/durabletask-go/api"
+	"github.com/dapr/durabletask-go/api/protos"
+	"github.com/dapr/durabletask-go/task"
+)
+
+func init() {
+	suite.Register(new(allfields))
+}
+
+type allfields struct {
```
acroca, line 37: It'd be good to include a comment explaining what this test covers

pkg/actors/targets/workflow/activity/invoke.go
```diff
@@ -56,31 +57,68 @@ func (a *activity) handleInvoke(ctx context.Context, req *internalsv1pb.Internal
...
+		invocation.HistoryEvent = &legacy
+	}
+
+	his := invocation.GetHistoryEvent()
+	if his == nil {
+		return errors.New("activity reminder missing history event")
 	}
 
-	err := a.executeActivity(ctx, reminder.Name, &state)
+	err := a.executeActivity(ctx, reminder.Name, his, invocation.GetPropagatedHistory())
```
acroca: We could be passing the `ActivityInvocation` to the executeActivity now, it contains both `his` and the propagated history.

pkg/actors/targets/workflow/orchestrator/rerun.go
```diff
@@ -167,7 +167,7 @@ func (o *orchestrator) rerunWorkflowInstanceRequest(ctx context.Context, request
 	startedEvent := o.getExecutionStartedEvent(newState)
 	if err = errors.Join(
 		o.callChildWorkflows(ctx, startedEvent.GetName(), childWFs),
-		o.callActivities(ctx, activities, newState).err,
+		o.callActivities(ctx, activities, newState, nil).err,
```
acroca: Does this mean a rerun doesn't have any history propagation capabilities?

pkg/actors/targets/workflow/orchestrator/run.go
```diff
@@ -98,9 +98,10 @@ func (o *orchestrator) runWorkflow(ctx context.Context, reminder *actorapi.Remin
 		NewEvents:  state.Inbox,
 		RetryCount: -1, // TODO
 		State:      rs,
-		Properties: make(map[string]any, 1),
+		Properties: make(map[string]any, 2),
```
acroca: hm I don't think we're adding a new property in this change, right?

pkg/runtime/wfengine/wfengine.go
```diff
@@ -166,7 +166,11 @@ func New(opts Options) (Interface, error) {
 		Backend:  abackend,
 		Executor: executor,
 		Logger:   wfBackendLogger,
-		AppID:    opts.AppID,
+		AppID: opts.AppID,
```
acroca: I think this needs `go fmt`

tests/integration/suite/daprd/workflow/propagation/nomtls.go
```diff
@@ -0,0 +1,100 @@
...
+
+func init() {
+	suite.Register(new(nomtls))
+}
+
+// nomtls tests that history propagation is blocked when mTLS is disabled.
+// Even though the workflow configures WithHistoryPropagation(PropagateLineage()),
+// the child should receive nil propagated history because w/o mTLS the
+// runtime cannot sign/verify the history chain.
+type nomtls struct {
```
acroca, line 42: strictly speaking, history propagation requires signing and not mtls. But signing is enabled through MTLS. Maybe it's worth being clear about this?

cicoyle: I think I had this test before and forgot to update it when I merged in latest and updated to include signing - will update

pkg/actors/targets/workflow/activity/invoke.go
```diff
@@ -56,31 +57,68 @@ func (a *activity) handleInvoke(ctx context.Context, req *internalsv1pb.Internal
...
+// empty.
+func decodeActivityInvocation(data []byte) (*protos.ActivityInvocation, *string, error) {
+	var invocation protos.ActivityInvocation
+	if err := proto.Unmarshal(data, &invocation); err == nil && invocation.GetHistoryEvent() != nil {
+		return &invocation, taskScheduledName(invocation.GetHistoryEvent()), nil
+	}
+
+	var legacy backend.HistoryEvent
+	if err := proto.Unmarshal(data, &legacy); err != nil {
+		return nil, nil, err
```
JoshVanL: Should we also include the error from the parsing of `ActivityInvocation`-- or at least mention that both attempts failed?

JoshVanL (changes requested):

tests/integration/suite/daprd/workflow/propagation/allfields.go
cicoyle: I thought allfields was descriptive enough? It tests all propagation fields

pkg/actors/targets/workflow/orchestrator/rerun.go
cicoyle: should be done now.

pkg/actors/targets/workflow/activity/invoke.go
cicoyle: should be addressed in the latest commit(s)

pkg/actors/targets/workflow/activity/invoke.go
cicoyle: added the todo - should be addressed in the latest commit(s)

tests/integration/suite/daprd/workflow/propagation/nomtls.go
cicoyle: should be addressed in the latest commit(s)

pkg/runtime/wfengine/state/state.go
```diff
@@ -655,6 +695,15 @@ func LoadWorkflowState(ctx context.Context, state state.Interface, actorID strin
 		}
 	}
 
+	if len(bulkRes[propagatedHistoryKey]) > 0 {
+		var ph protos.PropagatedHistory
+		if err = proto.Unmarshal(bulkRes[propagatedHistoryKey], &ph); err != nil {
+			wfLogger.Warnf("Failed to unmarshal propagated history for '%s': %v", actorID, err)
```
JoshVanL: This should be a returned error.

JoshVanL (changes requested):

JoshVanL (approved):

PR #9812, opened 2026-04-21
Actors: app-initiated gRPC stream for actor callbacks

### LINKED ISSUE #927: Implement actor APIs for gRPC
Today, the actor functionality (invoking actors, reminders, timers and receiving callbacks from the Dapr runtime) is only possible using HTTP.

We should expose this functionality using gRPC so that apps can call Dapr Actors using gRPC, and allow for language specific SDKs to implement a gRPC server for callbacks.

acroca:
closes #927

Delivers actor app callbacks — invoke, reminder, timer, deactivate — over a single **app-initiated bidirectional gRPC stream** on the Dapr service. Apps that host actors no longer need to expose an HTTP or gRPC server port: they dial daprd, open the stream, advertise their registered actor types, and exchange request/response pairs correlated by id.

Aligns actor callbacks with the pattern Dapr already uses for pubsub (`SubscribeTopicEventsAlpha1`), config watch, and scheduler job streams — one way to receive callbacks across every Dapr app-callback API.


tests/integration/framework/process/grpc/app/app.go
```diff
@@ -71,15 +82,230 @@ func New(t *testing.T, fopts ...Option) *App {
...
 	a.runOnce.Do(func() {
 		a.grpc.Run(t, ctx)
+
+		// Spawn the actor callback stream dialer if the test wired a daprd
+		// address. The stream runs in the background until Cleanup.
+		if a.opts.daprdGRPCAddrFn != nil {
+			a.streamCtx, a.streamCancel = context.WithCancel(context.Background())
+			a.streamWG.Go(func() {
+				a.runActorCallbackStream(t)
+			})
```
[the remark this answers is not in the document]
acroca: Wrong. It was added in Go 1.25. This repo is on Go 1.26.2

dapr/proto/runtime/v1/actors.proto
```diff
@@ -166,3 +166,185 @@ message UnregisterActorRemindersByTypeRequest {
...
+  repeated string entities = 1;
+
+  // actor_idle_timeout is the default idle timeout for all actor types as
+  // a duration string (e.g. "1h"). Empty means use Dapr's default.
+  string actor_idle_timeout = 2;
+
+  // drain_ongoing_call_timeout is the duration Dapr waits for ongoing
+  // invocations to finish when an actor is rebalanced (e.g. "30s"). Empty
+  // means use Dapr's default.
+  string drain_ongoing_call_timeout = 3;
```
JoshVanL: Make optional, and use `Duration` proto type.

dapr/proto/runtime/v1/actors.proto
```diff
@@ -166,3 +166,185 @@ message UnregisterActorRemindersByTypeRequest {
...
+
+  // drain_rebalanced_actors, when true, instructs Dapr to drain invocations
+  // from a rebalanced actor rather than aborting them.
+  optional bool drain_rebalanced_actors = 4;
+
+  // reentrancy is the default reentrancy configuration applied to all
+  // actor types.
+  ActorReentrancyConfig reentrancy = 5;
```
JoshVanL:
```suggestion
  // reentrancy is the default reentrancy configuration applied to all
  // actor types.
  optional ActorReentrancyConfig reentrancy = 5;
```

pkg/actors/callbackstream/callbackstream.go
```diff
@@ -0,0 +1,283 @@
...
+var ErrNoConnection = errors.New("actor host is not connected")
+
+// ErrDisconnected is returned when an in-flight callback fails because the
+// app closed the stream.
+var ErrDisconnected = errors.New("actor host disconnected before responding")
+
+// Manager holds the set of active actor callback streams. It is safe for
+// concurrent use by the gRPC server (which registers streams) and the
+// actor transport (which sends callbacks).
+type Manager struct {
```
JoshVanL, line 76: It looks like we should be using `kit/events/loop` for this (?) This would make it more consistent with other parts of the project, remove the need for locks, and we know works well in other queue like systems.

JoshVanL (changes requested):

pkg/api/grpc/actorcallbacks.go
```diff
@@ -0,0 +1,201 @@
...
+			InitialResponse: &runtimev1pb.SubscribeActorEventsResponseInitialAlpha1{},
+		},
+	}); err != nil {
+		return err
+	}
+
+	errCh := make(chan error, 2)
+
+	// Recv loop: route every inbound response to the pending caller.
+	go func() {
```
cicoyle, line 119: Can this go routine and the one below use the runner manager - we should be able to match the scheduler `streamer.go` logic here. The current logic leaks recv go routines if the send loop returns first, itll sit in the `stream.Recv` until gRPC tears down. For consistency, we could align with the streamer logic using a runner manager with shared ctx so cleanup is explicit.

pkg/api/grpc/actorcallbacks.go
```diff
@@ -0,0 +1,201 @@
...
+		DrainRebalancedActors:   cfg.DrainRebalancedActors,
+		DrainOngoingCallTimeout: cfg.DrainOngoingCallTimeout,
+		HostedActorTypes:        cfg.Entities,
+		DefaultIdleTimeout:      cfg.ActorIdleTimeout,
+		Reentrancy:              cfg.Reentrancy,
+		AppChannel:              a.channels.AppChannel(),
+	}); err != nil {
+		return status.Errorf(codes.Internal, "failed to register hosted actors: %v", err)
+	}
+	defer a.Actors().UnRegisterHosted(cfg.Entities...)
```
cicoyle: I think we want the `UnRegisterHosted` and `mgr.Close` in a single defer block or flip the order
```suggestion
	defer func() {
		a.Actors().UnRegisterHosted(cfg.Entities...)
		mgr.Close(conn, nil)
	}
```

pkg/actors/callbackstream/callbackstream.go
```diff
@@ -0,0 +1,363 @@
...
+// Send issues a callback request over an available connection and blocks
+// until the correlated response arrives, ctx is cancelled, or the
+// connection closes. The caller supplies a builder that stamps the
+// request id onto the concrete oneof payload — this keeps id generation
+// under the Connection's control while leaving payload shape to callers.
+func (m *Manager) Send(
+	ctx context.Context,
+	build func(id string) *runtimev1pb.SubscribeActorEventsResponseAlpha1,
+) (*runtimev1pb.SubscribeActorEventsRequestAlpha1, error) {
+	conn := m.latest.Load().latestConn
```
cicoyle, line 269: The doc at the top of the file says:
`multiple concurrent connections from the same app are permitted`, but `publishSnapshot` always picks `m.conns[len(m.conns)-1]` and `Send` only reads `latest.latestConn`. This means earlier conns can still receive responses via `Deliver`, but they never receive a new `Send` once a  2nd conn registers and all new traffic moves to the newest stream.

During rolling restarts like if pod A has in flight invocations and pod B opens a new stream, A finishes its in flight work, but every new callback goes to B. Is this the intended behavior?

Can we add a test for this?

pkg/actors/targets/app/transport/grpc/grpc.go
```diff
@@ -0,0 +1,278 @@
...
+func buildInternalResponse(resp *runtimev1pb.SubscribeActorEventsRequestInvokeResponseAlpha1) *internalv1pb.InternalInvokeResponse {
+	headers := expandSingleValue(resp.GetMetadata())
+	if resp.GetError() {
+		headers[errorResponseHeader] = &internalv1pb.ListStringValue{Values: []string{"true"}}
+	}
+	return &internalv1pb.InternalInvokeResponse{
+		Status:  &internalv1pb.Status{Code: 200},
+		Headers: headers,
+		Message: &commonv1pb.InvokeResponse{
+			ContentType: resp.GetMetadata()["content-type"],
```
cicoyle: Should this be "Content-Type"?

pkg/actors/targets/app/transport/grpc/grpc.go
```diff
@@ -0,0 +1,278 @@
...
+
+	"google.golang.org/protobuf/types/known/anypb"
+)
+
+// errorResponseHeader is the HTTP-protocol marker that downstream code
+// (notably pkg/actors/router/router.go) inspects to decide whether a
+// response represents an application-level actor error. The streaming
+// transport synthesizes it on responses when error=true so router-side
+// code keeps working unchanged.
+const errorResponseHeader = "X-Daprerrorresponseheader"
```
cicoyle: Is this an exported const declared somewhere else we can reuse? I think the HTTP transport has the same constant, so would be good to pull to a shared package so the two transports can't drift.

cicoyle (changes requested):

pkg/actors/targets/app/app.go
```diff
@@ -59,160 +51,73 @@ type app struct {
...
 func (a *app) InvokeReminder(ctx context.Context, reminder *api.Reminder) error {
-	invokeMethod, err := method.NormalizeMethod("remind/" + reminder.Name)
-	if err != nil {
+	// Reject names that could alter the synthetic path the HTTP transport
+	// builds ("remind/<name>" → "actors/{type}/{id}/method/remind/<name>").
+	// Guards against legacy persisted reminder names containing "/", "..",
+	// or control characters.
+	if err := method.ValidateName(reminder.Name); err != nil {
 		return fmt.Errorf("invalid reminder name: %w", err)
 	}
```
JoshVanL: Why? I think '/' were supported in reminder/timer names (?)

pkg/api/grpc/actorcallbacks.go
```diff
@@ -0,0 +1,248 @@
...
+		DrainOngoingCallTimeout: cfg.DrainOngoingCallTimeout,
+		HostedActorTypes:        cfg.Entities,
+		DefaultIdleTimeout:      cfg.ActorIdleTimeout,
+		Reentrancy:              cfg.Reentrancy,
+		AppChannel:              a.channels.AppChannel(),
+	}); err != nil {
+		return status.Errorf(codes.Internal, "failed to register hosted actors: %v", err)
+	}
+	conn := mgr.Register(stream.Context(), cfg)
+	// Cleanup runs as one deferred block so ordering is explicit: drop the
```
JoshVanL, line 74: Do we have a test whereby we have actors hosted on the HTTP config, then call `SubscribeActorEventsAlpha1`? What happens? We are calling `RegisteredHosted` twice.

pkg/api/grpc/actorcallbacks.go
```diff
@@ -0,0 +1,248 @@
...
+	}
+	results := make(chan recvResult, 1)
+	go func() {
+		defer close(results)
+		for {
+			msg, err := stream.Recv()
+			results <- recvResult{msg, err}
+			if err != nil {
+				return
+			}
```
JoshVanL: Isn't checking ctx, seems like a go routine leak.

JoshVanL (changes requested):

pkg/api/grpc/actorcallbacks.go
acroca: Added a test to cover this, but it was already handled by returning a PreconditionFailed error

JoshVanL (approved):

cicoyle (approved):

jeremylcarter:
This is exciting! Well done everyone

Issue #5915, opened 2023-02-09
Flaky unit tests


artursouza:
We have noticed significant increase in flakiness in our unit tests in 1.10 release. This issue intends to fix in 2 ways:
- Stabilization of existing unit tests; OR
- Rewrite unit tests as end-to-end tests when they are doing "too much"

mukundansundar:
This is currently still the case

PR #9813, opened 2026-04-22
test(operator): fix flaky TestDaprWatchdog_Start by removing sleep

Kunalbehbud:
Related to #5915

## Summary

`TestDaprWatchdog_Start` used timing-based synchronization by starting `dw.Start(ctx)` in a goroutine, sleeping for 1 second, and then cancelling the context. That makes the test dependent on scheduler timing and machine load, which can become flaky under CI and especially under race instrumentation.

This change makes the test deterministic and single-iteration based instead of wall-clock based. It sets `dw.interval` below `singleIterationDurationThreshold`, so `DaprWatchdog.Start` returns after the first successful iteration. With that in place, the test can call `dw.Start(ctx)` directly and remove the extra goroutine, `time.Sleep`, and manual cancellation path.

## Verification

Ran:

```bash
go test -race -count=50 -run TestDaprWatchdog_Start ./pkg/operator
```

Result: passed successfully.

### DIFF
--- a/pkg/operator/watchdog_test.go
+++ b/pkg/operator/watchdog_test.go
@@ -242,14 +242,7 @@ func Test_patchPodLabel(t *testing.T) {
 }
 
 func TestDaprWatchdog_Start(t *testing.T) {
-	// simple test of start
-	ctx, cancel := context.WithCancel(t.Context())
-	cancelled := false
-	defer func() {
-		if !cancelled {
-			cancel()
-		}
-	}()
+	ctx := t.Context()
 
 	singleIterationDurationThreshold = 100 * time.Millisecond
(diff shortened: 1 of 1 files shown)


Kunalbehbud:
Could a maintainer take a look when convenient? It's a small, test-only change (+3/-20) that makes `TestDaprWatchdog_Start` deterministic â€” it drops the `time.Sleep` + goroutine + manual cancel and instead sets `dw.interval` below `singleIterationDurationThreshold` so `Start` returns after the first iteration. Ran `go test -race -count=50 -run TestDaprWatchdog_Start ./pkg/operator` locally with no flake. Happy to rebase if the branch has drifted.

PR #9814, opened 2026-04-22
fix(runtime): align bulk HTTP pubsub empty status response with single delivery

### LINKED ISSUE #8737: Subscription Fallthru Inconsistency

Kunalbehbud:
Fixes #8737

Bulk HTTP pubsub delivery currently treats an app response with `status == ""` as RETRY even when the app returned a response entry for that message. Single HTTP delivery and gRPC flows both interpret an empty status as SUCCESS, so the HTTP bulk path is inconsistent and can trigger unnecessary retries.

This change aligns that contract by moving `case ""` onto the SUCCESS branch in `DeliverBulk`. Missing response entries are still handled as RETRY, so the behavior change is limited to explicit empty-status responses from the app.

A regression test covers the empty-status bulk response case, and unit tests passed:

- `go test -tags=unit,allcomponents ./pkg/runtime/subscription/...`

### DIFF
--- a/pkg/runtime/subscription/postman/http/http.go
+++ b/pkg/runtime/subscription/postman/http/http.go
@@ -295,9 +295,11 @@ func (h *http) DeliverBulk(ctx context.Context, req *postman.DeliverBulkRequest)
 		for _, response := range appBulkResponse.AppResponses {
 			if _, ok := (*bscData.EntryIdIndexMap)[response.EntryId]; ok {
 				switch response.Status {
-				case "":
-					// When statusCode 2xx, Consider empty status field OR not receiving status for an item as retry
-					fallthrough
+				case "", contribpubsub.Success:
+					// When statusCode is 2xx, consider empty status field as success.
+					bscData.BulkSubDiag.StatusWiseDiag[string(contribpubsub.Success)]++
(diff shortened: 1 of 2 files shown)


pkg/runtime/subscription/postman/http/http.go
```diff
@@ -296,19 +296,19 @@ func (h *http) DeliverBulk(ctx context.Context, req *postman.DeliverBulkRequest)
 			if _, ok := (*bscData.EntryIdIndexMap)[response.EntryId]; ok {
 				switch response.Status {
 				case "":
```
mikeee: Rather than the below case, perhaps combining it with this case might make this a bit easier to read than the current fallthrough. Wdyt?
```go
				case "", contribpubsub.Success:
```

Kunalbehbud:
Good call, that reads better. I combined them into `case "", contribpubsub.Success:` and dropped the `fallthrough`, so the empty and `Success` statuses land on the same branch. Behavior is unchanged â€” missing entries still fall through to RETRY â€” and the targeted tests (`go test -tags=unit,allcomponents ./pkg/runtime/subscription/...`) still pass.

PR #9819, opened 2026-04-22
Integration: fix flakes for stream dep jobs

JoshVanL:
Fix test flakes which rely on a stable stream to scheduler.

### DIFF
--- a/pkg/api/grpc/server.go
+++ b/pkg/api/grpc/server.go
@@ -253,8 +253,6 @@ func (s *server) StartNonBlocking() error {
 }
 
 func (s *server) Close() error {
-	defer s.wg.Wait()
-
 	s.htarget.NotReady()
 
 	if s.api != nil {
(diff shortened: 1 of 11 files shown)


tests/integration/suite/actors/reminders/failurepolicy/constant/interval.go
```diff
@@ -60,6 +60,7 @@ func (i *interval) Setup(t *testing.T) []framework.Option {
 
 func (i *interval) Run(t *testing.T, ctx context.Context) {
 	i.actors.WaitUntilRunning(t, ctx)
+	i.actors.Scheduler().WaitUntilSidecarsConnected(t, ctx, 3)
```
acroca, line 63: Isn't this something we want to do always when we use a scheduler in a test?

JoshVanL: @acroca not necessarily- I think it is useful to have a bit of natural chrun in other tests which absolutely don't need a stable stream to have some signal there.

cicoyle (dismissed):

pkg/api/grpc/server.go
```diff
@@ -263,14 +261,18 @@ func (s *server) Close() error {
 		}
 	}
 
-	s.wg.Add(len(s.servers))
+	var stopWg sync.WaitGroup
+	stopWg.Add(len(s.servers))
```
acroca, line 265: you can also use `stopWg.Go` instead, it'll simplify the code

acroca (approved):

PR #9822, opened 2026-04-23
add version to workflow metadata

### LINKED ISSUE #105: Add workflow version to OrchestrationMetadata
Processes using the `EventSink` for getting workflow status updates have no way to know what version the particular workflow instance is running. This information is already available to the orchestrator, but is not being propagated.

pablochacin:
# Description

Adds missing fields to WorkflowMetadata
- Adds the workflow version
- Adds the parent's appId to child work flows

Requires https://github.com/dapr/durabletask-go/pull/86 to be merged

## Issue reference

close: https://github.com/dapr/proposals/issues/105

## Checklist

- [x] Code compiles correctly
- [x] Created/updated tests
- [x] Unit tests passing


acroca:
I think it'd be good to have integration tests covering this change.

pablochacin:
@acroca Do you have an idea where to add them?  I couldn't find any place to add such test. This seems to be an area not covered.

acroca:
> @acroca Do you have an idea where to add them? I couldn't find any place to add such test. This seems to be an area not covered.

There's a test tests/integration/suite/daprd/workflow/fetch.go where you could add it, or add a new test in tests/integration/suite/daprd/workflow/get if you prefer to have a more custom test for this specific change.

pablochacin:
@acroca test added and they found what seems [a bug](https://github.com/dapr/dapr/issues/9901)? I implemented a workaround using the same logic used in durable tasks for getting the version.

acroca (dismissed):

pkg/runtime/wfengine/state/version.go
```diff
@@ -0,0 +1,39 @@
...
+func WorkflowVersion(events []*backend.HistoryEvent) *wrapperspb.StringValue {
+	for i := len(events) - 1; i >= 0; i-- {
+		ws := events[i].GetWorkflowStarted()
+		if ws == nil {
+			continue
+		}
+		if name := ws.GetVersion().GetName(); name != "" {
+			return wrapperspb.String(name)
+		}
+	}
```
JoshVanL, line 35: Should we not always return after we find `WorkflowStarted` ?

JoshVanL (changes requested):

famarting (approved):

pkg/runtime/wfengine/state/version.go
pablochacin: good catch. aligned with changes in  https://github.com/dapr/dapr/pull/9898

acroca (dismissed):

JoshVanL:
@pablochacin see file conflicts

tests/integration/suite/daprd/workflow/fetch.go
```diff
@@ -91,7 +91,8 @@ func (f *fetch) Run(t *testing.T, ctx context.Context) {
 		})
 
 		t.Run("http", func(t *testing.T) {
-			req, err := http.NewRequestWithContext(ctx,
+			req, err := http.NewRequestWithContext(
+				ctx,
```
cicoyle, line 95: Is this PR needed? I don't see a diff here, just looks like ctx was moved to the line below.

pablochacin: There were more changes in the initial commits.
Something happened when I resolved the merge conflicts. let me check what was it

pablochacin:
The changes in the original PR were incorporated into #9898.

PR #9823, opened 2026-04-23
Integration: Workflow: scheduler disk exhaustion tests

JoshVanL:
Covers four failure modes of the scheduler's embedded etcd and asserts that writes surface errors rather than being silently dropped:

  quota           - etcd space quota exceeded before workflow starts
  midrun          - etcd space quota exceeded while a workflow is running
  diskfull        - backing filesystem fills before workflow starts
  diskfullmidrun  - backing filesystem fills while a workflow is running


tests/integration/framework/os/unshare.go
```diff
@@ -0,0 +1,123 @@
...
+// outer test runner is NOT root.
+func InUnshareNamespace() bool {
+	return os.Geteuid() == 0
+}
+
+// MountTmpfs mounts a size-capped tmpfs at a fresh t.TempDir() and registers
+// umount via t.Cleanup. Requires InUnshareNamespace() to be true.
+func MountTmpfs(t *testing.T, sizeMiB int) string {
+	t.Helper()
+	mount := t.TempDir()
```
acroca, line 42: Should this add some precondition checks and skip otherwise? Like, would this work on Windows?

tests/integration/framework/process/scheduler/quota.go
```diff
@@ -0,0 +1,96 @@
...
+	"github.com/stretchr/testify/require"
+	clientv3 "go.etcd.io/etcd/client/v3"
+)
+
+// FillQuota writes 1MiB values under the given etcd key prefix until etcd
+// returns "database space exceeded". Requires the scheduler to have been
+// started with a low --etcd-space-quota (see WithEtcdSpaceQuota).
+func (s *Scheduler) FillQuota(t *testing.T, ctx context.Context, prefix string) {
+	t.Helper()
+	cli := s.ETCDClient(t, ctx)
```
acroca, line 36: Should we ensure somehow the etcd instance has a quota? If it doesn't, it might just write every 10ms for 30 seconds and fill the disk right?

tests/integration/suite/daprd/workflow/disk/midrun.go
```diff
@@ -0,0 +1,114 @@
...
+	"github.com/dapr/dapr/tests/integration/framework/process/scheduler"
+	"github.com/dapr/dapr/tests/integration/framework/process/workflow"
+	"github.com/dapr/dapr/tests/integration/suite"
+)
+
+func init() {
+	suite.Register(new(midrun))
+}
+
+type midrun struct {
```
acroca: shouldn't this test be `quotamidrun`?

acroca (approved):

PR #9825, opened 2026-04-24
Workflow: terminate tampered in-flight workflows

JoshVanL:
When the orchestrator actor loads a workflow whose persisted history, signatures, or inbox have been tampered with, the actor now appends a terminal ExecutionCompleted event marked with the well-known error type `DAPR_WORKFLOW_HISTORY_TAMPERED`. This stops the executor from acting on forged input and gives clients a stable signal to match on. The original (untrusted) state is left intact for forensics. Only an unsigned terminal event is appended, and LoadWorkflowState bypasses signature verification when it sees the tamper marker so subsequent loads succeed and surface FAILED.

The append is scoped to workflows still in custody of the executor. A workflow that had already finished is left untouched: there is nothing for the actor to do, and the verification error is surfaced to readers who are responsible for detecting the tampering at read time.

Configuration mismatches (signed history loaded by a non-signing host, or vice versa) are split out into a new ConfigurationError type so the tamper-recovery path does not run against intact-but-unloadable workflows.

### DIFF
--- a/pkg/actors/targets/workflow/orchestrator/state.go
+++ b/pkg/actors/targets/workflow/orchestrator/state.go
@@ -41,17 +41,21 @@ func (o *orchestrator) loadInternalState(ctx context.Context) (*wfenginestate.St
 		return o.state, o.ometa, nil
 	}
 
-	// state is not cached, so try to load it from the state store
-	state, err := wfenginestate.LoadWorkflowState(ctx, o.actorState, o.actorID, wfenginestate.Options{
+	opts := wfenginestate.Options{
 		AppID:             o.appID,
 		WorkflowActorType: o.actorType,
 		ActivityActorType: o.activityActorType,
 		Signer:            o.signer,
-	})
+	}
+
+	// state is not cached, so try to load it from the state store
+	state, err := wfenginestate.LoadWorkflowState(ctx, o.actorState, o.actorID, opts)
 	if err != nil {
 		var verifyErr *wferrors.VerificationError
 		if errors.As(err, &verifyErr) {
-			o.failSignatureVerification(ctx)
+			if !isTerminal(state) {
+				return o.tombstoneTamperedState(ctx, opts, state, err)
+			}
 		}
 		return nil, nil, err
 	}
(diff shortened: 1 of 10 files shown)


pkg/actors/targets/workflow/orchestrator/state.go
```diff
@@ -83,6 +87,38 @@ func (o *orchestrator) loadInternalState(ctx context.Context) (*wfenginestate.St
...
+// tombstoneTamperedState appends an unsigned ExecutionCompleted(FAILED)
+// tamper marker to the workflow's history (see [wfenginestate.MarkAsFailed])
+// so it surfaces as terminally FAILED on every subsequent load. The original
+// (untrusted) history, inbox, signatures, and certs are left intact for
+// forensics. The actor's reminders are deleted to stop further activations
+// from firing against the dead workflow.
+func (o *orchestrator) tombstoneTamperedState(ctx context.Context, opts wfenginestate.Options, prior *wfenginestate.State, cause error) (*wfenginestate.State, *backend.WorkflowMetadata, error) {
+	log.Warnf("Workflow actor '%s': tampering detected, marking workflow as FAILED: %s", o.actorID, cause)
+
+	failed, err := wfenginestate.MarkAsFailed(ctx, o.actorState, o.actorID, opts, prior, cause)
```
acroca: I think we should rename `MarkAsFailed` to something like `MarkAsTamperFailed` or similar, as the naming (and package) is very generic.

pkg/actors/targets/workflow/orchestrator/state.go
```diff
@@ -83,6 +87,38 @@ func (o *orchestrator) loadInternalState(ctx context.Context) (*wfenginestate.St
...
+
+	return failed, o.ometa, nil
+}
+
+// isTerminal reports whether the loaded workflow's history ends in an
+// ExecutionCompleted event of any kind (the runtime's terminal marker).
+// There is nothing for the orchestrator actor to do on a terminal workflow,
+// so the tamper-marker append is skipped — the reader path is responsible
+// for surfacing the verification error to clients.
+func isTerminal(s *wfenginestate.State) bool {
```
acroca: We have more terminal states like Failed, Terminated and Cancelled. I think this function should be `isCompleted` no?

pkg/actors/targets/workflow/orchestrator/state.go
JoshVanL: Done :+1:

pkg/actors/targets/workflow/orchestrator/state.go
JoshVanL: Changed to `isCompleted`.

cicoyle (approved):
lgtm

PR #9833, opened 2026-04-27
fix: preserve trailing slash in service invocation method path

### LINKED ISSUE #7686: Service invocation via request header removes trailing slash
Hi,

this week we tried to implement a Non-Dapr endpoint via HTTP. Normally we use the `dapr-app-id` in the header instead of the `/v1.0/invoke/{appId}/method/{methodName}`, but ran into an issue with one of our endpoints. For this endpoint we need to have a trailing slash, which doesn't arrive at our server. Strangely enough, the dapr logs show the trailing slash. Calling the endpoint with the `/v1.0/invoke/{appId}/method/{methodName}/` version works. So my guess is, that it has something to to with the `dapr-app-id` header version. (Maybe the log is not correct?)

## In what area(s)?

/area runtime

## What version of Dapr?

> 1.13.0

## Expected Behavior

URL should be send to the endpoint as specified.

## Actual Behavior

Trailing slash is missing. 

Dapr Kubernetes Sidecar
Calling the HTTPEndpoint via `/v1.0/invoke/{appId}/method/{methodName}/` (including trailing slash), the server returns 200.

Dapr Log
```
time="2024-04-10T13:47:41.873135982Z" level=info msg="HTTP API Called" app_id=xy code=200 duration=1159 instance=xy-6d5bb584b-s4trc method="GET /v1.0/invoke/appId/method/my-demo/route/2024-04-08T00:00:00.0000000/2024-04-08T23:59:59.9990000/" scope=dapr.runtime.http-info size=4994 type=log ver=1.13.0
```

Server Log
```
2024-04-10 13:47:41 10.21.1.10 GET /my-demo/route/2024-04-08T00:00:00.0000000/2024-04-08T23:59:59.9990000/ - 443 - 127.0.0.1 Go-http-client/2.0 - 200 0 0 1065 400 6610 800d
```

Calling the HTTPEndpoint via `/{methodName}/` and `dapr-app-id` header, the server returns 404 (as the trailing slash is missing)

Dapr Log
```
time="2024-04-10T13:50:03.933427538Z" level=info msg="HTTP API Called" app_id=xy code=404 duration=11 instance=xy-6d5bb584b-s4trc method="GET /my-demo/route/2024-04-08T00:00:00.0000000/2024-04-08T23:59:59.9990000/" scope=dapr.runtime.http-info size=5848 type=log ver=1.13.0
```

Server Log 
```
2024-04-10 13:50:03 10.21.1.10 GET /my-demo/route/2024-04-08T00:00:00.0000000/2024-04-08T23:59:59.9990000 - 443 - 127.0.0.1 Go-http-client/2.0 - 404 0 2 10 400 6610 800d
```

## Release Note

RELEASE NOTE:

mail2sudheerobbu-oss:
## Description

Fixes #7686

When a client invokes a service using the `dapr-app-id` request header (or Basic Auth or the `/v1.0/invoke/` URL form), any trailing slash in the request path was silently stripped before being forwarded to the target app. This caused routes like `GET /api/events/2024-04-08T00:00:00/` to arrive as `GET /api/events/2024-04-08T00:00:00` — resulting in 404s on apps that treat the trailing slash as significant.

## Root cause

Go's `path.Clean` strips trailing slashes by design. It was used at three independent points in the invocation pipeline:

1. `findTargetIDAndMethod` in `pkg/api/http/directmessaging.go` — extracts method from header / Basic Auth / URL
2. `NormalizeMethod` in `pkg/messaging/method/normalize.go` — normalizes on every dispatch
3. `constructRequest` in `pkg/channel/http/http_channel.go` — defense-in-depth clean before building the outbound URL

## Fix

Introduce a `cleanPath` helper in each package that calls `path.Clean` but re-attaches the trailing slash when the original path ended with one:

```go
func cleanPath(p string) string {
    if p == "" {
        return p
    }
    hasTrailing := p[len(p)-1] == '/'
    cleaned := path.Clean(p)
    if hasTrailing && cleaned != "/" {
        cleaned += "/"
    }
    return cleaned
}
```

All three call sites are updated to use `cleanPath` instead of `path.Clean`, so a trailing slash survives end-to-end regardless of which entry point is used. Path traversal sequences (`../`) are still resolved correctly.

## Tests

- `TestFindTargetIDAndMethod`: 3 new cases — dapr-app-id header, Basic Auth, and `/v1.0/invoke/` URL all preserve trailing slash
- `TestNormalizeMethod`: 3 new cases — trailing slash preserved, trailing slash preserved after traversal resolution, root-only `/` stays empty string


mail2sudheerobbu-oss:
Hi dapr team — gentle ping on this PR! It fixes a bug where trailing slashes in service invocation method paths are silently stripped (fixes #7686). The fix is minimal — a one-line path join change with a test. No conflicts. Would appreciate a review when you have a moment. Thanks!

mail2sudheerobbu-oss:
Hi team — gentle ping on this PR. It fixes a bug where service invocation incorrectly stripped trailing slashes from method paths (closes #7686). Includes unit tests covering the slash-preservation behavior. CI is green and the branch is synced with master. Would love a review when convenient! 🙏

mail2sudheerobbu-oss:
Hi dapr team - branch just rebased onto latest master (no conflicts). This fixes trailing slashes being silently stripped in service invocation method paths (closes #7686). Minimal change - adds a cleanPath helper that preserves trailing slashes in 3 places. No reviews yet - would appreciate a look when you get a chance!

pkg/messaging/method/normalize.go
```diff
@@ -39,6 +39,21 @@ func ValidateName(name string) error {
...
+// unchanged rather than having the slash silently stripped.
+func cleanPath(p string) string {
+	if p == "" {
+		return p
+	}
+	hasTrailing := p[len(p)-1] == '/'
+	cleaned := path.Clean(p)
+	if hasTrailing && cleaned != "/" {
+		cleaned += "/"
+	}
```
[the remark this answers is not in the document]
mail2sudheerobbu-oss: Fixed in commit 15b0908a — all 4 Copilot comments addressed:

1. **cleanPath dot-segment guard** (normalize.go, http_channel.go, directmessaging.go): added `&& cleaned != "." && cleaned != ".."` to the trailing-slash condition so inputs like `"./"` or `"../"` no longer have the slash re-appended after `path.Clean` resolves them to dot-segments.

2. **NormalizeMethod root-slash** (normalize.go): added `if cleaned == "/" { cleaned = "" }` after `cleanPath` so a root-only `"/"` method normalises to an empty string rather than being forwarded as-is.

mikeee (changes requested):
Please also add test(s) for `constructRequest` or construct a http server to show that the complete outbound http channel path query string is preserved.

mail2sudheerobbu-oss:
Thanks @mikeee — added TestConstructRequestTrailingSlash in 809019c. It covers both angles you mentioned:

Direct constructRequest — asserts the built request has URL.Path == "/api/events/2024-04-08T00:00:00/" (trailing slash intact) and URL.RawQuery == "start=1&end=2".
End-to-end via an httptest server — runs a real request through InvokeMethod and asserts the receiving handler sees both the trailing-slash path and the query string preserved.

go test ./pkg/channel/http/ passes locally. Could you take another look when you have a moment? 🙏

mail2sudheerobbu-oss:
@mikeee — added TestConstructRequestTrailingSlash in 809019c, covering both cases you asked for: a direct constructRequest assertion (URL.Path == "/api/events/2024-04-08T00:00:00/", RawQuery == "start=1&end=2") and an end-to-end httptest server check that the receiving handler sees the trailing slash and query preserved. go test ./pkg/channel/http/ passes. Could you re-review when you have a moment? 🙏

pkg/messaging/method/normalize.go
```diff
@@ -39,6 +39,21 @@ func ValidateName(name string) error {
 	return nil
 }
 
+// cleanPath cleans a URL path like path.Clean but preserves a single trailing
+// slash, so that method names ending with "/" are forwarded to the target app
+// unchanged rather than having the slash silently stripped.
+func cleanPath(p string) string {
```
wasniksudesh:
```suggestion
func CleanPath(p string) string {
```
how about importing this in directmessaging and http_channel instead of copy-pasting ? 
(duplicate code might diverge)

pkg/messaging/method/normalize.go
```diff
@@ -39,6 +39,21 @@ func ValidateName(name string) error {
...
+	if p == "" {
+		return p
+	}
+	hasTrailing := p[len(p)-1] == '/'
+	cleaned := path.Clean(p)
+	if hasTrailing && cleaned != "/" && cleaned != "." && cleaned != ".." {
```
wasniksudesh, line 51: nit - UTs to test when cleaned might be "." and ".." - since it is handled specially.

wasniksudesh (commented):
Hello, minor comments. 👍

pkg/messaging/method/normalize.go
mail2sudheerobbu-oss: Done — deduped the three cleanPath copies into an exported method.CleanPath (imported into directmessaging.go and http_channel.go as invokemethod to avoid colliding with the local method variables), and added TestCleanPath covering the ./.. dot-segment cases. Thanks for the review!

mail2sudheerobbu-oss:
@wasniksudesh @mikeee — addressed both points in 47685b7: cleanPath is now an exported method.CleanPath reused by directmessaging.go and http_channel.go (no more copy-paste), and TestCleanPath covers the ./.. dot-segment cases. Could you re-review when you get a chance? 🙏

wasniksudesh:
thank you - looks all good - no more comments from my end 🚀

mail2sudheerobbu-oss:
@mikeee — the constructRequest/httptest coverage you asked for is in 809019c, and wasniksudesh's dedup comments are addressed. Could you re-review when you have a moment? 🙏

mail2sudheerobbu-oss:
Update: resolved the merge conflict in http_channel_test.go (a clean two-way addition — my TestConstructRequestTrailingSlash landed at the same spot as master's TestConcurrencyLimiterContext, both now coexist) and rebased on master. @mikeee the httptest coverage you asked for is in 809019c and @wasniksudesh's dedup comments are addressed in 47685b7 - both reviewers' feedback should be fully resolved now. CI is green. Would appreciate a re-review when you have a chance.

cicoyle (changes requested):
Hi,

Thanks for keeping up with this PR 🙏🏻 

I recognize that there are several duplicates trying to fix this issue. This is the furthest along and earliest PR, so I'd like to consolidate into this one.

Before being able to merge this, I would like to see integration tests added testing this inside of `tests/integration/suite/daprd/serviceinvocation/` for both `grpc` & `http` - equal test coverage.
- Path correctness (`http/reservedchars/pathcorrectness.go`, & the gRPC equivalent):
  - /foo/bar/ reaches the app as /foo/bar/ via the dapr-app-id header
  - /v1.0/invoke/<app>/method/foo/bar/, and Basic auth
  - /foo/bar without a slash still arrives without one
  - /admin/../public/ -> /public/, and admin%2F..%2Fpublic%2F -> /public/
  - //foo//bar// -> /foo/bar/
  - root: /v1.0/invoke/<app>/method/ and the header form with /
- gRPC API -> gRPC app (`grpc/reservedchars/pathcorrectness.go`): the existing `trailing slash cleaned` case from #9805 asserts `public/` -> `public`. Please update it to assert `public/` is preserved, and add `foo/bar/` -> `foo/bar/` (gRPC apps receive the method as-is, no leading `/`)
- gRPC API -> HTTP app: `InvokeService` with method `foo/bar/` delivers `/foo/bar/` to the app
- Two daprds: caller daprd -> callee daprd -> app keeps the slash, since the callee normalizes again in callLocalValidateACL
- HTTPEndpoint: the original #7686 scenario, where invoking a non-Dapr endpoint with a trailing slash delivers the slash (http/httpendpoints.go)
- ACL, for both HTTP and gRPC (reservedchars/aclbypass.go):
  - default allow with deny on /admin: /admin/ is denied
  - deny on /admin/*: admin/ is denied
  - default deny with allow on /public: public/ is allowed, consistent with how /public rules match today
  - traversal + trailing slash, over both HTTP and raw gRPC - the ACL is evaluated against the same normalized path the app receives::
    - admin/../public/ -> public/
    - admin/../../public/ -> public/
    - ../admin/ -> admin/ (so a deny on /admin must still apply)
    - admin/../ -> empty

Actors: the Actors API also uses NormalizeMethod, so with this change actor methods with a trailing slash also get forwarded with the slash. Please add integration tests for actor method invocation with and without a trailing slash over both the HTTP and gRPC actor APIs, and note this behavior change in the PR description.

On the [advisory](https://github.com/advisories/GHSA-85gx-3qv6-4463): the [fix](https://github.com/dapr/dapr/pull/9805) stripped the trailing slash as a side effect of using `path.Clean`. I think it's fine to add it back when one was sent, as long as it's re-appended after `path.Clean` so path traversal is still resolved, and the ACL and the app keep seeing the same normalized method.

mail2sudheerobbu-oss:
Thanks for the detailed breakdown — added integration test coverage for every scenario listed, in `tests/integration/suite/daprd/serviceinvocation/` and `tests/integration/suite/actors/trailingslash/` (commits `d78992c`, `d818c3e`):

**Path correctness (HTTP + gRPC, `reservedchars/pathcorrectness.go`)**: trailing slash via `dapr-app-id` header, `/v1.0/invoke/` URL form, and Basic Auth; no-slash paths stay without one; traversal (`/admin/../public/`, encoded `admin%2F..%2Fpublic%2F`) resolves and preserves the slash; duplicate slashes (`//foo//bar//` → `/foo/bar/`) collapse correctly; root-only cases (`/v1.0/invoke//method/` and the header form with `/`) normalize as expected.

**gRPC API → gRPC app**: updated the pre-existing "trailing slash cleaned" case to assert `public/` is now preserved (not stripped), and added `foo/bar/` → `foo/bar/` (confirmed gRPC apps receive the method as-is, no leading `/`).

**gRPC API → HTTP app** (new `pathcorrectnesshttpapp.go`): `InvokeService` with method `foo/bar/` delivers `/foo/bar/` to the HTTP app.

**HTTPEndpoint** (`http/httpendpoints.go`): added the original #7686 scenario directly — invoking a non-Dapr HTTP endpoint with a trailing slash delivers the slash.

**ACL, HTTP + gRPC** (`reservedchars/aclbypass.go` and new `acladmindeny.go`): extended the existing default-deny/allow-`/public` policy with trailing-slash and traversal+trailing-slash cases (`admin/../public/`, `admin/../../public/`, `../admin/`, `admin/../`), and added a new default-allow/deny-`/admin` policy (exact + wildcard) to cover the deny-side assertions you listed. All confirmed the callee's deny-path handler was never invoked.

**Actors**: confirmed via code that `pkg/actors/targets/app/transport/http/http.go` calls `method.NormalizeMethod` before building the outbound callback URL, so actor invocation gets the same trailing-slash preservation as service invocation. Added `tests/integration/suite/actors/trailingslash/` covering actor method invocation with and without a trailing slash via the gRPC actor API (`InvokeActor`), confirming the app receives the slash intact.

**Behavior change note**: this fix changes actor invocation behavior — previously an actor method invoked with a trailing slash (e.g. `method/foo/`) would arrive at the actor app as `foo`; after this fix it arrives as `foo/`, consistent with regular service invocation. I've also added this note to the PR description since it's an actor-specific side effect of the shared `NormalizeMethod` fix, not something scoped to direct service invocation.

All tests run and pass locally via `go test ./tests/integration/...`. Let me know if you'd like additional coverage anywhere.

PR #9837, opened 2026-04-28
placement: post-round coalesce window + stable scheduler reload

JoshVanL:
Introduces two opt-in optimizations on top of the per-type lock granularity work:

1. Stable scheduler reload (always on) actorTable.Types() iterates a sync.Map and returned types in non-deterministic order, so slices.Equal in scheduler.ReloadActorTypes misfired on every round and triggered spurious reconnects. Sort the slice at the disseminator call site before passing it to the scheduler.

2. Post-round coalesce window on the placement server (opt-in) New flag --disseminate-coalesce-window (default 0 = disabled). When set, after a dissemination round transitions back to REPORT with queued host registers/disconnects, the disseminator arms a one-shot timer instead of immediately starting the next round. Additional host events arriving inside the window accumulate in waitingToDisseminate / waitingToDelete and fold into a single follow-up round when CoalesceFire arrives. The first event in REPORT state with no timer pending still fires a round immediately, so cold-starts and isolated churn stay responsive; only sustained bursts pay the (configurable) latency to save rounds.

For 100+ replica scale-ups or rolling restarts this collapses what would have been O(N) follow-up rounds into O(1), removing N-1 rounds of drain + halt cost from every existing daprd in the namespace.

Recommended values are 100-250ms for clusters with frequent rapid churn. Default 0 preserves existing behavior end-to-end and is verified by the broad workflow / dissemination / hotreload suite passing unchanged.


pkg/actors/internal/placement/loops/disseminator/order.go
```diff
@@ -119,14 +143,30 @@ func (d *disseminator) handleOrder(ctx context.Context, order *loops.StreamOrder
...
-		d.scheduler.ReloadActorTypes(d.actorTable.Types())
-
-		d.inflight.Unlock(ctx)
+		// Sort to keep ReloadActorTypes' slices.Equal fast path stable:
+		// actorTable.Types() iterates a sync.Map and returns types in
+		// non-deterministic order, otherwise spurious scheduler
+		// reconnects fire on every routine round.
+		hostedTypes := d.actorTable.Types()
+		slices.Sort(hostedTypes)
+		d.scheduler.ReloadActorTypes(hostedTypes)
```
cicoyle: Should we move the sort into the reload func to prevent future callers from running into the same issue?

pkg/placement/internal/loops/disseminator/closeconn.go
```diff
@@ -116,12 +128,115 @@ func (d *disseminator) advancePhase() {
...
+// them in the store, and either sends one-shot table pushes (no store change)
+// or starts a single follow-up dissemination round covering them all. Must
+// only be called when currentOperation is REPORT and waitingToDelete has
+// already been drained (delete-rounds take priority over add-rounds).
+//
+// forceRound is set by callers that have already mutated the store (eg, by
+// applying queued deletions) and therefore must emit a full round even when
+// the queued adds themselves do not change the store - otherwise existing
+// streams would never see those prior store mutations.
+func (d *disseminator) processWaitingDisseminate(ctx context.Context, forceRound bool) {
```
cicoyle, line 160: Can we add the rest of these funcs to a unit test?

pkg/actors/internal/placement/loops/disseminator/inflight/inflight.go
```diff
@@ -87,46 +111,115 @@ func (i *Inflight) Lock(err error) {
 	lock.LoopFactory.CacheLoop(lo)
 }
 
-func (i *Inflight) Set(in *v1pb.PlacementTables, version uint64) {
-	entries := make(map[string]*hashing.Consistent)
+// Open ensures the claim-tracking loop is running and flushes any queued
+// requests whose actor type is not currently blocked. Idempotent.
```
cicoyle, line 115: Can we also add the rest of these funcs to the unit test?

cicoyle (changes requested):
mostly lgtm

cicoyle (approved):

PR #9844, opened 2026-04-29
feat(subscription): pause-and-drain pubsub components on graceful shutdown

javier-aliaga:
# Description
When the underlying pubsub component implements `pubsub.PausableSubscriber` and `Subscription.Stop()` is called with cause `pubsub.ErrGracefulShutdown`, the runtime now executes the textbook graceful-drain sequence:

1. **SIGTERM received** — daprd starts graceful shutdown
2. **Block shutdown period starts** — `runtime.go` block-shutdown closer fires
3. **Stop consuming new messages, keep heartbeat alive** — `Pause()` calls Sarama's `consumerGroup.PauseAll()`. Broker stops sending FetchResponses, but session and heartbeat loop stay alive
4. **Wait for ACK from app for all in-flight messages** — `closed` stays `false`, handlers continue delivering buffered messages via `postman.Deliver`. App returns SUCCESS / DROP / RETRY; handler returns based on status. Atomic poll on `inflight` waits for every handler to return
5. **Commit all pending offsets** — on SUCCESS, contrib's `doCallback` runs `session.MarkMessage`. Sarama's auto-commit + the final commit during `session.release()` flush them to the broker
6. **Stop heartbeat loop** — part of `consumerGroup.Close()`
7. **Leave consumer group / close session** — `consumerGroup.Close()` sends `LeaveGroup`, broker rebalances
8. **Block shutdown ends** — timer fires or app reports unhealthy
9. **Graceful termination** — runner's `graceful-shutdown-duration` phase waits for any remaining in-flight work (HTTP servers, components)
10. **Container terminates**

This delivers the in-flight transactions to the exiting pod's app before partition handoff, instead of relying purely on broker redelivery to a peer.

## Issue reference

## Checklist

- [ ] Code compiles correctly
- [ ] Created/updated tests
- [ ] Unit tests passing


pkg/runtime/subscription/subscription.go
```diff
@@ -412,9 +437,74 @@ func New(opts Options) (*Subscription, error) {
...
+	// close (e.g. Kafka's consumerGroup.Close → LeaveGroup) happens via
+	// the existing s.cancel path at the end.
+	paused := false
+	if graceful {
+		if pausable, ok := s.pubsub.Component.(contribpubsub.PausableSubscriber); ok {
+			log.Infof("pausing subscription on topic %s during graceful shutdown", s.topic)
+			if perr := pausable.Pause(context.Background()); perr != nil {
+				log.Warnf("failed to pause subscription on topic %s during graceful shutdown: %v", s.topic, perr)
+			} else {
+				paused = true
```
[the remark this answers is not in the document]
javier-aliaga: Pause interface is marked to be idempotent, calling multiple times should not be a probmel

pkg/runtime/subscription/subscription.go
```diff
@@ -412,9 +428,64 @@ func New(opts Options) (*Subscription, error) {
...
+	// will be flushed by Sarama's auto-commit and the final commit during
+	// session.release(). For non-pausable components or non-graceful Stop,
+	// fall back to close-first behavior — without Pause we have no way to
+	// bound the buffer growth, so blocking incoming work is the only safe
+	// option.
+	if !paused {
+		s.closed.Store(true)
+	}
+
 	inflight := s.inflight.Load() > 0
```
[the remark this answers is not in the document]
javier-aliaga: we monitor the inflight count for 100ms to check nothing is flowing to the app

pkg/runtime/subscription/subscription.go
```diff
@@ -412,9 +437,74 @@ func New(opts Options) (*Subscription, error) {
...
+	// session and partition assignments stay alive — and importantly, the
+	// heartbeat loop continues — so any messages already buffered locally
+	// can still be delivered to the app via postman.Deliver. The actual
+	// close (e.g. Kafka's consumerGroup.Close → LeaveGroup) happens via
+	// the existing s.cancel path at the end.
+	paused := false
+	if graceful {
+		if pausable, ok := s.pubsub.Component.(contribpubsub.PausableSubscriber); ok {
+			log.Infof("pausing subscription on topic %s during graceful shutdown", s.topic)
+			if perr := pausable.Pause(context.Background()); perr != nil {
```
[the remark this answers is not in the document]
javier-aliaga: Introduced a timeout context

pkg/runtime/subscription/subscription.go
```diff
@@ -412,9 +437,98 @@ func New(opts Options) (*Subscription, error) {
...
+	for stable < drainStableCount {
+		if s.inflight.Load() > 0 {
+			inflight = true
+			stable = 0
+		} else {
+			stable++
+		}
+		time.Sleep(drainPollInterval)
+	}
+
```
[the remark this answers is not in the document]
javier-aliaga: this is not a real problem, and in theory messages will take longer than 20 ms to be processed

pkg/runtime/subscription/subscription.go
```diff
@@ -412,9 +437,104 @@ func New(opts Options) (*Subscription, error) {
...
+	// caps the whole close phase, so a stuck handler can't hang shutdown
+	// indefinitely.
+	const drainPollInterval = 20 * time.Millisecond
+	requiredStable := 1
+	if paused {
+		requiredStable = 5 // 100ms of stable quiet
+	}
+	stable := 0
+	for stable < requiredStable {
+		if s.inflight.Load() > 0 {
```
cicoyle: Can we add a case for an early exit? Like if `inflight == 0`, I don't think we need to wait the  `drainPollInterval` time.

cicoyle (dismissed):

pkg/runtime/subscription/subscription.go
```diff
@@ -412,9 +437,107 @@ func New(opts Options) (*Subscription, error) {
...
+	}
+
+	// When pausing succeeded, leave s.closed=false during the drain so
+	// handlers continue delivering buffered messages to the app and we get
+	// SUCCESS/DROP/RETRY back from the app. Pending offsets get marked and
+	// will be flushed by Sarama's auto-commit and the final commit during
+	// session.release(). For non-pausable components or non-graceful Stop,
+	// fall back to close-first behavior — without Pause we have no way to
+	// bound the buffer growth, so blocking incoming work is the only safe
+	// option.
```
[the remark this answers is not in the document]
javier-aliaga: Added max drain time

go.mod
```diff
@@ -11,7 +11,7 @@ require (
 	github.com/cenkalti/backoff/v4 v4.3.0
 	github.com/cloudevents/sdk-go/v2 v2.15.2
 	github.com/coreos/go-oidc/v3 v3.17.0
-	github.com/dapr/components-contrib v1.17.3
+	github.com/dapr/components-contrib v1.16.2-0.20260504135955-0f1825477d7e
```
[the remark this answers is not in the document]
javier-aliaga: we will fix this when there is a new version of the lib

acroca:
Can this be covered in integration tests?

javier-aliaga:
> Can this be covered in integration tests?

Not sure it is worth it, we need a component that implements the new PausableSubscriber capability as the current one grpcPubSub is not implementing it (and it is not straightforward).

pkg/runtime/subscription/subscription.go
```diff
@@ -412,9 +442,116 @@ func New(opts Options) (*Subscription, error) {
...
+		requiredStable = 5 // 100ms of stable quiet
+	}
+	deadline := time.Now().Add(drainMaxDuration)
+	stable := 0
+	for {
+		if s.inflight.Load() > 0 {
+			inflight = true
+			stable = 0
+		} else {
+			stable++
```
cicoyle: I dont see that my comment ^ was addressed unless Im mistaken- can we add something like the following
```suggestion
			if !inflight {
			    break
			}
			stable++
```
That way if we've never seen any inflight work, we can exit immediately and not wait.

javier-aliaga: the fix is there, but maybe not clear, let me rework it

javier-aliaga: Ok, I kept it in but only if the component is paused, it could be the case that if it is paused you get 0 the first time so we wait next iteration . But for the non paused path it makes no sense to wait at all.

cicoyle:
Is the copilot feedback addressed?

javier-aliaga:
I decided to drop wg.Wait — it was adding complexity (RWMutex, bounded-wait wrapper, force-cancel) just to avoid the "WaitGroup reused before Wait returned" panic from contrib retry loops. Polling the atomic inflight can't panic, and is much simpler. ctx propagation handles stuck handlers via s.cancel.

dapr/proto/components/v1/pubsub.proto
```diff
@@ -128,4 +138,13 @@ message PullMessagesResponse {
...
+}
+
+// reserved for future-proof extensibility
+message PauseRequest {}
+// reserved for future-proof extensibility
+message PauseResponse {}
+// reserved for future-proof extensibility
+message ResumeRequest {}
+// reserved for future-proof extensibility
+message ResumeResponse {}
```
JoshVanL: fix end of file new line

JoshVanL (changes requested):

tests/integration/framework/process/pubsub/component.go
```diff
@@ -29,13 +34,38 @@ type component struct {
...
+
+	// paused gates PullMessages from forwarding new messages from
+	// pmrRespCh while the component is paused — faithfully simulating
+	// Kafka's PauseAll semantics for integration tests of the runtime's
+	// pause-and-drain shutdown path.
+	paused        atomic.Bool
+	pauseCalled   atomic.Int64
+	resumeCalled  atomic.Int64
+	pauseStartCh  chan struct{}
+	pauseStartOnc atomicOnce
```
JoshVanL:
```suggestion
	pauseStartOnc sync.Once
```

pkg/runtime/subscription/subscription.go
```diff
@@ -412,10 +410,90 @@ func New(opts Options) (*Subscription, error) {
...
+	s.drainSealed.Store(true)
+	s.closed.Store(true)
+
+	// On ceiling hit, force-cancel so stuck handlers' HTTP/gRPC calls
+	// error out via ctx propagation. Join with the original cause so
+	// downstream cause-based checks still see the graceful-shutdown
+	// signal alongside the ceiling-exceeded reason.
+	if hitCeiling {
+		s.cancel(errors.Join(cause, errors.New("subscription drain ceiling exceeded")))
+	}
```
JoshVanL, line 488: Should we not immediately return here?
```suggestion
		s.cancel(errors.Join(cause, errors.New("subscription drain ceiling exceeded")))
		return
	}
```

JoshVanL (changes requested):

pkg/runtime/subscription/subscription.go
javier-aliaga: No because then we miss closing the adapterStrem on line 498. But then on line 504 we only wait 400ms if we did not hit the ceiling, doing it

JoshVanL (approved):

PR #9847, opened 2026-04-30
workflow: gracefully stall on oversized payload

### LINKED ISSUE #9846: Detect workflow high payload size and stall workflow
## In what area(s)?

 /area runtime

> /area operator

> /area placement

> /area docs

> /area test-and-release

> /area injector

> /area scheduler

> /area sentry

## Describe the feature

There is a footgun in Dapr workflow today that allows users to pass payloads that are larger than the maximum dapr.io/max-body-size ([default 4MB](https://docs.dapr.io/reference/arguments-annotations-overview/)) or even hit the limitation that is being set to the application in some SDKs DAPR_GRPC_MAX_INBOUND_MESSAGE_SIZE_BYTES ([set to 4MB in Java as well](https://docs.dapr.io/developing-applications/sdks/java/java-client/properties/#grpc-configuration)). 

The stalled runtime status is clearly exposed on all workflow outputs (cli, API) and would allow workflows that are getting close to this max payload size to gracefully stall the workflow in place until action is taken.

This would greatly improve the user experience when users write workflows.

## Release Note

RELEASE NOTE: **ADD** Auto-stall workflow for large payload size.

JoshVanL:
When a workflow's history payload (PastEvents + NewEvents + PropagatedHistory) exceeded the dapr API gRPC server's MaxSendMsgSize (= --max-body-size, default 4 MiB), stream.Send on the GetWorkItems server stream returned ResourceExhausted, the entire stream was torn down, every other workflow pending on that stream was cancelled, and the SDK reconnected only to repeat the same failure on the same offending workflow.

The orchestrator now precomputes the proto size of the WorkflowRequest / ActivityRequest the executor will build for each dispatch and compares it to a 95% safety threshold of --max-body-size (the headroom covers the engine's WorkflowStarted event injection plus gRPC framing).

Fixes: dapr/dapr#9846
Protobuf: dapr/durabletask-protobuf#45
durabletask-go: https://github.com/dapr/durabletask-go/pull/91

Backport for v1.17.7

Adds missing v1.17.6 release notes


olitomlinson:
@JoshVanL What is an operator supposed to do when a workflow gracefully stalls? Is it effectively terminated at this point and nothing can be done with it?

JoshVanL:
@olitomlinson the workflow isn't terminal, just stalled- this is the same state as when a workflow version isn't available (due to a new Deployment roll out or similar). https://docs.dapr.io/developing-applications/building-blocks/workflow/workflow-versioning/

Users can either [increase the max body size](https://docs.dapr.io/operations/configuration/increase-request-size/) as is described in the stalled status description- which will be logged as well as be in `$ dapr wf list` output etc, or they can `purge --force` the workflow if something has gone haywire and they don't want to process it due to an application error/bug producing too much data.

JoshVanL:
We specifically don't want to make the workflow terminal here, since the workflow may legitimately need this size of payload, so we want to gives users the ability to increase the max body size, which will let the workflow continue execution.

JoshVanL:
FYI, we unload the workflow state from memory when stalled, so there shouldn't be a problem with OOM issues when there is a storm of payload max size exceeded https://github.com/dapr/dapr/blob/306901b55b6f3caeedef0b29e29f379420aacbec/pkg/actors/targets/workflow/orchestrator/versioning.go#L98-L100

pkg/actors/targets/workflow/orchestrator/activity.go
```diff
@@ -64,6 +64,10 @@ func (o *orchestrator) callActivity(ctx context.Context, e *backend.HistoryEvent
 		return nil
 	}
 
+	if err := o.activityPayloadOversize(e, ph); err != nil {
```
cicoyle: Should this move down to include the `payload` from below otherwise I think we may under count the actual wire msg.

cicoyle (changes requested):

pkg/actors/targets/workflow/orchestrator/activity.go
JoshVanL: Done :+1:

olitomlinson:
> @olitomlinson the workflow isn't terminal, just stalled- this is the same state as when a workflow version isn't available (due to a new Deployment roll out or similar). https://docs.dapr.io/developing-applications/building-blocks/workflow/workflow-versioning/
> 
> Users can either [increase the max body size](https://docs.dapr.io/operations/configuration/increase-request-size/) as is described in the stalled status description- which will be logged as well as be in `$ dapr wf list` output etc, or they can `purge --force` the workflow if something has gone haywire and they don't want to process it due to an application error/bug producing too much data.

Got it thank you :) 

Does the 4mb default limit apply to the total size of the history for a given workflow instance, or is it 4mb per history event?

JoshVanL:
> > @olitomlinson the workflow isn't terminal, just stalled- this is the same state as when a workflow version isn't available (due to a new Deployment roll out or similar). https://docs.dapr.io/developing-applications/building-blocks/workflow/workflow-versioning/
> 
> > 
> 
> > Users can either [increase the max body size](https://docs.dapr.io/operations/configuration/increase-request-size/) as is described in the stalled status description- which will be logged as well as be in `$ dapr wf list` output etc, or they can `purge --force` the workflow if something has gone haywire and they don't want to process it due to an application error/bug producing too much data.
> 
> 
> 
> Got it thank you :) 
> 
> 
> 
> Does the 4mb default limit apply to the total size of the history for a given workflow instance, or is it 4mb per history event?

The full history log size passed to the SDK

cicoyle (approved):

PR #9850, opened 2026-04-30
WorkflowAccessPolicy: Extend to all workflow operations

JoshVanL:
Adds per-operation access control for every workflow operation (schedule, terminate, raise, pause, resume, purge, get, rerun) and splits WorkflowOperationRule into separate WorkflowRule and ActivityRule blocks - activities only have schedule, so the unified shape was misleading.

Schema:
  - WorkflowAccessPolicyRule.Operations -> Workflows (with operations list) + Activities. CRD enum widened to all eight values; CEL rule blocks empty rules.

Enforcement moves into the actor:
  - Per-operation policy evaluation happens inside the orchestrator and activity actors under the actor lock, after loadInternalState. This eliminates the resolve/execute TOCTOU race that an upstream check has when the workflow name is fetched outside the lock.
  - The gRPC handler and actor router now only authenticate the caller (mTLS/SPIFFE for remote, local appID for same-sidecar) and stamp identity into req.Metadata; the actor reads it back.
  - For AddWorkflowEvent, the HistoryEvent is unmarshaled once in handleInvoke and threaded through both the policy check and addWorkflowEvent to avoid a duplicate proto.Unmarshal.

Holder added in pkg/acl/workflow shared between the gRPC API and the workflow actors so both read the same atomic snapshot.

---

```yaml
apiVersion: dapr.io/v1alpha1
kind: WorkflowAccessPolicy
metadata:
  name: orders-policy
  namespace: production
scopes:
  - orders-target
spec:
  # defaultAction is taken when no rule matches. "deny" (default) or "allow".
  defaultAction: deny
  rules:
    # Each rule pairs a set of callers with workflow and/or activity rules.
    # At least one of `workflows` or `activities` must be set.
    - callers:
        - appID: frontend
        - appID: ops-console
      workflows:
        # Exact name, all eight operations.
        - name: OrderWF
          operations:
            - schedule
            - terminate
            - raise
            - pause
            - resume
            - purge
            - get
            - rerun
          action: allow

        # Glob pattern - read-only access to anything under "Report*".
        - name: "Report*"
          operations: [get]
          action: allow

        # Wildcard with a deny - frontend can never purge ProcessSecret*.
        - name: "ProcessSecret*"
          operations: [purge]
          action: deny

      activities:
        # Activities only have the schedule operation, so there is no
        # `operations` field. Glob patterns and exact names both work.
        - name: ChargePayment
          action: allow
        - name: "*"
          action: deny

    # A second rule for a different caller. The compiler unions rules
    # across all matching policies; deny wins ties at the same specificity.
    - callers:
        - appID: orders-target
      workflows:
        - name: "*"
          operations: [schedule, terminate, get]
          action: allow
      activities:
        - name: "*"
          action: allow
```

```
┌───────────┬────────────────────────────────────────┐
│ Operation │              Triggered by              │
├───────────┼────────────────────────────────────────┤
│ schedule  │ StartWorkflow / CreateWorkflowInstance │
├───────────┼────────────────────────────────────────┤
│ terminate │ TerminateWorkflow                      │
├───────────┼────────────────────────────────────────┤
│ raise     │ RaiseEventWorkflow                     │
├───────────┼────────────────────────────────────────┤
│ pause     │ PauseWorkflow                          │
├───────────┼────────────────────────────────────────┤
│ resume    │ ResumeWorkflow                         │
├───────────┼────────────────────────────────────────┤
│ purge     │ PurgeWorkflow                          │
├───────────┼────────────────────────────────────────┤
│ get       │ GetWorkflow / WaitForRuntimeStatus     │
├───────────┼────────────────────────────────────────┤
│ rerun     │ RerunWorkflowFromEvent                 │
└───────────┴────────────────────────────────────────┘
```

Specificity rules (deterministic when multiple rules match):

1. Longest literal prefix wins (Process* beats *).
2. Exact match beats glob at the same prefix length (OrderWF beats OrderWF*).
3. At equal specificity, deny beats allow (fail-closed).
4. With no matching rule, defaultAction applies.


pkg/acl/workflow/extract.go
```diff
@@ -58,42 +59,36 @@ func ParseActorType(actorType string) (OperationType, bool) {
...
-		// Only CreateWorkflowInstance is subject to access control (schedule
-		// operation).
-		return "", false, nil
-	}
+	case todo.AddWorkflowEventMethod:
+		if parsedAddEvent == nil {
+			return "", true, errors.New("AddWorkflowEvent: parsed event is required to derive the operation")
+		}
+		op, err := operationFromHistoryEvent(parsedAddEvent)
+		return op, true, err
```
cicoyle: Should the bool be removed here bc I see we return true for both the err and nil case

pkg/acl/workflow/extract.go
```diff
@@ -28,8 +30,7 @@ const (
 	suffixWorkflow  = ".workflow"
 	suffixActivity  = ".activity"
 
-	methodCreateWorkflowInstance = "CreateWorkflowInstance"
-	methodExecute                = "Execute"
+	methodExecute = "Execute"
```
cicoyle: Should this also be moved to the `todo` pkg?

cicoyle (changes requested):

JoshVanL:
Thanks @cicoyle, please take another look

cicoyle (approved):

PR #9731, opened 2026-03-31
feat(mcp): add built-in mcp wf orchestration engine for mcpserver crd

sicoyle:
# Description

NOTE: Build will fail as I depend on this: https://github.com/dapr/durabletask-go/pull/83

This PR builds on top of my initial MCPServer resource PR here: https://github.com/dapr/dapr/pull/9685

This implements the runtime machinery behind the MCPServer CRD defined in this proposal: https://github.com/dapr/proposals/pull/100

When a sidecar loads one or more `MCPServer` resources (gated behind the `MCPServerResource` preview feature flag), daprd registers built-in `dapr.mcp.<name>.ListTools` and `dapr.mcp.<name>.CallTool` workflow orchestrations that execute MCP protocol calls as durable activities. This means that callers interact with MCP servers through the standard Dapr Workflow API which means they don't need to introduce the MCP SDK into their application, there's no connection boilerplate, or credential management required in application code. That's all done transparently through Dapr.

The executor is installed lazily via a `RoutingExecutor` wrapper so sidecars with no `MCPServer` resources incur zero MCP-related allocations - I did this for the `sidecar-validations` build step :) . 
- Transport support covers `streamable_http`, `sse`, and `stdio`.
- Auth covers static header injection, OAuth2 client credentials (secret fetched at call time from Dapr secret stores with retryable errors on transient failures), and SPIFFE JWT SVID injection.
- Tool result handling covers all MCP spec content types (text, image, audio, resource_link, embedded resource) with a JSON fallback for unknown/future types. I cross checked these with the MCP docs on tool results here: https://modelcontextprotocol.io/specification/2025-11-25/server/tools#tool-result

Hook | Error behavior | Mutate behavior
-- | -- | --
beforeCallTool | Workflow COMPLETES, isError=true in output | Replaces arguments for next hook + tool call
afterCallTool | Workflow FAILS | Replaces result for next hook + caller
beforeListTools | Workflow FAILS | N/A (no mutate field)
afterListTools | Workflow FAILS | Replaces tool list for next hook + caller

[...]

PR #9857, opened 2026-05-04
fix(workflow-acl): allow same-sidecar self-invocations to bypass policy

sicoyle:
# Description

I was seeing integration test failures here https://github.com/dapr/dapr/pull/9731 that this PR fixes. 

WorkflowAccessPolicy was denying same-sidecar self-invocations on non-subject workflow methods (e.g. `AddWorkflowEvent`, `PurgeWorkflowState`), breaking any workflow that uses sub-workflows on a sidecar that has a policy defined.

## Root cause

When a workflow finishes and emits `AddWorkflowEvent` back to its parent workflow actor on the same sidecar, the call goes through the gRPC `CallActor` endpoint (this is how durabletask routes parent/child events even within the same app). 

The ACL hook then:
1. Sees `method=AddWorkflowEvent` — a non-subject method whose target instance name cannot be extracted from the request payload.
2. Falls back to `IsCallerKnown(callerAppID, opType)`, which requires the caller to have a wildcard-allow rule (and no deny rules) to be considered trusted for non-subject methods.
3. The sidecar invoking its own actors is rarely listed in its own `WorkflowAccessPolicy.callers` (no reason to write a rule about yourself), so `IsCallerKnown` returns false → `PermissionDenied`.

The orchestrator treats this as a recoverable error and schedules a retry; the reminder fires every second forever, the parent workflow never receives the child's result, and `WaitForWorkflowCompletion` hangs.

## Fix

Add `isSelfCall(callerAppID, callerNamespace)` and bail out early in both `callActorValidateWorkflowACL` and `callActorReminderValidateWorkflowACL` when the caller's mTLS-verified SPIFFE identity matches this sidecar's own appID and namespace. mTLS guarantees the SPIFFE identity is not spoofable, so equality is a sound signal that the call originated from this sidecar's own actors.

WorkflowAccessPolicy is meant to gate **cross-app** access; applying it to a sidecar talking to its own actors was never the intent.

## Issue reference

## Checklist

- [ ] Code compiles correctly
- [ ] Created/updated tests
- [ ] Unit tests passing


sicoyle:
closing this as we are awaiting a maintainers decision for how we want to handle this case for the workflow access policy resource

PR #9870, opened 2026-05-04
WorkflowAccessPolicy: pure allow-list with self-call exemption

JoshVanL:
Reshapes WorkflowAccessPolicy from an allow/deny rules engine with specificity ranking into a simple allow-list, and exempts self-calls from enforcement.

Schema:
- Drop `WorkflowAccessPolicySpec.DefaultAction`.
- Drop `Action` from `WorkflowRule` and `ActivityRule`. A rule's presence now grants access; absence denies. The `PolicyAction` type is removed.

Semantics:
- When 0 policies are loaded for the target app, all calls are allowed (open by default, backward compatible).
- When 1+ policies are loaded, the target defaults to deny; only matching rules grant access.
- Self-calls (caller appID == target appID) are always allowed. The policy is a cross-app gate only and the local daprd cannot be denied access to its own actors.
- Cross-app callers are also denied for non-subject methods on workflow/activity actors and for reminder injection. These paths are reachable only by the local daprd in normal operation.

Compiler/evaluator:
- `compiledOp` reduces to {opType, operation, pattern}.
- `CompiledPolicies` reduces to {rules}.
- `IsCallerKnown`, specificity ranking and the deny-wins/longest- prefix tiebreakers are removed, they have no meaning under a pure allow-list.
- `Evaluate` returns true on the first rule whose caller, opType, operation and glob pattern all match.

Enforcement:
  - In-actor `checkAccessPolicy` returns nil immediately when the caller equals the local appID, regardless of method.
  - Non-subject method calls from cross-app callers are denied (subsumes the old partialauthbypass invariant).
  - The reminder gRPC handler requires `callerAppID == a.AppID()` when policies are loaded.

---

```yaml
apiVersion: dapr.io/v1alpha1
kind: WorkflowAccessPolicy
metadata:
  name: orders-policy
  namespace: production
scopes:
  # The target apps that load and enforce this policy.
  - orders-target

spec:
  # Allow-list. A call is permitted iff some rule below matches the caller,
  # the operation and the workflow/activity name. With this policy loaded
  # the target defaults to deny for everything not listed; remove all
  # WorkflowAccessPolicy resources to revert to allow-all. Self-calls
  # (orders-target -> orders-target) are always allowed and need no rule.
  rules:
    # Frontend and ops-console can drive OrderWF end-to-end.
    - callers:
        - appID: frontend
        - appID: ops-console
      workflows:
        - name: OrderWF
          operations:
            - schedule
            - terminate
            - raise
            - pause
            - resume
            - purge
            - get
            - rerun

    # Reporting tools get read-only access to anything matching Report*.
    - callers:
        - appID: reporting
      workflows:
        - name: "Report*"
          operations: [get]

    # Frontend can also schedule the ChargePayment activity directly.
    # Activities only have one operation (schedule), so no operations field.
    - callers:
        - appID: frontend
      activities:
        - name: ChargePayment
        - name: "RefundEvent*"
```


tests/integration/suite/daprd/workflow/accesspolicy/deny.go
```diff
@@ -67,14 +69,12 @@ metadata:
 scopes:
 - wfacl-target
 spec:
```
acroca, line 71: I think it'd be good to test with `spec: {}`, to make sure just by having a policy we get a deny, regardless of the rules.

acroca (commented):
LGTM, I like how much simpler it got :)

tests/integration/suite/daprd/workflow/accesspolicy/deny.go
JoshVanL: :+1: Thanks, done

cicoyle (approved):

PR #9873, opened 2026-05-04
Injector/Operator: use RSA keys for webhook serving certs

JoshVanL:
The Kubernetes API server on some managed cloud distributions rejects Ed25519 serving certificates when invoking admission webhooks. After the workload-identity key algorithm was switched to Ed25519, the Dapr injector (mutating sidecar webhook) and operator (CRD conversion, validating, and mutating webhooks) could no longer be reached by the apiserver in those environments.

The injector and operator each have a single SPIFFE identity that backs both their webhook TLS cert and any other mTLS they perform, so switching the entire identity algorithm is sufficient. Sentry's CA already accepts CSRs of any supported algorithm, so no signing-side changes are needed.

This commit plumbs a new spiffe.KeyAlgorithm option through security.Options into spiffe.New (added in dapr/kit) and sets it to ptr.Of(spiffe.KeyAlgorithmRSA) in the injector and operator binaries. All other control plane components (sentry, placement, scheduler) and daprd workloads keep the Ed25519 default, since the new option is nil-defaulted.


cmd/injector/app/app.go
```diff
@@ -109,6 +111,10 @@ func Run() {
 		MTLSEnabled:             true,
 		Mode:                    modes.KubernetesMode,
 		Healthz:                 healthz,
+		// The injector serves a mutating admission webhook to the Kubernetes
+		// API server, which on some cloud distributions rejects Ed25519
+		// serving certs.
+		KeyAlgorithm: ptr.Of(spiffe.KeyAlgorithmRSA),
```
acroca, line 117: I guess we can't know which cert would be allowed, right? It'd be cool to use ed25519 when possible, but I'm not sure we can check if it's supported dynamically.

acroca (dismissed):

cmd/injector/app/app.go
JoshVanL: Yeah it's tricky because as the webhook server it is difficult to tell why the peer didn't connect, and it would be a lot of machinery to rotate with a different key in these scenarios

cicoyle (approved):
changes lgtm

PR #9876, opened 2026-05-05
fix(placement): always arm coalesce timer after round to fix dissemination race

sicoyle:
# Description

My integration tests fail consistently here https://github.com/dapr/dapr/pull/9731 due to a race my tests uncovered in placement under `Test_Integration/placement/dissemination/coalesce.`.

Fixes a race in the placement disseminator where a fresh dissemination round could fire immediately after a previous round completed, defeating the `--disseminate-coalesce-window` setting.

## Root cause

`handleReportedUnlock` only armed the coalesce timer if `waitingToDisseminate` or `waitingToDelete` was non-empty at the moment the final UNLOCK ack arrived:

```go
if d.coalesceWindow > 0 && (len(d.waitingToDelete) > 0 || len(d.waitingToDisseminate) > 0) {
    d.startCoalesceTimer()
    return
}
```

When an UNLOCK ack and a ConnAdd for a new stream both land in the disseminator's event queue concurrently, the queue order is non-deterministic. If UNLOCK is popped first:

1. Round completes, currentOperation → REPORT.
2. The check above sees both queues empty and skips the timer.
3. The pending ConnAdd is then popped, handleAdd finds currentOperation == REPORT && coalesceTimer == nil, and starts a new round immediately — no coalescing.

Fix
Always arm the coalesce timer after a round completes when coalesceWindow > 0, regardless of whether anything is currently queued. handleCoalesceFire already handles the empty-queue case as a no-op, so unconditional arming is safe and is already covered by TestHandleCoalesceFire_EmptyQueues_ClearsTimer.

## Issue reference

## Checklist

- [ ] Code compiles correctly
- [ ] Created/updated tests
- [ ] Unit tests passing

### DIFF
--- a/pkg/placement/internal/loops/disseminator/host.go
+++ b/pkg/placement/internal/loops/disseminator/host.go
@@ -181,12 +181,15 @@ func (d *disseminator) handleReportedUnlock(ctx context.Context, streamIDx uint6
 			return ok
 		}, &d.waitingToDelete)
 
-		// If a coalesce window is configured and there is something queued,
-		// defer the next round so additional churn arriving in the window
-		// folds into a single follow-up round. The first event that arrived
-		// in REPORT state already triggered THIS round immediately, so cold
-		// starts stay responsive.
-		if d.coalesceWindow > 0 && (len(d.waitingToDelete) > 0 || len(d.waitingToDisseminate) > 0) {
+		// If a coalesce window is configured, always arm the timer after a
+		// round completes so any churn arriving inside the window folds
+		// into a single follow-up round. Without this, an UNLOCK ack and a
+		// new ConnAdd that race through the disseminator's event queue can
+		// be processed in either order: if the ConnAdd loses the race, the
+		// queue is empty here and a cold-start round fires immediately for
+		// the late arrival, defeating coalescing. handleCoalesceFire is a
+		// no-op when queues are empty when the timer elapses.
+		if d.coalesceWindow > 0 {
 			d.startCoalesceTimer()
 			return
 		}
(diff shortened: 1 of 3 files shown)


pkg/placement/internal/loops/disseminator/host.go
```diff
@@ -181,12 +181,15 @@ func (d *disseminator) handleReportedUnlock(ctx context.Context, streamIDx uint6
...
-		if d.coalesceWindow > 0 && (len(d.waitingToDelete) > 0 || len(d.waitingToDisseminate) > 0) {
+		// If a coalesce window is configured, always arm the timer after a
+		// round completes so any churn arriving inside the window folds
+		// into a single follow-up round. Without this, an UNLOCK ack and a
+		// new ConnAdd that race through the disseminator's event queue can
+		// be processed in either order: if the ConnAdd loses the race, the
+		// queue is empty here and a cold-start round fires immediately for
+		// the late arrival, defeating coalescing. handleCoalesceFire is a
+		// no-op when queues are empty when the timer elapses.
+		if d.coalesceWindow > 0 {
```
JoshVanL, line 192: Please can we have an integration test covering this please?

JoshVanL (changes requested):

pkg/placement/internal/loops/disseminator/host.go
sicoyle: yes adding :)

JoshVanL (changes requested):
Please can we make the same change here https://github.com/dapr/dapr/blob/11015aee64ad9e93eed136fdb811486f13e3e84b/pkg/placement/internal/loops/disseminator/closeconn.go#L136

cicoyle (approved):
lgtm

JoshVanL (approved):

PR #9877, opened 2026-05-05
Integration: fix flaky tests via metadata-API polling and tuning

JoshVanL:
metadata API- add MetadataResiliency to GetMetadataResponse, populated from compStore.ListResiliencyResources(). Mirrors the existing MetadataWorkflowAccessPolicy entry.

Stop scraping logs to detect SIGHUP-driven reloads. Poll the metadata API for the loaded resiliency name instead. Three subtests (add / rename / delete) each assert the expected resource list, removing the timing dependency on log buffering.

Add `get` to the WAP policy.

Increase requememory client timeout & increase sampler poll.

Reduce amend error workflow func.

Add single activity execution int tests for correctness validation.


famarting (approved):
I understand that when resiliencies or wap show up in the metadata response it is guaranteed that the runtime has completely loaded them

cicoyle (dismissed):

cicoyle (approved):

PR #9884, opened 2026-05-05
Add requires to WorkflowAccessPolicy: gate operations on caller history

cicoyle:
Extends WorkflowAccessPolicy with a `requires` block that gates an operation on specific events appearing in the caller's propagated workflow history.


cicoyle:
Will fix conflicts

pkg/apis/workflowaccesspolicy/v1alpha1/types.go
```diff
@@ -147,47 +140,89 @@ const (
...
+
+// RequiredEvent is a single entry in a rule's Requires list. It matches when
+// the caller's propagated history contains an event with the given Status
+// produced by the named activity / workflow / external event. Exactly one
+// of activityName, workflowName, or eventName must be set.
+//
+// +kubebuilder:validation:XValidation:rule="(has(self.activityName) ? 1 : 0) + (has(self.workflowName) ? 1 : 0) + (has(self.eventName) ? 1 : 0) == 1",message="exactly one of activityName, workflowName, or eventName must be set"
+type RequiredEvent struct {
+	// +kubebuilder:validation:Enum=Started;Completed;Failed;Raised
+	Status RequiredStatus `json:"status"`
```
JoshVanL: What would be the use case for wanting anything other than this event being successful- it seems like a foot gun to match on failed events?

pkg/apis/workflowaccesspolicy/v1alpha1/types.go
```diff
@@ -147,47 +140,89 @@ const (
...
+
+	// WorkflowName restricts the match to events produced by the named workflow
+	// (own or child).
+	// +optional
+	WorkflowName string `json:"workflowName,omitempty"`
+
+	// EventName restricts the match to an external event with the given name.
+	// Only valid with Status=Raised.
+	// +optional
+	EventName string `json:"eventName,omitempty"`
```
JoshVanL: Instead of multiple fields which end up dead on being mutually exclusive, I think we should have an `EventType` where the user would put `workflow`, `activity` etc .

pkg/apis/workflowaccesspolicy/v1alpha1/types.go
```diff
@@ -147,47 +140,89 @@ const (
...
+
+	// AppID restricts the match to events produced by the named app.
+	// +optional
+	AppID string `json:"appID,omitempty"`
```
JoshVanL: Optional fields should be pointers. 
```suggestion
	// AppID restricts the match to events produced by the named app.
	// +optional
	AppID *string `json:"appID,omitempty"`
```

pkg/apis/workflowaccesspolicy/v1alpha1/types.go
```diff
@@ -147,47 +140,89 @@ const (
...
 	// Action is "allow" or "deny".
 	// +kubebuilder:validation:Enum=allow;deny
 	Action PolicyAction `json:"action"`
+
+	// Requires is a list of history events that must all be present in the
+	// caller's propagated workflow history for this rule to apply. When set,
+	// the rule only matches if every entry can be located in the caller's
+	// history. When the field is omitted/empty, means nothing is required.
+	// +optional
+	Requires []RequiredEvent `json:"requires,omitempty"`
```
JoshVanL: Does this test order? I assume so, but should be written in the comments.

JoshVanL (changes requested):

pkg/apis/workflowaccesspolicy/v1alpha1/types.go
cicoyle: To express compensation patterns declaratively. So `Status: Failed` enables ppl to run an activity only if xyz failed on a cross app instance, where a try/catch wouldn't work bc its a cross app invocation.

cicoyle: After more thought, I think you're right - matching on failed to enable compensation will require more work than I was originally thinking. I think it will be a great future add to have some sort of `WorkflowCompensationPolicy` CRD with a `TaskCompensationFired` history event or similar and enabling that declaratively. Ill noodle more on it

cicoyle: I did remove the `Failed` in the last commit 👍🏻

cicoyle:
Will figure out why CI is all red and fix shortly

tests/integration/suite/daprd/workflow/accesspolicy/deny.go
```diff
@@ -142,4 +142,7 @@ func (d *deny) Run(t *testing.T, ctx context.Context) {
 	require.NotNil(t, metadata.GetFailureDetails(),
```
JoshVanL, line 142: Do we have non-cross app requires tests?

tests/integration/suite/daprd/workflow/accesspolicy/requires/crossapp.go
```diff
@@ -0,0 +1,395 @@
...
+        status: Completed
+        name: FraudCheckPassed
+        appID: wfacl-requires-caller
+    - name: ProcessPaymentAppMismatch
+      requires:
+      - eventType: activity
+        status: Completed
+        name: FraudCheckPassed
+        appID: nonexistent-app
+    - name: LogReceipt
```
JoshVanL: No  workflow or raise event tests?

tests/integration/suite/daprd/workflow/accesspolicy/requires/crossapp.go
```diff
@@ -0,0 +1,395 @@
...
+	require.NoError(t, os.WriteFile(configFile, []byte(`
+apiVersion: dapr.io/v1alpha1
+kind: Configuration
+metadata:
+  name: wfaclrequiresconfig
+spec:
+  features:
+  - name: WorkflowAccessPolicy
+    enabled: true
+`), 0o600))
```
JoshVanL: Feature gate doesn't exist

pkg/apis/workflowaccesspolicy/v1alpha1/types.go
```diff
@@ -169,6 +180,57 @@ type ActivityRule struct {
...
+	Status RequiredStatus `json:"status"`
+
+	// Name is the activity/workflow name (eventType=activity|workflow) or
+	// the external event name (eventType=event).
+	// +kubebuilder:validation:MinLength=1
+	Name string `json:"name"`
+
+	// AppID restricts the match to events produced by the named app.
+	// +optional
+	AppID *string `json:"appID,omitempty"`
```
JoshVanL: Need int tests covering these CEL ruless

pkg/acl/workflow/policy.go
```diff
@@ -128,11 +160,162 @@ func (cp *CompiledPolicies) Evaluate(callerAppID string, opType OperationType, o
...
+	// completion events can resolve their scheduled name.
+	var events []decodedEvent
+	taskScheduledByID := make(map[int32]*protos.TaskScheduledEvent)
+	childCreatedByID := make(map[int32]*protos.ChildWorkflowInstanceCreatedEvent)
+	for _, chunk := range history.GetChunks() {
+		for _, raw := range chunk.GetRawEvents() {
+			e := &protos.HistoryEvent{}
+			if err := proto.Unmarshal(raw, e); err != nil {
+				continue
+			}
```
JoshVanL, line 292: This means dapr runtime is going to unmarshal multiple times for a single cycle?

pkg/acl/workflow/policy.go
```diff
@@ -128,11 +160,162 @@ func (cp *CompiledPolicies) Evaluate(callerAppID string, opType OperationType, o
...
+	// Decode every chunk's raw events once; index by event ID so
+	// completion events can resolve their scheduled name.
+	var events []decodedEvent
+	taskScheduledByID := make(map[int32]*protos.TaskScheduledEvent)
+	childCreatedByID := make(map[int32]*protos.ChildWorkflowInstanceCreatedEvent)
+	for _, chunk := range history.GetChunks() {
+		for _, raw := range chunk.GetRawEvents() {
+			e := &protos.HistoryEvent{}
+			if err := proto.Unmarshal(raw, e); err != nil {
+				continue
```
JoshVanL: Why do we continue here, and not error?

pkg/acl/workflow/policy.go
```diff
@@ -128,11 +160,162 @@ func (cp *CompiledPolicies) Evaluate(callerAppID string, opType OperationType, o
...
+			}
+			events = append(events, decodedEvent{e: e, chunk: chunk})
+			if ts := e.GetTaskScheduled(); ts != nil {
+				taskScheduledByID[e.GetEventId()] = ts
+			}
+			if cc := e.GetChildWorkflowInstanceCreated(); cc != nil {
+				childCreatedByID[e.GetEventId()] = cc
```
JoshVanL: This looks like it will cause collisions between chunks?

pkg/acl/workflow/policy.go
```diff
@@ -128,11 +160,162 @@ func (cp *CompiledPolicies) Evaluate(callerAppID string, opType OperationType, o
...
+
+			return true, DenialReasonNone
+		}
+	}
+
+	if requiresNotMet {
+		return false, DenialReasonRequiresUnmet
+	}
+	return false, DenialReasonNotAllowed
+}
```
JoshVanL, line 209: Is it safe for us to be propagate the reason to the caller? Probably not.

JoshVanL (changes requested):

tests/integration/suite/daprd/workflow/accesspolicy/deny.go
cicoyle: yeah most of if not all of the integration requires tests are cross app 👍🏻

tests/integration/suite/daprd/workflow/accesspolicy/requires/crossapp.go
cicoyle: added

tests/integration/suite/daprd/workflow/accesspolicy/requires/crossapp.go
cicoyle: I missed removing this when I merged in master after your PR got merged. Its removed now, thx!

pkg/acl/workflow/policy.go
cicoyle: should be addressed now

pkg/apis/workflowaccesspolicy/v1alpha1/types.go
```diff
@@ -160,6 +162,15 @@ type WorkflowRule struct {
...
 	Operations []WorkflowOperation `json:"operations"`
+
+	// Requires is a list of history events that must all be present in the
+	// caller's propagated workflow history for this rule to apply. Entries
+	// are evaluated as a set — order is not significant. Only meaningful
+	// for operations that carry propagated history (currently `schedule`);
+	// rules whose `requires` cover other operations will fail-closed
+	// because no history is available to satisfy them.
+	// +optional
+	Requires []RequiredEvent `json:"requires,omitempty"`
```
JoshVanL: Requires here doesn't make sense in the context of operation which are not workflow schedules, only schedule has propagated context - how should be reconcile this? I think this is a fundamental issue with the API shape.

```
  workflows:
  - name: TargetWF
    operations: [raise]
    requires: [...]
```

JoshVanL (changes requested):

pkg/apis/workflowaccesspolicy/v1alpha1/types.go
cicoyle: this should be addressed now in the last commit

pkg/actors/targets/workflow/activity/accesspolicy.go
```diff
@@ -23,8 +23,6 @@ import (
 	internalsv1pb "github.com/dapr/dapr/pkg/proto/internals/v1"
 )
 
-const workflowACLDeniedMsg = "access denied by workflow access policy"
-
 func (a *activity) checkAccessPolicy(method string, data []byte, md map[string]*internalsv1pb.ListStringValue) error {
```
JoshVanL, line 26: We need to ensure that this is executed _after_ we have done signature checks on the propogated history, rather than after.

pkg/acl/workflow/policy.go
```diff
@@ -16,12 +16,26 @@ package workflow
...
+	"google.golang.org/protobuf/proto"
+
 	wfaclapi "github.com/dapr/dapr/pkg/apis/workflowaccesspolicy/v1alpha1"
+	"github.com/dapr/durabletask-go/api/protos"
 	"github.com/dapr/kit/logger"
 )
 
 var log = logger.NewLogger("dapr.acl.workflow")
 
+const DeniedMessageBase = "access denied by workflow access policy"
```
JoshVanL, line 29: Duplicated in pkg/api/grpc/workflow_acl.go

pkg/diagnostics/service_monitoring.go
```diff
@@ -552,7 +553,8 @@ func (s *serviceMetrics) WorkflowACLActionDenied(callerAppID, opType, operation
 				appIDKey, s.appID,
 				sourceAppIDKey, callerAppID,
 				operationKey, operation,
-				typeKey, opType)...),
+				typeKey, opType,
+				failReasonKey, reason)...),
```
JoshVanL: Not sure failReasonKey should be a new cardinality key?

JoshVanL (changes requested):

pkg/apis/workflowaccesspolicy/v1alpha1/types.go
```diff
@@ -157,9 +159,28 @@ type WorkflowRule struct {
 
 	// Operations is the set of operations this rule applies to.
 	// +kubebuilder:validation:MinItems=1
-	// +kubebuilder:validation:items:Enum=schedule;terminate;raise;pause;resume;purge;get;rerun
-	// +listType=set
-	Operations []WorkflowOperation `json:"operations"`
+	// +listType=map
+	// +listMapKey=name
+	Operations []WorkflowRuleOperation `json:"operations"`
+}
```
JoshVanL: Operations are keyed by name, meaning you can't have two different rules for the same workflow operation (with different requires).

JoshVanL (changes requested):

tests/integration/suite/daprd/workflow/accesspolicy/requires/activity.go
```diff
@@ -0,0 +1,200 @@
...
+	r.sched = scheduler.New(t, scheduler.WithSentry(r.sentry), scheduler.WithID("dapr-scheduler-server-0"))
+	r.db = sqlite.New(t, sqlite.WithActorStateStore(true), sqlite.WithCreateStateTables())
+
+	callerID := "caller-" + uuid.NewString()
+	targetID := "target-" + uuid.NewString()
+	ns := uuid.NewString()
+
+	r.gatedActivity = "gated-" + uuid.NewString()
+	r.wfWithPreflight = "wf-with-preflight-" + uuid.NewString()
+	r.wfNoPreflight = "wf-no-preflight-" + uuid.NewString()
```
acroca: Why does it need uuid in the name? I think most tests have fixed names, and I find it easier to read.

tests/integration/suite/daprd/workflow/accesspolicy/requires/child.go
```diff
@@ -0,0 +1,270 @@
...
+	require.NoError(t, callerClient.StartWorkItemListener(ctx, callerReg))
+	targetClient := client.NewTaskHubGrpcClient(r.target.GRPCConn(t, ctx), logger.New(t))
+	require.NoError(t, targetClient.StartWorkItemListener(ctx, targetReg))
+
+	assert.EventuallyWithT(t, func(c *assert.CollectT) {
+		assert.GreaterOrEqual(c, len(r.caller.GetMetadata(t, ctx).ActorRuntime.ActiveActors), 1)
+		assert.GreaterOrEqual(c, len(r.target.GetMetadata(t, ctx).ActorRuntime.ActiveActors), 1)
+	}, 20*time.Second, 10*time.Millisecond)
+
+	t.Run("child workflow allowed when ChildWorkflowInstanceCompleted in propagated history", func(t *testing.T) {
```
acroca, line 177: I'd prefer to see only one scenario per test file, wdyt? I find having multiple scenarios in a single file, even though they are related, it's harder to follow.

cicoyle: These are all related tests - I am hoping to leave it as is instead of breaking it up. There is a t.run for each instead of a new test file, but its all the same and requires less boiler plate code being copy/pasted around.

tests/integration/suite/daprd/workflow/accesspolicy/requires/activity.go
cicoyle: I can rm the uuids 👍🏻

acroca (dismissed):

pkg/actors/targets/workflow/orchestrator/accesspolicy.go
```diff
@@ -43,60 +42,77 @@ func (o *orchestrator) checkAccessPolicy(ctx context.Context, method string, dat
...
-	if !policies.Evaluate(callerAppID, workflowacl.OperationTypeWorkflow, operation, name) {
-		log.Warnf("Workflow actor '%s': workflow access policy denied app '%s' operation '%s' on '%s'", o.actorID, callerAppID, operation, name)
+	if err := o.signing.VerifyPropagatedHistoryStateless(history); err != nil {
+		log.Warnf("Workflow actor '%s': workflow access policy denied app '%s' operation '%s' on '%s': propagated history verification failed: %v", o.actorID, callerAppID, operation, name, err)
+		diag.DefaultMonitoring.WorkflowACLActionDenied(callerAppID, string(workflowacl.OperationTypeWorkflow), string(operation))
+		return status.Errorf(codes.PermissionDenied, "%s: app '%s' operation '%s' on workflow '%s' (instance '%s')", workflowacl.DeniedMessageBase, callerAppID, operation, name, o.actorID)
+	}
+
+	allowed, reason := policies.Evaluate(callerAppID, workflowacl.OperationTypeWorkflow, operation, name, history)
+	if !allowed {
```
JoshVanL: Shouldn't we require that history signing be enabled in order to allow requires?

pkg/actors/targets/workflow/orchestrator/messages.go
```diff
@@ -154,10 +154,10 @@ func (o *orchestrator) callStateMessage(ctx context.Context, m proto.Message, hi
 		if isPermissionDenied(err) && historyEvent != nil {
 			if es := historyEvent.GetExecutionStarted(); es != nil {
 				if es.GetParentInstance() != nil {
-					if fErr := o.failChildWorkflowACL(ctx, es.GetParentInstance().GetTaskScheduledId(), err); fErr != nil {
+					log.Warnf("Workflow actor '%s': child workflow start denied by access policy on '%s': %v", o.actorID, method, err)
+					if fErr := o.failChildWorkflowACL(ctx, es.GetParentInstance().GetTaskScheduledId()); fErr != nil {
 						return fmt.Errorf("failed to record child workflow failure: %w (original: %v)", fErr, err)
 					}
-					return nil
```
JoshVanL, line 192: Why has this been removed?

pkg/actors/targets/workflow/orchestrator/childwf.go
```diff
@@ -125,22 +126,26 @@ func isPermissionDenied(err error) bool {
 	return false
 }
 
+func aclFailureType() (string, string) {
+	return "WorkflowAccessPolicyDenied",
+		"access denied by workflow access policy"
+}
+
 // failChildWorkflowACL creates a ChildWorkflowInstanceFailed event on the
```
JoshVanL: This already exists  in `accesspolicy.go`

pkg/acl/workflow/policy.go
```diff
@@ -128,11 +169,167 @@ func (cp *CompiledPolicies) Evaluate(callerAppID string, opType OperationType, o
...
+		case wfaclapi.RequiredStatusCompleted:
+			// ExecutionCompleted carries no name on the event itself, so a
+			// name filter only matches a child-workflow completion.
+			if cwc := e.GetChildWorkflowInstanceCompleted(); cwc != nil {
+				cc, ok := childCreatedByID[cwc.GetTaskScheduledId()]
+				return ok && cc.GetName() == req.Name
+			}
+			return false
+		}
+
```
JoshVanL: I think we can match requires on a workflow name which is itself, which is likely not what we want or what the policy author wanted.

```
requires:
- eventType: workflow
  status: Started
  name: OrderWF        # the caller's own workflow name
  appID: caller-app
```

pkg/acl/workflow/policy.go
```diff
@@ -128,11 +169,167 @@ func (cp *CompiledPolicies) Evaluate(callerAppID string, opType OperationType, o
...
+				dc.taskScheduledByID[e.GetEventId()] = ts
+			}
+			if cc := e.GetChildWorkflowInstanceCreated(); cc != nil {
+				dc.childCreatedByID[e.GetEventId()] = cc
+			}
+		}
+		chunks = append(chunks, dc)
+	}
+	return chunks, true
+}
```
JoshVanL: Skip decoding a chunk if the app ID is not in the requires.

JoshVanL: Try this to improve the maps

```go
// Lookups are scoped per-chunk: event IDs are only unique within their producer.
type decodedChunk struct {
      chunk  *protos.PropagatedHistoryChunk
      events []*protos.HistoryEvent

      // Built lazily on the first Completed lookup; nil until then. Started/Raised
      // requirements never consult these, so a requires list with no Completed
      // entry never allocates them.
      taskScheduledByID map[int32]*protos.TaskScheduledEvent
      childCreatedByID  map[int32]*protos.ChildWorkflowInstanceCreatedEvent
}

// indexByID populates the per-chunk ID maps on first use. Cheap no-op once built.
func (dc *decodedChunk) indexByID() {
      if dc.taskScheduledByID != nil {
              return
      }
      dc.taskScheduledByID = make(map[int32]*protos.TaskScheduledEvent)
      dc.childCreatedByID = make(map[int32]*protos.ChildWorkflowInstanceCreatedEvent)
      for _, e := range dc.events {
              if ts := e.GetTaskScheduled(); ts != nil {
                      dc.taskScheduledByID[e.GetEventId()] = ts
              }
              if cc := e.GetChildWorkflowInstanceCreated(); cc != nil {
                      dc.childCreatedByID[e.GetEventId()] = cc
              }
      }
}

// Fails closed: unmarshal failure means a tampered/corrupted chunk.
func decodeHistoryChunks(history *protos.PropagatedHistory) ([]decodedChunk, bool) {
      if history == nil {
              return nil, true
      }
      chunks := make([]decodedChunk, 0, len(history.GetChunks()))
      for _, chunk := range history.GetChunks() {
              if chunk == nil {
                      continue
              }
              dc := decodedChunk{chunk: chunk}
              for _, raw := range chunk.GetRawEvents() {
                      e := &protos.HistoryEvent{}
                      if err := proto.Unmarshal(raw, e); err != nil {
                              log.Warnf("WorkflowAccessPolicy: malformed raw event in tAppId(), err)
                              return nil, false
                      }
                      dc.events = append(dc.events, e)
              }
              chunks = append(chunks, dc)
      }
      return chunks, true
}

// Iterate by index so the lazy indexByID() build persists on the cached chunk.
func findMatchingEvent(req *wfaclapi.RequiredEvent, chunks []decodedChunk) bool {
      for i := range chunks {
              dc := &chunks[i]
              if dc.chunk.GetAppId() != req.AppID {
                      continue
              }
              for _, e := range dc.events {
                      if eventStatusMatches(req, e, dc) {
                              return true
                      }
              }
      }
      return false
}

// eventStatusMatches reports whether e is the kind of history event described by req.
func eventStatusMatches(req *wfaclapi.RequiredEvent, e *protos.HistoryEvent, dc *decodedChunk) bool {
      switch req.EventType {
      case wfaclapi.RequiredEventTypeActivity:
              switch req.Status {
              case wfaclapi.RequiredStatusStarted:
                      ts := e.GetTaskScheduled()
                      return ts != nil && ts.GetName() == req.Name
              case wfaclapi.RequiredStatusCompleted:
                      tc := e.GetTaskCompleted()
                      if tc == nil {
                              return false
                      }
                      dc.indexByID()
                      ts, ok := dc.taskScheduledByID[tc.GetTaskScheduledId()]
                      return ok && ts.GetName() == req.Name
              }

      case wfaclapi.RequiredEventTypeWorkflow:
              switch req.Status {
              case wfaclapi.RequiredStatusStarted:
                      if es := e.GetExecutionStarted(); es != nil && es.GetName() == req.Name {
                              return true
                      }
                      if cwc := e.GetChildWorkflowInstanceCreated(); cwc != nil && cwc
                              return true
                      }
                      return false
              case wfaclapi.RequiredStatusCompleted:
                      if cwc := e.GetChildWorkflowInstanceCompleted(); cwc != nil {
                              dc.indexByID()
                              cc, ok := dc.childCreatedByID[cwc.GetTaskScheduledId()]
                              return ok && cc.GetName() == req.Name
                      }
                      return false
              }

      case wfaclapi.RequiredEventTypeEvent:
              if req.Status != wfaclapi.RequiredStatusRaised {
                      return false
              }
              er := e.GetEventRaised()
              return er != nil && er.GetName() == req.Name
      }
      return false
}
```

pkg/apis/workflowaccesspolicy/v1alpha1/types.go
```diff
@@ -149,17 +156,35 @@ const (
...
+	// +kubebuilder:validation:MaxItems=8
 	// +kubebuilder:validation:items:Enum=schedule;terminate;raise;pause;resume;purge;get;rerun
 	// +listType=set
 	Operations []WorkflowOperation `json:"operations"`
+
+	// Requires is a list of history events that must all be present in the
+	// caller's propagated workflow history for this rule to apply. The order of
+	// entries is not significant, and the events are not required to have
+	// occurred in any particular order. Only valid when the rule's only
+	// operation is `schedule`.
```
JoshVanL: Please make ordering a requirement. This becomes far more valuable when ordering is respected.

JoshVanL (changes requested):

pkg/apis/workflowaccesspolicy/v1alpha1/types.go
```diff
@@ -168,7 +193,64 @@ type WorkflowRule struct {
...
+//
+// +kubebuilder:validation:XValidation:rule="self.eventType == 'event' ? self.status == 'Raised' : self.status != 'Raised'",message="eventType=event requires status=Raised; eventType=activity|workflow forbids status=Raised"
+type RequiredEvent struct {
+	// EventType is the category of history event to match.
+	// +kubebuilder:validation:Enum=activity;workflow;event
+	EventType RequiredEventType `json:"eventType"`
+
+	// Status is the lifecycle phase the matched event must have.
+	// +kubebuilder:validation:Enum=Started;Completed;Raised
+	Status RequiredStatus `json:"status"`
```
JoshVanL: Collapse these into just `EventType`. The Status is redundant, it is based on the type of event.

JoshVanL (changes requested):

pkg/actors/targets/workflow/orchestrator/childwf.go
cicoyle: I may have missed this from having the PR open for so long, thx will rm 👍🏻

pkg/actors/targets/workflow/orchestrator/messages.go
cicoyle: I did this on accident

pkg/actors/targets/workflow/orchestrator/childwf.go
cicoyle: actually I double checked and Im consolidating them. That err msg was in a few places, so Im consolidating them. From: 
https://github.com/dapr/dapr/blob/master/pkg/actors/targets/workflow/activity/accesspolicy.go#L26
https://github.com/dapr/dapr/blob/master/pkg/actors/targets/workflow/orchestrator/accesspolicy.go#L29
https://github.com/dapr/dapr/blob/master/pkg/api/grpc/workflow_acl.go#L28

To just this one location:
dapr/pkg/actors/targets/workflow/orchestrator/accesspolicy.go

pkg/acl/workflow/policy.go
cicoyle: fixed. workflow.started now matches only child wfs so the rule doesnt get satisfied by the callers own execution.

pkg/actors/targets/workflow/orchestrator/accesspolicy.go
```diff
@@ -43,60 +64,77 @@ func (o *orchestrator) checkAccessPolicy(ctx context.Context, method string, dat
...
-	name, err := o.workflowNameForOperation(ctx, method, data, preLoadedMeta)
+	name, history, err := o.workflowNameForOperation(ctx, method, data, preLoadedMeta)
 	if err != nil {
 		log.Errorf("Workflow actor '%s': failed to resolve workflow name for policy check on '%s': %v", o.actorID, method, err)
 		return status.Error(codes.Internal, "failed to evaluate workflow access policy")
 	}
 
-	if !policies.Evaluate(callerAppID, workflowacl.OperationTypeWorkflow, operation, name) {
-		log.Warnf("Workflow actor '%s': workflow access policy denied app '%s' operation '%s' on '%s'", o.actorID, callerAppID, operation, name)
+	if err := o.signing.VerifyPropagatedHistoryStateless(history); err != nil {
```
acroca: Isn't this verification also called in the `execute.go`? Are we verifying it twice? Maybe same happens in the activity check.

pkg/actors/targets/workflow/orchestrator/accesspolicy.go
```diff
@@ -43,60 +64,77 @@ func (o *orchestrator) checkAccessPolicy(ctx context.Context, method string, dat
 	// Self-calls are exempt: the policy is a cross-app gate.
 	callerAppID := workflowacl.CallerAppID(md)
 	if callerAppID == o.appID {
+		if policies.ListsCaller(o.appID) {
```
acroca: We could skip this check if the warning has already been sent.

tests/integration/suite/daprd/workflow/accesspolicy/accesspolicy.go
```diff
@@ -0,0 +1,18 @@
+/*
+Copyright 2024 The Dapr Authors
```
acroca:
```suggestion
Copyright 2026 The Dapr Authors
```

acroca (dismissed):

acroca (approved):

PR #9890, opened 2026-05-05
Workflow: cryptographically verify history propagation

JoshVanL:
Extends workflow history signing and child/activity attestation to cover propagated history. Previously, when A scheduled B and A's signed history was forwarded to C as IncomingHistory under PropagateLineage, C had no way to detect a tampered or forged payload because verifySignatureChain only covers state.History, not the propagated-history state-store key.

Each PropagatedHistoryChunk now carries a fresh chunk-local signature + DER cert chain produced by the chunk's app at dispatch time. Lineage chunks pass through verbatim - the original producer's bytes travel with the chain. Receivers verify every chunk against:

- chain-of-trust to a Sentry trust anchor (cached per cert digest via the existing per-orchestrator certVerifyCache)
- leaf SPIFFE ID app component matches chunk.appId
- signatures form a contiguous chain over the chunk's events
- bounds, empty-with-sigs, missing-sigs, missing-certs branches rejected explicitly

Verified foreign certs are absorbed into the ext-sigcert table so downstream attestation lookups can content-address them.


pkg/runtime/wfengine/state/state.go
```diff
@@ -61,13 +61,15 @@ var wfLogger = logger.NewLogger("dapr.runtime.actor.target.workflow.state")
...
 	AppID             string
+	Namespace         string
 	WorkflowActorType string
 	ActivityActorType string
 	Signer            *signer.Signer
 }
 
 type State struct {
 	appID             string
+	Namespace         string
```
cicoyle: I cant tell why is this exported? Am I missing a ref for it, or should it be unexported?

pkg/actors/targets/workflow/activity/execute.go
```diff
@@ -50,6 +50,18 @@ func (a *activity) executeActivity(ctx context.Context, name string, invocation
 	}
 	workflowID := a.actorID[0:endIndex]
 
+	// Cryptographically verify any propagated history before letting the
+	// activity see it. Activities are stateless workers with no
+	// ext-sigcert table to absorb certs into, so this is a verify-or-
+	// reject gate. No-op when signing is disabled. On failure, abort
+	// activity execution: the caller (parent workflow) gets a recoverable
+	// error and the activity never runs.
+	if a.signing != nil && invocation.GetPropagatedHistory() != nil {
```
cicoyle: What happens in a mixed environment? The gate here is `a.signing != nil`, so when the receiver has signing disabled, any attached signatures on incoming propagated history are ignored - not verified, not logged and the payload is accepted as-is. The lineage chunks get forwarded verbatim downstream, signatures and all.

An app with signing disabled loses protection for any signed payload flowing in from a signing enabled peer app, with no signal that something configured upstream was unverified. Should the receiver log a warning when carrying non-empty `RawSignatures` and signing is off? Or fail closed?

Either way, can we add tests for both directions: signing-disabled receiver & signing-enabled sender, and vice versa?

pkg/actors/targets/workflow/orchestrator/signing/propagation.go
```diff
@@ -0,0 +1,138 @@
...
+//
+// Failure: returns an error and absorbs nothing. Caller must reject the
+// inbound message (do not persist or run).
+//
+// No-op when Signer is nil. Each chunk owns its own rawEvents and signing
+// material; verification rejects mismatches such as empty rawEvents with
+// signatures attached, or signatures that do not chain over the chunk's
+// rawEvents.
+func (s *Signing) VerifyAndAbsorbPropagatedHistory(ph *protos.PropagatedHistory, state *wfenginestate.State) error {
+	if s.Signer == nil || ph == nil {
```
cicoyle: I think these should be separate cases. `ph == nil` means there's nothing to do. But `s.Signer == nil` with `ph != nil` is interesting to think about. When signing is disabled on the receiver, the receiver doesn't verify signatures regardless of whether the sender attached them and current behavior silently accepts without verifying. An app with signing disabled loses protection for any signed payload flowing in, with no log line.

Should we at minimum log a warning when chunks carry non-empty `RawSignatures` and signing is `disabled`? Or fail closed and reject? Same question applies to `VerifyPropagatedHistoryStateless` func.

pkg/actors/targets/workflow/orchestrator/signing/propagation.go
```diff
@@ -0,0 +1,138 @@
...
+		}
+
+		rawEvents := chunk.GetRawEvents()
+		if len(rawEvents) == 0 {
+			return fmt.Errorf("own-app propagated chunk has empty rawEvents (app %q)", chunk.GetAppId())
+		}
+
+		// Single-batch fresh signature. StartEventIndex is 0 (chunk-
+		// local indexing) and there is no previous signature.
+		res, err := historysigning.Sign(s.Signer, historysigning.SignOptions{
```
cicoyle, line 154: Should `InstanceId` and wf `Name` be bound into the signature input so they can't be swapped after signing? Right now they sit on the chunk outside the signed bytes, so a legit chunk could be lifted and relabeled to a different instance under the same app and still verify.

cicoyle (changes requested):

tests/integration/suite/daprd/workflow/propagation/signed/activity.go
```diff
@@ -0,0 +1,95 @@
...
+
+func (s *activity) Setup(t *testing.T) []framework.Option {
+	s.workflow = procworkflow.New(t,
+		procworkflow.WithDaprds(2),
+		procworkflow.WithMTLS(t),
+	)
+	return []framework.Option{framework.WithProcesses(s.workflow)}
+}
+
+func (s *activity) Run(t *testing.T, ctx context.Context) {
```
acroca: I don't see anything special in this test, am I missing something? Doesn't seem to be testing signing explicitly, it's just a standard workflow run, with history propagation. I guess it's something that happens behind the scenes and not something we can test explicitly, right?
In that case, aren't the existing tests good enough? This one feels it should be already covered elsewhere.
This apply for other tests that don't have any specific signing assertions...

pkg/runtime/wfengine/state/state.go
JoshVanL: :+1:

pkg/actors/targets/workflow/activity/execute.go
JoshVanL: Thanks, we are now passing `signing` though and the warning logs gets hit. Added an int test for a signing disabled receiver.

pkg/actors/targets/workflow/orchestrator/signing/propagation.go
JoshVanL: Done- 

`ph == nil` -> return nil

`ph != nil && s.Signer == nil` -> run new warnIfSigned(ph, caller)

pkg/actors/targets/workflow/orchestrator/signing/propagation.go
JoshVanL: Thanks- I agree. I have opened this PR up to cover this https://github.com/dapr/durabletask-go/pull/101

tests/integration/suite/daprd/workflow/propagation/signed/activity.go
JoshVanL: Yeah, fair enough
Removing

cicoyle (approved):

PR #9897, opened 2026-05-06
Actors: Clamp drain timeout to dissemination budget; honor per-entity drain

JoshVanL:
When the configured drainOngoingCallTimeout meets or exceeds the daprd-side placement dissemination timeout, drain holds the LOCK -> UPDATE -> UNLOCK round long enough that daprd resets its placement stream (e.g. an SDK that defaulted drain to 60s with the runtime's 30s dissemination default).

Add api.ClampDrainOngoingCallTimeout: when `drain >= dissemination`, log a warning and clamp to (`dissemination * drainTimeoutBudgetRatio`), floored at DefaultOngoingCallTimeout. Using a ratio (0.8) rather than a fixed margin keeps the proportion stable across small and large dissemination timeouts. Apply at both registration sites: the global drain in actors.RegisterHosted and the per-entity drain in api.TranslateEntityConfig. While there, fix two inverted-conditional bugs in TranslateEntityConfig that logged the warn-then-default branch on parse success instead of failure.


pkg/actors/internal/placement/loops/disseminator/inflight/lock/lock.go
```diff
@@ -172,46 +178,64 @@ func (l *lock) handleAcquire(event *Acquire) {
...
+		wg.Go(func() {
+			timer := time.NewTimer(timeout)
+			defer timer.Stop()
+			for _, claim := range claims {
+				select {
+				case <-claim.Context.Done():
+				case <-timer.C:
+					log.Errorf("Timed out waiting for actor in-flight lock claims to be released for rebalanced types, force cancelling remaining claims")
+					for _, claim := range claims {
+						claim.Cancel(event.Error)
```
acroca, line 233: Q: It's possible the context of some of the claims is done at this point, and we're cancelling them. For example, claim1 could be done, but claim2 times out. Is this fine?

pkg/actors/internal/placement/loops/disseminator/inflight/lock/lock.go
```diff
@@ -172,46 +178,64 @@ func (l *lock) handleAcquire(event *Acquire) {
...
 		}
+
+		wg.Go(func() {
+			timer := time.NewTimer(timeout)
+			defer timer.Stop()
+			for _, claim := range claims {
+				select {
+				case <-claim.Context.Done():
+				case <-timer.C:
+					log.Errorf("Timed out waiting for actor in-flight lock claims to be released for rebalanced types, force cancelling remaining claims")
```
acroca: This log line can potentially trigger multiple times now. I think we should either ensure it shows only once, or add the actor type so the N lines are different.

acroca (commented):
Both tests only cover happy path. I think it's fine as they make sure the clamp is happening, but do you think we should also test the timeout scenario?

pkg/actors/internal/placement/loops/disseminator/inflight/lock/lock.go
JoshVanL: Yeah this is fine- context cancellation is idempotent. It's not worth checking if there are already cancelled imo

pkg/actors/internal/placement/loops/disseminator/inflight/lock/lock.go
JoshVanL: Updated the log line to output the type which is being cancelled- this is actually quite helpful to understand which one is being problematic :)

acroca (dismissed):

cicoyle (approved):

PR #9902, opened 2026-05-07
Implement detached workflows

### LINKED ISSUE #9261: [Workflows] Add first-class support to trigger new, orphaned child workflows from a parent
## In what area(s)?

> /area runtime

## Describe the feature
Something that I ran into recently:  I wanted to start a workflow with its history and any failures separately from the one invoking it.  The solution ended up being that I had to create an activity that only has the purpose of calling the Dapr SDK to begin another workflow.

Main two issues with this I feel are:

- It makes for some tricky reading when going over the workflow history because I now have activities named something like `trigger-other-name-workflow`
- Language-specific code and semantics boilerplate, duplication and ambiguity where an activity _"looks like a workflow"_
- It makes spawning detached child workflows seem more like a side effect
- There's a missed opportunity for Dapr to track the invocation as a first class entry in the parent workflows history, making it more explicit, rather than burying it in an activity

This is most likely to come up during fan-outs, especially in a multi-tenant scenario:
You can imagine a workflow that spins through all active tenants/organizations/customers in an application that then goes through and triggers off individual operations.
Of utmost importance, one would want hard isolation between the tenants to be enforceable at a code and standards-and-practices level.  You might also want other types of downstream work to be isolated from one another as well.  Sky is the limit really in terms of when someone might want to fork that history chain off into a new one.

Additionally, it might be nice to come up with or identify an existing formal name for this concept. Detached, orphaned, whatever... 😆 

## Release Note

RELEASE NOTE: **ADD** Support to spawn disconnected/detached/orphaned workflows from an existing one

acroca:
Integrates [dapr/durabletask-go#100](https://github.com/dapr/durabletask-go/pull/100)
Adds first-class support for spawning detached (decoupled) workflows from a running workflow via `ctx.ScheduleNewWorkflow(...)`.

Closes #9261.

## Why

Today, spawning an independent workflow requires wrapping `WorkflowClient.ScheduleNewWorkflow` inside an activity. That bloats history with bookkeeping activities, hides intent, and conflates "side effect" and "spawn" semantics. With this change, an orchestration can spawn a fully decoupled instance with one call — no parent linkage, no completion or failure flow-back.

## Runtime changes

`pkg/actors/targets/workflow/orchestrator/`:
- **`dispatch.go`** — `DetachedWorkflowInstanceCreated` is now a dispatchable event.
- **`messages.go`** — `callCreateWorkflowStateMessage` builds an `instanceID → action.Id` index from `rs.NewEvents` to correlate detached spawn dispatch failures back to the caller's audit event (the spawned `StartEvent` carries `EventId=-1` and no `ParentInstance`).
- **`messages.go`** — when a detached spawn is rejected by the target's `WorkflowAccessPolicy`, the parent is failed terminally with `FailureDetails{ErrorType: "WorkflowAccessPolicyDenied"}`. The orchestrator has already returned synchronously from `ScheduleNewWorkflow`, so the parent's terminal status is the only operator-visible signal — silent dropping would hide policy violations and retrying is fruitless.
- **`fork/fork.go`** — explicit `InvalidArgument` when a `DetachedWorkflowInstanceCreated` event is targeted by a rerun (no replayable continuation).
- **`run.go`** — passes `rs.GetNewEvents()` into the new correlation path; routes `detachedSpawnDeniedError` to the parent-fail path before recoverable-retry fallback.

Recursive purge / terminate, history signing, and access-policy evaluation already do the right thing without changes — locked in by tests.


go.mod
```diff
@@ -528,7 +528,8 @@ replace (
 //
 // replace github.com/dapr/components-contrib => ../components-contrib
 // replace github.com/dapr/kit => ../kit
-// replace github.com/dapr/durabletask-go => ../durabletask-go
+replace github.com/dapr/durabletask-go => github.com/acroca/dapr-durabletask-go v0.0.0-20260507092108-f4b60dc9f7c8
```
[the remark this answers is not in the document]
acroca: This is temporary until durabletask-go PR is merged.

pkg/actors/targets/workflow/orchestrator/fork/fork.go
```diff
@@ -165,6 +165,11 @@ func (f *Fork) handleBefore(his *backend.HistoryEvent) {
 		delete(f.unfinishedChildWorkflows, his.GetChildWorkflowInstanceFailed().GetTaskScheduledId())
 
 	default:
+		// DetachedWorkflowInstanceCreated falls into this branch: detached
+		// spawns are fire-and-forget, so there is no completion or failure to
+		// pair them with — copying the audit event into the new history is
+		// the right thing to do, mirroring how a freshly client-scheduled
+		// workflow has no upstream linkage on rerun.
```
JoshVanL:
```suggestion
```

pkg/actors/targets/workflow/orchestrator/messages.go
```diff
@@ -19,19 +19,118 @@ import (
...
+	instanceID  string
+	targetAppID string
+	cause       error
+}
+
+func (e *detachedSpawnDeniedError) Error() string {
+	return fmt.Sprintf("detached workflow spawn '%s' denied by access policy on app '%s': %v", e.instanceID, e.targetAppID, e.cause)
+}
+
+func (e *detachedSpawnDeniedError) Unwrap() error { return e.cause }
```
JoshVanL: Move to new `errors` sub package?

pkg/actors/targets/workflow/orchestrator/messages.go
```diff
@@ -19,19 +19,118 @@ import (
...
+	}
+
+	state.ApplyRuntimeStateChanges(rs)
+	if err := o.signAndSaveState(ctx, state); err != nil {
+		return fmt.Errorf("failed to persist terminal failure for denied detached spawn: %w", err)
+	}
+
+	log.Warnf("Workflow actor '%s': failing parent terminally because detached workflow spawn was denied by access policy: %v", o.actorID, denyErr)
+	return nil
+}
```
JoshVanL: Move to new `detachesd.go` file

pkg/actors/targets/workflow/orchestrator/messages.go
```diff
@@ -19,19 +19,118 @@ import (
...
+
+// failParentDetachedDenied marks the parent orchestration as terminally
+// FAILED in response to a detached spawn that was rejected by the target
+// app's WorkflowAccessPolicy. The orchestrator function may have already
+// staged an ExecutionCompletedEvent (success) for this run; we override
+// it with a FAILED variant so the persisted terminal state reflects the
+// fact that the orchestration could not fulfill its scheduling intent.
+// Returns nil so the caller can mark the reminder as completed and stop
+// retrying — the denial is permanent.
+func (o *orchestrator) failParentDetachedDenied(ctx context.Context, state *wfenginestate.State, rs *backend.WorkflowRuntimeState, denyErr *detachedSpawnDeniedError) error {
```
JoshVanL:
```suggestion
func (o *orchestrator) failParentDetachedWorkflowACL(ctx context.Context, state *wfenginestate.State, rs *backend.WorkflowRuntimeState, denyErr *detachedSpawnDeniedError) error {
```

pkg/actors/targets/workflow/orchestrator/messages.go
```diff
@@ -19,19 +19,118 @@ import (
...
+	failureDetails := &protos.TaskFailureDetails{
+		ErrorType:    "WorkflowAccessPolicyDenied",
+		ErrorMessage: denyErr.Error(),
+	}
+	completed := &protos.ExecutionCompletedEvent{
+		WorkflowStatus: protos.OrchestrationStatus_ORCHESTRATION_STATUS_FAILED,
+		FailureDetails: failureDetails,
+	}
```
JoshVanL: Move to `events/events.go`

JoshVanL (changes requested):

go.mod
```diff
@@ -528,7 +528,8 @@ replace (
 //
 // replace github.com/dapr/components-contrib => ../components-contrib
 // replace github.com/dapr/kit => ../kit
-// replace github.com/dapr/durabletask-go => ../durabletask-go
+replace github.com/dapr/durabletask-go => github.com/acroca/dapr-durabletask-go v0.0.0-20260508084356-8d054d2aff65
```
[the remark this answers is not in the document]
acroca: As said earlier, will remove it after merging https://github.com/dapr/durabletask-go/pull/100

pkg/actors/targets/workflow/orchestrator/detached.go
```diff
@@ -0,0 +1,79 @@
...
+)
+
+// failParentDetachedWorkflowACL marks the parent orchestration as
+// terminally FAILED in response to a detached spawn that was rejected by
+// the target app's WorkflowAccessPolicy. The orchestrator function may
+// have already staged an ExecutionCompletedEvent (success) for this run;
+// we override it with a FAILED variant so the persisted terminal state
+// reflects the fact that the orchestration could not fulfill its
+// scheduling intent. Returns nil so the caller can mark the reminder as
+// completed and stop retrying — the denial is permanent.
```
JoshVanL: Do we want the parent workflow to fail because of a denied detached workflow? Surely since it is detached we don't consider this failed?

JoshVanL (changes requested):

pkg/actors/targets/workflow/orchestrator/detached.go
acroca: I think it's important to at least ensure the detached workflow is allowed to run. Otherwise the users might assume their detached workflow started correctly but failed silently and was not even created.

acroca: We had a sync chat and decided to do fire-and-forget, as it's what we think the user would expect.

JoshVanL (dismissed):

JoshVanL (approved):

PR #9923, opened 2026-05-09
placement: fix dissemination timeout not firing for slow mid-round connections

### LINKED ISSUE #9885: [1.18] placement: dissemination timeout does not fire when slow replica connects mid-round, leaving stream undisconnected
Int test logs https://github.com/dapr/dapr/actions/runs/25386690605/job/74450635775?pr=9731
PR seeing this: https://github.com/dapr/dapr/pull/9731

## Describe the bug

When a slow replica connects after a dissemination round has started and never responds with a LOCK ack, the placement service's `--disseminate-timeout` does not fire to force-disconnect that slow stream. The integration test `Test_Integration/placement/timeout/slowreplica` asserts the slow stream is disconnected after the configured timeout and currently fails: `stream2 (slow) was not disconnected`.

## To Reproduce

Run the integration test:

```bash
make test-integration ARGS="-run Test_Integration/placement/timeout/slowreplica"
```

The test configures placement with `--disseminate-timeout=2s` and expects the slow stream to be disconnected ~2s after a round it doesn't ack. Instead nothing fires for 10s and the test gives up.

## Expected behavior

After `--disseminate-timeout` elapses on a round that hasn't completed, `handleTimeout` should close streams that have not reached the current target operation. The slow stream should be disconnected.

## Actual behavior

The dissemination timeout for the round started by the slow replica's `REPORT` (which advanced currentVersion via `doReport`) never fires. The slow stream stays connected until the test framework SIGINTs placement.

## Logs

Test output excerpt:
\`\`\`
slowreplica.go:160: 
    Error: stream2 (slow) was not disconnected
    Test:  Test_Integration/placement/timeout/slowreplica/run
--- FAIL: Test_Integration/placement/timeout/slowreplica/run (10.10s)
\`\`\`

Placement log timeline (annotated):
\`\`\`
56:23.393827  Dissemination of version 1 in default complete
56:23.393855  Received status report connection from new namespace=default id=slow-app
56:23.393917  Received report from stream (idx:1) (ns=default) (appID=slow-app) (op=REPORT) (ver=1)
56:23.393930  Sending LOCK for version 2 to stream default:0   <-- round 2 starts
56:23.393983  Sending LOCK for version 2 to stream default:1
56:23.394293  Received report from stream (idx:0) (ns=default) (appID=healthy-app) (op=LOCK) (ver=2)
[ no further events for ~10s — neither healthy-app moves on without slow-app, nor does the disseminate-timeout fire to evict slow-app ]
56:33.399    Received signal 'interrupt'; beginning shutdown
\`\`\`

Expected: at \`56:25.394\` (2s after round 2 started) \`handleTimeout\` should fire and close slow-app's stream. It never does.

## Suspected root cause

In \`pkg/placement/internal/loops/disseminator/host.go:doReport\`, when a new host's REPORT triggers a new round:

\`\`\`go
d.timeoutQ.Dequeue(d.currentVersion)
d.currentVersion++
d.timeoutQ.Enqueue(d.currentVersion)
\`\`\`

Two suspected causes:

1. **Wrong-direction filter in `handleTimeout`.** It returns early when `timeout.Version != d.currentVersion`. If `Enqueue` runs before the timer task fires, but the version has been incremented again later, the timer fires for an old version and gets ignored. Worth verifying the filter direction matches the scheduler intent.
2. **Race in Dequeue/Enqueue ordering** when REPORT advances the round mid-flight; the new version's timer may be the one being dequeued.

A targeted unit test for the disseminator's timeout queue under the "round advance during arrival" scenario would help pin this down.

## Steps the test exercises (for repro context)

- Start placement with \`--disseminate-timeout=2s\`.
- Connect \`healthy-app\`; complete round 1.
- Connect \`slow-app\` (with no LOCK ack); round 2 starts.
- \`healthy-app\` acks LOCK v2. \`slow-app\` never does.
- Test waits up to 10s for the slow stream to be disconnected by the timeout.

## Environment

- Affects the placement disseminator's timeout queue path on master.
- Reproduces on macOS arm64 and Linux x86_64 CI runners.

## Additional context

Found while triaging an unrelated placement coalesce-timer fix. This bug is independent — the failing test does not configure \`--disseminate-coalesce-window\`, so the new coalesce-timer logic is dormant.

kimjune01:
`doReport` dequeues the old timeout before enqueuing the new one. A stale timer firing in the gap between those two calls gets rejected (wrong version), and now there's nothing armed for the current round. Swap the order so the new timeout is live before the old one is removed.

Fixes #9885

### DIFF
--- a/pkg/placement/internal/loops/disseminator/host.go
+++ b/pkg/placement/internal/loops/disseminator/host.go
@@ -89,9 +89,9 @@ func (d *disseminator) doReport(streamIDx uint64, host *v1pb.Host) {
 		return
 	}
 
-	d.timeoutQ.Dequeue(d.currentVersion)
 	d.currentVersion++
 	d.timeoutQ.Enqueue(d.currentVersion)
+	d.timeoutQ.Dequeue(d.currentVersion - 1)
 	d.currentOperation = v1pb.HostOperation_LOCK
 	d.streamsInTargetState = 0
 


pkg/placement/internal/loops/disseminator/host.go
```diff
@@ -89,9 +89,9 @@ func (d *disseminator) doReport(streamIDx uint64, host *v1pb.Host) {
 		return
 	}
 
-	d.timeoutQ.Dequeue(d.currentVersion)
 	d.currentVersion++
 	d.timeoutQ.Enqueue(d.currentVersion)
+	d.timeoutQ.Dequeue(d.currentVersion - 1)
```
JoshVanL, line 94: Thanks @kimjune01, but this is all in the same thread, so not sure this achieves anything? 🤔

JoshVanL (changes requested):

kimjune01:
You're right, I missed that this runs single-threaded. Closing — thanks for the pointer.

PR #9929, opened 2026-05-11
Workflow payload size metrics

JoshVanL:
The orchestrator stalls workflows whose payloads would exceed 95% of the gRPC server's max body size on the way to the SDK, but operators have no way to see how close payloads are to the threshold until a workflow actually trips it. By then the workflow is stuck and the only remediation is to raise --max-body-size and restart daprd or force purge the workflow.

Record the payload size as a fraction of the configured max body size on every workflow and activity dispatch from the same code path that performs the stall precheck, so the value fed into the threshold comparison is the value exported as a metric. A ratio (rather than an absolute byte count) keeps the metric portable across daprds with different --max-body-size settings, scales beyond the 4 GiB ceiling of the existing size distribution, and makes proximity-to-stall queries trivial: histogram_quantile(0.99, ...) > 0.95.

Two new histograms in DefaultWorkflowMonitoring:

- runtime/workflow/payload/size_ratio, tagged by app_id, namespace, workflow_name. Recorded once per runWorkflow dispatch.
- runtime/workflow/activity/payload/size_ratio, tagged by app_id, namespace, workflow_name, activity_name. Recorded once per activity dispatch, including the dispatch that trips the stall.

Both share a new payloadRatioDistribution with buckets concentrated near the 0.95 stall threshold (0.1, 0.25, 0.5, 0.75, 0.9, 0.95, 0.99, 1.0, 1.5, 2.0). Recording is skipped when --max-body-size is not configured, since the ratio is undefined without a limit.

Refactors workflowPayloadOversize and activityPayloadOversize to short-circuit on no-limit, then compute the size, record the ratio, and evaluate the stall threshold. Plumbs workflowName from runWorkflow and callActivities into both prechecks.


tests/integration/suite/daprd/metrics/workflowpayloadsize.go
```diff
@@ -0,0 +1,101 @@
+/*
+Copyright 2026 The Dapr Authors
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
```
cicoyle, line 6: I think `tests/integration/suite/daprd/metrics/pa` was unintentionally added and can be removed?

pkg/diagnostics/workflow_monitoring.go
```diff
@@ -300,3 +320,29 @@ func (w *workflowMetrics) AttestationCertCacheLookup(ctx context.Context, outcom
 		stats.WithTags(diagUtils.WithTags(w.attestationCertCacheCount.Name(), appIDKey, w.appID, namespaceKey, w.namespace, certCacheOutcomeKey, outcome)...),
 		stats.WithMeasurements(w.attestationCertCacheCount.M(1)))
 }
+
+// WorkflowPayloadSizeRatio records a workflow payload size as a fraction
+// of the configured gRPC max body size. Callers should skip recording
+// when no max body size is configured (ratio is undefined).
+func (w *workflowMetrics) WorkflowPayloadSizeRatio(ctx context.Context, workflowName string, ratio float64) {
+	if !w.IsEnabled() || ratio <= 0 {
```
cicoyle:
```suggestion
	if !w.IsEnabled() {
```
Should we drop the ` || ratio <= 0` bc thats still a valid data point, right? It just means the wf has no incoming history/no inputs so we should see 0 and record the metric anyways?

cicoyle (changes requested):

tests/integration/suite/daprd/metrics/workflowpayloadsize.go
JoshVanL: Yep 👍

pkg/diagnostics/workflow_monitoring.go
JoshVanL: Make sense 👍

cicoyle (approved):

PR #9940, opened 2026-05-12
MCPServer: correct workflow actor refcount

JoshVanL:
The wfengine MCP actor refcount could go negative or stay positive indefinitely under several failure paths, leaking actor type registrations or skipping unregistration. Hot-reload updates with a bad spec also left stale servers in compstore. Fixes four discrete bugs:

1. mcp_registry.closeAll read entry.holder without holding the entry mutex while register/unregister wrote it under the mutex, racing during the shutdown path.

2. wfe.UnregisterMCPServer always decremented mcpRegistrationCount, even when the inprocess registration never succeeded (e.g. IgnoreErrors=true bootstrap path where validation rejected the spec before any holder was installed). A later hot-reload delete drove the count to -1.

3. wfe.RegisterMCPServer left the inprocess registration in place when EnsureActorsRegistered failed, while EnsureActorsRegistered's own defer had already rolled its Add(1) back. A subsequent UnregisterMCPServer then decremented twice for the same server.

4. Executor.Run's ctx.Done path tore down inprocess holders during shutdown without notifying wfengine, so the actor refcount stayed positive and actor types were never unregistered.


famarting (approved):

sicoyle (approved):
lgtm - thank. you!!

cicoyle (approved):

PR #9943, opened 2026-05-13
workflow: fix lost activity results under transient errors

JoshVanL:
Three paths in pkg/actors/targets/workflow/orchestrator could leave a workflow stuck in RUNNING after a transient state-store or scheduler hiccup, because the in-memory cache and the wake-up reminder fell out of sync with the durable state.

Fixes:
* add.go: create the wake-up reminder BEFORE state.AddToInbox and the signAndSaveState that follows it. Failure of either step no longer leaves an orphan event in the cached inbox that a redelivery would dedup-drop. The reminder name is now derived deterministically from the event so retries collapse via overwrite-by-name instead of accumulating under random suffixes.
* run.go: same reorder and deterministic name for the continue-as-new carryover-save path.
* state.go: invalidate the cached o.state on any failed signAndSaveState so a follow-up loadInternalState reloads from the durable store.
* events/remindername.go (new): hosts the deterministic reminder-name builder. Keyed event types use the stable identifier on the event; non-keyed types use the event timestamp so identical retries still collapse while genuinely distinct events stay separate.


pkg/actors/targets/workflow/orchestrator/events/remindername.go
```diff
@@ -0,0 +1,63 @@
...
+	case *protos.HistoryEvent_ChildWorkflowInstanceFailed:
+		return fmt.Sprintf("%s-cwf-%d", prefix, evt.ChildWorkflowInstanceFailed.GetTaskScheduledId())
+	case *protos.HistoryEvent_EventRaised:
+		// Hash the user-supplied event name so the reminder name stays bounded and
+		// never contains scheduler-key separators (e.g. "||"). Empty names fall
+		// back to the event timestamp so genuinely distinct unnamed events still
+		// produce distinct reminders.
+		name := evt.EventRaised.GetName()
+		if name == "" {
+			return fmt.Sprintf("%s-er-%d", prefix, e.GetTimestamp().AsTime().UnixNano())
```
acroca, line 46: Will two events raised at the same time end up with the same name? Should we use some unique ID instead?
Same applies for other events using non-unique fields, like names or time.

pkg/actors/targets/workflow/orchestrator/events/remindername.go
```diff
@@ -0,0 +1,63 @@
...
+
+	"github.com/dapr/durabletask-go/api/protos"
+	"github.com/dapr/durabletask-go/backend"
+)
+
+// NewEventReminderName builds a deterministic name for the wake-up reminder
+// that drains the workflow inbox after an event is appended. Retries of the
+// same inbox append must collapse onto a single scheduler entry (the scheduler
+// overwrites by name) instead of accumulating under random suffixes.
+func NewEventReminderName(prefix string, e *backend.HistoryEvent) string {
```
acroca: Not sure `New...` is the right construct here, maybe just `EventReminderName`?

pkg/actors/targets/workflow/orchestrator/reminder.go
```diff
@@ -61,6 +61,44 @@ func (o *orchestrator) createRetentionReminder(ctx context.Context, name string,
 	})
 }
 
+// createWorkflowReminderWithName creates a reminder with a caller-supplied
+// deterministic name. Use this when retries must collapse onto a single
+// scheduler entry (overwrite-by-name) rather than piling up under random
+// suffixes.
+func (o *orchestrator) createWorkflowReminderWithName(ctx context.Context, reminderName string, data proto.Message, start time.Time, targetAppID string, concurrencyKey *string) error {
```
acroca: I'm wondering if we could have improved the existing `createWorkflowReminder` instead so from the caller side we don't need to worry about the logic behind the creation of the reminders... wdyt?

pkg/actors/targets/workflow/orchestrator/state.go
```diff
@@ -154,14 +154,25 @@ func (o *orchestrator) notifyStreams() {
...
 	}
-	return o.saveInternalState(ctx, state)
+	if err := o.saveInternalState(ctx, state); err != nil {
+		o.invalidateCachedState()
+		return err
+	}
+	return nil
+}
+
+func (o *orchestrator) invalidateCachedState() {
```
acroca, line 172: We have those 3 lines in the orchestrator.go and versioning.go, it'd be good if we could unify.

tests/integration/framework/process/scheduler/proxy/proxy.go
```diff
@@ -0,0 +1,299 @@
...
+// armConfig captures the per-method failure injection state.
+type armConfig struct {
+	remaining int
+	code      codes.Code
+	notify    chan struct{}
+}
+
+// New returns a proxy that wraps the given scheduler. The proxy must be
+// added to the framework's process list (after the scheduler) so it boots
+// once the upstream is reachable. daprd should be configured with
```
acroca: It seems it actually waits for it to be running, no? so order shouldn't matter right?

tests/integration/framework/process/scheduler/proxy/proxy.go
```diff
@@ -0,0 +1,299 @@
...
+				return
+			}
+			if serr := stream.Send(msg); serr != nil {
+				errCh <- serr
+				return
+			}
+		}
+	}()
+
+	err = <-errCh
```
javier-aliaga: Can both goroutine fail? If yes, should we read both errors from the channel?

tests/integration/framework/process/scheduler/proxy/proxy.go
```diff
@@ -0,0 +1,299 @@
...
+		return err
+	}
+
+	errCh := make(chan error, 2)
+
+	go func() {
+		for {
+			msg, rerr := stream.Recv()
+			if rerr != nil {
+				errCh <- rerr
```
javier-aliaga, line 247: if this fails on Recn() it will never send and the second goroutine will stay blocked, maybe is not a problem, but should we clean them?

pkg/actors/targets/workflow/orchestrator/run.go
```diff
@@ -226,16 +227,18 @@ func (o *orchestrator) runWorkflow(ctx context.Context, reminder *actorapi.Remin
...
+					if err = o.createWorkflowReminderWithName(ctx, reminderName, nil, time.Now(), o.appID, &workflowName); err != nil {
+						o.rstate = rstateSnapshot
 						return todo.RunCompletedFalse, err
 					}
 				}
+
+				if err = o.signAndSaveState(ctx, state); err != nil {
+					o.rstate = rstateSnapshot
+					return todo.RunCompletedFalse, err
+				}
```
javier-aliaga: It seems after this reorder the state has been mutated but not restored or saved if the reminder fails, like the rstate in line 233.

```
  - runWorkflow obtains state via loadInternalState (run.go:41), which returns o.state directly (state.go:40-41). Mutations to state are mutations to the cached o.state.
  - The CAN branch mutates state at lines 208-227 (ApplyRuntimeStateChanges, ClearInbox, AddToInbox(carryover), Generation++, SetIncomingHistory) before the reminder-create at line 232.
  - On reminder-create failure (line 234), only o.rstate = rstateSnapshot is restored. o.state retains the CAN mutations.
 ```

it makes sense to me, wdyt?

File: pkg/actors/targets/workflow/orchestrator/events/remindername.go (hunk 1 above)
Comment (JoshVanL) [reply pkg/actors/targets/workflow/orchestrator/events/remindername.go:46]: This is intentional. The reminder built here is a wake-up tick for the workflow inbox — it carries no event payload, it just causes runWorkflow to drain state.Inbox. If two distinct events end up sharing a reminder name (same event name, or empty name at the same timestamp), both events are still appended to the inbox and processed when the single wake-up fires.

File: pkg/actors/targets/workflow/orchestrator/run.go (hunk 8 above)
Comment (JoshVanL) [reply pkg/actors/targets/workflow/orchestrator/run.go]: On reminder-create failure in the ContinueAsNew carryover path, replaces `o.rstate = rstateSnapshot` with `o.invalidateCachedState()`. The ContinueAsNew state changes apply to the cached o.stat but not yet committed. Invalidation forces a reload from the store on the next access, restoring consistency.

File: tests/integration/suite/daprd/workflow/chaos/canfail.go [hunk 9]
```diff
@@ -0,0 +1,141 @@
...
+	activityStarted := make(chan struct{})
+	releaseActivity := make(chan struct{})
+
+	r := c.workflow.Registry()
+
+	require.NoError(t, r.AddActivityN("act", func(actx task.ActivityContext) (any, error) {
+		select {
+		case activityStarted <- struct{}{}:
+		default:
+		}
```
acroca: I don't think we want the default case

```suggestion
		activityStarted <- struct{}{}:
```

acroca: Same for the other tests

acroca (approved):

PR #9944, opened 2026-05-13
workflow: fix stuck-RUNNING under chaos

JoshVanL:
addWorkflowEvent's dedup branch returned nil without re-asserting the
new-event wake-up reminder. Under chaos an inbox row could be durable
while the matching reminder had been deleted (a prior empty-inbox
SUCCESS ack racing a partial save); retried activity reminders kept
hitting dedup, acking SUCCESS, and stranding the workflow in RUNNING.
Re-create the deterministic new-event-<kind>-<id> reminder on every
dedup hit, so retries collapse onto a single scheduler entry
(overwrite-by-name) and drain the inbox.

notifyStreams only signalled errCh when the stream callback returned
ok=true. Remote-routed WaitForWorkflowCompletion subscribers reach the
workflow actor via daprinternal CallActorStream, whose proxy wrapper
always returns ok=false: the COMPLETED snapshot reached the wire but
the subscriber was never released, and the upstream
WaitForWorkflowCompletion hung. Signal errCh on every terminal save
regardless of the callback's ok return. handleStream's initial-send
path gets the matching short-circuit so a wait that arrives after the
workflow is already terminal also exits cleanly.

Extracted assertNewEventReminder helper consolidating three call sites
(dedup branch, addWorkflowEvent main path, run.go CAN carryover).

Branched from https://github.com/dapr/dapr/pull/9943

### DIFF
--- a/pkg/actors/targets/workflow/orchestrator/add.go
+++ b/pkg/actors/targets/workflow/orchestrator/add.go
@@ -17,7 +17,6 @@ import (
 	"context"
 
 	"github.com/dapr/dapr/pkg/actors/targets/workflow/orchestrator/dedup"
-	"github.com/dapr/dapr/pkg/actors/targets/workflow/orchestrator/events"
 	wfenginestate "github.com/dapr/dapr/pkg/runtime/wfengine/state"
 	"github.com/dapr/durabletask-go/api"
 	"github.com/dapr/durabletask-go/backend"
@@ -62,11 +61,11 @@ func (o *orchestrator) addWorkflowEvent(ctx context.Context, e *backend.HistoryE
 
 	// Drop completion events whose resolution is already in history or the
 	// inbox; otherwise an inbox redelivery (e.g. an activity actor reminder
-	// firing twice during pod migration) would pin the workflow in a
-	// replay/spin loop.
+	// firing twice during pod migration) would pin the workflow in a replay/spin
+	// loop.
 	if dedup.IsDuplicateCompletion(e, state.History, state.Inbox) {
-		log.Debugf("Workflow actor '%s': dropping duplicate completion event already present in history/inbox", o.actorID)
-		return nil
+		log.Debugf("Workflow actor '%s': dropping duplicate completion event already present in history/inbox; re-asserting wake-up reminder so the inbox row is not stranded", o.actorID)
+		return o.assertNewEventReminder(ctx, e, state)
 	}
 
 	if e.GetTaskCompleted() != nil || e.GetTaskFailed() != nil {
@@ -111,15 +110,7 @@ func (o *orchestrator) addWorkflowEvent(ctx context.Context, e *backend.HistoryE
 	// source app. For cross-app events (e.g. ExecutionTerminated from a
 	// parent in another app), router.SourceAppID is the sender's app and
 	// would route the reminder to a non-existent remote actor.
-	sourceAppID := o.appID
-	dueTime := e.Timestamp.AsTime()
-	if len(state.History) > 0 {
-		dueTime = state.History[0].Timestamp.AsTime()
-	}
-	wfName := o.getExecutionStartedEvent(state).GetName()
-	reminderName := events.EventReminderName(reminderPrefixNewEvent, e)
-
-	if err := o.createWorkflowReminder(ctx, reminderName, nil, dueTime, sourceAppID, &wfName); err != nil {
+	if err := o.assertNewEventReminder(ctx, e, state); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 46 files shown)


tests/integration/suite/daprd/workflow/chaos/remotestream.go
```diff
@@ -0,0 +1,203 @@
...
+
+type remotestream struct {
+	workflow *workflow.Workflow
+}
+
+func (r *remotestream) Setup(t *testing.T) []framework.Option {
+	r.workflow = workflow.New(t, workflow.WithDaprds(2))
+	return []framework.Option{
+		framework.WithProcesses(r.workflow),
+	}
```
javier-aliaga, line 80: is this true? 

```
setup gives each of the 2 daprds a different default UUID appID, so they host different workflow actor types and no cross-daprd routing happens. The comment at line 85-87 ("Workflows
  hashed to daprd-1 will therefore exercise the cross-daprd CallActorStream proxy path") is the assumption the test rests on; it isn't satisfied. Compare with
  tests/integration/suite/daprd/workflow/loadbalance/config.go:38 which explicitly sets shared appID + WorkflowsClusteredDeployment. As written, this test passes with the bug present.
```

JoshVanL: Fixed 👍

pkg/actors/targets/workflow/orchestrator/add.go
```diff
@@ -62,11 +61,11 @@ func (o *orchestrator) addWorkflowEvent(ctx context.Context, e *backend.HistoryE
...
 	// inbox; otherwise an inbox redelivery (e.g. an activity actor reminder
-	// firing twice during pod migration) would pin the workflow in a
-	// replay/spin loop.
+	// firing twice during pod migration) would pin the workflow in a replay/spin
+	// loop.
 	if dedup.IsDuplicateCompletion(e, state.History, state.Inbox) {
-		log.Debugf("Workflow actor '%s': dropping duplicate completion event already present in history/inbox", o.actorID)
-		return nil
+		log.Debugf("Workflow actor '%s': dropping duplicate completion event already present in history/inbox; re-asserting wake-up reminder so the inbox row is not stranded", o.actorID)
+		return o.assertNewEventReminder(ctx, e, state)
```
yaron2, line 68: Could this create a retry storm upstream in cases where scheduler is degraded?

JoshVanL: @yaron2 we retry here for a minute before we give up and error out:

https://github.com/JoshVanL/dapr/blob/6326d44247b1bb7288df37cf97d274f47e62c451/pkg/actors/targets/workflow/common/retry.go#L45-L60

pkg/actors/targets/workflow/orchestrator/events/remindername.go
```diff
@@ -26,6 +26,14 @@ import (
 // that drains the workflow inbox after an event is appended. Retries of the
 // same inbox append must collapse onto a single scheduler entry (the scheduler
 // overwrites by name) instead of accumulating under random suffixes.
+//
+// EventRaised: a named user event keys on the SHA-256 of its name (truncated
```
cicoyle: Do we want this comment here on the event reminder name func?

pkg/actors/targets/workflow/orchestrator/add.go
cicoyle: +1, Wont this generate more load on an already degraded scheduler tho? I see the retry is for `Unavailable`/`DeadlineExceeded`. Can/should we maybe add some jitter or a longer backoff with a higher cap? The jitter could help so the retries arent all in sync. Or can we try another scheduler instance each time?

cicoyle (changes requested):

pkg/actors/targets/workflow/orchestrator/add.go
JoshVanL: We already have an exponential backoff here, and we have natural jitter by the nature of the actor IDs working concurrently- is it really necessary? 
We are already cycling scheduler instances lower in the stack- this is handled by the scheduler client transparently.

JoshVanL: Which settings would you recommend changing? 1 minute is already a long time for _all three_ scheduler instances to be unavailable

pkg/actors/targets/workflow/orchestrator/events/remindername.go
JoshVanL: Removed 👍

pkg/actors/targets/workflow/orchestrator/add.go
cicoyle: fair, I dont think there is much else to do then 👍🏻

cicoyle (approved):

PR #9949, opened 2026-05-14
Fix daprd shutdown API and add force option

JoshVanL:
The Shutdown HTTP and gRPC APIs were no-ops: the handler spawned a goroutine that waited on the request context's Done channel before calling the runtime shutdown function, but that cancellation does not reliably fire after the response is sent. Call the shutdown function directly in a goroutine instead, so graceful shutdown actually happens.

Add a "dapr-force-shutdown" gRPC metadata key (and matching "Dapr-Force-Shutdown" HTTP header), which causes daprd to skip graceful shutdown and exit via os.Exit(1) after a short delay so the response can flush. The HTTP handler bridges the header onto the gRPC metadata context so Universal.Shutdown keeps a single code path.


pkg/api/http/shutdown.go
```diff
@@ -15,13 +15,19 @@ package http
...
+	grpcMetadata "google.golang.org/grpc/metadata"
 
 	"github.com/dapr/dapr/pkg/api/http/endpoints"
+	"github.com/dapr/dapr/pkg/api/universal"
 	runtimev1pb "github.com/dapr/dapr/pkg/proto/runtime/v1"
 )
 
+// forceShutdownHeader is the HTTP header that maps onto the
+// universal.ForceShutdownMetadataKey gRPC metadata key.
+const forceShutdownHeader = "Dapr-Force-Shutdown"
```
cicoyle: I see both the metadata key and this header which are the same almost but one is `"Dapr-Force-Shutdown"` and the other is `"dapr-force-shutdown"` - can we have it once as a shared const in a transport file or similar?

cicoyle (changes requested):

pkg/api/http/shutdown.go
JoshVanL: Done 👍

JoshVanL:
Thanks @cicoyle please take another look

cicoyle (approved):

PR #9958, opened 2026-05-14
workflow: serialize saves with metadata ETag

JoshVanL:
Under placement leader churn, two daprds can briefly believe they each own the same workflow actor. Both load state at the same row version, each mutates a separate in-memory copy, and each writes back the result. With blind upserts on every history-NNNNNN row, the later write silently overwrote the earlier write's events at the same offset, leaving the workflow with a TaskScheduled and no matching TaskCompleted, frozen in RUNNING forever.

Workflow state saves now use optimistic concurrency on a single version anchor: the metadata row's ETag.

1. State.metadataETag is captured at load time from the metadata Get's ETag and cached alongside the rest of the state.

2. GetSaveRequest attaches the cached ETag to the metadata TransactionalUpsert. Other rows in the same Multi stay blind, but Multi is atomic, so a mismatch on metadata aborts the entire transaction without touching any row.

3. signAndSaveState classifies an ETagMismatch return as recoverable and invalidates the cache. The existing reminder-retry path re- fires the operation, the next load reads fresh state with the new ETag, and the retry succeeds against the up-to-date base state.

4. After a successful save, the orchestrator refreshes the metadata ETag via a single-key Get so the next save has the correct token without paying for a full state reload. MarkAsTamperFailed (the tombstone-on-tamper path) applies the same refresh after its direct write so a subsequent operation by the same actor does not collide with its own just-committed change.

api.StateResponse and api.BulkStateResponse now carry the per-key ETag the state store returned, so any actor-state caller performing optimistic concurrency has the token available without an extra round-trip. Existing callers that consume the response by .Data are unaffected; the ETag is an additive field.

---

Also cleans up logs under placement reconnect storms.


pkg/actors/internal/placement/connector/dnslookup/dnslookup.go
```diff
@@ -97,7 +97,7 @@ func (r *dnsLookUpConnector) Connect(ctx context.Context) (*grpc.ClientConn, err
 		return nil, err
 	}
 
-	log.Infof("Connected to placement %s", hostPort)
+	log.Debugf("Connected to placement %s", hostPort)
```
cicoyle, line 100: Should this remain `info` on success?

pkg/actors/internal/placement/loops/placement/placement.go
```diff
@@ -245,7 +255,16 @@ func (p *placement) handleCloseStream(ctx context.Context, closeStream *loops.Co
...
+	// then demote that path's reconnect line to debug so the cycle doesn't
+	// spam the log. The flag is read by handleReconnect to mirror the
+	// downgrade on the connect-side info lines.
+	p.lastCloseTransient = loops.IsTransientLeaderError(closeStream.Error)
+	if p.lastCloseTransient {
+		log.Debugf("Placement stream closed: %v. Reconnecting...", closeStream.Error)
+	} else {
+		log.Infof("Placement stream closed: %v. Reconnecting...", closeStream.Error)
+	}
 	return p.handleReconnect(ctx, new(loops.PlacementReconnect))
```
cicoyle: Should we reset the bool once we successfully reconnect? Or maybe pass it as an argument to the func?

cicoyle (changes requested):

pkg/actors/internal/placement/connector/dnslookup/dnslookup.go
JoshVanL: It's a duplicate of the placement loop  `Connected to placement service` log, which has the same address and stays at info on first/real connect (only demotes for transient "not a leader" reconnects). Don't think it makes sense to keep this at info

cicoyle (approved):

PR #9959, opened 2026-05-14
refactor(mcp): replace hand-defined MCP proto types with upstream go-sdk types

sicoyle:
# Description

Drop the proto messages I made for MCP spec messages in favor of using the official Go SDK types directly. We now carry MCP payloads as bytes of the spec JSON. 

I tried generating Go types from the published MCP JSON Schema first:
- `quicktype --src-lang schema --lang go` crashes on internal errors against MCP's `$ref` + `oneOf` patterns
- `go-jsonschema` produces a 3700-line flattened mega-struct with no `Type` discriminator field, so output JSON would not match the spec.

MCP's content unions use `oneOf` without an explicit discriminator property, so no off-the-shelf codegen produces spec-correct types. The upstream go-sdk has hand-written `MarshalJSON` per variant. Using it is the only path to spec-compliance without writing those marshalers ourselves.

This means SDKs can each use their MCP pkg types accordingly.

## Issue reference

## Checklist

- [ ] Code compiles correctly
- [ ] Created/updated tests
- [ ] Unit tests passing


pkg/proto/workflows/v1/mcp.pb.go
```diff
@@ -13,7 +13,7 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
 // 	protoc-gen-go v1.32.0
-// 	protoc        v7.34.1
+// 	protoc        v4.25.4
```
cicoyle: Can you update your version and regenerate the protos?

pkg/runtime/wfengine/inprocess/mcp/v1/worker.go
```diff
@@ -163,15 +161,10 @@ func DiscoverTools(ctx context.Context, holder *SessionHolder) ([]*wfv1.MCPToolD
 			return nil, fmt.Errorf("ListTools failed: %w", err)
 		}
 		for _, t := range result.Tools {
```
cicoyle: Can we do this here?:
```
tools = append(tools, result.Tools...)
```

cicoyle (changes requested):
mostly lgtm, just 2 comments

pkg/proto/workflows/v1/mcp.pb.go
sicoyle: oops yes

cicoyle (dismissed):

cicoyle (approved):

PR #9967, opened 2026-05-15
[1.17] scheduler: reconnect on WatchHosts first-Recv error instead of tearing down daprd

JoshVanL:
When the scheduler control plane is briefly unreachable at the moment daprd opens or re-opens its WatchHosts stream, the first Recv() can return a non-Canceled, non-Unimplemented status code (Unavailable, Internal, server-side Canceled, etc.). The runtime's WatchHosts loop returned that error directly, so it propagated up Scheduler.Run into the top-level RunnerCloserManager, which treats any runner exit as terminal and cancelled every sibling: the actor runtime, the workflow backend, both gRPC servers.

Now, always reconnect the stream to prevent going into zombie unready state.

### DIFF
--- a/pkg/runtime/scheduler/internal/watchhosts/watchhosts.go
+++ b/pkg/runtime/scheduler/internal/watchhosts/watchhosts.go
@@ -86,9 +86,7 @@ func (w *WatchHosts) Run(ctx context.Context) error {
 		}
 
 		resp, err := stream.Recv()
-		code := status.Code(err)
-		switch code {
-		case codes.Unimplemented:
+		if status.Code(err) == codes.Unimplemented {
 			if err = w.clients.Reload(ctx, w.allAddrs); err != nil {
 				return err
 			}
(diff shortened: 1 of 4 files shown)


pkg/runtime/scheduler/internal/watchhosts/watchhosts.go
```diff
@@ -102,14 +100,20 @@ func (w *WatchHosts) Run(ctx context.Context) error {
...
-			return err
+			if ctx.Err() != nil {
+				return ctx.Err()
+			}
+			log.Warnf("Scheduler WatchHosts stream error, reconnecting: %s", err)
+			select {
+			case <-ctx.Done():
+				return ctx.Err()
+			case <-time.After(time.Second):
+				continue
```
acroca, line 115: Should we limit the amount of reconnection attempts?

JoshVanL: Don't think so- to what end does that achieve?

cicoyle (approved):

PR #9968, opened 2026-05-15
scheduler: reconnect on WatchHosts first-Recv error instead of tearing down daprd

JoshVanL:
When the scheduler control plane is briefly unreachable at the moment daprd opens or re-opens its WatchHosts stream, the first Recv() can return a non-Canceled, non-Unimplemented status code (Unavailable, Internal, server-side Canceled, etc.). The runtime's WatchHosts loop returned that error directly, so it propagated up Scheduler.Run into the top-level RunnerCloserManager, which treats any runner exit as terminal and cancelled every sibling: the actor runtime, the workflow backend, both gRPC servers.

Now, always reconnect the stream to prevent going into zombie unready state.


acroca (commented):
LGTM, fix lint issues

cicoyle (changes requested):
same, lgtm just some linter issues

javier-aliaga (approved):

cicoyle (approved):

PR #9974, opened 2026-05-19
feat: search and vector blocks

mikeee:
# Description

## Issue reference

## Checklist

- [ ] Code compiles correctly
- [ ] Created/updated tests
- [ ] Unit tests passing


nelson-parente (commented):
I read the full diff at 1340575 minus the generated `*.pb.go`, built every touched package, and ran the universal tests. The protos and the HTTP routes match dapr/proposals#107 field for field. The wiring mirrors the Conversation block at every touch point I checked: registry, processor, compstore, allowlist groups, error codes, and resiliency on all 17 RPCs. The inline-close path even has a test. Four things to fix, one decision, then housekeeping.

**1. The `replace` to `github.com/mikeee/dapr_components-contrib` cannot merge.** It sits under the comment that says not to commit one, and it is why `go.mod` bumps sarama, the AWS SDK, tencentcloud and a dozen others this feature never touches. The path is: land dapr/components-contrib#4613, pin contrib to a commit or tag that includes it, drop the replace, run `make modtidy-all`. Copilot raised this too and the thread is still open.
https://github.com/dapr/dapr/blob/134057580d3e6c5cfd1b918f5a8cb8ec29cc2cfa/go.mod#L536-L546

**2. The integration harness pulls `getmeili/meilisearch:latest` from Docker Hub at test time.** Nothing else in `tests/integration/framework/process` reaches the network: sqlite is embedded, scheduler is built from this repo, kubernetes is an in-process mock. This helper is the first to depend on a registry, an unpinned tag, and a Docker daemon. A missing daemon skips the test, a failed pull fails it, and Windows is skipped outright. Two asks. Pin the tag to what the contrib PR tests against (`v1.43`). Then consider an in-memory search and vector component in contrib, the way `conversation.echo` and `state.in-memory` serve this suite. That keeps daprd's tests hermetic and leaves Meilisearch to contrib's conformance and certification runs. @JoshVanL owns this framework, so the second one is his call.
https://github.com/dapr/dapr/blob/134057580d3e6c5cfd1b918f5a8cb8ec29cc2cfa/tests/integration/framework/process/meilisearch/meilisearch.go#L52-L62
https://github.com/dapr/dapr/blob/134057580d3e6c5cfd1b918f5a8cb8ec29cc2cfa/tests/integration/framework/process/meilisearch/meilisearch.go#L93-L98

**3. Search and vector disagree on an empty query.** `QueryVectorsAlpha1` rejects a request with neither `vector` nor `by_id`. `SearchAlpha1` forwards a request with neither `text` nor `native` straight to the component, and the unit test at `search_test.go:270` relies on that. The contrib contract says "exactly one", the Meilisearch component treats empty as match-all, and its conformance suite depends on match-all. Pick one. If empty means match-all, say so in `search.proto` and in the contrib doc, and keep the runtime as is. If not, validate here the way vector does.
https://github.com/dapr/dapr/blob/134057580d3e6c5cfd1b918f5a8cb8ec29cc2cfa/pkg/api/universal/search.go#L265-L268
https://github.com/dapr/dapr/blob/134057580d3e6c5cfd1b918f5a8cb8ec29cc2cfa/pkg/api/universal/vector.go#L399-L401

**4. Batch queries silently overwrite the inner `collection` and never read the inner `store_name`.** The HTTP handler copies both onto every query, then the universal layer replaces `Collection` with the batch value. A caller who sets a different collection on one query gets results from another collection with no error. This is Cassie's open thread on the proposal. Either reject a non-empty inner value that differs from the batch, or drop the two fields from the inner message while the API is alpha.
https://github.com/dapr/dapr/blob/134057580d3e6c5cfd1b918f5a8cb8ec29cc2cfa/pkg/api/universal/vector.go#L292-L296
https://github.com/dapr/dapr/blob/134057580d3e6c5cfd1b918f5a8cb8ec29cc2cfa/pkg/api/http/vector.go#L217-L221

**5. `respondWithError` now renders any gRPC status error with the `google.rpc.Code` name as `errorCode`.** That puts `NOT_FOUND` next to `ERR_STATE_STORE_NOT_FOUND` in the same JSON field across the whole HTTP API. Today only search and vector produce bare status errors, and pluggable components convert theirs before this point, so nothing else changes behavior. It is still a new HTTP error contract, so it needs a sentence in the docs and a note in this description.
https://github.com/dapr/dapr/blob/134057580d3e6c5cfd1b918f5a8cb8ec29cc2cfa/pkg/api/http/responses.go#L180-L186

**Coverage.** The suites cover request validation well. They do not touch four things the protos document: a missing index or collection, the continuation token lifecycle, a vector whose length differs from the collection's dimensions, and `INDEXING_MODE_WAIT_FOR_COMPLETION` with either timeout action. The wait-for-completion path is the one most likely to differ per provider.

**Housekeeping.**
- Six of the eight Copilot threads are fixed at this head. Resolve them so the two live ones stand out.
- No `SearchInvoked` or `VectorInvoked` metrics exist in `pkg/diagnostics`. The proposal defers observability to beta, which is fine, but the description needs to say that.
- The description is still the empty template and there is no dapr/docs PR. Both are needed before merge.

PR #9994, opened 2026-05-21
fix(ci): use Hyper-V Gen 1-compatible Windows VM size for AKS test pool

nelson-parente:
## Summary

Every `dapr-test` scheduled run since #9983 merged (2026-05-21 00:08) has failed at "Deploy the test cluster", e.g. https://github.com/dapr/dapr/actions/runs/26225340722/job/77170747378. The Windows AKS node pool (`akswinpol`) cannot be created:

```
BadRequest: Create or update VMSS akswinpol failed.
The selected VM size 'Standard_D4ads_v6' cannot boot Hypervisor Generation '1'.
If this was a Create operation please check that the Hypervisor Generation of the Image
matches the Hypervisor Generation of the selected VM Size.
```

## Why

#9983 (continuation of #9971) bumped `windowsVMSize` from `Standard_DS3_v2` (Gen 1 + Gen 2) to `Standard_D4ads_v6` (**Gen 2 only**). The Linux + ARM bumps to `*_v6` are fine because their osSKUs support Gen 2 images. The Windows pool, however, uses `osSKU: 'Windows2022'` which AKS provisions as a **Hyper-V Gen 1 image** — so a Gen-2-only VM size cannot boot it.

## Fix

Switch to `Standard_D4s_v5`:
- Supports both Hyper-V Gen 1 **and** Gen 2 — compatible with the existing `Windows2022` osSKU
- Same vCPU / memory as `Standard_D4ads_v6` (4 vCPU, 16 GiB)
- Modern, widely-available SKU (not on Azure's near-term deprecation list)

Also expanded the param `@description` to call out the Gen 1 constraint so the next person picking a SKU avoids the same trap.

## Alternatives considered

- **Keep `D4ads_v6` and switch osSKU to a Gen 2 Windows image.** AKS's stable `osSKU` enum (`Windows2019`, `Windows2022`, `WindowsAnnual`) all resolve to Gen 1 images — there is no public `osSKU` value that forces Gen 2 today. Would require an AKS API/version upgrade plus careful image testing; out of scope for a same-day fix.
- **Revert to `Standard_DS3_v2`.** Mike picked `_v6` intentionally because the old SKUs are being deprecated; reverting would just re-hit the original deprecation issue.

## Test plan

- [ ] dapr-test scheduled run completes "Deploy the test cluster" successfully
- [ ] Windows node pool reaches Ready in AKS
- [ ] Subsequent e2e jobs that depend on the Windows pool unblock

## References

- Last green dapr-test (any conclusion): 2026-02-11 (>3 months)
- Regression introduced: #9983 (this PR's parent commit)
- Failing run example: https://github.com/dapr/dapr/actions/runs/26225340722/job/77170747378

cc @mikeee


cicoyle (dismissed):

nelson-parente:
running here: https://github.com/dapr/dapr/actions/runs/26234356579 🤞

nelson-parente:
Update: the first attempt picked `Standard_D4s_v5`, but a workflow_dispatch run against this branch ([26233473860](https://github.com/dapr/dapr/actions/runs/26233473860)) showed `_v5` is **not allowed** in the test subscription's westus quota:

> The VM size of Standard_D4s_v5 is not allowed in your subscription in location 'westus'.

Inspecting the preflight "available VM sizes" list:
- **All D4 `_v6` SKUs** in this region are Gen-2-only (no Gen-1 capability)
- **No D4 `_v5` SKU** is on the allow list at all
- Only Gen-1-capable D4 SKUs available: `Standard_D4_v3` and `Standard_D4s_v3`

Pushed `fc5ea59`: switch to `Standard_D4s_v3` (4 vCPU, 16 GiB, Gen 1 + Gen 2, premium SSD — matches the pool's `osDiskType: 'Ephemeral'`). Updated the param `@description` to record why `_v5` was ruled out.

nelson-parente:
Second update: the run on \`fc5ea59\` ([26234356579](https://github.com/dapr/dapr/actions/runs/26234356579)) hit yet another preflight error:

> VMCannotFitEphemeralOSDisk: Standard_D4s_v3 has a cache size of 100 GiB and temporary disk size of 32 GiB, but the OS disk requires 128 GiB. Use a VM size with larger cache, larger temp disk, or disable ephemeral OS.

The original \`Standard_DS3_v2\` got away with ephemeral OS because it carried a 172 GiB local SSD. No Gen-1-compatible D4-class SKU available in the test subscription has cache or temp >= 128 GiB. The smallest size up that would satisfy ephemeral (\`Standard_D8s_v3\`, 200 GiB cache) doubles compute cost across two Windows nodes — wasteful for a short-lived CI cluster.

Pushed \`140b096\`: drop \`osDiskType: 'Ephemeral'\` on the \`winpol\` agent pool so AKS provisions a managed OS disk instead. Latency overhead is negligible for a 30-min CI cluster and the cost delta is single-digit dollars per cluster lifetime. Left an explanatory comment in the bicep so the next person reading it doesn't re-add ephemeral without checking the SKU constraint.

nelson-parente:
https://github.com/dapr/dapr/actions/runs/26235181578

nelson-parente:
> https://github.com/dapr/dapr/actions/runs/26235181578

"Deploy test infrastructure" ✅

nelson-parente:
https://github.com/dapr/dapr/actions/runs/26237844417 on https://github.com/dapr/dapr/pull/9994/commits/034bfd44ab55e8db4f48765e17731d5f7ec7a5e5

nelson-parente:
Re-ran `dapr-test` against this branch: https://github.com/dapr/dapr/actions/runs/26280788289

**The fix works.** For the first time since 2026-02-11, `Deploy test infrastructure` succeeds — the Windows AKS node pool boots with `Standard_D4s_v5`. ✅

The run still ends red, but on **pre-existing failures unrelated to this PR** — they've been hidden for 3+ months because the cluster never deployed. None of them touch the Windows VM SKU or anything in `tests/test-infra/azure-aks.bicep`:

| Test | Failure | Likely cause |
|---|---|---|
| `TestCrypto/{grpc,http}/azurekeyvault/*` + `TestSubtleCrypto/{grpc,http}/azurekeyvault/*` (linux + windows) | KV calls return 0/non-200; ciphertext empty | Azure Key Vault not reachable — same Azure subscription/infra migration cc @mikeee is driving |
| `TestPubSubHTTP/built-in_publish_with_subscriber_error_test_redelivery_of_messages_http` (linux + windows) | `subscriber did not receive all dead letter messages within timeout` after ~398s | Dead-letter redelivery flake or backend timing |
| `TestWorkflowRetentionPolicy` (linux) | `workflow should have been purged by retention policy, but still returned: Completed` | Workflow retention not purging — product/test bug |
| `TestDaprSslCertInstallation` (windows injector) | expected 200, got 500 | Windows-specific injector issue |

## Proposal

Merge this PR as-is so master gets past the deploy step. Then file follow-ups for each failure category — they need separate investigation and (in the AKV case) the broader Azure infra migration to land first. I'll open the follow-up issues once this merges.

javier-aliaga (approved):

cicoyle (approved):

cicoyle:
Ill fix the dep review issue in a follow up PR 👍🏻

PR #9995, opened 2026-05-21
[1.18] ci(version-skew): patch v1.17 framework with actors-disseminate-timeout

nelson-parente:
## What

Adds `004-clamp-disseminate-timeout.patch` to the `dapr-sidecar-master` version-skew patch set for `release-1.17`. The patch backports `daprd.WithActorsDisseminateTimeout` into the v1.17 test framework and updates the three actor drain clamp tests to use a 4s disseminate timeout — mirroring the equivalent change already present on master/release-1.18.

## Why

The Version Skew workflow has been red on `release-1.18` (and `master`) since 2026-05-12, exclusively on the `dapr-sidecar-master` axis. The failing tests are:

- `Test_Integration/actors/dissemination/drain/clamp/clamp`
- `Test_Integration/actors/dissemination/drain/clamp/clampentity`
- `Test_Integration/actors/dissemination/drain/clamp/timeout`

All three time out with:
> `call should be cancelled after the clamped drain timeout, not the configured 60s`

### Root cause

The clamp feature ([#9897](https://github.com/dapr/dapr/pull/9897)) was backported to v1.17.7 via [#9916](https://github.com/dapr/dapr/pull/9916), so both v1.17.7 and 1.18 daprd implement the clamp. After v1.17.7 shipped, however, the master/release-1.18 copies of `clamp.go`, `clampentity.go`, and `timeout.go` picked up a change adding `daprd.WithActorsDisseminateTimeout(time.Second*4)` to both daprds in each test.

The version-skew runner mounts the v1.17.7 source into `latest-release/` and runs **that** test source against the master daprd binary. The v1.17.7 test source doesn't include the timeout override, so placement is launched with its default `--disseminate-timeout=30s`. The 80% clamp lands at ~24s while the test's `assert.Eventually(..., 20*time.Second, ...)` deadline is 20s — so the clamp fires *after* the deadline and the assertion fails.

### The fix

This patch teaches the v1.17.7 framework about `WithActorsDisseminateTimeout` (adds the option field, the `With…` helper, and the corresponding `--actors-disseminate-timeout` arg generation in `daprd.go`), then updates the three clamp tests to pass a 4s timeout to both daprds. The clamp now lands at ~3.2s, well inside the 20s test deadline.

The `timeout.go` test also expects the per-type log line (`"Timed out waiting for actor type 'abc' in-flight lock claims to be released"`) that master/release-1.18 daprd emits; this patch updates the expected substring accordingly.

## Test plan

- [ ] Version Skew workflow against this branch goes green on the `dapr-sidecar-master` axis for v1.17.7
- [ ] `control-plane-master` axis remains green
- [ ] Other patches in the same directory continue to apply cleanly

Locally verified: `git apply --check --ignore-space-change --ignore-whitespace` succeeds against a fresh v1.17.7 checkout, and `go build ./tests/integration/suite/actors/dissemination/drain/clamp/...` compiles cleanly after the patch is applied.

### DIFF
--- a/.github/scripts/version-skew-test-patches/integration/release-1.17/dapr-sidecar-master/004-clamp-disseminate-timeout.patch
+++ b/.github/scripts/version-skew-test-patches/integration/release-1.17/dapr-sidecar-master/004-clamp-disseminate-timeout.patch
@@ -1,0 +1,126 @@
+diff --git a/tests/integration/framework/process/daprd/daprd.go b/tests/integration/framework/process/daprd/daprd.go
+index a2836a509..df440617d 100644
+--- a/tests/integration/framework/process/daprd/daprd.go
++++ b/tests/integration/framework/process/daprd/daprd.go
+@@ -152,6 +152,9 @@ func New(t *testing.T, fopts ...Option) *Daprd {
+ 	if len(opts.schedulerAddresses) > 0 {
+ 		args = append(args, "--scheduler-host-address="+strings.Join(opts.schedulerAddresses, ","))
+ 	}
++	if opts.actorsDisseminateTimeout != nil {
++		args = append(args, "--actors-disseminate-timeout="+opts.actorsDisseminateTimeout.String())
++	}
+ 	if opts.controlPlaneTrustDomain != nil {
+ 		args = append(args, "--control-plane-trust-domain="+*opts.controlPlaneTrustDomain)
+ 	}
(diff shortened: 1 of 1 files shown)


nelson-parente:
Closing in favour of an upstream branch on `dapr/dapr` so the Version Skew workflow can be triggered via `workflow_dispatch`. Same content, just relocated.

PR #9996, opened 2026-05-21
[1.18] ci(version-skew): patch v1.17 framework with actors-disseminate-timeout

nelson-parente:
Branch lives on the upstream repo so the Version Skew workflow can be triggered manually via `workflow_dispatch`. This obsoletes #9995 (same diff) and is an alternative to #9978 which still showed `clamp/timeout` failing in its skew run.

## What

Adds `004-clamp-disseminate-timeout.patch` to the `dapr-sidecar-master` version-skew patch set for `release-1.17`. The patch backports `daprd.WithActorsDisseminateTimeout` into the v1.17 test framework and updates all three actor drain clamp tests (`clamp.go`, `clampentity.go`, `timeout.go`) to set a 4s disseminate timeout — mirroring the equivalent change already present on master/release-1.18.

## Why

The Version Skew workflow has been red on `release-1.18` (and `master`) since 2026-05-12, exclusively on the `dapr-sidecar-master` axis. The failing tests are:

- `Test_Integration/actors/dissemination/drain/clamp/clamp`
- `Test_Integration/actors/dissemination/drain/clamp/clampentity`
- `Test_Integration/actors/dissemination/drain/clamp/timeout`

All three time out with:
> `call should be cancelled after the clamped drain timeout, not the configured 60s`

### Root cause

The clamp feature (#9897) was backported to v1.17.7 via #9916, so both v1.17.7 and 1.18 daprd implement the clamp. After v1.17.7 shipped, master/release-1.18 picked up updates to `clamp.go`, `clampentity.go`, and `timeout.go` that add `daprd.WithActorsDisseminateTimeout(time.Second*4)` to both daprds in each test.

The version-skew runner mounts the v1.17.7 source into `latest-release/` and runs **that** test source against the master daprd binary. The v1.17.7 test source doesn't include the timeout override, so placement is launched with its default `--disseminate-timeout=30s`. The 80% clamp lands at ~24s while the test's `assert.Eventually(..., 20*time.Second, ...)` deadline is 20s, so the clamp fires *after* the deadline and the assertion fails.

### The fix

This patch teaches the v1.17.7 framework about `WithActorsDisseminateTimeout` (adds the option field, the `With…` helper, and the corresponding `--actors-disseminate-timeout` arg generation in `daprd.go`), then updates all three clamp tests to pass a 4s timeout to both daprds. The clamp now lands at ~3.2s, well inside the 20s test deadline.

The `timeout.go` test also expects the per-type log line (`"Timed out waiting for actor type 'abc' in-flight lock claims to be released"`) that master/release-1.18 daprd emits; this patch updates the expected substring accordingly.

### Difference from #9978

#9978 takes the same approach but only updates the log-line string in `timeout.go` — it does **not** add `WithActorsDisseminateTimeout` to the daprds in that file. #9978's most recent Version Skew run still showed `clamp/timeout` failing with `expected to log line: Timed out waiting for actor in-flight lock claims to be released` (the old string), suggesting either that hunk didn't apply or the timeout test still needs the explicit timeout option. This PR matches the master `timeout.go` 1:1.

## Test plan

- [ ] Trigger Version Skew via workflow_dispatch against this branch
- [ ] Confirm `dapr-sidecar-master` integration leg goes green for v1.17.7
- [ ] Confirm `control-plane-master` axis remains green
- [ ] Confirm `clamp/timeout` no longer fails on the OLD log-line expectation
- [ ] Other patches in the same directory continue to apply cleanly

Locally verified: `git apply --check --ignore-space-change --ignore-whitespace` succeeds against a fresh v1.17.7 checkout, and `go build ./tests/integration/suite/actors/dissemination/drain/clamp/...` compiles cleanly after the patch is applied.


nelson-parente:
https://github.com/dapr/dapr/actions/runs/26235727154

nelson-parente:
green here 💚 
https://github.com/dapr/dapr/actions/runs/26283493893

.github/workflows/version-skew.yaml
```diff
@@ -245,11 +245,11 @@ jobs:
     - name: Set up for Dapr lastest release
       run: |
         echo "DAPR_PREV_VERSION=$(curl -s https://api.github.com/repos/dapr/dapr/releases/latest | jq -r '.tag_name' | cut -c 2-)" >> $GITHUB_ENV
-    - name: Set up for master
+    - name: Set up for push/dispatch
       if: github.event_name != 'repository_dispatch'
       run: |
         echo "CHECKOUT_REPO=${{ github.repository }}" >> $GITHUB_ENV
-        echo "CHECKOUT_REF=refs/heads/master" >> $GITHUB_ENV
+        echo "CHECKOUT_REF=${{ github.ref }}" >> $GITHUB_ENV
```
JoshVanL, line 252: Why are we doing this?

JoshVanL (changes requested):

JoshVanL:
@nelson-parente version skew (and all new changes really) are merged into master and backported to v1.18- please can we do that here

.github/workflows/version-skew.yaml
nelson-parente: so I could run via workflow dispatch as I didn't had the perms with the bot

nelson-parente: I'm fine with reverting :)

PR #10012, opened 2026-05-26
fix: always start pubsub span and propagate context in subscription r…

### LINKED ISSUE #9990: PubSub: no root span created when inbound message has no traceparent (breaks observability for DLQ re-queues / external publishers)
## In what area(s)?

/area runtime

## Expected Behavior

When the sidecar delivers a pub/sub message whose inbound CloudEvent carries **no** W3C trace context (no `traceparent`, no legacy `traceid`), Dapr should still start a **new root span** for the delivery, named `pubsub/<topic>`, and inject it into the metadata passed to the app. Every delivered message then has an end-to-end trace, whether or not the producer set trace headers.

## Actual Behavior

No span is created at all. The message is delivered normally, but the hop is invisible in tracing and the app receives an **empty** `traceparent` header — so it has nothing to continue from and nothing to cleanly detect as absent. The gap propagates through the whole downstream call chain, not just the one hop.

Span creation is gated on the trace context already existing, rather than falling back to a new root. There are **four independent copies** of this gate, pinned to `v1.18.4` because the line numbers drift:

| Gate | Delivery paths it covers |
|---|---|
| [`pkg/runtime/pubsub/subscriptions.go:389-408`](https://github.com/dapr/dapr/blob/v1.18.4/pkg/runtime/pubsub/subscriptions.go#L389-L408) (`GRPCEnvelopeFromSubscriptionMessage`) | gRPC single ([`postman/grpc/grpc.go:75`](https://github.com/dapr/dapr/blob/v1.18.4/pkg/runtime/subscription/postman/grpc/grpc.go#L75)) and streaming subscriptions ([`pubsub/streamer/streamer.go:189`](https://github.com/dapr/dapr/blob/v1.18.4/pkg/runtime/pubsub/streamer/streamer.go#L189)) |
| [`postman/grpc/grpc.go:195-207`](https://github.com/dapr/dapr/blob/v1.18.4/pkg/runtime/subscription/postman/grpc/grpc.go#L195-L207) | gRPC bulk |
| [`postman/http/http.go:83-93`](https://github.com/dapr/dapr/blob/v1.18.4/pkg/runtime/subscription/postman/http/http.go#L83-L93) | HTTP single |
| [`postman/http/http.go:242-254`](https://github.com/dapr/dapr/blob/v1.18.4/pkg/runtime/subscription/postman/http/http.go#L242-L254) | HTTP bulk |

All four share the same shape:

```go
iTraceID := cloudEvent[contribpubsub.TraceParentField]
if iTraceID == nil {
    iTraceID = cloudEvent[contribpubsub.TraceIDField]
}

if traceID, ok := iTraceID.(string); ok {          // <-- gate
    ctx, span = diag.StartInternalCallbackSpan(ctx, "pubsub/"+msg.Topic, sc, spec)
}
```

When neither field is present the branch is skipped, `span` stays nil, and every later `if span != nil` guard silently no-ops.

1.18 added a guard on some paths for a *non-string* trace id, which logs at debug level and still creates no span. The nil case remains silent everywhere.

Secondary issues in the same area, unchanged from the original report:

- [`subscriptions.go:396`](https://github.com/dapr/dapr/blob/v1.18.4/pkg/runtime/pubsub/subscriptions.go#L396) silently discards `SpanContextFromW3CString` parse errors — a malformed `traceparent` yields an invalid `SpanContext` and the code continues as if it were valid.
- [`pkg/runtime/subscription/subscription.go:348`](https://github.com/dapr/dapr/blob/v1.18.4/pkg/runtime/subscription/subscription.go#L348) seeds the resiliency runner with `context.Background()`, severing cancellation/deadline (and span) propagation from the outer handler `ctx`. Not the proximal cause, but worth fixing alongside.

Dapr 1.18.4, self-hosted, Redis pub/sub, Zipkin, `samplingRate: "1"`. A subscriber on topic `orders` that logs its inbound `traceparent` header.

The Redis component uses the topic as the stream key and a `data` field, so a raw `XADD` is a faithful stand-in for any non-Dapr publisher — no cloud broker needed.

1. Control, published through Dapr:

```bash
curl -X POST http://localhost:3510/v1.0/publish/pubsub/orders \
  -H 'Content-Type: application/json' -d '{"orderId":"CASE1-via-dapr"}'
```

2. External publisher, straight to the broker with no trace context:

```bash
docker exec dapr_redis redis-cli XADD orders '*' data '{"specversion":"1.0","type":"com.dapr.event.sent","source":"external-system","id":"CASE2-no-traceparent","datacontenttype":"application/json","data":{"orderId":"CASE2-external"},"pubsubname":"pubsub","topic":"orders"}'
```

**Observed.** Both messages are delivered; only the control is traced:

```
APP-RECEIVED traceparent="00-20fa547b905bdb1cfa166aa0d13ad37e-a1dbda0979694042-01" body={... "traceparent":"00-20fa547b905bdb1cfa166aa0d13ad37e-80a94f78e9645aab-01" ...}
APP-RECEIVED traceparent="" body={"specversion":"1.0","source":"external-system","id":"CASE2-no-traceparent", ...}
```

Zipkin holds one trace for the delivery path — the control's, a publish span with a `pubsub/orders` child:

```json
{"traceId":"20fa547b905bdb1cfa166aa0d13ad37e","id":"80a94f78e9645aab","name":"/v1.0/publish/pubsub/orders"}
{"traceId":"20fa547b905bdb1cfa166aa0d13ad37e","parentId":"80a94f78e9645aab","id":"a1dbda0979694042","name":"pubsub/orders"}
```

The external message produced no span.

This also reproduces via a manual DLQ re-queue on Azure Service Bus — moving messages from the DLQ back to the active queue through the portal/CLI drops the AMQP application properties carrying `traceparent` — which is how it was originally hit in production. The runtime code path is broker-agnostic.

## Proposed Fix

Call `StartInternalCallbackSpan` unconditionally and parse a parent only when one is present. `StartInternalCallbackSpan` ([`pkg/diagnostics/tracing.go:167`](https://github.com/dapr/dapr/blob/v1.18.4/pkg/diagnostics/tracing.go#L167)) already handles an invalid parent correctly; the callers never reach it.

```go
var sc trace.SpanContext
    sc, _ = diag.SpanContextFromW3CString(traceID)
} else if iTraceID != nil {
    log.Debugf("...")
}
ctx, span = diag.StartInternalCallbackSpan(ctx, "pubsub/"+msg.Topic, sc, spec)
```

Checked against a real SDK tracer provider — a zero parent yields a new root, a valid parent is still inherited:

```
zero parent  -> valid=true traceID=1518101967406ff7cd618e473a1e85b0 spanID=58e84f5a5f714031
valid parent -> traceID=20fa547b905bdb1cfa166aa0d13ad37e (inherited) spanID=c15543ca6e312712
spans exported: 2
  name="pubsub/orders" traceID=1518101967406ff7cd618e473a1e85b0 parentValid=false   <- new root
  name="pubsub/orders" traceID=20fa547b905bdb1cfa166aa0d13ad37e parentValid=true    <- child of remote parent
```

Applied at all four gates. Sampling is still governed by the tracing spec, so no spans are created where tracing is disabled.

## Release Note

RELEASE NOTE: **FIX** PubSub subscription delivery now starts a new root span when the inbound message has no `traceparent`, so externally-published or manually re-queued (DLQ) messages remain traceable end-to-end.

RituGupta23:
Fixes https://github.com/dapr/dapr/issues/9990

### 1. Pubsub tracing span not started when `traceparent` is absent

When an incoming pubsub message did not contain a `traceparent` or `traceID` in the CloudEvent, `StartInternalCallbackSpan` was never called.

As a result, messages without trace context were completely invisible to tracing, reducing observability for those pubsub flows.

### 2. Context cancellation not respected in subscription delivery

In `New()`, the resiliency `policyRunner` was created using `context.Background()` instead of the caller-provided `ctx`.

This meant that if the runtime shut down or the parent context was cancelled, in-flight delivery retries continued running instead of being cancelled properly, leading to delayed or unclean shutdown behavior.

---

### `pkg/runtime/pubsub/subscriptions.go`

- Moved `StartInternalCallbackSpan` outside the `if iTraceID != nil` block so a tracing span is always started.
- An empty `SpanContext` now correctly creates a new root span when no trace context is provided.
- Added a warning log when a `traceparent` is present but fails to parse, surfacing malformed trace context without blocking message processing.

### `pkg/runtime/subscription/subscription.go`

- Changed:

```go
resiliency.NewRunner[any](context.Background(), policyDef)
```

to:

```go
resiliency.NewRunner[any](ctx, policyDef)
```

- This ensures the caller’s context, including cancellation signals and deadlines, is correctly propagated into the delivery runner.

---

## Tests

Existing tests in:

- `pkg/runtime/subscription/postman/grpc/grpc_test.go`

already cover the tracing path through `GRPCEnvelopeFromSubscriptionMessage` via `Deliver()`, including scenarios with:

- a valid `traceparent`
- no `traceparent`

No new tests were added because the current coverage is sufficient for these changes.
### Existing tests are passing.


RituGupta23:
@dapr/maintainers-dapr @dapr/approvers-dapr Could u please review this PR?

pkg/runtime/pubsub/subscriptions.go
mikeee: If there is an issue with the extensions, you'll be left with an unended span.. I'd suggest re-ordering or ending the span (with a preference on the former).

mikeee (changes requested):
One issue I've noticed, also please also add tests, particularly surrounding the cancellation behaviour.

cicoyle (changes requested):
bump on this @RituGupta23 please - Can you add unit tests and integration tests for the changes?

alicejgibbons:
@RituGupta23 following up on this one.  Can you add unit tests and integration tests for the changes?

MyMirelHub:
Thanks @RituGupta23 contributing to original fix, and @nelson-parente for the tests and the reordering on top of it.

I've opened [dapr/dapr#10511](https://github.com/dapr/dapr/pull/10511) to carry this forward. Working through it, the same gate turned out to exist in four places rather than one: GRPCEnvelopeFromSubscriptionMessage covers gRPC single and streaming subscriptions, but the gRPC bulk, HTTP single and HTTP bulk postmen each carry their own copy and still skip span creation entirely. [dapr/dapr#10511](https://github.com/dapr/dapr/pull/10511) routes all four through one shared helper, and adds the unit and integration tests asked for here.

Both of your commits are cherry-picked into it unchanged, so the authorship is preserved. Review very welcome if you have time 

@dapr/maintainers-dapr happy for this one to be closed in favour of [dapr/dapr#10511](https://github.com/dapr/dapr/pull/10511).

PR #10013, opened 2026-05-26
tests: stabilize flaky CI tests

JoshVanL:
workflow_retention: anyTerminal 1s->5s, Eventually 60s->120s.

pubsub redelivery: error-mode sleep 30s->90s so retries exhaust to DLQ.

crypto: skip azurekeyvault subtests when sidecar returns 500.

hotreload selfhosted resiliency: retry os.Remove on Windows fsnotify lock.

version-skew: new 004 patch loosening clamp timings against v1.17.7 placement.


cicoyle (approved):
lgtm, just want one more green run before merging 👍🏻

PR #10019, opened 2026-05-27
Workflow: retentioner ignore not completed errors

JoshVanL:
Workflow retentioner should ignore not completed errors, as the workflow has been superseded by a newer workflow. This fixes a small race condition where the retentioner is trying to delete a workflow that has recently been superseded, and that new workflow will therefore recreate the retentioner reminder upon completion.

---

Workflow stuck and unrecoverable after re-scheduling an instance ID that has already completed

Problem

When an application schedules a workflow with a deterministic instance ID and a later request re-schedules the same ID after the previous run has reached a terminal state, the previous run's retention reminder can keep firing in the scheduler indefinitely instead of draining.

Impact

Any deployment running workflows that are scheduled with stable, repeatable instance IDs is affected whenever the same ID is scheduled again within the configured retention window. The most visible trigger is an at-least-once delivery against an idempotent handler that uses `REUSE_ID_ACTION_TERMINATE`, but plain `ScheduleNewWorkflow` calls against an already-completed instance ID follow the same path.

Visible symptoms include:

- A workflow appears stuck: subsequent re-scheduling attempts for the same ID do not run to completion on their own, and the workflow's status oscillates between non-terminal and terminal states without ever settling.
- The scheduler accumulates one retention reminder per affected workflow that does not drain after the configured retention TTL elapses.
- The `dapr_runtime_workflow_operation_count{operation=purge_workflow,status=failed}` metric increments at exactly one tick per second per affected workflow, for as long as the bug is active.
- The workflow recovers only after an operator manually deletes the retentioner job out of the scheduler.

Root Cause

When a workflow is scheduled with an instance ID that already corresponds to a completed run, the workflow runtime resets the existing state and starts a fresh run. The retention reminder queued by the previous run is anchored to the previous run's completion time and is intentionally left in the scheduler so it can fire on the previous run's terminal-state retention TTL.

When that retention reminder fires, the workflow has been put back into a non-completed state by the fresh run, so the workflow runtime's purge path correctly refuses to delete state that is in flight, and signals "this run has not completed". The retentioner treated that signal as a hard failure and re-raised it. The retention reminder is created with a failure policy of `Constant{Interval: 1s, MaxRetries: nil}` (retry every second, forever), so the scheduler then retried the same reminder once per second indefinitely. Only an operator-driven scheduler delete could clear it.

Solution

The retentioner now treats the "this run has not completed" signal the same way it already treats "this run no longer exists": as a drain-silently signal that the retention reminder is no longer applicable. Either the run has been purged (nothing to retain) or the run has been superseded by a fresh run (the fresh run will queue its own retention reminder on completion). Either way, the previous run's retention reminder is obsolete and the scheduler can let it drop.

Workflows already stuck in this state on existing 1.17 deployments recover automatically once the sidecar is upgraded to 1.17.8: the next retention reminder fire drains, the reminder is removed from the scheduler, and the workflow proceeds normally. No operator intervention or manual scheduler delete is required after the upgrade.


javier-aliaga (dismissed):

cicoyle (dismissed):
lgtm

JoshVanL:
Re-running int test: 

```
2026-05-27T14:08:09.8226783Z         	Error Trace:	/home/runner/work/dapr/dapr/tests/integration/framework/process/scheduler/proxy/proxy.go:111
2026-05-27T14:08:09.8227379Z         	            				/opt/hostedtoolcache/go/1.26.3/x64/src/sync/once.go:78
2026-05-27T14:08:09.8227959Z         	            				/opt/hostedtoolcache/go/1.26.3/x64/src/sync/once.go:69
2026-05-27T14:08:09.8228894Z         	            				/home/runner/work/dapr/dapr/tests/integration/framework/process/scheduler/proxy/proxy.go:96
2026-05-27T14:08:09.8229710Z         	            				/home/runner/work/dapr/dapr/tests/integration/framework/process/once/once.go:52
2026-05-27T14:08:09.8230445Z         	            				/home/runner/work/dapr/dapr/tests/integration/framework/framework.go:45
2026-05-27T14:08:09.8231109Z         	            				/home/runner/work/dapr/dapr/tests/integration/integration.go:94
2026-05-27T14:08:09.8231311Z         	Error:      	Received unexpected error:
2026-05-27T14:08:09.8231787Z         	            	listen tcp 127.0.0.1:13011: bind: address already in use
2026-05-27T14:08:09.8232098Z         	Test:       	Test_Integration/daprd/scheduler/watchhostsrecover
```

cicoyle (approved):

PR #10023, opened 2026-05-27
fix(mcp): gate /healthz/outbound on mcpserver wf registration

sicoyle:
# Description

The outbound healthz "app" target was marked ready before we loaded mcpservers so `/v1.0/healthz/outbound` returned.a 204 while the workflows were still being registered. Clients who saw this false-positive readiness signal who then went to schedle new workflow calls inside the window would get a CANCELLED response. This now blocks until after we wrap up MCP setup before allowing workflow calls to be invoked and clients to be signaled that all is ready.

## Issue reference

## Checklist

- [ ] Code compiles correctly
- [ ] Created/updated tests
- [ ] Unit tests passing


JoshVanL (changes requested):
> new workflow calls inside the window would get a CANCELLED response

I am surprised that we would get a CANCELLED here. I would have expected to see the schedule workflow call to hang until the workflow runtime was ready.

Can we add an int test for this? We shouldn't return app level error codes while dapr is still booting.

sicoyle:
adding the test rn @JoshVanL :)

JoshVanL:
@sicoyle e2e tests are failing on actor_sdk

pkg/runtime/runtime.go
```diff
@@ -830,6 +824,12 @@ func (a *DaprRuntime) initRuntime(ctx context.Context) error {
 
 	a.runtimeConfig.outboundHealthz.AddTarget("app").Ready()
 
+	if err := a.blockUntilAppIsReady(ctx); err != nil {
```
JoshVanL, line 827: Previously we were blocking until the app was ready before initialising actors- which itself will call into the app when app actors are enabled. Blocking for the app to be ready before initialising actors I think is intentional.

JoshVanL (changes requested):

pkg/runtime/runtime.go
sicoyle: initActors today only sets up infrastructure from what i see. The app-callback path is appHealthChanged → appHealthReadyInit → loadAppConfiguration/RegisterHosted, which still runs only after blockUntilAppIsReady. this move is needed to unblock SDK preflights that hit the workflow engine during app startup; reverting it re-introduces a deadlock seen in e2e tests a commit up.

JoshVanL: Makes sense, thanks

JoshVanL (approved):

PR #10053, opened 2026-06-02
test: validate pubsub CloudEvent IDs are UUIDs

### LINKED ISSUE #5962: E2E to validade CloudEvent UUID due to regression in 1.10
Why I think E2E tests in this case?

Although there is unit test for the regression found in 1.10 on cloudevent UUID, it is not uncommon for unit test to be modified or not cover problems in our releases. This is because unit test are targeting specific code paths and not the end-to-end feature and bugs can happen at multiple layers, causing the end user to see an issue even when all unit tests are passing. Having it as E2E test can cover the whole feature's vertical and not specific code paths. Plus, E2E tests are refactor-proof - meaning black-box testing targets the API contract and not the internal interfaces.

The person picking this up should use a version in master that does not contain the hotfix first to make sure the E2E test would catch it before rebasing into the latest master version.

https://github.com/dapr/dapr/issues/5959

SivagurunathanV:
Fixes #5962.

Adds E2E coverage for the Pub/Sub CloudEvent ID regression from Dapr 1.10.0 by capturing CloudEvent IDs received by the HTTP subscriber and asserting they are valid UUIDs.

Validation:
- go test ./tests/apps/pubsub-subscriber ./tests/apps/pubsub-publisher
- go test -c -tags=e2e ./tests/e2e/pubsub -o /private/tmp/dapr-pubsub-e2e.test


cicoyle (changes requested):
Hi thanks for the PR 🎉 

Mind addressing copilots feedback?

cicoyle:
bump on this @SivagurunathanV 🙏🏻

SivagurunathanV:
Missed it. I will work on the comments and publish a new PR.

PR #10059, opened 2026-06-03
fix(wfengine): do not unregister workflow actor types on SDK stream disconnect

javier-aliaga:
# Description

The OnGetWorkItemsDisconnect callback previously unregistered the workflow/activity/retentioner actor types from the local actor table whenever the SDK's work-item gRPC stream worker count hit zero. Under transient infrastructure pressure (network latency, gRPC keepalive timeout, app container restart) that stream churns and reconnects within milliseconds, but during the disconnect window:

  - the placement table still reports this daprd as hosting those actor types (placement has no visibility into the local table)
  - peer daprds keep routing dispatch calls here based on placement
  - every such dispatch fails locally with "actor type ... not registered"
  - the orchestrator's recoverable retries hit the same wall and workflows whose activity dispatch happened to land in the window strand indefinitely

The race is timing-sensitive. With a single brief disconnect a small fraction of workflows strands; under sustained stream churn the fraction grows toward 100%. Either way it requires the actor-table state to diverge from the placement view, which is the underlying defect.

Keep the actor-type registrations for the daprd's lifetime once established. UnRegisterActors still runs from the MCP teardown path and on daprd shutdown, which are the paths that should actually drop the types.

## Issue reference

## Checklist

- [ ] Code compiles correctly
- [ ] Created/updated tests
- [ ] Unit tests passing

### DIFF
--- a/pkg/runtime/wfengine/wfengine.go
+++ b/pkg/runtime/wfengine/wfengine.go
@@ -208,21 +208,23 @@ func New(opts Options) (Interface, error) {
 			wfe.actorRegLock.Lock()
 			defer wfe.actorRegLock.Unlock()
 
-			if ctx.Err() != nil {
-				ctx = context.Background()
-			}
+			wfe.getWorkItemsCount.Add(-1)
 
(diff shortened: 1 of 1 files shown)


javier-aliaga:
Close as we do not need it because https://github.com/dapr/dapr/pull/10058 already solves the problem

PR #10062, opened 2026-06-04
Respect resiliency retry matching on pub/sub publish errors

cicoyle:
Resiliency retry `matching` is currently ignored on the pubsub publish path - the retry runner only checks the configured status codes when the error is a `resiliency.CodeError`, but unlike the other runners (service invocation, gRPC proxy, bindings, subscriber delivery) the publish path returns the raw broker error, so the match never hits and every publish error retries up to `maxRetries` no matter what codes you configure.


atrauzzi:
Closes #10068

JoshVanL (approved):

PR #10072, opened 2026-06-09
Workflow: Harden external-event delivery

JoshVanL:
A workflow's wake-up reminder is created on demand when an event arrives and deleted once consumed, so a waiting workflow holds no standing reminder. When an external event (RaiseEvent) is saved to the inbox but the wake-up reminder's ScheduleJob then fails, the inbox row is left with no driver. Unlike an activity result, whose activity actor keeps a retry-forever reminder that re-asserts the wake-up, an external event has no such backstop, so the workflow stalls in RUNNING until (if ever) a timeout timer fires. Re-sending the same event under placement churn can also append it to history twice.

This adds the external-event equivalents of the durability activities already have:

- Retry the new-event wake-up reminder Create forever (bounded only by the actor context) instead of giving up after one minute on a non-transient code. The reminder name is deterministic, so retries are idempotent overwrite-by-name. New common.CreateReminderWithRetryForever retries every code except clearly-permanent request errors; wired into assertNewEventReminder via orchestrator.createWorkflowReminderForever.

- Dedup a redelivered external event. dedup.IsDuplicateExternalEvent matches an EventRaised already in history or the inbox by event name and ingestion timestamp: distinct RaiseEvents carry distinct timestamps and are never deduped, while a genuine redelivery (the same built event re-sent) is dropped and the wake-up reminder re-asserted, mirroring the completion dedup-and-reassert path.

### DIFF
--- a/tests/integration/suite/daprd/workflow/chaos/eventdedup.go
+++ b/tests/integration/suite/daprd/workflow/chaos/eventdedup.go
@@ -1,0 +1,133 @@
+/*
+Copyright 2026 The Dapr Authors
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package chaos
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"google.golang.org/protobuf/proto"
+
+	rtv1 "github.com/dapr/dapr/pkg/proto/runtime/v1"
+	"github.com/dapr/durabletask-go/api"
+	"github.com/dapr/durabletask-go/api/protos"
+	"github.com/dapr/durabletask-go/task"
+
+	"github.com/dapr/dapr/tests/integration/framework"
+	"github.com/dapr/dapr/tests/integration/framework/process/workflow"
+	"github.com/dapr/dapr/tests/integration/suite"
+)
+
+func init() {
+	suite.Register(new(eventdedup))
+}
+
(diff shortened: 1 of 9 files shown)


tests/integration/suite/daprd/workflow/chaos/eventdedup.go
```diff
@@ -0,0 +1,134 @@
...
+		ActorType: actorType,
+		ActorId:   wfID,
+		Method:    "AddWorkflowEvent",
+		Data:      dupBytes,
+	})
+	require.NoError(t, err)
+
+	// The redelivery must NOT complete the workflow: it is the same single
+	// event, so only the first wait is satisfied.
+	require.Never(t, func() bool {
```
acroca, line 117: How is this better than just waiting 3 seconds? To fail fast in case of failure?

tests/integration/suite/daprd/workflow/chaos/eventdedup.go
```diff
@@ -0,0 +1,134 @@
...
+			InstanceId:        wfID,
+			WorkflowComponent: "dapr",
+		})
+		return gerr == nil && resp.GetRuntimeStatus() == "COMPLETED"
+	}, 3*time.Second, 200*time.Millisecond,
+		"redelivered external event was applied twice and wrongly completed the workflow")
+
+	// A genuinely distinct RaiseEvent (new ingestion timestamp) is NOT deduped
+	// and satisfies the second wait, so the workflow completes.
+	require.NoError(t, bc.RaiseEvent(ctx, api.InstanceID(wfID), "go", api.WithEventPayload("v2")))
```
acroca: It'd be good to create an event with the same payload just to make sure it's not considered as dup even though the event name and payload is exactly the same (because we also compare timestamp).

acroca (dismissed):
LGTM, just left a couple of comments but feel free to ignore :)

javier-aliaga (dismissed):

tests/integration/suite/daprd/workflow/chaos/eventdedup.go
JoshVanL: Yep :+1: 
I suppose there is a slight thing about checking in case the status flaps but that's not really likely..

JoshVanL:
Thanks @acroca, please take another look

cicoyle (approved):

PR #10075, opened 2026-06-09
feat(sentry): expose extraArgs in sentry Helm chart

nelson-parente:
## Description

The `dapr_sentry` subchart hardcodes every container arg and only exposes `extraEnvVars`, so there is no supported way to pass additional sentry flags via Helm. Sentry flags that aren't surfaced as dedicated values (for example `--jwt-ttl`) can currently only be set by patching the Deployment by hand.

This adds an `extraArgs` list value to the sentry subchart, rendered after the built-in args — mirroring the existing `extraEnvVars` pattern:

```yaml
dapr_sentry:
  extraArgs:
    - "--jwt-ttl=48h"
```

It defaults to `[]`, so there is **no behavior change** when unset.

## How tested

- `helm template` with `extraArgs` set → flags render at the end of the sentry container args.
- `helm template` with the default (`extraArgs: []`) → output unchanged.
- `helm lint` passes; full chart templates without error.

## Release note

```release-note
Add `extraArgs` value to the sentry Helm subchart to allow passing additional command-line flags to the sentry container.
```


acroca (commented):
I think this should be tested in an integration test in `tests/integration/suite/helm`

JoshVanL (approved):

PR #10098, opened 2026-06-14
fix(subscriptions): align empty-status fallthrough in Deliver with DeliverBulk

### LINKED ISSUE #8737: Subscription Fallthru Inconsistency

woodens:
Fixes #8737

Deliver and DeliverBulk in the http postman both handle an empty AppResponse.Status, but in opposite directions: Deliver fell through to success, DeliverBulk to retry. Same response, same status code, two different outcomes.

This aligns them. Empty status now falls through to retry in both paths — if the app didn't say success, treat it as a retryable error and let the resiliency policy decide.


mikeee:
Thanks for your contribution. As there are several PRs currently proposing the same changes this is being closed as a dupe

PR #10114, opened 2026-06-17
fix(resiliency): update circuit breaker state gauge on transitions

### LINKED ISSUE #10113: Bug Report: dapr_resiliency_cb_state remains half-open after successful recovery
## In what area(s)?

Image

Image

Image

Image

Image

/area runtime

## What version of Dapr?

Dapr v1.17.9

## Expected Behavior

After a Circuit Breaker transitions from OPEN → HALF-OPEN, and all probe requests succeed, the Circuit Breaker should transition to CLOSED.

The metric should reflect the new state:

```promql
dapr_resiliency_cb_state{status="closed"} = 1
dapr_resiliency_cb_state{status="half-open"} = 0
```

## Actual Behavior

The runtime behavior appears correct, but the metric remains stuck in HALF-OPEN.

Observed metric:

```promql
dapr_resiliency_cb_state{status="closed"} = 0
dapr_resiliency_cb_state{status="half-open"} = 1
dapr_resiliency_cb_state{status="open"} = 0
```

Even after waiting 5-10 minutes after successful probe requests.

### Resiliency Policy

```yaml
apiVersion: dapr.io/v1alpha1
kind: Resiliency
metadata:
  name: checkout-resiliency
spec:
  policies:
    circuitBreakers:
      circuitBreaker1:
        interval: 300s
        maxRequests: 5
        timeout: 25s
        trip: consecutiveFailures > 25

    retries:
      retry1:
        duration: 3s
        matching:
          httpStatusCodes: 500,501,502,503,504
        maxRetries: 2
        policy: constant

    timeouts:
      timeout1: 7s

  targets:
    apps:
      order:
        circuitBreaker: circuitBreaker1
        retry: retry1
        timeout: timeout1
```

### Test Service

Force CB to OPEN:

```python
if order_id == "500":
    response.status_code = 500
```

Half-open validation endpoint:

```python
if order_id == "201":
    await asyncio.sleep(5)
    return {"message": "success"}
```

### Reproduction Flow

1. Send enough HTTP 500 requests to trip the Circuit Breaker.
2. Verify requests are rejected with:

```text
circuit breaker is open
```

3. Wait more than 25 seconds (CB timeout).
4. Run a load test with:

   * 100 concurrent requests
   * orderId=201

### Observed Behavior

Only 5 requests reach the service:

```text
START
START
START
START
START
```

5 seconds later:

```text
END
END
END
END
END
```

All 5 requests return HTTP 200.

The remaining 95 requests are rejected.

This confirms that:

```yaml
maxRequests: 5
```

is working correctly.

### Additional Observation

Immediately after the successful probe requests:

* Metric still reports HALF-OPEN.
* Waiting 5-10 minutes does not change the metric.

However, running the same load test again results in:

```text
100 requests
100 successes
```

which suggests the runtime Circuit Breaker has already transitioned to CLOSED.

If the Circuit Breaker were truly still HALF-OPEN, I would expect only 5 requests to be allowed through.

## Additional Context

Based on the observed behavior, the Circuit Breaker runtime state appears to transition correctly, but the `dapr_resiliency_cb_state` metric may not be updated accordingly.

Could this be an issue with metric reporting or metric state synchronization?

anneheartrecord:
# Description

`dapr_resiliency_cb_state` is a gauge that is supposed to reflect a circuit breaker's current state. It was only written by `PolicyWithStatusExecuted`, which runs when a policy is instantiated and captures the breaker state at that moment. When a breaker changes state *during* request execution — most notably half-open → closed after a successful probe — the change was reported to the activations counter (`PolicyWithStatusActivated`) but the gauge was never updated. So the gauge stayed on the previously observed state (e.g. `half-open`) even though the runtime breaker had already recovered, until some later policy instantiation happened to observe the new state.

This records the gauge from the circuit breaker state-change callback (`pkg/resiliency/resiliency.go`) so it reflects the new state as soon as the transition happens. The gauge-writing loop in `PolicyWithStatusExecuted` is pulled into a small helper that the new `RecordCircuitBreakerState` method reuses; the execution counter is left untouched, so there is no double counting.

Tests:
- New regression case `EndpointPolicyHalfOpenToClosedState` drives a breaker open → half-open → closed and asserts the gauge ends on `closed`.
- The existing `EndpointPolicyHalfOpenState` case ends with the breaker tripping back to `open` (its probe fails). Its expected last value is updated from `half-open` to `open`, which is the breaker's real state — the old expectation was the stale-gauge behaviour this PR fixes.

`go test -tags unit ./pkg/diagnostics/... ./pkg/resiliency/...` passes locally.

## Issue reference

Please reference the issue this PR will close: #10113

## Checklist

- [x] Code compiles correctly
- [x] Created/updated tests
- [x] Unit tests passing
- [ ] Extended the documentation / Created issue in the <https://github.com/dapr/docs/> repo
- [ ] Specification has been updated / Created issue in the <https://github.com/dapr/docs/> repo
- [ ] Provided sample for the feature / Created issue in the <https://github.com/dapr/docs/> repo

### DIFF
--- a/pkg/diagnostics/resiliency_monitoring.go
+++ b/pkg/diagnostics/resiliency_monitoring.go
@@ -117,24 +117,47 @@ func (m *resiliencyMetrics) PolicyWithStatusExecuted(resiliencyName, namespace s
 
 		// Record cb gauge, 4 metrics, one for each cb state, with the active state having a value of 1, otherwise 0
 		if policy == CircuitBreakerPolicy {
-			for _, s := range cbStatuses {
-				if s == status {
-					_ = stats.RecordWithOptions(
-						m.ctx,
-						stats.WithRecorder(m.meter),
-						stats.WithTags(diagUtils.WithTags(m.circuitbreakerState.Name(), append(commonTags, s)...)...),
-						stats.WithMeasurements(m.circuitbreakerState.M(1)),
-					)
-				} else {
-					_ = stats.RecordWithOptions(
-						m.ctx,
-						stats.WithRecorder(m.meter),
-						stats.WithTags(diagUtils.WithTags(m.circuitbreakerState.Name(), append(commonTags, s)...)...),
-						stats.WithMeasurements(m.circuitbreakerState.M(0)),
-					)
-				}
-			}
+			m.recordCircuitBreakerStateGauge(commonTags, status)
 		}
+	}
+}
+
+// RecordCircuitBreakerState updates the cb_state gauge to reflect a circuit
+// breaker's current state. It is used when the state changes during request
+// execution (e.g. half-open -> closed on a successful probe) so the gauge does
+// not stay stuck at the value that was snapshotted when the policy was
+// instantiated.
+func (m *resiliencyMetrics) RecordCircuitBreakerState(resiliencyName, namespace string, flowDirection PolicyFlowDirection, target string, status string) {
+	if m.enabled {
+		commonTags := []any{
+			appIDKey, m.appID,
+			resiliencyNameKey, resiliencyName,
+			policyKey, string(CircuitBreakerPolicy),
+			namespaceKey, namespace,
+			flowDirectionKey, string(flowDirection),
+			targetKey, target,
+			statusKey, // status appended on each recording
+		}
+		m.recordCircuitBreakerStateGauge(commonTags, status)
+	}
+}
+
+// recordCircuitBreakerStateGauge records the cb_state gauge as one series per
+// possible circuit breaker state, setting the active state to 1 and all others
+// to 0. commonTags must end with the statusKey tag key; the status value is
+// appended here for each recorded series.
+func (m *resiliencyMetrics) recordCircuitBreakerStateGauge(commonTags []any, status string) {
+	for _, s := range cbStatuses {
+		value := int64(0)
+		if s == status {
+			value = 1
+		}
+		_ = stats.RecordWithOptions(
+			m.ctx,
+			stats.WithRecorder(m.meter),
+			stats.WithTags(diagUtils.WithTags(m.circuitbreakerState.Name(), append(commonTags, s)...)...),
+			stats.WithMeasurements(m.circuitbreakerState.M(value)),
+		)
 	}
 }
 
--- a/pkg/diagnostics/resiliency_monitoring_test.go
+++ b/pkg/diagnostics/resiliency_monitoring_test.go
@@ -267,7 +267,43 @@ func TestResiliencyCountMonitoringCBStates(t *testing.T) {
 				diag.NewTag(diag.StatusKey.Name(), string(breaker.StateOpen)):     1,
 				diag.NewTag(diag.StatusKey.Name(), string(breaker.StateHalfOpen)): 1,
 			},
-			wantCbStateLastValue: diag.NewTag(diag.StatusKey.Name(), "half-open"),
+			// The failing probe trips the breaker back to open (half-open -> open).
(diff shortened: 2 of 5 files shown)


pkg/diagnostics/resiliency_monitoring.go
```diff
@@ -117,27 +117,50 @@ func (m *resiliencyMetrics) PolicyWithStatusExecuted(resiliencyName, namespace s
...
+		}
+		_ = stats.RecordWithOptions(
+			m.ctx,
+			stats.WithRecorder(m.meter),
+			stats.WithTags(diagUtils.WithTags(m.circuitbreakerState.Name(), append(commonTags, s)...)...),
+			stats.WithMeasurements(m.circuitbreakerState.M(value)),
+		)
+	}
+}
+
```
JoshVanL, line 163: Thanks @anneheartrecord - please can we have a new integration test for this change? https://github.com/dapr/dapr/tree/master/tests/integration/suite/daprd/metrics

JoshVanL (changes requested):

anneheartrecord:
Thanks for the direction, @JoshVanL!

I've added the integration test you requested:

**New test**: `tests/integration/suite/daprd/metrics/resiliencycbstate.go`
- Two sub-tests that directly cover the fix:
  1. _"gauge shows open after consecutive failures"_ — trips the breaker with 3 failures, asserts `dapr_resiliency_cb_state{status="open"}=1`
  2. _"gauge immediately reflects closed after probe succeeds"_ — waits for CB timeout (1 s), sends one successful probe, then asserts `status="closed"=1` and `status="half-open"=0` via `EventuallyWithT` — this is the core regression check for #10113

**Updated test**: `tests/integration/suite/daprd/resiliency/apps/defaultcircuitbreaker.go`
- Replaced the stale `status="half-open"=1` assertion (which encoded the old broken behaviour) with an immediate `status="closed"=1` check to match what the fixed code now produces.

Both files pass `gofmt` and `go vet`.

JoshVanL:
@anneheartrecord please can you fix the DCO?

https://github.com/dapr/dapr/pull/10114/checks?check_run_id=82367777993

anneheartrecord:
Fixed — rebased with `--signoff` to add `Signed-off-by:` to both commits. The DCO check should now pass.

The only other failing check (`Integration tests macos-14`) is in `Test_Integration/actors/reminders/scheduler/precision` — unrelated to the circuit breaker change.

JoshVanL (approved):

naot7:
It seems like Issue #10113 is still not fixed in version v1.18.1, as I am still encountering the same behavior. Do I need to wait for a newer release, or is there a specific patch I should look out for?

anneheartrecord:
v1.18.1 won't include this PR if that release was cut before #10114 merged. The fix is on master now via #10114 (merged Jun 23, 2026), so you'll need a later Dapr release, or a backport if the maintainers decide to include it in a patch release. I don't control the release/backport timing; the safest thing to watch is the release notes or any backport PR linked from #10114.

JoshVanL:
@naot7 @anneheartrecord - if you would like this PR released in the v1.18 release channel (that makes sense to me), then we will need a new entry to the v1.18.2 (unreleased) release notes, see https://github.com/dapr/dapr/blob/master/docs/release_notes/v1.18.2.md

Once open, I can mark this PR for v1.18 backporting, as well as that new release notes PR. We will take care of the backporting and release itself.

marcduiker:
@holopin-bot @anneheartrecord runtime-badge Thank you! Here's a digital badge as a small token of appreciation.

Issue #9563, opened 2026-02-26
Sentry always regenerating and replacing dapr-trust-bundle


dicki:
## In what area(s)?

> /area runtime

> /area operator

> /area placement

> /area docs

> /area test-and-release

> /area injector

> /area scheduler

/area sentry

## What version of Dapr?

1.16.9

## Expected Behavior

After running `dapr mtls renew-certificate -k --restart` command to renew the expired `dapr-trust-bundle` used for mTLS, the new `dapr-trust-bundle` should persist, even through Sentry restarts, until it is renewed again in due time.

## Actual Behavior

Shortly after running `dapr mtls renew-certificate -k --restart` command to renew the expired `dapr-trust-bundle` the Sentry service appears detects the updated bundle and tries to reload it but fails.
This results in it regenerating the certs and requiring manual restarting of the Dapr control plane and my applications as well.

Log lines indicating the behaviour:

```
Root and issuer certs not found: generating self signed CA
Issuer credentials changed; reloading
```

```
{"instance":"dapr-sentry-5fc8c77d9c-tbgmt","level":"warning","msg":"Issuer credentials changed; reloading","scope":"dapr.sentry","time":"2026-02-26T11:51:45.609923041Z","type":"log","ver":"1.16.9"}

{"instance":"dapr-sentry-5fc8c77d9c-tbgmt","level":"info","msg":"Root and issuer certs not found: generating self signed CA","scope":"dapr.sentry.ca","time":"2026-02-26T11:51:45.631510905Z","type":"log","ver":"1.16.9"}
```

## Steps to Reproduce the Problem

1. Run `dapr mtls renew-certificate -k --restart`
2. Wait for Sentry to reject the newly created bundle and regenerate new ones

## Release Note

RELEASE NOTE: **FIX** Sentry regenerating renewed certificate bundle

dicki:
I suspect this (https://github.com/dapr/dapr/blob/c353a56031ca489424cb759da0992509c2f1ae8e/pkg/sentry/server/ca/kube.go#L70) condition to be incorrect and therefore the bundle gets regenerated.

Issue #8407, opened 2025-01-23
Allow bootstrapping mTLS certificates outside of the helm chart


vit-zikmund:
## In what area(s)?
/area sentry

### Help the operators read
This proposal is based on my production experience running Dapr (v1.13) via its Helm chart on a custom k8s cluster. The [documentation says](https://docs.dapr.io/operations/security/mtls/) (at the very top), you _can_ bring  your own certificates, but you might as well use the default ones.
> Dapr allows operators and developers to bring in their own certificates, or instead let Dapr automatically create and persist self-signed root and issuer certificates.

Not loving to read all the letters of a documentation for a feature you don't care for, you might as well stop reading right there. The sentence doesn't say you shouldn't let Dapr handle this for you unless you opt for a lovely cluster disaster one year later. Well, it does, but so subtly, one might tend to think of this being later handled somehow more gracefully (than ending up with the `dapr-operator` perpetually killing the dapr-enabled pods, for which the `dapr-injector` failed to inject the sidecar, because it saw an expired certificate in `dapr-sentry` 😮‍💨).
>If custom certificates have not been provided, Dapr automatically creates and persist self-signed certs **valid for one year**.

The warning about your inevitable downtime - if you use the default configuration - is [much farther in the docs](https://docs.dapr.io/operations/security/mtls/#updating-root-or-issuer-certs-using-kubectl).
> Avoiding downtime when rotating certificates
To avoid downtime when rotating expiring certificates your new certificates must be signed with the same private root key as the previous certificates. **This is not currently possible using self-signed certificates generated by Dapr.**

Correcting this by warning readers in the very 1st section should be pretty easy and it's my first mini-proposal.

### Make their life easier
[...]

PR #10131, opened 2026-06-25
Allow a pre-existing dapr-trust-bundle to be preserved

karljaxon:
# Description

Prior to a major overhaul of identity management (https://github.com/dapr/dapr/pull/6171) the helm release of dapr would preserve the existing trust bundle (using a lookup). This had some issues if the certificate was undefined (see https://github.com/dapr/dapr/issues/2808), and the overhaul moved to the existing secret ALWAYS being replaced. This includes replacing a secret which has been auto-generated by Sentry itself.

This PR would a user to bring their own trust bundle and avoid the helm upgrade overwriting it. In my use case I want cert-manager to manage the certificates and use Flux/Helm to restructure this into the secret and configmap expected by Dapr. This small change would allow me to stop Dapr helm upgrades from replacing the secret and cm generated by cert-manager.

## Issue reference

Partially addresses:
- https://github.com/dapr/dapr/issues/8407
- https://github.com/dapr/dapr/issues/3968

May also be related to why this bundle regeneration issue occurred:
- https://github.com/dapr/dapr/issues/9563

## Checklist

All N/A - no compiled code or affected tests and proposing this is NOT surfaced as top-level documented value

- [ ] Code compiles correctly
- [ ] Created/updated tests
- [ ] Unit tests passing


charts/dapr/charts/dapr_sentry/values.yaml
```diff
@@ -23,6 +23,9 @@ ports:
   targetPort: 50001
 
 tls:
+  # Controls whether this chart renders the dapr-trust-bundle Secret/ConfigMap manifests.
+  # WARNING: Switching this from true -> false on an existing release will cause Helm to delete those resources during upgrade.
```
mikeee: Would it be worth adding a resource-policy to keep the trust bundle when the flag is toggled the other way?

mikeee (commented):
Please add docs to surface this in the README 👍

JoshVanL (approved):

JoshVanL (dismissed):

charts/dapr/charts/dapr_sentry/values.yaml
```diff
@@ -23,6 +23,9 @@ ports:
...
+  generateBundleSecret: true
```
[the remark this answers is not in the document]
karljaxon: I don't regard the configmap as a separate resource. It is simply an extract of the public information from the secret. I feel 'generateBundleSecret' is concise/readable and users would expect the configmap to match

charts/dapr/charts/dapr_sentry/values.yaml
karljaxon: Sorry for slow response - have been on holiday  :)
The idea being we could flip the value? So use the chart to create the initial value and then we don't need to worry about deletion when we change our values. I like it - can probably remove the copilot-generated warning too

charts/dapr/charts/dapr_sentry/templates/dapr_sentry_deployment.yaml
```diff
@@ -1,3 +1,4 @@
+{{ if .Values.tls.generateBundleSecret }}
```
JoshVanL, line 1: Sentry will still regenerate root and issuer and write over your cert-manager bundle whenever the ConfigMap ca.crt doesn't byte match the Secret's (kube.go (https://github.com/dapr/dapr/blob/master/pkg/sentry/server/ca/kube.go#L64)). Needs #10290 first to actually "preserve".

charts/dapr/charts/dapr_sentry/templates/dapr_sentry_deployment.yaml
```diff
@@ -8,6 +9,8 @@ metadata:
     {{- range $key, $value := .Values.global.k8sLabels }}
     {{ $key }}: {{ tpl $value $ }}
     {{- end }}
+  annotations:
+    helm.sh/resource-policy: keep
```
JoshVanL, line 13: Today a dapr reinstall generates a new CA- this must be opt in.

charts/dapr/charts/dapr_sentry/templates/dapr_sentry_deployment.yaml
```diff
@@ -25,10 +28,13 @@ metadata:
...
```
JoshVanL, line 32: Same here

charts/dapr/charts/dapr_sentry/values.yaml
```diff
@@ -23,6 +23,8 @@ ports:
   targetPort: 50001
 
 tls:
+  # Controls whether this chart renders the dapr-trust-bundle Secret/ConfigMap manifests.
+  generateBundleSecret: true
```
JoshVanL, line 27: This needs a fail when the value is false and the objects aren't already present

JoshVanL (changes requested):
Before continuing with this, I think we need https://github.com/dapr/dapr/pull/10290 first.

charts/dapr/charts/dapr_sentry/templates/dapr_sentry_deployment.yaml
karljaxon: Agreed sentry does blat the bundle if they don't match. For my use case I generate them both at the same time (same HelmRelease) so they are guaranteed to match so it didn't matter too much, but happy to wait for 10290 if you think that's best?

charts/dapr/charts/dapr_sentry/templates/dapr_sentry_deployment.yaml
karljaxon: The keep resource policy would just prevent the delete if a future helm release would remove. It wouldn't change the behaviour if updated or if the secret is over-ridden by sentry, I believe? Does this affect whether we need this to be opt in? Also happy to remove the resource policy entirely - was a previous review comment to add it

charts/dapr/charts/dapr_sentry/templates/dapr_sentry_deployment.yaml
karljaxon: The keep resource policy would just prevent the delete if a future helm release would remove. It wouldn't change the behaviour if updated or if the secret is over-ridden by sentry, I believe? Does this affect whether we need this to be opt in? Also happy to remove the resource policy entirely - was a previous review comment to add it

charts/dapr/charts/dapr_sentry/values.yaml
karljaxon: OK. Heads up this would be helm 'lookup' and would mean dry runs would fail if client-side

PR #10132, opened 2026-06-25
fix(operator): only hot-reload a sidecar for its assigned Configuration

JoshVanL:
In Kubernetes mode the operator's ConfigurationUpdate stream sent every Configuration in a namespace to every connected sidecar. Configurations carry no scopes, so the informer filter treated them as belonging to all apps, and the sidecar's SIGHUP reconciler restarted the runtime on any Configuration create/update in the namespace, including ones it does not use.

The operator now resolves, server side, which Configuration is assigned to a connecting sidecar from its pod's dapr.io/config annotation, keyed off the app identity authenticated over mTLS, and only streams updates for that Configuration. The sidecar is never trusted to self-report its own config.

The pod lookup uses a dedicated metadata-only pod cache (PartialObjectMetadata, as sentry does) so no pod specs are fetched or retained, reusing the manager's REST mapper and HTTP client to avoid a second API discovery. The config is resolved before the informer watcher is registered so a slow pod cache cannot back up the watcher's event channel.

Also bound the operator gRPC GracefulStop with a forceful Stop fallback so a streaming handler that does not return promptly can no longer hang operator shutdown.


pkg/operator/api/api.go
```diff
@@ -153,7 +164,19 @@ func (a *apiServer) Run(ctx context.Context) error {
...
+			// handler does not return promptly, so shutdown cannot hang.
+			done := make(chan struct{})
+			go func() {
+				s.GracefulStop()
+				close(done)
+			}()
+			select {
+			case <-done:
+			case <-time.After(5 * time.Second):
+				s.Stop()
```
acroca, line 177: This call to `Stop` will be sent at the same time as we wait for the `GracefulStop`. Checking the gRPC implementation, both stop functions en up in the same private `stop` ([ref](https://github.com/grpc/grpc-go/blob/d74861eb76cc713fae3ffd381e4f25c5d1324111/server.go#L1969)) function and the difference between graceful and not graceful is in a piece of code that has a mutex. So I don't think the `Stop` will really force the stop...

JoshVanL: Stop does force close the gRPC server without gracefully waiting for all client connections to have closed, like in s.GracefulStop(). There are some int tests which cover this being needed, shown by the test hangining.

pkg/operator/api/configurations.go
```diff
@@ -113,6 +116,28 @@ func (a *apiServer) ConfigurationUpdate(in *operatorv1pb.ConfigurationUpdateRequ
 
 	ctx := srv.Context()
 
+	// Verify authorization and resolve the connecting app's identity.
+	id, err := authz.Request(ctx, in.GetNamespace())
+	if err != nil {
+		return err
```
lrascao: wrap error

pkg/operator/api/configurations.go
```diff
@@ -113,6 +116,28 @@ func (a *apiServer) ConfigurationUpdate(in *operatorv1pb.ConfigurationUpdateRequ
...
+
+	// Determine, server side, which configuration is assigned to the connecting
+	// app from its pod's dapr.io/config annotation. The sidecar is not trusted to
+	// self-report this. Only updates to that configuration are streamed, so the
+	// sidecar is not restarted by changes to configurations that do not belong to
+	// it. This is resolved before registering the informer watcher so a slow pod
+	// cache lookup cannot back up the watcher's event channel.
+	assigned, err := a.appAssignedConfiguration(ctx, in.GetNamespace(), id.AppID())
+	if err != nil {
+		return err
```
lrascao: wrap error

pkg/operator/cache/podmeta.go
```diff
@@ -0,0 +1,132 @@
...
+// sinkholing everything else. It is used to resolve, server side, which
+// configuration is assigned to a connecting sidecar from its pod's dapr.io/config
+// annotation, without holding full pod objects in memory.
+func NewPodMetadataCache(podOpts PodMetadataCacheOptions) (ctrlcache.Cache, error) {
+	cfg := rest.CopyConfig(podOpts.RestConfig)
+	cfg.ContentType = runtime.ContentTypeJSON
+
+	scheme := runtime.NewScheme()
+	if err := corev1.AddToScheme(scheme); err != nil {
+		return nil, err
```
lrascao: wrap error

lrascao (approved):
g2g, minor nit is not enough error wrapping

cicoyle (approved):
lgtm

PR #10138, opened 2026-06-30
workflow: do not expose failure message as workflow output

### LINKED ISSUE #9085: [Workflows] Incorrect value being set on output during failed workflow
When a Workflow throws an exception which puts the workflow into a `failed` state, the value of `dapr.workflow.failure.error_message` is copied into `dapr.workflow.output"` which I believe is wrong behaviour.

IMO If a workflow has not completed successfully, the `output` should be undefined/null

## In what area(s)?

/area runtime

## What version of Dapr?

1.16

## Expected Behavior

`"dapr.workflow.output"` is set to empty value or something that represents null e.g.`""`

## Actual Behavior

```json
{
    "instanceID": "0-210",
    "workflowName": "FanOutWorkflow",
    "createdAt": "2025-09-26T21:48:05.974667500Z",
    "lastUpdatedAt": "2025-09-26T21:48:07.122787625Z",
    "runtimeStatus": "FAILED",
    "properties": {
        "dapr.workflow.failure.error_message": "Task 'SuperSlowActivity' (#0) failed with an unhandled exception: bomb",
        "dapr.workflow.failure.error_type": "Dapr.Workflow.WorkflowTaskFailedException",
        "dapr.workflow.input": "{\"RandomData\":\"b4cf9ba7-f925-4190-831f-fa1bdf77d40f\",\"Itterations\":200,\"Data\":null}",
        "dapr.workflow.output": "Task 'SuperSlowActivity' (#0) failed with an unhandled exception: bomb"
    }
}
```

cc @JoshVanL 

## Release Note

RELEASE NOTE:

anxkhn:
# Description

A workflow that ends in `FAILED` was copying the failure message into the
`dapr.workflow.output` property returned by the workflow `Get` API. The failure
is already reported separately via `dapr.workflow.failure.error_message` and
`dapr.workflow.failure.error_type`, so the same error text was surfaced twice
and a failed workflow appeared to have a successful output.

Root cause: `durabletask-go` stores the failure message in the completed-event
result, which `runtimestate.Output()` returns. `pkg/runtime/wfengine/client.go`
`Get()` copied that value into `dapr.workflow.output` unconditionally.

The fix gates the `dapr.workflow.output` assignment on `metadata.FailureDetails
== nil`, which is exactly the "completed with failure" signal (it is nil for
`RUNNING` and for a successful `COMPLETED`, and is not set for `TERMINATED` /
`CANCELED`). So only the `FAILED` case changes: it no longer exposes the failure
text as output. The failure is still reported via the existing
`dapr.workflow.failure.*` properties immediately below the changed block.

Why fix it here rather than upstream in `durabletask-go`: Dapr owns the
`dapr.workflow.output` API contract. `durabletask-go` storing the error text in
the completed-event result is general behaviour shared by other consumers;
suppressing the duplicate at Dapr's response builder is self-contained and
avoids an upstream dependency change.

Before:

```json
"dapr.workflow.failure.error_message": "Task 'SuperSlowActivity' (#0) failed ... boom",
"dapr.workflow.output": "Task 'SuperSlowActivity' (#0) failed ... boom"
```

After: `dapr.workflow.output` is omitted for a failed workflow;
`dapr.workflow.failure.error_message` is unchanged.

RELEASE NOTE: **FIX** Do not set `dapr.workflow.output` to the failure message for failed workflows.

## Issue reference

Please reference the issue this PR will close: #9085

## Checklist

- [x] Code compiles correctly
- [x] Created/updated tests
- [x] Unit tests passing
- [ ] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_
- [ ] Specification has been updated / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_
- [ ] Provided sample for the feature / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_

### DIFF
--- a/pkg/runtime/wfengine/client.go
+++ b/pkg/runtime/wfengine/client.go
@@ -200,7 +200,11 @@ func (c *client) Get(ctx context.Context, req *workflows.GetRequest) (*workflows
 		res.Workflow.Properties["dapr.workflow.input"] = metadata.GetInput().GetValue()
 	}
 
-	if metadata.Output != nil {
+	// A failed workflow has no successful output: durabletask-go stores the
+	// failure message in the completed-event result (surfaced here as Output),
+	// but the failure is already reported via dapr.workflow.failure.* below, so
+	// it must not also be exposed as dapr.workflow.output.
+	if metadata.Output != nil && metadata.FailureDetails == nil {
 		res.Workflow.Properties["dapr.workflow.output"] = metadata.GetOutput().GetValue()
 	}
 
--- a/pkg/runtime/wfengine/client_test.go
+++ b/pkg/runtime/wfengine/client_test.go
@@ -1,0 +1,134 @@
+/*
+Copyright 2026 The Dapr Authors
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package wfengine
+
+import (
+	"context"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"google.golang.org/protobuf/types/known/wrapperspb"
+
+	"github.com/dapr/durabletask-go/api"
+	"github.com/dapr/durabletask-go/api/protos"
+	"github.com/dapr/durabletask-go/backend"
+
+	"github.com/dapr/components-contrib/workflows"
+	"github.com/dapr/kit/logger"
+)
+
+// fakeTaskHubClient is a minimal backend.TaskHubClient whose only behaviour is
+// to return a pre-built WorkflowMetadata from FetchWorkflowMetadata. Every
+// other method is an unused no-op required to satisfy the interface.
+type fakeTaskHubClient struct {
+	metadata *backend.WorkflowMetadata
+}
+
+func (f *fakeTaskHubClient) FetchWorkflowMetadata(context.Context, api.InstanceID) (*backend.WorkflowMetadata, error) {
+	return f.metadata, nil
+}
+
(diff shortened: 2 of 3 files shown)


pkg/runtime/wfengine/client.go
```diff
@@ -200,7 +200,11 @@ func (c *client) Get(ctx context.Context, req *workflows.GetRequest) (*workflows
 		res.Workflow.Properties["dapr.workflow.input"] = metadata.GetInput().GetValue()
 	}
 
-	if metadata.Output != nil {
+	// A failed workflow has no successful output: durabletask-go stores the
+	// failure message in the completed-event result (surfaced here as Output),
+	// but the failure is already reported via dapr.workflow.failure.* below, so
+	// it must not also be exposed as dapr.workflow.output.
+	if metadata.Output != nil && metadata.FailureDetails == nil {
 		res.Workflow.Properties["dapr.workflow.output"] = metadata.GetOutput().GetValue()
```
JoshVanL, line 208: Please can we have a new dedicated integration test for this change? https://github.com/dapr/dapr/tree/master/tests/integration/suite/daprd/workflow

JoshVanL (changes requested):

anxkhn:
Thanks Josh. I added a dedicated integration test in `tests/integration/suite/daprd/workflow/failureoutput.go`. It runs a workflow whose activity fails to a FAILED state and asserts, over both the gRPC `GetWorkflowBeta1` and the HTTP Get endpoints, that the failure surfaces via `dapr.workflow.failure.error_message`/`error_type` and that `dapr.workflow.output` is not present.

One thing worth calling out: with the Go durabletask SDK a failed workflow's completed-event result is `nil`, so the actor backend already returns a nil output and `dapr.workflow.output` is absent regardless of this change (I confirmed this end to end). The original report was against the .NET SDK, whose worker copies the error text into the completed-event result, which is the case the `FailureDetails == nil` guard actually suppresses; that path is covered directly by the unit test in `pkg/runtime/wfengine/client_test.go`. So the integration test locks in the observable Get-API contract from #9085 end to end, and the unit test guards the `client.go` condition itself. Happy to push this further (for example a raw work-item worker that reproduces the .NET-style FAILED+result completion) if you would prefer the integration test to exercise that exact path.

JoshVanL (approved):

JoshVanL:
Thank you @anxkhn

PR #10139, opened 2026-06-30
messaging: stamp caller/callee headers on self-invocation

### LINKED ISSUE #10061: Service invocation: self-invocation (invoking your own app-id) delivers no dapr-caller-app-id / dapr-caller-namespace / dapr-callee-app-id headers
## In what area(s)?

/area runtime

## What version of Dapr?

Observed on 1.17.7 (Kubernetes). The code path is unchanged on master.

## Expected Behavior

The [service invocation API docs](https://docs.dapr.io/reference/api/service_invocation_api/#headers-in-service-invocation-requests) document `dapr-caller-app-id`, `dapr-caller-namespace`, and `dapr-callee-app-id` unconditionally ("The ID of the calling application"). We expected an app invoking *itself* through its own sidecar (`POST /v1.0/invoke/<own-app-id>/method/<m>`) to receive these headers like any other callee.

## Actual Behavior

When the target app-id and namespace match the sidecar's own, `directMessaging.Invoke` short-circuits to `invokeLocal`, which dispatches straight to the app channel:

```go
// pkg/messaging/direct_messaging.go
if app.id == d.appID && app.namespace == d.namespace {
    return d.invokeLocal(ctx, req)
}
return d.invokeWithRetry(ctx, ..., d.invokeRemote, req)
```

Only `invokeRemote` calls `addCallerAndCalleeAppIDHeaderToMetadata` (and `addForwardedHeadersToMetadata` / `addDestinationAppIDHeaderToMetadata`). `invokeLocal` adds nothing — so on a self-invoke the app receives **none** of `dapr-caller-app-id`, `dapr-caller-namespace`, `dapr-callee-app-id`, or the `Forwarded`/`X-Forwarded-*` set.

Source: [pkg/messaging/direct_messaging.go](https://github.com/dapr/dapr/blob/master/pkg/messaging/direct_messaging.go)

## Why it matters

A documented use of these headers is identifying the calling app — e.g. allowlisting which internal apps may hit an internal route. With this gap, a callee cannot distinguish "my own sidecar's loopback" from "an anonymous direct hit on my app port":

- Fail-closed services reject **only** the self-invoke leg (cross-app calls carry the headers and work), which makes this confusing to diagnose — the identical request succeeds from every other app's sidecar.
- The tempting workaround — treating "no caller headers" as trusted — fails open.

We were bitten by this twice in production-like environments before root-causing it, and ended up bypassing Dapr for self-calls entirely (direct loopback + an app-minted boot token).

In-pod evidence matrix (same route, same payload):

| Call | Result |
|---|---|
| Self-invoke via own sidecar (`/v1.0/invoke/myapp/method/...`) | 401 (no `dapr-caller-app-id` delivered) |
| Identical call from another app's sidecar | 200 (`dapr-caller-app-id: otherapp` present) |
| Self-invoke with a manually injected `dapr-caller-app-id: myapp` | 200 |

## Inconsistency within Dapr

The actor router already solves this exact problem for placement-local actor calls — `stampLocalCallerIdentity` in [pkg/actors/router/router.go](https://github.com/dapr/dapr/blob/master/pkg/actors/router/router.go) stamps `(appID, namespace)` on local calls, preserving identity already stamped upstream:

```go
func (r *router) stampLocalCallerIdentity(req *internalv1pb.InternalInvokeRequest) {
    ...
    // Preserve identity stamped upstream (e.g. by the CallActor gRPC
    // handler from a remote SPIFFE peer) so a remote-routed-local call
    // keeps the remote caller's identity.
    if workflowacl.CallerAppID(req.GetMetadata()) != "" {
        return
    }
    workflowacl.SetCallerIdentity(req, r.appID, r.namespace)
}
```

Service invocation's `invokeLocal` has no equivalent.

1. Run any app `myapp` with a Dapr sidecar; add a route that echoes request headers.
2. Cross-app: `curl localhost:3500/v1.0/invoke/myapp/method/echo` from *another* app's sidecar → `dapr-caller-app-id: otherapp` (and `dapr-caller-namespace`, `dapr-callee-app-id`) present.
3. Self: the same call from inside `myapp` via its own sidecar → none of the `dapr-caller-*` / `dapr-callee-*` headers are present.

## Proposed fix

Stamp caller/callee identity in `invokeLocal` (mirroring `stampLocalCallerIdentity` semantics: skip if already stamped upstream), e.g. calling `addCallerAndCalleeAppIDHeaderToMetadata(d.namespace, d.appID, d.appID, req)` before dispatching to the app channel — or, at minimum, document the self-invocation exception in the service invocation API reference.

anxkhn:
# Description

Service invocation to your own app id (`POST /v1.0/invoke/<own-app-id>/method/<m>`)
short-circuits to `invokeLocal`, which dispatched straight to the app channel
without setting any caller/callee identity headers. Only `invokeRemote` calls
`addCallerAndCalleeAppIDHeaderToMetadata`, so a self-invoke reached the app with
none of `dapr-caller-app-id`, `dapr-caller-namespace` or `dapr-callee-app-id`,
while cross-app invocations carry them. Services that allowlist by
`dapr-caller-app-id` therefore reject only the self leg.

This stamps the same headers in `invokeLocal` (caller == callee == this app,
own namespace), preserving any caller already stamped upstream so the local leg
sees the same headers as a cross-app call. It mirrors `invokeRemote` and the
actor router's `stampLocalCallerIdentity`.

## Issue reference

Please reference the issue this PR will close: #10061

## Checklist

- [x] Code compiles correctly
- [x] Created/updated tests
- [x] Unit tests passing
- [x] Extended the documentation / Created issue in the <https://github.com/dapr/docs/> repo: n/a - behaviour now matches the documented headers
- [ ] Specification has been updated / Created issue in the <https://github.com/dapr/docs/> repo: n/a - no API change
- [ ] Provided sample for the feature / Created issue in the <https://github.com/dapr/docs/> repo: n/a - bug fix

## Release note

```release-note
FIX: self-invocation now stamps dapr-caller-app-id, dapr-caller-namespace and dapr-callee-app-id, matching cross-app service invocation.
```

### DIFF
--- a/pkg/messaging/direct_messaging.go
+++ b/pkg/messaging/direct_messaging.go
@@ -277,6 +277,21 @@ func (d *directMessaging) invokeLocal(ctx context.Context, req *invokev1.InvokeM
 		return nil, errors.New("cannot invoke local endpoint: app channel not initialized")
 	}
 
+	// A self-invocation bypasses invokeRemote, so stamp the caller/callee
+	// identity headers here too. caller == callee == this app. Any
+	// caller-supplied identity headers are stripped first and re-stamped
+	// unconditionally, mirroring invokeRemote, so an app cannot spoof its own
+	// identity headers on the local leg. The strip is case-insensitive because
+	// HTTP-origin metadata retains the request's original header casing.
+	md := req.Metadata()
+	for k := range md {
+		switch strings.ToLower(k) {
+		case invokev1.CallerIDHeader, invokev1.CallerNamespaceHeader, invokev1.CalleeIDHeader:
+			delete(md, k)
+		}
+	}
+	d.addCallerAndCalleeAppIDHeaderToMetadata(d.namespace, d.appID, d.appID, req)
+
 	return appChannel.InvokeMethod(ctx, req, "")
 }
 
--- a/pkg/messaging/direct_messaging_test.go
+++ b/pkg/messaging/direct_messaging_test.go
@@ -31,6 +31,7 @@ import (
 	"google.golang.org/protobuf/types/known/anypb"
 
 	"github.com/dapr/dapr/pkg/channel"
+	channelfake "github.com/dapr/dapr/pkg/channel/fake"
 	invokev1 "github.com/dapr/dapr/pkg/messaging/v1"
 	commonv1pb "github.com/dapr/dapr/pkg/proto/common/v1"
 	internalv1pb "github.com/dapr/dapr/pkg/proto/internals/v1"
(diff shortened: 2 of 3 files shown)


pkg/messaging/direct_messaging.go
```diff
@@ -277,6 +277,14 @@ func (d *directMessaging) invokeLocal(ctx context.Context, req *invokev1.InvokeM
 		return nil, errors.New("cannot invoke local endpoint: app channel not initialized")
 	}
 
+	// A self-invocation bypasses invokeRemote, so stamp the caller/callee
+	// identity headers here too. Mirrors invokeRemote: caller == callee == this
+	// app. Preserve any caller already stamped upstream so allowlist-by-caller
+	// services see the same headers for the local leg as for cross-app calls.
+	if req.Metadata()[invokev1.CallerIDHeader] == nil {
+		d.addCallerAndCalleeAppIDHeaderToMetadata(d.namespace, d.appID, d.appID, req)
+	}
```
JoshVanL, line 292: This should be done unconditionally to prevent spoofing.

```suggestion
      // A self-invocation bypasses invokeRemote, so stamp the caller/callee
      // identity headers here too. caller == callee == this app. Stamped
      // unconditionally, mirroring invokeRemote, so an app cannot spoof its own
      // identity headers on the local leg.
      md := req.Metadata()
      delete(md, invokev1.CallerIDHeader)
      delete(md, invokev1.CallerNamespaceHeader)
      delete(md, invokev1.CalleeIDHeader)
      d.addCallerAndCalleeAppIDHeaderToMetadata(d.namespace, d.appID, d.appID, req)
```

JoshVanL (changes requested):
Thanks @anxkhn, please can you add an integration test for this, as well as covering the cases where we prevent spoofing.

JoshVanL (approved):

Issue #1026, opened 2020-02-11
Actors: dapr runtime should not arm/queue timer before timer method completes


LMWF:
## In what area(s)?

> /area runtime

## What version of Dapr?

0.3

## Expected Behavior

dapr runtime should arm actor timers again only after the actor method/timer callback/reminder callback has finished.  

## Actual Behavior

Today it may arm and queue them before the current method finishes, for example if `period` is set to something small.

## Steps to Reproduce the Problem

1. register a timer with 0 period
2. implement a timer callback method that sleeps a minute before returning

LMWF:
They aren't queued, no change needed.

PR #10145, opened 2026-06-30
Actors: run in-memory timer callbacks concurrently with a bounded pool

### LINKED ISSUE #10144: Actor timer callbacks for different actors don't run concurrently
## In what area(s)?

/area runtime

## What version of Dapr?

Observed on 1.17.7. Confirmed still present on `master` (1.18 dev, dapr/kit v0.18.1). Introduced in 1.15.0 by the actors runtime rewrite (#8310).

## Expected Behavior

Per the actor concurrency model, "turn-based concurrency is enforced on a
per-actor basis and not across actors," and "Actor methods and timer/reminder
callbacks can execute simultaneously on behalf of different actors"
(https://docs.dapr.io/developing-applications/building-blocks/actors/actors-features-concepts/).

A slow in-memory timer callback for one actor should therefore not delay timer
callbacks for other actors on the same sidecar — the same way actor methods and
scheduler-backed reminders already behave.

## Actual Behavior

All in-memory actor timer callbacks for an entire sidecar are executed by a single serial goroutine, so one slow/long callback blocks every other actor's timer on that pod. There is also no per-callback timeout.

Source: the sidecar builds one `inmemory` timer store (one `queue.Processor`) in `pkg/actors/actors.go`. Its `processorExecuteFn` calls `router.CallReminder(context.TODO(), reminder)` **synchronously** and only arms the next tick afterward; dapr/kit's `queue.Processor` runs `executeFn` on a single `processLoop` goroutine (no `go`). So every actor's timer callbacks round-robin through one goroutine, and the callback runs with no timeout (`context.TODO()`).

This contradicts the documented model (which the reminder path satisfies via `pkg/runtime/scheduler/internal/cluster/streamer.go`, dispatching one goroutine per fire).

1. On a single sidecar, register a repeating timer on two different actors (A and B), both due immediately.
2. Make actor A's timer callback take ~2s (e.g., `sleep`).
3. Observe that actor B's timer callback does not fire until A's completes — the two callbacks strictly round-robin through one goroutine; there is zero cross-actor concurrency.

Production observation (daprd 1.17.7): 8 sidecars, ~4 shard-actors each, every shard a 100ms repeating timer whose callback takes ~3.7s. The four co-located shards ran strict round-robin through one goroutine (~5ms gaps, zero cross-shard concurrency); the single timer goroutine was 74–86% saturated and the effective per-shard cadence was ~15s instead of 100ms.

## Release Note

RELEASE NOTE: **FIX** In-memory actor timer callbacks for different actors now run concurrently instead of serializing on a single goroutine.

MrMint:
# Description

In-memory actor timer callbacks for an entire sidecar were executed by the single goroutine of one `queue.Processor` (`processorExecuteFn` called `router.CallReminder` synchronously and only armed the next tick afterward). One slow callback therefore blocked every other actor's timer on the pod, and there was no per-callback cancellation on shutdown (`context.TODO()`).

This dispatches each due timer callback to a bounded worker pool so callbacks for different actors run concurrently — matching the documented turn-based model and the scheduler-backed reminder path — while preserving correctness:

- **Per-actor turn-based execution is unchanged.** An actor's timers and methods still serialize on the existing per-actor lock (`pkg/actors/targets/app/lock`); the lock guarantees mutual exclusion, not scheduled-time ordering.
- **A repeating timer never overlaps its own next tick** — the next tick is enqueued only after the callback returns (dapr/dapr#1026).
- **The `activeTimers` ↔ processor transition is atomic.** `Create`, `Delete`, and the in-flight re-enqueue hold a `queueLock`, so a concurrent `Delete`/replace cannot interleave between the membership check and the `Enqueue` (which could otherwise resurrect a deleted timer or silently lose a replacement). This also let `Create` drop its previous spin-retry loop.
- **No in-place mutation of a queued reminder.** The schedule is advanced on a copy and the copy is published as the new active timer, so the processor's heap never reads a reminder that a pool goroutine is mutating (the old code was race-free only because callbacks ran on the processor's own goroutine; this is verified under `go test -race`).
- **Bounded.** Concurrency is capped at an internal default of 100 (matching `defaultBulkPublishMaxConcurrency`, the runtime's other bounded worker pool), so a backlog of slow callbacks cannot grow goroutines without bound. In-flight callbacks are cancelled and drained on `Close`.

### Looking for feedback

1. **The concurrency bound and whether it should be configurable.** ~~I hardcoded the cap at 100 which is fairly arbitrary. For comparison the analogous reminder path is effectively *unbounded* on the daprd side (the scheduler streamer dispatches one goroutine per fire; scheduler `--workers` defaults to 2048). I looked for existing examples of hardcoded concurrency limits and found the `defaultBulkPublishMaxConcurrency` as an example.~~

2. **Bounded pool vs. per-actor sharding.** ~~A pool slot is held for a callback's whole lifetime, including time blocked on the per-actor lock. So an actor with many distinct timers all firing at once can occupy several slots with lock-waiters (it degrades gracefully — roughly 100 distinct concurrently-due timers on a *single* actor to saturate the pool). An alternative is per-actor-key sharded serial queues (similar to the per-name FIFO loops in #9938), which bounds a hot actor's blast radius and preserves intra-actor scheduled order, at the cost of capping concurrency at the shard count. I chose the pool for simplicity and for I/O-bound parallelism; happy to revisit if you'd rather have the sharded model.~~

3. **Minor:** ~~advancing the schedule on a copy adds one small allocation per tick (vs. the prior in-place mutation). It's required for race-freedom now that callbacks run off the processor goroutine, but flagging the tradeoff.~~

## Tests

- Unit (`-race`): cross-actor concurrency, the concurrency bound (incl. that parked timers still run), no self-overlap of a repeating timer, the delete-during-fire guard, graceful `Close` cancellation, and a concurrent `Create`/`Delete`/fire stress test.
- Integration (`tests/integration/suite/actors/timers/concurrency`): with a real daprd, a held timer callback on one actor does not block another actor's timer.

## Issue reference

Closes #10144

## Checklist

- [x] Code compiles correctly
- [x] Created/updated tests
- [x] Unit tests passing


pkg/actors/internal/timers/inmemory/inmemory.go
```diff
@@ -33,8 +30,17 @@ import (
 
 var log = logger.NewLogger("dapr.runtime.actors.timers.inmemory")
 
+// defaultMaxConcurrency bounds concurrent timer callbacks when
+// Options.MaxConcurrency is unset, capping goroutine growth under a backlog of
+// slow callbacks. Matches defaultBulkPublishMaxConcurrency.
+const defaultMaxConcurrency = 100
+
```
JoshVanL: I don't think we should have a default concurrency limit here, and should be left unbounded.

JoshVanL (changes requested):
Re: bounded pool vs. per-actor sharding: As you have mentioned, please can you look into doing per-actor sharding so we have correct intra-actor ordering. Ideally we would use the kit loops package you linked to be consistent with the rest of the project. Let me know, I'd be happy to draft this and PR to you branch?

JoshVanL:
Thank you for the contribution @MrMint!

MrMint:
> Re: bounded pool vs. per-actor sharding: As you have mentioned, please can you look into doing per-actor sharding so we have correct intra-actor ordering. Ideally we would use the kit loops package you linked to be consistent with the rest of the project. Let me know, I'd be happy to draft this and PR to you branch?

Appreciate the quick feedback @JoshVanL , I'll make those updates!

pkg/actors/internal/timers/inmemory/inmemory.go
MrMint: Sounds good to me, removed it.

MrMint:
Think I got all of the edge cases, let me know if I missed anything.

JoshVanL:
Thanks @MrMint, I think this is worth a backport to 1.18- please can you add an entry here https://github.com/dapr/dapr/blob/master/docs/release_notes/v1.18.2.md ?

JoshVanL (approved):

marcduiker:
@holopin-bot @MrMint runtime-badge Thank you! Here's a digital badge as a small token of appreciation.

PR #10155, opened 2026-07-03
Fix workflows becoming permanently unrecoverable after a save with a timer in the inbox

acroca:
`TimerFired` events are appended to the in-memory inbox when a timer reminder fires, but are by design never persisted as `inbox-NNNNNN` keys, their durable home is the scheduler job, which is only acked once the run that consumed them commits. `GetSaveRequest` however counted them in the metadata `inboxLength` and in the upsert index math.

Any save taken while a timer event was still in the inbox, a workflow stall (version mismatch or payload oversize) on a timer wake, or an abandoned ContinueAsNew, therefore declared an inbox key that was never written. Since #10058 made a metadata/key mismatch a hard load error, every subsequent load fails: status queries error, the wake-up reminder retries forever, and both normal and force purge fail because they load state first. The instance is permanently stuck, with no recovery short of manual state-store surgery. The index shift could also persist the timer event at the wrong key and silently drop a real event added while the timer sat in the inbox.

The fix: `GetSaveRequest` now builds its inbox view via `persistableInbox()`, excluding timer events from both the persisted operations and the metadata count, the same rule `ClearInbox` already applied.

### DIFF
--- a/tests/integration/suite/daprd/workflow/versioning/stalled/timerinbox.go
+++ b/tests/integration/suite/daprd/workflow/versioning/stalled/timerinbox.go
@@ -1,0 +1,142 @@
+/*
+Copyright 2026 The Dapr Authors
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package stalled
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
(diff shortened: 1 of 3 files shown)


tests/integration/suite/daprd/workflow/versioning/stalled/timerinbox.go
```diff
@@ -0,0 +1,142 @@
...
+		assert.True(c, found, "expected an ExecutionStalled event to be persisted to the state store")
+	}, time.Second*40, time.Millisecond*50)
+
+	// The stall save ran while the in-memory inbox still held the TimerFired
+	// event that woke the workflow. TimerFired events are never written as
+	// inbox-NNNNNN keys, so the metadata must not count them either.
+	_, metaRaw := d.workflow.DB().ReadStateValue(t, ctx, string(id), "metadata")
+	var metadata backend.BackendWorkflowStateMetadata
+	require.NoError(t, proto.Unmarshal(metaRaw, &metadata))
+	inboxKeys := d.workflow.DB().ReadStateValues(t, ctx, string(id), "inbox")
```
cicoyle, line 122: Can you add a check that loops over `inboxKeys` and asserts that there is no persisted inbox row with `TimerFired`?

cicoyle (changes requested):
mostly lgtm, just one comment

cicoyle (approved):

PR #10157, opened 2026-07-03
workflow: persist terminal state before recursive terminate cascade

### LINKED ISSUE #10154: Terminating a running workflow leaves it as a "zombie" (status stays RUNNING, terminate not durably applied, the terminate call hangs) - Dapr 1.18.1
### In what area(s)?

/area runtime
(Workflows - `wfengine` / durabletask-go backend; actor reminders via Scheduler)

### What version of Dapr?

- Dapr runtime (daprd sidecar): **1.18.1** (`ghcr.io/dapr/dapr:1.18.1`)
- Dapr .NET SDK: **1.18.4** (`Dapr.Workflow` / `Dapr.Client`)

### Expected Behavior

After a `TerminateWorkflow` call on a running root workflow (recursion left at the Dapr default):
- an `ExecutionCompleted{TERMINATED}` event is durably appended to the workflow history,
- the actor inbox is drained (`inboxLength = 0`),
- the runtime status becomes `TERMINATED`,
- and any subsequent `TerminateWorkflow` / `GetWorkflow` calls are idempotent and return promptly.

### Actual Behavior

`daprd` logs the workflow as terminated and recursively drops all child work items, **but the
workflow actor's durable state is never finalized**:

- **no** `ExecutionCompleted{TERMINATED}` event is appended to history (last event is a normal
  `TaskCompleted`),
- the `ExecutionTerminated` event is left **undrained** in the actor inbox (`metadata.inboxLength = 1`),
- so `GetWorkflow` keeps returning **RUNNING** indefinitely, and
- the **`TerminateWorkflow` call itself hangs and never returns** - the API waits for the terminate to
  be durably applied to the workflow actor, which never happens. This affects the **first** call too
  (confirmed app-side: the caller's post-terminate step - marking the task terminal - never ran, so the
  first `TerminateWorkflowAsync` never returned).

The app pod and daprd sidecar stay healthy the entire time (0 restarts); placement and scheduler are
healthy. This looks like a **non-atomic terminate finalize** in the workflow engine - same class as
#9258 ("Non-atomic workflow creation causes zombie workflows").

**Restarting the app pod does NOT recover it (verified).** After a restart the Scheduler re-fires an
actor reminder and the workflow actor reactivates; daprd **re-logs** the parent
`'<Workflow>' completed with a TERMINATED status.` and re-drops children - but the durable state is
**still never written** (state store unchanged, `inboxLength` still 1). On the restarted pod the
**child** workflows get both the backend log and the orchestrator-actor persist line
(`Workflow Actor '<child>': workflow completed with status 'ORCHESTRATION_STATUS_COMPLETED'`), while
the **terminated parent** gets **only** the `durabletask.backend` line and **no**
`actors.targets.orchestrator` completion at all. So the backend computes the terminated orchestration
but the **workflow orchestrator actor target never persists the terminal state** for the root.

### Steps to Reproduce the Problem (Occurred rarely and only once after upgrading to v1.18)

1. Dapr 1.18.1 with **Scheduler-based actor reminders** and a **SQL Server** actor state store.
2. Start a **root** workflow (no parent) that fans out to many child workflows/activities and let it
   reach steady mid-run progress (in our case: 16 child workflows × 8-10 activities each, ~72% done).
3. Call the .NET SDK `DaprWorkflowClient.TerminateWorkflowAsync(instanceId)` **once** - the call
   **hangs and never returns** (daprd still logs the server-side lines below).
4. Observe in daprd:
   - `'<Workflow>' completed with a TERMINATED status.`
   - one `... : workflow already completed; dropping work item` per child.
5. Despite that log, the durable state is never written:
   - `GetWorkflow(instanceId)` → **RUNNING** (never becomes TERMINATED),
   - any further `TerminateWorkflow(instanceId)` also **hangs**.
6. Inspect the actor state store for the instance: `metadata.inboxLength = 1`, an undrained
   `inbox-000000` = `ExecutionTerminated{recurse=true}`, and **no** `ExecutionCompleted{TERMINATED}`
   in history.
7. **Restart the app pod → the workflow is still RUNNING.** daprd re-logs
   `'<Workflow>' completed with a TERMINATED status.` on reactivation, but the state store is
   unchanged; only `PurgeInstance` clears it.

### Evidence

Anonymized. Instance id: `ABC_10d59ed4-86c3-4780-af19-de60cab10c0a`; app id `com-proj1-abc`;
workflow actor type `dapr.internal.<namespace>.com-proj1-abc.workflow`.

**daprd sidecar - the decisive lines:**

```
06:39:58.661  ABC_10d59ed4-...: 'AbcWorkflow' completed with a TERMINATED status.   scope=dapr.wfengine.durabletask.backend
06:39:58.712  <childId>: workflow already completed; dropping work item              scope=dapr.wfengine.durabletask.backend
              (repeats per child) ... last at 06:39:59.146
              (after this: total silence for this workflow actor - no reminder, no turn, no error)
```

**Durable actor state (SQL Server state store), decoded protobuf:**

```
metadata      = 0x08 01 10 5C 18 01  ->  inboxLength = 1, historyLength = 92, generation = 1
inbox-000000  (31 bytes)             ->  ExecutionTerminated{ recurse = true }   -- UNDRAINED
history-000091 (last event)          ->  TaskCompleted                            -- NO ExecutionCompleted{TERMINATED}
```

**On restart, parent skips the orchestrator-actor persist (children don't):**

```
# child (finalizes): both backend + orchestrator-actor persist
scope=dapr.wfengine.durabletask.backend          <child>: '<ChildWorkflow>' completed with a TERMINATED status.
scope=dapr.runtime.actors.targets.orchestrator   Workflow Actor '<child>': workflow completed with status 'ORCHESTRATION_STATUS_COMPLETED'

# terminated parent (never persists): backend line only, no orchestrator-actor line at all
scope=dapr.wfengine.durabletask.backend          ABC_10d59ed4-...: '<Workflow>' completed with a TERMINATED status.
(no "Workflow Actor 'ABC_10d59ed4-...': workflow completed with status ORCHESTRATION_STATUS_TERMINATED")
```

**Narrowing (what it is NOT):**

- **Not a state-store write failure** - no etag/optimistic-concurrency/transaction/"failed to save"
  error appears in daprd during the finalize window; the completion write is simply never committed.
- **Not placement** - single app replica, 0 pod/sidecar restarts, no rebalancing.
- **Scheduler owns the workflow-actor reminders** (etcd keys like
  `dapr/jobs/actorreminder||<ns>||dapr.internal.<ns>.com-proj1-abc.workflow||<instance>||new-event-tc-8`)
  and logs no error at terminate. (On a pod restart the reminder *does* re-fire and the actor
  reactivates - but it re-processes the terminate without persisting; see the restart section above.)

<details>
<summary>Attached files</summary>

- `daprd-sidecar.log` - start, TERMINATED, child drops, scheduler init/stream lines.
- `daprd-after-restart.log` - after an app-pod restart: parent re-logs TERMINATED (backend only) while
  children also get the `actors.targets.orchestrator` persist line; proves the parent persist is skipped.
- `daprd-errors-warnings.log` - non-benign daprd warnings for the terminate window.
- `scheduler-actorreminders.log` - Scheduler (etcd) actor-reminder keys for the instance.
- `statestore-evidence.md` - full SQL queries/results + protobuf decoding.
- `app-abc.log` - application workflow log (start → 72%, then no completion).

</details>

### Suspected area / ask

Primary suspect is the **workflow orchestrator actor target persist path for a *terminated*
orchestration** in the durabletask-go dapr backend: the backend logs `completed with a TERMINATED
status` but no `ORCHESTRATION_STATUS_TERMINATED` is ever persisted for the root (no
`actors.targets.orchestrator` completion, unlike children), so history/inbox/metadata are never
updated. This repeats on every reactivation (pod restart does not help).

Requests:
- Persist the terminal `ExecutionCompleted{TERMINATED}` (and drain the inbox) for a terminated root
  workflow, atomically with logging "completed with a TERMINATED status".
- On workflow-actor reactivation with a pending `ExecutionTerminated` in the inbox, actually apply it
  rather than re-logging completion without persisting.
- Make `TerminateWorkflow` not block on the durable apply - the **first** call currently hangs
  forever; it should return promptly once the terminate is accepted, and repeated calls should be
  idempotent.
- Provide a way to **force** purge a workflow (recursive with children)  from the statestore and from Scheduler 

### Workaround

- **Restarting the app pod does NOT recover it** (verified) - the actor reactivates and re-processes
  the terminate but never persists.
- `PurgeInstance` - `PurgeInstance` doesn't fail as result or workflow not in  terminal state.
- Manually deleting the Workflow from the statestore, cause the dapr sidecar kept logging warning and errors about the missing workflow, flooding the central logging.
- Clean reinstall of Dapr (including PVCs) - resolve the issue

### Release Note

RELEASE NOTE: FIX Workflow terminate is now applied atomically and the terminate call returns
promptly; terminating a running workflow no longer hangs the `TerminateWorkflow` call or leaves the
instance stuck in RUNNING with an undrained termination event.

[daprd-sidecar.log](https://github.com/user-attachments/files/29635119/daprd-sidecar.log)
[statestore-evidence.md](https://github.com/user-attachments/files/29635118/statestore-evidence.md)
[scheduler-actorreminders.log](https://github.com/user-attachments/files/29635117/scheduler-actorreminders.log)
[app-abc.log](https://github.com/user-attachments/files/29635120/app-abc.log)
[daprd-errors-warnings.log](https://github.com/user-attachments/files/29635121/daprd-errors-warnings.log)

JoshVanL:
Terminating a running workflow with child workflows could leave the parent permanently stuck in RUNNING. The engine delivered the recursive ExecutionTerminated to every child inline, before the parent's terminal state was persisted, so one failed child delivery (unreachable host, transient scheduler error on the child's wake-up reminder create) abandoned the work item and rolled back the parent's completion. The ExecutionTerminated stayed undrained in the parent's inbox, no ExecutionCompleted was written to history, and every reactivation repeated the same failing fan-out; only a purge could clear the instance. The fan-out also ran while holding the parent actor's lock, deadlocking against children concurrently reporting their completion back to the parent, which could hang the TerminateWorkflow call itself.

durabletask-go now emits the cascading messages instead of delivering them inline. The workflow actor persists its terminal state and drains its cascade as a separate idempotent step, re-derived from durable history and retried through the empty-inboeached. Same-app children receive a cascade-terminate reminder carrying the event as data (like timer reminderkes no actor locks and keeps retrying while the child is unavailable; the reminder is created and consumed bire format crosses a version boundary. Cross-app children keep the existing AddWorkflowEvent delivery, bounded parent/child lock cycle resolves and retries.

Fixes #10154

### DIFF
--- a/pkg/actors/targets/workflow/orchestrator/add.go
+++ b/pkg/actors/targets/workflow/orchestrator/add.go
@@ -26,6 +26,9 @@ const (
 	reminderPrefixStart    = "start"
 	reminderPrefixNewEvent = "new-event"
 	reminderPrefixTimer    = "timer-"
+	// Created on each child workflow actor by a recursively-terminated
+	// parent, carrying the ExecutionTerminated event as reminder data.
+	reminderPrefixCascadeTerminate = "cascade-terminate"
 )
 
 func (o *orchestrator) addWorkflowEvent(ctx context.Context, e *backend.HistoryEvent) error {
--- a/pkg/actors/targets/workflow/orchestrator/add.go
+++ b/pkg/actors/targets/workflow/orchestrator/add.go
@@ -26,9 +26,9 @@ const (
 	reminderPrefixStart    = "start"
 	reminderPrefixNewEvent = "new-event"
 	reminderPrefixTimer    = "timer-"
-	// Created on each child workflow actor by a recursively-terminated
-	// parent, carrying the ExecutionTerminated event as reminder data.
-	reminderPrefixCascadeTerminate = "cascade-terminate"
+	// Created on each child workflow actor by a recursively-terminated parent,
+	// carrying the ExecutionTerminated event as reminder data.
+	reminderCascadeTerminate = "cascade-terminate"
 )
 
 func (o *orchestrator) addWorkflowEvent(ctx context.Context, e *backend.HistoryEvent) error {
--- a/docs/release_notes/v1.18.2.md
+++ b/docs/release_notes/v1.18.2.md
@@ -15,6 +15,7 @@ This update contains the following bug fixes:
 - [Parent workflow becomes stuck when creating a child workflow with an instance ID that is already in use](#parent-workflow-becomes-stuck-when-creating-a-child-workflow-with-an-instance-id-that-is-already-in-use)
 - [Sub-millisecond latencies are omitted from latency metrics](#sub-millisecond-latencies-are-omitted-from-latency-metrics)
 - [Querying workflow instance history crashes the sidecar when no actor state store is configured](#querying-workflow-instance-history-crashes-the-sidecar-when-no-actor-state-store-is-configured)
+- [Terminating a running workflow could leave it permanently stuck in RUNNING](#terminating-a-running-workflow-could-leave-it-permanently-stuck-in-running)
 
 ## SPIFFE SVID component operation propagation
 
@@ -343,3 +344,30 @@ The workflow read and purge paths in the actor backend (`GetInstanceHistory`, `l
 ### Solution
 
 The read and purge paths now return the same actionable error the write path already surfaces (`the state store is not configured to use the actor runtime. Have you set the - name: actorStateStore value: "true" in your state store component file?`) instead of dereferencing a nil state. The history query fails cleanly and the sidecar stays up.
+
+## Terminating a running workflow could leave it permanently stuck in RUNNING
+
+### Problem
+
+Terminating a running workflow that had fanned out to child workflows could leave the parent workflow permanently stuck in the RUNNING status.
+The termination event remained undrained in the workflow's inbox, no `ExecutionCompleted` event was written to history, and repeated `TerminateWorkflow` calls could hang.
+Restarting the application did not recover the instance; only purging it did.
+
+### Impact
+
+You were affected if you terminated running workflows that had child workflows, most likely under load or during transient Scheduler or host unavailability.
+Affected instances could never be terminated and reported RUNNING forever, even though the runtime logged them as terminated.
+Termination of sibling child workflows could also be delayed indefinitely while the parent retried.
+
+### Root Cause
+
+On a recursive terminate, the workflow engine delivered the termination event to every child workflow before the parent's own terminal state was persisted.
+If delivery to any child failed, for example an unreachable host or a transient Scheduler error while creating the child's wake-up reminder, the whole operation was rolled back, discarding the parent's computed terminal state, and every retry repeated the same failing sequence.
+Child delivery also ran while holding the parent workflow actor's lock, which could deadlock against children concurrently reporting their completion back to the parent, causing the terminate call to hang.
+
+### Solution
+
+The parent workflow now persists its terminal state and drains its inbox first, then delivers the termination to its children as a separate, idempotent step that is retried until every child is reached.
+Same-app children are terminated through a durable reminder carrying the termination event, which takes no actor locks and keeps retrying while a child's host is unavailable.
+Cross-app children continue to be terminated through the existing event delivery method.
+A failure to reach a child can no longer roll back the parent's terminal state or block its completion.
--- a/go.mod
+++ b/go.mod
@@ -12,7 +12,7 @@ require (
 	github.com/cloudevents/sdk-go/v2 v2.15.2
 	github.com/coreos/go-oidc/v3 v3.17.0
 	github.com/dapr/components-contrib v1.18.1
-	github.com/dapr/durabletask-go v0.12.2-0.20260630111707-26034a1c0eef
+	github.com/dapr/durabletask-go v0.12.2
 	github.com/dapr/kit v0.18.2
 	github.com/diagridio/go-etcd-cron v0.12.6
 	github.com/evanphx/json-patch/v5 v5.9.0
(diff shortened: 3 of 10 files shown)


pkg/actors/targets/workflow/orchestrator/add.go
```diff
@@ -26,6 +26,9 @@ const (
 	reminderPrefixStart    = "start"
 	reminderPrefixNewEvent = "new-event"
 	reminderPrefixTimer    = "timer-"
+	// Created on each child workflow actor by a recursively-terminated
+	// parent, carrying the ExecutionTerminated event as reminder data.
+	reminderPrefixCascadeTerminate = "cascade-terminate"
```
acroca: This doesn't seem to be used as a prefix anywhere, right?

acroca (commented):
LGTM, just a small comment.
We need to remove the `replace` before merging.

pkg/actors/targets/workflow/orchestrator/add.go
JoshVanL: Yep, it is: 

```go
$ grep -nir . -e "reminderPrefixCascadeTerminate"
./run.go:88:	if reminder.Name == reminderPrefixCascadeTerminate && !runtimestate.IsCompleted(o.rstate) {
./terminate.go:91:		Name:      reminderPrefixCascadeTerminate,
./invoke.go:131:		reminder.Name == reminderPrefixCascadeTerminate:
./add.go:31:	reminderPrefixCascadeTerminate = "cascade-terminate"
```

acroca (dismissed):

cicoyle (dismissed):

cicoyle (approved):

PR #10159, opened 2026-07-05
fix(workflow): trim whitespace from instance ID in HTTP APIs

### LINKED ISSUE #8971: [Workflows] trim whitespace from the end of instanceID when using HTTP APIs
Some people use Workflows via HTTP when beginning to experiment with Workflows. I recommend making this small quality of Life improvement to stop users falling into this trap.

## In what area(s)?

/area runtime

## What version of Dapr?

all

## Expected Behavior

White space is trimmed from the end of the Instance ID when using the HTTP APIS

## Actual Behavior

Whitespace is not trimmed which can easily lead to querying a non-existant workflow instance, when whitespace is accidentally pasted to the end of an instance ID (see example in screenshot)

Image

## Release Note

RELEASE NOTE:

anxkhn:
# Description

The workflow HTTP handlers read the instance ID straight from the URL path or
query string without any normalization. A trailing newline or space accidentally
pasted onto an instance ID is preserved and never matches a stored instance, so
the request silently resolves to a non-existent workflow (for example a 404 on
get) instead of the intended one.

This applies `strings.TrimSpace` to the instance ID in every workflow HTTP
handler that reads one: start, get, terminate, raiseEvent, pause, resume, and
purge (get, pause, and resume go through the shared `workflowInModifier`). Only
the instance ID is trimmed; `workflowComponent`, `workflowName`, and `eventName`
are left untouched.

Trimming both ends (rather than only the trailing side) is safe here because a
valid instance ID cannot legitimately contain surrounding whitespace: instance
IDs are validated on creation to contain only letters, digits, `-`, and `_`, so
whitespace is never part of a real ID.

Tests: added regression subtests that send an instance ID with a trailing space
(`%20`) and a trailing newline (`%0A`) through the Start (query parameter), Get,
and Purge (path parameter) endpoints and assert the handler receives the trimmed
ID and the request succeeds. Because production wraps routes that contain path
parameters so their values are URL-unescaped before the handler runs, the test
server is given a small additive option to mirror that behavior for these cases;
it defaults to off, so existing tests are unaffected.

## Issue reference

Please reference the issue this PR will close: #8971

## Checklist

- [x] Code compiles correctly
- [x] Created/updated tests
- [x] Unit tests passing

### DIFF
--- a/pkg/api/http/http_test.go
+++ b/pkg/api/http/http_test.go
@@ -2686,6 +2686,80 @@ func TestV1Workflow(t *testing.T) {
 		// assert
 		assert.Nil(t, resp.ErrorBody)
 	})
+
+	/////////////////////////////////
+	// INSTANCE ID WHITESPACE TESTS //
+	/////////////////////////////////
+
+	// A trailing newline or space is easily pasted onto an instance ID by
+	// accident. Without trimming it is passed through verbatim and silently
+	// resolves to a non-existent instance. See issue #8971. These tests build
+	// the server with parameter unescaping enabled, mirroring how production
+	// registers routes that contain path parameters.
+
+	fakeServerUnescaped := func(t *testing.T) *fakeHTTPServer {
+		f := newFakeHTTPServer()
+		t.Cleanup(f.Shutdown)
+		f.StartServer(testAPI.constructWorkflowEndpoints(), &fakeHTTPServerOptions{unescapeParameters: true})
+		return f
+	}
+
+	t.Run("Start trims surrounding whitespace from instance ID", func(t *testing.T) {
+		var gotInstanceID string
+		wf.WithClient(func() workflows.Workflow {
+			return fake.NewClient().WithStart(func(ctx context.Context, req *workflows.StartRequest) (*workflows.StartResponse, error) {
+				require.NotNil(t, req.InstanceID)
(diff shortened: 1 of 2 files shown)


pkg/api/http/http_test.go
```diff
@@ -2686,6 +2686,80 @@ func TestV1Workflow(t *testing.T) {
...
+
+		// Trailing space (%20) encoded in the path segment.
+		apiPath := "v1.0/workflows/dapr/myInstanceID%20"
+		resp := fakeServerUnescaped(t).DoRequest("GET", apiPath, nil, nil)
+		assert.Equal(t, 200, resp.StatusCode)
+		assert.Nil(t, resp.ErrorBody)
+		assert.Equal(t, "myInstanceID", gotInstanceID)
+	})
+
+	t.Run("Purge trims surrounding whitespace from instance ID", func(t *testing.T) {
```
mikeee:
```suggestion
	t.Run("Purge trims trailing whitespace from instance ID", func(t *testing.T) {
```

mikeee (commented):
Just a small nit

anxkhn:
friendly ping: mikeee's test-name nit is addressed, and CI looks good aside from the linux integration job (happy to re-run if that was flaky). still looking for a review when someone has a moment. thanks!

PR #10165, opened 2026-07-07
ci: add codespell spell-check for changed files

anxkhn:
Dapr currently has no automated spell checking: there is no `.codespellrc`,
no `.pre-commit-config.yaml`, and no codespell / typos / misspell reference in
any workflow or the Makefile. Identifier, header, message and comment typos are
only caught by chance during review.

This adds a `codespell` step to the existing `static checks` job in
`.github/workflows/dapr.yml`. To avoid retroactively gating the large body of
pre-existing typos, the step spell-checks only the files changed in the pull
request, mirroring the neighbouring "Check white space in .md files" step's
changed-files approach. It lands green on the current tree and blocks new typos
from being introduced going forward.

Details:

- New `.codespellrc` is the single source of truth for both CI and local runs.
  It skips generated code (`*.pb.go`, `*.pb.json`), checksums (`go.sum`,
  `*.sum`), `vendor/*`, `node_modules/*`, the frozen `docs/release_notes/*`, and
  `*.svg`. Skips are written as trailing globs so they still apply when codespell
  is given an explicit file list, as the CI step does.
- A short `ignore-words-list` (`aks`, `aci`, `te`, `notin`, `showin`) covers
  correct tokens codespell can flag: Azure product names (AKS, ACI), the HTTP
  `TE` header, the Kubernetes `NotIn` selector operator, and the Grafana
  `showIn` schema key. Each is a real token in the tree, not a typo.
- The step is gated `if: github.event_name == 'pull_request'` and runs
  `git diff --name-only --diff-filter=ACMR FETCH_HEAD HEAD | xargs
  --no-run-if-empty pipx run codespell --` after fetching the PR base. `pipx`
  and Python are preinstalled on `ubuntu-latest`.

I am happy to instead scope this to a filed issue first, or to adjust the
skip / ignore lists, if the maintainers prefer.

## Issue reference

## Checklist

- [x] Code compiles correctly
- [ ] Created/updated tests
- [ ] Unit tests passing

### DIFF
--- a/.codespellrc
+++ b/.codespellrc
@@ -1,0 +1,10 @@
+[codespell]
+# Generated code, vendored deps, checksums and frozen release notes are not
+# hand-authored, so exclude them from spell checking. Directory skips use a
+# trailing glob so they also apply when codespell is passed an explicit file
+# list, as the CI step does for the files changed in a pull request.
+skip = .git,*.pb.go,*.pb.json,go.mod,go.sum,*.sum,vendor/*,node_modules/*,docs/release_notes/*,*.svg
+# Words codespell flags that are correct here: Azure product names (AKS, ACI),
+# the HTTP "TE" header, the Kubernetes "NotIn" selector operator and the Grafana
+# "showIn" schema key.
+ignore-words-list = aks,aci,te,notin,showin
--- a/.github/workflows/dapr.yml
+++ b/.github/workflows/dapr.yml
@@ -80,6 +80,14 @@ jobs:
             echo >&2
             exit 1
           fi
+      - name: Check spelling with codespell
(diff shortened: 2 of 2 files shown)


mikeee (commented):
I’m not too keen on adding unpinned dependencies to CI — this could cause extra work down the line _when_ changes happen. It also adds a bit too much friction, and hard-fail checks tend to invite reruns. We already get a fair bit of coverage from Copilot agent reviewers on PRs, so I’m not convinced this extra step earns its keep.

TL;DR: I’m not really in favour of this addition, unless we’re solving a very specific problem.

anxkhn:
thanks for the feedback, i'll close this.

Issue #8236, opened 2024-10-24
Add more descriptive logging for component startup failure


lor1113:
## In what area(s)?

> /area runtime

## Describe the feature
Currently, at least in the scenario I am currently experiencing, component startup failures (when the dapr sidecar fails) do not explicitly log what component has caused the logged failure. As an example (directly taken from our own logs)

```
time="2024-10-24T14:16:38.96601361Z" level=info msg="Loading components…"
time="2024-10-24T14:16:38.966165869Z" level=debug msg="Loading component: kubernetes (secretstores.kubernetes/v1)" 
time="2024-10-24T14:16:38.966859016Z" level=info msg="Component loaded: kubernetes (secretstores.kubernetes/v1)"
time="2024-10-24T14:16:38.968072409Z" level=info msg="Dapr is shutting down"
time="2024-10-24T14:16:38.968092427Z" level=debug msg="Graceful shutdown timeout: 5s"
time="2024-10-24T14:16:38.968190794Z" level=info msg="Shutting down component kubernetes (secretstores.kubernetes/v1)"
time="2024-10-24T14:16:38.968228115Z" level=info msg="Dapr runtime stopped" 
time="2024-10-24T14:16:38.968274223Z" level=debug msg="stopping workload cert expiry watcher"
time="2024-10-24T14:16:38.96829895Z" level=fatal msg="Fatal error from runtime: failed to load components: rpc error: code = Unknown desc = Secret XXX not found" 
```
Note that this is running ver=1.13.4, I have looked through some changelogs and don't see anything that implies this has been changed in later versions, but if it has then feel free to close this obviously.

As you can see the logs do not explicitly specify which component has failed: "kubernetes" is marked as having loaded, but there is no followup "loading component" for any other component before it fails.

I propose that the component responsible for the failure is added to the final "fatal" log, example:

```Fatal error from runtime: failed to load components: Failed to load component {COMPONENT}: errormessage```

or just 

```Fatal error from runtime: Failed to load component {COMPONENT}: errormessage```

## Release Note
RELEASE NOTE: More descriptive logging for component startup failures

Xander-run:
This looks like a great first-time issue to learn from. Could I take it on?

SpiffyEight77:
I would like to work on this issue.
/assign

RohanMishra315:
Hey @SpiffyEight77 , you workin' on this, let me know

SpiffyEight77:
[...]

PR #10175, opened 2026-07-10
feat: add more descriptive logging for component startup failures

magic-peach:
## Summary

Addresses #8236

Improves component startup logging to help operators diagnose initialization issues:

1. **Add 'Initializing component' log message**: Added at the start of component init in the root loop handler, giving operators visibility into when a component begins initialization.

2. **Include timeout value in error messages**: The timeout error message now includes the configured timeout value (e.g., 'init timeout for component X after 5s'), helping operators understand if they need to adjust timeouts.

3. **Fix misleading error message in state sub-processor**: The state sub-processor had a copy-paste error from the lock sub-processor, incorrectly saying 'failed to save lock keyprefix' instead of 'failed to save state configuration'.

4. **Include component name in error messages**: Error messages in lock and state sub-processors now include the component name for easier identification.

These changes make it significantly easier to track component lifecycle and diagnose initialization failures in production environments.

### DIFF
--- a/pkg/runtime/processor/components.go
+++ b/pkg/runtime/processor/components.go
@@ -108,7 +108,7 @@ func (p *Processor) initInline(ctx context.Context, comp compapi.Component) erro
 
 	initErr := p.runInlineInit(initCtx, comp, mgr)
 	if errors.Is(initCtx.Err(), context.DeadlineExceeded) && initErr == nil {
-		initErr = fmt.Errorf("init timeout for component %s", comp.LogName())
+		initErr = fmt.Errorf("init timeout for component %s after %s", comp.LogName(), timeout)
 	}
 	p.reportInline(ctx, comp, operatorv1.EventType_EVENT_INIT, initErr)
 	// Match legacy proc.Init: return the sub-processor's wrapped error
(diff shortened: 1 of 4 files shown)


pkg/runtime/processor/loops/root/component.go
```diff
@@ -32,6 +32,8 @@ const DefaultComponentInitTimeout = time.Second * 5
 func (r *Root) handleInit(ctx context.Context, ev *loops.Init) {
 	comp := ev.Component
 
+	log.Infof("Initializing component: %s", comp.LogName())
+
```
mikeee:
```suggestion
```

This will be a duplicate logging line: - https://github.com/dapr/dapr/blob/e575df7b19234e7094c3aeb3875d26a155eaf428/pkg/runtime/processor/loops/root/component.go#L97

mikeee (changes requested):

magic-peach:
Addressed all review feedback:\n1. Removed the duplicate "Initializing component" log (as flagged by @mikeee)\n2. Changed `%s` to `%w` for proper error wrapping in lock.go and state.go (as suggested by Copilot)\n3. Added timeout value to the inline init path in components.go for consistency (as suggested by Copilot)\n\n@mikeee could you please re-review?

PR #10176, opened 2026-07-10
fix: improve error message when conversation component type not found

magic-peach:
## Summary

When registry.Create returns nil, nil for an unrecognized conversation component type, the error message was wrapping a nil error which produced a confusing message containing "<nil>".

Now returns a clear error message: "conversation component type X not found".

This helps operators quickly identify that the component type is not registered, rather than seeing a nil error.

### DIFF
--- a/pkg/runtime/processor/conversation/conversation.go
+++ b/pkg/runtime/processor/conversation/conversation.go
@@ -17,6 +17,7 @@ package conversation
 
 import (
 	"context"
+	"fmt"
 	"io"
 
 	contribconversation "github.com/dapr/components-contrib/conversation"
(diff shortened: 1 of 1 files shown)


pkg/runtime/processor/conversation/conversation.go
```diff
@@ -59,7 +60,7 @@ func (c *conversation) Init(ctx context.Context, comp compapi.Component) error {
 	}
 
 	if conversate == nil {
-		return rterrors.NewInit(rterrors.CreateComponentFailure, fName, err)
+		return rterrors.NewInit(rterrors.CreateComponentFailure, fName, fmt.Errorf("conversation component type %s not found", comp.Spec.Type))
```
mikeee:
```suggestion
		return rterrors.NewInit(rterrors.CreateComponentFailure, fName, fmt.Errorf("conversation component %s returned a nil instance", comp.Spec.Type))
```

mikeee (commented):
I think this change is misleading - it's guarding against a nil response from a registered component.

Perhaps something like: `conversation component %s returned a nil instance` might make sense, wdyt?

magic-peach:
@mikeee I thik you are right, I made the necessary changes please merge it if there is nothing else blocking the merge

magic-peach:
@nelson-parente could you please merge if nothing else is needed?

Issue #10186, opened 2026-07-10
Make workflow execution latency buckets configurable


pablochacin:
## In what area(s)?

/area runtime

## Describe the feature

Presently, all latency metrics collected by dapr runtime use the same bucket definition, which defaults to a range from `1ms` to `100.000 ms`.  The buckets can be configured in the metrics specs but this will affect all latency metrics.

This range make sense for grpc and http requests, as well as internal operations.  However, workflows can be long running, spanning from seconds to days. 

It would be therefore convenient to allow defining a different bucket values for workflows

## Release Note

RELEASE NOTE: **ADD** Configurable workflow latency bucket. 

RELEASE NOTE:

PR #10187, opened 2026-07-10
feat(metrics): add workflow execution buckets configuration

pablochacin:
# Description

Allows configuring buckets for workflow execution metrics independently of the buckets used for other metrics. This allows adapting to the long-running workflows (e.g. hours)

## Issue reference

https://github.com/dapr/dapr/issues/10186

## Checklist

- [X] Code compiles correctly
- [X] Created/updated tests
- [X] Unit tests passing


pkg/config/configuration.go
```diff
@@ -397,14 +397,49 @@ func (o *OtelSpec) UnmarshalJSON(data []byte) error {
...
 	// Latency distribution buckets. If not set, the default buckets are used.
-	LatencyDistributionBuckets *[]int        `json:"latencyDistributionBuckets,omitempty" yaml:"latencyDistributionBuckets,omitempty"`
-	Rules                      []MetricsRule `json:"rules,omitempty" yaml:"rules,omitempty"`
+	LatencyDistributionBuckets *[]int `json:"latencyDistributionBuckets,omitempty" yaml:"latencyDistributionBuckets,omitempty"`
+	// Workflow execution distribution buckets. If not set or empty, workflow latency
+	// histograms fall back to the shared LatencyDistributionBuckets.
+	WorkflowLatencyDistributionBuckets *[]int `json:"workflowLatencyDistributionBuckets,omitempty" yaml:"workflowLatencyDistributionBuckets,omitempty"`
+	// WorkflowLatencyDistributionUnits is the unit the WorkflowLatencyDistributionBuckets
+	// values are expressed in. It defaults to milliseconds;
+	WorkflowLatencyDistributionUnits *time.Duration `json:"workflowLatencyDistributionUnits,omitempty" yaml:"workflowLatencyDistributionUnits,omitempty"`
```
JoshVanL: This should go into a new `Workflow` block like `HTTP` is if there are multiple workflow fields.

JoshVanL: Is this really necessary? Why is just using milliseconds not enough?

JoshVanL (changes requested):
Please also include integration tests for this change

pkg/config/configuration.go
pablochacin: expressing hours in milliseconds would make the histogram configuration hard to read. in any case is optional.

pablochacin:
> Please also include integration tests for this change

@JoshVanL what kind of integration test are you considering? this is a configuration change, no new functionality (metrics are collected and reported in the same way)

pkg/config/configuration.go
JoshVanL: In which case, should we then accept Go duration strings instead?

JoshVanL:
> > Please also include integration tests for this change
> 
> @JoshVanL what kind of integration test are you considering? this is a configuration change, no new functionality (metrics are collected and reported in the same way)

An integration test whereby we change the configuration and are able to see that the histogram is different when pull the metrics. See https://github.com/dapr/dapr/blob/master/tests/integration/suite/daprd/metrics/http/distributionbuckets/high.go

pkg/config/configuration.go
pablochacin: I considered that, but
a) this will be inconsistent with existing buckets (two ways to define buckets) 
b) then it would make harder (more code) to use the default buckets (now we just have to copy it)

Not sure this is justified. The proposed solution is simple and helps simplify definitions. But not strong opinion against changing it as suggested.

pablochacin:
@JoshVanL Implemented the integration test but it has resulted flaky and can't identify why. In my machine works most of the time, but eventually fails to retrieve the metrics. In the CI fails most of the time. Any hint is appreciated.

```
        	Error Trace:	/Users/runner/work/dapr/dapr/tests/integration/suite/daprd/metrics/workflow/buckets.go:118
        	            				/Users/runner/hostedtoolcache/go/1.26.5/arm64/src/runtime/asm_arm64.s:1447
        	Error:      	Should NOT be empty, but was []
```

tests/integration/suite/daprd/metrics/http/distributionbuckets/util.go
```diff
@@ -11,7 +11,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package distributionbuckets
```
JoshVanL, line 14: Can we move these package files into framework, probably here https://github.com/dapr/dapr/tree/master/tests/integration/framework/metrics

tests/integration/suite/daprd/metrics/workflow/buckets.go
```diff
@@ -0,0 +1,141 @@
...
+func init() {
+	suite.Register(new(buckets))
+}
+
+// buckets tests daprd metrics buckets for workflows
+type buckets struct {
+	daprd     *daprd.Daprd
+	place     *placement.Placement
+	scheduler *scheduler.Scheduler
+}
```
JoshVanL:
```suggestion
  workflow *workflow.Workflow
}
```

tests/integration/suite/daprd/metrics/workflow/buckets.go
```diff
@@ -0,0 +1,141 @@
...
+	app := app.New(t)
+
+	b.daprd = daprd.New(
+		t,
+		daprd.WithPlacementAddresses(b.place.Address()),
+		daprd.WithAppProtocol("http"),
+		daprd.WithAppID("myapp"),
+		daprd.WithInMemoryActorStateStore("mystore"),
+		daprd.WithSchedulerAddresses(b.scheduler.Address()),
+		daprd.WithConfigManifests(t, `
```
JoshVanL:
```suggestion
b.workflow = workflow.New(t,
		workflow.WithDaprds(2),
  workflow.WithDaprdOptions(0, 
		daprd.WithConfigManifests(t, `
```

tests/integration/suite/daprd/metrics/workflow/buckets.go
```diff
@@ -0,0 +1,141 @@
...
+		if strings.HasPrefix(m, metric) && strings.Contains(m, name) && strings.Contains(m, status) {
+			bucket := util.GetBucketFromKey(t, m)
+			buckets = append(buckets, bucket)
+		}
+	}
+
+	slices.Sort(buckets)
+
+	return buckets
+}
```
JoshVanL: Move to framework

tests/integration/suite/daprd/metrics/workflow/custombuckets.go
```diff
@@ -0,0 +1,141 @@
...
+
+			assert.NotEmpty(c, workflowLatencyBuckets)
+			assert.NotEmpty(c, activityLatencyBuckets)
+		}, time.Second*10, time.Millisecond*100)
+
+		expected := []float64{1_000, 5_000, 10_000, 30_000}
+		assert.ElementsMatch(t, expected, workflowLatencyBuckets[:len(workflowLatencyBuckets)-1])
+		assert.ElementsMatch(t, expected, activityLatencyBuckets[:len(activityLatencyBuckets)-1])
+	})
+}
```
JoshVanL, line 113: Please can we cover the default values, as well as a custom one

tests/integration/suite/daprd/metrics/workflow/buckets.go
```diff
@@ -0,0 +1,141 @@
...
+
+	var buckets []float64
+	for m := range metrics {
+		if strings.HasPrefix(m, metric) && strings.Contains(m, name) && strings.Contains(m, status) {
+			bucket := util.GetBucketFromKey(t, m)
+			buckets = append(buckets, bucket)
+		}
+	}
+
+	slices.Sort(buckets)
```
JoshVanL: Shouldn't they always be sorted anyway?

JoshVanL (changes requested):
Thanks @pablochacin - please can we ensure we have a docs PR with this, explaining why one would want to use this and how?

tests/integration/suite/daprd/metrics/workflow/buckets.go
pablochacin: Not sure what you mean "be sorted anyway". Should the test assume they are sorted?

pablochacin:
documentation https://github.com/dapr/docs/pull/5238

JoshVanL (approved):

Issue #7633, opened 2024-03-16
Creating a child workflow with an ID used by another active workflow should properly fail


olitomlinson:
**dapr 1.13.0**
**dotnet SDK 1.13.0**

_as discussed on Discord with @cgillum_ 

**Actual**
Given a workflow instance, and it's instance Id is `123`, and its state is `running`
When a parent workflow performs `WorkflowContext.CallChildWorkflowAsync` on workflow instance `123`
**Then no exception is propagated to the parent workflow.**

**Expected**
Given a workflow instance, and it's instance Id is `123`, and its state is `running`
When a parent workflow performs `WorkflowContext.CallChildWorkflowAsync` on workflow instance `123`
**Then an exception is propagated to the parent workflow (in the same way that `DaprWorkflowClient.ScheduleNewWorkflowAsync` will throw an exception if the instance is already running)**

---

It looks to me that dapr is internally swallowing the exception in the runtime, and not allowing it to propagate to the parent/caller. Internally the runtime looks to be in a retry loop trying to create the Workflow. Expressed by the logs attached in the screen shot.

![image](https://github.com/dapr/dapr/assets/4224880/7140e808-a1d3-471a-8b1d-4fe473706368)

olitomlinson:
cc @philliphoff @cgillum

olitomlinson:
Keep open

olitomlinson:
Keep alive

cgillum:
Updated title to reflect the impacted scenario. Also added the `P1` label: it's important to fix this, but as an edge case I don't consider it a release blocker.

olitomlinson:
Bump

olitomlinson:
Bump

olitomlinson:
bump

olitomlinson:
Bump

acroca:
@olitomlinson https://github.com/dapr/dapr/pull/9537 may fix this issue, will you be able to try again using the upcoming RC?

olitomlinson:
Hey @acroca - I think my original assumption that these issues shared the same root cause was probably **wrong** tbh.

I think the wider question here is, what should the expected behaviour be for this particular use-case? I don't think that has been agreed by anyone yet?

> Actual
>
> Given a workflow instance, and it's instance Id is 123, and its state is running
> When a parent workflow performs WorkflowContext.CallChildWorkflowAsync on workflow instance 123
> Then an error is continually produced in the runtime, and the parent Workflow does not advance forward
>
> Expected
[...]

PR #10192, opened 2026-07-13
workflow: gate instance ID reuse on child workflows being terminal

JoshVanL:
Creating a workflow with the instance ID of a completed, failed, or terminated workflow succeeded even while child workflows started by that previous execution were still running. A live child belongs to the old execution but reports its completion to the parent instance ID, so its events could be delivered into the new execution, and a new execution pinning the same child IDs could collide with the old execution's running children. The create path only checked the runtime status of the instance being recreated; the children recorded in its history were never consulted.

Recreating a terminal workflow now verifies that every child workflow of the previous execution, checked recursively and across app boundaries, is also in a terminal state or has been purged, before the state is reset. A still-running or unverifiable descendant rejects the create with AlreadyExists naming the blocking child. Purge continues to free an instance ID unconditionally, and creates with a fresh instance ID take the same path as before with no additional work.

The check is carried on the existing WaitForRuntimeStatus stream via a new request metadata flag rather than a new actor method, so it is compatible across a version boundary: an older daprd hosting a child ignores the flag and reports that child's own status, truncating the recursion there instead of failing the create. The verdict rides in the response status code (404 purged, 409 non-terminal subtree, 200 regular metadata) rather than a handler error, because CallStream retries handler errors through the built-in resiliency policy. Each per-child hop is bounded by a 5 second timeout because the walk runs under the creating actor's lock while an abandoned child may concurrently be delivering its completion back into the terminal parent; on timeout the create is rejected and can be retried.

### DIFF
--- a/pkg/actors/targets/workflow/orchestrator/create.go
+++ b/pkg/actors/targets/workflow/orchestrator/create.go
@@ -17,12 +17,14 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"strings"
 
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 	"google.golang.org/protobuf/proto"
 
 	wfenginestate "github.com/dapr/dapr/pkg/runtime/wfengine/state"
+	"github.com/dapr/durabletask-go/api"
 	"github.com/dapr/durabletask-go/api/protos"
 	"github.com/dapr/durabletask-go/backend"
 	"github.com/dapr/durabletask-go/backend/runtimestate"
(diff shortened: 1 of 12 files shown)


pkg/actors/targets/workflow/orchestrator/create.go
```diff
@@ -101,17 +101,29 @@ func (o *orchestrator) createIfCompleted(ctx context.Context, rs *backend.Workfl
...
 	if o.activityResultAwaited.Load() {
 		return fmt.Errorf("a terminated workflow with ID '%s' is already awaiting an activity result", o.actorID)
 	}
+
+	// An ID is reusable only once the previous execution's entire child
+	// workflow tree is terminal: a still-running descendant could deliver
+	// events from the old execution into the new one.
+	if err := o.childrenTerminalCheck(ctx, state); err != nil {
+		return status.Errorf(codes.AlreadyExists, "cannot recreate workflow with ID '%s': %s", o.actorID, err.Error())
+	}
```
javier-aliaga, line 124: Should we treat all errors as AlreadyExists? Can the context be canceled or timeout? Would it even matter?

cicoyle (dismissed):

cicoyle:
Once CI and the above feedback is addressed this is g2g 👍🏻

pkg/actors/targets/workflow/orchestrator/create.go
JoshVanL: Done 👍

olitomlinson:
I'd be interested to see how this change effects this issue here :

https://github.com/dapr/dapr/issues/7633

cicoyle (approved):

JoshVanL:
> I'd be interested to see how this change effects this issue here :
> 
> #7633

Not quite, I have just tested it and the parent will still wait in a RUNNING state until the underlying workflow instance ID is purged.

JoshVanL:
@olitomlinson see https://github.com/dapr/dapr/pull/10197

PR #10197, opened 2026-07-13
Workflow: Fail child when target child ID is already in use

### LINKED ISSUE #7633: Creating a child workflow with an ID used by another active workflow should properly fail

JoshVanL:
Creating a child workflow with an explicit instance ID that belongs to
another active workflow left the parent stuck forever: the dispatch
error was treated as recoverable and the CreateWorkflowInstance call was
retried indefinitely via the wake-up reminder, with no error ever
surfaced to the workflow code.

The parent's awaited child workflow task now fails immediately with a
ChildWorkflowInstanceFailed event (errorType
WorkflowInstanceAlreadyExists) carrying the callee's "already exists"
message, mirroring the existing access-policy denial handling. The
failure is retriable, so a retry policy on the child workflow call can
re-attempt and succeed once the instance ID becomes free. The same
handling applies to the terminal-parent-with-live-subtree rejection,
cross-app child workflows, and child workflows re-driven by rerun.
Detached spawns onto an occupied instance ID are dropped with a
warning, matching their fire-and-forget semantics. The occupying
workflow is never affected.

Fixes #7633

---

    Move state message dispatch into orchestrator/messages package

    Moves the outbound state message dispatch out of the orchestrator
    package into a new orchestrator/messages sub package, following the
    same per-actor-instance pattern as the signing sub package.

    The Messages struct carries the dispatch dependencies (app ID, actor
    identity, router, actor type builder, signer) and is wired in
    initOrchestrator alongside Signing. Failing a child workflow task is
    injected as a callback so the package has no dependency on the
    orchestrator's reminder machinery.

    childwf.go is split accordingly: the gRPC status helpers
    (HasGRPCStatusCode, IsPermissionDenied, IsAlreadyExists,
    GRPCStatusMessage) and the child workflow failure error type constants
    move to the messages package, while callChildWorkflows and
    failChildWorkflowTask stay in the orchestrator. DispatchResult also
    moves there as the shared result type for message and activity
    dispatch, and activity.go now reuses the shared access policy error
    constants instead of duplicating the literals.


pkg/actors/targets/workflow/orchestrator/messages/status.go
```diff
@@ -0,0 +1,70 @@
...
+	return HasGRPCStatusCode(err, codes.PermissionDenied)
+}
+
+func IsAlreadyExists(err error) bool {
+	return HasGRPCStatusCode(err, codes.AlreadyExists)
+}
+
+// HasGRPCStatusCode checks whether the error (possibly wrapped) contains the
+// given gRPC status code. Walks both single-error and multi-error chains.
+func HasGRPCStatusCode(err error, code codes.Code) bool {
```
cicoyle: Are we wanting this to be exported to be used elsewhere outside this pkg? As-is I dont see it used elsewhere even tho its exported.

pkg/actors/targets/workflow/orchestrator/childwf.go
```diff
@@ -90,8 +88,15 @@ func (o *orchestrator) callChildWorkflows(ctx context.Context, startEventName st
...
-				return o.failChildWorkflowACL(ctx, e.GetEventId(), err)
+			if messages.IsPermissionDenied(err) {
+				log.Warnf("Workflow actor '%s': child workflow denied by access policy: %v", o.actorID, err)
+				return o.failChildWorkflowTask(ctx, e.GetEventId(), messages.ErrorTypeAccessPolicyDenied, messages.ErrorMessageAccessPolicyDenied)
+			}
+			// The target instance ID is occupied by another workflow. Fail the
+			// awaited child task rather than retrying forever.
+			if messages.IsAlreadyExists(err) {
+				log.Warnf("Workflow actor '%s': child workflow instance ID '%s' already exists: %v", o.actorID, id, err)
+				return o.failChildWorkflowTask(ctx, e.GetEventId(), messages.ErrorTypeAlreadyExists, messages.GRPCStatusMessage(err))
```
cicoyle: This return exits the whole `or _, e := range es` loop. Shall we join errs and continue instead, then return errs after the loop? Otherwise, I think we silently drop remaining children.

cicoyle (changes requested):

tests/integration/suite/daprd/workflow/reuseid/workflow.go
```diff
@@ -0,0 +1,18 @@
+/*
+Copyright 2024 The Dapr Authors
```
acroca:
```suggestion
Copyright 2026 The Dapr Authors
```

acroca (approved):

PR #10216, opened 2026-07-16
scheduler: recover triggered reminder/job name from metadata prefix

anxkhn:
The scheduler stores each job under a key built by `serialize.buildJobName`
as `"<prefix>||<name>"`, where the prefix is derived from the job metadata
(`actorreminder||<ns>||<type>||<id>` for actor reminders, `app||<ns>||<appid>`
for jobs). When a job fires, `triggerHandler` recovered the reminder/job name to
deliver to the app by splitting the key on the **last** `||` and taking the
trailing segment.

`||` is a permitted character inside reminder and job names: the API edge
(`method.ValidateName`) rejects only `/`, `\`, `#`, `?`, NUL, control characters
and the exact sequences `.` and `..`, and #10120 deliberately made `||` legal
within names. Splitting on the last `||` therefore corrupts any name that itself
contains `||`:

- `foo||bar` was delivered to the app as `bar`
- `a||b||c` was delivered as `c`
- a name ending in `||` (e.g. `name||`) tripped the malformed-name guard and was
  returned `UNDELIVERABLE`, so a validly registered reminder silently never
  fired.

The `ListJobs` read path already moved away from this key splitting in #10120
and recovers the name via `serialize.PrefixFromMetadata` + `strings.TrimPrefix`.
This applies the same approach in the trigger/delivery path, so the delivered
name is unambiguous even when the name (or the actor id) contains `||`, and names
ending in `||` are no longer dropped. `serialize.PrefixFromMetadata` is already
exported and used by the `ListJobs` path, so no new API surface is added.

Only the delivered **name** was affected: actor type, actor id and namespace are
taken from the protobuf `JobMetadata`, not from the key, so they were already
correct.

### Tests

- New unit test `Test_nameFromKey` (`pkg/scheduler/server/internal/cron/cron_test.go`)
  covering names that contain `||`, names ending in `||`, an actor id that
  itself contains `||`, and the error cases (key not matching the metadata
  prefix, unknown target type). This is a regression guard: it fails on the old
  `strings.LastIndex` logic and passes with the fix.
- Extended `tests/integration/suite/daprd/jobs/specialchars.go` to schedule a job
  named `my||job@name` and assert the triggered app callback is
  `job/my||job@name` (previously these suites only exercised a single `|`).
- Extended `tests/integration/suite/actors/reminders/specialchars.go` with a
  subtest that fires a reminder named `remind||me||now` and asserts the app
  receives `/method/remind/remind||me||now`.

## Issue reference

## Checklist

- [x] Code compiles correctly
- [x] Created/updated tests
- [x] Unit tests passing
  
- [ ] Extended the documentation / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_
  
- [ ] Specification has been updated / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_
  
- [ ] Provided sample for the feature / Created issue in the https://github.com/dapr/docs/ repo: dapr/docs#_[issue number]_
  
```

---

## Diff summary (what ships)

4 files, +142/-7 (net; unit test +87, cron.go +28/-7, two integration suites +36/-3):

- `pkg/scheduler/server/internal/cron/cron.go` (+28/-7)
  `triggerHandler` now derives the delivered name via a new unexported helper
  `nameFromKey(key, meta)` that does
  `serialize.PrefixFromMetadata(meta)` -> `HasPrefix` guard -> `strings.TrimPrefix`,
  replacing `strings.LastIndex(name, "||")` + slice + the `len<=idx+2` guard.
  Adds one import (`.../internal/serialize`); `fmt` and `strings` were already
  imported.
- `pkg/scheduler/server/internal/cron/cron_test.go` (+87, new test only)
  `Test_nameFromKey`, 8 table cases.
- `tests/integration/suite/daprd/jobs/specialchars.go` (+9/-3)
  job name `my|job@name` -> `my||job@name`; callback assertion updated; doc
  comment updated.
- `tests/integration/suite/actors/reminders/specialchars.go` (+27)
  new firing subtest for reminder name `remind||me||now`.

## Verification (from the independent verify phase; re-confirmed at draft)

- Unit: `go test ./pkg/scheduler/server/internal/cron` -> `Test_nameFromKey` 8/8
  pass; whole cron + serialize packages pass. RED on the old `LastIndex` algo
  (`foo||bar`->`bar`, `a||b||c`->`c`, `myjob||` dropped), GREEN after.
- Integration (in-process, no Docker):
  `go test -tags=integration ./tests/integration --focus '(daprd/jobs|actors/reminders)/specialchars'`
  passes; RED for the right reason when cron.go is reverted to the old logic
  (reminder fired as `.../remind/now`; job callback `job/job@name`).
- Lint: golangci-lint pinned to v2.10.1 (the version the Makefile pins; the
  machine default is v2.12.2) -> 0 issues on the cron package and both
  integration packages (`--build-tags=integration`). `go vet` clean, `gofmt -l`
  clean on all 4 files.
- Correctness: `key == prefix + name` byte-for-byte because `buildJobName` and
  `PrefixFromMetadata` build the same leading segments (incl. the id segment) via
  the same `joinStrings`; `TrimPrefix` recovers `name` exactly, even with `||`
  inside. Confirmed `req.GetName()` in `triggerHandler` is the composed key
  WITHOUT the `<ns>/jobs/` prefix (go-etcd-cron informer sets it via
  `key.JobName`, which strips the namespace), so `nameFromKey` correctly does not
  strip a leading `/` (unlike the ListJobs read path).

### DIFF
--- a/pkg/scheduler/server/internal/cron/cron.go
+++ b/pkg/scheduler/server/internal/cron/cron.go
@@ -31,6 +31,7 @@ import (
 	"github.com/dapr/dapr/pkg/scheduler/monitoring"
 	"github.com/dapr/dapr/pkg/scheduler/server/internal/etcd"
 	"github.com/dapr/dapr/pkg/scheduler/server/internal/pool"
+	"github.com/dapr/dapr/pkg/scheduler/server/internal/serialize"
 	"github.com/dapr/kit/concurrency"
 	"github.com/dapr/kit/events/broadcaster"
 	"github.com/dapr/kit/events/loop"
@@ -235,19 +236,38 @@ func (c *cron) triggerHandler(req *api.TriggerRequest, fn func(*api.TriggerRespo
 		return
 	}
 
-	idx := strings.LastIndex(req.GetName(), "||")
-	if idx == -1 || len(req.GetName()) <= idx+2 {
-		log.Errorf("Job name is malformed: %s", req.GetName())
+	name, err := nameFromKey(req.GetName(), &meta)
+	if err != nil {
+		log.Errorf("Job name is malformed: %s", err)
 		fn(&api.TriggerResponse{Result: api.TriggerResponseResult_UNDELIVERABLE})
 		return
 	}
 
 	c.connectionPool.Trigger(&internalsv1pb.JobEvent{
 		Key:      req.GetName(),
-		Name:     req.GetName()[idx+2:],
+		Name:     name,
 		Data:     req.GetPayload(),
 		Metadata: &meta,
 	}, c.respHandler(req.GetName(), &meta, fn))
+}
+
+// nameFromKey recovers the reminder/job name delivered to the app from the
+// stored job key. The key is the metadata prefix (built by serialize) followed
+// by the name, so the name is recovered by trimming that prefix rather than by
+// re-splitting the key on the last "||". This is unambiguous even when the name
+// (or the actor id) itself contains "||", which is a permitted character, and
+// it does not drop names that end in "||".
+func nameFromKey(key string, meta *schedulerv1pb.JobMetadata) (string, error) {
+	prefix, err := serialize.PrefixFromMetadata(meta)
+	if err != nil {
+		return "", err
+	}
+
+	if !strings.HasPrefix(key, prefix) {
+		return "", fmt.Errorf("key %q does not match expected prefix %q from its metadata", key, prefix)
+	}
+
+	return strings.TrimPrefix(key, prefix), nil
 }
 
 func (c *cron) respHandler(name string, meta *schedulerv1pb.JobMetadata, fn func(*api.TriggerResponse)) func(api.TriggerResponseResult) {
--- a/pkg/api/universal/actors.go
+++ b/pkg/api/universal/actors.go
@@ -141,6 +141,12 @@ func (a *Universal) RegisterActorReminder(ctx context.Context, in *runtimev1pb.R
 		}
 	}
 
+	if in.GetName() == "" {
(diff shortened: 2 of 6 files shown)

### CODE REVIEW

File: pkg/scheduler/server/internal/cron/cron.go [hunk 1]
```diff
@@ -235,21 +236,40 @@ func (c *cron) triggerHandler(req *api.TriggerRequest, fn func(*api.TriggerRespo
 		return
 	}
 
-	idx := strings.LastIndex(req.GetName(), "||")
-	if idx == -1 || len(req.GetName()) <= idx+2 {
-		log.Errorf("Job name is malformed: %s", req.GetName())
+	name, err := nameFromKey(req.GetName(), &meta)
```
JoshVanL, line 239: Please can we include a check to ensure that the reminder name is not an empty string. This should live in `RegisterActorReminder`, specifically https://github.com/dapr/dapr/blob/master/pkg/api/universal/actors.go#L144

```go
if in.GetName() == "" {
    err = messages.ErrBadRequest.WithFormat("reminder name cannot be empty")
    a.logger.Debug(err)
    return nil, err
}
```

Please can we also add an integration test which verifies you cannot create a reminder with an empty name.

JoshVanL (changes requested):
Thanks @anxkhn

anxkhn:
Thanks, I added the empty reminder name validation in `RegisterActorReminder` and extended the existing invalid-names integration test to cover it. The scoped unit and integration tests and lint checks pass locally.

JoshVanL (approved):

PR #10218, opened 2026-07-16
fix(actors): return error from State when no actor state store is configured

### LINKED ISSUE #10217: Runtime panic (nil pointer) in wfengine LoadWorkflowState when querying workflow history on a sidecar without an actor state store
## In what area(s)?

/area runtime

## What version of Dapr?

Runtime **1.18.1** (CLI 1.18.0), self-hosted / slim mode on macOS (darwin/arm64).
durabletask-go **v0.12.1**.

## Expected Behavior

Calling the Workflow **GetInstanceHistory** API against a sidecar whose configured state
store is **not** an actor state store (no `actorStateStore: "true"`) — or where the
instance's workflow state cannot be loaded — should return a gRPC error to the caller.

This is how the sibling **ListInstanceIDs** call on the same TaskHub gRPC API already
behaves: against the same sidecar it returns a clean error, `rpc error: code = Unknown
desc = no state store with actor support found`.

## Actual Behavior

`daprd` **panics with SIGSEGV (nil pointer dereference)** and the process exits (status 2),
taking the app down with it. A client that can reach the workflow gRPC API can therefore
crash the sidecar — an availability / DoS concern, since the crash is remote-triggerable via
a normal read-only API call.

The nil dereference is in `pkg/runtime/wfengine/state/state.go:685` (`LoadWorkflowState`),
reached from `pkg/runtime/wfengine/backends/actors/actors.go:861` (`GetInstanceHistory`).
`LoadWorkflowState` is invoked with a nil state argument (`{0x0, 0x0}` in the trace) and
dereferences it without a nil check.

Looking at the code, `GetInstanceHistory` does:

```go
ss, err := abe.actors.State(ctx)
if err != nil {
    return nil, err
}
resp, err := state.LoadWorkflowState(ctx, ss, ...)
```

so it appears `abe.actors.State(ctx)` can return a nil state interface with a nil error
when no actor state store is configured, and `LoadWorkflowState` then calls
`state.Get(...)` on it. Either `State()` should return an error in that case (as the
list path effectively surfaces) or the callers need a nil guard.

<details>
<summary>Full stack trace</summary>

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x18 pc=0x109709594]

goroutine 418 [running]:
github.com/dapr/dapr/pkg/runtime/wfengine/state.LoadWorkflowState({0x110e741d8, 0xab132ce79e0}, {0x0, 0x0}, {0xab132a9dbc0, 0x20}, {{0x16b35e6f9, 0xf}, {0x109a29db7, 0x7}, ...})
	/Users/runner/work/dapr/dapr/pkg/runtime/wfengine/state/state.go:685 +0x114
github.com/dapr/dapr/pkg/runtime/wfengine/backends/actors.(*Actors).GetInstanceHistory(0xab131ab3950, {0x110e741d8, 0xab132ce79e0}, 0xab1331e6940)
	/Users/runner/work/dapr/dapr/pkg/runtime/wfengine/backends/actors/actors.go:861 +0xc0
github.com/dapr/durabletask-go/backend.(*grpcExecutor).GetInstanceHistory(0x1112c8e10?, {0x110e741d8?, 0xab132ce79e0?}, 0x104b5b18c?)
	/Users/runner/go/pkg/mod/github.com/dapr/durabletask-go@v0.12.1/backend/executor.go:569 +0x2c
github.com/dapr/durabletask-go/api/protos._TaskHubSidecarService_GetInstanceHistory_Handler.func1(...)
	/Users/runner/go/pkg/mod/github.com/dapr/durabletask-go@v0.12.1/api/protos/orchestrator_service_grpc.pb.go:675 +0xcc
... (grpc middleware chain) ...
github.com/dapr/durabletask-go/api/protos._TaskHubSidecarService_GetInstanceHistory_Handler(...)
	/Users/runner/go/pkg/mod/github.com/dapr/durabletask-go@v0.12.1/api/protos/orchestrator_service_grpc.pb.go:677 +0x140
google.golang.org/grpc.(*Server).processUnaryRPC(...)
	/Users/runner/go/pkg/mod/google.golang.org/grpc@v1.80.0/server.go:1430 +0xc90
google.golang.org/grpc.(*Server).handleStream(...)
	/Users/runner/go/pkg/mod/google.golang.org/grpc@v1.80.0/server.go:1856 +0x89c
created by google.golang.org/grpc.(*Server).serveStreams.func2 in goroutine 263

❌  The daprd process exited with error code: exit status 2
```

The caller (a `durabletask-go` workflow client calling `GetInstanceHistory`) received
`rpc error: code = Unavailable desc = error reading from server: EOF` as the sidecar died.
</details>

1. `dapr init` (self-hosted, runtime 1.18.1).
2. Create a state store component that is **not** an actor state store (e.g. plain
   `state.redis` with **no** `actorStateStore: "true"` metadata):
   ```yaml
   apiVersion: dapr.io/v1alpha1
   kind: Component
   metadata:
     name: statestore
   spec:
     type: state.redis
     version: v1
     metadata:
       - name: redisHost
         value: localhost:6379
       - name: redisPassword
         value: ""
   ```
3. Start a sidecar with those components (no app needed):
   `dapr run --app-id demo --resources-path ./components`
4. Query workflow instance history against that sidecar for any instance ID
   (CLI equivalent of the gRPC `GetInstanceHistory` call):
   `dapr workflow history -a demo <any-instance-id>`
5. The `demo` sidecar (`daprd`) panics with the stack trace above and exits with
   status 2; the client sees `rpc error: code = Unavailable desc = error reading
   from server: EOF`.

**Contrast:** a `ListInstanceIDs` call over the same TaskHub gRPC API against the same
sidecar returns a clean error (`no state store with actor support found`) instead of
crashing — only the history path is missing the nil guard.

Originally observed while pointing a workflow client (an MCP server) at a sidecar whose
app-id did not own the queried instance; that sidecar had a non-actor state store. The
minimal repro above triggers the same panic without any workflow app involved.

## Proposed fix

`actors.State(ctx)` returns a nil interface with a nil error when the runtime is ready
but no actor state store is configured (`a.state` is only set when `storeEnabled`).
`Reminders()` in the same file already guards against this with
`messages.ErrActorRuntimeNotFound`; `State()` is missing the equivalent guard.

I have a one-line fix (mirroring the `Reminders()` guard) plus a regression test ready,
verified against this repro: with the patch, the same `GetInstanceHistory` call returns
`rpc error: code = Internal desc = the state store is not configured to use the actor
runtime. Have you set the - name: actorStateStore value: "true" in your state store
component file?` and the sidecar stays up. PR follows.

## Release Note

RELEASE NOTE: **FIX** Prevent runtime panic (nil pointer dereference) when querying workflow instance history on a sidecar without an actor state store; return an error instead.

jongep86:
# Description

Querying workflow instance history against a sidecar without an actor state store crashed `daprd` with a nil pointer dereference. The three workflow read/purge paths in the actor backend (`GetInstanceHistory`, `loadInternalState`, `purgeWorkflowForce`) call `actors.State(ctx)`, which returns a nil state interface with a nil error when no actor state store is configured, and pass it straight to `state.LoadWorkflowState`, which dereferences it. A single `dapr workflow history` call took the whole sidecar down (exit status 2), so the panic was remotely triggerable through a read-only API.

The write path already guards against exactly this: `orchestrator.executeMethod` returns `messages.ErrActorRuntimeNotFound` when its `actorState` is nil. This mirrors that guard on the three read/purge sites, so they return the same actionable error the schedule path already does (`the state store is not configured to use the actor runtime. Have you set the - name: actorStateStore value: "true" in your state store component file?`) instead of crashing.

`actors.State()` is intentionally left returning `(nil, nil)`: the actor registration path (`orchestrator.New`/`activity.New`) calls it during `RegisterActors` and tolerates a nil state, surfacing the error at use rather than registration. Guarding `State()` itself would break that contract (verified: it makes the existing `nostatestore/{grpc,http,workerclient}` suite fail with `DeadlineExceeded` instead of the clean error).

# Verification

Added integration test `tests/integration/suite/daprd/workflow/nostatestore/history`: a sidecar with a non-actor state store returns `codes.Internal` "the state store is not configured to use the actor runtime" for a `GetInstanceHistory` call and stays up (a second call over the same connection succeeds in returning the same error rather than a transport failure). The full `nostatestore` suite (grpc, http, workerclient, history) passes. Verified end-to-end against the repro in the linked issue: unpatched 1.18.1 panics at `pkg/runtime/wfengine/state/state.go:685`; the patched build returns the error and survives.

# Issue reference

Fixes #10217

# Checklist

- [x] Code compiles correctly
- [x] Created/updated tests
- [x] Extended the documentation / added release note (v1.18.2)


acroca (changes requested):
Thanks for the contribution!
Please:
- Cover this change in an integration test
- Add release notes for 1.18

jongep86:
Both done, and the fix is reworked while I was at it.

- **Integration test**: added `tests/integration/suite/daprd/workflow/nostatestore/history` — a sidecar with a non-actor state store returns `codes.Internal` "the state store is not configured to use the actor runtime" for a `GetInstanceHistory` call and stays up. The full `nostatestore` suite (grpc/http/workerclient/history) passes.
- **Release note**: added to `docs/release_notes/v1.18.2.md`.

While writing the test I found the original approach (guarding `actors.State()`) was too broad — it broke the existing `nostatestore/{grpc,http,workerclient}` suite, which relies on `State()` returning `(nil, nil)` so actor registration succeeds and the error surfaces at use. The fix now mirrors the guard `orchestrator.executeMethod` already uses, on the three read/purge sites in the actor workflow backend (`GetInstanceHistory`, `loadInternalState`, `purgeWorkflowForce`), and leaves `State()` unchanged. Force-pushed.

acroca (changes requested):
Thanks for the rework.
Can you make sure all the paths are tested? At the moment only the `GetInstanceHistory` is covered, we need to also cover the fetch metadata, and purge (forced, recursive and plain).

jongep86:
Added integration coverage for all the affected paths. The `nostatestore` suite now also has:

- `metadata` — `FetchWorkflowMetadata` (`loadInternalState`)
- `purge` — plain, forced (`WithForcePurge`), and recursive (`WithRecursivePurge`)

Forced and recursive purge and the metadata fetch load the actor state directly and return the actor-runtime error. Plain purge dispatches to the workflow actor, which is never located without an actor state store, so it returns a `FailedPrecondition` instead — the test asserts it's a clean gRPC status (i.e. the sidecar answered rather than crashed) and that a follow-up call still succeeds. Full `nostatestore` suite (grpc, http, workerclient, history, metadata, purge) passes.

acroca (approved):

JoshVanL (approved):

cicoyle (approved):

marcduiker:
@holopin-bot @jongep86 runtime-badge Thank you! Here's a digital badge as a small token of appreciation.

PR #10220, opened 2026-07-16
fix(runtime): make binding OPTIONS probe timeout configurable

### LINKED ISSUE #10185: Make the timeout for bindings OPTIONS request configurable
## In what area(s)?

> /area runtime

## What version of Dapr?

`> 1.15.x`

see: [1.15.x](https://github.com/dapr/dapr/blob/v1.15.0/pkg/runtime/processor/binding/send.go#L89)
see: [1.18.x](https://github.com/dapr/dapr/blob/v1.18.1/pkg/runtime/processor/binding/send.go#L106)

## Expected Behavior

The timeout for binding `OPTIONS` discovery requests to be configurable.

## Actual Behavior

A fixed timeout of three seconds was introduced in the `1.16.x` series.  For applications that have a slower startup time (JIT), this timeout can be exceeded.  Particularly if they are running in resource constrained environments.

 - Configure any kind of input binding for an application that has a slow startup ( > 3 seconds )
 - When the application starts up, the Dapr `OPTIONS` probe will time out

## Release Note

RELEASE NOTE: **FIX** Allow binding OPTIONS probe timeout to be configurable.

Advait2912:
A hardcoded 3-second timeout was introduced in Dapr 1.16.x for the HTTP OPTIONS request used to discover input binding subscriptions. Applications with slow startup (JVM, JIT, resource-constrained environments) can exceed this window, causing the binding to not be activated.

This change makes the timeout configurable via:
- CLI flag: --binding-options-timeout (default: 3s)

The default value is preserved so existing behaviour is unchanged.

The constant DefaultBindingOptionsTimeout lives in pkg/config to avoid import cycles with the binding sub-package.

Fixes #10185

# Description

This PR makes the timeout used for the HTTP OPTIONS request that discovers input binding subscriptions configurable.

### Changes

- Add `--binding-options-timeout` CLI flag (default: `3s`)
- Add `DefaultBindingOptionsTimeout` in `pkg/config`
- Thread the timeout through the runtime configuration to the binding processor
- Replace the hardcoded `3 * time.Second` timeout with the configured value
- Add unit tests covering the default timeout fallback and custom timeout configuration

The existing default behavior is preserved.
## Issue reference
Closes #10185


pkg/runtime/processor/binding/send.go
```diff
@@ -103,7 +103,7 @@ func (b *binding) startInputBinding(comp componentsV1alpha1.Component, binding b
 	if isBindingOfExplicitDirection(ComponentTypeInput, m) {
 		isSubscribed = true
 	} else {
-		ctx, cancel := context.WithTimeout(context.Background(), time.Second*3)
+		ctx, cancel := context.WithTimeout(context.Background(), b.bindingOptionsTimeout)
 		defer cancel()
```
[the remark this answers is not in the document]
Advait2912: It's right that using context.Background() means the probe wont inherit cancellation from the caller context. The existing implementation used context.Background() before this change, so I kept this PR focused on making the timeout configurable while preserving existing behavior. I'll look into threading the parent context through the binding startup flow as a separate improvement.

Advait2912:
Hi @nelson-parente ! I've addressed the review feedback, including the test improvements, CI stability update, and documentation clarifications. The workflows also need to be re-approved after the latest push. When you have a chance, I'd appreciate another review. Thanks!

pkg/runtime/processor/binding/send.go
```diff
@@ -103,7 +103,7 @@ func (b *binding) startInputBinding(comp componentsV1alpha1.Component, binding b
 	if isBindingOfExplicitDirection(ComponentTypeInput, m) {
 		isSubscribed = true
 	} else {
-		ctx, cancel := context.WithTimeout(context.Background(), time.Second*3)
+		ctx, cancel := context.WithTimeout(context.Background(), b.bindingOptionsTimeout)
```
cicoyle:
```suggestion
		probeCtx, cancel := context.WithTimeout(ctx, b.bindingOptionsTimeout)
```

pkg/runtime/processor/binding/send.go
```diff
@@ -103,7 +103,7 @@ func (b *binding) startInputBinding(comp componentsV1alpha1.Component, binding b
...
 		defer cancel()
 
 		isSubscribed, err = b.isAppSubscribedToBinding(ctx, comp.Name)
```
cicoyle:
```suggestion
		isSubscribed, err = b.isAppSubscribedToBinding(probeCtx, comp.Name)
```

cicoyle (changes requested):
mostly lgtm, can you address the feedback and add an [integration test](https://github.com/dapr/dapr/tree/master/tests/integration/suite/daprd) for this configurable timeout?

Advait2912:
Hi @cicoyle, thanks for the review! I've addressed the requested feedback by propagating the parent context for the binding probe, adding the integration test for the configurable timeout, and incorporating the remaining documentation update. When you have a chance, I'd appreciate another look. Thanks!

Advait2912:
Hi @cicoyle ! All the requested changes have been addressed and the CI checks have passed. When you have some time, I'd appreciate another review. Thanks!

cicoyle (dismissed):

JoshVanL (approved):

marcduiker:
@holopin-bot @Advait2912 runtime-badge Thank you! Here's a digital badge as a small token of appreciation.

PR #10222, opened 2026-07-17
fix(placement): prevent nil pointer dereference and infinite loop in consistent hash

magic-peach:
## Summary

Fixes multiple safety issues in the consistent hash ring implementation (`pkg/placement/hashing/consistent_hash.go`):

### 1. Infinite loop in `GetLeast` (line 293)

The `for` loop iterating through hosts had no termination condition. If no host passed `loadOK()` (e.g., all hosts were above the load threshold), this would loop forever. Added a bounded iteration (`for range len(c.hosts)`) and returns `ErrNoHosts` if no suitable host is found.

### 2. Nil pointer dereference in `Inc` (line 337)

`Inc()` directly accessed `c.loadMap[host].Load` without checking if the host exists. If a host was removed between the call to `GetLeast` and `Inc`, this would panic. Added the same nil guard that `Done()` already has.

### 3. Panic in `loadOK` on removed host (line 423)

When `loadOK` was called with a host that had been removed from `loadMap`, it panicked with an error message. Changed to `return false` so the caller can try the next host gracefully.

### 4. Divide-by-zero in `MaxLoad` (line 402)

`MaxLoad()` divides `c.totalLoad` by `len(c.loadMap)` without checking if `loadMap` is empty. Added an early return of 1 when there are no hosts.

### DIFF
--- a/pkg/placement/hashing/consistent_hash.go
+++ b/pkg/placement/hashing/consistent_hash.go
@@ -24,7 +24,6 @@ package hashing
 import (
 	"encoding/binary"
 	"errors"
-	"fmt"
 	"maps"
 	"math"
 	"slices"
@@ -291,7 +290,7 @@ func (c *Consistent) GetLeast(key string) (string, error) {
 	idx := c.search(h)
 
 	i := idx
-	for {
+	for range len(c.hosts) {
 		host := c.hosts[c.sortedSet[i]]
 		if c.loadOK(host) {
 			return host, nil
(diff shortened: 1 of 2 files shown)


magic-peach:
I've added unit tests covering the new behaviors:\n- `TestGetLeastBoundedLoad`: empty ring, all hosts overloaded, at least one under load\n- `TestIncMissingHost`: no panic when incrementing nonexistent host\n- `TestMaxLoadEmptyRing`: returns 1 for empty ring\n- `TestLoadOKRemovedHost`: returns false instead of panicking\n\nAll tests pass. @copilot-pull-request-reviewer could you please re-review?

magic-peach:
@nelson-parente

PR #10237, opened 2026-07-19
fix(runtime): only log in-scope subscriptions

### LINKED ISSUE #8319: Only emit the "Found Subscription" log when the subscription is within scope
## In what area(s)?

/area runtime

## Describe the feature

Since 1.13 (I think?) Dapr emits the following `info` level log line per declarative `subscription` :

`Found Subscription: xxxxxx` 

https://github.com/dapr/dapr/blob/24ac3b84a96c399e998e797a0a09719ec98ad1bc/pkg/runtime/runtime.go#L1182

**However this log line is output even for subscriptions that are out scope.** 

From a user/operator perspective, I only want to know about subscriptions that are _within_ scope, as it leads to unnecessary confusion / alarm to see a Dapr emitting one or more logs for a subscriptions that are unexpected on a particular App.

In my example below there are lots of subscriptions being logged out entirely unnecessarily as they are out of scope.

Found Subscription us

AkashKumar7902:
# Description

Only emit the startup `Found Subscription` log for declarative subscriptions that are global or scoped to the current app.

The subscription processor already ignores out-of-scope subscriptions, but the loader logged every discovered resource before that filter ran. The log now uses the shared `common.Scoped.IsAppScoped` predicate, so logging and subscription processing follow the same scope semantics without changing ordering or hot-reload behavior.

The self-hosted scope integration suite captures startup output from two sidecars that share global, empty-scope, per-app, and multi-app subscription manifests. Both HTTP and gRPC variants require each sidecar to log its visible subscriptions and use a bounded `assert.Never` window to reject an out-of-scope log.

Validation at commit `a375f28dfc47803633598d518860acfbb976de3a`:

- `make lint` with the repository-pinned golangci-lint v2.10.1
- `go test -count=1 -tags=unit,allcomponents ./pkg/runtime/...`
- `go test ./tests/integration -timeout=5m -count=1 -v -tags=integration -integration-parallel=false -focus='^daprd/subscriptions/declarative/selfhosted/v2alpha1/scopes/(http|grpc)$'`
- Manual process validation built `daprd` from both the PR and its opening base, then launched real standalone HTTP and gRPC sidecars against shared subscription resources. The base logged an out-of-scope subscription; the PR sidecars logged only global, empty-scope, shared, and matching app-specific resources, and their public metadata endpoints reported the same in-scope topics. All processes shut down cleanly with no listener leaks.

## Issue reference

Please reference the issue this PR will close: Closes #8319

## Checklist

- [x] Code compiles correctly
- [x] Created/updated tests
- [x] Unit tests passing

Focused Dapr integration scenarios and manual process validation pass. The full Kubernetes end-to-end suite was not run. Documentation, specification, and sample changes are not applicable to this logging-only fix.


pkg/runtime/subscriptions.go
```diff
@@ -49,7 +50,9 @@ func (a *DaprRuntime) loadDeclarativeSubscriptions(ctx context.Context) error {
 	}
 
 	for _, s := range subs {
-		log.Infof("Found Subscription: %s", s.Name)
+		if subscriptionIsInScope(a.runtimeConfig.id, s) {
```
cicoyle: Can we use this existing func https://github.com/dapr/dapr/blob/master/pkg/apis/common/scoped.go#L27 - as it looks to be identical logic as this new helper func.

tests/integration/suite/daprd/subscriptions/declarative/selfhosted/v2alpha1/scopes/http.go
```diff
@@ -38,24 +41,36 @@ type http struct {
...
+	h.log1 = logline.New(t, logline.WithCaptureAll())
+	h.log2 = logline.New(t, logline.WithCaptureAll())
 
 	resDir := t.TempDir()
 
 	h.daprd1 = daprd.New(t,
 		daprd.WithAppPort(h.sub.Port()),
 		daprd.WithAppProtocol("http"),
 		daprd.WithResourcesDir(resDir),
+		daprd.WithExecOptions(
```
cicoyle, line 61: Can you add the equivalent test to [the grpc test?](https://github.com/dapr/dapr/blob/master/tests/integration/suite/daprd/subscriptions/declarative/selfhosted/v2alpha1/scopes/grpc.go)

cicoyle (changes requested):
Hey @AkashKumar7902 thx for the PR

Mostly lgtm, but can you address the feedback from me + copilot?

tests/integration/suite/daprd/subscriptions/declarative/selfhosted/v2alpha1/scopes/http.go
```diff
@@ -128,14 +143,24 @@ scopes:
...
 }
 
 func (h *http) Run(t *testing.T, ctx context.Context) {
 	h.daprd1.WaitUntilRunning(t, ctx)
 	h.daprd2.WaitUntilRunning(t, ctx)
 
+	for _, name := range []string{"sub1", "sub2", "sub3", "sub5"} {
+		h.log1.EventuallyContains(t, "Found Subscription: "+name, time.Second*15, time.Millisecond*10)
+	}
+	assert.False(t, h.log1.Contains("Found Subscription: sub4"))
```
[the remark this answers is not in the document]
AkashKumar7902: Fixed in a375f28df. The bounded negative assertion now uses assert.Never, directly expressing that app-one must not log the out-of-scope app-two subscription during the observation window.

tests/integration/suite/daprd/subscriptions/declarative/selfhosted/v2alpha1/scopes/http.go
AkashKumar7902: Fixed in a375f28df. The integration suite now includes equivalent gRPC-mode sidecar coverage, including global, empty, matching, multi-app, and out-of-scope manifests with bounded negative log assertions.

pkg/runtime/subscriptions.go
AkashKumar7902: Fixed in a375f28df. The local scope helper was removed and the log filter now reuses common.Scoped.IsAppScoped, matching the subscription processor semantics for global, empty, single-app, and multi-app scopes.

tests/integration/suite/daprd/subscriptions/declarative/selfhosted/v2alpha1/scopes/http.go
```diff
@@ -128,14 +143,24 @@ scopes:
...
+
+	for _, name := range []string{"sub1", "sub2", "sub4", "sub5"} {
+		h.log2.EventuallyContains(t, "Found Subscription: "+name, time.Second*15, time.Millisecond*10)
+	}
+	assert.False(t, h.log2.Contains("Found Subscription: sub3"))
```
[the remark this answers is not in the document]
AkashKumar7902: Fixed in a375f28df. This side now also uses assert.Never, directly expressing that app-two must not log the out-of-scope app-one subscription during the bounded window.

AkashKumar7902:
@cicoyle PTAL

cicoyle (approved):

PR #10238, opened 2026-07-19
fix(placement): reliably fire dissemination timeouts

### LINKED ISSUE #9885: [1.18] placement: dissemination timeout does not fire when slow replica connects mid-round, leaving stream undisconnected

AkashKumar7902:
# Description

Replace the placement dissemination timeout queue with a single re-armable timer shared by the placement server and actor runtime.

Placement has at most one active dissemination round. A new round therefore supersedes the previous timer, while `Dequeue` cancels only the version that still owns it. The timer uses scheduled-callback identity under a mutex, waits for any already-fired callback during `Close`, and never lets a callback outlive its owning event loop.

The server ignores a queued timeout after returning to `REPORT`, while the actor accepts one only during `LOCK`. This prevents a same-version timeout event that was already queued before successful phase completion from closing a healthy stream. The actor-specific queue implementation is removed in favor of the shared primitive.

Validation at commit `0532f8dbbe085a7bcc936fab4ead28cec1182019`:

- `make lint` with the repository-pinned golangci-lint v2.10.1
- `go vet -tags=unit,allcomponents ./pkg/internal/placement/timeout/... ./pkg/placement/internal/loops/disseminator/... ./pkg/actors/internal/placement/loops/disseminator/...`
- `go test -race -count=10 -tags=unit,allcomponents ./pkg/internal/placement/timeout/... ./pkg/placement/internal/loops/disseminator/... ./pkg/actors/internal/placement/loops/disseminator/...`
- `go test ./tests/integration -timeout=15m -count=10 -v -tags=integration -integration-parallel=false -focus='^(placement/timeout/slowreplica|daprd/placement/daprddisstimeout/flag)$'`
- Manual server-boundary validation launched three fresh real placement processes. In every run, an external gRPC client completed a healthy round, withheld the slow stream's acknowledgement, observed only that stream close with `DeadlineExceeded`, and completed recovery on the healthy stream.
- Manual actor-boundary validation launched real placement and `daprd` processes plus an external actor application. Across three timeout/reconnect cycles, `daprd` closed and re-established the placement stream and public actor invocation returned HTTP 200 after every recovery. All processes shut down with no listener leaks.

## Issue reference

Please reference the issue this PR will close: Closes #9885

## Checklist

- [x] Code compiles correctly
- [x] Created/updated tests
- [x] Unit tests passing

Focused Dapr integration scenarios and manual process validation pass. The full Kubernetes end-to-end suite was not run. Documentation, specification, and sample changes are not applicable to this internal concurrency fix.


pkg/internal/placement/timeout/timeout.go
```diff
@@ -14,47 +14,114 @@ limitations under the License.
...
 	"github.com/dapr/kit/events/loop"
-	"github.com/dapr/kit/events/queue"
 )
 
 type Options struct {
 	Loop    loop.Interface[loops.EventDisseminator]
 	Timeout time.Duration
 }
 
+type scheduledTimer struct {
```
JoshVanL, line 28: This file is byte-for-byte the pre-PR placement implementation and carries the same worker-exit race this PR fixes. The actor-side disseminator drives it with the identical back-to-back pattern in `order.go:66-68`:

```go
d.timeoutQ.Dequeue(d.timeoutVersion)
d.timeoutVersion++
d.timeoutQ.Enqueue(d.timeoutVersion)
```

and its `handleTimeout` still has only the old `// Ignore old timeouts.` guard, so it's also missing the `currentOperation == REPORT` fix. Rather than fixing placement as a special case and letting the two copies diverge, could we extract the new re-armable `Timeout` into a shared component used by both packages (or fix `kit/events/queue`)? Otherwise this race stays live on the actor path and the copies drift.

pkg/placement/internal/loops/disseminator/timeout/timeout.go
```diff
@@ -14,47 +14,114 @@ limitations under the License.
...
+	t.version = version
+	generation := t.generation
+	scheduled := new(scheduledTimer)
+	t.callbacks.Add(1)
+	t.timer = scheduled
+	scheduled.timer = time.AfterFunc(t.timeout, func() {
+		defer scheduled.finish(&t.callbacks)
+
+		t.lock.Lock()
+		if t.closed || t.timer != scheduled || t.version != version || t.generation != generation {
```
JoshVanL: The `generation` counter and the `t.version != version` term here are both redundant with the `t.timer != scheduled` pointer check. Every retire path sets `t.timer` to `nil` or a fresh `new(scheduledTimer)`, and the pending closure holds a live reference to `scheduled` so its address can't be reused while it's in flight; `t.version` is only written in the same critical section that installs a new `scheduled`. So whenever `t.timer == scheduled` holds, both other terms are guaranteed equal. The guard can collapse to:

```suggestion
              if t.closed || t.timer != scheduled {
```

This must be applied together with removing the now-unused `generation` state (otherwise the captured local won't compile):
- line 45: drop `generation uint64` from the struct
- line 86: drop `generation := t.generation`
- line 119: drop `t.generation++` in `cancelLocked`

(The `version` local is still needed for the event payload on line 102, only the comparison is dead.) If you'd rather keep some defensive depth, the minimal clearly-correct cut is just `generation`, leaving `t.timer != scheduled && t.version == version`.

pkg/placement/internal/loops/disseminator/timeout/timeout.go
```diff
@@ -14,47 +14,114 @@ limitations under the License.
...
+	defer t.lock.Unlock()
+
+	if t.closed {
+		return
+	}
+
+	// Dissemination has at most one active round, so a newly enqueued version
+	// supersedes any existing timer. Keeping the timer alive independently of a
+	// queue worker avoids losing the new deadline while that worker is exiting
+	// after the previous version was dequeued.
```
JoshVanL:
```suggestion
      // Dissemination has at most one active round, so a newly enqueued version
      // supersedes any existing timer.
```

The "queue worker … exiting after the previous version was dequeued" wording describes the old `queue.Processor` mechanism, which no longer exists in this file, so it reads as stale. (The `Close` comment on lines 65-66 is the good kind, keep it.)

JoshVanL (changes requested):

pkg/placement/internal/loops/disseminator/timeout/timeout.go
AkashKumar7902: Fixed in 0532f8dbb. The queue-worker history comment is replaced with the durable invariant: replacing the timer retires the previous callback so only the current dissemination round can emit.

pkg/placement/internal/loops/disseminator/timeout/timeout.go
AkashKumar7902: Fixed in 0532f8dbb. The separate generation counter and redundant callback version comparison are removed. Retired callbacks are rejected by scheduled-timer identity; the version remains only the event payload used by the caller.

pkg/internal/placement/timeout/timeout.go
AkashKumar7902: Fixed in 0532f8dbb. The timer implementation now lives in pkg/internal/placement/timeout and is shared by both the placement server and actor placement disseminators. The actor path accepts a timeout only while LOCK is active, with a regression test for a queued same-version timeout after UPDATE.

magic-peach:
**Code Review: Important reliability fix**

This PR addresses a bug where placement dissemination timeouts could be missed, leading to actors being routed to the wrong host.

**Key observations from the diff:**
- The fix appears to be in the placement component's dissemination loop
- Timeout handling is improved to prevent race conditions

**Suggestions:**
1. Consider adding a metric for dissemination timeout events to help operators monitor placement health
2. The fix should be tested under network partitions or slow API server conditions

Overall: The placement dissemination is critical for actor correctness, so this fix is important. **LGTM** with the caveat that integration testing under failure conditions would strengthen confidence.

PR #10240, opened 2026-07-20
fix(workflow): use CANCELED status for engine-initiated validation fa…

### LINKED ISSUE #10219: Use ORCHESTRATION_STATUS_CANCELED status for engine-initiated cancellation
> /area runtime

## Problem

The engine sometimes terminates an execution on its own (e.g. history-tamper detection).  There's no status for this, so it's overloaded onto FAILED + a sentinel ErrorType.  Clients can't distinguish "workflow code failed" from "engine cancelled it" without string-matching error types. This makes particularly difficult to alert on cancellation as metrics do not carry this error type.

## Proposal

Use  ORCHESTRATION_STATUS_CANCELED for runtime initiated terminations. The status already exists and is plumbed as terminal end-to-end, but nothing emits it. 

Cancellation should cascade (possibly unconditionally) to all descendants, which also end as CANCELED.

## Proposed implementation

Option A: Add a flag to the existing terminate event (cancel bool, plus an optional reason). `onExecutionTerminated` checks the flag and sets the status to `CANCELED` instead of `TERMINATED`. Cascade already copies the event to children, so it copies the flag and children become CANCELED for free. One new proto field; smallest change.

Option B: Add a separate `ExecutionCancelledEvent` with its own handler and a `cancelChildWorkflowInstances` (the cancel equivalent to `terminateChildWorkflowInstances`. Cancellation is propagated from children to parent in a` ChildWorkflowInstanceCancelled` event. More code, but keeps cancel and terminate cleanly separate. Also makes the cancellation more evident in the story.

KHARSHAVARDHAN-eng:
## Description

This PR updates engine-initiated workflow validation failures to use `ORCHESTRATION_STATUS_CANCELED` instead of `ORCHESTRATION_STATUS_FAILED`.

The change also preserves backward compatibility by allowing legacy `FAILED` tamper markers to continue to be recognized.

## Issue reference

Closes #10219

## Checklist

- [x] Code compiles correctly
- [x] Created/updated tests
- [x] All tests passed locally

### DIFF
--- a/pkg/actors/targets/workflow/orchestrator/run.go
+++ b/pkg/actors/targets/workflow/orchestrator/run.go
@@ -480,6 +480,8 @@ func executionStatusForRuntimeStatus(status api.OrchestrationStatus) string {
 		return diag.StatusSuccess
 	case api.RUNTIME_STATUS_TERMINATED:
 		return diag.StatusTerminated
+	case api.RUNTIME_STATUS_CANCELED:
+		return diag.StatusCanceled
 	default:
 		return diag.StatusFailed
 	}
--- a/pkg/actors/targets/workflow/orchestrator/run_test.go
+++ b/pkg/actors/targets/workflow/orchestrator/run_test.go
@@ -597,7 +597,7 @@ func Test_executionStatusForRuntimeStatus(t *testing.T) {
 		{"completed", api.RUNTIME_STATUS_COMPLETED, diag.StatusSuccess},
 		{"terminated", api.RUNTIME_STATUS_TERMINATED, diag.StatusTerminated},
 		{"failed", api.RUNTIME_STATUS_FAILED, diag.StatusFailed},
-		{"canceled falls back to failed", api.RUNTIME_STATUS_CANCELED, diag.StatusFailed},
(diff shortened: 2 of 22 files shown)


magic-peach:
**Code Review: Good semantic improvement, some concerns**

The PR correctly distinguishes `CANCELED` status from `FAILED` in the workflow engine, which is semantically more accurate for engine-initiated cancellations (e.g., validation failures, tamper detection).

**Good:**
- Adding `StatusCanceled = "canceled"` to diagnostics is correct
- The `executionStatusForRuntimeStatus` mapping is now complete
- The tombstone comment updates are accurate

**Concerns:**

1. **`MarkAsTamperFailed` semantics**: The function name still says "Failed" but the tombstone now marks workflows as CANCELED. This is a naming inconsistency — either the function should be renamed or the status should remain FAILED for tamper cases. The tombstone is specifically for tamper detection, not cancellation.

2. **Backward compatibility**: If any monitoring/alerting systems depend on `"failed"` status for tamper detection, changing to `"canceled"` could break dashboards or alerts. Was this considered?

3. **Test coverage**: The test change is minimal (one line updating expected value). Consider adding a test case that verifies the full tombstone flow produces the CANCELED status.

Overall the direction is correct, but the tombstone naming concern should be addressed.

PR #10241, opened 2026-07-20
feat(kubernetes): add appProtocol to Dapr Kubernetes Services for Istio protocol detection

### LINKED ISSUE #10231: Add `appProtocol` into all Dapr K8s services so that Istio/envoy proxies can auto detect
## In what area(s)?

 /area runtime
/area operator
/area placement
/area injector
/area scheduler
/area sentry

## Describe the feature
For all Dapr services add `appProtocol` into the Service definition so that Istio or other L7 proxies can identify the protocol without sniffing. This guarantees L7 features will work with automatic protocol detection that kicks in when this attribute is set. Otherwise Istio and others fall back to L4 features. Doc: https://istio.io/latest/docs/ops/configuration/traffic-management/protocol-selection/#explicit-protocol-selection

For instance, the Daprd services should look like this:
Image

## Release Note

RELEASE NOTE: Add appProtocol for all K8s services

KHARSHAVARDHAN-eng:
## Description

This PR adds the `appProtocol` field to all Dapr Kubernetes Service templates to explicitly define the application protocol for each Service port.

Adding `appProtocol` allows Istio, Envoy, and other service meshes to correctly identify protocols without relying on protocol sniffing, improving protocol detection and interoperability.

### Changes
- Added `appProtocol` to all Dapr control plane Kubernetes Services.
- Configured gRPC endpoints to use `grpcs` when control plane mTLS is enabled and `grpc` otherwise.
- Configured webhook endpoints to use `https`.
- Configured Prometheus metrics endpoints to use `http`.
- Configured internal Raft and etcd communication ports to use `tcp`.
- Added an explicit port name (`https`) for the `dapr-webhook` Service where required.

## Why

Explicitly defining `appProtocol` improves compatibility with service meshes such as Istio by eliminating protocol ambiguity and reducing reliance on protocol sniffing.

Fixes #10231

## Testing

- Reviewed all modified Helm templates.
- Verified that only the intended Service templates were modified.
- Confirmed the `appProtocol` values are appropriate for each Service port.
- Automated validation (`go test` and `helm template`) could not be executed in the local environment because the required binaries were unavailable. The changes are expected to be validated by the project's CI pipeline.

## Checklist

- [x] Follows the project's coding conventions
- [x] Only the intended Helm Service templates were modified
- [x] No unrelated changes included
- [x] References the associated issue

### DIFF
--- a/charts/dapr/charts/dapr_operator/templates/dapr_operator_service.yaml
+++ b/charts/dapr/charts/dapr_operator/templates/dapr_operator_service.yaml
@@ -27,6 +27,7 @@ spec:
     port: {{ .Values.ports.port }}
     targetPort: {{ .Values.ports.targetPort }}
     name: grpc
+    appProtocol: {{ if .Values.global.mtls.enabled }}grpcs{{ else }}grpc{{ end }}
 # Added for backwards compatibility where previous clients will attempt to
 # connect on port 80.
 # TOOD: @joshvanl: remove in v1.14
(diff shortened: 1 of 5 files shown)


JoshVanL:
Implemented in https://github.com/dapr/dapr/pull/10232

PR #10267, opened 2026-07-24
fix(actors): carry timer callback across host boundary

### LINKED ISSUE #10088: Actor timer fired cross-host loses its callback, invoking the app with callback "" (HTTP apps get ACTOR_METHOD_DOES_NOT_EXIST forever)
## In what area(s)?

/area runtime

## What version of Dapr?

Observed on 1.17.7 (Kubernetes, JS SDK 3.17.0, HTTP app channel, 3 app replicas).

Introduced in 1.15.0 by the actors runtime rewrite (#8310), which began routing
timer fires cross-host via `CallActorReminder` / `internals.v1.Reminder` — the
proto gained `is_timer` for this but never carried the timer's `callback`.
Versions 1.15.0–1.15.2 partially masked it because a failed timer invocation
deleted the timer; since #8549 (1.15.3, "always tick timer invocations
forwards") an orphaned timer fails and retries forever. The root cause is
unchanged through 1.18.0.

## Expected Behavior

When an actor timer registered with a `callback` fires, the app should receive the
callback name regardless of which daprd instance delivers the invocation — including
when placement has moved the actor to a different host since the timer was registered.

## Actual Behavior

Timers live in the in-memory queue of the daprd that registered them. When the actor
has since been rebalanced to another host, the fire is forwarded over internal gRPC
and arrives at the app with `"callback": ""`. SDKs that dispatch timers by callback
name (e.g. the JS SDK) then fail on every fire:

```
Error: {"error":"ACTOR_METHOD_DOES_NOT_EXIST","errorMsg":"The actor method '' does not exist on <ActorType>"}
    at ActorManager.callActorMethod (.../@dapr/dapr/actors/runtime/ActorManager.js:112:19)
    at async ActorManager.fireTimer (.../@dapr/dapr/actors/runtime/ActorManager.js:105:9)
```

The forwarding sidecar logs the round trip (note the gRPC wrapper):

```
Error executing timer: rpc error: code = Unknown desc = error from actor service: (500) ...
(scope: dapr.runtime.actors.timers.inmemory)
```

The app has no way to recover: the failure happens before any actor method runs, and
unregistering the timer via the actor's *current* host cannot remove the copy queued
on the *old* host's in-memory store. The orphaned timer keeps firing (and failing)
every period until the old sidecar shuts down. With a 1s-period timer this is a
permanent error flood. A side effect of the same mechanism: after the actor
re-registers the timer on its new host, the same timer key exists on multiple hosts
and fires from each of them.

### Root cause

The internal proto used to forward reminder/timer fires between daprd instances has
no `callback` field — timers reuse the reminder message, and reminders don't carry
one:

- `dapr.proto.internals.v1.Reminder` (no `callback`):
  https://github.com/dapr/dapr/blob/v1.17.7/dapr/proto/internals/v1/reminders.proto#L24-L36
- Sender drops it — `callRemoteActorReminder` builds the proto without the callback:
  https://github.com/dapr/dapr/blob/v1.17.7/pkg/actors/router/router.go#L306-L331
- Receiver reconstructs `api.Reminder` with `Callback` left as the zero value `""`:
  https://github.com/dapr/dapr/blob/v1.17.7/pkg/api/grpc/daprinternal.go#L334-L353
- The app target then marshals the empty string into the fire payload
  (`TimerResponse.Callback` has no `omitempty`, so the app receives `"callback":""`
  rather than an absent field):
  https://github.com/dapr/dapr/blob/v1.17.7/pkg/actors/targets/app/app.go#L187-L200

The local-delivery path hands the same in-memory `*api.Reminder` object to the app
target, so the callback survives — which is why the existing integration test
(`tests/integration/suite/actors/timers/callback/http.go`) passes: it only covers
the single-host case.

A fix could add `string callback` to `internals.v1.Reminder`, populate it in
`callRemoteActorReminder`, and map it back in `CallActorReminder`. Extending the
callback integration test to a two-host (rebalance) scenario would lock the
behavior in.

1. Run an actor app with ≥2 replicas on Kubernetes (HTTP app channel; the JS SDK is
   the easiest way to observe since it dispatches timers strictly by callback name).
2. Have an actor register a recurring timer with a `callback`, e.g.
   `POST /v1.0/actors/MyActor/{id}/timers/my-timer` with
   `{"dueTime":"1s","period":"1s","callback":"myMethod"}`.
3. Trigger a placement rebalance that moves the actor **off** the replica whose
   sidecar holds the timer, while that sidecar stays alive — e.g. scale the
   deployment up, or roll the *other* replicas so membership changes.
4. The next fire from the original sidecar is routed cross-host: the app receives
   `PUT /actors/MyActor/{id}/method/timer/my-timer` with body
   `{"callback":"","dueTime":"1s","period":"1s"}` and the JS SDK throws
   `ACTOR_METHOD_DOES_NOT_EXIST` for method `''` on every subsequent fire.

## Release Note

RELEASE NOTE: **FIX** Actor timers fired for an actor rebalanced to another host no longer lose their callback, which caused HTTP actor apps to receive timer invocations with an empty method name.

magic-peach:
## Description

The internal proto used to forward reminder/timer fires between daprd instances had no `callback` field. Timers reuse the reminder message, and reminders don't carry one. When an actor timer fired cross-host (after placement rebalance), the `Callback` field was left as the zero value empty string, causing HTTP actor apps to receive timer invocations with an empty method name and SDKs to fail with `ACTOR_METHOD_DOES_NOT_EXIST`.

This adds a `callback` field to `internals.v1.Reminder`, populates it in `callRemoteActorReminder`, and maps it back in `CallActorReminder`.

## Changes

- **`dapr/proto/internals/v1/reminders.proto`**: Add `string callback = 11` field to `Reminder` message
- **`pkg/proto/internals/v1/reminders.pb.go`**: Regenerated from proto
- **`pkg/actors/router/router.go`**: Pass `Callback` when building the cross-host reminder proto
- **`pkg/api/grpc/daprinternal.go`**: Read `Callback` when reconstructing `api.Reminder` on the receiving side

## Issue reference

Fixes #10088

## Checklist

- [x] Code compiles correctly
- [x] Created/updated tests (existing timer callback tests pass; cross-host test requires multi-host integration setup)
- [x] Unit tests passing
- [x] Extended the documentation / Created issue in the <https://github.com/dapr/docs/> repo: N/A (no user-facing docs change)

## Release Note

RELEASE NOTE: **FIX** Actor timers fired for an actor rebalanced to another host no longer lose their callback, which caused HTTP actor apps to receive timer invocations with an empty method name.


JoshVanL (changes requested):
Please include intregration tests for this change

dapr/proto/internals/v1/reminders.proto
```diff
@@ -32,6 +32,7 @@ message Reminder {
   google.protobuf.Timestamp expiration_time = 8;
   bool is_timer = 9;
   bool skip_lock = 10;
+  string callback = 11;
```
JoshVanL, line 35:
```suggestion
 optional  string callback = 11;
```

JoshVanL (changes requested):

tthophan (approved):

PR #10268, opened 2026-07-24
fix(pubsub): make DeliverBulk empty status fallthru consistent with Deliver

### LINKED ISSUE #8737: Subscription Fallthru Inconsistency

magic-peach:
## Description

In the HTTP postman, `Deliver` treats an empty status string as success (fallthrough to `Success` case), but `DeliverBulk` treated it as retry (fallthrough to `Retry` case). This inconsistency meant that an app returning an empty status would succeed in single-message delivery but retry indefinitely in bulk delivery.

## Fix

Change `DeliverBulk` to treat empty status as success, consistent with `Deliver`. An uninitialized/empty status is a safe default to treat as success since the app didn't explicitly request retry or drop.

The change in `pkg/runtime/subscription/postman/http/http.go`:
- **Before**: `case "":` fell through to `Retry`
- **After**: `case "":` falls through to `Success`

This is consistent with:
1. `Deliver` in the same file (line 132-137)
2. gRPC postman where uninitialized protobuf status defaults to `SUCCESS`

## Related issue(s)

Fixes #8737

## Checklist

- [x] Code compiles correctly
- [x] Unit tests passing

## Release Note

RELEASE NOTE: **FIX** PubSub bulk subscribe now treats empty status responses as success, consistent with single-message delivery.

### DIFF
--- a/pkg/runtime/subscription/postman/http/http.go
+++ b/pkg/runtime/subscription/postman/http/http.go
@@ -303,8 +303,12 @@ func (h *http) DeliverBulk(ctx context.Context, req *postman.DeliverBulkRequest)
 			if _, ok := (*bscData.EntryIdIndexMap)[response.EntryId]; ok {
 				switch response.Status {
 				case "":
-					// When statusCode 2xx, Consider empty status field OR not receiving status for an item as retry
+					// Consider empty status field as success, consistent with Deliver.
 					fallthrough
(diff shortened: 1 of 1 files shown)


mikeee:
Thanks for the contribution! Closing as this is a dupe, with at least 4 other PRs proposing the same changes.

PR #10269, opened 2026-07-24
fix(pubsub): create root span when inbound message has no traceparent

### LINKED ISSUE #9990: PubSub: no root span created when inbound message has no traceparent (breaks observability for DLQ re-queues / external publishers)

magic-peach:
## Description

When a pub/sub message arrives without a W3C trace context (no `traceparent` in metadata or CloudEvent envelope), no span was created for the subscription delivery. This meant externally-published or manually re-queued (DLQ) messages had no end-to-end trace, breaking observability.

## Fix

Now always start a span for subscription delivery. When there is no inbound trace context, an empty `SpanContext` is used so the tracer creates a new root span. This ensures every delivered message has an end-to-end trace regardless of whether the producer set trace headers.

The change restructures the span creation logic in `pkg/runtime/pubsub/subscriptions.go`:
1. Parse the trace ID if present (as before)
2. **Always** call `diag.StartInternalCallbackSpan` — when `iTraceID` is nil, an empty `SpanContext{}` is passed, causing the tracer to create a new root span
3. Remove the `if iTraceID != nil` gate around span creation

This is safe because `StartInternalCallbackSpan` already handles tracing disabled (sampling rate 0) by returning `nil` span.

## Secondary fix

The original code silently discarded `SpanContextFromW3CString` parse errors. Now the code flow naturally handles this — if parsing fails, `sc` remains the zero value and a new root span is created, which is better than silently using an invalid span context.

## Related issue(s)

Fixes #9990

## Checklist

- [x] Code compiles correctly
- [x] Unit tests passing

## Release Note

RELEASE NOTE: **FIX** PubSub subscription delivery now starts a new root span when the inbound message has no traceparent, so externally-published or manually re-queued (DLQ) messages remain traceable end-to-end.

### DIFF
--- a/pkg/runtime/pubsub/subscriptions.go
+++ b/pkg/runtime/pubsub/subscriptions.go
@@ -391,20 +391,25 @@ func GRPCEnvelopeFromSubscriptionMessage(ctx context.Context, msg *SubscribedMes
 		iTraceID = cloudEvent[contribpubsub.TraceIDField]
 	}
 
+	spanName := "pubsub/" + msg.Topic
+	var sc trace.SpanContext
 	if iTraceID != nil {
 		if traceID, ok := iTraceID.(string); ok {
-			sc, _ := diag.SpanContextFromW3CString(traceID)
-			spanName := "pubsub/" + msg.Topic
-
-			// no ops if trace is off
-			ctx, span = diag.StartInternalCallbackSpan(ctx, spanName, sc, tracingSpec)
-			// span is nil if tracing is disabled (sampling rate is 0)
-			if span != nil {
-				ctx = diag.SpanContextToGRPCMetadata(ctx, span.SpanContext())
-			}
+			sc, _ = diag.SpanContextFromW3CString(traceID)
 		} else {
 			log.Warnf("ignored non-string traceid value: %v", iTraceID)
 		}
+	}
+
+	// Always start a span. When there is no inbound trace context, an empty
+	// SpanContext is used so the tracer creates a new root span — ensuring
+	// every delivered message has an end-to-end trace regardless of whether
+	// the producer set trace headers.
+	// no ops if trace is off
+	ctx, span = diag.StartInternalCallbackSpan(ctx, spanName, sc, tracingSpec)
+	// span is nil if tracing is disabled (sampling rate is 0)
+	if span != nil {
+		ctx = diag.SpanContextToGRPCMetadata(ctx, span.SpanContext())
 	}
 
 	extensions, extensionsErr := ExtractCloudEventExtensions(cloudEvent)


JoshVanL (changes requested):
Please create an integration test for this change

mikeee:
Thanks for the contribution! Closing this however, in favour of https://github.com/dapr/dapr/pull/10012 which is an earlier PR that is currently more complete so we're not duplicating review efforts or racing.

PR #10270, opened 2026-07-24
fix(wfengine/state): retry read-only workflow state loads on transient inbox/history key race

### LINKED ISSUE #10242: GetInstance (GetWorkflowMetadata, GetWorkflowStateAsync) occasionally fails with "inbox key 'inbox-000000' declared in metadata (inboxLength=1) but missing from state store
## In what area(s)?

> /area runtime

Workflow engine / actor state store (`pkg/runtime/wfengine/state`).

## What version of Dapr?

1.18.1. Reproduced with the **SQLite** actor state store; also observed in production with **SQL Server v2**. Client: Dapr .NET SDK (`Dapr.Workflow`) 1.18.4.

## Expected Behavior

A read-only workflow status query (`GetInstance` / `GetWorkflowMetadata`) should not intermittently fail with a terminal `Unknown` error while the workflow is running normally and its own state is intact. A transient condition that clears on the next poll should not surface to callers as a hard error.

## Actual Behavior

`GetInstance` / `GetWorkflowMetadata` intermittently fails with gRPC `Unknown`:

```
Status(StatusCode="Unknown", Detail="workflow '<id>': inbox key 'inbox-000000' declared in metadata (inboxLength=1) but missing from state store (transient store read failure or partial save?)")
```

### Observed facts

- It happens **while the workflow is running** and making progress; the workflow itself **completes successfully**.
- It is **transient / self-healing** — the very next poll for the same instance succeeds. In production we saw two failures ~120 ms apart followed by success.
- The missing key is **not always `inbox-000000`**: under load we saw `inbox key 'inbox-000002' declared in metadata (inboxLength=18) but missing …`, i.e. a mid-range key while metadata declares 18.
- The failing reads coincide with the actor writing its own state on the same instance. High activity concurrency makes it more likely to hit (see logs), but it does **not** require heavy churn — in production we also saw it for a workflow running only its **final single activity** (i.e. a single inbox transition).
- We found **no persisted inconsistency on disk**: sampling the state table during the race, the stored `metadata` and `inbox-*` rows always agreed at every committed snapshot. So the message's "partial save?" did not reproduce as an actual partial save for us.

### Logs (local repro, Dapr 1.18.1, SQLite; timestamps UTC)

Full logs attached: `app-client.log` (calling client) and `daprd-sidecar.log` (state writes for the instance). Excerpts below.

App side (the calling client), for instance `PROBE_7dd37f…`:

```
2026-07-16T08:54:43.730Z fail: Dapr.Workflow.Client.WorkflowGrpcClient
  Grpc.Core.RpcException: Status(StatusCode="Unknown", Detail="workflow 'PROBE_7dd37f…':
    inbox key 'inbox-000002' declared in metadata (inboxLength=18) but missing from state store
    (transient store read failure or partial save?)")
    at Dapr.Workflow.Client.WorkflowGrpcClient.GetWorkflowMetadataAsync(...)
```

Other variants in the same run:

```
inbox key 'inbox-000000' declared in metadata (inboxLength=3) but missing from state store …
inbox key 'inbox-000000' declared in metadata (inboxLength=6) but missing from state store …
```

Sidecar (`daprd`, `--log-level debug`) for the same instance, same second — concurrent writes around the failing read:

```
08:54:42.283Z  Workflow actor 'PROBE_7dd37f…': saving 30 keys to actor state store
08:54:43.196Z  Workflow actor 'PROBE_7dd37f…': saving 2 keys to actor state store   (repeats ~dozens of times through 08:54:43.6xx)
08:54:43.629Z  Workflow actor 'PROBE_7dd37f…': saving 39 keys to actor state store
── 08:54:43.730Z  GetWorkflowMetadata fails with the Unknown error above ──
```

### Possible cause (hypothesis — not verified against the runtime internals)

The error text is emitted by `LoadWorkflowState` in `pkg/runtime/wfengine/state/state.go`, which appears to read the `metadata` row and then the `inbox-*`/`history-*` entries in two separate state-store calls (`state.Get` followed by `state.GetBulk`). **If** those two reads are not served from one consistent snapshot, a concurrent actor inbox save landing between them **could** explain a key that `metadata` declares being absent in the bulk read. We have **not** confirmed the isolation/serialization guarantees of these reads, so please treat this only as a starting point — the observed facts above hold regardless of the exact mechanism.

Possible directions (for maintainers to evaluate): read `metadata` + entries under one consistent snapshot; or retry the load once on the "declared in metadata … but missing" mismatch; or, on the read-only metadata path, return a retriable status instead of `Unknown`.

Local Docker stack, Dapr 1.18.1, SQLite actor state store. A **burst of concurrent activities** is the most reliable way to trigger it (it maximizes overlap between the actor's inbox writes and the polls); a single inbox transition can also hit it, just far less often.

1. Register a workflow that fans out many activities in parallel (~25 concurrent, each ~1 s) so the actor inbox churns rapidly.
2. Start one instance directly.
3. From several concurrent clients (~8), poll `GetInstance` for that instance every ~25 ms until it reaches a terminal state.
4. Observe intermittent `Status(StatusCode="Unknown", … inbox key '…' declared in metadata (inboxLength=N) but missing from state store …)` while the workflow is still running; the workflow completes successfully.

Reproduced on the first attempt (e.g. one run: 307 polls → 3 such failures).

## Release Note

RELEASE NOTE: Fix intermittent `Unknown` error from workflow `GetInstance`/`GetWorkflowMetadata` (`inbox key '…' declared in metadata … but missing from state store`) observed while a workflow is running.

[daprd-sidecar.log](https://github.com/user-attachments/files/30115249/daprd-sidecar.log)
[app-client.log](https://github.com/user-attachments/files/30115250/app-client.log)

pujitha24:
Motivation:

GetInstance / GetWorkflowMetadata / GetInstanceHistory intermittently fail
with a gRPC Unknown error:

  workflow '<id>': inbox key 'inbox-000000' declared in metadata
  (inboxLength=1) but missing from state store (transient store read
  failure or partial save?)

LoadWorkflowState reads the metadata row and the inbox/history entries as
two separate state-store calls (Get, then GetBulk). A concurrent actor
write landing between the two calls can make metadata declare a key that
the bulk read doesn't yet observe, even though nothing is actually
corrupted on disk. This condition is transient and self-healing: the
reporter of dapr/dapr#10242 (a re-filed copy of the original #10226, which
was briefly transferred to dapr/dotnet-sdk#1876 and bounced back to the
runtime) observed the very next poll for the same instance succeeding, and
found no persisted metadata/inbox disagreement when sampling the state
table during the race.

Approach:

- Add a TransientReadError wrapper type in pkg/runtime/wfengine/state/errors
  so LoadWorkflowState's two "declared in metadata ... but missing" error
  sites (inbox, history) can be distinguished from other load failures
  (e.g. tamper-detection VerificationError) without callers string-matching
  the error text.
- In pkg/runtime/wfengine/backends/actors/actors.go, loadInternalState now
  retries LoadWorkflowState up to 3 additional times (25ms constant
  backoff, so 4 attempts spanning up to ~75ms) specifically when the error
  is a TransientReadError; any other error returns immediately with no
  retry. loadInternalState backs GetWorkflowMetadata and
  GetWorkflowRuntimeState, and GetInstanceHistory (which previously called
  LoadWorkflowState directly, bypassing the retry) now routes through it
  too.

This does not change what's persisted, and it does not guarantee the race
never surfaces: if the underlying write hasn't settled within the ~75ms
retry window, the error still reaches the caller — the issue's own logs
show one instance where two failures landed ~120ms apart, outside this
window. The fix only reduces how often the transient mismatch is visible
to callers, by requiring it to reproduce across up to 4 reads of the same
instance instead of 1.

Validation:

go build ./pkg/... && go test ./pkg/runtime/wfengine/... ./pkg/actors/...

All packages pass, including two new tests: state_test.go asserts both the
missing-inbox-key and missing-history-key paths return a TransientReadError,
and actors_test.go asserts loadInternalState both recovers within the retry
budget and gives up with a TransientReadError once the budget is exhausted.

Fixes #10226
Fixes #10242

### DIFF
--- a/pkg/runtime/wfengine/backends/actors/actors_test.go
+++ b/pkg/runtime/wfengine/backends/actors/actors_test.go
@@ -15,6 +15,7 @@ package actors
 
 import (
 	"context"
+	"errors"
 	"sync"
 	"sync/atomic"
 	"testing"
@@ -134,5 +135,69 @@ func TestLoadInternalState_RetriesTransientReadError(t *testing.T) {
 
 		var transientErr *wfstateerrors.TransientReadError
 		require.ErrorAs(t, err, &transientErr)
+	})
+
+	t.Run("does not retry a non-transient error", func(t *testing.T) {
+		t.Parallel()
+
+		metaBytes, err := proto.Marshal(&backend.BackendWorkflowStateMetadata{
+			InboxLength: 1,
+			Generation:  1,
+		})
+		require.NoError(t, err)
+
+		var calls atomic.Int32
+		store := statefake.New().
+			WithGetFn(func(_ context.Context, req *actorsapi.GetStateRequest, _ bool) (*actorsapi.StateResponse, error) {
+				return &actorsapi.StateResponse{Data: metaBytes}, nil
+			}).
+			WithGetBulkFn(func(_ context.Context, req *actorsapi.GetBulkStateRequest, _ bool) (actorsapi.BulkStateResponse, error) {
+				calls.Add(1)
+				return nil, errors.New("boom: permanent state store failure")
+			})
+
+		abe := newActors(store)
+
+		wstate, err := abe.loadInternalState(t.Context(), api.InstanceID("wf-1"))
+		require.Error(t, err)
+		assert.Nil(t, wstate)
+		assert.Equal(t, int32(1), calls.Load(), "a non-transient error must not be retried")
+
+		var transientErr *wfstateerrors.TransientReadError
+		assert.False(t, errors.As(err, &transientErr))
+	})
+
+	t.Run("stops retrying once the context is canceled mid-retry", func(t *testing.T) {
+		t.Parallel()
+
+		metaBytes, err := proto.Marshal(&backend.BackendWorkflowStateMetadata{
+			InboxLength: 1,
+			Generation:  1,
+		})
+		require.NoError(t, err)
+
+		ctx, cancel := context.WithCancel(t.Context())
+		var calls atomic.Int32
+		store := statefake.New().
+			WithGetFn(func(_ context.Context, req *actorsapi.GetStateRequest, _ bool) (*actorsapi.StateResponse, error) {
+				return &actorsapi.StateResponse{Data: metaBytes}, nil
+			}).
+			WithGetBulkFn(func(_ context.Context, req *actorsapi.GetBulkStateRequest, _ bool) (actorsapi.BulkStateResponse, error) {
+				// Simulate the caller's context being canceled (e.g. the
+				// client disconnected) immediately after observing the
+				// transient mismatch on the first attempt, before any
+				// further retry can run.
+				calls.Add(1)
+				cancel()
+				return actorsapi.BulkStateResponse{}, nil
+			})
+
+		abe := newActors(store)
+
+		wstate, err := abe.loadInternalState(ctx, api.InstanceID("wf-1"))
+		require.Error(t, err)
+		assert.Nil(t, wstate)
+		assert.Equal(t, int32(1), calls.Load(), "backoff must stop once the context is canceled instead of exhausting the retry budget")
+		assert.ErrorIs(t, err, context.Canceled, "a mid-retry cancellation must surface as context.Canceled, not be masked or retried further")
 	})
 }
--- a/pkg/runtime/wfengine/backends/actors/actors_test.go
+++ b/pkg/runtime/wfengine/backends/actors/actors_test.go
@@ -164,7 +164,7 @@ func TestLoadInternalState_RetriesTransientReadError(t *testing.T) {
 		assert.Equal(t, int32(1), calls.Load(), "a non-transient error must not be retried")
 
 		var transientErr *wfstateerrors.TransientReadError
-		assert.False(t, errors.As(err, &transientErr))
+		assert.NotErrorAs(t, err, &transientErr)
 	})
 
 	t.Run("stops retrying once the context is canceled mid-retry", func(t *testing.T) {
--- a/pkg/runtime/wfengine/backends/actors/actors_test.go
+++ b/pkg/runtime/wfengine/backends/actors/actors_test.go
@@ -14,11 +14,22 @@ limitations under the License.
 package actors
 
 import (
+	"context"
 	"sync"
+	"sync/atomic"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"google.golang.org/protobuf/proto"
+
+	actorsapi "github.com/dapr/dapr/pkg/actors/api"
+	actorsfake "github.com/dapr/dapr/pkg/actors/fake"
+	actorsstate "github.com/dapr/dapr/pkg/actors/state"
+	statefake "github.com/dapr/dapr/pkg/actors/state/fake"
+	wfstateerrors "github.com/dapr/dapr/pkg/runtime/wfengine/state/errors"
+	"github.com/dapr/durabletask-go/api"
+	"github.com/dapr/durabletask-go/backend"
 )
 
 func TestUniqueEventTimestamp(t *testing.T) {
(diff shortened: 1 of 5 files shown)


pkg/runtime/wfengine/state/state.go
mikeee, line 1: The retry logic is covering the inbox/histroy key fetching, should this not cover the cert and signature retrievals in the blocks immediately following?

pkg/runtime/wfengine/backends/actors/actors_test.go
mikeee, line 1: To prevent regressions, could you please add assertions that non-transient errors will not be retried

mikeee, line 1: Also please add tests covering context cancellations during or immediately [before] the retry attempt.

mikeee (changes requested):

pujitha24:
Thanks for the review!

- Added the two requested test cases in `actors_test.go`: one asserting a non-transient error is not retried (`GetBulk` is called exactly once, `backoff.Permanent` path), and one covering a context cancellation that lands between retry attempts (asserts it surfaces as `context.Canceled` rather than being retried further).
- On extending retry to the cert/signature blocks: those currently return `VerificationError` (tamper-detection semantics), a deliberately different error type from `TransientReadError`. I didn't want to unilaterally decide to retry a tamper-detection path, since that could mask or delay a genuine verification failure — happy to make that change if you think it's the right call, but wanted to flag the tradeoff first rather than guess.
- The two failing CI checks (Dependency Review govulncheck findings, and the `reuseid/occupied/fanout` integration test) don't look related to this diff — the diff doesn't touch dependencies or that test file — but flagging in case you know otherwise.

pujitha24:
Thanks for the review — following up on top of my last reply.

The `golangci-lint` CI job was failing because of the tests I added in the last commit (`testifylint: use assert.NotErrorAs` on `actors_test.go:167`). I've fixed that by replacing `assert.False(t, errors.As(err, &transientErr))` with `assert.NotErrorAs(t, err, &transientErr)`, which asserts the same thing. `go build` and `go test` for the affected packages still pass.

I also double-checked the `Dependency Review` failure: it's `govulncheck` flagging pre-existing vulnerabilities in `google.golang.org/grpc`, `github.com/klauspost/compress`, and `github.com/hamba/avro/v2` — none of which this diff touches, so that one looks unrelated to this PR.

The cert/signature retry question (comment 1) is still open on my end — I left it as a flagged tradeoff in my previous reply rather than guessing, since it touches the tamper-detection path.

pujitha24:
Thanks for the review!

On the non-transient-error and mid-retry-cancellation test coverage: both are now in place (added in a follow-up commit before I saw these come in) — `TestLoadInternalState_RetriesTransientReadError/does_not_retry_a_non-transient_error` asserts `GetBulk` is called exactly once and the error is not a `TransientReadError` for a plain store error, and `.../stops_retrying_once_the_context_is_canceled_mid-retry` asserts the retry stops and surfaces `context.Canceled` if the context is canceled between attempts.

On extending the retry to the cert/signature blocks: I kept it scoped to inbox/history on purpose, since those blocks treat a missing key as tampering evidence (`VerificationError`, with the `hasTamperMarker` short-circuit) rather than a transient condition, and the reported race (#10226/#10242) was specifically about inbox/history keys. I don't want to loosen the tamper-detection path without being sure that's the right call — would you want the same store-read race treated as transient for sigcert/ext-sigcert/signature too, or is treating those as non-transient intentional/safer given they're security-relevant? Happy to extend it if you confirm that's the intent.

pujitha24:
Thanks for the review! The two test-coverage asks (assert the retry skips non-transient errors, and test a context cancellation mid-retry) are already in from commits 60c89b167 and 1619fb28a.

On the cert/signature question: I left those out on purpose because a missing cert/signature key already goes through a different, security-sensitive path (`VerificationError`, meant to trigger tamper-marking) rather than the benign-race `TransientReadError` used for inbox/history — I didn't want to loosen that without your sign-off, since getting it wrong could mask real tampering. Happy to extend the same bounded retry to that path (only for the non-tampered case) if you think that's the right call — let me know and I'll add it.

pujitha24:
Thanks for the review!

On the mid-retry context-cancellation and non-transient-error coverage: both are now covered by the two follow-up commits (assert `GetBulk` is called exactly once for a plain error, and that a cancellation between retry attempts surfaces as `context.Canceled` rather than being retried further).

On the cert/signature question: good catch — the same metadata-Get/GetBulk race window does technically reach the sigcert/ext-sigcert/signature blocks too, not just inbox/history. I didn't extend the retry there because those paths return `VerificationError`, which `orchestrator.loadInternalState` treats as tamper evidence and permanently tombstones the workflow (`tombstoneTamperedState`/`MarkAsTamperFailed`). Reclassifying a missing key there as transient isn't a mechanical change to me — it touches the tamper-detection guarantee, and I don't want to weaken that unilaterally in this PR.

A safer follow-up would be: retry those key-missing cases too, but only fall back to `VerificationError` (and tombstoning) once the retry budget is exhausted, so genuine tampering is still caught while the same race window is covered. I'd rather do that as a separate, carefully-reviewed change than fold it into this fix — let me know if you'd prefer it here instead, or if I'm missing something about why the distinction is intentional.

pujitha24:
Thanks for the review! The non-transient-error and mid-retry-cancellation test coverage you asked for is in, in 60c89b167 and 1619fb28a.

On the cert/signature question: you're right that the same metadata/bulk-read race could in principle hit the sigcert/ext-sigcert/signature keys too, since they're fetched in the same `GetBulk` call as inbox/history. I didn't extend the retry there because those "missing key" branches currently return `VerificationError`, which `orchestrator/state.go`'s `loadInternalState` specifically checks for to tombstone the workflow as tampered (`tombstoneTamperedState`). Reclassifying them as `TransientReadError` would make that tamper path retry-then-swallow instead of tombstoning, which felt like a bigger, security-relevant behavior change than I wanted to make unilaterally in this PR. Happy to take it on if you think it's the right call, or would prefer it split into its own PR given the different blast radius — let me know which you'd prefer.

Separately, the two failing checks (Standalone validations, Unit tests) look unrelated to this diff — the unit test failure is in `pkg/runtime/processor` (TestReporter), a package this PR doesn't touch, and the resource-validation benchmark is failing the same way on other unrelated branches right now.

pujitha24:
Thanks for the review, @mikeee.

- Non-transient errors not being retried: added in `TestLoadInternalState_RetriesTransientReadError/does_not_retry_a_non-transient_error` (actors_test.go) — it asserts `GetBulk` is called exactly once and the returned error is not a `TransientReadError`.
- Context cancellation mid-retry: added in the `stops_retrying_once_the_context_is_canceled_mid-retry` subtest — it asserts the backoff stops after the cancellation (no further retry) and the error surfaces as `context.Canceled`.

Both landed in an earlier commit on this branch and are passing locally.

On the cert/signature retrieval question: you're right that `LoadWorkflowState` does the same two-call (Get metadata, then GetBulk) sequencing for the signing-cert/signature keys, so the same kind of race is structurally possible there. I didn't extend `TransientReadError` to those blocks because a missing cert/signature key is currently treated as a possible-tampering signal (`VerificationError`, with a `hasTamperMarker` escape hatch) rather than a benign availability gap, and I don't want to weaken that guarantee without being sure it's safe. Would you rather I (a) apply the same transient-retry treatment there too, trusting `hasTamperMarker` to still catch genuine tampering, or (b) leave certs/signatures on the stricter path since they're written far less often mid-workflow than inbox/history? I can follow up with a scoped commit once we're aligned on which.

WhitWaldo:
@pujitha24 Please check that you're actually committing anything. You're referring to changes made and I'm not seeing any changes being committed.

JoshVanL:
Duplicate of https://github.com/dapr/dapr/pull/10288

PR #10282, opened 2026-07-27
sentry: stamp container image references into SVIDs

JoshVanL:
Embed the requesting pod's container image references (tag plus best effort resolved digest) into workload identity certificates as a non-critical X.509 extension, so downstream consumers can link an mTLS identity or attestation record back to the exact software that was running, for example to flag executors on a CVE-affected image.

The data is sourced from the pod object the Kubernetes validator already fetches from the API server, never from the requestor: the gRPC request carries no image fields, and the pod is bound to the requestor by the existing TokenReview, service account, and app-id checks. Regular containers are always included; init containers only when they are native sidecars (restartPolicy Always), which captures daprd when native sidecar mode is enabled. Digests may be empty for containers which have not started at issuance time and are re-stamped on certificate rotation.

The extension lives under OID 1.3.6.1.4.1.62884.1.1 and its value is a DER OCTET STRING wrapping a JSON array of
{role, containerName, image, digest} entries, where role is "daprd" or "app". Being non-critical, existing SPIFFE consumers ignore it. The JWKS and insecure validators have no pod to read, so certificates issued via those paths omit the extension.


pkg/sentry/server/images/images.go
```diff
@@ -0,0 +1,98 @@
...
+import (
+	"crypto/x509/pkix"
+	"encoding/asn1"
+	"encoding/json"
+	"errors"
+	"fmt"
+)
+
+// OIDContainerImages is the Dapr container images certificate extension OID.
+var OIDContainerImages = asn1.ObjectIdentifier{1, 3, 6, 1, 4, 1, 62884, 1, 1}
```
javier-aliaga: Why this number? Specially why 62884? 

https://www.iana.org/assignments/enterprise-numbers/?q=62884

mikeee: 👍 can we use the cncf one?

JoshVanL: Thanks both- updated to use https://www.iana.org/assignments/enterprise-numbers/?q=57683

javier-aliaga (approved):

mikeee (approved):

PR #10283, opened 2026-07-27
workflows: remove the clustered-deployment completion rendezvous hops

JoshVanL:
Under `WorkflowsClusteredDeployment`, task completions rendezvous with their pending work-item waiter through an "executor" actor: the waiter opens a WatchComplete stream to it at dispatch, and the completion is forwarded to it with a unary actor call. The waiter always runs on the daprd hosting the workflow or activity actor, but the executor actor is placed by hashing its own ID: for workflow tasks (ID = workflow instance ID) it already lands on the waiter's host, while for activity tasks the ID ("instanceID/taskID") differs from the activity actor ID ("instanceID::taskID::generation") and lands on an arbitrary host. Every activity therefore pays a cross-node stream setup at dispatch, two cross-node transfers of the completion payload, an executor actor lifecycle on a third node, and placement lookups.

After this change a completion costs at most one direct internal call (zero when the load balancer happens to land it on the waiter's daprd, and always zero on a single-replica deployment), with no per-work-item streams and no third host involved. Two coordinated changes achieve this in the steady state, keeping the executor actor wire contract (type, methods, payloads) unchanged as the fallback:

1. Co-locate the rendezvous with the waiter. The activity actor ID becomes "instanceID::taskID" (common.ActivityActorID) and the executor actor for an activity task uses the same string. Placement hashes only the actor ID and all workflow actor types are registered by the same hosts, so equal IDs resolve to equal hosts: the executor actor is now always co-located with the activity actor and its waiter. The durabletask execution key ("instanceID/taskID") cannot be used as the actor ID because the ID is embedded in the run-activity reminder job name and the scheduler rejects job names containing "/". The "::" form keeps the existing workflow-ID parse and the existing "instanceID::" reminder purge prefixes working unchanged for both the new and the legacy format.

2. Deliver completions through a process-local pending map. Waiters register in the map at WaitFor*Completion time, before the work item is dispatched, so no completion can be reported before the waiter exists. A completion arriving on the waiter's own daprd is delivered straight into the map with no actor machinery; a completion arriving on another daprd is forwarded via a single unary call tor, which delivers into the map in-process and deactivates. Waiters whose executor actor is not local per placement (lrs created by pre-upgrade daprds, placement disagreement windows) deregister and fall back to the existing Watchutor actor still parks completions in its channel when no local waiter is registered.

Dropping the generation from the activity actor ID is safe for result integrity: the completion rendezvous already keyed on (instanceID, taskID) only, and stale results from orphaned previous-generation activities are discarded by the payload-level dedup guards on the workflow actor. What changes is that an orphaned previous-generation activity that is still executing now shares its actor with the continued-as-new generation's same-task-ID activity: the new generation serializes behind the orphan's actor lock and converges via the durable reminder retry, and its reminder upsert replaces the orphan's reminder, stopping pointless redelivery of the orphaned execution. Covered by the new canorphan integration test.

Rolling upgrades are bridged. The activity rendezvous ID differs between versions ("instanceID/taskID" before, "instanceID::taskID" after), so a completion for an activity in flight across the upgrade boundary can be routed with one version's key while its waiter rendezvouses under the other's. Workflow-task rendezvous (bare instance ID) is unchanged and unaffected. To keep such completions flowing, the executor actor bridges the two formats: on a completion that finds neither a local waiter nor a parked watch stream, it forwards the completion once (loop-guarded via request metadata) to the sibling-format rendezvous actor, where the other version's waiter is watching or, on an upgraded host, the co-located pending map delivers it. The bridge is dormant on a homogeneous cluster and covers both directions of skew whenever the rendezvous actor lands on an upgraded daprd; in the shrinking residue of cases where it lands on a pre-upgrade daprd, the activity's durable reminder retries the execution, the same self-healing path that already covers placement churn during upgrades today. No RPC or persisted format changes: reminders created by pre-upgrade daprds keep their legacy actor IDs and are parsed, executed and purged unchanged.

The rendezvous routing was previously unmeasured: a new dapr_runtime_workflow_completion_route_count metric records how each completion is routed (wait_local/wait_watch on the wait side, complete_local/complete_actor on the completion side). The new routes integration test asserts completions rendezvous through the co-located pending map with zero watch-stream fallbacks on a two-daprd cluster.


pkg/runtime/wfengine/backends/actors/clustertasks.go
```diff
@@ -96,119 +154,130 @@ func (be *ClusterTasksBackend) CancelActivityTask(ctx context.Context, id api.In
...
+// returned function blocks until the completion is delivered.
+//
+// The pending map is only consulted by completions arriving on this daprd or
+// forwarded here by the executor actor, so blocking on it is only correct
+// when placement resolves the executor actor for key to this host. That is
+// the steady state by construction (the executor actor ID equals the
+// waiter's actor ID); when it does not hold, the waiter deregisters and falls
+// back to a watch stream on the executor actor, wherever it is placed.
+func (be *ClusterTasksBackend) waitForCompletion(taskType, key string) func(context.Context, proto.Message) error {
+	waitCh, deregister := be.pending.Register(key)
```
mikeee: Can the key be the same across the two task types leading to a collision? For example - deriving these from the tests, if you had a workflow with an instance id set to `abc::5`, could another workflow with id `abc` and activity id `5` could potentially register r `abc::5` as a key here too. If so could this be addressed by constructing a key/identity that includes the type which seems only used for metrics at the mo

mikeee (changes requested):

tests/integration/suite/daprd/workflow/loadbalance/routes.go
```diff
@@ -0,0 +1,126 @@
...
+
+func (r *routes) Setup(t *testing.T) []framework.Option {
+	r.workflow = newClusteredDeployment(t, 2)
+
+	return []framework.Option{
+		framework.WithProcesses(r.workflow),
+	}
+}
+
+func (r *routes) Run(t *testing.T, ctx context.Context) {
```
javier-aliaga, line 60: Is the test exercising the improvement? 

1. A daprd only registers workflow actor types when a worker connects to it
2.  The test gives only daprd 0 a worker.
3.  Therefore the placement ring for every workflow actor type contains exactly one host

pkg/runtime/wfengine/backends/actors/clustertasks.go
JoshVanL: Good catch, thank you! Yep, I've updated the key to include type to prevent collisions now.

tests/integration/suite/daprd/workflow/loadbalance/routes.go
JoshVanL: Fixed 👍

pkg/actors/targets/workflow/executor/executor.go
```diff
@@ -86,7 +106,12 @@ func (e *executor) complete(ctx context.Context, req *internalsv1pb.InternalInvo
...
 }
 
-func (e *executor) cancel() error {
+func (e *executor) cancel(req *internalsv1pb.InternalInvokeRequest) error {
+	if e.pending != nil && e.pending.Cancel(PendingKey(taskTypeOf(req, e.actorID), e.actorID)) {
+		e.deactivateCh <- e
+		return nil
+	}
+
 	close(e.cancelCh)
```
javier-aliaga: complete()/cancel() now do a bare e.deactivateCh <- e (no closeCh/ctx arm) while holding e.wg, and complete() can also make an uninterruptible cross-node call (forwardSibling) under that same wg. Since deactivateCh is drained by a single goroutine whose per-item work is Deactivate → wg.Wait(), one slow forward (actor-not-found retries + dial, inbound ctx usually has no deadline) stalls all deactivations on the host — and rolling upgrades, when forwards fire, are the worst time for that.

JoshVanL: Thank you, should be fixed now

JoshVanL: Thanks @javier-aliaga, I think this should be resolved now, PTAL

javier-aliaga (approved):

mikeee (approved):

cicoyle (approved):

PR #10290, opened 2026-07-28
Trust anchor distribution and automatic append-only CA renewal by Sentry

JoshVanL:
Implements trust distribution proposal (20231024-CIR-trust- distribution) and builds automatic CA renewal on top of it, so the Dapr control plane CA can be rotated with zero downtime and without any workload restarts.

Trust distribution:
- The operator runs a new trust distribution controller which writes the CA trust anchors into a `dapr-root-ca.crt` ConfigMap in every namespace, reconciles new namespaces, repairs drift, and re-distributes whenever the trust anchors change. Leader-only, metadata-only watches, with the cluster-wide ConfigMap informer restricted by a field selector. Gated by `--trust-distribution` and disabled for namespaced RBAC installs.
- The injector mounts the per-namespace ConfigMap into daprd at /var/run/secrets/dapr.io/tls (optional volume) and injects DAPR_TRUST_ANCHORS_FILE, so sidecars consume trust anchors from a hot-reloadable file instead of a baked env var. daprd gains --trust-anchors-file and the DAPR_TRUST_ANCHORS_FILE env var; the literal DAPR_TRUST_ANCHORS env var is deprecated but still honoured.

CA renewal:
- Sentry automatically renews the CA once the issuer certificate passes `--ca-renewal-threshold`, a fraction of its lifetime (default 0.9). Renewal is append-only: a fresh self-signed root (fresh key, the root key is never persisted or reused) is appended to the trust anchors and the new issuer pair is stored as issuer.next.crt/issuer.next.key beside the active pair.
- Sentry keeps signing with the old issuer key for `--trust-anchor-propagation-grace` (default 24h) so the appended anchor can propagate everywhere, then promotes the pending issuer. The state machine is derived purely from certificate timestamps, so restarts resume the pending state without re-renewing, and the switchover is clamped to never sign past the old issuer's expiry. An orphan anchor guard refuses to renew when the bundle contains an unexplained anchor.
- The kube store no longer regenerates the entire CA when the trust bundle ConfigMap is out of sync with the Secret; it heals the ConfigMap from the Secret instead.
- New flags --ca-ttl, --ca-renewal-enabled, --ca-renewal-threshold and --trust-anchor-propagation-grace, helm values under dapr_sentry.caRenewal (defaulting to disabled for bring-your-own-CA installs), and sentry/ca/renewal_total, switchover_total, renewal_pending and switchover_timestamp metrics.


tthophan (commented):
PR implements trust anchor distribution and automatic CA renewal. Logic appears sound, but `pkg/operator/trustdistribution/trustdistribution.go` is missing from the checkout. Verify file existence and content.

### Additional findings

- **WARNING** `pkg/operator/trustdistribution/trustdistribution.go (file-level)` — File missing from checkout. Cannot verify implementation.

PR #10293, opened 2026-07-29
Workflow: cross-app operations

JoshVanL:
Extends cross-app workflow support from in-workflow scheduling (child workflows, activities, detached workflows) to every client-level operation: start, get, raise event, pause, resume, terminate, purge and rerun. A caller app can target a workflow instance owned by another app in the same namespace; the target app's WorkflowAccessPolicy governs whether each operation is permitted, per operation and per workflow name. Bumps durabletask-go to v0.12.2-0.20260728154718-758d20072ebc for the TaskRouter request fields and per-operation app ID options.

API surface: the workflow request protos gain an optional app_id field (json appID), and the HTTP endpoints accept an appID query parameter. The Universal handlers validate the target (instance-ID charset, rejecting '.' so a caller cannot smuggle segments into the derived actor type name) and append the matching durabletask api.With*AppID option. An empty or local app ID preserves the existing local behaviour. Rerun remains SDK-only via the TaskHubSidecarService, which routes through the same backend.

Actors backend routing:
- CreateWorkflowInstance and RerunWorkflowFromEvent resolve the target app's workflow actor type from the event/request router instead of hardcoding the local one, stamping the local app as sourceAppID. Terminate, raise, pause and resume already routed via AddNewWorkflowEvent; purge already delegated via the recursive purge actor method, now also honouring non-recursive delegation.
- GetWorkflowMetadata and WatchWorkflowRuntimeStatus implement the new router-aware Backend signatures. A cross-app get is a one-shot fetch over the WaitForRuntimeStatus stream using a new MetadataFetchOnly flag: the target replies immediately with metadata, or a not-found status when the instance does not exist, instead of parking the stream. Older daprds ignore the flag and degrade to a wait.
- The actor router's CallStream now classifies errors like the unary path: only placement churn is retried, so access policy denials and not-found fail fast instead of burning the retry budget.

WorkflowAccessPolicy: all eight operations in the CRD enum now take real effect. RecursivePurgeWorkflowState is mapped to the purge operation, fixing a latent denial of cross-app recursive purges whenever policies were loaded. Non-schedule operations match rules against the workflow name recorded in the target instance's state; an operation on a nonexistent instance resolves to an empty name so only wildcard rules match. The requires mechanism remains schedule-only.


pkg/runtime/wfengine/backends/actors/actors.go
```diff
@@ -338,8 +339,16 @@ func (abe *Actors) RerunWorkflowFromEvent(ctx context.Context, req *backend.Reru
 		return "", fmt.Errorf("failed to marshal RerunWorkflowFromEvent: %w", err)
 	}
 
+	// A router targeting another app means the source instance (and therefore
+	// the forked instance) lives on that app: fork on its workflow actor. The
+	// subsequent RerunWorkflowInstance hop is a same-app self-call there.
+	actorType := abe.workflowActorType
+	if target := req.GetRouter().GetTargetAppID(); target != "" && target != abe.appID {
+		actorType = common.NewActorTypeBuilder(abe.namespace).Workflow(target)
```
cicoyle: Should we validate the target first?

cicoyle (changes requested):
mostly lgtm, just one comment

cicoyle (approved):
thx!

PR #10294, opened 2026-07-29
Fix streaming service invocation connections closed mid-stream by idle-purge

### LINKED ISSUE #10291: SSE - Persistent Connection across Side car timeout
## Issue Description: 
The call to subscribe to a server sent event goes through multiple services (and their sidecars). 
Between 3 minutes and 3:45 minutes the connection abruptly closes and we see following error in the calling service's sidecar

{
   "app_id":"api-gateway",
   "instance":"foundation-platform-api-gateway-6954994c94-2xvsb",
   "level":"warning",
   "msg":"HTTP service invocation failed to complete with error: error receiving message: rpc error: code = Canceled desc = grpc: the client connection is closing",
   "scope":"dapr.runtime.http",
   "time":"2026-07-16T10:22:58.027176558Z",
   "type":"log",
   "ver":"edge"
}

When we bypass the sidecar and directly connect with the main container, the connection stays open for a longer duration till it is closed by downstream service.

## Probable RCA: 
After a quick investigation I could see that dapr runtime has set a maximum threshold (maxConnIdle) of 3 minutes after which an idle connection is marked for cleanup. A scheduler that runs every 45 seconds closes the connection that is marked idle. 
Ideally if the downstream API is sending data then the connection should not be marked as Idle. We would need a way to keep the connection open indefinitely so that we can keep streaming data to clients.

pujitha24:
Motivation

Long-lived streaming responses proxied across sidecars (e.g. Server-Sent
Events) were being cut off roughly 3-3:45 minutes into the stream, with
the source sidecar logging: "error receiving message: rpc error: code =
Canceled desc = grpc: the client connection is closing".

invokeRemoteStream (pkg/messaging/direct_messaging.go) reads only the
leading chunk of a streamed response synchronously and then spawns a
background goroutine that keeps calling stream.RecvMsg in a loop to pump
the remaining body into an io.Pipe, returning to its caller as soon as
the leading chunk arrives - well before that goroutine is done. The
caller (invokeWithRetry) released the underlying grpc.ClientConn back to
the connection pool (pkg/api/grpc/manager) immediately once
invokeRemoteStream returned, even though the background goroutine was
still actively receiving data on that same connection. This let the
connection's reference count drop to zero and get marked idle almost
immediately after the call started. A periodic collector
(Manager.StartCollector, ticking every 45s) then force-closes any
connection whose reference count is <= 0 and has been idle for more than
3 minutes - killing the connection out from under the still-in-flight
streaming goroutine. This matches the reported ~3:00-3:45 minute window
exactly (3 minute idle threshold plus up to one 45s collector tick).

This only affects cross-sidecar (remote) service invocation: the
connection pool used for local sidecar-to-app calls always keeps at
least one connection active (minActiveConns=1) and is not vulnerable,
while the remote pool used for sidecar-to-sidecar calls allows all
connections to be purged (minActiveConns=0). Non-streaming/unary
invocations and short-lived streaming responses are unaffected, since
the connection was already released promptly after those complete;
user-visible behavior only changes for streaming responses that are
still being read past the idle threshold.

Approach

Thread the connection-pool teardown function into invokeRemoteStream so
it can decide when to actually release the connection. Every early
return (dial/send/receive errors, and the fallback to a unary call for
older sidecars that don't support CallLocalStream) keeps releasing the
connection immediately, exactly as before. Only for the success path,
where the background goroutine takes over reading the rest of the
stream, does the real teardown get deferred: the goroutine now releases
the connection itself once it finishes draining the stream, and
invokeRemoteStream returns a no-op teardown to its caller so the
pre-existing immediate call becomes a harmless no-op instead of a
premature release.

Validation

go build ./...
go vet -tags unit ./pkg/messaging/...
go test -tags unit -race ./pkg/messaging/...
go test -tags unit -race ./pkg/api/grpc/manager/...

All pass. Added a regression test
(TestInvokeRemote/streaming_keeps_the_connection_checked_out_until_the_response_body_is_fully_drained)
that wraps the connection-pool teardown to detect when the connection is
actually released, calls invokeRemote the way invokeWithRetry does
(including its immediate teardown(false) call), and asserts the
connection stays checked out until the response body is fully drained.
Verified this test fails on the pre-fix code and passes on the
post-fix code.

Fixes #10291

### DIFF
--- a/pkg/api/grpc/manager/manager.go
+++ b/pkg/api/grpc/manager/manager.go
@@ -18,6 +18,7 @@ import (
 	"crypto/tls"
 	"fmt"
 	"net"
+	"os"
 	"strconv"
 	"sync"
 	"sync/atomic"
@@ -37,19 +38,26 @@ import (
 	"github.com/dapr/dapr/pkg/modes"
 	"github.com/dapr/dapr/pkg/security"
 	securityConsts "github.com/dapr/dapr/pkg/security/consts"
+	"github.com/dapr/kit/logger"
 )
 
 const (
 	// needed to load balance requests for target services with multiple endpoints, ie. multiple instances.
 	grpcServiceConfig = `{"loadBalancingPolicy":"round_robin"}`
 	dialTimeout       = 30 * time.Second
-	maxConnIdle       = 3 * time.Minute
 
 	// appConnectTimeout is the budget gRPC gives a single connection attempt to
 	// the app, covering the TCP connect and the HTTP/2 handshake. This matches
 	// gRPC's own default; it does not bound how long a request waits, which
 	// remains governed by the caller's context.
 	appConnectTimeout = 20 * time.Second
+)
+
+var log = logger.NewLogger("dapr.runtime.grpc.manager")
+
+var (
+	maxConnIdle       = durationFromEnv("DAPR_GRPC_MAX_CONN_IDLE", 3*time.Minute)
+	collectorInterval = durationFromEnv("DAPR_GRPC_CONN_COLLECTOR_INTERVAL", 45*time.Second)
 )
 
 // ConnCreatorFn is a function that returns a gRPC connection
@@ -253,7 +261,7 @@ func (g *Manager) connTeardownFactory(address string, conn *grpc.ClientConn) fun
 // StartCollector starts a background goroutine that periodically watches for expired connections and purges them.
 func (g *Manager) StartCollector() {
 	g.wg.Go(func() {
-		t := time.NewTicker(45 * time.Second)
+		t := time.NewTicker(collectorInterval)
 		defer t.Stop()
 
 		for {
@@ -292,4 +300,19 @@ func (g *Manager) AddAppTokenToContext(ctx context.Context) context.Context {
 		return md.AppendToOutgoingContext(ctx, securityConsts.APITokenHeader, g.channelConfig.AppAPIToken)
 	}
 	return ctx
+}
+
+func durationFromEnv(name string, def time.Duration) time.Duration {
+	v, ok := os.LookupEnv(name)
+	if !ok {
+		return def
+	}
+
+	d, err := time.ParseDuration(v)
+	if err != nil || d <= 0 {
+		log.Warnf("Ignoring invalid duration %q in %s; using default %s", v, name, def)
+		return def
+	}
+
+	return d
 }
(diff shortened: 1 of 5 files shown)


JoshVanL (changes requested):
@pujitha24 is it possible to create an integration test for this change, where the test reproduces the bug without these implementation changes?

pujitha24:
Thanks for the feedback! I looked into adding an integration test for this, but the actual purge timing (`maxConnIdle = 3 * time.Minute` and the 45s collector tick in `pkg/api/grpc/manager`) is hardcoded and not currently configurable, so a real end-to-end reproduction would need to sleep ~3-4 minutes to trigger the purge — much longer than the existing streaming integration tests (e.g. `sserelay.go`) take today.

To make this fast enough for CI, I'd need to add a way to inject a shorter idle/collector interval into the connection pool for tests, which is a production-code change beyond this fix's scope, and I don't want to guess at the right shape for that without your input.

The unit test I added does reproduce the exact mechanism (it fails on pre-fix code and passes on post-fix code) by asserting the connection isn't released until the stream is drained, but I recognize that's not the same as an integration-level reproduction. Would you like me to add a test-only knob for the idle/collector timing so we can write a real integration test, or is the unit-level regression coverage sufficient here?

nelson-parente (dismissed):

cicoyle (dismissed):

mikeee (dismissed):

olitomlinson:
will this get back ported to 1.17? Thanks

pujitha24:
@olitomlinson Thanks for asking! Backporting to 1.17 is a maintainer/release-process decision, not something I control as the PR author, so I don't have visibility into those plans. This is a bug fix rather than a new feature, so I'd have no objection if the maintainers choose to backport it once it merges.

@JoshVanL, still standing by for your guidance on the integration-test question from my July 30 reply above — happy to add a test-only idle/collector-timing knob if that's the direction you'd like, just didn't want to make that production-code change without your input first.

pujitha24:
@JoshVanL I've pushed changes addressing your review — the branch is now at `d6e7610` and CI is green. Could you take another look when you have a moment? Happy to keep iterating if anything is still off.

pujitha24:
Also flagging for @cicoyle @mikeee @nelson-parente — the rebase that cleared the merge conflict dismissed your earlier approvals. The branch is at `d6e7610` and the diff is unchanged from what you approved. Could you re-approve when you get a chance? Sorry for the extra round trip.

JoshVanL (approved):

Issue #7485, opened 2024-02-02
Error Standardization: Configuration API


cicoyle:
Richer errors in Dapr for the **Configuration API**.

**Note**: Please clone this issue and assign the clone of this issue to yourself if you would like to contribute to the error standardization efforts.

# Scope of work

- [ ] Choose 5-10 errors in the Configuration API to enrich. Ideally, cover as many errors as possible from the API.
- [ ] Follow the steps outlined in [this README](https://github.com/dapr/dapr/blob/master/pkg/api/errors/README.md) for each error being updated

# Definition of done

- [ ] Update 5-10 errors in the Configuration API
- [ ] Update the unit tests, if necessary
- [ ] Write integration tests to cover each updated errors

# Keep in mind

Unless a major version is being released and there is a solid reason for introducing backwards incompatibility, we can't change the API, meaning the HTTP and gRPC status codes, and the error codes must stay the same. An exception to this is the gRPC `StatusCode.UNKNOWN` which should be converted to a specific status.

See the [kit errors pkg README](https://github.com/dapr/kit/blob/main/errors/README.md) for another usage example. For each error being updated, you can use the following helper functions to add error details, see the kit [errors pkg](https://github.com/dapr/kit/tree/main/errors) for more details:

- `WithHelpLink()`
-  `WithHelp()`
-  `WithErrorInfo()` - **Required**
-  `WithFieldViolation()`
-  `WithDetails()`

# Reference PRs
Existing error standardization for: 
- [State API](https://github.com/dapr/dapr/pull/7257)
- [PubSub API](https://github.com/dapr/dapr/pull/7322)

stackanthony:
Hi @cicoyle, does this issue still need to be worked on? I haven't contributed to the project yet, but looking forward to working on this one, if possible.

cicoyle:
> Hi @cicoyle, does this issue still need to be worked on? I haven't contributed to the project yet, but looking forward to working on this one, if possible.

Yes this is available to be worked on. I'm happy to hear you are interested in contributing to Dapr. Please clone this issue and assign yourself to it.

Please lmk if you need any assistance along the way 🙂

stackanthony:
Forsure, will do. I'll need some time to get acclimated to the codebase. Will let you know if I have any questions.

FYI - I worked at HashiCorp too, awesome to see someone from there contributing to the project @cicoyle

stackanthony:
[...]

PR #9782, opened 2026-04-11
feat: standardize Configuration API errors using rich error types

ogulcanaydogan:
## What problem does this PR solve?

Closes #7485 — Error Standardization: Configuration API

The Configuration API used ad-hoc error construction via `apierrors.Basic()` (gRPC) and `messages.NewAPIErrorHTTP()` (HTTP), while other APIs like State and PubSub already use structured, domain-specific error types.

## Changes

- **New**: `pkg/api/errors/configuration.go` — `ConfigurationStoreError` struct with the same pattern as `StateStoreError`
  - `NotFound(appID string) error`
  - `NotConfigured(appID string) error`
  - `GetFailed(keys []string, err error) error`
  - `SubscribeFailed(keys []string, err error) error`
  - `UnsubscribeFailed(id string, err error) error`

- **Updated** `pkg/api/http/http.go` — replaced 3 error sites in `getConfigurationStoreWithRequestValidation`, `onSubscribeConfiguration`, and `onGetConfiguration`

- **Updated** `pkg/api/grpc/grpc.go` — replaced 3 error sites in `getConfigurationStore`, `GetConfiguration`, and `subscribeConfiguration`

- **Updated** `pkg/api/http/http_test.go` — adjusted assertions for the new standardized error messages

## Testing

All existing HTTP API configuration tests pass:
```
go test ./pkg/api/http/... -run "Configuration" -v
--- PASS: TestConfigurationGet
--- PASS: TestV1Alpha1ConfigurationUnsubscribe
```

PR #10306, opened 2026-07-30
Standardize Configuration API errors with the rich error model

ogulcanaydogan:
Migrates the Configuration API errors to the gRPC richer error model, following the pattern in `pkg/api/errors/state.go`.

Part of #7485.

### What this does
- Adds `pkg/api/errors/configuration.go` with a `ConfigurationError` builder covering `NotConfigured`, `NotFound`, `GetFailed`, `SubscribeFailed` and `UnsubscribeFailed`.
- Fills in the previously empty gRPC reason codes for the Configuration error codes (`DAPR_CONFIGURATION_*`) in `pkg/messages/errorcodes/errorcodes.go`.
- Updates the gRPC handlers in `pkg/api/grpc/grpc.go` to use the new builders instead of the generic `apierrors.Basic(...)` calls.
- Adds unit tests for the builders.

### Scope / follow-up
This covers the gRPC side. The HTTP handlers in `pkg/api/http/http.go` still use the older `NewAPIErrorHTTP` flow. I'm happy to migrate those in this PR or a follow-up, whichever you prefer. I'm also happy to adjust the `DAPR_CONFIGURATION_*` reason names if there's a preferred convention.

### DIFF
--- a/pkg/api/errors/configuration.go
+++ b/pkg/api/errors/configuration.go
@@ -1,0 +1,124 @@
+/*
+Copyright 2024 The Dapr Authors
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


ogulcanaydogan:
Closing this — it overlaps my earlier #9782, and the error standardization effort has not been getting maintainer review. Happy to revisit if the campaign becomes active again.

PR #10307, opened 2026-07-31
workflows: fix lost completion rendezvous

### LINKED ISSUE #10305: Workflow cancel-before-register race is lost on the local fast path introduced by #10283
## In what area(s)?

> /area runtime

## What version of Dapr?

> edge: v1.18.0-rc.1-180-g2cf451e2a (commit 2cf451e2a, same day as #10283 merged)

## Expected Behavior

When a workflow/activity task's execution is cancelled before the corresponding wait call has registered (a real, plausible ordering: durabletask-go's `grpcExecutor.Shutdown`/stream-disconnect cleanup iterates `pendingWorkflows`/`pendingActivities` and calls `CancelActivityTask`/`CancelWorkflowTask` on entries stored *before* the wait closure returned by `WaitForActivityCompletion`/`WaitForWorkflowTaskCompletion` is ever invoked), the waiter should observe the cancellation promptly and return `api.ErrTaskCancelled`, the same as it did before #10283 and the same as the current watch-stream fallback path still does today.

## Actual Behavior

Under `WorkflowsClusteredDeployment`, PR #10283 ("workflows: remove the clustered-deployment completion rendezvous hops") added a process-local fast path in `ClusterTasksBackend.waitForCompletion` (`pkg/runtime/wfengine/backends/actors/clustertasks.go`) that, when placement resolves the executor actor locally, only selects on the process-local `pending.Pending` map's channel and `ctx.Done()`. It never inspects the executor actor's own `cancelCh` fallback (`pkg/actors/targets/workflow/executor/executor.go`'s `cancel()`), which is the only place an at-least-once cancel lands when no waiter is registered yet in the pending map.

Concretely: if `Cancel` reaches the executor actor before `Register()` runs for that key, `pending.Pending.Cancel()` finds no live waiter and returns `false`, so `executor.cancel()` falls through to closing `cancelCh` - a signal the new local fast path never looks at. The waiter then hangs until its context deadline (or forever, if the context has no deadline) instead of promptly observing the cancellation.

The PR's own author was careful about the symmetric "completion arrives before the waiter registers" race (documented in `waitForCompletion`'s own godoc, with an explicit note that a lost completion self-heals via the durable reminder retry), and reviewers separately caught and fixed three other subtle issues in the same diff (a nil-panic, a key collision between workflow/activity ID spaces, and a deactivation-stalling goroutine issue) before merge. The cancel-before-register case appears to have been missed.

Practical impact: the scenario that triggers this (connection churn, worker shutdown, stream disconnects) is also when Cancel calls fire, so this turns what should be a fast, bounded cleanup path into a context-timeout-bounded hang, or an unbounded goroutine leak if the caller's context has no deadline - concentrated in exactly the window this PR's own fast path targets for its performance benefit.

1. Enable the `WorkflowsClusteredDeployment` feature, with placement resolving the executor actor for a given key to the local host (the fast-path condition).
2. Call `Cancel` for that key (e.g. via `ClusterTasksBackend.CancelActivityTask`) *before* the corresponding `WaitForActivityCompletion`/`WaitForWorkflowTaskCompletion` closure has been invoked for that same key.
3. Invoke the wait closure. Expected: it returns `api.ErrTaskCancelled` immediately. Actual: it blocks until `ctx` is done.

I have two unit tests demonstrating this precisely (one showing the bug on the local fast path, one showing the same sequence works correctly via the watch-stream fallback as a positive control), plus a design for a minimal fix in `pending.go` that records a pre-registration cancellation instead of dropping it, verified against the existing test suite with no regressions. Happy to open a PR once this issue is confirmed/triaged, per CONTRIBUTING.md's "open an issue before a PR" policy.

## Release Note

RELEASE NOTE: **FIX** Workflow/activity task cancellation issued before the corresponding wait call registers is no longer lost on the clustered-deployment local fast path.

JoshVanL:
Under WorkflowsClusteredDeployment, the pending-map waiter registers when the wait function is invoked, which happens after the work item has been dispatched to the worker. A fast worker can report the completion before the waiter registers: the completion finds no pending-map entry, is forwarded to the co-located executor actor, finds no entry there either, and is parked in the actor's channel, which the wait_local path never consults. The waiter then registers and blocks on the pending map forever. The durable reminder retry cannot redispatch the work item because the reminder invocation is the very call chain that is blocked, so the workflow hangs permanently.

Fix with register-then-claim: after registering in the pending map on the wait_local path, the waiter makes one non-blocking Claim call to the co-located executor actor to drain a completion or cancellation parked before registration. Claim pairs with complete()'s deliver-then-park: whichever side loses the race, one side always observes the other. A parked completion of the colliding other task type is put back for its own waiter, and a stale watch stream is fed a copy on claim, mirroring the existing deliver-success behaviour. A claim error is returned rather than waited out so the work item is abandoned and the durable retry converges.

Fixes https://github.com/dapr/dapr/issues/10305

### DIFF
--- a/pkg/actors/fake/fake.go
+++ b/pkg/actors/fake/fake.go
@@ -17,8 +17,10 @@ import (
 	"context"
 
 	"github.com/dapr/dapr/pkg/actors"
+	"github.com/dapr/dapr/pkg/actors/api"
 	"github.com/dapr/dapr/pkg/actors/hostconfig"
 	"github.com/dapr/dapr/pkg/actors/internal/placement"
+	placementfake "github.com/dapr/dapr/pkg/actors/internal/placement/fake"
 	"github.com/dapr/dapr/pkg/actors/reminders"
 	remindersfake "github.com/dapr/dapr/pkg/actors/reminders/fake"
 	"github.com/dapr/dapr/pkg/actors/router"
@@ -124,6 +126,20 @@ func (f *Fake) WithReminders(fn func(context.Context) (reminders.Interface, erro
 
 func (f *Fake) WithPlacement(fn func(context.Context) (placement.Interface, error)) *Fake {
 	f.fnPlacement = fn
+	return f
+}
+
+// WithPlacementLookupActor configures the fake's placement to answer
+// LookupActor with the given function. Unlike WithPlacement, callers outside
+// the actors tree can use this without importing the internal placement
+// package.
+func (f *Fake) WithPlacementLookupActor(fn func(context.Context, *api.LookupActorRequest) (*api.LookupActorResponse, error)) *Fake {
+	f.fnPlacement = func(context.Context) (placement.Interface, error) {
+		return placementfake.New().WithLookupActor(func(ctx context.Context, req *api.LookupActorRequest) (*api.LookupActorResponse, context.Context, context.CancelCauseFunc, error) {
+			lar, err := fn(ctx, req)
+			return lar, ctx, func(error) {}, err
+		}), nil
+	}
 	return f
 }
 
--- a/pkg/actors/targets/workflow/executor/executor.go
+++ b/pkg/actors/targets/workflow/executor/executor.go
@@ -50,6 +50,15 @@ type executor struct {
 
 	watchLock chan struct{}
 
+	// displaced holds a parked completion of the colliding other task type
+	// that a claim drained from completeCh. It cannot be put back on the
+	// channel: a completer blocked on a full channel (the only sender
+	// outside mu) can refill the freed slot first, and the non-blocking
+	// put-back would silently drop the payload. Guarded by mu; consumed by
+	// the first matching claim or watch stream. While it is non-nil the
+	// actor must not deactivate, or the payload would be stranded.
+	displaced *internalsv1pb.InternalInvokeResponse
+
 	// mu serializes each side's check-then-act pair of the rendezvous:
 	// complete's pending-map miss followed by its channel park, cancel's map
 	// miss followed by closing cancelCh, claim's drain of both, and close on
@@ -114,8 +123,11 @@ func (e *executor) complete(ctx context.Context, req *internalsv1pb.InternalInvo
 		case e.completeCh <- d:
 		default:
 		}
+		deactivate := e.displaced == nil
 		e.mu.Unlock()
-		e.tryDeactivate()
+		if deactivate {
+			e.tryDeactivate()
+		}
 		return nil
 	}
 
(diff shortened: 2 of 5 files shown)


nelson-parente (dismissed):

javier-aliaga (changes requested):
@JoshVanL I think this narrows the race window but still does not close it in all cases:

```
The completer's two steps are not atomic. There's a gap between its pending-map miss (executor.go:96) and its park (executor.go:121), and the waiter's entire register+claim sequence can fit inside that gap:

completer: Deliver → MISS          (waiter not registered yet)
waiter:    Register                 (too late — completer already checked)
waiter:    Claim → NOT_FOUND        (too early — completer hasn't parked yet)
waiter:    blocks on pending map
completer: parks into completeCh    (nobody will ever read it)
completer: returns nil → app told "success"

Both sides did their steps in the prescribed order, yet neither observed the other. The completion sits in a channel with no reader; the waiter sits in the pending map that already got its (missed) check. The app thinks it succeeded, so nothing retries from that side. And the durable reminder retry can't rescue it for the same reason as the original bug — the reminder invocation is the blocked call chain. Permanent hang, identical symptom to #10305.
```

and also:

```The escape hatch in the not-found path (the comment at executor.go:186-187) is racy: tryDeactivate is async, so complete() usually parks before closeCh closes; and when it doesn't, the select has two ready cases and picks randomly. The closed-actor retry only fires on the lucky half of an already-narrow sub-case.```

zanarellidev:
Thanks for the fix direction in #10307. I opened #10305 with repro unit tests for the cancel/completion-before-register race on the local fast path. Happy to help here with regression tests (including the deliver→park gap @javier-aliaga described) or a follow-up patch if useful. Let me know what you prefer.

JoshVanL:
@javier-aliaga please can you take another look? https://github.com/dapr/dapr/pull/10307/commits/16a9c80a20646880f4c48164680643a962d34ee2

javier-aliaga (approved):

cicoyle (approved):

zanarellidev:
Thanks for the fix direction in #10307. I opened #10305 with repro unit tests for the cancel/completion-before-register race on the local fast path. Happy to help here with regression tests (including the deliver→park gap @javier-aliaga described) or a follow-up patch if useful. Let me know what you prefer.

zanarellidev:
Thanks for the fix direction in #10307. I opened #10305 with repro unit tests for the cancel/completion-before-register race on the local fast path. Happy to help here with regression tests (including the deliver→park gap @javier-aliaga described) or a follow-up patch if useful. Let me know what you prefer.

zanarellidev:
Thanks for the fix direction in #10307. I opened #10305 with repro unit tests for the cancel/completion-before-register race on the local fast path. Happy to help here with regression tests (including the deliver→park gap @javier-aliaga described) or a follow-up patch if useful. Let me know what you prefer.

zanarellidev:
Thanks for the fix direction in #10307. I opened #10305 with repro unit tests for the cancel/completion-before-register race on the local fast path. Happy to help here with regression tests (including the deliver→park gap @javier-aliaga described) or a follow-up patch if useful. Let me know what you prefer.

zanarellidev:
Thanks for the fix direction in #10307. I opened #10305 with repro unit tests for the cancel/completion-before-register race on the local fast path. Happy to help here with regression tests (including the deliver→park gap @javier-aliaga described) or a follow-up patch if useful. Let me know what you prefer.

zanarellidev:
Thanks for the fix direction in #10307. I opened #10305 with repro unit tests for the cancel/completion-before-register race on the local fast path. Happy to help here with regression tests (including the deliver→park gap @javier-aliaga described) or a follow-up patch if useful. Let me know what you prefer.

zanarellidev:
Thanks for the fix direction in #10307. I opened #10305 with repro unit tests for the cancel/completion-before-register race on the local fast path. Happy to help here with regression tests (including the deliver→park gap @javier-aliaga described) or a follow-up patch if useful. Let me know what you prefer.

zanarellidev:
Thanks for the fix direction in #10307. I opened #10305 with repro unit tests for the cancel/completion-before-register race on the local fast path. Happy to help here with regression tests (including the deliver→park gap @javier-aliaga described) or a follow-up patch if useful. Let me know what you prefer.

zanarellidev:
Thanks for the fix direction in #10307. I opened #10305 with repro unit tests for the cancel/completion-before-register race on the local fast path. Happy to help here with regression tests (including the deliver→park gap @javier-aliaga described) or a follow-up patch if useful. Let me know what you prefer.

zanarellidev:
Thanks for the fix direction in #10307. I opened #10305 with repro unit tests for the cancel/completion-before-register race on the local fast path. Happy to help here with regression tests (including the deliver→park gap @javier-aliaga described) or a follow-up patch if useful. Let me know what you prefer.

zanarellidev:
Thanks for the fix direction in #10307. I opened #10305 with repro unit tests for the cancel/completion-before-register race on the local fast path. Happy to help here with regression tests (including the deliver→park gap @javier-aliaga described) or a follow-up patch if useful. Let me know what you prefer.

zanarellidev:
Thanks for the fix direction in #10307. I opened #10305 with repro unit tests for the cancel/completion-before-register race on the local fast path. Happy to help here with regression tests (including the deliver→park gap @javier-aliaga described) or a follow-up patch if useful. Let me know what you prefer.

zanarellidev:
Thanks for the fix direction in #10307. I opened #10305 with repro unit tests for the cancel/completion-before-register race on the local fast path. Happy to help here with regression tests (including the deliver→park gap @javier-aliaga described) or a follow-up patch if useful. Let me know what you prefer.

zanarellidev:
Thanks for the fix direction in #10307. I opened #10305 with repro unit tests for the cancel/completion-before-register race on the local fast path. Happy to help here with regression tests (including the deliver→park gap @javier-aliaga described) or a follow-up patch if useful. Let me know what you prefer.

zanarellidev:
Thanks for the fix direction in #10307. I opened #10305 with repro unit tests for the cancel/completion-before-register race on the local fast path. Happy to help here with regression tests (including the deliver→park gap @javier-aliaga described) or a follow-up patch if useful. Let me know what you prefer.

zanarellidev:
Thanks for the fix direction in #10307. I opened #10305 with repro unit tests for the cancel/completion-before-register race on the local fast path. Happy to help here with regression tests (including the deliver→park gap @javier-aliaga described) or a follow-up patch if useful. Let me know what you prefer.

zanarellidev:
Thanks for the fix direction in #10307. I opened #10305 with repro unit tests for the cancel/completion-before-register race on the local fast path. Happy to help here with regression tests (including the deliver→park gap @javier-aliaga described) or a follow-up patch if useful. Let me know what you prefer.

zanarellidev:
Thanks for the fix direction in #10307. I opened #10305 with repro unit tests for the cancel/completion-before-register race on the local fast path. Happy to help here with regression tests (including the deliver→park gap @javier-aliaga described) or a follow-up patch if useful. Let me know what you prefer.

zanarellidev:
Thanks for the fix direction in #10307. I opened #10305 with repro unit tests for the cancel/completion-before-register race on the local fast path. Happy to help here with regression tests (including the deliver→park gap @javier-aliaga described) or a follow-up patch if useful. Let me know what you prefer.

zanarellidev:
Thanks for the fix direction in #10307. I opened #10305 with repro unit tests for the cancel/completion-before-register race on the local fast path. Happy to help here with regression tests (including the deliver→park gap @javier-aliaga described) or a follow-up patch if useful. Let me know what you prefer.

zanarellidev:
Thanks for the fix direction in #10307. I opened #10305 with repro unit tests for the cancel/completion-before-register race on the local fast path. Happy to help here with regression tests (including the deliver→park gap @javier-aliaga described) or a follow-up patch if useful. Let me know what you prefer.

zanarellidev:
Thanks for the fix direction in #10307. I opened #10305 with repro unit tests for the cancel/completion-before-register race on the local fast path. Happy to help here with regression tests (including the deliver→park gap @javier-aliaga described) or a follow-up patch if useful. Let me know what you prefer.

zanarellidev:
Thanks for the fix direction in #10307. I opened #10305 with repro unit tests for the cancel/completion-before-register race on the local fast path. Happy to help here with regression tests (including the deliver→park gap @javier-aliaga described) or a follow-up patch if useful. Let me know what you prefer.

zanarellidev:
Thanks for the fix direction in #10307. I opened #10305 with repro unit tests for the cancel/completion-before-register race on the local fast path. Happy to help here with regression tests (including the deliver→park gap @javier-aliaga described) or a follow-up patch if useful. Let me know what you prefer.

zanarellidev:
Thanks for the fix direction in #10307. I opened #10305 with repro unit tests for the cancel/completion-before-register race on the local fast path. Happy to help here with regression tests (including the deliver→park gap @javier-aliaga described) or a follow-up patch if useful. Let me know what you prefer.

zanarellidev:
Thanks for the fix direction in #10307. I opened #10305 with repro unit tests for the cancel/completion-before-register race on the local fast path. Happy to help here with regression tests (including the deliver→park gap @javier-aliaga described) or a follow-up patch if useful. Let me know what you prefer.

zanarellidev:
Thanks for the fix direction in #10307. I opened #10305 with repro unit tests for the cancel/completion-before-register race on the local fast path. Happy to help here with regression tests (including the deliver→park gap @javier-aliaga described) or a follow-up patch if useful. Let me know what you prefer.

zanarellidev:
Thanks for the fix direction in #10307. I opened #10305 with repro unit tests for the cancel/completion-before-register race on the local fast path. Happy to help here with regression tests (including the deliver→park gap @javier-aliaga described) or a follow-up patch if useful. Let me know what you prefer.

zanarellidev:
Thanks for the fix direction in #10307. I opened #10305 with repro unit tests for the cancel/completion-before-register race on the local fast path. Happy to help here with regression tests (including the deliver→park gap @javier-aliaga described) or a follow-up patch if useful. Let me know what you prefer.

zanarellidev:
Thanks for the fix direction in #10307. I opened #10305 with repro unit tests for the cancel/completion-before-register race on the local fast path. Happy to help here with regression tests (including the deliver→park gap @javier-aliaga described) or a follow-up patch if useful. Let me know what you prefer.

zanarellidev:
Thanks for the fix direction in #10307. I opened #10305 with repro unit tests for the cancel/completion-before-register race on the local fast path. Happy to help here with regression tests (including the deliver→park gap @javier-aliaga described) or a follow-up patch if useful. Let me know what you prefer.

zanarellidev:
Thanks for the fix direction in #10307. I opened #10305 with repro unit tests for the cancel/completion-before-register race on the local fast path. Happy to help here with regression tests (including the deliver→park gap @javier-aliaga described) or a follow-up patch if useful. Let me know what you prefer.

zanarellidev:
Thanks for the fix direction in #10307. I opened #10305 with repro unit tests for the cancel/completion-before-register race on the local fast path. Happy to help here with regression tests (including the deliver→park gap @javier-aliaga described) or a follow-up patch if useful. Let me know what you prefer.

zanarellidev:
Thanks for the fix direction in #10307. I opened #10305 with repro unit tests for the cancel/completion-before-register race on the local fast path. Happy to help here with regression tests (including the deliver→park gap @javier-aliaga described) or a follow-up patch if useful. Let me know what you prefer.

PR #10309, opened 2026-07-31
actors: allow hot reloading the actor state store

JoshVanL:
Remove the hot reload restriction on components marked as the actor state store. The actors runtime now reconciles hosting with the store at runtime:

- Adding a store enables actor hosting and the workflow APIs, including for workflow workers connected before the store existed.
- Removing it drains and deactivates hosted actors, de-advertises actor types from placement and scheduler, and actor state and workflow APIs error until a store returns. daprd keeps running and hosting resumes automatically on re-add.
- Updating it in place (e.g. a reload picking up a rotated secret) swaps the store instance without interrupting actor hosting; the store is resolved per call. Draining is reserved for the store being removed, unmarked, or replaced by a differently-named component.

The component store is now the single source of truth: the actor state store is resolved per call, the hot reload reconciler notifies the actors runtime once a store change settles, and the actor table suspends or resumes hosting while retaining registered factories. Actor types are only advertised to placement while a store is configured. Only one actor state store is allowed; hot loading a second is skipped with an error log.


tests/integration/suite/daprd/hotreload/selfhosted/actorstateupdate.go
```diff
@@ -0,0 +1,209 @@
...
+		case act := <-a.deactivatedCh:
+			assert.Failf(t, "unexpected actor deactivation", "actor %s", act)
+		default:
+		}
+		require.EventuallyWithT(t, func(c *assert.CollectT) {
+			code, body := invokeActor(c)
+			assert.Equalf(c, http.StatusOK, code, "body: %s", body)
+		}, time.Second*10, time.Millisecond*10)
+	})
+
```
javier-aliaga, line 187: Are we sure at this point the deactivation has happened? Should we wait to a log event saying the state store has been reloaded?

pkg/runtime/hotreload/reconciler/components.go
```diff
@@ -117,39 +127,38 @@ func (c *components) update(ctx context.Context, comp compapi.Component) error {
...
+// changed.
+func (c *components) notifyActorStateStoreChanged() func() {
+	//nolint:dogsled
+	_, _, before, _ := c.store.GetStateStoreActorWithRevision()
+	return func() {
+		if _, _, after, _ := c.store.GetStateStoreActorWithRevision(); after != before {
+			c.proc.OnActorStateStoreChanged()
+		}
 	}
+}
```
javier-aliaga, line 200: It seems this is not run when there is a secretRef in the component definition

```
The actor runtime only learns of store-slot changes through this defer, but a marked store parked behind 
a not-yet-loaded secret store is flushed via handleInstanceInitDone → loops.Init{Internal: true} (root/component.go:151-158), 
reaching AddStateStoreActor without passing through update() — so the rev bumps but no kick fires, 
and hosting stays suspended despite a live store (the periodic reconcile won't repair it: committed 
component → no diff).
```

pkg/runtime/hotreload/reconciler/components.go
```diff
@@ -54,10 +54,20 @@ func NewComponents(opts Options[compapi.Component]) *Reconciler[compapi.Componen
 }
 
 func (c *components) update(ctx context.Context, comp compapi.Component) error {
-	if !c.verify(comp) {
+	// Only a single actor state store may be configured. Skip a component
+	// which would become a second actor state store, rather than failing its
+	// init and exiting daprd.
+	if _, name, ok := c.store.GetStateStoreActor(); ok && name != comp.Name && isMarkedActorStateStore(comp) {
+		log.Errorf("Aborting hot reload of %s: %s is already the actor state store, only one actor state store is allowed", comp.LogName(), name)
```
javier-aliaga: This still could happens in 2 places:

(a) Two marked stores in one reconcile batch. The reconciler processes each event group concurrently — reconciler.go:208-221 spawns a goroutine per resource and only wg.Wait()s between groups. So two new marked stores arriving in the same Created batch (one kubectl apply -f dir/, or the re-list after an operator stream reconnect) both read an empty slot, both pass the guard, both proceed to init. state.Init serializes them on its lock; the loser gets the duplicate error.

(b) The parked-dependent flush from the secret reference — that path never enters update(), so it never sees the guard at all.

javier-aliaga: It seems events could arrive in any order. When receiving delete and then create it works as expected. But what happens if create is received the first? It seems we let daprd errroing until the reconciliations runs.

```
kubectl apply -k/-f dir/, a Helm upgrade, or a GitOps sync that removes the old component and adds the new one in one shot → both events hit the API server milliseconds apart → arrival order at daprd is effectively a coin flip. This is the realistic trigger, and it's exactly how a store migration would ship.
- A human running kubectl delete … waits … kubectl create → ordering is almost certainly preserved in practice, though nothing formally guarantees it.
```

pkg/actors/actors_test.go
```diff
@@ -49,3 +56,123 @@ func TestHostValidation(t *testing.T) {
...
+		require.NoError(t, cs.AddStateStoreActor("mystore", store(t)))
+		a.convergeHosting(t.Context())
+
+		// A different component becoming the actor state store is a
+		// different backing store, so hosted actors are drained.
+		cs.DeleteStateStore("mystore")
+		require.NoError(t, cs.AddStateStoreActor("otherstore", store(t)))
+		a.convergeHosting(t.Context())
+		assert.Equal(t, 1, *suspends)
+		assert.Equal(t, 2, *resumes)
```
javier-aliaga: Can we assert ordering too not only count?

tests/integration/suite/daprd/hotreload/selfhosted/actorstateupdate.go
JoshVanL: Thanks, done

pkg/runtime/hotreload/reconciler/components.go
JoshVanL: Thanks, done

pkg/runtime/hotreload/reconciler/components.go
JoshVanL: I think this is fine behaviour, what do you think we should do instead?

javier-aliaga: yeah, probably. I was looking different ways to solve it, but all of them are complex and does not solve the problem of 2 actors states stores in the same batch, just mitigate them.

javier-aliaga (dismissed):

pkg/runtime/hotreload/reconciler/components.go
mikeee, line 114: With this PR, the component appears to be dropped before the authorizaiton gate. Is this intended or should the existing component live if the update is unsuccessful?

JoshVanL: This is intentional- the way an update becomes unauthorized is the component being rescoped away from this app ID, so the component should be removed for this daprd.

mikeee (dismissed):

javier-aliaga (approved):

yaron2 (approved):

alicejgibbons:
Does this need docs @JoshVanL ??

JoshVanL:
@alicejgibbons thanks, good catch, done
https://github.com/dapr/docs/pull/5277
https://github.com/dapr/docs/pull/5278

PR #10311, opened 2026-07-31
Bound injector admission request body size to prevent unbounded reads

### LINKED ISSUE #10310: Injector admission handler reads the full request body without a size limit
## In what area(s)?

/area injector

## What version of Dapr?

edge: tested against commit 8fd31d1. I also checked current master at afb22f809c5915f7682880f6e360a9a2a1a68b5d and the same code path is still present.

## Expected Behavior

The injector admission handler should apply a request body size limit before reading the admission request body into memory.

## Actual Behavior

The injector admission handler reads the full request body with io.ReadAll.

Relevant code in pkg/injector/service/handler.go:

    body, err := io.ReadAll(r.Body)

The injector HTTP server sets ReadHeaderTimeout, but I did not find a request body size limit on this server path.

Relevant code in pkg/injector/service/injector.go:

    server: &http.Server{
        Addr:              fmt.Sprintf("%s:%d", opts.ListenAddress, opts.Port),
        Handler:           mux,
        ReadHeaderTimeout: 10 * time.Second,
    }

Local validation output:

    H1 = TRUE POSITIVE: readRequestBody read all 2097152 bytes via io.ReadAll with NO LimitReader

1. In a local checkout, add or run a package-level test for pkg/injector/service that calls readRequestBody with a large request body.
2. Use an httptest request whose Body contains a large byte slice.
3. Call readRequestBody(req).
4. Observe that the function reads and returns the full body instead of stopping at a configured or fixed maximum size.

In my local Docker-based test, readRequestBody read all 2097152 bytes.

## Release Note

RELEASE NOTE: **FIX** Add a request body size limit to the injector admission handler.

pujitha24:
Motivation:
The injector's admission webhook handler (pkg/injector/service/handler.go)
read the entire HTTP request body via io.ReadAll(r.Body) with no size
limit. The server sets ReadHeaderTimeout but nothing bounds the size of
the body itself, so a client able to reach this endpoint could send an
arbitrarily large request and force the injector to buffer it all into
memory. This is a resource-exhaustion hardening gap in an admission
webhook handler, not a demonstrated crash or outage.

Approach:
Wrap r.Body with the existing github.com/dapr/kit/streams.LimitReadCloser
helper (already used by pkg/api/http/middlewares.go's MaxBodySizeMiddleware
and pkg/channel/http/http_channel.go for the same purpose) before reading
it, capping reads at a new maxAdmissionRequestBodySize constant (4 MiB,
matching runtime.DefaultMaxRequestBodySize, Dapr's own existing HTTP body
size default). When the limit is exceeded, readRequestBody now returns
streams.ErrStreamTooLarge, and handleRequest responds with 413 Request
Entity Too Large instead of falling through to the generic "empty body"
400 response. AdmissionReview payloads embed a single Pod object, which
is normally well under this ceiling (Kubernetes' own default etcd object
size limit is ~1.5 MiB), so this should not affect legitimate traffic.

Validation:
- go build ./...
- go test ./pkg/injector/... (all packages pass)
- golangci-lint run ./pkg/injector/service/... — 0 new findings; the 8
  pre-existing goconst findings in this package are unchanged from master
- Added TestReadRequestBody, a table test exercising readRequestBody
  directly for within-limit, empty, nil, and over-limit bodies
- Added TestHandleRequestRejectsOversizedBody, an end-to-end test that
  POSTs an oversized body through a real httptest.NewServer connection
  and asserts the handler now responds 413 instead of reading the full
  body

Fixes #10310

### DIFF
--- a/pkg/injector/service/handler.go
+++ b/pkg/injector/service/handler.go
@@ -48,7 +48,7 @@ func (i *injector) handleRequest(w http.ResponseWriter, r *http.Request) {
 			http.Error(w, "request body too large", http.StatusRequestEntityTooLarge)
 			return
 		}
-		log.Error("Empty body")
+		log.Errorf("Failed to read request body: %v", err)
 		http.Error(w, "empty body", http.StatusBadRequest)
 		return
 	}
@@ -107,7 +107,7 @@ func readRequestBody(r *http.Request) ([]byte, error) {
 		if errors.Is(err, streams.ErrStreamTooLarge) {
 			return nil, err
 		}
-		return nil, errors.New("empty body")
+		return nil, fmt.Errorf("failed to read request body: %w", err)
 	}
 	if len(body) == 0 {
 		return nil, errors.New("empty body")
--- a/pkg/injector/service/handler.go
+++ b/pkg/injector/service/handler.go
@@ -29,13 +29,25 @@ import (
 
 	"github.com/dapr/dapr/pkg/injector/annotations"
 	"github.com/dapr/dapr/utils"
+	"github.com/dapr/kit/streams"
 )
+
+// maxAdmissionRequestBodySize is the maximum size, in bytes, of an admission
+// request body the injector will read. AdmissionReview payloads embed the
+// full Pod object, which is normally small, so this is a generous ceiling
+// meant to bound memory use against oversized or malicious requests.
+const maxAdmissionRequestBodySize = 4 << 20 // 4 MiB
 
 func (i *injector) handleRequest(w http.ResponseWriter, r *http.Request) {
 	RecordSidecarInjectionRequestsCount()
 
 	body, err := readRequestBody(r)
 	if err != nil {
+		if errors.Is(err, streams.ErrStreamTooLarge) {
+			log.Errorf("Request body exceeds the %d byte limit", maxAdmissionRequestBodySize)
+			http.Error(w, "request body too large", http.StatusRequestEntityTooLarge)
+			return
+		}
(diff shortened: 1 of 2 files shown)


pkg/injector/service/handler.go
```diff
@@ -81,15 +93,23 @@ func (i *injector) handleRequest(w http.ResponseWriter, r *http.Request) {
...
-	body, err := io.ReadAll(r.Body)
-	if err != nil || len(body) == 0 {
+	limited := streams.LimitReadCloser(r.Body, maxAdmissionRequestBodySize)
+	defer limited.Close()
+	body, err := io.ReadAll(limited)
+	if err != nil {
+		if errors.Is(err, streams.ErrStreamTooLarge) {
+			return nil, err
+		}
+		return nil, errors.New("empty body")
```
mikeee: I think propoagating the error here might be more helpful than just constructing a new error, wdyt?

mikeee (changes requested):

pujitha24:
Good catch, thanks! I've updated `readRequestBody` to wrap and return the underlying `io.ReadAll` error (`fmt.Errorf("failed to read request body: %w", err)`) instead of discarding it into a generic `errors.New("empty body")`, and updated `handleRequest` to log that propagated error instead of a static "Empty body" message. The client-facing HTTP response is unchanged (still 400 "empty body") — only the server-side error/log now carries the real cause. Existing tests pass.

tthophan (approved):

PR #10336, opened 2026-08-10
grpc: give the app connection realistic connect budget

### LINKED ISSUE #10335: Pub/Sub error use of closed network connection not seen in previous released, 1.13 tests ok
## In what area(s)?

> /area runtime
Get pub/sub errors when running 1.18 and 1.18.3-rc.2 using RabbitMQ

[15:51:00 ERR] rabbitmq pub/sub error: handling message from topic 'TopicTestStatus', error returned from app while processing pub/sub event d6c32369-1158-490e-aca7-945271f0c18c: retriable error occurred: rpc error: code = Unavailable desc = connection error: desc = "error reading server preface: read tcp 127.0.0.1:56264->127.0.0.1:14208: use of closed network connection" {"SourceContext": "Man.Dapr.Sidekick.DaprSidecarHost", "DaprVersion": "1.18.3-rc.2", "DaprAppId": "main-analysisinvoker", "DaprInstance": "CPWS06129", "DaprScope": "dapr.contrib", "DaprTime": "2026-08-10T15:51:00.8666739-05:00", "DaprType": "log", "AppId": "main-analysisinvoker"}

Errors do not happen on 1.13 and possible other releases (not tested). This error appears on the latest release 1.18 but could be with other releases.  Using RabitMQ for pub/sub messages.

> /area operator

> /area placement

> /area docs

> /area test-and-release

> /area injector

> /area scheduler

> /area sentry

## What version of Dapr?

> 1.1.x
> 1.0.x

## Release Note

RELEASE NOTE:

JoshVanL:
Pub/sub delivery to a gRPC app intermittently failed before the app saw the message, with:

```
  error returned from app while processing pub/sub event <id>: retriable
  error occurred: rpc error: code = Unavailable desc = connection error:
  desc = "error reading server preface: read tcp ...: use of closed
  network connection"
```

daprd dialed the app with a MinConnectTimeout of one second. In gRPC that is the budget for an entire connection attempt, covering the TCP connect and the HTTP/2 handshake, not just the TCP connect. When it expires gRPC hard-closes the socket, so a request already riding on that connection fails while reading the server preface.

Up to 1.17 this was almost never reachable: daprd held a single app connection for the lifetime of the process and only dialed at startup. Since #9492 the app connection lives in the same pool used for sidecar-to-sidecar connections, so connections are established on demand while messages are being delivered, and each new one had one second to complete its handshake.

The pool made this more frequent than it needed to be. It keeps one warm connection to the app, but once that connection had been idle longer than the pool's idle window it was neither handed out again (treated as expired) nor closed (it was the connection being kept warm). It sat in the pool unusable and open, so every request after an idle period dialed a new connection instead of reusing the warm one.

Give a connection attempt to the app gRPC's default 20 second budget, matching what daprd already used for sidecar-to-sidecar connections, and stop expiring out connections the pool holds to satisfy its warm connection minimum. This does not change how long a request waits for an unreachable app: a refused connection still fails immediately, and the wait remains governed by the caller's context and resiliency policy.

Fixes #10335

### DIFF
--- a/docs/release_notes/v1.18.3.md
+++ b/docs/release_notes/v1.18.3.md
@@ -14,6 +14,7 @@ This update contains the following bug fixes:
 - [Workflow `continue_as_new` iterations sharing a single unbounded trace](#workflow-continue_as_new-iterations-sharing-a-single-unbounded-trace)
 - [Stalled workflows permanently stuck after the last workflow worker disconnects](#stalled-workflows-permanently-stuck-after-the-last-workflow-worker-disconnects)
 - [Fixes orphaned workflow activity-result reminders retrying forever](#fixes-orphaned-workflow-activity-result-reminders-retrying-forever)
+- [Pub/sub delivery to a gRPC app failing with "use of closed network connection"](#pubsub-delivery-to-a-grpc-app-failing-with-use-of-closed-network-connection)
 
 ## Actor state store components cannot be hot reloaded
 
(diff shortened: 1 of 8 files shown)


cicoyle (dismissed):
lgtm

cicoyle:
changes lgtm, but I see the test failure is related - mind taking a look at it?
```
daprd/serviceinvocation/grpc/appconnredial/daprd
```

JoshVanL:
Thanks @cicoyle, please take another look

cicoyle (approved):

Issue #10337, opened 2026-08-11
daprd never recovers after placement/scheduler control-plane restart. Stuck permanently on `dapr is not ready: [grpc-api-server grpc-internal-server]`


pf56:
## In what area(s)?

/area runtime

## What version of Dapr?

1.16.4-msft.11

## Expected Behavior

After a control-plane restart, once the control plane becomes reachable again, the sidecar should either:

1. re-initialize the actor runtime, workflow engine, and gRPC API servers and resume serving, or
2. exit with a non-zero code so the orchestrator replaces it.

## Actual Behavior

The runtime shuts down permanently and never attempts re-initialization. The container remains running.

### Impact

Complete, silent loss of Dapr functionality for the replica until manual intervention. Nothing surfaces it: the container is up, restart count is zero, and the only signal is a debug-level log line. Note the readiness failure is logged at debug level, so on any normal log level the failure is entirely invisible.

### Workaround

Manually restart the replica/revision. The sidecar does not self-heal.

## Steps to Reproduce the Problem

We are experiencing this issue with the managed Dapr offering on Azure Container Apps. This seems to (sometimes, but not always?) happen when the Dapr control-plane is restarted. After a Dapr control-plane restart (placement leader change + scheduler server restart), the sidecar shuts down its actor runtime and workflow engine and never re-initializes them. The container stays alive and keeps logging, but its gRPC API servers are never restored, so the app's gRPC endpoint on localhost:50001 is dead for the remaining lifetime of the replica (multiple hours in our case, with no self-recovery and no crash to trigger a restart).

Logs of the ACA using Dapr Workflows:

```
[Healthy baseline. Heartbeart is successful]
time="2026-08-09T11:00:13.43874308Z" level=info msg="gRPC API Called" app_id=my-workflow-app code=0 duration=0 instance=my-workflow-container-app--ep9ksm9-697bdc5975-vsfps method=/TaskHubSidecarService/Hello scope=dapr.runtime.grpc.api-info type=log useragent="grpc-dotnet/2.76.0 (.NET 9.0.18; CLR 9.0.18; net9.0; linux; x64) dapr-sdk-dotnet/v1.17.9+08d9e625f4ebf480025e3f4e4684456654b64575" ver=1.16.4-msft.11
...
[...]

PR #10338, opened 2026-08-11
runtime: surface readiness-check failures at warn instead of debug

pujitha24:
# Description

Today, when daprd's readiness check fails (`/v1.0/healthz` or `/v1.0/healthz/outbound`), the failure is only logged at `Debug` level. On any normal (default) log verbosity, a stuck-not-ready sidecar produces zero log output — the only way to notice is to already be polling the health endpoint from outside. This PR raises those readiness-failure log lines to `Warn`, logged once per not-ready streak (not once per poll, since readiness probes can run as often as every second), with a matching `Info` log when the target recovers.

This does **not** fix the underlying reason a sidecar can get stuck not-ready in the first place (that is a separate, larger architectural question about whether/how the actor runtime and workflow engine subsystems should restart after certain fatal errors) — it only makes an already-failing readiness check visible in logs at default settings, instead of requiring `--log-level debug`.

## Issue reference

Relates to #10337, but does not fix it. That issue's logs show the scheduler `WatchHosts` stream failing to reconnect, what looks like the same root cause #9968 addressed on master; #10337 was reported against 1.16.4-msft.11, so the actual fix there is likely a backport of #9968 rather than this logging-only change.

## Checklist

- [x] Code compiles correctly
- [x] Created/updated tests
- [x] Unit tests passing

### DIFF
--- a/pkg/api/http/healthz.go
+++ b/pkg/api/http/healthz.go
@@ -59,8 +59,15 @@ func (a *api) onGetHealthz(w http.ResponseWriter, r *http.Request) {
 	if !a.healthz.IsReady() {
 		msg := messages.ErrHealthNotReady.WithFormat(a.healthz.GetUnhealthyTargets())
 		respondWithError(w, msg)
-		log.Debug(msg)
+		// Only log once per not-ready streak, at a level visible on default log settings,
+		// so the failure isn't silent but readiness polling doesn't spam the logs.
+		if a.healthzNotReadyLogged.CompareAndSwap(false, true) {
+			log.Warn(msg)
+		}
 		return
+	}
+	if a.healthzNotReadyLogged.CompareAndSwap(true, false) {
+		log.Info("dapr is ready again")
 	}
 
 	// If we have an "appid" parameter in the query string, we will return an error if the ID of this app is not the value of the requested "appid"
@@ -80,8 +87,13 @@ func (a *api) onGetOutboundHealthz(w http.ResponseWriter, r *http.Request) {
 	if !a.outboundHealthz.IsReady() {
 		msg := messages.ErrOutboundHealthNotReady
 		respondWithError(w, msg)
-		log.Debug(msg)
+		if a.outboundNotReadyLogged.CompareAndSwap(false, true) {
+			log.Warn(msg)
+		}
 		return
+	}
+	if a.outboundNotReadyLogged.CompareAndSwap(true, false) {
+		log.Info("dapr outbound is ready again")
 	}
 
 	respondWithEmpty(w)
--- a/pkg/api/http/healthz.go
+++ b/pkg/api/http/healthz.go
@@ -59,13 +59,20 @@ func (a *api) onGetHealthz(w http.ResponseWriter, r *http.Request) {
 	if !a.healthz.IsReady() {
 		msg := messages.ErrHealthNotReady.WithFormat(a.healthz.GetUnhealthyTargets())
 		respondWithError(w, msg)
-		// Only log once per not-ready streak, at a level visible on default log settings,
-		// so the failure isn't silent but readiness polling doesn't spam the logs.
-		if a.healthzNotReadyLogged.CompareAndSwap(false, true) {
-			log.Warn(msg)
+		// A startup that hasn't become ready yet is expected and stays quiet at Debug.
+		// Once we've been ready at least once, a not-ready poll means a regression, so
+		// surface it at a level visible on default log settings. Still only log once per
+		// not-ready streak, since readiness polling can run as often as every second.
+		if a.healthzEverReady.Load() {
+			if a.healthzNotReadyLogged.CompareAndSwap(false, true) {
+				log.Warn(msg.Message())
+			}
+		} else {
+			log.Debug(msg)
 		}
 		return
 	}
+	a.healthzEverReady.Store(true)
 	if a.healthzNotReadyLogged.CompareAndSwap(true, false) {
 		log.Info("dapr is ready again")
 	}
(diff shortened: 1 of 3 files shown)


pkg/api/http/healthz.go
```diff
@@ -59,9 +59,16 @@ func (a *api) onGetHealthz(w http.ResponseWriter, r *http.Request) {
 	if !a.healthz.IsReady() {
 		msg := messages.ErrHealthNotReady.WithFormat(a.healthz.GetUnhealthyTargets())
 		respondWithError(w, msg)
-		log.Debug(msg)
+		// Only log once per not-ready streak, at a level visible on default log settings,
+		// so the failure isn't silent but readiness polling doesn't spam the logs.
+		if a.healthzNotReadyLogged.CompareAndSwap(false, true) {
+			log.Warn(msg)
```
JoshVanL: This will fire a WARN on essentially every healthy pod start, not just on the
stuck-sidecar case this is targeting.  Keep Debug until the not-ready condition has *persisted*, and only then
escalate to Warn. Either track the streak start time and warn past a threshold
(~30s), or gate on "has been ready at least once" so that a never-yet-ready startup
stays quiet and only a regression after ready warns. The second option also fixes
the "ready again" wording below.

pkg/api/http/http_test.go
```diff
@@ -4180,6 +4180,83 @@ func TestV1HealthzEndpoint(t *testing.T) {
...
+		testAPI.healthzNotReadyLogged.Store(false)
+		htarget.NotReady()
+		t.Cleanup(htarget.NotReady)
+
+		server := fakeServer(t)
+
+		resp := server.DoRequest("GET", apiPath, nil, nil)
+		assert.Equal(t, 500, resp.StatusCode)
+		assert.True(t, testAPI.healthzNotReadyLogged.Load(), "flag should be set after the first not-ready poll")
+
```
JoshVanL, line 4219: Nothing verifies the log level actually changed

JoshVanL (changes requested):

pkg/api/http/healthz.go
```diff
@@ -59,9 +59,23 @@ func (a *api) onGetHealthz(w http.ResponseWriter, r *http.Request) {
 	if !a.healthz.IsReady() {
 		msg := messages.ErrHealthNotReady.WithFormat(a.healthz.GetUnhealthyTargets())
 		respondWithError(w, msg)
-		log.Debug(msg)
+		// A startup that hasn't become ready yet is expected and stays quiet at Debug.
+		// Once we've been ready at least once, a not-ready poll means a regression, so
+		// surface it at a level visible on default log settings. Still only log once per
+		// not-ready streak, since readiness polling can run as often as every second.
+		if a.healthzEverReady.Load() {
```
JoshVanL: This gate fixes the startup-noise problem, but it also reintroduces the original one
for the case the issue actually describes.

If daprd never reaches ready at all, `healthzEverReady` stays false forever and this
path logs only at Debug for the entire outage. That is reachable: the HTTP server is
serving from runtime.go:831 while init continues, and `blockUntilAppIsReady`
(runtime.go:1344) loops indefinitely if the app never opens its port, as does a
component or actor-runtime hang before `rtHealthz.Ready()` at runtime.go:443. A
sidecar stuck never-ready is arguably the more common "stuck" shape, and it is now
silent again.

A persistence threshold handles both cases without the startup noise: track when the
not-ready streak began and only escalate to Warn once it has lasted longer than, say,
30s. A normal 2s startup stays quiet, a stuck startup warns, and a post-ready
regression warns too. Keeping `everReady` alongside it is reasonable (warn
immediately on regression, warn after the threshold otherwise), but on its own it
does not cover the reported scenario.

pkg/api/http/healthz.go
```diff
@@ -80,9 +94,19 @@ func (a *api) onGetOutboundHealthz(w http.ResponseWriter, r *http.Request) {
 	if !a.outboundHealthz.IsReady() {
 		msg := messages.ErrOutboundHealthNotReady
 		respondWithError(w, msg)
-		log.Debug(msg)
+		if a.outboundEverReady.Load() {
```
JoshVanL: Worth checking whether this branch is reachable in practice. The injector's readiness
probe only targets /v1.0/healthz (injector/consts/consts.go:30, wired at
patcher/sidecar_container.go:256); nothing polls /v1.0/healthz/outbound by default.
The only in-tree caller is `WaitUntilDaprOutboundReady` (pkg/runtime/wait.go:31),
which is the `daprd wait` path.

Since `outboundEverReady` is only set by a successful poll of this same endpoint, in
a default k8s deployment it stays false for the whole process lifetime and this Warn
never fires. The outbound half of the PR is effectively a no-op unless the user wired
their own outbound probe.

JoshVanL (changes requested):

pujitha24:
Good catch on both. I added a persistence threshold: track when a not-ready streak started, and escalate to Warn once it's been not-ready for more than 30s, keeping the immediate-warn-on-regression path via `everReady` alongside it. So a normal ~2s startup stays quiet, a sidecar that never comes up warns after 30s, and a post-ready regression still warns right away. Same mechanism is used for the outbound endpoint, so it now warns after 30s of not-ready polling regardless of whether it's ever reached ready, rather than depending on `outboundEverReady` (which, as you noted, never gets set without a custom probe). Added tests for both new cases by backdating the streak-start timestamp rather than sleeping 30s in the test.

pujitha24:
The golangci-lint failure was a real dupword lint error from a typo I introduced ("should warn exactly once once the threshold is crossed") in the two new threshold tests — fixed that. The e2e (v1.33.4, ha) failure looks unrelated to this change: the log shows ACR registry auth failures and an SSL cert error building the actorphp/actorjava/actordotnet test images, nothing touching healthz.go or its tests.

mikeee (changes requested):
The pr body incorrectly states that it fixes #10337 

Perhaps backporting #9968 is a better idea as this issue was resolved in 1.17.x with the stream now reconnecting

PR #10340, opened 2026-08-11
Fix devcontainer image rebuild trigger

### LINKED ISSUE #10313: Devcontainer ships Go 1.24.6 while go.mod requires Go 1.26.2, breaking go mod tidy/build
## In what area(s)?

/area test-and-release

## What version of Dapr?

edge: output of `git describe --dirty`

## Expected Behavior

Opening the dapr/dapr devcontainer should provide a Go toolchain that satisfies the `go` directive in `go.mod` (currently `go 1.26.5`). `docker/Dockerfile-dev` already declares `ARG GOVERSION=1.26.2`, so the published `ghcr.io/dapr/dapr-dev:latest` image should reflect that and provide a working Go toolchain out of the box.

## Actual Behavior

The published devcontainer image (`ghcr.io/dapr/dapr-dev:latest`, referenced directly in `.devcontainer/devcontainer.json`) still ships Go 1.24.6, even though the Dockerfile it's supposed to be built from already specifies `GOVERSION=1.26.2`. Since `GOTOOLCHAIN=local` is set in the container, Go won't self-heal by downloading a newer toolchain, so any command touching `go.mod` fails immediately:

$ go version
go version go1.24.6 linux/amd64
$ go mod tidy
go: go.mod requires go >= 1.26.5 (running go 1.24.6; GOTOOLCHAIN=local)

This reproduces even after deleting the local devcontainer image and pulling fresh, confirming the issue is with the **published** `:latest` image on ghcr.io, not local Docker caching — the image simply hasn't been rebuilt/republished since `Dockerfile-dev` was updated to `GOVERSION=1.26.2`.

1. Clone `dapr/dapr` at `master`.
2. Open the repo in VS Code and select "Reopen in Container" (or run `devcontainer up`), which pulls `ghcr.io/dapr/dapr-dev:latest` as defined in `.devcontainer/devcontainer.json`.
3. Inside the container, run `go version` — it reports `go1.24.6`.
4. Run `go mod tidy` or `go build ./...` — fails with `go.mod requires go >= 1.26.5 (running go 1.24.6; GOTOOLCHAIN=local)`.

## Release Note

RELEASE NOTE: **FIX** Rebuild and republish ghcr.io/dapr/dapr-dev:latest to match Dockerfile-dev's Go version.

Advait2912:
## Description

Updated the devcontainer publishing workflow to automatically run on changes to the devcontainer configuration, Docker files, or the workflow itself. This ensures the published devcontainer image is rebuilt when relevant changes are made and prevents the image from becoming stale.

## Issue reference

Please reference the issue this PR will close: #10313

### DIFF
--- a/.github/workflows/dapr-dev-container.yml
+++ b/.github/workflows/dapr-dev-container.yml
@@ -15,6 +15,22 @@ name: dapr-dev-container
 
 on:
   workflow_dispatch:
+  push:
+    branches:
+      - master
+      - release-*
+    paths:
+      - '.github/workflows/dapr-dev-container.yml'
+      - 'docker/**'
+      - '.devcontainer/**'
+  pull_request:
+    branches:
+      - master
+      - release-*
+    paths:
+      - '.github/workflows/dapr-dev-container.yml'
+      - 'docker/**'
+      - '.devcontainer/**'
   schedule:
     # Run weekly on Tuesdays
     - cron: "22 03 * * 2"
--- a/docker/Dockerfile-dev
+++ b/docker/Dockerfile-dev
@@ -2,7 +2,7 @@
 
 # [Choice] Go version: 1, 1.26.5, etc
 ARG GOVERSION=1.26.5
-FROM golang:${GOVERSION}-bullseye
+FROM golang:${GOVERSION}-bookworm
 
 # [Option] Install zsh
 ARG INSTALL_ZSH="true"


Advait2912:
@cicoyle  @JoshVanL  I investigated the remaining devcontainer build failure and verified the base image independently.
go.mod requires Go 1.26.5, so the Go version cannot be downgraded. I also verified that golang:1.26.5-bookworm is published for both amd64 and arm64.
image
I then inspected the Bookworm image directly. The image does not contain the legacy /etc/apt/sources.list; instead, its repositories are defined in /etc/apt/sources.list.d/debian.sources.
image
This matches the failure in CI. The existing common-debian.sh directly modifies /etc/apt/sources.list when ADD_NON_FREE_PACKAGES=true.
image
I have not modified common-debian.sh yet. I wanted to confirm the intended approach before adding a shared-script change to this PR.
Based on this, the remaining fix appears to be making common-debian.sh support the DEB822 debian.sources layout while preserving the existing /etc/apt/sources.list path for older Debian-based images.

I can make the minimal backward-compatible change and validate it against the multi-architecture devcontainer build. Before adding that as the next commit, does this approach align with how you'd like the shared Debian setup script handled?

PR #10342, opened 2026-08-12
perf: grpc: raise the pooled connection stream cap to 2048

JoshVanL:
grpcMaxConcurrentStreams capped every pooled connection at 100 concurrent shared refs, and exhausting all pooled connections dropped callers into a write-locked growth path whose new-connection mTLS handshakes serialize per peer.
cross-host in-flight count (call rate times parent-turn hold time) reaches 100 at moderate rates: goroutine dumps at the best clean rate show 97 callers in the dial path and 48 parked on stream quota, and past the cap the dial storm plus quota waits collapse throughput while every other resource idles. This cap, not any engine or storage resource, was the measured cluster throughput ceiling across the cycle-12 investigation.

The internal HTTP/2 servers advertise no 100-stream limit; the value was self-imposed. 2048 keeps a growth bound while taking the cap out of the operating range.


cicoyle (approved):

cicoyle:
Changes lgtm, but can you check the failing static check in ci?

PR #10345, opened 2026-08-12
scheduler: retry client creation instead of fatal exit

JoshVanL:
The scheduler run loop now recreates the server after a jittered backoff (base 500ms, cap 10s) when a server incarnation stops with a runtime error, instead of exiting the process. Context cancellation still stops the loop, and errors from server construction remain fatal (they are configuration failures). Healthz reports not-ready between incarnations, so Kubernetes probes still observe the outage truthfully.

Branched from https://github.com/dapr/dapr/pull/10344

### DIFF
--- a/cmd/scheduler/app/app.go
+++ b/cmd/scheduler/app/app.go
@@ -16,8 +16,10 @@ package app
 import (
 	"context"
 	"os"
+	"time"
 
 	"github.com/dapr/dapr/cmd/scheduler/options"
+	wfcommon "github.com/dapr/dapr/pkg/actors/targets/workflow/common"
 	"github.com/dapr/dapr/pkg/buildinfo"
 	"github.com/dapr/dapr/pkg/healthz"
 	healthzserver "github.com/dapr/dapr/pkg/healthz/server"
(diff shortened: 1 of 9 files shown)


cmd/scheduler/app/app.go
```diff
@@ -171,3 +162,60 @@ func Run() {
...
+		if ctx.Err() != nil {
+			return ctx.Err()
+		}
+
+		server, err := getServer()
+		if err != nil {
+			return err
+		}
+
+		err = server.Run(ctx)
```
mikeee, line 210: In k8s mode it reads as though the controller is initialised outside of this retry loop on L99-108, so after one failure there will be a cancellation propagated in the context - this retry loop attempts to run against an exited controller?

mikeee (changes requested):

cmd/scheduler/app/app.go
JoshVanL: Thanks @mikeee, good catch- done!

cicoyle (approved):

mikeee (approved):
dismissing the blocking status - lgtm

PR #10346, opened 2026-08-12
actors: survive transient state store outage

JoshVanL:
A failed init of the component marked as the actor state store is now retried with jittered backoff (base 500ms, cap 10s) instead of being recorded as a fatal init error. Each attempt goes through the normal category/instance init path with the same per-attempt timeout, the caller stays blocked so runtime readiness keeps reporting not-ready, and shutdown cancels the loop cleanly (an abandoned retry is not recorded as fatal). All other components keep the existing fail-fast behavior, as does the actor state store when IgnoreErrors is set.

Branched from https://github.com/dapr/dapr/pull/10344

### DIFF
--- a/pkg/runtime/processor/loops/root/component.go
+++ b/pkg/runtime/processor/loops/root/component.go
@@ -19,8 +19,8 @@ import (
 	"strings"
 	"time"
 
-	wfcommon "github.com/dapr/dapr/pkg/actors/targets/workflow/common"
 	compapi "github.com/dapr/dapr/pkg/apis/components/v1alpha1"
+	"github.com/dapr/dapr/pkg/backoff"
 	"github.com/dapr/dapr/pkg/components"
 	rterrors "github.com/dapr/dapr/pkg/runtime/errors"
 	"github.com/dapr/dapr/pkg/runtime/processor/loops"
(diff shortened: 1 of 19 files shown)


pkg/runtime/processor/loops/root/component.go
```diff
@@ -108,15 +113,28 @@ func (r *Root) handleInit(ctx context.Context, ev *loops.Init) {
...
 		innerErr := <-intercept
+		// The actor state store retries in place of failing fatally: a
+		// transient outage of its backing database (for example Postgres
+		// restarting) must leave the runtime alive and unready rather than
+		// crash it, which under Kubernetes turns a seconds-long database blip
+		// into a crash-restart cycle of every replica that lasts until the
+		// database returns. The caller stays blocked while retrying, so
+		// readiness keeps reporting the degraded state.
+		var abandoned bool
+		if innerErr != nil && shouldRetryInit(comp) {
```
mikeee: With the new shouldRetryInit function checking only the component config; does this not mean _all_ init errors are retried rather than selectively... e.g. could permanent misconfiguration continue retrying indefinitely? / is this intentional

pkg/runtime/processor/loops/root/component.go
```diff
@@ -108,15 +113,28 @@ func (r *Root) handleInit(ctx context.Context, ev *loops.Init) {
...
+			innerErr, abandoned = r.retryInit(catLoop, comp, timeout, innerErr)
```
mikeee: Could this retry be superseded by a newer component update or deletion, or alterantively scoped to only on init events? It appears to stop only when the original configuration succeeds or when the runtime shuts down. Would a failed hot-reload update prevent the hot-reloader from applying an updated configuration without restarting the sidecar?

JoshVanL:
Thanks @mikeee, please can you take another look?

mikeee (dismissed):
lgtm

cicoyle (dismissed):

PR #10348, opened 2026-08-12
scheduler: honor invoke cancellation, tolerate unknown job acks

JoshVanL:
Two independent robustness fixes found while auditing the scheduler job delivery path.

App channel: the max-concurrency limiter acquires its slot with a bare channel send, so a caller whose context is already cancelled queues on the limiter indefinitely: a dead scheduler stream or an aborted invocation still consumes queue position and blocks forever. All four acquire sites (HTTP and gRPC, invokeMethodV1 and the job trigger path) now select on ctx.Done() and return the context error. The gRPC invokeMethodV1 path additionally released its slot twice, once via defer and once inline after OnInvoke: with app-max-concurrency set and a gRPC app, the second release blocked the return path on the empty limiter channel until the next request's acquire fed it, holding completed responses hostage and defeating the concurrency bound. The inline release is removed; the deferred release runs exactly once.

Scheduler: a WatchJobs result whose id matches no inflight entry (a duplicate or retried ack from a sidecar, or an ack racing the failed-Send resolution) was treated as a protocol error, exiting the receive loop and tearing down the whole stream: every other inflight trigger on that stream was aborted and redelivered because of one stray message. Drop the unknown result with a debug log instead.

Branched from https://github.com/dapr/dapr/pull/10347

### DIFF
--- a/pkg/scheduler/server/internal/pool/loops/stream/recv.go
+++ b/pkg/scheduler/server/internal/pool/loops/stream/recv.go
@@ -68,7 +68,7 @@ func (s *stream) recv() error {
 		// failed-Send resolution which already resolved the trigger. Drop
 		// it rather than treating it as a protocol error: closing the
 		// stream would abort every other inflight trigger on it.
-		log.Debugf("Dropping unknown trigger response %d from %s/%s", result.GetId(), s.ns, s.appID)
+		log.Warnf("Dropping unknown trigger response %d from %s/%s", result.GetId(), s.ns, s.appID)
 		return nil
 	}
 
--- a/pkg/scheduler/server/internal/pool/loops/stream/recv.go
+++ b/pkg/scheduler/server/internal/pool/loops/stream/recv.go
@@ -63,7 +63,13 @@ func (s *stream) recv() error {
 
 	inf, ok := s.inflight.LoadAndDelete(result.GetId())
 	if !ok {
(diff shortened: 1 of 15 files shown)


mikeee (dismissed):

pkg/scheduler/server/internal/pool/loops/stream/recv.go
```diff
@@ -63,7 +63,13 @@ func (s *stream) recv() error {
 
 	inf, ok := s.inflight.LoadAndDelete(result.GetId())
 	if !ok {
-		return errors.New("received unknown trigger response from stream")
+		// A result whose id matches no inflight entry is a duplicate or
+		// stale ack, for example a client retry, or an ack racing the
+		// failed-Send resolution which already resolved the trigger. Drop
+		// it rather than treating it as a protocol error: closing the
+		// stream would abort every other inflight trigger on it.
+		log.Debugf("Dropping unknown trigger response %d from %s/%s", result.GetId(), s.ns, s.appID)
```
mikeee:
```suggestion
		log.Warnf("Dropping unknown trigger response %d from %s/%s", result.GetId(), s.ns, s.appID)
```

Should this be at least a warning to show when the stream is misbehaving rather than obscuring by default?

mikeee (approved):

cicoyle (approved):
lgtm

PR #10350, opened 2026-08-13
integration: make the test output readable

JoshVanL:
A full suite run printed 25,444 lines. 8,073 of those were `go test -v` framing, roughly 17,000 were process logs dumped straight into the terminal, and the rest were libraries logging to stderr behind the framework's back. The failures were somewhere in the middle, and finding the logs explaining one meant scrolling past the logs of the 1,346 tests that passed.

Process logs now go to one file per failed test under /tmp/dapr_integration_logs. The terminal gets the assertion and a pointer:

```
    daprd.go:60: Error: Not equal: expected 1234, actual 1027
    logs: /tmp/dapr_integration_logs/ports.daprd.log
```

and the end of the run lists every failure with its log path. Inside a file, framework events come first, then one block per process, every line stamped with how long after the test started it was written. Dapr log lines are parsed down to level, scope and message instead of repeating a full logfmt prefix on each one, which takes a typical line from about 180 columns to 70. Anything that does not parse, panics, stack traces, etcd output, is passed through untouched. Where a test runs several of the same process they are indexed, so two sidecars read as daprd and daprd-1.

Dropping the process logs left nothing on screen during a run, so the framework now draws a progress line:

```
    512/1347  ok 510  fail 2  1m04s  actors/reminders/period
```

It is written to /dev/tty rather than stdout, because `go test` captures the binary's output and discards it entirely for a passing package. That also means `go test > out.txt` still gets clean output while progress stays on screen. Failures print above the line as they happen. It turns itself off with no terminal, on CI, and under -v, where `go test` is already printing a line per test.

Three separate channels were bypassing the framework and writing to stderr, where the output belongs to no test and cannot be suppressed:

  - dapr's own logger, from packages such as pkg/security running in-process, 217 lines a run;
  - the durabletask client, iting to the process's stderr, 1,742 lines a run. Its logger is captured at package init so it cannothe injection of logger.New(t) at every call site, which is most of this diff;
  - the standard library logest apps and, less obviously, by net/http for TLS handshake errors when a server has no ErrorLog of its o

The first and last are rediress.log beside
the per-test reports. The etcd client gets a logger for the same reason. The disk tests re-exec the tch was piping a
whole second test run to the terminal; that child is now captured like any other process, and given its start of
run reset was deleting every other test's report.

Also fixes a data race in actors/deactivation/move, where an HTTP handler wrote a variable thee fails every
test running in parallel when the detector fires, so this alone accounted for 18 failures in

# Description

## Issue reference

## Checklist

- [ ] Code compiles correctly
- [ ] Created/updated tests
- [ ] Unit tests passing

### DIFF
--- a/tests/integration/framework/iowriter/collector.go
+++ b/tests/integration/framework/iowriter/collector.go
@@ -1,0 +1,379 @@
+/*
+Copyright 2026 The Dapr Authors
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package iowriter
+
+import (
+	"fmt"
+	"io"
+	"os"
+	"sort"
+	"strconv"
+	"strings"
+	"sync"
+	"sync/atomic"
+	"time"
+
+	kitstrings "github.com/dapr/kit/strings"
+)
+
+// eventsBlock is the name of the block holding framework events, i.e. what the
+// test harness did rather than what a process printed.
+const eventsBlock = "framework"
+
+// bannerWidth is the width the log banner is padded to.
+const bannerWidth = 78
+
+// collectors holds the collector for each test. Entries outlive their test so
+// that late output from a process finds a collector to discard it, but the
+// captured lines themselves are released once the test has reported.
+var collectors sync.Map
+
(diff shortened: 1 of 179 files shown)


mikeee (dismissed):

cicoyle (dismissed):

tthophan (commented):
AI review findings are included with this review.

tests/integration/framework/iowriter/collector.go
```diff
@@ -0,0 +1,379 @@
...
+		head.WriteString(" " + key + "=" + vals[key])
+	}
+	head.WriteString(" " + rule(2))
+
+	return colorize(head.String(), ansiDim, ansiBold) + "\n", lifted
+}
+
+func offset(d time.Duration) string {
+	return fmt.Sprintf("%7.3fs", d.Seconds())
+}
```
tthophan, line 379: collectorFor function creates a new collector for each test, which is good.

cicoyle (approved):

PR #10364, opened 2026-08-17
Placement -> Scheduler Part 3: Server

cicoyle:
This PR has been broken up into smaller chunks per the following. It now contains the scheduler-side placement service only.

move placement -> scheduler via a scheduler flag rather than a sidecar feature flag since this is a cluster wide toggle. The only toggle is `global.scheduler.placement.enabled`, where sidecars follow whichever authority the control plane advertises. Prevent split brain scenarios. Schedulers withhold placement advertising whenever a placement service is present, and old sidecars hold the gate, meaning schedulers wont advertise while any connected sidecar is too old to follow. 

Presence detection and the cutover/rollback handling land in [PR Part 5](https://github.com/dapr/dapr/pull/10416): a pod informer in k8s and address probing (dns & sidecar reported addrs) in self hosted - nothing is stored, so leader crashes/restarts cant bring back up a zombie 2nd authority. Toggling the helm flag either way does not require a sidecar restart. Scheduler now uses a rendezvous hashing algo for placement (https://en.wikipedia.org/wiki/Rendezvous_hashing). perf tests run twice for wf and actor tests to test the suite for both algos (old in placement and new in scheduler). Should have full test coverage: unit/IT/e2e/perf

-----------
Breaking this PR up into the following ordered chunks:
[Part 0: Rendezvous Hashing Algo](https://github.com/dapr/dapr/pull/10437)
[Part 1: Protos](https://github.com/dapr/dapr/pull/10438)
[Part 2: Scheduler & Helm Flags](https://github.com/dapr/dapr/pull/10439)

Next after this PR is merged
[Part 4: Client logic](https://github.com/dapr/dapr/pull/10440)
[Part 5: Handover logic](https://github.com/dapr/dapr/pull/10416)


cicoyle:
/ok-to-perf actor_activation actor_double_activation actor_id_scale actor_reminder actor_timer actor_type_scale workflows scheduler-placement

cicoyle:
/ok-to-perf actor_activation actor_double_activation actor_id_scale actor_reminder actor_timer actor_type_scale workflows scheduler-placement

cicoyle:
/ok-to-perf actor_activation actor_double_activation actor_id_scale actor_reminder actor_timer actor_type_scale workflows scheduler-placement

tests/integration/suite/actors/call/badmethod.go
```diff
@@ -33,21 +34,37 @@ func init() {
...
-		framework.WithProcesses(b.actors),
+		framework.WithProcesses(procs...),
 	}
 }
 
 func (b *badmethod) Run(t *testing.T, ctx context.Context) {
+	t.Run("placement", func(t *testing.T) { b.place.run(t, ctx) })
+	t.Run("scheduler", func(t *testing.T) { b.sched.run(t, ctx) })
+}
+
```
JoshVanL: Rather than updating every suite to have both tests- please can we instead update the framework so we don't have to update every test code?
We can even have another job which tests for the placement in scheduler, like we have done for clustered workflows https://github.com/dapr/dapr/pull/10382

pkg/scheduler/server/internal/placement/loops/connections/connections.go
```diff
@@ -0,0 +1,322 @@
...
+// from every in-flight round. It returns the seqs of rounds whose phase may
+// now be complete; the caller must call advanceRounds with them.
+func (c *connections) removeStream(streamIDx uint64, err error) []uint64 {
+	conn, ok := c.streams[streamIDx]
+	if !ok {
+		return nil
+	}
+
+	delete(c.streams, streamIDx)
+	conn.loop.Close(&loops.StreamShutdown{Error: err})
```
JoshVanL: This can deadlock everything if the `Send` is blocked on the daprd. Instead we should pipe a context `cancel` for the `streamConn` to prevent the lock.

pkg/placement/internal/loops/namespaces/namespaces.go
```diff
@@ -82,6 +87,13 @@ func (n *namespaces) Handle(ctx context.Context, event loops.EventNamespace) err
 		return n.handleReportedHost(e)
 	case *loops.Shutdown:
 		return n.handleShutdown(e)
+	case *loops.StandDown:
+		return n.handleStandDown(e)
+	case *loops.StandUp:
+		n.standDown = nil
+		n.drainPending = 0
+	case *loops.DrainComplete:
```
JoshVanL: `StandUp` sets `n.standDown = nil` and `drainPending = 0` while disseminators may still be draining (the drain can take up to the 8s disseminate timeout). A `DrainComplete` arriving after this finds its disseminator still in the map and dereferences `n.standDown.Error` in `handleDrainComplete` (line 212): nil pointer panic inside the loop handler, crashing the placement process. Trigger: a WatchHosts response showing rollback (or a transient view during a scheduler rolling restart) lands mid-drain.

Needs a nil guard in `handleDrainComplete`, plus a decision on whether the drained disseminator should be closed or resurrected on stand-up.

JoshVanL (changes requested):

cicoyle:
/ok-to-perf actor_activation actor_double_activation actor_id_scale actor_reminder actor_timer actor_type_scale workflows scheduler-placement

cicoyle:
/ok-to-perf actor_activation actor_id_scale actor_reminder actor_timer actor_type_scale workflows scheduler-placement

pkg/actors/actors.go
```diff
@@ -154,35 +166,37 @@ type actors struct {
 // New create a new actors runtime with given config.
 func New(opts Options) Interface {
 	var disabled atomic.Pointer[error]
-	if len(opts.PlacementAddresses) == 0 ||
-		(len(opts.PlacementAddresses) == 1 && strings.TrimSpace(strings.Trim(opts.PlacementAddresses[0], `"'`)) == "") {
+	noPlacementAddresses := len(opts.PlacementAddresses) == 0 ||
+		(len(opts.PlacementAddresses) == 1 && strings.TrimSpace(strings.Trim(opts.PlacementAddresses[0], `"'`)) == "")
+	if noPlacementAddresses && !opts.SchedulerPlacementEnabled {
```
tthophan: The log message here is quite long. Consider shortening it while keeping the essential information.

pkg/actors/hashing/rendezvous/rendezvous.go
```diff
@@ -0,0 +1,119 @@
...
+		if s := t.score(i, key); s > best {
+			owner, best = t.hosts[i], s
+		}
+	}
+
+	return owner, true
+}
+
+// score is xxhash64("<host>||<key>") for hosts[i], resumed from the host's
+// precomputed prefix digest.
```
tthophan, line 84: The `Equal` method uses `len(t.Hosts())` which calls `t.hosts`. This is fine, but `t.hosts` is already accessible directly. Consider using `t.hosts` directly for consistency.

pkg/actors/internal/placement/connector/leader/leader.go
```diff
@@ -0,0 +1,153 @@
...
+		l.conn = nil
+	}
+	l.lock.Unlock()
+
+	for {
+		addr, unsupported, changed := l.leadership.Leader()
+		if unsupported {
+			return nil, ErrSchedulerPlacementUnsupported
+		}
+
```
tthophan: The `grpc.DialContext` call is marked with `//nolint:staticcheck`. Ensure this is necessary and that the context usage is correct.

tthophan (approved):

tthophan (approved):

tthophan (commented):
WIP PR migrating actor placement from the standalone placement service into the Scheduler. Adds v2 wire protocol (proto + scheduler leadership + ReportActorTypes stream), rendezvous hashing, server-side placement disseminator/connections/store, daprd-side v2 client + connector, Helm gate via global.scheduler.placement.enabled, and integration suites. Diff is large (80 files, +6640/-593) and titled "wip" with many new files; the patch is truncated in the excerpt so surrounding code in new files cannot all be verified inline.

### Additional findings

- **WARNING** `charts/dapr/charts/dapr_rbac/templates/placement.yaml:2` — All four placement RBAC objects (SA/Role/RoleBinding) are skipped when scheduler placement is on. Confirm the scheduler service account already grants the same permissions to host actors; otherwise sidecars will fail authorization.
- **INFO** `cmd/scheduler/options/options.go:121` — `placement-disseminate-coalesce-window` default 0 with comment "0 disseminates immediately". Doc says "Marks the sidecars are not given a placement address". Validate (or document) interaction with `--scheduler-placement-enabled` on injector: sidecars still need *some* hint where to dial; otherwise they fall back to nothing.
- **INFO** `pkg/actors/internal/placement/loops/disseminator/disseminator.go:148` — V2 routing inside Handle dispatches to handleOrderV2. Confirm v1 still functions identically (v2Rounds is unused for v1 path) and the `currentOperation` is no longer mutated by v2 path; v2 tracks per-seq round state in `v2Rounds`.
- **WARNING** `pkg/actors/internal/placement/loops/disseminator/disseminator.go:184` — v2 timeout handler keys by `timeout.Version` interpreted as `seq`. Confirm timeout.Version is set to seq in v2 emitter (not the legacy monotonic version). Mixing v1 and v2 timeouts will misroute.
- **WARNING** `pkg/actors/internal/placement/placement.go:1` — +172 lines in placement.go. Cannot review internal wiring of SchedulerPlacementEnabled plumbing from excerpt; flagging.
- **WARNING** `pkg/runtime/scheduler/internal/watchhosts/watchhosts.go:1` — +134 changes. Need to confirm: leader election excludes placement_enabled=false hosts; at-most-one leader invariant; resync on leadership loss; backwards-compat with old WatchHostsResponse (no leader/placement_enabled fields).
- **WARNING** `pkg/runtime/runtime.go:1` — +43 changes. Plumb SchedulerPlacementEnabled + SchedulerLeadership. Verify the wiring: when SchedulerPlacementEnabled=true, placement client must be constructed with the leader connector (not standalone) and reminders partition uses rendezvous tables.
- **WARNING** `pkg/scheduler/server/internal/cron/cron.go:1` — +16 lines. Confirm cron leadership wiring still functions when placement is enabled; verify placement leader is independent from cron leader (per PR description), or that they're consolidated intentionally.

charts/dapr/charts/dapr_placement/templates/dapr_placement_poddisruptionbudget.yaml
```diff
@@ -1,4 +1,4 @@
-{{- if and (eq .Values.global.ha.enabled true) (eq .Values.global.actors.enabled true) (eq .Values.global.actors.serviceName "placement") }}
+{{- if and (eq .Values.global.ha.enabled true) (eq .Values.global.actors.enabled true) (eq .Values.global.actors.serviceName "placement") (ne .Values.global.scheduler.placement.enabled true) }}
```
tthophan, line 1: Gate now suppresses PDB when `global.scheduler.placement.enabled=true`. Standalone placement may still exist transiently during mixed-version upgrades; consider also gating on a per-release feature/version check so an upgrade where only some schedulers serve placement doesn't strand actors.

charts/dapr/charts/dapr_placement/templates/dapr_placement_statefulset.yaml
```diff
@@ -1,4 +1,13 @@
-{{- if and (eq .Values.global.actors.enabled true) (eq .Values.global.actors.serviceName "placement") }}
+{{- if and (eq .Values.global.scheduler.placement.enabled true) (ne .Values.global.scheduler.enabled true) }}
```
tthophan, line 1: Standalone StatefulSet gated on `global.scheduler.placement.enabled != true`. If users flip the flag mid-upgrade, in-flight placements and stream sessions break. Document or block the transition, or handle overlapping lifetimes.

charts/dapr/charts/dapr_scheduler/templates/dapr_scheduler_statefulset.yaml
```diff
@@ -148,6 +148,7 @@ spec:
         - "--etcd-initial-cluster"
         - "dapr-scheduler-server-0=https://dapr-scheduler-server-0.dapr-scheduler-server.{{ .Release.Namespace }}.svc{{ .Values.global.dnsSuffix }}:2380,dapr-scheduler-server-1=https://dapr-scheduler-server-1.dapr-scheduler-server.{{ .Release.Namespace }}.svc{{ .Values.global.dnsSuffix }}:2380,dapr-scheduler-server-2=https://dapr-scheduler-server-2.dapr-scheduler-server.{{ .Release.Namespace }}.svc{{ .Values.global.dnsSuffix }}:2380"
         - "--etcd-embed={{ .Values.etcdEmbed }}"
+        - "--placement-enabled={{ .Values.global.scheduler.placement.enabled | toString }}"
```
tthophan, line 151: `--placement-enabled` is forwarded from values but the chart's values.yaml only sets `scheduler.placement.enabled` (default false). Confirm a default-on / upgrade-default path so existing users aren't silently kept on the old placement service.

dapr/proto/scheduler/v1/placement.proto
```diff
@@ -0,0 +1,119 @@
...
+  string app_id = 2;
+
+  // namespace is the namespace of the sidecar, validated against its SPIFFE
+  // identity.
+  string namespace = 3;
+
+  // actor_types are the actor types hosted by the sidecar. May be empty for a
+  // sidecar which performs actor lookups without hosting any types.
+  repeated string actor_types = 4;
+}
```
tthophan, line 56: HashAlgorithm is set by leader; receivers must reject unknown algorithms (enforced in inflight.Merge via HashAlgorithm_HASH_ALGORITHM_RENDEZVOUS check). Good. Consider documenting wire-compat policy when adding future algorithms: must round-trip identical owner for every (host,key) with the previous algorithm's table.

dapr/proto/scheduler/v1/scheduler.proto
```diff
@@ -135,6 +143,19 @@ message WatchJobsRequestInitial {
   // a key which groups triggers, and an optional concurrency_key for per-name
   // sub-limits within a group.
   repeated ConcurrencyLimit concurrency_limits = 5;
+
```
tthophan, line 146: `optional string actor_address = 6` in WatchJobsRequestInitial — `optional` in proto3 is non-default for field-optionality; downstream Connect-Go generated code reflects this. Confirm all clients (older daprd) still compile; `optional` keyword is supported on this proto3 syntax.

dapr/proto/scheduler/v1/scheduler.proto
```diff
@@ -289,6 +310,17 @@ message WatchHostsResponse {
 message Host {
```
tthophan, line 310: `Host.leader` and `Host.placement_enabled` semantics: leader is at most one; placement_enabled may be true on multiple hosts. Mixed-version clusters: only a subset advertise placement_enabled; leader election must exclude non-capable hosts (the leader connector already uses this). Verify no scheduler-side code path assigns leader=true to a placement_enabled=false host.

go.mod
```diff
@@ -9,6 +9,7 @@ require (
 	github.com/aavaz-ai/pii-scrubber v0.0.0-20220812094047-3fa450ab6973
 	github.com/argoproj/argo-rollouts v1.4.1
 	github.com/cenkalti/backoff/v4 v4.3.0
+	github.com/cespare/xxhash/v2 v2.3.0
```
tthophan, line 12: `cespare/xxhash/v2` promoted from indirect to direct require. Fine. Confirm vendoring/CI cache reflects this (transitive bump may invalidate build cache for downstream consumers).

pkg/actors/hashing/rendezvous/rendezvous.go
```diff
@@ -0,0 +1,119 @@
...
+	prefixes := make([]xxhash.Digest, len(sorted))
+	for i, host := range sorted {
+		prefixes[i].Reset()
+		//nolint:errcheck // Digest.WriteString never returns an error.
+		prefixes[i].WriteString(host)
+		//nolint:errcheck
+		prefixes[i].WriteString(separator)
+	}
+
+	return &Table{hosts: sorted, prefixes: prefixes}
```
tthophan, line 58: Lookup uses xxhash64 over `<host>||<key>`. Wire contract comment says "<host address>||<actor id>". Verify the scheduler leader computes scores with the exact same string and separator; any drift yields divergent ownership and actors routed to wrong hosts.

pkg/actors/hashing/rendezvous/rendezvous.go
```diff
@@ -0,0 +1,119 @@
...
+		}
+	}
+
+	return owner, true
+}
+
+// score is xxhash64("<host>||<key>") for hosts[i], resumed from the host's
+// precomputed prefix digest.
+func (t *Table) score(i int, key string) uint64 {
+	d := t.prefixes[i]
```
tthophan, line 86: Equal() compares both hosts and table-internal dedup. Tests cover Equal. Good. But the merge detector in inflight.Merge compares `existing.table.Equal(entry.table) && maps.Equal(existing.hosts, hosts)` — order of map iteration in proto is non-deterministic; ensure Entry decode yields deterministic map. Using slices.Compact on sorted addrs in New() already gives deterministic hosts ordering.

pkg/scheduler/server/internal/placement/placement.go
```diff
@@ -0,0 +1,202 @@
+/*
```
tthophan, line 1: New placement server entrypoint; truncated. Verify: authz on ReportActorTypes (authorizer package), namespace SPIFFE validation, graceful shutdown drains streams.

pkg/scheduler/server/internal/placement/authorizer/authorizer.go
```diff
@@ -0,0 +1,56 @@
+/*
```
tthophan, line 1: New authorizer for placement RPCs. Verify: validates app_id matches SPIFFE identity's namespace (per ActorHost.namespace comment), rejects mismatches, and short-circuits before allocation.

pkg/scheduler/server/internal/placement/loops/connections/connections.go
```diff
@@ -0,0 +1,334 @@
+/*
```
tthophan, line 1: +301 lines. Truncated; flagging for in-depth review. Specifically: per-namespace tracking, race-free host add/remove under concurrent streams, ordering of LOCK/UPDATE/UNLOCK emissions per actor type.

pkg/scheduler/server/internal/placement/loops/connections/rounds.go
```diff
@@ -0,0 +1,282 @@
+/*
```
tthophan, line 1: Round coalescing & timeout logic. Confirm coalesce window edge cases (timer cancelation, partial merges, key reset on leader change), and that the 8s default timeout (set in options) is sufficient for large clusters under slow networks.

pkg/scheduler/server/internal/placement/loops/store/store.go
```diff
@@ -0,0 +1,178 @@
+/*
```
tthophan, line 1: Placement state store. Verify: concurrent map access safety, version monotonicity, delete-on-empty-hosts behavior matches inflight.Merge contract, and namespace isolation.

pkg/scheduler/server/internal/placement/loops/stream/stream.go
```diff
@@ -0,0 +1,194 @@
+/*
```
tthophan, line 1: Server-side stream handler. Truncated; verify ack dispatch (seq lookup, ack ordering), error propagation (close on bad hash alg or version regression), and graceful close on shutdown.

pkg/actors/hashing/rendezvous/rendezvous_test.go
```diff
@@ -0,0 +1,199 @@
+/*
```
tthophan, line 1: Tests are thorough (lookup, equal, balance, minimal movement). Add: nil-vs-empty-table distinction in Equal (current code returns true on both nil — could hide bugs), and a wire-format compatibility test that pins the exact xxhash64 output for a known (host,key) pair.

charts/dapr/values.yaml
```diff
@@ -95,6 +95,27 @@ global:
   scheduler:
     # Enables scheduler functionality in the cluster
     enabled: true
+    placement:
```
tthophan, line 98: Document upgrade sequence and flag-rollout procedure. Currently silent on whether flipping `scheduler.placement.enabled` requires coordinated upgrade of schedulers + daprd + injector.

tthophan:
### Additional findings

- **WARNING** `cmd/injector/options/options.go:65` — `scheduler-placement-enabled` default false. With no placement address patched and the feature off, actor runtime is silently disabled. Injector must validate: if SchedulerPlacementEnabled && !SchedulerEnabled, fail boot rather than producing sidecars with no placement.
- **WARNING** `pkg/actors/actors.go:171` — Actors runtime now enabled without placement addresses when SchedulerPlacementEnabled is set. Confirm all subsequent code paths (hashing/lookup, reminders, reminders partition) handle the v2 rendezvous table in addition to v1 consistent hash. Also: warning message should mention both resolution paths.
- **WARNING** `pkg/actors/internal/placement/connector/leader/leader.go:96` — Connect() closes prior conn and redials in a loop waiting on `changed`. After successful dial, `watchLeader` closes the conn on leader change, forcing the placement client to reconnect — but Connect() is called by the client only on (re)connect. Verify the placement client's backoff/reconnect loop honors ErrSchedulerPlacementUnsupported and doesn't busy-loop forever in clusters where no scheduler serves placement.
- **INFO** `pkg/actors/internal/placement/connector/leader/leader.go:121` — watchLeader polls `Leadership.Leader()` and compares addr. On race where leader moves between samples, the connection closes. No `unsupported` handling for `leader addr != ""` path on shutdown — covered. But: leadership may report unsupported mid-connection; ensure conn.Close() in that branch does not deadlock with a concurrent DialContext.
- **WARNING** `pkg/actors/internal/placement/loops/disseminator/inflight/inflight.go:197` — Merge accepts `versions map[string]uint64` — partial: missing types are not version-bumped. Verify scheduler always includes a version for every changed actor type; otherwise two equal-content updates could collapse and miss notifying waiters.
- **WARNING** `pkg/actors/internal/placement/loops/disseminator/inflight/inflight.go:198` — Merge returns error on unknown hash algorithm or version regression; both must close the stream per proto comment. Verify the caller (v2 disseminator) actually closes the stream on this error and not just logs.
- **INFO** `pkg/actors/internal/placement/loops/disseminator/inflight/inflight.go:222` — `changed` accumulation: identical-content re-merge returns empty. Subsequent rounds may rely on `changed` to know which types were updated; verify that's the intended signal (lookups don't re-block when nothing changed).
- **INFO** `pkg/actors/internal/placement/loops/disseminator/inflight/inflight.go:260` — HasTables checks both v2 and v1 maps. A process must use only one family — coexistence is safe but indicates misconfig. Consider a strict-mode assert that v2Entries is empty when SchedulerPlacement=false.
- **INFO** `pkg/actors/internal/placement/loops/disseminator/inflight/inflight.go:365` — IsBlocked exposes internal isBlocked. Confirm no internal caller still reaches in via the unexported form.
- **WARNING** `pkg/actors/internal/placement/loops/disseminator/inflight/inflight.go:414` — resolve() prefers v2Entries over v1 hashTable.Entries. Confirm this precedence is intentional and documented — once v2 is installed for a type, v1 should never again serve that type (otherwise actors could bounce).
- **WARNING** `pkg/actors/internal/placement/loops/disseminator/order_v2.go:1` — File truncated in excerpt; cannot review logic. Flagging for reviewer to inspect handleOrderV2 / seq lifecycle / ack wiring in detail.
- **INFO** `pkg/actors/internal/placement/loops/stream/transport/v1/v1.go:1` — New transport abstraction. v1 path must remain bit-compatible with existing placement v1 protocol; confirm no proto change slipped into v1pb.
- **WARNING** `pkg/actors/internal/placement/loops/stream/transport/v2/v2.go:1` — Truncated. Needs in-depth review of: stream open message (ActorHost fields), ack semantics vs. seq, partial-merge handling, ordering guarantees, and reconnect/reset semantics on leader change.
- **WARNING** `pkg/runtime/scheduler/leadership/leadership.go:1` — New package — leadership tracking from WatchHosts. Verify thread-safety of Leader() under concurrent update; connector.leader relies on it. Also verify the `unsupported` signal is set only when WatchHosts reports no host with placement_enabled=true AND leader never set.
- **WARNING** `tests/integration/suite/actors/schedulerplacement/basic.go:1` — New integration suite. Verify it runs in CI (gated behind SchedulerPlacement feature flag) and covers: actor invocation, reminder routing, hash algorithm negotiation, multi-stream ordering.
- **INFO** `pkg/actors/internal/placement/loops/disseminator/inflight/inflight_merge_test.go:1` — Unit test for Merge. Coverage: install/partial-merge, no-op identical, empty-hosts removal, version regression + reset. Add: actor lookup via rendezvous for both pre- and post-merge states; concurrent Merge calls.

charts/dapr/charts/dapr_placement/templates/dapr_placement_poddisruptionbudget.yaml
cicoyle: [I have pulled out logic into multi part PRs. This logic is covered in PR part 5](https://github.com/dapr/dapr/pull/10416)

charts/dapr/charts/dapr_placement/templates/dapr_placement_statefulset.yaml
cicoyle: [This is handled in a separate PR here.](https://github.com/dapr/dapr/pull/10416)

pkg/scheduler/server/internal/placement/loops/namespaces/namespaces.go
```diff
@@ -0,0 +1,193 @@
...
+			Authorizer:         n.authorizer,
+			DisseminateTimeout: n.disseminateTimeout,
+			CoalesceWindow:     n.coalesceWindow,
+		})
+
+		n.wg.Go(func() {
+			if err := cl.Run(ctx); err != nil && ctx.Err() == nil {
+				log.Errorf("Error running placement connections loop for namespace %s: %v", ns, err)
+			}
+		})
```
JoshVanL, line 120: Shouldn't we be not be doing something with this error?

pkg/scheduler/server/internal/placement/loops/stream/stream.go
```diff
@@ -0,0 +1,194 @@
...
+	s.loop = loop.New[loops.EventStream](64).NewLoop(s)
+
+	s.wg.Go(func() {
+		err := s.recvLoop()
+		s.nsLoop.Enqueue(&loops.ConnCloseStream{
+			StreamIDx: s.idx,
+			Namespace: s.ns,
+			Error:     err,
+		})
+	})
```
JoshVanL, line 90: A stream can emit `ConnCloseStream` more than once: once per failed `Send` in `Handle` (line 105, the loop keeps processing queued sends after the first failure) and once from this goroutine when `removeStream` cancels the RPC. `namespaces.handleCloseStream` decrements its per-namespace counter once per event with no per-stream dedupe, so with the order `[Close{A} from Send, ConnAdd C, Close{A} from recv]` the counter goes 1 -> 0 -> 1 -> 0 and `handleCloseNamespace` closes the loop that now owns live stream C (I reproduced this 50/50 under `-race` with a harness test). A stale duplicate for idx 0 can likewise land in a recreated loop whose `streamIDx` restarted at 0 and kill the new stream.

Emitting exactly one close per stream fixes both:

```suggestion
      s.wg.Go(func() {
              s.closeStream(s.recvLoop())
      })
```

with the same call replacing the `Enqueue` at lines 105-109, and:

```go
// closeStream reports the stream closed to the namespace loop exactly once,
// however many send failures and the receive loop exit race to report it.
func (s *stream) closeStream(err error) {
      s.closeOnce.Do(func() {
              s.nsLoop.Enqueue(&loops.ConnCloseStream{
                      StreamIDx: s.idx,
                      Namespace: s.ns,
                      Error:     err,
              })
      })
}
```

plus a `closeOnce sync.Once` field on `stream`.

pkg/scheduler/server/internal/cron/leadership.go
```diff
@@ -18,48 +18,139 @@ import (
...
+		if !h.incapableWarned {
+			h.incapableWarned = true
+			log.Warn("A sidecar running an older Dapr version is connected while actor placement is served by the scheduler. Its actor APIs stall unless it can reach a placement service the control plane cannot detect, such as one under a custom service name or outside the cluster, which would place its actors as a second authority. Upgrade the sidecar, and remove any such placement service.")
+		}
+	} else {
+		h.incapableWarned = false
+	}
+
+	electedAddr := placementLeader(hosts)
+	leaderAddr := electedAddr
```
JoshVanL: This fires on every scheduler in every cluster, including ones started without `--placement-enabled`: `gateIncapable` isn't conditioned on any scheduler serving placement, and at this head no daprd sets `supports_scheduler_placement` yet, so every sidecar counts as incapable and the first `WatchJobs` connection logs a WARN telling operators to upgrade sidecars and remove their placement service. Gate it on an elected placement leader existing:

```suggestion
      electedAddr := placementLeader(hosts)

      // An old sidecar cannot take scheduler placement, so warn, but only when
      // a scheduler actually serves placement.
      if gateIncapable && electedAddr != "" {
              if !h.incapableWarned {
                      h.incapableWarned = true
                      log.Warn("A sidecar running an older Dapr version is connected while actor placement is served by the scheduler. Its actor APIs stall unless it can reach a placement service the codetect, such as one under a custom service name or outside the cluster, which would place its actors as a second authority. Upgrade the sidecar, and remove any such placement service.")
              }
      } else {
              h.incapableWarned = false
      }

      leaderAddr := electedAddr
```

The `placement_incapable_sidecars` gauge in `pool.go:128,145` has the same problem (permanently non-zero on placement-disabled clusters, and its description says to alert on it); it probably wants the same gate or a description change.

tests/integration/suite/scheduler/placement/disabled.go
```diff
@@ -0,0 +1,74 @@
...
+	require.NoError(t, rstream.Send(&schedulerv1pb.ReportActorTypesRequest{
+		Msg: &schedulerv1pb.ReportActorTypesRequest_Report{Report: &schedulerv1pb.ActorHost{
+			Address:    "127.0.0.1:40001",
+			AppId:      "myapp",
+			Namespace:  "default",
+			ActorTypes: []string{"myactortype"},
+		}},
+	}))
+	_, err = rstream.Recv()
+	require.Equal(t, codes.Unimplemented, status.Code(err))
```
JoshVanL, line 65:
```suggestion
      rstream, err := d.sched.Client(t, ctx).ReportActorTypes(ctx)
      require.NoError(t, err)
      _, err = rstream.Recv()
      require.Equal(t, codes.Unimplemented, status.Code(err))
```

JoshVanL (changes requested):
Please can we also include the following integration tests:

- Host leave and rebalance. Only join is exercised. No test closes a stream and asserts the survivors receive a round dropping that host, that a type whose last host left is removed from the tables, or that a stream closing mid-round lets the round complete 
- Dissemination timeout and coalesce window. WithPlacementDisseminateTimeout and WithPlacementDisseminateCoalesceWindow were added to the framework but nothing uses them. Missing: a sidecar that never acks is evicted after the timeout (stream closed, with what error), the round restarts, and the others converge; and that churn within the coalesce window collapses into one round.
- Multi-scheduler cluster. Every test runs one scheduler.
- Reminder trigger routing to the actor owner.  two WatchJobs streams for one app with actor_address set, placement tables installed, an actor reminder fires on the owner's stream rather than round-robin, and falls back to round-robin when any stream is addressless. Only TestGetByKey covers it at unit level.
- Mixed-version gate and latch. No IT connects a sidecar without SupportsSchedulerPlacement.
- Actor type re-report on an existing stream.
- Connection churn. A test that opens and closes many streams concurrently in one namespace (including killing a stream while the server is mid-Send) while asserting the survivors are never cancelled and keep receiving orders.
- Authorization under mTLS.
- Scheduler shutdown and restart with live placement streams: streams closed with Unavailable "placement is shutting down", and a reconnect after restart gets a fresh snapshot.

pkg/scheduler/server/internal/placement/loops/namespaces/namespaces.go
cicoyle: oh, yes. good catch

tests/integration/suite/scheduler/placement/restart.go
```diff
@@ -42,6 +43,10 @@ type restart struct {
 }
 
 func (r *restart) Setup(t *testing.T) []framework.Option {
+	if runtime.GOOS == "windows" {
+		t.Skip("Skipping test on windows which relies on unix process signals")
+	}
+
```
JoshVanL: os.SkipWindows(t)

JoshVanL (changes requested):

tests/integration/suite/scheduler/placement/authz.go
```diff
@@ -0,0 +1,117 @@
...
+
+	report := func(host *schedulerv1pb.ActorHost) error {
+		rstream, rerr := a.sched.ClientMTLS(t, ctx, "myapp").ReportActorTypes(ctx)
+		require.NoError(t, rerr)
+		require.NoError(t, rstream.Send(&schedulerv1pb.ReportActorTypesRequest{
+			Msg: &schedulerv1pb.ReportActorTypesRequest_Report{Report: host},
+		}))
+		_, rerr = rstream.Recv()
+		return rerr
+	}
```
JoshVanL, line 86: `report` is called from inside the `require.EventuallyWithT` callback at line 81, and testify runs that callback on its own goroutine, so these `require.NoError` calls invoke `t.FailNow` off the test goroutine, which the `testing` package does not allow. It also defeats the retry the `Eventually` is there for: until the gate elects the leader, `ReportActorTypes` returns `FailedPrecondition` before reading anything (`placement.go:162`), so a `Send` racing that termination returns `io.EOF` (grpc-go: `SendMsg` returns `io.EOF` on a server-aborted stream, the status is only available from `RecvMsg`). Same race class as the `disabled.go` fix. It passed in my 15 local runs only because leadership settles quickly.

Dial once on the test goroutine and return errors so `Eventually` can retry, while the rejection asserts still see the status from `Recv` (needs `errors` and `io` imports):

```suggestion
	client := a.sched.ClientMTLS(t, ctx, "myapp")
	report := func(host *schedulerv1pb.ActorHost) error {
		rstream, rerr := client.ReportActorTypes(ctx)
		if rerr != nil {
			return rerr
		}
		if rerr = rstream.Send(&schedulerv1pb.ReportActorTypesRequest{
			Msg: &schedulerv1pb.ReportActorTypesRequest_Report{Report: host},
		}); rerr != nil && !errors.Is(rerr, io.EOF) {
			return rerr
		}
		_, rerr = rstream.Recv()
		return rerr
	}
```

tests/integration/suite/scheduler/placement/coalesce.go
```diff
@@ -0,0 +1,174 @@
...
+// into one dissemination round: 2 hosts joining back to back land on the
+// observer in a single update, never one at a time.
+type coalesce struct {
+	sched *scheduler.Scheduler
+}
+
+func (o *coalesce) Setup(t *testing.T) []framework.Option {
+	o.sched = scheduler.New(t,
+		scheduler.WithPlacementEnabled(true),
+		scheduler.WithPlacementDisseminateCoalesceWindow(time.Second*8),
```
JoshVanL, line 46: This test passes unchanged with `WithPlacementDisseminateCoalesceWindow(0)`: I ran it that way and it passed in 1.6s instead of 9.9s. What batches C and D into one round is not the window but the lock: A stops reading in `ackOrdersUntil` right after acking round seq2's UPDATE and leaves that round's UNLOCK unread, so `typeA` is still locked when C and D join, and both are queued as pending by the lock (`rounds.go` pending-types path). When A resumes and acks the UNLOCK, `completeRound` sees pending types and starts one round with members {A,B,C,D}, whatever the window is.

To exercise the window itself: keep A acking on a goroutine for the whole test so every round completes before C joins; join C; join D while C's round is still in flight; then assert the observer's next `typeA` UPDATE carries 4 hosts. With window 0 that UPDATE carries 3 and the test fails, which is the signal this test is missing today.

tests/integration/suite/scheduler/placement/timeout.go
```diff
@@ -0,0 +1,235 @@
...
+// after the disseminate timeout: its stream is closed with DeadlineExceeded,
+// the aborted round restarts and the surviving hosts converge.
+type timeout struct {
+	sched *scheduler.Scheduler
+}
+
+func (o *timeout) Setup(t *testing.T) []framework.Option {
+	o.sched = scheduler.New(t,
+		scheduler.WithPlacementEnabled(true),
+		scheduler.WithPlacementDisseminateTimeout(time.Second),
```
JoshVanL: A's join round is started, and its eviction timer armed, at A's `ConnAdd`, but A only acks its snapshot LOCK inside `openAndReport`; the LOCK for the join round is not acked until the goroutine at line 145 starts, which is after B's entire `openAndReport` on the test goroutine (blocking dial, `NewStream`, `Send`, `Recv`, possibly a 100ms `Eventually` tick). On a loaded `-race` CI runner that can exceed one second, A is evicted with `DeadlineExceeded`, and the test fails at line 181 with "the acking survivor must not be evicted". 45 clean runs locally, so this is hardening against CI load rather than a reproduced failure. The principled fix is to start A's ack goroutine before joining B; the cheap one is more headroom:

```suggestion
		scheduler.WithPlacementDisseminateTimeout(time.Second*3),
```

tests/integration/suite/scheduler/placement/gate.go
```diff
@@ -0,0 +1,134 @@
...
+		defer stream.CloseSend()
+		resp, err := stream.Recv()
+		require.NoError(t, err)
+		for _, host := range resp.GetHosts() {
+			if host.GetLeader() {
+				return true
+			}
+		}
+		return false
+	}
```
JoshVanL, line 71: `leader()` uses `require.NoError` twice and is called from the `require.EventuallyWithT` callback at line 86, which testify runs on a separate goroutine, so a transient `WatchHosts` or `Recv` error calls `t.FailNow` off the test goroutine and aborts instead of retrying. `g.sched.Client(t, ctx)` inside it has the same problem (the framework's `Client` does `require.NoError` on the dial) and also leaks one connection per `Eventually` tick. Dial once and take an `assert.TestingT`, which both `*testing.T` and `*assert.CollectT` satisfy; callers become `leader(t)` at lines 76, 119 and 133 and `leader(c)` at line 86:

```suggestion
	client := g.sched.Client(t, ctx)
	leader := func(c assert.TestingT) bool {
		stream, err := client.WatchHosts(ctx, new(schedulerv1pb.WatchHostsRequest))
		if !assert.NoError(c, err) {
			return false
		}
		//nolint:errcheck
		defer stream.CloseSend()
		resp, err := stream.Recv()
		if !assert.NoError(c, err) {
			return false
		}
		for _, host := range resp.GetHosts() {
			if host.GetLeader() {
				return true
			}
		}
		return false
	}
```

tests/integration/suite/scheduler/placement/churn.go
```diff
@@ -0,0 +1,217 @@
...
+	// closing cleanly, so some die mid round and mid Send.
+	var wg sync.WaitGroup
+	for c := range 6 {
+		wg.Go(func() {
+			for i := range 4 {
+				// Bounded: a churner blocked in Recv otherwise waits out the
+				// disseminate timeout, stacking iterations toward the suite's
+				// per test kill.
+				cctx, cancel := context.WithTimeout(ctx, time.Second*2)
+				stream, err := h.sched.Client(t, cctx).ReportActorTypes(cctx)
```
JoshVanL: `Scheduler.Client` does `require.NoError` on a blocking `grpc.DialContext` and registers a `t.Cleanup` (framework `process/scheduler/scheduler.go:346-347`). Called from these churner goroutines with a 2s `cctx` during heavy stream churn, a dial that misses the window calls `t.FailNow` off the test goroutine, even though the `ReportActorTypes` error right below is handled with `continue`. The per-iteration dial also buys nothing: `cctx` only bounds the dial, the connection itself lives until test cleanup, and `cancel()` already tears the stream down mid-round exactly as before. Dial once on the test goroutine:

```suggestion
	client := h.sched.Client(t, ctx)
	var wg sync.WaitGroup
	for c := range 6 {
		wg.Go(func() {
			for i := range 4 {
				// Bounded: a churner blocked in Recv otherwise waits out the
				// disseminate timeout, stacking iterations toward the suite's
				// per test kill.
				cctx, cancel := context.WithTimeout(ctx, time.Second*2)
				stream, err := client.ReportActorTypes(cctx)
```

JoshVanL (changes requested):

tests/integration/suite/scheduler/placement/cluster.go
```diff
@@ -0,0 +1,143 @@
...
+		assert.Equal(c, schedulerv1pb.Operation_OPERATION_LOCK, order.GetOperation())
+	}, time.Second*20, time.Millisecond*100)
+
+	// A follower refuses report streams.
+	followerN := (leaderN + 1) % 3
+	fstream, err := l.cluster.ClientN(t, ctx, followerN).ReportActorTypes(ctx)
+	require.NoError(t, err)
+	_, err = fstream.Recv()
+	require.Equal(t, codes.FailedPrecondition, status.Code(err))
+}
```
JoshVanL, line 143: Leader failover is untested. This test elects a leader and checks that a follower refuses report streams, then stops.

The failover path is the one that matters most in production. A scheduler restart or a rolling upgrade removes the current leader while sidecars hold live placement streams to it. Three things must happen: those streams close with `FailedPrecondition` and the message about lost placement leadership, the remaining schedulers advertise a new leader on `WatchHosts`, and the new leader rebuilds its tables from the sidecars that reconnect.

Killing the leader here would cover all three. Call `Cleanup` on the leader scheduler, assert the open report stream closes with that code, wait for a new leader on `WatchHosts`, then report a host to it and assert the table comes back.

tests/integration/suite/scheduler/placement/cluster.go
```diff
@@ -0,0 +1,143 @@
...
+// report streams, and a follower refuses them with FailedPrecondition.
+type clustered struct {
+	cluster *cluster.Cluster
+}
+
+func (l *clustered) Setup(t *testing.T) []framework.Option {
+	l.cluster = cluster.New(t,
+		cluster.WithCount(3),
+		cluster.WithSchedulerOptions(scheduler.WithPlacementEnabled(true)),
+	)
```
JoshVanL, line 48: Every scheduler in this cluster has placement enabled, so the election never has to exclude anyone.

`placementLeader` picks the lowest address among hosts with `SchedulerPlacementEnabled` set. A cluster where only some schedulers run with `--placement-enabled` is what an in-progress rolling upgrade looks like. If the election ever picked a scheduler that does not serve placement, every sidecar would follow a leader that answers `Unimplemented`, and actors would stall cluster wide.

Enabling placement on only two of the three schedulers here would cover it. Assert the elected leader is one of the two, and that the third is never advertised as leader.

tests/integration/suite/scheduler/placement/namespaces.go
```diff
@@ -0,0 +1,140 @@
...
+	streamB := openAndReport(&schedulerv1pb.ActorHost{
+		Address:    "127.0.0.1:40002",
+		AppId:      "app-b",
+		Namespace:  "ns2",
+		ActorTypes: []string{"shared"},
+	})
+	updateB := ackOrdersUntilUpdate(streamB)
+	assert.Equal(t, "ns2", updateB.GetNamespace())
+	assert.ElementsMatch(t, []string{"127.0.0.1:40002"}, sharedHosts(updateB))
+}
```
JoshVanL: This test proves two namespaces stay isolated, but no namespace is ever emptied.

When a namespace's last stream closes, the connections loop is shut down and the namespace is deleted, along with its store, its per type versions, and its sequence counter. A sidecar reconnecting afterwards has to rebuild all of it. Nothing exercises that path, and a bug there would show up as a sidecar that connects but never receives a table.

Closing both streams here, then opening a new one in `ns1` and asserting it receives a table with only itself, would cover it.

JoshVanL (changes requested):

JoshVanL (approved):

PR #10365, opened 2026-08-18
perf(diagnostics): cache tag maps for hot-path metric records

JoshVanL:
Every hot-path metric record went through stats.RecordWithOptions, rebuilding the tag mutators and options struct per call. Cache one prebuilt recorder per measure at Init (direct meter.Record through a prebuilt tag map), so per-measure metric rules resolve exactly as before while the per-record allocation chain disappears. Applies to the workflow and activity operation/execution/latency records and the gRPC interceptor records.

### DIFF
--- a/pkg/diagnostics/utils/metrics_utils.go
+++ b/pkg/diagnostics/utils/metrics_utils.go
@@ -19,6 +19,7 @@ import (
 	"regexp"
 	"strings"
 	"sync"
+	"sync/atomic"
 
 	"go.opencensus.io/stats"
 	"go.opencensus.io/stats/view"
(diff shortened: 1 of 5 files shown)


pkg/diagnostics/utils/metrics_utils.go
```diff
@@ -105,6 +315,7 @@ func AddNewTagKey(views []*view.View, key *tag.Key) []*view.View {
 
 // CreateRulesMap generates a fast lookup map for metrics regex.
 func CreateRulesMap(rules []config.MetricsRule) error {
+	defer rulesGeneration.Add(1)
```
cicoyle: Shouldnt this be moved to just above the `return nil` line so we only add one if it doesnt err?

cicoyle (changes requested):

pkg/diagnostics/utils/metrics_utils.go
JoshVanL: Thanks, done :+1:

cicoyle (approved):
lgtm just a vuln issue. Im fine fixing in this PR or a separate one.

PR #10367, opened 2026-08-18
Binary storebuilding block support

WhitWaldo:
# Description

This PR represents the first pass at implementing the [binary store building block](https://github.com/dapr/proposals/pull/88) I proposed. I will follow up with PRs in dapr/components-contrib with corresponding implementations as well.

I still need to follow up with some tests, so this should be considered a WIP in the short-term, but I intend to complete it for integration and release alongside 1.19.

## Issue reference

## Checklist

- [ ] Code compiles correctly
- [ ] Created/updated tests
- [ ] Unit tests passing


dapr/proto/runtime/v1/binarystore.proto
```diff
@@ -0,0 +1,76 @@
...
+message SetBinaryFileRequestOptions {
+  // Name of the component. Required.
+  string component_name = 1 [json_name="componentName"];
+  // Name of the file to store. Required.
+  string file_name = 2 [json_name="fileName"];
+  // When false (POST semantics), the operation fails if a file with the same
+  // name already exists. When true (PUT semantics), any existing file is
+  // replaced. Defaults to false.
+  bool overwrite = 3 [json_name="overwrite"];
+}
```
JoshVanL, line 50: We should change the SetBinaryFileRequest to be a `oneof`, where the first stream request would be the initial request with the options, and the following request stream items would be the payloads. Same as we do for streaming subscriptions. https://github.com/dapr/dapr/blob/86ff68c47f757f52991979a056e03ced5ca9953d/dapr/proto/runtime/v1/pubsub.proto#L95

JoshVanL (changes requested):

tthophan (approved):

dapr/proto/runtime/v1/binarystore.proto
WhitWaldo: Addressed in recent commit

mikeee (commented):
https://github.com/dapr/dapr/blob/337bc91ca4e015ee201ea396ed768577734822ae/tests/integration/framework/binary/binary.go#L38

Bit nuanced how the tests are included/constructed - you'll need to add the fake binary store to the build tags for int tests to pass

nelson-parente (commented):
I reviewed e4150ec with dapr/components-contrib#4559 at 9b85810 next to it. The wiring matches the Conversation block file for file, the protos match dapr/proposals#88, and Josh's `oneof` ask landed in fb349397b. Start with the red CI, because it is one line.

**1. Every check fails on one stale `replace`.** The fork commit `01eb2796` is no longer on the `binary-state-store` branch. The branch was rewritten after Sep 4, and `binarystore/in-memory` only exists on the new line, so all 39 failures come from this:

```
cmd/daprd/components/binarystore_in-memory.go:21:2: no required module provides package github.com/dapr/components-contrib/binarystore/in-memory
```

I pointed the replace at the #4559 head locally. `daprd` builds with `-tags allcomponents` and the universal tests pass. Bump it now so the integration and e2e jobs report something real. Before merge the replace has to go: land #4559, pin contrib to a commit that includes it, run `make modtidy-all`. Copilot's thread on this line shows resolved only because the line moved.
https://github.com/dapr/dapr/blob/e4150ec6206a69b8ecd2991da047e37fba6d3c43/go.mod#L546-L548

**2. Any timeout policy breaks `Get` for every cloud provider.** `GetBinaryFileAlpha1` opens the stream inside the resiliency runner. With a timeout policy, the runner wraps the call in `context.WithTimeout` and cancels it when `component.Get` returns, which is before anyone reads the body. The S3 provider returns `resp.Body` from `GetObject(ctx)`, and Go's HTTP client uses that context for reading the body. So the first `Read` fails with `context canceled`, and Blob and GCS behave the same way. The tests miss it because in-memory and fake ignore `ctx` and no test sets a policy. Open the stream outside the runner with `policyDef.ComponentContext(ctx)`, the way `subscription.go` does for `Subscribe`. That also removes the missing disposer Copilot noted on a retried `Get`.
https://github.com/dapr/dapr/blob/e4150ec6206a69b8ecd2991da047e37fba6d3c43/pkg/api/universal/binarystore.go#L88-L96
https://github.com/dapr/dapr/blob/e4150ec6206a69b8ecd2991da047e37fba6d3c43/pkg/resiliency/policy.go#L143-L148
https://github.com/dapr/components-contrib/blob/9b858106d10c3b30341ae5e3083fe30e0e1bf333/binarystore/aws/s3/s3.go#L253-L261

**3. `Set` skipped the runner to avoid replaying a half-read stream. That part is right, but it dropped two other things.** It no longer calls `ComponentContext`, which is where daprd attaches the workload SPIFFE identity for component auth. The resiliency docs say every call that bypasses the runner must apply it. It also has no bound after the first frame. The 5-second first-chunk timeout covers only the header, so a client that sends one chunk and goes quiet holds the handler and the component call until transport keepalive gives up. Apply `ComponentContext`. Then add an idle deadline between frames rather than a total timeout, which also kills a healthy multi-gigabyte upload.
https://github.com/dapr/dapr/blob/e4150ec6206a69b8ecd2991da047e37fba6d3c43/pkg/api/universal/binarystore.go#L62-L67
https://github.com/dapr/dapr/blob/e4150ec6206a69b8ecd2991da047e37fba6d3c43/pkg/api/grpc/binarystore.go#L167-L189

**4. The proposal's YAML does not load.** The proposal documents `type: state.binarystore.<provider>`. The category matcher checks `state.` before `binarystore.`, so that type goes to the state store registry and fails with a state-store-not-found error. The code is right to use `binarystore.*`, which matches `conversation.*` and `crypto.*`. Update the proposal to match, and decide whether the HTTP path keeps its `state/` segment while the API is alpha. Set also returns 204 where the proposal says 200. 204 matches the state API, so the proposal is the thing to fix.
https://github.com/dapr/dapr/blob/e4150ec6206a69b8ecd2991da047e37fba6d3c43/pkg/runtime/processor/manager.go#L96-L114

**5. Set and Get spans never get their attributes.** They are streaming RPCs, and the stream interceptor passes `ss.Context()` to `spanAttributesMapFromGRPC`, never the request message. So the new `AppendSpanAttributes` methods for Set and Get never run, and those spans carry no `db.name`. Delete is unary and works. Set the attributes from the handler once the options frame arrives.
https://github.com/dapr/dapr/blob/e4150ec6206a69b8ecd2991da047e37fba6d3c43/pkg/diagnostics/grpc_tracing.go#L222-L226
https://github.com/dapr/dapr/blob/e4150ec6206a69b8ecd2991da047e37fba6d3c43/pkg/proto/runtime/v1/dapr_tracing.go#L83-L107

**6. The streaming code has no unit tests.** No test covers `pkg/api/grpc/binarystore.go` or `pkg/api/http/binarystore.go`, and every integration test sends exactly one chunk. The sequence check, the rejection of a late options frame, the first-chunk timeout, cancel mid-upload, empty files and multi-chunk files all go untested. The 1 MiB test in the universal package calls `SetBinaryFileAlpha1` with a `bytes.Reader` directly, so it never touches the framing. #9974 also added an inline-close test for its new categories. The same test for `CategoryBinaryStore` is cheap.

**7. Errors use the old `messages.APIError` style.** State, pubsub, and the search and vector blocks in #9974 use the `pkg/api/errors` kit builders, which give SDKs `ErrorInfo` reasons. Two blocks landing in the same release with two error styles is worth avoiding while the API is alpha.

**8. The Data Lake race ships in daprd too.** This PR's `go.mod` selects `azdatalake v1.6.0`, while #4559 builds and tests against `v1.1.0`. Both versions, and Azure SDK `main`, have the unsynchronized `actualSize +=` inside the upload goroutines that Josh found. The `Concurrency: 1` fix on #4559 is needed either way. Bump contrib to `v1.6.0` so conformance runs against the SDK daprd ships, and open an issue on Azure/azure-sdk-for-go. I found none.

**Housekeeping.**
- Commit c8d6fe4ae moves `nolint` comments in eight files that have nothing to do with the binary store. Drop it or split it out.
- The runtime only rejects an empty `file_name`. My #4559 review asks for `..` and empty segments to be rejected in `ObjectPath`. If that lands, the runtime needs nothing more.
- tthophan's approval predates the `oneof` rewrite and 24 commits. @JoshVanL, your only inline ask is addressed, so a re-review clears your block.
- No `BinaryStoreInvoked` metrics exist. #9974 has the same gap, and the description needs to say so.
- The description still says WIP, and there is no dapr/docs PR yet.

nelson-parente:
One more, found while reading the proposal against the code: HTTP uploads larger than 4 MiB fail by default.

`MaxBodySizeMiddleware` wraps every request body on the HTTP server, binary store PUT and POST included, in a `LimitReadCloser` sized by `--max-body-size`, which defaults to 4 MiB. `onBinaryStoreSet` hands `r.Body` straight to the component, so at 4 MiB the component reads `ErrStreamTooLarge` and `mapBinaryStoreError` turns it into a 500. gRPC is fine because every frame is one small chunk. The proposal's use case is files up to several gigabytes, so the binary store routes need their own limit or an exemption from the global one. `ErrStreamTooLarge` also needs to map to 413, so a client can tell a size limit from a store failure. One HTTP integration test with a body over 4 MiB covers both.
https://github.com/dapr/dapr/blob/e4150ec6206a69b8ecd2991da047e37fba6d3c43/pkg/api/http/server.go#L101-L105
https://github.com/dapr/dapr/blob/e4150ec6206a69b8ecd2991da047e37fba6d3c43/pkg/api/http/middlewares.go#L29-L36
https://github.com/dapr/dapr/blob/e4150ec6206a69b8ecd2991da047e37fba6d3c43/pkg/api/http/binarystore.go#L82-L99

WhitWaldo:
@nelson-parente This review is premature since, as you note, all of CI is red and I haven't finished addressing Josh's comments. Might you review it again next week after I've refactored dapr/dapr and wrapped up the already known work here?

PR #10368, opened 2026-08-18
workflow: add WorkflowsFastPath preview feature

JoshVanL:
Adds the workflow scheduler fast-path stack behind a single preview feature gate, disabled by default.

Wake-ups drive eagerly on the arming host instead of creating a per-event scheduler reminder. Durability moves to a per-instance repeating janitor reminder plus on-failure escalation to the durable per-event reminder; delayed starts keep their scheduler due time and the start reminder remains durable.

Activities certified by that janitor execute locally without their run-activity reminder: the dispatching orchestrator certifies the janitor backstop is armed, and the target host drives the execution in place of the reminder fire. Local drive retries back off with the shared decorrelated jitter, matching the randomized per-job intervals of the reminder failure policy they replace. Crash recovery moves to the janitor, which re-dispatches unresolved scheduled tasks within one period and escalates a task still unresolved after a re-dispatch to the durable reminder, restoring the non-fast-path recovery chain.

Sender-retried completion events fold straight into the next turn's single state commit instead of the durable inbox: the completion is held in memory, the turn persists it into history, and the sender is acked only after that commit, making the sender's retry the durability. External raised events keep the durable inbox path. An empty-inbox no-op reminder fire no longer deactivates a running instance while the gate is enabled: deactivation there could race a concurrent fold submit and flush its held completion into a sender retry.

Each leg removes scheduler job upsert/delete commits and trigger round trips from the workflow hot path. At-least-once execution is unchanged throughout. New local_wake, local_activity and completions_fold metrics expose drive outcomes, with the rare rescue series pre-registered at zero so absence is distinguishable from never-fired. Integration suites cover job accounting, restarts, stream loss, captive-fold recovery and lost re-dispatch escalation under fault injection.


pkg/actors/targets/workflow/activity/execute.go
```diff
@@ -52,7 +60,7 @@ func (a *activity) executeActivity(ctx context.Context, name string, invocation
 	}
 
 	key := inflight.Key(a.actorID, taskEvent)
-	call, owner := a.inflight.Acquire(key)
+	call, owner := a.claim(key, workflowID, taskEvent.GetEventId())
```
cicoyle: shall we re-check the staleness with a ticker while following in the select case below so the lock is released sooner bc it looks to only be re evaluated at claim time?

pkg/runtime/wfengine/backends/actors/actors.go
```diff
@@ -1008,16 +1016,30 @@ func (abe *Actors) callWithBackoff(ctx context.Context, fn func() error) error {
 		), ctx))
 }
 
-// WaitForActivityCompletion implements backend.Backend.
+// WaitForActivityCompletion implements backend.Backend. The registration is
+// mirrored into activityExecs so the activity target's stale-claim eviction
+// can see the work item as engine-held for the duration of the wait.
 func (abe *Actors) WaitForActivityCompletion(request *protos.ActivityRequest) func(context.Context) (*protos.ActivityResponse, error) {
-	return abe.pendingTasksBackend.WaitForActivityCompletion(request)
+	release := abe.activityExecs.add(activityExecutionKey(request.GetWorkflowInstance().GetInstanceId(), request.GetTaskId()))
```
cicoyle: I think we want `abe.activityExecs.add...` inside the returned func below so we dont leak any `adds` in case durabletask abandons the activity.

pkg/actors/targets/workflow/orchestrator/reminder.go
```diff
@@ -74,6 +77,131 @@ func (o *orchestrator) createRetentionReminder(ctx context.Context, name string,
...
+	o.janitorAsserted.Store(true)
+	return nil
+}
+
+// deleteJanitor removes the janitor reminder. NotFound is tolerated: the
+// janitor may never have been asserted (fast path never taken this
+// residency), or an older binary may already have swept it via
+// DeleteByActorID.
+func (o *orchestrator) deleteJanitor(ctx context.Context) {
+	if err := o.reminders.Delete(ctx, &actorapi.DeleteReminderRequest{
```
cicoyle, line 198: Shall we store false before deleting the reminder?
```suggestion
	o.janitorAsserted.Store(false)
	if err := o.reminders.Delete(ctx, &actorapi.DeleteReminderRequest{
```

pkg/runtime/wfengine/backends/actors/actors.go
```diff
@@ -243,6 +249,8 @@ func (abe *Actors) RegisterActors(ctx context.Context) error {
 		WorkflowAccessPolicies:          abe.workflowAccessPolicies,
 		Signer:                          abe.signer,
 		WorkflowsRemoteActivityReminder: abe.workflowsRemoteActivityReminder,
+		FastPath:                        abe.workflowsFastPath,
+		ExecutionHeld:                   abe.ActivityExecutionHeld,
```
cicoyle, line 253: Should `ExecutionHeld` be gated on the feature flag?
```
		ExecutionHeld:                   nil,
}

		if abe.workflowsFastPath {
				aopts.ExecutionHeld = 		abe.ActivityExecutionHeld
		}		
```

pkg/actors/targets/workflow/orchestrator/run.go
```diff
@@ -509,6 +591,12 @@ func (o *orchestrator) runWorkflow(ctx context.Context, reminder *actorapi.Remin
 			if err = o.deleteAllReminders(ctx); err != nil {
 				return todo.RunCompletedFalse, err
 			}
+		} else if o.janitorAsserted.Load() {
```
cicoyle:
```suggestion
		} else if o.fastPath || o.janitorAsserted.Load() {
```

cicoyle (changes requested):

pkg/runtime/wfengine/backends/actors/actors.go
JoshVanL: I don't think think so because the stale claim class of issue can arise outside of the fast path as well

cicoyle (approved):

PR #10372, opened 2026-08-19
perf(workflow): bounded actor residency, janitor hysteresis, scoped activity lock

JoshVanL:
Actor residency is bounded: an idle reaper walks the actor table and releases orchestrators that have taken no turn within the idle TTL, stamped via lastActive on every turn-lock acquisition, with lastProgress (durable commit time) distinguishing alive-and- progressing from merely-polled so a recreated actor reads as stalled and the durable backstops recover it. The reaper scan interval and idle TTL are resolved in the factory constructor (env-overridable for tests via a shared EnvDurationOr helper); the TTL defaults to half the janitor period, since the janitor fire re-stamps lastActive every period and a TTL at or above it would leave the reaper inert for exactly the instances holding memory. The deactivation worker pool and reaper exit on the engine context and are joined by the final HaltAll, so engine shutdown leaks no goroutines.

Escalation and the janitor gain pressure-aware hysteresis: a failed drive against an instance showing recent life is not escalated to a durable reminder, and the janitor skips re-dispatch checks for instances with fresh progress, so overload sheds to the next period instead of amplifying into scheduler churn.

The activity actor lock is scoped to the execution claim rather than the whole invocation: the app roundtrip and result delivery run outside it, with the inflight entry as the execution guard. The inflight call gains a resolving phase covering the window between the engine releasing its held registration and the result publish settling the call, so the stale-claim check can never evict a healthy execution mid-publish. Lock waits are recorded as a per-factory sampled lock_wait histogram splitting queueing from turn body.

### DIFF
--- a/pkg/actors/targets/workflow/orchestrator/factory.go
+++ b/pkg/actors/targets/workflow/orchestrator/factory.go
@@ -373,7 +373,7 @@ func (f *factory) deactivate(orchestrator *orchestrator) {
 
 	select {
 	case f.deactivateCh <- orchestrator:
-	default:
+	case <-f.deactivateCtx.Done():
 	}
 }
 
--- a/pkg/actors/targets/workflow/orchestrator/factory.go
+++ b/pkg/actors/targets/workflow/orchestrator/factory.go
@@ -18,6 +18,7 @@ import (
 	"errors"
 	"sync"
 	"sync/atomic"
+	"time"
 
 	workflowacl "github.com/dapr/dapr/pkg/acl/workflow"
 	"github.com/dapr/dapr/pkg/actors"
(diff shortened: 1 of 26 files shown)


pkg/actors/targets/workflow/orchestrator/factory.go
```diff
@@ -326,5 +370,35 @@ func (f *factory) deactivate(orchestrator *orchestrator) {
...
+		}
+		cutoff := time.Now().Add(-f.reaperIdleTTL).UnixNano()
+		f.table.Range(func(_, v any) bool {
+			o, ok := v.(*orchestrator)
+			if !ok {
+				return true
+			}
+			// driveRunning actors are mid-drive by definition; their lastActive is
+			// refreshed on the next lock acquisition.
+			if o.lastActive.Load() < cutoff && !o.driveRunning.Load() {
```
cicoyle: Should the `lastActive` time also be refreshed on the unlock? I wonder if there will be a case where there is a slow app roundtrip where the turn runs longer than the `reaperIdleTTL` and gets queued for deactivation.

cicoyle (changes requested):
mostly lgtm

pkg/actors/targets/workflow/orchestrator/factory.go
JoshVanL: Thank you- turns are now bracketed by an in-turn counter the reaper skips, and the unlock re-stamps `lastActive` so the idle clock starts at turn end

cicoyle (approved):
thx! lgtm

Issue #9914, opened 2026-05-08
Flaky tests: TestV1DirectMessagingEndpointsWithResiliency subtests failing intermittently


sicoyle:
## Description

The `TestV1DirectMessagingEndpointsWithResiliency` test in `pkg/api/http` is flaking with multiple subtest failures. Errors indicate `context canceled` rather than the expected resiliency behavior (retries, circuit breaker tripping, etc.), suggesting a timing/race condition in the test setup.

## Where it showed up

Showing up on PR #9731 (unrelated changes):
- PR: https://github.com/dapr/dapr/pull/9731
- Failing job: https://github.com/dapr/dapr/actions/runs/25569842345/job/75063083221?pr=9731

## Failing subtests

All under `pkg/api/http TestV1DirectMessagingEndpointsWithResiliency`:

- `Test_invoke_direct_message_retries_with_resiliency` — got HTTP 500 instead of 200; only 1 call made instead of 2
- `Test_invoke_direct_message_retries_on_unsuccessful_statuscode_with_resiliency` — got 500 instead of 450; 0 calls instead of 2; response body shows `ERR_DIRECT_INVOKE: failed to invoke, id: failingApp, err: context canceled`
- `Test_invoke_direct_message_fails_with_timeout` — 0 calls instead of 2; `superfluous response.WriteHeader call` warning at directmessaging.go:255
- `Test_invoke_direct_messages_fails_after_exhausting_retries` — 0 calls instead of 2; same `superfluous response.WriteHeader` warning
- `Test_invoke_direct_messages_can_trip_circuit_breaker` — 1 call instead of 5; never received "circuit breaker is open" error; got `context canceled` instead

## Sample failure output

```
=== FAIL: pkg/api/http TestV1DirectMessagingEndpointsWithResiliency/Test_invoke_direct_message_retries_on_unsuccessful_statuscode_with_resiliency
    directmessaging_test.go:973:
        Error: Not equal:
            expected: "failingKey2"
            actual  : "{\"errorCode\":\"ERR_DIRECT_INVOKE\",\"message\":\"failed to invoke, id: failingApp, err: context canceled\"}"

=== FAIL: pkg/api/http TestV1DirectMessagingEndpointsWithResiliency/Test_invoke_direct_messages_can_trip_circuit_breaker
    directmessaging_test.go:1015:
        Error: "{...err: context canceled\"}" does not contain "circuit breaker is open"
        Messages: Should have received a circuit breaker open error.
```

## Impact

[...]

PR #10373, opened 2026-08-20
fix: prevent sidecar panic on timeout cancellation during response streaming (#10371)

### LINKED ISSUE #10371: panic: runtime error after HTTP service invocation failed to complete with error: context canceled
Hi. I am only somewhat familiar with your code, and I am no Go expert, but I have been hitting an issue recently with sidecars crashing. I have had a couple of agents digging into it. Below is the Fable 5's diagnosis. Hopefully, it's more helpful than me just reporting the crash.

---

## In what area(s)?

/area runtime

## What version of Dapr?

1.18.3 (also reproduced on 1.18.2). The code involved is unchanged on `master`. By inspection it affects every release carrying #9638: v1.17.2 onward and all of 1.18.x.

## Expected Behavior

When an HTTP service invocation is cancelled while the response body is still being streamed to the local  application — because the application disconnected, or because a resiliency timeout elapsed — the sidecar should stop copying, log the cancellation (it already does), and survive:

```
HTTP service invocation failed to complete with error: context canceled
```

## Actual Behavior

The sidecar panics with a nil pointer dereference and the `daprd` container exits with code 2, taking every other in-flight invocation on that pod with it. The application container is unaffected and does not restart, so from the application's side this looks like unexplained 500s and circuit-breaker trips rather than a Dapr fault.

The warning and the panic are adjacent in the log (verbatim apart from the app id and pod name; Dapr 1.18.3, Go 1.26.6):

```
{"level":"warning","msg":"HTTP service invocation failed to complete with error: context canceled",
 "scope":"dapr.runtime.http","ver":"1.18.3"}
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x16983c]

goroutine 6228 [running]:
bufio.(*Writer).Flush(0xc550852a240)
	/opt/hostedtoolcache/go/1.26.6/x64/src/bufio/bufio.go:642 +0x3c
net/http.(*response).FlushError(0xc55080861e0)
	/opt/hostedtoolcache/go/1.26.6/x64/src/net/http/server.go:1718 +0x38
net/http.(*response).Flush(0xc5509473b28?)
	/opt/hostedtoolcache/go/1.26.6/x64/src/net/http/server.go:1711 +0x1c
github.com/dapr/dapr/pkg/responsewriter.(*responseWriter).Flush(0xc5509f65b80?)
	/home/runner/work/dapr/dapr/pkg/responsewriter/response_writer.go:150 +0x48
github.com/dapr/dapr/pkg/api/http.(*flushWriter).Write(0xc550ab2b4a0, {0xc5509976000?, 0xc5509473bc8?, 0x336dc?})
	/home/runner/work/dapr/dapr/pkg/api/http/directmessaging.go:478 +0x50
io.copyBuffer({0x70e86a0, 0xc550ab2b4a0}, {0x70e3b60, 0xc55087211a0}, {0x0, 0x0, 0x0})
	/opt/hostedtoolcache/go/1.26.6/x64/src/io/io.go:431 +0x1cc
io.Copy(...)
	/opt/hostedtoolcache/go/1.26.6/x64/src/io/io.go:388
github.com/dapr/dapr/pkg/api/http.(*api).onDirectMessage.func1({0x7182420?, 0xc5508148fc0?})
	/home/runner/work/dapr/dapr/pkg/api/http/directmessaging.go:264 +0xd2c
github.com/dapr/dapr/pkg/resiliency.NewRunnerWithOptions[...].func2.1.1()
	/home/runner/work/dapr/dapr/pkg/resiliency/policy.go:152 +0x40
created by github.com/dapr/dapr/pkg/resiliency.NewRunnerWithOptions[...].func2.1 in goroutine 6222
	/home/runner/work/dapr/dapr/pkg/resiliency/policy.go:151 +0x114
```

## Analysis

Checked against the `v1.18.3` sources and Go 1.26.6.

**1. The response copy runs on a goroutine the handler does not wait for.** When a resiliency timeout applies to the invocation (`def.t > 0`), `NewRunnerWithOptions` runs the operation on its own goroutine and, on `ctx.Done()`, returns `ctx.Err()` immediately without joining it (`pkg/resiliency/policy.go:146-192` — the `created by … policy.go:151 in goroutine 6222` line in the trace is that goroutine; 6222 is the handler). For `onDirectMessage` the operation is not just `Invoke`: it also streams the response body into the `http.ResponseWriter` — `io.Copy(dst, reader)` at `directmessaging.go:264`. So once the response has started (`success == true`), a cancellation makes the handler log the warning at line 313 and **return while the goroutine is still writing into `w`**. That is a use of the `ResponseWriter` after `ServeHTTP` has returned, which `net/http` forbids.

**2. Why it is a nil dereference at `addr=0x18`.** When the handler returns, `net/http`'s `finishRequest` calls `putBufioWriter(w.w)`, which does `bw.Reset(nil)` — the response's `*bufio.Writer` now has a nil underlying `io.Writer` — and returns it to a `sync.Pool`. Response chunks from the remote sidecar are at most `StreamBufferSize` (2 KiB), which fits the response writer's 2 KiB buffer, so the orphaned goroutine's next `fw.w.Write(p)` succeeds into the reset buffer (`n > 0`), and the unconditional `fw.f.Flush()` in `flushWriter.Write` reaches `bufio.(*Writer).Flush` → `b.wr.Write(b.buf[0:b.n])` on a nil interface. `bufio.go:642` in Go 1.26.6 is exactly that line, and `0x18` is the method slot in the itab.

**3. Why it is intermittent.** The timeout wrapper's `defer cancel()` also cancels the context `Invoke` used, which tears down the remote stream, so usually the copy's next `Read` fails before it can write again. If a chunk was already in hand at the moment of cancellation, the write wins and the process dies. Under sustained streaming that is not rare: in one 50-minute batch, 7 mid-stream cancellations across 4 sidecars produced 2 panics; over roughly a day of load the `daprd` restart counts on the 4 replicas reached 17, 17, 7 and 7 while the application containers stayed at 0.

**4. A worse variant is possible by reading `net/http`, though we have not observed it.** The reset `*bufio.Writer` is pooled. If another connection on the same server acquires it before the orphaned write lands, the bytes go into an unrelated client's response instead of crashing the process. The panic is the benign outcome.

**5. Preconditions.** A resiliency timeout must apply to the invocation — ours is a namespace-wide `DefaultAppTimeoutPolicy: 30s`. Without one the operation runs inline on the handler goroutine and cannot outlive it. Either trigger reaches the same code: the local application disconnecting/cancelling mid-body (`context canceled` — our case, an upstream cancellation propagated through the app's request handler into its outbound call), or the timeout itself elapsing mid-body (`context deadline exceeded`).

**6. Provenance.** The orphaned goroutine predates #9638: before it, `io.Copy(w, reader)` would buffer the stray ≤2 KiB chunk in the reset writer and never flush it (the next `Reset` discards it), so the contract violation was silent. #9638 ("Fix ServiceInvocation response stream buffering", master 2026-03-16; backported to `release-1.17` as #9643) added `flushWriter`, whose flush after every write is what turns the stray write into a SIGSEGV. First affected release is v1.17.2; all 1.18.x carry it; `pkg/api/http/directmessaging.go` and `pkg/resiliency/policy.go`
are unchanged on `master` since.

**7. Related.** The `superfluous response.WriteHeader call` at `directmessaging.go:255` in #9914 is the same orphaned goroutine touching `w` after the handler has moved on — in the unit test it lands on `WriteHeader`; in production it lands on `Write` + `Flush`.

1. Two HTTP applications, `app-a` invoking `app-b` through the HTTP API    (`POST http://localhost:3500/v1.0/invoke/app-b/method/…`).
2. A `Resiliency` whose timeout covers that call — a `DefaultAppTimeoutPolicy`, or a  `targets.apps.app-b` timeout.
3. `app-b` returns a response body of tens of megabytes, streamed at full speed.
4. Either:
   - **timeout variant (deterministic):** make the timeout shorter than the transfer (for example  `2s` against a body that takes 10s). `app-a`'s sidecar logs `HTTP service invocation failed   to complete with error: context deadline exceeded` and panics; or
   - **disconnect variant:** leave the timeout generous and have `app-a` close its    connection to its sidecar after the first bytes arrive. The sidecar logs `… context canceled`    and panics.
5. Observe the **caller's** (`app-a`'s) `daprd` exit with code 2 and the trace above. The callee's    sidecar and both application containers are unaffected.

## Possible Fix

The invariant to restore is that `onDirectMessage` cannot return while anything can still write to its `ResponseWriter`. A few shapes:

- Run the policy only around `Invoke` — the part that can legitimately be retried or timed out —  and stream the response synchronously on the handler goroutine. The comment at  `directmessaging.go:234` explains why the copy was put inside the policy function (returning  cancels the context `Invoke` used, which tears down the stream); that coupling would need  lifting so the response stream outlives the policy function.
- Or have the timeout wrapper join the operation goroutine after cancelling when the operation  has already started writing.
- At minimum, guard the writer with a mutex and a "handler returned" flag set before the handler  returns. A bare `ctx.Err()` check before `Flush` is not enough — it is racy, and it leaves the  pooled-writer hazard in (4) open.

## Release Note

RELEASE NOTE: **FIX** Sidecar no longer panics when an HTTP service invocation with a resiliency timeout is cancelled while the response body is streaming.

waterWang:
## Description

Fixes #10371 — the Dapr sidecar (daprd) panics with a nil pointer dereference when an HTTP service invocation with a resiliency timeout is cancelled while the response body is still streaming.

### Root cause

In `NewRunnerWithOptions` (pkg/resiliency/policy.go), when a timeout fires, the wrapper returns `ctx.Err()` immediately and **abandons the operation goroutine**. The operation can then outlive the caller: in the HTTP service invocation path, `onDirectMessage` streams the response body with `io.Copy(w, reader)` inside the policy function, so on cancellation the handler returns while the orphaned goroutine is still writing to the `http.ResponseWriter`. `net/http` then recycles the `*bufio.Writer` (returned to a `sync.Pool` after `Reset(nil)`), and the goroutine's next `Flush()` dereferences a nil writer — SIGSEGV, daprd exits code 2, killing every in-flight invocation on the pod. The flush-every-write behavior that turns the stray write into a crash was introduced by #9638, on top of the timeout contract from #7270.

### Fix

On timeout, **wait for the operation goroutine to finish before returning to the caller** (one of the options suggested in the issue). The operation shares the same now-cancelled context, so it observes the cancellation and returns promptly (the next `Read` on the stream fails). This closes the use-after-recycle window: the caller never returns while anything can still write to its resources.

### Changes

- `pkg/resiliency/policy.go`: the timeout branch now blocks on the `done` channel (which always receives a value, since the goroutine's send no longer races a timeout marker), disposes the result, and only then returns `ctx.Err()`.
- `pkg/resiliency/policy_test.go`: `TestPolicyTimeout` now asserts the returned error type (`context.DeadlineExceeded`) and that the operation goroutine has been joined (`called == true`) even on timeout.

### Related

- #9638 added the `flushWriter` with an unconditional `Flush()` after every write (first release v1.17.2)
- #9914 flaky direct-messaging resiliency tests — same orphaned-goroutine code path

## Release Note

RELEASE NOTE: **FIX** Sidecar no longer panics when an HTTP service invocation with a resiliency timeout is cancelled while the response body is streaming.

### DIFF
--- a/pkg/resiliency/policy.go
+++ b/pkg/resiliency/policy.go
@@ -150,39 +150,28 @@ func NewRunnerWithOptions[T any](ctx context.Context, def *PolicyDefinition, opt
 				done := make(chan doneCh[T], 1)
 				go func() {
 					rRes, rErr := operCopy(ctx)
-
-					// If the channel is full, it means we had a timeout
-					select {
-					case done <- doneCh[T]{rRes, rErr}:
-						// No timeout, all good
-					default:
-						// The operation has timed out
-						// Invoke the disposer if we have a non-zero return value
-						// Note that in case of timeouts we do not invoke the accumulator
-						if opts.Disposer != nil && !isZero(rRes) {
-							opts.Disposer(rRes)
-						}
-					}
+					done <- doneCh[T]{rRes, rErr}
 				}()
 
 				select {
 				case v := <-done:
 					return v.res, v.err
 				case <-ctx.Done():
-					// Because done has a capacity of 1, adding a message on the channel signals that there was a timeout
-					// However, the response may have arrived in the meanwhile, so we need to also check if something was added
-					select {
-					case done <- doneCh[T]{}:
-						// All good, nothing to do here
-					default:
-						// The response arrived at the same time as the context deadline, and the channel has a message
-						v := <-done
+					// The operation has timed out. Wait for the operation
+					// goroutine to finish before returning to the caller:
+					// the operation may still be writing to caller-owned
+					// resources (e.g. the http.ResponseWriter used to stream
+					// HTTP service invocation responses), and returning while
+					// it is still running lets the caller recycle those
+					// resources underneath it, causing a panic
+					// (dapr/dapr#10371).
+					v := <-done
 
-						// Invoke the disposer if the return value is non-zero
-						// Note that in case of timeouts we do not invoke the accumulator
-						if opts.Disposer != nil && !isZero(v.res) {
-							opts.Disposer(v.res)
-						}
+					// Invoke the disposer if the return value is non-zero.
+					// Note that in case of timeouts we do not invoke the
+					// accumulator.
+					if opts.Disposer != nil && !isZero(v.res) {
+						opts.Disposer(v.res)
 					}
 					if def.addTimeoutActivatedMetric != nil && timeoutMetricsActivated.CompareAndSwap(false, true) {
 						def.addTimeoutActivatedMetric()
--- a/pkg/resiliency/policy_test.go
+++ b/pkg/resiliency/policy_test.go
@@ -178,19 +178,19 @@ func TestPolicyTimeout(t *testing.T) {
 		name      string
 		sleepTime time.Duration
 		timeout   time.Duration
-		expected  bool
+		wantErr   bool
 	}{
 		{
 			name:      "Timeout expires",
 			sleepTime: time.Millisecond * 100,
 			timeout:   time.Millisecond * 10,
-			expected:  false,
+			wantErr:   true,
 		},
 		{
 			name:      "Timeout OK",
 			sleepTime: time.Millisecond * 10,
 			timeout:   time.Millisecond * 100,
-			expected:  true,
+			wantErr:   false,
 		},
 	}
 
@@ -209,9 +209,17 @@ func TestPolicyTimeout(t *testing.T) {
 				name: "timeout",
 				t:    test.timeout,
 			})
-			policy(fn)
+			_, err := policy(fn)
 
-			assert.Equal(t, test.expected, called.Load())
+			if test.wantErr {
+				assert.ErrorIs(t, err, context.DeadlineExceeded)
+			} else {
+				assert.NoError(t, err)
+			}
+			// The operation goroutine is always joined before the policy
+			// returns, even on timeout: this prevents the operation from
+			// outliving a caller that owns its resources (dapr/dapr#10371).
+			assert.True(t, called.Load())
 		})
 	}
 }


JoshVanL (changes requested):
Thanks @waterWang - please can we have integration tests covering this change, reproducing the panic https://github.com/dapr/dapr/tree/master/tests/integration/suite/daprd/resiliency

tthophan (approved):

PR #10374, opened 2026-08-20
feat(sentry): support jwt typing via optional typ header

sicoyle:
# Description

Adds an optional `Type` field to the Sentry JWT issuer's `Request`. When set, its value is written to the JWS `typ` protected header (RFC 7515); when empty, behavior is unchanged and jwx's default `typ` of `JWT` is used.

This enables *explicit typing* as recommended by RFC 8725 §3.11 (JWT Best Current Practices) and used by RFC 9068 (`at+jwt` for OAuth 2.0 access tokens). Setting a distinct media type lets a verifier require a specific `typ`, so a token minted for one purpose cannot be silently accepted in a context that expects another.

## Changes

- `Request` gains a `Type string` field (documented; optional).
- `Generate` sets the `typ` protected header via `jws.WithProtectedHeaders` only when `Type` is non-empty, otherwise leaves the existing default untouched.
- Unit test covering both cases: a custom `Type` is written to the `typ` header, and an empty `Type` still defaults to `JWT`.

## Backward compatibility

Fully backward compatible. Existing callers that don't set `Type` get the exact same output as before (`typ: JWT`).

## Issue reference

## Checklist

- [X] Code compiles correctly
- [X] Created/updated tests
- [X] Unit tests passing


pkg/sentry/server/ca/jwt/jwt.go
```diff
@@ -55,6 +56,12 @@ type Request struct {
 
 	// TTL is the time-to-live for the token in seconds
 	TTL time.Duration
+
+	// Type, when set, is written to the JWS "typ" header (RFC 7515) to enable
+	// explicit typing per RFC 8725 §3.11 — as RFC 9068 does with "at+jwt". A
+	// verifier can then require a specific media type, so a token minted for one
+	// purpose is not accepted in another. Empty leaves jwx's default "JWT".
+	Type string
```
JoshVanL: Use point to signal field is optional

```suggestion
	Type option string
```

JoshVanL (changes requested):
Please also add an integration for this addition- doesn't look like the option is exposed to a client (?)

sicoyle:
> Please also add an integration for this addition- doesn't look like the option is exposed to a client (?)

on it!

JoshVanL (approved):

PR #10378, opened 2026-08-20
fix: protect Scheduler.currentActorTypes with mutex to prevent data race

### LINKED ISSUE #10352: Data race on `Scheduler.currentActorTypes` can crash daprd
## In what area(s)?

> /area runtime

> /area placement

## What version of Dapr?

> v1.18.2

## Expected Behavior

`Scheduler.currentActorTypes` can be updated from the app-health path while the placement
disseminator reads it, without a data race and without crashing the sidecar.

## Actual Behavior

`Scheduler.currentActorTypes` (`*[]string`, `pkg/runtime/scheduler/scheduler.go:58`) is
accessed from two goroutines with no synchronization:

- `StartApp()` / `StopApp()` set it to `nil` (`scheduler.go:119`, `:126`) — called from the
  app-health probe goroutine (`pkg/runtime/runtime.go:992`, `:1000`).
- `ReloadActorTypes()` reads and derefs it (`scheduler.go:135`, `:139`) — called from the
  placement disseminator goroutine
  (`pkg/actors/internal/placement/loops/disseminator/order.go:159`).

`ReloadActorTypes` does a check-then-deref on that pointer:

```go
if s.currentActorTypes != nil && slices.Equal(*s.currentActorTypes, actorTypes) {
```

If `StartApp`/`StopApp` nils the pointer between the `!= nil` check and the
`*s.currentActorTypes` deref, the placement goroutine panics and daprd crashes. The
window is easiest to hit at startup, when the app reports healthy while placement is
disseminating actor types.

Add this test (`pkg/runtime/scheduler/scheduler_race_test.go`, same package, builds a
`Scheduler` directly — no network needed) which drives `ReloadActorTypes` from one
goroutine and `StartApp`/`StopApp` from four others:

```go
package scheduler

import (
	"sync"
	"testing"

	"github.com/dapr/dapr/pkg/runtime/scheduler/internal/loops/connector"
)

func TestCurrentActorTypesRace(t *testing.T) {
	s := &Scheduler{connector: connector.New(connector.Options{})}

	const iters = 200000

	var wg sync.WaitGroup

	// placement disseminator goroutine: reads/dereferences currentActorTypes
	wg.Add(1)
	go func() {
		defer wg.Done()
		for i := 0; i < iters; i++ {
			s.ReloadActorTypes([]string{"actorA", "actorB"})
		}
	}()

	// app-health callback goroutines: write nil into currentActorTypes
	for g := 0; g < 4; g++ {
		wg.Add(1)
		go func(healthy bool) {
			defer wg.Done()
			for i := 0; i < iters; i++ {
				if healthy {
					s.StartApp()
				} else {
					s.StopApp()
				}
			}
		}(g%2 == 0)
	}

	wg.Wait()
}
```

Run with the race detector:

```sh
go test -race -run TestCurrentActorTypesRace -count=1 ./pkg/runtime/scheduler/
```

Result — fails:

```
==================
WARNING: DATA RACE
Read at 0x00c00054e7f8 by goroutine 33:
  scheduler.(*Scheduler).ReloadActorTypes()
      pkg/runtime/scheduler/scheduler.go:135
Previous write at 0x00c00054e7f8 by goroutine 36:
  scheduler.(*Scheduler).StartApp()
      pkg/runtime/scheduler/scheduler.go:119
==================
FAIL	github.com/dapr/dapr/pkg/runtime/scheduler	0.180s
FAIL
```

The full report (trimmed above) flags the same field racing across all four lines:
writes at `119`/`126`, read at `135`, write at `139`.

Without `-race` it passes:

```sh
go test -run TestCurrentActorTypesRace -count=1 ./pkg/runtime/scheduler/
```

```
ok  	github.com/dapr/dapr/pkg/runtime/scheduler	3.055s
```

It passes because the crash is timing-dependent — the panic only fires when a write lands
in the few instructions between the check and the deref. The race detector, not the panic,
is the evidence: the unsynchronized access is guaranteed by the two call paths and
confirmed by TSAN; the nil-deref is what it can turn into in production.

## Release Note

RELEASE NOTE: **FIX** Data race on `Scheduler.currentActorTypes` between app-health and placement dissemination can crash daprd.

---

I'd like to work on this. I'll open a PR with the fix (guard `currentActorTypes` with a mutex, or `atomic.Pointer[[]string]`) and keep the race test as a regression test.

waterWang:
## Description

Fixes a data race on `Scheduler.currentActorTypes` that can crash daprd.

`StartApp()` / `StopApp()` (called from the app-health probe goroutine) set `currentActorTypes` to `nil` without synchronization. `ReloadActorTypes()` (called from the placement disseminator goroutine) checks `currentActorTypes != nil` then dereferences it. If the write between the check and deref, the placement goroutine panics.

## Fix

Adds a `sync.Mutex` to the `Scheduler` struct and protects all accesses to `currentActorTypes` in `StartApp()`, `StopApp()`, and `ReloadActorTypes()`.

## Test

Includes a race-detector test (`TestCurrentActorTypesRace`) that drives concurrent `ReloadActorTypes` + `StartApp`/`StopApp` calls. Before the fix, this test fails with `-race`.

Fixes #10352

### DIFF
--- a/pkg/runtime/scheduler/scheduler_race_test.go
+++ b/pkg/runtime/scheduler/scheduler_race_test.go
@@ -1,0 +1,44 @@
+package scheduler
+
+import (
+	"sync"
+	"testing"
+
+	"github.com/dapr/dapr/pkg/runtime/scheduler/internal/loops/connector"
+)
+
+func TestCurrentActorTypesRace(t *testing.T) {
+	t.Parallel()
+
+	s := &Scheduler{connector: connector.New(connector.Options{})}
+
+	const iters = 200000
+
+	var wg sync.WaitGroup
+
+	// placement disseminator goroutine: reads/dereferences currentActorTypes
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		for i := 0; i < iters; i++ {
+			s.ReloadActorTypes([]string{"actorA", "actorB"})
+		}
+	}()
+
+	// app-health callback goroutines: write nil into currentActorTypes
+	for g := 0; g < 4; g++ {
+		wg.Add(1)
+		go func(healthy bool) {
+			defer wg.Done()
+			for i := 0; i < iters; i++ {
+				if healthy {
+					s.StartApp()
+				} else {
+					s.StopApp()
+				}
+			}
+		}(g%2 == 0)
+	}
+
+	wg.Wait()
+}


JoshVanL (approved):

JoshVanL:
@waterWang golangci-lint failing

PR #10380, opened 2026-08-20
actors: delete timers on ownership loss

JoshVanL:
In-memory actor timers previously survived dissemination events that moved their actor to another host: the old host kept firing them forever and forwarded each fire to the new host over CallActorReminder. Re-registering the same timer on the new host created a second independent copy, double invoking the actor. This contradicts the documented contract that timers are stopped on deactivation and only fire while the actor is active.

Timers are now bound to the host that owns the actor:

- The timer storage gains DeleteFunc, and the actor table sweeps it before halting actors in HaltAll, HaltNonHosted, and UnRegisterActorTypes. A rebalance, placement disconnect, hosting suspension, or type unregistration deletes the timers of every actor this host no longer owns, including actors that were idle-deactivated or never activated locally. Idle-timeout deactivation is unchanged and keeps timers, so a timer with a period longer than the idle timeout still reactivates its actor.

- The router drops a timer fire whose placement lookup resolves to a remote host instead of forwarding it, closing the window for a fire already in flight when ownership moves. Reminder forwarding is unchanged.

- Timer registration and unregistration now verify ownership via a placement lookup and are rejected with ERR_ACTOR_TIMER_NOT_OWNED (HTTP 403, gRPC PermissionDenied) on a sidecar that does not own the actor, matching the owner-only semantics of the actor state API. Previously a non-owning sidecar accepted the registration into its own store, where no lifecycle event could ever clean it up.


cicoyle (dismissed):

cicoyle:
lgtm, but CI is pretty red. plz ensure ci failures are not from this PR 🙏🏻

pkg/actors/internal/timers/inmemory/inmemory.go
m3nax, line 282: `i.router.CallReminder(ctx, reminder)` return nil also for dropped timers and the evaluation of `err == nil` return true (aka success = true) this cause the increment of **ActorTimerFired** metrics when a dropped timer is called.

pkg/actors/router/router.go
``` GOLANG
func (r *router) callReminder(ctx context.Context, req *api.Reminder) error {
--- code ---
        // Ownership moved while the fire was in flight and the timer is being
		// deleted with it.
		if req.IsTimer {
			log.Debugf("Dropping timer %s: actor is no longer hosted on this instance", req.Key())
			return nil
		}
--- code ---
}
```

In this case I think a dedicated metric should be incremented to record the invocation of a dropped timer, and the timer should be removed from the list of timers managed by that specific daprd instance.

m3nax, line 334: The placement claim guarding this registration may have been force-cancelled by the dissemination drain timeout (passed with`_ context.Context`).

``` golang
	if err := ctx.Err(); err != nil {
		return err
	}
```

I think we need to add a check here to handle cancellation.

m3nax (commented):
Is rejecting (for PUT and DELETE timer methods) with 403 preferred over forwarding the registration to the owning host? Actor invocation is transparently routed today, so an app calling its local sidecar will now get a 403 where it previously succeeded.

This changes the behaviour of these methods when called from a daprd that doesn't host the actor. Should this be documented and flagged as a breaking change?

pkg/actors/timers/timers.go
```diff
@@ -61,9 +68,42 @@ func (t *timers) Create(ctx context.Context, req *api.CreateTimerRequest) error
...
 	reminder.IsTimer = true
 
-	return t.storage.Create(ctx, reminder)
+	cctx, cancel, err := t.claimLocal(ctx, req.ActorType, req.ActorID)
+	if err != nil {
+		return err
+	}
+	defer cancel(nil)
+
+	return t.storage.Create(cctx, reminder)
```
mikeee, line 77: There's a race here between the type check and the inmemory write. `SuspendHsoting` or `UnRegisterActorTypes` can run and clear the timers, but a create that passed before the cleanup can insert a new timer after the cleanup. Sure it's a small window but this leftover timer could continue to fire against the new actor.

mikeee (changes requested):

pkg/actors/internal/timers/inmemory/inmemory.go
JoshVanL: :+1:

JoshVanL: :+1:

pkg/actors/timers/timers.go
JoshVanL: During a rebalance, dissemination cancels in-flight placement claims before it sweeps timers. Create checks the claim's ctx.Err() while holding the same queueLock the sweep uses. So a late insert is either rejected, or it lands before the sweep and the sweep deletes it.

For SuspendHosting and UnRegisterActorTypes, a timer can slip in after the sweep, but it is harmless. Every fire re-checks ownership in the router. Locally, GetOrCreate fails because the type is gone. Once the new placement table arrives, the first fire is dropped, the timer deletes itself, and `timers_dropped_total` records it. Both operations also broadcast a type change that triggers a dissemination, and that sweep removes the timer anyway.

mikeee (approved):

cicoyle (approved):

m3nax:
Hi @cicoyle, the `automerge` label didn't work.

image

Yesterday I opened a PR to try to address the problem. https://github.com/dapr/.github/pull/15

PR #10394, opened 2026-08-21
healthz: honour --healthz-listen-address in the healthz server

JoshVanL:
Every control-plane binary (placement, scheduler, sentry, injector, operator) parses --healthz-listen-address but never wires it anywhere: the shared healthz server hardcodes a wildcard bind, so the flag has been dead since its introduction. On shared CI runners a wildcard bind races other processes for the port across every interface and address family, and a healthz port squatted on another interface (or on IPv6, as on the macOS runners) kills the process at startup with "bind: address already in use", failing whole integration jobs on tests that never got a healthy placement or scheduler.

Add ListenAddress to the healthz server options, bind "<address>:<port>" with the empty default preserving today's wildcard behavior, and pass the parsed flag through in all five apps. The integration framework already sets the flag to 127.0.0.1, so control plane healthz servers under test now bind loopback only and stop colliding with foreign sockets. A unit test pins the loopback bind and that /healthz serves on it.

### DIFF
--- a/cmd/operator/app/app.go
+++ b/cmd/operator/app/app.go
@@ -79,9 +79,10 @@ func Run() {
 		metricsExporter.Start,
 		op.Start,
 		server.New(server.Options{
-			Log:     log,
-			Port:    opts.HealthzPort,
-			Healthz: healthz,
+			Log:           log,
+			ListenAddress: opts.HealthzListenAddress,
+			Port:          opts.HealthzPort,
+			Healthz:       healthz,
 		}).Start,
 	).Run(ctx)
 	if err != nil {
--- a/cmd/injector/app/app.go
+++ b/cmd/injector/app/app.go
@@ -143,9 +143,10 @@ func Run() {
 		metricsExporter.Start,
 		secProvider.Run,
 		healthzserver.New(healthzserver.Options{
-			Log:     log,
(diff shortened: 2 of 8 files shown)


cmd/operator/app/app.go
```diff
@@ -79,9 +79,10 @@ func Run() {
 		metricsExporter.Start,
 		op.Start,
 		server.New(server.Options{
-			Log:     log,
-			Port:    opts.HealthzPort,
-			Healthz: healthz,
+			Log:           log,
+			ListenAddress: opts.HealthzListenAddress,
```
cicoyle, line 83: I noticed we already have operator options of `HealthzListenAddress` [here](https://github.com/dapr/dapr/blob/master/pkg/operator/operator.go#L82) that is never used. Since you are wiring in the use of `opts.HealthzListenAddress` can you rm the unused field?

cicoyle (changes requested):
mostly lgtm, but one comment on something I noticed while looking at the service options

cmd/operator/app/app.go
JoshVanL: Thanks, removed these!

cicoyle (approved):

PR #10399, opened 2026-08-24
Retry workflow reminder creates on all transient scheduler errors

JoshVanL:
A workflow create commits actor state (status PENDING, inbox row) and then registers its start reminder as a separate scheduler call. The bounded reminder-create retry only retried Unavailable and DeadlineExceeded; every other error was treated as permanent and dropped after a single attempt. When a create landed while a scheduler replica's cron was shutting down (a single pod restart or rolling upgrade is enough), ScheduleJob failed with the plain error "cron is closed", which crosses the wire as gRPC code Unknown. The one call that could wake the workflow was dropped: the instance stayed PENDING forever with no reminder, never transitioned to FAILED, and was invisible to error-based monitoring.

Invert the classifier: CreateReminderWithRetry now retries every error except clearly-permanent request errors, sharing the isPermanentCreateError denylist (InvalidArgument, PermissionDenied, Unauthenticated, FailedPrecondition, Unimplemented, NotFound, and local non-status errors) with CreateReminderWithRetryForever. No allowlist of codes can anticipate every transient failure: the scheduler surfaces both cron shutdown and etcd errors as Unknown. The retry stays bounded, so a genuinely permanent unlisted error still surfaces within the existing budget. This covers every bounded create site at once: start, timer, retention, terminate, janitor, and activity reminders.

Also include the actor type and ID on the reminder scheduling error log, so a dropped reminder can be attributed to its workflow instance.


cicoyle (approved):
lgtm

cicoyle:
changes look good to go, but CI is a little 🔴

PR #10403, opened 2026-08-25
workflow: fix spurious failures under history signing

JoshVanL:
Force purge failed for every signed workflow because purgeWorkflowForce loaded state without the runtime's Signer and Namespace, tripping the loader's tamper protection; the call site now passes all five state.Options fields.

Completions that no longer correspond to a scheduling event in signed history (a straggler after ContinueAsNew, a save rolled back by a transient store fault, or a collision with a task id the new generation reused) were tombstoned as tampering; verification now returns the signing.ErrUnknownTaskScheduledID sentinel when a genuinely signed, correctly bound attestation merely fails to match the recorded invocation's io digest or activity name, and the orchestrator drops the event unpersisted, mirroring the unsigned path, while bad signatures, untrusted certificates, wrong bindings, and status mismatches still tombstone.

The drop is terminal for the sender, with the child's completion dispatch treating ErrInstanceNotFound as delivered instead of retrying forever, so every verification failure is first corroborated against freshly loaded durable state and a stale cache leads to a retry rather than a lost completion or a condemned workflow.

### DIFF
--- a/pkg/actors/targets/workflow/orchestrator/add.go
+++ b/pkg/actors/targets/workflow/orchestrator/add.go
@@ -15,8 +15,12 @@ package orchestrator
 
 import (
 	"context"
+	"errors"
+
+	"google.golang.org/protobuf/proto"
 
 	"github.com/dapr/dapr/pkg/actors/targets/workflow/orchestrator/dedup"
+	"github.com/dapr/dapr/pkg/actors/targets/workflow/orchestrator/signing"
 	wfenginestate "github.com/dapr/dapr/pkg/runtime/wfengine/state"
 	"github.com/dapr/durabletask-go/api"
 	"github.com/dapr/durabletask-go/backend"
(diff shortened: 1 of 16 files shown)


javier-aliaga (dismissed):

pkg/actors/targets/workflow/orchestrator/add.go
```diff
@@ -152,15 +155,49 @@ func (o *orchestrator) verifyAndAbsorbAttestation(ctx context.Context, state *wf
...
 		AppID:             o.appID,
 		Namespace:         o.namespace,
 		WorkflowActorType: o.actorType,
 		ActivityActorType: o.activityActorType,
 		Signer:            o.signer,
 	}
-	if _, _, terr := o.tombstoneTamperedState(ctx, opts, state, verr); terr != nil {
+	fresh, lerr := wfenginestate.LoadWorkflowState(ctx, o.actorState, o.actorID, opts)
+	if lerr != nil {
+		return wferrors.NewRecoverable(fmt.Errorf("failed to reload state to classify attestation failure (%s): %w", verr, lerr))
```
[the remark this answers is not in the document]
cicoyle: I don't think we want to recover from a `VerificationError` - I think we want to check the err here first before returning a recoverable err.

pkg/runtime/wfengine/backends/actors/actors.go
```diff
@@ -1198,8 +1198,10 @@ func (abe *Actors) purgeWorkflowForce(ctx context.Context, id api.InstanceID) er
 
 	s, err := state.LoadWorkflowState(ctx, astate, id.String(), state.Options{
```
cicoyle, line 1200: `LoadWorkflowState` can return a `VerificationError`, so I think we want to check that below and maybe continue to purge or at least log the err?

cicoyle (changes requested):

pkg/runtime/wfengine/backends/actors/actors.go
JoshVanL: Thanks- fixed

cicoyle (approved):

PR #10407, opened 2026-08-25
workflow: prevent activity double-execution across placement handoff

JoshVanL:
Under WorkflowsFastPath the exactly-once-in-flight claim for an activity execution is host-local, so when placement moves the actor mid-execution the janitor re-dispatch or escalated run-activity reminder fires on the new owner and runs the body a second time while the old host is still executing. Result dedup converges, but the body observably runs twice.

The fix is a durable execution-claim record (a lease) under the activity actor's state key, in the new activity/claim subpackage, written only around the handoff: when HaltNonHosted deactivates an actor with an unsettled claim, a guard writes the record and heartbeats it every janitor period. A clean finish marks it Completed (retained for the inflight cache TTL), a failed finish deletes it, and a dead host goes stale after two periods. HaltAll spawns no guards; at shutdown the execution dies with the process. The steady-state fast path never touches the record.

Recovery arrivals on the new owner (janitor re-dispatches, marked via Execute metadata, and scheduler-fired run-activity reminders) gate on the record: live defers recoverably, Completed acks without executing, stale or missing proceeds, and a stale record from another scheduling is reaped on read. A local inflight entry acks the re-dispatch only while the stale-claim oracle reads it live; a stranded entry falls through so stale eviction re-executes (a blanket ack livelocked the rescue chain by also swallowing the escalation). The drive-failure escalation is suppressed while the aborted drive's claim is live locally. Reminder creates clamp a stale dueTime to now, since the scheduler paces failure retries from the scheduled time and a past dueTime replayed the whole elapsed backlog as a burst on every deferred trigger.

Contract: at most one live body per activity across hosts during handoff; at-least-once preserved (a failed record write degrades to today's behavior, a dead guard host is reclaimed after the grace).

### DIFF
--- a/pkg/actors/targets/workflow/activity/claim/check.go
+++ b/pkg/actors/targets/workflow/activity/claim/check.go
@@ -1,0 +1,105 @@
+/*
+Copyright 2026 The Dapr Authors
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package claim
+
+import (
+	"context"
+	"time"
+)
+
+// Check classifies the record for a fresh-owner recovery arrival about to
+// execute taskKey.
+func (g *Guards) Check(ctx context.Context, actorID, taskKey string) (Outcome, error) {
+	rec, etag, err := g.read(ctx, actorID)
+	if err != nil {
+		return Defer, err
+	}
+	if rec == nil {
+		g.forget(actorID)
+		return Proceed, nil
+	}
+	stale := g.observeStale(actorID, rec)
+	if rec.TaskKey != taskKey {
+		// Another scheduling's record: ignore it, and reap it once its
+		// guard host reads dead so an old run's row cannot leak forever.
+		if stale && g.delete(ctx, actorID, etag) != nil {
+			// The row changed under the read (a guard heartbeat won the
+			// race): defer so the retry re-observes it. Not an error, a
+			// classification.
+			return Defer, nil //nolint:nilerr
+		}
+		if stale {
+			g.forget(actorID)
+		}
+		return Proceed, nil
+	}
+	if rec.Completed {
+		g.forget(actorID)
+		return Completed, nil
+	}
+	if !stale {
+		return Defer, nil
+	}
+	// The guarding host stopped heartbeating: it is dead, reclaim. The
+	// ETag-conditional delete closes the read/delete race: a heartbeat
+	// landing in between fails the delete, and the arrival defers instead
+	// of duplicating the revived execution.
+	if g.delete(ctx, actorID, etag) != nil {
+		// Reclaim lost the race; classify as live rather than error.
+		return Defer, nil //nolint:nilerr
+	}
+	g.forget(actorID)
+	return Proceed, nil
+}
+
+// observeStale reports whether rec's heartbeat has been observed unchanged
+// for StaleAfter on this reader's clock. HeartbeatMs is another host's wall
+// clock, so it is never compared against local time (skew of the grace or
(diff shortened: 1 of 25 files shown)


pkg/actors/targets/workflow/activity/claim/guard.go
```diff
@@ -0,0 +1,119 @@
...
+				g.deleteOwned(ctx, actorID, taskKey)
+				return
+			}
+			g.write(ctx, actorID, taskKey, true)
+			select {
+			case <-ctx.Done():
+			case <-time.After(g.opts.Retention):
+				g.deleteOwned(ctx, actorID, taskKey)
+			}
+			return
```
javier-aliaga, line 101: It seems this part can left completed records not purged

```
The retention wait's `<-ctx.Done()` arm returns without deleting, and being nested inside
`<-call.Done()` it can't reach the shutdown delete above. Doesn't a host restart inside the
retention window leave the `Completed` row behind for good? I couldn't find anything that
reclaims it later — purge looks scoped to the workflow actor.
```

pkg/actors/targets/workflow/activity/claim/check.go
```diff
@@ -0,0 +1,105 @@
...
+			// classification.
+			return Defer, nil //nolint:nilerr
+		}
+		if stale {
+			g.forget(actorID)
+		}
+		return Proceed, nil
+	}
+	if rec.Completed {
+		g.forget(actorID)
```
cicoyle: Shall we check for staleness here too and delete if true?
```suggestion
		if stale {
				_ = g.delete(ctx, actorID, etag)
		}
		g.forget(actorID)
```

pkg/actors/targets/workflow/activity/claim/guard.go
```diff
@@ -0,0 +1,119 @@
...
+			g.deleteOwned(context.WithoutCancel(ctx), actorID, taskKey)
+			return
+		case <-call.Done():
+			if call.Err() != nil {
+				g.deleteOwned(ctx, actorID, taskKey)
+				return
+			}
+			g.write(ctx, actorID, taskKey, true)
+			select {
+			case <-ctx.Done():
```
cicoyle, line 94: I think we should also call `deleteOwned` here too.
```suggestion
			case <-ctx.Done():
					g.deleteOwned(context.WithoutCancel(ctx), actorID, taskKey)
```

cicoyle (changes requested):

pkg/actors/targets/workflow/activity/claim/guard.go
JoshVanL: No bc the Completed row is what lets recovery arrivals on the new owner ack instead of re-executing an already-published body

pkg/actors/targets/workflow/activity/claim_test.go
```diff
@@ -714,6 +714,36 @@ func Test_gateJanitorRedispatch(t *testing.T) {
...
+		// First read opens the reader-side observation window and acks.
+		handled, err := a.gateJanitorRedispatch(t.Context(), testInvocation())
+		assert.True(t, handled)
+		require.NoError(t, err)
+		_, ok := store.get(t, actorID)
+		assert.True(t, ok, "a fresh completed row is retained for in-flight recovery arrivals")
+
+		// A later read past the grace still acks and reaps the row.
+		time.Sleep(5 * time.Millisecond)
+		handled, err = a.gateJanitorRedispatch(t.Context(), testInvocation())
```
javier-aliaga, line 741: Why do we need two `gateJanitorRedispatch` calls here? What happens if there's no second read
— e.g. the task already resolved, so the janitor stops re-dispatching?

JoshVanL: We need two reads because it's based on observed time diff, so we need a start observation then a second to check is stale

cicoyle (approved):
mostly lgtm but I think this ci failure is related - can you plz verify:
```
pkg/actors/targets/workflow/activity Test_checkClaimRecord/stale_record_from_another_scheduling_is_reaped_on_read (0.14s)
    claim_test.go:425: 
        	Error Trace:	D:/a/dapr/dapr/pkg/actors/targets/workflow/activity/claim_test.go:425
        	Error:      	Should be false
        	Test:       	Test_checkClaimRecord/stale_record_from_another_scheduling_is_reaped_on_read
        	Messages:   	a dead old-run record must not leak
```

javier-aliaga (approved):

PR #10410, opened 2026-08-26
fix: handle NewRequest error in pathMatching.match to prevent nil deref panic

### LINKED ISSUE #10409: Service invocation decodes percent-encoded reserved characters in the path, and daprd panics on some of them
**Dapr version:** 1.18.1 (`ghcr.io/dapr/operator:1.18.1`, sidecar `ver=1.18.1`)
**Deployment:** Kubernetes (k3s v1.36.2), sidecar-injected, HTTP service invocation
**Impact:** a callee cannot be addressed with a path segment containing a percent-encoded reserved
character. The encoding is lost, so the request reaches a *different* resource than the caller named;
for two characters the sidecar panics instead.

## What happens

Caller sends, from inside the app container, to its own sidecar:

```
POST http://127.0.0.1:3500/v1.0/invoke/catalog/method/v1/table/<segment>/describe
```

| `<segment>` sent | result at the caller | what the callee actually received |
| --- | --- | --- |
| `x%2Fy` | 404 | `POST /v1/table/x/y/describe` — **decoded; one segment became two** |
| `x%3Fy` | 500 | — |
| `x%23y` | 500 | — |
| `acme%1Fbronze` | connection closed (`RemoteDisconnected`) | — |

The first row is the core defect and is the quiet one: `%2F` is a literal character *inside* one path
segment, never a separator. Promoting it silently addresses a different object, and the caller gets a
plausible 404 rather than an error.

The callee's own access log is the proof — it logs `/v1/table/x/y/describe` for a request that named
the single segment `x%2Fy`.

## The panic

`%1F` (and `%3F`/`%23` on the 500 path) reaches a nil dereference in the sidecar's HTTP **metrics**
path-matching, which then kills the connection:

```
http: panic serving 127.0.0.1:35052: runtime error: invalid memory address or nil pointer dereference
goroutine 33455 [running]:
net/http.(*conn).serve.func1()
	/opt/hostedtoolcache/go/1.26.4/x64/src/net/http/server.go:1907 +0xbd
panic({0x6a4d280?, 0xda8e0e0?})
	/opt/hostedtoolcache/go/1.26.4/x64/src/runtime/panic.go:860 +0x13a
github.com/dapr/dapr/pkg/diagnostics.(*pathMatching).match(0xf4d845cb940, {0xf4d84dadd00?, 0xf4d84dadd00?})
	/home/runner/work/dapr/dapr/pkg/diagnostics/http_monitoring_path_matching.go:96 +0x85
github.com/dapr/dapr/pkg/diagnostics.(*httpMetrics).getMetricsPath(0xf4d84f163f0, {0xf4d84dadd00, 0x39})
	/home/runner/work/dapr/dapr/pkg/diagnostics/http_monitoring.go:127 +0x65
github.com/dapr/dapr/pkg/diagnostics.(*httpMetrics).ServerRequestCompleted(...)
	/home/runner/work/dapr/dapr/pkg/diagnostics/http_monitoring.go:164 +0xa5
github.com/dapr/dapr/pkg/diagnostics.(*httpMetrics).HTTPMiddleware-fm...
	/home/runner/work/dapr/dapr/pkg/diagnostics/http_monitoring.go:375 +0x2ba
```

Two things worth separating here: the metrics middleware panicking on a control character is a bug on
its own (an observability path should never be able to fail a request), and it is downstream of the
decoding, which is the behaviour that makes the address wrong in the first place.

The sidecar also logs the decoded form, which shows the loss before the callee is ever reached:

```
"Error processing operation endpoint[catalog, catalog:v1/table/acmebronze/describe]. Retrying in 2s…"
```

Note `acmebronze` — the `%1F` is already a raw control character in daprd's own log line.

## Expected

The path should reach the callee byte-identical to the one the caller sent. RFC 3986 §2.2 is explicit
that a percent-encoded reserved character is *not* equivalent to its decoded form — `a%2Fb` is one
segment named `a/b`, and `a/b` is two segments. A proxy that normalises between them changes which
resource is addressed.

## Why this matters beyond one API

This is not an exotic input. The [Lance Namespace
spec](https://lance.org/format/namespace/) joins a multipart namespace identifier with the **unit
separator, `0x1F`** — so `["acme","bronze"]` travels as `acme%1Fbronze`, and the root namespace is
`%1F` alone. Any catalog implementing that spec behind Dapr service invocation cannot address a nested
namespace at all. More generally, any object store or catalog whose keys may contain `/` is affected
by the `%2F` row.

## Reproduction

Minimal, with any HTTP callee that echoes its request path:

```bash
# from inside an app container with an injected sidecar
curl -s -o /dev/null -w '%{http_code}\n' -X POST \
  http://127.0.0.1:3500/v1.0/invoke/<callee-app-id>/method/v1/table/x%2Fy/describe
# callee logs: POST /v1/table/x/y/describe     <- decoded

curl -s -o /dev/null -w '%{http_code}\n' -X POST \
  http://127.0.0.1:3500/v1.0/invoke/<callee-app-id>/method/v1/table/acme%1Fbronze/describe
# sidecar: panic in pkg/diagnostics/http_monitoring_path_matching.go:96
```

## Notes

Confirmed that the caller is not the source of the decoding: the calling proxy was fixed to build its
upstream URL from the raw ASGI path and verified to emit `.../v1/table/x%2Fy/describe` on the wire.
The loss occurs at the service-invocation hop.

waterWang:
### Problem

When an HTTP service invocation path contains a percent-encoded reserved or control character (e.g. `%1F`), the decoded path reaches `pathMatching.match()` with an unescaped control byte. Calling `http.NewRequest(method, path, nil)` on such a path returns an **error and a nil request** (`net/url: invalid control character in URL`), so the subsequent `req.URL != nil` check dereferences a nil pointer and panics the sidecar's HTTP server, killing the connection:

```
http: panic serving 127.0.0.1:35052: runtime error: invalid memory address or nil pointer dereference
github.com/dapr/dapr/pkg/diagnostics.(*pathMatching).match(...)
	http_monitoring_path_matching.go:96
```

### Fix

Capture the error returned by `http.NewRequest` and bail out early (return no match) when the path cannot be parsed, instead of dereferencing the nil request. Paths that parse cleanly are unaffected.

Fixes #10409

### DIFF
--- a/pkg/diagnostics/http_monitoring_path_matching.go
+++ b/pkg/diagnostics/http_monitoring_path_matching.go
@@ -92,7 +92,10 @@ func (pm *pathMatching) match(path string) (string, bool) {
 	}
 
 	cleanPath := NormalizeHTTPPath(path)
-	req, _ := http.NewRequest(http.MethodGet, cleanPath, nil)
+	req, err := http.NewRequest(http.MethodGet, cleanPath, nil)
+	if err != nil {
+		return "", false
+	}
 	if req.URL != nil {
 		req.URL.Path = cleanPath
 	}


pkg/diagnostics/http_monitoring_path_matching.go
```diff
@@ -92,7 +92,10 @@ func (pm *pathMatching) match(path string) (string, bool) {
 	}
 
 	cleanPath := NormalizeHTTPPath(path)
-	req, _ := http.NewRequest(http.MethodGet, cleanPath, nil)
+	req, err := http.NewRequest(http.MethodGet, cleanPath, nil)
+	if err != nil {
+		return "", false
+	}
```
JoshVanL, line 98: Please include an integration test covering this change/issue https://github.com/dapr/dapr/tree/master/tests/integration/suite/daprd/serviceinvocation

JoshVanL (changes requested):

PR #10411, opened 2026-08-26
fix (resiliency): eliminate data race between circuit-breaker cache reads and writes

adashexplorer:
### Description

Root cause
---------------------

EndpointPolicy and ActorPreLockPolicy in pkg/resiliency/resiliency.go each have two branches:

- A named-policy branch, used when the app/actor type has an explicit target entry in the resiliency config. This branch read r.serviceCBs[app] (line 637) / r.actorCBCaches[actorType] (line 771) directly, with no lock held.
- A default-policy branch, used when no explicit target entry exists. This branch calls getServiceCBCache / getActorCBCache, which read and, on first use, write the same maps, correctly guarded by r.serviceCBsMu / r.actorCBsCachesMu.

Because both branches operate on the same underlying map[string]*lru.Cache[...], a goroutine taking the named-policy path (unsynchronized read) can run concurrently with a goroutine taking the default-policy path (locked write, e.g. when a new app/actor type is seen for the first time). Go's built-in maps are not safe for concurrent read/write even on different keys, so this is a real data race, not just a benign one.

How to reproduce
---------------------

Run two kinds of concurrent traffic against the same Resiliency instance:
1. Repeated calls to EndpointPolicy/ActorPreLockPolicy for an app/actor type that has an explicit target entry with a circuit breaker configured (hits the unlocked read at line 637/771).
2. Repeated calls to EndpointPolicy/ActorPreLockPolicy for a new/unconfigured app/actor type on every iteration, so a default circuit breaker is resolved and getServiceCBCache/getActorCBCache writes a new map entry each time (hits the locked write path).

Under go test -race, this reliably surfaces:
fatal error: concurrent map read and map write
Under production load, the same condition crashes the daprd process outright — it's a fatal runtime error, not a recoverable panic.

What is the fix
---------------------

Wrap the two unsynchronized reads with the same RWMutex already used to guard writes to that map:
- r.serviceCBs[app] read in EndpointPolicy → guarded by r.serviceCBsMu.RLock()/RUnlock().
- r.actorCBCaches[actorType] read in ActorPreLockPolicy → guarded by r.actorCBsCachesMu.RLock()/RUnlock().

No other changes were needed: the *lru.Cache values themselves (hashicorp/golang-lru/v2) are already internally synchronized, so cache.Get/cache.Add calls after the lookup don't need additional locking — only the map lookup itself was unsafe.

```
// EndpointPolicy — before
cache, ok := r.serviceCBs[app]

// EndpointPolicy — after
r.serviceCBsMu.RLock()
cache, ok := r.serviceCBs[app]
r.serviceCBsMu.RUnlock()
// ActorPreLockPolicy — before
cache, ok := r.actorCBCaches[actorType]

// ActorPreLockPolicy — after
r.actorCBsCachesMu.RLock()
cache, ok := r.actorCBCaches[actorType]
r.actorCBsCachesMu.RUnlock()
```

Impact
---------------------

- Severity: Critical — this is a fatal, unrecoverable Go runtime error (concurrent map read and map write), not an error/panic that can be caught. It terminates the entire daprd process immediately.
- Trigger conditions: Any deployment with a resiliency policy that (a) defines an explicit circuit breaker for at least one app or actor type, and (b) also serves traffic to apps/actor types that only match a default circuit breaker policy — i.e. any reasonably active service-invocation or actor workload using resiliency circuit breakers, since the two branches run naturally in separate goroutines per request.
- Blast radius: Process-wide crash, so it takes down all in-flight requests handled by that sidecar, not just the request that triggered it.

Tests covered
---------------------

Added TestEndpointAndActorCBCacheConcurrentAccess in pkg/resiliency/resiliency_test.go:
- Spins up concurrent goroutines exercising both the named-policy branch (fixed app/actor type with an explicit target entry) and the default-policy branch (a fresh app/actor type name generated on every iteration, forcing repeated map writes via on-demand cache creation).
- Run under go test -race:
  - Verified against pre-fix code: reproduces the exact fatal error: concurrent map read and map write crash at resiliency.go:771.
  - Verified against fixed code: passes cleanly.
- Full package regression: go test -race ./pkg/resiliency/... passes (both pkg/resiliency and pkg/resiliency/breaker).

Files changed
---------------------

- pkg/resiliency/resiliency.go — locking fix (+4 lines)
- pkg/resiliency/resiliency_test.go — new race regression test (+89 lines)

---


adashexplorer:
This PR fixes issue https://github.com/dapr/dapr/issues/10413.

adashexplorer:
@JoshVanL , Can you please review / suggest on the changes ? or redirect to someone who can review.

